PR #651, opened 2020-02-06
Implementing the secondary click for a button

palikar:
A very simple change to `widget/button.go` which enables the use of the secondary mouse button. I needed this functionality for a project and I would like to use the official fyne repository as dependency and not my own fork. There are no braking changes and one can do something like

``` go
b := widget.NewButton("Button", func() {})
b.OnSecondaryTapped = func(){ }
```
in order to attach a callback for the right mouse button.

#### Tests 
- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/button.go
+++ b/widget/button.go
@@ -148,9 +148,10 @@ type Button struct {
 	Icon         fyne.Resource
 	disabledIcon fyne.Resource
 
-	OnTapped   func() `json:"-"`
-	hovered    bool
-	HideShadow bool
+	OnTapped          func() `json:"-"`
+	OnSecondaryTapped func() `json:"-"`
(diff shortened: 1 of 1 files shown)


andydotxyz:
The standard widgets provide only standard use-cases and we don't think that right clicking or tap-old buttons is a standard interaction. Thanks for your contribution but I don't think we will merge it. 

There is no need to fork the project to get this functionality - a custom button in your code that extends the standard button can implement TappedSecondary(). See the example illustrated at https://fyne.io/develop/extending-widgets.html, with the exception that you would not add Tapped() as that is handled by widget.Button already.

PR #666, opened 2020-02-09
Adds a command to print the version of the Fyne library that would be…

### LINKED ISSUE #656: fyne should have a "version" command
The bug report template asks for the `fyne` version, but there's no way to do it except to go to the git checkout and query git.

If there's a way to get the version, it'd be great if that could be documented in the bug template, because it's not obvious.

```
$ fyne version          
Unsupported command version                                                                     
$ fyne --version                                                                                
Usage: fyne [command] [parameters], where command is one of:                       
  install, bundle, get, package or help                                                         
... (trimmed help text)
$ fyne -v
Usage: fyne [command] [parameters], where command is one of:
  bundle, get, package, install or help
... (trimmed help text)
$ fyne -version
Usage: fyne [command] [parameters], where command is one of:
  package, install, bundle, get or help
... (trimmed help text)
```

xxxserxxx:
… used to compile packages, if go modules are used.

### Description:
Adds a `lib-version` command that uses the `go/build` library to introspect on the `fyne.io/fyne` package being used in the current environment, and prints out the version number (if any).  As version information is only available for Go modules, this will only produce useful information if modules are disabled, or if Fyne is being somehow otherwise referenced.  E.g., it doesn't work well within the Fyne project itself.

If modules are not being used, it merely reports that it can't get version information when none exists.

E.g.:
```
$ cd ~/workspace/fyne
$ fyne lib-version
Malformed package info /home/ser/workspace/fyne/pkg/linux_amd64/fyne.io/fyne.a; not using modules?
$ cd ~/workspace/healthdiary
$ fyne lib-version
v1.2.2
```

Fixes #656

### Checklist:

- [ ] Tests included.  **Nope.  I didn't see any test patterns for other commands, so I didn't provide one for this.**
- [X] Tests all pass. **Yes, but see above.**

#### Where applicable:

- [X] Public APIs match existing style. **I guess?**
- [X] Any breaking changes have a deprecation path or have been discussed. **N/A**

### DIFF
--- a/cmd/fyne/main.go
+++ b/cmd/fyne/main.go
@@ -52,6 +52,7 @@ func loadCommands() {
 	commands["get"] = &getter{}
 	commands["package"] = &packager{}
 	commands["install"] = &installer{}
+	commands["lib-version"] = &versioner{}
 }
 
 func main() {
--- a/cmd/fyne/version.go
+++ b/cmd/fyne/version.go
@@ -1,0 +1,63 @@
+package main
+
+import (
+	"fmt"
+	"go/build"
+	"os"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+const PKG = "fyne.io/fyne"
+
+// Declare conformity to command interface
+var _ command = (*versioner)(nil)
+
+type versioner struct {
+	context build.Context
+}
+
+func (v *versioner) run(_ []string) {
+	err := v.validate()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err.Error())
+		os.Exit(1)
+	}
+
+	err = v.doVersion()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%s\n", err.Error())
+		os.Exit(1)
+	}
+}
+
+func (v *versioner) validate() error {
+	return nil
+}
+
+func (v *versioner) doVersion() error {
+	c := build.Default
+	p, e := c.Import(PKG, ".", build.FindOnly)
+	if e != nil {
+		return e
+	}
+	// p.PkgObj =~ /home/ser/go/pkg/mod/fyne.io/fyne@v1.2.2/pkg/li
+	ps := strings.Split(p.PkgObj, "@")
+	if len(ps) < 2 {
+		return errors.New(fmt.Sprintf("Malformed package info %s; not using modules?", p.PkgObj))
+	}
+	vs := strings.Split(ps[1], "/")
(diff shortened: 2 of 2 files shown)


andydotxyz:
Thanks so much for this, a good idea.
I have a couple of comments on the code that I hope seem reasonable.

I also wonder if a clever fallback would be to try and get the current git tag (or SHA) if it's not a project using modules.

xxxserxxx:
Oh, dammit.  I hate git.  I changed that PKG declaration before I made the pull request.  I'm going to wipe my fork and try again.  I'll use the suggestions in the next pull.

xxxserxxx:
I'm going to resubmit my pull request from a clean fork; I have no idea why git decided to use the wrong version for this one.

I'm going to add a unit test.  It'd be *great* if you guys could *also* add unit tests for commands, so that contributors have patterns to work with, and it'd set a good example.  "Do as I say, not as I do" tends to turn off contributors.

Moving the string comment to a test case is a good idea.

I'm not competent, and don't have the time to become competent, in figuring out how to get the git hash.  I can think of three solutions, none of which I particularly like.  They all depend on more introspection using the package library to find out where fyne lives, and then:

1. navigate to the `.git` directory, and dump out the contents of `FETCH_HEAD`.  I'm not even sure if that's correct; I really dislike git and have no interest in learning about its internals.
2. `exec` out to git to dump out the checked-out revision.  Assuming the user has git installed; this seems particularly fragile and hacky.
3. Use a git library that removes a dependency on an executable. Maybe there's one in the standard libs, or maybe I'd have to introduce a new dependency; the latter would not be great.

Is there another way to get at this information?

andydotxyz:
You're absolutely right about the unit tests - we should have them in commands as well.
We can look at the .git fallback later on. I think it's reasonable to assume "git" is installed as compiling with Go requires it :). Your suggestion about .git/FETCH_HEAD is probably the cleanest though.

xxxserxxx:
> I think it's reasonable to assume "git" is installed as compiling with Go requires it 

Well... only if you're compiling code that has dependencies being pulled from git. Same can be said for Mercurial, Bazaar, and Subversion.  But, yeah.  Since fyne is a git project, and is unlikely to change, we could do that thing.  It would cause problems from anyone trying to fork into a different VCS.

I think this also may break with modules, which don't check out repositories but fetch bundles.  `go get` still checks stuff out, but this fails as soon as a distribution releases a binary fyne executable package, and the fyne library (dependency) is pulled purely with modules.

PR #668, opened 2020-02-10
lib-version implementation

### LINKED ISSUE #656: fyne should have a "version" command

xxxserxxx:
### Description:
Adds a `lib-version` command to print out the version of fyne that would be used to compile the project.

Fixes #656 

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

#### Where applicable:

- [ ] Public APIs match existing style. **N/A**
- [ ] Any breaking changes have a deprecation path or have been discussed. **N/A**

### DIFF
--- a/cmd/fyne/version.go
+++ b/cmd/fyne/version.go
@@ -1,0 +1,63 @@
+package main
+
+import (
+	"fmt"
+	"go/build"
+	"os"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+const pkg = "fyne.io/fyne"
+
+// Declare conformity to command interface
+var _ command = (*versioner)(nil)
+
+type versioner struct {
+	context build.Context
+}
+
+func (v *versioner) run(_ []string) {
+	err := v.validate()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err.Error())
+		os.Exit(1)
+	}
+
+	err = v.doVersion()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%s\n", err.Error())
+		os.Exit(1)
+	}
+}
+
+func (v *versioner) validate() error {
+	return nil
+}
+
+func (v *versioner) doVersion() error {
+	c := build.Default
+	p, e := c.Import(pkg, ".", build.FindOnly)
+	if e != nil {
+		return e
+	}
+	// p.PkgObj =~ /home/ser/go/pkg/mod/fyne.io/fyne@v1.2.2/pkg/li
+	ps := strings.Split(p.PkgObj, "@")
+	if len(ps) < 2 {
+		return errors.New(fmt.Sprintf("Malformed package info %s; not using modules?", p.PkgObj))
+	}
+	vs := strings.Split(ps[1], "/")
+	if len(vs) < 2 {
+		return errors.New(fmt.Sprintf("Missing version information %s; not using modules?", p.PkgObj))
+	}
+	fmt.Println(vs[0])
+	return nil
+}
+
+func (v *versioner) printHelp(indent string) {
+	fmt.Println(indent, "Print the version of the Fyne library that would be used to build the package.")
+}
+
+func (v *versioner) addFlags() {
+}
--- a/cmd/fyne/version.go
+++ b/cmd/fyne/version.go
@@ -42,17 +42,28 @@ func (v *versioner) doVersion() error {
 	if e != nil {
 		return e
 	}
-	// p.PkgObj =~ /home/ser/go/pkg/mod/fyne.io/fyne@v1.2.2/pkg/li
-	ps := strings.Split(p.PkgObj, "@")
-	if len(ps) < 2 {
-		return errors.New(fmt.Sprintf("Malformed package info %s; not using modules?", p.PkgObj))
+	ve, e := parsePkgString(p.PkgObj)
+	if e != nil {
+		return e
 	}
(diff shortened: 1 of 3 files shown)


cmd/fyne/version_test.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParsePkgString(t *testing.T) {
+	type test struct {
+		in string // Test input
+		es string // Expected output
```
andydotxyz: Can you use meaningful variable names? If it was "expectedOutput" then you wouldn't need the comment...

cmd/fyne/version_test.go
```diff
@@ -0,0 +1,32 @@
...
+		{in: "/home/user/go/pkg/mod/fyne.io/fyne@v1.2.2/pkg/li", es: "v1.2.2", ee: false},
+		{in: "fyne.io/fyne@v1.1", es: "v1.1", ee: false},
+		{in: "/home/user/fyne/pkg/linux_amd64/fyne.io/fyne.a", es: "", ee: true},
+		{in: "v1.2.2", es: "", ee: true},
+		{in: "/home/user/go/pkg/mod/github.com/peterbourgon/diskv@v2.0.1+incompatible/pkg/linux_amd64/github.com/peterbourgon/diskv.a", es: "v2.0.1+incompatible", ee: false},
+		{in: "/home/ser/go/pkg/mod/github.com/peterbourgon/diskv/v3@v3.0.0/pkg/linux_amd64/github.com/peterbourgon/diskv/v3.a", es: "v3.0.0", ee: false},
+	}
+
+	for _, tc := range ts {
+		gs, ge := parsePkgString(tc.in)
```
andydotxyz: It's not clear from the context what gs and ge are - could they be named more clearly?

cmd/fyne/version_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParsePkgString(t *testing.T) {
+	type test struct {
+		in string // Test input
+		es string // Expected output
+		ee bool   // Expecting error?
```
andydotxyz: expectedError would imply an error type, perhaps shouldError is closer to the truth for a bool?

cmd/fyne/version.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+}
+
+func (v *versioner) validate() error {
+	return nil
+}
+
+func (v *versioner) doVersion() error {
+	c := build.Default
+	p, e := c.Import(pkg, ".", build.FindOnly)
```
andydotxyz: c, p, e and ve in this method are confusing - what are they short for?
Go normally uses "err" for an error var so that's a start. Could the other variables be more clearly named?

cmd/fyne/version.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	err = v.doVersion()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%s\n", err.Error())
+		os.Exit(1)
+	}
+}
+
+func (v *versioner) validate() error {
```
andydotxyz: Why do we need a method that does nothing? The call to it above also seems un-needed.

cmd/fyne/version_test.go
xxxserxxx: So, the Fyne codebase doesn't follow Go stdlib conventions?

- https://github.com/golang/go/wiki/CodeReviewComments#variable-names
- https://talks.golang.org/2014/names.slide#6
- and, of course, [Rob Pike's comments](http://doc.cat-v.org/bell_labs/pikestyle) on the matter.

cmd/fyne/version.go
xxxserxxx: That's my bad.  It was in the other commands, so I thought it was a required method on the `command` interface; I should have checked.

cmd/fyne/version_test.go
andydotxyz: I would not go so far as to say Fyne does not follow Go conventions, but I would say that the project follows the belief that if you have to have inline comments then your code could be clearer. Of course "i" is an iterator and "b" could be a buffer - where the context is clear and a convention is followed. I do not think that "es" is common shorthand for "expected output" - looking at slide#8 It would seem to indicate that "expected" could be completely inline with their recommendation.

cmd/fyne/version.go
andydotxyz: Perhaps in this instance build.Default could be set to "b" but I don't follow "c" as a choice. If "e" were "err" I still don't know what "p" and "ve" indicate?

andydotxyz:
Private feedback from the developer is that they don't want to make the requested changes. Given that code style is very important to this project the PR will be closed without merge.

PR #676, opened 2020-02-14
bugfix: blankscreen on some android phones

cerisara:
### Description:

On some android phones, fyne apps sometimes are just rendered as a blank screen. This occurs when the current window size is set to 0x0. So when this happens, the fix consists in not doing an actual rendering of the app. Also, when a new size.Event is observed, a paint event must be sent to redraw the app.

There is another (related) fix in this commit: when the app is pushed in the background and called back in the foreground later on, then the objects in the canvas were not redrawn (the refreshQueue list of objects to redraw was empty); so I push back all objects in the queue to be refreshed whenever a lifecycle event makes the app visible again.


andydotxyz:
This change looks great but the formatting is failing the Travis builds.
Please correct the formatting (gofmt or goimports will do it) then I'll have another look.

cerisara:
Thank you, I prefer to close this PR and open another one to the develop branch, I guess it's what is expected anyway.

PR #686, opened 2020-02-18
provide default window title

spatocode:
### Description:
Provide default window title. This replaces the default title GLFW-Application from glfw driver.

Fixes #(issue)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/internal/driver/glfw/window_test.go
+++ b/internal/driver/glfw/window_test.go
@@ -39,7 +39,9 @@ func TestMain(m *testing.M) {
 
 func TestGLDriver_CreateWindow(t *testing.T) {
 	w := NewGLDriver().CreateWindow("Test").(*window)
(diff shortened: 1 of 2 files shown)


internal/driver/glfw/window_test.go
```diff
@@ -39,7 +39,9 @@ func TestMain(m *testing.M) {
 
 func TestGLDriver_CreateWindow(t *testing.T) {
 	w := NewGLDriver().CreateWindow("Test").(*window)
+	wt := NewGLDriver().CreateWindow("").(*window)
```
andydotxyz: This line effectively sets up a new test case unrelated to the test in here already. Could you move it to a new test, perhaps named TestGLFriver_CreateWindow_EmptyTitle? Then the window bar could return to the common “w”

PR #690, opened 2020-02-20
Update GLFW to 3.3.2

rubensayshi:
### Description:

GLFW v3.3 has some new flags such as `FocusOnShow` so an update could be nice (https://www.glfw.org/docs/latest/news.html for all stuff).  

However in v3.3.1 there's a MacOS bug that's really painful; https://github.com/go-gl/glfw/issues/262  
So atm I'm using the `update-to-3.3.2` branch that hasn't been merged yet; https://github.com/go-gl/glfw/pull/268

Tbh the MacOS bug in v3.3.1 is pretty painful and somewhat scary it slipped through, maybe some extensive testing with v3.3.2 would be good before actually merging this ... unfortunately my usage of fyne & GLFW is fairly limited so I don't have much to test it with ... 

So for anyone who wants to use v3.3.2 or wants to test using v3.3.2, they can use this branch.  
I'll update the branch as soon as the upstream PR is merged, until then it obviously shouldn't be merged ;)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


rubensayshi:
closed, so I can open it as draft PR

andydotxyz:
As this upgrade is pretty big we are testing it already on the develop branch. If you can PR on that you should find much smaller differences :)

rubensayshi:
oh sweet, thx @andydotxyz , I'll test my stuff using develop and keep an eye on the progress there!

rubensayshi:
not sure why, but your `go.mod` has the commit of the `update-3.2.2` branch, but in your `vendor/` there's still 3.3.1 code ... 

however I can't seem to fix it by running `go mod vendor`, for w/e reason it just deletes code when I do that :/

andydotxyz:
Vendor doesn’t like glfw unfortunately - we have a tool internally to do it - ‘go run tools/modvendor/main.go’

Issue #520, opened 2019-11-08
Blurry content on VirtualBox


kroksys:
Windows on VirtualBox have problems detecting correct DPI so the content becomes blurry (see the picture below - open it in full screen).
When I run the same app from macOS (host) the app is not blurry anymore.

![Screen Shot 2019-11-08 at 12 56 48](https://user-images.githubusercontent.com/15876796/68472153-8914a380-0228-11ea-996a-0ff25d997f10.png)

 - OS: Windows 10 on VirtualBox hosted on Mac
 - Windows 10, macOS Catalina v10.15.1
 - Go 1.13.1
 - Fyne version 1.1.2

I played with changing  Fonts and Scale, but that didn't solve the problem.

andydotxyz:
If you click on the advanced tab what scale does it say is being used?

kroksys:
Scale 1.00

andydotxyz:
can you report if that value changes, or the display is more clear if you set the environment FYNE_SCALE to "auto"?

kroksys:
After setting up FYNE_SCALE to "auto" or "0" Result Scale was 0.7
It was still blurry (looked more blurry than before).
I believe this happens because of the VirtualBox virtual screen.

andydotxyz:
Yes, that makes sense. I cannot figure out why the Fyne content was blurry but other content was not.
Does windows internally have some scaling settings?
Can you confirm the virtual box window size vs the internal windows desktop size please?

andydotxyz:
This should be resolved now as we are using window's builtin scaling value..?
Please re-open if it is still an issue.

Issue #635, opened 2020-01-27
Scale does not match user configuration in Windows 10


andydotxyz:
### Describe the bug:
There is a user configuration for a percentage scale in Windows 10.
Fyne apps do not respect this

### To Reproduce:
Steps to reproduce the behaviour:
1. Load Fyne app on Windows 10
2. Right click desktop -> desktop settings
3. Scroll to scale and set it to, for example, 200%
4. See that fyne app does not update
5. Launch another instance of the app and see it is the same size as the first one.

### Device (please complete the following information):
 - OS: Windows
 - Version: 10
 - Go version: 1.12.5
 - Fyne version: 1.2.1

perqin:
There is [an API](https://github.com/glfw/glfw/commit/16bf872117896fb88493403157fcd367303fd1f6#diff-7a6c70f06dc23e69c5a9bfe9c3967e64R1957) for retrieving system scaling, does it help?

andydotxyz:
Absolutely!
We just need to get GLFW 3.3.2 into the go bindings, which recently update to 3.3.
As soon as it's available we could use that API to resolve this ticket - good catch.

DamnWidget:
Also happening to me on Linux with version v1.2.2 it was fine on v1.1.1 Scaling feels broken in general on v1.2.2

![image](https://user-images.githubusercontent.com/329766/74575200-f3528800-4f7d-11ea-966f-b06e27a4c385.png)
![image](https://user-images.githubusercontent.com/329766/74575248-1bda8200-4f7e-11ea-84ca-57d7154df7ff.png)

Is there any way to enforce an scale in the meanwhile?

andydotxyz:
The behaviour on Linux is changed in 1.2.2 @DamnWidget. For all platforms we added a “system scale”, for macOS this is “1” due to the way it handles scaling. On Windows we seem to have the bug that is the topic of this issue.

On Linux (as there is no standard for scaling) it calculates the DPI of a screen and updates system scale so the base is 96DPI. A user scale of 1.2 will larger than this or 0.9 will be smaller. The readout in fyne_demo is the result of system*user (we should update this to display the details). It looks like you have a fairly dense screen and so the system scale is calculated as 1.5.

I hope this helps. In an updated version of our settings UI I think we will update it to “tiny” “small” “normal” “big” “huge” so the maths is hidden completely.

This enables windows to maintain real size when dragged from one screen to another when they are very different sizes.

DamnWidget:
[...]

PR #698, opened 2020-02-23
Use auto-detected scaling on Windows

hineH:
### Description:

Use auto-detected scaling on Windows.
There are two possible situation with incorrect scaling according to Windows display settings:
1. Everything displayed very small with original scaling.
2. Windows force scale it that display in correct size but blurry.

The detectScale() method detected correct scaling so use it when user do not explicitly specify the scaling.
Tested ok on my Windows10, 3840x2160  200% scaling display.
Fixes #(issue)
#520 
### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/internal/driver/glfw/device.go
+++ b/internal/driver/glfw/device.go
@@ -25,7 +25,7 @@ func (*glDevice) HasKeyboard() bool {
 }
 
 func (*glDevice) SystemScale() float32 {
-	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
+	if  runtime.GOOS == "darwin" {
 		return 1.0
 	}
 


andydotxyz:
Because windows defines a scale factor in the operating system we left auto-detect off. Unfortunately the “return 1.0” clearly does not work like it does on macOS. I think the correct solution would be to load the OS configuration, fixing the bug at https://github.com/fyne-io/fyne/issues/635

hineH:
> Because windows defines a scale factor in the operating system we left auto-detect off. Unfortunately the “return 1.0” clearly does not work like it does on macOS. I think the correct solution would be to load the OS configuration, fixing the bug at #635

@andydotxyz  Yeah, the best solution is using system configuration.

Due to entirely diffrent mechanism of HiDPI implementation between macOS and Windows:
* For macOS, developers don't care about different resolutions, any resolitions are "HiDPI"-ed to a user settings value ( usually 1920x1080 for macbook), everything worked as same as on that physical resolution.
* For Windows , the system display everything in a particular dpi according to system scaling settings. If the application is designed for lower dpi(less pixels per inch) will display a very small size.To adapt diffrent resolutions application need to detect(or read from system) the "scaling rate" and scale display more pixels in every inch.

It could be expedient to use the detected scaling by default on Windows before you are finishing implement reading from system configure, otherwire it's unable to develop  resolution self-adaption program.

In my opinion the Enviroment variable "FYNE_SCALE" should override the "system scale" instead of (currently in use) "user scale".The value could be "auto"(auto detect),"system"(read from system),"{value}"(explicitly specific value)or ""(default can be any one of "auto","system","1.0"). "user scale" should be only used when user want to scale particular object.

andydotxyz:
I fortunately the additional complication is that multiple screens can have (vastly) different DPIs, so overriding to a specific value doesn’t make sense - on HiDPI the app would be half the size.

That is why the user setting (or FYNE_SCALE) override multiplies on top of system detected scale. So “1.5” means “make all of my apps 50% bigger than standard size”. And the OS defines what standard is. MacOS makes this easy and Linux has no support at all, so we calculate DPI.

On Windows it is somewhere between the two. The OS tells us what scale to use, and we should respect that instead of ignoring (current issue) or overriding (what your patch would do).

FYNE_SCALE allows developers to try out different user settings easily. We would consider adding a FYNE_SYSTEM_SCALE_IGNORE or something similar, but we really need to get the correct behaviour before we decide if people need to disable it.

andydotxyz:
Any further thoughts following up from the messages above?

hineH:
It's no problem that you already had designed scheme, this is just an expedient patch which is more proper than current behavior if you can't implement it as soon as possible.

andydotxyz:
I appreciate the thought, however we found that many monitors were not reporting size parameters on the Windows OS so we were getting crazy detected DPI values. Probably best to get the correct fix in :)

andydotxyz:
Replaced by #805 - It's working nicely and adjust window scale as you move from one screen to another.
Sorry to close this, but I'm glad we have a complete fix in for 1.2.4 (releasing soon).

PR #699, opened 2020-02-23
Refactor PopUp size and position handling

toaster:
### Description:

The `PopUp` size and position handling is a difficult thing.
I tried to improve the implementation and to reduce the visibility of internal structure.
This refactoring also fixes the bug which my previous bugfix of `Select`'s drop-down size adjustment fixed. That's why I reverted the previous fix.
The refactoring reduces unwanted dependencies and also fixes the `Select` drop-down size on mobile.

### Checklist:

- ~~[ ] Tests included.~~
- [x] Tests all pass.

### DIFF
--- a/widget/popup_test.go
+++ b/widget/popup_test.go
@@ -3,11 +3,11 @@ package widget
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
-
 	"fyne.io/fyne"
 	"fyne.io/fyne/test"
 	"fyne.io/fyne/theme"
+
+	"github.com/stretchr/testify/assert"
 )
 
 func TestNewPopUp(t *testing.T) {
--- a/widget/popup_test.go
+++ b/widget/popup_test.go
@@ -3,10 +3,12 @@ package widget
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
 	"fyne.io/fyne"
+	"fyne.io/fyne/internal/cache"
 	"fyne.io/fyne/test"
 	"fyne.io/fyne/theme"
-	"github.com/stretchr/testify/assert"
 )
 
 func TestNewPopUp(t *testing.T) {
@@ -48,6 +50,7 @@ func TestPopUp_Move(t *testing.T) {
 
 	pos := fyne.NewPos(10, 10)
 	pop.Move(pos)
+	cache.Renderer(pop).Layout(pop.Size())
 
 	innerPos := pop.Content.Position()
 	assert.Equal(t, pos.X+theme.Padding(), innerPos.X)
@@ -66,6 +69,7 @@ func TestPopUp_Move_Constrained(t *testing.T) {
 
 	pos := fyne.NewPos(30, 20)
 	pop.Move(pos)
+	cache.Renderer(pop).Layout(pop.Size())
 
 	innerPos := pop.Content.Position()
 	assert.Less(t, innerPos.X-theme.Padding(), pos.X,
@@ -86,6 +90,7 @@ func TestPopUp_Move_ConstrainedWindowToSmall(t *testing.T) {
 
 	pos := fyne.NewPos(20, 10)
 	pop.Move(pos)
+	cache.Renderer(pop).Layout(pop.Size())
 
 	innerPos := pop.Content.Position()
 	assert.Equal(t, theme.Padding(), innerPos.X, "content X position is adjusted but the window is too small")
--- a/widget/popup_test.go
+++ b/widget/popup_test.go
@@ -6,7 +6,6 @@ import (
 	"github.com/stretchr/testify/assert"
 
 	"fyne.io/fyne"
-	"fyne.io/fyne/internal/cache"
 	"fyne.io/fyne/test"
 	"fyne.io/fyne/theme"
 )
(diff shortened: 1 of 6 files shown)


widget/popup_test.go
```diff
@@ -48,6 +50,7 @@ func TestPopUp_Move(t *testing.T) {
 
 	pos := fyne.NewPos(10, 10)
 	pop.Move(pos)
+	cache.Renderer(pop).Layout(pop.Size())
```
andydotxyz: This concerns me. This seems to be required because the Move function no longer lays out content to accommodate hitting the edge of the screen.
And so any code that calls PopUp.Move() may not work any more.

widget/select_test.go
```diff
@@ -83,6 +83,7 @@ func TestSelect_Tapped_Constrained(t *testing.T) {
 	canvas := fyne.CurrentApp().Driver().CanvasForObject(combo)
 	canvas.(test.WindowlessCanvas).Resize(fyne.NewSize(100, 100))
 
+	combo.Resize(combo.MinSize())
```
andydotxyz: The addition of this line is problematic, it seems to reflect a change in behaviour.
In fyne_demo go to Graphics, then tap the icon name combo...
You will see that it's sizing up to full size instead of staying the correct width.

widget/popup_test.go
```diff
@@ -3,10 +3,12 @@ package widget
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
 	"fyne.io/fyne"
+	"fyne.io/fyne/internal/cache"
 	"fyne.io/fyne/test"
 	"fyne.io/fyne/theme"
-	"github.com/stretchr/testify/assert"
```
andydotxyz: I think we order with github after fyne. I'm surprised goimports did this...

widget/popup_test.go
toaster: The `Move` method of a widget should not lay out anything. That's the responsibility of the renderer. That's why the adjustments are now made there.

widget/popup_test.go
toaster: GoLand did this. It imports local packages after remote after system.

widget/popup_test.go
andydotxyz: In that case if the move could cause a need to layout then it should trigger it - that way the tests (and all other code) don't need this line added...

widget/popup_test.go
andydotxyz: Have you told it to use goimports sort order?

toaster: grafik

Honestly, I like the idea of separating external and internal imports.

widget/select_test.go
toaster: This line is not the cause of the drop.down size bug. I reported it as #706.
This line is here because the test set-up is uncommon. The combo is not set as canvas' content and therefore never laid out and therefore does have size 0x0.

widget/popup_test.go
toaster: There is no “all other code” and the tests are not normal code. I don't think that a `Move` should run `Layout` immediately. This is for several reasons:
1. I consider `Layout` as quite expensive
2. The `Move` of an element X is not important for the layout of X but for its parent p(X). Also `Move` is normally called from parent's renderer's `Layout` because the parent decides where the element has to be placed..
3. The `PopUp` is special because it mixes up two things: a container (an overlay with canvas size and 0,0 pos) and a content (a freely sizeable and positionable element). #707 describes a proposal to reduce or even eliminate the effects of this special behaviour.
4. Position adjustment due to outer constraints (window size) _is clearly renderer responsibility_. That's why the test that tests this adjustment has to ensure that the renderer is called. If #707 was addressed, the container and its content would be separated and the position won't be sent to container's `Move` but to a specific container function like `MoveContent` or alike. This would also make the test more clear because it would call `MoveContent` which then _might_ trigger a layout. But honestly I don't like the idea of the latter because it adds internal dependencies to the widget. This could be avoided by performing `Refresh` and call `Layout` from the renderer's `Refresh` if needed.

After writing all this I think I might add the `Refresh` mentioned in point 4 to `PopUp`. The `Layout` line then would vanish from the test again. Stay tuned :).

andydotxyz: Yes I think you're right. And this should get better now that #707 is landed.
Thanks

widget/select_test.go
andydotxyz: Oh yes, thanks. in that case is the canvas resize really helping? If the component is not inside the canvas then the parent size may have no impact?

widget/popup_test.go
andydotxyz: Me too :) but github is external and fyne is internal, so surely the previous order is correct (a space probably belonged in there between them though)

widget/select_test.go
toaster: I don't resize the canvas but the combo :).

widget/popup_test.go
toaster: Okay, I resorted it.

widget/select_test.go
andydotxyz: I was meaning the resize call above, which is a canvas resize (I think it was perhaps put in accidentally?)

andydotxyz:
much nicer now, thanks

PR #700, opened 2020-02-24
Only trigger DoubleTapped if the tap before occured on the same canvas object

swbartosz:
### Description:
Only trigger the DoubleTapped event if the double click happens on the same object. 

Currently, click anywhere on the window, then click on an object and it will trigger DoubleTapped on that object, while the object wasn't technically double tapped.

### Checklist:

- [ ] Tests included.
- [X] Tests all pass.


andydotxyz:
Good catch, thanks.
We should probably test this area more, but button timing is tricky :-\

andydotxyz:
On release/v1.2.3 for release this week

PR #708, opened 2020-02-24
Add Shift+Insert support for Paste shortcut

pymq:
One of my users noticed that Shift+Insert is a common shortcut for paste and should work - I think it's reasonable, as it works almost everywhere (QT, GTK, browsers what I tested), what do you say?

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -900,6 +900,17 @@ func (w *window) keyPressed(viewport *glfw.Window, key glfw.Key, scancode int, a
 			shortcut = &fyne.ShortcutSelectAll{}
 		}
 	}
+
+	if keyDesktopModifier == desktop.ShiftModifier {
+		switch keyName {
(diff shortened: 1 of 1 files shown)


swbartosz:
What about also adding:
CTRL + Insert for Copy
Shift + Delete for CUT

andydotxyz:
This looks like a good change. I agree with @swbartosz though, it would be good to add the group rather than just a single one?

pymq:
Didn't know about them :)
Thanks, added.

andydotxyz:
Did you pull in the develop branch somehow? it seems to have added those commits to the history here but it doesn't normally do that...
Can you pick your 2 commits on top of the latest develop instead?

andydotxyz:
It looks good from a code point of view though :)

pymq:
Yeah, I tried to rebase on develop but messed up.
Fixed now

andydotxyz:
Excellent thanks

Issue #704, opened 2020-02-24
Spinner Widget 


DerThorsten:
Hi I was playing around with your library and tried to implement a  Spinner widget since it was not yet available (and I want to implemented this as custom type which implements the widget interface, st I can define whatever method I want on  the spinner instance, but it should still be a fyne.widget )
This spinner is in particular nice to use in conjunction with the form widget:
![Untitled](https://user-images.githubusercontent.com/904752/75173389-15db6280-572f-11ea-93d6-ef02f5343fa4.png)

While the widget is looking fyne(ish), I am not yet happy with the code.
I ended up writing a custom widget (`Spinner`) which extends the fyne container.
The custom renderer (`spinnerRenderer`) then mostly calls functions from the container. 
Is there an easier way to have a custom widget struct which basically just composes existing widgets with a  (border) layout?

Greetings

```golang
package widget

import (
	"fmt"
	"image/color"
	"math"
	"strconv"

	"fyne.io/fyne"
	"fyne.io/fyne/layout"
	"fyne.io/fyne/theme"
	"fyne.io/fyne/widget"
)

func sgn(a float64) float64 {
	switch {
	case a < 0:
		return -1
	case a > 0:
		return +1
	}
	return 0
}

// extend entry to detect enter
type customEntry struct {
	widget.Entry
	onEnter    func()
	onScrolled func(s *fyne.ScrollEvent)
}

func newCustomEntry() *customEntry {
	cutomEntry := &customEntry{}
	cutomEntry.ExtendBaseWidget(cutomEntry)
	return cutomEntry
}

// KeyDown handler for keypress events - used to store shift modifier state for text selection
func (e *customEntry) KeyDown(key *fyne.KeyEvent) {

	if key.Name == fyne.KeyReturn {
		if e.onEnter != nil {
			e.onEnter()
		}
	} else {
		e.Entry.KeyDown(key)
	}
}

func (e *customEntry) Scrolled(s *fyne.ScrollEvent) {
	if e.onScrolled != nil {
		e.onScrolled(s)
	}
}

type Spinner struct {
	fyne.Container
	value    float64
	Min      float64
	Max      float64
	Step     float64
	startVal float64

	buttonUp   *widget.Button
	buttonDown *widget.Button
	entry      *customEntry
	integer    bool
}

func NewSpinner(min, max, step float64) *Spinner {
	return newSpinnerImpl(min, max, step, false)
}

func NewIntegralSpinner(min, max, step int) *Spinner {
[...]

Issue #709, opened 2020-02-24
Add a "SimpleWidget" or "StaticRenderer" so I can make canvas primitives interactive


andydotxyz:
### Is your feature request related to a problem? Please describe:
The amount of code required to make a simple widget is high, especially if I just want to wrap an existing CanvasObject.

### Is it possible to construct a solution with the existing API?
Yes, but it's a lot of code.

### Describe the solution you'd like to see:
Something like the following would allow me to make a tappable image: 

```go
type tappableImage struct {
    SimpleWidget
}

func (t *tappableImage) Tapped(_ *PointEvent) {
... etc 
}

func newTappableImage(img *canvas.Image) *tappableImage {
    ret := &tappableImage{}
    ret.SimpleWidget := widget.NewSimpleWidget(img) 
    return ret
}
```

gooid:
#### suggest:
```go
index 9d021e03..ab37f920 100644
--- a/internal/painter/gl/draw.go
+++ b/internal/painter/gl/draw.go
@@ -153,5 +153,9 @@ func (p *glPainter) drawObject(o fyne.CanvasObject, pos fyne.Position, frame fyn
                p.drawGradient(obj, p.newGlRadialGradientTexture, pos, frame)
        case fyne.Widget:
                p.drawWidget(obj, pos, frame)
+       default:
+               if obj, ok := o.(interface{ superCanvas() fyne.CanvasObject }); ok {
+                       p.drawObject(obj.superCanvas(), pos, frame)
+               }
        }
 }
```

#### cmd/fyne_demo/main.go
```go
// Package main loads a very basic Hello World graphical application.
package main

import (
	"fmt"

	"fyne.io/fyne"
	"fyne.io/fyne/app"
	"fyne.io/fyne/driver/desktop"
	"fyne.io/fyne/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Hello")

	hello := widget.NewLabel("Hello Fyne!")
	w.SetContent(widget.NewVBox(
		hello,
		//widget.NewButton("Hi!", func() {
		NewCustomButton("Hi!", func() {
			hello.SetText("Welcome :)")
		}),
	))

	w.ShowAndRun()
}

type CustomButton struct {
	*widget.Button
}

func (button *CustomButton) superCanvas() fyne.CanvasObject {
	return button.Button
}

// MouseIn is called when a desktop pointer enters the widget
func (button *CustomButton) MouseIn(ev *desktop.MouseEvent) {
[...]

PR #711, opened 2020-02-25
Spinner Widget [draft/work in progress]

DerThorsten:
### Description:
Spinner Widget implementation composed of button and entry widget + border layout

Fixes #(issue)
Missing Spinner Widget

- In #704  @andydotxyz  already hinted that I should rather use `entry.OnChanged` instead of relying on a press of `Enter`. But `OnChanged` does not work well with a spinner, because while entering a valid number one generates text which does not encode a number inbetween (for instance when typing `10.0` one has a `10.` at some point, or  for `1e-4` one has `1e`. Therefore I think enter is more appropriate. When spinner.GetValue() is called the current text is also evaluated (even without a press of enter).
- Also   @andydotxyz hinted to arrange the buttons vertically instead of horizontally, but this makes the height of the widget the height of two buttons which looks shitty (how would I resize  two buttons which are arranged horizontally to have exactly the height of the entry widget?)
- atm the widget extends a container canvas object, so all the public methods of the container are also visible in the spinner widget, which does not feel right, but not sure how to address this without manually adding all the methods from container which are needed by hand.
-   @andydotxyz suggested to implement the spinner by extending entry widget, but the entry widget is already very complicated and I did not like to read 1200 LOC just to implement something like a spinner which is clearly a composition of a text entry and two buttons (also I wanted to see how easy/hard it is to implement a custom widget by composing existing widgets)

### DIFF
--- a/cmd/fyne_demo/screens/widget.go
+++ b/cmd/fyne_demo/screens/widget.go
@@ -77,7 +77,8 @@ func makeFormTab() fyne.Widget {
 	password := widget.NewPasswordEntry()
 	password.SetPlaceHolder("Password")
 	largeText := widget.NewMultiLineEntry()
-
+	age := widget.NewIntegralSpinner(0, 100, 1)
+	weight := widget.NewSpinner(0.0, 100.0, 0.25)
 	form := &widget.Form{
 		OnCancel: func() {
 			fmt.Println("Cancelled")
@@ -88,13 +89,16 @@ func makeFormTab() fyne.Widget {
 			fmt.Println("Email:", email.Text)
 			fmt.Println("Password:", password.Text)
 			fmt.Println("Message:", largeText.Text)
+			fmt.Println("Age:", age.GetValue())
+			fmt.Println("Weight:", weight.GetValue())
 		},
 	}
 	form.Append("Name", name)
 	form.Append("Email", email)
 	form.Append("Password", password)
 	form.Append("Message", largeText)
-
+	form.Append("Age", age)
+	form.Append("Weight", weight)
 	return form
 }
 
--- a/widget/spinner.go
+++ b/widget/spinner.go
@@ -1,0 +1,197 @@
+package widget
+
+import (
+	"fmt"
+	"image/color"
+	"math"
+	"strconv"
+
+	"fyne.io/fyne"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/theme"
+)
+
+func sgn(a float64) float64 {
+	switch {
+	case a < 0:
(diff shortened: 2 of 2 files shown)


Kvaz1r:
> Therefore I think enter is more appropriate.

I would also handled focus lost event with Enter.  

> Also  @andydotxyz hinted to arrange the buttons vertically instead of horizontally, but this makes the height of the widget the height of two buttons which looks shitty (how would I resize  two buttons which are arranged horizontally to have exactly the height of the entry widget?)

That's because it's usual view for Spinner (although it doesn't mean that every GUI library should use only such way). As for size - each button just takes ((height of entry)/2 - minus some border value)

DerThorsten:
> As for size - each button just takes ((height of entry)/2 - minus some border value)
Sure, this is what we want, but not sure how to enforce this.

andydotxyz:
Thanks for working on this. As you'll be seeing I guess that getting API right is a lot more work than composing existing APIs.
I haven't gone through the code in detail, but in response to your comments introducing this I hope the following helps:

* OnChanged could work if you ignore the times when the text is invalid, then do the full update when it is valid. Also we should add OnChanged for this widget so people could subscribe to change notifications.
* You can see that we don't typically use getter functions but fields instead, that may play into the onChanged vs enter press conversation
* Yes if you use a basic button 2 of them are twice the height, the complete widget will probably have to be slightly more complex to get a better user experience?
* You're right about extending Container - we can't do this as every API will be exposed - we are very careful about only exposing API that absolutely is required. Extending Entry would be closer to the desired API I guessed - but perhaps it will have to be a complete custom widget to manage the API.

Perhaps some of this would be made easier if we implement the proposed #709 Simple Widget (as you could pass a container in to that so it's fully encapsulated?)

andydotxyz:
Anything we can do to help with this pull request?

Jacalz:
Hi @DerThorsten. Have you had the time to look at this again? You have done some good work on this spinner widget and it is definitely something that would be useful in the toolkit if you ask my me :)

andydotxyz:
The Entry widget now has ActionItem that might allow this to be implemented more easily (like the password show/hide button).
Are you still interested in picking this up @DerThorsten ?

Jacalz:
Hi @DerThorsten. I hope that you are doing well and continuing to enjoy using Fyne. We have recently opened a new community contributions repository (https://github.com/fyne-io/fyne-x) where new widgets can be added faster and then possibly move into here over time. We would recommend that this PR be updated and re-opened over there. Hopefully we can continue the discussion there 🙂

PR #715, opened 2020-02-28
Add Stroke Attributes and Omit if Empty

stuartmscott:
The SVG tool would skip stroke attributes because they were not defined in the struct with an xml annotation. The "omitempty" flag prevents unspecified attributes appearing in the output.

The following attributes were added;
- stroke
- stroke-width
- stroke-linecap
- stroke-linejoin
- stroke-dasharray

### DIFF
--- a/theme/svg.go
+++ b/theme/svg.go
@@ -24,40 +24,66 @@ type svg struct {
 }
 
 type pathObj struct {
-	XMLName xml.Name `xml:"path"`
-	Fill    string   `xml:"fill,attr"`
-	D       string   `xml:"d,attr"`
+	XMLName         xml.Name `xml:"path"`
+	Fill            string   `xml:"fill,attr,omitempty"`
+	Stroke          string   `xml:"stroke,attr,omitempty"`
+	StrokeWidth     string   `xml:"stroke-width,attr,omitempty"`
+	StrokeLineCap   string   `xml:"stroke-linecap,attr,omitempty"`
+	StrokeLineJoin  string   `xml:"stroke-linejoin,attr,omitempty"`
+	StrokeDashArray string   `xml:"stroke-dasharray,attr,omitempty"`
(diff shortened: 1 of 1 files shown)


stuartmscott:
To reduce repetition an Attributes struct could be declared;
```
type Attributes struct {
	Fill            string   `xml:"fill,attr,omitempty"`
	Stroke          string   `xml:"stroke,attr,omitempty"`
	StrokeWidth     string   `xml:"stroke-width,attr,omitempty"`
	StrokeLineCap   string   `xml:"stroke-linecap,attr,omitempty"`
	StrokeLineJoin  string   `xml:"stroke-linejoin,attr,omitempty"`
	StrokeDashArray string   `xml:"stroke-dasharray,attr,omitempty"`
}
```
And included in pathObj, rectObj, circleObj, polygonObj, and objGroup as an anonymous field;
```
type pathObj struct {
	XMLName xml.Name `xml:"path"`
	Attrs   Attributes
	D       string   `xml:"d,attr"`
}
```
According to https://golang.org/pkg/encoding/xml/ it is possible;
```
an anonymous struct field is handled as if the fields of its value were part of the outer struct.
```
Although this idea still remains to be tested

andydotxyz:
Thanks for the fix.
We do need more tests in this area, but this code is no less tested than before I guess.
Reducing the repetition as you indicate would be great if you'd like to open another PR.
If it's possible to add to the tests in that area as well we would be appreciative :)

stuartmscott:
You're welcome! I agree tests would be useful and it would be good to reduce the repetition, but before all of that I think its worth asking if svg tools should remain here or should be spun out into a new repo to keep this repo clean and focused.

andydotxyz:
That's a fair question. Given that it's just a thin layer over oksvg right now I think we're OK - but if it grows to something larger we probably should consider splitting.

PR #717, opened 2020-03-02
Refactoring: reduce widget dependencies to internal/cache

toaster:
### Description:

Some widget renderers implemented `Destroy` and used the internal cache in there.
These implementations were replaced by a proper implementation of `Show` and `Hide`.

### Checklist:

- ~~[ ] Tests included.~~
- [x] Tests all pass.

### DIFF
--- a/canvasobject.go
+++ b/canvasobject.go
@@ -17,7 +17,9 @@ type CanvasObject interface {
 
 	// visibility
 	Visible() bool
+	// Show shows this object.
 	Show()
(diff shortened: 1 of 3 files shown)


andydotxyz:
I have merged this but then it occurred to me - should it really re-show the popups on widget show?

toaster:
Maybe not :) … I'll make a PR today where show does not re-show the pop-up.

Issue #469, opened 2019-09-29
modal PopUp containing a Select closes early


seanpringle:
According to [the docs](https://godoc.org/fyne.io/fyne/widget#PopUp) a modal PopUp can contain any "standard elements", which I take to mean any CanvasObject.

Placing a Select inside a modal PopUp causes the latter to close when the former is opened, and the options then appear alone at the top-left. Other controls, like a Button, work fine and don't automatically close the modal.

```go
package main

import (
	"fyne.io/fyne/app"
	"fyne.io/fyne/widget"
	"log"
)

func main() {
	app := app.New()

	w := app.NewWindow("Hello")
	w.SetContent(widget.NewVBox(
		widget.NewLabel("Hello Fyne!"),
		widget.NewButton("Popup", func() {
			widget.NewModalPopUp(
				widget.NewVBox(
					widget.NewSelect([]string{"one", "two"}, func(item string) {
						log.Println(item)
					}),
				),
				w.Canvas(),
			)
		}),
	))

	w.ShowAndRun()
}
```

Real bug, or docs bug, or "just don't do that"? :-)

 - OS: Ubuntu
 - Version 18.04
 - Go version 1.13
 - Fyne version 1.1.1

andydotxyz:
Well found.
This is due to our simplistic implementation of overlays - the select box popup overwrites the modal popup.
This is indeed a bug but fixing it may not be trivial.

andydotxyz:
Moving to 1.2 release sorry, this will be quite complex to fix and we are working in that area for the larger release.

andydotxyz:
On develop for testing

Issue #707, opened 2020-02-24
inconsistent size/position handling for overlays


toaster:
The size and position handling of overlays is inconsistent and breaks things or causes dirty hacks.
An overlay is supposed to always have its canvas' size. That's why a canvas resizes its overlay to its size if it gets resized. Thus giving an overlay a different size doesn't make sense because it will be resized on a window resize.
The `PopUp` is designed to be an overlay which behaves like normal content (settable size and position) but under the hood it is an overlay and tries to satisfy the requirement of pos(0,0) and canvas size.
That's why `PopUp.Size()` doesn't report what `PopUp.Resize()` sets. But the latter is must not be used with the canvas size because it affects the `PopUp`'s content.
The same is true for the `PopUp`'s position.
Because of this there is `PopUp` specific code in the canvas which is obviously bad.

This whole issue also is the root cause of #706.

I propose to wrap overlay objects into an special (probably internal) structure. This should handle the size/position correctly while the size/position handling of the overlay's content works normally.
The details still need to be carved out but this would remove the cause of #706 and the special code in the canvas. It also will make the fix of #706 obsolete if one has been put in place.

andydotxyz:
is this resolved now @toaster ?

toaster:
No, no work has been done on this.

andydotxyz:
This is finally resolved on develop. It results in a slight change in position for some popups due to corrected shadow calculations. In addition the Overlays applied have different internal stack to be noted in release notes, but if `Show()` and `Hide()` of popup/dialog are used there will be no noticeable change - other than Move / Resize work as expected :)

andydotxyz:
Moved to H/2.8 due to slight behaviour change

PR #720, opened 2020-03-03
Bugfix/469 overlay stack interface

toaster:
### Description:

In preparation for a fix of #469 this PR introduces a new overlay interface to `fyne.Canvas`. The old methods (`Overlay`, `SetOverlay`) are deprecated.

### Checklist:

- ~~[ ] Tests included.~~
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/canvas.go
+++ b/canvas.go
@@ -22,8 +22,28 @@ type Canvas interface {
 	// Deprecated: Settings are now calculated solely on the user configuration and system setup.
 	SetScale(float32)
 
+	// Overlay returns the current overlay.
+	//
+	// Deprecated: Overlays are stacked now.
+	// This method returns the top of the overlay stack.
+	// Use Overlays() instead.
 	Overlay() CanvasObject
+	// Overlays returns the overlays currently on the overlay stack.
+	Overlays() []CanvasObject
+	// PopOverlay removes the top-most object of the overlay stack and returns it.
+	PopOverlay() CanvasObject
+	// PushOverlay pushes an overlay on the top of the overlay stack.
+	PushOverlay(CanvasObject)
+	// RemoveOverlay removes the given object and all objects above it from the overlay stack.
+	RemoveOverlay(CanvasObject)
+	// SetOverlay sets the overlay for the canvas.
+	//
+	// Deprecated: Overlays are stacked now.
+	// This method replaces the whole stack by the given overlay.
+	// Use PushOverlay() instead.
 	SetOverlay(CanvasObject)
+	// TopOverlay returns the top-most object of the overlay stack.
+	TopOverlay() CanvasObject
 
 	OnTypedRune() func(rune)
 	SetOnTypedRune(func(rune))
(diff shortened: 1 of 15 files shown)


andydotxyz:
Is there a way to do this without adding the long list of methods to the Canvas interface?

One more specific concern is that PopOverlay looks dangerous - for example something that pushes and later pops may be popping the wrong item. I like that RemoveOverlay removes all the child ones too, which makes me think that perhaps AddOverlay and RemoveOverlay is sufficient.

Back to the point above maybe Overlays() returns an OverlayStack or something? That is basically how it's implemented under the hood and would mean a single new method added to the interface. Just a thought.

Or even further outside of the box... Is it a ContentStack? The lowest level is the real content and the rest are overlays... but I just wondered are they not really the same thing?

toaster:
I'm not sure about the `ContentStack` what would be the benefits? How will the API look like? Would `window.SetContent()` still be in place?
Honestly, I think it would be better to not mix content and overlays for now.

Regarding `func Overlays() *OverlayStack`:
Do you mean that the canvas will not have methods for directly adding/removing the overlays?
So `canvas.PushOverlay(x)` will become `canvas.Overlays().Push(x)`?

Regarding `PopOverlay()`: I thought about this too. I just kept it because I didn't wanted to change `NewPopUpMenuAtPosition` :). So for me the three methods `AddOverlay(x)`, `RemoveOverlay(x)` and `TopOverlay()` would be sufficient.
Accessing these via `canvas.Overlays()` is acceptable for me, too.

andydotxyz:
Yeah ContentStack was a bit of a curve ball - SetContent() would basically be like ContentStack().Remove(oldContent); ContentStack().Add(newContent) - causing all overlays open to be dismissed like your current RemoveOverlay(). For some reason I imagined it could make some driver code simpler as it would remove the divide between content and overlay content. Maybe I was just getting carried away here :).

Yes, the Overlays() *OverlayStack would indeed then have the Push/Remove etc methods on it. I don't know why I thought of it, probably just to reduce the number of new calls on Canvas() when they would probably all be implemented by the same stack code anyway...

Your thoughts on PopOverlay make sense, but I think in this instance bigger changes to the existing widgets allow for a cleaner and less ambiguous API, so let's do that :)

andydotxyz:
Is it intended that this version of the stack only has 1 element? (or maybe I am reading it wrong?)
For the location of the stack implementation I thought we were avoiding new files directly in internal/?

Also, and I'm not sure about this: Given the methods on StackOverlay refers to overlays should they be just Add(), Remove() and Top()? The code using them can seem quite repetitive otherwise? (a bit more of an open question here).

toaster:
Yes, the real stack implementation will follow in a separate PR.
This PR should only introduce the necessary interfaces without changing anything on the functionality. It is meant as a “refactoring”.

I can put the the stack implementation deeper but not into `driver` directly because the `test` package depends on it and the `driver` package still depends on `widget` (because of #707) and `widget` tests are within the `widget` package which depends on `driver`.

andydotxyz:
Thanks. I guess the file can stay where it is until we get a solution to the circular dep, or find a nice sub-package for such things.

I had not thought about "All()" instead of "Overlays()" - I wonder if that implies an unsorted collection though, would "List()" be more explicit? (Maybe this is not an issue with Go where there is no "set" primitive?)

PR #722, opened 2020-03-05
Don't lay out parent object more than required

andydotxyz:
### Description:
Work on textgrid revealed this issue - in ensureMinSize the parent is laid out for every child.
Instead let's try to only do it once per parent.

### Checklist:

- [ ] Tests included. - not sure how to test an optimisation like this, suggestions?
- [x] Tests all pass.


andydotxyz:
Thanks. I’ll leave optimisation tests out until we start to focus on that work.

PR #724, opened 2020-03-05
fix calculating executable name in package command for apps using go modules

pymq:
### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


pymq:
Good idea, done

andydotxyz:
The code and tests look ready, thanks.
However the files removed from the vendor file don’t seem right, do you know why this has happened?

pymq:
I don't know, really. I just used `fyne vendor` to pull new dependencies. Maybe because of `+build ignore` tag in those files?
Reverted changes.

andydotxyz:
Many thanks. Can you confirm what platform you were running the vendor command on? We probably need to look into it further if our tooling is causing issues

pymq:
I am using Ubuntu Linux, Go 1.13.8, project is located outside of GOPATH

andydotxyz:
Thanks, I see the same on an Arch Linux box. Will look into this more and see what's going on.

andydotxyz:
It seems like these files are needed by macOS, I can't find a way to make the vendor command include them, I'll open another issue so we can track this.

Issue #723, opened 2020-03-05
Allow resetting the Select Widget selected field


Ariemeth:
### Is your feature request related to a problem? Please describe:

The Select Widget does not have a way to reset the currently selected option back to being empty.  The only options that can be set using `SetSelected` are ones that are in the options slice.  This means that one cannot set the selected value to be empty unless you have an empty string as part of the options slice.

### Is it possible to construct a solution with the existing API?

Including an empty string in the options slice allows you to set the selected value back to an empty string which resets back to the placeholder.

### Describe the solution you'd like to see:

I see two possible options right now.  
1.  Allow the SetOption function on the Select Widget to accept an empty string even if an empty string does not exist in the option slice effectively resetting the widget.
2. Add a Reset() type function that clears the currently selected value.

andydotxyz:
This is on develop branch for testing

PR #725, opened 2020-03-05
Support resetting a Select widget's current selection

Ariemeth:
### Description:

Adding the ability to reset the current selection on a Select widget causing it to use the PlaceHolder text.  This feature came out while having cascading Select widgets where one Select's selection determined what should be in the other Select.  When changing the first selection, the text being shown would stay the last chosen value.  The only way around this was to set the current selection to a value that is in the Select's option slice.  What is desired is that you could easily reset the current selection.  The other option would be to keep separate Selects around for each sub list and swap them out, but this is a cumbersome approach.

Fixes #(issue)
https://github.com/fyne-io/fyne/issues/723

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/select.go
+++ b/widget/select.go
@@ -199,19 +199,19 @@ func (s *Select) CreateRenderer() fyne.WidgetRenderer {
 // clearing the current option, the Select widget's PlaceHolder will
 // be displayed.
 func (s *Select) ClearSelected() {
-	s.updateSelect("")
+	s.updateSelected("")
 }
 
 // SetSelected sets the current option of the select widget
 func (s *Select) SetSelected(text string) {
 	for _, option := range s.Options {
 		if text == option {
-			s.updateSelect(text)
+			s.updateSelected(text)
 		}
 	}
 }
 
-func (s *Select) updateSelect(text string) {
+func (s *Select) updateSelected(text string) {
 	s.Selected = text
 
 	if s.OnChanged != nil {
--- a/widget/select.go
+++ b/widget/select.go
@@ -196,7 +196,7 @@ func (s *Select) CreateRenderer() fyne.WidgetRenderer {
 }
 
 // ClearSelected clears the current option of the select widget.  After
-// clearing the current option, the Select widgets PlaceHolder will
+// clearing the current option, the Select widget's PlaceHolder will
 // be displayed.
 func (s *Select) ClearSelected() {
 	s.Selected = ""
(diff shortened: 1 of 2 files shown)


widget/select.go
```diff
@@ -195,6 +195,19 @@ func (s *Select) CreateRenderer() fyne.WidgetRenderer {
 	return &selectRenderer{icon, text, shadow, objects, s}
 }
 
+// ClearSelected clears the current option of the select widget.  After
+// clearing the current option, the Select widgets PlaceHolder will
```
toaster:
```suggestion
// clearing the current option, the Select widget's PlaceHolder will
```

widget/select.go
```diff
@@ -195,6 +195,19 @@ func (s *Select) CreateRenderer() fyne.WidgetRenderer {
...
+// be displayed.
+func (s *Select) ClearSelected() {
```
toaster: The functionality of `ClearSelected()` should be supported by `SetSelected()` by always accepting the empty string. `ClearSelected()` could either be removed or kept as a convenience call for `SetSelected("")` then.
I'm undecided if the `ClearSelected()` should stay then. @andydotxyz what do you think?

The consolidation into `SetSelected()` would remove the double implementation of the `OnChanged` logic.

Ariemeth: Great feedback.  I wasn't sure which path would be better here.

andydotxyz:
I’d say that ClearSelected makes a nicer public API than SetSelected(“”) (especially as we move toward typed options ;) ). Agreed that the duplication should be resolved but the api looks good.

Ariemeth:
How about something like this
```golang
// ClearSelected clears the current option of the select widget.  After
// clearing the current option, the Select widget's PlaceHolder will
// be displayed.
func (s *Select) ClearSelected() {
	s.updateSelected("")
}

// SetSelected sets the current option of the select widget
func (s *Select) SetSelected(text string) {
	for _, option := range s.Options {
		if text == option {
			s.updateSelected(text)
		}
	}
}

func (s *Select) updateSelected(text string) {
	s.Selected = text

	if s.OnChanged != nil {
		s.OnChanged(s.Selected)
	}

	s.Refresh()
}
```
The change this would cause as is, would be that if a value passed into SetSelected doesn't exist in the options slice, it gets ignored completed.

Ariemeth:
or 
```golang
func (s *Select) selectChanged() {
	if s.OnChanged != nil {
		s.OnChanged(s.Selected)
	}

	s.Refresh()
}
```

PR #729, opened 2020-03-06
Bugfix: absolute position computation + Select.Move()

toaster:
### Description:

Moving a `Select` with an open pop-up now moves the pop-up, too.
This implies a fix to `driver.AbsolutePositionForObject`.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
Yup, thanks for fixing the TODO in test driver too :)

PR #732, opened 2020-03-08
Feature/drop down entry

toaster:
### Description:

This PR introduces `SelectEntry`: an `Entry` field with options to select from.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/test/testcanvas.go
+++ b/test/testcanvas.go
@@ -167,10 +167,10 @@ func (c *testCanvas) Capture() image.Image {
 	return img
 }
 
-// LaidOutObjects returns all fyne.CanvasObject starting at the given fyne.CanvasObject which is laid out with the given size.
+// LaidOutObjects returns all fyne.CanvasObject starting at the given fyne.CanvasObject which is laid out previously.
 func LaidOutObjects(o fyne.CanvasObject) (objects []fyne.CanvasObject) {
 	if o != nil {
-		objects = layoutAndCollect(objects, o, o.MinSize())
+		objects = layoutAndCollect(objects, o, o.MinSize().Union(o.Size()))
 	}
 	return objects
 }
--- a/test/testcanvas.go
+++ b/test/testcanvas.go
@@ -3,6 +3,7 @@ package test
 import (
 	"image"
 	"image/draw"
+	"reflect"
 
 	"fyne.io/fyne"
 	"fyne.io/fyne/internal"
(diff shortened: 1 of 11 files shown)


Jacalz:
Thought of this a few days ago. Nice job, looking forward to having this in fyne :+1:

test/testcanvas.go
```diff
@@ -167,6 +168,32 @@ func (c *testCanvas) Capture() image.Image {
 	return img
 }
 
+// CanvasItem is a part of a canvas and returned by InspectCanvasItems().
+type CanvasItem struct {
```
andydotxyz: Is it worth adding a new type for this? The resulting is string comparison instead of compile time type checks - in the same number of lines of code.

test/testcanvas.go
```diff
@@ -167,6 +168,32 @@ func (c *testCanvas) Capture() image.Image {
...
+	Type   string
+	Object fyne.CanvasObject
+}
+
+// InspectCanvasItems returns all CanvasItems starting at the given CanvasObject which is laid out with the given size.
+func InspectCanvasItems(o fyne.CanvasObject, size fyne.Size) (objects []CanvasItem) {
```
andydotxyz: This name is quite abstract, really what this is doing is simply flattening a tree of objects?
Every usage of this method passes an item and it's size - could it be simplified to not pass the size at all?
I wonder if "ListCanvasObjects(o fyne.CanvasObject) []fyne.CanvasObject" or "ListObjectTree(o fyne.CanvasObject) []fyne.CanvasObject" is all that's needed?

In fact, seeing that in every usage you are then iterating over them perhaps a foreach pattern instead?
"ForEachCanvasObject(o fyne.CanvasObject, func(c fyne.CanvasObject))"? in essence it's a simpler version of the internal walk method we use.

toaster:
Like so? :)

andydotxyz:
Thanks. The docs say "laid out with the given size" but there is no size given.
Also I noticed that it moved from using .Size() to .MinSize() - is that right?
Which made me wonder - were the objects not already at MinSize? If so that seems like a bug...

andydotxyz:
Oh, also some of the tests use .String() then a string comparison for type - it could be risky if the code gets refactored... but I guess that is unlikely

toaster:
> Thanks. The docs say "laid out with the given size" but there is no size given.

Ooops

> Also I noticed that it moved from using .Size() to .MinSize() - is that right?

Yes.

> Which made me wonder - were the objects not already at MinSize? If so that seems like a bug...

Not necessarily, many tests simply create objects without ever set their size or position. If needed this helper can be extended to not shrink the object if its size is larger than its min size.

toaster:
> Oh, also some of the tests use .String() then a string comparison for type - it could be risky if the code gets refactored... but I guess that is unlikely

Hm, I don't understand what you mean here. What is a “string comparison for type”?

andydotxyz:
With regards to string comparison I guess I'm referring to the usages of reflect: 1 of the two is https://github.com/fyne-io/fyne/pull/732/files#diff-c13dbcb895beb1136edf77adbbcd4318R92

It's just a bit more fragile than doing an actual type cast check like you do on other lines in this patch.

toaster:
Ah, now I understand. You are right, the usage of `reflect.TypeOf()` is no longer necessary because I changed the drop-down button to an exported widget which simply be can checked via type assertion in the `widget_test` package.
Thanks!

PR #738, opened 2020-03-10
Split Tappable interface

andydotxyz:
### Description:
Remove the requirement for people to implement TappedSecondary() when they only want Tapped().

### Checklist:

- [ ] Tests included. - no new tests
- [x] Tests all pass

- [ ] Any breaking changes have a deprecation path or have been discussed. - it removes empty methods, is this breaking or not? :)

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -579,7 +579,7 @@ func (w *window) mouseOut() {
 	})
 }
 
-func (w *window) mouseClicked(viewport *glfw.Window, btn glfw.MouseButton, action glfw.Action, mods glfw.ModifierKey) {
+func (w *window) mouseClicked(_ *glfw.Window, btn glfw.MouseButton, action glfw.Action, mods glfw.ModifierKey) {
 	co, pos := w.findObjectAtPositionMatching(w.canvas, w.mousePos, func(object fyne.CanvasObject) bool {
 		if _, ok := object.(fyne.Tappable); ok {
 			return true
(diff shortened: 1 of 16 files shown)


internal/driver/glfw/window.go
```diff
@@ -658,17 +660,19 @@ func (w *window) mouseClicked(viewport *glfw.Window, btn glfw.MouseButton, actio
...
 		} else if action == glfw.Release {
-			if wid == w.mousePressed {
+			if co == w.mousePressed {
 				switch button {
 				case desktop.RightMouseButton:
-					w.queueEvent(func() { wid.TappedSecondary(ev) })
+					w.queueEvent(func() { co.(fyne.SecondaryTappable).TappedSecondary(ev) })
 				default:
-					w.queueEvent(func() { wid.Tapped(ev) })
+					w.queueEvent(func() { co.(fyne.Tappable).Tapped(ev) })
```
toaster: You should better add some test cases :).
This will panic if the obj search finds a `SecondaryTappable` but the mouse event was a primary click and vice versa.

andydotxyz: I'm actually going to defend this - the two objects will not be equal so 5 lines above will be false and this code will not execute...

I certainly could add tests to show the code doesn't run if that helps ;)

toaster: Please write a test. I still think it will fail if you right-click on a left-click-only object and if you left-click on a right-click-only object.

andydotxyz:
Quite right, sorry for pushing back. I'll get right on this :)

andydotxyz:
Oops, the commit title should have mentioned crash that @toaster predicted, obviously!

PR #739, opened 2020-03-10
Remove Focus func from Focused interface

andydotxyz:
### Description:
I saw that Focused() is never used, so let's get rid of it!?

### Checklist:

- [ ] Tests included. - no new tests
- [x] Tests all pass.

- [x] Any breaking changes have a deprecation path or have been discussed. - this could be considered breaking.

It removes Entry and CheckBox Focus() funcs as they are not actually needed.
Should I re-add them as deprecated in case anyone was calling them, despite their lack of use internally?

### DIFF
--- a/canvasobject.go
+++ b/canvasobject.go
@@ -65,6 +65,7 @@ type Draggable interface {
 type Focusable interface {
 	FocusGained()
 	FocusLost()
+	Focused() bool // Deprecated: this is an internal detail, canvas tracks current focused object
 
 	TypedRune(rune)
 	TypedKey(*KeyEvent)
--- a/canvasobject.go
+++ b/canvasobject.go
@@ -65,7 +65,6 @@ type Draggable interface {
 type Focusable interface {
 	FocusGained()
 	FocusLost()
-	Focused() bool
 
 	TypedRune(rune)
 	TypedKey(*KeyEvent)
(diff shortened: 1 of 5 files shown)


toaster:
I'm wondering if this is still a breaking change. Imagine someone wrote the following code:
```go
if f, ok := o.(fyne.Focusable); ok && f.Focused() {
  // do something
}
```
it will break, won't it?

andydotxyz:
Yes, that is true.
I'm torn on this now. If we need to re-add the method so that the above code continues to work then it shows that people actually are using the feature. I assumed that as it was never used internally that it wasn't needed at all...

The reason behind it was I have found myself implementing the interface a number of times and usually you don't care if you are focused or not - but to satisfy the requirement you must track state you don't care about.

What do you think? Should I actually just close this instead? 
It's in the vicinity of the Tappable split - aiming to make external code easier, but perhaps this was a step too far?

toaster:
I'm against dropping this. I agree that `.Focused()` is _probably_ superfluous. That's why we should deprecate it _in the interface_. So we can remove it in 2.0. This won't help developers with the current version, that's right. But if we don't deprecate it, it will be an annoyance forever.

andydotxyz:
OK, I think that fixes it

PR #745, opened 2020-03-13
Add file dialog to open and save(as) files

### LINKED ISSUE #225: Open file dialog
I'm not sure what the limitations are with opengl but is there a way to get the open file dialog?

andydotxyz:
### Description:
This PR opens a traditional file dialog overlaying the app in either open or save mode.
It returns the file path through a callback like the other dialogs.

On mobile there are many challenges to file management, and so we launch OS specific screens instead. This is just open, no save support on mobile yet.
The paths returned on mobile may not be to writeable (or readable) file space. Fixing that is the topic of a further item of work on file management...

Fixes #225 

### Checklist:

- [x] Tests included. - except the mobile code which spans 4 languages and is rather tricky to test at all
- [x] Tests all pass.


andydotxyz:
p.s. screenshots are on #225 for reference

dialog/file_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+func TestShowFileOpen(t *testing.T) {
+	chosen := ""
+	win := test.NewWindow(widget.NewLabel("Content"))
+	ShowFileOpen(func(file string) {
+		chosen = file
+	}, win)
+
+	popup := win.Canvas().Overlays().Top().(*widget.PopUp).Content
+	defer win.Canvas().Overlays().Remove(popup)
```
toaster: The `defer` won't work because the pop-up's content is not the overlay.

dialog/file.go
```diff
@@ -0,0 +1,244 @@
...
+	ui := d.makeUI()
+	dir, err := os.UserHomeDir()
+	if err != nil {
+		fyne.LogError("Could not load user home dir", err)
+		dir, _ = os.Getwd() //fallback
+	}
+	d.setDirectory(dir)
+
+	spacer := canvas.NewRectangle(theme.BackgroundColor())
+	spacer.SetMinSize(fyne.NewSize(436, 300))
```
toaster: A fixed min size? Where come these numbers from?

dialog/file_test.go
```diff
@@ -0,0 +1,117 @@
...
+		return
+	}
+
+	// This will only execute if we have a file in the home path.
+	// Until we have a way to set the directory of an open file dialog.
+	test.Tap(target)
+	assert.Equal(t, filepath.Base(target.path), nameLabel.Text)
+	assert.False(t, open.Disabled())
+
+	test.Tap(open)
```
toaster: You should assert that the pop-up has been removed here.

dialog/file_test.go
```diff
@@ -0,0 +1,117 @@
...
+	assert.Equal(t, target.path, chosen)
+}
+
+func TestShowFileSave(t *testing.T) {
```
toaster: the same comments as for the “open” test apply here

dialog/fileicon.go
```diff
@@ -0,0 +1,120 @@
...
+
+const (
+	fileIconSize = 64
+	fileTextSize = 24
+)
+
+type fileIcon struct {
+	widget.BaseWidget
+	parent  *fileDialog
+	current bool
```
toaster: I'd prefer `isCurrent` for better reading.

dialog/file.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+	}
+
+	f.files.Objects = icons
+	f.files.Refresh()
+}
+
+func (f *fileDialog) setFile(file *fileIcon) {
+	if f.current != nil {
+		f.current.current = false
```
toaster: `.current.current` stutters. Even with the later proposed change of the latter to `isCurrent` the resulting `.current.isCurrent` wouldn't read good. A current thing is supposed to be current.

I suggest to rename `f.current` to something more expressive. Maybe `f.selection`?

The name of `setFile` should be adjusted accordingly (e.g. `setSelection`).

And the name of the type `fileIcon` is misleading. How about `fileDialogItem`?

dialog/fileicon.go
```diff
@@ -0,0 +1,120 @@
...
+	"fyne.io/fyne/theme"
+	"fyne.io/fyne/widget"
+)
+
+const (
+	fileIconSize = 64
+	fileTextSize = 24
+)
+
+type fileIcon struct {
```
toaster: This name is misleading (see comment on `fileDialog`).

dialog/fileicon.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"fyne.io/fyne"
+	"fyne.io/fyne/canvas"
+	"fyne.io/fyne/theme"
+	"fyne.io/fyne/widget"
+)
+
+const (
+	fileIconSize = 64
+	fileTextSize = 24
```
toaster: Shouldn't the theme be the only thing to hold hard-coded sizes?

dialog/fileicon.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+type fileIcon struct {
+	widget.BaseWidget
+	parent  *fileDialog
```
toaster: The name `parent` suggests a line of ancestors (dir->dir->dir->file) which does not exist.
I would prefer `dialog` to refer to the dialog.

dialog/fileicon.go
```diff
@@ -0,0 +1,120 @@
...
+	text    *widget.Label
+	objects []fyne.CanvasObject
+}
+
+func (s fileIconRenderer) Layout(size fyne.Size) {
+	iconAlign := (size.Width - fileIconSize) / 2
+	s.img.Resize(fyne.NewSize(fileIconSize, fileIconSize))
+	s.img.Move(fyne.NewPos(iconAlign, 0))
+	s.ext.Resize(fyne.NewSize(fileIconSize, fileTextSize))
+	ratioDown := 0.45
```
toaster: hard-coded size (adjustment) outside a theme

dialog/file.go
andydotxyz: A fair question - in my testing it was just the right size to show 4 file icons wide and 2 high.
You're right though this could be calculated. And now I think about it 2.2 or 2.5 high is probably better so you can more easily see to scale!

dialog/file.go
```diff
@@ -0,0 +1,244 @@
...
+			ShowConfirm("Overwrite?", "Are you sure you want to overwrite the file\n"+name+"?",
+				func(ok bool) {
+					if !ok {
+						f.callback("")
+						return
+					}
+
+					f.callback(path)
+					f.win.Hide()
+				}, f.parent)
```
spatocode: Could be better if we check if the existed file is really a file and not a dir. Replacing a dir with file is not common.

dialog/fileicon.go
andydotxyz: As this relates specifically to icons in the file dialog grid (currently) I don't think it's as heavily related to theme as the items above...

dialog/fileicon.go
andydotxyz: OK, this is the only thing I have not resolved in the updated commits.
Is it worth breaking any existing custom themes (as the API would have to be extended) ?
Can we re-incorporate it in a future (also 1.3) theme API that is extensible so it's not an ongoing challenge?

andydotxyz:
I think this is ready to re-review.
I am adding more tests to the (dialog)fileIcon and realised that some of the code could be refactored.
I might wait before pushing that though, so it's easier to review the recent changes?

andydotxyz:
Oh, this also includes the addition of ScrollContainer.SetMinSize discussed on slack - it is a more elegant solution than the "spacer" struct and meant that the correct sizes could be easily calculated.

widget/scroller.go
```diff
@@ -371,10 +372,20 @@ func (s *ScrollContainer) Dragged(e *fyne.DragEvent) {
 
 // MinSize returns the smallest size this widget can shrink to
 func (s *ScrollContainer) MinSize() fyne.Size {
+	if s.minSize.Width > 0 && s.minSize.Height > 0 {
```
toaster:
```suggestion
	if s.minSize != fyne.Size{} {
```

widget/scroller_test.go
```diff
@@ -27,6 +27,18 @@ func TestNewScrollContainer(t *testing.T) {
...
+	rect.SetMinSize(fyne.NewSize(500, 50))
+	scroll := NewScrollContainer(rect)
+	scroll.SetMinSize(fyne.NewSize(100, 100))
+	test.WidgetRenderer(scroll).Layout(scroll.minSize)
+
+	assert.Equal(t, fyne.NewSize(500, 100), rect.Size())
+	assert.Equal(t, 0, scroll.Offset.X)
+	assert.Equal(t, 0, scroll.Offset.Y)
+}
+
```
toaster: I miss the test of `scroll.MinSize()` here.
Also you access the scroll's private field `minSize`.

toaster: N.B.: The `MinSize()` test should also test the fallback code path.

dialog/fileicon.go
```diff
@@ -11,41 +11,43 @@ import (
...
 const (
-	fileIconSize = 64
-	fileTextSize = 24
+	fileIconSize      = 64
+	fileTextSize      = 24
+	fileIconCellWidth = fileIconSize * 1.25
 )
 
-type fileIcon struct {
+type fileDialogIcon struct {
```
toaster: While this is a better name than the old one, I'm still confused by it because this is a complete widget while an icon is just some kind of resource IMO.
Maybe that's just my biased view?
I often name variables representing complex structures by their type like `win := Window.create()` or `dialog := NewDialog()`. Doing this with the `fileDialogIcon` and accessing its `icon` property would lead to stuttering:
```go
icon := fileDialogIcon{…}
if icon.icon != nil {
  …
}
```
I think an icon _is_ an icon and doesn't _have_ an icon :).

widget/scroller.go
toaster: P.S.: Yes, this is different to prior test but I think it meets expectations better.

dialog/fileicon.go
toaster: 1. No, breaking custom themes doesn't sound like a good idea.
2. Yeah, I think so.

widget/scroller_test.go
andydotxyz: Apologies that the test was not explicit - layout will use the minsize value, but I will update this to be clearer.

dialog/fileicon.go
andydotxyz: Good point - how about fileDialogItem?

widget/scroller.go
andydotxyz: I had to revert this, it doesn't compile

widget/scroller_test.go
andydotxyz: Pushed  new version of the tests

widget/scroller_test.go
```diff
@@ -31,9 +31,13 @@ func TestScrollContainer_MinSize(t *testing.T) {
 	rect := canvas.NewRectangle(color.Black)
 	rect.SetMinSize(fyne.NewSize(500, 50))
 	scroll := NewScrollContainer(rect)
-	scroll.SetMinSize(fyne.NewSize(100, 100))
+	assert.Equal(t, fyne.NewSize(scrollContainerMinSize, scrollContainerMinSize), scroll.MinSize())
```
toaster: Please don't use internal constants for test expectations. Use literals instead. As long as the constant is not exported it is an implementation detail.
For exported constants I would recommend to test the constant's value separately and test functions returning its value using the constant.

widget/scroller.go
```diff
@@ -372,7 +372,7 @@ func (s *ScrollContainer) Dragged(e *fyne.DragEvent) {
 
 // MinSize returns the smallest size this widget can shrink to
 func (s *ScrollContainer) MinSize() fyne.Size {
-	if s.minSize != fyne.Size{} {
```
toaster: What was the issue with this?

andydotxyz: It doesn't compile:
```
syntax error: unexpected { at end of statement
syntax error: non-declaration statement outside function body
```

andydotxyz:
Ok, pushed changes to respond to comments.
Apologies for previous formatting issues, I must have accidentally skipped commit hooks :(.

Still to do:
* fileDialogItem tests
* resolve conversation about return types (in next message)

widget/scroller_test.go
andydotxyz: That hadn't really crossed my mind, but happy to use it as a rule

andydotxyz:
In further investigation it seems unlikely that returning a filepath is going to work on mobile devices (some types of file return a URI and others provide access to a stream. In other instances access to the path needs to be surrounded by security elevation as well!)

So I think we need a more sophisticated return type to handle file i/o.
I am considering something like: `fyne.UserFile` (or using a new package `storage.File`):
```
func Name() string
func ReadOnly() bool
func OpenRead() FileReader
func OpenWrite() FileWriter
```

And the supporting types are basically `io.Reader` or `io.Writer` with an additional `Close()` method. This allows us to manage access to files on behalf of the user and serialise to/from various sources. I think that we can also use these same types for managing the storage described in the "App.Documents" section of https://github.com/fyne-io/fyne/wiki/Settings-API.

What I have not yet figured out is how we can re-open the same file in a new run of the application...

andydotxyz:
How about using URIs? files and streams all have a unique ID of some sort.
so `URI() string` could be used to store a reference and `FileFromURI(string)` could re-populate a file type manually.
Does this work for people?

widget/scroller.go
```diff
@@ -371,10 +373,20 @@ func (s *ScrollContainer) Dragged(e *fyne.DragEvent) {
 
 // MinSize returns the smallest size this widget can shrink to
 func (s *ScrollContainer) MinSize() fyne.Size {
+	if s.minSize.Width > 0 && s.minSize.Height > 0 {
```
toaster:
```suggestion
	if if s.minSize != (fyne.Size{}) {
```

toaster: second try :)

andydotxyz: Oh, good catch! is the deliberate mistake there just to keep me on my toes? ;)

andydotxyz:
I'm happy to work on the return type as a separate PR, I had forgotten that we're not committing to final API just by merging it :)

andydotxyz:
OK, missing tests added now, I think if we are pushing return type to another PR then this one is ready to go.

widget/scroller.go
toaster: Ehm … yes, yes of course :).

toaster: Fixed the proposal.

toaster: Silly me, I should read ahead.

andydotxyz:
> * I think it would make more sense to have it pop out of the window into a new window.

That's complex - we currently cannot block input on 1 window whilst popping out another...

> Currently having too small window since makes it unusable.

I may be able to make it possible to be smaller... dialogs are currently always minsize which is why it doesn't shrink - really the minsize I set is desired size, so this could probably be improved.

> Some kind of right click context menu for stuff 

Feature request or future PR :) (but I think you're right - though not certain of whether it's a context menu)

> It doesn't really appreciate long folder and file names (needs some sort of text wrapping and I know that it is work in progress).

Yeah, I left it for that reason... could do a "poor man's truncation" if it's required, let's see if wrap/elipsis code fixes it first but it should not be released like this you're right.

andydotxyz:
OK, I have an approach that will ensure it can appear smaller, but only as small as sensible ;).
Need to land #757 as the modal dialog was not respecting resize after the non-modal changes...

ThembaniSt:
my machine is unable to install ,go v1.13.7

andydotxyz:
Sorry @ThembaniSt but unless you provide a little more information there isn’t anything we can do for you.

Can’t install what?
What error message?
What type of computer?
Should this be a new bug report?

andydotxyz:
Just for @Jacalz I pushed a small commit after this was merged that allows the dialog to go smaller if the window cannot accomodate.
Smaller than this size and it will do it's best but can't promise anything ;)

Screenshot 2020-03-25 at 12 13 23

tataDan:
If the code for the ShowFileOpen function that is currently in the feature/filehandling brach is adopted, will there be a method in the FileReader (and possibly FileWriter) interface named something like "Path()"?  This would be similar to Name() and URI() except that it would return the full path as opposed to the file name only and would not contain "file://". This could help eliminate using code like the following when getting the path from ShowFileOpen():

`path := file.URI()

path = strings.ReplaceAll(path, "file://", "")`

Instead it would be something like: `path := file.Path()`

andydotxyz:
Probably not. The code that you illustrate is not advisable, as we cannot guarantee that the data came from a filesystem resource.

Mobile devices make it almost impossible to read from a path.
You should use the Open() function to get access to the underlying data.

Uldiniad:
I have seen mention in the code comments that the save file dialog is supposed to allow saving to a new file. I don't see an option for that (I do see the option to overwrite an existing file though). Am I missing something?

andydotxyz:
Browse to the directory you want to save and write the  filename into the entry at the bottom left?

Uldiniad:
ah yeah. i couldn't see that was a textbox on white theme sorry. thanks :)

Issue #726, opened 2020-03-06
Allow widgets to influence mouse cursor style


AhmedMKamal:
Is there a way to set the cursor style on a widget, like pointer, text, grab, etc?

andydotxyz:
We currently do not support this sorry. If you could provide more information about what you’re looking to do we could try and add something

AhmedMKamal:
@andydotxyz I'm trying to build a custom entry widget for a source code editor app, for now, my custom widget is working quite well but I can't set the cursor style to text when I hover over it.

here's an example from `.Net WPF` which I used to work with
```c#
// listen to mouse enter event.
private void MyWidgetMouseEnter(object sender, SelectionChangedEventArgs e)
{
  // set the cursor to pen.
  DisplayArea.Cursor = Cursors.Pen;
}

// listen to mouse leave event.
private void MyWidgetMouseLeave(object sender, SelectionChangedEventArgs e)
{
  // set the cursor to cursor.
  DisplayArea.Cursor = Cursors.Cursor;
}
```

andydotxyz:
If you are extending the Entry widget this should "just work" (but I think there is a bug that causes it not to). I guess we could create some desktop extension that can hint at a different cursor type. It's not something we can support cross-platform for obvious reasons.

AhmedMKamal:
@andydotxyz, I understand, I'll try to dive deeper into `fyne`'s source code and see if I can help.

andydotxyz:
This is on develop for testing

PR #750, opened 2020-03-15
Handle window cursor

spatocode:
### Description:
This PR creates a better way of handling the window cursor with respect to canvasobjects.

Fixes #(726)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/canvasobject.go
+++ b/canvasobject.go
@@ -79,9 +79,3 @@ type Focusable interface {
 type Shortcutable interface {
 	TypedShortcut(Shortcut)
 }
-
-// Cursorable describes any CanvasObject that needs a cursor change
-type Cursorable interface {
-	Cursor() Cursor
-	SetCursor(Cursor)
-}
--- a/canvasobject.go
+++ b/canvasobject.go
@@ -79,3 +79,9 @@ type Focusable interface {
 type Shortcutable interface {
 	TypedShortcut(Shortcut)
 }
+
+// Cursorable describes any CanvasObject that needs a cursor change
+type Cursorable interface {
+	Cursor() Cursor
+	SetCursor(Cursor)
+}
--- a/driver/desktop/cursor.go
+++ b/driver/desktop/cursor.go
@@ -1,4 +1,4 @@
-package fyne
+package desktop
 
 // Cursor represents a standard fyne cursor
 type Cursor string
@@ -12,8 +12,13 @@ const (
 	CrosshairCursor Cursor = "CrosshairCursor"
 	// PointerCursor is the cursor often used to indicate a link
 	PointerCursor Cursor = "PointerCursor"
-	// HResizeCursor is the cursor often used for horizontal resize
+	// HResizeCursor is the cursor often used to indicate horizontal resize
 	HResizeCursor Cursor = "HResizeCursor"
-	// VResizeCursor is the cursor often used for vertical resize
+	// VResizeCursor is the cursor often used to indicate vertical resize
 	VResizeCursor Cursor = "VResizeCursor"
 )
+
+// Cursorable describes any CanvasObject that needs a cursor change
+type Cursorable interface {
+	Cursor() Cursor
+}
--- a/driver/desktop/cursor.go
+++ b/driver/desktop/cursor.go
@@ -1,21 +1,21 @@
 package desktop
 
 // Cursor represents a standard fyne cursor
-type Cursor string
+type Cursor int
 
 const (
 	// DefaultCursor is the default cursor typically an arrow
-	DefaultCursor Cursor = "DefaultCursor"
+	DefaultCursor Cursor = iota
 	// TextCursor is the cursor often used to indicate text selection
-	TextCursor Cursor = "TextCursor"
+	TextCursor
 	// CrosshairCursor is the cursor often used to indicate bitmaps
-	CrosshairCursor Cursor = "CrosshairCursor"
+	CrosshairCursor
 	// PointerCursor is the cursor often used to indicate a link
-	PointerCursor Cursor = "PointerCursor"
+	PointerCursor
 	// HResizeCursor is the cursor often used to indicate horizontal resize
-	HResizeCursor Cursor = "HResizeCursor"
+	HResizeCursor
 	// VResizeCursor is the cursor often used to indicate vertical resize
-	VResizeCursor Cursor = "VResizeCursor"
+	VResizeCursor
 )
 
 // Cursorable describes any CanvasObject that needs a cursor change
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -18,7 +18,6 @@ import (
 	"fyne.io/fyne/internal/cache"
 	"fyne.io/fyne/internal/driver"
 	"fyne.io/fyne/internal/painter/gl"
-	"fyne.io/fyne/widget"
 
 	"github.com/go-gl/glfw/v3.3/glfw"
 )
@@ -29,16 +28,9 @@ const (
 )
 
 var (
-	defaultCursor, entryCursor, hyperlinkCursor *glfw.Cursor
-	initOnce                                    = &sync.Once{}
-	defaultTitle                                = "Fyne Application"
+	initOnce     = &sync.Once{}
+	defaultTitle = "Fyne Application"
 )
-
-func initCursors() {
-	defaultCursor = glfw.CreateStandardCursor(glfw.ArrowCursor)
-	entryCursor = glfw.CreateStandardCursor(glfw.IBeamCursor)
-	hyperlinkCursor = glfw.CreateStandardCursor(glfw.HandCursor)
-}
 
 // Declare conformity to Window interface
 var _ fyne.Window = (*window)(nil)
@@ -502,17 +494,31 @@ func (w *window) findObjectAtPositionMatching(canvas *glCanvas, mouse fyne.Posit
 	return driver.FindObjectAtPositionMatching(mouse, matches, canvas.Overlays().Top(), roots...)
 }
 
+var cursorMap = map[fyne.Cursor]glfw.StandardCursor{
+	fyne.DefaultCursor:   glfw.ArrowCursor,
+	fyne.TextCursor:      glfw.IBeamCursor,
+	fyne.CrosshairCursor: glfw.CrosshairCursor,
+	fyne.PointerCursor:   glfw.HandCursor,
+	fyne.HResizeCursor:   glfw.HResizeCursor,
+	fyne.VResizeCursor:   glfw.VResizeCursor,
+}
+
+func cursorToName(cursor fyne.Cursor) *glfw.Cursor {
+	cur, ok := cursorMap[cursor]
+	if !ok {
+		return glfw.CreateStandardCursor(glfw.ArrowCursor)
+	}
+	return glfw.CreateStandardCursor(cur)
+}
+
 func (w *window) mouseMoved(viewport *glfw.Window, xpos float64, ypos float64) {
 	w.mousePos = fyne.NewPos(internal.UnscaleInt(w.canvas, int(xpos)), internal.UnscaleInt(w.canvas, int(ypos)))
 
-	cursor := defaultCursor
+	cursor := cursorToName(fyne.DefaultCursor)
 	obj, pos := w.findObjectAtPositionMatching(w.canvas, w.mousePos, func(object fyne.CanvasObject) bool {
-		entry, entOk := object.(*widget.Entry)
-		_, selOk := object.(*widget.SelectEntry)
-		if (entOk && !entry.Disabled()) || selOk {
-			cursor = entryCursor
-		} else if _, ok := object.(*widget.Hyperlink); ok {
-			cursor = hyperlinkCursor
+		if cursorable, ok := object.(fyne.Cursorable); ok {
+			fyneCursor := cursorable.Cursor()
+			cursor = cursorToName(fyneCursor)
 		}
 
 		_, hover := object.(desktop.Hoverable)
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -47,6 +47,7 @@ var _ fyne.Window = (*window)(nil)
 
 type window struct {
 	viewport *glfw.Window
+	cursor   *glfw.Cursor
 	painted  int // part of the macOS GL fix, updated GLFW should fix this
 	canvas   *glCanvas
 	title    string
@@ -533,6 +534,7 @@ func (w *window) mouseMoved(viewport *glfw.Window, xpos float64, ypos float64) {
 		return hover
 	})
 
+	w.cursor = cursor
 	viewport.SetCursor(cursor)
 	if obj != nil && !w.objIsDragged(obj) {
 		ev := new(desktop.MouseEvent)
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -507,10 +507,10 @@ func (w *window) mouseMoved(viewport *glfw.Window, xpos float64, ypos float64) {
 
 	cursor := defaultCursor
 	obj, pos := w.findObjectAtPositionMatching(w.canvas, w.mousePos, func(object fyne.CanvasObject) bool {
-		if wid, ok := object.(*widget.Entry); ok {
-			if !wid.Disabled() {
-				cursor = entryCursor
-			}
+		entry, entOk := object.(*widget.Entry)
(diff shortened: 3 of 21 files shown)


internal/driver/glfw/window.go
```diff
@@ -507,10 +507,10 @@ func (w *window) mouseMoved(viewport *glfw.Window, xpos float64, ypos float64) {
 
 	cursor := defaultCursor
 	obj, pos := w.findObjectAtPositionMatching(w.canvas, w.mousePos, func(object fyne.CanvasObject) bool {
-		if wid, ok := object.(*widget.Entry); ok {
-			if !wid.Disabled() {
-				cursor = entryCursor
-			}
+		entry, entOk := object.(*widget.Entry)
+		_, selOk := object.(*widget.SelectEntry)
+		if (entOk && !entry.Disabled()) || selOk {
```
andydotxyz: This ignores the disabled state of the select entry

andydotxyz:
I think tests here could have caught the error whereby disabled state was ignored.
I know there are no cursor tests yet, so I have attached a diff that puts them in place for you.
Perhaps combining this with additions for your patch will round out the test side as well?
[cursor-test.diff.txt](https://github.com/fyne-io/fyne/files/4335234/cursor-test.diff.txt)

spatocode:
> Good catch of another dependency of an internal package from `widget`.
> These should be removed where found because they introduce several kinds of problems.
> For instance nobody can create a custom entry widget with this cursor support. This should be solved the right way, i.e. we have to think about it.
> 
> Currently I can see two solutions to this:
> 
>     1. introduce special interfaces for cursor adjustment (currently entries and links: `fyne.Entry`, `fyne.Hyperlink`) and leave the selection of the cursor in the internal package
> 
>     2. introduce a new interface for canvas objects that want a cursor change (on hover) (`fyne.Cursable` wink) with the method `HoverCursor() fyne.Cursor` which returns a Fyne constant which can be translated to the according glfw constant by the internal package
> 
> 
> I like idea 2 most.
> Any better ideas?

You're right. I've thought of this already. If crafted well, it could help solve #726 .

internal/driver/glfw/window.go
spatocode: The disabled state was already ignored before now. Could it be better to exempt the disabled state? I think it makes sense.

spatocode:
I will begin work on refactoring this pull request inline with the discussion and suggestion by @toaster .

internal/driver/glfw/window.go
andydotxyz: I'm not sure what you mean - the test above checks the disabled state of Entry - it only sets the bar cursor if the entry is not disabled.
The same should surely hold true for the selectEntry?
It seems fairly standard to not indicate an edit cursor over items that cannot be edited.

internal/driver/glfw/window_test.go
```diff
@@ -59,6 +59,34 @@ func TestGLDriver_CreateSplashWindow(t *testing.T) {
...
 
+func cursorForObject(obj fyne.CanvasObject) *glfw.Cursor {
+	if wid, ok := obj.(*widget.Entry); ok {
+		if !wid.Disabled() {
+			return entryCursor
+		}
+	} else if _, ok := obj.(*widget.Hyperlink); ok {
+		return hyperlinkCursor
+	} else if _, ok := obj.(*widget.SelectEntry); ok {
+		return entryCursor
```
andydotxyz: I think this is a perfect spot to check disabled state like entry above :)

internal/driver/glfw/window_test.go
```diff
@@ -59,6 +59,34 @@ func TestGLDriver_CreateSplashWindow(t *testing.T) {
...
+
+	assert.Same(t, defaultCursor, cursorForObject(widget.NewLabel("")))
+
+	entry := widget.NewEntry()
+	assert.Same(t, entryCursor, cursorForObject(entry))
+	entry.Disable()
+	assert.Same(t, defaultCursor, cursorForObject(entry))
+
+	selectEntry := widget.NewSelectEntry([]string{"A", "B", "C"})
+	assert.Same(t, entryCursor, cursorForObject(selectEntry))
```
andydotxyz: I think this is a perfect spot to check disabled state like entry above :)

canvasobject.go
```diff
@@ -79,3 +79,9 @@ type Focusable interface {
 type Shortcutable interface {
 	TypedShortcut(Shortcut)
 }
+
+// Cursorable describes any CanvasObject that needs a cursor change
+type Cursorable interface {
+	Cursor() Cursor
+	SetCursor(Cursor)
```
toaster: I'm with @andydotxyz here: don't add `SetCursor()`. I don't think there is a real use case. IMO the cursor depends on the object's purpose and the purpose of an object does not change without the object being changed.

widget/widget.go
```diff
@@ -12,6 +12,7 @@ type BaseWidget struct {
 	size     fyne.Size
 	position fyne.Position
 	Hidden   bool
+	cursor	 fyne.Cursor
```
toaster: I don't think that every widget should implement `fyne.Cursor`. Only the entries and links should, for now. Every non-cursorable object uses the default cursor. This saves a lot of redundant tests.

internal/driver/glfw/window.go
```diff
@@ -502,17 +494,31 @@ func (w *window) findObjectAtPositionMatching(canvas *glCanvas, mouse fyne.Posit
...
 	obj, pos := w.findObjectAtPositionMatching(w.canvas, w.mousePos, func(object fyne.CanvasObject) bool {
-		entry, entOk := object.(*widget.Entry)
-		_, selOk := object.(*widget.SelectEntry)
-		if (entOk && !entry.Disabled()) || selOk {
-			cursor = entryCursor
-		} else if _, ok := object.(*widget.Hyperlink); ok {
-			cursor = hyperlinkCursor
+		if cursorable, ok := object.(fyne.Cursorable); ok {
+			fyneCursor := cursorable.Cursor()
+			cursor = cursorToName(fyneCursor)
```
toaster: AFAICS this creates a new glfw default cursor on every call. I'm not sure if this is save with glfw memory usage (might leak) and at least new Go objects are allocated every time.
I propose to return `initCursors` and create a map for the (immutable) default cursors.

internal/driver/glfw/window_test.go
```diff
@@ -59,32 +59,26 @@ func TestGLDriver_CreateSplashWindow(t *testing.T) {
...
-	assert.Same(t, entryCursor, cursorForObject(entry))
+	assert.Equal(t, fyne.TextCursor, cursorForObject(entry))
 	entry.Disable()
-	assert.Same(t, defaultCursor, cursorForObject(entry))
+	assert.Equal(t, fyne.TextCursor, cursorForObject(entry))
 
 	selectEntry := widget.NewSelectEntry([]string{"A", "B", "C"})
-	assert.Same(t, entryCursor, cursorForObject(selectEntry))
+	assert.Equal(t, fyne.TextCursor, cursorForObject(selectEntry))
 }
```
toaster: This test does not really test anything on glfw.window. Instead it tests the `Cursor()` implementation of a couple of widgets.
The `Cursor()` implementation should be tested in the object specific tests (like entry_test and so on).
The window test should test the following cases:
* applied cursor for a non-cursorable object
* applied cursor for a `fyne.Cursorable` object which is _not_ `fyne.Disableable`(I suggest a minimal test object instead of an existing wigdet to not introduce a dependency from `widget` for this)
* applied cursor for a cursorable object which _is_ `fyne.Disableable` and enabled (again use a minimal test object)
* applied cursor for a disabled cursorable object

toaster:
N.B.: Please ensure to run the glfw tests locally. They aren't part of the CI run, yet.

andydotxyz:
You might find our suggested git hooks useful, to check formatting and lint before commit: https://github.com/fyne-io/fyne/wiki/GitHooks

internal/driver/glfw/window_test.go
spatocode: Am not sure i understood your suggestion but i hope the new test works fine?

internal/driver/glfw/window.go
```diff
@@ -494,31 +504,24 @@ func (w *window) findObjectAtPositionMatching(canvas *glCanvas, mouse fyne.Posit
...
+				cursor = textCursor
+			} else if fyneCursor == desktop.PointerCursor {
+				cursor = pointerCursor
+			} else if fyneCursor == desktop.CrosshairCursor {
+				cursor = crosshairCursor
+			} else if fyneCursor == desktop.HResizeCursor {
+				cursor = hResizeCursor
+			} else if fyneCursor == desktop.VResizeCursor {
+				cursor = vResizeCursor
+			}
```
toaster: Why didn't you keep a (constant) map which is initialised in `initCursors()`? This would reduce the amount of global vars and also simplify this code. Map lookup might be slower but I don't believe that this would be an issue.

internal/driver/glfw/window_test.go
```diff
@@ -59,26 +60,32 @@ func TestGLDriver_CreateSplashWindow(t *testing.T) {
...
-	entry := widget.NewEntry()
-	assert.Equal(t, fyne.TextCursor, cursorForObject(entry))
-	entry.Disable()
-	assert.Equal(t, fyne.TextCursor, cursorForObject(entry))
-
-	selectEntry := widget.NewSelectEntry([]string{"A", "B", "C"})
-	assert.Equal(t, fyne.TextCursor, cursorForObject(selectEntry))
+	assert.Same(t, textCursor, cursorForObject(widget.NewEntry()))
+	assert.Same(t, pointerCursor, cursorForObject(widget.NewHyperlink("Testing", u)))
+	assert.Same(t, defaultCursor, cursorForObject(widget.NewButton("Test", nil)))
```
toaster: You are still testing that the widgets have a `Cursor()` implementation and which cursor they want.
This should be tested in the widgets' tests!

Here you should test that mouse moving over a `CanvasObject` which implements `Cursorable` changes the actual cursor.

If you want we can do a Google Meet session to talk about this and develop the tests together. Just contact me on Slack for that.

internal/driver/glfw/window.go
```diff
@@ -504,24 +504,29 @@ func (w *window) findObjectAtPositionMatching(canvas *glCanvas, mouse fyne.Posit
 	return driver.FindObjectAtPositionMatching(mouse, matches, canvas.Overlays().Top(), roots...)
 }
 
+func cursorToName(cursor desktop.Cursor) *glfw.Cursor {
+	var cursorMap = map[desktop.Cursor]*glfw.Cursor{
```
toaster: Doesn't this create a new map on every call?

spatocode: Yes, it does but that shouldn't be an issue. We need to reduce the amount of global variables.

toaster: Why?

widget/select_entry_test.go
```diff
@@ -1,4 +1,4 @@
-package widget_test
+package widget
```
toaster: I'm very biased regarding this. Why did you move the tests into the `widget` package?

widget/select_entry.go
```diff
@@ -57,3 +58,19 @@ func (e *SelectEntry) SetOptions(options []string) {
 	dropDownButton.SetIcon(theme.MenuDropDownIcon())
 	e.ActionItem = dropDownButton
 }
+
+type cursorableButton struct {
+	Button
+}
+
```
toaster: Why not implement `Cursorable` in `Button` instead?

internal/driver/glfw/window_test.go
```diff
@@ -60,32 +60,23 @@ func TestGLDriver_CreateSplashWindow(t *testing.T) {
...
+	assert.Same(t, textCursor, w.cursor)
+
+	w.mouseMoved(w.viewport, 10, float64(h.Position().Y+10))
+	assert.Same(t, pointerCursor, w.cursor)
 
-	assert.Same(t, textCursor, cursorForObject(widget.NewEntry()))
-	assert.Same(t, pointerCursor, cursorForObject(widget.NewHyperlink("Testing", u)))
-	assert.Same(t, defaultCursor, cursorForObject(widget.NewButton("Test", nil)))
+	w.mouseMoved(w.viewport, 10, float64(b.Position().Y+10))
+	assert.Same(t, defaultCursor, w.cursor)
```
toaster: 👍

internal/driver/glfw/window.go
andydotxyz: Yes, this could be called a lot and I don't see the harm in a single map global to the package that is set up in init()

widget/select_entry_test.go
spatocode: All widgets and corresponding tests in this package has the same package name `widget`. So i don't see any reason of not maintaining the the same package name throughout all widget test. Moreover, it's better calling functions from the same package than making redundant imports. 

For the special cursorable button, we already discussed this on slack. SelectEntry dropdown button cursor is overriden by it's parent's (SelectEntry) cursor. This solves that issue.

widget/select_entry.go
spatocode: IMHO, I don't think that's the best thing to do. By default, any non-cursorable widget has `fyne.DefaultCursor`. I see no need of allowing `Button` implement `Cursorable` for the sake of `SelectEntry`. If button implement `Cursorable`, it's cursor will still remain `fyne.DefaultCursor`. Does it really make sense?

toaster: > Does it really make sense?

IMO yes. The implementation of `Cursorable` means that a component has a specific wish regarding the cursor. Any component that does not implement `Cursorable` simply doesn't care about the cursor.

toaster: Plus, I see no need to implement a special button just for the sake of the cursor in the `SelectEntry` :).

toaster: The drop-down button should behave like every other button. That's why it should not be a special button.

widget/select_entry_test.go
toaster: I consider testing a package via its exported API the better approach. The only exception from this are unit tests for complex internal methods. But these are rare in my experience.

So moving tests into the tested package just to be able to implement an unnecessary internal struct is a bad idea.

internal/driver/glfw/window.go
```diff
@@ -29,22 +28,26 @@ const (
 )
 
 var (
-	defaultCursor, entryCursor, hyperlinkCursor *glfw.Cursor
-	initOnce                                    = &sync.Once{}
-	defaultTitle                                = "Fyne Application"
+	defaultCursor, textCursor, pointerCursor, crosshairCursor, hResizeCursor, vResizeCursor *glfw.Cursor
```
andydotxyz: can you move this to two lines please - otherwise it makes the following aligned items harder to read

widget/select_entry.go
andydotxyz: I have thought this over more and agree with @toaster - all buttons should have the same cursor, so let's just put this on widget.Button :)

stuartmscott:
Excited to see the progress here, SplitContainer (#789) could use this to set the cursor to HResize/VResize when the user hovers over the divider.

driver/desktop/cursor.go
```diff
@@ -0,0 +1,24 @@
+package desktop
+
+// Cursor represents a standard fyne cursor
+type Cursor string
+
+const (
+	// DefaultCursor is the default cursor typically an arrow
+	DefaultCursor Cursor = "DefaultCursor"
```
andydotxyz, line 8: It just occurred to me also that this would probably be more efficient as an enum... using string comparison for an internal-only lookup seems like we're burning cycles for no benefit.

internal/driver/glfw/window.go
andydotxyz: actually this probably doesn't matter, feel free to ignore that preference

internal/driver/glfw/window.go
andydotxyz: Also I have noticed this method name is strange. Maybe it should be "nativeCursor" or "lookupNativeCursor"?

andydotxyz:
I'd love to see this land - if possible in time for the 1.2.4 release I am about to start preparing. It would enable fixing issues that developers have with custom widgets not getting the right cursor.
Can you let us know if you'll be able to address the comments @spatocode?
Thanks for the work and sorry if this puts on pressure, really I just want to know for planning purposes.

andydotxyz:
Great work thanks, I'm happy - how about @toaster ?

toaster:
Awesome!

PR #751, opened 2020-03-15
Return a Dialog from ShowCustom() and ShowCustomConfirm()

fiam:
This allows the caller to close the dialog programatically in response
to other events. Also, refactor dialog.Hide() and the OnTapped handlers
for closing the dialogs to simplify the code a bit.

This also fixes a goroutine leak in dialog.wait().

### Checklist:

- [X] Tests included
- [X] Tests all pass.

### DIFF
--- a/dialog/base.go
+++ b/dialog/base.go
@@ -3,6 +3,7 @@ package dialog // import "fyne.io/fyne/dialog"
 
 import (
 	"image/color"
+	"sync"
 
 	"fyne.io/fyne"
 	"fyne.io/fyne/canvas"
@@ -37,19 +38,33 @@ type dialog struct {
 	dismiss        *widget.Button
 
 	response  chan bool
+	mu        sync.Mutex
 	responded bool
 	parent    fyne.Window
 }
 
 func (d *dialog) wait() {
-	select {
-	case response := <-d.response:
-		d.responded = true
-		d.win.Hide()
-		if d.callback != nil {
-			d.callback(response)
-		}
+	d.mu.Lock()
+	if d.response != nil {
+		// Already waiting
+		d.mu.Unlock()
+		return
 	}
+	d.response = make(chan bool)
+	d.mu.Unlock()
+	go func() {
+		select {
+		case response := <-d.response:
+			d.responded = true
+			d.win.Hide()
+			if d.callback != nil {
+				d.callback(response)
+			}
+		}
+
+		close(d.response)
+		d.response = nil
+	}()
 }
 
 func (d *dialog) setButtons(buttons fyne.CanvasObject) {
@@ -120,7 +135,6 @@ func (d *dialog) applyTheme() {
 func newDialog(title, message string, icon fyne.Resource, callback func(bool), parent fyne.Window) *dialog {
 	d := &dialog{content: newLabel(message), title: title, icon: icon, parent: parent}
 
-	d.response = make(chan bool, 1)
 	d.callback = callback
 
 	return d
@@ -141,16 +155,16 @@ func newButtonList(buttons ...*widget.Button) fyne.CanvasObject {
 }
 
 func (d *dialog) Show() {
-	go d.wait()
+	d.wait()
 	d.win.Show()
 }
 
 func (d *dialog) Hide() {
-	d.win.Hide()
-
-	if !d.responded && d.callback != nil {
-		d.callback(false)
+	if d.response == nil {
+		// Already hidden
+		return
 	}
+	d.response <- false
 }
 
 // SetDismissText allows custom text to be set in the confirmation button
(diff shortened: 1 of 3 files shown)


andydotxyz:
Oh yes we did miss this.
The model used on other dialogs is that a "New___()" would return the dialog but "Show___()" does not, perhaps it would be better to match that format?

andydotxyz:
It is unusual to see a single test failure in the run (1/4) so I restarted it.
Unfortunately the dialog tests are hanging on 1 configuration so maybe there is something wrong after all.
I suspect the refactoring may hit an issue with the channel on Go 1.12?

fiam:
@andydotxyz Just noticed the problem: it's the busy wait loop. Go before < 1.14 mostly does preemption only at function call sites (except the usual points that would cause a grouting to block). Lets' see if it passes CI with these changes and we'll go from there.

andydotxyz:
It works again, but as there is more code than before I’m not sure that it meets your goal of “ simplify the code a bit”, sorry.
Why not just limit this to the original intent of providing access to the Dialog type?

fiam:
@andydotxyz There's a goroutine leak in the original code and I took the opportunity to fix it and make `Show()/Hide()` concurrency safe. I had amended the commit message, but I forgot to edit the PR message.

Regarding the API changes, I've been looking at the rest of the dialog related functions and it seems to me that adding new functions with the `New___()` prefix would just increase API surface without too much benefit. What about making all `Show___()` functions return the Dialog? Changing the signature to return a value would be a very small incompatible change. The only affected case I can think if would be passing any of the `Show__()` functions around to other functions, but I'd expect that be a very niche use case.

andydotxyz:
Hmm. It’s great that the code changes fix a goroutine leak - but I don’t think arranging it so the method called “wait” no longer waits...

As for the dialog returns, I would like to get other developers feedback. We initially followed a pattern inspired by the have JOptionPane where create...() returns a dialog but show...() does not... https://docs.oracle.com/javase/8/docs/api/javax/swing/JOptionPane.html

toaster:
Hi @fiam, I've started to review this PR (by reading the current discussion) and am wondering: did you force push twice _after_ the PR has got some review?

If so, I would kindly ask you to _not_ do this.

Force pushes in branches that are being reviewed make it harder for the reviewer to keep track of the changes. Sometimes even simple PRs collect a long history of back-and-forth commits while feedback is implemented. That's totally okay. There is no reason to be ashamed of making mistakes and correcting them. There is no reason to hide this from the git history. But it might be disappointing for the reviewer if changes are force pushed and he has to start from the beginning.

dialog/base.go
```diff
@@ -37,19 +38,33 @@ type dialog struct {
 	dismiss        *widget.Button
 
 	response  chan bool
+	mu        sync.Mutex
 	responded bool
 	parent    fyne.Window
 }
 
 func (d *dialog) wait() {
-	select {
```
toaster: AFAICS the changes regarding the goroutine leak are independent from the change to the `ShowCustom*` functions.
Please separate different topics into different commits in future contributions, which we greatly appreciate.

dialog/base.go
```diff
@@ -141,16 +155,16 @@ func newButtonList(buttons ...*widget.Button) fyne.CanvasObject {
 }
 
 func (d *dialog) Show() {
-	go d.wait()
+	d.wait()
```
toaster: `wait` sounds like a synchronous call but now it runs a goroutine. Maybe it should be renamed to reflect this?

fiam:
@toaster I force pushed once to fix the deadlock on Go 1.12 on Linux, but no one had written any code reviews at that point. However, Travis failed to launch the macOS builders and I had to force push twice then to make it restart the build (just changed the commit date, no code changes).

I agree that fixing the goroutine leak is probably a PR on its own. It was just that it came up while I was refactoring the code to replace those in place functions with a calls to `Hide()`, so I ended adding it up to the initial commit.

Regarding the API changes, if I were to write this API from the scratch I would have skipped the `Show__()` methods entirely, they're not worth the increase in API surface (note that with your current design you need 2x functions per dialog kind/conveniency function). However, the API right now has several `Show__()` methods that have no `New__()` equivalent, hence there's no way to e.g. show a custom dialog and get a pointer to it since the `dialog` type is private or to show a dialog with a warning icon without passing by `ShowError()`.

The `Dialog` interface is also problematic as it stands, since it contains a `SetDismissText()` method that does not apply to all dialogs (the most obvious example is a dialog for a non cancellable long running operation that can't be dismissed by the user). 

After reviewing the whole `dialog` package, I think adapting it to suit my needs (hiding custom dialogs manually, update any dialog element dynamically) while avoiding even the most minimal breaking change is just too much work, or likely just not possible. I'm gonna proceed by using an internal copy of the package.

fiam:
Fix is in #754, closing this one.

andydotxyz:
I would have thought that `dialog.NewCustom(...) fyne.Dialog` would have worked and not been a breaking change as it's a whole new call that internally the old Show version could utilise?

andydotxyz:
I have been considering whether the dialogs should be returning their own types.
If Dialog were a struct instead, then it could have the Show() and Hide() as well as things like a Title string and OnDismiss func() field and those that have dismiss text could have a mixin for SetDismissedText, for example?

fiam:
@andydotxyz I think interfaces are fine for this use case. In fact, I think exposing the dialogs as interfaces is better than concrete types. This way the user can have generic methods that handle dialogs.

andydotxyz:
Yes, the concrete types would still implement the basic dialog interface - but it would allow adding fields and methods to individual dialogs which is not currently possible...

PR #754, opened 2020-03-18
Fix goutine leak in dialog's Show()/Hide()

fiam:
### Description:

Instead of spawning a blocked goroutine in Show(), just hide the popup and call the callback directly from Hide().

Make all dialog button callbacks use Hide() or hideWithResp(), to encapsulate the
logic better.

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

### DIFF
--- a/dialog/base.go
+++ b/dialog/base.go
@@ -189,9 +189,7 @@ func ShowCustom(title, dismiss string, content fyne.CanvasObject, parent fyne.Wi
 	d := &dialog{content: content, title: title, icon: nil, parent: parent}
 
 	d.dismiss = &widget.Button{Text: dismiss,
-		OnTapped: func() {
-			d.response <- false
-		},
+		OnTapped: d.Hide,
 	}
 	d.setButtons(widget.NewHBox(layout.NewSpacer(), d.dismiss, layout.NewSpacer()))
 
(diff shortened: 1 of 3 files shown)


dialog/base.go
```diff
@@ -183,18 +203,13 @@ func ShowCustom(title, dismiss string, content fyne.CanvasObject, parent fyne.Wi
...
 
 	d.dismiss = &widget.Button{Text: dismiss, Icon: theme.CancelIcon(),
-		OnTapped: func() {
-			d.response <- false
-		},
+		OnTapped: d.Hide,
 	}
 	ok := &widget.Button{Text: confirm, Icon: theme.ConfirmIcon(), Style: widget.PrimaryButton,
-		OnTapped: func() {
-			d.response <- true
```
andydotxyz: This is surely broken? It will now return false instead of true

dialog/base.go
```diff
@@ -159,17 +174,22 @@ func (d *dialog) SetDismissText(label string) {
 	widget.Refresh(d.win)
 }
 
+func (d *dialog) hideWithResp(resp bool) {
```
andydotxyz: We don't usually truncate function names - surely hideWithResponse is not too verbose?

dialog/base.go
```diff
@@ -159,17 +174,22 @@ func (d *dialog) SetDismissText(label string) {
...
+	if d.response == nil {
+		// Already hidden
+		return
+	}
+	d.response <- resp
```
andydotxyz: Everywhere else `d.response` is guarded by `mu`, should it not be here as well?

dialog/base.go
```diff
@@ -37,19 +38,38 @@ type dialog struct {
...
 	response  chan bool
+	mu        sync.Mutex
 	responded bool
 	parent    fyne.Window
 }
 
+func (d *dialog) setupWait() {
+	d.mu.Lock()
+	if d.response != nil {
+		// Already shown
```
andydotxyz: A comment in code probably indicates it could be clearer. How about a function `isShown` that checks (with a lock) the value of `d.response`? Calling that would remove the need for this, and the later (inverted) case, comments.

andydotxyz:
Wow, what a simplification, thanks!

andydotxyz:
Thanks!
I have merged this to master as it's almost a hotfix. For larger items or anything with API changes please use the develop branch.

fiam:
@andydotxyz Thanks for the merge! I hadn't noticed you do use a develop branch. I'll submit my PRs against it in the future.

PR #755, opened 2020-03-18
Add support for -appVersion and -buildNumber in fyne package

fiam:
Add support for -appVersion and -buildNumber in fyne package
    
For macOS, -appVersion is mapped to CFBundleVersionShort while
-buildNumer is used for CFBundleVersion. If no CFBundleVersion
is provided, it's set to the same value as CFBundleVersionShort.
   
For Windows, -appVersion is expanded to 3 components, then
-buildNumber (which must have 1 component at most) is appended
to it to form the 4 component assembly version that Windows
expects. If no -buildNumber is provided, we append a zero to
get to 4 components.
    
No other platforms use these 2 new flags for now.

### Description:

These changes allow specifying the app version in macOS and Windows. This is important on macOS where users expect an "About" dialog that shows the app version. It also makes automatic updates a bit easier.

At some point, this should probably be extended to support iOS and Android too, since at least the App Store does enforce increasing version numbers on the bundle version (not sure about Android, it's been a long time since I've written software for it).

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne/package.go
+++ b/cmd/fyne/package.go
@@ -12,6 +12,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 
 	"fyne.io/fyne"
@@ -22,6 +23,10 @@ import (
 	"github.com/pkg/errors"
 	"golang.org/x/mod/modfile"
 	"golang.org/x/mod/module"
+)
+
+const (
+	defaultAppVersion = "1.0"
 )
 
 func exists(path string) bool {
@@ -83,13 +88,59 @@ func runAsAdminWindows(args ...string) error {
 	return exec.Command("powershell.exe", "Start-Process", "cmd.exe", "-Verb", "runAs", "-ArgumentList", cmd).Run()
 }
 
+func parseNumericVersion(vers string) ([]int, error) {
+	parts := strings.Split(vers, ".")
+	if len(parts) == 0 {
+		return nil, errors.New("version is empty")
+	}
+	numbers := make([]int, len(parts))
+	for ii, v := range parts {
+		n, err := strconv.Atoi(v)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing component %q in version %q: %v", v, vers, err)
+		}
+		numbers[ii] = n
+	}
+	return numbers, nil
+}
+
+// buildNumericComponentsVersion allows creating a version identifier
+// based on dot separated numeric components with the number of in
+// the given [min, max] range. If the version has less than minimum
+// required components, zeroes are appended at the end. If the version
+// has more components than the maximum, an error is returned.
+func buildNumericVersion(vers string, min, max int) (string, error) {
+	numbers, err := parseNumericVersion(vers)
+	if err != nil {
+		return "", err
+	}
+	for len(numbers) < min {
+		numbers = append(numbers, 0)
+	}
+	if len(numbers) > max {
+		return "", fmt.Errorf("version %q has %d components but the maximum is %d", vers, len(numbers), max)
+	}
+	parts := make([]string, len(numbers))
+	for ii, n := range numbers {
+		parts[ii] = strconv.Itoa(n)
+	}
+	return strings.Join(parts, "."), nil
+}
+
 // Declare conformity to command interface
 var _ command = (*packager)(nil)
 
 type packager struct {
-	name, srcDir, dir, exe, icon string
-	os, appID                    string
-	install, release             bool
+	name, srcDir, dir, exe, icon       string
+	os, appID, appVersion, buildNumber string
+	install, release                   bool
+}
+
+func (p *packager) getAppVersion() string {
+	if p.appVersion == "" {
+		return defaultAppVersion
+	}
+	return p.appVersion
 }
 
 func (p *packager) packageLinux() error {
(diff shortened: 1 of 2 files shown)


andydotxyz:
I like the idea of being able to specify these on the commandline, but not so sure about auto-changing them. Trying to be too clever hear could backfire if metadata does not contain what developers specify.

Also note that macOS docs say the version can be 1,2 or 3 integers (and also that extra ones will be ignored) so there is no need to append 0s there. Microsoft seems to be more strict here, but I don’t know if it’s enforced.

I think you’d get away with using the appVersion for both fields in Mac manifest, but not certain.

fiam:
@andydotxyz `CFBundleShortVersionString` is documented as requiring 3 components. The fact that on macOS it works with 2 components right now is a implementation detail that might change at any point in the future. Only `CFBundleVersion` is allowed to use 1 to 3 components. I could change `buildNumericVersion()` to accept min/max arguments instead of a fixed count, but it felt a bit overkill. Note that `103.0.0` should be interpreted as equal to `103` by tools/libraries doing version comparisons, so the extra zeroes should be fine. 

Also, specifying the same value for both `CFBundleShortVersionString` and `CFBundleVersion` is perfectly fine. If the developer wants to use CFBundleVersion as a build number they can run `fyne package -appVersion 42 -appShortVersion 1.2.3`.

The underlying problem is that `CFBundleShortVersionString` and `CFBundleVersion` are badly named. The former is what the user sees and is displayed in system provided widgets while the latter is just an internal numeric identifier (note, however, that some 3rd party libraries and tools will erroneously use them interchangeably). I understand that for someone without macOS experience, mapping `-appVersion` to what's essentially and build number and `-appShortVersion` to the user-displayed version might feel strange. However, if you map `-appVersion` to `CFBundleShortVersionString` and e.g. `-appBuildNumber` to `CFBundleVersion` then you'd have to document how they map to macOS bundle keys. With the naming in this PR anyone with macOS experience will recognise what `-appVersion` and `-appShortVersion` map to.

andydotxyz:
Thanks for your explanation, yes I see now that one of the variables is stated as x.y.z whilst the other is not. In my experience with the App Store Connect I don’t think it interprets “10.0” as being equal to “10.0.0”.

I agree that Mac has really bad naming here, which is one reason I don’t think we should replicate it at all. I absolutely disagree that we should expect people using “fyne package” to have any macOS development experience at all - we are aiming for a platform agnostic system here so cannot rely on prior knowledge.

I noticed that Microsoft lists the 4th digit as “build” so how about this for a proposal:

* -appVersion is a 3 digit string, 1.2.3, used for Mac’s short version
* -buildNumber (or similar) is a single integer that is used for the Mac version number (which as you say is often the build id).
* the windows version is then made up of <appVersion>.<buildNumber>

With this proposal we have semantic naming that can be adapted to other platforms. And to avoid mutating people’s version numbers the semantics / format described could just be enforced during parse.

fiam:
@andydotxyz At the end of the day, I think there's no perfect solution because the numbers have different semantics. For example, on Windows you probably want to reset the build number (4th component) to 0 with any patch number (3rd component) increase while on macOS you probably want to increase the build number (CFBundleVersion) on any published build regardless of the version number.

I think the solution you proposed is an acceptable compromise, so I'll change this PR to implement those semantics. I'm gonna implement `-buildNumber` as a string, since we want different defaults for macOS and Windows (1 vs 0).

fiam:
Fixed conflicts by rebasing and pushed it again

andydotxyz:
> we probably should find a way to be less liberal about changing app identifiers if it can be avoided

This point still bothers me - is the magic to rebuild version numbers required?
If I specify a version number it should either be acceptable, or not - if you rewrite it then there is a decent chance it won't match the metadata I set in the store. And it has to (for mac uploads at least).

I'm not sure why having different default ints (0 vs 1) means that a value should be a string?...

fiam:
> > we probably should find a way to be less liberal about changing app identifiers if it can be avoided
> 
> This point still bothers me - is the magic to rebuild version numbers required?

I've changed `CFBundleVersion` to accept 1, 2 or 3 components. `CFBundleVersionShort` is expanded to 3 components as required. Windows assembly version must have 4 components and is built as you requested by expanding the version to 3 components, then appending the build number.

> If I specify a version number it should either be acceptable, or not - if you rewrite it then there is a decent chance it won't match the metadata I set in the store. And it has to (for mac uploads at least).

You need to make up your mind on this. You made me change the PR to make it more platform agnostic, but it can't be made platform agnostic without massaging the version numbers because each platform has different version number semantics and/or requirements. Do you expect users to define different version numbers for each platform? That would be very unpleasant to work with and totally defeats the point of having platform agnostic flags for setting the version numbers.

> 
> I'm not sure why having different default ints (0 vs 1) means that a value should be a string?...

Because otherwise what are you gonna set as default, `-1`? It will show up in `fyne package -h` and it will be confusing. An empty string doesn't show any default values when you invoke the help.

andydotxyz:
Apologies that I appear to have been indecisive - that is not my intention at all. You are quite right that platform agnostic at the core of every decision we are making. To be able to do this sometimes the system needs to be "more opinionated" to ensure we can satisfy all platform requirements.

The code in this PR offers a lot of flexibility to account for the differing requirements of platforms, I think this could be made more simple in one of two possible ways:

* use the parse code just to validate that the number is a valid version (x, x.y, x.y.z) and that the build is also valid (x), with this validated then the only possible operation required is appending ".0" in the cases where there are not enough digits for a specific platform.
* require that version numbers are always x.y.z and then there is no version change code required at all (except appending build to version for windows).

I see what you mean now about the build parameter - I had not thought about the command line default readout. However this means that the user will have to think about platforms and what build number to use. Could we instead specify that "0" is not a valid build version with the downside that "1.0.0.0" simply will not be possible for windows? It may not be ideal but I think it is truly platform agnostic to the user.

Sorry for the frustrations, this is certainly not an easy task!

andydotxyz:
Do the above compromises seem sensible or would you rather not continue this PR @fiam ?

andydotxyz:
This has been replaced by #1429, thank you for your contribution and im sorry we didn't get this verison landed.
I hope that the replacement addresses the use-case you have too.

PR #757, opened 2020-03-21
Update modal popup to respect innerSize as well

andydotxyz:
### Description:
Update the modal layout to respect the popup innerSize changes.
(as it is centred it will ignore innerPosition).

Added Size.Min and Max, deprecated confusingly named Size.Union

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/geometry_test.go
+++ b/geometry_test.go
@@ -39,6 +39,32 @@ func TestSizeUnion(t *testing.T) {
 	assert.Equal(t, size3.Height, maxH)
 }
 
+func TestSizeMax(t *testing.T) {
+	size1 := NewSize(10, 100)
+	size2 := NewSize(100, 10)
+
+	size3 := size1.Max(size2)
+
+	maxW := Max(size1.Width, size2.Width)
+	maxH := Max(size1.Height, size2.Height)
+
+	assert.Equal(t, size3.Width, maxW)
+	assert.Equal(t, size3.Height, maxH)
+}
+
+func TestSizeMin(t *testing.T) {
+	size1 := NewSize(10, 100)
+	size2 := NewSize(100, 10)
+
+	size3 := size1.Min(size2)
+
+	minW := Min(size1.Width, size2.Width)
+	minH := Min(size1.Height, size2.Height)
+
+	assert.Equal(t, size3.Width, minW)
+	assert.Equal(t, size3.Height, minH)
+}
+
 func TestPosition_Add(t *testing.T) {
 	pos1 := NewPos(10, 10)
 	pos2 := NewPos(25, 25)
--- a/widget/popup.go
+++ b/widget/popup.go
@@ -5,7 +5,6 @@ import (
 
 	"fyne.io/fyne"
 	"fyne.io/fyne/canvas"
-	"fyne.io/fyne/layout"
 	"fyne.io/fyne/theme"
 )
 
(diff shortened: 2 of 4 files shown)


geometry_test.go
```diff
@@ -39,6 +39,32 @@ func TestSizeUnion(t *testing.T) {
...
+	size1 := NewSize(10, 100)
+	size2 := NewSize(100, 10)
+
+	size3 := size1.Max(size2)
+
+	maxW := Max(size1.Width, size2.Width)
+	maxH := Max(size1.Height, size2.Height)
+
+	assert.Equal(t, size3.Width, maxW)
+	assert.Equal(t, size3.Height, maxH)
```
toaster: I would prefer a literal expectation instead of computing such a trivial thing like the maximum of 10 and 100. It would make more obvious what the test expects.
Also the expectation comes first and the actual value second.
```suggestion
	assert.Equal(t, NewSize(100, 100), size1.MaxSize(size2))
	assert.Equal(t, NewSize(100, 100), size2.MaxSize(size1))
```

geometry_test.go
```diff
@@ -39,6 +39,32 @@ func TestSizeUnion(t *testing.T) {
...
+	size1 := NewSize(10, 100)
+	size2 := NewSize(100, 10)
+
+	size3 := size1.Min(size2)
+
+	minW := Min(size1.Width, size2.Width)
+	minH := Min(size1.Height, size2.Height)
+
+	assert.Equal(t, size3.Width, minW)
+	assert.Equal(t, size3.Height, minH)
```
toaster: The same as above:
```suggestion
	assert.Equal(t, NewSize(10, 10), size1.Min(size2))
	assert.Equal(t, NewSize(10, 10). size2.Min(size1))
```

widget/popup_test.go
```diff
@@ -201,3 +201,22 @@ func TestModalPopUp_TappedSecondary(t *testing.T) {
...
+	win.Resize(fyne.NewSize(80, 80))
+	pop := NewModalPopUp(label, win.Canvas())
+	defer test.Canvas().Overlays().Remove(pop)
+
+	size := fyne.NewSize(70, 50)
+	pop.Resize(size)
+
+	innerSize := pop.Content.Size()
+	assert.Equal(t, size.Width-theme.Padding()*2, innerSize.Width)
+	assert.Equal(t, size.Height-theme.Padding()*2, innerSize.Height)
```
toaster: You should add an assertion to check if the size was different before.

widget/popup.go
```diff
@@ -181,17 +179,20 @@ func (r *popUpRenderer) Destroy() {
...
 	popUp   *PopUp
 	bg      *canvas.Rectangle
 	objects []fyne.CanvasObject
 }
 
-func (r *modalPopUpRenderer) Layout(size fyne.Size) {
-	r.center.Layout(r.objects, size)
+func (r *modalPopUpRenderer) Layout(canvasSize fyne.Size) {
+	size := r.popUp.innerSize.Subtract(fyne.NewSize(theme.Padding()*2, theme.Padding()*2))
+	size = size.Min(canvasSize).Max(r.popUp.MinSize())
```
toaster: This one is hard to understand.
We want the content's size to be at least `r.popUp.MinSize`.
If `size` is bigger, we want it to be `size`.
But if `size` is bigger than the canvas, we want the content's size to be the canvas' size.
Maybe splitting the line and change the order helps the reader to understand what happens here?

```suggestion
	requestedSize := r.popUp.innerSize.Subtract(fyne.NewSize(theme.Padding()*2, theme.Padding()*2))
	size = r.popUp.MinSize().Max(requestedSize)
	size = size.Min(canvasSize)
```

Please not how I changed the order of the operations to the order of the expectations I wrote above.

N.B.: There are no tests for these computations. Please add some.

andydotxyz:
> This commit are two things and should have been split: first introduce `Min()`& `Max()` and deprecate `Union()`, second change the pop-up.

Quite right, my bad. I will replay them separately on a different branch, with the fixes requested as well.

andydotxyz:
Replaced by #759 where I think all of these points have been addressed.

PR #758, opened 2020-03-21
Initial security fixes

### LINKED ISSUE #742: META: Fix security issues reported by gosec
### Describe the bug:
~~Running [gosec](https://github.com/securego/gosec) on the code base returns between 66 and 79 issues-- depending on if v.1.23 or the develop version is used at the time of checking. These issues should in general be fixed as soon as possible in the MEDIUM severity parts, but the LOW can probably be knocked of when the time is there to do it.~~

#### Update as of 2020-03-28:
After the merge of #758, and #801 we are now down to 20 potential security issues reported by [gosec](https://github.com/securego/gosec). All but two case of `unhandled errors` are left and it is mostly MEDIUM severity issues that are left to be tackled.

### Future work:
When all potential security issues have fixed, we should ideally make sure to run `gosec` on Travis to make sure that no more bad behaviour slips in unnoticed.

### To Reproduce:
1. Get gosec using `go get github.com/securego/gosec/cmd/gosec`.
2. Run gosec on the fyne code from the repo root using `gosec ./...`.
3. Look through the reported issues.

### Statistics of reported issues on `develop` as of https://github.com/fyne-io/fyne/commit/7df22f0c81f45ed8abefaa0b790aa4f835eb9c95

```
Summary:
   Files: 187
   Lines: 23788
   Nosec: 7
  Issues: 20
```

#### Security Report (zipped html file for easy viewing and sorting in the browser):
[security.zip](https://github.com/fyne-io/fyne/files/4438783/security.zip)

Jacalz:
### Description:
This is a big PR aimed at fixing a lot of potential issues that are being reported by gosec analysing. the issues fixed here are mostly related to error handling, but there are also four fixes for file permissions.

We are now down from 69 to 33 issues on develop.

Partial fix for #742 

### Checklist:

- [-] Tests included.
- [x] Tests all pass.

## Note to anyone reviewing this: 

Please look at the error handling and say if we need to return or not. There might be places where we don't need to return when an error occurs (would not be possible to exploit) or the other way around where it might be strictly necessary to actually return.

### DIFF
--- a/app/settings_desktop.go
+++ b/app/settings_desktop.go
@@ -41,7 +41,11 @@ func watchFile(path string, callback func()) *fsnotify.Watcher {
 	go func() {
 		for event := range watcher.Events {
 			if event.Op&fsnotify.Remove != 0 { // if it was deleted then watch again
-				watcher.Remove(path)
+				err := watcher.Remove(path)
+				if err != nil {
+					fyne.LogError("Failed to stop watching settings file", err)
+					return
+				}
 				watchFileAddTarget(watcher, path)
 			} else {
 				callback()
@@ -67,5 +71,9 @@ func (s *settings) stopWatching() {
 		return
 	}
 
-	s.watcher.(*fsnotify.Watcher).Close()
+	err := s.watcher.(*fsnotify.Watcher).Close()
+	if err != nil {
+		fyne.LogError("Failed to close watcher", err)
+		return
+	}
 }
--- a/app/settings_desktop.go
+++ b/app/settings_desktop.go
@@ -5,6 +5,7 @@ package app
 import (
 	"os"
 	"path/filepath"
+	"runtime"
 
 	"fyne.io/fyne"
 	"github.com/fsnotify/fsnotify"
@@ -42,7 +43,7 @@ func watchFile(path string, callback func()) *fsnotify.Watcher {
 		for event := range watcher.Events {
 			if event.Op&fsnotify.Remove != 0 { // if it was deleted then watch again
 				err := watcher.Remove(path)
-				if err != nil {
+				if err != nil && runtime.GOOS == "darwin" { // fsnotify returns false positives on Windows and mac (possibly BSD too), see https://github.com/fsnotify/fsnotify/issues/268
 					fyne.LogError("Failed to stop watching settings file", err)
 					return
 				}
--- a/app/settings_desktop.go
+++ b/app/settings_desktop.go
@@ -5,7 +5,6 @@ package app
 import (
 	"os"
 	"path/filepath"
-	"runtime"
 
 	"fyne.io/fyne"
 	"github.com/fsnotify/fsnotify"
@@ -42,11 +41,8 @@ func watchFile(path string, callback func()) *fsnotify.Watcher {
 	go func() {
 		for event := range watcher.Events {
 			if event.Op&fsnotify.Remove != 0 { // if it was deleted then watch again
-				err := watcher.Remove(path)
-				if err != nil && runtime.GOOS == "darwin" { // fsnotify returns false positives on Windows and mac (possibly BSD too), see https://github.com/fsnotify/fsnotify/issues/268
-					fyne.LogError("Failed to stop watching settings file", err)
-					return
-				}
+				watcher.Remove(path) // fsnotify returns false positives, see https://github.com/fsnotify/fsnotify/issues/268
+
 				watchFileAddTarget(watcher, path)
 			} else {
 				callback()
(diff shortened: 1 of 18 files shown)


app/settings_desktop.go
```diff
@@ -41,7 +41,11 @@ func watchFile(path string, callback func()) *fsnotify.Watcher {
 	go func() {
 		for event := range watcher.Events {
 			if event.Op&fsnotify.Remove != 0 { // if it was deleted then watch again
-				watcher.Remove(path)
+				err := watcher.Remove(path)
+				if err != nil {
+					fyne.LogError("Failed to stop watching settings file", err)
+					return
```
Jacalz: This is the reason that the app test is failing. Removing the `return` call seems to fix things, but I have a feeling that something might be wrong with the underlying code because it looks like we always get an error saying `Fyne error:  Failed to stop watching settings file2020/03/22 09:20:41   Cause: can't remove non-existent inotify watch`.

Jacalz:
A little weird side-note here, it looks like some tests really aren't reproducible from time to time. Yesterday I had the following error on my branch:
```
2020/03/21 20:55:13 Fyne error:  Failed to decode image for window icon
2020/03/21 20:55:13   Cause: image: unknown format
2020/03/21 20:55:13   At: /home/jacob/fyne/internal/driver/glfw/window.go:215
--- FAIL: TestWindow_Focus (0.07s)
    window_test.go:726: 
                Error Trace:    window_test.go:726
                Error:          Not equal: 
                                expected: "abcd"
                                actual  : "bcd"
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -abcd
                                +bcd
                Test:           TestWindow_Focus
FAIL
```
This morning I had the same error, so I decided to run tests against develop too. I got the same error there. I then opened a new terminal and decided to run tests again so I could have develop beside my branch and then none of them had that test failing...

app/settings_desktop.go
andydotxyz: I think you are right this return is not correct. If we can't stop watching a deleted file (entirely probable) we should return to trying again.
Not sure why this is only an issue on this PR though - the others are working as expected. Are you certain nothing has changed?

Jacalz: I think you might understand me wrong here. I added that return statement in my PR to handle the error and the reason that it doesn't cause errors in other ones is because of that.

What I tried saying was that it throws an error here even though I don't think it should and we can see it now because we check the error. :wink:

andydotxyz: Oh, OK. How do you propose fixing this then? We can't merge with a new break ;)

Jacalz: I honestly have no idea that might be causing this. Do you want me to just log the error and not return or just not handle the error at all and leave it for another day?

andydotxyz: Let me know what you think after reading https://github.com/fsnotify/fsnotify/issues/40 and https://github.com/fsnotify/fsnotify/issues/238

Jacalz: The issue seems to be specific to Windows and Linux, perhaps we could only throw an error on operating systems that aren’t those?

Jacalz: Pushed new commits with the fix. Might need testing on MacOS, but tests are not failing for me anymore at least :)

Jacalz: Looks like MacOS tests on travis still fails. Looks like we always get an error no matter the case. I say that we ignore this for now and get back to it when it is fixed upstream.

andydotxyz:
Just restarting the second windows build to see what's going on. It may be that there's a legitimate failure here - can you see if it relates to the discussion above?

Jacalz:
> Just restarting the second windows build to see what's going on. It may be that there's a legitimate failure here - can you see if it relates to the discussion above?

Are you possibly commenting on the wrong PR? All the tests seem to be completing now :)

Jacalz:
I have done testing with the `cmd/fyne` tool without any new issues and I have done testing  the `fyne_demo` application and everything there seems to be working correct. I also tested it with the changes from #773 and on desktop that's no issues, but building with `-tags mobile` spits out this:
```go
2020/03/27 20:00:57 Fyne error:  Preferences load error:
2020/03/27 20:00:57   Cause: mkdir /data: permission denied
2020/03/27 20:00:57   At: /home/jacob/fyne/app/preferences.go:67
2020/03/27 20:01:00 Fyne error:  Failed on saving preferences
2020/03/27 20:01:00   Cause: mkdir /data: permission denied
2020/03/27 20:01:00   At: /home/jacob/fyne/app/preferences.go:96
```
Will probably need to look at this more, but it might not be entirely relevant because the preferences code does not work on mobile right?

andydotxyz:
Yeah you are probably right. If you could put that error log in a “preferences api not working on mobile” ticket we can reference that here and move on :).
I’d love to get mobile prefs into 1.2.4 in the next couple of weeks

Jacalz:
Here you go: https://github.com/fyne-io/fyne/issues/779 :)

PR #759, opened 2020-03-21
Update modal popup to respect innerSize

andydotxyz:
### Description:
Update the modal layout to respect the popup innerSize changes.
(as it is centred it will ignore innerPosition).

Added Size.Min and Max, deprecated confusingly named Size.Union.

Replaces #757 as it was too much in 1 commit.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/popup_test.go
+++ b/widget/popup_test.go
@@ -209,8 +209,8 @@ func TestModalPopUp_Resize(t *testing.T) {
 	pop := NewModalPopUp(label, win.Canvas())
 	defer win.Canvas().Overlays().Remove(pop)
 
-	assert.Less(t, pop.Content.Size().Width, 70)
-	assert.Less(t, pop.Content.Size().Height, 50)
+	assert.Less(t, pop.Content.Size().Width, 70-theme.Padding()*2)
+	assert.Less(t, pop.Content.Size().Height, 50-theme.Padding()*2)
 
 	pop.Resize(fyne.NewSize(70, 50))
 	assert.Equal(t, 70-theme.Padding()*2, pop.Content.Size().Width)
--- a/widget/popup_test.go
+++ b/widget/popup_test.go
@@ -227,22 +227,8 @@ func TestModalPopUp_Resize_Constrained(t *testing.T) {
 	defer win.Canvas().Overlays().Remove(pop)
 
 	pop.Resize(fyne.NewSize(90, 100))
-	assert.Equal(t, 80, pop.Content.Size().Width)
-	assert.Equal(t, 80, pop.Content.Size().Height)
+	assert.Equal(t, 80-theme.Padding()*2, pop.Content.Size().Width)
+	assert.Equal(t, 80-theme.Padding()*2, pop.Content.Size().Height)
 	assert.Equal(t, 80, pop.Size().Width)
 	assert.Equal(t, 80, pop.Size().Height)
-}
-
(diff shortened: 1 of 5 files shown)


andydotxyz:
This replaces the older, single commit, PR - apologies.
The build failure is an intermittent Travis issue on macOS - you can see that one did pass...

widget/popup_test.go
```diff
@@ -201,3 +201,48 @@ func TestModalPopUp_TappedSecondary(t *testing.T) {
...
+
+func TestModalPopUp_Resize(t *testing.T) {
+	label := NewLabel("Hi")
+	win := test.NewWindow(NewLabel("OK"))
+	win.Resize(fyne.NewSize(80, 80))
+	pop := NewModalPopUp(label, win.Canvas())
+	defer test.Canvas().Overlays().Remove(pop)
+
+	assert.Less(t, pop.Content.Size().Width, 70)
+	assert.Less(t, pop.Content.Size().Height, 50)
```
toaster: If `theme.Padding()` is greater than zero, that's true for the final content size, too.

widget/popup_test.go
```diff
@@ -201,3 +201,48 @@ func TestModalPopUp_TappedSecondary(t *testing.T) {
...
+	win := test.NewWindow(NewLabel("OK"))
+	win.Resize(fyne.NewSize(80, 80))
+	pop := NewModalPopUp(label, win.Canvas())
+	defer test.Canvas().Overlays().Remove(pop)
+
+	win.Resize(fyne.NewSize(30, 20))
+	assert.Greater(t, pop.Content.Size().Width, 30)
+	assert.Greater(t, pop.Content.Size().Height, 20)
+	assert.Greater(t, pop.Size().Width, 30)
+	assert.Greater(t, pop.Size().Height, 20)
```
toaster: We probably should assert the canvas size here, shouldn't we?

andydotxyz: I don't see why - canvas size is controlled by window, a whole different area of tests (driver...)

widget/popup_test.go
andydotxyz: OK. somewhat academic given that the window was 80x80 but I have fixed it anyhow :)

toaster: My comment has nothing to do with the window size.
I requested that the test should ensure that a size change was made. Setting the size to 70x50 changes the content's size to 70-<pad>x50-<pad>. Thus the resulting size is probably smaller than 70x50. Therefore a test that the previous size was smaller than 70x50 is not an indicator for a content size change.

widget/popup_test.go
toaster: Okay, my request was probably not the correct one.
Let's try this:
Every test of `popUp.Resize()` should assert the final `popUp.Size()` and the final `popUp.Content.Size()` with literal values (+/- `theme.Padding()`).

widget/popup_test.go
andydotxyz: I think I follow - did the updated test fix the issue?

toaster: Yes

andydotxyz:
OK, following discussion on Slack I have removed the minsize constrained check as overlays do not exhibit this behaviour.
Hopefully this can be merged now - let me know :)

andydotxyz:
Phew, Travis seems to be working again :)

PR #760, opened 2020-03-22
Support Empty Tabber

stuartmscott:
Previously tabcontainer would panic if rendered without any children

```
panic: runtime error: index out of range [0] with length 0

fyne.io/fyne/widget.(*tabContainerRenderer).Layout(0xc0001aadb0, 0x0, 0x6b)
	fyne.io/fyne/widget/tabcontainer.go:272 +0x781
```

### DIFF
--- a/widget/tabcontainer.go
+++ b/widget/tabcontainer.go
@@ -268,59 +268,60 @@ func (t *tabContainerRenderer) Layout(size fyne.Size) {
 		tabLocation = TabLocationBottom
 	}
 
+	tabBarMinSize := t.tabBar.MinSize()
+	var tabBarPos fyne.Position
+	var tabBarSize fyne.Size
+	var linePos fyne.Position
+	var lineSize fyne.Size
+	var childPos fyne.Position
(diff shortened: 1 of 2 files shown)


andydotxyz:
I’m not sure about this, is an empty tab bar a legitimate state?

If so is there much point laying out at all?
I agree it shouldn’t crash either way :)

stuartmscott:
I think an empty tab bar is a legitimate state - my use case is a text editor with a file tree and tabs for each open file, if there are no open files the tabber should not crash. It should be laid out so it takes up space, and rendered so it can draw its line.

stuartmscott:
Hey @toaster 

Added a test to ensure an empty TabContainer doesn't crash, and that CurrentTab, CurrentTabIndex, and MinSize return appropriate values.

Refactored the Layout method so the switch calculates the measurements and the objects are Moved and Resized afterwards.

toaster:
Cool!

andydotxyz:
Much neater, thanks :)

PR #762, opened 2020-03-24
BaseWidget should not implement CreateRenderer

toaster:
### Description:

The `widget.BaseWidget` should not implement a useless `CreateRenderer` method. It prevents the compiler from raising a meaningful error message if one forgot to implement it. Instead a GLFW error is logged which is almost impossible to debug.

### Checklist:

- ~~[ ] Tests included.~~
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/loop.go
+++ b/internal/driver/glfw/loop.go
@@ -1,6 +1,7 @@
 package glfw
 
 import (
+	"fmt"
 	"runtime"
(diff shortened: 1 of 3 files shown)


andydotxyz:
This passed, travis API seems flakey https://travis-ci.org/github/fyne-io/fyne/builds/666483098

PR #712, opened 2020-02-25
Support Text Wrapping

stuartmscott:
Support Text Wrapping

Fixes #332

Applies to Label, Hyperlink, and Entry

There are three text wrapping modes;
- TextWrapOff (default)
  - Widgets are expanded to fit the text.
  - Consistent with previous behaviour.
- TextTruncate
  - Text longer than the widget's width is not displayed.
  - An elipse is NOT added if text gets truncated, although this
    should be considered for future.
- TextWrapOn
  - Text longer than the widget's width is wrapped onto the next line.
  - A hyphen is NOT added if a word gets broken, although this
    should be considered for future (perhaps TextWrapOnHypenate).

This commit has some issues and is therefore not yet ready to merge;
- Truncation doesn't work with Selection;
  - Cursor position overflows row boundary and gets set to 0,0.
  - Select All only selects visible text.
- Truncation doesn't work with Scrollers;
  - Text does not get truncated, scroller just keeps expanding.
- Cursor moves to the wrong position after wrapping.
- Unclear how to handle Entry.multiline=false with TextWrapOn;
  - Maybe fallback to TextTruncate?.
- Test execution result in;
  - Warnings about multiple methods names 'center'
  - Warnings about the use of deprecated NSOpenGLContext

Issue #471, opened 2019-09-29
Scrollbars for multiline input(entry) widget


Kvaz1r:
**Is it possible to construct a solution with the existing API?**
Yes, it possible now if put the widget into ScrollContainer. 

**Describe the solution you'd like**
Entry widget itself can contain scrollbars (vertical and horizontal). It can be set automatically(after resize, after text changing) or always to be shown if corresponding property is set.

andydotxyz:
What are the benefits to adding the scrollers into the entry, above doing it the way you describe by putting the entry inside a scrollcontainer?

Kvaz1r:
Flexibility(allow use different modes), simplicity/convenient(no need create container by myself).

andydotxyz:
Fair enough. I think this is probably duplicate of #378 - the right way may be complex as we look at mobile as you can't really have a scrollcontainer within a scrollcontainer which might mean entry has to be limited based on it's context - any thoughts?

Kvaz1r:
I don't think that it's duplicate. 378 describes actual unexpected (and incorrect) behaviour i.e. bug. And this issue about settings i.e. feature request. But it's not really important.

andydotxyz:
I think the reverse was true ;) if we fix this the bug goes away.
You're probably right this is something that people will expect

andydotxyz:
As per #338 we will need to improve pageup/down handling one scrolling is in.
Obviously changing cursor position should update scroll position so it's visible.

andydotxyz:
It was agreed today that this should be added.
However it is a behaviour break so we're not comfortable putting it in until 2.0.
That is currently expected around July so should not be too long to wait.

fleshin:
"Obviously changing cursor position should update scroll position so it's visible."
Is this going to happen with single line Entry as well??

andydotxyz:
Yes, single and multi line entries are the same widget

fleshin:
Great, anxiously waiting for this to appear on the develop branch! Thanks! :)

andydotxyz:
On `develop` for testing

PR #764, opened 2020-03-24
Limit Scroll Directions

### LINKED ISSUE #763: Allow Disabling of ScrollContainer Direction
By default a ScrollContainer can scroll in both horizontal and vertical directions. When reading text scrolling horizontally is tedious.

Not possible at this time

A way to create a ScrollContainer for a given direction only whose MinSize takes into account the child content's MinSize.

For example, a ScrollContainer can be created for only the vertical direction. An Entry with text wrapping enabled can be added to ScrollContainer so that it can only be scrolled vertically and long lines of text are wrapped to the next line.

stuartmscott:
### Description:
Adds a Direction field to ScrollContainer to specify the allowed directions which defaults to both.

Fixes #763

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

- [x] Any breaking changes have a deprecation path or have been discussed. (no breaking changes)

### DIFF
--- a/cmd/fyne_demo/screens/widget.go
+++ b/cmd/fyne_demo/screens/widget.go
@@ -116,12 +116,14 @@ func makeScrollTab() fyne.CanvasObject {
 		}))
 	}
 
-	horiz := widget.NewScrollContainer(list)
-	vert := widget.NewScrollContainer(list2)
+	horiz := widget.NewHorizontalScrollContainer(list)
+	vert := widget.NewVerticalScrollContainer(list2)
+	vert.SetMinSize(fyne.NewSize(32, 128))
 
-	scrolls := fyne.NewContainerWithLayout(layout.NewBorderLayout(horiz, nil, nil, nil),
-		horiz, vert)
-	return fyne.NewContainerWithLayout(layout.NewAdaptiveGridLayout(2), scrolls, makeScrollBothTab())
+	return fyne.NewContainerWithLayout(layout.NewAdaptiveGridLayout(2), widget.NewVBox(
+		horiz,
+		vert,
+	), makeScrollBothTab())
 }
 
 func makeScrollBothTab() fyne.CanvasObject {
--- a/cmd/fyne_demo/screens/widget.go
+++ b/cmd/fyne_demo/screens/widget.go
@@ -118,12 +118,10 @@ func makeScrollTab() fyne.CanvasObject {
 
 	horiz := widget.NewHorizontalScrollContainer(list)
 	vert := widget.NewVerticalScrollContainer(list2)
-	vert.SetMinSize(fyne.NewSize(32, 128))
 
-	return fyne.NewContainerWithLayout(layout.NewAdaptiveGridLayout(2), widget.NewVBox(
-		horiz,
-		vert,
-	), makeScrollBothTab())
+	return fyne.NewContainerWithLayout(layout.NewAdaptiveGridLayout(2),
+		fyne.NewContainerWithLayout(layout.NewBorderLayout(horiz, nil, nil, nil), horiz, vert),
+		makeScrollBothTab())
 }
 
 func makeScrollBothTab() fyne.CanvasObject {
--- a/widget/entry_test.go
+++ b/widget/entry_test.go
@@ -57,36 +57,6 @@ func TestMultiLineEntry_MinSize(t *testing.T) {
 	assert.Equal(t, singleMin.Height, multiMin.Height)
 }
 
-func TestMultiLineEntry_Scroller_MinSize(t *testing.T) {
-	text := "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
-	size := textMinSize(text, theme.TextSize(), fyne.TextStyle{})
-
-	multi := NewMultiLineEntry()
-	multi.SetText(text)
-	multiMin := multi.MinSize()
-	assert.Equal(t, size.Width+2*theme.Padding()+theme.ScrollBarSize(), multiMin.Width)
-	assert.Equal(t, size.Height*multiLineRows+2*theme.Padding(), multiMin.Height)
-
-	t.Run("Both", func(t *testing.T) {
-		scroller := NewScrollContainer(multi)
-		scrollerMin := scroller.MinSize()
-		assert.Equal(t, 32, scrollerMin.Height)
-		assert.Equal(t, 32, scrollerMin.Width)
-	})
-	t.Run("HorizontalOnly", func(t *testing.T) {
-		scroller := NewHorizontalScrollContainer(multi)
-		scrollerMin := scroller.MinSize()
-		assert.Equal(t, size.Height*multiLineRows+2*theme.Padding(), scrollerMin.Height)
-		assert.Equal(t, 32, scrollerMin.Width)
-	})
-	t.Run("VerticalOnly", func(t *testing.T) {
-		scroller := NewVerticalScrollContainer(multi)
-		scrollerMin := scroller.MinSize()
-		assert.Equal(t, 32, scrollerMin.Height) // TODO FIXME Should equal size.Height*multiLineRows+2*theme.Padding()
-		assert.Equal(t, size.Width+2*theme.Padding()+theme.ScrollBarSize(), scrollerMin.Width)
-	})
-}
-
 func TestEntry_SetPlaceHolder(t *testing.T) {
 	entry := NewEntry()
 
--- a/widget/entry_test.go
+++ b/widget/entry_test.go
@@ -57,6 +57,36 @@ func TestMultiLineEntry_MinSize(t *testing.T) {
 	assert.Equal(t, singleMin.Height, multiMin.Height)
 }
 
+func TestMultiLineEntry_Scroller_MinSize(t *testing.T) {
+	text := "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+	size := textMinSize(text, theme.TextSize(), fyne.TextStyle{})
+
+	multi := NewMultiLineEntry()
+	multi.SetText(text)
+	multiMin := multi.MinSize()
+	assert.Equal(t, size.Width+2*theme.Padding()+theme.ScrollBarSize(), multiMin.Width)
+	assert.Equal(t, size.Height*multiLineRows+2*theme.Padding(), multiMin.Height)
+
+	t.Run("Both", func(t *testing.T) {
+		scroller := NewScrollContainer(multi)
+		scrollerMin := scroller.MinSize()
+		assert.Equal(t, 32, scrollerMin.Height)
+		assert.Equal(t, 32, scrollerMin.Width)
+	})
+	t.Run("HorizontalOnly", func(t *testing.T) {
+		scroller := NewHorizontalScrollContainer(multi)
+		scrollerMin := scroller.MinSize()
+		assert.Equal(t, size.Height*multiLineRows+2*theme.Padding(), scrollerMin.Height)
+		assert.Equal(t, 32, scrollerMin.Width)
+	})
+	t.Run("VerticalOnly", func(t *testing.T) {
+		scroller := NewVerticalScrollContainer(multi)
+		scrollerMin := scroller.MinSize()
+		assert.Equal(t, 32, scrollerMin.Height) // TODO FIXME Should equal size.Height*multiLineRows+2*theme.Padding()
+		assert.Equal(t, size.Width+2*theme.Padding()+theme.ScrollBarSize(), scrollerMin.Width)
+	})
+}
+
 func TestEntry_SetPlaceHolder(t *testing.T) {
 	entry := NewEntry()
 
--- a/widget/scroller.go
+++ b/widget/scroller.go
@@ -9,12 +9,16 @@ import (
 	"fyne.io/fyne/theme"
 )
 
-type scrollDirection int
+// ScrollDirection represents the directions in which a ScrollContainer can scroll its child content.
+type ScrollDirection int
 
 const (
-	scrollBoth = iota
-	scrollHorizontalOnly
-	scrollVerticalOnly
+	// ScrollBoth allows the ScrollContainer to scroll its child in both horizontal and vertical directions.
+	ScrollBoth ScrollDirection = iota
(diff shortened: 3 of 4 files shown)


widget/scroller.go
```diff
@@ -306,52 +326,59 @@ func (r *scrollContainerRenderer) handleShadowVisibility(offset int, contentSize
...
+	}
 
-	canvas.Refresh(r.vertArea)  // this is required to force the canvas to update, we have no "Redraw()"
-	canvas.Refresh(r.horizArea) // this is required like above but if we are horizontal
+	if r.scroll.Direction != scrollHorizontalOnly {
+		canvas.Refresh(r.vertArea) // this is required to force the canvas to update, we have no "Redraw()"
+	}
+	if r.scroll.Direction != scrollVerticalOnly {
+		canvas.Refresh(r.horizArea) // this is required like above but if we are horizontal
+	}
```
toaster: Will any test fail if you remove the `if`s here?

widget/scroller.go
```diff
@@ -306,52 +326,59 @@ func (r *scrollContainerRenderer) handleShadowVisibility(offset int, contentSize
...
 }
 
 // ScrollContainer defines a container that is smaller than the Content.
 // The Offset is used to determine the position of the child widgets within the container.
 type ScrollContainer struct {
 	BaseWidget
-
-	Content fyne.CanvasObject
-	Offset  fyne.Position
+	Direction scrollDirection
```
pymq: I think `scrollDirection` type should be public (and constants too), or `Direction` field should be private. Because in current implementation user can't assign anything to `Direction`. I like first option more.

andydotxyz:
Agreed, it's not a good idea to export a field that has a private type.
This may impact how the consts are set up to - I think the first one needs an explicit type set.

andydotxyz:
It can be helpful to use the fyne_demo code to test how this would work for end users.
In fact you could update the two button scrollers to be a single direction and remove the workaround code in there as a real test.
I think that will demonstrate the issue with the private field definition.

widget/scroller.go
stuartmscott: Added some more tests and now commenting out most 'ifs' and 'switches' will cause test failures, except:
```
if r.scroll.Direction != ScrollHorizontalOnly {
	canvas.Refresh(r.vertArea) // this is required to force the canvas to update, we have no "Redraw()"
}
if r.scroll.Direction != ScrollVerticalOnly {
	canvas.Refresh(r.horizArea) // this is required like above but if we are horizontal
}
```
Since canvas.Refresh is happy to take a nil arg.

widget/scroller.go
stuartmscott: Thanks @pymq, these types and constants are now public.

stuartmscott:
@toaster I see you added minSize to scrollBarRenderer:
```
cb056f4f8 (Tilo Prütz      2019-06-13 17:55:10 +0200  32)       minSize fyne.Size
```
Which gets read:
```
func (r *scrollBarRenderer) MinSize() fyne.Size {
	return r.minSize
}
```
But I don't see where it gets written, am I missing something?

@andydotxyz The first constant didn't need a type to compile (guess the compiler infers the type), but I added it anyway for consistency and clarity.

I also updated fyne_demo to use NewHorizontalScrollContainer and NewVerticalScrollContainer, on the latter I had to use SetMinSize because the BorderLayout forced it to fill the screen, but BoxLayout wraps it to its MinSize (which defaults to 32x32)

andydotxyz:
I suspect the content worked because the type is a type of int and 0 is an int... it may be what allowed the gl code to accept 0 instead of a typed value (but that’s fixed now).

I don’t understand your comment about the vertical scroller - should the minsize not be set to the width of the contents in vertical mode? If this is done correctly we should not need setMinSize workarounds in the majority of cases..?

stuartmscott:
> I don’t understand your comment about the vertical scroller - should the minsize not be set to the width of the contents in vertical mode? If this is done correctly we should not need setMinSize workarounds in the majority of cases..?

Width yes, height no. With this change the width of a vertical scroller is the same as the MinSize.Width of the child, however the height of a vertical scroller falls back to the default (32). Hence this change calls SetMinWidth but only actually changes the min height:
```
vert.SetMinSize(fyne.NewSize(32, 128))
```
Similarly the height of a horizontal scroller is set to the child's height, but the width falls back to the default (32).

This behaviour depends on the layout as well. Most layouts squash the scroller to the MinSize, but others like MaxLayout and BorderLayout(center) set their child to fill all the available space.

This presents a problem with Multiline Entry in Vertical Scrollers - the height will be 32, however Multiline Entries wants at least multiLineRows visible.

widget/scroller.go
toaster: Okay, `canvas.Refresh(nil)` will probably panic in a real world app with a totally incomprehensible error. So it's better to leave the `if` where they are :)

toaster:
> @toaster I see you added minSize to scrollBarRenderer:
…
> But I don't see where it gets written, am I missing something?

Oops :). You are right. That might need some clean-up.

toaster:
@stuartmscott Please, _never ever_ do a force push on a PR that already has been reviewed! Now I have to start from the beginning. That's annoying and costs my time.

andydotxyz:
> Width yes, height no. With this change the width of a vertical scroller is the same as the MinSize.Width of the child, however the height of a vertical scroller falls back to the default (32). Hence this change calls SetMinWidth but only actually changes the min height:

The vertical scroller in fyne_demo is packed to the left of a borderlayout so the height should automatically expand. I don't see how minSize is needed - are you sure there is not a bug in here somewhere?

cmd/fyne_demo/screens/widget.go
```diff
@@ -116,12 +116,14 @@ func makeScrollTab() fyne.CanvasObject {
...
-	vert := widget.NewScrollContainer(list2)
-
-	scrolls := fyne.NewContainerWithLayout(layout.NewBorderLayout(horiz, nil, nil, nil),
-		horiz, vert)
-	return fyne.NewContainerWithLayout(layout.NewAdaptiveGridLayout(2), scrolls, makeScrollBothTab())
+	horiz := widget.NewHorizontalScrollContainer(list)
+	vert := widget.NewVerticalScrollContainer(list2)
+	vert.SetMinSize(fyne.NewSize(32, 128))
+
+	return fyne.NewContainerWithLayout(layout.NewAdaptiveGridLayout(2), widget.NewVBox(
```
andydotxyz: Oh I see why you are having problems with VBox now - why don't you use the container/layout that was already here that did not suffer the problem you describe?

stuartmscott: That's fine for this case, but adding a scroller to a borderlayout to force it to take up all the space is not a solution that can be applied to all use cases. We need something better.

In general, layouts measure the min size of their objects, but if that layout is invoked to layout objects in a bigger size, shouldn't it expand its objects to take up all the space? Granted there are cases you wouldn't want this, eg toolbars should have a max height.

stuartmscott:
> Oops :). You are right. That might need some clean-up.

@toaster I'll leave that with you.

> Please, never ever do a force push on a PR that already has been reviewed! Now I have to start from the beginning. That's annoying and costs my time.

Sorry @toaster, didn't mean to cost your time - I pulled the develop branch and rebased this scroller branch on top of it, but there were some recent changes to ScrollContainer which caused conflicts with mine. After resolving them git would not let me do a normal push, hence the force push. What is your normal workflow to keep PRs up to date?

widget/scroller.go
stuartmscott: True, it might be worth adding a nil check inside canvas.Refresh

toaster:
> Sorry @toaster, didn't mean to cost your time - I pulled the develop branch and rebased this scroller branch on top of it, but there were some recent changes to ScrollContainer which caused conflicts with mine. After resolving them git would not let me do a normal push, hence the force push. What is your normal workflow to keep PRs up to date?

A PR branch must not be rebased. Of course you can't normal push a rebased branch: you changed the history.
Instead simply merge the develop branch into the PR branch.

widget/scroller.go
```diff
@@ -9,12 +9,16 @@ import (
...
 const (
-	scrollBoth = iota
-	scrollHorizontalOnly
-	scrollVerticalOnly
+	// ScrollBoth allows the ScrollContainer to scroll its child in both horizontal and vertical directions.
+	ScrollBoth ScrollDirection = iota
+	// ScrollHorizontalOnly allows the ScrollContainer to only scroll its child in the horizontal direction.
+	ScrollHorizontalOnly
+	// ScrollVerticalOnly allows the ScrollContainer to only scroll its child in the vertical direction.
+	ScrollVerticalOnly
```
toaster: FYI: I think for constants that are mostly self-describing (which of course they should be ;)), a simple comment before the `const` block should be sufficient:

```go
// ScrollDirection represents the directions in which a ScrollContainer can scroll its child content.
type ScrollDirection int

// Constants for valid values of ScrollDirection.
const (
	ScrollBoth ScrollDirection = iota
	ScrollHorizontalOnly
	ScrollVerticalOnly
)
```

stuartmscott:
> Instead simply merge the develop branch into the PR branch.

Not sure I follow. You're saying every time there is update to develop, run;
```
git checkout scroller
git merge develop
git push
```
Surely this changes the history too? And what if there are conflicts?

widget/scroller.go
stuartmscott: Done

stuartmscott:
I already rebased before your last comment, I tried rebasing back onto the parent commit but again got a new commit hash, so I'm going to need to force push again;
```
$ git push
To github.com:stuartmscott/fyne.git
 ! [rejected]          scroller -> scroller (non-fast-forward)
error: failed to push some refs to 'git@github.com:stuartmscott/fyne.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

toaster:
Ouch! You did it again.

Please. Do. Not. Force-Push. To. PR-Branches.

toaster:
> > Instead simply merge the develop branch into the PR branch.
> 
> Not sure I follow. You're saying every time there is update to develop, run;
> 
> ```
> git checkout scroller
> git merge develop
> git push
> ```
> 
> Surely this changes the history too? And what if there are conflicts?

No, this adds a merge commit on top of the history.
If there are conflicts, you fix them in the merge commit.

And: you don't need to do this every time `develop` is updated. Actually we at Fyne are relaxed with merging branches that are not up-to-date.
If there are conflicts, you should merge `develop` of course, to fix the conflicts.

toaster:
I'll try another approach on review tomorrow.

stuartmscott:
I tried your approach and merged develop into scroller and pushed that. Thanks for helping me understand your workflow.

Unfortunately my other PR #712 was rebased before all this, so I will need to force push that, but that'll be the last time, promise :)

stuartmscott:
> This presents a problem with Multiline Entry in Vertical Scrollers - the height will be 32, however Multiline Entries wants at least multiLineRows visible.

Added a test to demonstrate this

stuartmscott:
> Unfortunately my other PR #712 was rebased before all this, so I will need to force push that, but that'll be the last time, promise :)

Actually I managed to get around this by checking out the remote branch to a new local branch, cherry-picking the commits across from the old local branch, then pushing.

toaster:
I have no more objections 👍. @andydotxyz are you happy, too?

cmd/fyne_demo/screens/widget.go
andydotxyz: Any widget that is expanded can ask it's content to expand. HBox and VBox are special in that they pack items to the left, or the top. That is how they are designed and so by moving to use VBox you require the workaround above. I don't think there is a good reason to keep it this way.

If you feel layouts need improvements please open a new issue so we can discuss separately.

andydotxyz:
I'm happy with the change and the API - but I would like the demo code to go back to using the previous layout to avoid use of a MinSize workaround.

widget/entry_test.go
```diff
@@ -57,6 +57,36 @@ func TestMultiLineEntry_MinSize(t *testing.T) {
...
+	t.Run("HorizontalOnly", func(t *testing.T) {
+		scroller := NewHorizontalScrollContainer(multi)
+		scrollerMin := scroller.MinSize()
+		assert.Equal(t, size.Height*multiLineRows+2*theme.Padding(), scrollerMin.Height)
+		assert.Equal(t, 32, scrollerMin.Width)
+	})
+	t.Run("VerticalOnly", func(t *testing.T) {
+		scroller := NewVerticalScrollContainer(multi)
+		scrollerMin := scroller.MinSize()
+		assert.Equal(t, 32, scrollerMin.Height) // TODO FIXME Should equal size.Height*multiLineRows+2*theme.Padding()
```
stuartmscott: @andydotxyz @toaster can I draw you attention to this test which shows scrollers do not respect multiline entries multiLineRows? If a height of 32 is okay then I can remove the comment, but if we want multiline entries to always display at least multiLineRows then there will need to be more coordination between entries and scrollers.

andydotxyz: Absolutely. The conversation about how an entry should be presented, and how it should scroll is ongoing, you can see it in the discussion about test wrapping and the issue #471.
The minSize for entry in multiline mode is correct - but when you put it in a scroller that overrides the minSize. Scrollers do not know anything about their content, and so could not query the Entry.multilineRows property...
Surely this is a separate issue to the discussion in hand?

stuartmscott: You're right - this is a separate issue to limiting scroll directions. For now I'll leave the test as is (with comment) and we can come back to it when we have an answer for #471

andydotxyz: Sorry somehow I missed that this was new tests. I’d be tempted to say don’t add entry scroll tests just now it’s a whole other topic that belongs in one of the other discussions. Could this perhaps be added as a patch attachment to an appropriate discussion on entry and left out of this PR for now? 

Thanks for bringing it to your attention, sorry I didn’t notice the addition of new entry tests.

stuartmscott: Reverted

stuartmscott:
Travis appears stuck again

andydotxyz:
Thanks for all of this - looks like Travis returned after all

PR #766, opened 2020-03-25
Allow Linear or Nearest Scaling Filter on canvas.Image

laullon:
### Description:

Adds the option to choose between Linear or Nearest filters for textures

```
	display := canvas.NewImageFromImage(machine.Display())
	display.FillMode = canvas.ImageFillOriginal
	display.ScalingFilter = canvas.NearestFilter
```

By default will use `Linear` filter for backwards compatibility.
 
* canvas.NearestFilter
Screenshot 2020-03-25 at 13 18 28

* canvas.LinearFilter
Screenshot 2020-03-25 at 13 18 03

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


laullon:
Better now?

internal/painter/gl/gl_common.go
```diff
@@ -117,7 +117,7 @@ func (p *glPainter) newGlTextTexture(obj fyne.CanvasObject) Texture {
 	d.Dot = freetype.Pt(0, height-face.Metrics().Descent.Ceil())
 	d.DrawString(text.Text)
 
-	return p.imgToTexture(img)
+	return p.imgToTexture(img, 0)
```
andydotxyz: This should be LinearFilter rather than 0 should it not?

andydotxyz: (and a couple of other instances below)

andydotxyz:
The image illustration above is incredible - I wonder if this is the source of the fuzzy text if your scale gets too small? Part of me wonders if LINEAR is the best default afterall?
I agree that probably that should remain the default for images - but maybe NEAREST makes more sense for our other types of texture (or maybe specifically text?)
Would be interested in @toaster, @lucor and @okratitan thoughts here...

laullon:
The fuzzy text is because the original text (image) is too small (is on ZX Spectrum resolution), NEAREST makes senes in a pixel art application.

andydotxyz:
I think this is failing to build due to an out of date vendor folder?

andydotxyz:
Apologies, the fuzzy text I was referring to was *canvas.Text, we have noticed that at some scale values it is not well rendered, I wonder if it's related.

laullon:
you see text, but is not text, both screenshots text are image

andydotxyz:
Yes indeed. But inside Fyne everything is a texture, so text becomes an image before it is rendered...
It was just a thought - I could have been way off.

toaster:
I think this is a good addition. @laullon do you think you could add some tests? AFAIR there is a `Capture()` function to capture the current state of a canvas in an image. This should make it possible to making a test which expects different outcome for the different filters.

laullon:
@toaster can you point me to a similar test? I can implement a new one for this

toaster:
> @toaster can you point me to a similar test? I can implement a new one for this

Unfortunately not. This is unknown area. You have the honour of being the first to implement one. ;)

andydotxyz:
Just thought I would revisit this ticket - it seems like something good to land but has got stuck on testing perhaps? Is there something we can do to help set this up @laullon?

laullon:
I just need time to figure out how to implement the test ;-)

laullon:
This is my app: https://github.com/laullon/b2t80s

toaster:
@laullon I just finished a `test.AssertImagesEqual` test helper which allows to implement render tests based on master images. It might land within the next one or two weeks (have other stuff to do first). It’s basically this:
```go
// AssertImagesEqual asserts that the master image at given filename (below testdata) and the given image are equal.
// It fails the test if not. In this case the actual image is written to disk and reported.
func AssertImagesEqual(t *testing.T, masterFilename string, got image.Image) {
	wd, err := os.Getwd()
	require.NoError(t, err)
	masterPath := filepath.Join("testdata", masterFilename)
	var expected image.Image
	_, err = os.Stat(masterPath)
	if os.IsNotExist(err) {
		fmt.Printf("Master image does not exist at %s.\nAssume initial run and use empty image for comparison.\n", filepath.Join(wd, masterPath))
		expected = image.NewRGBA(image.Rectangle{})
	} else {
		file, err := os.Open(masterPath)
		require.NoError(t, err)
		defer file.Close()
		raw, _, err := image.Decode(file)
		require.NoError(t, err)
		normalized := image.NewRGBA(raw.Bounds())
		draw.Draw(normalized, normalized.Bounds(), raw, image.Pt(0, 0), draw.Src)
		expected = normalized
	}
	if !assert.Equal(t, expected, got) {
		path := filepath.Join("testdata/failed", masterFilename)
		require.NoError(t, os.MkdirAll(filepath.Dir(path), 0755))
		file, err := os.Create(path)
		require.NoError(t, err)
		defer file.Close()
		require.NoError(t, png.Encode(file, got))
		fmt.Println("Images were not equal. Actual image written to:", filepath.Join(wd, path))
	}
}
```

And it is used in tests like this:

```go
	canvas := fyne.CurrentApp().Driver().CanvasForObject(o)
	test.AssertImagesEqual(t, "master.png", canvas.Capture())
```

In GL tests you probably already have the canvas or some better way to get it than the one used above.

andydotxyz:
I suspect a nice utility or two may be written on top of that too, such as `test.AssertObjectMatchesImage(*testing.T, fyne.CanvasObject, string)` to avoid the canvas lookup :)

steveoc64:
Im really liking this PR

Let me know if there is anything I can do to help get it over the line.

That Z80 project is simply Awesome !

laullon:
Turns out @andydotxyz already implemented the image tests, so I did take advantage of it.

laullon:
@andydotxyz I did some change on the `drawImage` method to reduce the number of image paint, can you take a look?

andydotxyz:
Sorry @laullon but you removed some important code that is causing the test failures you can see in the Travis runs.
If you look carefully you have replaced test images with ones that are clearly not the intended output...

Screenshot 2020-05-09 at 18 22 41

Perhaps you can comment on why those were changed, or was it just an oversight?
Normally we don't do reviews on PRs that are failing like this, but I will run through quickly to try and explain where you went wrong.

.travis.yml
```diff
@@ -20,6 +20,7 @@ before_script:
   - GO111MODULE=off go get golang.org/x/tools/cmd/goimports
   - GO111MODULE=off go get github.com/fzipp/gocyclo
   - GO111MODULE=off go get golang.org/x/lint/golint
+  - GO111MODULE=off go get golang.org/x/image/draw
```
andydotxyz, line 23: This should not be needed - we only need to get tools that will be used in the built process.
Dependencies are handled by `go mod` or the `vendor` folder.

canvas/base.go
```diff
@@ -3,7 +3,6 @@ package canvas // import "fyne.io/fyne/canvas"
 
 import (
 	"sync"
-
```
andydotxyz, line 6: `goimports` wants this line to remain. I would recommend using our pre-commit git hook so you don't accidentally commit things that break the build. https://github.com/fyne-io/fyne/wiki/GitHooks

canvas/image.go
```diff
@@ -24,6 +24,16 @@ const (
 	ImageFillOriginal
 )
 
+// TextureFilter defines the different scaling filters used to scaling images
+type TextureFilter int32
```
andydotxyz, line 28: The word Filter tripped me up here - is it really a filter? The go docs seem to use the term "Interpolator" (which is quite ungainly).
The reason that I am unsure is that we are about to add FileFilter (or similar) to our file listing UI and the two different uses of the term could be confusing.
What would you think about "Sample" instead? That seems to be suggested higher up in google results than "filter" if I search for "image bilinear"... (open question for comment)

canvas/image.go
```diff
@@ -40,6 +50,8 @@ type Image struct {
 
 	Translucency float64   // Set a translucency value > 0.0 to fade the image
 	FillMode     ImageFill // Specify how the image should scale to fill or fit
+
+	ScalingFilter TextureFilter // set how the image is scaled
```
andydotxyz, line 54: This name may relate to the question above. It's not great to repeat if it can be avoided also, but may depend on what the opinions are on the other naming question.

go.sum
```diff
@@ -53,6 +53,8 @@ golang.org/x/image v0.0.0-20190802002840-cff245a6509b h1:+qEpEAPhDZ1o0x3tHzZTQDA
 golang.org/x/image v0.0.0-20190802002840-cff245a6509b/go.mod h1:FeLwcggjj3mMvU+oOTbSwawSJRM1uh48EjtB4UJZlP0=
 golang.org/x/image v0.0.0-20200119044424-58c23975cae1 h1:5h3ngYt7+vXCDZCup/HkCQgW5XwmSvR/nA2JmJ0RErg=
```
andydotxyz, line 54: There is now a duplicate in here - running "go mod tidy" will remove the old references.

internal/painter/gl/gl_common.go
```diff
@@ -131,7 +131,7 @@ func (p *glPainter) newGlImageTexture(obj fyne.CanvasObject) Texture {
 		return NoTexture
 	}
 
-	return p.imgToTexture(tex)
+	return p.imgToTexture(tex, img.ScalingFilter)
```
andydotxyz, line 134: We should sanity check this value I think - if the user sets it to "2" we will crash with index out of bounds in the draw code...

internal/painter/software/draw.go
```diff
@@ -34,29 +32,33 @@ func drawImage(c fyne.Canvas, img *canvas.Image, pos fyne.Position, frame fyne.S
...
 	height := internal.ScaleInt(c, bounds.Height)
-	tex := painter.PaintImage(img, c, width, height)
 
 	if img.FillMode == canvas.ImageFillStretch {
 		width = internal.ScaleInt(c, img.Size().Width)
 		height = internal.ScaleInt(c, img.Size().Height)
-		tex = resize.Resize(uint(width), uint(height), tex, resize.Lanczos3)
 	} else if img.FillMode == canvas.ImageFillContain {
-		imgAspect := painter.GetAspect(img)
+		imgAspect := float32(img.Size().Width) / float32(img.Size().Height)
```
andydotxyz, line 40: Sorry - what you have calculated here is the output aspect ratio - what the `GetAspect` is returning is a cache of the source image ratio. With this change the `objAspect` below will always be 1, this is the source of all your test failures.

internal/painter/software/draw.go
```diff
@@ -34,29 +32,33 @@ func drawImage(c fyne.Canvas, img *canvas.Image, pos fyne.Position, frame fyne.S
...
 			pos.Y += (height - newHeight) / 2
-			height = internal.ScaleInt(c, newHeight)
+			height = newHeight //internal.ScaleInt(c, newHeight)
 		}
-		tex = resize.Resize(uint(width), uint(height), tex, resize.Lanczos3)
 	}
 
-	drawTex(c, pos, width, height, base, tex)
+	scaledX, scaledY := internal.ScaleInt(c, pos.X), internal.ScaleInt(c, pos.Y)
+	outBounds := image.Rect(pos.X, pos.X, pos.X+scaledX+width, pos.Y+scaledY+height)
```
andydotxyz, line 55: This line looks wrong - you are adding a scaled x value to an unscaled value...
I wonder if it should be `Rect(scaledX, scaledY, scaledX+width, scaledY+height)`

internal/painter/software/draw.go
```diff
@@ -34,29 +32,33 @@ func drawImage(c fyne.Canvas, img *canvas.Image, pos fyne.Position, frame fyne.S
...
+	switch img.ScalingFilter {
+	case canvas.LinearFilter:
```
andydotxyz, line 57: Thanks for adding to software painter too, this should work once the calculations above are fixed

internal/painter/software/draw_test.go
```diff
@@ -35,6 +35,26 @@ func TestDrawImage(t *testing.T) {
 	test.AssertImageMatches(t, "draw_image_default.png", target)
 }
 
+func TestDrawImageNearestFilter(t *testing.T) {
```
andydotxyz, line 38: These tests are what we needed I think, thanks. 
The output images we compare with make the value of this clear :)

andydotxyz:
Thanks. The tests added are great and we're close on the original code.
Unfortunately the optimisations added since have broken lots of other tests, evident in the image comparisons.
Additionally I have started to wonder if "filter" is the right term (all commented inline).
Very close, thanks for coming back to this to get it wrapped up.

canvas/image.go
laullon: Maybe we can use `ScalingMode`, same way as `FillMode` ?

andydotxyz:
Have you missed a file change in vendor (the modules.txt listing)? I think that is why the tests are failing

canvas/image.go
andydotxyz: That's a good suggestion. Probably should keep the tense the same which would indicate `ScaleMode` instead of `ScalingMode`. That said I wonder if it clashes with the concept of scaling for size / position values. 
Anyone got thoughts?

internal/painter/software/draw_test.go
```diff
@@ -65,8 +86,9 @@ func TestDrawImage_Contain(t *testing.T) {
 
 func TestDrawImage_ContainX(t *testing.T) {
 	target := image.NewNRGBA(image.Rect(0, 0, 100, 50))
-	img := canvas.NewImageFromImage(makeTestImage())
+	img := canvas.NewImageFromImage(makeTestImage(3, 5))
```
andydotxyz, line 89: Are these values the wrong way round? I assumed that when you created the rectangular pattern it was intended to be in the same orientation as the contains test?

canvas/image.go
toaster: What about `InterpolationMode` with `Linear` and `NearestNeighbour` as currently possible values?

internal/painter/software/draw_test.go
laullon: not, I just want a not 1:1 aspect ratio

canvas/image.go
laullon: GL call it Texture filter (`gl.TexParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR)`)

andydotxyz: Very true, but I think that doesn't really follow outside of GL and this is a toolkit global name

internal/painter/software/draw_test.go
andydotxyz: I guess I don't understand why this changed from before. The new parameters were added for this, and the following, function - so I was curious why the change.

andydotxyz:
Travis is green, not sure why the status is not reported yet

canvas/image.go
laullon: so? ScalingMode ?

internal/painter/software/draw_test.go
laullon: Still, I'm not sure that we cover all possibilities... maybe I'll add more test cases.

laullon:
Travis hate me.

canvas/image.go
stuartmscott: Its long, but I'd go with TextureScalingFilter

andydotxyz:
Travis is green now (the last failure is a temporary windows thing that I think is fixed on latest develop).

canvas/image.go
andydotxyz: I'd drop Texture becauase that's a GL name.  It could be ImageScalingFilter I suppose. 
Looking again at the `Image.FillMode ImageFill` type it could indicate `image.ScaleMode ImageScale` with `ImageScale[Linear|NearestNeighbour]`.
It's a tough call: `Interpolation` seems correct but too complicated, `Filter` is clearly image related but in an instagram world means something else. `Scale` too has other connotations - all that is left from the suggestions is really `Sample` but that didn' seem popular.

I'll open a vote on the #fyne slack channel to see what our API consumers think.

andydotxyz: Possibly some food for thought from other environments:

iOS uses an InterpolationQuality setting which is another way to go
QT’s QImage uses a parameter to image.scale() such as Qt::FastTransformation
EFL seems to opt for “smooth_scale” bool
In general it seems like the implementation detail is hidden which is probably a good idea
And GDK-pixbuf does expose the algorithm choice under “GDK_INTERP_” prefix, so we wouldn’t be alone

andydotxyz: I wonder also about the values, perhaps they should mean more and/or not be implementation specific.
How about Pixel[Scale] and Smooth[Scale] perhaps? I guess there is also the "slow but best" option that is not being added in this PR.

andydotxyz: Looks like the channel has spoken and `ScaleMode` is the best of what we have just now.
So that's an indicator of what the field should be, then the field type would follow as `ImageScale` if we copy `FillMode`.
Then I would propose that we name the consts to match intent so perhaps `ImageScalePixels`, `ImageScaleSmooth`?

internal/painter/gl/gl_core.go
```diff
@@ -26,24 +27,30 @@ type Texture uint32
...
 var NoTexture = Texture(0)
 
-func newTexture() Texture {
+var textureFilterToGL = []int32{gl.LINEAR, gl.NEAREST}
+
+func newTexture(textureFilter canvas.ImageScale) Texture {
 	var texture uint32
 
+	if int(textureFilter) >= len(textureFilterToGL) {
+		panic(fmt.Sprintf("Invalid canvas.ImageScale value (%d)", textureFilter))
```
andydotxyz, line 36: As a library we should never panic.
I would recommend using fyne.LogError and then setting it to the default value and continuing

internal/painter/gl/gl_es.go
```diff
@@ -27,24 +28,28 @@ type Texture uint32
 // NoTexture is the zero value for a Texture
 var NoTexture = Texture(0)
 
-func newTexture() Texture {
+func newTexture(textureFilter canvas.ImageScale) Texture {
 	var texture uint32
 
+	if int(textureFilter) >= len(textureFilterToGL) {
+		panic(fmt.Sprintf("Invalid canvas.ImageScale value (%d)", textureFilter))
```
andydotxyz, line 35: As above, don't panic

internal/painter/gl/gl_gomobile.go
```diff
@@ -33,23 +34,27 @@ func (p *glPainter) glctx() gl.Context {
 	return p.context.Context().(gl.Context)
 }
 
-func (p *glPainter) newTexture() Texture {
+func (p *glPainter) newTexture(textureFilter canvas.ImageScale) Texture {
 	var texture = p.glctx().CreateTexture()
 
+	if int(textureFilter) >= len(textureFilterToGL) {
+		panic(fmt.Sprintf("Invalid canvas.ImageScale value (%d)", textureFilter))
```
andydotxyz, line 41: one more :)

internal/painter/software/draw.go
```diff
@@ -34,29 +33,32 @@ func drawImage(c fyne.Canvas, img *canvas.Image, pos fyne.Position, frame fyne.S
...
+			scaledY += (height - newHeight) / 2
 			height = internal.ScaleInt(c, newHeight)
 		}
-		tex = resize.Resize(uint(width), uint(height), tex, resize.Lanczos3)
 	}
 
-	drawTex(c, pos, width, height, base, tex)
+	outBounds := image.Rect(scaledX, scaledY, scaledX+width, scaledY+height)
+	switch img.ScaleMode {
+	case canvas.ImageScaleSmooth:
```
andydotxyz, line 57: In line with the addition of panic lines in the GL driver I wonder if this should be in a default block? just a thought, open to discussion :)

andydotxyz:
This seems very clean now, I think the naming works well - how do you feel about this?
Other than the panics you added I reckon it's really close if others are happy with the names.

internal/painter/gl/gl_core.go
laullon: I think this is a dev error and falling back to a default value can give the impression that any value is ok, which is not...

But is a philosophical thing, so if you want a log msg I can change.

internal/painter/software/draw.go
laullon: Same, I think default should panic, but is up to you (or the team ;-) ).

internal/painter/gl/gl_core.go
andydotxyz: Is it the sort of developer error that is irrevocable and must force the application to close?
By leaving this in you are opening the possibility that a functioning app may force close in the middle of use. If we can fall back to a decent default we could save the user's sanity ;)

andydotxyz:
Also I think you included some test code in `cmd/img` by mistake

canvas/image.go
```diff
@@ -38,8 +48,9 @@ type Image struct {
 	Resource fyne.Resource // Load the image from an in-memory resource
 	Image    image.Image   // Specify a loaded image to use in this canvas object
 
-	Translucency float64   // Set a translucency value > 0.0 to fade the image
-	FillMode     ImageFill // Specify how the image should scale to fill or fit
+	Translucency float64    // Set a translucency value > 0.0 to fade the image
+	FillMode     ImageFill  // Specify how the image should scale to fill or fit
+	ScaleMode    ImageScale // set how the image is scaled
```
stuartmscott, line 53: nit: the comment should be capitalized, and maybe be a bit more descriptive. a user may ask what is the difference between "how the image should scale to fill or fit" and "how the image is scaled"

cmd/img/main.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"fyne.io/fyne"
+	"fyne.io/fyne/app"
+	"fyne.io/fyne/canvas"
+	"fyne.io/fyne/cmd/fyne_demo/data"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/widget"
+)
+
+func main() {
```
stuartmscott, line 13: I think we could keep this in if it was able to showcase the different scale modes, maybe add a button so the user can flip through the different modes?

stuartmscott: Or maybe this should be integrated into fyne_demo?

andydotxyz: Yeah if it needs to be kept then pop in graphics demo - we don’t need another App I don’t think

laullon: I was committed by mistake ;-)

internal/painter/gl/gl_core.go
laullon: something like:
```
	if int(textureFilter) >= len(textureFilterToGL) {
		fyne.LogError(fmt.Sprintf("Invalid canvas.ImageScale value (%d), using 0 (canvas.ImageScaleSmooth) as defualt value", textureFilter), nil)
		textureFilter = canvas.ImageScaleSmooth
	}
```

canvas/image.go
laullon: **Specify how the image should be scaled** is ok?

andydotxyz: "Specify the type of scaling applied to the image content"?

toaster:
> This seems very clean now, I think the naming works well - how do you feel about this?
> Other than the panics you added I reckon it's really close if others are happy with the names. 

The names are okay. But the panics should be removed, yes.

internal/painter/gl/gl_core.go
andydotxyz: Yes, except I would avoid `using 0 (canvas.ImageScaleSmooth)` and just say `using canvas.ImageScaleSmooth` - the specific int value is not something we should encourage.

canvas/image.go
stuartmscott: Hmm I think the dev could get confused by the two different uses of the word scale, but that may be more the fault of the FillMode comment above.
```suggestion
	FillMode	ImageFill // Specify how the image should expand to fill or fit the available space
	ScaleMode	ImageScale // Specify how type of scaling applied to the image
```

canvas/image.go
```diff
@@ -38,8 +48,9 @@ type Image struct {
 	Resource fyne.Resource // Load the image from an in-memory resource
 	Image    image.Image   // Specify a loaded image to use in this canvas object
 
-	Translucency float64   // Set a translucency value > 0.0 to fade the image
-	FillMode     ImageFill // Specify how the image should scale to fill or fit
+	Translucency float64    // Set a translucency value > 0.0 to fade the image
+	FillMode     ImageFill  // Specify how the image should expand to fill or fit the available space
+	ScaleMode    ImageScale // Specify how type of scaling applied to the image
```
andydotxyz, line 53: This sentence doesn’t make sense - did you mean “Specify the type” instead of “Specify how type”?

laullon: I just merged @stuartmscott change ;-)

andydotxyz: I don’t think that what exists in the result is what either of us suggested

stuartmscott: There is a typo in the change I suggested, but let's not merge it into the repo

internal/painter/software/draw_test.go
```diff
@@ -0,0 +1,127 @@
+package software
```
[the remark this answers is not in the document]
andydotxyz: It looks like this file was commented out in it's entirety during a merge commit.
Please don't do this, it was lucky that @stuartmscott caught this as merges are often not checked as thoroughly based on the assumption that you're not making changes outside the merge code change.

canvas/image.go
andydotxyz: Oh yes I see it now. Probably best to sanity check suggestions ;)

andydotxyz:
I don't know how others feel about this but I wonder if this PR has gone through so many changes and develop merges that it would probably be a lot cleaner to re-apply the work on top of current develop.

laullon:
Just Squashing all commits during merge will create a clean history

internal/painter/software/painter_test.go
```diff
@@ -12,37 +12,65 @@ import (
...
+
+	target := p.Paint(c)
+	test.AssertImageMatches(t, "draw_image_ImageScaleSmooth.png", target)
+}
+
 func TestPainter_PaintImage_StretchX(t *testing.T) {
 	c := test.NewCanvas()
 	c.SetPadded(false)
-	c.SetContent(canvas.NewImageFromImage(makeTestImage()))
+	c.SetContent(canvas.NewImageFromImage(makeTestImage(4, 3)))
```
andydotxyz, line 73: Just scanning through the images and this new 4-3 ratio is hard to tell if it is stretched in the 1-2 ratio of the output image (same below). Can you go back to the 3-3 please?

andydotxyz:
Fair point.
I was about to approve this but then I scanned the test images. The old ScaleX and ScaleY are far harder to visually inspect since the change in aspect ratio. Sorry I didn't catch this sooner but 1:1 -> 1:2 was much easier to read than the new 3:4 -> 1:2

internal/painter/software/painter_test.go
laullon: Sure

internal/painter/gl/gl_es.go
```diff
@@ -27,24 +28,29 @@ type Texture uint32
 // NoTexture is the zero value for a Texture
 var NoTexture = Texture(0)
 
-func newTexture() Texture {
+func newTexture(textureFilter canvas.ImageScale) Texture {
 	var texture uint32
 
+	if int(textureFilter) >= len(textureFilterToGL) {
+		fyne.LogError(fmt.Sprintf("Invalid canvas.ImageScale value (%d), using canvas.ImageScaleSmooth as defualt value", textureFilter), nil)
```
stuartmscott, line 35: s/defualt/default

internal/painter/gl/gl_gomobile.go
```diff
@@ -33,23 +34,28 @@ func (p *glPainter) glctx() gl.Context {
 	return p.context.Context().(gl.Context)
 }
 
-func (p *glPainter) newTexture() Texture {
+func (p *glPainter) newTexture(textureFilter canvas.ImageScale) Texture {
 	var texture = p.glctx().CreateTexture()
 
+	if int(textureFilter) >= len(textureFilterToGL) {
+		fyne.LogError(fmt.Sprintf("Invalid canvas.ImageScale value (%d), using canvas.ImageScaleSmooth as defualt value", textureFilter), nil)
```
stuartmscott, line 41: s/defualt/default

internal/painter/software/draw.go
andydotxyz: I guess this one is still up for discussion? The behaviour is not consistent with the GL painter...

laullon: default ?

andydotxyz: That would fix it I guess, maybe error and fallback like the Gl?

andydotxyz:
Thanks so much for this - sorry it was a long journey to the finished API.

laullon:
It was fun !!!

andydotxyz:
I just thought I would add this as a little thanks - it made the emulator look way clearer :)
gameboy-pixels

PR #772, opened 2020-03-27
Improve native menus

toaster:
### Description:

This PR adds some improvements regarding native menus (MacOS):

- allow menu items to be added to the native app menu (Settings for instance)
- allow menu items to have a “key equivalent” which is a way to define a short-cut
- allow menus to be placed before (default) or after the additional native menus (“Window” on MacOS)
- allow items to wish for a separator to the prior items

All those changes only affect the native menus because there is no support for menu short-cuts or separators in Fyne menus, currently.
Because of this, there seems to be no good way to test this automatically for now :(.

Menu short-cuts interfere with Fyne short-cuts. That's why one has to be careful to not use `"c"`, `"x"` or `"v"` as key equivalent for now.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/menu.go
+++ b/menu.go
@@ -16,6 +16,7 @@ func NewMenu(label string, items ...*MenuItem) *Menu {
 type MenuItem struct {
 	Label             string
 	PlaceInNativeMenu bool
+	Separate          bool
 	Action            func()
 }
 
--- a/menu.go
+++ b/menu.go
@@ -3,9 +3,8 @@ package fyne
 // Menu stores the information required for a standard menu.
 // A menu can pop down from a MainMenu or could be a pop out menu.
 type Menu struct {
-	Label            string
-	Items            []*MenuItem
-	AfterNativeMenus bool
+	Label string
+	Items []*MenuItem
 }
 
 // NewMenu creates a new menu given the specified label (to show in a MainMenu) and list of items to display.
@@ -15,10 +14,9 @@ func NewMenu(label string, items ...*MenuItem) *Menu {
 
 // MenuItem is a single item within any menu, it contains a display Label and Action function that is called when tapped.
 type MenuItem struct {
-	IsSeparator       bool
-	Label             string
-	PlaceInNativeMenu bool
-	Action            func()
+	IsSeparator bool
+	Label       string
+	Action      func()
 }
 
 // NewMenuItem creates a new menu item from the passed label and action parameters.
--- a/menu.go
+++ b/menu.go
@@ -17,7 +17,6 @@ func NewMenu(label string, items ...*MenuItem) *Menu {
 type MenuItem struct {
 	IsSeparator       bool
 	Label             string
-	KeyEquivalent     string
 	PlaceInNativeMenu bool
 	Action            func()
 }
--- a/menu.go
+++ b/menu.go
@@ -14,13 +14,14 @@ func NewMenu(label string, items ...*MenuItem) *Menu {
 
 // MenuItem is a single item within any menu, it contains a display Label and Action function that is called when tapped.
 type MenuItem struct {
-	Label  string
-	Action func()
+	Label             string
+	PlaceInNativeMenu bool
+	Action            func()
 }
 
 // NewMenuItem creates a new menu item from the passed label and action parameters.
 func NewMenuItem(label string, action func()) *MenuItem {
-	return &MenuItem{label, action}
+	return &MenuItem{Label: label, Action: action}
 }
 
 // MainMenu defines the data required to show a menu bar (desktop) or other appropriate top level menu.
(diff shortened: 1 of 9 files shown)


menu.go
```diff
@@ -3,24 +3,28 @@ package fyne
...
 
-// MenuItem is a sligne item within any menu, it contains a dispay Label and Action function that is called when tapped.
+// MenuItem is a single item within any menu, it contains a display Label and Action function that is called when tapped.
 type MenuItem struct {
-	Label  string
-	Action func()
+	Label             string
+	KeyEquivalent     string
+	PlaceInNativeMenu bool
+	Separate          bool
```
stuartmscott: This approach is different from the rest of the Fyne API in which separators are implemented as independent objects, eg "ToolbarSeparator". How about defining "MenuItemSeparator", or "MenuSeparator" instead?

andydotxyz:
The shortcuts issue sounds weird. If someone registers the CopyShortcut in a menu can we not somehow see that it fired and execute the appropriate code as if it were called from the keyboard or Entry popup?

menu.go
```diff
@@ -3,24 +3,28 @@ package fyne
...
+	return &Menu{Label: label, Items: items}
 }
 
-// MenuItem is a sligne item within any menu, it contains a dispay Label and Action function that is called when tapped.
+// MenuItem is a single item within any menu, it contains a display Label and Action function that is called when tapped.
 type MenuItem struct {
-	Label  string
-	Action func()
+	Label             string
+	KeyEquivalent     string
```
andydotxyz: This is mac-specific terminology isn't it? It would be much better if we could use our Shortcut types...

menu.go
```diff
@@ -3,24 +3,28 @@ package fyne
...
+	PlaceInNativeMenu bool
```
andydotxyz: "NativeMenu" seems like a term that may not be familiar to people.
How often will it be used? I don't think that mac guidelines support putting lots of stuff in this category, though I can't find the doc at this time.
Isn't it just the "About", "Quit" and the "Preferences" items? In which case perhaps there is a cleaner way where these are special types, so "placeInNative" becomes an internal detail when running on mac in native menu mode?

toaster: I don't think that there is a standard set of app menu entries beside the system entries (About, Services, Hide, Hide Others, Show All, Quit) and Settings. I checked five different apps and they all have different app menu entries and some even have different app menu layout.
I think that we are too inflexible if we restrict the set of menu items that can be put into the native app menu. If a developer doesn't care for menu item layout on MacOS, he simply creates his menus with the default constructor and doesn't think about native menus.

menu.go
toaster: Probably, that was the lazy developer approach :). What about leaving it this way until I do the short-cut support (which then would include Fyne default short-cuts available in menus) which probably would show what is needed here anyhow?

toaster:
> The shortcuts issue sounds weird. If someone registers the CopyShortcut in a menu can we not somehow see that it fired and execute the appropriate code as if it were called from the keyboard or Entry popup?

The developer, that is aware of this current shortcoming, could of course define the menu action to do whatever the Fyne short-cut does.
However, the Copy/Cut/Paste short-cuts are not that trivial because they depend on the current focus.

menu.go
toaster: Yes, you are right.

menu.go
andydotxyz: > If a developer doesn't care for menu item layout on MacOS, he simply creates his menus with the default constructor and doesn't think about native menus.

I disagree with this. We should not expect developers to “care” about macOS, we should do they for them. For example the “Quit” item is automatically in the correct place. If the developer wants a “Preferences” item in their menu then on macOS it should be in the app-menu as that is what the OS expects. As in other places I’d rather stick to the semantics rather than direct configuration. “This is a quit item” vs “this goes at the bottom of native menu if on macOS”.

menu.go
andydotxyz: Or hold back on the shortcut portion of the PR for now?

andydotxyz:
Hmm, not sure about having folk work around menu shortcomings. What about the cut/copy/paste menus being special predefined things as well - afterall they do the same thing in any app, and we could disable automatically if nothing is focused?

toaster:
> Hmm, not sure about having folk work around menu shortcomings. What about the cut/copy/paste menus being special predefined things as well - afterall they do the same thing in any app, and we could disable automatically if nothing is focused?

No need to introduce special things here. I will address the short-cut issue in a clean way. I just wanted to do this in a follow-up PR to keep the PR size small.

menu.go
toaster: I agree with you in this two cases and probably in some others (check for updates, help and about). However, a lot of apps place additional app-specific entries in the app menu. Your approach does not allow this.

menu.go
toaster: ack

menu.go
andydotxyz: Very true. Is it worth looking at the specific cases and trying to figure out an approach when we know more?

PR #773, opened 2020-03-27
Add preferences demo code to fyne_demo

Jacalz:
### Description:
This adds the support for storing the dark or light theme choice between restarts of the demo application.

### Checklist:

### DIFF
--- a/cmd/fyne_demo/main.go
+++ b/cmd/fyne_demo/main.go
@@ -49,9 +49,11 @@ func welcomeScreen(a fyne.App) fyne.CanvasObject {
 			fyne.NewContainerWithLayout(layout.NewGridLayout(2),
 				widget.NewButton("Dark", func() {
 					a.Settings().SetTheme(theme.DarkTheme())
+					a.Preferences().SetString("Theme", "Dark")
 				}),
 				widget.NewButton("Light", func() {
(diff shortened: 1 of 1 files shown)


andydotxyz:
Not sure I agree about this... The preferred theme is already stored in global settings that are managed by fyne_settings.
I agree that this could be cleaner - I don't think we should remove the setting, but we could embed the settings panel (like fynedesk does) so that it will preview the widgets and also store the outcome? A new "settings" tab might make this easier?

Jacalz:
I still think it makes sense to be able to change application themes on a per app basis for cases where the user might want to not use the default, there are cases like that for some apps.

Yeah I think your settings panel idea might be a good option. How about having light, dark and default where default uses the default theme from `fyne_settings`?

andydotxyz:
I agree that some apps need to override theme. I just don't think that fyne_demo is one of them.
If we import the settings panel there is no "default" required, as it would be changing the system config directly.
And having a third option would be confusing anyway IMHO.
That said for mobile/macOS there is an additional option of "system configuration" but we don't support that yet so it's a separate issue...

Jacalz:
I think I have something better planned. Not possibly in the same form but you will see soon enough :)

PR #774, opened 2020-03-27
Test the codebase using Go 1.14.x also

Jacalz:
### Description:
This adds travis matrixes to test using Go version 1.14.x.
I don't think there is any good way to test this other than letting travis test it live here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,10 +31,23 @@ script:
       # check that go mod tidy does not change go.mod/go.sum
       go mod tidy && git diff-index --quiet HEAD -- || ( echo "go.mod/go.sum not up-to-date"; git diff-index HEAD --; false )
     fi
-  - test -z "$(goimports -e -d $GO_FILES | tee /dev/stderr)"
+  - |
+    if [ "$TRAVIS_OS_NAME" != "windows" ]
+    then
+      test -z $(goimports -e -d $GO_FILES | tee /dev/stderr)
+    fi
+# goimports reports a lot of pseudo-diffs on Windows
+#      echo > goimports.out
+#      tail -f goimports.out &
+#      test -z "$(goimports -e -d $GO_FILES | tee goimports.out)"
   - go test -tags ci ./...
   - go vet -tags ci -unsafeptr=false ./...
-  - golint -set_exit_status $(go list -tags ci ./...)
+# Windows golint alarm probably related to https://github.com/golang/lint/issues/157
+  - |
+    if [ "$TRAVIS_OS_NAME" != "windows" ]
+    then
+      golint -set_exit_status $(go list -tags ci ./...)
+    fi
   - gocyclo -over 50 $GO_FILES
   - |
     set -e
@@ -56,9 +69,15 @@ matrix:
     - go: 1.12.x
       os: osx
       env: GO111MODULE=off GOFLAGS=
+    - go: 1.12.x
+      os: windows
+      env: GO111MODULE=off GOFLAGS=
     - go: 1.13.x
       os: linux
       env: GO111MODULE=off GOFLAGS=
     - go: 1.13.x
       os: osx
       env: GO111MODULE=on GOFLAGS=-mod=vendor
+    - go: 1.13.x
+      os: windows
+      env: GO111MODULE=off GOFLAGS=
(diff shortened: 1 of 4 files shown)


Jacalz:
Looks like MacOS is failing over incorrect vendoring:
```bash
The command "if [ "$TRAVIS_OS_NAME" != "windows" ]

then

fi

" exited with 0.

0.03s$ go test -tags ci ./...

go: inconsistent vendoring in /Users/travis/gopath/src/fyne.io/fyne:

	golang.org/x/mobile@v0.0.0-20190719004257-d2bd2a29d028: is marked as replaced by github.com/fyne-io/mobile@v0.0.0-20200218100723-2a24c20a57c6 in vendor/modules.txt, but replaced by github.com/fyne-io/mobile@v0.0.0-20200313204317-b97e3cad9468 in go.mod

run 'go mod vendor' to sync, or use -mod=mod or -mod=readonly to ignore the vendor directory

The command "go test -tags ci ./..." exited with 1.

0.06s$ go vet -tags ci -unsafeptr=false ./...

go: inconsistent vendoring in /Users/travis/gopath/src/fyne.io/fyne:

	golang.org/x/mobile@v0.0.0-20190719004257-d2bd2a29d028: is marked as replaced by github.com/fyne-io/mobile@v0.0.0-20200218100723-2a24c20a57c6 in vendor/modules.txt, but replaced by github.com/fyne-io/mobile@v0.0.0-20200313204317-b97e3cad9468 in go.mod

run 'go mod vendor' to sync, or use -mod=mod or -mod=readonly to ignore the vendor directory

The command "go vet -tags ci -unsafeptr=false ./..." exited with 1.

0.12s$ if [ "$TRAVIS_OS_NAME" != "windows" ]

then

  golint -set_exit_status $(go list -tags ci ./...)

fi

go: inconsistent vendoring in /Users/travis/gopath/src/fyne.io/fyne:

	golang.org/x/mobile@v0.0.0-20190719004257-d2bd2a29d028: is marked as replaced by github.com/fyne-io/mobile@v0.0.0-20200218100723-2a24c20a57c6 in vendor/modules.txt, but replaced by github.com/fyne-io/mobile@v0.0.0-20200313204317-b97e3cad9468 in go.mod


```

andydotxyz:
Are you sure you were up to date? that Feb replace date is not what I see in the develop go.mod...?

Also, I don't think we can handle the delay of so many test runs. To go from 4 to 9 in 1 week is going to slow things a lot.
Perhaps we can switch to just the oldest and newest? (i.e. 1.12 and 1.14)

Jacalz:
Yeah you are right, I think I forgot to push the rebase to my fork.
How about dropping 1.12? Looking at golang.org it looks like only the two latest versions are supported and there is no point in testing with a dead release in my opinion.

Jacalz:
I checked and it looks like the branch is up to date with develop. I removed 1.12.x and it is still failing on MacOS using golang 1.14.x.

andydotxyz:
If we do that then we could accidentally start requiring 1.13 APIs, which feels like it would be a mistake. The reality is that many platforms don't update as fast as GoLang wants them to, so I'd rather we didn't drop support for systems before we have to.

Jacalz:
Well then. How about just leaving the Linux test for 1.12.x? They seem to complete fastest and that gives us the check that we are not using APIs introduced in Go 1.13.x.

andydotxyz:
Sounds like a plan - try that and see how we do on speed

Jacalz:
It looks like the vendor files indeed are out of sync. I have no idea why go1.14.x is the first to report it though :confused: See [module.txt](https://github.com/fyne-io/fyne/blob/053ad610d39f2c65201b184853738d4cb6d6f649/vendor/modules.txt#L42) and [go.mod](https://github.com/fyne-io/fyne/blob/053ad610d39f2c65201b184853738d4cb6d6f649/go.mod#L26), they are indeed out of sync.

Time wise it is not too bad, but not amazing either. Miles better than having all 1.12.x builds running ;)

andydotxyz:
OK, I have updated develop to the latest of all our dependencies.
All tests pass.
If you can base your changes on top of this the issue should go away...

Jacalz:
Managed to wrangle this branch into a borked state. Will open a new PR later today.

PR #780, opened 2020-03-28
Travis: Test with Go 1.14.x

Jacalz:
We should ideally be testing using the two latest supported releases.
However, we still want to keep one 1.12.x running to ensure no 1.13.x APIs are installed just yet.

### Description:
We should ideally be testing using the two latest supported releases.
However, we still want to keep one 1.12.x running to ensure no 1.13.x APIs are installed just yet.

Follow up to #774, due to me managing to put the branch in an unusable state :confused:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,12 +66,6 @@ matrix:
     - go: 1.12.x
       os: linux
       env: GO111MODULE=on GOFLAGS=
-    - go: 1.12.x
-      os: osx
-      env: GO111MODULE=off GOFLAGS=
-    - go: 1.12.x
-      os: windows
-      env: GO111MODULE=off GOFLAGS=
     - go: 1.13.x
       os: linux
       env: GO111MODULE=off GOFLAGS=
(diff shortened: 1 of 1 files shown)


Jacalz:
This very weird. Windows build on 1.13.x and 1.14.x are failing now on stuff that isn't even related to this PR and the commit before this one on develop completed on Windows :confused:

.travis.yml
```diff
@@ -66,12 +66,6 @@ matrix:
     - go: 1.12.x
       os: linux
       env: GO111MODULE=on GOFLAGS=
-    - go: 1.12.x
-      os: osx
-      env: GO111MODULE=off GOFLAGS=
```
andydotxyz: In removing this config we don't test osx without modules any more - I would recommend setting these flags onto the 1.13 version

.travis.yml
```diff
@@ -81,3 +75,12 @@ matrix:
     - go: 1.13.x
       os: windows
       env: GO111MODULE=off GOFLAGS=
+    - go: 1.14.x
+      os: linux
+      env: GO111MODULE=off GOFLAGS=
```
andydotxyz: I'm not sure that this is right - should it not be module=on? It doesn't make sense to test 1.13 and 1.14 with the same config

Jacalz:
The review comments have been fixed and tests are once again not failing :+1:

Issue #478, opened 2019-10-07
TreeView widget


lkarlslund:
It could be awesome if Fyne shipped with a treeview like widget.

For my use case, I'd like to be able to handle dynamically filling entries when a tree node is expanded

andydotxyz:
Duplicate of #254

Issue #206, opened 2019-04-04
Add Accordion Widget


pbarnum:
Accordions are great for displaying Parent/Child data relationships

andydotxyz:
Can you add some use cases or descriptions of how a thing like this may work? There are various examples around the web but I would like to understand what we think it would provide.

pbarnum:
Accordians provide a nice way to display headers with subsequent detail, while also clearing the clutter of too much text on the screen at once. The default functionality should be opening a header to reveal it's information, clicking another header should close the first section while at the same time open the new one. Some options can be provided like having a section opened by default, or keeping multiple sections open at the same time. Helper functions like `GetOpened()` or `OpenAll()/CloseAll()` would be nice.

Some use cases would be like FAQs, Navigation sections, sections and subsections of data about one or many topics. The underlying functionality could also be used to create hidden panels that open and close with a "+" icon.

andydotxyz:
On develop for testing - thanks @stuartmscott

PR #781, opened 2020-03-28
Add Tree Widget

### LINKED ISSUE #254: Tree control
A tree control displays its items in a tree like structure.
Each item has a label and optional an icon. Item may be either collapsed when children are not visible or expanded when children are visible.

stuartmscott:
### Description:
Adds widget for displaying and interacting with hierarchical data, eg. file system.

Fixes #254 #478

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

- [?] Public APIs match existing style. @andydotxyz let's discuss the API and how it will work with the future data binding API.


andydotxyz:
Would it be possible to add a screen capture of this so folk can feedback on the design without cloning the fork?

stuartmscott:
Sure, here's a screen recording
out


The travis failure doesn't appear related to this PR:
> --- FAIL: TestWatchFile (0.20s)
>    settings_test.go:92: File watcher callback was not called

stuartmscott:
The icon changing from open and closed folder needs some work.

stuartmscott:
Here's a better one
out

andydotxyz:
I was just thinking about this and the design and how the closest to a material design for tree may be react (https://material.angular.io/components/tree/overview) and it's quite different to this.

Then it occurred to me that the intent of it's usage may impact this - is it primarily for displaying grouped content, or is it for some side-view selecting categories for example.
I wonder if it's possible to serve both cases with one widget design?

stuartmscott:
I had a play with the react tree and noticed a couple of differences:
- icons: react uses arrows, this PR uses folder/file but this is user-configurable, perhaps I should add another tree to the demo showing a different set of icons?
- touch target: react only opens/closes if you tap on the icon, this PR open/closes if you tap icon, text, or the full width background.
- flat trees: react supports flat trees for scrolling performance, this PR does not and I am not convinced it should.

@andydotxyz any other differences you noticed?

This PR is intended to support the user navigating and selecting from hierarchical data, eg;
- a side panel in an IDE showing the components of the project
- a file open/save dialog showing the file system contents

I think the other use case of displaying grouped content and the react flat tree is typically called an accordion or concertina layout eg #206. I wouldn't mix the two in one widget - it is better to have two widgets that each do one thing well.

andydotxyz:
I think the other big difference is spacing - this feels very condensed (think about use on a phone too). And the indentation seems very subtle.

- icons yes - great to have the option of configuring - perhaps the default could be arrows as that’s more generic?
- touch target - happy with your approach I think, subject to padding notes above (buttons have 2*padding vertically and 4* horizontally)
- not sure what flat trees are, is that like the concertina?

I think overall this is close, your answers make sense. It would probably be good for this and the menu (@toaster is improving just now) to have similar padding / layout which would help with a consistent feel.

The bigger question is whether it should wait for data binding or not? Any widget landed before that may result in breaking changes later on (I expect radio and check will suffer this fate).
That said it could be landed in develop for ongoing work as data api lands, we don’t promise that unreleased APIs won’t change ;)

stuartmscott:
Agree on the spacing note, will add more padding and now that @toaster 's changes are merged I can match the padding.

Will make arrow icons the default.

If data binding will definitely arrive before the next release then sure let's get this in knowing it will be updated before release.

### Tangent
If radio is going to change any way there are some changes I'd like to see;
 - support custom icons
 - support adding to toolbars

The use case is a text editor with a toolbar holding a set of radio buttons controlling text alignment and another set controlling text wrapping. Ie. if I select a radio, rather than the filled circle icon, its background color changes to primary and the other alignments are unselected. I uploaded the icons in a draft PR.

Similar use case could warrant changes to check, eg bold and italic checks in toolbar.

stuartmscott:
Travis failure appears unrelated to this PR:

> --- FAIL: TestWatchFile (0.14s)
    settings_test.go:92: File watcher callback was not called
FAIL

Jacalz:
> Travis failure appears unrelated to this PR:
> 
> > --- FAIL: TestWatchFile (0.14s)
> > settings_test.go:92: File watcher callback was not called
> > FAIL

Yeah it looks like it is related to my PR with some potential security issue fixes. However, this error is so weird and seems to only come from time to time. It is very strange and it isn’t coming up for all the PRs so there is something weird here. Both your PRs fail but others don’t. Is your develop branch fully up to date?

stuartmscott:
Hey @Jacalz this branch "tree" was created from "develop" at a85639c (your commit). I pulled develop again and the only changes "tree" doesn't have are from @toaster and @andydotxyz, what do you think is missing?

Jacalz:
> Hey @Jacalz this branch "tree" was created from "develop" at [a85639c](https://github.com/fyne-io/fyne/commit/a85639c2c5ffba3cbaf962a156b5961b3f3349d4) (your commit). I pulled develop again and the only changes "tree" doesn't have are from @toaster and @andydotxyz, what do you think is missing?

I don’t think it is your fault. It looks like a weird error that creeps up sometime and sometimes not 🤔

andydotxyz:
> * support adding to toolbars

You can add any CanvasObject to a toolbar already, so this isn’t something that we would adapt the widgets for. You just need to add a custom ToolbarItem that wraps the widget.

When it comes to using radio and check for the formatting items I think that to use in a toolbar I think it would make more sense to wrap toolbar buttons in a bit of logic - much less customisation work required ;)

stuartmscott:
> It looks like a weird error that creeps up sometime and sometimes not 🤔

Flakey tests can be hard to track down. Have you seen this locally or could it be related to the Travis environment?

> wrap toolbar buttons in a bit of logic

Yeah that's the approach I was going to take, it unfortunately means there will be duplicate functionality between the Radio and RadioToolbarItem.

stuartmscott:
Is there a way I can re-trigger Travis?

andydotxyz:
Kicked it, hopefully it helps

Jacalz:
@stuartmscott Just pinging you to tell that you have some conflicts in the PR. It looks like a great new widget and I would hope to see it reviewed and landed :)

pbrown12303:
I have been looking at incorporating the tree widget into a graphical diagram editor I am creating (similar in nature to UML, BPMN, or Entity-Relationship diagram editors). In considering the tree widget, I have some thoughts I would like to share and hopefully reach consensus about before attempting any code modifications.

**LEAVES and BRANCHES**
The current distinction between leaves and branches requires a priori knowledge regarding whether a given node is going to be a branch node or a leaf of the tree. Consider an empty folder: at the time it is added, it is a leaf. But the moment a file is added, it becomes a branch node. To me this argues that all nodes should be represented uniformly and that all nodes can potentially have children. This allows the tree to be assembled incrementally without regard for the eventual position of a new node in the tree.

**IDENTIFIERS**
At present, each tree entry appears to be identified by a path of names and (at least at the leaf level) duplicate names appear to be allowed. The implication is that there may not be a unique identifier for any given node. Furthermore, names are typically editable at the application level (e.g. I can change the name of a folder or file). If the editing originates within the tree widget, this may be OK, but if the editing occurs elsewhere and the tree needs to be updated to reflect the change, I can foresee problems identifying the node requiring update. My inclination is to add an optional user-defined identifier to each node and the logic to find a node in the tree using the identifier. 

**ICONS**
I think two concepts have been conflated in the current icon implementation: _open/closed status_ and _nature of the node_. I think they need to be separated, and each may require its own icons. 

The _open/closed status_ is pretty simple: there are at most (I believe) three states here: 1) node with no children; 2) node with children that is presently closed; and 3) node with children that is presently open. I think a default set of icons for these states and the ability for the developer to specify an alternate set of icons (that would be automatically applied) would be useful.

The _nature of the node_, on the other hand, is open-ended (think file types). Here I think the developer should be able to optionally add an icon to each node. I don't think defaults here would be particularly useful.

**POPULATING THE TREE**
Real trees are likely to be much larger that what can practically be rendered in a widget. As such, the widget is only going to display a subset of the actual tree. If this is the case, there needs to be an event protocol that the developer can use to update the widget tree as nodes are opened and closed. A NODE_OPENED event can be used to add the children to the widget tree, and a NODE_CLOSED event can be used to (optionally) remove children from the widget tree. This approach could also benefit from an additional HAS_CHILDREN attribute on a node that controls whether or not the widget presents the ability to open the node.

pbrown12303:
Thanks for the clarification @andydotxyz  - I see I was conflating the old and new designs.

As far as the tree editing is concerned, I don't necessarily think that all the logic has to  be built into the widget - only that it can be accomplished with some reasonable combination of application code and widget code. I think it makes a good thought experiment for evaluating the emerging widget design.

Along those lines, let me suggest the following as a related thought experiment:

_Given_: An application with chunks of data identified by a UUID. Each chunk also has a string associated with it. The data chunks have parent-child relationships between them. The relationships form a tree.

_Application requirements_:

1. Display a tree, labeling the nodes with the strings in the chunks. The strings are NOT unique. 
2. Update the tree structure: The application modifies some of the relationships between the chunks and now needs to update the tree. The old tree structure is not known at this point - only the UUIDs of the chunks and which parent-child relationships have changed. What does the application need to do in order to update the tree? How does it locate the nodes requiring update? A pseudo-code example would be helpful.
3. Update a tree node: The string in the chunk has changed. How does the application update the tree entry. Again, finding the node to edit is the challenge.

With respect to locating nodes, I can see three options:
1) Have a user-defined field in a tree node that can be used for an application-defined identifier and some logic in the tree widget to locate the node with a given value. This is my preferred approach.
2) Have a widget-defined identifier for the node (not dependent upon its position in the tree) and the application maintains a map between the widget identifier and its identifiers. Logic is provided by the widget to locate a node using its identifiers. To me, this is an acceptable approach.
3) The application creates a map  using the addresses of the widget nodes. This approach has many issues. I would consider this only as a last resort.

stuartmscott:
The main distinction between Branches and Leaves as far as Tree Widget is concerned is how they respond to user interaction; when a Branch is tapped, its state will toggle (Open/Closed), whereas when a Leaf is tapped it would trigger an action. For example, in the use case of a Tree displaying a File System hierarchy, tapping a branch would open/close that directory, whereas tapping a leaf may open the file.

One thing to note, which wasn't immediately obvious in my previous comment was that the algorithm is run whenever the tree refreshes, so a node that was once a leaf, could later become a branch - that decision is not made by Tree, but rather by the dev's implementation of the callbacks.

There is also an intrinsic assumption that Branches and Leaves will be rendered differently, and thus callbacks exist to create, update, and differentiate between them. While this assumption may be true in most cases it may not be true for all. We could remove this assumption and instead just have Nodes.

> Unless I am reading the code wrong, the add() function always treats the last item in the path as a leaf.

There will no longer be an Add function - the Tree widget will not store the data, instead the Tree uses the callbacks to create the UI elements to represent the data.

I'm not convinced about using identifiers. The path of a Node within a Tree is unique, whereas the Tree or dev will need to ensure an identifier is unique.

> Application requirements:
> 1. Display a tree, labeling the nodes with the strings in the chunks. The strings are NOT unique.

You'd create a tree where branches and leaves are represented with a widget.Label containing the string. `path []int` would be used for addressing, and the Chunk's string would only be used for rendering.

> 2. Update the tree structure: The application modifies some of the relationships between the chunks and now needs to update the tree. The old tree structure is not known at this point - only the UUIDs of the chunks and which parent-child relationships have changed. What does the application need to do in order to update the tree? How does it locate the nodes requiring update? A pseudo-code example would be helpful.

> 3. Update a tree node: The string in the chunk has changed. How does the application update the tree entry. Again, finding the node to edit is the challenge.

I think there is a misunderstanding here, the dev doesn't update the tree structure. Instead the dev calls tree.Refresh(), the tree then recycles all existing CanvasObjects and traverses the tree again, building a new structure to represent the data that is currently visible in the tree's viewport at that moment.

To address your point about moving nodes, let's revisiting the File System example from earlier. When a user wants to move a file, let's say on a Linux system via the command line, they'd enter something along the lines of `mv /foo/bar /foobar` note the path is the unique identifier of the files, but under the hood that command will resolve the paths to the actual inode. This Tree widget would work in a similar way, using paths to represent the tree structure and the dev is responsible for resolving the path into their chunks of data.

pbrown12303:
Some thoughts:

**Node Appearance**
While I don't think it is a bad assumption that applications will likely want to render interior nodes and leaf nodes differently, I think it equally likely that there may be different types of interior nodes and that these would want to be rendered differently as well. So the most general case requires the application to specify on a node-by-node basis. 

How this can be done will depend upon the Widget strategy. One strategy is for the widget to allow the application to tag each node with a "type" (application-defined string" and the Widget will then prompt (through a callback) the application to supply the appearance definition to be used for that node via whatever mechanism you decide on.

Another strategy is slightly more general, but requires the application to tag each node with an "identifier" (application-defined string). It is then the application's responsibility to determine which appearance definition to use (make its own type determination based on the identifier).

The general question here is what information does the Widget provide to the application when asking for the appearance definition to be used for a particular node.

**Rendering Strategy With Truncated Trees**
I think it is worth examining the rendering strategy in more detail for the use case in which the Widget tree is only a subset of the actual data tree (e.g. the Widget tree does not contain any of the children of closed nodes). In this case walking the tree of the original data will not tell you which nodes are closed (i.e. when to stop populating the Widget tree). The Widget tree is actually defining the subset of the actual data tree to be displayed. The Widget tree is analogous to a database view.

Alternatively, you can require the application to derive its own view (truncated as desired) and then drive the Widget from that data structure. This would be more work on the application side. My preference would be a strategy that allows application-defined identifiers in Widget tree nodes and some parent-child editing operations for the Widget Tree. Of course, any use of these editing operations requires a full rendering of the Widget.

_An aside:_
I strongly urge you to use a more challenging tree example for your thought experiments than file system or URL hierarchies: they are overly simplistic. I would suggest perhaps an organizational chart. Think about the malleability of an organizational structure and the mechanics of what an application would have to do to update the organizational hierarchy as represented in the Widget tree.

andydotxyz:
To display multiple different renderings of otherwise similar nodes is probably a future item of work. 
Similarly in our new List implementation we are asserting that items are homogenous at this time, with the ability to add different types of items in the future once we are happy with the caching and performance of the single style version.

With regards to your point about widget tree being a subset of the data tree I don't understand.
Closed nodes will not be asked for the child data until a user expands to show those children.
The Tree widget is not copying the data or mirroring it, simply asking the developer how to show an item when required.
Expanding and collpasing the tree is handled entirely in the widget and has no impact on the data held within the application.

pbrown12303:
OK. I think I misunderstood your description of the rendering above to mean that the application data is traversed (as opposed to the Widget tree structure). I understand now.

Unfortunately I will not be able to use the tree widget until the ability to depict interior and leaf nodes differently, but I will wait for that. Happy to do reviews, however, as the design evolves.

stuartmscott:
We arrived at the following API, and would love more feedback from the wider Fyne community;
```go
tree := &widget.Tree{
    Children: func(id string) []string {
        // Return a sorted slice of Children IDs for the given Node ID
        // Initially called with "" to get children of root node
    },
    IsBranch: func(id string) bool {
        // Return true if the given ID represents a Branch
    },
    NewNode: func(branch bool) fyne.CanvasObject {
        // Return a CanvasObject that can represent any Branch or Leaf in the Tree
    },
    UpdateNode: func(id string, branch bool, node fyne.CanvasObject) {
        // Update the given Branch CanvasObject to represent the data at the given ID
    },
}
// Optionally
tree.OnNodeSelected = func(id string, branch bool) {
    // Called when the Node with the given ID is selected.
}
tree.OnBranchOpened = func(id string) {
    // Called when a Branch is opened
}
tree.OnBranchClosed = func(id string) {
    // Called when a Branch is closed
}
// Additionally, Tree will have:
IsBranchOpen(id string) bool
OpenBranch(id string)
CloseBranch(id string)
OpenAllBranches()
CloseAllBranches()
SetSelectedNode(id string)
```
And here's a mock of how the Tree will be rendered, again we welcome feedback;
```
Closed Branch:
+---+---------------------------------------+
| > |    dev-defined-branch-canvasobject    |
+---+---------------------------------------+

Opened Branch:
+---+---------------------------------------+
| v |    dev-defined-branch-canvasobject    |
+---+---------------------------------------+
    |    dev-defined-leaf-canvasobject      |
    +---------------------------------------+
    |    dev-defined-leaf-canvasobject      |
    +---------------------------------------+
```

stuartmscott:
Still a lot of work to do, but here's a preview:

Screen Shot 2020-09-03 at 11 23 54 AM

stuartmscott:
After some UI tweaks:
Screen Shot 2020-09-03 at 1 41 00 PM

andydotxyz:
This is looking slick, great work.
I think some of the new public API don't quite fit, some subtle things like we have used "WithStrings" rather than "OfStrings" - should we be consistent here or would it read poorly?
Some other things like "widget.AddPath" don't really seem like they belong, though I get that it's helpful - perhaps such helpers go in the demo app or unexported for now, we don't know much about how people will create the data etc.

stuartmscott:
Travis failure appears unrelated:
```
--- FAIL: TestEntry_Select (0.26s)
    --- FAIL: TestEntry_Select/delete_select_down_with_Shift_still_hold (0.06s)
        util.go:47: 
            	Error Trace:	util.go:47
            	            				util.go:32
            	            				entry_test.go:1324
            	Error:      	Not equal: 
            	Test:       	TestEntry_Select/delete_select_down_with_Shift_still_hold
```

andydotxyz:
I got a demo working locally and love the feel of this - great work!.
Padding / line height is slightly different to List, I wonder if the default (i.e. just label content) should be consistent but I did not look into the difference beyond noticing (not sure if List or Tree is correct - list is a little shorter).

I am not sure about TreeContainer as a name (but could be convinced), mostly I think that containers should be really light on chrome / interaction - but then I remember TabContainer. Tough one...
Also I wondered if the default MinSize was maybe too small?

The code I finally used was:
```go
func makeTree() *widget.TreeContainer {
	data := make(map[string][]string)
	data[""] = []string{"Cars", "Trains"}
	data["Cars"] = []string{"Ford", "Tesla", "Jaguar"}
	data["Trains"] = []string{"Rocket", "TGV", "Shinkansen"}

	 t := widget.NewTreeWithStrings(data)
	 t.OpenBranch("Trains")
	 return t
}
```

Initially I had not understood the need for the top `data[""]` node - let's make sure its clear in the docs :).
And the need for all the keys to be globalle unique - though that would soon become apparant I am sure!

stuartmscott:
> I got a demo working locally and love the feel of this - great work!.

Great news!

> Padding / line height is slightly different to List, I wonder if the default (i.e. just label content) should be consistent but I did not look into the difference beyond noticing (not sure if List or Tree is correct - list is a little shorter).

I agree Tree & List should be consistent in terms of cell height, especially for simple cases such as labels. Previous reviews critiqued the spacing as "condensed" so this add padding around the cell and makes the touch targets larger and easier to hit on mobile. Should this be removed, or should List do the same?

> I am not sure about TreeContainer as a name (but could be convinced), mostly I think that containers should be really light on chrome / interaction - but then I remember TabContainer. Tough one...

I feel the same, I was following suite (AccordionContainer, ScrollContainer, SplitContainer, TabContainer, et al.) but "Tree" would be a perfectly good name (and inline with "List"), as would "Accordion". However Scroll, Split, and Tab don't work as well and are usually suffixed with "er": Scroller, Splitter, Tabber. It would be great to have a consistent naming scheme across the widgets.

> Also I wondered if the default MinSize was maybe too small?

This falls back to ScrollContainer's magic number (32), if Tree should also have a magic number too what should it be, 64?

> Initially I had not understood the need for the top `data[""]` node - let's make sure its clear in the docs :).

Ah yeah, the AddTreePath helper does this for you. Having `data[""]` in the map doesn't apply in all cases (as the dev could override the value of Root), but perhaps it would be sufficient to say the given map must contain a mapping for Root?

> And the need for all the keys to be globalle unique - though that would soon become apparant I am sure!

True, I thought the nomenclature of "ID" was sufficient, but I can add to the docs that IDs must be globally unique.

andydotxyz:
> This falls back to ScrollContainer's magic number (32), if Tree should also have a magic number too what should it be, 64?

ScrollContainer knows nothing about it's content whereas here we do. The list implementation makes space for the template cell (admittedly it is only scrolling vertically) - should we do something similar here?
If not the default size should at least be big enough to display the controls plus a bit - i.e. if we can only go with a constant value (like 32x32) then it should probably be 32+indicator+expansionwidthx32 - but I think we can be smarter.

andydotxyz:
> Ah yeah, the AddTreePath helper does this for you.

True, but I'm still not confinced that `widget.AddTreePath(map[string][]string, ...)` is an API that fits our design. Additionally it only applies to the specific `...WithStrings` helper for tree. As convenient as it is I'm wondering if the string helpers could be considered separately to the main tree PR - and possibly look at ListWithStrings at the same time?

widget/tree.go
```diff
@@ -0,0 +1,751 @@
...
+	"fyne.io/fyne"
+	"fyne.io/fyne/canvas"
+	"fyne.io/fyne/driver/desktop"
+	"fyne.io/fyne/internal/widget"
+	"fyne.io/fyne/storage"
+	"fyne.io/fyne/theme"
+)
+
+const (
+	indentation = 4 // Multiplier of theme.Padding() to indent children by. // TODO consider moving to theme.Indentation()
```
andydotxyz, line 18: Given the recent discussion about alignment I wonder if this should not be something separate but instead theme.IconSize+theme.Padding?

andydotxyz (commented):
I finally got my computer booted back up so here are the comments I had tried to make yesterday before power-outage. Removed items we otherwise discussed.

widget/tree.go
```diff
@@ -0,0 +1,781 @@
...
+	} else {
+		t.OpenBranch(uid)
+	}
+}
+
+// OpenAllBranches opens all branches in the tree.
+func (t *Tree) OpenAllBranches() {
+	t.Walk(func(uid string, branch bool, depth int) {
+		if branch {
+			// TODO this triggers a refresh for each branch
```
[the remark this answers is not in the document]
stuartmscott: This should be resolved

widget/tree.go
```diff
@@ -0,0 +1,781 @@
...
+			if err != nil {
+				fyne.LogError("Unable to list "+luri.String(), err)
+			} else {
+				// TODO sort.Slice(uris, sorter)
+				for _, u := range uris {
+					c = append(c, u.String())
+				}
+			}
+		}
+		//log.Println("FileTree.Children:", uid, c)
```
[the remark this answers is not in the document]
stuartmscott: This should be resolved

widget/tree.go
stuartmscott: Ack. This would tie the child indentation to what the branch contains (the arrow icon + padding)

widget/tree.go
```diff
@@ -0,0 +1,781 @@
...
+		l.SetText(uid)
+		//log.Println("StringTree.Label:", l.Text)
+	}
+	t.ExtendBaseWidget(t)
+	return
+}
+
+// NewTreeWithFiles creates a new tree with the given file system URI.
+// TODO contents of a directory are sorted with the given sorter
+func NewTreeWithFiles(root fyne.URI) (t *Tree) {
```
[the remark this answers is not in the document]
stuartmscott: It doesn't make much difference implementation-wise as it is immediately converted to a string, which `Children` recreates into a URI, and then Listable URI.

It was originally a string parameter, but it was changed to fyne.URI to indicate it can't be just any string. It could be changed to ListableURI so it is obvious to the dev that it should be listable, but that conveys an incorrect requirement - it doesn't have to be listable, a single leaf tree is still a valid state.

widget/tree.go
```diff
@@ -0,0 +1,781 @@
...
+			b = fi.IsDir()
+		}
+		//log.Println("FileTree.IsBranch:", uid, b)
+		return
+	}
+	t.NewNode = func(branch bool) (o fyne.CanvasObject) {
+		//log.Println("FileTree.NewNode:", branch, o)
+		i := NewIcon(theme.FileIcon())
+		l := NewLabel("Name")
+		o = NewHBox(i, l)
```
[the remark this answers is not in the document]
stuartmscott: Ack. Fixed

widget/tree.go
```diff
@@ -0,0 +1,781 @@
...
+		//log.Println("FileTree.Children:", uid, c)
+		return
+	}
+	t.IsBranch = func(uid string) (b bool) {
+		if strings.HasPrefix(uid, "file://") {
+			uid = uid[7:]
+		}
+		fi, err := os.Lstat(uid)
+		if err != nil {
+			fyne.LogError("Unable to stat path "+uid, err)
```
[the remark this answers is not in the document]
stuartmscott: Ack. Fixed.

widget/tree_test.go
```diff
@@ -0,0 +1,531 @@
...
+				leaves = append(leaves, uid)
+			}
+		})
+		assert.Equal(t, 2, len(branches))
+		assert.Equal(t, "", branches[0]) // Root
+		assert.Equal(t, "foo", branches[1])
+		assert.Equal(t, 1, len(leaves))
+		assert.Equal(t, "foobar", leaves[0])
+	})
+	/* TODO Not currently possible as testDriver doesn't support listable URIs:
```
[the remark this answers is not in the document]
stuartmscott: Ack. Will re-merge develop after #1108 lands

andydotxyz (changes requested):
It looks great but the latest version seems quite slow.
Looks like new nodes/leafs are recreated for every refresh/scroll/hover event

andydotxyz:
Thanks, getting closer. hover is slick now. Selecting for some reason creates a leaf and a branch node after updating min sizes no matter what I click on which is a bit strange.

What is most strange though is that scrolling down releases all the elements (according to log) but scrolling up creates them all - so scrolling up you see empty nodes get filled but scrolling down it looks smooth... not sure what is going on here unfortunately.

stuartmscott:
> Selecting for some reason creates a leaf and a branch node after updating min sizes no matter what I click on which is a bit strange.

Selection works differently to hover; whereas hover can just update a bool in the node and refresh just that node, tree supports SetSelectedNode which takes the UID and refreshes the whole tree, then when the node with matching UID is refreshed it will update the indicator accordingly, while all other nodes will render it normally. Can you suggest an improvement? Should Tree support SetHoveredNode?

> What is most strange though is that scrolling down releases all the elements (according to log) but scrolling up creates them all - so scrolling up you see empty nodes get filled but scrolling down it looks smooth... not sure what is going on here unfortunately.

This should be fixed now

cmd/fyne_demo/screens/container.go
```diff
@@ -143,3 +148,34 @@ func makeSplitTab() fyne.CanvasObject {
...
+	l.OnNodeSelected = func(id string, node fyne.CanvasObject) {
+		fmt.Println("LeftTreeNodeSelected:", id)
+	}
+	l.OpenBranch("A")
+	l.OpenBranch("D")
+	l.OpenBranch("E")
+	l.OpenBranch("L")
+	l.OpenBranch("M")
+
+	dir := filepath.Join(os.Getenv("GOPATH"), "src", "fyne.io", "fyne")
```
andydotxyz, line 175: Probably not a good path for a binary that could be run on any computer with or without Go or Fyne installed..

widget/tree.go
```diff
@@ -0,0 +1,792 @@
...
+				r = theme.FolderOpenIcon()
+			} else {
+				// Set folder icon
+				r = theme.FolderIcon()
+			}
+		} else {
+			// Set file icon
+			r = theme.FileIcon()
+
+			/* TODO Copied from dialog/fileicon.go - perhaps this should be a utility?
```
andydotxyz, line 118: Probably just a leftover here?

widget/tree.go
```diff
@@ -0,0 +1,792 @@
...
+				res = theme.FileIcon()
+			}
+			*/
+		}
+		i.SetResource(r)
+		l := c.Objects[1].(*Label)
+		if t.Root == uid {
+			l.SetText(uid)
+		} else {
+			l.SetText(filepath.Base(uid))
```
andydotxyz, line 141: As these are all URIs can you use the URI and get Name() to avoid the dependency on filepath?

widget/tree_test.go
```diff
@@ -0,0 +1,577 @@
+package widget_test
+
+import (
+	//"io/ioutil"
```
andydotxyz, line 4: leftovers here too

widget/tree_test.go
```diff
@@ -0,0 +1,577 @@
...
+		tree.Walk(func(uid string, branch bool, depth int) {
+			if branch {
+				branches = append(branches, uid)
+			} else {
+				leaves = append(leaves, uid)
+			}
+		})
+		assert.Equal(t, 3, len(branches))
+		assert.Equal(t, tempDir, branches[0]) // Root
+		assert.Equal(t, filepath.Join(tempDir, "A"), branches[1])
```
andydotxyz, line 116: When merging the URI Child() feature is probably usable to do this.

widget/tree.go
```diff
@@ -0,0 +1,792 @@
...
+	t.Refresh()
+}
+
+// MinSize returns the size that this widget should not shrink below.
+func (t *Tree) MinSize() fyne.Size {
+	t.ExtendBaseWidget(t)
+	return t.BaseWidget.MinSize()
+}
+
+// Walk visits every open node of the tree and calls the given callback with node Unique ID, whether node is branch, and the depth of node.
```
andydotxyz, line 225: Is this needed for the public API? It feels like an internal detail

widget/tree.go
```diff
@@ -0,0 +1,792 @@
...
+	treeContent *treeContent
+	dividers    []*canvas.Rectangle
+	branches    map[string]*branch
+	leaves      map[string]*leaf
+	branchPool  pool
+	leafPool    pool
+}
+
+func (r *treeContentRenderer) MinSize() (min fyne.Size) {
+	r.treeContent.propertyLock.Lock()
```
andydotxyz, line 357: Does the minsize matter as it's inside the scroll container? This looks like an expensive operation so thought I would ask :)

andydotxyz (changes requested):
Thanks @stuartmscott - apart from the complex Layout functions this is really easy to read :).
Code in general looks good but there are a few specifiecs still to resolve.

widget/tree.go
stuartmscott: Not sure I follow. Doesn't the scroller need to know how big its content is?

widget/tree_test.go
stuartmscott: Sure, need #1108 to land first

widget/tree.go
andydotxyz: I think it will use the max of MinSize and Size.
Changing MinSize has impacts in the driver as it will immediately invalidate caches that would remain if it was just Size changing. (IIRC - this is related to how a window knows if it needs to expand).

stuartmscott: Okay, are you proposing this MinSize return 0,0?

andydotxyz: I wasnt't thinking that.
On reflection though I think my concerns were unfounded, we are inside a scroller so the propagation should not occur.

widget/tree.go
```diff
@@ -0,0 +1,781 @@
...
+				for _, u := range uris {
+					c = append(c, u.String())
+				}
+			}
+		}
+		//log.Println("FileTree.Children:", uid, c)
+		return
+	}
+	t.IsBranch = func(uid string) (b bool) {
+		if strings.HasPrefix(uid, "file://") {
```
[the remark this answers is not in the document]
stuartmscott: Hmm, are you suggesting:
```go
_, err := storage.ListerForURI(storage.NewURI(uid))
return err != nil
```

okratitan (approved):
Thanks for cleaning up the assignment + null checking where unnecessary within the code as we discussed on slack.  Looks great.

This is great and coming through months of hard work.  Can't wait to use it for file browsing stuff in my apps!

widget/tree_test.go
stuartmscott: Done

widget/tree.go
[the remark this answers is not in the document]
andydotxyz: Yeah I guess so. That would work right?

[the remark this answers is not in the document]
stuartmscott: Done

cmd/fyne_demo/screens/container.go
```diff
@@ -156,3 +160,37 @@ func makeTableTab() fyne.CanvasObject {
...
+	}
+	l.OpenBranch("A")
+	l.OpenBranch("D")
+	l.OpenBranch("E")
+	l.OpenBranch("L")
+	l.OpenBranch("M")
+
+	dir, err := os.UserHomeDir()
+	if err != nil {
+		fyne.LogError("Could not get user home dir", err)
```
andydotxyz, line 189: I think this is missing a fallback, it wil try to list "file://" which is not valid

widget/tree.go
```diff
@@ -0,0 +1,791 @@
...
+
+// Tree widget displays hierarchical data.
+// Each node of the tree must be identified by a Unique ID.
+type Tree struct {
+	BaseWidget
+	Root     string
+	Selected string
+	Offset   fyne.Position
+
+	Children       func(uid string) (c []string)                         // Return a sorted slice of Children Unique IDs for the given Node Unique ID
```
andydotxyz: Can we use a name here that better indicates that it refers to IDs and not the actual child nodes or their content?
(discussion in slack but we've not reached any agreement)
Perhaps ChildIDs?

widget/tree.go
```diff
@@ -0,0 +1,791 @@
...
+}
+
+func (n *treeNode) Tapped(*fyne.PointEvent) {
+	n.tree.SetSelectedNode(n.uid)
+	if f := n.tree.OnNodeSelected; f != nil {
+		f(n.uid)
+	}
+}
+
+func (n *treeNode) Update(uid string, depth int) {
```
andydotxyz, line 620: This plus another few internal types have an exported Upate function, I'm not sure it should be - the exported methods above are satisfying an interface.

widget/tree.go
```diff
@@ -0,0 +1,791 @@
...
+	}
+	b.ExtendBaseWidget(b)
+	return
+}
+
+func (b *branch) DoubleTapped(*fyne.PointEvent) {
+	b.tree.ToggleBranch(b.uid)
+}
+
+func (b *branch) Update(uid string, depth int) {
```
andydotxyz, line 732: As before I think this does not need to be exported

widget/tree.go
```diff
@@ -0,0 +1,791 @@
...
+		i.Resource = theme.NavigateNextIcon()
+	}
+	i.Icon.Refresh()
+}
+
+func (i *branchIcon) Tapped(*fyne.PointEvent) {
+	i.tree.ToggleBranch(i.uid)
+}
+
+func (i *branchIcon) Update(uid string, depth int) {
```
andydotxyz, line 771: As above

andydotxyz (changes requested):

andydotxyz (approved):
Great stuff, thanks @stuartmscott

okratitan (approved):
Looks great!

PR #784, opened 2020-03-29
Add Alignment and Wrapping Icons

stuartmscott:
### Description:
Adds icons for;
- Alignment
  - Leading
  - Center
  - Trailing
  - Justified
- Wrapping
  - Off
  - Truncate
  - WrapBreak
  - WrapWord

### Checklist:
- [n/a] Tests included.
- [y] Tests all pass.

#### Where applicable:


andydotxyz:
I was curious about this being a draft PR. Is there more to be added?

stuartmscott:
Its a draft because I'm not quite happy with theme/icons/text-wrap-break.svg and theme/icons/text-wrap-word.svg - the differences in behaviour are slight and so it is hard to convey in an icon.

I also wanted to use them in the text entry demo, but this comes back to radio buttons in toolbars and another change I wanted to make to the svg tool: currently the tool loads an svg and sets the fill color to theme.IconColor() but what I would like to be able to do is colourize the svg with the entire palette. For example 'fill="primary"' would get replaced with theme.PrimaryColor() and 'stroke="icon"' would get replaced by theme.IconColor(). With this change the icons could be colored to show more information while still being in-keeping with the theme, like the boundary line, text-truncate-icon.svg could show the cut off text in red, et al.

Otherwise I'd like to get some feedback on them!

andydotxyz:
Oh, I had assumed you grabbed them from the material resources - https://material.io/resources/icons/?search=align&style=baseline, that is where the rest were downloaded from.

As for the different types of wrap I don't think that is likely something we expose as a user setting - they probably only care about wrap on or off... Truncating is a developer option and the type of wrap is a nuance that I expect is only known to the developer of an application...?

Lastly I don't think that using primary colour on an icon makes sense, as that will disappear when the button has primary state. I don't see a strong reason why these icons need to break from our tradition of single colour material icons.

andydotxyz:
Should we pick this up for the release?

stuartmscott:
Not for 1.3, maybe 2.0 but still more work to be done and would like some more feedback

andydotxyz:
After chat on the call it seems like fyne-x might be the place that this could go.
But no decision made.

AlbinoGeek:
I suppose this didn't make 2.0 -- but that being said, extra icons really could go in `fyne-x` -- until the related feature is 100% implemented in the core framework. When considering this, I would say since word break and alignment are currently not accessible to end-users whatsoever (we have no rich editor, etc), we don't need this in the core repository yet.

E.g.: At the moment, only a develop is concerned with alignment or wrapping, not the end-user.

andydotxyz:
@stuartmscott what do you think, should this move? It would be good to either merge or close in preference of a new home before it hits the 1 year mark :).

PR #785, opened 2020-03-30
Feature/mobilepreferences

### LINKED ISSUE #779: Preferences API not working on mobile
### Describe the bug:
During testing of #773 within #758, I noticed that the Preferences API throws out errors on when built using `-tags mobile` because mobile does not yet have support for this API (look [here](https://github.com/fyne-io/fyne/blob/master/app/settings_mobile.go) for example).

These errors were not getting displayed before, but it  was not working either before, so now with better error handling, they are simply getting displayed for us to see:
```go
2020/03/27 20:00:57 Fyne error:  Preferences load error:
2020/03/27 20:00:57   Cause: mkdir /data: permission denied
2020/03/27 20:00:57   At: /home/jacob/fyne/app/preferences.go:67
2020/03/27 20:01:00 Fyne error:  Failed on saving preferences
2020/03/27 20:01:00   Cause: mkdir /data: permission denied
2020/03/27 20:01:00   At: /home/jacob/fyne/app/preferences.go:96
```

### To Reproduce:
1. Apply the changes from #773 over #758 (or possibly develop when it lands).
2. Build like normal and make sure that the theme choice is saved between starts.
3. Rebuild with `-tags mobile` and look at the error messages in console.

### Example code:
Differential from #773:
image

 - OS: Linux
 - Version: 5.5.11
 - Go version: 1.14.1
 - Fyne version: https://github.com/fyne-io/fyne/pull/758/commits/829a94a8259c609eac25f3bba1fc0193d3c3f022

andydotxyz:
### Description:
Implement the missing preferences system for Android and iOS.
On Android we can use the same code, but with an appropriate storage location.
For iOS we delegate to the NSUserDefaults so that iCloud users can get their config saved to cloud.

Fixes #779

### Checklist:

- [x] Tests included. -> minimal tests on mobile side, but it does fix text failures in mobile sim mode
- [x] Tests all pass.


app/app_mobile_and.go
```diff
@@ -25,8 +25,11 @@ func (app *fyneApp) OpenURL(url *url.URL) error {
 }
 
 func rootConfigDir() string {
-	homeDir := "/data" //, _ := os.UserHomeDir()
+	filesDir := os.Getenv("FILESDIR")
+	if filesDir == "" {
+		log.Println("FILESDIR env was not set by android native code")
+		return "/data/data" // probably won't work, but we can't make a better guess
```
stuartmscott: This path will almost certainly not be readable or writable, but if you can get the package name instead return "data/data/<package>/files/"?

andydotxyz: Indeed, that is what FILESDIR is providing. The fallback is for when it was not possible to look up the package info (or if the packager l didn’t yet support that feature) so we can’t look into the android runtime when falling back...

stuartmscott: Where does FILESDIR get set?

andydotxyz: https://github.com/fyne-io/fyne/commit/e8ddd407ac01dbd772d84039c2183f6a091aa64f#diff-9930725eda019b4f278d281d5d77ab2fR89

stuartmscott: Ah kk, so it is getting set to internalDataPath and if that fails there will be a log entry. Is returning /data/data any better than returning an empty string?

andydotxyz: Well, as it gets appended to its a choice between desperately hoping that /data/data is writeable vs hoping that the file system root is writeable...

stuartmscott: I think you're out of luck either way - but it doesn't matter too much because if a dev encounters an unwritable error they need to realize the issue is not with the unwritable error and instead they need to figure out why the NDK did not set the path, which your log helps them realize

andydotxyz:
I just need a thumbs up from someone with the power :)

app/preferences_ios.go
```diff
@@ -0,0 +1,126 @@
...
+func (p *iOSPreferences) SetBool(key string, value bool) {
+	cKey := C.CString(key)
+	cBool := 0
+	if value {
+		cBool = 1
+	}
+	C.setPreferenceBool(cKey, C.int(cBool))
+
+	C.free(unsafe.Pointer(cKey))
+}
```
toaster, line 57: I propose to use `C.bool` for the bool values. You might need to import AppKit.h on Obj-C side.

app/preferences_ios.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (p *iOSPreferences) BoolWithFallback(key string, fallback bool) bool {
+	cKey := C.CString(key)
+	cBool := 0
+	if fallback {
+		cBool = 1
+	}
+	ret := C.getPreferenceBool(cKey, C.int(cBool))
+
+	C.free(unsafe.Pointer(cKey))
```
toaster, line 44: place the `free` near the `alloc`by using `defer` (applies to all following cKeys, too).
```suggestion
	cKey := C.CString(key)
	defer C.free(unsafe.Pointer(cKey))
	cBool := 0
	if fallback {
		cBool = 1
	}
	ret := C.getPreferenceBool(cKey, C.int(cBool))
```

app/preferences_ios.m
```diff
@@ -0,0 +1,77 @@
+// +build darwin
+// +build ios
+
+#import <Foundation/Foundation.h>
+
+int getPreferenceBool(const char* key, int fallback) {
```
toaster, line 6: Use `bool` for booleans.

app/preferences_ios.m
```diff
@@ -0,0 +1,77 @@
...
+
+const char* getPreferenceString(const char* key, const char* fallback) {
+    NSString *nskey = [NSString stringWithUTF8String:key];
+    NSString *ret = [[NSUserDefaults standardUserDefaults] stringForKey:nskey];
+
+    if (ret == nil) {
+        return fallback;
+    } else {
+        return [ret UTF8String];
+    }
```
toaster, line 70: Could be shorten, couldn't it? Applies to other fallbacks, too.
```suggestion
    return ret ? [ret UTF8String] : fallback
```

app/app_mobile_and_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+func Test_StoragePath(t *testing.T) {
+	oldEnv := os.Getenv("FILESPATH")
+	os.Setenv("FILESPATH", "/tmp")
+
+	p := &preferences{}
+	expected := "/tmp/storage/preferences.json"
+	assert.Equal(t, expected, p.storagePath())
+
+	os.Setenv("FILESPATH", oldEnv)
```
toaster, line 30: use `defer` to place this near to the fetch of `oldEnv`.

app/preferences_ios.go
andydotxyz: Sorry, I tried but objC.BOOL is not C.bool, so it still requires additional conversion.
I will give it another shot... Is the usage of “stdbool.h” truly universal for all CGo targets?
A cursory glance indicates that MS only made it available in 2013

app/preferences_ios.m
andydotxyz: I suppose - I have a general rule of not using ternary if though

toaster: I'm not dogmatic about it.

app/preferences_ios.go
toaster: Who cares about M$ ;). This is MacOS code anyhow.
Maybe `BOOL` and `bool` are interoperable … give it a try.
If not, I still would prefer to have the conversion in the Obj-C code and use `bool` for Go<->ObjC communication.

andydotxyz: Haha, yes this is just for mac aspects - I forgot! Coding in 3 different languages concurrently seems to take it's toll :(

andydotxyz: I think that BOOL is "signed char" whereas bool is an "int", will need some conversion.

app/app_mobile_and_test.go
andydotxyz: cunning :)

andydotxyz:
Great feedback thanks. The C.bool intermediary and also the defers helped clean up the code quite a bit.
There are 2 lines that look like BOOL could be simplified, but it shows the conversion so feels clearer.

app/preferences_ios.go
toaster: I still think it's interoperable. The compiler probably will do the correct thing.

app/preferences_ios.m
```diff
@@ -1,26 +1,23 @@
...
 
     if (obj == nil) {
         return fallback;
     } else {
-        if ([[NSUserDefaults standardUserDefaults] boolForKey:nskey]) {
-            return 1;
-        } else {
-            return 0;
-        }
+        return [[NSUserDefaults standardUserDefaults] boolForKey:nskey] != FALSE;
```
toaster, line 14: Prefer `==` over `!=`.
Prefer `YES`/`NO` over `TRUE`/`FALSE`? Where do these come from?
https://developer.apple.com/documentation/objectivec/objective-c_runtime/boolean_values?language=objc
```suggestion
        return [[NSUserDefaults standardUserDefaults] boolForKey:nskey] == YES;
```

app/preferences_ios.m
```diff
@@ -1,26 +1,23 @@
...
 }
 
-void setPreferenceBool(const char* key, int value) {
+void setPreferenceBool(const char* key, bool value) {
     NSString *nskey = [NSString stringWithUTF8String:key];
-    if (value == 0) {
+    if (value == false) {
         [[NSUserDefaults standardUserDefaults] setBool:FALSE forKey:nskey];
     } else {
         [[NSUserDefaults standardUserDefaults] setBool:TRUE forKey:nskey];
```
toaster: Use `YES`/`NO` instead?

app/preferences_ios.m
andydotxyz: Quite right, thanks. TRUE/FALSE work just as well but are not correct. I think they are a hangover from the old Carbon "Boolean" type, and are designed to be interoperable.

andydotxyz:
Thanks, it's been too long since I did proper iOS development!
Fixed now

PR #788, opened 2020-03-30
Add env command to the fyne tool

lucor:
### Description:

This commit adds the env command to the fyne tool.
It prints the Fyne module and environment information that can be reported into issues to help debugging.

It uses the github.com/lucor/goinfo library to retrieve the information

Usage: fyne env

Output example:
```
## Fyne info
cli_version="(devel)"
go_mod="true" 
go_path="false"
imported="false"
path="/path/to/project/fyne"
version="v1.2.3-254-ga85639c2-dirty"

## Go version info
version="go1.14 linux/amd64"

## OS info
architecture="x86_64"
home_url="https://www.ubuntu.com/"
kernel="Linux 5.3.0-42-generic #34-Ubuntu SMP Fri Feb 28 05:49:40 UTC 2020"
name="Ubuntu"
version="19.10 (Eoan Ermine)"

## Go environment info
CGO_ENABLED="1"
GO111MODULE=""
GOARCH="amd64"
GOOS="linux"
```

Note:
- `path` is the project path. It can be related to a clone of the `fyne.io/fyne` project or a third-party project that uses it as module.
- `go_mod` report if a go.mod file is present into `path` folder
- `go_path` is true if the `fyne.io/fyne` module is located under the GOPATH
- `imported` is true for third-party projects

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
Nice thanks. Should it include GOOS and GOARCH as well? They can be overridden but I guess probably not in the user env?

To meet the overall requirements for an env command I think we wanted to include the version of the “fyne” command itself too - this is important when packaging android as the bootstrap is embedded in the command.

Also please don’t delete vendor packages if you’re not explicitly removing them - macOS needs more than Linux to vendor correctly

lucor:
> Nice thanks. Should it include GOOS and GOARCH as well? They can be overridden but I guess probably not in the user env?

Added the go environment report available by default in `goinfo` that basically reports the output of `go env`. The variables should not contain sensitive data, anyway let me know if we want report only a subset :)

> 
> To meet the overall requirements for an env command I think we wanted to include the version of the “fyne” command itself too - this is important when packaging android as the bootstrap is embedded in the command.

Added also a `fyne version` command.
Note: it requires go module support enabled (GO111MODULE=on)

>
> Also please don’t delete vendor packages if you’re not explicitly removing them - macOS needs more than Linux to vendor correctly

Reverted the deleted packages.

andydotxyz:
> let me know if we want report only a subset :)

Yes, I think that's a lot to scroll through on every report ;) How about "GOOS", "GOARCH", "CGO_ENABLED" and "GO111MODULE"?

Looking great though, thanks. Let's tidy it into the 1.2.4 release and update our bug report template :)

PR #791, opened 2020-03-31
Look at key name for shortcuts

### LINKED ISSUE #790: Shortcuts not properly interpreted on non en_US keyboard
### Describe the bug:
Shortcut keys are based on en_US keyboard layout (qwerty).

When using Crtl+a on an azerty keyboard nothing happens. Ctrl+q, on the other hand, triggers select all, because "q" is on the same place on azerty as "a" would be on qwerty. 

### To Reproduce:
1. Use an azerty keyboard
2. Add an entry widget to your application
3. Type some text
4. Try to select it with Ctrl+a
5. Nothing happens
6. Try to select it with Ctrl+q
7. All text is selected

 - OS: Ubuntu
 - Version: 18.10
 - Go version: 1.12.7
 - Fyne version: 79219c23682e4cec4246dead2d7e4c7370470d21

pieter-biomerieux:
### Description:
Shortcuts where based on key, not on keyname, which causes issues for non-qwerty keyboards.

Fixes #790

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

I am new to fyne development and couldn't find any test for the current method.  I don't really know how to make a test for this.

I couldn't find instructions on lint to use. `gofmt -d internal/driver/glfw/window.go` returned nothing

Some tests in window_test.go fail, but this seems unrelated.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -885,23 +885,23 @@ func (w *window) keyPressed(viewport *glfw.Window, key glfw.Key, scancode int, a
 		ctrlMod = desktop.SuperModifier
 	}
 	if keyDesktopModifier == ctrlMod {
-		switch keyName {
-		case fyne.KeyV:
+		switch glfw.GetKeyName(key, 0) {
+		case "v":
 			// detect paste shortcut
 			shortcut = &fyne.ShortcutPaste{
 				Clipboard: w.Clipboard(),
 			}
-		case fyne.KeyC:
+		case "c":
 			// detect copy shortcut
 			shortcut = &fyne.ShortcutCopy{
 				Clipboard: w.Clipboard(),
 			}
-		case fyne.KeyX:
+		case "x":
 			// detect cut shortcut
 			shortcut = &fyne.ShortcutCut{
 				Clipboard: w.Clipboard(),
 			}
-		case fyne.KeyA:
+		case "a":
 			// detect selectAll shortcut
 			shortcut = &fyne.ShortcutSelectAll{}
 		}
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -830,24 +830,18 @@ var keyCodeMap = map[glfw.Key]fyne.KeyName{
 	glfw.KeyMenu:         desktop.KeyMenu,
 }
 
-func keyToName(code glfw.Key) fyne.KeyName {
-	ret, ok := keyCodeMap[code]
-	if !ok {
-		return ""
-	}
-
-	return ret
-}
 
 func (w *window) keyPressed(viewport *glfw.Window, key glfw.Key, scancode int, action glfw.Action, mods glfw.ModifierKey) {
-	keyName := keyToName(key)
-	if keyName == "" {
+	keyCode, found := keyCodeMap[key]
+	if !found {
 		return
 	}
-	keyEvent := &fyne.KeyEvent{Name: keyName}
+	keyName := glfw.GetKeyName(key, scancode)
+
+	keyEvent := &fyne.KeyEvent{Name: keyCode}
 	keyDesktopModifier := desktopModifier(mods)
 
-	if keyName == fyne.KeyTab {
+	if keyCode == fyne.KeyTab {
 		if keyDesktopModifier == 0 {
 			if action != glfw.Release {
 				w.canvas.focusMgr.FocusNext(w.canvas.focused)
@@ -885,7 +879,7 @@ func (w *window) keyPressed(viewport *glfw.Window, key glfw.Key, scancode int, a
 		ctrlMod = desktop.SuperModifier
 	}
 	if keyDesktopModifier == ctrlMod {
-		switch glfw.GetKeyName(key, 0) {
+		switch keyName {
 		case "v":
 			// detect paste shortcut
 			shortcut = &fyne.ShortcutPaste{
@@ -908,7 +902,7 @@ func (w *window) keyPressed(viewport *glfw.Window, key glfw.Key, scancode int, a
 	}
 	if shortcut == nil && keyDesktopModifier != 0 && keyDesktopModifier != desktop.ShiftModifier {
 		shortcut = &desktop.CustomShortcut{
-			KeyName:  keyName,
+			KeyName:  keyCode,
 			Modifier: keyDesktopModifier,
 		}
 	}


andydotxyz:
Thanks for this. It potentially highlights a bigger issue. Can you please load fyne_demo, go to “Advanced” and the press “Q”. What do you see in the output?
I worry that this fix may hide a bigger issue... Fyne_KeyA should refer to the key with “A” written on it ;)

pieter-biomerieux:
When I press "Q" I get

> Rune: q
> Key: A

> KeyUp: Q
> KeyDown: Q

This is as I expected.

`fyne.KeyA` maps to `glfw.KeyA`, which is `C.GLFW_KEY_A`
As per the GLFW documentation (https://www.glfw.org/docs/latest/group__keys.html)

> The naming of the key codes follow these rules:
> 
> - The US keyboard layout is used

Also https://www.glfw.org/docs/latest/input_guide.html#input_keyboard

> Key events relate to actual physical keyboard keys

Key events are meant for usage in games, so they can be at the physical position you want them to be. This way the WASD control scheme on qwerty nicely translates to a ZQSD control scheme on azerty.

Shortcuts on the other hand mostly refer to the starting letter of the action. So then you need to use the keyname: https://www.glfw.org/docs/latest/input_guide.html#input_key_name. Which you can 't use to act on special keys like Tab, Ctrl ... So you'll always need either the key or the keyname, or both, depending on the use case. There doesn't seem to be a way to just apply the keyboard layout and get on with it.

In `internal/driver/glfw/window.go` you have the poorly named `func keyToName(code glfw.Key) fyne.KeyName`, but that's just a mapping from glfw constants to fyne constants. Not really sure what the use is.

andydotxyz:
> In `internal/driver/glfw/window.go` you have the poorly named `func keyToName(code glfw.Key) fyne.KeyName`, but that's just a mapping from glfw constants to fyne constants. Not really sure what the use is.

Right, I worried that this was the case. I think that the method is well named, but it's content is not working correctly. I don't know what others feel but I think that Fyne.KeyA should refer to the key which generates 'A' rather than a specific keyboard position. Can we move the bug-fix into that method instead?

pieter-biomerieux:
Mapping in advance gets complicated really fast.

On azerty:
The key `C.GLFW_KEY_M` will result in the name `","`
The key `C.GLFW_KEY_SEMICOLON` will result in the name `"m"`
Do you map those or not ?

You can't map one of them, because you would end up with two times `"m"` or two times `","`.

If you map them both, it means you start start mapping symbols too. But in azerty `GLFW_KEY_RIGHT_BRACKET` has the name `"$"` which no button in US English has.

And what do you do with Japanese or Hindi keyboard layouts ?

I think you always have to make the distinction between key by position and keyname and make a deliberate decision every time you use keyboard input.

internal/driver/glfw/window.go
```diff
@@ -830,24 +830,18 @@ var keyCodeMap = map[glfw.Key]fyne.KeyName{
...
-	keyName := keyToName(key)
-	if keyName == "" {
+	keyCode, found := keyCodeMap[key]
+	if !found {
 		return
 	}
-	keyEvent := &fyne.KeyEvent{Name: keyName}
+	keyName := glfw.GetKeyName(key, scancode)
+
+	keyEvent := &fyne.KeyEvent{Name: keyCode}
```
andydotxyz, line 841: Thanks, this is clearer up to this point.
The KeyEvent "Name" should be the name, or we have not fixed the bug for other app developers.
It also does not fix the issue for custom keyboard shortcuts.

pieter-biomerieux: Okay, this goes a bit beyond the scope of what I was trying to solve :P

I would suggest some refactoring:

change `type KeyName string` to `type KeyPosName string`

on `type CustomShortcut struct`:
- change `fyne.KeyName` to `fyne.KeyPosName`
- add `KeyName string`

on `type KeyEvent struct`:
- change `Name KeyName` to `Position KeyPosName`
- add `Name string`

This way, the code written based on the current situation doesn't break (it does change however). And for future use, developers can decide what is needed in their use case.

So, basically the diff below plus all necessary refactoring in the uses of `CustomShortcut` and `KeyEvent`

How should I continue? Both the issue and the pull request titles don't really represent the scope of the change we're talking about.

(By the way, I noticed `CustomShortcut` only supports 1 modifier key, while I use plenty of shortcuts that involve two modifier keys. But that should probably be put in another issue and pull request :P )

```diff
diff --git a/driver/desktop/shortcut.go b/driver/desktop/shortcut.go
index 0cc9da04..bd27fef1 100644
--- a/driver/desktop/shortcut.go
+++ b/driver/desktop/shortcut.go
@@ -12,7 +12,8 @@ var _ fyne.Shortcut = (*CustomShortcut)(nil)
 
 // CustomShortcut describes a shortcut desktop event.
 type CustomShortcut struct {
-	fyne.KeyName
+	fyne.KeyPosName // The name of the position of the key on the US_en keyboard
+	KeyName string  // The name of the key, i.e. the rune it would produce when pressed without modifiers
 	Modifier
 }
 
diff --git a/event.go b/event.go
index ddd62703..13ed617a 100644
--- a/event.go
+++ b/event.go
@@ -2,7 +2,8 @@ package fyne
 
 // KeyEvent describes a keyboard input event.
 type KeyEvent struct {
-	Name KeyName
+	Position KeyPosName // The name of the position of the key on the US_en keyboard
+	Name	 string     // The name of the key, i.e. the rune it would produce when pressed without modifiers
 }
 
 // PointEvent describes a pointer input event. The position is relative to the
diff --git a/internal/driver/glfw/window.go b/internal/driver/glfw/window.go
index 457e6dbc..f6c06f30 100644
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -734,7 +734,7 @@ func convertMouseButton(btn glfw.MouseButton, mods glfw.ModifierKey) (desktop.Mo
 	return button, modifier
 }
 
-var keyCodeMap = map[glfw.Key]fyne.KeyName{
+var keyPosMap = map[glfw.Key]fyne.KeyPosName{
 	// non-printable
 	glfw.KeyEscape:    fyne.KeyEscape,
 	glfw.KeyEnter:     fyne.KeyReturn,
@@ -832,16 +832,19 @@ var keyCodeMap = map[glfw.Key]fyne.KeyName{
 
 
 func (w *window) keyPressed(viewport *glfw.Window, key glfw.Key, scancode int, action glfw.Action, mods glfw.ModifierKey) {
-	keyCode, found := keyCodeMap[key]
+	keyPosName, found := keyPosMap[key]
 	if !found {
 		return
 	}
 	keyName := glfw.GetKeyName(key, scancode)
 
-	keyEvent := &fyne.KeyEvent{Name: keyCode}
+	keyEvent := &fyne.KeyEvent{
+		Position: keyPosName,
+		Name: keyName,
+	}
 	keyDesktopModifier := desktopModifier(mods)
 
-	if keyCode == fyne.KeyTab {
+	if keyPosName == fyne.KeyTab {
 		if keyDesktopModifier == 0 {
 			if action != glfw.Release {
 				w.canvas.focusMgr.FocusNext(w.canvas.focused)
@@ -902,8 +905,9 @@ func (w *window) keyPressed(viewport *glfw.Window, key glfw.Key, scancode int, a
 	}
 	if shortcut == nil && keyDesktopModifier != 0 && keyDesktopModifier != desktop.ShiftModifier {
 		shortcut = &desktop.CustomShortcut{
-			KeyName:  keyCode,
-			Modifier: keyDesktopModifier,
+			KeyPosName: keyPosName,
+			KeyName:    keyName,
+			Modifier:   keyDesktopModifier,
 		}
 	}
 
diff --git a/key.go b/key.go
index a156145e..8a9da580 100644
--- a/key.go
+++ b/key.go
@@ -1,7 +1,7 @@
 package fyne
 
-// KeyName represents the name of a key that has been pressed
-type KeyName string
+// KeyName represents the name of the position of a key on a US_en keyboard
+type KeyPosName string
 
 const (
 	// KeyEscape is the "esc" key
```

andydotxyz: Thanks, I agree that this is a bigger issue than you realised - but I think we should fix it the right way rather than pushing a workaround into the library.

I don't want KeyName to become KeyPosName - these are supposed to describe semantic keys, not the specific position on a keyboard. The "KeyA" could refer to a virtual keyboard which has no concept of positions or hardware actions, so it needs to be a high level concept to work.

I think that the correct fix is that we need to lookup the Fyne key based on resolved key name from GLFW rather than the key code. Close to your original fix, but with a wider scope :)

andydotxyz: > (By the way, I noticed `CustomShortcut` only supports 1 modifier key, while I use plenty of shortcuts that involve two modifier keys. But that should probably be put in another issue and pull request :P )

Good catch. Please open an issue for this and I can fix for 1.2.4 - it's just a mistake in how the consts are defined, the API is OK otherwise I think. (i.e. `ControlModifier | ShiftModifier` should work, but it doesn't).

andydotxyz: I take that back - the modifiers do not have a bug, I misread the code - use 
`ControlModifier | ShiftModifier`.

andydotxyz:
Replaced by #809

PR #792, opened 2020-03-31
Feature: mandatory radio

toaster:
### Description:

This PR adds the ability of a `Radio` to be mandatory, i.e. it requires an option to be selected.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/radio.go
+++ b/widget/radio.go
@@ -279,19 +279,8 @@ func (r *Radio) CreateRenderer() fyne.WidgetRenderer {
 	return &radioRenderer{items, objects, r}
 }
 
-// SetMandatory sets if the radio must have a selection.
-func (r *Radio) SetMandatory(m bool) {
-	r.Mandatory = m
-	if r.Selected == "" && len(r.Options) > 0 {
-		r.Selected = r.Options[0]
-	}
-}
-
 // SetSelected sets the radio option, it can be used to set a default option.
 func (r *Radio) SetSelected(option string) {
-	if option == "" && r.Mandatory && len(r.Options) > 0 {
-		option = r.Options[0]
-	}
 	if r.Selected == option {
 		return
 	}
--- a/widget/radio.go
+++ b/widget/radio.go
@@ -164,11 +164,11 @@ func (r *radioRenderer) Destroy() {
 // Changing the selection (only one can be selected) will trigger the changed func.
 type Radio struct {
 	DisableableWidget
-	Options  []string
-	Selected string
-
-	OnChanged  func(string) `json:"-"`
 	Horizontal bool
+	Mandatory  bool
+	OnChanged  func(string) `json:"-"`
+	Options    []string
+	Selected   string
 
 	hoveredItemIndex int
 	hovered          bool
@@ -238,6 +238,9 @@ func (r *Radio) Tapped(event *fyne.PointEvent) {
 	clicked := r.Options[index]
 
 	if r.Selected == clicked {
+		if r.Mandatory {
+			return
+		}
 		r.Selected = ""
 	} else {
 		r.Selected = clicked
@@ -276,8 +279,19 @@ func (r *Radio) CreateRenderer() fyne.WidgetRenderer {
 	return &radioRenderer{items, objects, r}
 }
 
+// SetMandatory sets if the radio must have a selection.
+func (r *Radio) SetMandatory(m bool) {
+	r.Mandatory = m
+	if r.Selected == "" && len(r.Options) > 0 {
+		r.Selected = r.Options[0]
+	}
+}
+
 // SetSelected sets the radio option, it can be used to set a default option.
 func (r *Radio) SetSelected(option string) {
+	if option == "" && r.Mandatory && len(r.Options) > 0 {
+		option = r.Options[0]
+	}
 	if r.Selected == option {
 		return
 	}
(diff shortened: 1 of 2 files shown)


widget/radio.go
```diff
@@ -276,8 +279,19 @@ func (r *Radio) CreateRenderer() fyne.WidgetRenderer {
 	return &radioRenderer{items, objects, r}
 }
 
+// SetMandatory sets if the radio must have a selection.
+func (r *Radio) SetMandatory(m bool) {
+	r.Mandatory = m
+	if r.Selected == "" && len(r.Options) > 0 {
```
andydotxyz: Would it not be better to return from this and do nothing rather than resetting it to a different value?

widget/radio.go
```diff
@@ -164,11 +164,11 @@ func (r *radioRenderer) Destroy() {
 // Changing the selection (only one can be selected) will trigger the changed func.
 type Radio struct {
 	DisableableWidget
-	Options  []string
-	Selected string
-
-	OnChanged  func(string) `json:"-"`
 	Horizontal bool
+	Mandatory  bool
```
andydotxyz: Isn't it more commonly known as Required?

widget/radio_test.go
```diff
@@ -343,6 +344,39 @@ func TestRadio_FocusIndicator_Centered_Vertically(t *testing.T) {
...
+	radio = NewRadio([]string{"Hi", "There"}, func(string) {})
+	radio.SetSelected("There")
+	radio.SetMandatory(true)
+
+	assert.True(t, radio.Mandatory)
+	assert.Equal(t, "There", radio.Selected, "radio becoming mandatory does not affect a valid selection")
+
+	radio.SetSelected("")
+	assert.True(t, radio.Mandatory)
+	assert.Equal(t, "Hi", radio.Selected, "unsetting selection of mandatory radio selects the first option")
```
andydotxyz: This doesn't feel right. If I have previously made a selection and something triggers a reset then it should either clear, or remain the same. Setting it to something different feels unexpected. So I think this should check for "There".

widget/radio.go
toaster: Then you would not need to have `SetMandatory` at all :).
I wanted the `Radio` to ensure a valid option is selected even if I don't pre-select one.
I use it for dark/light theme selection on my settings page and really don't care which one is pre-selected.

A mandatory `Radio` does not allow to deselect an option. I think it is a pragmatic approach to simply chose one if the `Radio` is made mandatory and no option has been selected. I don't wanted to add form input validation (because I don't need it ;)).

In the light of future input validation it might be better to not pre-select automatically.
Therefore I would remove `SetMandatory` again, okay?

widget/radio.go
toaster: Dunno … I always say “notwendig” 😜. Will rename it.

widget/radio_test.go
toaster: With the removal of `SetMandatory` I will also remove the adjustment of the selection if it is set to `""`.

toaster:
> Will there be follow-on to add a required state for Entry, Check (i.e. "you must select this") and also the opposite for Select (i.e. "(no selection)")? :) (I'm just kidding really, but it might be worth capturing in a ticket that this puts radio ahead of other input in terms of this functionality.

Haha … I don't believe you 😉. While form input validation surely is worth a ticket, I don't think that `Radio.Required` would be part of this. IMO it only defines if the `Radio` allows the user to select no option by clicking on the selected one.

andydotxyz:
Hmm, yes I think most radio required input would be mostly in validation but you’re right we need the widget behaviour side too. (I think it is reflected in Option widget whilst the others perhaps less important.

However I think pre-selecting option 1 is probably not the expected default. If the developer wants one to always be selected they can specify a default option and then it can’t be unselected.

toaster:
> Hmm, yes I think most radio required input would be mostly in validation but you’re right we need the widget behaviour side too. (I think it is reflected in Option widget whilst the others perhaps less important.
> 
> However I think pre-selecting option 1 is probably not the expected default. If the developer wants one to always be selected they can specify a default option and then it can’t be unselected.

Yes, that's why I already removed the pre-selection.

widget/radio.go
andydotxyz: I was worried it was personal preference but when I googled "mandatory input field" every result used the word "required" :)

widget/radio.go
```diff
@@ -279,19 +279,8 @@ func (r *Radio) CreateRenderer() fyne.WidgetRenderer {
...
-func (r *Radio) SetMandatory(m bool) {
-	r.Mandatory = m
-	if r.Selected == "" && len(r.Options) > 0 {
-		r.Selected = r.Options[0]
-	}
-}
-
 // SetSelected sets the radio option, it can be used to set a default option.
 func (r *Radio) SetSelected(option string) {
-	if option == "" && r.Mandatory && len(r.Options) > 0 {
```
andydotxyz: I think removing this completely may have been too far - it now allows programattic setting of an invalid value ("").
Should it perhaps be re-instated without the len(r.Options) check to simply return, thus ensuring that a selected value cannot be unselected?

I guess I don't know for sure - but I wanted to ask the question in case this was an oversight

andydotxyz: (from slack conversation - yes, this is not an oversight)

PR #793, opened 2020-03-31
Refactoring: generic shadowing renderer

toaster:
### Description:

The extraction of a `baseRenderer` similar to the `BaseWidget` allows to consolidate and simplify the appliance of shadows to a widget.

### Checklist:

- ~~[ ] Tests included.~~
- [x] Tests all pass.


widget/widget_renderer.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"fyne.io/fyne"
+	"fyne.io/fyne/theme"
+)
+
+// When using the baseRenderer the embedding renderer should call
+// layoutShadow(contentSize, contentPos) to lay out the shadow.
+type baseRenderer struct {
+	objects []fyne.CanvasObject
+	sh      fyne.CanvasObject
```
stuartmscott: Can we full name this to shadow?

widget/widget_renderer.go
```diff
@@ -0,0 +1,34 @@
...
+func newBaseRenderer(objects []fyne.CanvasObject, level int) *baseRenderer {
+	var shadow fyne.CanvasObject
+	if level > 0 {
+		shadow = newShadow(shadowAround, level*theme.Padding()/2)
+		objects = append(objects, shadow)
+	}
+	return &baseRenderer{objects, shadow}
+}
+
+func (r *baseRenderer) Objects() []fyne.CanvasObject {
```
stuartmscott: Can baseRenderer also implement BackgroundColor and Destroy since most of these implementations are "return theme.BackgroundColor()" and empty respectively?

widget/widget_renderer.go
```diff
@@ -0,0 +1,34 @@
...
+// layoutShadow(contentSize, contentPos) to lay out the shadow.
+type baseRenderer struct {
+	objects []fyne.CanvasObject
+	sh      fyne.CanvasObject
+}
+
+func newBaseRenderer(objects []fyne.CanvasObject, level int) *baseRenderer {
+	var shadow fyne.CanvasObject
+	if level > 0 {
+		shadow = newShadow(shadowAround, level*theme.Padding()/2)
```
stuartmscott: It isn't immediately obvious what level means here. Can this be 'depth' to match newShadow and push this calculation out to the callers?

widget/widget_renderer.go
```diff
@@ -0,0 +1,34 @@
+package widget
```
stuartmscott: There aren't any other _renderer.go files in widget, typically the renderer is in the same file as the widget. I actually prefer it being in a separate file, but for consistency should this be moved into widget.go?

widget/popup.go
```diff
@@ -90,10 +90,9 @@ func (p *PopUp) CreateRenderer() fyne.WidgetRenderer {
 		return &modalPopUpRenderer{popUp: p, bg: bg, objects: []fyne.CanvasObject{bg, p.Content}}
 	}
 
-	shadow := newShadow(shadowAround, theme.Padding()*2)
 	bg := canvas.NewRectangle(theme.BackgroundColor())
-	objects := []fyne.CanvasObject{shadow, bg, p.Content}
-	return &popUpRenderer{popUp: p, shadow: shadow, bg: bg, objects: objects}
+	objects := []fyne.CanvasObject{bg, p.Content}
+	return &popUpRenderer{baseRenderer: newBaseRenderer(objects, 4), popUp: p, bg: bg}
```
stuartmscott: Should this '4' be a constant, or theme.Padding()?

widget/button.go
```diff
@@ -206,16 +191,15 @@ func (b *Button) CreateRenderer() fyne.WidgetRenderer {
 	objects := []fyne.CanvasObject{
 		text,
 	}
-	var shadow *shadow
-	if !b.HideShadow {
-		shadow = newShadow(shadowAround, theme.Padding()/2)
-		objects = append(objects, shadow)
+	shadowLevel := 1
+	if b.HideShadow {
```
stuartmscott: What if HideShadow is changed after the renderer is created, should buttonRenderer.Refresh notify baseRenderer?

widget/widget_renderer.go
stuartmscott: Also, can depth be updated after the renderer is created? Eg. the theme gets updated and now theme.Padding() is different

widget/widget_renderer.go
toaster: Of course! This is an artifact from preparations I made.

widget/widget_renderer.go
toaster: I thought of it as the _level_ of elevation (https://material.io/design/environment/elevation.html) which is then used to compute the _depth_ of the shadow.
The depth depends on the theme while the level doesn't.
I don't want to use the same name for both but I'm open for alternatives.

toaster: Regarding the update: Probably the whole theme dependency should be moved into the shadow which will be refreshed on theme change and can adjust its size, then.
WDYT?

widget/popup.go
toaster: As the level is a kind of elevation it should not depend on the theme.

widget/button.go
toaster: Probably yes.

widget/widget_renderer.go
toaster: I agree that these two  are good candidates for the base renderer. No I think that I chose the wrong name and the renderer should be named `shadowingBaseRenderer` or alike. The simple implementation of common methods would then go into the `baseRenderer` which will be embedded by the `shadowingBaseRenderer`.
WDYT?

widget/widget_renderer.go
toaster: I prefer to have a separate file because the shadowingRenderer is not bound to the baseWidget. Even the proposed baseRenderer is not.
Specific renderers of a specific widget are another kind of story.
However, I would prefer widget specific packages where the widget and the renderer would reside in different files. But this is not only my personal preference but would also be a breaking change.

widget/button.go
toaster: In the previous implementation the shadow could be hidden after the renderer was created but it could not be shown.
To do this right adds a lot of code without a real use case.

I think that shadows will be moved out of the renderers and into the rendering engine itself in the long run.
That's because shadows should be cast by higher elevated elements according to the material design. For this we have to introduce layers into the rendering process and then we are able to cast shadows correctly.
Until then I would leave this as it is now, with the shadow not be hideable/showable after renderer creation.

widget/widget_renderer.go
stuartmscott: How about calling it elevation?

So this would change newShadow to accept elevation and it will calculate the depth using the current theme and respond to theme changes?

widget/widget_renderer.go
stuartmscott: Makes sense. Yeah that would be my preference too

widget/button.go
stuartmscott: > into the rendering engine itself

Yes yes yes

> Until then I would leave this as it is now, with the shadow not be hideable/showable after renderer creation.

Okay for now, maybe file an issue so we don't forget about it

widget/widget_renderer.go
andydotxyz: Elevation seems to be the name that Material uses, but it seems a fairly long name https://material.io/design/environment/light-shadows.html#

widget/popup.go
andydotxyz: I agree that it should not be defined by theme. But I also think a const makes sense to avoid magic numbers. Something like "elevationButton" or similar? Maybe we should reference these elevations? https://storage.googleapis.com/spec-host/mio-staging%2Fmio-design%2F1584058305895%2Fassets%2F0B6xUSjjSulxceF9udnA4Sk5tdU0%2Fbaselineelevation-chart.png

widget/button.go
andydotxyz: In theory that's OK - but it is possible that a renderer is created before widget setup is complete (I mention it out of experience). However if all the windgets that set this to true seem to be working then maybe we can leave it as-is.

widget/popup.go
```diff
@@ -90,10 +90,9 @@ func (p *PopUp) CreateRenderer() fyne.WidgetRenderer {
 		return &modalPopUpRenderer{popUp: p, bg: bg, objects: []fyne.CanvasObject{bg, p.Content}}
 	}
 
-	shadow := newShadow(shadowAround, theme.Padding()*2)
 	bg := canvas.NewRectangle(theme.BackgroundColor())
-	objects := []fyne.CanvasObject{shadow, bg, p.Content}
-	return &popUpRenderer{popUp: p, shadow: shadow, bg: bg, objects: objects}
+	objects := []fyne.CanvasObject{bg, p.Content}
+	return &popUpRenderer{shadowingRenderer: newShadowingRenderer(objects, 4), popUp: p, bg: bg}
```
andydotxyz, line 95: I think this was in the discussion, but use a const for the magic '4'

widget/widget_renderer.go
toaster: That's why I used `level` … it shares 80% of its letters with `elevation` :).

widget/base_renderer.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+func (r *baseRenderer) Destroy() {
+}
+
+func (r *baseRenderer) Objects() []fyne.CanvasObject {
+	return r.objects
+}
+
+func (r *baseRenderer) SetObjects(objects []fyne.CanvasObject) {
```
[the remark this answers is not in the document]
toaster: In this PR it could but my queue is long and unexporting it would just be temporarily.

widget/popup.go
toaster: ack

widget/button.go
toaster: I checked all and didn't found any issues.

widget/button.go
```diff
@@ -206,16 +188,15 @@ func (b *Button) CreateRenderer() fyne.WidgetRenderer {
 	objects := []fyne.CanvasObject{
 		text,
 	}
-	var shadow *shadow
-	if !b.HideShadow {
-		shadow = newShadow(shadowAround, theme.Padding()/2)
-		objects = append(objects, shadow)
+	shadowLevel := baseLevel
```
andydotxyz, line 191: Looking at the diff aren't baseLevel and buttonLevel in this snippet the wrong way round?

PR #797, opened 2020-04-01
Text Size

stuartmscott:
### Description:
Allows labels and hyperlinks to have different font sizes (and colors)

### Checklist:

- [y] Tests included.
- [y] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne_demo/screens/widget.go
+++ b/cmd/fyne_demo/screens/widget.go
@@ -38,6 +38,9 @@ func makeButtonTab() fyne.Widget {
 
 func makeTextTab() fyne.CanvasObject {
 	label := widget.NewLabel("Label")
+	labelSmallPrimary := widget.NewLabel("Small Label")
+	labelSmallPrimary.TextSize = theme.TextSize() * 3 / 4
+	labelSmallPrimary.Color = theme.PrimaryColor()
 
 	link, err := url.Parse("https://fyne.io/")
 	if err != nil {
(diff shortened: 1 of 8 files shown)


andydotxyz:
This is a popular request but honestly I am enclined to reject this, here is why:

Our widgets present a standard user interface (guaranteed) - it aims to make it easy to do great UI and very difficult to make bad decisions.
For people that really want to make custom things like blue text on a grey background there is the 'canvas' api.
Additionally an application's theme can change at any time, according to user preference - and custom colours do not respond well in such an environment.

Sorry. Happy to take contrary opinions but that has been the line we have drawn in the past.

andydotxyz:
After discussion with the team I think this is too much configuration for standard widgets.
Allowing specific colours is very dangerous when themes can be loaded and changed at runtime.
With regards to text size the standard size can change at any time so absolute values could be dangerous.
I suppose it could be interesting to support different "standard sizes" through an enum or similar that multiplies the standard size? A future discussion and/or different PR I would expect.

stuartmscott:
Andy I see your point about making it easy to make a good UI and too much configuration can get in the way.

For simple UIs a single standard text size is fine, but for more complex UIs you still want a single theme, but a theme that encapsulates the different parts of your UI. A good design uses different sizes, shapes, and colours to visually convey importance and group and separate items - eg. heading text should be bigger and bolder than a status label. And while theme currently supports different colours, it does not support different font faces and font sizes.

>I suppose it could be interesting to support different "standard sizes" through an enum or similar that multiplies the standard size? A future discussion and/or different PR I would expect.

That's the direction I initially took - adding theme.TextSizeSmall(). But label, hyperlink, et al. were hardcoded to theme.TextSize() so there needed to be a way to say "I want this label to be small" which is how I arrived with this PR

andydotxyz:
I agree that heading text and highlighting is something we should add - but we should expose that through API rather than configuration.
As you might notice the fyne APIs are semantic based, so "Heading" or "primary text" could make sense in a way that "setColor" doesn't.
Perhaps we could look more at TextStyle?

stuartmscott:
> expose that through API rather than configuration

So you mean add theme.TextSizeHeading() and theme.TextSizePrimary()?

I think the bit I'm missing is how do you get tell a label that you want its text to be theme.TextSizeX()?

You could move this all into TextStyle and setup the label and hyperlink that way, but entry is hardcoded to fyne.TextStyle{}

PR #801, opened 2020-04-01
More security fixes and improvements

Jacalz:
### Description:
This PR adds file-path sanitation using `filepath.Clean()` for files that has user interaction and those that does not are marked with `// #nosec` to make `gosec` ignore them while scanning. I also corrected a place where I missed that there was incorrect folder permissions. This means that we now are down to 20 potential security issues being reported.

I also removed error handling from one of the watcher functions because it was making random errors due to the incorrect cross platform behaviour talked about before.

### Checklist:

- [-] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/preferences.go
+++ b/app/preferences.go
@@ -50,7 +50,7 @@ func (p *preferences) saveToFile(path string) error {
 		if !os.IsExist(err) {
 			return err
 		}
-		file, err = os.Open(filepath.Clean(path))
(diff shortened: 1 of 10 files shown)


andydotxyz:
Not sure I agree about putting workarounds like the fsnotify commit in something entitled "security fixes". Just a note for future work I guess, this looks OK otherwise.

PR #802, opened 2020-04-01
fyne vendor command: exec go mod tidy before go mod vendor

lucor:
### Description:

This commit executes the go mod tidy before the go mod vendor one into the fyne vendor command.
This should help to have go.mod and go.sum aligned with the vendor/modules.txt file and make travis happy :)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
This is a nice enhancement, but by doing it we no longer match the way that `go vendor` works.
Is that something we want to do? I can't see a downside of running tidy first, but I thought I would check.

lucor:
To be honest I was thinking it was not required too but with #788 we had an issue on travis related to a missed tidy after updating a dependency... hence this PR :)
I can't see a downside too, the aim is to help catching edge cases as the above and ensure tidy is always performed.

andydotxyz:
Yes, tricky. It works well for us, as we are developing fyne.
However others who want to vendor fyne now don't have the choice of leaving in old dependencies... is this a bad thing?

lucor:
Not a bad thing, I'd slightly prefer to always apply a go mod tidy to remove the unused deps but I see your point about leaving the possibility to third-party project to leave them there.
What about leaving open for a while for user's feedback ?

In any case changes are really minimal so we'd close and re-apply later if any need arise :-)

andydotxyz:
Given the number of times that contributions have forgotten this step I am becoming tempted to say let's merge it as a best practice...

PR #803, opened 2020-04-02
Modify function to avoid launching a subprocess with variable

Jacalz:
### Description:
The function was always called with the same string everywhere in the code base, so we should be able to just remove the function parameter. We are now down to 19 issues reported by `gosec`.

### Checklist:

- [-] Tests included.
- [x] Tests all pass.

#### Where applicable:

- [-] Public APIs match existing style.


andydotxyz:
Hmm, I agree that this is valie. However this package is a trimmed down version of the gomobile builder - so if we deviate further from their code it will make it harder to pull in patches from upstream.

Jacalz:
> Hmm, I agree that this is valie. However this package is a trimmed down version of the gomobile builder - so if we deviate further from their code it will make it harder to pull in patches from upstream.

I see your point. It is your call though. Feel free to close this if you feel it necessary to do so...

PR #804, opened 2020-04-02
Data Binding API

stuartmscott:
### Description:
Adds support for two-way bindings between widgets and data

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


stuartmscott:
Demo:
out

binding/gen.go
```diff
@@ -0,0 +1,126 @@
...
+type {{ .Name }}Binding struct {
+	BaseBinding
+	Value {{ .Type }}
+}
+
+func (b *{{ .Name }}Binding) Get{{ .Name }}() {{ .Type }} {
+	return b.Value
+}
+
+func (b *{{ .Name }}Binding) Set(value interface{}) {
```
andydotxyz, line 33: Is it possible that this line was typed, avoiding the varying named methods below?

binding/gen.go
```diff
@@ -0,0 +1,126 @@
...
+	ht := template.Must(template.New("header").Parse(headerTemplate))
+	writeHeader(f, ht, []string{
+		"fmt",
+		"fyne.io/fyne",
+		"net/url",
+	})
+
+	ubt := template.Must(template.New("binding").Parse(unitBindingTemplate))
+
+	writeUnitBinding(f, ubt, "Bool", "bool")
```
andydotxyz, line 99: I'm not very familiar with the go generate command - but could this perhaps hook into that?

binding/binding.go
```diff
@@ -0,0 +1,27 @@
+package binding
+
+import "sync"
+
+type UnitBinding interface {
+	addListener(func(interface{}))
+	Set(interface{})
+}
+
+type BaseBinding struct {
```
andydotxyz, line 10: I think we could keep this un-exported - unless we want to support third party bind types? (I think that shouldn't be needed)

binding/button.go
```diff
@@ -0,0 +1,39 @@
+package binding
+
+import (
+	"fyne.io/fyne"
+	"fyne.io/fyne/widget"
+)
+
+func BindButtonTapped(button *widget.Button, data *BoolBinding) {
```
andydotxyz, line 8: Wrapping on-tapped like this is dangerous because it only works if it's called after the custom OnTapped function is set.
I'm not sure that a data binding to a user action anyway - developers can just call data.Set(true) in their OnTapped ;)

binding/check.go
```diff
@@ -0,0 +1,33 @@
+package binding
+
+import (
+	"fyne.io/fyne/widget"
+)
+
+func BindCheckChanged(check *widget.Check, data *BoolBinding) {
+	existing := check.OnChanged
```
andydotxyz, line 8: Where the "main state" is different to something else (like label below) maybe one of them is primary so BindCheck() is the one associated with the main Value? having "Changed" on the end seems like it's binding to an event.

binding/icon.go
```diff
@@ -0,0 +1,14 @@
+package binding
+
+import (
+	"fyne.io/fyne"
+	"fyne.io/fyne/widget"
+)
+
+func BindIconResource(icon *widget.Icon, data *ResourceBinding) {
```
andydotxyz, line 8: Here there is only one type, so even more than above I wonder if it could be BindIcon(...)

binding/button_test.go
```diff
@@ -0,0 +1,63 @@
...
+	a := test.NewApp()
+	defer a.Quit()
+	tapped := false
+	button := widget.NewButton("button", func() {
+		tapped = true
+	})
+	data := &binding.BoolBinding{}
+	binding.BindButtonTapped(button, data)
+	data.Set(true)
+	time.Sleep(time.Second)
```
andydotxyz, line 24: This is a substantial delay - what is it waiting for?
Adding timing in tests is usually a warning flag ;)

cmd/fyne_demo/screens/bindings.go
```diff
@@ -0,0 +1,183 @@
...
+	"fyne.io/fyne/binding"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/theme"
+	"fyne.io/fyne/widget"
+)
+
+// BindingsScreen loads a data bindings example panel for the demo app
+func BindingsScreen() fyne.CanvasObject {
+	// TODO entry
+	// TODO form
```
andydotxyz, line 18: I don't think we have to bind everything - really it's aimed at data display elements isn't it?
form/group/scroller for example surely have no bindable content?

stuartmscott: Sure. (thinking out loud) binding to scroller could be a way to implement the "scroll to cursor" functionality we talked about

binding/button.go
stuartmscott: Would you agree that the reason to use 2-way data binding is so that if the data changes the UI is updated _and_ if the UI is changed the data is updated?

binding/button_test.go
stuartmscott: A very good suggestion Andy - I changed to a chan bool and select w/ timeout, and test execution dropped from 18.280s to 0.388s

binding/button.go
andydotxyz: For elements that can change data, then yes 2 way binding's aim is to update data appropriately.
When discussing this with @steveoc64 I think we came to the conclusion that we can't use public hooks because it causes confusion. If a developer changes OnTapped or OnChanged (which they are entitled to do) this breaks.

Just as I suggested widgets could have a simple .Bind(type) method that calls out to data binding code I don't think it would be a big deal for a widget to call out to databinding "ValueChanged(...)" or similar before it calls the user hooks. That way it does not impact the previous function of callbacks.

binding/check.go
andydotxyz: I guess I'm not sure of the intent for this method, so it's tough to suggest better names. I assume you are not literally binding to OnChanged as that is an event, not a state. Can you provide some context for this?

cmd/fyne_demo/screens/bindings.go
andydotxyz: Hmm, I'm not sure about this. It seems very strange that an item of data setting scroll position would make sense - scrolling is a user controlled activity (for the most part).
As before I think that data binding is about presenting and adjusting data primarily - to venture into UI state is a very different project and would need appropriate planning over and above the simple data binding concept.

andydotxyz:
p.s. I looked at Travis and there are a lot of issues, mostly cosmetic. I know that this is a WiP but you might find it helpful to use our git hooks (https://github.com/fyne-io/fyne/wiki/GitHooks).

stuartmscott:
The Travis error is odd:
> GitHub payload is missing a merge commit (mergeable_state: "draft", merged: false)

andydotxyz:
Where are you seeing the error? I could not find the build...

stuartmscott:
Its on the request page (https://travis-ci.org/github/fyne-io/fyne/requests). As the request was denied, it doesn't appear on the build page.

andydotxyz:
Maybe it is trying not to build draft PRs after all? :)

stuartmscott:
I didn't know Travis was on the call 😛

If that's the case, we need to tell GitHub not to wait for a check that isn't coming

stuartmscott:
It happened again;
> Add Notifiable and Support Listener De-Registration
> GitHub payload is missing a merge commit (mergeable_state: "draft", merged: false)

After some quick searching it appears to occur if the PR can't merge cleanly. I pulled&merged develop and there were many conflicts. I'll fix those and we'll see if that fixes the issue.

stuartmscott:
Grrrr:
>--- FAIL: TestFileWatcher_FileDeleted (0.38s)
    settings_desktop_test.go:87: File watcher callback was not called
FAIL

stuartmscott:
@steveoc64 I like the idea of using init() so that only packages that import binding will start the updater thread

steveoc64:
yeah, its easy to abuse init() - its sort of a red flag in code review ...
but this is a valid case for sure :)

stuartmscott:
After mulling this, I'm on the fence.
Pros;
- updater thread only started if the package is used
- no visible API as nothing about the updater needs to be exported, ie no Start/Stop

Cons;
- the channel is never closed and the thread never ends gracefully. Maybe it is ok that this is killed when the program exists, but it feels better to have the app start the thread when it is created, and stop the thread when it quits.
- things that happen 'automagically' are harder to learn, test, and debug.

steveoc64:
```go
var notifierUpdateChan = make(chan *Binding, 24000)
var notifierQuitChan = make(chan struct{}, 1)

func StopNotifier() {
	notifierQuitChan <- struct{}{}
}

func init() {
	go func() {
		for {
			select {
			case b := <-notifierUpdateChan:
				b.Element.Notify(b)
			case <-notifierQuitChan():
				retur
			}
		}
	}()
}
```

provides a hook to turn it off - could then call that from exit ?

Or add an os signal watcher in here, and terminate the if a SIGTERM or something is seen.

I dont see it as a big drama - its a goroutine, is its only going to be doing work if the rest of the system is in a valid state, and not doing other processing (like catching a panic or calling exit)

stuartmscott:
Yeah that would work, but then the API has StopNotifier. I like the os.Signal route - importing the package starts it, and quitting the program stops it. It is still 'automagical', but it is neat and self-contained.

On semi-related note, binding to address is still not quite right with regards to collections.

While this works for unit bindings:
```
value = 1.0
data := binding.NewFloat64(&value)
...
value = 25.0
data.Update()
```

And this works with list bindings:
```
values := []float64{0.1, 0.2}
data := binding.NewFloat64List(&values)
...
values[1] = 0.4
data.Update()
```

This does not:
```
values := []float64{0.1, 0.2}
data := binding.NewFloat64List(&values)
...
values = append(values, 0.3)
data.Update()
```

Because the binding.BaseList doesn't keep the reference to the original slice. Instead the constructor creates []binding.Binding containing bindings for each of the elements in the original slice. So if an individual element is changed binding.BaseList can see the update, but if the original slice is changed binding.BaseList cannot see the update.

binding.BaseList could instead store both the reference to the original slice, and the slice of element bindings, but this requires type-specific list bindings to be generated like they are for unit bindings.

andydotxyz:
It sounds like an implementation detail and that the api supports this but the code doesn’t. How much more type specific would it need to be than the ‘ binding.NewFloat64List(&values)’ you already indicated?

stuartmscott:
Got it working! Making lists type-specific resulted in bindings.go growing from 426 to 1251 lines and I haven't made maps type-specific yet so it will grow more. Radio and Select are both a little more complex now as they bind to both the list and each individual element, however their tests show how flexible and versatile this implementation is.

stuartmscott:
Travis failure is already filed as #854 / #874:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x12d2d33]
goroutine 268 [running]:
sync.(*RWMutex).RLock(...)
	/Users/travis/.gimme/versions/go1.13.10.darwin.amd64/src/sync/rwmutex.go:48
fyne.io/fyne/test.(*testDriver).CanvasForObject(0x0, 0x14e6140, 0xc00329d140, 0x0, 0x0)
	/Users/travis/gopath/src/fyne.io/fyne/test/testdriver.go:30 +0x33
fyne.io/fyne/canvas.Refresh(0x14e6140, 0xc00329d140)
	/Users/travis/gopath/src/fyne.io/fyne/canvas/base.go:69 +0x81
fyne.io/fyne/widget.(*infProgressRenderer).Refresh(0xc00329d180)
	/Users/travis/gopath/src/fyne.io/fyne/widget/progressbarinfinite.go:91 +0x6c
fyne.io/fyne/widget.(*infProgressRenderer).infiniteProgressLoop(0xc00329d180)
	/Users/travis/gopath/src/fyne.io/fyne/widget/progressbarinfinite.go:115 +0x48
created by fyne.io/fyne/widget.(*infProgressRenderer).start
	/Users/travis/gopath/src/fyne.io/fyne/widget/progressbarinfinite.go:100 +0xe9
FAIL	fyne.io/fyne/widget	3.471s
FAIL
The command "go test -tags ci ./..." exited with 1.
```

andydotxyz:
Sorry, I missed this - have kicked Travis now.

PR #805, opened 2020-04-02
Windows on Windows now respect system scale.

### LINKED ISSUE #635: Scale does not match user configuration in Windows 10

andydotxyz:
### Description:
This PR fixes windows on Windows to use the correct (configured) system scale.
It also updates the window scale as it moves to different screens. Nifty :)

Unfortunately it is technically a breaking change - the old Device API asserted that a system scale is global. I realise now it clearly is not - it's per screen, and a window can move screens. So this needed a parameter added :(.

Fixes #635 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.
- [x] Any breaking changes have a deprecation path or have been discussed. (discussion below, sorry)

Options here:

1) Accept this breaking change, which should have no external impact
2) Add a new method "SystemScaleForWindow" and deprecate the old one...

Option 2 is technically still a breaking change, but would only impact custom driver implementations, of which I am not aware of.

Thoughts?

### DIFF
--- a/internal/driver/gomobile/canvas.go
+++ b/internal/driver/gomobile/canvas.go
@@ -77,7 +77,7 @@ func (c *mobileCanvas) Refresh(obj fyne.CanvasObject) {
 }
 
 func (c *mobileCanvas) edgePadding() (topLeft, bottomRight fyne.Size) {
-	scale := fyne.CurrentDevice().SystemScale(nil) // we don't need a window parameter on mobile
+	scale := fyne.CurrentDevice().SystemScaleForWindow(nil) // we don't need a window parameter on mobile
 
 	return fyne.NewSize(int(float32(c.insetLeft)/scale), int(float32(c.insetTop)/scale)),
 		fyne.NewSize(int(float32(c.insetRight)/scale), int(float32(c.insetBottom)/scale))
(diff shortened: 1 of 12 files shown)


andydotxyz:
Updated to option 2, this is almost non-breaking now. Will only impact custom drivers. Seemss like a fair compromise.

internal/driver/gomobile/canvas.go
```diff
@@ -77,7 +77,7 @@ func (c *mobileCanvas) Refresh(obj fyne.CanvasObject) {
 }
 
 func (c *mobileCanvas) edgePadding() (topLeft, bottomRight fyne.Size) {
-	scale := fyne.CurrentDevice().SystemScale()
```
stuartmscott, line 80: I don't see the benefit of changing this file at all - since mobile isn't affected by the issues with SystemScale, and anyway mobile implementation calls SystemScaleForWindow(nil) underneath

andydotxyz: Because I didn't want to keep the "calling deperecated methods" in the codebase if I could avoid it :)

stuartmscott: Okay, go ahead

PR #809, opened 2020-04-03
Fix key lookup on AZERTY

### LINKED ISSUE #790: Shortcuts not properly interpreted on non en_US keyboard

andydotxyz:
### Description:
Our previous lookup was returning position based key names instead of their actual name.
This broke shortcuts for AZERTY users... but also it was declaring KeyQ but printing 'a', rather confusing.
Change to actual key names, not name encoded position codes

Whils this fixes the issue It does leave un-mapped the following AZERTY keys: ':', 'ù', '$', '^', ')', '<'.
They work fine through rune callbacks but they do not generate key events as we don't currently recognise them as keys. I expect that we could add them, but am I doing the wrong thing?

Fixes #790 

### Checklist:

- [ ] Tests included. - I could not figure out how to simulate AZERTY in glfw keyboard events
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -766,6 +766,10 @@ var keyCodeMap = map[glfw.Key]fyne.KeyName{
 	glfw.KeyHome:      fyne.KeyHome,
 	glfw.KeyEnd:       fyne.KeyEnd,
 
+	glfw.KeySpace:   fyne.KeySpace,
+	glfw.KeyKPEnter: fyne.KeyEnter,
+
+	// functions
 	glfw.KeyF1:  fyne.KeyF1,
 	glfw.KeyF2:  fyne.KeyF2,
 	glfw.KeyF3:  fyne.KeyF3,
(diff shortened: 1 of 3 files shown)


andydotxyz:
Thanks. Regarding numbers unfortunately I tried that first and it doesn’t work. In AZERTY it reports the numbers as “shift” characters and so it doesn’t correctly map unless you use the keycode. It’s almost a hack but it seemed like alt-5 is more likely to be desirable than alt-é (or whatever it was)...

andydotxyz:
FYI it is not just shortcut that is affected but also any widget that listen to desktop.KeyDown/Up events

lucor:
> Thanks. Regarding numbers unfortunately I tried that first and it doesn’t work. In AZERTY it reports the numbers as “shift” characters and so it doesn’t correctly map unless you use the keycode. 

Right, my bad (I was thinking that glfw.Key0 was a string type)

> It’s almost a hack but it seemed like alt-5 is more likely to be desirable than alt-é (or whatever it was)...

I agree with you

PR #815, opened 2020-04-05
Bugfix/toolbar

toaster:
### Description:

Toolbar's `CreateRenderer` added the item's objects to the toolbar's objects even if they already existed there.
Besides the fix for this behaviour, this PR contains a refactoring which moves the object handling of toolbar to its renderer.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/test/testcanvas.go
+++ b/test/testcanvas.go
@@ -174,7 +174,7 @@ func (c *testCanvas) Capture() image.Image {
 // LaidOutObjects returns all fyne.CanvasObject starting at the given fyne.CanvasObject which is laid out previously.
 func LaidOutObjects(o fyne.CanvasObject) (objects []fyne.CanvasObject) {
 	if o != nil {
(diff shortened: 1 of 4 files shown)


test/testcanvas.go
```diff
@@ -183,6 +183,7 @@ func layoutAndCollect(objects []fyne.CanvasObject, o fyne.CanvasObject, size fyn
 	objects = append(objects, o)
 	if w, ok := o.(fyne.Widget); ok {
 		r := w.CreateRenderer()
+		r.Refresh()
```
andydotxyz, line 186: This line seems curious. Why was a renderer needing to be refreshed immediately after being created?

toaster: Because the toolbar renderer was strange. The objs weren't set before a refresh.
I changed this with the following commit but left the refresh here. It can be removed now, I think.

PR #818, opened 2020-04-06
Feature/wasm

Bluebugs:
### Description:

This introduce a new supported platform for Fyne: Webassembly. Technically it should also be close to working with gopherjs, but since it hasn't been update in 2 go version, I haven't bothered with it. This wasm port should be considered to stay beta or technical preview for the time being as go support for wasm isn't very robust. I have encountered a few compiler bugs during my development.

### Checklist:

- [ ] Tests included.
- [X ] Tests all pass.


andydotxyz:
Can you clarify the build tags to use? I see 3 new ones `js`, `wasm`, `web` - which should we be using to build? I can't remember the right commands. perhaps a README_web.md or similar would be in order?

Also there seem to be a lot of deleted vendor files - including glfw, probably a mistake?

cmd/fyne/package.go
```diff
@@ -297,14 +297,80 @@ func (p *packager) packageIOS() error {
...
 func (p *packager) addFlags() {
-	flag.StringVar(&p.os, "os", "", "The operating system to target (android, darwin, ios, linux, windows)")
+	flag.StringVar(&p.os, "os", "", "The operating system to target (android, darwin, ios, linux, windows, js)")
 	flag.StringVar(&p.exe, "executable", "", "The path to the executable, default is the current dir main binary")
 	flag.StringVar(&p.srcDir, "sourceDir", "", "The directory to package, if executable is not set")
 	flag.StringVar(&p.name, "name", "", "The name of the application, default is the executable file name")
 	flag.StringVar(&p.icon, "icon", "Icon.png", "The name of the application icon file")
 	flag.StringVar(&p.appID, "appID", "", "For ios or darwin targets an appID is required, for ios this must \nmatch a valid provisioning profile")
 	flag.BoolVar(&p.release, "release", false, "Should this package be prepared for release? (disable debug etc)")
+	flag.BoolVar(&p.debug, "debug", false, "Define if a debugging package should be generated")
```
andydotxyz, line 373: The other packagers use "release" as !debug (as indicated in the docs).
Adding debug will be a bigger change, to 3 states (debug, normal, release).
Is it true that debug should normally be disabled for development?

andydotxyz:
I see that packaging uses `js` - but is that intuitive? should it not be `web` or something more semantic?

app/app_goxjs.go
```diff
@@ -0,0 +1,108 @@
...
+// +build js wasm web
+
+package app
+
+import (
+	"net/url"
+	"sync"
+	"errors"
+
+        "fyne.io/fyne"
```
andydotxyz, line 13: Go uses tab indenting :)

app/app_js.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"syscall/js"
+)
+
+func (s *settings) Scale() float32 {
+        s.themeLock.RLock()
+        defer s.themeLock.RUnlock()
+	// Get the scale information from the web browser directly
+	return float32(js.Global().Get("devicePixelRatio").Float())
```
andydotxyz, line 16: I think that this is SystemScale not Scale

cmd/fyne/build.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 type builder struct {
-	os, srcdir string
+	os, srcdir, target string
```
andydotxyz, line 11: I cannot see where this new field is set

cmd/fyne/main.go
```diff
@@ -61,6 +61,7 @@ func loadCommands() {
 		{"install", &installer{}},
 		{"vendor", &vendor{}},
 		{"version", &version{}},
+		{"server", &server{}},
```
andydotxyz, line 64: Nice, but should probably be "serve" to match verb based commands ("package", "install")

internal/driver/glfw/device_web.go
```diff
@@ -0,0 +1,35 @@
...
+
+func (*glDevice) IsMobile() bool {
+	if mobileCheck {
+		return true
+	}
+	return false
+}
+
+func (*glDevice) SystemScaleForWindow(w fyne.Window) float32 {
+	return fyne.SettingsScaleAuto
```
andydotxyz, line 34: As mentioned above I think the scale calculations go in here - it's not really auto if we are going to match teh browser config,

internal/driver/glfw/gl_other.go
```diff
@@ -15,5 +15,5 @@ func updateGLContext(w *window) {
 	w.canvas.painter.SetOutputSize(winWidth, winHeight)
 }
 
-func forceWindowRefresh(_ *window) {
+func forceWindowRefresh(w *window) {
```
andydotxyz, line 18: I don't think this is needed

Bluebugs:
> Can you clarify the build tags to use? I see 3 new ones `js`, `wasm`, `web` - which should we be using to build? I can't remember the right commands. perhaps a README_web.md or similar would be in order?

js -> gopherjs
wasm -> webassembly via go
web -> desktop build of the backend via go

At the moment, only wasm is to be really advertised and I included that in fyne build, but I can add some documentation for it in a README with some more explanation.

> Also there seem to be a lot of deleted vendor files - including glfw, probably a mistake?

I will have to check them after the rebase as there is a double include of glfw, there is some change, but it shouldn't remove it.

internal/driver/glfw/loop_desktop.go
```diff
@@ -0,0 +1,90 @@
+// +build !js,!wasm,!web
```
andydotxyz, line 1: Possibly here as well the duplication could be less. The core loop is valid, the specific files perhaps could be just helper files (with functions such as init, poll and stop - the rest is identical I think.

internal/driver/glfw/window_generic.go
```diff
@@ -0,0 +1,530 @@
+package glfw
```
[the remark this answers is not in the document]
andydotxyz: And as above, can the core logic be left intact with platform specific functions where it differs?

internal/painter/gl/gl_core.go
```diff
@@ -40,6 +41,19 @@ func newTexture() Texture {
...
 }
 
+func getTexture(object fyne.CanvasObject, creator func(canvasObject fyne.CanvasObject) Texture) (Texture, error) {
+	texture, ok := textures[object]
+
+	if !ok {
+		texture = creator(object)
+		textures[object] = texture
+	}
+	if texture == NoTexture {
```
andydotxyz, line 51: This does not seem like a change - the code pulled out from gl_common is essentially the same but you are creating an eror here instead of the common code?

internal/scale.go
```diff
@@ -9,6 +9,8 @@ import (
 // ScaleInt converts a fyne coordinate in the given canvas to a screen coordinate
 func ScaleInt(c fyne.Canvas, v int) int {
 	switch c.Scale() {
+	case 0.0:
```
andydotxyz, line 12: Having a 0 scale is an error case - the problem may be deeper in your code.

internal/driver/glfw/window_generic.go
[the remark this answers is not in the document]
Bluebugs: The problem is that the abstraction layer is also named glfw and isn't 1:1 compatible with the native glfw. It will also always be lagging behind. So I kind of choose to live with the more duplicated code and less work to do on upstream JS glfw. Still I agree, there is potential for more refactoring and better naming. Will take care of that.

internal/painter/gl/gl_core.go
Bluebugs: I am actually changing the common code to use an error. The issue is that the wasm backend can not compare texture like that as they are js.Value and == does not work on them /o\. Maybe I should split this change in two step. One modifying core to get an error and then use that change in the web backend.

internal/driver/glfw/window_generic.go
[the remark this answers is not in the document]
andydotxyz: Check out internal/painter/gl where we solve the same problem

internal/painter/gl/gl_core.go
andydotxyz: But your code gets to define NoTexture - can’t a suitable sentinel value be of type js.Value?

Bluebugs: It was possible before with 1.13, but with 1.14 and forward you just can't compare two js.Value without using Equal. The problem is that in the case of a structure with a js.Value inside it, you can not compare the structure itself. I think we could do manual Equal on each members, but then it break portability also. So returning an error did seems like the cleaner solution here.

Bluebugs:
> I see that packaging uses `js` - but is that intuitive? should it not be `web` or something more semantic?

I have choosen `wasm` for the time being.

cmd/fyne/build.go
Bluebugs: The issue is that for web server to recognize a wasm file it needs to be terminated by the '.wasm' extention, but go build will happily generate a executable with no extention that would actually be a webassembly binary. It make it a pain to work with a native and a wasm build at the same time. Also the package command is kind of confused to see a binary that might not be a wasm file with no clue.

Fale:
This is a very interesting feature to me. Is there any way I can help around this?

Bluebugs:
> This is a very interesting feature to me. Is there any way I can help around this?

Actually there is. The current code is running on https://fyne.ddlm.me/ to test fyne_demo. It does generate a few log. If you could try a few different mobile browser (ideally on both Android and iOS, but any will help) and try to get the debug output generated on them it will help.

Most mobile browser require you to do some contortion to get the webdeveloper console output. On a PC, you get the output pretty easily by going in the webdeveloper console menu (Just to check what I am talking about).

Bluebugs:
Will do a new PR when this is ready. Github behavior isn't great in this long standing branch.

andydotxyz:
Most of the support is already in - did you try "fyne serve" recently? https://docs.fyne.io/started/webapp

PR #819, opened 2020-04-06
Add support for removing values in the Preferences API

Jacalz:
### Description:
It may be necessary to be able to delete values stored in the preferences. For example, I am planning of storing the hashed passwords on a per user basis and if a user deletes their account, the hashed password can be deleted too (hashed passwords are only used for comparison and noting is encrypted using the hash).

This PR basically adds that feature and adds the relevant test.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

- [x] Public APIs match existing style (well kind of, but it did not seem appropriate to have a remove function per data type).

### DIFF
--- a/app/preferences_ios.go
+++ b/app/preferences_ios.go
@@ -3,6 +3,7 @@
 package app
 
 import (
+	"errors"
 	"unsafe"
 
 	"fyne.io/fyne"
@@ -104,6 +105,11 @@ func (p *iOSPreferences) SetString(key string, value string) {
 	defer C.free(unsafe.Pointer(cValue))
 
 	C.setPreferenceString(cKey, cValue)
+}
+
+// TODO: Implement the support for RemoveValue on iOS to make it work across all platforms.
+func (p *iOSPreferences) RemoveValue(key string) {
+	fyne.LogError("Calling RemoveValue() on iOS", errors.New("It is currently not supported to use RemoveValue on iOS"))
 }
 
 func newPreferences() *iOSPreferences {
--- a/internal/preferences.go
+++ b/internal/preferences.go
@@ -33,6 +33,13 @@ func (p *InMemoryPreferences) get(key string) (interface{}, bool) {
 	return v, err
 }
 
+func (p *InMemoryPreferences) remove(key string) {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+
+	delete(p.values, key)
+}
+
 // Values provides access to the underlying value map - for internal use only...
 func (p *InMemoryPreferences) Values() map[string]interface{} {
 	p.lock.RLock()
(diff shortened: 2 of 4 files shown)


andydotxyz:
It seems likely that this will not compile for ios or android

Jacalz:
> It seems likely that this will not compile for ios or android

I don't really have a way to test that currently. What is it that might hinder it compiling and if it does, how can we possibly work around it? Would very much like to have this support inside fyne.

andydotxyz:
Well, the mobile driver provides a different preferences, so this PR is missing some code.
Use `fyne package -os ios` will likely complain about missing methods.

Jacalz:
> Well, the mobile driver provides a different preferences, so this PR is missing some code.
> Use `fyne package -os ios` will likely complain about missing methods.

Sorry, but that requires xcode and I think we both can conclude that it won't work on Linux :slightly_frowning_face:

andydotxyz:
Actually I’d disagree with you on that one. You can extract the SDK from an XCode download and use a Mac tool chain for linux. I have Mac builds working here from Linux and iOS is just a few steps further I think https://docs.godotengine.org/en/3.2/development/compiling/cross-compiling_for_ios_on_linux.html.

Jacalz:
Sorry but that does not seem to lead anywhere useful. None of the links that I can found online for downloading the SDK leads to any download page for it. All apple pages just redirect to downloading Xcode for MacOS. The only thing I find that might be usable is https://github.com/tpoechtrager/cctools-port/ and I don't even know if that works with Fyne (or gomobile for that matter). It really looks like a dead end sadly...

andydotxyz:
To get XCode you have to go to Apple, Download the installer .dmg and extract the SDK - more hints are at https://github.com/tpoechtrager/osxcross#packaging-the-sdk (or in the Go GUIs book).

To be clear I was not meaning you should do this, I was just contradicting the suggestion that it was not possible. 

What we need in the meantime is a travis build that checks iOS and Android. I’ll look into that. In the meantime I’ll try and think about the right way to land PRs that don’t support all our platforms. (And/or provide compile errors ;) ).

andydotxyz:
fyi, here's the output if you try to build ios:

```
# fyne.io/fyne/app
../../app/app.go:95:9: cannot use newPreferences() (type *iOSPreferences) as type fyne.Preferences in assignment:
	*iOSPreferences does not implement fyne.Preferences (missing RemoveValue method)
../../app/app.go:97:9: cannot use loadPreferences(id) (type *iOSPreferences) as type fyne.Preferences in assignment:
	*iOSPreferences does not implement fyne.Preferences (missing RemoveValue method)
../../app/preferences_ios.go:31:5: cannot use (*iOSPreferences)(nil) (type *iOSPreferences) as type fyne.Preferences in assignment:
	*iOSPreferences does not implement fyne.Preferences (missing RemoveValue method)
```

Perhaps if there were stubs in place with some suitable "incomplete" marker or error logged it would be OK to merge this as it wouldn't break any builds.

Jacalz:
I see. Will see what I can do about it. Never saw those files before, now I know that :)

I am sorry if I have been sounding negative or rude in my latest comments. It has not been the intention. Every link I clicked more or less felt like they pointed me to out of date documentation...

Jacalz:
I have cooked something now that hopefully solves iOS builds and adds relevant notes :)

andydotxyz:
iOS code added on top of this in `develop`

PR #825, opened 2020-04-07
Wrapping speed has been increased

fralx:
At the moment, the wrapping speed is dramatically low for large texts. This commit increases the wrapping speed by several times.

### DIFF
--- a/widget/text.go
+++ b/widget/text.go
@@ -348,6 +348,7 @@ func lineBounds(text []rune, wrap fyne.TextWrap, maxWidth int, measurer func([]r
 	if maxWidth == 0 || wrap == fyne.TextWrapOff {
 		return lines
 	}
+	maxWidth -= 2 * theme.Padding()
 	var bounds [][2]int
 	for _, l := range lines {
 		low := l[0]
@@ -361,7 +362,6 @@ func lineBounds(text []rune, wrap fyne.TextWrap, maxWidth int, measurer func([]r
 			for {
 				curWidth := measurer(text[low:high])
 				symbWidth := (curWidth / len(text[low:high]))
-				curWidth += symbWidth
 				if maxWidth/symbWidth < len(text[low:high]) {
 					high = low + (maxWidth / symbWidth)
 				}
@@ -376,7 +376,6 @@ func lineBounds(text []rune, wrap fyne.TextWrap, maxWidth int, measurer func([]r
 			for low < high {
 				curWidth := measurer(text[low:high])
 				symbWidth := (curWidth / len(text[low:high]))
-				curWidth += symbWidth
 				if curWidth <= maxWidth {
 					bounds = append(bounds, [2]int{low, high})
 					low = high
@@ -394,7 +393,6 @@ func lineBounds(text []rune, wrap fyne.TextWrap, maxWidth int, measurer func([]r
 				sub := text[low:high]
 				curWidth := measurer(sub)
 				symbWidth := (curWidth / len(sub))
-				curWidth += symbWidth
 				if curWidth <= maxWidth {
 					bounds = append(bounds, [2]int{low, high})
 					low = high
--- a/widget/text.go
+++ b/widget/text.go
@@ -89,7 +89,7 @@ func (t *textProvider) updateRowBounds() {
 	textWrap := t.presenter.textWrap()
 	textStyle := t.presenter.textStyle()
 	textSize := theme.TextSize()
-	maxWidth := t.Size().Width
+	maxWidth := t.Size().Width - 2*theme.Padding()
 
 	t.rowBounds = lineBounds(t.buffer, textWrap, maxWidth, func(text []rune) int {
 		return fyne.MeasureText(string(text), textSize, textStyle).Width
@@ -345,10 +345,9 @@ func splitLines(text []rune) [][2]int {
 // lineBounds returns a slice containing the start and end indicies of each line with the given wrapping applied.
 func lineBounds(text []rune, wrap fyne.TextWrap, maxWidth int, measurer func([]rune) int) [][2]int {
 	lines := splitLines(text)
-	if maxWidth == 0 || wrap == fyne.TextWrapOff {
+	if maxWidth <= 0 || wrap == fyne.TextWrapOff {
 		return lines
 	}
-	maxWidth -= 2 * theme.Padding()
 	var bounds [][2]int
 	for _, l := range lines {
 		low := l[0]
@@ -361,9 +360,11 @@ func lineBounds(text []rune, wrap fyne.TextWrap, maxWidth int, measurer func([]r
 		case fyne.TextTruncate:
 			for {
 				curWidth := measurer(text[low:high])
-				symbWidth := (curWidth / len(text[low:high]))
-				if maxWidth/symbWidth < len(text[low:high]) {
-					high = low + (maxWidth / symbWidth)
+				textLength := high - low
+				symbWidth := curWidth / textLength
+				if maxWidth/symbWidth < textLength {
+					high = low + maxWidth/symbWidth
+					curWidth = measurer(text[low:high])
 				}
 				if curWidth <= maxWidth {
 					bounds = append(bounds, [2]int{low, high})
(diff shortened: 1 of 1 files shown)


fralx:
Some tests failed because I added one space at the end, as the line was cut too close to the borders.

widget/text.go
```diff
@@ -359,7 +359,13 @@ func lineBounds(text []rune, wrap fyne.TextWrap, maxWidth int, measurer func([]r
 		switch wrap {
 		case fyne.TextTruncate:
 			for {
-				if measurer(text[low:high]) <= maxWidth {
+				curWidth := measurer(text[low:high])
+				symbWidth := (curWidth / len(text[low:high]))
+				curWidth += symbWidth
```
stuartmscott, line 364: Rather than adding an arbitrary (text containing only 'i's, 'l's, or 'j's will give a smaller symbWidth than text containing only 'm's, or 'w's) symbol width, you could reduce maxWidth. Currently 'maxWidth := t.Size().Width' but it should probably be 'maxWidth := t.Size().Width - 2*theme.Padding()'

fralx: Hi @stuartmscott, thanks for your reply!
>   Any data to back up this claim? The PR doesn't reduce the big O complexity so I can't see where your speed increase comes from

First of all, It is unfortunately visible to the eye.
Then I have made some tests, I just simply calculated loops count and got next results:

fyne.TextTruncate
Exists method
Text len: 1409 loops count: 1361 
Text len: 1441 loops count: 1393 
Text len: 1473 loops count: 1425
New method
Text len: 1409 loops count: 3 
Text len: 1441 loops count: 3 
Text len: 1473 loops count: 3

fyne.TextWrapBreak
Exists method
Text len: 1441 loops count: 17632 
Text len: 1505 loops count: 19253 
Text len: 1537 loops count: 20085
New method
Text len: 1441 loops count: 243 
Text len: 1473 loops count: 250 
Text len: 1473 loops count: 250

fyne.TextWrapWord
Exists method
Text len: 1377 loops count: 1676 
Text len: 1441 loops count: 1837 
Text len: 1473 loops count: 1918
New method
Text len: 1441 loops count: 82 
Text len: 1441 loops count: 82 
Text len: 1473 loops count: 82

> you could reduce maxWidth

Totally agree with you :) I have made the changes.

widget/text.go
```diff
@@ -348,6 +348,7 @@ func lineBounds(text []rune, wrap fyne.TextWrap, maxWidth int, measurer func([]r
 	if maxWidth == 0 || wrap == fyne.TextWrapOff {
 		return lines
 	}
+	maxWidth -= 2 * theme.Padding()
```
stuartmscott, line 351: Correct calculation, wrong location - could you move this to the call site? (line 92)

It is better for the caller to pass the unpadded width than hidden behaviour in this helper function because some callers may want padding, others may want double padding, and tests don't care about padding

widget/text.go
```diff
@@ -359,7 +360,12 @@ func lineBounds(text []rune, wrap fyne.TextWrap, maxWidth int, measurer func([]r
 		switch wrap {
 		case fyne.TextTruncate:
 			for {
-				if measurer(text[low:high]) <= maxWidth {
+				curWidth := measurer(text[low:high])
+				symbWidth := (curWidth / len(text[low:high]))
+				if maxWidth/symbWidth < len(text[low:high]) {
+					high = low + (maxWidth / symbWidth)
```
stuartmscott, line 366: If you change high, shouldn't you re-calculate curWidth?

widget/text.go
```diff
@@ -359,7 +360,12 @@ func lineBounds(text []rune, wrap fyne.TextWrap, maxWidth int, measurer func([]r
 		switch wrap {
 		case fyne.TextTruncate:
 			for {
-				if measurer(text[low:high]) <= maxWidth {
+				curWidth := measurer(text[low:high])
+				symbWidth := (curWidth / len(text[low:high]))
```
stuartmscott, line 364: Unnecessary slicing; len(text[low:high]) == high-low

widget/text.go
fralx: Done

widget/text.go
fralx: Done

widget/text.go
fralx: Done

fralx:
> symbWidth will be the width of 'h'. If high was 3, symbWidth will be the average width of 'h', 'e', and 'l'. 'l' is much thinner than 'h' and 'e'. My concern is that if the text is a lot of 'i's, 'j's, and 'l's followed by a lot of 'm's and 'w's your initial calculation may be off.

If calculation is wrong it won't pass the measurer check.  

> Second, this value is getting calculated for every line. It would be better if the caller could pass charWidth as a parameter, like:

I think this won't increase speed radically.

stuartmscott:
Let me try to explain with a practical example;
- text is 'iiiiiimmmmmm" (6 i's and 6 m's)
- width of 'i' is 1
- width of 'm' is 4
- max width is 10
- wrap is TextWrapBreak

```
			for low < high {
				curWidth := measurer(text[low:high])
				symbWidth := curWidth / (high - low)
				if curWidth <= maxWidth {
					bounds = append(bounds, [2]int{low, high})
					low = high
					high = l[1]
				} else {
					if (high - low) <= maxWidth/symbWidth {
						high--
					} else {
						high = low + maxWidth/symbWidth
					}
				}
			}
```

###Iteration 1
- low = 0
- high = 12
- curWidth = 30
- symbWidth = 2
- curWidth <= maxWidth is false
- high-low <= maxWidth/symWidth is (0-12) <= 10/2 is 12 <= 5 is false
- high = 0 + 10/2 is 5

###Iteration 2
- low = 0
- high = 5
- curWidth = 5
- curWidth <= maxWidth is true
- bounds = append(bounds, [2]int{0, 5}) <- this is wrong, should be {0, 7}

andydotxyz:
I think our tests must be missing some more corner case areas because this is an over simplication.

The quickest way to see it is if you open `cmd/fyne_demo/screens/widgets.go` and replace the lorem text:

```
	loremIpsum = `iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiimmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm`
```

Then launch the fyne_demo and look at Widgets -> Text. It looks OK when it's wider but as you get closer to the i/m divide the wrapping goes off by a long distance.

I must say that it's surprising how well the averaging works for most text, but the one above illustrates that it's perhaps not quite solid enough?
Maybe the averaging code could be used to get an approximate location and then it could be measured (backward or forward?) until the exact position is found?

andydotxyz:
> Let me try to explain with a practical example;
> 
> * text is 'iiiiiimmmmmm" (6 i's and 6 m's)

Do you think this could be written up as an additional unit test @stuartmscott?
The point @fralx makes about all tests passing is completely valid - with green lights on the tests it wasn't clear that this calculation fails in some situations?...

fralx:
I have made the forward space check. I think it looks better now.

stuartmscott:
@andydotxyz absolutely agree we need more tests in this area. The example I gave above is extremely contrived - the values in a real test would depend on the textProvider's text size, which is hardcoded to theme.TextSize().

The essence of this text wrapping problem is finding the optimal index within a linear data structure. My initial solution simply started at the end and iterated back until it found the biggest index that fitted, however it is inefficient for strings that are much longer than the width. @fralx's solution first tries with the whole line, then uses the average character width to estimate the index. If this is an overestimate it can iterate again and an underestimate will be addressed with 46ab67a.

However, it occurs to me that the optimal solution to this problem is actually a binary search. First check if the whole line will fit, if not check 1/2 the line, if not check 1/4 of the line, and so on. If you find out 1/2 the line wont fit, but 1/4 of the line will, try 3/8 and so on.

fralx:
> However, it occurs to me that the optimal solution to this problem is actually a binary search. First check if the whole line will fit, if not check 1/2 the line, if not check 1/4 of the line, and so on. If you find out 1/2 the line wont fit, but 1/4 of the line will, try 3/8 and so on.

I have tried to combine the average character width and binary search in the checkForwardSpace function.

stuartmscott:
I appreciate that @fralx , your solution is much more efficient than the previous one (and now we can compare them with #839). I'm just wondering if a pure binary search would be better than a hybrid.

fralx:
I have finished wrapping based on pure binary search should I push it here?

stuartmscott:
Wow that's awesome @fralx ! You could either push it to a new PR so we can compare, or just push it here - up to you :)

I added a test in #840, could you merge develop into your binary search branch and ensure it passes? I also added some benchmarks in #839, but this is not yet merged. Could you copy it locally and run it so we have some data to compare?

andydotxyz:
This has been replaced by #845

PR #832, opened 2020-04-08
Improve url input checking to only get valid urls

Jacalz:
### Description:
This fixes a potential security issues reported by gosec (now down to 19).
We are now actually verifying the url to make sure that it is valid before sending the http request.

Partial fix for #742 

### Checklist:

- [ ] Tests included.
  - Might add more tests for these functions in a future PR.
- [x] Tests all pass.

#### Where applicable:

- [-] Public APIs match existing style.

### DIFF
--- a/resource.go
+++ b/resource.go
@@ -3,6 +3,7 @@ package fyne
 import (
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"path/filepath"
 )
 
@@ -59,7 +60,12 @@ func LoadResourceFromPath(path string) (Resource, error) {
 
 // LoadResourceFromURLString creates a new StaticResource in memory using the body of the specified URL.
 func LoadResourceFromURLString(urlStr string) (Resource, error) {
-	res, err := http.Get(urlStr)
+	u, err := url.ParseRequestURI(urlStr)
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := http.Get(u.String())
 	if err != nil {
 		return nil, err
 	}


pymq:
I am sorry @Jacalz , but it seems like url.Parse is already called under the hood in [request.go#L854](https://github.com/golang/go/blob/master/src/net/http/request.go#L854) so extra parsing -> String() is unnecessary

Jacalz:
> I am sorry @Jacalz , but it seems like url.Parse is already called under the hood in [request.go#L854](https://github.com/golang/go/blob/master/src/net/http/request.go#L854) so extra parsing -> String() is unnecessary

Alright. Everywhere I read online suggested that url strings should be verified this way, but they might not be using http.Get(). Do you have a better suggestion to do this then?

Jacalz:
It looks like url.ParseRequestURI() is much more strict with checking this, but ain’t don’t really know how this would be solved otherwise.

pymq:
> It looks like url.ParseRequestURI() is much more strict with checking this, but ain’t don’t really know how this would be solved otherwise.

You had any issues with how plain `url.Parse` validate urls?
As I see from docs `url.ParseRequestURI` is intended for use on a server side, to process incoming HTTP requests.
So I think `url.Parse` should be enough, as it is used in stdlib.

Jacalz:
Yeah, I guess so. The whole issues is that gosec complains about using http.Get() with a variable and that it apparently something you should avoid 🤷‍♂️

Issue #829, opened 2020-04-08
Rendering issues on a window running on a separate goroutine when resizing


parkerqueen:
### Describe the bug:
If I try to resize a window with a container having multiple canvas objects, many of the container's canvas objects are not rendered. The window is using a Grid Layout of 50 Columns.

If I instead use a layout with a smaller number of columns (say 10), it doesn't break as often (only if you wildly resize the window).

Additionally, this only happens when I run the window as a goroutine `go w.ShowAndRun()`. If I dont, it works perfectly fine.

Lastly, sometimes nothing would render from the get go. Just a blank screen would appear. (See the last screenshot). Again, unlikely to happen in a grid with 10 columns but very likely in case of 50 columns.

### To Reproduce:
Look at the code below.

### Screenshots:
Before:
![f](https://user-images.githubusercontent.com/25511740/78787756-449e5700-79c4-11ea-9c2f-7d90909bb63c.png)

After Resizing:
![f2](https://user-images.githubusercontent.com/25511740/78787776-4c5dfb80-79c4-11ea-8c10-f3150f7e53cd.png)

The Blank Screen:
![f3](https://user-images.githubusercontent.com/25511740/78788504-592f1f00-79c5-11ea-8e1f-653dc4212d70.png)

### Example code:
```go
package main

import (
	"image/color"
	"time"

	"fyne.io/fyne"
	"fyne.io/fyne/app"
	"fyne.io/fyne/canvas"
	"fyne.io/fyne/layout"
)

var guiApp fyne.App
var guiWin fyne.Window

func main() {
	guiApp = app.New()

	guiWin = guiApp.NewWindow("A Window")
	guiWin.Resize(fyne.Size{Width: 1200, Height: 800})

	container := fyne.NewContainerWithLayout(layout.NewGridLayout(50))
	for r := uint(0); r < 50; r++ {
		for c := uint(0); c < 50; c++ {
			container.AddObject(canvas.NewRectangle(color.White))
		}
	}

	guiWin.SetContent(container)
	go guiWin.ShowAndRun()
	time.Sleep(50 * time.Second)
}
```

### Device (please complete the following information):
 - OS: Pop!_OS (A wrapper around Ubuntu honestly)
 - Version: 18.04 LTS
 - Go version: 1.14.1
 - Fyne version: 1.2.3

andydotxyz:
Sorry, but you cannot run Window.ShowAndRun() from a goroutine.
The UI drawing requires to be run in main().

[...]

PR #833, opened 2020-04-08
Panic on bad start

andydotxyz:
### Description:
We regularly get bug reports like #829 that say things don't work well if started from a goroutine.
This is true on Linux, but on macOS it will just crash. We don't support it as graphics must be kicked off on the main goroutine.

This change kills the application with a `panic()` if this occurs.
I considered a `fyne.LogError` but it seems like this is a fundamental issue that should not continue.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/driver.go
+++ b/internal/driver/glfw/driver.go
@@ -4,12 +4,16 @@ package glfw
 
 import (
 	"runtime"
+	"strconv"
+	"strings"
 	"sync"
 
 	"fyne.io/fyne"
 	"fyne.io/fyne/internal/driver"
 	"fyne.io/fyne/internal/painter"
 )
+
+const mainGoroutineID = 1
 
 var canvasMutex sync.RWMutex
 var canvases = make(map[fyne.CanvasObject]fyne.Canvas)
@@ -59,11 +63,20 @@ func (d *gLDriver) Quit() {
 }
 
 func (d *gLDriver) Run() {
-	count := runtime.Callers(4, make([]uintptr, 1))
-	if count == 0 {
-		panic("Run() must be called from main goroutine")
+	if goroutineID() != mainGoroutineID {
+		panic("Run() or ShowAndRun() must be called from main goroutine")
 	}
 	d.runGL()
+}
+
+func goroutineID() int {
+	b := make([]byte, 64)
+	b = b[:runtime.Stack(b, false)]
+	// string format expects "goroutine X [running..."
+	id := strings.Split(strings.TrimSpace(string(b)), " ")[1]
+
+	num, _ := strconv.Atoi(id)
+	return num
 }
 
 // NewGLDriver sets up a new Driver instance implemented using the GLFW Go library and OpenGL bindings.
--- a/internal/driver/glfw/driver.go
+++ b/internal/driver/glfw/driver.go
@@ -3,6 +3,7 @@
 package glfw
 
 import (
+	"runtime"
 	"sync"
 
 	"fyne.io/fyne"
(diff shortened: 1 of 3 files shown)


andydotxyz:
The stack reports whether or not we are on a goroutine rather than launching from main (it's got a shorter history).

As you can see from the linked bug report there is no crash on many systems.
The code is correct except the thread it is running on. Not sure what other approach works.

andydotxyz:
I guess also it’s possible that different goroutines could be on the same thread but we want to know that this call is blocking main. If our stack is too short we know that it’s spawned a new goroutine so can’t be blocking.

toaster:
> The stack reports whether or not we are on a goroutine rather than launching from main (it's got a shorter history).

Sorry, but that's technically wrong. The stack reports the calls up to here. It might be longer than main calls but a method call is not a goroutine and you can get large stacks on the main goroutine, too. You also can get shorter stacks on non-main goroutines. See https://play.golang.org/p/BMqZ4Ngkl_4

toaster:
> I guess also it’s possible that different goroutines could be on the same thread

That's not only possible but almost for sure.

andydotxyz:
I wanted to add a little certainty here. We cannot use thread ID checks, even if we are in C land - they will not help:

```go
package main

/*
#include <pthread.h>
#include <stdint.h>

uint64_t threadID() {
        uint64_t id;
        if (pthread_threadid_np(pthread_self(), &id)) {
                abort();
        }
        return id;
}
 */
import "C"
import (
	"log"
	"time"
)

func main() {
	log.Println("OK = ", C.threadID)
	go func() {
		log.Println("NOT OK = ", C.threadID)
	}()
	time.Sleep(time.Second)
}

```

Output is:

```
2020/04/11 20:07:34 OK =  0x4096670
2020/04/11 20:07:34 NOT OK =  0x4096670
```

So we are limited to doing something on the Go side surely?

andydotxyz:
> The stack reports the calls up to here. It might be longer for goroutine calls than for main calls but a method call is not a goroutine and you can get large stacks on the main goroutine, too

Yes, it is possible for people to work around this, or break this, or just not see it at all.
It is not supposed to be any kind of guarantee - it is aimed at helping newbies not make this common mistake.
People seem to have working code, want to do something more complex so add "go" prefix to the `Run()` or `ShowAndRun()`. It does catch that, and from someone who recently hit this trap they said it would have been helpful.

toaster:
I’m still not convinced that we should try to trick Go here.

Please have a look at https://github.com/golang/go/issues/22770 and https://blog.sgmansfield.com/2015/12/goroutine-ids on this.
If you still insist that checking the thread is not enough (which I disagree with) you should at least use the goroutine ID instead of some random stack size.

BTW: I think that a good documentation is a better way to approach this caveat than tricky hacks in Fyne’s code.

andydotxyz:
After our discussion earlier I tried the C version (`g->goid`) but I could not get it to compile... I think we have to link directly to the go runtime or build... any ideas?

```
/*
#include "runtime.h"

int goroutineID() {
	return g->goid;
}
 */
import "C"
```
gives:
```
fatal error: 'runtime.h' file not found
#include "runtime.h"
         ^~~~~~~~~~~
1 error generated.
```

toaster:
We could use github.com/davecheney/junk/id :).

andydotxyz:
Can you compile it? I still can’t

toaster:
Dunno, have paid work to do :(. I’ll try in the evening.

andydotxyz:
No hurry. It may be using some fancy C include that I don't know about, but I suspect it's just a demo that doesn't compile:
```
$ go get github.com/davecheney/junk/id
package github.com/davecheney/junk/id: C source files not allowed when not using cgo or SWIG: id.c
```

andydotxyz:
Dropped back to the string parse approach, unless folk find a better way that is compilable by an average developer ;)

PR #840, opened 2020-04-09
Add text wrapping test for variable character widths

stuartmscott:
### Description:
Existing tests mocked the text measurement function to avoid a dependency on the font face and size, however this created a hole in the test coverage. This PR intends to close the hole by wrapping text composed of characters with different widths and measuring with fyne.MeasureText at a fixed font size (10), but assumes the default font face will not change.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

- [n/a] Public APIs match existing style.

### DIFF
--- a/widget/text_test.go
+++ b/widget/text_test.go
@@ -650,3 +650,59 @@ func TestText_lineBounds(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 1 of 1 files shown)


andydotxyz:
Thanks, this helps understand the details I think

PR #845, opened 2020-04-10
Binary search wrapping

fralx:


toaster:
Hi @fralx, thank you very much for your contribution which we really appreciate.

Before starting the review: would you mind to clean up your PR a little bit by squashing the commits that fix an earlier mistake into the commits where this mistake was introduced?
This would really help to review this PR.
The overall diff of the PR doesn't look that big. I'm wondering why it takes 9 commits for this. Probably you'll end up with 2 or 3 or even a single commit :).

toaster:
I just saw that this is a follow-up PR on an already discussed PR. Sorry, that was my fault.
However, what I said stays true for _this_ PR :).

andydotxyz:
Thanks for this. Is it possible to run the text benchmark that was pushed to develop so we can compare the two approaches in this PR?
I agree that binary search is faster for the detail as suggesed - but I have a hunch that for certain lengths of text (long!) that the average approach might be a good seed to the calculation to narrow search further.
Of course at this point it's just getting into more optimisation - probably not important given how much faster this is than 'develop'.

fralx:
develop
BenchmarkText_splitLines-12              	   58532	     23049 ns/op	    1008 B/op	       6 allocs/op
BenchmarkText_lineBounds_WrapOff-12      	   53094	     19834 ns/op	    1008 B/op	       6 allocs/op
BenchmarkText_lineBounds_Truncate-12     	       2	 549031400 ns/op	2109503004 B/op	  204551 allocs/op
BenchmarkText_lineBounds_WrapBreak-12    	       1	7931453600 ns/op	33730313368 B/op	 3265282 allocs/op
BenchmarkText_lineBounds_WrapWord-12     	       1	1201068700 ns/op	5166824064 B/op	  500855 allocs/op

average
BenchmarkText_splitLines-12              	   49582	     24910 ns/op	    1008 B/op	       6 allocs/op
BenchmarkText_lineBounds_WrapOff-12      	   59996	     22886 ns/op	    1008 B/op	       6 allocs/op
BenchmarkText_lineBounds_Truncate-12     	      92	  11098461 ns/op	58263499 B/op	    5597 allocs/op
BenchmarkText_lineBounds_WrapBreak-12    	       6	 200511467 ns/op	1082831016 B/op	  104011 allocs/op
BenchmarkText_lineBounds_WrapWord-12     	       6	 185843967 ns/op	994320218 B/op	   95903 allocs/op

binary
BenchmarkText_splitLines-12              	   53094	     20663 ns/op	    1008 B/op	       6 allocs/op
BenchmarkText_lineBounds_WrapOff-12      	   76213	     19617 ns/op	    1008 B/op	       6 allocs/op
BenchmarkText_lineBounds_Truncate-12     	     136	   8721088 ns/op	46580759 B/op	    4457 allocs/op
BenchmarkText_lineBounds_WrapBreak-12    	       6	 180176967 ns/op	944955708 B/op	   90624 allocs/op
BenchmarkText_lineBounds_WrapWord-12     	       6	 175843383 ns/op	928487141 B/op	   89238 allocs/op

andydotxyz:
Thanks that is really impressive. So for the test cases the bunary search wins by about 15% over average approach?
it's impressive that it is reduced compute time by about 80% compared to develop :)

stuartmscott:
Wow @fralx this is amazing, well done! Average was already much quicker than develop, but now you've out done yourself with binary 😃 

Curious that splitLines saw a speed increase too - as it wasn't changed in either PR 😕

stuartmscott:
Consider moving "binarySearch" and "findSpaceIndex" out into top level functions - this will make it easier to test. Also, please can you add tests?

fralx:
Hi all! 
@stuartmscott are you going to use binary search for something else? In another case, I see no reason to move it outside the function. Honestly, I see no reason to test this function separately. But if you insist, I can do it :smiley:

andydotxyz:
When adding functionality we are trying to be good about having tests that cover it, and also to keep the complexity down. Having two embedded functions (binarySearch and findSpaceIndex) vastly increases the complexity of a single method and makes it harder to see the crux of it's intent.

I agree that the binary search (and indeed the findspace helper func) should be moved out to their own functions and tested. It think doing so should improve readability and test coverage.

widget/text_test.go
```diff
@@ -706,3 +706,106 @@ func TestText_lineBounds_variable_char_width(t *testing.T) {
...
+		{
+			name: "case_2",
+			text: "foobar foobar",
+			last: 12,
+			want: 6,
+		},
+		{
+			name: "case_3",
+			text: "ww wwww www",
+			last: 9,
```
stuartmscott, line 794: 10?

widget/text_test.go
```diff
@@ -706,3 +706,106 @@ func TestText_lineBounds_variable_char_width(t *testing.T) {
...
+			want: 7,
+		},
+		{
+			name: "case_4",
+			text: "ww wwww www wwwww",
+			last: 5,
```
stuartmscott, line 800: 16?

widget/text_test.go
```diff
@@ -706,3 +706,106 @@ func TestText_lineBounds_variable_char_width(t *testing.T) {
...
+			want: 2,
```
stuartmscott, line 801: 11?

widget/text_test.go
```diff
@@ -706,3 +706,106 @@ func TestText_lineBounds_variable_char_width(t *testing.T) {
...
+
+func TestText_findSpaceIndex(t *testing.T) {
+	tests := []struct {
+		name string
+		text string
+		last int
+		want int
+	}{
+		{
+			name: "case_1",
```
stuartmscott, line 780: "case_n" is a poor naming scheme, what is it that you are testing?

case_1 is testing that findSpaceIndex returns the fallback if the string has no spaces, so how about "no_space_fallback"?
case_2 is testing that findSpaceIndex returns the correct index, so "single_space"
case_3 is testing that findSpaceIndex returns the last (second) index, so "double_space"
case_4 is testing that findSpaceIndex returns the last (third) index, so "last_space", "third_space", or "many_spaces"

widget/text_test.go
```diff
@@ -706,3 +706,106 @@ func TestText_lineBounds_variable_char_width(t *testing.T) {
...
 }
+
+func TestText_binarySearch(t *testing.T) {
+	tests := []struct {
+		name string
+		text string
+		want int
+	}{
+		{
+			name: "IM_Test",
```
stuartmscott, line 717: The result for this test will be reported as "TestText_binarySearch_IM_Test" - the "_Test" suffix is unnecessary

widget/text.go
```diff
@@ -341,11 +341,55 @@ func splitLines(text []rune) [][2]int {
...
+		}
+		for (high < maxHigh) && measurer(text[low:high+1]) <= maxWidth {
+			high++
+		}
+		return high
+	}
+}
+
+// findSpaceIndex accept a slice of runes and the last rune index
+// findSpaceIndex returns an index of space in the text or if not found the last rune index
```
stuartmscott, line 371: findSpaceIndex returns the index of the last space in the text, or fallback if there are no spaces

widget/text_test.go
toaster: If you touch it anyways: using `map[string]struct{…}` for table tests instead of `[]struct{name string, …}` gives a better readability _and_ runs the tests in random order (which is good to detect order dependencies).

Like so:
```go
for name, tt := range map[string]struct {
	…
} {
	"test description": { … },
} {
	t.Run(name, func(t *testing.T) {
		…
	})
}
```

andydotxyz:
The travis errors look like bad whitespace

widget/text_test.go
```diff
@@ -706,3 +706,85 @@ func TestText_lineBounds_variable_char_width(t *testing.T) {
...
+		"W": {
+			text: "WWWWW",
+			want: 4,
+		},
+		"Empty": {
+			text: "",
+			want: 0,
+		},
+	} {
+		maxWidth := 50
```
stuartmscott, line 753: Move out of 'for' loop, and put next to textSize & textStyle

widget/text.go
```diff
@@ -341,11 +341,57 @@ func splitLines(text []rune) [][2]int {
...
+			}
+		}
+		for (high < maxHigh) && lessMaxWidth(low, high+1) {
+			high++
+		}
+		return high
+	}
+}
+
+// findSpaceIndex accepts a slice of runes and the last rune index
```
stuartmscott, line 370: ...and a fallback index

widget/text.go
```diff
@@ -341,11 +341,57 @@ func splitLines(text []rune) [][2]int {
 	return append(lines, [2]int{low, length})
 }
 
+// newBinarySearch accepts a function that checks if the text width between the start and end index less the maximum width
+// newBinarySearch returns a function that accepts the start and end rune index and returns the index of rune
+// located as close to the maximum line width as possible
+func newBinarySearch(lessMaxWidth func(int, int) bool) func(low int, maxHigh int) int {
```
stuartmscott, line 347: I'm not sure this needs to be a func that returns a func. Why couldn't this be binarySearch(func, int, int) int?

Rather than:

```
binarySearch := newBinarySearch(func(...
...
binarySearch(low, high)
```

It would be:

```
ruler := func(...
...
binarySearch(ruler, low, high)
```

I think it would look cleaner, but am I missing something?

widget/text.go
```diff
@@ -341,11 +341,57 @@ func splitLines(text []rune) [][2]int {
...
+		return fallback
+	}
+	return curIndex
+}
+
 // lineBounds accepts a slice of runes, a wrapping mode, a maximum line width and a function to measure line width.
 // lineBounds returns a slice containing the start and end indicies of each line with the given wrapping applied.
 func lineBounds(text []rune, wrap fyne.TextWrap, maxWidth int, measurer func([]rune) int) [][2]int {
+
+	binarySearch := newBinarySearch(func(low int, high int) bool {
```
stuartmscott, line 389: BenchmarkText_lineBounds_WrapOff is faster if you move this initialization after line 396

fralx:
The 2f14b4dc commit broke the Label wrapping :(

stuartmscott:
Yes it did, luckily a fix is ready #859

stuartmscott:
The Travis error seems to be something wrong with the formatting of the binarySearch comment;
```
$ if [ "$TRAVIS_OS_NAME" != "windows" ]
then
fi
diff -u ./widget/text.go.orig ./widget/text.go
--- ./widget/text.go.orig	2020-04-16 13:53:32.366789999 +0000
+++ ./widget/text.go	2020-04-16 13:53:32.366789999 +0000
@@ -341,8 +341,8 @@
 	return append(lines, [2]int{low, length})
 }
 
-// binarySearch accepts a function that checks if the text width less the maximum width and the start and end rune index 
-// binarySearch returns the index of rune located as close to the maximum line width as possible 
+// binarySearch accepts a function that checks if the text width less the maximum width and the start and end rune index
+// binarySearch returns the index of rune located as close to the maximum line width as possible
 func binarySearch(lessMaxWidth func(int, int) bool, low int, maxHigh int) int {
 	if low >= maxHigh {
 		return low
/home/travis/.travis/functions: line 111: test: too many arguments
```

andydotxyz:
Looks like trailing whitespace

andydotxyz:
So fast, thanks :)

PR #847, opened 2020-04-11
Preview: menu refactoring

toaster:
This is a discussion-only PR.
The most work still has to be cleaned-up and tested.
It will then be split into smaller PRs.

### DIFF
--- a/cmd/fyne_demo/main.go
+++ b/cmd/fyne_demo/main.go
@@ -64,7 +64,18 @@ func main() {
 
 	w := a.NewWindow("Fyne Demo")
 
-	newItem := fyne.NewMenuItem("New", func() { fmt.Println("Menu New") })
+	// newItem := fyne.NewMenuItem("New", func() { fmt.Println("Menu New") })
+	newItem := fyne.NewMenuItem("New", nil)
+	otherItem := fyne.NewMenuItem("Other", nil)
+	otherItem.ChildMenu = fyne.NewMenu("",
+		fyne.NewMenuItem("Project", func() { fmt.Println("Menu New->Other->Project") }),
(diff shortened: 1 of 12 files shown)


toaster:
I opened this PR mainly for discussing one topic:
In the top-most commit (1a43677) most of the menu bar widget stuff is in the `widget` package and therefore public. I don't think it should because this is desktop (Linux, Windows) specific code.
Currently I cannot move it into the driver's menu file because then I have to export the `shadowingRenderer` and the `baseRenderer`.
My first approach was to create an internal widget package which would have contained all internal widgets (exported). The driver then could use this for its menus without making the menu bar widgets public.

However, maybe it is reasonable to make the mentioned renderers public instead?

WDYT?

andydotxyz:
The MenuBar is desktop specific - but the Menu should not be surely? Isn’t that what PopUpMenu should be using as well?

toaster:
Yes, exactly.

toaster:
Also I want to change the naming. Currently I named the widgets `Menu*Widget` to avoid the same name as the menu definition structures (`fyne.Menu`, `fyne.MenuItem`).
But I actually prefer names without `Widget`. We then would have `widget.Menu` and `widget.MenuItem` for the widgets and `fyne.Menu` and `fyne.MenuItem` for the definition.

andydotxyz:
> Yes, exactly.

You lost me here... You said things cant move because of the shadowrenderer but MenuBar doesn't use a shadow.

toaster:
Ah, yes! My brain is tangled too much with this stuff.
Unfortunately this enlightenment does not help:
To move `MenuBar` and `MenuBarItem` into the driver package, I need to export renderer constructors for `menuItemRenderer` and `menuRenderer`.

PR #848, opened 2020-04-11
Refactoring: deprecate NewPopUp*

toaster:
### Description:

The `NewPopUp*` methods not only construct a new pop-up widget but also immediately display it as overlay. This makes it impossible to extend the pop-up or pop-up menu without flickering.

With the new `ShowPopUp*` methods the old comfort remains while it is now possible to extend pop-up functionality.
Until 2.0 a developer has to build the pop-up with `widget.PopUp{…}` if he wants to do stuff with it before showing. Internally we can use the private constructors which will replace the deprecated ones in 2.0.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/geometry.go
+++ b/geometry.go
@@ -23,6 +23,11 @@ func (p1 Position) Add(p2 Position) Position {
 	return Position{p1.X + p2.X, p1.Y + p2.Y}
 }
 
+// IsZero returns whether the Size has zero width and zero height.
+func (p Position) IsZero() bool {
+	return p.X == 0 && p.Y == 0
+}
+
 // Subtract returns a new Position that is the result of offsetting the current
 // position by p2 -X and -Y.
 func (p1 Position) Subtract(p2 Position) Position {
--- a/geometry.go
+++ b/geometry.go
@@ -19,8 +19,8 @@ type Position struct {
 
 // Add returns a new Position that is the result of offsetting the current
 // position by p2 X and Y.
-func (p1 Position) Add(p2 Position) Position {
-	return Position{p1.X + p2.X, p1.Y + p2.Y}
+func (p Position) Add(p2 Position) Position {
+	return Position{p.X + p2.X, p.Y + p2.Y}
 }
 
 // IsZero returns whether the Size has zero width and zero height.
@@ -30,8 +30,8 @@ func (p Position) IsZero() bool {
 
 // Subtract returns a new Position that is the result of offsetting the current
 // position by p2 -X and -Y.
-func (p1 Position) Subtract(p2 Position) Position {
-	return Position{p1.X - p2.X, p1.Y - p2.Y}
+func (p Position) Subtract(p2 Position) Position {
+	return Position{p.X - p2.X, p.Y - p2.Y}
 }
 
 // Size describes something with width and height.
--- a/geometry.go
+++ b/geometry.go
@@ -37,8 +37,8 @@ type Size struct {
 
 // Add returns a new Size that is the result of increasing the current size by
 // s2 Width and Height.
-func (s1 Size) Add(s2 Size) Size {
-	return Size{s1.Width + s2.Width, s1.Height + s2.Height}
+func (s Size) Add(s2 Size) Size {
+	return Size{s.Width + s2.Width, s.Height + s2.Height}
 }
 
 // IsZero returns whether the Size has zero width and zero height.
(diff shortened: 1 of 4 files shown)


widget/popup.go
```diff
@@ -53,15 +53,24 @@ func (p *PopUp) Resize(size fyne.Size) {
 	p.Refresh()
 }
 
-// Show this widget, if it was previously hidden
+// Show this pop-up as overlay if not already shown.
 func (p *PopUp) Show() {
 	if !p.overlayShown {
+		if (p.Size() == fyne.Size{}) {
```
stuartmscott, line 59: nit: unnecessary parenthesis

widget/popup_test.go
```diff
@@ -20,6 +21,80 @@ func TestNewPopUp(t *testing.T) {
 	assert.Equal(t, pop, test.Canvas().Overlays().List()[0])
 }
 
+func TestShowPopUp(t *testing.T) {
+	require.Nil(t, test.Canvas().Overlays().Top())
```
stuartmscott, line 25: test.Canvas() is called multiple times in this test, does it change or can it be a variable?

widget/popup.go
toaster: No, the `fyne.Size{}` requires the parenthesis here.

widget/popup_test.go
toaster: It's immutable, I think. But honestly, I don't think it is of much importance in tests if it does not affect the test's comprehensibility.

widget/popup.go
stuartmscott: Ah ok, creating a new object to compare with Size to see if it is 0,0 is not very elegant. Maybe Size should have an "isEmpty" or "isZero" method?

widget/popup_test.go
stuartmscott: Fair enough

andydotxyz:
Looks good, but I have a question: Should the canvas be attached to the constructor or the show request? It is needed to set the overlay so either works.
Not sure how much it matters, but for situations where the popup is used for other things like menu might we want to construct it and assign the canvas / position together later on?

widget/popup.go
toaster: I would assume that the Go compiler is able to optimise comparisons with zero values by only allocating the needed value once (or at least not that often) or by applying an optimised zero value test procedure. Of course it might not do such stuff.

Before adding a very special function for a limited purpose I want to see benchmarks that prove its benefits.
Is calling a function which compares two `int` values really faster than the current comparison?

toaster:
> Looks good, but I have a question: Should the canvas be attached to the constructor or the show request? It is needed to set the overlay so either works.
> Not sure how much it matters, but for situations where the popup is used for other things like menu might we want to construct it and assign the canvas / position together later on?

I agree regarding the position. That’s why it is not assigned via the new constructors but only by the convenience methods.
It might be useful for the canvas, too. But I don’t see the use-case here. Also, it would break the `PopUp` because its `Widget` methods (`Show`, `Hide`, `Resize`) depend on the canvas. Therefore I think the canvas should remain a constructor argument.

widget/popup.go
stuartmscott: #849

toaster: Do you agree that the function isn’t significantly faster?

stuartmscott: Hahaha you didn't ask for a benchmark showing it is significantly faster :)

As far as execution time goes 10ns is nothing to a human, and a handful of opcodes to a CPU.

As far as code goes
```
if p.Size().Empty() {
```
is significantly more elegant and readable than
```
if (p.Size() == fyne.Size{}) {
```

andydotxyz: I don’t really have an opinion on this, but if a method were added I imagine that Zero() fits the requirement more than Empty()

toaster: I just introduced `IsZero()`…

stuartmscott: Thanks, it is much more readable 👍

andydotxyz:
Looks great overall - but in the geometry reorder, was it not supposed to be:
* Type definition
* Constructors
* Exported methods
...?

geometry.go
```diff
@@ -1,5 +1,39 @@
...
+	Y int // The position from the parent's top edge
+}
+
+// Add returns a new Position that is the result of offsetting the current
+// position by p2 X and Y.
+func (p Position) Add(p2 Position) Position {
+	return Position{p.X + p2.X, p.Y + p2.Y}
+}
+
+// IsZero returns whether the Size has zero width and zero height.
```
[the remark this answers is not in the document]
toaster: 😬

toaster:
> Looks great overall - but in the geometry reorder, was it not supposed to be:
> 
>     * Type definition
> 
>     * Constructors
> 
>     * Exported methods
>       ...?

Yes :/ … will fix it tomorrow.

toaster:
The build failed because of a Travis hick-up. @andydotxyz please restart the failed task.

PR #850, opened 2020-04-12
Add OnTabSelected callback to tab container

pymq:
### Description:
Add `OnTabSelected` to TabContainer widget. Also removed unnecessary computations if tab already selected.

My primary use-case was to update data from external source and avoid background polling. I think it would be useful for others as well.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne_demo/main.go
+++ b/cmd/fyne_demo/main.go
@@ -88,7 +88,9 @@ func main() {
 		widget.NewTabItemWithIcon("Graphics", theme.DocumentCreateIcon(), screens.GraphicsScreen()),
 		widget.NewTabItemWithIcon("Windows", theme.ViewFullScreenIcon(), screens.DialogScreen(w)),
 		widget.NewTabItemWithIcon("Advanced", theme.SettingsIcon(), screens.AdvancedScreen(w)))
-	tabs.SetTabLocation(widget.TabLocationLeading)
+	tabs.OnTabSelected = func(tab *widget.TabItem) {
+		fmt.Printf("%s tab selected\n", tab.Text)
+	}
 	tabs.SelectTabIndex(a.Preferences().Int(preferenceCurrentTab))
 	w.SetContent(tabs)
 
--- a/cmd/fyne_demo/main.go
+++ b/cmd/fyne_demo/main.go
@@ -88,6 +88,7 @@ func main() {
 		widget.NewTabItemWithIcon("Graphics", theme.DocumentCreateIcon(), screens.GraphicsScreen()),
 		widget.NewTabItemWithIcon("Windows", theme.ViewFullScreenIcon(), screens.DialogScreen(w)),
 		widget.NewTabItemWithIcon("Advanced", theme.SettingsIcon(), screens.AdvancedScreen(w)))
+	tabs.SetTabLocation(widget.TabLocationLeading)
 	tabs.OnChanged = func(tab *widget.TabItem) {
 		fmt.Printf("%s tab selected\n", tab.Text)
 	}
--- a/cmd/fyne_demo/main.go
+++ b/cmd/fyne_demo/main.go
@@ -88,7 +88,7 @@ func main() {
 		widget.NewTabItemWithIcon("Graphics", theme.DocumentCreateIcon(), screens.GraphicsScreen()),
 		widget.NewTabItemWithIcon("Windows", theme.ViewFullScreenIcon(), screens.DialogScreen(w)),
 		widget.NewTabItemWithIcon("Advanced", theme.SettingsIcon(), screens.AdvancedScreen(w)))
-	tabs.OnTabSelected = func(tab *widget.TabItem) {
+	tabs.OnChanged = func(tab *widget.TabItem) {
 		fmt.Printf("%s tab selected\n", tab.Text)
 	}
 	tabs.SelectTabIndex(a.Preferences().Int(preferenceCurrentTab))
--- a/widget/tabcontainer.go
+++ b/widget/tabcontainer.go
@@ -46,9 +46,10 @@ const (
 type TabContainer struct {
 	BaseWidget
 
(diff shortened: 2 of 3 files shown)


widget/tabcontainer.go
```diff
@@ -80,6 +81,8 @@ func (t *TabContainer) CurrentTab() *TabItem {
 func (t *TabContainer) SelectTabIndex(index int) {
 	if index < 0 || index >= len(t.Items) {
 		return
+	} else if t.current == index {
```
stuartmscott, line 84: This could be merged into the if above:
```
if index < 0 || index >= len(t.Items) || t.current == index {
  return
}
```

widget/tabcontainer.go
```diff
@@ -95,6 +98,11 @@ func (t *TabContainer) SelectTabIndex(index int) {
 	r := cache.Renderer(t).(*tabContainerRenderer)
 	r.Layout(t.size)
 	t.Refresh()
+
+	callback := t.OnTabSelected
```
stuartmscott, line 102: This could be inlined:
```
if callback := t.OnTabSelected; callback != nil {
  callback(t.Items[t.current])
}
```

widget/tabcontainer.go
```diff
@@ -46,9 +46,10 @@ const (
 type TabContainer struct {
 	BaseWidget
 
-	Items       []*TabItem
-	current     int
-	tabLocation TabLocation
+	Items         []*TabItem
+	OnTabSelected func(tab *TabItem)
```
stuartmscott, line 50: Rename to OnChanged to be consistent with Check, Radio, and Select.

I'm not sure whether this parameter should be the TabItem, or just the current index. Anyone have a strong opinion one way or the other?

stuartmscott:
@Jacalz seeing the same failure on Windows, go 1.13:
> --- FAIL: TestWatchFile (0.33s)
FAIL

widget/tabcontainer.go
pymq: I thought it would be more readable this way, but probably you are right.

stuartmscott: Its a tough call, if the bounds checking returned an error then it would be better to keep them separate because you're adding an optimization, but since the outcome is the same we might as well aim for brevity.

widget/tabcontainer.go
Jacalz: Is on OnChanged the best name? I know that it doesn’t provide as much consistency, but I think OnSelect might be a better name?

pymq: I think TabItem is more useful - you can directly set different icon/title on selected tab. Index is rarely needed, and also you can get it in one call - so I don't see a problem.

Jacalz:
> @Jacalz seeing the same failure on Windows, go 1.13:
> 
> > --- FAIL: TestWatchFile (0.33s)
> > settings_desktop_test.go:62: File watcher callback was not called
> > FAIL

Sorry, but I have no idea why this is happening now. I have verified that all my changes to the file have been reverted and I can't think of why this would be happening anymore. Might be time to stop putting the blame on me now, but I don't know...

widget/tabcontainer.go
andydotxyz: I’m not really a fan of setting up inside an if like this - unless it’s an ok check for casting. Either way in this instance you don’t need the variable set anyhow :)

widget/tabcontainer.go
andydotxyz: I agree TabItem - i think the rule of thumb is prefer the typed version when possible :)

widget/tabcontainer.go
stuartmscott: Fair enough, this is shorter anyway;
```
if t.OnTabSelected != nil {
  t.OnTabSelected(t.Items[t.current])
}
```

stuartmscott:
@Jacalz sorry mate!

@andydotxyz can we pull these tests out of Travis?

widget/tabcontainer.go
pymq: I thought about this too. But I have a question: is it possible a race condition between checking for nil and calling a function? I understand that we still have no lock on set/read, but still... I think I've had this problem once before.

andydotxyz: If there is a race condition on read then setting it to another variable sadly does not remove that. I appreciate that it’s “less likely” but technically still there unless you guard access to the variable.

pymq: Sorry, I was not clear, I meant to protect only one special case when at one time function was set and later on it set as nil and calling nil func after `if` will cause panic.

Just out of curiosity I tried to reproduce this case and failed, probably on a low level it doesn't work like that.

cmd/fyne_demo/main.go
```diff
@@ -88,7 +88,9 @@ func main() {
 		widget.NewTabItemWithIcon("Graphics", theme.DocumentCreateIcon(), screens.GraphicsScreen()),
 		widget.NewTabItemWithIcon("Windows", theme.ViewFullScreenIcon(), screens.DialogScreen(w)),
 		widget.NewTabItemWithIcon("Advanced", theme.SettingsIcon(), screens.AdvancedScreen(w)))
-	tabs.SetTabLocation(widget.TabLocationLeading)
```
stuartmscott, line 91: Should this be removed by this PR? It maybe an unnecessary call as leading is the default location, but this code is mean for showing devs how to use the framework so could be helpful to leave

stuartmscott: *correction: TabLocationTop is the default, so this line is needed

pymq: @stuartmscott Sorry, it was deleted by accident. Will revert later.

stuartmscott:
Looks good to me, thanks @pymq !

PR #851, opened 2020-04-13
Callbacks for Information and Error dialogs

### LINKED ISSUE #752: Callback on error dialog dismiss
Whenever I display an error message to the user in the error dialog, I'd like to have some way of handling when this window is dismissed. 

In the case of my app, any error that I display in the error dialog is a critical one, after which I do not expect the user to continue using the app, thus dismiss action must call: window.Close() and some other cleanup code (e.g. closing DB connection, stopping goroutines).

I think callback function on the error dialog dismiss would be helpful to accomplish this task.

obsti8383:
### Description:
Motivation: (from #752): "Whenever I display an error message to the user in the error dialog, I'd like to have some way of handling when this window is dismissed.
In the case of my app, any error that I display in the error dialog is a critical one, after which I do not expect the user to continue using the app, thus dismiss action must call: window.Close() and some other cleanup code (e.g. closing DB connection, stopping goroutines).
I think callback function on the error dialog dismiss would be helpful to accomplish this task."

This pull request implements this behaviour for information and error dialogs with new "xxxWithCallback" methods.

Fixes #752 

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

#### Where applicable:


dialog/base.go
```diff
@@ -39,6 +40,14 @@ type dialog struct {
 	parent         fyne.Window
 }
 
+// SetOnClosed allows to set a callback function that is called when
+// the dialog is closed
+func (d *dialog) SetOnClosed(closed func()) {
+	d.callback = func(dummy bool) {
```
andydotxyz, line 46: One sticking point... It could be possible that this field is set already (like in a confirm dialog).
I think that probably this callback should call the other, if it exists (i.e. save a reference to it and call before the closed() callback)?

toaster: Having a parameter named `dummy` is a waste of space. You can simply omit the parameter name if you don't intend to use it. If you want to express more explicitly that you didn't forgot the name but intentionally don't use the param, you can use a `_`.

dialog/information_test.go
```diff
@@ -16,3 +18,42 @@ func TestDialog_MinSize(t *testing.T) {
...
+	d.SetOnClosed(func() { tapped <- true })
+	d.Show()
+
+	information := d.(*dialog)
+	assert.False(t, information.win.Hidden)
+	go test.Tap(information.dismiss)
+	func() {
+		select {
+		case <-tapped:
+		case <-time.After(1 * time.Second):
```
toaster, line 54: That's way too long IMO.
```suggestion
		case <-time.After(100 * time.Millisecond):
```

toaster: Oh, when using `test.Tap` you don't have to do concurrency, it's synchronous. However, using a `chan` and the `select` makes the test independent from the implementation of `test.Tap`.

obsti8383:
@andydotxyz  @toaster Should be fixed now. Hope the solution for saving the callback for Confirm dialogs is OK?

dialog/base.go
```diff
@@ -39,6 +41,22 @@ type dialog struct {
 	parent         fyne.Window
 }
 
+// SetOnClosed allows to set a callback function that is called when
+// the dialog is closed
+func (d *dialog) SetOnClosed(closed func()) {
+	// if there is already a callback set, remember it and call both
+	if d.callback != nil {
```
andydotxyz, line 48: You don’t need a new field here, just set up ‘originalCallback := callback’

dialog/base.go
```diff
@@ -21,16 +21,18 @@ type Dialog interface {
...
 // Declare conformity to Dialog interface
 var _ Dialog = (*dialog)(nil)
 
 type dialog struct {
-	callback     func(bool)
-	sendResponse bool
-	title        string
-	icon         fyne.Resource
+	callback         func(bool)
+	originalCallback func(bool) // used by SetOnClosed() if there is already a callback set
```
andydotxyz, line 32: This isn’t needed - see comment below

dialog/information_test.go
andydotxyz: It looks like this is still 1 second

obsti8383: Sorry, fixed.

dialog/base.go
obsti8383: Ah, sorry, pretty sure I had it that way at first but got me into problems with nil pointers. One shouldn't program to late in the night... Fixed.

dialog/information_test.go
```diff
@@ -51,7 +51,7 @@ func TestDialog_ErrorCallback(t *testing.T) {
 	func() {
 		select {
 		case <-tapped:
-		case <-time.After(1 * time.Second):
+		case <-time.After(100 * time.Second):
```
toaster, line 54: Now you have raised the timeout to 100 seconds instead of lowering it to 100 milliseconds…

obsti8383: Sorry again. Overlooked the milliseconds from the beginning. Not may day(s). Should now really be fixed.

andydotxyz:
Thanks.
There was another instance of a 1 second wait that we all missed - but I'm going to merge now and tweak that area a little anyway.

andydotxyz:
That's great thanks.
I missed this before unfortunately but we try to limit master work to hotfixes, so please do larger work off 'develop' branch.

PR #858, opened 2020-04-14
Refactoring/menu prerequisites

toaster:
### Description:

This PR contains more refactorings in preparation of the non-native menu overhaul.
It also includes a bugfix for the `shadowingRenderer`.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/popup.go
+++ b/widget/popup.go
@@ -14,9 +14,9 @@ import (
 type PopUp struct {
 	BaseWidget
 
-	Content        fyne.CanvasObject
-	Canvas         fyne.Canvas
-	WithoutPadding bool
+	Content   fyne.CanvasObject
+	Canvas    fyne.Canvas
+	NotPadded bool
 
 	innerPos     fyne.Position
 	innerSize    fyne.Size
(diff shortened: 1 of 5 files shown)


widget/popup.go
```diff
@@ -14,8 +14,9 @@ import (
 type PopUp struct {
 	BaseWidget
 
-	Content fyne.CanvasObject
-	Canvas  fyne.Canvas
+	Content        fyne.CanvasObject
+	Canvas         fyne.Canvas
+	WithoutPadding bool
```
andydotxyz, line 19: In Canvas we use `Padded` - helps to differentiate from the theme `Padding`.
I wonder if this should be consistent with that, such as `NotPadded` or `UnPadded`.
Brain going slow but I think you get the gist?

andydotxyz: It's not entirely clear what will use this as I think it's just called from tests so far, I assume we can't flip the logic as that would remove padding from the &PopUp{} creations?

toaster: > It's not entirely clear what will use this as I think it's just called from tests so far, I assume we can't flip the logic as that would remove padding from the &PopUp{} creations?

Yes, unfortunately this is true.
But it all will be consistent and good when we add elevation support some day.

PR #864, opened 2020-04-15
Fix error of backspace keydown to MultiLineEntry

lusingander:
### Description:

Pressing the backspace key several times in MultiLineEntry may cause panic.

For example, press BS twice in the following situation:
```
aaa
b|
```
When the cursor is at the beginning of a line, the value of row/col is wrong.
The calculation of the boundaries modified by #712 seemed to be little wrong, so I fixed it.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
Great catch thank you

PR #872, opened 2020-04-16
Refactoring/menu prerequisites

toaster:
### Description:

The next prerequisite PR for the upcoming menu change. We get closer!
In this PR the `menuItemWidget` no longer embeds `Label` and handles the dismiss action separately from the menu's action. It also knows about the `fyne.MenuItem`.

In addition the padding in pop-up's is optional now.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/entry_test.go
+++ b/widget/entry_test.go
@@ -448,8 +448,8 @@ func TestEntry_TappedSecondary(t *testing.T) {
 	pos := fyne.CurrentApp().Driver().AbsolutePositionForObject(over)
 
 	cont := over.(*PopUp).Content
-	assert.Equal(t, pos.X+theme.Padding()+tapPos.X, cont.Position().X)
-	assert.Equal(t, pos.Y+theme.Padding()+tapPos.Y, cont.Position().Y)
+	assert.Equal(t, pos.X+tapPos.X, cont.Position().X)
+	assert.Equal(t, pos.Y+tapPos.Y, cont.Position().Y)
 
 	items := cont.(*Box).Children
 	assert.Equal(t, 4, len(items)) // Cut, Copy, Paste, Select All
--- a/widget/select_test.go
+++ b/widget/select_test.go
@@ -5,7 +5,6 @@ import (
 
 	"fyne.io/fyne"
 	"fyne.io/fyne/test"
-	"fyne.io/fyne/theme"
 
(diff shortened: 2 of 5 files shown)


widget/entry_test.go
```diff
@@ -448,8 +448,8 @@ func TestEntry_TappedSecondary(t *testing.T) {
 	pos := fyne.CurrentApp().Driver().AbsolutePositionForObject(over)
 
 	cont := over.(*PopUp).Content
-	assert.Equal(t, pos.X+theme.Padding()+tapPos.X, cont.Position().X)
```
andydotxyz, line 451: This doesn't smell right. Are entry popups now not padded, or is the position handling changed?

widget/select_test.go
```diff
@@ -151,8 +150,8 @@ func TestSelect_Tapped(t *testing.T) {
 	pos := fyne.CurrentApp().Driver().AbsolutePositionForObject(over)
 
 	cont := over.(*PopUp).Content
-	assert.Equal(t, cont.Position().X, pos.X+theme.Padding())
```
andydotxyz, line 154: Here it seems like we are testing that the select drop down no longer has padding - is that correct?

widget/entry_test.go
toaster: The padding is now in the menuItemWidget and therefore the pop-up must not add additional padding.

widget/select_test.go
toaster: Yes, because it is a menu pop-up and these are no longer padded because the items already have padding now.

widget/entry_test.go
andydotxyz: I don't follow - this code is asking for the X position of the top overlay. Why does where the padding is change that?

andydotxyz: Oh, I think I see - you are comparing to the content position, and that is now different?

I guess I missed the implicit design change "popup menus no longer have padding".

toaster: Its explicit in the commit comment:
“… menu pop-up is without padding”

andydotxyz: Haha, yes I guess. Perhaps changes in design could be called out with an image?
I know it's not a big deal just for some reason it hadn't sunk in.

PR #873, opened 2020-04-16
Text Grid - attempt 2

andydotxyz:
### Description:
Add a TextGrid widget, basically a 2D array of runes with styles attached.
Some helper APIs for applying style.
And started work on semantic styles :)

This is a replaement for #746 as develop had diverged too far in text handling.
This does not include full history unfortunatey but also I realised it did not need the textHandler refactor so this is slimmer than before.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/textgrid.go
```diff
@@ -0,0 +1,384 @@
...
+		bufCols += t.lineCountWidth()
+	}
+
+	t.cols = int(math.Max(float64(sizeCols), float64(bufCols)))
+	t.rows = int(math.Max(float64(sizeRows), float64(bufRows)))
+}
+
+func (t *textGridRenderer) Layout(size fyne.Size) {
+	t.updateGridSize(size)
+	t.ensureGrid()
```
toaster, line 344: This name is confusing … as discussed.

widget/textgrid.go
```diff
@@ -0,0 +1,384 @@
...
+
+		line++
+	}
+	for ; x < len(t.objects)/2; x++ {
+		t.setCellRune(' ', x, TextGridStyleDefault) // trailing cells and blank lines
+	}
+	canvas.Refresh(t.text)
+}
+
+func (t *textGridRenderer) lineCountWidth() int {
```
toaster, line 318: Possible `lineNumberWidth`?

widget/textgrid_test.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func TestTextGrid_LineNumbers(t *testing.T) {
+	grid := NewTextGridFromString("1\n2\n3\n4\n5\n6\n7\n8\n9\n10")
+	grid.ShowLineNumbers = true
+
+	grid.Resize(fyne.NewSize(100, 250))
+	r := test.WidgetRenderer(grid).(*textGridRenderer)
+
+	assert.Equal(t, ' ', rendererCellRune(r, 0, 0))
```
toaster, line 30: This would be cool to be just a comparison to an ASCII string representing the rendered grid.

widget/textgrid_test.go
```diff
@@ -0,0 +1,134 @@
...
+	grid.SetStyleRange(0, 1, 1, 0, &CustomTextGridStyle{FGColor: color.White, BGColor: color.Black})
+
+	assert.Nil(t, grid.Content[0][0].Style)
+	assert.Equal(t, color.White, grid.Content[0][1].Style.TextColor())
+	assert.Equal(t, color.Black, grid.Content[0][1].Style.BackgroundColor())
+	assert.Equal(t, color.White, grid.Content[1][0].Style.TextColor())
+	assert.Equal(t, color.Black, grid.Content[1][0].Style.BackgroundColor())
+	assert.Nil(t, grid.Content[1][1].Style)
+}
+
```
toaster, line 79: Add tests for index overflows, please.

theme/gen.go
```diff
@@ -85,7 +85,7 @@ func main() {
 	bundleFont(fontFace, "Bold", f)
 	bundleFont(fontFace, "Italic", f)
 	bundleFont(fontFace, "BoldItalic", f)
-	bundleFont("NotoMono", "Regular", f)
+	bundleFont("DejaVuSansMono", "Powerline", f)
```
stuartmscott, line 88: Does this TextGrid PR require the new font or can it be added separately? Is there an issue with NotoMono?

widget/textgrid.go
```diff
@@ -0,0 +1,384 @@
...
+}
+
+// CreateRenderer is a private method to Fyne which links this widget to it's renderer
+func (t *TextGrid) CreateRenderer() fyne.WidgetRenderer {
+	t.ExtendBaseWidget(t)
+	render := &textGridRenderer{text: t}
+
+	cell := canvas.NewText("M", color.White)
+	cell.TextStyle.Monospace = true
+	render.cellSize = cell.MinSize()
```
stuartmscott, line 191:
```go
render.cellSize = fyne.MeasureText("M", theme.TextSize(), fyne.TextStyle{Monospace:true})
```

theme/gen.go
andydotxyz: Good point. As my test app was a terminal the NotoMono didn't really cut the mustard.
Powerline was needed to make this good enough IMHO.
The move from noto mono to dejavu sans was just a personal choice that I forgot about.
There is a noto powerline if you'd prefer.

widget/textgrid.go
andydotxyz: Thanks, this was not an option when I first coded it up. In now :)

widget/textgrid.go
```diff
@@ -156,6 +166,14 @@ func (t *TextGrid) SetStyle(row, col int, style TextGridStyle) {
...
 // SetStyleRange sets a grid style to all the cells between the start row and column through to the end row and column.
 func (t *TextGrid) SetStyleRange(startRow, startCol, endRow, endCol int, style TextGridStyle) {
+	if startRow >= len(t.Content) {
+		return
+	}
+	if endRow >= len(t.Content) {
+		endRow = len(t.Content) - 1
+		endCol = len(t.Content[endRow]) - 1
+	}
+
```
toaster, line 176: I think `startRow` and `endRow` also should be `>= 0`.

theme/gen.go
stuartmscott: If DejaVu is a better font we should use it, but maybe it should be added in its own PR. If further down the road we decide we don't like it we can revert it without reverting TextGrid.

andydotxyz: True, but reverts are done at a commit level not a PR level, so creating a new PR makes no difference in that regard :)

Issue #827, opened 2020-04-07
Consistent widget file layout


stuartmscott:
Every widget seems to have a different file layout; some define the widget struct, widget constructors, widget functions, then widget renderer. Others define the renderer first with widget constructors at the bottom. Others like Toolbar define their auxiliary objects first. Can we define and adhere to a consistent file layout?

My preferred file layout would be;
- widget struct
- widget constructor(s)
- widget exported functions
- widget unexported functions
- widget auxiliary objects
- widget renderer
- widget renderer auxiliary objects

Existing files can be updated without breaking the API although they will lose git blame-ability.

andydotxyz:
This works for me. Though I would be tempted to put the renderer with the object it renders rather than after it's auxiliary objects.

The other possible change would be putting renderers in another file. It could make it a little more annoying to develop but it would further encourage the separation :)

stuartmscott:
That makes sense;
- widget struct
- widget constructor(s)
- widget exported functions
- widget unexported functions
- widget renderer
- widget auxiliary objects
- widget renderer auxiliary objects

Yeah @toaster mentioned this as well. It would be good to separate them, but if most changes to a widget affect its renderer too then it will be annoying

toaster:
I like the idea!

Regarding separated files:
1. If we do this then the auxiliary objects should have their own files, too, shouldn't they?
2. This would create a lot of more files in the widget folder.

What about this:
- keep renderer and widget together in a file (for now)
- extract auxiliary widgets into separate files
- keep other auxiliary objects in their widget's file

andydotxyz:
Changes such as this probably have a relatively small window to put in as reordering files is very disruptive to potential merges of PRs, bugs or branches. If this is to be in as part of 1.3 push then I would say we likely need it in the next 3 or 4 days, otherwise it gets pushed to the (longer) 2.0 window that opens around May 5th.

Any thoughts or volunteers?

stuartmscott:
2.0 seems more realistic and sensible

andydotxyz:
Can I request that we have [Set|Is] prefix ignored in the orderring so that MyVal and SetMyVal are next to each other?

stuartmscott:
On a separate, but related topic, tests should also be consistent.

A feature XYZ should have all public APIs checked by test cases located in a separate file and package. Some internal behaviour may also need to be tested, these tests should be in a separate file with the same package:
- XYZ.go (package xyz) - main feature being tested
[...]

PR #875, opened 2020-04-16
Refactoring: internal widgets

toaster:
### Description:

This PR moves several internal widgets (which are needed for the upcoming menu refactoring) into a new internal `widget` package.

### Checklist:

- ~~[ ] Tests included.~~
- [x] Tests all pass.


internal/widget/base.go
```diff
@@ -0,0 +1,83 @@
...
+	r.Layout(size)
+}
+
+func (b *base) show(w fyne.Widget) {
+	if !b.hidden {
+		return
+	}
+
+	b.hidden = false
+	w.Refresh()
```
stuartmscott, line 82: how come show uses w.Refresh(), but hide uses canvas.Refresh(w)?

internal/widget/base.go
```diff
@@ -0,0 +1,83 @@
...
+	b.size = size
+	r := cache.Renderer(w)
+	if r == nil {
+		return
+	}
+
+	r.Layout(size)
+}
+
+func (b *base) show(w fyne.Widget) {
```
stuartmscott, line 76: I was going to ask if show can be moved up next to hide, then I realized you have made all of these files declare members in alphabetical order and I think I prefer that, danke!

internal/widget/base_renderer.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"image/color"
+
+	"fyne.io/fyne"
+	"fyne.io/fyne/theme"
+)
+
+// NewBaseRenderer creates a new BaseRenderer.
+func NewBaseRenderer(objects []fyne.CanvasObject) BaseRenderer {
```
stuartmscott, line 11: #827 says struct before constructor, but this looks okay too. I'd rather the codebase was consistent so should #827 be changed to this order?

widget/base_renderer.go
```diff
@@ -1,27 +0,0 @@
-package widget
-
-import (
-	"image/color"
-
-	"fyne.io/fyne"
-	"fyne.io/fyne/theme"
-)
-
-type baseRenderer struct {
```
stuartmscott, line 10: I understand why this was moved into internal, however this convenience is no longer unaccessible for developers creating custom widgets. We could expose it as:
```
package widget

import "fyne.io/fyne/internal/widget"

type BaseRenderer struct {
  widget.baseRenderer
}
```
What do you think?

internal/widget/base.go
toaster: That’s because a hidden widget won’t be rendered. Therefore `w.Refresh()` is a waste of time there.

internal/widget/base.go
toaster: Gern geschehen! (My pleasure!)

internal/widget/base_renderer.go
toaster: Oops … I wrote this code before #827. Because everybody™ agreed to #827 I already did some other refactorings with structs before constructors.
I will adjust this PR.

widget/base_renderer.go
toaster: I think the advantage is too small to do so.
@andydotxyz convinced me to _not_ add such a construct for the base widget where the advantage would be much bigger. Therefore I would not do that here and now.

andydotxyz: I think this was not available to any custom widget creation because it was not exposed. The conversation I had with @toaster largely came to the conclusion that someone writing a widget could be expected to write a complete renderer, they do not have the same inheritance properties that widget objects do. Additionally if we make it too "simple" or "static" then there is a much higher chance that third party widgets would not respond well on refresh or theme change (I guess?).

There is another discussion about making it easier to create widgets with trivial renderers at #709 but honestly I am coming to the view that such enablers might encourage things that we don't think are good ideas (such as realising that customisation is difficult resorting to everything becoming a clickable image using a png that does not reflect theme colours)...

It's probably quite a big discussion but given that this was not exposed we can always decide to add your proposed solution in the future.

PR #877, opened 2020-04-17
Feature: menu widget

toaster:
### Description:

This PR contains the first part of the menu change: `internal/widget/menu`.
This widget replaces the box that is used for menus up to now.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
Hmm, thanks for all of this, it's really coming together. I see that the reason for making popup shadow optional is to allow the menu to draw it's own shadow - this seems a little strange. 
Can you expand on why this is needed?

widget/popup_test.go
```diff
@@ -273,6 +276,56 @@ func TestPopUp_Stacked(t *testing.T) {
...
+			pop.Resize(size)
+			r := cache.Renderer(pop)
+			expectedObjectCount := 2
+			if tt.castShadow {
+				expectedObjectCount++
+			}
+			require.GreaterOrEqual(t, len(r.Objects()), expectedObjectCount)
+
+			bgIdx := 0
+			if tt.castShadow {
```
andydotxyz, line 312: This doesn't test the no-shadow state does it?

toaster:
> Hmm, thanks for all of this, it's really coming together. I see that the reason for making popup shadow optional is to allow the menu to draw it's own shadow - this seems a little strange.
> Can you expand on why this is needed?

Because menus will be allowed to have submenus in the next PR (tests just finished). These have to draw another shadow.

widget/popup_test.go
toaster: not explicitly, good catch
if the pop-up would cast a shadow with `hideShadow == true`, the following assertion regarding the background would fail. I’ll add an explicity assertion to clarify the intention.

andydotxyz:
I don’t follow sorry - won’t submenus look like top level menus? A list of items with a shadow around it...

toaster:
Exactly, and who should draw the shadow?

internal/widget/shadow.go
```diff
@@ -27,6 +27,7 @@ const (
 	ButtonLevel           ElevationLevel = 2
 	PopUpLevel            ElevationLevel = 8
 	SubmergedContentLevel ElevationLevel = 8
+	MenuLevel             ElevationLevel = 4
```
stuartmscott, line 30: nit: move up 2 lines

andydotxyz:
I'm having difficulty reading this PR now sorry. The modifications to PopUp coming and going seem to be interleaved with the commits that get to this finished menu.
Is it possible to somehow pick the required commits onto a clean branch?

My other question is around the internal Overlay widget - it is only used by the new PopUpMenu - my normal rule of thumb is to not create an abstraction until there are multiple uses of it, so is the additional code really needed at this time?

toaster:
> My other question is around the internal Overlay widget - it is only used by the new PopUpMenu - my normal rule of thumb is to not create an abstraction until there are multiple uses of it, so is the additional code really needed at this time?

I don’t think this is an abstraction. As I already mentioned in #707 a widget should not be an overlay and another widget at the same time. The `PopUpMenu` can be moved and resized while an overlay can’t really (it should be canvas’ size at 0,0).
However, there will be another usage of the `Overlay` by the main menu widget, soon.

toaster:
> I'm having difficulty reading this PR now sorry. The modifications to PopUp coming and going seem to be interleaved with the commits that get to this finished menu.
> Is it possible to somehow pick the required commits onto a clean branch?

Technically: yes.
Should I do it: I don’t think so.
It’s not that I’m too lazy (I am, of course), but the revert of the `PopUp` changes includes `NotPadded` which is already on the develop branch. In the end we would save like two small commits after investing a couple of hours.

andydotxyz:
> widget should not be an overlay and another widget at the same time

Hmm, I see what you mean. And yet we publicly expose the AddOverlay but hide the Overlay widget.
I think this may be a naming issue. To avoid confusing developers could something less generic and more meaningful help? Like OverlayWrapper, or OverlayContainer?
I would like to avoid people thinking they need to use this and can't, if they want an overlay. The reality is that any widget could be one, even though (as you say) it shouldn't be both.

Alternatively we could make a much more sweeping change that requires Overlays are in fact Overlays...

toaster:
I think using a different name is the best way.
WDYT about `OverlayBackground` or `OverlayPresenter`?

andydotxyz:
> WDYT about `OverlayBackground` or `OverlayPresenter`?

Presenter is a bit of a trigger word as it's used for defined purposes in MVP https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93presenter.

OverlayBackground sounds servicable, I just proposed Container as we have used that elsewhere. It is both a background and a container I guess. Perhaps background sounds alittle more passive than the object is?

toaster:
It’s fyne for me if you prefer `Container` over `Background`. It’s an internal widget for now and we are free to change the name if a better one occurs :).

andydotxyz:
Good by me, though I think Stuart had an open review too?

toaster:
Now everything should be in @stuartmscott @andydotxyz

PR #879, opened 2020-04-17
Feature/filehandling

### LINKED ISSUE #822: File dialog returns do not work on mobile
The simple `path string` return type from file dialogs is not sufficient for mobile (or web) file picking.
The solution os to define a new, stream based, file type that will be usable elsewhere as well.

Proposal at https://github.com/fyne-io/fyne/wiki/Settings-API

andydotxyz:
### Description:
This is an almost-complete file handler implementation for dialogs and re-accessing files later.
It follows the updated API discussion at [Settings-API](https://github.com/fyne-io/fyne/wiki/Settings-API)

This is fully working on desktop and on iOS.
Android is still to come, hence why this is draft.
Comments welcome.

Open question: Should we implicity create files on write?
The (default) dialog will warn the user you will overwrite if the file exists - but should the developer have to decide whether or not this is to be a new file?
I think it's OK this way but it's certainly non-trivial question.

This is a breaking API change, as discussed. It is only on develop so the break is not a released one.

Fixes #822

### Checklist:

- [x] Tests included. - yes, some. TODO add a few more, especially edge cases
- [x] Tests all pass.

*ALSO NOTE* To test please turn off go modules - I have temporary changes in the vendor package that will get merged upstream before this lands!!!


andydotxyz:
OK, that is iOS and Android file opening completed.
This adds the refactor as discussed for files in the Settings-API wiki page.
Questions at the top may still need discussed.

file.go
```diff
@@ -0,0 +1,31 @@
+package fyne
+
+import "io"
+
+// FileReader represents a cross platform data stream from a file or provider of data.
+// It may refer to an item on a filesystem or data in another application that we have access to.
+type FileReader interface {
+	io.ReadCloser
```
toaster, line 8: `FileReader` doesn’t read as if it could `Close()`, too.
I think this should be called `FileReadCloser()`.

file.go
```diff
@@ -0,0 +1,31 @@
...
+	Name() string
+	URI() string
+}
+
+// FileWriter represents a cross platform data writer for a file resource.
+// This will normally refer to a local file resource.
+type FileWriter interface {
+	io.WriteCloser
```
toaster, line 16: `FileWriteCloser`

internal/driver/gomobile/file.go
```diff
@@ -1,14 +1,90 @@
...
+
+type fileOpen struct {
+	io.ReadCloser
+	uri  string
+	done func()
+}
+
+//func (f *fileOpen) Save() (io.WriteCloser, error) {
+//	return nativeFileSave(f)
+//}
```
toaster, line 20: Is this an artifact?

internal/driver/gomobile/file.go
```diff
@@ -1,14 +1,90 @@
...
+	uri := uriOrPath
+	if strings.Index(uriOrPath, "://") == -1 {
+		uri = "file://" + uriOrPath
+	}
+
+	file := &fileSave{uri: uri}
+	write, err := nativeFileSave(file)
+	file.WriteCloser = write
+	return file, err
+}
```
toaster, line 65: I thought writing is unsupported on mobile?
If so, please add a clarifying comment (or delete).

andydotxyz: Quite right, I was filling the blanks where really that wasn't necessary. Gone now

PR #882, opened 2020-04-20
Infrastructure/image based tests

toaster:
### Description:

This PR adds shadow support for the software painter and a test helper for asserting equal images.

I need this PR for better tests in #877.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


test/util.go
```diff
@@ -1,10 +1,53 @@
...
 	"fyne.io/fyne"
 	"fyne.io/fyne/internal/cache"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
+// AssertImagesEqual asserts that the master image at given filename (below testdata) and the given image are equal.
+// It fails the test if not. In this case the actual image is written to disk and reported.
+func AssertImagesEqual(t *testing.T, masterFilename string, got image.Image) {
```
andydotxyz, line 21: I think that this method name could be better - as a developer I would expect an equals check to operate on two of the same type. How about AssertImageMatchesMaster/Golden or something similar, though the order of the parameters would be reversed.

test/util.go
```diff
@@ -1,10 +1,53 @@
...
+	wd, err := os.Getwd()
+	require.NoError(t, err)
+	masterPath := filepath.Join("testdata", masterFilename)
+	var expected image.Image
+	_, err = os.Stat(masterPath)
+	if os.IsNotExist(err) {
```
andydotxyz, line 27: If it does not exist how about writing the image out to disk and failing straigt off - possibly even erroring instead of failing, as that is really what is happening.

This would allow the "else" below to drop to the main body of the method where I think it belongs.
Perhaps the writing out of a failed image moves to a helper method and it can be called in this error case and the fail case below.

test/util.go
```diff
@@ -1,10 +1,53 @@
...
+	"testing"
+
 	"fyne.io/fyne"
 	"fyne.io/fyne/internal/cache"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
+// AssertImagesEqual asserts that the master image at given filename (below testdata) and the given image are equal.
```
andydotxyz, line 19: This documentation needs more information with regards to the filename. It should specify that it is loaded from testdata relative to the test - but more importantly the final part "written to disk" should probably specify location as well.

andydotxyz:
Some requests to change inline - but also I had these two thoughts:

1) Shouldn't this be tested? even comparing with an empty image we could assert that the pass/fail/create behaviour works.
2) I wonder if the test canvas should always use the software painter? I am not sure that the additional WindowWithPainter should be needed. I wonder if CanvasWithPainter was a mistake - it should probably hook that up by default, leaving the driver init with an opportunity to specify a different driver if the user so desires?

Those seemed more like conversation pieces rather than direct feedback, sorry for the multple conversations.

toaster:
>     1. Shouldn't this be tested? even comparing with an empty image we could assert that the pass/fail/create behaviour works.

Yeah, I’ll work on this.

>     2. I wonder if the test canvas should always use the software painter? I am not sure that the additional WindowWithPainter should be needed. I wonder if CanvasWithPainter was a mistake - it should probably hook that up by default, leaving the driver init with an opportunity to specify a different driver if the user so desires?

I think the test package’s app/driver/windows/canvas stuff should as easy as possible. I like the idea of using the software painter by default.

toaster:
Unfortunately the test package currently cannot create a software painter due to cyclic dependencies. To resolve all of these we have to come to change either the way the software painter and the test packages are implemented or come to an agreement regarding support of the test package in tests outside of `*_test` packages. And the usages of this are a lot.

PR #887, opened 2020-04-21
Refactoring: select tests

toaster:
### Description:

This PR makes the select tests independent from the implementation details.
This makes the menu refactoring a lot easier.

Also included:
* fix a refresh bug in the shadow implementation
* testApp applies theme to overlays
* move testTheme into test package
* overlay support for software painter
* `test.WaitForThemeToBeApplied()`

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


test/testapp.go
```diff
@@ -50,29 +52,31 @@ func (a *testApp) UniqueID() string {
...
 		return
 	}
 	content.Refresh()
 
-	if wid, ok := content.(fyne.Widget); ok {
-		for _, o := range wid.CreateRenderer().Objects() {
-			a.applyThemeTo(o, canv)
+	switch x := content.(type) {
+	case fyne.Widget:
+		for _, o := range cache.Renderer(x).Objects() {
```
andydotxyz, line 63: Wouldn't `test.WidgetRenderer()` be better, not adding dependency on cache package.

test/util.go
```diff
@@ -96,12 +97,34 @@ func TypeOnCanvas(c fyne.Canvas, chars string) {
 	typeChars([]rune(chars), c.OnTypedRune())
 }
 
+// WaitForThemeToBeApplied waits for the current theme to be applied to the current app.
+func WaitForThemeToBeApplied(t *testing.T) {
```
andydotxyz, line 101: How about making this `ApplyTheme`, take the theme parameter and do the settings apply as well.
Every call of it follows a setTheme, so why not bring it into this method :)

test/util.go
```diff
@@ -96,12 +97,34 @@ func TypeOnCanvas(c fyne.Canvas, chars string) {
...
+}
+
 // WidgetRenderer allows test scripts to gain access to the current renderer for a widget.
 // This can be used for verifying correctness of rendered components for a widget in unit tests.
 func WidgetRenderer(wid fyne.Widget) fyne.WidgetRenderer {
 	return cache.Renderer(wid)
 }
 
+// WithTestTheme runs a function with the testTheme temporarily set.
+func WithTestTheme(t *testing.T, f func()) {
```
andydotxyz, line 116: Open question here (and I know I named it this way, but) with the prefix now would `test.WithTheme(...)` be clear enough?

test/testapp.go
toaster: Well, it would not really reduce the dependency of the package but okay.

test/util.go
toaster: I don’t think so. When reading `test.WithTheme(f)` I read that this is the test helper `WithTheme` and it expects at least a `Theme` to use.

test/testapp.go
andydotxyz: I meant that this has a new dependency on cache, that line of code would go away - hiding the internal details.

test/util.go
andydotxyz: Ah I see your point. I had imagined that but missed the parameter list sorry :)

PR #890, opened 2020-04-21
Add List Widget

stuartmscott:
### Description:
Adds a List Widget to display a List Data Binding.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/internal/driver/glfw/canvas.go
+++ b/internal/driver/glfw/canvas.go
@@ -361,8 +361,7 @@ func (c *glCanvas) paint(size fyne.Size) {
 
 	paint := func(node *renderCacheNode, pos fyne.Position) {
 		obj := node.obj
-		// TODO should this be somehow not scroll container specific?
-		if _, ok := obj.(*widget.ScrollContainer); ok {
+		if _, ok := obj.(fyne.Scrollable); ok {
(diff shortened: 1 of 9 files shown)


internal/driver/glfw/canvas.go
```diff
@@ -361,8 +361,7 @@ func (c *glCanvas) paint(size fyne.Size) {
 
 	paint := func(node *renderCacheNode, pos fyne.Position) {
 		obj := node.obj
-		// TODO should this be somehow not scroll container specific?
-		if _, ok := obj.(*widget.ScrollContainer); ok {
+		if _, ok := obj.(fyne.Scrollable); ok {
```
toaster, line 364: `fyne.Scrollable` does not mean that this is a container with clipped and scrollable content but that this is an object which handles scroll events.
For a number entry with up/down arrows on the right `fyne.Scrollable` might provide a way to increase/decrease using the mouse.

stuartmscott: Good point, thanks

andydotxyz:
Should this be closed now we have #1253 @stuartmscott ?

PR #898, opened 2020-04-23
Refactoring/entry tests

toaster:
This PR refactors the `Entry` tests to not depend on internal implementation details any more. A few exceptions were moved into `entry_internal_test.go`.

The refactored tests use image comparison a lot.

Also a new test helper was added to tap anywhere on a test canvas. It will fail the test if there’s no object to tap there. This allows to tap on pop-ups without knowing the object.


test/util.go
```diff
@@ -49,36 +62,43 @@ func AssertImageMatches(t *testing.T, masterFilename string, img image.Image) bo
...
 	TapAt(obj, fyne.NewPos(1, 1))
 }
 
 // TapAt simulates a left mouse click on the passed object at a specified place within it.
-func TapAt(obj fyne.Tappable, pos fyne.Position) {
-	if focus, ok := obj.(fyne.Focusable); ok {
-		if focus != Canvas().Focused() {
-			Canvas().Focus(focus)
+func TapAt(obj TappableCanvasObject, pos fyne.Position) {
+	c := fyne.CurrentApp().Driver().CanvasForObject(obj)
```
andydotxyz, line 71: This is a big behavioural change. I can currently simulate tapping any canvasobject disconnected from a parent canvas.
That behaviour will crash here.

test/util.go
```diff
@@ -49,36 +62,43 @@ func AssertImageMatches(t *testing.T, masterFilename string, img image.Image) bo
...
+func Tap(obj TappableCanvasObject) {
 	TapAt(obj, fyne.NewPos(1, 1))
 }
 
 // TapAt simulates a left mouse click on the passed object at a specified place within it.
-func TapAt(obj fyne.Tappable, pos fyne.Position) {
-	if focus, ok := obj.(fyne.Focusable); ok {
-		if focus != Canvas().Focused() {
-			Canvas().Focus(focus)
+func TapAt(obj TappableCanvasObject, pos fyne.Position) {
```
andydotxyz, line 70: Not a fan of adding new interfaces that developers won't recognise here - Can't we use Tappable?

PR #905, opened 2020-04-26
Feature: submenus

### LINKED ISSUE #395: Add submenu support
As noticed in #283 we should add submenu support at some point.
attaching a menu to another menu should create this cascade.

toaster:
### Description:

This PR introduces the ability to create menus with submenus.

It also includes:
* a bugfix for menu position and size handling for edge cases.
* a modification to `SelectEntry` to let the pop-up size match the entry size if it changes (like for `Select`)

Fixes #395

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


toaster:
artifact detected … will clean-up

toaster:
Okay, I think I will remove it by extra commit :).

andydotxyz:
I'm not sure I follow the need for `widget.Label{Text: "secondary-tap me for context menu"}` in advanced. It's already available on entry. I see that this has a submenu - but that is now in the menu anyway so I don't think it adds anything.

toaster:
> I'm not sure I follow the need for `widget.Label{Text: "secondary-tap me for context menu"}` in advanced. It's already available on entry. I see that this has a submenu - but that is now in the menu anyway so I don't think it adds anything.

You can check the placement on the far right which also shows that submenus are placed left on insufficient space to the right :).
And on MacOS there is no main menu bar by default.

internal/driver/glfw/menu_darwin.go
```diff
@@ -80,18 +81,22 @@ func addNativeMenu(w *window, menu *fyne.Menu, nextItemID int, prepend bool) int
...
+				nextItemID = addNativeSubMenu(w, nsMenuItem, item.ChildMenu, nextItemID)
+			}
 		}
 		if !item.IsSeparator {
-			action := item.Action // catch
-			callbacks = append(callbacks, func() { w.queueEvent(action) })
-			nextItemID++
+			if action := item.Action; action != nil { // catch action value
+				callbacks = append(callbacks, func() { w.queueEvent(action) })
+				nextItemID++
```
andydotxyz, line 73: I don't know for sure but it seems like nextItemID should not be subject to action != nil check?

internal/driver/glfw/menu_darwin.go
```diff
@@ -100,3 +105,27 @@ func addNativeMenu(w *window, menu *fyne.Menu, nextItemID int, prepend bool) int
...
+			C.CString(item.Label),
+			C.int(nextItemID),
+			C.int(-1),
+			C.bool(item.IsSeparator),
+		)
+		if item.ChildMenu != nil {
+			nextItemID = addNativeSubMenu(w, nsMenuItem, item.ChildMenu, nextItemID)
+		}
+		if !item.IsSeparator {
+			if action := item.Action; action != nil { // catch action value
```
andydotxyz, line 98: this callback handling is a duplicate of above, maybe shouldn't be?

andydotxyz:
> shows that submenus are placed left on insufficient space to the right :)

Fair enough. Can you find a place to put it that does not take up so much space?

toaster:
> Fair enough. Can you find a place to put it that does not take up so much space?

I hoped, you would suggest one :). I’ll have a look.

internal/driver/glfw/menu_darwin.go
toaster: oh, yes … this happens when I get interrupted.

internal/driver/glfw/menu_darwin.go
toaster: I’ll DRY up …

andydotxyz:
> I hoped, you would suggest one :). I’ll have a look.

Maybe buttons -> Actions and it's added to the bottom of that panel?

internal/driver/glfw/menu_darwin.go
```diff
@@ -86,10 +86,13 @@ func createNativeMenu(w *window, menu *fyne.Menu, nextItemID int) (unsafe.Pointe
 
 func registerCallback(w *window, item *fyne.MenuItem, nextItemID int) int {
 	if !item.IsSeparator {
-		if action := item.Action; action != nil { // catch action value
-			callbacks = append(callbacks, func() { w.queueEvent(action) })
-			nextItemID++
-		}
+		action := item.Action // catch action value
```
andydotxyz, line 89: I'm unsure here, the action is being captured on menu creation so it cannot change - should instead the item be captured so the latest value for Action is used in the callback? I only noticed it as the nil check is on a value that doesn't change.

andydotxyz:
Just 1 inline comment if you could check it out.

I also realised something (from a previous commit I think): macOS apps use "Preferences..." instead of "Settings" in the label - should we be rewriting it,  should we move to Preferences or should we check for both?

Issue #899, opened 2020-04-23
TabContainer add a spacer (horizontal line)


s77rt:
Hi,
It would be nice if we can add an horizontal line to the tab container to separate between different tabs :)

andydotxyz:
hmm, not too sure about this - can you add more information about why this design would be desirable over what exists?

s77rt:
"desirable over what exists?"
not desirable over what exists, this won't replace anything, it would be just an added feature.
it's good for ux, to list some tabs then a spacer then other tabs
so the user knows that the first tabs have something common and so the last tabs...

check [github's menu](https://i.postimg.cc/2jt9K38N/github-menu.png), it have some items and a spacer between every few common items

zyfdegh:
An sneaky way to do this can be

```go
import (
    "fyne.io/fyne/layout"
    "fyne.io/fyne/widget"
)

tabs := widget.NewTabContainer(
    widget.NewTabItem("____", layout.NewSpacer()),
)
```
Check the [image here](https://postimg.cc/BX99chVN), hope it helps.

andydotxyz:
I see what you are referring to. Toolbars have spacers and separators. Menus (in 1.3 will) have separators. Tabcontainers do not as they are designed to have the same sort content throughout - I.e. top level application areas or files in a panel.

andydotxyz:
See: https://material.io/components/tabs#placement

We decided to move to a more prominent “current tab” highlight and also that scrolling isn’t the way we want to handle overflow, otherwise it’s the spec

andydotxyz:
Let's stick to the spec for now.

PR #907, opened 2020-04-26
New layout to demo app

andydotxyz:
### Description:
As discussed in Slack shuffling round content to have fewer tabs on widget screen

### Checklist:

- [ ] Tests included. <- just moving code really
- [x] Tests all pass.


cmd/fyne_demo/screens/container.go
```diff
@@ -0,0 +1,109 @@
...
+	"fyne.io/fyne/theme"
+	"fyne.io/fyne/widget"
+)
+
+// ContainerScreen loads a tab panel for containers and layouts
+func ContainerScreen() fyne.CanvasObject {
+	return widget.NewTabContainer(
+		widget.NewTabItem("Split", makeSplitTab()),
+		widget.NewTabItem("Scroll", makeScrollTab()),
+		// layouts
```
stuartmscott, line 19: If we had tab grouping (#899) we could surface this distinction

cmd/fyne_demo/screens/container.go
```diff
@@ -0,0 +1,109 @@
...
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/theme"
+	"fyne.io/fyne/widget"
+)
+
+// ContainerScreen loads a tab panel for containers and layouts
+func ContainerScreen() fyne.CanvasObject {
+	return widget.NewTabContainer(
+		widget.NewTabItem("Split", makeSplitTab()),
+		widget.NewTabItem("Scroll", makeScrollTab()),
```
stuartmscott, line 18: Tabber is missing from this list, but since the whole thing is in a tabber I think it's ok

cmd/fyne_demo/screens/container.go
```diff
@@ -0,0 +1,109 @@
...
+		// layouts
+		widget.NewTabItem("Border", makeBorderLayout()),
+		widget.NewTabItem("Box", makeBoxLayout()),
+		widget.NewTabItem("Grid", makeGridLayout()),
```
stuartmscott, line 22: Center, Fixed Grid, and Max layouts are missing here. (Form Layout too, but Form is shown in widget, so I think it's ok)

cmd/fyne_demo/screens/container.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return fyne.NewContainerWithLayout(layout.NewGridLayout(2),
+		box1, box2, box3, box4)
+}
+
+func makeScrollTab() fyne.CanvasObject {
+	list := widget.NewHBox()
+	list2 := widget.NewVBox()
+
+	for i := 1; i <= 20; i++ {
```
stuartmscott, line 72: On my external display 20 is not enough buttons to reach the bottom, can we increase this?

cmd/fyne_demo/screens/container.go
andydotxyz: I removed the FixedGrid because it is going away/deprecated and confusing.
Unfortunately max and center look almost the same. I think center is a good shout though. Max is less use externally.

cmd/fyne_demo/screens/container.go
Bluebugs: Maybe with a slider to define it or something like that.

cmd/fyne_demo/main.go
```diff
@@ -84,8 +84,9 @@ func main() {
 
 	tabs := widget.NewTabContainer(
```
stuartmscott, line 85: I agree with Welcome as the first, and Advanced as the last. Windows (&Dialogs) towards the end also makes sense. Containers (&Layouts) are about surfacing to the users so being next to Windows makes sense.
I'm not sure about Widgets and Graphics. Graphics (canvas) are the building blocks so maybe that should be right after Welcome. Widgets build on Graphics so I'd put that next.

- Welcome
- Graphics
- Widgets
- Containers
- Windows
- Advanced

(just noticed it goes W, not W, W, not W, W, not W which is appealing for absolutely no reason :P)

cmd/fyne_demo/main.go
```diff
@@ -84,8 +84,9 @@ func main() {
...
 		widget.NewTabItemWithIcon("Welcome", theme.HomeIcon(), welcomeScreen(a)),
-		widget.NewTabItemWithIcon("Widgets", theme.ContentCopyIcon(), screens.WidgetScreen()),
+		widget.NewTabItemWithIcon("Widgets", theme.CheckButtonCheckedIcon(), screens.WidgetScreen()),
 		widget.NewTabItemWithIcon("Graphics", theme.DocumentCreateIcon(), screens.GraphicsScreen()),
+		widget.NewTabItemWithIcon("Containers", theme.ViewRestoreIcon(), screens.ContainerScreen()),
```
stuartmscott, line 89: Would "Layout" be a better name for this, since it is more about laying out components than containing them?

stuartmscott: Actually disregard - If accordion, list, table, tree will be in here too then containers is fine

cmd/fyne_demo/screens/container.go
andydotxyz: I don't understand - add a slider to the scroll demo to define how many items are inside a container?

cmd/fyne_demo/main.go
andydotxyz: I had the same thought, but then I wasn't sure about what we should be presenting if people thought of this as order of importance. The canvas is really an enabling API rather than the main thrust for the library?

cmd/fyne_demo/screens/container.go
andydotxyz: Haha, that did occur to me - but without the context it would be meaningless.
Honestly I didn't like this a huge amount because it's not clear that some are layouts and that is an important distinction.
But when the layouts were all in a layout tab we had 3 tabs deep navigation which seemed very wrong.
We could put 2 tab controllers next to each other but that also feels bad.
Not sure that there is a perfect solution to this right now, maybe it needs more thought for layout?

cmd/fyne_demo/main.go
stuartmscott: Yes I see your point, we're not just making a graphics library, we're trying to make a library for creating beautiful UIs and widgets are the heart of that.
I saw the demo kind of like a tour, starting with welcome, basic building blocks, functional elements, positional elements, and so on. Maybe it should match tour.fyne.io, but I noticed that goes Introduction, Basics, Canvas, Layout, Widgets :/

cmd/fyne_demo/screens/container.go
stuartmscott: This would be a nice demo for bindings, not sure it is necessary here

cmd/fyne_demo/screens/container.go
stuartmscott: Hmm the deeper the navigation the less discoverable the content - definitely not what we want for a demo. Two tabbers next to each other feels bad to me too - there is no visual separation between the groups of tabs making it a confusing UI. You could use a splitter as that has a big divider, but again that doesn't feel right, maybe we should have Layouts and Containers as separate tabs?

andydotxyz: Too many tabs I think. I am happy enough with this as a compromise for now, this app will continue to evolve.

cmd/fyne_demo/main.go
andydotxyz: Well, I went with it - consistency with the tour is a good idea, perhaps some language tweaking on the other side will even that up

andydotxyz:
Pushed one additional item that is to move entries off the text tab back to input and move the textgrid to text tab as it was in the wrong place.

The entries didn't really add much to the wrap/truncate demo - lorem is still in there to show off :)

cmd/fyne_demo/screens/container.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func makeCell() fyne.CanvasObject {
+	rect := canvas.NewRectangle(&color.RGBA{128, 128, 128, 255})
+	rect.SetMinSize(fyne.NewSize(30, 30))
+	return rect
+}
+
+func makeCenterLayout() *fyne.Container {
+	middle := widget.NewButton("CenterLayout", func() {})
```
stuartmscott, line 72: nit: to be consistent this should be a label like the other ones

cmd/fyne_demo/screens/widget.go
```diff
@@ -30,6 +29,14 @@ func makeButtonTab() fyne.Widget {
...
 	disabled.Disable()
 
+	return widget.NewVBox(
+		widget.NewButton("Button (text only)", func() { fmt.Println("tapped text button") }),
+		widget.NewButtonWithIcon("Button (text & icon)", theme.ConfirmIcon(), func() { fmt.Println("tapped text & icon button") }),
+		disabled,
+	)
+}
+
+func makeTextGrid() *widget.TextGrid {
```
stuartmscott, line 39: You could expand this TextGrid demo if you wanted to show it off more (two lines is a little small, especially if it follows Lorem Ipsum)

cmd/fyne_demo/screens/container.go
andydotxyz: If it were a label you would not be able to see the edge of the widget. so I made it a button

cmd/fyne_demo/screens/widget.go
andydotxyz: True. I will update that in a different commit where TestGrid gets some love

stuartmscott:
Travis failed because of a golden image test:
```
--- FAIL: TestEntry_Select (1.12s)
    --- FAIL: TestEntry_Select/replace (0.18s)
        util.go:45: 
            	Error Trace:	util.go:45
            	            				entry_test.go:1272
            	Error:      	Not equal: 
            	Test:       	TestEntry_Select/replace
            	Messages:   	Image did not match master. Actual image written to /Users/travis/gopath/src/fyne.io/fyne/widget/testdata/failed/entry_selection_initial.png.
        util.go:45: 
            	Error Trace:	util.go:45
            	            				entry_test.go:1282
            	Error:      	Not equal: 
            	Test:       	TestEntry_Select/replace
            	Messages:   	Image did not match master. Actual image written to /Users/travis/gopath/src/fyne.io/fyne/widget/testdata/failed/entry_selection_replace.png.
```
@toaster can Travis surface the failed images?

andydotxyz:
Looks like the failure was a temporary one, kicked travis as it was just 1 of the jobs that failed

toaster:
> @toaster can Travis surface the failed images?

Jein (German for Yes and No). It can copy artifacts to S3 (or maybe anywhere with the artifacts plugin) but AFAIK it doesn’t work for PR builds.

toaster:
> Looks like the failure was a temporary one, kicked travis as it was just 1 of the jobs that failed

The image based tests revealed some race conditions regarding rendering with the test infrastructure. That never was a problem because no one ever rendered anything there :).
I already fixed one of these in #925 but there are still some left :/.

PR #914, opened 2020-04-28
Adds Button Alignment and Icon Placement

### LINKED ISSUE #895: Button Alignment and Ordering
1. It is not possible to change the alignment of the components of a button.
2. It is not possible to change the relative positions of the components of a button.

Currently not possible without creating a Custom Widget that embeds Button and overrides Button.CreateRenderer

1. The ability to set the alignment of a button to:
    - Centered - (default) - the current behaviour
        ```
        +------------------------+
        |       Icon Text        |
        +------------------------+
        ```
    - Leading - components are aligned to the button's leading edge
        ```
        +------------------------+
        | Icon Text              |
        +------------------------+
        ```
    - Trailing - components are aligned to the button's trailing edge
        ```
        +------------------------+
        |              Icon Text |
        +------------------------+
        ```
    - Justified, or Spaced - components are aligned to opposite edges
        ```
        +------------------------+
        | Icon              Text |
        +------------------------+
        ```
2. The ability to set the relative positions of the components of a button to:
    - Icon Above Text
        ```
        +------------------------+
        |          Icon          |
        |          Text          |
        +------------------------+
        ```
    - Icon Below Text
        ```
        +------------------------+
        |          Text          |
        |          Icon          |
        +------------------------+
        ```
    - Icon Leading Text
        ```
        +------------------------+
        |       Icon Text        |
        +------------------------+
        ```
    - Icon Trailing Text
        ```
        +------------------------+
        |       Text Icon        |
        +------------------------+
        ```

stuartmscott:
### Description:
Adds two public fields to Button for controlling the alignment and ordering of its subcomponents.

This PR also introduces Golden Image Tests which required sorting the button tests into button_internal_test.go (widget package) and button_test.go (widget_test package).

Fixes #895

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
This is a huge amount of code aht the massive switch statement shows how much this increases complexity. Given that all buttons are the same height right now (not including tab bar) I think that the vertical arrangement option is a much bigger change than just adding the code.

From the material spec:

> Don’t vertically align an icon and text in the center of a contained button.

https://material.io/components/buttons#contained-button

widget/button.go
```diff
@@ -29,45 +29,152 @@ func (b *buttonRenderer) padding() fyne.Size {
...
+		return
+	}
+	iconSize := fyne.NewSize(theme.IconInlineSize(), theme.IconInlineSize())
+	labelSize := b.label.MinSize()
+	padding := b.padding()
+	innerSize := size.Subtract(padding)
+	innerOffset := fyne.NewPos(padding.Width/2, padding.Height/2)
+	var iconPos fyne.Position
+	var labelPos fyne.Position
+	if hasLabel {
```
andydotxyz, line 67: I can't help but feel that this code could be simplified if an intermediate container was added.
Conceptually a bigger change, but I think it will be vastly simpler.

I.e. a content box that takes an icon and/or text in the right order... then all you have to do is align it.
Trying to find ways to avoid large indents of layout code like this

widget/button.go
```diff
@@ -29,45 +29,152 @@ func (b *buttonRenderer) padding() fyne.Size {
...
+		b.icon.Move(alignedPosition(b.button.Alignment, padding, iconSize, size))
+		b.icon.Resize(iconSize)
+	}
+}
+
+func alignedPosition(align ButtonAlign, padding, objectSize, layoutSize fyne.Size) (pos fyne.Position) {
+	pos.Y = (layoutSize.Height - objectSize.Height) / 2
+	switch align {
+	case ButtonAlignCenter:
+		// +------------------------+
```
andydotxyz, line 151: Hopefully these comments aren't really needed?
Usually they are helpful when the code is not clear, but this seems clear.

widget/button.go
```diff
@@ -29,45 +29,152 @@ func (b *buttonRenderer) padding() fyne.Size {
...
+		pos.X = layoutSize.Width - objectSize.Width - padding.Width/2
+	}
+	return
+}
+
+// applyAlignment updates the button label alignment
+func (b *buttonRenderer) applyAlignment() {
+	switch b.button.Alignment {
+	case ButtonAlignLeading:
```
andydotxyz, line 172: If the suggestion about a content container is picked up then this code goes away as the text is always minsize

widget/button.go
stuartmscott: Switched to using an HBoxLayout to layout at min size when both icon and label are present, then reused the alignedPosition to offset the items correctly. The code is definitely shorter.

The downside with this approach is that HBoxLayout forces all children to be the same height, and since the min height of the label is 21, the icon is resized to 20x21, causing the golden image tests and TestButtonRenderer_Layout_Stretch to fail. 1 pixel is not that noticeable, but if the text size is increased, this will cause the button image to get stretched in height.

widget/button.go
stuartmscott: Removed

widget/button.go
andydotxyz: Oh bummer sorry, didn't see that coming. 
I think this may be a bug in icon - it should always maintain ratio in the space.
What if the icon's default image fill is ImageFillContain?

stuartmscott: Fixed now, thanks to your work on #919

PR #923, opened 2020-05-01
fix tests

toaster:
### Description:

Small refactorings and a fix for the darwin menu tests.

### Checklist:

- ~[ ] Tests included.~
- [x] Tests all pass.

### DIFF
--- a/layout/boxlayout.go
+++ b/layout/boxlayout.go
@@ -107,35 +107,32 @@ func (g *boxLayout) Layout(objects []fyne.CanvasObject, size fyne.Size) {
 // For a BoxLayout this is the width of the widest item and the height is
 // the sum of of all children combined with padding between each.
 func (g *boxLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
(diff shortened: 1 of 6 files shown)


layout/boxlayout.go
```diff
@@ -107,35 +107,32 @@ func (g *boxLayout) Layout(objects []fyne.CanvasObject, size fyne.Size) {
 // For a BoxLayout this is the width of the widest item and the height is
 // the sum of of all children combined with padding between each.
 func (g *boxLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
-	spacerCount := 0
 	minSize := fyne.NewSize(0, 0)
-	added := false
+	addPadding := false
```
stuartmscott, line 111: I think it would be cleaner to remove this flag and just do:
```suggestion
	for i, child := range object {
```
```suggestion
	if i != 0 {
```

toaster: This won’t work because the first element might be ignored (invisible or a spacer).

PR #924, opened 2020-05-02
Update baseline DPI to 120 for autodetecting systems

### LINKED ISSUE #862: Applications appear to large on Linux
Hi,
First, thanks for your all your work, i like it :)
I just noticed that the scaling described here 
https://github.com/fyne-io/fyne/wiki/Scaling-Revisited

may not be the best practice, as I'm having trouble using my apps on Linux, they are so big and sometimes does not fit on the screen; while on Windows works just fine

Is it possible to disable the feature that adjust the scaling factor for Linux and keep it just like Windows/MacOS ? (or even better control the scaling factor from within the app itself)

andydotxyz:
### Description:
Update the baseline DPI from 96 to 120 for Linux, as discussed in #862.
Also refactor out the scale handling to a separate file that is well tested

Fixes #862

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/scale.go
+++ b/internal/driver/glfw/scale.go
@@ -1,0 +1,61 @@
+package glfw
+
+import (
+	"math"
+	"os"
+	"strconv"
+
+	"fyne.io/fyne"
+)
+
+const (
+	baselineDPI = 120.0
+	scaleEnvKey = "FYNE_SCALE"
+)
+
+func calculateDetectedScale(widthMm, widthPx int) float32 {
+	dpi := float32(widthPx) / (float32(widthMm) / 25.4)
+	if dpi > 1000 || dpi < 10 {
+		dpi = baselineDPI
+	}
+
+	scale := float32(float64(dpi) / baselineDPI)
+	if scale < 1.0 {
+		return 1.0
+	}
+	return scale
+}
+
+func calculateScale(user, system, detected float32) float32 {
+	if user == fyne.SettingsScaleAuto {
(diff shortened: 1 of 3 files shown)


internal/driver/glfw/scale.go
```diff
@@ -0,0 +1,61 @@
...
+	if user == fyne.SettingsScaleAuto {
+		user = 1.0
+	}
+
+	if system == fyne.SettingsScaleAuto {
+		system = detected
+	}
+
+	raw := system * user
+	return float32(math.Round(float64(raw*10.0))) / 10.0
```
toaster, line 39: The `.0` is not necessary here because the compiler infers the numbers’ types.

internal/driver/glfw/scale.go
```diff
@@ -0,0 +1,61 @@
...
+		scale, err := strconv.ParseFloat(env, 32)
+		if err == nil && scale != 0 {
+			return float32(scale)
+		}
+		fyne.LogError("Error reading scale", err)
+	}
+
+	if env != "auto" {
+		setting := fyne.CurrentApp().Settings().Scale()
+		if setting != fyne.SettingsScaleAuto && setting != 0.0 {
```
toaster, line 55: unnecessary `.0`

internal/driver/glfw/scale.go
andydotxyz: May be unnecessary but I like having decimal places when talking about floats :)

Issue #930, opened 2020-05-04
Initially shown file dialog does not fill content


andydotxyz:
### Describe the bug:
When first shown, and until resize, the file dialog content is not filling the scrollable area.
This is new

### To Reproduce:
Steps to reproduce the behaviour:
1. fyne_demo
2. Windows
3. ShowFileOpen

Looks like it's just the clipping that is not correct the content seems about right and scrolls as expected.

### Screenshots:
<img width="790" alt="Screenshot 2020-05-04 at 15 48 37" src="https://user-images.githubusercontent.com/294436/80979269-06f3e900-8e1f-11ea-9b4e-52ea8018e0fd.png">

### Device (please complete the following information):
 - OS: macOS 10.15
 - Version: latest
 - Go version: 1.12
 - Fyne version: develop

stuartmscott:
I have a feeling this is related to #834, try adding logs to your renderer.Refresh and renderer.Layout functions then reproducing this issue.

andydotxyz:
I will try... but this has happened in the last week or two, but the other issue goes back a month or more...

andydotxyz:
I think this relates to scroller. It is also visible on the containers tab of fyne_demo when you look at the scroll tab

stuartmscott:
Interestingly this is fixed by resizing the window, opening the demo on a second monitor, or setting the demo window size before adding the content;
```go
w := a.NewWindow("Fyne Demo")
w.Resize(fyne.NewSize(1,1))
```
Maybe glScissor can't be set before the window has position/size.

andydotxyz:
I agree that it's weird that those actions fix it.
However I don't think it can be a glScissor issue as this has worked for nearly a year before it started breaking this month.

andydotxyz:
I was randomly testing other stuff and noticed that this goes away with latest develop if you `-tags no_native_menus`. It does flicker sometimes noticeably on load...
So I suspect that the menu bar is causing a refresh that fixes an underlying issue that has gone un-noticed.
Using macOS native menus this refresh is skipped and the screen does not re-calculate something.
It may be a duplicate of #938 after all...

stuartmscott:
Interesting, so it is adding an extra refresh? But this is a patch fix - the root cause is much deeper.
The fact that resizing the window _before_ creating and showing a file dialog fixes this issue makes me think that the scroller is dependent on some internal state (possibly around scale/size) is wrong and which a resize fixes.

andydotxyz:
[...]

PR #932, opened 2020-05-05
Truncate Dialog File Name and Extension

stuartmscott:
### Description:
Fixes file dialog text issues with truncation

Fixes https://gophers.slack.com/archives/CK5U4BU87/p1588630548028400

Screen Shot 2020-05-04 at 6 24 09 PM

Screen Shot 2020-05-04 at 6 21 00 PM

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


stuartmscott:
Extension still extends beyond the icon file bounds resulting in grey-on-blue text. 10 is also quite an arbitrary number and may not work at all resolutions. Unfortunately canvas.Text does not implement text wrapping. widget.Label has text wrapping, but it renders text in theme.TextColor and does not support custom text color.

dialog/file.go
```diff
@@ -42,9 +42,12 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 				f.open.Enable()
 			}
 		}
+		saveName.Wrapping = fyne.TextTruncate
```
andydotxyz, line 45: We should not truncate an entry, the user needs to be able to edit the whole thing. A ScrollContainer is probably the right solution.

dialog/fileitem.go
```diff
@@ -57,6 +57,9 @@ func fileParts(path string) (name, ext string) {
 	if len(ext) > 1 {
 		ext = ext[1:]
 	}
+	if len(ext) > 10 {
```
andydotxyz, line 60: From your screenshots it looks unlikely that more than 5 characters will fit... it’s probably better to have it sometimes truncates too much (iiiii) than still overlap at times (mmmmmmmmmm). I’d also consider defining a local const or naming a variable for this number.

dialog/file.go
stuartmscott: Ack

dialog/fileitem.go
stuartmscott: Yeah okay. 5 is another arbitrary number and may not work for all resolutions but you're right that it is better to over truncate than spill over into other elements. I can't think of a better solution

stuartmscott:
5 char limit looks better than 10

Screen Shot 2020-05-05 at 8 12 47 AM

Unfortunately there is something weird happening with the scroller - it doesn't initially respond to events, but then suddenly works. @andydotxyz can you try locally?

stuartmscott:
Also the scroll bar is far too thick

dialog/fileitem.go
andydotxyz: I agree it is arbitrary, but not sure what you mean about not working on some resolutions? This is all scaled proportionately so the only thing that would change the proportions is a custom theme that changes font size or icon size.

andydotxyz:
I can't seem to replicate the issue with scrolling - perhaps you can provide more detailed steps as there are now a few scrollers on the screen. Additionally I expect there could be a difference between initial state and loading other directories / file names.

The entry scroll bar works as expected here - that large size is the hover state. Was that showing by default at your end?

dialog/fileitem.go
stuartmscott: Yup, I was thinking that really small screens, or really large screens could surface the limitations of this approach, but you're right that it _should_ all get scaled properly. Custom themes pose more of an issue, but I don't think we can protect against all customizations. This arbitrary number fixes _most_ issues, so let's go with it for now, and we can always improve it later

stuartmscott:
I captured a video, the first part of it I am interacting with the scroller and scrollbar but does not respond. Then it suddenly starts to work :/

I should turn on visible taps and scrolls so you can see, but rest assured I'm tapping, scrolling, and dragging!

Open Dialog
out

Save Dialog
This seems to not work, until I tap the entry and it gains focus, then scroll works
out

andydotxyz:
Sorry I realised this one got left behind. Is there something additional we need to do with the scrolling issue, or can it be replicated elsewhere?

stuartmscott:
I'm still trying to figure out the scroll issue, but if it is related to the scroll issue in #930 then I think we can get this in and fix the issue separately.

andydotxyz:
Hmm, I wondered the same but it might be wishful thinking as the content looks correct in this case.

andydotxyz:
I’d recommend rebasing on develop just in case :)

stuartmscott:
You're right, I could not repro after merging develop

andydotxyz:
Good to merge then?

andydotxyz:
Green from Travis

PR #937, opened 2020-05-05
Introduce simple mime type handling and add an api for devs to get file icons

### LINKED ISSUE #405: Add Mime Type Support to file pickers
Hello -

I would like to propose that we add some mime type support that can be provided through fyne and especially helpful for things like a file chooser that we have in upcoming releases after we have a data binding and list/grid widgets that cache.  I will begin the work on this and follow it up with a PR.

okratitan:
### Description:
This adds simple mimetype handling to the fyne file dialog.  It also exposes the file icon creation so that developers who may wish to have icons for their files can retrieve those icons in a uniform way that Fyne does.

Tests still need to be added.

dialog.NewFileIcon returns a canvas object which is the constructed file icon - is dialog the right place for this API though?

The icons look poor with text overlaid.

Fixes #405 

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

#### Where applicable:


andydotxyz:
Looks like there is a flaw in the arc handling of oksvg. The files are attempting to make a circle out of 4 arcs but instead it is drawing separate circles causing the issue you see.
Usually raising a ticket against the upstream project with the icon and a rendering they are usually quite helpful.

dialog/file.go
```diff
@@ -165,7 +165,7 @@ func (f *fileDialog) refreshDir(dir string) {
 		if file.IsDir() {
 			icons = append(icons, f.newFileItem(theme.FolderIcon(), itemPath))
 		} else {
-			icons = append(icons, f.newFileItem(theme.FileIcon(), itemPath))
+			icons = append(icons, f.newFileItem(nil, itemPath))
```
andydotxyz, line 168: Passing nil as information is a bit of an anti-pattern. From what I can tell the icon parameter is now initialising a cache so it probably should be an internal detail.
Instead the folder icon could be handled by the mime code so we don't need this special case.

dialog/fileicon.go
```diff
@@ -0,0 +1,88 @@
...
+		res = theme.FileImageIcon()
+	case "text":
+		res = theme.FileTextIcon()
+	case "video":
+		res = theme.FileVideoIcon()
+	default:
+		res = theme.FileIcon()
+	}
+	img := canvas.NewImageFromResource(res)
+	extText := canvas.NewText(i.mimeSubType, theme.BackgroundColor())
```
andydotxyz, line 35: I don't think that mimeSubType is the same as extension.
This is user facing so we should present what they are familiar with... the visible extension.

dialog/fileitem.go
```diff
@@ -35,44 +41,67 @@ func (i *fileDialogItem) TappedSecondary(_ *fyne.PointEvent) {
 }
 
 func (i *fileDialogItem) CreateRenderer() fyne.WidgetRenderer {
-	img := canvas.NewImageFromResource(i.icon)
+	var img fyne.CanvasObject
+	if i.icon == nil {
+		mimeSub := i.mimeSubType
+		if len(mimeSub) > 5 {
```
andydotxyz, line 47: This doesn't seem to be the right place to trim a string - the trimming is a display detail and so should not pass in partial information

dialog/fileitem.go
```diff
@@ -35,44 +41,67 @@ func (i *fileDialogItem) TappedSecondary(_ *fyne.PointEvent) {
...
-		img: img, text: text, ext: extText, objects: []fyne.CanvasObject{img, text, extText}}
+		img: img, text: text, objects: []fyne.CanvasObject{img, text}}
 }
 
 func (i *fileDialogItem) isDirectory() bool {
 	return i.icon == theme.FolderIcon() || i.icon == theme.FolderOpenIcon()
 }
 
-func fileParts(path string) (name, ext string) {
+func fileParts(path string) (name, ext, mimeType, mimeSubType string) {
```
andydotxyz, line 65: I don't think that adding 2 more return types to `fileParts` is clean. A new method `fileMime` would be cleaner I think.

dialog/fileitem.go
```diff
@@ -35,44 +41,67 @@ func (i *fileDialogItem) TappedSecondary(_ *fyne.PointEvent) {
...
 	name = filepath.Base(path)
 	ext = filepath.Ext(name[1:])
 	name = name[:len(name)-len(ext)]
 
-	if len(ext) > 1 {
-		ext = ext[1:]
+	mimeTypeFull := mime.TypeByExtension(ext)
+	if mimeTypeFull == "" {
```
andydotxyz, line 71: How often does this happen?
If we open a directory of 1000 files then opening all the files to peek will likely be very slow.
Perhaps it would be better to leave out this more detailed check until we know it is required?

dialog/fileitem.go
```diff
@@ -35,44 +41,67 @@ func (i *fileDialogItem) TappedSecondary(_ *fyne.PointEvent) {
...
+
+	mimeTypeSplit := strings.Split(mimeTypeFull, "/")
+	mimeType = mimeTypeSplit[0]
+	mimeSubType = mimeTypeSplit[1]
+
 	return
 }
 
 func (f *fileDialog) newFileItem(icon fyne.Resource, path string) *fileDialogItem {
-	name, ext := fileParts(path)
```
andydotxyz, line 64: Perhaps moving this code into fileDialogItem is tidier? As far as I can see they are only used in the CreateRenderer of the child object, so don't have to be passed in and managed as state of the item.
I might be wrong, just looks that way from this diff

theme/icons.go
```diff
@@ -305,11 +311,36 @@ func WarningIcon() fyne.Resource {
 	return warning
 }
 
-// FileIcon returns a resource containing the standard file icon for the current theme
+// FileIcon returns a resource containing the appropriate file icon for the current theme
```
andydotxyz, line 314: I think "default" is more accurate than "appropriate" as this method call is not dynamic so it can't really tell what is appropriate...

andydotxyz:
Oh also the icons should be added to the graphics tab in demo :)

andydotxyz:
Nice work though. Looks like a great balance of info and visuals.
I like the layout but when you open the images in the larger graphics demo tab you'll see that alignment isn't quite right - particularly noticeable on the image file.

dialog/fileicon.go
```diff
@@ -0,0 +1,88 @@
...
+	img     *canvas.Image
+	objects []fyne.CanvasObject
+}
+
+func (s fileIconRenderer) Layout(size fyne.Size) {
+	iconAlign := (size.Width - fileIconSize) / 2
+	s.img.Resize(fyne.NewSize(fileIconSize, fileIconSize))
+	s.img.Move(fyne.NewPos(iconAlign, 0))
+	s.ext.Resize(fyne.NewSize(fileIconSize, fileTextSize))
+	ratioDown := 0.45
```
toaster, line 67: Maybe this should be a `const`?

dialog/fileitem.go
```diff
@@ -35,44 +41,67 @@ func (i *fileDialogItem) TappedSecondary(_ *fyne.PointEvent) {
...
 }
 
-func fileParts(path string) (name, ext string) {
+func fileParts(path string) (name, ext, mimeType, mimeSubType string) {
 	name = filepath.Base(path)
 	ext = filepath.Ext(name[1:])
 	name = name[:len(name)-len(ext)]
 
-	if len(ext) > 1 {
-		ext = ext[1:]
```
andydotxyz, line 58: P.S. remoiving this line iis what seems to be making the tests fail...

dialog/fileitem.go
okratitan: Only on files with no extensions... so not that often in a directory with 1000s of files hopefully.

dialog/fileicon.go
```diff
@@ -0,0 +1,127 @@
+package dialog
```
stuartmscott, line 1: nit: please order this file like #827 to make it easier to read

theme/icons/file-application.svg
```diff
@@ -0,0 +1,31 @@
...
+</svg>
```
stuartmscott, line 31: Still a lot of metadata
```svg
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="24" height="24" viewBox="0 0 24 24">
    <path style="fill:#000000;stroke-width:0.758587" d="m 8.5489811,5.1365526 c 0.6291226,0 1.136968,0.5086607 1.136968,1.1387931 0,0.6301315 -0.5078454,1.138793 -1.136968,1.138793 -0.636702,0 -1.1369682,-0.5086615 -1.1369682,-1.138793 0,-0.6301324 0.5078458,-1.1387931 1.1369682,-1.1387931 z M 8.12769,4.1000094 c -0.054902,0 -0.1013548,0.038698 -0.1098008,0.090302 L 7.9376517,4.7579098 C 7.8109575,4.8138129 7.6884879,4.8826089 7.5786873,4.9600091 L 7.0550226,4.7450081 c -0.046454,0 -0.1013542,0 -0.130916,0.055903 L 6.5017968,5.5448094 c -0.025338,0.047296 -0.016893,0.1031993 0.025338,0.1375997 L 6.974783,6.0350078 c -0.00845,0.073098 -0.012669,0.1418993 -0.012669,0.2149999 0,0.073098 0.00422,0.1418994 0.012669,0.214999 L 6.5271353,6.8176064 c -0.038008,0.034399 -0.050677,0.090302 -0.025338,0.1375997 l 0.4223099,0.7438975 c 0.025338,0.055903 0.080239,0.055903 0.1309161,0.055903 l 0.523664,-0.2150057 c 0.1098007,0.077396 0.2280475,0.1504996 0.3589638,0.2020989 l 0.080237,0.5675985 c 0.00844,0.051604 0.050679,0.090302 0.1098008,0.090302 h 0.8446195 c 0.046453,0 0.092906,-0.038698 0.101355,-0.090302 L 9.1539069,7.7420995 C 9.2806,7.6861968 9.3946235,7.6174007 9.5086472,7.5400006 l 0.5194388,0.2149989 c 0.05491,0 0.109802,0 0.139363,-0.055903 L 10.58976,6.9551982 C 10.6151,6.9079022 10.60243,6.8519989 10.56442,6.8175984 L 10.11255,6.4649991 c 0.0085,-0.073098 0.01689,-0.1418994 0.01689,-0.214999 0,-0.073098 -0.0043,-0.1418993 -0.01689,-0.2149999 l 0.447648,-0.3525988 c 0.03801,-0.0344 0.05068,-0.090302 0.02535,-0.1375998 l -0.42232,-0.7438974 c -0.02534,-0.055903 -0.08024,-0.055903 -0.135141,-0.055903 L 9.5086472,4.9599994 C 9.3946235,4.8826031 9.2806,4.8094997 9.1496836,4.7579005 L 9.069441,4.1903019 C 9.065281,4.1386979 9.018766,4.0999999 8.97231,4.0999999 H 8.12769 M 13,9 h 5.5 L 13,3.5 V 9 M 6,2 h 8 l 6,6 v 12 a 2,2 0 0 1 -2,2 H 6 C 4.89,22 4,21.1 4,20 V 4 C 4,2.89 4.89,2 6,2" />
</svg>
```

theme/icons/file-video.svg
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
stuartmscott, line 1: The camcorder looks a little squashed, I think it could be a little better if the left side moved in a bit away from the edge

dialog/file.go
```diff
@@ -164,9 +174,11 @@ func (f *fileDialog) refreshDir(dir string) {
 
 		itemPath := filepath.Join(dir, file.Name())
 		if file.IsDir() {
-			icons = append(icons, f.newFileItem(theme.FolderIcon(), itemPath))
+			icons = append(icons, f.newFileItem(canvas.NewImageFromResource(theme.FolderIcon()),
+				itemPath, file.Name(), true))
 		} else {
-			icons = append(icons, f.newFileItem(theme.FileIcon(), itemPath))
+			icons = append(icons, f.newFileItem(NewFileIcon(itemPath), itemPath,
```
andydotxyz, line 180: Why does the constructor take an icon derived from the path, a name derived form the path, and the path itself?
It doesn't seem right that the filedialog needs helpers to setup data to pass into a constructor

dialog/fileicon.go
andydotxyz: I'm going to up this from "nit" - the code may not all be ordered correctly, but new files probably should be or we have an increasing tidyup later.

dialog/fileicon.go
```diff
@@ -0,0 +1,127 @@
...
+			scanner := bufio.NewScanner(file)
+			if scanner.Scan() && !utf8.Valid(scanner.Bytes()) {
+				mimeTypeFull = "application/octet-stream"
+			}
+		}
+	}
+
+	mimeTypeSplit := strings.Split(mimeTypeFull, "/")
+	mimeType = mimeTypeSplit[0]
+	mimeSubTypeSplit := strings.Split(mimeTypeSplit[1], ";")
```
andydotxyz, line 53: This will panic if we have not got a "/" in the string. perhaps we should be more safe?

theme/icons/file-audio.svg
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
andydotxyz, line 1: This image is smaller than the others, not sure why or how. Look at them blown up in the fyne_demo Graphics tab

andydotxyz:
A few issues inline. I think that the constructors require too much state passed in, causing utility functions to be created in places that wouldn't need them otherwise.

I think that code duplication could be reduced by usage of the URI type, but if you think that is beyond the scope of this PR that's OK, it can be tidied up later.

storage/uri.go
```diff
@@ -22,7 +26,23 @@ func (u *uri) Extension() string {
 }
 
 func (u *uri) MimeType() string {
-	return "x/unknown" // TODO - added in #405
+	ext := filepath.Ext(u.raw[1:])
```
andydotxyz, line 29: Why is the first character being trimmed from the URI?

andydotxyz: More to the point, why not use u.Extension() ;)

dialog/fileicon.go
```diff
@@ -34,24 +30,13 @@ func (i *fileIcon) CreateRenderer() fyne.WidgetRenderer {
...
-	}
+	uri := storage.NewURI(path)
+	ext = uri.Extension()
+	mimeTypeFull := uri.MimeType()
 
 	mimeTypeSplit := strings.Split(mimeTypeFull, "/")
 	mimeType = mimeTypeSplit[0]
-	mimeSubTypeSplit := strings.Split(mimeTypeSplit[1], ";")
-	mimeSubType = mimeSubTypeSplit[0]
+	mimeSubType = mimeTypeSplit[1]
```
andydotxyz: I don't know if we discussed before, is it OK to panic if the mime does not contain a "/"?

storage/uri.go
okratitan: the first character isn't being trimmed from the URI.  The first character, which is "." is being trimmed from the extension.  We can use uri.Extension.

dialog/fileicon.go
okratitan: It would definitely make sense to add some safety here.

storage/uri.go
andydotxyz: Are you sure? It looks very much like the raw uri string is being sliced then passed to the extension function of filepath

storage/uri.go
```diff
@@ -22,7 +26,23 @@ func (u *uri) Extension() string {
...
+			scanner := bufio.NewScanner(file)
+			if scanner.Scan() && !utf8.Valid(scanner.Bytes()) {
+				mimeTypeFull = "application/octet-stream"
+			}
+		}
+	}
+	mimeSubTypeSplit := strings.Split(mimeTypeFull, ";")
+	mimeTypeFull = mimeSubTypeSplit[0]
+
+	return mimeTypeFull
```
stuartmscott, line 45: Can this be one line?
```go
return strings.Split(mimeTypeFull, ";")[0]
```

storage/uri.go
okratitan: on second thought I see this didn't translate well in the move and now is indeed just removing the first character of the URI. Will fix.

dialog/fileitem.go
```diff
@@ -35,47 +36,40 @@ func (i *fileDialogItem) TappedSecondary(_ *fyne.PointEvent) {
...
-	return i.icon == theme.FolderIcon() || i.icon == theme.FolderOpenIcon()
+	return &fileItemRenderer{item: i,
+		img: i.icon, text: text, objects: []fyne.CanvasObject{i.icon, text}}
 }
 
-func fileParts(path string) (name, ext string) {
+func fileName(path string) (name string) {
 	name = filepath.Base(path)
-	ext = filepath.Ext(name[1:])
+	ext := filepath.Ext(path[1:])
```
andydotxyz, line 48: This looks like an unnecessary slice off the front?

PR #942, opened 2020-05-05
Do not create a new test app inside the assert

andydotxyz:
Cleaner version of the assert notification helper.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -55,8 +55,12 @@ func TestFyneApp_OpenURL(t *testing.T) {
 
 func TestFyneApp_SendNotification(t *testing.T) {
 	n := fyne.NewNotification("Test Title", "Some content")
+	testApp := test.NewApp()
+	oldApp := fyne.CurrentApp() // we are testing the app package so don't use the test app
+	fyne.SetCurrentApp(testApp)
 
-	test.AssertNotificationSent(t, n, func(a fyne.App) {
-		a.SendNotification(n)
+	test.AssertNotificationSent(t, n, func() {
+		fyne.CurrentApp().SendNotification(n)
 	})
+	fyne.SetCurrentApp(oldApp)
 }
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -7,7 +7,6 @@ import (
 	"testing"
 
 	"fyne.io/fyne"
-	"fyne.io/fyne/test"
 	_ "fyne.io/fyne/test"
 	"github.com/stretchr/testify/assert"
 )
@@ -51,16 +50,4 @@ func TestFyneApp_OpenURL(t *testing.T) {
 	for opened == "" {
 	}
 	assert.Equal(t, urlStr, opened)
-}
-
-func TestFyneApp_SendNotification(t *testing.T) {
-	n := fyne.NewNotification("Test Title", "Some content")
-	testApp := test.NewApp()
-	oldApp := fyne.CurrentApp() // we are testing the app package so don't use the test app
-	fyne.SetCurrentApp(testApp)
-	defer fyne.SetCurrentApp(oldApp)
-
-	test.AssertNotificationSent(t, n, func() {
-		fyne.CurrentApp().SendNotification(n)
-	})
 }
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -58,9 +58,9 @@ func TestFyneApp_SendNotification(t *testing.T) {
 	testApp := test.NewApp()
 	oldApp := fyne.CurrentApp() // we are testing the app package so don't use the test app
 	fyne.SetCurrentApp(testApp)
+	defer fyne.SetCurrentApp(oldApp)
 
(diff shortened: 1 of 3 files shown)


app/app_test.go
```diff
@@ -55,8 +55,12 @@ func TestFyneApp_OpenURL(t *testing.T) {
...
+	testApp := test.NewApp()
+	oldApp := fyne.CurrentApp() // we are testing the app package so don't use the test app
+	fyne.SetCurrentApp(testApp)
 
-	test.AssertNotificationSent(t, n, func(a fyne.App) {
-		a.SendNotification(n)
+	test.AssertNotificationSent(t, n, func() {
+		fyne.CurrentApp().SendNotification(n)
 	})
+	fyne.SetCurrentApp(oldApp)
```
stuartmscott, line 65: nit: use defer to keep this near the other

app/app_test.go
```diff
@@ -55,8 +55,12 @@ func TestFyneApp_OpenURL(t *testing.T) {
 
 func TestFyneApp_SendNotification(t *testing.T) {
 	n := fyne.NewNotification("Test Title", "Some content")
+	testApp := test.NewApp()
+	oldApp := fyne.CurrentApp() // we are testing the app package so don't use the test app
+	fyne.SetCurrentApp(testApp)
+	defer fyne.SetCurrentApp(oldApp)
```
toaster, line 61: Looks strange. You
* build a new test app which is implicitly set as current app -> `testApp`
* get the current app which is at this point the same as `testApp` -> `oldApp == testApp`
* set the `testApp` as current app which does nothing because it already is
* defer setting `oldApp` as current app which doesn’t change anything because `oldApp == testApp`

andydotxyz: Or more simply "I think line 58 amd 59 are the wrong way round".
Agreed, apologies.

test/notification_test.go
```diff
@@ -0,0 +1,28 @@
+package test
```
toaster, line 1:
```suggestion
package test_test
```

test/notification_test.go
```diff
@@ -0,0 +1,28 @@
...
+	})
+	assert.Equal(t, myApp, fyne.CurrentApp())
+}
+
+func TestAssertNotificationSent_NotSent(t *testing.T) {
+	myApp := fyne.CurrentApp()
+
+	AssertNotificationSent(t, nil, func() {
+		// don't send anything
+	})
```
toaster, line 26:
```suggestion
	n := fyne.NewNotification("Test Title", "Some content")
	myApp := fyne.CurrentApp()
	tt := &testing.T{}
	
	AssertNotificationSent(tt, n, func() {
		// don't send anything
	})
	assert.True(t, tt.Failed(), "test failed")
```

andydotxyz:
> And I think the test should fail if `n` is `nil` but a notification was sent.

It will:

```
Error:      	Expected nil, but got: &fyne.Notification{Title:"Test Title", Content:"Some content"}
```

test/notification.go
```diff
@@ -4,21 +4,25 @@ import (
...
-	a := NewApp().(*testApp)
-	fyne.SetCurrentApp(a)
-
-	f(a)
+	f()
 	if n == nil {
+		assert.Nil(t, a.lastNotification)
+		return
+	} else if a.lastNotification == nil {
+		t.Error("No notification sent")
```
stuartmscott, line 25: Consider moving this message into the NonNil on 29 and drop this "else if"

andydotxyz: We have to return early. But the NotNil assertion is not needed any more - would removing that be sufficient?

test/notification_test.go
```diff
@@ -12,17 +13,23 @@ func TestAssertNotificationSent(t *testing.T) {
...
 func TestAssertNotificationSent_NotSent(t *testing.T) {
-	myApp := fyne.CurrentApp()
+	tt := &testing.T{}
 
-	AssertNotificationSent(t, nil, func() {
+	test.AssertNotificationSent(tt, &fyne.Notification{}, func() {
 		// don't send anything
 	})
-	assert.Equal(t, myApp, fyne.CurrentApp())
+	assert.True(t, tt.Failed(), "A notification was sent incorrectly")
```
toaster, line 34: The explanation confuses me. We test the notification assert helper. If this `assert.True` fails, we assume that the helper is broken. This means that no notification was sent (as expected) but that the helper did something wrong.
Maybe the explanation should be like “asserting notification should fail if no notification was sent”.

toaster:
> > And I think the test should fail if `n` is `nil` but a notification was sent.
> 
> It will:
> 
> ```
> Error:      	Expected nil, but got: &fyne.Notification{Title:"Test Title", Content:"Some content"}
> ```

Ah, yes, my fault.

andydotxyz:
Thanks all, suggestions all merged in

Issue #814, opened 2020-04-05
Key events not blocked by overlays


toaster:
### Describe the bug:

When opening a pop-up (like the menu) the overlay should catch all events. This works well for mouse events but keyboard events are still forwarded to the window's content.

Maybe this isn't fixable by generic event blocking in overlays. When a pop-up belongs to a form widget (`Select` or `EntrySelect`), the `<TAB>`-key should probably close the pop-up.

### To Reproduce:

Steps to reproduce the behaviour:
1. Open demo app without native menus
2. Open the Widget->Input pane
3. Click on a menu
4. Press `<TAB>` until the entry widget has focus
5. Enter some text
6. See error

### Screenshots:

![keyboard_events_with_open_menu](https://user-images.githubusercontent.com/73959/78469295-1d276000-7720-11ea-9142-2cd2527cd6c1.gif)

toaster:
This is partly fixed with #1318 which fixes the focus handling for overlays.
However, the menu part is still open.

andydotxyz:
This has a pretty solid solution on `develop` now, thanks @toaster.
Closing

PR #945, opened 2020-05-06
Bugfix: 922

### LINKED ISSUE #922: Menu clears focus so cut/copy/paste doesn't work
Open the fyne_demo app, (with non-native menus macOS users ;) ).
Then go to an input field and get focus. Then tap menu bar and see focus is lost.

toaster:
### Description:

This PR changes the tap handling on non-native menus to not change the canvas' focus.
It allows to build menu items which interact with the focused elements (e.g. “Cut”, “Copy” and ”Paste”).

IMO this is a workaround. The visual focus should be handled layer specific (content, menu, overlay(s)). The input focus should be one one of these layers. #814 is related to this topic.

Fixes #922

### Checklist:

- ~~[ ] Tests included.~~
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -472,15 +472,8 @@ func (w *window) refresh(viewport *glfw.Window) {
 	w.canvas.setDirty(true)
 }
 
-func (w *window) findObjectAtPositionMatching(canvas *glCanvas, mouse fyne.Position,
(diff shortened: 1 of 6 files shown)


andydotxyz:
Seems like a decent workaround, but I think that introducing a new layer counter like this should be tested.
I assume they were not included for a reason, but you didn't say what that might have been.

PR #947, opened 2020-05-06
Switch to NRGBA in all areas except the openGL painter 

### LINKED ISSUE #909: Colours are not correctly calculated when alpha is low
It looks like there is a mismatch with pre-mul vs regular alpha handling.

I would propose that we should hide any GL requirement to be pre-multiplied so the API works more like the HTML RGB handling which I think is more common.

Sadly making this change (one way or the other) will likely break some people's app colouring if they have picked bespoke colour values where Alpha < R or G or B.

andydotxyz:
GL requires it pre-multiplied, everywhere else expects RGBA like HTML where any channel can be changed independently.

This changed some of the test images, I guess that the alpha channel differences affected their represntation.
Also update the image comparison to check the bytes as the RGBA vs NRGBA colour models was breaking things
but PNG is always NRGBA, so something strange with the decoder there.

This may result in very subtle changes in how shadows are applied but will not impact the API in general. If people pass in a color.RGBA or color.NRGBA it will still be respected.

Fixes #909

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


test/util.go
```diff
@@ -3,19 +3,19 @@ package test
...
 	"fmt"
 	"image"
-	"image/draw"
 	"image/png"
 	"os"
 	"path/filepath"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
```
stuartmscott, line 12: nit: move this back next to require

andydotxyz: Fixed. No idea why GoLand occasionally does stupid things like that. 
I swear I fixed it at least twice!

PR #949, opened 2020-05-06
WIP: Add test for dialog background

stuartmscott:


dialog/base_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"testing"
+)
+
+func TestDialog_Background_Group(t *testing.T) {
+	a := test.NewApp()
+	defer test.NewApp()
+	test.ApplyTheme(t, theme.LightTheme())
+
+	w := a.NewWindow("")
```
andydotxyz, line 19: Apologies I realise now, the software painter is not wired up by default due to a circular dependency we are working to resolve. You'll need to create a window this way in the meantime:

```
	test.NewApp().Settings().SetTheme(theme.LightTheme())
	defer test.NewApp()
	w := test.NewWindowWithPainter(canvas.NewRectangle(theme.BackgroundColor()), software.NewPainter())
```

andydotxyz:
With the suggested change this is what I get:

dialog_background_group

andydotxyz:
This should have worked automatically, it's a known issue that it requires a wiring up in your test code :(

stuartmscott:
Thanks, I tried this and it rendered!

dialog/base_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"fyne.io/fyne"
+	"fyne.io/fyne/canvas"
+	"fyne.io/fyne/dialog"
+	"fyne.io/fyne/internal/painter/software"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/test"
+	"fyne.io/fyne/theme"
+	"fyne.io/fyne/widget"
+
+	"testing"
```
andydotxyz, line 13: import ordering

dialog/base_test.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+func TestDialog_Background_Group(t *testing.T) {
```
andydotxyz, line 16: I know this came from background being suspicious, but in this instance I don't know if the test actually cares about background so the name is a little misleading.

Should we leave the content as is? just an open question - it doesn't matter much one way or another.

Issue #506, opened 2019-10-29
Massive data race detect!!


prospero78:
**Describe the bug**
Im compile my soft with option -race. After start im loock in console 23 event data race

**Device (please complete the following information):**
 - OS: Ubuntu Linux
 - Version 18.04 LTS
 - Go version 1.13.1
 - Fyne version: last stable, public download from github.

andydotxyz:
You are quite right - there are lots of areas where this needs to be improved.
The biggest of which are the widget properties and the manner in which canvasobjects are refreshed.
Both of these should be addressed in v1.2 which means that the races detected should be down to a manageable number of fixes required.

andydotxyz:
The codebase is now passing with `-race` for the most part.
We know that the redenderers can still cause some races but that requires a behaviour change to how they are maintained - moved that item to 2.0.
I think we can close this ticket though. Feel free to test on `develop` if you would like.

stuartmscott:
Tested on develop @ 3135c8894de0342b7f67bbf615c8ed1399287168 with `go test -race -count 10 ./...`

```
==================
WARNING: DATA RACE
Write at 0x00c000001800 by main goroutine:
  fyne.io/fyne/internal/driver/glfw.(*window).create.func1()
      /Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/internal/driver/glfw/window.go:1137 +0x1c3
  fyne.io/fyne/internal/driver/glfw.(*gLDriver).runGL()
      /Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/internal/driver/glfw/loop.go:96 +0x64d
  fyne.io/fyne/internal/driver/glfw.(*gLDriver).Run()
      /Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/internal/driver/glfw/driver.go:70 +0x44
  fyne.io/fyne/internal/driver/glfw.TestMain()
      /Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/internal/driver/glfw/window_test.go:47 +0xab
  main.main()
      _testmain.go:136 +0x223

Previous read at 0x00c000001800 by goroutine 9:
  fyne.io/fyne/internal/driver/glfw.(*gLDriver).startDrawThread.func1()
      /Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/internal/driver/glfw/loop.go:171 +0x234

Goroutine 9 (running) created at:
[...]

PR #951, opened 2020-05-07
Improve thread safety of all test utilities and the singleton CurretApp

andydotxyz:
### Description:

This goes a long way to improving the ability to turn on race checking again.

- [x] Remove race conditions from test utilities

[[[ proposed moving these two to another PR - 210 race conditions in `widget` package
- [ ] Add utility to BaseWidget to provide safe property access
- [ ] Fix race conditions in our codebase :) (this one may not be 100% in this PR)

]]]

This is mostly attempting to address #506 

### Checklist:

- [x] Tests included. <- well, current tests with "-race" added
- [x] Tests all pass.

Tracking packages in which "-race" passes

- [x] .
- [x] app
- [x] canvas
- [x] cmd
- [x] dialog
- [x] driver
- [x] internal
- [x] internal/driver
- [x] internal/painter
- [x] internal/widget
- [x] layout
- [x] test
- [x] theme
- [x] tools
- [ ] widget - for another PR?

### DIFF
--- a/app.go
+++ b/app.go
@@ -1,6 +1,9 @@
 package fyne
 
-import "net/url"
+import (
+	"net/url"
+	"sync"
+)
 
 // An App is the definition of a graphical application.
 // Apps can have multiple windows, it will exit when the first window to be
@@ -54,13 +57,20 @@ type App interface {
 }
 
 var app App
+var appLock sync.RWMutex
 
 // SetCurrentApp is an internal function to set the app instance currently running.
 func SetCurrentApp(current App) {
+	appLock.Lock()
+	defer appLock.Unlock()
+
 	app = current
 }
 
 // CurrentApp returns the current application, for which there is only 1 per process.
 func CurrentApp() App {
+	appLock.RLock()
+	defer appLock.RUnlock()
+
 	return app
 }
--- a/internal/painter/gl/gl_core.go
+++ b/internal/painter/gl/gl_core.go
@@ -187,7 +187,7 @@ func (p *glPainter) glCreateBuffer(points []float32) Buffer {
 
 	texCoordAttrib := uint32(gl.GetAttribLocation(uint32(p.program), gl.Str("vertTexCoord\x00")))
 	gl.EnableVertexAttribArray(texCoordAttrib)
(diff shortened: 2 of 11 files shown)


andydotxyz:
This PR now makes it possible to `go test -race` in every package except `widget`.
It could be merged now, or wait for my widget code.

The work in `widget` is more contentious as field access vs `Refresh()` needs a new helper imho.
I have the work locally but wasn't sure if I should push to this PR...

app.go
```diff
@@ -54,13 +57,20 @@ type App interface {
 }
 
 var app App
+var appLock sync.RWMutex
```
stuartmscott, line 60: Pity we don't have more benchmarks so we can see the performance impact of all this locking

internal/painter/gl/gl_core.go
```diff
@@ -187,7 +187,7 @@ func (p *glPainter) glCreateBuffer(points []float32) Buffer {
 
 	texCoordAttrib := uint32(gl.GetAttribLocation(uint32(p.program), gl.Str("vertTexCoord\x00")))
 	gl.EnableVertexAttribArray(texCoordAttrib)
-	gl.VertexAttribPointer(texCoordAttrib, 2, gl.FLOAT, false, 5*4, gl.PtrOffset(3*4))
+	gl.VertexAttribPointer(texCoordAttrib, 2, gl.FLOAT, false, 5*4, gl.PtrOffset(12))
```
stuartmscott, line 190: not sure this is part of this PR
why change 3 * 4 to 12 but not 5 * 4 to 20?
what are these magic numbers?

andydotxyz: 3*4 was creating a compile error due to it thinking this was bad pointer arithmatic, so I only changed what was needed.
The offset is, iirc, the 3 vertexes in to a 4 channel colour. Honestly I can't remember - it requires special concentration to get GL code to make sense ;)

app.go
andydotxyz: Agreed.
I was about to suggest that is write-once somehow - but other issues want to run multiple apps one after the other in the same runtime so I don't think that's an option. Unless we say "no" to running app2 after app1

stuartmscott: Hmm I think that would break our tests as we need multiple apps to run after each other

andydotxyz: oh yes :)

PR #955, opened 2020-05-07
Full race condition fixes

### LINKED ISSUE #506: Massive data race detect!!

andydotxyz:
### Description:

This will conclude by turning on race checking in our tests again :).

- [x] Add utility to BaseWidget to provide safe property access
- [x] Fix race conditions in our codebase :) (this one will be 100% in this PR 😄)
- [ ] Re-enable race checking in the traivs config

Fixes #506 

### Checklist:

- [x] Tests included. <- well, current tests with "-race" added
- [x] Tests all pass.

This PR will fix the race conditions in `widget` package.


andydotxyz:
OK, this PR now shows the proposed approach to widget safety.
It took the count to below 100 races (from around 200) in the widget package by just fixing BaseWidget.
Further commits will resolve the widget specific issues.

internal/widget/base.go
```diff
@@ -10,34 +12,72 @@ type base struct {
...
+func (b *base) readFields(f func()) {
+	b.propertyLock.RLock()
+	defer b.propertyLock.RUnlock()
+
+	f()
+}
+
+// SetFieldsAndRefresh helps to make changes to a widget that should be followed by a refresh.
+// This method is a guaranteed thread-safe way of directly manipulating widget fields.
+func (b *base) setFieldsAndRefresh(w fyne.Widget, f func()) {
```
stuartmscott, line 45: nit: the name ordering doesn't match parameter ordering.
```suggestion
func (b *base) setFieldsAndRefresh(f func(), w fyne.Widget) {
```

widget/widget.go
```diff
@@ -13,51 +15,95 @@ type BaseWidget struct {
...
+}
+
+// SetFieldsAndRefresh helps to make changes to a widget that should be followed by a refresh.
+// This method is a guaranteed thread-safe way of directly manipulating widget fields.
+func (w *BaseWidget) SetFieldsAndRefresh(f func()) {
+	w.propertyLock.Lock()
+	f()
+	w.propertyLock.Unlock()
+
+	w.Refresh()
```
stuartmscott, line 49: What if the basewidget has been extended?
```suggestion
	if w.impl == nil {
		w.Refresh()
	} else {
		w.impl.Refresh()
	}
```

internal/widget/base.go
```diff
@@ -10,34 +12,72 @@ type base struct {
...
 
 func (b *base) hide(w fyne.Widget) {
-	if b.hidden {
+	if !b.Visible() {
 		return
 	}
 
+	b.propertyLock.Lock()
 	b.hidden = true
+	b.propertyLock.Unlock()
```
stuartmscott, line 80: Use setFieldsAndRefresh(f func(){...}, nil) for consistency

internal/widget/base.go
```diff
@@ -60,11 +100,17 @@ func (b *base) refresh(w fyne.Widget) {
...
+	b.readFields(func() {
+		baseSize = b.size
+	})
+	if baseSize == size {
 		return
 	}
 
+	b.propertyLock.Lock()
 	b.size = size
+	b.propertyLock.Unlock()
```
stuartmscott, line 113: Use setFieldsAndRefresh(f func(){...}, nil) for consistency

widget/widget.go
```diff
@@ -13,51 +15,95 @@ type BaseWidget struct {
...
 
-	impl fyne.Widget
+	impl         fyne.Widget
+	propertyLock sync.RWMutex
 }
 
 // ExtendBaseWidget is used by an extending widget to make use of BaseWidget functionality.
 func (w *BaseWidget) ExtendBaseWidget(wid fyne.Widget) {
+	w.propertyLock.Lock()
+	defer w.propertyLock.Unlock()
```
stuartmscott, line 25: Use SetFieldsAndRefresh(f func(){...}) for consistency

widget/widget.go
```diff
@@ -13,51 +15,95 @@ type BaseWidget struct {
...
+	f()
+	w.propertyLock.Unlock()
+
+	w.Refresh()
+}
+
 // Size gets the current size of this widget.
 func (w *BaseWidget) Size() fyne.Size {
+	w.propertyLock.RLock()
+	defer w.propertyLock.RUnlock()
```
stuartmscott, line 55: Use ReadFields(f func(){...}) for consistency

widget/widget.go
```diff
@@ -13,51 +15,95 @@ type BaseWidget struct {
...
 		return
 	}
-	cache.Renderer(w.impl).Layout(size)
+	cache.Renderer(impl).Layout(size)
 }
 
 // Position gets the current position of this widget, relative to its parent.
 func (w *BaseWidget) Position() fyne.Position {
+	w.propertyLock.RLock()
+	defer w.propertyLock.RUnlock()
```
stuartmscott, line 86: Use the helper methods to avoid lock/unlock

internal/widget/base.go
andydotxyz: This is diferent. We don't actually refresh widget on hide - just the canvas refresh expressed below.

internal/widget/base.go
andydotxyz: But we are laying out not refreshing as a result of resize

widget/widget.go
andydotxyz: But we do not need to refresh

widget/widget.go
andydotxyz: I don't think that's how Refresh lookup works - the extended object might provide the Refresh call, and if not then basewidget will do the lookup.

widget/widget.go
andydotxyz: Yes, the return here meant that it would double the number of variables.
I'm thinking about a ReadAndReturn(func()) interface{} helper for these situations

widget/widget.go
andydotxyz: As above

widget/widget.go
stuartmscott: If impl does not define Refresh, then BaseWidget.Refresh will call Refresh on the impl's Renderer.
If impl does define Refresh, this will still call Refresh on the impl's Renderer

internal/widget/base.go
stuartmscott: Hence the nil - you already have a nil check in setFieldsAndRefresh

internal/widget/base.go
stuartmscott: Again the nil means you're not refreshing. You could also define setFields(func()) if you don't like passing nil

widget/widget.go
andydotxyz: Ah yes I see your point, thanks :)

widget/widget.go
stuartmscott: Not sure about that - the type cast may be uglier than extra variables

internal/widget/base.go
andydotxyz: Oh i missed that. The nil check was really safety rather than a feature I thought should be added. using a "nil" special value is generally an anti-pattern so I would prefer an additional method over telling people to pass nil

widget/widget.go
andydotxyz: Possible. I was going to play with it and see

widget/widget.go
stuartmscott: Define SetFields(func()) which SetFieldsAndRefresh(func(), fyne.Widget) uses

andydotxyz:
I have reverted the .travis update because the widget race conditions are not completely fixed.
They all pass on my computer but under other setups some are present.
The complete solution is around more safety in renderers (as you can see in #979 )
We can't merge the full solution in 1.3 without risking consumer apps locking up.
I think this PR takes it as far as possible. But we could also consider reverting the widget changes if we are concerned.

widget/entry.go
```diff
@@ -95,17 +93,20 @@ func (e *Entry) CreateRenderer() fyne.WidgetRenderer {
 	cursor := canvas.NewRectangle(theme.FocusColor())
 	cursor.Hide()
 
-	objects := []fyne.CanvasObject{line, e.placeholderProvider(), e.textProvider(), cursor}
+	var objects []fyne.CanvasObject
+	e.SetFields(func() {
```
stuartmscott, line 97: Why do you need the Write lock to read line, e.placeholderProvider, e.textProvider, and cursor?

widget/entry.go
```diff
@@ -170,35 +170,41 @@ func (e *Entry) Enable() { // TODO remove this override after ReadOnly is remove
 
 // ExtendBaseWidget is used by an extending widget to make use of BaseWidget functionality.
 func (e *Entry) ExtendBaseWidget(wid fyne.Widget) {
```
stuartmscott, line 172: Note: this function has slightly different behaviour now, before it would only call registerShortcut() if e.BaseWidget.impl was nil. Now it calls registerShortcut() every time. Tho I don't think that is an issue based on how ShortcutHandler is implemented - it will just replace the old shortcuts with the identical new ones.

widget/entry.go
```diff
@@ -296,25 +312,26 @@ func (e *Entry) SetReadOnly(ro bool) {
...
-		e.Refresh()
-	} else {
-		provider := e.textProvider()
+		e.SetFieldsAndRefresh(func() {
+			e.CursorColumn = 0
+			e.CursorRow = 0
+		})
+		return
+	}
+	provider := e.textProvider()
```
stuartmscott, line 325: Would it be safer to move this inside SetFields so it is called while holding the lock?

widget/entry.go
```diff
@@ -378,124 +395,131 @@ func (e *Entry) TypedKey(key *fyne.KeyEvent) {
...
+		e.SetFields(func() {
+			provider.insertAt(e.cursorTextPos(), []rune("\n"))
+			e.CursorColumn = 0
+			e.CursorRow++
+		})
 	case fyne.KeyUp:
-		if !e.MultiLine {
+		if e.GetField(func() interface{} {
+			return !e.MultiLine
+		}).(bool) {
```
stuartmscott, line 431: This is repeated a few times, perhaps add "func (e *Entry) IsMultiline() bool" ?

widget/entry.go
```diff
@@ -378,124 +395,131 @@ func (e *Entry) TypedKey(key *fyne.KeyEvent) {
 
 	switch key.Name {
 	case fyne.KeyBackspace:
-		e.RLock()
-		isEmpty := provider.len() == 0 || (e.CursorColumn == 0 && e.CursorRow == 0)
-		e.RUnlock()
+		isEmpty := e.GetField(func() interface{} {
+			return provider.len() == 0 || (e.CursorColumn == 0 && e.CursorRow == 0)
```
stuartmscott, line 399: Apparently I can't comment on lines that haven't changed, but on line 387, is it safe to get the textProvider without read lock?

widget/entry.go
```diff
@@ -818,15 +837,9 @@ func (e *Entry) textProvider() *textProvider {
...
-	defer e.Unlock()
-	if e.text != nil {
-		return e.text
-	}
-
 	text := newTextProvider(e.Text, e)
-	text.ExtendBaseWidget(&text)
-	e.text = &text
+	text.ExtendBaseWidget(text)
+	e.text = text
```
stuartmscott, line 842: Is this safe? Shouldn't this and placeholderProvider() be set with SetFields?

widget/entry.go
```diff
@@ -849,47 +862,51 @@ func (e *Entry) textWrap() fyne.TextWrap {
...
-	if changed && e.OnChanged != nil {
-		e.OnChanged(text)
-	}
+	var callback func(string)
+	e.SetFieldsAndRefresh(func() {
+		changed := e.Text != text
+		e.Text = text
 
-	e.Refresh()
+		if changed && e.OnChanged != nil {
```
stuartmscott, line 902: Nil check not needed here, it is already checked at 907
```suggestion
		if changed {
```

widget/widget.go
```diff
@@ -13,51 +15,108 @@ type BaseWidget struct {
...
+func (w *BaseWidget) ReadFields(f func()) {
+	w.propertyLock.RLock()
+	defer w.propertyLock.RUnlock()
+
+	f()
+}
 
-	w.impl = wid
+// GetField provides a guaranteed thread safe way to access fields and return a value.
+func (w *BaseWidget) GetField(f func() interface{}) interface{} {
```
stuartmscott, line 42: Nice

andydotxyz:
I will come back to this in detail tomorrow. However the answer to many questions is "textProvider is special, we cannot lock in some places we want to - so instead we have to ensure the calling method is locking on it's behalf". I'm sure I missed some.

widget/entry.go
andydotxyz: You don't this was just a shorthand to encapsulate a read, then possible write, then more reads. I can break it out into separate areas but the code will get rather ugly.

widget/entry.go
andydotxyz: That is a good point. This takes a bit of thinking but I think overall I should change how baseimpl is locked. In this instance the method may be called a lot so the write lock is probably a bad idea if not truly required.

widget/entry.go
andydotxyz: I'll just grab it at the top, seems seasier

widget/entry.go
andydotxyz: This is called so frequently and also initiates other code if a new one is created that making it lock itself was too deadlocky - so I opted for it must be locked outside this method, which as you noted above I have missed a few. They are fixed now so this doesn't have to be.

andydotxyz:
Thanks for that, all addressed and tests still passing (phew ;) )

widget/entry.go
stuartmscott: `must be locked outside this method`

Read or Write lock?

widget/entry.go
```diff
@@ -960,13 +977,22 @@ func (r *entryRenderer) Objects() []fyne.CanvasObject {
 
 // Refresh satisfies the fyne.WidgetRenderer interface.
 func (r *entryRenderer) Refresh() {
-	if r.entry.Text != string(r.entry.textProvider().buffer) {
-		r.entry.textProvider().SetText(r.entry.Text)
+	var provider, placeholder *textProvider
+	content := r.entry.GetField(func() interface{} {
+		provider = r.entry.textProvider()
```
stuartmscott, line 982: This could create the textProvider without holding a write lock

widget/text.go
```diff
@@ -90,7 +103,7 @@ func (t *textProvider) updateRowBounds() {
 	textWrap := t.presenter.textWrap()
 	textStyle := t.presenter.textStyle()
 	textSize := theme.TextSize()
-	maxWidth := t.Size().Width - 2*theme.Padding()
+	maxWidth := t.size.Width - 2*theme.Padding()
```
stuartmscott, line 106: The func doc should explain what lock callers must hold

widget/entry.go
andydotxyz: are you sure? - textProvider() can write the field if there is no provider...

widget/text.go
andydotxyz: In this instance I thought it was clear - the method was an exported method so is safe (locked). As I am obtaining the lock I can access the field but not call the method that would deadlock.

widget/entry.go
stuartmscott: But it would write the field while holding the read lock, but not the write lock

widget/text.go
stuartmscott: Yup, I mean updateRowBounds' doc should explain that callers hold lock

widget/entry.go
andydotxyz: It is my understanding that if you have RWMutex.Lock() then nobody can .RLock() I don't think that Lock() is write-only

widget/text.go
andydotxyz: Thanks. This is an ongoing discussion about documentation that I don't think has been resolved yet.

toaster:
Sorry for being so late to the party.
IMO methods like `GetField`, `ReadFields`, `SetFields` and `SetFieldsAndRefresh` are hard to understand. Especially they don’t relieve the developer of the burden to think about thread safety.
So, why do it this complicated way?
Why not simply provide the lock to the developer?
Why is
```go
func (t *textProvider) setText(text string) {
	t.SetFields(func() {
		t.buffer = []rune(text)
		t.updateRowBounds()
	})

	t.refreshTextRenderer()
}
```
better than
```go
func (t *textProvider) SetText(text string) {
	t.Lock()
	defer t.Unlock()
	t.buffer = []rune(text)
	t.updateRowBounds()

	t.refreshTextRenderer()
}
```

Why is
```go
func (p *infProgressRenderer) isRunning() bool {
	return p.progress.GetField(func() interface{} {
		return p.running
	}).(bool)
}
```
better than
```go
func (p *infProgressRenderer) isRunning() bool {
	t.RLock()
	defer t.RUnlock()
	return p.running
}
```

IMO the examples with the explicit lock/unlock are better to read and better to understand.

andydotxyz:
The general concept was that for any lock it should be set in as few places as possible.
If it’s exposed to the user then a lock they forget to unlock will lock up the app, taking care of it with utility methods should help avoid that.

And it makes lock scope explicit. Defer works well if you want the lock for the whole method but in many places there are locks that must be cleared before a callback or other final items in the method...

toaster:
Sorry, but I think the method solution is not elegant.
The methods don’t make a lock context explicit because they hide the fact that locking happens.
If I read the code which uses such a method, I am puzzled. I can’t see a lock but some kind of magic function and sometimes ugly casts.
If one wants to clarify lock context in a method where other stuff happens without the lock being in place, he should extract the locked part into a separate function. There `defer` can be used again. IMO this is much simpler and cleaner.

And I also don’t think that “forgetting to unlock a lock” is a problem that has to be solved by Fyne.

andydotxyz:
I think there is one pattern to take into consideration (that led me down this path):
Update some fields and call refresh. The lock must be unlocked before refresh is called which makes for defer being unusable. So either:

```
wid.Lock()
wid.Thing = x
wid.Other = y
wid.Unlock()
wid.Refresh()
```
vs
```
wid.UpdateAndRefresh(func() {
    wid.Thing.x
    wid.Other = y
})
```

toaster:
I don’t think the method is better here :).
If one doesn’t like the `Unlock` before the `Refresh` in the same method he can extract the update part:
```go
func (wid *MyWidget) updateThings(x, y …) {
  wid.Lock()
  defer wid.Unlock()
  wid.Thing = x
  wid.Other = y
}

…

  wid.updateThings(x, y)
  wid.Refresh()
```

andydotxyz:
So in this instance it's better to expose the details and have developers implement the methods you think are not a good pattern?

andydotxyz:
Updated in line with conversation last week. I kept an unexported setFieldsAndRefresh as it’s helpful.

widget/entry.go
```diff
@@ -1136,10 +1171,9 @@ func (r *passwordRevealer) Cursor() desktop.Cursor {
 
 // Tapped satisfies the fyne.Tappable interface.
 func (r *passwordRevealer) Tapped(*fyne.PointEvent) {
-	r.entry.Lock()
-	r.entry.Password = !r.entry.Password
-	r.entry.Unlock()
-	r.Refresh()
+	r.entry.setFieldsAndRefresh(func() {
```
stuartmscott, line 1174: Should this refresh here, doesn't 1177 call Focus which calls Refresh?

andydotxyz: Focus does not necessarily call refresh.
It might if the state has changed, but it's not part of the contract.
Here we are probably updating the display when we are already focused, so it's a driver detail whether or not the entry will refresh.

andydotxyz:
Label conflict dismissed review sorry @stuartmscott

andydotxyz:
Agreed, sorry. And now GitHub pull is not responding. Will come back to this shortly

andydotxyz:
I used the GitHub merge tool and it somehow kept the old and the new line. This is not the first time it's happened either. I wonder if I'm not using that function correctly!

andydotxyz:
Sigh, sorry, had to merge again, and it invalidated again sorry @stuartmscott

PR #957, opened 2020-05-08
Refactoring/testapp with software painter

toaster:
### Description:

This PR makes software.Painter default in tests and removes the `test.NewWindowWithPainter` function.

### Checklist:

- ~~[ ] Tests included.~~
- [x] Tests all pass.


toaster:
Please restart test:
```
--- FAIL: TestWatchFile (0.17s)
FAIL
FAIL	fyne.io/fyne/app	0.349s
```

andydotxyz:
Restarted. I think I fixed that issue just after this branch was based from. At least I hope, otherwise I don't know why this is happening!

internal/painter/software/draw.go
```diff
@@ -7,9 +7,9 @@ import (
 	"fyne.io/fyne"
 	"fyne.io/fyne/canvas"
 	"fyne.io/fyne/internal"
-	"fyne.io/fyne/internal/painter"
+	"fyne.io/fyne/internal/cache"
+	pnt "fyne.io/fyne/internal/painter"
```
andydotxyz, line 11: "pnt"? why did this move from "painter"? If it has to the name could be something more clear like "paint" or "intPainter"

toaster:
> Is it intentional that this removes transparency from the test image creation?

Not really. I think it is clear that the image does not cover this area because this is the theme background. I could apply a special custom theme with a transparent background to get the old image but I would question the use of this.

> Lastly I think the number of times a comment reads "// xxx satisfies the yyy interface." is becoming quite high - does this not make it harder for the reader of this code as they have to go elsewhere for the documentation?

I wish for a kind of `// @see MyInterface` to provide automatic links to follow.
This could also work as hint for the linter.
But even without this, I’m totally against duplicating documentation (“satisfies yyy” does not count as duplicate ;)).

toaster:
> A strange import name needs fixed I think.

fixed

internal/painter/software/painter_test.go
```diff
@@ -0,0 +1,103 @@
...
+	c.SetContent(img)
+	c.Resize(fyne.NewSize(50, 50))
+	p := software.NewPainter()
+
+	target := p.Paint(c)
+	test.AssertImageMatches(t, "draw_image_default.png", target)
+}
+
+func TestPainter_PaintImage_ContainX(t *testing.T) {
+	test.ApplyTheme(t, theme.LightTheme())
```
stuartmscott, line 78: Why is Theme is only applied to TestPainter_PaintImage_ContainX and TestPainter_PaintImage_ContainY, but not the others?

toaster: For the other it doesn’t matter because the image should fill the whole canvas.
The theme is set to ensure a deterministic background.

andydotxyz:
> I could apply a special custom theme with a transparent background to get the old image but I would question the use of this.

You're right - not worth it, the background is fine. I just wanted to make sure it was an intentional change.

andydotxyz:
> But even without this, I’m totally against duplicating documentation (“satisfies yyy” does not count as duplicate ;)).

Well, thinking from a developers point of view this is not particularly helpful. Many of the comments added in this PR are not necessary and I would prefer we leave no documentation for an internal method than one that has placeholder feel to it. That way someone who wants to help writing something more useful will see that it's missing.

Looking at the `testDriver.Run()` for example it would be far more meaningful to say "Run does nothing in the test driver as we do not handle an event loop".

toaster:
> Looking at the `testDriver.Run()` for example it would be far more meaningful to say "Run does nothing in the test driver as we do not handle an event loop".

Agreed. But this is only helpful if one knows that `Run` isn’t there to do nothing but to implement an interface which assumes that `Run` handles the event loop.
And many (probably the most) implementations of interface methods do exactly what they should from the interface user’s point of view.

andydotxyz:
Consider Widgets that extend BaseWidget, when they implement (override) a Widget method they are doing so because they have special behaviour.
The cases where there is literally nothing to say may exist, but I am against the concept of automatically filling in a placeholder into production code.

toaster:
Overwriting a method and implementing an interface are different things.

What about writing comments like:
```go
// CreateRenderer creates a new renderer for MyCoolWidget as requested by the fyne.Widget interface.
func (w *MyCoolWidget) CreateRenderer() fyne.WidgetRenderer {
…
}
```

andydotxyz:
We know the difference between overriding and implementing - but does the person who reads the docs? Documentation should be presented as is most beneficial for the reader, not the developer writing it.

Ending with the suggestted "... as requested by the fyne.Widget interface." sounds good. Then again the shorter and "kindof standard" version " @see fyne.Widget" or we could make up " @implements fyne.Widget" as the doc generator I am working on will be more sophisticated than the default godoc... ;)

toaster:
> Ending with the suggestted "... as requested by the fyne.Widget interface." sounds good. Then again the shorter and "kindof standard" version " @see fyne.Widget" or we could make up " @implements fyne.Widget" as the doc generator I am working on will be more sophisticated than the default godoc... ;)

Like so?
```go
// CreateRenderer creates a new renderer for MyCoolWidget.
// @implements fyne.Widget
func (w *MyCoolWidget) CreateRenderer() fyne.WidgetRenderer {
…
}
```

Or so?
```go
// CreateRenderer creates a new renderer for MyCoolWidget (@implements fyne.Widget).
func (w *MyCoolWidget) CreateRenderer() fyne.WidgetRenderer {
…
}
```

andydotxyz:
I thought about this more and realised that Go has sort of establised it's own format with the "// Deprecated:" lines that we have used in some places. Combine that with what I think is an honest need for a short explanation I think we should go with the following:

```
// CreateRenderer sets up the renderer for MyCoolWidget
// Implements: fyne.Widget
func (w *MyCoolWidget) CreateRendrer() fyne.WidgetRenderer {
...
}
```

That way we are able to match the godoc format and yet able to do clever things with our documentation generation.

toaster:
ack, will adjust _all code in the internet_ in the evening ;)

PR #959, opened 2020-05-09
Add OnSubmit and OnCancel funcs

### LINKED ISSUE #475: Form.OnSubmit does not add Submit button to the Form
**Describe the bug**
When Form.OnSubmit = func() {} Submit should be added to the Form but Submit does not appear

 - OS: MacOS Mojave
 - Version 10.14.6
 - Go version 1.13.1
 - Fyne version 1.1.1

steveoc64:
### Description:
Allows the user to reset OnSubmit / OnCancel on a form, and have it correctly reconstruct the Cancel/Submit buttons on the next form.Refresh()

## Concept 
https://youtu.be/O37njf6dvHo

## Usages:

form.OnSubmit = func() {...}
form.SubmitText = "Something New"
form.Refresh()  // repaints the screen with Submit button pointing to the new function

### Fixes #475 

This PR allows the user to dynamically reset the buttons on a form after the form is created / rendered.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

#### TODO

May need to update this code after 955 lands, to make use of the new propertyLock mechanism


andydotxyz:
Hmm, this (naming?) seems to go against patterns elsewhere, “On...” represents an event being called so “OnSubmit()” is called to indicate that Submit event is happening - like in check’s OnChanged. The field can be set using “OnChanged =“ to a new handler.

But here the setter looks like an event invocation “OnSubmit(handler)” so it’s different to the pattern that we have used in other widgets...

steveoc64:
That is true about the naming.

All widgets pretty much have an OnChanged func ptr as a public attrib. Hence the discussion about whether we were going to clean that up as part of 955.

I would prefer that all widgets had a method `OnChanged(func())` that was used to set handlers internally.  Otherwise its a race cond ... and anything that needs postprocessing (such as updating handlers on a form like this example) needs the user to do special things to get their change propogated onto the widget.

steveoc64:
... also, the existing OnChanged being a publically accessible func ptr means that none of the user level code that attaches handlers for events can be chained.  So they basically need to create local vars every time they want to use a widget that they want to attach an OnChanged func too.

Changing this around will greatly clean all this up

widget/form_test.go
```diff
@@ -46,9 +47,53 @@ func TestForm_CustomButtonsText(t *testing.T) {
...
+
+	go test.Tap(form.submitButton)
+	func() {
+		select {
+		case <-tapped:
+		case <-time.After(1 * time.Second):
+			assert.Fail(t, "Timed out waiting for submit button tap")
+		}
+	}()
+	assert.Equal(t, 1, scount)
```
toaster: Why do you run the tap in a goroutine? I think the whole async part can be removed:
```go
test.Tap(…)
assert.Equal(t, 1, scount)
```

widget/form_test.go
```diff
@@ -72,7 +72,8 @@ func TestForm_AddRemoveButton(t *testing.T) {
 			assert.Fail(t, "Timed out waiting for submit button tap")
 		}
 	}()
-	assert.Equal(t, 1, scount)
+	assert.Equal(t, 1, scount) // because we tapped submit
```
toaster, line 75: I suggest to use `assert`’s message instead of adding comments:
```suggestion
	assert.Equal(t, 1, scount, "tapping submit should increase scount")
```

widget/form_test.go
steveoc64: copypasta from the button_test code

I think its there to make sure that the if the button press never happens, it doesnt hang the test ?

But yeah, that looks much simpler, will do

widget/form_test.go
steveoc64: :+1:

widget/form_test.go
steveoc64: that works :+1:

widget/form.go
```diff
@@ -21,8 +21,8 @@ func NewFormItem(text string, widget fyne.CanvasObject) *FormItem {
 
 // Form widget is two column grid where each row has a label and a widget (usually an input).
 // The last row of the grid will contain the appropriate form control buttons if any should be shown.
-// Setting OnSubmit will set the submit button to be visible and call back the function when tapped.
```
andydotxyz, line 24: These docs should go back to what they were before I guess?

widget/form.go
```diff
@@ -32,7 +32,11 @@ type Form struct {
 	SubmitText string
 	CancelText string
 
-	itemGrid *fyne.Container
+	itemGrid     *fyne.Container
+	vbox         *Box
```
andydotxyz, line 36: I think this variable is no tneeded. if it is can it have a better name?

widget/form.go
andydotxyz: This still references "Calling OnSubmit" instead of "Setting OnSubmit"

widget/form.go
```diff
@@ -88,28 +126,18 @@ func (f *Form) CreateRenderer() fyne.WidgetRenderer {
...
-		buttons.Append(submitButton)
-	}
-	return cache.Renderer(NewVBox(f.itemGrid, buttons))
+	// Create the buttons and hide the box till they are needed
+	f.cancelButton = NewButtonWithIcon("", theme.CancelIcon(), f.OnCancel)
+	f.submitButton = NewButtonWithIcon("", theme.ConfirmIcon(), f.OnSubmit)
+	f.submitButton.Hide()
+	f.cancelButton.Hide()
+	f.buttonBox = NewHBox(layout.NewSpacer(), f.cancelButton, f.submitButton)
+	f.buttonBox.Hide()
```
toaster, line 135: All three `Hide` calls will also be done by `setButtons`. I suggest to remove them here.

widget/form.go
```diff
@@ -48,13 +54,13 @@ func (f *Form) ensureGrid() {
 }
 
 // Append adds a new row to the form, using the text as a label next to the specified Widget
-func (f *Form) Append(text string, widget fyne.CanvasObject) {
+func (f *Form) Append(text string, widget fyne.CanvasObject) *Form {
```
toaster, line 57: Is the return of the form still needed? AFAIK we don’t use chaining elsewhere. And ATM I doubt the business value of it :). Last but not least: If it should stay, it has to be covered by tests.

andydotxyz:
Looks good to me, @toaster?
I think this is candidate for a squash-and-merge if you don't mind.

PR #960, opened 2020-05-10
Bugfix/async resize

### LINKED ISSUE #938: App sometimes loads with broken aspect
Looks like some bad size calculations or resize.

The demo window sometimes loads too tall. This time it is the right size but with the wrong size of content.
Is this what you saw on WASM @Bluebugs?

OS: macOS 10.15.4
Fyne: develop

Screenshot 2020-05-05 at 16 51 36

Bluebugs:
### Description:
This should make Resize an always asynchronous operation and maybe solve the strange scaling issue shown on some platform. Needs testing on more system than just Linux.

Fixes #938

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -91,9 +91,11 @@ type window struct {
 	menuTogglePending       fyne.KeyName
 	menuDeactivationPending fyne.KeyName
 
-	xpos, ypos    int
-	width, height int
-	shouldExpand  bool
+	xpos, ypos                      int
+	width, height                   int
+	requestedWidth, requestedHeight int
+	shouldWidth, shouldHeight       int
+	shouldExpand                    bool
 
 	pending []func()
 }
@@ -185,10 +187,11 @@ func (w *window) Resize(size fyne.Size) {
 
 		width, height := internal.ScaleInt(w.canvas, bigEnough.Width), internal.ScaleInt(w.canvas, bigEnough.Height)
 		if w.fixedSize || !w.visible { // fixed size ignores future `resized` and if not visible we may not get the event
+			w.shouldWidth, w.shouldHeight = width, height
 			w.width, w.height = width, height
 		}
 		w.viewLock.Unlock()
-
+		w.requestedWidth, w.requestedHeight = width, height
 		w.viewport.SetSize(width, height)
 	})
 }
@@ -282,22 +285,21 @@ func (w *window) fitContent() {
 	w.viewLock.RLock()
 	view := w.viewport
 	w.viewLock.RUnlock()
+	w.shouldWidth, w.shouldHeight = w.width, w.height
 	if w.width < minWidth || w.height < minHeight {
 		if w.width < minWidth {
-			w.width = minWidth
+			w.shouldWidth = minWidth
 		}
 		if w.height < minHeight {
-			w.height = minHeight
+			w.shouldHeight = minHeight
 		}
 		w.viewLock.Lock()
 		w.shouldExpand = true // queue the resize to happen on main
 		w.viewLock.Unlock()
 	}
 	if w.fixedSize {
-		w.width = internal.ScaleInt(w.canvas, w.Canvas().Size().Width)
-		w.height = internal.ScaleInt(w.canvas, w.Canvas().Size().Height)
-
-		view.SetSizeLimits(w.width, w.height, w.width, w.height)
+		w.shouldWidth, w.shouldHeight = w.requestedWidth, w.requestedHeight
+		view.SetSizeLimits(w.requestedWidth, w.requestedHeight, w.requestedWidth, w.requestedHeight)
 	} else {
 		view.SetSizeLimits(minWidth, minHeight, glfw.DontCare, glfw.DontCare)
 	}
@@ -1457,11 +1459,13 @@ func (w *window) create() {
 			fn()
 		}
 
+		w.requestedWidth, w.requestedHeight = w.width, w.height
+
 		if w.fixedSize { // as the window will not be sized later we may need to pack menus etc
 			w.canvas.Resize(w.canvas.Size())
 		}
 		// order of operation matters so we do these last items in order
-		w.viewport.SetSize(w.width, w.height) // ensure we requested latest size
+		w.viewport.SetSize(w.shouldWidth, w.shouldHeight) // ensure we requested latest size
 	})
 }
 
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -531,7 +531,7 @@ func (w *window) moved(_ *glfw.Window, x, y int) {
 }
 
 func (w *window) resized(_ *glfw.Window, width, height int) {
-	canvasSize := fyne.NewSize(internal.UnscaleInt(w.canvas, width), internal.UnscaleInt(w.canvas, height))
+	canvasSize := w.computeCanvasSize(width, height)
 	if !w.fullScreen {
 		w.width = internal.ScaleInt(w.canvas, canvasSize.Width)
 		w.height = internal.ScaleInt(w.canvas, canvasSize.Height)
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -179,19 +179,17 @@ func (w *window) RequestFocus() {
 
 func (w *window) Resize(size fyne.Size) {
 	// we cannot perform this until window is prepared as we don't know it's scale!
-
+	bigEnough := size.Max(w.canvas.canvasSize(w.canvas.Content().MinSize()))
 	w.runOnMainWhenCreated(func() {
-		w.canvas.Resize(size)
 		w.viewLock.Lock()
 
-		width, height := internal.ScaleInt(w.canvas, size.Width), internal.ScaleInt(w.canvas, size.Height)
+		width, height := internal.ScaleInt(w.canvas, bigEnough.Width), internal.ScaleInt(w.canvas, bigEnough.Height)
 		if w.fixedSize || !w.visible { // fixed size ignores future `resized` and if not visible we may not get the event
 			w.width, w.height = width, height
 		}
 		w.viewLock.Unlock()
 
 		w.viewport.SetSize(width, height)
-		w.fitContent()
 	})
 }
 
@@ -533,10 +531,6 @@ func (w *window) moved(_ *glfw.Window, x, y int) {
 }
 
 func (w *window) resized(_ *glfw.Window, width, height int) {
-	if w.fixedSize {
-		return
-	}
-
 	canvasSize := fyne.NewSize(internal.UnscaleInt(w.canvas, width), internal.UnscaleInt(w.canvas, height))
 	if !w.fullScreen {
 		w.width = internal.ScaleInt(w.canvas, canvasSize.Width)
@@ -545,6 +539,12 @@ func (w *window) resized(_ *glfw.Window, width, height int) {
 
 	if !w.visible { // don't redraw if hidden
 		w.canvas.Resize(canvasSize)
+		return
+	}
+
+	if w.fixedSize {
+		w.canvas.Resize(canvasSize)
+		w.fitContent()
 		return
 	}
 
--- a/internal/driver/glfw/window_notwindows.go
+++ b/internal/driver/glfw/window_notwindows.go
@@ -1,6 +1,16 @@
+//go:build !windows
 // +build !windows
 
 package glfw
 
+import (
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/internal"
+)
+
 func (w *window) setDarkMode() {
+}
+
+func (w *window) computeCanvasSize(width, height int) fyne.Size {
+	return fyne.NewSize(internal.UnscaleInt(w.canvas, width), internal.UnscaleInt(w.canvas, height))
 }
--- a/cmd/fyne_demo/tutorials/window.go
+++ b/cmd/fyne_demo/tutorials/window.go
@@ -25,6 +25,16 @@ func windowScreen(_ fyne.Window) fyne.CanvasObject {
 			w.SetFixedSize(true)
 			w.Show()
 		}),
+		widget.NewButton("Toggle between fixed/not fixed window size", func() {
+			w := fyne.CurrentApp().NewWindow("Toggle fixed size")
+			w.SetContent(fyne.NewContainerWithLayout(layout.NewCenterLayout(), widget.NewCheck("Fixed size", func(toggle bool) {
+				if toggle {
+					w.Resize(fyne.NewSize(240, 180))
+				}
+				w.SetFixedSize(toggle)
+			})))
+			w.Show()
+		}),
 		widget.NewButton("Centered window", func() {
 			w := fyne.CurrentApp().NewWindow("Central")
 			w.SetContent(fyne.NewContainerWithLayout(layout.NewCenterLayout(), widget.NewLabel("Hello World!")))
--- a/internal/driver/glfw/loop.go
+++ b/internal/driver/glfw/loop.go
@@ -136,10 +136,13 @@ func (d *gLDriver) runGL() {
 				if expand && !fullScreen {
 					w.fitContent()
 					w.viewLock.Lock()
+					shouldExpand := w.shouldExpand
 					w.shouldExpand = false
 					view := w.viewport
 					w.viewLock.Unlock()
-					view.SetSize(w.width, w.height)
+					if shouldExpand {
+						view.SetSize(w.shouldWidth, w.shouldHeight)
+					}
 				}
 
 				newWindows = append(newWindows, win)
--- a/internal/driver/glfw/window_windows.go
+++ b/internal/driver/glfw/window_windows.go
@@ -6,6 +6,7 @@ import (
 	"unsafe"
 
 	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/internal"
 	"golang.org/x/sys/windows/registry"
 )
 
@@ -40,4 +41,11 @@ func isDark() bool {
 	}
 
 	return useLight == 0
+}
+
+func (w *window) computeCanvasSize(width, height int) fyne.Size {
+	if w.fixedSize {
+		return fyne.NewSize(internal.UnscaleInt(w.canvas, w.width), internal.UnscaleInt(w.canvas, w.height))
+	}
+	return fyne.NewSize(internal.UnscaleInt(w.canvas, width), internal.UnscaleInt(w.canvas, height))
 }


andydotxyz:
The failures are formatting I think - run goimports on your repo should fix it.
(But don't re-format vendor folder).

internal/scale.go
```diff
@@ -16,6 +16,10 @@ func ScaleInt(c fyne.Canvas, v int) int {
 	}
 }
 
+func ScaleSize(c fyne.Canvas, s fyne.Size) fyne.Size {
```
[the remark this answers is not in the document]
andydotxyz: Correct, pixel values should be (int, int) to clearly differentiate from fyne sizes.
If this is a little abstract we could add a "PixelSize" which is int, int - but I think this might pollute the API and encourage people to think that it's usable instead of normal Size

Bluebugs:
> Regarding the removal of `ignoreResize`: I can see how this would fix the scale issue eventually but it might add flickering. @andydotxyz surely knows more about this :).

`ignoreResize` is actually part of the problem here. The original code is build with the idea that when you resize the window, the viewport will instantly obey and you actually can fit the content in the new size you have choosen. This is incorrect. You are technically sending a request for resize to the window manager who can decide whatever it want. So you can only fit your current widgets inside the window size you know at the time you are there and you will get a second call later via the resize callback that will actually result in fitting the content in that requested size.
 
> What was the purpose of not setting the canvas size as window size for fixed size windows anymore?

The goal is that we should never change directly the size of the canvas and only change it on callback from the window manager. The resize code already does that, just we have some custom logic to bypass that case, which can result in problems when the window get resized by the window manager differently from what the code expect (or at a different time the code expect).

internal/scale.go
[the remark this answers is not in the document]
Bluebugs: Ok, I will propose to introduce a PixelSize then. The idea of going with an intermediate type is to discourage people to directly modify this value. As they should be considered read only for almost all of their use and should not be tempered with.

Bluebugs:
>     * the removal of `ignoreResize`

This was the buggy behavior.

>     * the change that fixed size windows don’t get their size adjusted to the canvas’ size in `fitContent` anymore

This was also the buggy behavior.

andydotxyz:
The removal of some of the code has broken certain types of window. Specifically fixed size windows no longer show at the requested size.
You can see this (at least on macOS?) with fyne_demo -> Windows -> Fixed size. This window should not be shown at min size.

andydotxyz:
As we're in code freeze time is this something that will have to slip?
The issue that it fixed has been fixed in other ways so it is lower priority than before.

toaster:
FYI: I take force push after a review as bad behaviour.

Why?
Because a reviewed PR should be modified by additional commits addressing the feedback.
If the feedback is squashed into the old commits and then force-pushed it is hard to see the difference between the reviewed and the current state. Thus it takes significantly more time to review the changes. And you don't want to waste others time, do you?

The PR might be closed and replaced by another one If the complete code is to be reworked because of the review.

Bluebugs:
Sorry @toaster, I force pushed as this is kind of the expected strategy with git as the goal of having a clean history is part of git design. If the project prefer to have additional patch that do address comment in further commit, I will do so in the future.

@andydotxyz sorry, I didn't have time to finish this and it is more invasive than I wanted. I do think their are bug that this fixes, but most people are not facing them, which make it low priority, I think.

toaster:
> Sorry @toaster, I force pushed as this is kind of the expected strategy with git as the goal of having a clean history is part of git design. If the project prefer to have additional patch that do address comment in further commit, I will do so in the future.

Thank you!

This is the second time I hear of this “clean history” and how some big project aims to achieve it. I don’t know if the git team wanted the git users to create clean histories. But I know that as a reviewer, I want to see if my objections and suggestions were addressed the right way. In this case it is very helpful to see the additional diff instead of squashed commits.

Regarding “clean history”: I think the concept is crap but everyone should do as he pleases as long as he doesn’t harm others.

P.S.: Read a bit about this “clean history” thing and it affirmed my opinion. I think it’s misguided. If the git people really wanted to make this the way to go, they wouldn’t have put it behind the `--force` option. Doing something forcefully is almost always against the intentional way.

andydotxyz:
Looks like a bad merge - you're getting compile errors and format errors in Travis

geometry.go
```diff
@@ -40,6 +40,17 @@ func NewSize(w int, h int) Size {
 	return Size{w, h}
 }
 
+// PixelSize describe an object width and height in Pixels
```
andydotxyz, line 43: I don't think that we should add a new public API for this as it's an internal detail.
If you think that our driver needs this type then please move it out of the public API.

internal/driver/glfw/canvas.go
```diff
@@ -381,7 +387,9 @@ func (c *glCanvas) ensureMinSize() bool {
 	shouldResize := windowNeedsMinSizeUpdate && (c.size.Width < min.Width || c.size.Height < min.Height)
 	c.RUnlock()
 	if shouldResize {
-		c.Resize(c.Size().Union(c.MinSize()))
+		// We need to resize the Window actually as the content doesn't fit at the moment
+		log.Println("Resize window ensureMinSize")
```
andydotxyz, line 391: Please remove this debug

internal/driver/glfw/window.go
```diff
@@ -115,7 +115,8 @@ func (w *window) SetFullScreen(full bool) {
 		if full {
 			w.viewport.SetMonitor(monitor, 0, 0, mode.Width, mode.Height, mode.RefreshRate)
 		} else {
-			w.viewport.SetMonitor(nil, w.xpos, w.ypos, w.width, w.height, 0)
+			w.viewport.SetMonitor(nil, w.xpos, w.ypos, w.size.Width, w.size.Height, 0)
+			w.fullScreen = false
```
andydotxyz, line 119: Isn't this set in line 106?

internal/driver/glfw/window.go
```diff
@@ -256,21 +252,29 @@ func (w *window) fitContent() {
...
+			// their is no point into doing anything
+			return
+		}
 
-		w.viewport.SetSizeLimits(w.width, w.height, w.width, w.height)
+		w.viewport.SetSizeLimits(limitSize.Width, limitSize.Height, limitSize.Width, limitSize.Height)
+		w.viewport.SetSize(limitSize.Width, limitSize.Height)
 	} else {
+		// Adjust size to be at least minSize
+		if w.size.Width < minWidth || w.size.Height < minHeight {
```
andydotxyz, line 267: This is no longer applying to FixedSize - is it your intent to allow FixedSize Resize() to be less than MinSize()?

internal/driver/glfw/canvas.go
```diff
@@ -110,10 +111,15 @@ func (c *glCanvas) SetContent(content fyne.CanvasObject) {
 
 	c.content.Resize(c.content.MinSize()) // give it the space it wants then calculate the real min
 	// the pass above makes some layouts wide enough to wrap, so we ask again what the true min is.
+
 	newSize := c.size.Union(c.canvasSize(c.content.MinSize()))
 	c.Unlock()
 
-	c.Resize(newSize)
+	c.Resize(c.size) // Resize the content to this canvas size
```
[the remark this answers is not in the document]
Bluebugs: I think it is necessary here. The canvas content got resized above with that comment about real min. We now need to get the content to fit inside the real canvas size for this rendering round. So I do think it is mandatory to resize the canvas back to its own size and get everything correct.

internal/driver/glfw/window.go
Bluebugs: Oops, this wasn't there before merging with develop, will fix.

internal/driver/glfw/window.go
Bluebugs: I believe it is, but in an indirect way. When canvas.SetContent is called, it resize the window to the new canvas content.MinSize. So at this point, I think the canvas size is already constrained to be MinSize().

geometry.go
Bluebugs: I moved it to internal/geometry.go.

andydotxyz:
I can't compile this now and neither can Travis

andydotxyz:
I pieced together the missing code.

Unfortunately this now breaks windows created with a specific size and position.
In fyne_demo choose WIndow->Fixed size or Window->Centered and you will see they both appear approx minsize, top left instead of the expected size and position.

internal/driver/glfw/canvas.go
```diff
@@ -110,10 +110,15 @@ func (c *glCanvas) SetContent(content fyne.CanvasObject) {
...
+
+	// If the canvas is to small, we need it to be resized up, but that should be done when the window is actually resized
+	c.context.(*window).Resize(newSize)
+
```
toaster, line 121: I think this is not quite right yet.
First: the canvas should not resize the window IMO. I think the caller of `canvas.SetContent` should handle this.
Second: the canvas should not type-assert its `context` to `*window` just because it is the same by chance. These are two different semantics.
Third: resizing to the current size and then indirectly resizing to the real size really seems to be the wrong thing. There might be something else wrong if this really fixes some misbehaviour.

internal/driver/glfw/canvas.go
```diff
@@ -381,7 +386,8 @@ func (c *glCanvas) ensureMinSize() bool {
 	shouldResize := windowNeedsMinSizeUpdate && (c.size.Width < min.Width || c.size.Height < min.Height)
 	c.RUnlock()
 	if shouldResize {
-		c.Resize(c.Size().Union(c.MinSize()))
+		// We need to resize the Window actually as the content doesn't fit at the moment
+		c.context.(*window).Resize(c.Size().Union(c.MinSize()))
```
toaster, line 390: See my previous comment.

internal/driver/glfw/window.go
```diff
@@ -256,21 +251,29 @@ func (w *window) fitContent() {
...
+			targetWidth, targetHeight := w.size.Width, w.size.Height
+			if targetWidth < minWidth {
+				targetWidth = minWidth
+			}
+			if targetHeight < minHeight {
+				targetHeight = minHeight
+			}
+			w.viewport.SetSize(targetWidth, targetHeight)
```
toaster, line 274: In the end `fitContent` now changes the viewport size without adjusting the window’s size as well. This feels wrong. Maybe some other code adjusts this later but I think the window’s size and the viewport size should be very thightly coupled. A method should not change one without changing the other.

toaster: If `viewport.GetSize` is cheap, we could drop the window’s size property at all.

toaster (dismissed):
This looks like a lot of distributed calculations to me. Maybe the responsibilities are not quite right yet?
See my comments.

internal/driver/glfw/window.go
andydotxyz: The viewport size should not be adjusted outside the render loop I think - this may be causing some of the resize flicker?

internal/driver/glfw/canvas.go
Bluebugs: 1. I am not sure it is a good idea to move that duty outside of SetContent for a few reason. First one, will be possible forgotten side call when using SetContent (as it would not be a function that work by itself, but require additional step not included). Second is that the logic for resizing the content to any size different from the canvas would have to be moved outside of the canvas logic and I am not sure this belong in window.go.
2. Hum, in that case, what do you recommend? I copied that approach from SetScale. This should also get corrected, I guess.
3. I agree, I just didn't want to undo the first content.Resize due to the comment attached to it and the fear to trigger unwarranted side effect. I feel like removing that double Resize should be part of a different PR later on once this one work properly.

internal/driver/glfw/window.go
Bluebugs: I think you are missing some context. When you do call viewport.SetSize, the outcome of that call is not that the window is resized once that call is done, but that a request has been made to resize the window. The window resize call is asynchronous in nature and any change can be denied by the window manager. So you can only update the window size once you get the callback from the window manager that tell your new window size. And any point in between your window is still the same size as it was when you initiated the viewport.SetSize.

Basically the entire point of this patch set is to make sure that the viewport size and the window size always reflect the true value that the window manager has given it at any point in time.

Bluebugs: Also I would not assume that the viewport.GetSize is cheap as on some system it might require a round with the window manager to get that information. I do not know if GLFW does cache that value for us.

andydotxyz: Oops, I was confusing viewport.SetSize and gl.Viewport() sorry!

internal/driver/glfw/canvas.go
andydotxyz: Is Window.Resize even needed here? ensureMinSize will maybe take care of it?

internal/driver/glfw/window.go
toaster: Thanks for the explanation. Then of course my comment is totally wrong. Alas, they could have named the method `RequestResize` :).

internal/driver/glfw/canvas.go
toaster: I don’t have a solution that resolve my objections at hand.
Maybe the whole windows/canvas/content (re)size handling is not right regarding responsibilities?
Probably your changes just revealed some shortcomings.

Bluebugs: I will try without the Window.Resize here, but indeed the ensureMinSize might do the right work, except with a potential delay. So we might have a frame that get the wrong size that we could have otherwise avoided. Will check.

internal/driver/glfw/canvas.go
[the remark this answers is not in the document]
andydotxyz: Not sure I follow. In this PR are we not asserting that canvas should always follow window size?
The comment in your code indicates that we are resizing content but it is actually resizing the canvas - to it's current size. What does such a line acheive, especially if the window size that comes quickly after will override it?

internal/driver/glfw/canvas.go
andydotxyz: The comment above this line almost explains why it shouldn't exist, not why it does...

internal/driver/glfw/canvas.go
andydotxyz: In this instance it seems OK - I suppose the code could be moved into window to tidy things up a little.
Context() returning *window is not a co-incidence, it is how this driver is designed. The generic nature is to support passing GL contexts in different drivers. We pass the value in and get it back again so it's pretty safe.

internal/driver/glfw/window.go
andydotxyz: Are you sure? Really we have instilled the minSize for window in the non-fixed scenario as well, so if constraints are respected it should not be needed in any instance.
Think about it programatically. If someone calls Resize() manually on a fixed size window this will now allow it to be smaller than minSize...

internal/geometry.go
```diff
@@ -0,0 +1,12 @@
+package internal
+
+// PixelSize describe an object width and height in Pixels
+type PixelSize struct {
+	Width  int // The number of pixels along the X axis.
```
andydotxyz, line 5: As per slack discussions, if we really need this to differentiate then the Width should probably be WidthPx so that it's clear when autocompling what the type is rather than just in the field type or parameter lists.

internal/driver/glfw/canvas.go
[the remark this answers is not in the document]
Bluebugs: Their is no guarantee that the window size change would come quickly enough that you would push a frame with the content at the right size.

The canvas Resize call does a bit more than just content.Resize, yes, but I choose to do so to be sure no code call is missing and reduce code duplication. From my understanding, most of the code of canvas.Resize is about resizing content.

internal/driver/glfw/canvas.go
Bluebugs: I guess that my english is not correct then. Could you suggest a better comment in your opinion?

internal/driver/glfw/window.go
Bluebugs: Hum, I have build this with the assumption that calling Resize on a fixed size window once it has a content would be a noop if the new requested size was below the SizeLimits. I guess I should check that before calling viewport.SetSize in case the window manager doesn't enforce it when coming from the application. I have also realized that the call to fitContent in Resize is unnecessary.

internal/geometry.go
Bluebugs: Ok, will do.

Bluebugs:
> Running fyne_demo - tapping Windows -> Fixed Size is not working.
> 
> On my linux desktop the window loads at the right size then shrinks back to minsize...

Ah sorry, missed to retest that after merging with latest develop. Will investigate.

andydotxyz:
Would it be easier to rebase the commits on a new branch off `develop`?

andydotxyz:
I tried rebasing this work on latest but I can't seem to make it work.
I keep hitting deadlocks and/or windows that don't respect requested size any more.

I do wonder if we should start again on this and define, as of 2.0.1, what bug this is aiming to fix.

internal/driver/glfw/window.go
```diff
@@ -533,10 +531,6 @@ func (w *window) moved(_ *glfw.Window, x, y int) {
 }
 
 func (w *window) resized(_ *glfw.Window, width, height int) {
-	if w.fixedSize {
```
andydotxyz, line 536: Unfortunately Windows fixed size did not work :(

I got it working locally by re-instating this `if w.fixedSize` and setting `width = w.width; height = w.height`.
Then it all worked on windows OS :).

andydotxyz (changes requested):
Windows being a pain - possibly a small change fixes it, mentioned inline.

internal/driver/glfw/window.go
andydotxyz: I also tested the fix on macOS and it works fine :)

andydotxyz:
On macOS this crashed as it triggered another bug  (good catch ;) ).
By updating the `fitContent` line inside `SetFixedSize` to
```
		w.runOnMainWhenCreated(w.fitContent)
```
that fixed up macOS well :).

The new checkbox is a great test - but why not just add it to the existing fixed size window so we're not adding another item to the list of window types?

internal/driver/glfw/window_notwindows.go
```diff
@@ -1,6 +1,16 @@
+//go:build !windows
```
andydotxyz, line 1: Let's not add new Go features while we still target 1.14

andydotxyz (changes requested):

internal/driver/glfw/window_notwindows.go
Jacalz: I think we should be able to add all of these in one go using `gofmt` later, which probably makes more sense. However, I don’t think they should be causing any issues if we have both.

andydotxyz (approved):
Nice, thanks :)
Good to have this in, looks like it could start resolving some resize crashes as well.

internal/driver/glfw/canvas.go
andydotxyz: Dismissing the stale comments so we can merge

meagar:
This definitely seems to still be an issue. My steps to reproduce are to simply follow the README and try to run the simple "hello world" example, resulting in a crash when I resize the window.

Jacalz:
> This definitely seems to still be an issue. My steps to reproduce are to simply follow the README and try to run the simple [Getting Started](https://github.com/fyne-io/fyne/tree/release/v2.1.x#getting-started) example, resulting in a crash when I resize the window.

Please don't post off topic information. This is a PR that was merged relatively long ago. If you are having an issue, I'd suggest opening an issue instead.

andydotxyz:
> My steps to reproduce are to simply follow the README and try to run the simple [Getting Started](https://github.com/fyne-io/fyne/tree/release/v2.1.x#getting-started) example

I think you are looking for #2188

PR #961, opened 2020-05-10
WIP / RFC - Chain all the things

steveoc64:
### Description:
For all current widgets, make Setter funcs return the object, so that calls can be chained.

Most useful when constructing widgets, as there is no need to create temp vars to hold the widget.

Extends existing API, whilst being backward compatible

TODO - add tests ?  (all current tests pass, but I suppose might need regression tests too ?)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


andydotxyz:
Too much has changed in the API since this was opened - let's come back to it later, as it can't go in the 1.4 release.

Issue #211, opened 2019-04-06
Proposal: Allow dynamic themes via JSON templates


pbarnum:
It would be nice to be able to load and run custom themes during runtime.
* The app's main theme must still be written in Go and compiled with the app.
* New themes must follow strict guidelines in order to be a valid and usable theme.
* A app's theme should run after successfully loading.
* The app can handle multiple themes
  * This would require a menu system to switch between themes. Though this might be needed anyway if a user wanted to switch back to default after loading a custom theme.

andydotxyz:
I like this proposal.
Theme loading is probably out of scope for the 1.1 release but if someone contributed some solid code we would probably accept it.

Part of the conversation may be about the actual storage format. JSON and XML are easy from a Go point of view but are they right for people who will edit the themes?

It may also relate to the [settings discussion](https://github.com/fyne-io/fyne/wiki/Settings-API).

pbarnum:
Good point. I agree that JSON and XML are easy to work with and provide a simple and familiar interface, but it could get out of hand depending on a few factors:
* Number of customizations
* Level of detail within customizations
* Relationships between customizations
* Versioning
Perhaps a look into how other applications handle custom theming could answer some questions. VSCode comes to mind as does Sublime Text.

asciifaceman:
VSCode uses json defined overrides of sane defaults, likely with a right to left merge somewhere in the line

andydotxyz:
Depending on whether or not we change how themes work I don't think this needs to be multiple files. 
Unless you include the icon assets too...

pbarnum:
That gets into `assets/` territory where these should be loaded separately IMO. Assets could follow a set scheme declared in the code, something like `assets/framework/right_border.png` or `assets/icons/home.png`

andydotxyz:
Given that we have started using TOML for the FyneApp.toml build metadata, I wonder if we should do the same here.
It could be the same file even, but that seems less wise as an app may want to switch themes.

PR #968, opened 2020-05-10
958 colors

### LINKED ISSUE #958: Subjective comment on default colour choices in provided Fyne themes

steveoc64:
### Description:
Update Fyne themes to be closer to Material Design recommendations.

WorkInProgress - still a lot of rough edges in this PR, and Ive managed to break any code that defines a custom theme as well.  However, the addition of a subtle on-hover highlight on primary buttons makes a world of difference to the feel of the app, so maybe its worth breaking ?

Adding a PrimaryTextColor allows themes to be created that look right, and still have the button text on primary buttons visible.

Would like to continue to play with some ideas whilst this is open. There are still some gaps between Fyne and MD recommendations, although we are very close.

If it gets too wild, I will split this into 2 PRs 

1. Dead simple update to the primary color in dark mode, ie - no breaking changes.
2. New ideas / easier access for the developer to set up aesthetically pleasing themes without having to have a masters degree in color science. 

Aso - Im thinking that as there is a mathematical relationship between the "main branding color" and things like backgrounds, hover highlights, dimmed disabled text, etc ... we should in theory be able to generate the correct and aesthetic colors based on a smaller set of info.  Could keep the existing set of colors, but could add "SetPrimary() / SetBrandingColor() / SetAccentColor()" or something like that which ends up calculating the derived colors and plugging them into the theme.  Not sure yet, will play and see what happens.

The current theme engine / resource pool for icons has some good code in there for colorizing SVG icons already, but its a bit hard to access, and not easy to toggle the color of an icon / primary icon within the toolkit.  Not sure how that affects the SVG cache either. Wouldnt mind having a closer look at that while Im in here.

Fixes #958 

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


steveoc64:
As a special bonus, I have made the custom theme in the fyne demo even uglier, by setting the text color on primary buttons to be a bright orange on a hot pink background.

Would be good to update the Fyne Demo to have a really nice Theme tweaker page that lets the user pick colors and apply them to the current theme.

ala - https://cimdalli.github.io/mui-theme-generator/

Can have some fun with this PR

steveoc64:
One more thought - would be good too if any helper funcs recognised the "#RRGGBB" format hex string, as that is what 99% of these designer tools spit out these days

andydotxyz:
> Would be good to update the Fyne Demo to have a really nice Theme tweaker page that lets the user pick colors and apply them to the current theme.

That is a fun idea - though it may relate to the 2.0 ticket for making themes easier to customise. If you get it "just right" at the moment it would still take a lot of coding to manage to have those values in the developer's own app... #211

steveoc64:
Didnt see #211 - that makes perfect sense.

Ive already added it to my experiment branch anyway - pass it a blob of json, and it decodes the bits it needs, so you can copypaste directly off that MD theme generator site.  Seems to work.

andydotxyz:
I'd say that this ticket doesn't really fix #211 - that's the other PR surely?
I don't know what others would say but we have generally taken the line that breaking common API usage before 2.0 is a no-no. There are a number of apps out there that use a custom theme as the API has been in for a long time.

steveoc64:
Yeah I know, hence my comment in the body of the PR
```
If it gets too wild, I will split this into 2 PRs

Dead simple update to the primary color in dark mode, ie - no breaking changes.
```

So Ive split it out in one direction with the other PR already.  Will trim this one back so that it doesnt do hover on primary, or text reversal on primary, and that gives you a zero-breaking-change PR

Not sure its worth doing though - just changing the primary color, because that will break a tonne of tests that now have golden images.  I mean I could ... but that would bloat this PR with a tonne of new bitmaps to update all the golden images for the test cases.

Suggest we close this for 1.3 and think about improving the theme management for the user in whatever the next release is that is allowed to alter the API.  (is that 2.0 ???)

andydotxyz:
I know that you said it might split in 2, that is fine - but this issue still had the old scope on it, so I thought it should be updated.

Given that the issue this PR resolves is about "default colour choice" it implies that the primary colour should be changed, and so I don't see why that is no longer worth changing?
If it breaks tests all we need to do is drag the new images over old.
If we agree a change should be made then the tests should not hold us back surely?

steveoc64:
Managed to get this into shape where the primary color matches on both light and dark theme, and implement MD recommended dark text when used on primary colored buttons.

and added hover highlights for primary buttons

no changes to existing theme interface

as you can imagine, shoehorning this in without touching the existing theme interface adds some somewhat dirty code.

havnt touched the golden test cases yet, because none of my tests pass anyway - even on vanilla develop, they just fail all over the place, so its a bit hard to correct tests locally.  Means iterating through using travis to point out where the errors are

theme/color.go
```diff
@@ -1,11 +1,48 @@
...
-// in the intensity of the color.  Negative values allow for 
-// dimming the color
-func Brighten(c color.Color, factor float64) color.Color {
+// amount. Each value in the amount represents a 10% increase
+// in the intensity of the color.  Negative values allow for
+// dimming the color. This is used to compute a color such as
+// a hightlight on hover based on a base color.
+func Brighten(clr color.Color, amount float64) color.Color {
+	factor := 1.0 + (amount / 10.0)
+	c, ok := clr.(color.NRGBA)
```
andydotxyz, line 16: Why only support brigtening of NRGBA this might be surprising for a public API and it's not documented?
Why not get the channels from `clr.RGBA()`?

theme/color.go
```diff
@@ -1,11 +1,48 @@
...
+	}
+	r := limit(int(float64(c.R) * factor))
+	g := limit(int(float64(c.G) * factor))
+	b := limit(int(float64(c.B) * factor))
+	return color.NRGBA{r, g, b, c.A}
+}
 
-}
\ No newline at end of file
+func rgb(c color.NRGBA, s string) color.NRGBA {
```
andydotxyz, line 26: Is this helper used? why does it take a NRGBA in?

theme/theme.go
```diff
@@ -40,21 +40,26 @@ func LightTheme() fyne.Theme {
...
+		disabledButton: color.NRGBA{0xff, 0xff, 0xff, 0x10},
+		text:           color.NRGBA{0xff, 0xff, 0xff, 0xd0},
+		disabledText:   color.NRGBA{0xff, 0xff, 0xff, 0x40},
+		icon:           color.NRGBA{0xff, 0xff, 0xff, 0xe0},
+		disabledIcon:   color.NRGBA{0xff, 0xff, 0xff, 0x60},
 		hyperlink:      color.NRGBA{0x99, 0x99, 0xff, 0xff},
 		placeholder:    color.NRGBA{0xb2, 0xb2, 0xb2, 0xff},
-		primary:        color.NRGBA{0x1a, 0x23, 0x7e, 0xff},
-		hover:          color.NRGBA{0x31, 0x31, 0x31, 0xff},
+		primary:        color.NRGBA{0x9f, 0xa8, 0xda, 0xd0}, // same as light mode
```
andydotxyz, line 58: This says same as light mode but they are different?

theme/theme.go
```diff
@@ -40,21 +40,26 @@ func LightTheme() fyne.Theme {
...
 		scrollBar:      color.NRGBA{0x0, 0x0, 0x0, 0x99},
 		shadow:         color.NRGBA{0x0, 0x0, 0x0, 0x66},
+		//primary:        color.NRGBA{0x33, 0x29, 0x40, 0xff},
```
andydotxyz, line 61: please remove

widget/tabcontainer.go
```diff
@@ -79,7 +79,7 @@ func (t *TabContainer) CurrentTab() *TabItem {
 
 // SelectTabIndex sets the TabItem at the specific index to be selected and its content visible.
 func (t *TabContainer) SelectTabIndex(index int) {
-	if index < 0 || index >= len(t.Items) || t.current == index {
```
andydotxyz, line 82: Why are no not ignoring tapping the current tab any more?

andydotxyz:
Dark mode looks great (except the icon colour on primary buttons isn't matching).
I'm not sure the contrast is right in light mode for primary buttons though - it seems very difficult to read compared to the dark mode.

widget/tabcontainer.go
steveoc64: 2 things :

- its not just taps that SelectTabIndex
- without this change, get a situation where the 1st tab in a container does not correctly set the text color, because this line skips the refresh below

You dont notice it when the PrimaryTextColor and TextColor are the same value

theme/color.go
steveoc64: just discovered the beauty of clr.RGBA() - yep, will refactor all this later on

theme/color.go
steveoc64: same same - color.Color is the way to go

theme/theme.go
steveoc64: that is true, refactor on the way later on

theme/theme.go
steveoc64: :+1:

steveoc64:
Ideally, we need to compute the icon color for primary backgrounds.  Can see some interesting refactors coming up for the way we handle icons - they work very well where the color never changes, but its a bit tricky to implement multiple colors on an SVG.  We can fix that.

widget/button.go
```diff
@@ -105,18 +105,27 @@ func alignedPosition(align ButtonAlign, padding, objectSize, layoutSize fyne.Siz
...
 
 func (b *buttonRenderer) BackgroundColor() color.Color {
 	switch {
 	case b.button.Disabled():
 		return theme.DisabledButtonColor()
 	case b.button.Style == PrimaryButton:
-		return theme.PrimaryColor()
+		c := theme.PrimaryColor()
+		if b.button.hovered {
+			c = theme.Brighten(c, 1)
```
stuartmscott, line 126: Shouldn't this brighten in light mode, and darken in dark mode, or visa versa?

theme/color.go
```diff
@@ -0,0 +1,40 @@
...
+func Brighten(c color.Color, amount float64) color.Color {
+	r, g, b, a := c.RGBA()
+	factor := 1.0 + (amount / 10.0)
+	fn := func(i uint32, f float64) uint8 {
+		return uint8(float64(i) * f / 255)
+	}
+	return color.RGBA{fn(r, factor), fn(g, factor), fn(b, factor), uint8(a)}
+}
+
+// PrimaryTextColor returns either {black, white} depending on
```
stuartmscott, line 21: Rather than comparing with Primary, shouldn't this compare the contrast ratio against the Background color - it is more likely that text is rendered over the background colour than the primary colour.

It also feels like we shouldn't need to check this each time, shouldn't each theme have a text colour that doesn't change?

steveoc64: Agree - that it should be part of the theme and not computed on the fly.  Thats what I did originally in this PR, but that requires adding entries to the interface, which in turn introduces a breaking change.

andydotxyz:
Some of these commit messages could be a little more on point ;)
"colors" removes an unneeded comment and "tweak colors" fixes formatting.

andydotxyz:
Let's not expose new public APIs to support our testing

theme/color.go
steveoc64: keep in mind that this is used only to compute the color of text that is rendered over a primary background.

It doesnt affect text that is rendered over normal backgrounds.

widget/button.go
steveoc64: added code to detect if the current theme is light/dark based on the predominant BG color, and then brighten/dim accordingly

Is subtle, but it looks pretty cool

steveoc64:
Added Tilo as a reviewer, because of all the changes to tests

test/util_test.go
```diff
@@ -37,6 +37,7 @@ func TestAssertCanvasTappableAt(t *testing.T) {
 }
 
 func TestAssertImageMatches(t *testing.T) {
+	fyne.CurrentApp().Settings().SetTheme(test.MonoTheme())
```
andydotxyz, line 40: I think this line isn't needed - the theme should come from the current app config

theme/color.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"image/color"
+)
+
+// Brighten takes a color and brightens / dims it by the given
+// amount.
+// A value of 1.0 == the same
+// 1.1 = 10% brighter
+// 0.9 = 10% dimmer, etc
+func Brighten(c color.Color, amount float64) color.Color {
```
andydotxyz, line 12: I don't think this should be a public API that we have to support ongoing.
If it needs to be expoted then how about somewhere in `internal` packages?

theme/color.go
```diff
@@ -0,0 +1,55 @@
...
+			x = 0
+		}
+		return uint8(x >> 8)
+	}
+	return color.RGBA{fn(r), fn(g), fn(b), uint8(a)}
+}
+
+// PrimaryTextColor returns either {black, white} depending on
+// the brightness of the primary color
+func PrimaryTextColor() color.Color {
```
andydotxyz, line 32: This is only used in the widget package, perhaps it can be a private fuction there instead?

theme/color.go
```diff
@@ -0,0 +1,55 @@
...
+	r, g, b, _ := PrimaryColor().RGBA()
+	if (r+g+b)/3 < 24000 {
+		// is a dark primary
+		return color.White
+	}
+	return color.Black
+}
+
+func IsDark() bool {
```
andydotxyz, line 41: this does not need to be public

widget/button_test.go
```diff
@@ -118,9 +118,10 @@ func TestButton_Disabled(t *testing.T) {
 }
 
 func TestButton_Layout(t *testing.T) {
+	fyne.CurrentApp().Settings().SetTheme(test.MonoTheme())
 	test.NewApp()
 	defer test.NewApp()
-	test.ApplyTheme(t, theme.LightTheme())
+	test.ApplyTheme(t, test.MonoTheme())
```
andydotxyz, line 124: This is the third line that sets test theme - the SetTheme above, the NewApp below it then this ApplyTheme

andydotxyz:
I wonder if this should be 2 PRs...
The #958 issue is not addressed by these changes (light mode and icon colour).
A lot of the changes to test handling that are unrelated to the issue that this PR names and I wouldn't like one to slow landing the other.

andydotxyz:
Travis is red across the board, looks like a combination of format, compile and test errors

andydotxyz:
I think this has been superceded by #1195, but closing this seemed rude.
Let me know what you think when you're back @steveoc64.

andydotxyz:
Further to my previous comment this is getting a longer list of conflicts. If we still want to take this path later it could be re-opened or cherry picked from. Closing to better focus our PR backlog.

PR #969, opened 2020-05-10
ignore vscode settings

steveoc64:
### Description:
Add .vscode to the gitignore list

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
Let's not say that something is tested or has new public APIs when that's not the case ;)

PR #970, opened 2020-05-11
WIP - Full material design 

### LINKED ISSUE #958: Subjective comment on default colour choices in provided Fyne themes

steveoc64:
### Description:
WIP - preview of upgrading the internals to support the full Material Design color palette.

Support for extending the current theme by applying JSON patches on top - sort of like CSS, but not CSS.

Can copy generated themes directly from here, and apply the generated JSON to load the theme. 
  https://cimdalli.github.io/mui-theme-generator/

Remarkably, the existing API isnt touched at all, and it doesnt grow the code base a great deal.

Needs a bit more polish  - but worthy as an RFC / experiment

Fixes #958 #211 

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


steveoc64:
Found some nasty bugs in the tabContainer as well - from a recent patch in the last month or so

andydotxyz:
Looks like tests are broken - I'd encourage running them before pushing any code even if it's WiP. They are there to remind us not to break API or functionality and might impact the viability of a proposed change.

andydotxyz:
As we progress the implementation of Theme2 for 2.0 we should check back here to make sure we don't miss anything.

andydotxyz:
Closing due to age and the number of conflicts against the new theme work.
Let's check back here as we roll out the updated theme to make sure we keep the work in mind.

PR #974, opened 2020-05-11
Include all Detail Widths in MinSize

stuartmscott:
### Description:
When used in a dialog where some accordion details were wider than others, the dialog size would change a lot as the user navigated the accordion's contents. With this change all details are included to calculate the min width, meaning the dialog will always be wide enough for the widest detail.

The height is only included if the detail is open so dialogs will still resize taller or shorter if the heights are different - a future PR could measure the max height of all details and include that in the MinSize however this would result in unused space if no details are open (or a short detail is open)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


stuartmscott:
Travis failure:
```
--- FAIL: TestEntry_Select (0.87s)
    --- FAIL: TestEntry_Select/replace_reverse (0.10s)
        util.go:60: 
            	Error Trace:	util.go:60
            	            				entry_test.go:1270
            	Error:      	Not equal: 
            	Test:       	TestEntry_Select/replace_reverse
            	Messages:   	Image did not match master. Actual image written to file:///home/travis/gopath/src/fyne.io/fyne/widget/testdata/failed/entry_selection_reverse_initial.png.
```

andydotxyz:
Seems like a good concept. Looking through the updated images however they all seem to be centered in the space. That didn't seem like part of the change or the correct behaviour - do you know what's happening there?

stuartmscott:
Yup, currently the accordion is stretched to the full width of the output image, this PR puts the accordion in a center layout to ensure the accordion was rendered at minsize so you can see the accordion is stretched to the max detail width

andydotxyz:
Oh, yes. I missed line 189 on the test file. Thanks

PR #975, opened 2020-05-11
ShowVirtualKeyboard on focused widgets

pymq:
### Description:
Current issue: hardcoded widgets assertion in driver prevent to show keyboard on extended Entry.
Solution: intruduce new interface for such usage. Naming suggestions are welcome!

The big question is: should this case be considered special at all? We can do the same like this:
```go
func (e *Entry) FocusGained() {
	if device, ok := fyne.CurrentDevice().(mobile.Device); ok {
		device.ShowVirtualKeyboard()
	}
}

func (e *Entry) FocusLost() {
	if device, ok := fyne.CurrentDevice().(mobile.Device); ok {
		device.HideVirtualKeyboard()
	}
}
```
Or even only in FocusGained, keyboard hiding can be left for driver.

This way has a lot more flexibility. What do you think?

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
Well caught. Maybe `Typeable`?
Whether this is the right way or not is interesting. 

The bigger picture question might be how do we transition from one to another? At the moment it looks a bit weird as one hides and another shows.
Is this also the right way to consider the /type/ of keyboard that should be shown? :)

andydotxyz:
I have thought more about this and although it's more challenging it seems like sending "hints" is more flexible than inserting the actualy hide/show requests.

Considering future requirements (and perhaps also fixing the current multiline issue where enter is needed vs single line "done") how about a different method signatture: `KeyboardType()` returning one of a (currently limited) enum of keyboard types (single line, multiline and password I guess?)...?

stuartmscott:
We already have a way to configure which mouse cursor is rendered (arrow, ibeam, resize), perhaps we can follow a similar pattern with keyboard. Whatever types we define will need to map to the OS, eg https://developer.android.com/reference/android/widget/TextView#attr_android:inputType

stuartmscott:
I like the idea of showing/hiding keyboard when the focus changed, but as has already been mentioned switching between two entries could cause the keyboard to hide and reshow unnecessarily. Adding an interface which devs can implement moves this decision into the driver so it can be clever and see you're switching from one input to another so keep the keyboard up. I wonder if this interface should combine "do you need a keyboard" and "what type of keyboard do you need"?

stuartmscott:
Something like:
```go
// Keyboard represents a standard fyne keyboard
type Keyboard int

const (
	// DefaultKeyboard is the default keyboard (primary key inserts newline)
	DefaultKeyboard Keyboard = iota
	// SingleLineKeyboard is the keyboard often used to indicate text entry of a single line (primary key submits)
	SingleLineKeyboard
	// NumberKeyboard is the keyboard often used to indicate numerical entry
	NumberKeyboard
	// EmailKeyboard is the keyboard often used to indicate email entry
	EmailKeyboard
	// PasswordKeyboard is the keyboard often used to indicate password entry
	PasswordKeyboard
)

// Keyboardable describes any CanvasObject that needs a keyboard
type Keyboardable interface {
	Keyboard() Keyboard
}
```

pymq:
I like it! Also, I will introduce new method to Device interface to show keyboard of specific type, do you have naming suggestions? (as I understand we do not break existing APIs, right?)
@stuartmscott, what do you mean by PasswordKeyboard? For android I see textPassword constant, but for [ios](https://developer.apple.com/documentation/uikit/uikeyboardtype) there is no a suitable one. What special keys does PasswordKeyboard have?
Also, I think it worth to add UrlKeyboard type.

andydotxyz:
Don’t get carried away :) each of those items needs the iOS and Android implementation in place too, so start the list short!

I’m pretty sure this would be a separate interface like you illustrated, so widgets can implement it, not devices?

pymq:
> I’m pretty sure this would be a separate interface like you illustrated, so widgets can implement it, not devices?

Yes, separate interface with Keyboard() Keyboard method like @stuartmscott suggested. I mean also to add to the existing Device interface the ability to specify what type of Keyboard to show.

```go
// Device describes functionality only available on mobile
type Device interface {
	ShowVirtualKeyboard() // Request that the mobile device show the touch screen keyboard (standard layout)
	HideVirtualKeyboard() // Request that the mobile device dismiss the touch screen keyboard
}
```

I'll try to prepare PR to mobile repo and update this tomorrow.

stuartmscott:
I think Android's textPassword is mainly a hint for the TextView to not show the characters, but since Entry already supports Password we can drop this one. It may also change behaviour such as not showing the popup above the recently pressed key, but I'm not sure.

I like URLKeyboard, so it has a ".com" button, but the only difference with EmailKeyboard is the "@" so I'd probably combine the two into one (Web?). I believe Android's textEmail will populate the suggestions with the email address(s) registered to the device.

I agree with Andy, we should start small - we can always add more later, but we can't easily remove. The core use cases here are text(singleline, multiline), number, and web(email,URL). We can leave things like autosuggest for now (may require binding a callback or slice), and wait until we have data validation for the date/time keyboard.

andydotxyz:
> I mean also to add to the existing Device interface the ability to specify what type of Keyboard to show.

Of course, that makes sense - thanks for working on this.

andydotxyz:
> not show the characters

I think it is also used to now use autocompletion and suggestions ;). We don't have them turned on right now so it's not noticable, but in the future who knows.

I would be hesitant to add number and web just now, it would rrequire extending these widgets to specify the keyboard type and in the future we may have a better solution (such as tying this to validation). It's not a blocker, just a consideration. If we start small it's really just single vs multiline and possibly password (though I see that iOS doesnt expose password keyboard type, I think in this instance it's more of a hint to our driver than to the OS underneath).

driver/mobile/device.go
```diff
@@ -2,6 +2,10 @@ package mobile
 
 // Device describes functionality only available on mobile
 type Device interface {
-	ShowVirtualKeyboard() // Request that the mobile device show the touch screen keyboard (standard layout)
-	HideVirtualKeyboard() // Request that the mobile device dismiss the touch screen keyboard
+	// Request that the mobile device show the touch screen keyboard (standard layout)
+	ShowVirtualKeyboard()
+	// Request that the mobile device show the touch screen keyboard (custom layout)
+	ShowVirtualKeyboardType(Keyboard)
```
pymq, line 8: Do we need new method or we can break compatibility? Seems pretty minor for me. Also can't think of a better name than what's already in use.

cmd/fyne/internal/mobile/gendex/gendex.go
```diff
@@ -14,7 +14,7 @@
 //
 // Requires ANDROID_HOME be set to the path of the Android SDK, and
 // javac must be on the PATH.
-package mobile
+package main
```
pymq, line 17: Have to move into different directory and to `main` package to use with go generate (possibly because of go modules?)

pymq:
Added initial support for different keyboard types. Added seamless transition from one `Keyboardable` widget to another. I hope someone can test iOS build.
I will write tests for focus changes later.

driver/mobile/device.go
andydotxyz: You're right we need a new method here.
If this is deemed ugly we could change in 2.0 work.

cmd/fyne/internal/mobile/gendex/gendex.go
andydotxyz: cunning. I was always just changing the package name manually ;)

driver/mobile/keyboard.go
```diff
@@ -0,0 +1,26 @@
+package mobile
+
+import (
+	"fyne.io/fyne"
+)
+
+// Keyboard represents a standard fyne keyboard
```
andydotxyz, line 7: This doc line could probably be more descriptive. I also wonder if the type name should be a little clearer?

driver/mobile/keyboard.go
```diff
@@ -0,0 +1,26 @@
...
+
+const (
+	// DefaultKeyboard is the keyboard with default input style and "Done" return key
+	DefaultKeyboard Keyboard = iota
+	// MultiLineKeyboard is the keyboard with default input style and "return" return key
+	MultiLineKeyboard
+	// NumberKeyboard is the keyboard with number input style and "Done" return key
+	NumberKeyboard
+	// WebKeyboard is the keyboard with web input style and "Done" return key
+	WebKeyboard
```
andydotxyz, line 18: Is web keyboard for URLs or emails? this feels a little ambiguous.

driver/mobile/keyboard.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+// Keyboardable describes any CanvasObject that needs a keyboard
+type Keyboardable interface {
+	fyne.Focusable
+
+	Keyboard() Keyboard
```
andydotxyz, line 25: The duplication of naming here implies we need to think better `Keyboard` is a shorthand for `GetKeyboard` but then the return type doesn't help understand the context, what is a `Keyboard` - a layout of buttons?

driver/mobile/keyboard.go
```diff
@@ -0,0 +1,26 @@
...
+	"fyne.io/fyne"
+)
+
+// Keyboard represents a standard fyne keyboard
+type Keyboard int32
+
+const (
+	// DefaultKeyboard is the keyboard with default input style and "Done" return key
+	DefaultKeyboard Keyboard = iota
+	// MultiLineKeyboard is the keyboard with default input style and "return" return key
```
andydotxyz, line 13: I wonder if multiline should be the default? Just a question for discussion I guess.
In the current implementation we show the carriage return key I think.

andydotxyz:
Great work thanks, a few comments inline.
Also I think that Travis is finding compile errors...

pymq:
Oops, I thought I pushed vendor replace. Now CI should be ok.

andydotxyz:
Thanks, looks like just module tests are failing now :)

cmd/fyne/internal/mobile/gendex/gendex.go
pymq: Yeah, but a couple of times later I decided to fix it :)

driver/mobile/device.go
pymq: Not sure. We can introduce `ShowVirtualKeyboardType` or something in upcoming release and remove in 2.0. But if nobody needs it, we can wait until 2.0.

driver/mobile/keyboard.go
pymq: KeyboardLayout maybe?

driver/mobile/keyboard.go
pymq: In code I chose hints for URL. But in manual tests with two keyboards they doesn't change at all. Maybe I did something wrong. Or at least on iOS it will be different.

pymq: We could remove `WebKeyboard` as I think the most important types for now is one/multi line and number keyboards.

driver/mobile/keyboard.go
pymq: I think one line entries are more common. Also, before this changes "Done" return key was default.

driver/mobile/keyboard.go
pymq: I agree, it doesn't look good.
I also had thoughts about more flexible api, something like that:
```go
const (
	// DefaultKeyboard is the keyboard with default input style and "Done" return key
	DefaultKeyboard Keyboard = 1 << iota
	// MultiLineKeyboard is the keyboard modifier which replaces return key with "return"
	MultiLineKeyboard
	// NumberKeyboard is the keyboard modifier which replaces input style to numbers only
	NumberKeyboard
	// PhoneKeyboard is the keyboard modifier which replaces input style to phone keys only
	PhoneKeyboard
)
```
And someone can use it like this: `ShowVirtualKeyboard(NumberKeyboard | MultiLineKeyboard)` or `ShowVirtualKeyboard(DefaultKeyboard)`. It is more complicated to use and implement but has its advantages.

driver/mobile/keyboard.go
andydotxyz: Only on some phones. Older Android and iOS phones are all on return keys by default (that was related to the previous keyboard bug)

andydotxyz: The struggle I have with the single line "Done" default is that you don't know what the right action is. "Done", "Go", "Search" etc are all possibilities and I'd rather not pick the wrong one if we can avoid it.

driver/mobile/keyboard.go
andydotxyz: I don't think that building up flags like this makese sense. I can't think what a "multiline number" keyboard is as the number keys don't have a carriage return.
Perhaps you can expand on what the advantages are that you hint at?

driver/mobile/keyboard.go
andydotxyz: Remember that when designing Fyne APIs we need to consider the semantic meaning here - if an OS chooses to display the same keyboard for URL and for Email is fine, but we know they are different and so present that in the API.

driver/mobile/keyboard.go
andydotxyz: I think that would work. Though I do wonder if this is truly a layout - people might associate that with QWERTY/AZERTY ;)

andydotxyz:
Did you think more about the naming? we are entering feature freeze here so it's decision time for the outstanding PRs

andydotxyz:
This is a great addition, thanks

PR #979, opened 2020-05-12
Fixes/race3

andydotxyz:
This is built on top of races2 because some of this is risky and should not go into 1.3 for the following reasons:

* Refresh calling Layout (or the other way round) will deadlock
* The render locks are in basewidget but they would be better in the cache management layer (bigger change)

Maybe this is a place to discuss render refactoring like removing the separate Layout etc...


andydotxyz:
I'm closing this PR - I will restart it properly when we look at renderers again

PR #980, opened 2020-05-13
Feature/storage

andydotxyz:
### Description:
Before releasing the dialog and other new file APIs they need to be tidied a little.
Having the constructors in the root was no good. And having uri as `string` was creating a little confusion (i.e. lots of os.Open(uri[7:]) which is unsafe for external code.

This leaves one hole which is the mimeType which I figure will be filled in by #405.
The URI class will be able to hand caching and other clever things to make peoples lives easier and apps quicker :).

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
You can see that Travis is green, seems the data coming through is slow today

file.go
```diff
@@ -7,25 +7,22 @@ import "io"
...
-// This is mostly provided so that file references can be saved using their URI and written to again later.
-func SaveFileToURI(uri string) (FileWriteCloser, error) {
-	return CurrentApp().Driver().FileWriterForURI(uri)
+// URI represents the identifier of a resource on a target system.
+// This resource may be a file or another data source such as an app or file sharing system.
+type URI interface {
+	Extension() string
+	MimeType() string
+	Scheme() string
+	String() string
```
stuartmscott, line 27: nit: embed Stringer

internal/driver/glfw/file.go
```diff
@@ -6,48 +6,49 @@ import (
...
-func (f *fileWriter) URI() string {
-	return "file://" + f.path
+func (f *fileWriter) URI() fyne.URI {
+	return storage.NewURI("file://" + f.path)
 }
 
-func openFile(uri string, create bool) (*file, error) {
-	if len(uri) < 8 || uri[:7] != "file://" {
+func openFile(uri fyne.URI, create bool) (*file, error) {
+	if uri.Scheme() != "file" {
```
stuartmscott, line 47: much neater!

dialog/file.go
```diff
@@ -65,7 +66,7 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 			info, err := os.Stat(path)
 			if os.IsNotExist(err) {
 				f.win.Hide()
-				callback(fyne.SaveFileToURI("file://" + path))
+				callback(storage.SaveFileToURI(storage.NewURI("file://" + path)))
```
stuartmscott, line 69: it might be worth adding storage.NewFileURI(path)

internal/driver/glfw/file.go
```diff
@@ -6,48 +6,49 @@ import (
...
 		return nil, fmt.Errorf("invalid URI for file: %s", uri)
 	}
 
-	path := uri[7:]
+	path := uri.String()[7:]
```
stuartmscott, line 51: I agree adding "func Path() string" to the interface would be confusing, but perhaps we can come up with something else to avoid string slicing all together?

andydotxyz: I'm not really concerned as this is just in driver code, but we should revisit when support for remote files is added later

dialog/file.go
andydotxyz: Maybe, can we leave it out until we see if anyone needs that? It would only be useful externally if people are constructinhg URIs to feed in - which is not a primary path ATM.

file.go
andydotxyz: OK

toaster:
LGTM but I think the string helper methods are superfluous.

PR #987, opened 2020-05-15
Fix scrolling of key names so they expand horizontally

andydotxyz:
Advanced screen would scroll keys horizontally and vertically making it hard to read.
This makes it so they only scroll vertically and the grid expands horizontally to fit content.

### DIFF
--- a/widget/group.go
+++ b/widget/group.go
@@ -64,7 +64,7 @@ func NewGroup(title string, children ...fyne.CanvasObject) *Group {
 // This group will scroll when the available space is less than needed to display the items it contains.
 func NewGroupWithScroller(title string, children ...fyne.CanvasObject) *Group {
 	box := NewVBox(children...)
-	group := &Group{BaseWidget{}, title, box, NewScrollContainer(box)}
+	group := &Group{BaseWidget{}, title, box, NewVScrollContainer(box)}
 
 	Renderer(group).Layout(group.MinSize())
 	return group
--- a/widget/group_test.go
+++ b/widget/group_test.go
@@ -42,3 +42,15 @@ func TestGroup_Text(t *testing.T) {
 
 	assert.Equal(t, "World", test.WidgetRenderer(group).(*groupRenderer).label.Text)
 }
+
+func TestGroup_Scroll(t *testing.T) {
+	group := NewGroupWithScroller("Test",
+		NewLabel("A label"), NewLabel("Another item"))
+	group.Resize(group.MinSize())
+
+	scroll := group.content.(*ScrollContainer)
+	assert.Equal(t, scroll.Direction, ScrollVerticalOnly)
+	assert.Equal(t, group.MinSize().Width, scroll.Content.Size().Width)
+	totalHeight := test.WidgetRenderer(group).(*groupRenderer).label.MinSize().Height + scroll.Content.Size().Height
+	assert.Less(t, group.MinSize().Height, totalHeight)
+}


cmd/fyne_demo/screens/advanced.go
```diff
@@ -56,6 +55,11 @@ func AdvancedScreen(win fyne.Window) fyne.CanvasObject {
 		})
 	}
 
+	label := widget.NewLabel("Just type...")
+	head := widget.NewVBox(label,
+		fyne.NewContainerWithLayout(layout.NewGridLayout(2),
+			widget.NewGroup("Generic"),
```
stuartmscott, line 61: This doesn't take advantage of a group, unless group is used for the line, could this be label instead?

cmd/fyne_demo/screens/advanced.go
```diff
@@ -56,6 +55,11 @@ func AdvancedScreen(win fyne.Window) fyne.CanvasObject {
 		})
 	}
 
+	label := widget.NewLabel("Just type...")
```
stuartmscott, line 58: nit: could be inlined

cmd/fyne_demo/screens/advanced.go
andydotxyz: The group's scrolling is generic unfortunately and it is what was collapsing before.
We could use labels - but I liked the styling :)

cmd/fyne_demo/screens/advanced.go
andydotxyz: k

andydotxyz:
Apologies, I decided the previous approach was daft - this is really fixing a behaviour break so I force pushed a completely different solution that fixes this for everyone.

cmd/fyne_demo/screens/advanced.go
andydotxyz: This is not a problem with my new patch :)

widget/group.go
```diff
@@ -64,7 +64,7 @@ func NewGroup(title string, children ...fyne.CanvasObject) *Group {
 // This group will scroll when the available space is less than needed to display the items it contains.
 func NewGroupWithScroller(title string, children ...fyne.CanvasObject) *Group {
 	box := NewVBox(children...)
-	group := &Group{BaseWidget{}, title, box, NewScrollContainer(box)}
+	group := &Group{BaseWidget{}, title, box, NewVScrollContainer(box)}
```
stuartmscott, line 67: It may be excessive but I think I'd prefer leaving NewGroupWithScroller as is, but add NewGroupWithHScroller and NewGroupWithVScroller. All three of which would call "newGroupWithChild" passing a scroll container holding the content.

andydotxyz: Hmm, not sure:
A) how useful is a vertically scrolling horizontal list
B) surely we should be correcting the accidentally changed behaviour?

stuartmscott: I don't see anything about Group only being vertical. It could be a HBox in a HScroll, or a VBox in a VScroll. So maybe it is just two APIs not three: NewVGroupWithVScroll or NewHGroupWithHScroll

andydotxyz: All groups use a vertical layout - see https://github.com/fyne-io/fyne/blob/3e1864e1922a21b28f98f1745b09b39dfea1e1b0/widget/group.go#L56

stuartmscott: Sure the implementation only has vertical, but nothing in the developer docs says this is the case. I don't see a downside to giving the dev a choice here

andydotxyz: Ah, OK. Sure I guess, but with new APIs, my point here was that I was returning it to it's previous functionality.

stuartmscott: > B) surely we should be correcting the accidentally changed behaviour?

> returning it to it's previous functionality

Maybe this is the misunderstanding. I don't know what it used to do and now doesn't. Was the previous functionality only scrolling vertically?

andydotxyz: Yes. All scrollers were previously vertical only. Horizontal scroll was added fairly recently and this widget should have been updated

PR #999, opened 2020-05-17
[fyne cli] Fix fyne package for android won't build with empty go cache

lucor:
### Description:
With Cgo enabled we need to ensure the C compiler is set via CC to avoid the error: 
`gcc: error: unrecognized command line option '-marm'`

To replicate:
- checkout fyne-io/fyne@develop (commit 11be77)
- go build ./cmd/fyne
- go clean -cache -modcache
- cd cmd/fyne_demo
- ../../fyne package -os android -icon ../../theme/icons/fyne.png -appID test

Got:
```
go [-e -json -compiled=true -test=false -export=false -deps=false -find=true -- .]: exit status 2: go: downloading github.com/fyne-io/mobile v0.0.2-0.20200511203347-825e44fe5a95
go: downloading golang.org/x/image v0.0.0-20200119044424-58c23975cae1
go: downloading github.com/srwiley/oksvg v0.0.0-20200311192757-870daf9aa564
go: downloading github.com/srwiley/rasterx v0.0.0-20200120212402-85cb7272f5e9
go: downloading github.com/goki/freetype v0.0.0-20181231101311-fa8a33aabaff
go: downloading golang.org/x/net v0.0.0-20200324143707-d3edc9973b7e
go: downloading golang.org/x/text v0.3.2
# runtime/cgo
gcc: error: unrecognized command line option '-marm'; did you mean '-mabm'?
```

Expect:
To package the android app

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


cmd/fyne/internal/mobile/bind.go
```diff
@@ -69,6 +69,11 @@ func packagesConfig(targetOS string) *packages.Config {
 	config := &packages.Config{}
 	// Add CGO_ENABLED=1 explicitly since Cgo is disabled when GOOS is different from host OS.
 	config.Env = append(os.Environ(), "GOARCH=arm", "GOOS="+targetOS, "CGO_ENABLED=1")
+	if targetOS == "android" {
+		// with Cgo enabled we need to ensure the C compiler is set via CC to
+		// avoid the error: "gcc: error: unrecognized command line option '-marm'"
+		config.Env = append(os.Environ(), androidEnv["arm"]...)
```
stuartmscott, line 75: Do we need the env additions from line 71?
```suggestion
		config.Env = append(config.Env, androidEnv["arm"]...)
```

andydotxyz, line 75: What if this is an android build for something other than arm?

andydotxyz:
Are you sure this is the right fix? this code is for looking up packages not the actual build commands. Executions of the `goBuild` function already get the android arch env passed in...

cmd/fyne/internal/mobile/bind.go
lucor: No, it is not needed. `anroidEnv` contains already all env settings from line 71: https://github.com/fyne-io/fyne/blob/master/cmd/fyne/internal/mobile/env.go#L116-L122
Adding will duplicate them.

lucor: Looking at the code, this is used only to create the config for `packages.Load` and previously was already hardcoded to `arm`.

lucor:
Right but it seems that the `packages.Load` used for looking up packages will build them if not already compliled / found in cache and this will produce the error above.

PR #998, opened 2020-05-17
Fix #828: tab container update handling

toaster:
### Description:

This PR refactors the tab container widget’s update handling and fixes issues when adding/removing items to an existing tab container.
It also removes the tab container widget’s dependency from the renderer cache.
As a side-effect it introduces `TabContainer.SetItems` which might be used to replace the tab items completely.

Fixes #828

### Checklist:

- [x] Tests included.
- [x] Lint and formatter run with no errors.
- [x] Tests all pass.

#### Where applicable:

- [x] Public APIs match existing style.

PR #1001, opened 2020-05-18
Wrap dialog labels

stuartmscott:
### Description:
Wraps dialog labels to avoid them being too wide

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
How are you testing this? It does not seem to do much at this end...
Screenshot 2020-05-19 at 12 06 03

stuartmscott:
Oh wow, Andy is your screenshot showing dialog not measuring children for minsize?!

As you can see from the golden images there is still a lot of work to do here;
- Should dialog titles wrap, or truncate?
- Should dialog buttons wrap, or truncate?
- Should dialog content that is too big be put inside a scroller?
- What is the max bounds of a dialog? The entire window, or smaller to ensure window content is visible around the dialog.

andydotxyz:
From the golden images it looks like the dialog is incorrectly being resized to the full window on the test driver. If that issue is resolved and they go back to minsize you'll probably see the same as in the screenshot I posted.

Dialogs are designed to show the content at minsize. They can be expanded programatically, but when you expand the window that does not happen automatically.
I think this change is much bigger than a bug fix or changing default text layout.

andydotxyz:
> Oh wow, Andy is your screenshot showing dialog not measuring children for minsize?!

That is with your patch applied. Without it the dialog will expand to fir the content, but (at least I think and from what I can see) with wrapping turned on the minsize shrinks, and so does the dialog?

I was just curious about how it looks on your machine... maybe I based it off the wrong branch?

stuartmscott:
It appears setting textwrap affects the min size calculation, but when the label is laid out the maxWidth is set to the window's width, not the dialog width.

stuartmscott:
Here's what I get locally

Screen Shot 2020-05-19 at 1 41 33 PM

stuartmscott:
Ooops I was on the wrong branch, actually this is what I get locally:

Screen Shot 2020-05-19 at 1 51 37 PM

stuartmscott:
After merging #998 some test images changed

stuartmscott:
I think the issue is that Label uses textProvider which is itself a BaseWidget. When the updateRowBounds is called the textProvider size is unset, so it does not wrap. Label either needs to resize textProvider when it is resized, or textProvider shouldn't be a BaseWidget. Thoughts?

andydotxyz:
Hmm, tricky. In the longrun I think you're right TextProvider should not be a BaseWidget - on reflection that looks like it was a bad choice - it is possibly better presented as a renderer or mixin - probably related to a refresh of how renderers are handled in 2.0.

In the meantime I gues that a label should ensure that it's textProvider remains the same size, yes. It's curious that truncating works but wrapping does not - do they walk different paths with regards to their bounds?

andydotxyz:
Apologies, I see that this requested reviews a while back - but I think it requires more work as shown in the conversation right?

stuartmscott:
Yes, more work is needed here. My bad, I thought marking it as draft would cancel the review requests. I'll take it out of draft when it is ready

andydotxyz:
It seems like this has been causing problems recently for some developers. If we wrap and then shrink as we normally do it will likely wrap to as many lines as possible which is likely not what we want.
This is a complex topic it turns out, and I think we should bring it near the top of the list if possible.
Perhaps we need special wrapping in `Dialog`... and if we do that is that in `Dialog` or in `Text`?

AlbinoGeek:
I suppose this didn't make 2.0 -- while I hit this issue repeatedly in #1570 it turned out the answer for me was "Stop using Dialog." As in, if you are writing extremely long strings into a Dialog, then you are doing something wrong. Exception: `dialog.ShowError()` and `dialog.ShowMessage()` -- things I would 100% expect to wrap and also become as tall as necessary to fit the text.

Outside of this however, I would not expect other Dialogs to automagically wrap strings I pass to them.

Also, using code such as the following, it is possible to get working label wrapping (but requires workarounds):

```go
	user := widget.NewEntry()
	pass := widget.NewPasswordEntry()
	
	warning := widget.NewLabel(loginWarning) // very long text
	warning.Wrapping = fyne.TextWrapWord
	
	vbox := 
	
	dlg := dialog.NewCustomConfirm(
		"Login to sc2replaystats","Login", "Cancel",
		widget.NewVBox(
			warning,
			layout.NewSpacer(),
			widget.NewForm(
				widget.NewFormItem("Email", user),
				widget.NewFormItem("Password", pass),
			),
			layout.NewSpacer(),
		),
		func(ok bool) {
			if ok {
				// ...
			}
		},
		w)

	dlg.Show()
	vbox.Resize(fyne.NewSize(999, 280)) // workaround #1565 by setting height to 999 (larger than window)
	dlg.Resize(fyne.NewSize(420, 280))  // workaround #1570 by resizing Dialog after container
```

---

Also, I think checks need to be re-run on a number of PRs/issues (they are stuck at "Waiting for status to be reported."

andydotxyz:
> Also, I think checks need to be re-run on a number of PRs/issues (they are stuck at "Waiting for status to be reported."

These PRs pre-date an infrastructure move (read the Travis OSS story for more) and so they need to be re-based onto latest develop for the checks to run.

PR #1002, opened 2020-05-18
Ensure renderers are up to date

### LINKED ISSUE #834: Layout called before Refresh
In some cases a widget can be laid out before refresh is called, resulting in incorrect renders. This can happen if a widget is resized, as Resize(size) calls Layout(size) in BaseWidget.

After discussion with the community the consensus is to expect CreateRenderer to return a renderer that is ready to go.

Affected Widgets:
- Check
- Icon
- Radio

stuartmscott:
### Description:
Ensure renderers returned by CreateRenderer are up to date and ready to render.

Fixes #834

### Checklist:

- [ ] Tests included.
- [y] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/check.go
+++ b/widget/check.go
@@ -53,8 +53,17 @@ func (c *checkRenderer) applyTheme() {
 
 func (c *checkRenderer) Refresh() {
 	c.applyTheme()
-	c.label.Text = c.check.Text
+	c.updateLabel()
+	c.updateResource()
+	c.updateFocusIndicator()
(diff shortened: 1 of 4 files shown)


andydotxyz:
fyi "x" is checked and " " is not, i.e.

- [x] checked with x
- [ ] unchecked with space.

andydotxyz:
Would the image testing framework not be ideal to show that this previously did not work but now does?

stuartmscott:
Moved to #1011

PR #1013, opened 2020-05-20
Fix mobile rendering 

### LINKED ISSUE #950: iOS app stops re-drawing mid frame after a while
### Describe the bug:
Run a Fyne app on an iOS device and it works well for a minute or two.
Then it freezed mid-frame (bottom half of the screen is bgcolour) and does not respond.
Rotating will turn the screen black. Force quit required.

### To Reproduce:
1. Load fyne_demo on iOS device
2. Tap around and cause lots of redraws (theme setting for example).
3. See drawing eventually stops
4. Window is only partially painted and window stops responding.

### Screenshots:
8DC1ABCE-AE15-4559-838D-B97576484091

 - OS: iOS
 - Version: 13.3.1
 - Go version: 1.12.4
 - Fyne version: develop

I think this happened on 1.2.x as well

andydotxyz:
### Description:
Attempts to pin this down:
* Remove shared buffer, it's possible this was causing issues

Fixes #950 

### Checklist:

- [ ] Tests included. <- no clue how!
- [x] Tests all pass.


andydotxyz:
This is slow progress.
I don't know if it's the shared buffer removal (here) or the debugging I added (not committed) but one way or another I am struggling to reproduce this now, would appreciate help!

andydotxyz:
With logging in place I see this some times:

default	11:57:45.649780+0100	main	GoLog: 2020/05/25 10:57:45 Error 501 in GL Renderer
default	11:57:45.650541+0100	main	GoLog: 2020/05/25 10:57:45   At: /Users/andy/Go/src/fyne.io/fyne/internal/painter/gl/gl_gomobile.go:74

But I have not seen a crash since adding it. 

That is: `TexImage2D` erroring with GL_INVALID_VALUE...

andydotxyz:
OK, now I have the error occur again. The bad news is ... no long messages at all.
This backs up the worry I had before that a GL method is simply not returning :(

andydotxyz:
The latest version of the mobile dep fixes this issue by having an increased buffer before it will block (from 3 to 12). I have not been able to replicate since.

andydotxyz:
Basically because we are 1 step further up the stack the line it prints would always be the error printing line. So I pretty much replicated the functionality of fyne.LogError but with the correct stack line printed.

okratitan:
> Basically because we are 1 step further up the stack the line it prints would always be the error printing line. So I pretty much replicated the functionality of fyne.LogError but with the correct stack line printed.

Good enough for me - In the future if the stack error printing is needed elsewhere we can consider centralizing it.

PR #1014, opened 2020-05-20
Correct and expand locking from races2

### LINKED ISSUE #506: Massive data race detect!!

andydotxyz:
### Description:
Expunge more races

Fixes #506

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


internal/driver/glfw/canvas.go
```diff
@@ -162,29 +164,45 @@ func (c *glCanvas) Refresh(obj fyne.CanvasObject) {
 }
 
 func (c *glCanvas) Focus(obj fyne.Focusable) {
-	if c.focused != nil {
-		c.focused.FocusLost()
+	c.RLock()
```
stuartmscott, line 167: Could this be simplified to a single lock/unlock?

```go
c.Lock()
prev := c.focused
c.focused = obj
c.Unlock()

if prev != nil {
    prev.FocusLost()
}
if obj != nil {
    obj.FocusGained()
}
```
Ditto for Unfocus

andydotxyz: It could, but somehow it felt "right" to notify FocusLost before the new focus item was set... Maybe I was overly cautious?

stuartmscott: It is currently two lines longer than the suggestion, and besides that I don't think there is anything to be gained so let's keep it as is

PR #1016, opened 2020-05-21
Fix Vertical Alignment of Software Rendered Text

stuartmscott:
### Description:
Software renderer was drawing text starting from top left. This change vertically aligns the text in the center resulting in similar behaviour to the GL renderer.

There is still an issue here where it appears the text is rendered at Position(theme.Padding(), theme.Padding()), which makes the text look too low and too much to the right (see *_small.png test images).

Fixes #none (see #1011 for context)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
Nice catch. Regarding the extra padding I'm not sure. The first thing I noticed was that the opts.DPI is hard coded and there is now a constant, but they are the same so it should not cause the error:

```
draw.go:83
	opts.DPI = painter.TextDPI
```

The other area I would consider looking at is `testdriver.go` does not use the `painter.RenderedTextSize` helper. Possibly this was due to import loops which may be why the above constant is not used either, but it's possible there are subtle differences in their that could impact the size calculations.

andydotxyz:
Random travis failure in entry_tapped_initial, restarted just in case

stuartmscott:
The latest commit has some text image deltas, and breaks TextGrid:
```
--- FAIL: TestTextGrid_CreateRendererRows (0.00s)
    textgrid_test.go:104: 
        	Error Trace:	textgrid_test.go:104
        	Error:      	Not equal: 
        	            	expected: 8
        	            	actual  : 10
        	Test:       	TestTextGrid_CreateRendererRows
--- FAIL: TestTextGridRender_Size (0.00s)
    textgrid_test.go:141: 
        	Error Trace:	textgrid_test.go:141
        	Error:      	Not equal: 
        	            	expected: 2
        	            	actual  : 3
        	Test:       	TestTextGridRender_Size
```

stuartmscott:
Latest commit adjusts TextGrid tests as text is no longer rendered bold

andydotxyz:
Thanks for getting this resolved - it's good to know our painters are now consistent :)

PR #1017, opened 2020-05-21
Tidy up some resize and Gl flicker on macOS

### LINKED ISSUE #994: Tearing, layout issues on resize
Having moved to render thread the layout-on-resize is not very smooth.
THis culminates with red flickering on iMac Catalina machines.

andydotxyz:
This improves overall speed and cleans up some resize code.
Strange issue with test meant code had to be reordered, does not affect real apps.

There is an underlying issue that can _some times_ be triggered if you refresh very, very fast to large window sizes on macOS Catalina. This seems to be down to the GLFW implementation wrapping vulkan and we may need to hook in some custom "swapchain" code in place of the swapping buffers. Needless to say that is a lot of work and in the meantime they may fix the issue, which is now very hard to replicate.

Fixes #994

### Checklist:

- [ ] Tests included. <- nah sorry, just load fyne_demo and resize lots, it's better
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -78,7 +78,7 @@ type window struct {
 
 	xpos, ypos    int
 	width, height int
-	ignoreResize  bool
+	shouldExpand  bool
 
 	eventLock  sync.RWMutex
 	eventQueue chan func()
@@ -160,13 +160,13 @@ func (w *window) RequestFocus() {
 func (w *window) Resize(size fyne.Size) {
 	w.canvas.Resize(size)
 	w.viewLock.Lock()
-	w.width, w.height = internal.ScaleInt(w.canvas, size.Width), internal.ScaleInt(w.canvas, size.Height)
+	if w.fixedSize {
+		w.width, w.height = internal.ScaleInt(w.canvas, size.Width), internal.ScaleInt(w.canvas, size.Height)
+	}
 	w.viewLock.Unlock()
 
 	w.runOnMainWhenCreated(func() {
-		w.ignoreResize = true
 		w.viewport.SetSize(w.width, w.height)
-		w.ignoreResize = false
 		w.fitContent()
 	})
 }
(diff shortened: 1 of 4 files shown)


internal/driver/glfw/window.go
```diff
@@ -78,7 +78,7 @@ type window struct {
 
 	xpos, ypos    int
 	width, height int
-	ignoreResize  bool
+	shouldExpand  bool
```
stuartmscott, line 81: Is there any possibility that a window shouldContract?

internal/driver/glfw/window.go
```diff
@@ -160,13 +160,13 @@ func (w *window) RequestFocus() {
 func (w *window) Resize(size fyne.Size) {
 	w.canvas.Resize(size)
 	w.viewLock.Lock()
-	w.width, w.height = internal.ScaleInt(w.canvas, size.Width), internal.ScaleInt(w.canvas, size.Height)
+	if w.fixedSize {
```
stuartmscott, line 163: If fixedSize, change window size. This seems wrong, but maybe I'm misunderstanding what fixedSize means?

internal/driver/glfw/window.go
andydotxyz: No change in content will cause the window to contract, we just don't do that.
A window can be contracted in response to moving to a screen with lower DPI, but that is handled in the main window thread.

internal/driver/glfw/window.go
andydotxyz: fixedSize will ignore "user generated" window sizes, so here we save the "correct size" as future events may be incorrect.

I.e. set a fixed size - then the WM may show window small before getting our expand signal - so we have to save the state explicitly.

stuartmscott: okay so fixedSize means not user resizable, but it can still be resized programmatically?

andydotxyz: Yes. it has to be, otherwise you are forcing fixed size at minsize.
You would do (for example) w.Resize(); w.SetFixedSize(true) (but in either order).

PR #1020, opened 2020-05-22
Fix visible and expand locking

### LINKED ISSUE #506: Massive data race detect!!

andydotxyz:
The remaining races that can be found all have xxxRenderer.Refresh() in them, which we will resolve in 2.0

Fixes #506

### Checklist:

- [x] Tests included. <- go test or go run fyne_demo with `-race`
- [x] Tests all pass.


internal/driver/glfw/window.go
```diff
@@ -398,8 +402,11 @@ func (w *window) Content() fyne.CanvasObject {
 }
 
 func (w *window) SetContent(content fyne.CanvasObject) {
+	w.viewLock.RLock()
+	visible := w.visible
+	w.viewLock.RUnlock()
 	// hide old canvas element
-	if w.visible && w.canvas.Content() != nil {
+	if visible && w.canvas.Content() != nil {
```
stuartmscott, line 409: should w.canvas.Content also be captured in a local variable?

andydotxyz: Canvas does it's own locking

stuartmscott: Ok so Content() should be safe, perhaps capture the canvas reference then?

stuartmscott: Also line 354 accesses the canvas content without going through canvas' locking

internal/driver/glfw/loop.go
```diff
@@ -120,8 +120,14 @@ func (d *gLDriver) runGL() {
 					continue
 				}
 
-				if w.shouldExpand {
+				w.viewLock.RLock()
```
stuartmscott, line 123: What if we captured the viewport once at the start of the loop?
```go
w.viewLock.Lock()
v := w.viewport
w.viewLock.Unlock()

if v == nil {
	continue
}
if v.ShouldClose() {
	//...
}
//...
```

andydotxyz: That means gaining a write lock 60 * windowCount times every second instead of only when windows are closing

andydotxyz:
Fixed the canvas.Content() access

PR #1033, opened 2020-05-24
Draw canvas.Rectangle border when StrokeColor/Width is set

lusingander:
### Description:

The `StrokeColor` and `StrokeWidth` settings of `canvas.Rectangle` do not seem to be enabled.

For example, in `cmd/fyne_demo/screens/graphics.go`, the code is written as follows:

```go
&canvas.Rectangle{FillColor: color.NRGBA{0x80, 0, 0, 0xff},
	StrokeColor: color.NRGBA{0xff, 0xff, 0xff, 0xff},
	StrokeWidth: 1},
```

And the display looks like this:
(Look at the rectangle in the top left corner)

However, the expected display would be like this:

It seemed that the drawing functions do not take these into account, so I fixed these.

- `buf`(L100) is used to ensure that the border does not protrude from the rectangle.
- The reason for drawing one side at a time is to avoid rounding the corners...

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
Thanks for doing this.
Are you sure that rounded corners are wrong? :)
Regarding the `buf` stroke - Line and Circle centre the stroke on the coordinate / edge - should rectangle not be the same?

internal/painter/gl/gl_common.go
```diff
@@ -81,16 +81,55 @@ func (p *glPainter) newGlLineTexture(obj fyne.CanvasObject) Texture {
 }
 
 func (p *glPainter) newGlRectTexture(rect fyne.CanvasObject) Texture {
+	if rect, ok := rect.(*canvas.Rectangle); ok {
```
andydotxyz, line 84: Complexity here is increased a lot - it's basically a new function inline.
Hopw about: going back to the previous type check, and if it's got no stroke then using the previous code (much faster to render!). Then if there is a stroke calling out to a new method newGlStrokedRectTexture()?
I think that would be easier to read.

lusingander: Thanks for the review!
Is it okay like this?

```go
func (p *glPainter) newGlRectTexture(rect fyne.CanvasObject) Texture {
	col := theme.BackgroundColor()
	if wid, ok := rect.(fyne.Widget); ok {
		widCol := cache.Renderer(wid).BackgroundColor()
		if widCol != nil {
			col = widCol
		}
	} else if rect, ok := rect.(*canvas.Rectangle); ok {
		// add if block like this
		if rect.Stroke {
			return newGlStrokedRectTexture()
		}
		if rect.FillColor != nil {
			col = rect.FillColor
		}
	}

	return p.imgToTexture(image.NewUniform(col), canvas.ImageScaleSmooth)
}
```

andydotxyz: Yeah I think that looks quite neat :)

lusingander:
I thought it would be somehow better if the corners weren't rounded.
These images are enlarged with thicker lines, which is better?





(off course the protruding background needs to be fixed)

----

And if I don't consider `buf` in my code, we'll see the following.
(The thickness of the line is the same as the image above.)
Am I doing something wrong, like how to use functions?

andydotxyz:
I think that the thing the last image is missing would be the vector overflow that circle and line have enabled - that allows items to draw outside their "bounds" to allow stroke thickness etc. (search for `vectorPad` - this may need to be added when stroke is on)

lusingander:
Thanks for the confirming.
I may not understand it right yet, I think that it's like this:

```go
func (p *glPainter) newGlStrokedRectTexture(obj fyne.CanvasObject) Texture {
	rect := obj.(*canvas.Rectangle)

	width := p.textureScaleInt(rect.Size().Width + vectorPad*2)
	height := p.textureScaleInt(rect.Size().Height + vectorPad*2)
	stroke := rect.StrokeWidth * p.canvas.Scale() * p.texScale

	raw := image.NewRGBA(image.Rect(0, 0, width, height))
	scanner := rasterx.NewScannerGV(rect.Size().Width, rect.Size().Height, raw, raw.Bounds())

	scaledPad := p.textureScaleInt(vectorPad)
	p1x, p1y := scaledPad, scaledPad
	p2x, p2y := p.textureScaleInt(rect.Size().Width)+scaledPad, scaledPad
	p3x, p3y := p.textureScaleInt(rect.Size().Width)+scaledPad, p.textureScaleInt(rect.Size().Height)+scaledPad
	p4x, p4y := scaledPad, p.textureScaleInt(rect.Size().Height)+scaledPad

	if rect.FillColor != nil {
		filler := rasterx.NewFiller(width, height, scanner)
		filler.SetColor(rect.FillColor)
		rasterx.AddRect(float64(p1x), float64(p1y), float64(p3x), float64(p3y), 0, filler)
		filler.Draw()
	}

	if rect.StrokeColor != nil && rect.StrokeWidth > 0 {
		dasher := rasterx.NewDasher(width, height, scanner)
		dasher.SetColor(rect.StrokeColor)
		dasher.SetStroke(fixed.Int26_6(float64(stroke)*64), 0, nil, nil, nil, 0, nil, 0)
		dasher.Start(rasterx.ToFixedP(float64(p1x), float64(p1y)))
		dasher.Line(rasterx.ToFixedP(float64(p2x), float64(p2y)))
		dasher.Line(rasterx.ToFixedP(float64(p3x), float64(p3y)))
		dasher.Line(rasterx.ToFixedP(float64(p4x), float64(p4y)))
		dasher.Stop(true)
		dasher.Draw()
	}

	return p.imgToTexture(raw, canvas.ImageScaleSmooth)
}
```

However, there is a gap as shown in the image.



`drawRectangle` does not seem to call the function with `pad` as `drawCircle` and `drawLine` do.
I can see that this is related, but I don't know what to do.

```go
func (p *glPainter) drawRectangle(rect *canvas.Rectangle, pos fyne.Position, frame fyne.Size) {
	p.drawTextureWithDetails(rect, p.newGlRectTexture, pos, rect.Size(), frame, canvas.ImageFillStretch, 1.0, 0.0, 0)
}
```

andydotxyz:
I think you are close, but it’s inset by the vectorPad as that isn’t factored into the drawing. Instead of passing 0 as it currently does I imagine you would need to pass the vectorPad const instead of 0 if stroke is on..?

stuartmscott:
To the "rounded corners" vs "drawing each edge separately" discussion, should canvas.Rect support stroke line join and line cap?

andydotxyz:
> should canvas.Rect support stroke line join and line cap?

At some point we could add more complex drawing features, but as it's not been requested and we don't use it probably not much need yet. Would need to be added to line and circle too I would think.

lusingander:
Thanks for the advice, I fixed to draw `canvas.Rectangle` considering with `vectorPad`, and fixed to have rounded corners in one stroke.

It seems to work as expected, but the overlap is noticeable in `fyne_demo` app due to the color, position and shape of the figure and so on...

andydotxyz:
Thanks for this. I guess that outcome is to be expected of the way we calculate sizes at the moment. Perhaps swapping the rectangle with the fyne image would be a good idea so it doesn't look wrong in the demo? :)

lusingander:
I think it's a very good. It seems more natural for the Fyne icon to come first and simple shapes to follow.

It looks like the image below.

PR #1034, opened 2020-05-24
Add the ability to filter files in the File Dialog

### LINKED ISSUE #893: Support filtering of file dialog
Probably filter by mime type - this will need to be translated to other mobile specific approaches for iOS, Android.
Provide standard filterings, but maybe the ability to pass a function for more specific filtering...

okratitan:
### Description:
This adds the ability to filter files in Fyne's File Dialog

Fixes #893

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

#### Where applicable:


okratitan:
Still needs some tests - but that will likely involve me expanding the file system in dialog/testdata ... Seeing as how the freeze is closing in - I wanted to make sure we had the big parts in agreement.

dialog/fileicon.go
```diff
@@ -31,7 +31,7 @@ const ratioDown float64 = 0.45
 // NewFileIcon takes a filepath and creates an icon with an overlayed label using the detected mimetype and extension
 func NewFileIcon(path string) fyne.CanvasObject {
 
-	ext, mimeType, mimeSubType := mimeTypeGet(path)
+	ext, mimeType, mimeSubType := mimeTypeGet(storage.NewURI(path))
```
andydotxyz, line 34: I hadn't noticed that this API was public last time - given that we use URI everywhere else and we need it internally, this should probably have taken a URI parameter not a path string.

dialog/file.go
```diff
@@ -251,21 +297,33 @@ func showFileDialog(save bool, callback interface{}, parent fyne.Window) {
...
 	}
-	showFileDialog(false, callback, parent)
+	showFileDialog(false, callback, parent, filter)
 }
 
 // ShowFileSave shows a file dialog allowing the user to choose a file to save to (new or overwrite).
 // If the user chooses an existing file they will be asked if they are sure.
 // The dialog will appear over the window specified.
-func ShowFileSave(callback func(fyne.FileWriteCloser, error), parent fyne.Window) {
+func ShowFileSave(callback func(fyne.FileWriteCloser, error), parent fyne.Window, filter FileFilter) {
```
andydotxyz, line 324: Given that all usages of this are passing nil in our demos it seems more like an optional parameter.
Perhaps we should update to mirror the design pattern of ShowxxxDialog() being a shortcut for NewxxxDialog().Show() - then the filter could be set on a simple struct representing the dialog creation state?

dialog/fileicon.go
okratitan: Alright I fixed that. And made NewFileIcon take a URI as well.

dialog/file.go
okratitan: I added a constructor for file and save dialogs that take a FileFilter.  What do you think?

dialog/file.go
```diff
@@ -30,10 +30,54 @@ type fileDialog struct {
...
+// Matches returns true if a file URI has one of the filtered mimetypes
+func (mt *mimeTypeFileFilter) Matches(uri fyne.URI) bool {
+	_, mimeType, mimeSubType := mimeTypeGet(uri)
+	for _, mimeTypeFull := range mt.mimeTypes {
+		mimeTypeSplit := strings.Split(mimeTypeFull, "/")
+		if len(mimeTypeSplit) <= 1 {
+			continue
+		}
+		mType := mimeTypeSplit[0]
+		mSubType := mimeTypeSplit[1]
```
stuartmscott, line 71: Should this strip ';*'?

dialog/file.go
```diff
@@ -165,11 +209,13 @@ func (f *fileDialog) refreshDir(dir string) {
...
-
 		itemPath := filepath.Join(dir, file.Name())
 		if file.IsDir() {
 			icons = append(icons, f.newFileItem(itemPath, true))
 		} else {
+			if f.filter != nil && !f.filter.Matches(storage.NewURI(itemPath)) {
+				continue
+			}
 			icons = append(icons, f.newFileItem(itemPath, false))
 		}
```
stuartmscott, line 220:
```suggestion
		} else if f.filter == nil || f.filter.Matches(storage.NewURI(itemPath)) {
			icons = append(icons, f.newFileItem(itemPath, false))
		}
```

dialog/file.go
okratitan: It already gets stripped by mimetype function

stuartmscott: What about mt.mimeTypes?

okratitan: Good point. Should probably strip in case the user provided mimetypes have it.

dialog/file.go
okratitan: 👍

cmd/fyne_demo/screens/window.go
```diff
@@ -136,24 +136,48 @@ func loadDialogGroup(win fyne.Window) *widget.Group {
 			prog.Show()
 		}),
 		widget.NewButton("File Open (try txt or png)", func() {
-			dialog.ShowFileOpen(func(reader fyne.FileReadCloser, err error) {
```
andydotxyz, line 139: ShowFileOpen should still be appropriate here, no configuration is being done?

cmd/fyne_demo/screens/window.go
```diff
@@ -136,24 +136,48 @@ func loadDialogGroup(win fyne.Window) *widget.Group {
...
 					return
 				}
 
 				fileOpened(reader)
 			}, win)
+			fd.SetFilter(dialog.NewExtensionFileFilter([]string{".png", ".txt"}))
+			fd.Show()
 		}),
 		widget.NewButton("File Save", func() {
-			dialog.ShowFileSave(func(writer fyne.FileWriteCloser, err error) {
```
andydotxyz, line 149: As before, use the Show...() shortcut?

dialog/file.go
```diff
@@ -30,10 +30,63 @@ type fileDialog struct {
 	win      *widget.PopUp
 	selected *fileDialogItem
 	callback interface{}
```
andydotxyz, line 32: callback, filter and parent are all duplicated in FileDialog, is that needed?

dialog/file.go
```diff
@@ -249,23 +301,55 @@ func showFileDialog(save bool, callback interface{}, parent fyne.Window) {
 	d.win.Resize(size)
 
 	d.win.Show()
+	return d
 }
 
-// ShowFileOpen shows a file dialog allowing the user to choose a file to open.
-// The dialog will appear over the window specified.
-func ShowFileOpen(callback func(fyne.FileReadCloser, error), parent fyne.Window) {
```
andydotxyz, line 256: Why not keep the ShowFileOpen as a shortcut to New...().Show() like we discussed?

dialog/file.go
```diff
@@ -249,23 +301,55 @@ func showFileDialog(save bool, callback interface{}, parent fyne.Window) {
...
+// NewMimeTypeFileFilter takes a string slice of mimetypes, including globs, and creates a filter for the file dialog.
+// Example: image/*, audio/mp3, text/plain, application/*
+func NewMimeTypeFileFilter(mimeTypes []string) FileFilter {
+	return &mimeTypeFileFilter{mimeTypes: mimeTypes}
 }
 
-// ShowFileSave shows a file dialog allowing the user to choose a file to save to (new or overwrite).
+// NewFileOpenDialog creates a file dialog allowing the user to choose a file to open.
+// The dialog will appear over the window specified.
+func NewFileOpenDialog(callback func(fyne.FileReadCloser, error), parent fyne.Window) *FileDialog {
```
andydotxyz, line 344: dialog.NewFileOpenDialog is repetition and the others haven't got ...Dialog suffix, is it needed?

dialog/file_test.go
```diff
@@ -17,10 +17,10 @@ func TestShowFileOpen(t *testing.T) {
 	var chosen fyne.FileReadCloser
 	var openErr error
 	win := test.NewWindow(widget.NewLabel("Content"))
-	ShowFileOpen(func(file fyne.FileReadCloser, err error) {
+	NewFileOpenDialog(func(file fyne.FileReadCloser, err error) {
```
andydotxyz, line 20: Why not have the shortcut so these lines don't change either

dialog/fileicon_test.go
```diff
@@ -10,42 +10,43 @@ import (
...
+	"fyne.io/fyne/storage"
 	"fyne.io/fyne/theme"
 )
 
 func TestNewFileIcon(t *testing.T) {
 	f := &fileDialog{}
 	_ = f.makeUI()
 
-	item := NewFileIcon("/path/to/filename.zip")
+	item := NewFileIcon(storage.NewURI("/path/to/filename.zip"))
```
andydotxyz, line 21: URI should really have a scheme as well, i.e. `file:///path/to/...`

dialog/fileitem.go
```diff
@@ -60,7 +61,7 @@ func (f *fileDialog) newFileItem(path string, dir bool) *fileDialogItem {
 	if dir {
 		icon = canvas.NewImageFromResource(theme.FolderIcon())
 	} else {
-		icon = NewFileIcon(path)
+		icon = NewFileIcon(storage.NewURI(path))
```
andydotxyz, line 64: Another case of needing `file://` prefix

okratitan:
> You removed the old ShowFileOpen API, but I thought it was discussed that would stay in as a shortcut? And the new APIs gained a `Dialog` suffix
> 
> I noticed that FileDialog does not implement Dialog, was that decided against?
> 
> Otherwise the API seems nice and clean

I added the old API back in as a shortcut and removed the Dialog suffix.  FileDialog, now File, does not implement Dialog because I thought it would be messy to do with native pickers on mobile.  For instance File is a type that can be used for both the desktop and mobile pickers to pass in configuration/settings.  On Desktop, fileDialog is created from this.  On mobile it will be used to setup options in the native picker.  That being said, while we have a way to handle Show() gracefully if Dialog was implemented.  I think other cases would get a little more complicated and I would need to look at the mobile side... For instance with Hide()... easiest enough to hide the fileDialog on desktop.  Do we have a handle to the native filepicker that we could use to hide it?  with SetOnClosed... The File type already takes in a callback that gets called on close as part of the design so that seems redundant.  SetDismissText wouldn't be used either.  So with all of that being the case and never reaching an agreement on it in the Slack channel - I just chose the simplest, cleanest way to solve the issue and that was just Show() without implementing the Dialog interface.  I'm sure if there are objections, we could implement it simply enough.

okratitan:
> Sorry I missed that a public API from before was not using URI, should probably change to be consistent.
> Mostly i am not sure if the filter parameter should be part of the constructing method, and perhaps we should split out a New version for people who want to customise?

Done. Thanks!

andydotxyz:
> Tests added.

This is all looking good, but I can't find the tests you mentioned above?

dialog/file.go
```diff
@@ -20,22 +20,72 @@ type textWidget interface {
...
 	fileScroll *widget.ScrollContainer
-	parent     fyne.Window
 
 	win      *widget.PopUp
 	selected *fileDialogItem
-	callback interface{}
 	dir      string
+}
+
+// File is a dialog containing a file picker for use in opening or saving files
```
toaster, line 35: It’s confusing to speak of a “dialog” here if `File` does not implement `Dialog`.

dialog/file.go
```diff
@@ -245,27 +294,82 @@ func showFileDialog(save bool, callback interface{}, parent fyne.Window) {
...
-	d.win = widget.NewModalPopUp(ui, parent.Canvas())
+	d.win = widget.NewModalPopUp(ui, file.parent.Canvas())
 	d.win.Resize(size)
 
 	d.win.Show()
+	return d
 }
 
-// ShowFileOpen shows a file dialog allowing the user to choose a file to open.
+// SetFilter sets a filter for limiting files that can be chosen in the file dialog
```
toaster, line 304: nit

```suggestion
// SetFilter sets a filter for limiting files that can be chosen in the file dialog.
```

dialog/file.go
toaster, line 35: nit
```suggestion
// File is a dialog containing a file picker for use in opening or saving files.
```

dialog/file.go
```diff
@@ -20,22 +20,72 @@ type textWidget interface {
...
+// File is a dialog containing a file picker for use in opening or saving files
+type File struct {
 	save     bool
+	callback interface{}
+	filter   FileFilter
+	parent   fyne.Window
+	dialog   *fileDialog
+}
+
+// FileFilter is an interface that can be implemented to provide a filter to a file dialog
```
toaster, line 44: nit
```suggestion
// FileFilter is an interface that can be implemented to provide a filter to a file dialog.
```

dialog/file.go
```diff
@@ -20,22 +20,72 @@ type textWidget interface {
...
+
+type extensionFileFilter struct {
+	extensions []string
+}
+
+type mimeTypeFileFilter struct {
+	mimeTypes []string
+}
+
+// Matches returns true if a file URI has one of the filtered extensions
```
toaster, line 57: nit
```suggestion
// Matches returns true if a file URI has one of the filtered extensions.
```

dialog/file.go
```diff
@@ -20,22 +20,72 @@ type textWidget interface {
...
+	extension := uri.Extension()
+	for _, ext := range e.extensions {
+		if extension == ext {
+			return true
+		}
+	}
+	return false
+}
+
+// Matches returns true if a file URI has one of the filtered mimetypes
```
toaster, line 68: nit
```suggestion
// Matches returns true if a file URI has one of the filtered mimetypes.
```

dialog/file.go
```diff
@@ -245,27 +294,82 @@ func showFileDialog(save bool, callback interface{}, parent fyne.Window) {
...
+		if fileOpenOSOverride(f.callback.(func(fyne.FileReadCloser, error)), f.parent) {
+			return
+		}
+		f.dialog = showFile(f)
+		return
+	}
+	if fileSaveOSOverride(f.callback.(func(fyne.FileWriteCloser, error)), f.parent) {
+		return
+	}
+	f.dialog = showFile(f)
```
toaster, line 324:
```suggestion
	if f.save {
		if fileSaveOSOverride(f.callback.(func(fyne.FileWriteCloser, error)), f.parent) {
			return
		}
	} else {
		if fileOpenOSOverride(f.callback.(func(fyne.FileReadCloser, error)), f.parent) {
			return
		}
	}
	f.dialog = showFile(f)
```

dialog/file.go
```diff
@@ -245,27 +294,82 @@ func showFileDialog(save bool, callback interface{}, parent fyne.Window) {
...
+	dialog.Show()
+	return dialog
 }
 
-// ShowFileSave shows a file dialog allowing the user to choose a file to save to (new or overwrite).
+// ShowFileSave creates and shows a file dialog allowing the user to choose a file to save to (new or overwrite).
 // If the user chooses an existing file they will be asked if they are sure.
 // The dialog will appear over the window specified.
-func ShowFileSave(callback func(fyne.FileWriteCloser, error), parent fyne.Window) {
+func ShowFileSave(callback func(fyne.FileWriteCloser, error), parent fyne.Window) *File {
```
toaster, line 368: The `Show*` methods I already know are just fire and forget and do not return the created entity.
It this about to change?
If so, why?
If not, why do the `ShowFile*` functions return the `File`?

dialog/fileicon.go
```diff
@@ -97,8 +96,7 @@ func (s fileIconRenderer) Refresh() {
 	canvas.Refresh(s.item)
 }
 
-func mimeTypeGet(path string) (ext, mimeType, mimeSubType string) {
-	uri := storage.NewURI(path)
+func mimeTypeGet(uri fyne.URI) (ext, mimeType, mimeSubType string) {
```
toaster, line 99: The name of this function confuses me.
AFAICS there is one usage for all delivered details and one new use which only needs the mime type information.
Because the extension and the mime type are two different things (one is a property of the file path the other one is a property of the file data (strictly speaking)), I suggest to split it into two methods like `extensionForURI` and `mimeTypeForURI`.

storage/uri_test.go
```diff
@@ -1,14 +1,18 @@
-package storage
+package storage_test
 
 import (
 	"testing"
 
+	"fyne.io/fyne/storage"
+
```
toaster, line 7:
```suggestion
```

dialog/fileicon.go
andydotxyz: extensionForURI would just be uri.Extension() ;)

dialog/file.go
andydotxyz: Good catch, I missed this - the Show functions should indeed not return the type, it's fire and forget as pointed out.

dialog/file.go
andydotxyz: This probably needs discussion in the channel. I queried this too, but due to some methods (Hide and SetDismissText) not being applicable to mobile @okratitan decided we should not use it.

dialog/fileicon.go
okratitan: yes both already exist uri.Mimetype and uri.Extension ... This function returns the uri.MimeType split into type and subtype for use in creating file icons and filtering on dialogs.  It returns the extension too because that was needed to build out the File Icon and it would involve one less call to uri.Extension since we were already calling it in that funciton. I'll remove extension from the return though and just call uri.Extension again.

okratitan:
> Looks good!
> Just a couple of nits, a question and a suggestion.

Thanks!  I think I resolved all of those points.

dialog/file.go
```diff
@@ -245,27 +294,80 @@ func showFileDialog(save bool, callback interface{}, parent fyne.Window) {
...
 }
 
-// ShowFileOpen shows a file dialog allowing the user to choose a file to open.
+// Show shows the file dialog
+func (f *FileDialog) Show() {
+	if !f.save {
+		if fileOpenOSOverride(f.callback.(func(fyne.FileReadCloser, error)), f.parent) {
+			return
+		} 
+	else {
```
stuartmscott, line 318: is there a missing '}' before else?

dialog/file.go
```diff
@@ -245,27 +294,80 @@ func showFileDialog(save bool, callback interface{}, parent fyne.Window) {
...
+	f.filter = filter
+	if f.dialog != nil {
+		f.dialog.refreshDir(f.dialog.dir)
+	}
 }
 
-// ShowFileOpen shows a file dialog allowing the user to choose a file to open.
+// Show shows the file dialog
+func (f *FileDialog) Show() {
+	if !f.save {
```
stuartmscott, line 314: I would swap these two cases to remove the '!'

dialog/file.go
```diff
@@ -20,22 +20,72 @@ type textWidget interface {
...
 	win      *widget.PopUp
 	selected *fileDialogItem
-	callback interface{}
 	dir      string
+}
+
+// FileDialog is a dialog containing a file picker for use in opening or saving files.
+type FileDialog struct {
 	save     bool
+	callback interface{}
```
stuartmscott, line 38: I wonder if we should have openCallback and saveCallback to avoid the type casts later on

dialog/fileicon.go
```diff
@@ -97,9 +97,8 @@ func (s fileIconRenderer) Refresh() {
 	canvas.Refresh(s.item)
 }
 
-func mimeTypeGet(path string) (ext, mimeType, mimeSubType string) {
-	uri := storage.NewURI(path)
-	ext = uri.Extension()
+func mimeTypeGet(uri fyne.URI) (mimeType, mimeSubType string) {
```
stuartmscott, line 100: It is odd having 'Get' as a suffix. Maybe "uriToMime" or "mimeTypeFromURI"?

okratitan: the mimeTypeFromURI is uri.MimeType() ... It already exists as discussed on Tilo's comment earlier. uri.MimeType() returns a full mimetype: "image/jpeg" for instance... For the purposes of filtering and building out the file icon however we need it split so that we have "image" and "jpeg" ... This isn't a public method or anything - just a utility to avoid duplication by splitting the mimetype returned from URIs.  I'm open for suggestions here.

okratitan: Perhaps mimeTypeSplit(uri fyne.URI) (mimetype, mimeSubType  string) would make sense

dialog/file.go
okratitan: I don't mind the type casts later, but perhaps others may have some thought here.

dialog/file.go
okratitan: Okay most recent commit is with the file dialog implementing Dialog.

dialog/fileicon.go
stuartmscott: Again the verb suffix is odd, "splitMime"?

dialog/file.go
```diff
@@ -50,28 +107,34 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
...
 	label := "Open"
-	if f.save {
+	if f.file.save {
 		label = "Save"
 	}
 	f.open = widget.NewButton(label, func() {
-		if f.callback == nil {
+		if f.file.callback == nil {
 			f.win.Hide()
+			if f.file.onClosedCallback != nil {
```
andydotxyz, line 116: Should this be conditional on `file.callback` not being nil?

dialog/file.go
```diff
@@ -245,27 +319,122 @@ func showFileDialog(save bool, callback interface{}, parent fyne.Window) {
...
+	if f.dialog != nil {
+		f.dialog.win.Show()
+		return
+	}
+	f.dialog = showFile(f)
+}
+
+// Hide hides the file dialog.
+func (f *FileDialog) Hide() {
+	if f.nativePicker {
```
andydotxyz, line 351: Do you need the extra flag? Why not just check f.dialog == nil?

dialog/file.go
```diff
@@ -245,27 +319,122 @@ func showFileDialog(save bool, callback interface{}, parent fyne.Window) {
...
+	}
+	f.dialog.win.Hide()
+	if f.onClosedCallback != nil {
+		f.onClosedCallback(false)
+	}
+}
+
+// SetDismissText allows custom text to be set in the confirmation button
+func (f *FileDialog) SetDismissText(label string) {
+	if f.nativePicker {
```
andydotxyz, line 362: again, checking f.dialog for nil tells you cannot set the text

dialog/file.go
```diff
@@ -245,27 +319,122 @@ func showFileDialog(save bool, callback interface{}, parent fyne.Window) {
...
+		return
+	}
+	f.dialog.dismiss.SetText(label)
+	widget.Refresh(f.dialog.win)
+}
+
+// SetOnClosed sets a callback function that is called when
+// the dialog is closed.
+func (f *FileDialog) SetOnClosed(closed func()) {
+	// If there is already a callback set, remember it and call both.
```
andydotxyz, line 372: I think you added code in the wrong place so this document no longer applies to the code it's next to

dialog/fileicon.go
andydotxyz: Hangover from C and EFL days I guess - they always put verbs on the end. I would agree that splitMime seems clean.

andydotxyz: I “unresolved” this as I think Stuart’s point was not responded to. Not sure if GitHub automatically was dismissing or not.

okratitan: Oh I was thinking of split as a noun and not a verb. As in the return is a mime type split. I could see that being confusing and will change it to splitMimeType

dialog/file.go
okratitan: file.callback is the read/write closer callback ... file.onClosedCallback is the onClosed callbacks ... They are unrelated as far as I can tell

okratitan: Oh I see you were actually referring to code above. In that space we only hide immediately if there is no file.callback ... so we wouldn't want to call the onClosedCallback until the dialog is closed... in this case when file.callback is not nil, we go through things like validation that may keep the dialog open still - if validation passes we call the Hide() and onClosedCallbacks later in the method...

So long story short - I think this is appropriate.

andydotxyz: Thanks for the explanation, but I don't quite follow - there is a return here so I don't see where it falls through for the extra processes?

andydotxyz: Oh I see what you mean now sorry

dialog/file.go
stuartmscott: Unresolving this to see if others have an opinion here

andydotxyz: I didn't see that this had been reopened, sorry for that.
I agree that it could be improved, it has evolved a little since the first commit and we learned a lot along the way. Let's do a follow-on following discussion in Slack

PR #1039, opened 2020-05-27
Start stop progress bars in demo when tab shown/hidden

andydotxyz:
### Description:
Start progress when tab appears, stop when it goes away.
Reset progress to 0 when shown again.

Makes for a better demo and stops background processor usage :)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
I could not think of any other way to clear a channel that may be empty.

andydotxyz:
Yeah, I didn’t want to add to much complexity. This was more important so it didn’t start background processing on load. So I think this is the right balance.

PR #1040, opened 2020-05-27
Add Tab Overflow Drop Down

### LINKED ISSUE #371: Implement overflow on tab container
When there are more than 7 tabs UI tends to overfit, it'll be nice if a scroll bar would be added to the container

stuartmscott:
### Description:
Adds "More" Button to Tab Bar when Tabs Overflow

Fixes #371

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


widget/tabcontainer.go
```diff
@@ -580,3 +695,150 @@ func (r *tabButtonRenderer) padding() fyne.Size {
...
+	}
+}
+
+func (b *tabOverflowButton) updateOverflowPopUp() {
+	b.container.propertyLock.RLock()
+	expanded := b.container.overflowExpanded
+	b.container.propertyLock.RUnlock()
+	if expanded {
+		log.Println("Overflow Expanded")
+		if b.popUp == nil {
```
andydotxyz, line 812: It seems like expanded might not be needed - the presence of the popup should be sufficient to know if it should be managed. The tap event of the button could be used to create the overflow, removing it's code from this refresh method. The overflow would be removed by it's dismissal or item selection so that should not have to be handled here either.

andydotxyz:
I wonder if when an icon is not present the text could be better as "More..." so it's evidently different to the others?

widget/tabcontainer.go
stuartmscott: The tabOverflowButton could manage the popup without the expanded flag, however it would not be possible to programmatically toggle the overflow popup without getting the renderer for the TabContainer to access the tabOverflowButton and tap it. Since the tests need to toggle the overflow they would need to be in the same package, or tabContainerRenderer, tabOverflowButton would need to be exported

andydotxyz: If this is an important path then another approach to support it would be for the TabContainer to be told which button is the more one, so it can be tapped.
i.e. `contaner.moreButton := more` in the CreateRenderer, or wherever is appropriate.

andydotxyz:
Next time develop is merged in can you please remove the "if mobile" check in cmd/fyne_demo/main.go that removes the advanced tab? thanks :)

stuartmscott:
On hold, pending #1310

andydotxyz:
Is this replaced by #1755 @stuartmscott ?

stuartmscott:
Yes, this PR is replaced by #1755

PR #1046, opened 2020-05-29
Bugfix: gradient color computation

toaster:
### Description:

The gradient colour computation was wrong. It computed the final colour components as alpha-premultiplied but then treated them as non-alpha-premultiplied. This resulted in a kind of shade when drawing a gradient from white to transparent over a white background.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


test/util.go
```diff
@@ -128,6 +130,23 @@ func MoveMouse(c fyne.Canvas, pos fyne.Position) {
 	}
 }
 
+// NewCheckedImage returns a new black/white checked image with the specified size
+// and the specified amount of horizontal and vertical tiles.
+func NewCheckedImage(w, h, hTiles, vTiles int) image.Image {
```
andydotxyz, line 135: I'm not sure this fits with the rest of the test utilities that help people to test their apps.
This is more like something that we need to test the framework, so does it belong internal instead?

internal/painter/software/painter_test.go
```diff
@@ -15,17 +15,7 @@ import (
...
-	c := 0
-	for y := 0; y < h; y++ {
-		c = y % 2
-		for x := 0; x < w; x++ {
-			src.Set(x, y, colors[c])
-			c = 1 - c
-		}
-	}
-	return src
+	return test.NewCheckedImage(w, h, w, h)
```
andydotxyz, line 18: Are you sure this test compiles with the new package location?

toaster: No, they didn’t. I shouldn’t shoot so quick.

PR #1048, opened 2020-05-29
When a menu is dismissed on mobile dismiss the bar too

andydotxyz:
### Description:
Close menu bar when an item is selected

Thanks to @toaster's menu work - this was not possible before.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/gomobile/menu_test.go
+++ b/internal/driver/gomobile/menu_test.go
@@ -9,7 +9,7 @@ import (
 
 	"fyne.io/fyne"
 	"fyne.io/fyne/canvas"
-	widget2 "fyne.io/fyne/internal/widget"
+	internalWidget "fyne.io/fyne/internal/widget"
 	"fyne.io/fyne/theme"
 	"fyne.io/fyne/widget"
 )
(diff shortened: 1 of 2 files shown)


internal/driver/gomobile/menu_test.go
```diff
@@ -5,12 +5,48 @@ package gomobile
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
 	"fyne.io/fyne"
```
toaster, line 10: I thought the testify imports should go behind the fyne.io imports. 😕 
However, I really don’t care as long as `gofmt` or `goimports` don’t touch it 😃.

internal/driver/gomobile/menu_test.go
```diff
@@ -5,12 +5,48 @@ package gomobile
...
+	c.SetContent(canvas.NewRectangle(theme.BackgroundColor()))
+	menu := fyne.NewMainMenu(
+		fyne.NewMenu("Test"))
+	c.showMenu(menu)
+	c.resize(fyne.NewSize(100, 100))
+
+	assert.NotNil(t, c.menu)
+	// simulate tap as the test util does not know about our menu...
+	c.tapDown(fyne.NewPos(80, 20), 1)
+	c.tapUp(fyne.NewPos(80, 20), 1, nil, nil, nil)
```
toaster, line 28: What about using `test.TapCanvas`?

internal/driver/gomobile/menu_test.go
andydotxyz: Oh yes you are right. I think I reset my goland config by mistake.

internal/driver/gomobile/menu_test.go
toaster: This should be usable to _show_ the menu, too.

internal/driver/gomobile/menu_test.go
andydotxyz: Fixed :)

internal/driver/gomobile/menu_test.go
andydotxyz: That is what I was commenting on in the code - it does not work because gomobile driver does special things to handle menu. The TapCanvas utility only checks Content() and Overlays() trees - the menu bar is neither :(

toaster: Ah … okay

PR #1058, opened 2020-06-01
Focusing logic

pymq:
### Description:
Made symmetric changes to focusing logic in glCanvas as in mobileCanvas from #975. Added tests. Focusing already focused object is no-op now and manual focusing disabled objects is now impossible.

Follow-up to #975

### Checklist:

- [x] Tests included.
- [ ] Tests all pass. --- Test_glCanvas_InsufficientSizeDoesntTriggerResizeIfSizeIsAlreadyMaxedOut doesn't pass on develop, but it's disabled in CI run. Seems like a known issue.

#### Where applicable:

### DIFF
--- a/internal/driver/glfw/canvas.go
+++ b/internal/driver/glfw/canvas.go
@@ -167,6 +167,16 @@ func (c *glCanvas) Focus(obj fyne.Focusable) {
 	c.RLock()
 	focused := c.focused
 	c.RUnlock()
+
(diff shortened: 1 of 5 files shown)


andydotxyz:
The test failure is not a known issue. The whole GL package is disabled on Travis due to not having an output device

andydotxyz:
Great thanks

PR #1059, opened 2020-06-01
Keep mobile apps within screen bounds whilst ensuring layout updates when required

### LINKED ISSUE #1056: mobile: content doesn't fit into screen
### Screenshots:
picture

### Device:
 - OS: Android
 > Pixel 3 API 27__8.1 emulator
hw.lcd.height: 1920
hw.lcd.width: 1080
hw.lcd.density: 420
5.0"
 - Go version: 1.14.2
 - Fyne version: develop, 85790ee40c7570efa530fa8bcd0d2a35bd0860b6

andydotxyz:
### Description:
Fixes the issue when ensureMinSize expanded window outside of screen size.

This stops minSize scans when nothing has changed as well.
Also remove unneeded OpenGL debug now we have error logging.

Fixes #1056

### Checklist:

- [ ] Tests included. <- run in -tags mobile mode and size down, or use a small mobile screen
- [x] Tests all pass.

### DIFF
--- a/internal/driver/gomobile/canvas.go
+++ b/internal/driver/gomobile/canvas.go
@@ -39,6 +39,7 @@ type mobileCanvas struct {
 	lastTapDownPos map[int]fyne.Position
 	dragging       fyne.Draggable
 	refreshQueue   chan fyne.CanvasObject
+	minSizeCache   map[fyne.CanvasObject]fyne.Size
 }
 
 func (c *mobileCanvas) Content() fyne.CanvasObject {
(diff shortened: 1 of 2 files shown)


internal/driver/gomobile/canvas.go
```diff
@@ -382,6 +406,7 @@ func (c *mobileCanvas) tapUp(pos fyne.Position, tapID int,
 
 	if c.menu != nil && c.overlays.Top() == nil && pos.X > c.menu.Size().Width {
 		c.menu.Hide()
+		c.menu.Refresh()
```
okratitan, line 409: Hide() doesn't refresh already?

andydotxyz: It normally would, yes.
However this is slightly complicated by the c.menu = nil below...
Because the menu is an object root it being removed means that it is not traversed and the object is not found in the normal refresh cycle - so I had to call the Refresh().

Alternative is to remove the nil and set it after the refresh completes, but that was more complex.

okratitan: Okay - Good enough reasoning for me.

PR #1060, opened 2020-06-01
Improve default apps for mobile

andydotxyz:
### Description:
This is so that we better fit on mobile devices (size and lifecycle wise).

### Checklist:

- [ ] Tests included. <- just cosmetic
- [x] Tests all pass.


andydotxyz:
Feedback was that it wouldn’t make it, so I did this instead. The alternative is having buttons that overlap each other as we went from 5 to 6 in this release

andydotxyz:
Good idea, commented on the issue

PR #1069, opened 2020-06-02
Add changelog for 1.3 release

andydotxyz:
### Description:
Changelog for release of 1.3 (date TBC)


andydotxyz:
I had to pick from the > 1000 commits so this (particularly in fixes) is just highlights. I don’t think calling out text fixes is necessary. Security warning fixes is a possibility - though the work on that is incomplete is it not?

Jacalz:
I think we can deem it more or less complete. All of the remaining issues are part of the code imported from gomobile. If anything, it should move over to the fork of gomobile and we can see from there what we can do. I will close the issue sometime soon. It is good to have in the release notes if you ask me 🙂

PR #1074, opened 2020-06-03
WIP:  Keyboard only operation

jkvatne:
### Description:
This PR implements keyboard only operation of most but not all widgets. This will improve usability on desktop computers, allowing operation without a mouse.

It passes all tests except the tests where images are compared. The reason is that I had to modify the coloring sheme in order to distinguish between focused and primary colors. The problem is similar to the ongoing changes to implement button animation.  What I have called PressedColor should perhaps be named TappedColor.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


andydotxyz:
Just looking at Travis it seems like failures are formatting, test failures and linting - I'd recommend the git hook we use https://github.com/fyne-io/fyne/wiki/GitHooks

internal/app/focus.go
```diff
@@ -13,7 +13,12 @@ type FocusManager struct {
 func (f *FocusManager) nextInChain(current fyne.Focusable) fyne.Focusable {
 	var first, next fyne.Focusable
 	found := current == nil // if we have no starting point then pretend we matched already
-	driver.WalkVisibleObjectTree(f.canvas.Content(), func(obj fyne.CanvasObject, _ fyne.Position, _ fyne.Position, _ fyne.Size) bool {
+	//OBS driver.WalkVisibleObjectTree(f.cavas.Content(), func(obj fyne.CanvasObject, _ fyne.Position, _ fyne.Position, _ fyne.Size) bool {
```
andydotxyz, line 16: OBS? try not to commit commented out code

andydotxyz: This may relate to bug https://github.com/fyne-io/fyne/issues/814

internal/driver/glfw/menu_bar_item.go
```diff
@@ -112,11 +112,22 @@ func (i *menuBarItem) Show() {
...
 // It shows the item’s menu if the bar is activated and hides it if the bar is deactivated.
 // Implements: fyne.Tappable
-func (i *menuBarItem) Tapped(*fyne.PointEvent) {
+func (i *menuBarItem) Tapped(pos *fyne.PointEvent) {
 	if i.Parent.active {
 		i.Parent.deactivate()
 	} else {
+		// Set then menu bar's index to the tapped items index and update focus
 		i.Parent.activateChild(i)
+		for j := 0; j < len(i.Parent.Items); j++ {
```
andydotxyz, line 121: I don't understand this loop...
`i.Parent.Items[i.Parent.index]` is `i` so the item that has just been activated is being defocussed (or "upped").
Possibly this is because I don't understand what `index` is

internal/driver/glfw/menu_bar.go
```diff
@@ -16,11 +16,41 @@ var _ fyne.Widget = (*MenuBar)(nil)
 // MenuBar is a widget for displaying a fyne.MainMenu in a bar.
 type MenuBar struct {
 	widget.Base
-	Items []fyne.CanvasObject
-
+	Items      []fyne.CanvasObject
+	index      int
```
andydotxyz, line 20: index is quite generic, not clear what this means

internal/driver/glfw/menu_bar.go
```diff
@@ -16,11 +16,41 @@ var _ fyne.Widget = (*MenuBar)(nil)
...
 	active     bool
-	activeItem *menuBarItem
 	canvas     fyne.Canvas
+	activeItem *menuBarItem
+}
+
+// HandleKey will process keys pressed when menu is open
+func (b *MenuBar) HandleKey(key fyne.KeyName) {
```
andydotxyz, line 27: This should not be an exported API

internal/driver/glfw/menu_bar.go
```diff
@@ -16,11 +16,41 @@ var _ fyne.Widget = (*MenuBar)(nil)
...
+	if key == fyne.KeyEscape {
+		b.deactivate()
+	} else if key == fyne.KeyEnter || key == fyne.KeyReturn || key == fyne.KeySpace {
+		// Simulate tapping on  menu entry
+		b.Items[b.index].(*menuBarItem).Child().Do()
+	} else if key == fyne.KeyRight {
+		// Move to the menu item at right
+		if b.index < len(b.Items)-1 {
+			b.Items[b.index].(*menuBarItem).Tapped(nil)
```
andydotxyz, line 36: Tapped should not be called on an item with an invalid parameter.
Would it not be suitable to call the Action()?

internal/driver/glfw/menu_bar.go
```diff
@@ -114,7 +145,7 @@ func (b *MenuBar) deactivate() {
 	if !b.active {
 		return
 	}
-
+	b.active = false
```
andydotxyz, line 148: dupe

internal/driver/glfw/menu_bar_test.go
```diff
@@ -182,7 +182,7 @@ func TestMenuBar(t *testing.T) {
 				{
 					actions:    []action{{"tap", fileNewPos}},
 					wantAction: "new",
-					wantImage:  "menu_bar_initial.png",
+					wantImage:  "menu_bar_initial1.png",
```
andydotxyz, line 185: why postfixing numbers? it is not clear what the intended differences are

internal/driver/glfw/window.go
```diff
@@ -621,11 +622,19 @@ func (w *window) mouseClicked(_ *glfw.Window, btn glfw.MouseButton, action glfw.
...
-		if wid, ok := co.(fyne.Focusable); ok {
-			w.canvas.Focus(wid)
-		} else {
-			w.canvas.Unfocus()
+		needsfocus = true
+
+		if wid := w.canvas.Focused(); wid != nil {
+			if wid.(fyne.CanvasObject) != co {
+				// Avoid changing focus when a toolbar button is pressed.
+				if _, isToolbarButton := co.(*widget.ToolbarButton); !isToolbarButton {
```
andydotxyz, line 632: This is confusing - if you made a new ToolbButton so that it didn't accept focus like normal buttons then this should not be needed

internal/driver/glfw/window.go
```diff
@@ -635,6 +644,13 @@ func (w *window) mouseClicked(_ *glfw.Window, btn glfw.MouseButton, action glfw.
 		w.mouseButton = 0
 	}
 
+git	// we cannot switch here as objects may respond to multiple cases
```
andydotxyz, line 647: typo I'm guessing

internal/driver/glfw/window.go
```diff
@@ -857,7 +873,26 @@ func (w *window) keyPressed(viewport *glfw.Window, key glfw.Key, scancode int, a
...
+		}
+	}
+	if isMenu {
+		if action == glfw.Release {
+			menu.HandleKey(keyName)
+		}
+		return
+	}
+
+	if keyName == fyne.KeyTab && action != glfw.Release && !isMenu {
```
andydotxyz, line 895: The release check appears duplicating 2 liones below?

internal/driver/glfw/window.go
```diff
@@ -857,7 +873,26 @@ func (w *window) keyPressed(viewport *glfw.Window, key glfw.Key, scancode int, a
 	keyEvent := &fyne.KeyEvent{Name: keyName}
 	keyDesktopModifier := desktopModifier(mods)
 
-	if keyName == fyne.KeyTab {
+	// Activate the menu when the Alt key is pressed
+	focused := w.canvas.Focused()
+	isMenu := false
```
andydotxyz, line 878: This variable is reduntant - inside the ok the menu.active flag could apply to both blocks - the return in the second block means the later check on 895 is not needed

theme/theme.go
```diff
@@ -61,6 +61,39 @@ func DarkTheme() fyne.Theme {
 	return theme
 }
 
+// shade will darken a light color and lighten a light color by the given % value
```
andydotxyz, line 64: % or pct should explain it's expected range better.
In the code it clamps it to < 50, but I couldn't say why

theme/theme.go
```diff
@@ -61,6 +61,39 @@ func DarkTheme() fyne.Theme {
...
+	if pct == 0 {
+		return color.NRGBA{uint8(r), uint8(g), uint8(b), uint8(a)}
+	}
+	if pct > 50 {
+		pct = 50
+	}
+	if r+g+b < 3*0x8080 {
+		// Lighten
+		return color.NRGBA{
+			R: uint8((r + (0x10000-r)*pct/50) >> 8),
```
andydotxyz, line 77: percent / 50 is a strange seemting thing to do?

theme/theme.go
```diff
@@ -61,6 +61,39 @@ func DarkTheme() fyne.Theme {
 	return theme
 }
 
+// shade will darken a light color and lighten a light color by the given % value
+// should use 4,8, 12 or 14 percent darken/lighten to implement google material design rules
+func Shade(c color.Color, pct uint32) color.Color {
+	r, g, b, a := c.RGBA()
+	if pct == 0 {
+		return color.NRGBA{uint8(r), uint8(g), uint8(b), uint8(a)}
```
andydotxyz, line 69: better would be `return c` surely?

theme/theme.go
```diff
@@ -61,6 +61,39 @@ func DarkTheme() fyne.Theme {
...
+	}
+	if pct > 50 {
+		pct = 50
+	}
+	if r+g+b < 3*0x8080 {
```
andydotxyz, line 74: it seems like this might be important, should it be extracted into something like an `isLight` helper?

theme/theme.go
```diff
@@ -61,6 +61,39 @@ func DarkTheme() fyne.Theme {
...
+		return color.NRGBA{
+			R: uint8((r * (100 - pct) / 100) >> 8),
+			G: uint8((g * (100 - pct) / 100) >> 8),
+			B: uint8((b * (100 - pct) / 100) >> 8),
+			A: uint8(a >> 8),
+		}
+	}
+}
+
+const PressedShade = 14
```
andydotxyz, line 93: Are these percentages? should they be public API?

widget/button.go
```diff
@@ -118,10 +118,16 @@ func (b *buttonRenderer) BackgroundColor() color.Color {
 	switch {
 	case b.button.Disabled():
 		return theme.DisabledButtonColor()
-	case b.button.Style == PrimaryButton:
-		return theme.PrimaryColor()
+	case b.button.pressed:
```
andydotxyz, line 121: I guess that primary and default buttons should have a different pressed colour?

widget/button.go
```diff
@@ -171,7 +177,10 @@ type Button struct {
 	OnTapped   func() `json:"-"`
 	HideShadow bool
 
-	hovered, tapped bool
+	focused  bool
```
andydotxyz, line 180: why 4 lines when 1 would do?

widget/button.go
```diff
@@ -182,6 +191,7 @@ const (
 	DefaultButton ButtonStyle = iota
 	// PrimaryButton that should be more prominent to the user
 	PrimaryButton
+	CancelButton
```
andydotxyz, line 194: It's not clear what the "CancelButton" style is - and no doc to explain

widget/button.go
```diff
@@ -221,6 +231,57 @@ func (b *Button) Tapped(*fyne.PointEvent) {
 	}
 }
 
+// TappedSecondary is called when a secondary pointer tapped event is captured
+func (b *Button) TappedSecondary(*fyne.PointEvent) {
```
andydotxyz, line 235: Why add a no-op?

widget/button.go
```diff
@@ -221,6 +231,57 @@ func (b *Button) Tapped(*fyne.PointEvent) {
...
+
+// TypedRune receives text input events when the Check is focused.
+func (b *Button) TypedRune(r rune) {
+}
+
+func (b *Button) TypedKey(key *fyne.KeyEvent) {
+	if b.Disabled() {
+		return
+	}
+	if key.Name == fyne.KeyReturn || key.Name == fyne.KeyEnter || key.Name == fyne.KeySpace {
```
andydotxyz, line 266: This logic is used multiple times - perhaps it should be de-duplicated?

widget/entry.go
```diff
@@ -247,6 +248,16 @@ func (e *Entry) KeyDown(key *fyne.KeyEvent) {
 		}
 		e.selectKeyDown = true
 	}
+	if key.Name == fyne.KeyEscape {
+		if cancelBtn := e.findButton(CancelButton); cancelBtn != nil {
+			cancelBtn.KeyDown(&fyne.KeyEvent{Name: fyne.KeyEnter})
```
andydotxyz, line 253: This looks like form code in Entry - perhaps you can find a way that better separates the concerns?

widget/entry.go
```diff
@@ -409,6 +432,36 @@ func (e *Entry) TappedSecondary(pe *fyne.PointEvent) {
...
+				btn = b
+				return true
+			}
+			return false
+		}, nil)
+
+	return btn
+}
+
+func (e *Entry) HandleOkCancel() {
```
andydotxyz, line 456: I don't know what this does - looks like it taps 2 found buttons?
Pretty certain it should not be a public Entry API

widget/menu.go
```diff
@@ -18,6 +18,66 @@ type Menu struct {
 	OnDismiss   func()
 	activeItem  *menuItem
 	customSized bool
+	index       int
+}
+
+func (m *Menu) Defocus() {
```
andydotxyz, line 24: I don't think all of these APIs should be public.

widget/menu.go
```diff
@@ -18,6 +18,66 @@ type Menu struct {
...
+	for _,o := range m.Items {
+		if mi, ok := o.(*menuItem); ok {
+			mi.FocusLost()
+		}
+	}
+}
+
+func (m *Menu) Do() {
```
andydotxyz, line 32: What does "Do" mean?

widget/menu.go
```diff
@@ -18,6 +18,66 @@ type Menu struct {
...
+		m.activeItem.child.Do()
+		return
+	}
+	if o, ok := m.Items[m.index].(*menuItem); ok {
+		o.MouseIn(nil)
+		o.Tapped(nil)
+	}
+}
+
+func (m *Menu) Up() {
```
andydotxyz, line 43: Again, naming unclear and not a sensible public API

widget/menu_item.go
```diff
@@ -15,14 +15,33 @@ var _ fyne.Widget = (*menuItem)(nil)
...
 	child           *Menu
 	hovered         bool
+	focused         bool
 	onActivateChild func(*menuItem)
 }
 
+// FocusGained is called when the Entry has been given focus.
+func (i *menuItem) FocusGained() {
+	i.focused = true
+	i.hovered = true
```
andydotxyz, line 29: I don't think hovered should be set on focus - it is a different status

widget/menu_item.go
```diff
@@ -78,8 +97,20 @@ func (i *menuItem) MinSize() fyne.Size {
 // MouseIn changes the item to be hovered and shows the submenu if the item has one.
 // The submenu of any sibling of the item will be hidden.
 // Implements: desktop.Hoverable
-func (i *menuItem) MouseIn(*desktop.MouseEvent) {
+func (i *menuItem) MouseIn(e *desktop.MouseEvent) {
 	i.hovered = true
+	if i.Child() != nil {
+		i.Child().Defocus()
+		if e == nil && i.Child().Items[0] != nil {
```
andydotxyz, line 104: This seems wrong - you are sending "nil" for keyboard events but simulating mouse events - gotta be a cleaner way.

widget/radio.go
```diff
@@ -164,6 +166,58 @@ type Radio struct {
...
+func (s *Radio) TypedKey(key *fyne.KeyEvent) {
+	index := -1
+	for i:=0; i<len(s.Options); i++ {
+		if s.Selected == s.Options[i] {
+			index = i;
+			break
+		}
+	}
+	if key.Name == fyne.KeyLeft || key.Name == fyne.KeyUp {
+		index--
```
andydotxyz, line 201: I don't think that "too far to the left" should mean "no radio selected" 
```go
if index > 1 {
    s.Selected = s.Options[index-1]
}
``` 
is cleaner

widget/select.go
```diff
@@ -107,20 +110,67 @@ type Select struct {
...
+
+func (s *Select) TypedKey(key *fyne.KeyEvent) {
+	if key.Name == fyne.KeyEnter || key.Name == fyne.KeyReturn || key.Name == fyne.KeySpace {
+		if s.popUp==nil || !s.popUp.Visible() {
+			s.Tapped(nil)
+		} else {
+			s.popUp.Menu.Do()
+		}
+	}
+	if s.popUp!=nil {
```
andydotxyz, line 150: With your overlay fix in place this code will not execute as the popup itself will get events?...

widget/select.go
```diff
@@ -107,20 +110,67 @@ type Select struct {
...
+			s.popUp.Menu.Up()
+		} else if key.Name == fyne.KeyDown {
+			s.popUp.Menu.Down()
+		} else if key.Name == fyne.KeyEscape {
+			s.popUp.Dismiss()
+		}
+	}
+}
+
+func (s *Select) KeyUp(key *fyne.KeyEvent) {
```
andydotxyz, line 161: Why adding no-ops?

widget/select.go
```diff
@@ -107,20 +110,67 @@ type Select struct {
...
+}
+
 // Hide hides the select.
 // Implements: fyne.Widget
 func (s *Select) Hide() {
 	if s.popUp != nil {
 		s.popUp.Hide()
 		s.popUp = nil
 	}
-	s.BaseWidget.Hide()
```
andydotxyz, line 123: Removing this seems unwise

widget/select.go
```diff
@@ -145,7 +195,7 @@ func (s *Select) Resize(size fyne.Size) {
 
 func (s *Select) optionTapped(text string) {
 	s.SetSelected(text)
-	s.popUp = nil
```
andydotxyz, line 148: code elsewhere checks for nil of c.popUp, is this safe?

widget/select.go
```diff
@@ -168,9 +218,12 @@ func (s *Select) Tapped(*fyne.PointEvent) {
 		items = append(items, item)
 	}
 
-	s.popUp = newPopUpMenu(fyne.NewMenu("", items...), c)
-	s.popUp.ShowAtPosition(s.popUpPos())
-	s.popUp.Resize(fyne.NewSize(s.Size().Width, s.popUp.MinSize().Height))
+	s.popUp = ShowPopUpMenuAtPosition(fyne.NewMenu("", items...), c, s.popUpPos())
+	if s.popUp != nil {
```
andydotxyz, line 222: how can an object you just created be nil?

widget/select.go
```diff
@@ -168,9 +218,12 @@ func (s *Select) Tapped(*fyne.PointEvent) {
 		items = append(items, item)
 	}
 
-	s.popUp = newPopUpMenu(fyne.NewMenu("", items...), c)
-	s.popUp.ShowAtPosition(s.popUpPos())
-	s.popUp.Resize(fyne.NewSize(s.Size().Width, s.popUp.MinSize().Height))
+	s.popUp = ShowPopUpMenuAtPosition(fyne.NewMenu("", items...), c, s.popUpPos())
```
andydotxyz, line 221: What was wrong with the previous, existing, API calls?

widget/select_entry.go
```diff
@@ -56,16 +56,21 @@ func (e *SelectEntry) SetOptions(options []string) {
 		items = append(items, fyne.NewMenuItem(option, func() { e.SetText(option) }))
 	}
 	e.dropDown = fyne.NewMenu("", items...)
-	dropDownButton := NewButton("", func() {
```
andydotxyz, line 59: cleaner is 
```go
var dropDownButton *widget.Button
dropDownButton = NewButton("", func() {
...
```
so you don' t have to create an empty func then overwrite it

widget/popup_menu.go
```diff
@@ -15,13 +15,47 @@ type PopUpMenu struct {
 	*Menu
 	canvas  fyne.Canvas
 	overlay *widget.OverlayContainer
+	Parent  fyne.CanvasObject
```
andydotxyz, line 18: This parent tracking seems like an anti-pattern - there is need for focus tracking across overlays, remembering the old one and re-instating it after popup removal would be cleaner.

widget/select.go
```diff
@@ -107,20 +110,67 @@ type Select struct {
 	PlaceHolder string
 	OnChanged   func(string) `json:"-"`
 
-	hovered, tapped bool
-	popUp           *PopUpMenu
+	hovered bool
+	focused bool
+	tapped  bool
+	popUp   *PopUpMenu
+	Parent  fyne.CanvasObject
```
andydotxyz, line 117: This does not seem to be used anywhere and should not be needed

widget/tabcontainer.go
```diff
@@ -411,13 +414,78 @@ var _ desktop.Hoverable = (*tabButton)(nil)
...
 	IconPosition buttonIconPosition
 	OnTap        func()
 	Style        ButtonStyle
 	Text         string
 }
 
+// FocusGained is called when the Entry has been given focus.
+func (b *TabContainer) FocusGained() {
+	b.focused = true
+	b.Refresh()
```
andydotxyz, line 429: Refreshing a whole container is expensive (as it invalidates all children).
Here you are not actually using the focused value in the container but in it's tabs.
If the tabs tracked focus and/or the need to refresh it would be much more performant

widget/tabcontainer.go
```diff
@@ -411,13 +414,78 @@ var _ desktop.Hoverable = (*tabButton)(nil)
...
+	if key.Name == fyne.KeyLeft || key.Name == fyne.KeyUp {
+		if b.current > 0 {
+			b.SelectTabIndex(b.current-1)
+		}
+	} else if key.Name == fyne.KeyRight || key.Name == fyne.KeyDown {
+		if b.current < len(b.Items)-1 {
+			b.SelectTabIndex(b.current+1)
+		}
+	}
+	b.Refresh()
```
andydotxyz, line 457: not needed - SelectTabIndex does this

widget/tabcontainer.go
```diff
@@ -20,6 +20,8 @@ type TabContainer struct {
 	OnChanged   func(tab *TabItem)
 	current     int
 	tabLocation TabLocation
+	focused     bool
+	renderer    tabContainerRenderer
```
andydotxyz, line 24: This is a no-no - a widget must not retain a reference to it's current renderer

widget/tabcontainer.go
```diff
@@ -411,13 +414,78 @@ var _ desktop.Hoverable = (*tabButton)(nil)
...
+	} else if key.Name == fyne.KeyRight || key.Name == fyne.KeyDown {
+		if b.current < len(b.Items)-1 {
+			b.SelectTabIndex(b.current+1)
+		}
+	}
+	b.Refresh()
+}
+
+func (b *TabContainer) KeyUp(key *fyne.KeyEvent) {
+	if b.renderer.tabBar==nil {
```
andydotxyz, line 461: Sorry, no - a widget must not manually manipulate the state of it's renderer.
Widgets manage state - renderers draw it.
Surely calling `SelectTab` or similar helper is the right thing to do?

widget/tabcontainer.go
```diff
@@ -463,15 +531,6 @@ func (b *tabButton) Tapped(e *fyne.PointEvent) {
 	b.OnTap()
 }
 
-func (b *tabButton) setText(text string) {
```
andydotxyz, line 466: Please don't rearrange methods unless they are in the wrong order.

widget/toolbar.go
```diff
@@ -23,9 +23,9 @@ type ToolbarAction struct {
 
 // ToolbarObject gets a button to render this ToolbarAction
 func (t *ToolbarAction) ToolbarObject() fyne.CanvasObject {
-	button := NewButtonWithIcon("", t.Icon, t.OnActivated)
-	button.HideShadow = true
-
+	//button := NewButtonWithIcon("", t.Icon, t.OnActivated)
```
andydotxyz, line 26: Please don't commit commented out code - we have version control to track changes.

widget/toolbar.go
```diff
@@ -68,14 +68,102 @@ func NewToolbarSeparator() ToolbarItem {
 // Toolbar widget creates a horizontal list of tool buttons
 type Toolbar struct {
 	BaseWidget
-	Items []ToolbarItem
+	Items    []ToolbarItem
+	focused  bool
```
andydotxyz, line 72: I don't understand why a toolbarButton should not be focused but the toolbar can be?
This code generally seems strange - why would a bar track which of it's non-focusable children is focused?

andydotxyz: In the various apps I have tried it seems that toolbars are not focusable.
If that's not right then we can come back to it.

internal/driver/glfw/menu_bar.go
jkvatne: index is the currently focused item on the menu bar. Using arrow keys, the focus is moved and index is updated. This is also used on toolbars and tabs.

widget/toolbar.go
jkvatne: You are right, but it makes for a lot of Tab presses to navigate through the tab bar. If a toolbar has 10 toolbar buttons, you must press Tab 10 times to go to the first entry in the form. I like it much better with the chosen approach, where the Tab continues to the next object and arrow keys are used to select the actual toolbar button. It is very similar to the tabbed container.

widget/toolbar.go
jkvatne: I know. I just missed it while cleaning up.

widget/tabcontainer.go
jkvatne: This is one of the difficult, deep issues. I could not find any other way to do it.

internal/driver/glfw/menu_bar.go
andydotxyz: So maybe `activeIndex` or `currentMenu` naming then?

widget/toolbar.go
andydotxyz: Looking at more applications I see that toolbars are not focusable at all.
It's like they only focus "input controls" and not every element of the interface.
I don't know if that is right either - but if we are to focus toolbar not items then it should probably appeare focused rather than an item in it?

widget/tabcontainer.go
andydotxyz: From the code that uses it it's hard to see the intent, was it to change buttons or highlight that one had been pressed? if it's the latter then maybe tabButton is the place for that code? otherwise the keyboard handling code can change selected item that the render then updates to reflect.

cmd/fyne_demo/screens/window.go
```diff
@@ -187,7 +187,12 @@ func loadWindowGroup() fyne.Widget {
 		}),
 		widget.NewButton("Fixed size window", func() {
 			w := fyne.CurrentApp().NewWindow("Fixed")
-			w.SetContent(fyne.NewContainerWithLayout(layout.NewCenterLayout(), widget.NewLabel("Hello World!")))
+			w.SetContent(fyne.NewContainerWithLayout(
```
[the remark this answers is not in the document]
jkvatne: It was introduced to test keyboard operation in a second window. There was no extra windows with any focusable widgets, so I had to introduce some buttons to test it. Of cause it failed the first time I tested it, but should be working ok now.

internal/app/focus.go
jkvatne: I use //OBS to easily do a global search for comments that should be removed before commit. This fell through the cracks.
As for but #814, I belive the added lines of code will be a fix for the bug. I am not able to reproduce it.

internal/driver/glfw/menu_bar_item.go
jkvatne: I have changed the name "index" to "currentItem", as it indicates the menu item selected. There is unfortunately a mix of "focused". The focused item receiveing the keyboard data is the MenuBar, while the indicated (focus-colored) item is one of the menu items in the list. Perhaps each menu item should recieve keyboard input?

internal/driver/glfw/menu_bar.go
jkvatne: Just changed it to currentItem

internal/driver/glfw/menu_bar.go
jkvatne: Ok

internal/driver/glfw/menu_bar.go
jkvatne: No, Action() is not possible, I colud of cause duplicate the code or extract it into a new function, but I think the trick of using a nil position is fine for signaling that the event comes from the keyboard and not the mouse. I have used it several places.

internal/driver/glfw/menu_bar.go
jkvatne: ok

internal/driver/glfw/menu_bar_test.go
jkvatne: This relates to my trying to fix the tests. It did not suceed. The problem is the color changes when a button is focused. Previously the buttons did not change color when focused, as it was not possible to focus them. So the menu_bar_initial.png can not be used in every case. It will be different depending on the previous simulated operations. Further changes to the tests must be done later on, introducing new png images. This must wait until the coloring sheme is settled.

internal/driver/glfw/window.go
jkvatne: Unfortunatly it cannot. The toolbar must be focusable by TAB key-presses, but should not be focused when tapped. On the other hand, the code can be simplified as I have now done.

internal/driver/glfw/window.go
jkvatne: With two screens, it is not easy to know which window is focused...

internal/driver/glfw/window.go
jkvatne: Correct. Will fix.

internal/driver/glfw/window.go
jkvatne: Correct - but the simplification was not easy - at least not for me. I an realy impressed that you spotted this so easily.

theme/theme.go
jkvatne: It comes from the material design recomendations. The actual percents to lighten is twice the percent to darken. The shade for focus is 12% darken, but 24% lighten. In this way we need only one constant and the doubling is done in Shade(). I have improved the comments.

theme/theme.go
jkvatne: Correct

theme/theme.go
jkvatne: See comment above about shades.

theme/theme.go
jkvatne: Yes, they are percentages. I belive they should be part of the customizable theme, but I do not see how to do it.

widget/button.go
jkvatne: I dont see how one line would do. There are different colors for all the combinations. But this should be simplified somehow. I will leave this for later.

widget/button.go
jkvatne: This is another hack to implement the functionality of default buttons associated with the enter key and escape key. Normaly, for dialogs there is a default "OK" or "Submit" key that is activated by pressing Enter/Return, and a default "Cancel" or "Abort" key that is activated by pressing "Escape". This is the way I found to distinguish the types. Perhaps there is a better name than "CancelButton", or we could have another property of the buttons. In Delphi there is both Cancel and Default boolean properties that makes the button respond to Enter/Escape. But I think the new style is a better solution.

widget/button.go
jkvatne: My misunderstanding. I have removed it.

widget/button.go
jkvatne: Perhaps. But I do not see any good way to do it.

widget/entry.go
jkvatne: Not sure what you mean here. Which concerns?

widget/entry.go
jkvatne: HandleOkCancel() is a orphaned function leftover from earlier tests. I have removed it.

widget/menu.go
jkvatne: I agree, but this was the only way I found to implement access from internal/driver/glfw/menu_bar_item. Is there any other way?

widget/menu.go
jkvatne: Do is called when the Enter key is pressed, and will simulate tapping. The comments are updated now.

widget/menu.go
jkvatne: See above. I have now tried to improve naming in next version coming soon.

widget/menu_item.go
jkvatne: I agree

widget/menu_item.go
jkvatne: I would also like a cleaner way. But I can not find any, except by duplicating a lot of code.

widget/popup_menu.go
jkvatne: Yes, there should be a way of focus tracking across overlays. When that is implemented, the parent can be removed. (By the way, I have changed Parent to parent)

widget/radio.go
jkvatne: But with the mouse, it is possible to deselect all radio buttons, so this code was added to make it possible using the keyboard alone. But generaly, I think the radio buttons should allways have one and only one button selected all the time. It should not be possible to deselect all radio buttons with the mouse.

widget/select.go
jkvatne: I agree.

widget/select.go
jkvatne: Yes, except for the Enter keys. The code is now simplified.

widget/select.go
jkvatne: Well, I do not know why I added them. Now I have removed them.

cmd/fyne_demo/screens/window.go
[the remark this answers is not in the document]
andydotxyz: All windows should behave the same though?

widget/select.go
jkvatne: Must have been removed by accident. I have replaced it.

internal/app/focus.go
andydotxyz: I agree. In the future it's good to have a bug fix in separate commit - possibly even PR - than a big feature addition

internal/driver/glfw/menu_bar_item.go
andydotxyz: THat might be cleaner, would probably help with code encapsulation, that said left/right may still make sense at the top level.

internal/driver/glfw/menu_bar.go
andydotxyz: By way of explanation here (sorry it was lacking before) we don't add a new exported API unless there is a solid use-case for developers.

widget/popup_menu.go
andydotxyz: In general it is preferable to put in place the correct implementation rather than workarounds waiting for complete solutions.
Of course every rule has an exception - but we'd want to know why this was required instead of fixing the underlying deficiency.

widget/select.go
jkvatne: Not sure. But I changed optionTapped() to be identical to the one on develop, and it is ok.

widget/select.go
jkvatne: My misreading. I thought the call newPopUpMenu was depreciated. But it was NewPopUpMenu that was so.
I have changed back.

widget/select.go
jkvatne: Added the test because I had a panic there in earlier code. Now removed.

widget/select_entry.go
jkvatne: Ok, but it is not a big difference.

widget/tabcontainer.go
jkvatne: Actualy, on the toolbar, this functionality is not needed and has been removed. But on the Toolbar, there is a similar use of the renderer.
The problem is that the toolbar must update the state of the boolbar buttons, as it receives key-presses. But it has no reference to its buttons. To avoid the renderer reference, I have instead added a buttons array in the toolbar.

widget/tabcontainer.go
jkvatne: Yes, it is actualy only the buttons that need to be refreshed. But since the TabContainer have no refrences to the buttons, it is difficult to refresh only the buttons. I will leave this for later.

widget/tabcontainer.go
jkvatne: Ok. Did not see that earlier.

widget/tabcontainer.go
jkvatne: This has been fixed now

widget/tabcontainer.go
jkvatne: I have tried to change as little as possible on the existing code, but is there a description of what is correct order?
Anyway, I have moved setText() back to where if was.

widget/toolbar.go
jkvatne: Usualy the toolbar buttons are associated with keyboard shortcuts, and then there is no need to focus them. To do this we need an obligatory shortcut for each toolbar button. I leave this for future updates.

widget/tabcontainer.go
andydotxyz: I don't think we have it written up the wiki yet, the information is in issue #827

widget/toolbar.go
andydotxyz: That is a good idea, apart from the mandatory part. Some work for Shortcuts is in place, hopwfully we can make use of that.

widget/tabcontainer.go
jkvatne: Oops. Is there a way to force detachement of all renderers, so this can be tested?

widget/radio.go
andydotxyz: There has been some discussion in this area, both radio and select could do with an API that makes it explicitly possible or not to unselect

widget/menu_item.go
andydotxyz: Why does the code need to be duplicated? Can’t common code be extracted into separate methods that are nicely named to make it clearer?

widget/menu.go
andydotxyz: If the menus are focusable then the usual focus methods should work - I’m not sure I understand what things the menu bar needed to signal into menus.

widget/entry.go
andydotxyz: Form and entry are separate. If we want to submit a form on entry press we need to figure some way for that code to exist in Form not in Entry. It’s not an easy problem to solve, but doing it this way would need form logic inside every input widget which is what I mean about not separating concerns? (https://en.m.wikipedia.org/wiki/Separation_of_concerns)

widget/button.go
andydotxyz: A private function called ‘isSubmitKey(key)’ perhaps?

widget/button.go
andydotxyz: Is this specific to forms as well?

internal/driver/glfw/menu_bar_test.go
andydotxyz: Please use appropriate naming instead of adding a number - it should be clear of the image content/state based on the name. We can’t have failing tests waiting for another PR or external change, if this PR depends on another then we have to land that first before this can complete.

widget/tabcontainer.go
andydotxyz: Currently we do not have that test capability, but it’s a good suggestion.

andydotxyz:
Looks like there is still a format issue in there, plus lots of test failures.

andydotxyz:
Hi @jkvatne I think this would be much easier to review and discuss in detail if you broke it into smaller PRs, this currently contains a lot of different areas of change and in our experience such changes can take exponentially longer than items that focus on one area at a time. There may be a dependency ordering consideration as well.
What do you think?

jkvatne:
Sure, you are correct. I will try to split it more. I was a bit optimistic, but now all tests passes so it is not so bad.
There is one thing that should be descided soon, and that is the coloring sheme to differentiate between primary color and focused color. Also a "tapped" or "pressed" color should be introduced. This relates very much to WIP - Full material design #970, so I will try to look into what steveoc64 has done.

andydotxyz:
This PR has a growing list of conflicts as the codebase moves on - what should we do with it @jkvatne ?
It might be easiest if smaller pieces were added - perhaps addressing one widget at a time?

toaster:
> It might be easiest if smaller pieces were added - perhaps addressing one widget at a time?

@andydotxyz I think that would definitely the way to go.
Any bugfix/refactoring for keyboard handling in overlays should be separated, too.
As I am beginning to work on #814, I think it would even be a good idea to have such fixes extracted first :).

internal/driver/glfw/window.go
```diff
@@ -15,6 +15,7 @@ import (
 	"fyne.io/fyne/internal/cache"
 	"fyne.io/fyne/internal/driver"
 	"fyne.io/fyne/internal/painter/gl"
+	"fyne.io/fyne/widget"
```
toaster, line 18: We must not add dependencies from `widget` here.

jkvatne:
I am sorry, but I do not have much time now for this prosject. I am an independant consultant, and have suddenly a lot of paying customers that demands my time. Perhaps a bit later.
But one of the big problems I find, is that the coloring scheme is insufficient. The primary color is used both for focused items and for primary buttons. You need to differentiate between primary/secondary and focused/hovered/nonfocused. Google uses overlays with different shades to indicate hovered/focused/pressed/seelcted, so this might be a way of doing it.
image

But buttons might be the simplest items. All other widgets that have manual interaction must also be focus-able, and indicate state. Menus also.

So I would like you to first find a way to calculate colors/shades that is widget independant, and with consistent coloring for different widgets. This is beyond my ability. I tried, but you didn't like it.

andydotxyz:
Yes making such an addition will break the theme API so we need to hold it back for 2.0 - but then we can get alt-background and highlighted colours etc

widget/toolbar.go
vmcrash: IMHO, a toolbar (or containing buttons) **never** should be focusable. Toolbars are for easy clicking with the mouse. All commands available in a toolbar should also be available as a menu item in the menu bar - hence keyboard-only users already use the menu or its accelerator.

jkvatne: I agree. But it depends on having keyboard shortcuts. And the menu must be focusable.

andydotxyz: It would be great to get this split into smaller pieces so we can see some of this functionality land - the conflict list is probably too long to get this PR landed in one go.

andydotxyz:
Radio (-> RadioGroup) and Select widgets have been taken care of by other work.

andydotxyz:
There is a lot of good work here and we should take it all into account as we move forward.
However there is a lot of conflict against base now and many widgets have had keyboard operation added separately.

I'm going to close this, if you find that time becomes available then let's look at how to fill in the blanks of the updated keyboard support.
Mostly I think we could do with feed-in on the theme updates. They are proposed on the wiki at [Theme2](/fyne-io/fyne/wiki/Theme2) and should cover things like the shade code added here, but built into the API for 2.0.

PR #1075, opened 2020-06-03
Address crash with empty main menu

### LINKED ISSUE #1073: Empty main menu causes panic
Setting the window's main menu to an empty main menu will cause a panic on glfw.
This line will cause a panic:
`w.SetMainMenu(fyne.NewMainMenu())`
The panic is:
`panic: runtime error: index out of range [0] with length 0`

This happens with the glfw driver on the develop branch.

The panic is caused by a line in [https://github.com/fyne-io/fyne/blob/develop/internal/driver/glfw/menu.go](this) file. The line begins with,

`if menus.Items[0].Items[len(menus.Items[0].Items)-1].Label`

This assumes that menus.Items has an entry at index zero.

andydotxyz:
I don't think this should reset release vote as it's minor, but if anyone objects I will hold off for 1.3.x

Fixes #1073

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


internal/driver/glfw/menu.go
```diff
@@ -5,7 +5,15 @@ import (
 )
 
 func buildMenuOverlay(menus *fyne.MainMenu, c fyne.Canvas) fyne.CanvasObject {
-	if menus.Items[0].Items[len(menus.Items[0].Items)-1].Label != "Quit" { // make sure the first menu always has a quit option
+	if len(menus.Items) == 0 {
+		fyne.LogError("Main menu must have at least one child menu", nil)
```
stuartmscott, line 9: I don't think this should be an error since an empty menu will get the default "Quit" option.

PR #1086, opened 2020-06-08
Fix Label Refresh error

morpheus747:
### Description:
The custom widget label don't refresh changes until you resize the application window

Fixes 

### Checklist:

- [ ] Tests included.
- [x ] Lint and formatter run with no errors.
- [ ] Tests all pass.

#### Where applicable:


andydotxyz:
Thanks for helping, but this is a subset of the previous PR #1085, so the issue is already resolved.

PR #1108, opened 2020-06-15
File dialog locations

### LINKED ISSUE #821: FileDialog: Set start directory
 - It's kinda annoying to always navigate the file system to the working dir to load the (test) data for my program.
 - If I want to specify a location to save a file to, I'd like my program to have a default location, which is not the $HOME dir (could be a specific dir or the dir from which the initial data was loaded)

It would be nice to set the (starting) dir of the FileDialog by calling a function or as a parameter for ShowFile(Open|Save). Or it could be specified with the Preferences-API.

charlesdaniels:
### Description:

Fixes #821 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

- ~~[ ] Any breaking changes have a deprecation path or have been discussed.~~ (N/A)


charlesdaniels:
I'm not sure how best to test this. Based on looking at `file_test.go`, it seems there wouldn't be a good way to get at the `FileDialog` object from the perspective of the test case. Even if we could, it there is no way to know what directories might exist on the system running the tests.

Aside from the uncertainty about testing, I think this can be considered for merging, pending any other feedback.

andydotxyz:
The added complexity in the function to show should probably be factored out into a new function. If that was responsible for looking up starting directory that probably also answers how can it be tested :).

One stylistic thought - SetStartingDirectory could be omitted if the startingDirectory field was exported - would that be better or not? :)

andydotxyz:
Im also not sure about the “where” string parameter to ShowFileOpenAt... - it’s not a very descriptive variable name but possibly more of a question is, should it not be a URI?

Jacalz:
I agree with Andy and I think that a better name for the “where” variable might be “startdir” or something along the lines. Apart from that, I think that it all looks really good 👍 Good job and congratulations for making your contribution to Fyne 🙂

charlesdaniels:
I agree with the proposed changes by @andydotxyz and @Jacalz and have implemented them in db857e1bf46ec5eaa2d8eb4020b16a05ffc83d93, I will add tests presently. Thank you for the feedback! Let me know if you think of anything else.

charlesdaniels:
Tests are implemented as of 6fbdb2fe, please let me know if there is further feedback.

charlesdaniels:
Seems the build is failing do to inconsistent receiver names. To my knowledge, this was not caused by my code changes. I can fix this issue if desired, but I think it is out of scope for this PR.

andydotxyz:
It looks like it is your new method - you used the reciever `file` instead of `f`.
An odd test failure can be chalked up to random - but here 3/4 of them are failing, probably indicates taking a closer look ;)

dialog/file.go
```diff
@@ -35,13 +35,14 @@ type fileDialog struct {
...
-	dialog           *fileDialog
-	dismissText      string
+	save              bool
+	callback          interface{}
+	onClosedCallback  func(bool)
+	filter            storage.FileFilter
+	parent            fyne.Window
+	dialog            *fileDialog
+	dismissText       string
+	StartingDirectory string
```
andydotxyz, line 45: I think we need a URI based approach really, paths may not work across platforms

charlesdaniels:
This is blocked while we determine how best to handle a URI based approach as @andydotxyz mentioned. I have opened #1111 to merge just changing the start directory to `./` for now.

andydotxyz:
It is blocked I agree - but probably worth fixing the reciever name and probably rebasing on top of the merged #1111 in preparation :)

charlesdaniels:
I have "rebased" (this was simple enough I just cherry picked) on top of the work one in #1111.

andydotxyz:
Strange, that Windows failure was a keyboard test. I have restarted it to see if the issue is persistent.

andydotxyz:
Woops, this should be off `develop`, I think you can just edit the PR, not certain...

andydotxyz:
Just to put some thoughts here from a discussion we had.

* Clearly we want to be using URIs so to support the same cross-platform strategies that opening/saving files does
* If the system is using URIs for location it realistically can't rely on file system lookups for getting a directory listing
* Therefore we probably need to convert the file dialog to use URI functionality only internnaly, which leads to the challenge that we cannot list URIs currently.

Proposal / suggestion (This was my takeaway from our discussion - I know there were alternatives posed - this is to keep the discussion going).

Create a new type that is able to list children from a given URI:

```
type ListableURI interface {
    URI
    ListURIs() []URI
}
```

OK the naming probably wants some thought, but using this approach we could identify "directories" and list them like so:

```
dir, isDir := myURI.(fyne.ListableURI)
...
for _, child := range dir.ListURIs() {
...
}
```

And the new `Location` field of the `FileDialog` could be of type `fyne.ListableURI`.

andydotxyz:
Seems like a good time to get this updated and ready to merge? :) @charlesdaniels

charlesdaniels:
This once again raises the issue of listable URIs in the test driver, since I can't test this PR without being able to create URIs, for obvious reasons. How should I proceed? We could simply restore the copy-pasted code from before? Or maybe the file-based URI handling as used in the glfw driver should be moved somewhere in `internal/` and called from both places?

andydotxyz:
I'm not sure that I follow the obvious requirement to copy the behaviour of GLFW here - we are in test land, so why not test with some specific, fake, listableURI.

I.e. a completely in-memory mock heirarchy coded into a dummy listable implementation? Easier and faster for tests :)

charlesdaniels:
It has just occurred to me that, like `uri.Parent()`, we also need a `uri.Join()` or something to that effect. I can implement this in `storage/uri.go` for the time being, but we should consider making this part of the URI interface in 2.0.

charlesdaniels:
I think I have this at a stage where I would like more commentary. These are the issues that I think warrant discussion:

* The code still pretty heavily assumes that `file://` is the only scheme that exists. It definitely will not work for other types of URIs yet. If we wanted to fix this, we would need two wrap at least `os.Stat()` in a more URI-friendly way. I think this issue is OK, but worth noting.

* I have moved the test URI driver to use the glfw list-able URI implementation again. I don't see how we can actually test any of this stuff without doing so. It can't really be made any simpler and still use the real filesystem APIs. If we mocked it out instead, `os.State()` and probably other sundry things would fail. I don't think this is a surmountable problem without further work on URIs. I understand the previous reluctance to re-using the GLFW lister code, I believe it is warranted in this case.

* I have added a "Join" function in `storage/uri.go`, which I think should be added to the URI interface itself in 2.0, by the same logic as `Parent()`. 

* The code as implemented fails tests because it does not include file extensions in the names displayed in the dialog. I think that hiding file extensions is an anti-pattern, especially on non-Windows platforms where it is not the default. I would like to take this opportunity to advocate for showing the file extensions by default, and possibly considering a setting to show or hide them (maybe matching the platform setting by default) in the future.

andydotxyz:
I think that Stat() is only needed for the "does it exist" check before saving a file...
So if we had URI.Exists() we could use that right?
In the meantime why not put this exists check in an `if Scheme() == "file"` then we could keep that behaviour without requiring file?

andydotxyz:
> * I think that hiding file extensions is an anti-pattern

We're not actually hiding them, we have moved them to be over the icon.
We did this because almost all file names are too long to fit at the font size we are using! Happy to hear clever solutions :)

dialog/file.go
```diff
@@ -78,15 +82,17 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
...
 			callback := f.file.callback.(func(fyne.URIWriteCloser, error))
 			name := f.fileName.(*widget.Entry).Text
-			path := filepath.Join(f.dir, name)
+			path := storage.Join(f.dir, name)
 
-			info, err := os.Stat(path)
+			// this assumes the file:// type, which is enforced
+			// when we set `dir`
+			info, err := os.Stat(path.String()[len(path.Scheme())+3:])
 			if os.IsNotExist(err) {
```
andydotxyz, line 90: I think this is the crux of Stat() requirements. We could just check existence when Scheme is file and I think that removes our relience on file URI...

dialog/fileitem.go
```diff
@@ -24,7 +23,7 @@ type fileDialogItem struct {
 
 	icon fyne.CanvasObject
 	name string
-	path string
+	path fyne.URI
```
andydotxyz, line 26: Calling this path makes some of the usages look strange - is there another name that could work?

dialog/fileitem_test.go
```diff
@@ -15,34 +16,34 @@ func TestFileItem_Name(t *testing.T) {
 	f := &fileDialog{file: &FileDialog{}}
 	_ = f.makeUI()
 
-	item := f.newFileItem("/path/to/filename.txt", false)
+	item := f.newFileItem(storage.NewURI("file:///path/to/filename.txt"), false)
```
andydotxyz, line 19: Did we agree to storage.NewFileURI()? this make it look pretty darn useful :)

storage/uri.go
```diff
@@ -109,3 +109,15 @@ func Parent(u fyne.URI) (fyne.URI, error) {
 	parent = parent + strings.Join(components[0:len(components)-1], "/") + "/"
 	return NewURI(parent), nil
 }
+
+// Join appends a new path element to a URI, separated by a '/' character.
+func Join(u fyne.URI, component string) fyne.URI {
```
andydotxyz, line 114: I know when discussing 2.0 URI we agreed that Child() would make sense, if there is an intermediate method that is public can it be consistent - something like storage.ChildURI()?

andydotxyz (changes requested):

dialog/fileitem.go
charlesdaniels: Maybe `parent`?

dialog/fileitem_test.go
charlesdaniels: We didn't agree to that in particular, but I would be happy to add it. I agree it would come in quite handy.

storage/uri.go
charlesdaniels: I think we already have `storage.Parent()`, so I would be inclined to go with `storage.Child()` to be consistent with that, moving it into the URI interface in 2.0. This would also apply to `storage.Exists()`.

charlesdaniels:
> We're not actually hiding them, we have moved them to be over the icon.

Well OK, fair enough. Maybe we could display `longfilename.txt` as `long⋯.txt`?

andydotxyz:
> Well OK, fair enough. Maybe we could display `longfilename.txt` as `long⋯.txt`?

Let's pick that up on a separate thread?

storage/uri.go
andydotxyz: Yeah, that sounds like a plan

dialog/fileitem_test.go
andydotxyz: Ah yes I remember now there was potential confusion because we use the word File in other URI places that might need tidying up...
Your call

dialog/fileitem.go
andydotxyz: Isn't it the item URI rather than the parent?
id, location, res or others like that maybe

charlesdaniels: I went with location in 78c6984e. I think we should start moving from `path` to `location/loc`. I could probably be sold on `reference`/`ref` if someone feels strongly.

dialog/fileitem_test.go
charlesdaniels: I agree that we should do this, but I think that is out of scope for this PR.

dialog/file.go
charlesdaniels: I added an `Exists()` function in 78c6984e497079ae24611d2496a576b2237cd726.

charlesdaniels:
This is causing a CI failure for Windows target: https://travis-ci.org/github/fyne-io/fyne/jobs/726752792#L139

In 2.0, this would be easy to solve since the `Parent()` for Windows style paths will know to use `\` as the path separator. I'm not sure how to solve this, in that I know Windows supports both `/` and `\` as path separators. It never occurred to me to test the `\` case. I would be concerned about simply splitting on the platform path separator, since either can be valid. Perhaps we should split on both `/` and the platform path separator? Maybe someone more familiar with Windows can comment?

I am surprised this didn't fail when `Parent()` was first introduced, but I guess it wasn't directly used by any of the Windows code.

andydotxyz:
For now I would say that switching on GOOS should be OK. Otherplases use windows file paths as the other slash and it works fine.
When using things like MSYS it will do the translation I think.
Unless I misunderstood what you meant by both are possible @charlesdaniels?

charlesdaniels:
AFAIK, you *can not* mix and match both in the same path, but you *can* use either or. This has been around for a while, and is for UNIX compatibility. It's not unimaginable that a developer might give us a URI with forward slashes on Windows, even with a drive letter at the start of the path.

Maybe my understanding is wrong, but I think we need to handle both.

I would propose that we should  check the OS, and if it is Windows, split on `\` if any appear in the URI, and otherwise split on `/`. Or perhaps it should be the other way around, so as to prefer using `\` in the case where the path has only one element so far.

`Child()` would use `\` on Windows unless the path already contains `/`, in which case it would do that (`C:/foo/bar` is valid, but `C:/foo/bar\baz` is not, AFAIK).

charlesdaniels:
It just occurred to me that the better approach would be to check if the scheme is `file://`, and if so `path/filepath` for path operations, failing over to the current implementation otherwise, since most other types of URIs should work with the current implementation.

Once 2.0 hits, each individual type of URI will be able to specify it's own implementation, so `file://` can directly use `filepath` unconditionally, and perhaps things like HTTP can use libraries that cater to them specifically under the hood. 

Unless there is an argument against, I am convinced this is the way to go.

andydotxyz:
Just to clarify here... when using URI there are not two possible slash formats - file:// URI should match https:// whereby all path separators are "/". 
The issue may be where we are converting to a local path by simply stripping "file://" so in windows if we also switch "/" for "\\" that might solve the issue, but it's worth noting that encoding is an issue as well - i.e. %20 for spaces etc we may be ignoring mistakenly...

charlesdaniels:
Ahh fair enough. So then here is what I would propose:

1. If the OS is Windows, replace `/` with `\` int he URI.
2. Remove the scheme.
3. Perform the requested path operation via the `filepath` module.
4. If the OS is Windows, replace `\` with `/` in the resulting path.
5. Re-add the scheme.

charlesdaniels:
I have written the code to use the `filepath` module for `file://` type URIs, and the existing implementation for other schemes. This seems to work on Linux, but the Windows tests are failing. It seems that the string replacement isn't working, or the `runtime.GOOS` check isn't working for one reason or another.

I'm afraid I can't really debug this. I don't have a working Windows development environment. I think this is blocked until someone with a working Windows setup can take a look. I think that this is ready to go aside from not working on Windows though.

charlesdaniels:
For reference, [this](https://github.com/fyne-io/fyne/blob/d46f9ff9eb4de61cd464332d9001171387c292fc/storage/uri.go#L133) is the problematic code. I think it should be easy enough for someone who has worked more with Go on Windows, and who actually has a working development environment for it to patch up.

PucklaMotzer09:
I will take a look at it. I have got **windows** environment. What are the tests I should take a look at?

PucklaMotzer09:
> For reference, this is the problematic code. I think it should be easy enough for someone who has worked more with Go on Windows, and who actually has a working development environment for it to patch up.

This code works as expected. The OS gets detected correctly and the slashes get replaced. **TestShowFileSave** fails because the  URI with only forward slashes gets compared with the chosen URI you get from the dialog. The URI of the dialog does have backwards slashes and so the string comparison fails.

PucklaMotzer09:
[This line](https://github.com/fyne-io/fyne/blob/d46f9ff9eb4de61cd464332d9001171387c292fc/storage/uri.go#L107) seems be the reason why **TestURI_Parent** fails. This also needs to check for backslashes if the OS is windows.

PucklaMotzer09:
So, the only remaining failing test is **TestShowFileSave**. To fix it you either need to change the test to also accept backslashes or you replace the backslashes with forwardslashes when a file is selected.

PucklaMotzer09:
I just created a test application to test if **StartingLocation** works. This seems to work, but when I choose a root directory (eg. C:\) **nothing** is shown and an error is printed to the console that says *Cannot take the parent of the root element in a URI*. This happens even if I don't use **StartingLocation**.

PucklaMotzer09:
> I just created a test application to test if StartingLocation works. This seems to work, but when I choose a root directory (eg. C:) nothing is shown and an error is printed to the console that says Cannot take the parent of the root element in a URI. This happens even if I don't use StartingLocation.

I fixed this by changing [these lines](https://github.com/charlesdaniels/fyne/blob/file_dialog_locations/dialog/file.go#L232-L242) to this:
```go
parent, err := storage.Parent(dir)
	if err != nil && err != storage.URIRootError {
		fyne.LogError("Unable to get parent of "+dir.String(), err)
		return
	}
	if parent != nil && parent.String() != dir.String() {
		fi := &fileDialogItem{picker: f,
			name: "(Parent)", location: parent, dir: true}
		fi.ExtendBaseWidget(fi)
		icons = append(icons, fi)
	}
```

PucklaMotzer09:
I created a [Pull Request](https://github.com/charlesdaniels/fyne/pull/1) which fixes all failing tests on windows. In my version the save and open dialogs always return paths with **/** instead of **\\**. Because of that I had adjust the test **TestShowFileOpen**. I also added that the **setDirectory** function returns when a *nil* directory gets opened. This was an issue for me, because when I tried to open the volume that corresponds to my disk drive the application would crash, since Fyne couldn't create a ListableURI.

andydotxyz:
Yes, the URIs should never have windows separators in them - a file URI needs to be something that we can share irrespective of operating system.

Of course the varying locations of <Home> and <Documents> is a fun challenge for the future :)...

charlesdaniels:
@PucklaMotzer09 

Thanks for checking this out. However as @andydotxyz the code I thought was broken was actually a red herring -- we need to make sure that any Windows-style paths get correctly changed to use `/` everywhere they are ingested, a change that I plan to implement.

Your PR looks good, but I was actually planning to move this inside of `NewURI`. I will merge it and make the needed tweaks on my end.

Thanks!

storage/uri.go
```diff
@@ -59,53 +61,131 @@ func (u *uri) String() string {
...
+		}
+	}
 	if s[len(s)-1] == '/' {
 		s = s[0 : len(s)-1]
 	}
 
-	// trim the scheme (and +1 for the :)
-	s = s[len(u.Scheme())+1:]
+	// trim the scheme
+	s = s[len(u.Scheme())+3:]
```
andydotxyz, line 117: Why +3? the comment about the +1 has gone so it's not clear?
If it's for :// then beware that the "//" is optional in URI spec

storage/uri.go
```diff
@@ -59,53 +61,131 @@ func (u *uri) String() string {
...
 // removing the last item.
 func Parent(u fyne.URI) (fyne.URI, error) {
 	s := u.String()
 
 	// trim trailing slash
+	if runtime.GOOS == "windows" {
+		if s[len(s)-1] == '\\' {
+			s = s[0 : len(s)-1]
+		}
+	}
```
andydotxyz, line 111: Should this not be in an else? Would a windows path with / at the end  expect to have it truncated?

storage/uri.go
```diff
@@ -59,53 +61,131 @@ func (u *uri) String() string {
...
+		// pointed to a UNIX-style root
+		if len(location) == 0 {
+			return "", URIRootError
+		}
+	}
+
+	components := strings.Split(location, "/")
+
+	if len(components) == 1 {
+		// trimmed <= 2 makes sure we handle UNIX-style paths on
```
andydotxyz, line 84: This comment seems out of date somehow?

storage/uri_test.go
```diff
@@ -52,23 +53,37 @@ func TestURI_Parent(t *testing.T) {
 
 	parent, err := storage.Parent(storage.NewURI("file:///foo/bar/baz"))
 	assert.Nil(t, err)
-	assert.Equal(t, "file:///foo/bar/", parent.String())
```
andydotxyz, line 55: The comment above notes the importance of trailing slashes, but this alters that behaviour...
I would tend to agree with the above and that a fileURI.Parent() would always end in a '/'.

storage/uri.go
andydotxyz: And if this method is explicitly for handling non-file URIs  why is windows path handling in here?

storage/uri.go
```diff
@@ -59,53 +61,131 @@ func (u *uri) String() string {
 	return u.raw
 }
 
+// parentGeneric is a generic function that returns the last element of a
+// path after splitting it on "/". It should be suitable for most URIs.
+func parentGeneric(location string) (string, error) {
```
andydotxyz, line 66: It seems like this has not got tests because we're testing only the file version

dialog/file_test.go
```diff
@@ -152,13 +203,21 @@ func TestShowFileOpen(t *testing.T) {
...
-	assert.Equal(t, filepath.Base(target.path), nameLabel.Text)
+	assert.Equal(t, target.location.Name(), nameLabel.Text)
 	assert.False(t, open.Disabled())
 
 	test.Tap(open)
 	assert.Nil(t, win.Canvas().Overlays().Top())
 	assert.Nil(t, openErr)
-	assert.Equal(t, "file://"+target.path, chosen.URI().String())
+
+	targetLocationString := target.location.String()
```
andydotxyz, line 213: This should not be needed as you are comparing a URI with a URI

dialog/file.go
```diff
@@ -77,17 +82,29 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 		if f.file.save {
 			callback := f.file.callback.(func(fyne.URIWriteCloser, error))
 			name := f.fileName.(*widget.Entry).Text
-			path := filepath.Join(f.dir, name)
+			path, _ := storage.Child(f.dir, name)
+
+			// On windows replace '\\' with '/'
+			if runtime.GOOS == "windows" {
+				pathString := path.String()
```
andydotxyz, line 89: path is now a URI so this should not be needed

dialog/file.go
```diff
@@ -151,54 +176,100 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
...
+		favoritesGroup, header, footer, body)
+
 }
 
-func (f *fileDialog) loadFavorites() []fyne.CanvasObject {
-	home, _ := os.UserHomeDir()
+func (f *fileDialog) loadFavorites() ([]fyne.CanvasObject, error) {
+	osHome, err := os.UserHomeDir()
+	if err != nil {
+		return nil, err
```
andydotxyz, line 197: For this and all those below I don't think we should bomb out if one fails - let's add the favourites that we can and return the last error along with what we build...

dialog/file.go
```diff
@@ -208,13 +279,27 @@ func (f *fileDialog) refreshDir(dir string) {
...
+	}
+
 	f.setSelected(nil)
 	f.dir = dir
 
 	f.breadcrumb.Children = nil
-	buildDir := filepath.VolumeName(dir)
-	for i, d := range strings.Split(dir, string(filepath.Separator)) {
+
+	if dir.Scheme() != "file" {
```
andydotxyz, line 292: Is this required? it seems like the dialog is abstracted now...

andydotxyz (changes requested):
Great work and really close.
A couple of things inline about tests mostly.
Also some windows code that looks like it can be removed and we're forcing file:// for filedialog but is that really needed now?

dialog/file.go
```diff
@@ -97,11 +114,11 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 				func(ok bool) {
 					f.win.Hide()
 					if !ok {
-						callback(nil, nil)
+						// callback(nil, nil)
```
Jacalz, line 117: Is this supposed to be commented out?

dialog/file.go
```diff
@@ -245,39 +338,85 @@ func (f *fileDialog) setSelected(file *fileDialogItem) {
...
+				fyne.LogError("Error with StartingLocation", err)
+			} else {
+				return startdir
+			}
+		}
+
+	}
+
+	// Try to get ./
+	wd, err := os.Getwd()
```
Jacalz, line 394: My view is still that HomeDir() should be preferred over PWD. I think that this also was what we agreed on when talking about this in earlier discussions because PWD makes more sense for the developer than what it does for a user. A user of the application would probably be more happy to have Home than PWD.

Jacalz (changes requested):
Mostly a commented out callback that I think might be a mistake and the change to default starting location being PWD. I think we agreed on HomeDir being default as it makes more sense for users and that PWD mostly makes more sense to developers.

dialog/file.go
squash: Just a comment from a user.. When an app is launched from an icon, it isn't obvious what PWD would be. This may be as simple as it being documented somewhere but being able to rely on $HOME would be straight forward.

Jacalz: My proposal would be to have HomeDir the default and then developers wanting PWD could just put in a temporary different starting directory now that we, with this PR, have support for changing starting directory.

stuartmscott:
@charlesdaniels consider using https://golang.org/pkg/strings/#TrimPrefix to remove the `scheme://` instead of slicing. It may require a few invocations (1 to remove the scheme, 1 to remove the colon, and 1 to remove the optional double forward slash) but I think it could improve readability.

dialog/file.go
charlesdaniels: This was a debate we had previously in #1111, at which point we decided that ./ would be the best starting location.

charlesdaniels:
Most of the comments raised here seem reasonable, although this wasn't ready to review yet (sorry, I should have set it back to draft when the Windows problems cropped up). Many of these were things I had intended to tidy up. I will look through in more detail when I get a chance.

storage/uri_test.go
```diff
@@ -82,3 +97,11 @@ func TestURI_Extension(t *testing.T) {
...
 	assert.Equal(t, "", storage.NewURI("file://C:/directory/").Extension())
 }
+
+func TestURI_Child(t *testing.T) {
+	p, _ := storage.Child(storage.NewURI("file:///foo/bar"), "baz")
+	assert.Equal(t, "file:///foo/bar/baz", p.String())
+
+	p, _ = storage.Child(storage.NewURI("file:///foo/bar/"), "baz")
+	assert.Equal(t, "file:///foo/bar/baz", p.String())
+}
```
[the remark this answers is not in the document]
PucklaMotzer09: Shouldn't URIs always have ```/``` instead of ```\```? If so then this test should work for all platforms.

PucklaMotzer09:
@charlesdaniels Could you give me write access to your fork so that I can directly commit to it.

storage/uri.go
PucklaMotzer09: > beware that the "//" is optional in URI spec

We are using ```len(Scheme())+3``` everywhere. There should be some functions like ```Path()``` which returns only the path of an URI to be complient with that.

storage/uri_test.go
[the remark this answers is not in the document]
PucklaMotzer09: I added a test that uses an URI with ```\``` as input.

storage/uri_test.go
```diff
@@ -52,23 +53,37 @@ func TestURI_Parent(t *testing.T) {
...
 
 	parent, err = storage.Parent(storage.NewURI("file://C:/foo/bar/baz/"))
 	assert.Nil(t, err)
-	assert.Equal(t, "file://C:/foo/bar/", parent.String())
+	assert.Equal(t, "file://C:/foo/bar", parent.String())
+
+	if runtime.GOOS == "windows" {
+		// Only the Windows version of filepath will know how to handle
+		// backslashes.
+		parent, err = storage.Parent(storage.NewURI("file://C:\\foo\\bar\\baz\\"))
```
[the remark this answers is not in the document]
PucklaMotzer09: I seperated it.

storage/uri.go
PucklaMotzer09: I removed the check for windows since all URIs now have ```/```

storage/uri.go
PucklaMotzer09: I removed the comment

storage/uri_test.go
PucklaMotzer09: I changed it so that ```Parent()``` always returns a path with a trailing ```/```

storage/uri.go
PucklaMotzer09: I added multiple tests for this functions

dialog/file_test.go
PucklaMotzer09: I changed it back since now all URIs just have ```/```

dialog/file.go
PucklaMotzer09: I changed it back

dialog/file.go
PucklaMotzer09: I rewrote the function. Now it always runs until the end and returns the last error.

dialog/file.go
PucklaMotzer09: I removed the check. ```StartingLocation``` is ignored if it doesn't have ```file``` as scheme.

PucklaMotzer09:
I made some changes, but I want to directly push them to this fork.

storage/uri_test.go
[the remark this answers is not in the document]
andydotxyz: Yes, the slashes should always be the same, but I thought it was worth testing with the file://C:/ prefix version for windows

storage/uri.go
andydotxyz: But this code is improving the very way that URIs are being parsed and created - so we should be trying to improve rather than replicating the same issues in the previous implementations :).

The reason that Path() was not added immediately iis that if people thought this was just about files then they might do os.File operations on the result of URI.Path(), which is often not going to work.
In the future we can add it, once we're cleanly using abstractions in our own code. But this is a breaking change as the URI interface is public so it would not happen before 2.0.

storage/uri.go
andydotxyz: Is that the right fix? If a developer thought a comment was needed then surely there is a wider issue at play?

dialog/file.go
andydotxyz: Why is it ignored? can we not allow nfs:// or ftp:// URIs in the directory picker?

PucklaMotzer09: > can we not allow nfs:// or ftp:// URIs in the directory picker?

I don't know. That's just the way @charlesdaniels implemented it.

storage/uri.go
PucklaMotzer09: I thought the same, but I couldn't think of a proper comment, since I don't even know why the comment was there in the first place.

storage/uri.go
PucklaMotzer09: What about ```path()``` for internal use?

andydotxyz:
Thanks for all the contributions @PucklaMotzer09, it's a great help.
I think that most developers around here use their own forks as only write access for themselves - the only collaborative spaces are really the upstream develop and occasional feature branches.
I know that @charlesdaniels is very busy this week but I am sure he will pull your changes into his branch soon so the result will be the same as if you'd written directly.
We're still a few weeks away from the proposed time of our next release so there is plenty of opportunity to get this all wrapped up in plenty of time :).

dialog/file.go
Jacalz: We discussed this at today’s call and we (@stuartmscott, @andydotxyz and me) concluded that HomeDir should be the default as it is “the least likely to surprise the user” and having PWD can easily be toggled with few lines of code now that this work has landed.

storage/uri.go
charlesdaniels: I have pitched the idea of having a standardized method for retrieving the parts of the URI that aren't the scheme in the past. I still think it would be a good ideas as @PucklaMotzer09 mentioned, even if it was only for internal use. We could also put it in `storage` like `Parent` for now, if we wanted it to be public.

storage/uri.go
charlesdaniels: Yeah, I think this is not a relevant comment anymore. This was part of the logic for handling Windows drive letters.

storage/uri.go
charlesdaniels: Yes, I think a trailing `\` is no longer relevant if we are normalizing everything to `/`.

storage/uri.go
charlesdaniels: Yeah, I never got around to testing it explicitly after I switched to using `filepath`. Thank you @PucklaMotzer09

dialog/file.go
charlesdaniels: I wrote it that way since we don't actually have support for those things yet, and at the time I was under the impression that it would just cause it to break later. That might not be the case anymore, so I think we can remove this check if others are under the impression it won't break.

I don't remember specifically setting it to ignore `StartingLocation` with non-`file` schemes. That was probably an oversight on my part, or again I was assuming it would break later and wanted to catch the error early. I should have commented it better.

dialog/file.go
charlesdaniels: Probably not. This wasn't really ready for review, I was probably testing something.

dialog/file.go
charlesdaniels: Well if that was the decision.

Honestly, I think there has been enough contention on this point that maybe we should consider making it a preference the user can set? I find it infuriating that a few programs I use regularly open `~` rather than my working directory, since I then have to nagivate through many levels of file hierarchy.

A compromise might be to add `./` to "favorites"?

charlesdaniels:
As Andy said, thanks for your help @PucklaMotzer09. I would be happy to review your changes and will probably merge them. Is there some reason you can't just open a PR against my fork, as you did before? That would make it a lot easier to review and comment on the changes.

PucklaMotzer09:
Okay I'll just open a PR. I just thought It would be easier to have write access. Because it would let me commit faster if there were still some changes I should make. But a PR is also fine.

dialog/file.go
Jacalz: Yeah. I think adding PWD to favorites if PWD isn’t HomeDir would be a good addition.

dialog/file.go
PucklaMotzer09: In ```effectiveStartingDir()``` the scheme gets checked against ```file``` and if it doesn't match the function just skips to ```PWD``` and ```HomeDir```.

andydotxyz: I don't think we should be enforcing the scheme type - if the StartingLocation is a valid ListableURI then we should use it.

charlesdaniels:
@andydotxyz per our Slack discussion, I think this is ready to turn over:

* As far as I am aware, it should not error for non-`file://` URIs, though it may not work. (e.g. startdir is ignored without error if it is non-`file://`. I think @PucklaMotzer09 addressed this in his changes.

* I re-enabled the commented out code that @Jacalz noted.

* I have reverted the default starting directory to be `~` rather than `.` (I would like to note for the record that I oppose this change, and think the start directory should continue to be the CWD, but the consensus is against me here).

* I have removed the "open at" type constructors. 

* I have also modified the `TestShowFileOpen` test to explicitly assert that the first element of the path bar is `/`, which I think is the intended behavior, but I would like someone else double-check the changes I made to it in 5b5c3a34.

andydotxyz:
> * I have also modified the `TestShowFileOpen` test to explicitly assert that the first element of the path bar is `/`, which I think is the intended behavior

That is correct on Unix :) (just not Windows)

andydotxyz:
Thanks - really, really close. Replaced by #1379 where we will complete it.

PR #1110, opened 2020-06-16
Add missing comment for SetOnClosed()

Jacalz:
### Description:

This PR adds a missing comment for the function callback to run on window close.
I think that it might be worth back-porting this to master and the v1.3.1 in my opinion.

### Checklist:

- [-] Tests included.
- [ ] Tests all pass.
 - I am seeing test failures but none that are due to my change.


Jacalz:
The error seems to not be caused by my changes. Can probably be ignored...

PR #1119, opened 2020-06-20
Fix SetOptions in SelectEntry

### LINKED ISSUE #1118: Issue updating SelectEntry in onChange method for filtered list
### Describe the bug:
I want to filter the drop-down when the user is typing something into the `SelectEntry` text field.
For that, I've created code like described in the `example code` section.

If I debug the code, I can see, that the elements on `optionSelectEntry` are correct, but the UI isn't refreshing. Also an `optionSelectEntry.Refresh()` has no effect.

### To Reproduce:
see example code

### Example code:
**widget code**
```go
optionSelectEntry := widget.NewSelectEntry(filteredListForSelect(nil))
optionSelectEntry.PlaceHolder = "Type or select an option"
optionSelectEntry.OnChanged = func(input string) {
	optionSelectEntry.SetOptions(filteredListForSelect(&input))
}
```

**example for `filterListForSelect` function**
```go
func filteredListForSelect(filter *string) (resultList []string) {
	allOptionsList := []string{"Option A", "Selection B", "Method C", "Button D"}
	if filter == nil || *filter == "" {
		return allOptionsList
	} else {
		for _, option := range allOptionsList {
			if strings.Contains(option, *filter) {
				resultList = append(resultList, option)
			}
		}
		return resultList
	}
}
```

 - OS: found on MacOS
 - Version: 10.15.5
 - Go version: 1.14.4
 - Fyne version: 1.3.0

andydotxyz:
### Description:
Actually update the menu correctly when SetOptions is called after initial value set

Fixes #1118

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Ragin-LundF:
This PR fixes the issue I've reported for my case.

Issue #916, opened 2020-04-29
Add input validation


andydotxyz:
This probably needs some requirements fleshed out.

* verifying the content of an input (with some indication that verification fails
* verifying the content of a form (so Submit can be disabled)
* restrict input types on fields and switch soft-keyboard where appropriate.

Atomotron:
Except for switching the soft keyboard, this can all be done by overriding KeyDown. See [issue 1038](https://github.com/fyne-io/fyne/issues/1038) for a suggestion for how failed validation could be indicated to the user.

andydotxyz:
It might be better for cross platform to use OnChanged (so you validate the content not they key event) - mobile does not generate key events.

The soft keyboard type may be addressed separately in #975

Atomotron:
Given our discussion, I think the best way to implement this would be with:

- Subclassing to override OnChanged.
- A guarantee that OnChanged will run before anything is drawn. (This ensures that if you want to scrub, say, letters out of a number input, you can do it without flashing invalid input on the screen.)
- An IsValid flag in the Entry struct that its renderer respects, along with new colors in the theme interface.
- Some new subclasses of Entry that implement typical validation stuff in their default OnChanged

To see why this is a good idea, let's use an integer range with a maximum length as an example. Suppose I need to ask for a range start and a range end, where end > start, and also end < start + 100. (Perhaps my API will refuse to return more than 100 records.) In this case...

1. The validity of an entry can result in an error status displayed anywhere, including the valid state of *another* entry. I might want to report an error by setting a label's text, opening a pop-up, or logging to a file. Further, changing the range start can make a previously valid range end invalid. As a result, the validator needs write access to the valid state of both entries.
2. The validity of an entry can depend on data that comes from anywhere. The range end's validity depends on the value of the range start. I might also want to validate it by doing something weird, like requesting from the network.

It's clear that the cleanest way to support the general case would be to allow the user to run arbitrary code whenever an widget changes. Closures and callbacks would be a natural way to do this, but Fyne doesn't use callbacks, which leaves subclassing and additional fields in the widget struct as the remaining alternative.

Specific entry types, like "integer," "natural number," or "phone number," would fit in to this scheme as derived widgets. Maybe they could be packaged in fyne/widget/validated to keep them from proliferating and taking up too much of the widget namespace.

andydotxyz:
I don't know that overriding widget methods like `OnChanged` is required, or particularly clean. 
[...]

PR #1126, opened 2020-06-23
Add input validation to entry widget

Jacalz:
### Description:
This PR adds the initial steps towards input validation within fyne. It adds the necessary support for entry widgets which should later be possible to expand to form and soft keyboards as proposed in #916.

Partial solution for #916 

The implementation also works with password entries as you can see:
valid-password

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


charlesdaniels:
I like this approach.

However I think it fails to give the user any information about whether or not the input was accepted. 

This is my suggestion:

IMG_0227

* The colors could be based on the theme. I haven't dug into the theming system much, but I think something like a `ThemeColorAccentPositiveConnotation()` and `ThemeColorAccentNegativeConnotation()` would make sense, if we don't have something equivalent already.

* I think the icons are important, rather than just changing color, since colorblind users may not be able to distinguish the colors.

* There should be some way to convey to the user what a valid input is. I don't like tooltips for this, since they aren't discoverable.

* If there isn't already an appropriate variant of `Entry`, having a generic `EntryWithIcon` would be a good place to cut the abstraction for this, as other things might want that.

* Note for clarity: I would not modify non-validate entries in appearance from how the look now, since the sketch above might insinuate the contrapositive.

Jacalz:
Just rebased on develop and added an initial version of a little icon for displaying if the input is valid or not:
validation-preview

cmd/fyne_demo/screens/widget.go
```diff
@@ -174,6 +175,11 @@ func makeInputTab() fyne.Widget {
 	entryDisabled := widget.NewEntry()
 	entryDisabled.SetText("Entry (disabled)")
 	entryDisabled.Disable()
+	entryValidated := widget.NewValidatedEntry(regexp.MustCompile(`\d`))
+	entryValidated.SetPlaceHolder("Validated, must contain a number")
+	entryValidated.OnChanged = func(validText string) {
```
andydotxyz, line 180: Is this correct? Is OnChanged only called when the input has been validated now?

widget/entry.go
```diff
@@ -41,6 +42,10 @@ type Entry struct {
 	MultiLine   bool
 	Wrapping    fyne.TextWrap
 
+	RegexValidation *regexp.Regexp
```
andydotxyz, line 45: I just thought - RegexValidation also implies there could be multiple types of validation (which I agree with) - but why bake this specific one into entry? Perhaps "Validation" can be provided by a "RegexpValidation" provider?

andydotxyz:
A few comments inline. Mostly the tying to Regex(p) concerns me.
Also, can we follow the design guide and use a red/error colour for invalid state?

widget/entry.go
```diff
@@ -41,6 +42,10 @@ type Entry struct {
...
+	regexStatus     *regexStatus
+	ValidInput      bool // Will be false if RegexValidation is empty
```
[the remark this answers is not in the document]
Jacalz: Yeah good call. Shouldn't really be modified by the user. Better to have a method for accessing it.

cmd/fyne_demo/screens/widget.go
Jacalz: Yeah that was the initial way that I had this set up. I think it can be quite neat to only have the callback run when it is valid. Gets rid of a lot of manual code for checking if we are correct. Do you prefer to always run it, even with invalid input?

Jacalz:
I don't know about having a constructor. The multiline entry is also just setting one value so that falls into the same category somewhat though ;) But yeah, it definitely isn't a necessity to have. I just need to make sure that tests work without it :)

theme/icons.go
```diff
@@ -78,6 +83,37 @@ func (res *InvertedThemedResource) Invert() *ThemedResource {
...
+	return "error-" + res.source.Name()
+}
+
+// Content returns the underlying content of the resource adapted to the current background color.
+func (res *ErrorThemedResource) Content() []byte {
+	clr := &color.NRGBA{0xf4, 0x43, 0x36, 0xff} // Should be current().ErrorColor() in the future
+	return colorizeResource(res.source, clr)
+}
+
+// // Warning adapts the color of the resource to the warning color.
```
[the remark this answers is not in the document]
Jacalz: It isn't there. You were checking through an old commit ;)

theme.go
```diff
@@ -21,6 +21,7 @@ type Theme interface {
 	FocusColor() color.Color
 	ScrollBarColor() color.Color
 	ShadowColor() color.Color
+	// ErrorColor() color.Color
```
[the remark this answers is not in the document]
Jacalz: It could have been used as a reminder for 2.0 when we can add it in, but I think we can keep track of that anyway. Better remove it then :)

theme/theme.go
```diff
@@ -54,6 +55,7 @@ func DarkTheme() fyne.Theme {
 		hover:          color.NRGBA{0xff, 0xff, 0xff, 0x0f},
 		scrollBar:      color.NRGBA{0x0, 0x0, 0x0, 0x99},
 		shadow:         color.NRGBA{0x0, 0x0, 0x0, 0x66},
+		errors:         color.NRGBA{0xf4, 0x43, 0x36, 0xff},
```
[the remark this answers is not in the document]
Jacalz: Indeed. Better to add it all in with theme changes later down the road to 2.0 :)

Jacalz:
The implementation here should now be pretty much ready for prime-time except for the missing helper text for explaining why validation failed.

andydotxyz (approved):

andydotxyz:
Looking good.
I think Reason might want a fallback text, but until we display it that really doesn't matter :)

Jacalz:
Pushed some changes now per the discussions in Slack. There is now an `entry.SetValidationError(error)` that sets a validation manually, wording has moved from "Regex" to "Regexp", internal status is now of error type and there are a few smaller test, formatting and linter changes.

widget/entry_validation.go
```diff
@@ -0,0 +1,85 @@
...
+// SetValidationError manually updates the validation status until the next input change
+func (e *Entry) SetValidationError(err error) {
+	e.validationError = err
+
+	if e.Validator != nil {
+		e.validationStatus.Refresh()
+	}
+}
+
+var erroricon = theme.NewErrorThemedResource(theme.ErrorIcon()) // Avoid having to parse XML on each status refresh
```
stuartmscott, line 19: Note: this will not update if the theme changes

validator.go
```diff
@@ -0,0 +1,27 @@
...
+type Validator interface {
+	Validate(string) error
+}
+
+// RegexpValidator is a regexp implementation of fyne.Validator
+type RegexpValidator struct {
+	Reason string
+	Regexp *regexp.Regexp
+}
+
```
stuartmscott, line 18: Consider adding `func NewRegexpValidator(regexp, reason string) *RegexpValidator` that can wrap the `regexp.MustCompile`

widget/entry_validation.go
andydotxyz: Moving this to the CreateRenderer should not be too slow as the XML parse and render phase result is cached using the logic in internal/painter/svg_cache.go.

validator.go
andydotxyz: Agreed - except I would not use MustCompile in that case as it will cause panic if errored.
Instead try the compile and log an error if not successful I think? (and don't perform any validation) - better than an app dying :)

andydotxyz:
Agreed with Stuart's comments except minor warning about MustCompile.
Otherwise I'm happy once his comments are addressed.

widget/entry_validation.go
Jacalz: I will just remove this step entirely and instead quote the wise @andydotxyz, by saying "Best to leave optimisations until we see that they are needed" :wink:

Jacalz:
Alright. Should be all set now hopefully :slightly_smiling_face:

theme/icons.go
```diff
@@ -78,6 +83,29 @@ func (res *InvertedThemedResource) Invert() *ThemedResource {
...
+}
+
+// Name returns the underlying resource name (used for caching).
+func (res *ErrorThemedResource) Name() string {
+	return "error-" + res.source.Name()
+}
+
+// Content returns the underlying content of the resource adapted to the current background color.
+func (res *ErrorThemedResource) Content() []byte {
+	clr := &color.NRGBA{0xf4, 0x43, 0x36, 0xff} // Should be current().ErrorColor() in the future
```
stuartmscott, line 105: nit: it is easy to forget this comment. Consider adding a "TODO" or "FIXME" prefix so it shows up in searches

cmd/fyne_demo/screens/widget.go
```diff
@@ -175,6 +175,8 @@ func makeInputTab() fyne.Widget {
 	entryDisabled := widget.NewEntry()
 	entryDisabled.SetText("Entry (disabled)")
 	entryDisabled.Disable()
+	entryValidated := &widget.Entry{Validator: fyne.NewRegexpValidator(`\d`, "Does not contain a number")}
```
stuartmscott, line 178: nit:
```suggestion
	entryValidated := &widget.Entry{Validator: fyne.NewRegexpValidator(`\d`, "Must contain a number")}
```

validator_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+func TestValidator_RegexValidator(t *testing.T) {
+	r := RegexpValidator{Reason: "Input is not valid"}
+
+	assert.Equal(t, nil, r.Validate("nothing")) // Nothing to validate with, same as no validator.
+
+	r.Regexp = regexp.MustCompile(`^\d{2}-\w{4}$`)
+
+	assert.Equal(t, nil, r.Validate("10-fyne"))
+	assert.Equal(t, errors.New(r.Reason), r.Validate("100-fynedesk"))
```
stuartmscott, line 19: The assert library has special assertions for errors
```suggestion
	assert.NoError(t, r.Validate("10-fyne"))
	assert.EqualError(t, r.Validate("100-fynedesk"), r.Reason)
```

widget/entry.go
```diff
@@ -1083,6 +1112,13 @@ func (r *entryRenderer) Refresh() {
 	if r.entry.ActionItem != nil {
 		r.entry.ActionItem.Refresh()
 	}
+	if r.entry.Validator != nil {
+		if !r.entry.Focused() && r.entry.Text != "" && r.entry.validationError != nil {
+			r.line.FillColor = &color.NRGBA{0xf4, 0x43, 0x36, 0xff}
```
stuartmscott, line 1117: nit: add a similar TODO comment here so the hardcoded colour is replaced in future

widget/entry_test.go
```diff
@@ -43,6 +43,18 @@ func TestEntry_passwordRevealerCursor(t *testing.T) {
...
+func TestEntry_ValidatedEntry(t *testing.T) {
+	r := fyne.NewRegexpValidator(`^\d{4}-\d{2}-\d{2}`, "Input is not a valid date")
+	entry := &widget.Entry{Validator: r}
+	entry.ExtendBaseWidget(entry)
+
+	test.Type(entry, "2020-02")
+	assert.Error(t, r.Validate(entry.Text))
+
+	test.Type(entry, "-12")
+	assert.Equal(t, nil, r.Validate(entry.Text))
```
stuartmscott, line 55:
```suggestion
	assert.NoEqual(t, r.Validate(entry.Text))
```

Jacalz: I think you meant assert.NoError ;)

Jacalz:
Thanks for those small suggestions @stuartmscott. Should be fixed now :)

andydotxyz (approved):

PR #1130, opened 2020-06-25
Add a simple circle renderer to software painter

andydotxyz:
This is a simple rasteriser, so share it with GL which also is just drawing it to a texture at this time...

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
The failures are strange... not sure how this happened - image diff shows:

diff

Issue #1132, opened 2020-06-25
crash at pasting non text data into a text entry


RustamGamidov:
### Describe the bug:
fyne_demo 1.3.0 is crashing pasting image to the text entry
it was not the case on 1.2.4

### To Reproduce:
1. go get fyne.io/fyne
2. go get fyne.io/fyne/cmd/fyne_demo
3. run mspaint, select image area and ctrl+c
4. fyne_demo
5. go to Widgets->Input->Entry
6. ctrl+v or right click -> paste
crash
console output:
```
2020/06/25 17:36:46 fyne-io.fyne.demo-notify-1.ps1
2020/06/25 17:36:46 PATH C:\Users\user\AppData\Local\Temp\fyne-io.fyne.demo-notify-1.ps1
2020/06/25 17:36:55 Cancelled
GLFW: An invalid error was not accepted by the caller: FormatUnavailable: Win32: Failed to convert clipboard to string: The operation completed successfully.
GLFW: Please report this bug in the Go package immediately.
panic: FormatUnavailable: Win32: Failed to convert clipboard to string: The operation completed successfully.

goroutine 19 [running, locked to thread]:
fyne.io/fyne/vendor/github.com/go-gl/glfw/v3.3/glfw.acceptError(0x0, 0x0, 0x0, 0xc000bffea0, 0xc000bffe60)
        C:/Users/user/go/src/fyne.io/fyne/vendor/github.com/go-gl/glfw/v3.3/glfw/error.go:178 +0x208
fyne.io/fyne/vendor/github.com/go-gl/glfw/v3.3/glfw.panicError(...)
        C:/Users/user/go/src/fyne.io/fyne/vendor/github.com/go-gl/glfw/v3.3/glfw/error.go:185
fyne.io/fyne/vendor/github.com/go-gl/glfw/v3.3/glfw.(*Window).MakeContextCurrent(0xc00047e000)
        C:/Users/user/go/src/fyne.io/fyne/vendor/github.com/go-gl/glfw/v3.3/glfw/context.go:18 +0x48
fyne.io/fyne/internal/driver/glfw.(*window).RunWithContext(0xc00015e000, 0xc000bffe88)
        C:/Users/user/go/src/fyne.io/fyne/internal/driver/glfw/window.go:1008 +0x35
fyne.io/fyne/internal/driver/glfw.(*gLDriver).repaintWindow(0xc000028340, 0xc00015e000)
[...]

Issue #1123, opened 2020-06-21
Ctrl+v into text input field crashes app. Presumably clipboard is empty.


dbriemann:
### Describe the bug:

I am on Ubuntu Mate 20.04.

Using `ctrl+v` on a text input field of the `fyne_demo` crashes the application when clipboard is empty with the following stacktrace:

```
GLFW: An invalid error was not accepted by the caller: FormatUnavailable: X11: Failed to convert selection to string
GLFW: Please report this bug in the Go package immediately.
panic: FormatUnavailable: X11: Failed to convert selection to string

goroutine 10 [running, locked to thread]:
github.com/go-gl/glfw/v3.3/glfw.acceptError(0x0, 0x0, 0x0, 0xc0007cff90, 0xc000216060)
	/home/dlb/Coding/go/pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20200222043503-6f7a984d4dc4/error.go:178 +0x201
github.com/go-gl/glfw/v3.3/glfw.panicError(...)
	/home/dlb/Coding/go/pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20200222043503-6f7a984d4dc4/error.go:185
github.com/go-gl/glfw/v3.3/glfw.(*Window).MakeContextCurrent(0xc00023c000)
	/home/dlb/Coding/go/pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20200222043503-6f7a984d4dc4/context.go:18 +0x41
fyne.io/fyne/internal/driver/glfw.(*window).RunWithContext(0xc00010cb60, 0xc0007cfe88)
	/home/dlb/Coding/go/src/fyne.io/fyne/internal/driver/glfw/window.go:1008 +0x2e
fyne.io/fyne/internal/driver/glfw.(*gLDriver).repaintWindow(0xc00007c2c0, 0xc00010cb60)
	/home/dlb/Coding/go/src/fyne.io/fyne/internal/driver/glfw/loop.go:147 +0x60
fyne.io/fyne/internal/driver/glfw.(*gLDriver).startDrawThread.func1(0xc000036480, 0xc00007e190, 0xc00007c2c0)
	/home/dlb/Coding/go/src/fyne.io/fyne/internal/driver/glfw/loop.go:193 +0x217
created by fyne.io/fyne/internal/driver/glfw.(*gLDriver).startDrawThread
[...]

Issue #743, opened 2020-03-12
Panic when trying to paste empty clipboard into entry


terminationshock:
### Describe the bug:
An app crashes if one tries to paste text into an entry when the clipboard is empty.

### To Reproduce:

1. Clean the clipboard first: `xsel -bc`
2. Build and run the application below.
3. Put the cursor into the entry and hit "Ctrl + v"
4. Observe the crash

### Stack trace:

```
GLFW: An invalid error was not accepted by the caller: FormatUnavailable: X11: Failed to convert selection to string
GLFW: Please report this bug in the Go package immediately.
panic: FormatUnavailable: X11: Failed to convert selection to string

goroutine 1 [running, locked to thread]:
fyne.io/fyne/vendor/github.com/go-gl/glfw/v3.3/glfw.acceptError(0x0, 0x0, 0x0, 0x745cb5, 0xc000172501)
	/home/tobi/Code/Go/src/fyne.io/fyne/vendor/github.com/go-gl/glfw/v3.3/glfw/error.go:178 +0x201
fyne.io/fyne/vendor/github.com/go-gl/glfw/v3.3/glfw.panicError(...)
	/home/tobi/Code/Go/src/fyne.io/fyne/vendor/github.com/go-gl/glfw/v3.3/glfw/error.go:185
fyne.io/fyne/vendor/github.com/go-gl/glfw/v3.3/glfw.PollEvents()
	/home/tobi/Code/Go/src/fyne.io/fyne/vendor/github.com/go-gl/glfw/v3.3/glfw/window.go:949 +0x38
fyne.io/fyne/internal/driver/glfw.(*gLDriver).runGL(0xc00009d170)
	/home/tobi/Code/Go/src/fyne.io/fyne/internal/driver/glfw/loop.go:84 +0x1ff
fyne.io/fyne/internal/driver/glfw.(*gLDriver).Run(0xc00009d170)
	/home/tobi/Code/Go/src/fyne.io/fyne/internal/driver/glfw/driver.go:61 +0x2b
fyne.io/fyne/internal/driver/glfw.(*window).ShowAndRun(0xc0000cc280)
	/home/tobi/Code/Go/src/fyne.io/fyne/internal/driver/glfw/window.go:386 +0x57
main.main()
```

### Example code:

```
package main

import (
    "fyne.io/fyne/app"
    "fyne.io/fyne/widget"
)

func main() {
    application := app.New()
    win := application.NewWindow("")

    entry := widget.NewEntry()

[...]

PR #1136, opened 2020-06-26
dependencies: update glfw to 20200625191551

RustamGamidov:
changes in the glfw:
  - pr-285 Fixes a compiler error by using null joystick on freebsd
  - pr-282 fix regression from pr-256 GetClipboardString No longer returns an error
  - pr-280 fix: go mod vendor
  - pr-273 Added missing functions for compatibility with Vulkan
  - pr-270 .travis.yml: use latest stable Go (1.13.8 and 1.12.17 at this time)

### Description:
pulling in external dependency with the fixes preventing fyne crashes

Fixes #(issue)
#1132
#1123
#743 (claimed as fixed but the real fix is in the update that is missing)

### Checklist:
- [ ] Tests included.
- [x] Lint and formatter run with no errors. (no code changes, just deps)
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
I think the vendor files must be out of date after this? Using “fyne vendor” will ensure the updated C and H files are included as well as the Go.

RustamGamidov:
if i understood you correctly that was my mistake in he first try.
i've made force-push to update `vendor\modules.txt` in the same commit.
strange that only one build has failed because of that.

andydotxyz:
If you look at the changes tab for this PR there are 0 code changes present, so something doesn’t quite add up... the vendor folder should reflect all the code improvements you have listed I would think?

RustamGamidov:
oh! i missed `vendor` is in the repo and not received basing on the txt file. sorry
made that change

PR #1137, opened 2020-06-26
Update and add package synopsis for where useful

Jacalz:
### Description:
This PR adds useful package synopsis to packages for displaying a description on for example pkg.go.dev.
It also updated a couple existing ones and some spelling mistakes in the code.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.
  - There are always a few tests that fails on my machine on develop. Probably need to look at that at some point.


storage/file.go
```diff
@@ -1,3 +1,4 @@
+// Package storage provides storage specific functionality.
```
andydotxyz, line 1: I think this one could be better "storage specific" sounds like those per-platform comments above.
This however is features for all platforms.
How about "storage access and management functionality."

Jacalz:
Good point. Reworded it after your suggestion :)

PR #1138, opened 2020-06-26
Fix/1105

### LINKED ISSUE #1105: Set SetFixedSize to true, the menus are overlapped.
There is an error like below.

```Go
w.SetFixedSize(true)
```
image

```Go
w.SetFixedSize(false)
```
image

 - OS: Windows 10 20H1
 - Go version: latest
 - Fyne version: 1.3.0

andydotxyz:
### Description:
Fixing window size / resize handling on some bad Linux window managers.
Can be tested using demo code in each of the related issues.

Fixes #1105

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


RustamGamidov:
there's small inconsistency present with this PR applied at least on windows
```
func main() {
	app := app.New()
	win := app.NewWindow("window title")
	win.Resize(fyne.NewSize(200, 100))
	win.SetFixedSize(true)

	win.SetContent(widget.NewVBox(
		widget.NewLabel("long text to exceed initial window width"),
	))
	win.ShowAndRun()
}
```
This code is giving 200x100 sized window and it's width is adjusted once the window is dragged.
Commenting out `Resize` call is giving auto-sized window.
Removing `SetFixedSize` is adjusting the window width from the start.
Nothing critical, just a finding

andydotxyz:
Unfortunately I cannot replicate what you describe.
I think you mean that if `SetFixedSize(true)` is removed that the window does not appear at the size you specified, is that correct?

Can you please:
* Check that you are running on the `develop` branch and confirm the latest commit so we can check
* Show a screenshot or video of the window showing the wrong size?

Thanks

RustamGamidov:
Yes, removing `SetFixedSize(true)` call is giving width based on the content.

Here is the log (later is below):
```
8f5ebcd3 Some systems don't reset Raster type on resize, but we must            Andy Williams     Tue Jun 30 13:24:11 2020 +0100
f7e1114b Missed another comment                                                 Andy Williams     Tue Jun 30 14:46:54 2020 +0100
 (upstream/fix/1114)
102ff9e4 Merge pull request #1150 from andydotxyz/fix/1114                      Andy Williams     Tue Jun 30 19:28:36 2020 +0100
 (HEAD -> develop, upstream/develop)
```
image

on the image from the top to the bottom:
1. resized and fixed size initially
2. resized and fixed size after i dragged it a bit (same binary)
3. resized and size is not fixed
4. not resized, fixed size

It's hard to tell the size is wrong because it's explicitly resized and set to fixed size. That's why i called it inconsistent. It's not clear what is the correct behavior here

andydotxyz:
Oh I see, the issue is because you set fixedSize to be smaller than content?
That is indeed puzzling. I don't know what the right thing to do here is, we could:
* Ensure fixedSize is always respected (and you will lose some content)
* Update the size on load so that if you specify FixedSize too small it will enlarge

Both have pros and cons, I am not sure what is right...

andydotxyz:
I think this needs a new ticket, essentially "Setting FixedSize smaller than window content gives incosistent results"?
Could you open that please?

PR #1139, opened 2020-06-26
Improve linux packaging and installation

Jacalz:
### Description:

This commit does the following things:
- Handle if the distribution uses `/usr/local` or `/usr` when installing data.
- Install icon to `pixmaps` to ensure that the icon is shown correctly from the desktop file.
- Use an extra directory when packing to avoid cluttering the application root dir.
- Remove the temporary file in the end of the packaging.
- Add a Makefile with an install command for users to be able to install packaged applications separately from running `fyne install`.
    
Tested by installing using `fyne install` and by first packaging and then running `sudo make install` from inside the package.

I also added myself to AUTHORS along with ordering it alphabetically.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
I don't quite follow the Makefile... the result of "fyne package" should be a full app binary that can be copied or uncompressed in place. Install scripts belong in package maintainer helpers I would have thought?

Jacalz:
Yeah absolutely, but like the story about /use and /use/local is an example of when it might not be entirely easy to just copy it into place. While, yes makefiles are generally for packagers, but this way a user can install it much easier than copying folder for folder. Having the makefile gives a user just one command to run and install and application. Not all apps are in repositories so this makes it easier if a user grabs a release from a release page on GitHub for example.

andydotxyz:
The app packages are designed for specific systems, so the package command taking a parameter to determine /usr vs /usr/local would avoid the need for makefiles.
My understanding was that /usr/local is default so that's what we do - if you want to download then install to /usr you can use a temporary space and cp -r mycheckout/usr/local/* /usr

Jacalz:
For future reference: In the talk today, we more or less concluded that the Makefile is a good option for giving users more flexibility of installing packaged applications. The future could see Appimages getting built or something else, but until then this might be a good option.

cmd/fyne/package.go
```diff
@@ -93,42 +93,75 @@ type packager struct {
...
-		"Name="+p.name+"\n"+
-		"Exec="+filepath.Base(p.exe)+"\n"+
-		"Icon="+p.name+"\n")
-	if err != nil {
-		return errors.Wrap(err, "Failed to write desktop entry string")
+	if p.install {
+		tempDir = ""
+	}
+
+	if _, err := os.Stat(filepath.Join("usr", "local")); os.IsNotExist(err) {
```
andydotxyz, line 104: This line does not work on my system - I guess it is rooted at PWD?
The suggestion, however, works.

```suggestion
	if _, err := os.Stat(filepath.Join("/", "usr", "local")); os.IsNotExist(err) {
```

cmd/fyne/package.go
```diff
@@ -93,42 +93,75 @@ type packager struct {
...
+		_, err := io.WriteString(makefile, `ifneq ("$(wildcard /usr/local)","")`+"\n"+
+			"DESTDIR ?= /usr/local\n"+
+			"else\n"+
+			"DESTDIR ?= /usr\n"+
+			"endif\n"+
+			"\n"+
+			"install:\n"+
+			"	install -Dm00644 usr/"+local+"share/applications/"+p.name+`.desktop $(DESTDIR)$(PREFIX)/share/applications/`+p.name+".desktop\n"+
+			"	install -Dm00644 usr/"+local+"bin/"+filepath.Base(p.exe)+` $(DESTDIR)$(PREFIX)/bin`+filepath.Base(p.exe)+"\n"+
+			"	install -Dm00644 usr/"+local+"share/pixmaps/"+p.name+filepath.Ext(p.icon)+` $(DESTDIR)$(PREFIX)/share/icons/hicolor/512x512/apps/`+p.name+filepath.Ext(p.icon)+"\n")
```
andydotxyz, line 150: This destination directory is not the same as is installed by `fyne install`.
Which is wrong?

cmd/fyne/package.go
```diff
@@ -93,42 +93,75 @@ type packager struct {
...
+			return errors.Wrap(err, "Failed to write Makefile string")
+		}
+
+		tarCmd := exec.Command("tar", "zcf", p.name+".tar.gz", "-C", tempDir, "usr", "Makefile")
+		err = tarCmd.Run()
 		if err != nil {
 			return errors.Wrap(err, "Failed to create archive with tar")
 		}
+
+		err = os.RemoveAll(filepath.Join(p.dir, tempDir))
```
andydotxyz, line 161: This is only going to run if there were no errors, is that intended?

cmd/fyne/package.go
Jacalz: Yeah looks like I forgot to update the makefile. Good catch :+1:

cmd/fyne/package.go
Jacalz: Yeah that true. Will get that fixes :)

cmd/fyne/package.go
Jacalz: I didn't think about that. Good point :+1:

Jacalz:
I commited the update during the GitHub server issues. The changes seem to not be visible in this PR. Will create a new one.

PR #1144, opened 2020-06-27
Fixes https://github.com/fyne-io/fyne/issues/1143

### LINKED ISSUE #1143: dialog.NewCustomConfirm() does not call the callback for the "ok" and "dismiss" button
### Describe the bug:
dialog.NewCustomConfirm() does not call the callback of the function after pressing a button.
Background:
dialog.hideWithResponse() method in the base.go file calls only the callback if dialog.sendResponse is true (and if a callback exists). This was not set while creating the new dialog.

### To Reproduce:
Create new NewCustomConfirm dialog with a simple function and print something there to the console -> method will never be called.

### Example code:

```go
dialog.NewCustomConfirm("Callback example",
	"Ok", "Cancel",
	widget.NewLabel("Hello World"),
	func(confirmed bool) {
		fmt.Println("Callback...")
	},
	window)
```

 - OS: all
 - Version: all
 - Go version: all
 - Fyne version: 1.3.0

Ragin-LundF:
This method requires a callback. To call a callback it is necessary to set sendResponse() to true. Else dialog.hideWithResponse() will not call the callback.

### Description:
Fixes #1143

Bugfix has no tests because it is an abstraction of a dialog.

### Checklist:

- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/dialog/base.go
+++ b/dialog/base.go
@@ -186,6 +186,7 @@ func NewCustomConfirm(title, confirm, dismiss string, content fyne.CanvasObject,
 	callback func(bool), parent fyne.Window) Dialog {
 	d := &dialog{content: content, title: title, icon: nil, parent: parent}
 	d.callback = callback
+	d.sendResponse = true
 
 	d.dismiss = &widget.Button{Text: dismiss, Icon: theme.CancelIcon(),
 		OnTapped: d.Hide,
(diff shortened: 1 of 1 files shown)


andydotxyz:
Looking at this more that flag is set in the Show() method.

This may relate to an issue where some of the old New...Dialog() calls accidentally showed them without calling Show()?

Ragin-LundF:
You're right. I added a TODO like we talked about on Slack to keep an eye on it.

PR #1148, opened 2020-06-29
Improve packaging on Linux

Jacalz:
### Description:
This improves the packaging by doing the following changes:
- The `fyne install` command now works on distributions without `/usr/local`.
- Running `fyne package` now packages into a directory inside the project folder.
- The directory is now removed after packaging is complete.
- A Makefile is now included for making installation easier.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
I would like the person who reviews this to test it on Linux as well. Different Linux desktops seem to handle icons differently so please make sure that the icon is visible and that it installs correctly using both `fyne install` and  by running `sudo make install` inside a packaged application.

Jacalz:
I am debating on wether or not we should hard-code the path in the desktop file. When testing this on KDE Plasma, it seemed like it didn't want to pick it up automatically.

andydotxyz:
Do you mean the Exec path? That should be optional, so I guess if it fixes compattibility we could add it.
https://developer.gnome.org/desktop-entry-spec/#exec-variables

andydotxyz:
This looks good, install works as expected, but when I went to test the makefile I got this:

```
$ make
install -Dm00644 usr/local/share/applications/fyne_demo.desktop /usr/local/share/applications/fyne_demo.desktop
install -Dm00755 usr/local/bin/fyne_demo /usr/local/bin/fyne_demo
install -Dm00644 usr/local/share/pixmaps/fyne_demo.png /usr/local/share/pixmaps/fyne_demo.png
install: cannot remove '/usr/local/share/pixmaps/fyne_demo.png': Permission denied
make: *** [Makefile:10: install] Error 1
```

I know that "make install" might require sudo - but i don't know why "make" would?

andydotxyz:
Oh I see - is it running "install" as a default task?

Jacalz:
Yeah install is a default task because there are no other targets in the Makefile. Theres is no build target as we don't need to rebuild it in any way. We could possibly have an uninstall target if you think that makes sense?

Regarding hard-coding the path, I was talking about the icon path and not the exec path. Right now the icon is just the name of the software i.e. wormhole-gui and we rely on the desktop environment to find us the correct icon. This might be a bit flaky at times and possibly might also be different between desktop environments. Hard-coding the path like /usr/share/pixmaps/wormhole-gui.png would force that icon, but the downside is that it wouldn't pick an icon from the current theme if that exists.

Jacalz:
It should however work to not specify the icon path and just have one icon in /usr/share/pixmaps. It is exactly how firefox are doing it and I don't see a reason why it otherwise shouldn't work for us. Might need to test more with KDE.

andydotxyz:
I just mentioned the default task thing because "sudo make" is totally a bad thing, so I guess I would have normally seen "make" just complete successfully, perhaps saying "now make install (you might need sudo)" or just have the main task print out "now running make install" so people can see what's happening?

andydotxyz:
Regarding iconPaths I think you're right the downside is that themes won't work.
Personally I would look into why KDE didn't work before making a change to how we package.
It seems to work everywhere else ;)

Jacalz:
I dug a bit deeper into KDE and it looks like all it too was logging out and back in again...
I might have a solution that can solve your issue. hang tight :)

Jacalz:
The default target now print useful information. I think this is in good shape now. I don't have anything against merging it :)

andydotxyz:
This file was already a bit ugly so I didn't block merge - but in the future please try to keep method length down.
The packageLinux is now very length and hard to read - but as packageWindows is just as bad the file probably needs a proper clean and we'll get it all at once.

For example, as methods should do just 1 thing the creation of a makefile should be in a different function. But so should the desktop file creation hence why I think we need to tidy the whole file up. 

Thanks for this

Jacalz:
I see. Might tackle the cleanup sometime in the future :)

PR #1152, opened 2020-07-01
Correctly reset texture cache

andydotxyz:
This includes 'missing' textures as refresh may fix them
Previously an empty image that has a resource set then called Refresh() could have stayed empty as the "0" texture ID was kept

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/painter/gl/gl_core.go
+++ b/internal/painter/gl/gl_core.go
@@ -85,13 +85,15 @@ func (p *glPainter) SetOutputSize(width, height int) {
 }
 
 func (p *glPainter) freeTexture(obj fyne.CanvasObject) {
-	texture := textures[obj]
-	if texture != 0 {
-		tex := uint32(texture)
-		gl.DeleteTextures(1, &tex)
-		logError()
-		delete(textures, obj)
+	texture, ok := textures[obj]
+	if !ok {
+		return
 	}
+
+	tex := uint32(texture)
+	gl.DeleteTextures(1, &tex)
+	logError()
+	delete(textures, obj)
 }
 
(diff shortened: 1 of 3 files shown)


andydotxyz:
And you’re sure this was not happening before this commit?

Jacalz:
Yeah I think so. I double checked latest develop and couldn’t reproduce it there.

andydotxyz:
On the same machine? I just ask because I saw this on a Linux WM a couple of days back but not seen it on any other systems.

Jacalz:
Yea it was the same machine. I saw it yesterday too on Fynedesk once but didn’t see it in any of my further testing. I will rebase this against develop and see if it disappears.

Jacalz:
I rebased against develop and I am still seeing this behaviour when using this branch. I also seem to be getting the following printed to the terminal:
```
Xlib:  extension "NV-GLX" missing on display ":0".
```

I also tested develop again and I am seeing the resizing issue there now too. I definitely didn't have it when I tested before this, but now it is back. Seems like you are right, it isn't an issue with this PR. However the Xblib warning might be...

andydotxyz:
Erm, I have never seen that Xlib message before - it’s saying it cant load GLX... surely that relates to your machine and not this code?

Jacalz:
Yeah it definitely sounds like something that would be on my end, I just find it weird that I only get it on this PR. It is always as I close the window.

andydotxyz:
It seems more likely that https://github.com/fyne-io/fyne/pull/1153 could have related to that?
But this does not occur on develop?

PR #1180, opened 2020-07-10
window. add close button interceptor to react on the window closing before anything is done

### LINKED ISSUE #467: Add ability to intercept / veto window close
Is there a way to turn off the "Quit" option added to the gl driver's first main-menu menu?

https://github.com/fyne-io/fyne/blob/0f43af5fe05b6afc09ce8f0c03df1919147e5cd6/internal/driver/gl/menu.go#L33

Alternatively, a way to:

* change the label to allow for language localisation
* specify a custom action function (confirmation, unsaved work dialog, etc)

RustamGamidov:
### Description:
adding wrapper to closing handler so that we can react on the window closing before the window is destroyed.
for example it can serve window closing confirmation

Fixes #467 

### Checklist:

- [x] Tests included.
- [x] Tests all pass on win10

#### Where applicable:

#### example:
```go
func main() {
	app := app.New()
	win := app.NewWindow("window title")
	win.Resize(fyne.NewSize(200, 100))

	win.SetCloseIntercept(func() {
		dialog.ShowConfirm("title", "close?",
			func(response bool) {
				if response {
					win.Close()
				}
			}, win)
	})

	win.SetContent(widget.NewHBox(
		widget.NewLabel("long text to exceed initial window width"),
	))

	win.ShowAndRun()
}
```

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -74,8 +74,7 @@ type window struct {
 	mouseLastClick     fyne.CanvasObject
 	mousePressed       fyne.CanvasObject
 	onClosed           func()
-	willClose          func() bool
-	closing            bool
+	willClose          func()
 
 	xpos, ypos    int
 	width, height int
@@ -294,7 +293,7 @@ func (w *window) SetOnClosed(closed func()) {
 	w.onClosed = closed
 }
 
-func (w *window) SetWillClose(willClose func() bool) {
+func (w *window) SetWillClose(willClose func()) {
 	w.willClose = willClose
 }
 
@@ -410,7 +409,7 @@ func (w *window) Close() {
 	if w.viewport == nil {
 		return
 	}
-	w.closed(w.viewport)
+	w.doClose(w.viewport)
 }
 
 func (w *window) ShowAndRun() {
@@ -453,24 +452,13 @@ func (w *window) Canvas() fyne.Canvas {
 
 func (w *window) closed(viewport *glfw.Window) {
 	viewport.SetShouldClose(false)
-	if w.closing {
+
+	if w.willClose != nil {
+		w.willClose()
 		return
 	}
-	w.closing = true
 
-	go func() {
-		defer func() {
-			w.closing = false
-		}()
-
-		if w.willClose != nil {
-			if !w.willClose() {
-				return
-			}
-		}
-
-		w.doClose(viewport)
-	}()
+	w.doClose(viewport)
 }
 
 func (w *window) doClose(viewport *glfw.Window) {
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -74,7 +74,7 @@ type window struct {
 	mouseLastClick     fyne.CanvasObject
 	mousePressed       fyne.CanvasObject
 	onClosed           func()
-	willClose          func()
+	onCloseIntercepted func()
 
 	xpos, ypos    int
 	width, height int
@@ -293,8 +293,8 @@ func (w *window) SetOnClosed(closed func()) {
 	w.onClosed = closed
 }
 
-func (w *window) SetWillClose(willClose func()) {
-	w.willClose = willClose
+func (w *window) SetCloseIntercept(callback func()) {
+	w.onCloseIntercepted = callback
 }
 
 func (w *window) getMonitorForWindow() *glfw.Monitor {
@@ -412,13 +412,6 @@ func (w *window) Close() {
 	w.closed(w.viewport)
 }
 
-func (w *window) DoClose() {
-	if w.viewport == nil {
-		return
-	}
-	w.doClose(w.viewport)
-}
-
 func (w *window) ShowAndRun() {
 	w.Show()
 	fyne.CurrentApp().Driver().Run()
@@ -460,15 +453,11 @@ func (w *window) Canvas() fyne.Canvas {
 func (w *window) closed(viewport *glfw.Window) {
 	viewport.SetShouldClose(false)
 
-	if w.willClose != nil {
-		w.willClose()
+	if w.onCloseIntercepted != nil {
+		w.onCloseIntercepted()
 		return
 	}
 
-	w.doClose(viewport)
-}
-
-func (w *window) doClose(viewport *glfw.Window) {
 	viewport.SetShouldClose(true)
 
 	w.canvas.walkTrees(nil, func(node *renderCacheNode) {
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -409,7 +409,20 @@ func (w *window) Close() {
 	if w.viewport == nil {
 		return
 	}
-	w.closed(w.viewport)
+
+	w.viewport.SetShouldClose(true)
+
+	w.canvas.walkTrees(nil, func(node *renderCacheNode) {
+		switch co := node.obj.(type) {
+		case fyne.Widget:
+			cache.DestroyRenderer(co)
+		}
+	})
+
+	// trigger callbacks
+	if w.onClosed != nil {
+		w.queueEvent(w.onClosed)
+	}
 }
 
 func (w *window) ShowAndRun() {
@@ -458,19 +471,7 @@ func (w *window) closed(viewport *glfw.Window) {
 		return
 	}
 
-	viewport.SetShouldClose(true)
-
-	w.canvas.walkTrees(nil, func(node *renderCacheNode) {
-		switch co := node.obj.(type) {
-		case fyne.Widget:
-			cache.DestroyRenderer(co)
-		}
-	})
-
-	// trigger callbacks
-	if w.onClosed != nil {
-		w.queueEvent(w.onClosed)
-	}
+	w.Close()
 }
 
 // destroy this window and, if it's the last window quit the app
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -409,6 +409,13 @@ func (w *window) Close() {
 	if w.viewport == nil {
 		return
 	}
+	w.closed(w.viewport)
+}
+
+func (w *window) DoClose() {
+	if w.viewport == nil {
+		return
+	}
 	w.doClose(w.viewport)
 }
 
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -74,6 +74,8 @@ type window struct {
 	mouseLastClick     fyne.CanvasObject
 	mousePressed       fyne.CanvasObject
 	onClosed           func()
+	willClose          func() bool
+	closing            bool
 
 	xpos, ypos    int
 	width, height int
@@ -292,6 +294,10 @@ func (w *window) SetOnClosed(closed func()) {
 	w.onClosed = closed
 }
 
+func (w *window) SetWillClose(willClose func() bool) {
+	w.willClose = willClose
+}
+
 func (w *window) getMonitorForWindow() *glfw.Monitor {
 	xOff := w.xpos + (w.width / 2)
 	yOff := w.ypos + (w.height / 2)
@@ -446,6 +452,28 @@ func (w *window) Canvas() fyne.Canvas {
 }
 
 func (w *window) closed(viewport *glfw.Window) {
+	viewport.SetShouldClose(false)
+	if w.closing {
+		return
+	}
+	w.closing = true
+
+	go func() {
+		defer func() {
+			w.closing = false
+		}()
+
+		if w.willClose != nil {
+			if !w.willClose() {
+				return
+			}
+		}
+
+		w.doClose(viewport)
+	}()
+}
+
+func (w *window) doClose(viewport *glfw.Window) {
 	viewport.SetShouldClose(true)
 
 	w.canvas.walkTrees(nil, func(node *renderCacheNode) {
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -467,7 +467,7 @@ func (w *window) closed(viewport *glfw.Window) {
 	viewport.SetShouldClose(false)
 
 	if w.onCloseIntercepted != nil {
-		w.onCloseIntercepted()
+		w.queueEvent(w.onCloseIntercepted)
 		return
 	}
 
--- a/internal/driver/glfw/window_test.go
+++ b/internal/driver/glfw/window_test.go
@@ -821,6 +821,42 @@ func TestWindow_Clipboard(t *testing.T) {
 	cb.SetContent(cliboardContent)
 }
 
+func TestWindow_WillClose(t *testing.T) {
+	closed := make(chan bool)
+
+	createLocaWindow := func() fyne.Window {
+		w := createWindow("Test")
+		w.SetOnClosed(func() {
+			closed <- true
+		})
+		return w
+	}
+
+	checkClosed := func() bool {
+		select {
+		case <-closed:
+			return true
+		case <-time.After(100 * time.Millisecond):
+			return false
+		}
+	}
+
+	w := createLocaWindow()
+	w.SetWillClose(func() {})
+	w.Close()
+	assert.False(t, checkClosed())
+
+	w.SetWillClose(func() {
+		w.DoClose()
+	})
+	w.Close()
+	assert.True(t, checkClosed())
+
+	w = createLocaWindow()
+	w.Close()
+	assert.True(t, checkClosed())
+}
+
 // This test makes our developer screens flash, let's not run it regularly...
 //func TestWindow_Shortcut(t *testing.T) {
 //	w := createWindow("Test")
--- a/internal/driver/glfw/window_test.go
+++ b/internal/driver/glfw/window_test.go
@@ -826,35 +826,31 @@ func TestWindow_CloseInterception(t *testing.T) {
 	w := d.CreateWindow("test").(*window)
 	w.create()
 
-	// Close is not calling the interceptor
-	intercepted := false
+	onIntercepted := false
+	onClosed := false
 	w.SetCloseIntercept(func() {
-		intercepted = true
+		onIntercepted = true
+	})
+	w.SetOnClosed(func() {
+		onClosed = true
 	})
 	w.Close()
 	w.waitForEvents()
-	assert.False(t, intercepted)
+	assert.False(t, onIntercepted) // The interceptor is not called by the Close.
+	assert.True(t, onClosed)
 
-	// closed is calling the interceptor
+	onIntercepted = false
+	onClosed = false
 	w.closed(w.viewport)
 	w.waitForEvents()
-	assert.True(t, intercepted)
+	assert.True(t, onIntercepted) // The interceptor is called by the closed.
+	assert.False(t, onClosed)     // If the interceptor is set Close is not called.
 
-	// the interceptor is not calling Close if set
-	intercepted = false
-	w.SetCloseIntercept(func() {})
-	w.SetOnClosed(func() {
-		intercepted = true
-	})
-	w.closed(w.viewport)
-	w.waitForEvents()
-	assert.False(t, intercepted)
-
-	// Close is called if the interceptor is not set
+	onClosed = false
 	w.SetCloseIntercept(nil)
 	w.closed(w.viewport)
 	w.waitForEvents()
-	assert.True(t, intercepted)
+	assert.True(t, onClosed) // Close is called if the interceptor is not set.
 }
 
 // This test makes our developer screens flash, let's not run it regularly...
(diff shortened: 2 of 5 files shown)


RustamGamidov:
Yes, the main benefit is to get the place to prevent the window closing, but there could be different cases where onClose is too late because window is already closed. I took name from the original discussion and if you see there's better option just write it.

_closing_ yes it should be there, otherwise it's possible to get several handlers running that can create races. For example if you build my example and try to close the window several times without this protection it will show several confirmation dialogs.

Simple function call without goroutine is ok for a plain go code but using _ShowConfirm_ in the handler is taking the app to a deadlock (checked on linux and windows). I'll keep old _closed_ as _doClose_ to simplify the diff

The only reason of _no-op_ is i'm not sure how to check that case. I'll add the code there as well

RustamGamidov:
Mobile part is implemented, checked on pixel 2 simulator. 
There's an issue related to the _ShowConfirm_ call already in the codebase.
Calling it from the button click handler for example is not showing the dialog so i was checking using new window with two buttons.

andydotxyz:
Isn’t the race condition that requires ‘closing’ only present because of the goroutine?

If a confirm is required I would
Imagine that showing a confirm then vetoing the close, followed by a programmatic close if the user confirmed is a pretty standard workflow - this should not require goroutines should it?

RustamGamidov:
I had no issues with the goroutine present but every time i've tried to remove it i'm experiencing a deadlock.
This could be related to the viewport. Closed is connected to it and looks like these handlers should not block the flow.

andydotxyz:
Clicking close should absolutely be blockable - isn't that the point of this?
A window will not close until the line `viewport.SetShouldClose(true)` is executed, at which point it is scheduled for cleanup.

andydotxyz:
Thinking out loud a simpler design might require that if WillClose is set then the developer must call Window.Close() if that is what they want? That way threading is not required as the dialog callback can simply call that method if appropriate...

RustamGamidov:
`willClose` is replacing the internal closing procedure if set and developer should call `Close()` explicitly in the `willClose`
It's not solving the deadlock and `closing` need, just making it client code responsibility

test/testwindow.go
```diff
@@ -112,6 +112,10 @@ func (w *testWindow) SetOnClosed(closed func()) {
 	w.onClosed = closed
 }
 
+func (w *testWindow) SetWillClose(_ func()) {
+	// no-op
```
andydotxyz, line 116: This should probably mirror the other windows so comply with the WillClose contract

window.go
```diff
@@ -64,6 +64,10 @@ type Window interface {
 	// SetOnClosed sets a function that runs when the window is closed.
 	SetOnClosed(func())
 
+	// SetWillClose sets a function that runs before the window is closed
```
andydotxyz, line 67: I think these docs are now out of date

test/testwindow.go
andydotxyz: The test windows would be easier to test as well, to define the behaviour in some code...

internal/driver/gomobile/window.go
```diff
@@ -130,6 +135,15 @@ func (w *window) Hide() {
 }
 
 func (w *window) Close() {
+	if w.willClose != nil {
```
andydotxyz, line 138: I think this will get into a loop here - the willClose check should not be on the Close() call or you will get in a loop of not closing I think?

andydotxyz:
I think it looks cleaner this way - but perhaps others could feed in?
@toaster @stuartmscott @okratitan @lucor

internal/driver/gomobile/window.go
RustamGamidov: yes, that was the purpose of `closing` in my initial suggestion. I had a hope you do not like this approach because it's not covering programmatic window closing

RustamGamidov:
closing this PR to reopen and get additional commit

RustamGamidov:
one of the tests selection tests failed on one node, not sure it's related:
```
?   	fyne.io/fyne/tools/playground	[no test files]
--- FAIL: TestEntry_Select (0.50s)
    --- FAIL: TestEntry_Select/delete_select_down_with_Shift_still_hold (0.11s)
        util.go:60: 
            	Error Trace:	util.go:60
            	            				entry_test.go:1287
            	Error:      	Not equal: 
            	Test:       	TestEntry_Select/delete_select_down_with_Shift_still_hold
2020/07/13 08:17:46 Fyne error:  Entry does not allow Truncation
```

andydotxyz:
I have restarted in case it was a temporary fault

internal/driver/glfw/window.go
```diff
@@ -74,6 +74,7 @@ type window struct {
 	mouseLastClick     fyne.CanvasObject
 	mousePressed       fyne.CanvasObject
 	onClosed           func()
+	willClose          func()
```
ababo, line 77: IMO more intuitive would be to have `willClose func() bool`, i.e. the handler decides if the window is closed.

internal/driver/glfw/window_test.go
```diff
@@ -821,6 +821,42 @@ func TestWindow_Clipboard(t *testing.T) {
 	cb.SetContent(cliboardContent)
 }
 
+func TestWindow_WillClose(t *testing.T) {
+	closed := make(chan bool)
```
ababo, line 825: I think we can simplify this test just using a boolean variable. Indeed it will be set in a closure passed by `SetOnClosed()`.

internal/driver/glfw/window.go
```diff
@@ -404,6 +409,13 @@ func (w *window) Close() {
 	w.closed(w.viewport)
 }
 
+func (w *window) DoClose() {
```
ababo, line 412: Do we really need that to be exported?

internal/driver/gomobile/window.go
ababo: Maybe we should retain the original logic, i.e. `Close()` means unconditionally close the window, and introduce `TryClose()` method that relies on `willClose`?

window.go
```diff
@@ -64,6 +64,9 @@ type Window interface {
 	// SetOnClosed sets a function that runs when the window is closed.
 	SetOnClosed(func())
 
+	// SetWillClose sets a function that runs instead of closing if defined
```
ababo, line 67: Please add period at the end.

internal/driver/gomobile/window.go
RustamGamidov: then we have different behavior closing window using window manager and from the code

internal/driver/glfw/window.go
RustamGamidov: that was the initial suggestion and it was replaced for simplification so user should close window explicitly in the new handler

internal/driver/glfw/window_test.go
RustamGamidov: i did't manage to get stable tests using simple variable, running single test everything is OK but running the full set was not stable

internal/driver/glfw/window.go
RustamGamidov: yes, the idea was that user should explicitly close the window in this new handler

internal/driver/glfw/window.go
ababo: Not very intuitive IMHO. But I'm OK with that as well. I would only change the name of the handler from `willClose` to `onCloseTry` or something similar.

internal/driver/glfw/window_test.go
ababo: Probably that's because of goroutines involved. Can you try use atomics instead?

internal/driver/glfw/window.go
ababo: IMO a difference between `Close()` and `DoClose()`isn't clear intuitively. I would stick with `TryClose()` and `onCloseTry` naming scheme. @andydotxyz What do you think?

RustamGamidov: DoClose is expected to close window without additional checks so TryClose is exactly opposite name

ababo: Yep, that what I meant. Don't these names look more intuitive?

RustamGamidov: with the opposite meaning? not sure

ababo: After some thinking I came up with two alternatives:
1. We have `onClosing func() bool` in window implementation. Now `Close()` closes the window only if `w.onClosing == nil || w.onClosing()`. And no more methods are exported. IMO this way is the most intuitive.
2. We have `onCloseTry func()` in window implementation. It's called when user presses the closing cross or when `w.TryClose()` is called (this is a new exported method). `onCloseTry` decides whether to close the window (via `w.Close()`). IMO this way is more Go-idiomatic, but less intuitive.

ababo: Regarding the second way. I'm not sure if we need to close window if `TryClose()` is called and no `onCloseTry` handler is defined. But I would choose the first way.

andydotxyz: I think this got away from me sorry.
I had imagined that "Close()" would ignore the "will" callbacks and just close the window - the developer has told the window to close.
Tapping a close icon or having the app's Quit() called would trigger closing on each window internally, then the developer would be notified, if they requested the hook. In the call-close-to-confirm approach the original Close() method would do just that.

andydotxyz: The challenge with the `1` option @ababo was that it requires goroutines and messaging - the "willClose" want's to get a boolean return, but the common use-case is to then ask the user... In this case you are blocking a thread whilst waiting for a comfirm callback then passing the value. It made for complicated code.

The alternative, which is basically your `2` above means that the method is a notification and you can confirm with a callback - so threading issues go away.

At least that is what I had thought?...

ababo: @andydotxyz That makes sense then.

> I'm not sure if we need to close window if `TryClose()` is called and no `onCloseTry` handler is defined.

I would like `Close()` to be called if no `onCloseTry` is defined, otherwise the handler is called. Do you have a preference?

ababo: @RustamGamidov Are you OK with my naming?

RustamGamidov: not really, let me repeat myself, `Close()` is checking for the additional handler because it's the point where software closing request and window manager close request are coming to. To it looks strange to handle these two cases differently but it's something Fyne team is happy with, i'll step aside

ababo: @andydotxyz Please make a decision and let's advance with this PR, it's very critical for my project.

RustamGamidov: goroutine is needed anyway, with the current PR this responsibility is moved to the client code.
at least _ShowConfirm_ is deadlocking the app without additional goroutine

internal/driver/gomobile/window.go
andydotxyz: Isn't that how glfw works? User attempting to close the window can be vetoed but the code executing Close() is an instruction to be respected?

>  introduce TryClose() method that relies on willClose

There is no need to expose a new method here, as the gomobile driver draws the window headers so can call any internal method it wants when the close button is tapped...

internal/driver/glfw/window.go
andydotxyz: perhaps SetCloseIntercept()? that almost implies that follow-on action is required if set...

internal/driver/glfw/window_test.go
andydotxyz: Yes, I think hte goroutines are adding code complexity - this is what my alternative suggestion was attempting to avoid.

internal/driver/glfw/window.go
andydotxyz: I think we have crossed lines here, in the original code tapping a window close did not call `Close()` - it calls the internal method `closed()`, which is what the public `Close()` method also does.

If we think that developers calling Close() are doing so to explicitly close the window then `closed()` could have new content, which checks the callback - the main content moved to the `Close()` method and executed if there is no callback. Then the `Close()` does what you'd expect...

ababo: @andydotxyz can you summarize your vision, including proposed names for the handler and newly exported function?

andydotxyz: I think that this alternative approach needs only a single new function, named like `SetCloseIntercept(func())` - with the intercept func called only on user action and, if present, will only continue to close if the function calls `Close()`. But this is just me thinking out loud. I have asked if some other contributors have thoughts on this too...

ababo: @RustamGamidov Can you do as @andydotxyz suggests, i.e. rename `willClose` into `onCloseIntercept`, `SetWillClose()` into `SetCloseIntercept()` and get rid of `DoClose()`? Then the only thing that will remain is an approve from another contributor.

RustamGamidov: yeah... made it if i understood the idea. I'm just confused with how many things are moved to the client code like checking for deadlocks or loops. Especially because there was a comment regarding possible loop

RustamGamidov:
same test failed on the same node as far as i remember
```
?   	fyne.io/fyne/tools/playground	[no test files]
--- FAIL: TestEntry_Select (0.25s)
    --- FAIL: TestEntry_Select/delete_select_down_with_Shift_still_hold (0.06s)
        util.go:60: 
            	Error Trace:	util.go:60
            	            				entry_test.go:1287
            	Error:      	Not equal: 
            	Test:       	TestEntry_Select/delete_select_down_with_Shift_still_hold
```

internal/driver/glfw/window.go
andydotxyz: It should not be neccessary to do any checking for loops or deadlocks in client code. The simplicity we are looking for is simplicity for the developers using the API.
They should not have to worry about threading at all, that is one of our design principles.

RustamGamidov: then i believe _onCloseIntercepted_ should be called in goroutine but it probably worth to wait for comments from other repo owners because we already used our arguments and most likely we'll stuck arguing

internal/driver/glfw/window.go
```diff
@@ -443,6 +448,13 @@ func (w *window) Canvas() fyne.Canvas {
 }
 
 func (w *window) closed(viewport *glfw.Window) {
+	viewport.SetShouldClose(false)
+
+	if w.onCloseIntercepted != nil {
```
andydotxyz, line 453: As `Close()` will call `closed()` this will likely cause a loop.
Following the discussion yesterday if this approach is taken the code to actually close the window would need to move into `Close()`, which could be called from here if no intercept hook is registered.

internal/driver/glfw/window_test.go
```diff
@@ -821,6 +821,49 @@ func TestWindow_Clipboard(t *testing.T) {
...
+
+	createLocaWindow := func() fyne.Window {
```
andydotxyz, line 827: Is this a typo? "Loca" doesn't make sense to me

internal/driver/glfw/window_test.go
```diff
@@ -821,6 +821,49 @@ func TestWindow_Clipboard(t *testing.T) {
...
+		w.SetCloseIntercept(func() {
+			go func() {
+				closed <- true
+				w.Close()
+			}()
+		})
+		return w
+	}
+
+	checkClosed := func() bool {
```
andydotxyz, line 838: Maybe this should be 'checkIntercepted'? It's actually checking that the window /didn't/ automatically close.

internal/driver/glfw/window_test.go
```diff
@@ -821,6 +821,49 @@ func TestWindow_Clipboard(t *testing.T) {
...
+			return false
+		}
+	}
+
+	w := createLocaWindow()
+	w.SetCloseIntercept(func() {})
+	w.Close()
+	assert.False(t, checkClosed())
+
+	w.SetCloseIntercept(func() {
```
andydotxyz, line 852: Isn't this duplicating the test below?

internal/driver/gomobile/window.go
```diff
@@ -130,6 +135,11 @@ func (w *window) Hide() {
 }
 
 func (w *window) Close() {
+	if w.onCloseIntercepted != nil {
```
andydotxyz, line 138: As in the GLFW this creates a loop. The close button on the mobile driver windows would have to call a different function to trigger the close intercept. I don't think that needs to be exposed as this is all in the same package.

test/testwindow.go
```diff
@@ -37,6 +38,11 @@ func (w *testWindow) Clipboard() fyne.Clipboard {
 }
 
 func (w *testWindow) Close() {
+	if w.onCloseIntercepted != nil {
```
andydotxyz, line 41: To match the pattern this should not really be here. Perhaps the test window might need a `simulateUserClose()` or similar function that would test the intercept.
Either that or I was wrong about this being testable here - maybe test windows always close when requested?

internal/driver/glfw/window.go
andydotxyz: I wonder if perhaps the event queue is the right place to put it? Just launching goroutines to work around possible deadlocks seems like it's not fixing the underlying issue? We used to do that with button taps and key presses and it caused a jumble of out-of-order events, so we don't do that any more.

internal/driver/glfw/window.go
RustamGamidov: ok, missed this idea, sorry. 
there's one more issue on this way. having interception logic in the _closed_ it's i don't see the way to call it from the test

andydotxyz: Yes, that's true. I left a comment inline about this - if it is to be tested then we might need something new like 'simulateUserCloseRequest' just for the test driver? Not sure if that is really needed or not.

okratitan:
So I think my feeling on this is - I like the style of WillClose with a return boolean.  The Intercept requiring the dev to call Close() is a little confusing/not obvious in my opinion.  However - WillClose is probably the wrong name and the name of that Method will need to be revisited.

On the flip side the implementation will be easier/cleaner with Intercept - as WillClose will have to involve some goroutines etc...  So there is a trade off either way.  I will be happy going with what the majority decides.

RustamGamidov:
fixed the example and rebased over the current _develop_

internal/driver/glfw/window.go
```diff
@@ -446,19 +464,14 @@ func (w *window) Canvas() fyne.Canvas {
...
-		case fyne.Widget:
-			cache.DestroyRenderer(co)
-		}
-	})
-
-	// trigger callbacks
-	if w.onClosed != nil {
-		w.queueEvent(w.onClosed)
+	if w.onCloseIntercepted != nil {
+		w.onCloseIntercepted()
```
andydotxyz, line 470: Earlier you mentioned that this could block - should it be passed to `w.queueEvent` instead?

window.go
```diff
@@ -64,6 +64,9 @@ type Window interface {
 	// SetOnClosed sets a function that runs when the window is closed.
 	SetOnClosed(func())
 
+	// SetCloseIntercept sets a function that runs instead of closing if defined
```
andydotxyz, line 67: Can you add a "." to this, and add another line instructing people to call `Close()` to confirm window closure?

internal/driver/glfw/window.go
RustamGamidov: i've tried that, and it's not helping unfortunately
i can do this to follow the pattern

andydotxyz: Can you confirm what the issue is please?
Looking at the example above I can't help but feel the channel should not be needed - just calling `Window.Close()` in the callback should work.

RustamGamidov: the issue is not reacting `ShowConfirm` 
Yes, it's solved with removing the channel with and without using `queueEvent`

RustamGamidov:
> is it possible to re-add testing to this version?

i don't see the way how to do this as the current solution is catching window manager closing only that is private

RustamGamidov:
```
# cd .; git clone -- https://github.com/fzipp/gocyclo /Users/travis/gopath/src/github.com/fzipp/gocyclo
Cloning into '/Users/travis/gopath/src/github.com/fzipp/gocyclo'...
fatal: unable to access 'https://github.com/fzipp/gocyclo/': Failed to connect to github.com port 443: Operation timed out
package github.com/fzipp/gocyclo: exit status 128
```
quite unlikely related to the PR

andydotxyz:
> i don't see the way how to do this as the current solution is catching window manager closing only that is private

But from tests you can call private methods...

RustamGamidov:
oh indeed! missed they are in the same package
made it in one test that makes tests dependent on the previous code but i see it's the common practice in the project

RustamGamidov:
ci failed on one of the nodes
```
--- FAIL: TestFileWatcher_FileDeleted (0.19s)
    settings_desktop_test.go:92: File watcher callback was not called
```

andydotxyz:
@okratitan I think that @toaster and I are happy here, can you approve the latest version too please?

andydotxyz:
Many thanks, this will be in 1.4 release later this year.

RustamGamidov:
:pray: thanks a lot for all your patience and comments! That was the long way!

PR #1183, opened 2020-07-11
Update scroller.go to add ScrollToEnd function.   It help to auto scr…

ScottHuangZL:
…oll container to last to show latest info which end user really cared about

//ScrollToEnd will help to auto scroll container to last to show latest info which end user really cared about
//Such as when user insert new widget into ScrollerContainer, or container contail label, and label continue add new text, 
//Usually users would apprecaite if the content auto scroll to end to show latest new added info 
func (s *ScrollContainer) ScrollToEnd() {
	s.Offset.Y = s.Content.Size().Height - s.Size().Height
}

### Description:

Fixes #(issue)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


okratitan:
Hello - I wanted to pop in and offer my 2 cents.  I think if we add scroll to bottom we need to add scroll to top, scroll to right, and scroll to left as well.  All are very useful and adding just one seems to fall short of completeness

andydotxyz:
Yeah good point. Though Left/Right vs Leading/Trailing is a little challenging - would we get away just with Top/Bottom at this time?

ScottHuangZL:
Adjust ScrollToBottom function; Add ScrollToTop function; add test for both;
Please help to verify and merge, thanks.

widget/scroller.go
```diff
@@ -354,6 +354,24 @@ func (s *ScrollContainer) CreateRenderer() fyne.WidgetRenderer {
 	return scr
 }
 
+
+//ScrollToBottom will scroll content to container bottom - to show latest info which end user just added
+func (s *ScrollContainer) ScrollToBottom(isNeedRefresh bool) {
+	s.Offset.Y = s.Content.Size().Height - s.Size().Height + s.MinSize().Height
```
andydotxyz, line 360: Why is MinSize() needed here? The scroller Size() should be sufficient...
Y should range from 0 to (contentHeight - scrollerHeight)...

widget/scroller.go
```diff
@@ -354,6 +354,24 @@ func (s *ScrollContainer) CreateRenderer() fyne.WidgetRenderer {
 	return scr
 }
 
+
+//ScrollToBottom will scroll content to container bottom - to show latest info which end user just added
+func (s *ScrollContainer) ScrollToBottom(isNeedRefresh bool) {
```
andydotxyz, line 359: I didn't notice the isNeedRefresh flag get added - it feels unexpected.
Normally widgets refresh after any change function, is there a problem with doing that here?

widget/scroller.go
```diff
@@ -354,6 +354,24 @@ func (s *ScrollContainer) CreateRenderer() fyne.WidgetRenderer {
...
+
+//ScrollToTop will scroll content to container top. It is easy if compare to ScrollToBottom
+func (s *ScrollContainer) ScrollToTop(isNeedRefresh bool) {
+	s.Offset.Y = 0
+	if isNeedRefresh {
+		s.Refresh()
+	}
+}
+
+
```
andydotxyz, line 374: I think this extra line is why Travis is failing - but there also might be trailing whitespace above (see travis errors for info).

widget/scroller_test.go
```diff
@@ -43,6 +43,25 @@ func TestScrollContainer_MinSize(t *testing.T) {
...
+	assert.Equal(t, 0, Y)
+}
+
+func TestScrollContainer_ScrollToBottom(t *testing.T) {
+	rect := canvas.NewRectangle(color.Black)
+	rect.SetMinSize(fyne.NewSize(500, 50))
+	scroll := NewScrollContainer(rect)
+	scroll.ScrollToBottom(true)
+	Y := 50 + scroll.MinSize().Height
+	ExpectedY := scroll.Content.Size().Height - scroll.Size().Height + scroll.MinSize().Height
```
andydotxyz, line 61: I think that here `Y` is the expected value and `ExpectedY` is the actual value. If that's correct can you please update naming and parameter order below?

andydotxyz:
Thanks @ScottHuangZL - a few things in the review. Naming and formatting improvements probably needed.

widget/scroller.go
ScottHuangZL: In one of my scenario, the refresh is needed. So, better let end user to select must refresh or not.

widget/scroller.go
ScottHuangZL: Both ok.   Original "s.Content.Size().Height - s.Size().Height" already good, but the end would have some space.  And if I use "s.Content.Size().Height - s.Size().Height + s.MinSize().Height" it will scroll to the real end.

widget/scroller.go
andydotxyz: But if you set this value and don't call refresh then it will not update the display - how can it be valid to not refresh?

widget/scroller.go
andydotxyz: The test `TestScrollContainer_Scrolled_Limit` shows that the furthest offset is content size - scroll size.
Adding MinSize should, in theory, make no difference, it will be clipped by `scroller.go:440`.
Do you have an example of how the gap is left when you don't adjust like you say?

widget/scroller_test.go
ScottHuangZL: Done.

ScottHuangZL:
Thank you Andy for the patient. I have both update the func and test, please double check. Hope this time meet both you and Travis's request.

widget/scroller.go
ScottHuangZL: Done to change back to (s.Content.Size().Height - s.Size().Height)

widget/scroller.go
ScottHuangZL: Remove that isNeedRefresh flag.

widget/scroller.go
ScottHuangZL: Thanks for point out this.

ScottHuangZL:
I find Travis CI success pass already.   Andy, can you merge the pull request now? Thanks.  Btw, you can direct change in case still have minor point need changed. Thanks.

andydotxyz:
Thanks. Updated to the `develop` branch (I didn't notice it was on master before), just waiting for it to test alongside the latest code but otherwise this looks good.
Thanks for your contribution.

Noofbiz:
Sorry not sure if this is the place, I think a scroll to element would be much more useful, like for FAQs and stuff you'd be able to auto scroll to any answer with like a table of contents at the top. It could accept a widget, and there could be funny widgets for "top" "bottom" etc.

andydotxyz:
I think the use-case for this @Noofbiz was that if you're appending or prepending content a simple API was missing to do the scroll.
Scrolling to an element somewhere in content may be useful as well, but I'm not sure how frequently it would be used - a table of contents is not hugely common in an application is it? You could consider using the `widget.Accordion` instead, which would expand the answer content of an FAQ under the question... might be a neater solution?

Noofbiz:
Oh no worries, just thought a single function like that would be more useful than a restricted one is all. And fixing it down the road would either be an API breaking change or clutter up the code base. Just a thought, not really a concrete need or anything

andydotxyz:
I think they are different use-cases (subtly). If you scroll to an element but that is bigger than the scroller it will probably top-left align it. If you are scrolling to the bottom then it's ignoring how large the last child in the list is.
Basically what I mean is you can't really do both with a single API so adding one now and another later is fine.

ScottHuangZL:
> Sorry not sure if this is the place, I think a scroll to element would be much more useful, like for FAQs and stuff you'd be able to auto scroll to any answer with like a table of contents at the top. It could accept a widget, and there could be dummy widgets for "top" "bottom" etc.

@Noofbiz You can put the FAQs into a label, and the label into a container, then you can scroll to top or bottom of the container  to show latest FAQs.   Actually, I create these 2 small func is for similar purpose:  User click any button to do the actions, and I want to show the auto scrolled log, the container default would not auto scroll, but you can call ScrollToBottom after you update the label.   For example:

```
var semaInfo = make(chan struct{}, 1)  // a binary signal
var infoBuffer bytes.Buffer // a buffer for update label contents
var content = widget.NewLabel("") 
var scroller = widget.NewScrollContainer(content)


//You can create above in your main UI, and then call below func to show log
func addInfo(note string) {
	semaInfo <- struct{}{}
	infoBuffer.WriteString(fmt.Sprintf("%s : %s\n", time.Now().Format("2006-01-02 15:04:05"), note))
	content.SetText(infoBuffer.String())
	<-semaInfo

	scrollToBottom(scroller)
}

func setInfo(note string) {
	semaInfo <- struct{}{}
	infoBuffer.Reset()
	content.SetText(note)
	<-semaInfo
}
```

PR #1186, opened 2020-07-12
Use icons for file dialog favourites

Jacalz:
### Description:
This PR adds icons to the file picker favourites buttons to make it look a bit better and to make it possible for the user to know what a favourite points to without reading the full name of it. To make that happen, I had to add icons for download, computer and document.

image

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
I added the drive icon that you requested. Should be fine for a re-review now :)

PR #1187, opened 2020-07-12
WiP: Input Dialog

charlesdaniels:
### Description:

This PR implements an `InputDialog` which can be used to prompt the user for input. This can be simpler than using a custom widget, if you just need the user to enter a simple value.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/progressbar.go
+++ b/widget/progressbar.go
@@ -65,7 +65,7 @@ func (p *progressRenderer) Refresh() {
 	p.applyTheme()
 	p.updateBar()
 
-	canvas.Refresh(p.progress)
+	canvas.Refresh(p.progress.super())
 }
 
 // ProgressBar widget creates a horizontal panel that indicates progress
--- a/widget/progressbarinfinite.go
+++ b/widget/progressbarinfinite.go
@@ -92,7 +92,7 @@ func (p *infProgressRenderer) doRefresh() {
 	p.bar.FillColor = theme.PrimaryColor()
 
 	p.updateBar()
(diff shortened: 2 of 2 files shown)


charlesdaniels:
ezgif-1-ae7e22e7dc21

Video of the new dialog type in action.

charlesdaniels:
RFC from @andydotxyz or anyone else who would care to chime in.

* The most elegant way I could think of to actually get the text the user entered into a callback was via the `onConfirm()` cb I just added, however and the validator are both `nil`-able to minimize friction.
* The dialog callback is set to a no-op, maybe I should add a way to allow setting that too? This might be a good idea, since right now there is no way to detect if the user cancelled, other than that the `onConfirm` callback doesn't fire.

Once we're happy with how this works, I will add some test cases and remove the WiP.

andydotxyz:
Which callback is a no-op? Do you mean OnClosed?
The dialogs on the whole do need a bit of a design cleanup but until the 2.0 work we need to be consistent with how it currently works.

charlesdaniels:
I meant that the callback that is part of the base dialog type is currently a no-op.

andydotxyz:
Ah, I see - well that's an internal detail so it's not a worry if the exposed API is consistent with the others?
With them they (currently) call with a zero value on cancel.
That's the current design - I think in 2.0 we may need to move to the OnCancel vs OnConfirm(?) rather than a single generic function (later this year).

PR #1188, opened 2020-07-15
Feature/styling

jkvatne:
### Description:
This is a proposal for extended styling/theming. 
The themeing includes the json load theme function from SteovOC64's PR #970 and has a very similar struct for storing the data.

In addition I have implemented a new styling struct for the buttons. This makes it possible to tailor each button in any way the user wants. You can create button styles that can be easily applied to any number of buttons.

Here is an example:
```
		customStyle := &theme.ButtonStyle{
		EnabledColor: theme.PrimaryColor(),
		TextColor:    colornames.Lightgoldenrodyellow,
		ShadowColor:  theme.ShadowColor(),
		TextSize:     theme.TextSize() * 3 / 2,
		Height:       40,
		MinWidth:     300,
	}
	customStyle.AddStateColors()  // Will automaticaly calculate the hovered/tapped/focused colors
	myButton1.SetStyling(customStyle)
	myButton2.SetStyling(customStyle)
```
Note that the styling can include colors from the current Theme as well as fixed custom colors, or modified Theme colors.
There is six predefined button styles in the form of six structs:
DefaultRaisedButton, DefaultFlatButton, PrimaryRaisedButton, PrimaryFlatButton, SecondaryRaisedButton, SecondaryFlatButton.
More defaults could be defined.

Later on we can extend this with new stylings for other widgets, and perhaps include more options for the buttons, like rounded corners.

This proposal is made backward compatible, and it passes all tests. But when I try to implement focused buttons, it will fail many tests, so that have to wait. 

### Checklist:

- [ ] Tests included.
- [x ] Lint and formatter run with no errors.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -114,6 +114,7 @@ func newAppWithDriver(d fyne.Driver, id string) fyne.App {
 			helper.ApplySettings(set, newApp)
 		}
 	}()
+	helper.ApplySettings(newApp.settings, newApp)
 	if !d.Device().IsMobile() {
 		newApp.settings.watchSettings()
 	}
--- a/cmd/fyne_demo/main.go
+++ b/cmd/fyne_demo/main.go
@@ -56,12 +56,15 @@ func welcomeScreen(a fyne.App) fyne.CanvasObject {
 		layout.NewSpacer(),
 
 		widget.NewGroup("Theme",
-			fyne.NewContainerWithLayout(layout.NewGridLayout(2),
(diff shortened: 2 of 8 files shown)


andydotxyz:
This is a challenging topic. Our standard widgets adhere to the current theme and so the direct manipulation of colours, sizes etc isn't something we want to encourage. The canvas API exposes these details if developers want to construct something specific.
Extending the theme to handle missing semantic variants (like validation, for example) fits this approach, but passing specific colours does not.

PR #1195, opened 2020-07-17
Feature/material2

### LINKED ISSUE #958: Subjective comment on default colour choices in provided Fyne themes

andydotxyz:
### Description:
Working on bringing the Fyne design more in-line with Material Design.

So far:
* Update colours to better match recommended palette
* Set Primary to be the same for both themes :)
* Added support to invert icons to work on primary buttons
* Deprecate old button/icon/disabledbutton/hyperlink colours that shouldn't be required

Still to do:
* Fix padding
* Update various widgets to better fit with this asthetic.

This PR is designed to be non-breaking and lays work towards the full 2.0 theme rework.
There are some behavioural changes (i.e. button background colour) but they should not be enough to require breaking change status.

Fixes #958

### Checklist:

- [x] Tests included. <- the image tests are sufficient. Lots changed of course, but you can see how the theme changes apply.
- [x] Tests all pass.


andydotxyz:
Current state on the front page (see how padding / shadow layout needs work):

Screenshot 2020-07-17 at 22 29 39
Screenshot 2020-07-17 at 22 29 35

Jacalz:
I think it might make sense to have a darker text colour for the text inside the selected tab:
image

andydotxyz:
> tab bar colours seem to not be updating correctly some times when switching themes

This is not a new bug and sadly I have not been able to track it down.

Is the issue you reported with custom dialogs about theme change, or does it not display correctly initially?

> it might make sense to have a darker text and icon colour for the text inside the selected tab

Agreed, that is part of moving to the complete material UI but I could not think how to do it without breaking the API, which I a, avoiding at this stage. "OnPrimary" is the name of that property but I think it needs the 2.0 theme breakage to bring in.

Good catch on the select button, I will fix that :)

Jacalz:
The issue with dialog was regarding changing the theme. It might have been correct to begin with (I will need to double check that), but it definitely didn’t update the theme like it should when switching the theme. Might be some code missing for that.

Another thought, could we improve the dialog styling in some way? It might be outside of the scope of this PR, but it feels like they are a bit flat. Changing them would be a good step forward as long as it is backwards compatible 😁

Jacalz:
This would also need to be rebased against the changes made by my PRs 😅

andydotxyz:
Rebased now :)
I will look into the dialog issue later on and let you know what I find.

PR #1199, opened 2020-07-20
match macOS/Windows system theme

andydotxyz:
### Description:
We should use the macOS theme.
And we should update apps when it changes.

Now we do :)

### Checklist:

- [ ] Tests included. <- on macOS remove your fyne settings and launch an app, then change your macOS theme
- [x] Tests all pass.

On Windows go to Settings -> Personalisation -> Color and choose light or dark. Launch a Fyne app and it will match.

If on either system you had previosuly chosen a theme then either remove your config or run fyne_settings and choose "system default".


andydotxyz:
Added Windows lookup as well. Though I cannot figure out a way to monitor the windows theme changes :(
I hope that we can leave this PR here, more work may be added in the future.

Jacalz:
Perhaps update the PR title to add Windows supprot also? :)

PR #1209, opened 2020-07-24
Addition of NewImageFromURI

andydotxyz:
### Description:
A discussion point reall. `NewImageFromFile` isn't really inline with the new `URI` based file access - so this seems sensible.
However should the `http` and `https` scheme support be in the driver, or the image code or somewhere else?
Basically: is `OpenFileAtURI` file specific or not - and if yes should there be another method call for non-files?

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


canvas/image.go
```diff
@@ -74,6 +79,49 @@ func NewImageFromFile(file string) *Image {
...
+// Images returned from this method will scale to fit the canvas object.
+// The method for scaling can be set using the Fill field.
+func NewImageFromURI(uri fyne.URI) *Image {
+	if uri.Scheme() == "file" && len(uri.String()) > 7 {
+		return &Image{
+			File: uri.String()[7:],
+		}
+	}
+
+	var read io.ReadCloser
```
stuartmscott, line 93: Could this be split out into `NewImageFromReader`?

andydotxyz: A cunning plan. Does the rest of the PR seem sensible? I was more keen to get feedback on the http/https part in the  description.

Jacalz (commented):
This is a great addition. My view on this is that the http/https probably belongs in the driver (unless the go stdlib has us covered in all situations). I also think a general URI method makes sense.

andydotxyz:
I don't think that every driver needs to define how http works - that is built into Go.
The uncertainty is because this would be the only usage of URI where that would work.
There is a discussion about how the toolkit (and developers) could provide URI scheme handling through somesort of register, the http code may belong in there and we could provide this scheme support universally.

Jacalz:
I agree. That sounds like a good plan to me.

canvas/image.go
andydotxyz: Added

andydotxyz:
OK, I have updated this PR to remove the HTTP/HTTPS handling as that belongs in the discussed URI handler registry.
Otherwise updated as per comments.

stuartmscott (approved):

stuartmscott (approved):

PR #1210, opened 2020-07-24
Scheme always lower case

andydotxyz:
### Description:
URI Scheme should be lower case - probably not a field that people would compare case insensitive

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


toaster:
Yeah, that’s quite a tricky one.

I think we should _not_ change the case of URLs (including plain FS paths).
In cases where the URL or a part of it is used for comparison (matching or filtering) it should be treated case-insensitive.

I don’t know if this is possible without fuss for all current use-cases.

andydotxyz:
Thanks @toaster - I have updated this to maintain case of the extension.
I think the PR to make scheme lower still makese sense though, as it is never part of a file path etc

PR #1212, opened 2020-07-24
Add Resize function for dialog.NewFileOpen

### LINKED ISSUE #1211: Can we allow user to resize the file open dialog by themselves?

The file open dialog is quite small in Window 10.     If the file with long name but similar prefix, then quite hard for user to select correct file.

NA

**Can we allow user to resize the file open dialog by themselves?**

Or can programmer use **fd.Resize(fyne.NewSize(1000,800))** to adjust file open dialog's size before show? 
I find the dialog have 4 file icons each row by default.   So, may be provide API to allow user to adjust to 4~10 icons.

```go
fd := dialog.NewFileOpen(func(reader fyne.URIReadCloser, err error) {
		if err == nil && reader == nil {
			return
		}
		if err != nil {
			dialog.ShowError(err, w)
			return
		}
		fileOpened(reader)
	}, w)
	fd.SetFilter(storage.NewExtensionFileFilter([]string{".xlsx"}))
        **fd.Resize(fyne.NewSize(1000,800)) //can we add here??**
	fd.Show()
```

ScottHuangZL:
Add resize function for file dialog.

### Description:

Please refer to [url](https://github.com/fyne-io/fyne/issues/1211)

Fixes #1211

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


ScottHuangZL:
For example:

```go
widget.NewButton("File Open With Filter (.txt or .png)", func() {
			fd := dialog.NewFileOpen(func(reader fyne.URIReadCloser, err error) {
				if err == nil && reader == nil {
					return
				}
				if err != nil {
					dialog.ShowError(err, win)
					return
				}

				fileOpened(reader)
			}, win)
			fd.SetFilter(storage.NewExtensionFileFilter([]string{".png", ".txt"}))
			fd.Show()
			fd.Resize(fyne.NewSize(600, 400))

		}),

```

ScottHuangZL:
> Should this be specific to FileDialog? If dialogs should be resizable I can't help but wonder if it should apply to all Dialogs?
> 
> Also, please provide tests. Either checking sizes or using the image comparison we have in other areas to check rendered output.

As for my senario, it is for FileDialog only, the reason is the file dialog too small.

Jacalz:
This does make a lot of sense, but I think that it would make even more sense to have it as a standard feature for all dialog types. I could see scenarios where the user has a custom dialog and wants to resize it to better fit the screen.

Fyi, it looks like the tests are failing.

ScottHuangZL:
> This does make a lot of sense, but I think that it would make even more sense to have it as a standard feature for all dialog types. I could see scenarios where the user has a custom dialog and wants to resize it to better fit the screen.
> 
> Fyi, it looks like the tests are failing.

Test should be passing now.  Either for bigger size or wrong small size.

andydotxyz:
Red across the board. I guess some merge issue at your end because these tests don't compile on Travis.

ScottHuangZL:
Let me close this and create a new pull request.

PR #1221, opened 2020-07-26
Update radio widget to run OnChanged() on SetSelected()

Jacalz:
### Description:
This changes the radio widget to be in line with the select widget to actually run the OnChanged() function on setting a value. This had me a bit confused as I was used to how the select widget works.

### Checklist:

- [x] Tests included.
  - I updated the test to take this into account :)
- [x] Tests all pass.


Jacalz:
I got an error that I don’t think should be related to my PR. Would you mind kicking travis? @andydotxyz

PR #1222, opened 2020-07-27
Add Resize func for FileDialog and all other dialogs

ScottHuangZL:
Resize dialog, call this function after dialog show

### Description:

Fixes #(issue)

### Checklist:

- [Y] Tests included.
- [Y] Lint and formatter run with no errors.
- [Y] Tests all pass.

#### Where applicable:

- [Y] Public APIs match existing style.


ScottHuangZL:
All checks have passed. Please help to merge accordingly.  Thanks.

ScottHuangZL:
Oh, may be you want merge to develop firstly and then master. Please adjust it accordingly. Thanks.

andydotxyz:
Yes it needs to apply on `develop`, I have changed it's base and the tests will need to run again.

However you did not respond to @Jacalz and my requests that this be a feature of all `Dialog` types rather than just `FileDialog` - can you comment on that?

dialog/file_test.go
```diff
@@ -57,6 +59,92 @@ func TestEffectiveStartingDir(t *testing.T) {
...
+
+	//Test resize - bigger size
+	size := fyne.NewSize(800, 600)
+	file.Resize(size)
+	maxSize := file.dialog.win.Size()
+	minSize := file.dialog.win.MinSize()
+	//Test width
+	expectedWidth := size.Width
+	if size.Width > maxSize.Width {
+		expectedWidth = maxSize.Width
```
andydotxyz, line 89: A test case should set things up and they verify against known values.
This should be checking that the dialog is the expected size, the bounds checking should not be needed.

dialog/file_test.go
```diff
@@ -57,6 +59,92 @@ func TestEffectiveStartingDir(t *testing.T) {
...
+	} else if size.Width < minSize.Width {
+		expectedWidth = minSize.Width
+	}
+	assert.Equal(t, expectedWidth, file.dialog.win.Size().Width)
+	//Test height
+	expectedHeight := size.Height
+	if size.Height > maxSize.Height {
+		expectedHeight = maxSize.Height
```
andydotxyz, line 97: as above

ScottHuangZL:
> Yes it needs to apply on `develop`, I have changed it's base and the tests will need to run again.
> 
> However you did not respond to @Jacalz and my requests that this be a feature of all `Dialog` types rather than just `FileDialog` - can you comment on that?

Yeah, it should be doable.

dialog/file_test.go
```diff
@@ -57,6 +59,92 @@ func TestEffectiveStartingDir(t *testing.T) {
...
+
+	//Test again - for extreme min size
+	size = fyne.NewSize(1, 1)
+	file.Resize(size)
+	maxSize = file.dialog.win.Size()
+	minSize = file.dialog.win.MinSize()
+	contentMinSize := file.dialog.win.Content.MinSize()
+
+	//Test width
+	if contentMinSize.Width < minSize.Width {
```
andydotxyz, line 111: It's not clear why there are so many `if` branches in this test.
It seems like it should be much simpler to test the correct size when you have asked for a) a resize, b) a size below minSize and c) a size above maxSize...

dialog/file_test.go
```diff
@@ -57,6 +59,92 @@ func TestEffectiveStartingDir(t *testing.T) {
...
+		expectedWidth = size.Width
+		if size.Width > maxSize.Width {
+			expectedWidth = maxSize.Width
+		} else if size.Width < minSize.Width {
+			expectedWidth = minSize.Width
+		}
+		assert.Equal(t, expectedWidth, file.dialog.win.Size().Width)
+	}
+	//Test height
+	if contentMinSize.Height < minSize.Height {
```
andydotxyz, line 129: As above - It's hard to know that this test is working as the code has more conditions than the code under test.

ScottHuangZL: Thanks for your review.  Not a problem, I have refactor the testing.

ScottHuangZL:
Hello @andydotxyz , can you double check again? Thanks.

Jacalz:
But @ScottHuangZL, it seems like it still is for file dialog and not for all dialogs, like we discussed earlier? Did you accidentally not push all of your commits?

ScottHuangZL:
> But @ScottHuangZL, it seems like it still is for file dialog and not for all dialogs, like we discussed earlier? Did you accidentally not push all of your commits? 🙃

It seems need change the Dialog interface to add this func, right?  Or else, another way is to add this function for confirm/file/information/progress/progressinfinite one by one. May I know your opinion? thanks.

andydotxyz:
> Hello @andydotxyz , can you double check again? Thanks.

The test is much cleaner thanks.

Jacalz:
I think it would be much better to change the dialog interface than to edit each dialog individually. That would be a lot of duplicated code 😅

andydotxyz:
It may be impossible to do directly as we don't expose different types for every dialog. FileDialog is special because it has some custom fields that can be set.

ScottHuangZL:
I add Resize into Dialog interface, and add necessary testing for Dialog/ConfirmDialog/ProgressInfiniteDialog, all pass.   Please help to merge the pull request accordingly. Thanks.

Btw, the Resize func usage:
```go
			fd := dialog.NewFileOpen(func(reader fyne.URIReadCloser, err error) {
				if err == nil && reader == nil {
					return
				}
				if err != nil {
					dialog.ShowError(err, win)
					return
				}

				fileOpened(reader)
			}, win)
			fd.SetFilter(storage.NewExtensionFileFilter([]string{".png", ".txt"}))
			fd.Show()
			fd.Resize(fyne.NewSize(500, 300))


			info := dialog.NewInformation("Information", "You should know this thing after resize...", win)
			info.Resize(fyne.NewSize(300, 200))
```

ScottHuangZL:
> I think it would be much better to change the dialog interface than to edit each dialog individually. That would be a lot of duplicated code 😅

Complete enhancement and add necessary testing.

ScottHuangZL:
> It may be impossible to do directly as we don't expose different types for every dialog. FileDialog is special because it has some custom fields that can be set.

Complete enhancement the Dialog interface; Add necessary testing, all passed, thanks.

ScottHuangZL:
@andydotxyz How are you doing? I have complete add Resize for all dialogs. Do you have time to review and merge? thanks.

andydotxyz:
Before merging I would like to get a 👍 from @toaster as well - he has a good view of the impact of API changes.

ScottHuangZL:
> I think this looks pretty good now thanks

Thanks for the comment!

ScottHuangZL:
@andydotxyz May I know when this develop branch would merge to master? In other word, when my 2 pull requests (this and https://github.com/fyne-io/fyne/pull/1183) would show in master?   thanks.

andydotxyz:
It will be at the next release. Probably 1.3.3 - I might start preparing that soon for release in a week or so

ScottHuangZL:
Thank you Andy.  Looking forward to see in master branch soon :)

ScottHuangZL:
> It will be at the next release. Probably 1.3.3 - I might start preparing that soon for release in a week or so

@andydotxyz I find my another pull request about ScrollToBottom already merge to 1.3.3 master branch, thanks.  Could you advise why this "Resize" pull request still not merge into same master branch? thanks.

andydotxyz:
Indeed, 1.3.3 was a minor bugfix release.
When discussing the release plans we thought this change was more complex than a bug fix.
The develop branch in its entirety will merge to master for 1.4 release.

ScottHuangZL:
Thanks for your explanation. It is make sense, thanks. /Scott

Try431:
Is there a way to get access to the branch that has this Dialog Resize functionality? I see that it was merged into a patch-1 branch, but that looks to be private. 

Thanks!

andydotxyz:
It is on the ‘develop’ branch with the other code that will release in 1.4

Try431:
Ah, thanks. I had done a go get of the develop branch, but I ended up needing to restart VSCode before the changes would load.

angelopolotto:
Amazing PR!!! How Can I access the develop version of Fyne?

andydotxyz:
If you are using modules just specify the latest commit hash from `develop` branch as your version.
If you are not then go to your ~/go/src/fyne.io/fyne and `git pull; git checkout develop`.

Otherwise you can wait until 30th October when this is all due to be released as 1.4.

angelopolotto:
I tried `require fyne.io/fyne develop` in go module, but I didn't have success.

andydotxyz:
You need to use an actual commit hash not just ‘develop’ sorry. Try fc4e4ad as that is current just now.

PR #1224, opened 2020-07-28
Input Dialog

charlesdaniels:
This is #1187, I goofed up and force pushed to my fork when I shouldn't. I later un-did that, but apparently GitHub isn't smart enough to realize that. Oops, sorry everybody.


dialog/input.go
```diff
@@ -0,0 +1,81 @@
...
+	d.content = content
+
+	// hide the dialog, and empty the text
+	d.dismiss = widget.NewButton("Cancel", func() {
+		entry.Text = ""
+		d.Hide()
+	})
+	d.dismiss.Icon = theme.CancelIcon()
+
+	// we will override the OnTapped function later once we've instanced
```
andydotxyz, line 52: I don't see why this is needed - if you define the inputdialog first then all you reference is the entry, which is already defined.

dialog/input.go
```diff
@@ -0,0 +1,81 @@
+package dialog
+
+import (
+	"fyne.io/fyne"
+	"fyne.io/fyne/theme"
+	"fyne.io/fyne/widget"
+)
+
+// InputDialog is a variation of a dialog which prompts the user to enter some
+// text, with an optional validation function.
```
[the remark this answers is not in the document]
charlesdaniels: Fixed in 923b85cf

dialog/input.go
```diff
@@ -0,0 +1,81 @@
...
+	*dialog
+
+	onConfirm func(string)
+
+	entry *widget.Entry
+}
+
+// SetText changes the current text value of the input dialog, this can
+// be useful for setting a default value.
+func (i *InputDialog) SetText(s string) {
```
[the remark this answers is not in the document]
charlesdaniels: Could you elaborate on what you mean by "placeholder"?

dialog/input.go
```diff
@@ -0,0 +1,81 @@
...
+	i.entry.SetText(s)
+}
+
+// NewInput creates a dialog over the specified window for the user to enter
+// a value, possibly with validation.
```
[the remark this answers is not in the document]
charlesdaniels: 923b85cf

dialog/input.go
```diff
@@ -0,0 +1,81 @@
...
+//
+// onConfirm is a callback that runs when the user enters a valid string of
```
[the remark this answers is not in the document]
charlesdaniels: 923b85cf

dialog/input.go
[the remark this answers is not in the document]
andydotxyz: entry.SetPlaceholder() - the (disabled) text that is shown if no text is specified...

dialog/input.go
charlesdaniels: I think my thought at the time was that `onConfirm` may have been changed after the fact, but that cannot happen since it is a private field.

Fixed in 87558a48

dialog/input.go
```diff
@@ -0,0 +1,81 @@
...
+// text and clicks the "confirm" button. May be nil.
```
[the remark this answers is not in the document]
charlesdaniels: `onConfirm` can be `nil`. AFAIK that is allowed in Golang's equivalent of a function pointer.

dialog/input.go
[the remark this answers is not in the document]
charlesdaniels: Ahh, good thinking. Fixed in ac258613.

dialog/input.go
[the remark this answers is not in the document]
andydotxyz: Sory, I misread that as referring to the string parameter

charlesdaniels:
Tests and linter pass on my box, ready for review when you are!

dialog/input.go
```diff
@@ -0,0 +1,87 @@
...
+	confirmButton *widget.Button
+}
+
+// SetText changes the current text value of the input dialog, this can
+// be useful for setting a default value.
+func (i *InputDialog) SetText(s string) {
+	i.entry.SetText(s)
+}
+
+// GetText returns the current text value entered in the input dialog.
```
andydotxyz, line 27: This access is not something we have done with other dialogs, can you share why it was added?

dialog/input_test.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func TestInputDialogConfirm(t *testing.T) {
+	ch := make(chan string)
+	i := NewInput("test1", "test2", func(s string) { ch <- s }, test.NewWindow(nil))
+	i.Show()
+
+	assert.False(t, i.win.Hidden)
+	test.Type(i.entry, "test3")
+	go test.Tap(i.confirmButton)
```
andydotxyz, line 18: I think the tests work with synchronos tapping - did it not work without the goroutine?

dialog/input.go
charlesdaniels: The rationale was that if the user cancelled, you might still want access to the text? If that's stylistically inconsistent, it can be removed. I think the use case is probably pretty niche.

dialog/input_test.go
charlesdaniels: I cargo-cult-ed some of the other tests which used goroutines, I can switch to synchronous if that is preferred.

dialog/input.go
andydotxyz: Aha! I think the other ones call the callback on cancel with a nil value.
If you look at the confirm dialog I am pretty sure how that works.
If people use the ShowxxxDialog() function there is no reference to get the text after cancelling without callback so that's probably why we took that approach.

dialog/input_test.go
andydotxyz: Ah well, if it is consistent just leave it ;)
When we review dialogs properly this could more easily get revised with everything else.

dialog/input.go
charlesdaniels: So how would you handle that in this case? I kinda feel like there should be a separate callback for if the dialog is cancelled, since presumably the one that receives the value entered cannot do it's job without having a proper value.

Alternatively, I could have it take a `bool` argument that indicates if the user pressed cancel or confirm?

I don't like the idea of just calling it with the empty string, since in some cases that might be a legitimate input.

andydotxyz: I agree that we could do better to separate confirm vs dismiss callbacks.
However I also think that it should be consistent.
To fix the others is the aim but this is a breaking (2.0) change.

There is an OnClosed() callback that the developer can use in the meantime - this should be called irrespective of confirm or cancel. I guess given that we could have the input callback (in this case - with clear documentation) only call on confirm might just be ok.

dialog/input.go
```diff
@@ -0,0 +1,87 @@
...
+func (i *InputDialog) GetText() string {
+	return i.entry.Text
+}
+
+// SetPlaceholder defines the placeholder text for the entry
```
andydotxyz, line 32: This mismatch on capitalisation here is causing the Travis failures

charlesdaniels:
> But I also wanted to ask... if there will ever be any other type if input dialog should this be named less generically? like NewEntryDialog or NewStringInputDialog?

I agree, I think I will rename it to `NewEntryDialog`. It think that encapsulates it's purpose well. 

Agree with the other suggestions, will implement.

dialog/input.go
charlesdaniels: I went ahead and implemented this in 11d2c35b, just in case it might be useful for someone.

dialog/input.go
charlesdaniels: Fixed in e20f0f7f

andydotxyz:
Can't you use the same pattern as the Dialog interface and simply:
`func (d *dialog) SetOnClosed(closed func())`

Passing another function on the parameter list seems rather confusing. 
For OnClosed we are not passing bool in other dialogs - why here?

charlesdaniels:
I have changed it to match the conventions as discussed in 93bd6115.

dialog/entry.go
```diff
@@ -0,0 +1,111 @@
...
+
+// SetPlaceholder defines the placeholder text for the entry
+func (i *EntryDialog) SetPlaceholder(s string) {
+	i.entry.SetPlaceHolder(s)
+}
+
+// SetOnClosed changes the onClosed callback, which is nil by default.
+//
+// onClosed is called unconditionally weather the user confirms or cancels. The
+// May be
```
andydotxyz, line 40: A copy/paste issue here maybe?

dialog/entry.go
```diff
@@ -0,0 +1,111 @@
...
+	onClosed func()
+}
+
+// SetText changes the current text value of the entry dialog, this can
+// be useful for setting a default value.
+func (i *EntryDialog) SetText(s string) {
+	i.entry.SetText(s)
+}
+
+// GetText returns the current text value entered in the entry dialog.
```
andydotxyz, line 27: I lot has changed in this area so I wanted to check before this wraps up - you're sure this API is useful?

dialog/entry.go
charlesdaniels: Fixed in 775dc541

dialog/entry.go
charlesdaniels: Agreed, I don't think there is much use case, so I went ahead and nuked it in 775dc541.

PR #1225, opened 2020-07-29
Reduce the binary size when packaging

lonng:
Signed-off-by: Lonng <heng@lonng.org>

### Description:

This PR tries to strip the DWARF tables needed for debuggers and reduce the binary size significantly. In may case, which can reduce the size from 30MB to 14MB. The size can be reduced to 5MB after applying upx.

See： https://blog.filippo.io/shrink-your-go-binaries-with-this-one-weird-trick/

### Checklist:

- [x] Tests included.
   manual test

- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne/build.go
+++ b/cmd/fyne/build.go
@@ -9,6 +9,7 @@ import (
 
 type builder struct {
 	os, srcdir string
+	release    bool
 }
 
 func (b *builder) build() error {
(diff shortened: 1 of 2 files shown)


lonng:
I have no idea about CI failure.

andydotxyz:
Looks like an intermittent error on Travis, I have kicked it to see if it helps

andydotxyz:
Great addition thanks.
Anything that helps reduce release file size, especially without slowing startup time :)

andydotxyz:
Oh no! this should not have been off master.
Cherry picking to develop then reverting the master change.
This needs more testing and will go in a future release.

PR #1227, opened 2020-07-30
Use glfw keycodes instead of hardcoded chars

### LINKED ISSUE #1220: Copy-paste via keyboard don't work translated keyboard mappings on Windows.
### Describe the bug:
Copy-paste via keyboard don't work for non-English keyboards on Windows. 

### To Reproduce:
1. Run fyne_demo
2. Select input tab from Widgets
3. Copy some text into clipboard
4. Change keyboard to something non-English via Alt-Shift
5. Try to paste
6. Nothing happens

 - OS: Windows
 - Version: 10
 - Go version: 1.13.5
 - Fyne version: current master - https://github.com/fyne-io/fyne/commit/aea152d547caebfc547dbad6ad0abdd35ad24b34

RustamGamidov:
### Description:
Current solutions is using English letters to get an internal key aliases. So there are features that can work on English keyboard only like shortcuts.
Proposed solution is reusing glfw  key-codes similar to functional and numeric keys.

Fixes #1220 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
Can you provide a little more information about why this change fixes the issue? Does the GLFW_A key not generate ‘a’? And if not, is it the correct key to trigger a ctrl+A shortcut?

RustamGamidov:
Logging `key` at the very begging of the `keyPressed` gives me localized char.
`keyNameMap` contains English letters so we have empty string as a result of `keyToName(...)` call
and this omits shortcuts detection code.

RustamGamidov:
sorry, messed up my tries a bit
` glfw.GetKeyName(code, scancode)` is giving localized char and it's not present on the `keyNameMap` so we are getting empty string with not working shortcuts as a result.

RustamGamidov:
just tried on archlinux with XFCE and it has same issue (shortcust are not working on non english keyboard) so it's not windows special handling

andydotxyz:
Can you be a bit more specific about what values /are/ being seen instead of 'a'?
I don't know much about how people expect shortcuts to work with non-english keyboards.
The current system was set up in a change to support some keyboards (AZERTY was tested primarily)
and your change largely reverts this fix: https://github.com/fyne-io/fyne/commit/7f3741ec609d3cc14f87b60e9ab04799928c41b6.

By design we are aiming to use logical characters instead of keyboard code based output which may introduce complications in the attempt to be more consistent across platforms.
Let me know what you think.

Kvaz1r:
AZERTY mismatching looks like glfw issue - [Key events do not map to physical keys](https://github.com/glfw/glfw/issues/114).

RustamGamidov:
image
that how English 'A' button looks on Russian keyboards. Image is better here because i'm not sure the font you are using has corresponding chars. Below is the output i have pressing same button in english and russian layout:
```
2020/07/31 08:39:57 keyName:  a ; code: 65 ; scancode:  30
2020/07/31 08:40:01 keyName:  ф ; code: 65 ; scancode:  30
```
everything as expected. 
I'm using QUERTY, not sure how to check if glfw keys are working on AZERTY and

Using localized chars to determine the key pressed requires map for each language supported. One more thing is number of letters is different and for example in russian there are letters on the `[];'` buttons so even `Key#` is not good.
It could be easier to base on key codes and map them to the human friendly char before showing to the user. This way fyne knows exact button pressed and user has local chars in the hints.

andydotxyz:
> AZERTY mismatching looks like glfw issue - [Key events do not map to physical keys](https://github.com/glfw/glfw/issues/114).

That issue was fixed a long time ago and we a new enough version that the fix is in.

andydotxyz:
I'm concerned that there is not a single approach that works here.
In essence what this fix is looking for is for the key to be 'A' when using a shortcut but 'ф' when typing.
Is that the way that other apps work? It seems confusing, but I don't really understand expectations on this stuff.

andydotxyz:
> I'm using QUERTY, not sure how to check if glfw keys are working on AZERTY

If you open your keyboard preferences and tell it you have an AZERTY keyboard you should be able to do that.
It's not in the same place I don't think because this is about hardware rather than locale.

RustamGamidov:
Yes, my expectations are tied to the key, not to the char and it's quite common among the people i'm usually communicationg
In addition i've met quite negative feedback to localized shortcuts in the video-gaming area.

In this particular case i would say it's more related to the framemwork inner logic of the keys representation.
And mapping to the local char can be done right before the displaying.
Alternative solution could be moving shortcuts detection to the keycodes than nothing is changed for the user

andydotxyz:
Remember that in the main toolkit we cannot use keycodes etc because some of the OS we support don't even use a hardware keyboard (iOS, Android for example). So I think this means that whatever logic we use needs to be inside the glfw driver.

For this reason I don't think the shortcut code can use keyboard codes, as shortcuts are created outside the driver keyboard code.

andydotxyz:
Sorry this has not moved forward. Can you please confirm when you have tested AZERTY to ensure that this does not break the previous fix? @RustamGamidov

RustamGamidov:
Sorry, i've switched to another project a bit.
I've just checked the French azerty on windows 10 and the behavior is changed, at least Ctrl+A because it's A char corresponds to another keycode. 
It looks like there should be querty/azerty/dvorak mapping in addition to the keycodes, i hoped this is done by the OS.

andydotxyz:
Yes, the problem is that key codes relate to the hardware, whereas the mapping we were doing before relates to the output character.
The change introduced as it stands ignores the mapping that was in place to support the OS mapping.
Perhaps another solution would be to add U+0410 to the character mapping table and both will work?

RustamGamidov:
Yes, this PR is absolutely not ready to be merged. 
And one more yes :) adding this new char code will solve the issue in short term. Then other shortcuts should also be mapped because they are different as well. In future all languages with non english chars will cause the same issue. That's why i made this attempt, quite naive as it appeared.

andydotxyz:
Thanks for working on this though - it's not an easy issue.
Would you like to keep this PR open to track through or start another one with a different fix later?

RustamGamidov:
i'll close this one not to create noise, it's still available so given arguments can be referenced

andydotxyz:
Tring to find a more complete solution that works for other languages too.. If you're interested in helping with this I'm pulling together a doc at https://github.com/fyne-io/fyne/wiki/Keyboard-API

PR #1228, opened 2020-07-30
Add URI.Name()

andydotxyz:
Feels like we missed this - used lots and even more as we start using URI in file listings

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
Yes, Query etc is an interesting idea once we move away from mostly file path etc.
URI can easily expand over time I think

PR #1236, opened 2020-08-05
Update fileitem.go

### LINKED ISSUE #1235: dialog filter the dir name start with .
### Describe the bug:

In the windows home directory
There are many directories starting with `.`
But dialog will filter it with a blank name 

This is the screenshot about the problems
dialog

The real names are these:
image

godcong:
fix issue #1235

### Description:

Fixes #1235

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne/build.go
+++ b/cmd/fyne/build.go
@@ -9,7 +9,6 @@ import (
 
 type builder struct {
 	os, srcdir string
-	release    bool
 }
 
 func (b *builder) build() error {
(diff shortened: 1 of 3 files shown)


Jacalz:
It looks like the Travis tests failed due to incorrect formatting. You can run the [git commit hooks](https://github.com/fyne-io/fyne/wiki/GitHooks) for easily fixing it or running formating and linters manually.

andydotxyz:
Yeah - go formatting uses tabs for indent, in this commit they are converted to spaces

andydotxyz:
Also I don't think this is the correct fix - it will only improve visibility of directory names.
I have another patch that will resolve it for files as well

godcong:
> Also I don't think this is the correct fix - it will only improve visibility of directory names.
> I have another patch that will resolve it for files as well

Ok, if you solved this problem.
you can close it.

andydotxyz:
Some directories do have suffixes - like "MyApp.app" on MacOS etc.
Take a look at my fix and see what you think.

andydotxyz:
Thanks, upon discussion the team agree that your change should also be landed :).

PR #1240, opened 2020-08-09
Make buttons rely on shadow more than colour

andydotxyz:
More like material design.
This meant adding more padding so shadows come into the body

Follows the thread of "incremental design improvements"

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz:
Just a little strange thing with the second image of my examples where it looks slightly weird. I just remembered that there is an empty label in between the buttons, but it certainly doesn’t like like so now. I will of course need to remove it to look better, but it almost feels like the empty label disappears between the buttons on the logon screen.

andydotxyz:
In the first of your "bad" images it is perhaps a little off... but I think that some revision of the entry widget can help with that - it currently draws right up to the edge, but the spec shows a thinner underline inset from the boundary.
From the second image I can't really see what you mean about it looking bad, sorry.

Jacalz:
It is probably no big deal, but what I mean with the second image is that the two buttons have a strange distance between them. I had a label between them to get some distance between them because it looked better with the old button style. I am just concerned that the second image looks almost as though there isn’t any empty label between the buttons 😅

andydotxyz:
Hmm. The space between two buttons is now more than it used to be (between the inside edges of the shadows is 3*padding whereas it used to be 1*. So the gap between the shadows is around 2*padding instead of almost zero.
I suspect that you have found a bug if the label between them is not spacing them out...

andydotxyz:
Looks like button text minsize has a glitch, will look into it.
Adding a line for separation should not be any more or less needed than before, so I'd like to keep that separate...

andydotxyz:
The errors for the ugly theme relate to the comments in the tests about MinSize not being updated.
I have forced a refresh to resolve this.

andydotxyz:
Updating the branch - test image failures relate to another merge

andydotxyz:
Screenshot 2020-08-12 at 10 35 03
Screenshot 2020-08-12 at 10 35 27

andydotxyz:
Request for changes were resolved over a week ago, so expiring the block.
No other objections were raised, so merging.

PR #1243, opened 2020-08-10
Add very beginning of storage API

### LINKED ISSUE #1109: Allow the user to read the location of app configuration directory

Yes. For the default implementation of `fyne.App`, it uses the fixed path to store the settings and preferences by the different OS, like this,
```go
func rootConfigDir() string {
	homeDir, _ := os.UserHomeDir()

	desktopConfig := filepath.Join(homeDir, ".config")
	return filepath.Join(desktopConfig, "fyne")
}
```

For settings, the path of its configuration file is that as following,
```go
// StoragePath returns the location of the settings storage
func (sc *SettingsSchema) StoragePath() string {
	return filepath.Join(rootConfigDir(), "settings.json")
}

func (s *settings) load() {
	err := s.loadFromFile(s.schema.StoragePath())
	if err != nil {
		fyne.LogError("Settings load error:", err)
	}

	s.setupTheme()
}
```

That's, for all kinds of fyne apps, those settings files will use the same one.  It may be not what we want.

I hope that there is an API to allow the user to customize it.

Yes.

Add a implicit global variable to save the setting of the user, and a exported function to modify it. If the variable is not empty, `rootConfigDir` returns it. Or return the one supplied by the default implementation. Like this
```go
var rootDir string

func SetRootConfigDir(dir string) {
	rootDir = dir
}

func rootConfigDir() string {
	if rootDir != "" {
		return rootDir
	}

	homeDir, _ := os.UserHomeDir()
	desktopConfig := filepath.Join(homeDir, ".config")
	return filepath.Join(desktopConfig, "fyne")
}
```

andydotxyz:
Currently just an abstraction to get information that preferences uses
Fixes #1109

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app.go
+++ b/app.go
@@ -54,6 +54,9 @@ type App interface {
 
 	// Preferences returns the application preferences, used for storing configuration and state
 	Preferences() Preferences
+
+	// Storage returns a storage handler specific to this application.
+	Storage() Storage
 }
 
 var app App
--- a/app/app.go
+++ b/app/app.go
@@ -21,6 +21,7 @@ type fyneApp struct {
 	uniqueID string
 
 	settings *settings
+	storage  *store
 	prefs    fyne.Preferences
 	running  bool
 	runMutex sync.Mutex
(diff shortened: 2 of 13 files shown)


andydotxyz:
This is mininmal. Not normally worth creating an interface for 1 call - but I plan on adding the proposed `Create()`, `Open()`, `Remove()` etc to this interface as well.

Currently just App.Storage().RootURI() to close the feature request.

PR #1249, opened 2020-08-14
RFC: Add feature to set custom text for ProgressBarIninite

Jacalz:
### Description:
This is a work in progress effort of adding support for setting a custom text instead of the percentage on the progress bar. This could come in handy for saying that a specific event that tracks it's progress either failed or something like that.

The plan is to also add the feature to progressBarInifinite and then add some tests also.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/progressbar.go
+++ b/widget/progressbar.go
@@ -58,7 +58,7 @@ func (p *progressRenderer) applyTheme() {
 }
 
 func (p *progressRenderer) BackgroundColor() color.Color {
-	return theme.ButtonColor()
+	return theme.ShadowColor()
 }
 
 func (p *progressRenderer) Refresh() {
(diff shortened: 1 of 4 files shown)


andydotxyz:
Hmm, this is a good thing to add. But two possible considerations:

1) Should it be a format rather than a simple text? (i.e. "%f.0 bananas")
2) If not, then maybe just "ProgressBar.Text" could work - not sure if "Custom" is needed

Jacalz:
This probably doesn't make a whole lot of sense actually. Closing for now.

develar:
But how to set custom text for infinite progressbar?

andydotxyz:
There is no text feature on the infinite progress bar. You can put a label or text component over the bar by adding both to a container and using the MaxLayout so they fill the same space with one over the other.

develar:
I really like the way you are thinking. As result, it builds a solid solution. Thanks.

PR #1251, opened 2020-08-16
Add TextFormatter callback for ProgressBar

Jacalz:
### Description:
This adds a function callback for using a custom farmat on progress text.

progressbar-formatting

Surpasses #1249 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne_demo/screens/widget.go
+++ b/cmd/fyne_demo/screens/widget.go
@@ -201,11 +201,25 @@ func makeInputTab() fyne.Widget {
 
 func makeProgressTab() fyne.Widget {
 	progress = widget.NewProgressBar()
+	
+	format := widget.NewProgressBar()
+	format.TextFormatter = func() string {
+		return fmt.Sprintf("%.2f out of %.2f", format.Value, format.Max)
+	}
+	
+	go func() {
+		for format.Value < format.Max {
+			time.Sleep(time.Second)
+			format.SetValue(format.Value+0.01)
+		}	
+	}()
+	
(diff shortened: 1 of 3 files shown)


andydotxyz:
Hmm, not sure about "TextFormatter" as the field name, as it's not really a text formatter that returns...
Not got any suggestions right now, will think. Anyone else got thoughts?

Jacalz:
Honestly, me neither. I used that name because it was what @stuartmscott used in his example on Slack and because it was miles ahead of _TextOnValueChanged()_ that I had come up with 😅

andydotxyz:
I'm still thinking. Not come up with anything better - just reasons that despite my reaction it might be a good fit...

cmd/fyne_demo/screens/widget.go
```diff
@@ -201,11 +201,25 @@ func makeInputTab() fyne.Widget {
 
 func makeProgressTab() fyne.Widget {
 	progress = widget.NewProgressBar()
+
+	format := widget.NewProgressBar()
+	format.TextFormatter = func() string {
+		return fmt.Sprintf("%.2f out of %.2f", format.Value, format.Max)
+	}
+
+	go func() {
```
andydotxyz, line 210: Why not use the same loop as the other two progress bars?

Jacalz: That's a good question. I honestly hadn't seen it. Thanks for pointing it out  :sweat_smile:

PR #1253, opened 2020-08-20
Add a list widget that caches items for performance

okratitan:
This is a PR of a list widget implementation that I have created to be considered for Fyne 1.4
It still needs tests and perhaps some cleanup so I'll mark it as draft for now

APIs:
Methods:
OnNewItem func() fyne.CanvasObject
OnUpdateItem func(index int, item fyne.CanvasObject)
OnItemSelected func(index int, item fyne.CanvasObject)
Fields:
Length


cmd/fyne_demo/screens/widget.go
```diff
@@ -274,6 +274,34 @@ func stopProgress() {
...
 
+func makeListTab() fyne.CanvasObject {
+	var data []string
+	for i := 0; i < 1000; i++ {
+		data = append(data, fmt.Sprintf("Test Item %d", i))
+	}
+
+	icon := widget.NewIcon(nil)
+	label := widget.NewLabel("Select An Item From The List")
+	hbox := fyne.NewContainerWithLayout(layout.NewHBoxLayout(), icon, label)
```
stuartmscott, line 285: Could try using a Split Container to see how the list copes with many resizes as the divider is moved

widget/list.go
```diff
@@ -0,0 +1,304 @@
...
+
+	"fyne.io/fyne"
+	"fyne.io/fyne/canvas"
+	"fyne.io/fyne/driver/desktop"
+	"fyne.io/fyne/internal/widget"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/theme"
+)
+
+type listItemRenderer struct {
```
stuartmscott, line 15: Please follow https://github.com/fyne-io/fyne/issues/827

cmd/fyne_demo/screens/widget.go
```diff
@@ -274,6 +274,34 @@ func stopProgress() {
...
+
+	list := widget.NewList()
+	list.Length = len(data)
```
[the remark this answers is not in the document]
okratitan: This should be resolved now.

cmd/fyne_demo/screens/widget.go
okratitan: This should be resolved.

widget/list.go
```diff
@@ -0,0 +1,413 @@
...
+	l.ExtendBaseWidget(l)
+
+	return l.BaseWidget.MinSize()
+}
+
+// Refresh updates list to match the current theme
+func (l *List) Refresh() {
+	if l.background != nil {
+		l.background = theme.BackgroundColor()
+	}
```
stuartmscott, line 74: Not sure about this, perhaps `if l.background == nil {`? But is this needed since listRenderer.BackgroundColor() already does this check?

widget/list.go
```diff
@@ -0,0 +1,413 @@
...
+
+		cachePoolCount := 5
+		if l.list.Length()-len(l.children) < 5 {
+			cachePoolCount = l.list.Length() - len(l.children)
+		}
+		for i := l.itemCache.Count(); i < cachePoolCount; i++ {
+			// Make sure to keep extra items in the cache
+			item := newListItem(l.list.CreateItem(), nil)
+			l.itemCache.Release(item)
+		}
```
stuartmscott, line 171: How much change in performance do you see without these lines?

widget/list.go
```diff
@@ -0,0 +1,412 @@
...
+func (li *listItemRenderer) BackgroundColor() color.Color {
+	if li.item.background == nil {
+		return theme.BackgroundColor()
+	}
+
+	return li.item.background
+}
+
+func (li *listItemRenderer) Refresh() {
+	if li.item.selected == true {
```
stuartmscott, line 321: `if <boolean-expression> == true {` seems overly verbose

widget/list.go
```diff
@@ -0,0 +1,412 @@
...
+	}
+	objects[len(objects)-1].Resize(fyne.NewSize(l.list.size.Width, l.list.itemMin.Height))
+}
+
+func (l *listLayout) itemPrepended(objects []fyne.CanvasObject) {
+	objects[0].Move(fyne.NewPos(0, objects[1].Position().Y-l.list.itemMin.Height))
+	objects[0].Resize(fyne.NewSize(l.list.size.Width, l.list.itemMin.Height))
+}
+
+type cachePool interface {
```
stuartmscott, line 377:
```suggestion
type pool interface {
```

widget/list.go
```diff
@@ -0,0 +1,412 @@
...
+	Count() int
+	Obtain() fyne.CanvasObject
+	Release(fyne.CanvasObject)
+}
+
+type listCache struct {
```
stuartmscott, line 383:
```suggestion
type listPool struct {
```

widget/list.go
```diff
@@ -0,0 +1,412 @@
...
+	availableCache []fyne.CanvasObject
```
stuartmscott, line 384:
```suggestion
	contents []fyne.CanvasObject
```
A slice holding the contents of the pool

widget/list.go
okratitan: Honestly I look at other related widgets to see what they do and this was code that Refresh() in widget/box.go had

widget/list.go
okratitan: I would say performance is definitely noticeable visually without these lines.

widget/list.go
```diff
@@ -0,0 +1,412 @@
...
+	BaseWidget
+
+	Length         func() int
+	CreateItem     func() fyne.CanvasObject
+	UpdateItem     func(index int, item fyne.CanvasObject)
+	OnItemSelected func(index int, item fyne.CanvasObject)
+
+	background      color.Color
+	offsetY         int
+	previousOffsetY int
```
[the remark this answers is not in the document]
okratitan: I like it for readability of understanding what offset we are using if just looking at the code. And horizontal is always possible in the future.

okratitan:
> Please remove the changes to theme/ and vendor/ from this PR

Done

widget/list.go
```diff
@@ -0,0 +1,412 @@
...
+	"fyne.io/fyne"
+	"fyne.io/fyne/canvas"
+	"fyne.io/fyne/driver/desktop"
+	"fyne.io/fyne/internal/widget"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/theme"
+)
+
+// List is a widget that caches list items for performance and
+// lays the items out in a vertical direction inside of a scroller
```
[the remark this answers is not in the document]
okratitan: This should now be resolved

widget/list.go
okratitan: This should now be resolved

widget/list.go
okratitan: This should now be resolved

widget/list.go
okratitan: This should now be resolved

widget/list.go
```diff
@@ -0,0 +1,415 @@
...
+func (l *listRenderer) BackgroundColor() color.Color {
+	if l.list.background == nil {
+		return theme.BackgroundColor()
+	}
+
+	return l.list.background
+}
+
+func (l *listRenderer) Layout(size fyne.Size) {
+	if size != l.size {
```
andydotxyz, line 93: Is this needed? Layout should only be called when size has changed...

widget/list.go
```diff
@@ -0,0 +1,415 @@
...
+		return
+	}
+	if l.itemCache == nil {
+		l.itemCache = &listPool{}
+	}
+
+	offsetChange := l.list.previousOffsetY - l.list.offsetY
+
+	if offsetChange < 0 {
+		// Scrolling Down
```
andydotxyz, line 112: It feels like perhaps these could be broken out to other methods? It's a fairly beefy chunk of code

widget/list.go
```diff
@@ -0,0 +1,415 @@
...
+		for i := l.itemCache.Count(); i < poolCapacity; i++ {
+			// Make sure to keep extra items in the cache
+			item := newListItem(l.list.CreateItem(), nil)
+			l.itemCache.Release(item)
+		}
+	}
+}
+
+func (l *listRenderer) MinSize() fyne.Size {
+	return l.scroller.MinSize()
```
andydotxyz, line 176: This doesn't seem right - should it not be as wide as the minSize.Width of the template element?

widget/list.go
```diff
@@ -0,0 +1,415 @@
...
+}
+
+func (l *listRenderer) Refresh() {
+	l.Layout(l.list.Size())
```
andydotxyz, line 180: I'm not sure why the refresh causes a layout here - it may pair with the "size" cache that was added above.

widget/list.go
```diff
@@ -0,0 +1,415 @@
...
+
+// Count returns the number of available items in the cache
+func (lc *listPool) Count() int {
+	return len(lc.contents)
+}
+
+// Obtain returns an item from the cache for use
+func (lc *listPool) Obtain() fyne.CanvasObject {
+	cacheLength := len(lc.contents)
+	if cacheLength == 0 {
```
andydotxyz: This variable isn't needed along with the check below...
If you check len(lc.contents) and return nil then you don't need to also check it later.

andydotxyz (changes requested):
A few things inline - stuff can be simplified in a few places.

Having tests around this will help - particularly around Pool which feels like it should have been built test first.

widget/list.go
```diff
@@ -0,0 +1,415 @@
...
+// List is a widget that caches list items for performance and
+// lays the items out in a vertical direction inside of a scroller
+// List requires that all items are the same size
+type List struct {
+	BaseWidget
+
+	Length         func() int
+	CreateItem     func() fyne.CanvasObject
+	UpdateItem     func(index int, item fyne.CanvasObject)
+	OnItemSelected func(index int, item fyne.CanvasObject)
```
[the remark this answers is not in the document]
okratitan: Update one of their items that they might want to change visually on selection?

widget/list.go
```diff
@@ -0,0 +1,415 @@
...
+
+	background      color.Color
```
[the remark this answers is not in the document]
okratitan: this follows widget/box.go which keeps background in the same way.

widget/list.go
okratitan: layout is also called when the scroller offset has changed.

widget/list.go
okratitan: the min width of the template element, being inside of a vscroller, would force the scroller's min size to be that width anyway

widget/list.go
okratitan: because it refreshes on scroll change and calls layout

widget/list.go
okratitan: agree

widget/list.go
```diff
@@ -0,0 +1,415 @@
...
+	offsetY         int
+	previousOffsetY int
+	selectedItem    *listItem
+	selectedIndex   int
+	itemMin         fyne.Size
```
stuartmscott, line 31: `itemMin` is initialized in `CreateRenderer` and never updated. Consider this scenario: a list of items where each item has an icon and text. After the list is initialized and `CreateRenderer` is called, the theme gets changed to one with a larger `TextSize` and `InlineIconSize`. `itemMin` size is now too small

widget/list.go
stuartmscott: Note:
- `scroller.MinSize` returns `32 x 32` (`scrollContainerMinSize`) unless `scroller.SetMinSize` is called.
- `verticalscroller.MinSize` returns `scroller.Content.MinSize().Width x 32`

It might be better to return `l.layout.MinSize()` in this scenario.

stuartmscott: > It might be better to return l.layout.MinSize() in this scenario.

Actually this wont work as it will force the list to be big enough to hold all the content - no point having a scroller. I think Stephen's original code is the best approach

widget/scroller.go
```diff
@@ -444,6 +448,9 @@ func (s *ScrollContainer) updateOffset(deltaX, deltaY int) bool {
 	}
 	s.Offset.X = computeOffset(s.Offset.X, -deltaX, s.Size().Width, s.Content.MinSize().Width)
 	s.Offset.Y = computeOffset(s.Offset.Y, -deltaY, s.Size().Height, s.Content.MinSize().Height)
+	if s.onOffsetChanged != nil {
+		s.onOffsetChanged()
+	}
```
stuartmscott, line 453:
```suggestion
	if f := s.onOffsetChanged; f != nil {
		f()
	}
```
It is possible, though unlikely, that the field could be set nil after it is nil checked and before it is called. This suggestion guarantees that cannot happen.

widget/list.go
okratitan: Indeed fixed.

widget/list.go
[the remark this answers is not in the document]
okratitan: Also I suppose it is needed so that when items use theme.BackgroundColor() it doesn't stand out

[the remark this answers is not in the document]
okratitan: that said instead of saving the value we can just return theme.BackgroundColor() and don't need that field.  This should be resolved now.

widget/list.go
okratitan: a chunk of this code has already been broken out to l.appendItem() and l.prependItem() and layout.itemAppended and layout.itemPrepended.  To me this feels like the appropriate place to do the math based on the layout.  But I'm not going to complain about passing maths into two more append/prepend functions.

widget/list.go
okratitan: Yep

widget/list.go
okratitan: This should now be fixed

widget/scroller.go
okratitan: This should be fixed

widget/list.go
okratitan: This should be addressed now.

widget/list.go
```diff
@@ -394,16 +398,11 @@ func (lc *listPool) Count() int {
...
 		return nil
 	}
 	item := lc.contents[0]
-	if poolLength > 0 {
-		lc.contents = lc.contents[1:]
-	} else {
-		lc.contents = nil
-	}
+	lc.contents = lc.contents[1:]
 	item.Show()
```
stuartmscott, line 406: I'm not sure it is necessary to Hide/Show pool items since they will not be traversed when the list is rendered. Additionally as Show triggers Refresh and Hide triggers canvas.Refresh, there is a cost here and I have found better performance by instead just calling Refresh after obtaining and updating the item.

okratitan:
> This is a PR of a list widget implementation that I have created to be considered for Fyne 1.4
> It still needs tests and perhaps some cleanup so I'll mark it as draft for now
> 
> APIs:
> Methods:
> OnNewItem func() fyne.CanvasObject
> OnUpdateItem func(index int, item fyne.CanvasObject)
> OnItemSelected func(index int, item fyne.CanvasObject)
> Fields:
> Length

Tests have now been added

widget/list_test.go
```diff
@@ -0,0 +1,161 @@
...
+	visibleCount := len(test.WidgetRenderer(list).(*listRenderer).children)
+
+	assert.Equal(t, 1000, list.Length())
+	assert.GreaterOrEqual(t, list.MinSize().Width, template.MinSize().Width)
+	assert.Equal(t, list.MinSize(), test.WidgetRenderer(list).(*listRenderer).scroller.MinSize())
+	assert.Equal(t, 0, firstItemIndex)
+	assert.Equal(t, visibleCount, lastItemIndex-firstItemIndex+1)
+}
+
+func TestListResize(t *testing.T) {
```
stuartmscott, line 32: nit: `Test<struct-name>_<method-name>`
```suggestion
func TestList_Resize(t *testing.T) {
```

widget/list_test.go
```diff
@@ -0,0 +1,161 @@
...
+	children[0].(*listItem).Tapped(&fyne.PointEvent{})
+	assert.Equal(t, children[0].(*listItem).statusIndicator.FillColor, theme.FocusColor())
+	assert.Equal(t, list.selectedIndex, 0)
+	children[1].(*listItem).Tapped(&fyne.PointEvent{})
+	assert.Equal(t, children[1].(*listItem).statusIndicator.FillColor, theme.FocusColor())
+	assert.Equal(t, list.selectedIndex, 1)
+	assert.Equal(t, children[0].(*listItem).statusIndicator.FillColor, theme.BackgroundColor())
+}
+
+func testListDataChange(t *testing.T) {
```
stuartmscott, line 125: Where is this used? It look identical to the function above. If this is a test it should be exported.

widget/list.go
```diff
@@ -0,0 +1,414 @@
...
+
+// Layout the components of the button widget
+func (li *listItemRenderer) Layout(size fyne.Size) {
+	li.LayoutShadow(size, fyne.NewPos(0, 0))
+	li.layout.Layout([]fyne.CanvasObject{li.item.statusIndicator, li.item.child}, size)
+}
+
+func (li *listItemRenderer) BackgroundColor() color.Color {
+	return theme.BackgroundColor()
+}
```
stuartmscott, line 325: Duplicate of BaseRenderer

widget/list.go
```diff
@@ -0,0 +1,414 @@
...
+	children         []fyne.CanvasObject
+	visibleItemCount int
+	firstItemIndex   int
+	lastItemIndex    int
+	size             fyne.Size
+}
+
+func (l *listRenderer) BackgroundColor() color.Color {
+	return theme.BackgroundColor()
+}
```
stuartmscott, line 85: ditto

widget/list.go
```diff
@@ -0,0 +1,414 @@
...
+
+// MouseOut is called when a desktop pointer exits the widget
+func (li *listItem) MouseOut() {
+	li.hovered = false
+	li.Refresh()
+}
+
+// Tapped is called when a pointer tapped event is captured and triggers any tap handler
+func (li *listItem) Tapped(*fyne.PointEvent) {
+	if li.onTapped != nil && !li.Disabled() {
```
stuartmscott, line 295: In what scenario would a listItem be disabled? Is this something the dev would control?

widget/list.go
```diff
@@ -0,0 +1,414 @@
...
+		li.item.statusIndicator.FillColor = theme.FocusColor()
+		canvas.Refresh(li.item.super())
+		return
+	}
+	if li.item.hovered {
+		li.item.statusIndicator.FillColor = theme.HoverColor()
+	} else {
+		li.item.statusIndicator.FillColor = theme.BackgroundColor()
+	}
+	canvas.Refresh(li.item.super())
```
stuartmscott, line 338:
```suggestion
	if li.item.selected {
		li.item.statusIndicator.FillColor = theme.FocusColor()
	} else if li.item.hovered {
		li.item.statusIndicator.FillColor = theme.HoverColor()
	} else {
		li.item.statusIndicator.FillColor = theme.BackgroundColor()
	}
	canvas.Refresh(li.item.super())
```

cmd/fyne_demo/screens/widget.go
```diff
@@ -289,6 +289,37 @@ func stopProgress() {
...
+		},
+	)
+	list.OnItemSelected = func(index int, item fyne.CanvasObject) {
+		icon.Resource = item.(*fyne.Container).Objects[0].(*widget.Icon).Resource
+		label.Text = data[index]
+		icon.Refresh()
+		label.Refresh()
+	}
+	split := widget.NewHSplitContainer(list, fyne.NewContainerWithLayout(layout.NewCenterLayout(), hbox))
+	return fyne.NewContainerWithLayout(layout.NewBorderLayout(nil, nil, nil, nil), split)
```
stuartmscott, line 320: Not really leveraging border layout here, would a center or max layout work?

cmd/fyne_demo/screens/widget.go
```diff
@@ -289,6 +289,37 @@ func stopProgress() {
...
+		},
+		func(index int, item fyne.CanvasObject) {
+			item.(*fyne.Container).Objects[1].(*widget.Label).SetText(data[index])
+		},
+	)
+	list.OnItemSelected = func(index int, item fyne.CanvasObject) {
+		icon.Resource = item.(*fyne.Container).Objects[0].(*widget.Icon).Resource
+		label.Text = data[index]
+		icon.Refresh()
+		label.Refresh()
```
stuartmscott, line 317: Consider using the Setters which call Refresh for you

widget/list.go
okratitan: This should be resolved now.

widget/list_test.go
okratitan: This should be resolved now

widget/list_test.go
okratitan: These should be resolved now

widget/list.go
okratitan: Resolved

widget/list.go
okratitan: Resolved

widget/list.go
okratitan: Good question.  I put it there as good planning for the future.  As of right now - I don't think an item will be disabled.

cmd/fyne_demo/screens/widget.go
okratitan: This should be resolved now

cmd/fyne_demo/screens/widget.go
okratitan: This should be resolved now

widget/list.go

widget/list.go
```diff
@@ -0,0 +1,405 @@
...
+		// Scrolling Up
+		l.scrollUp(offsetChange)
+		return
+	}
+
+	// Relayout What Is Visible - no scroll change - initial layout or possibly from a resize
+	for _, child := range l.children {
+		l.itemPool.Release(child)
+	}
+	l.children = nil
```
andydotxyz, line 117: Sorry I am confused by this code block.
the comment above says no scroll change so I don't quite get why it is reset. 
However, even if the reset is needed then l.children is set to nil, and then you check if it is longer than the visible items - but that will never be true will it?

widget/list.go
```diff
@@ -0,0 +1,405 @@
...
+	l.layout.Objects = l.children
+	l.layout.Layout.Layout(l.layout.Objects, l.list.itemMin)
+	l.lastItemIndex = l.firstItemIndex + len(l.children) - 1
+
+	poolCapacity := 5
+	if l.list.Length()-len(l.children) < 5 {
+		poolCapacity = l.list.Length() - len(l.children)
+	}
+	for i := l.itemPool.Count(); i < poolCapacity; i++ {
+		// Make sure to keep extra items in the pool
```
andydotxyz, line 134: Each time I read an inline comment I think "should this be another method with a clear name".
Should this, with a few lines above, be in an "ensurePoolCapacity" method or something like that?

widget/list.go
andydotxyz: I'm not sure that having exported API here just for future purposes is good planning - we can be a simple BaseWidget for now, and in the future become a DisablableWidget I think?

widget/list.go
```diff
@@ -0,0 +1,405 @@
...
+func (li *listItem) Tapped(*fyne.PointEvent) {
+	if li.onTapped != nil && !li.Disabled() {
+		li.selected = true
+		li.Refresh()
+		li.onTapped()
+	}
+}
+
+type listItemRenderer struct {
+	*widget.ShadowingRenderer
```
andydotxyz, line 303: I know that we are moving away from this for now, but it's worth noting for future that this is drawing 8 shadow elements (sides and corners) despite us only wanting to show 1 edge. Performance increase can be expected when moving to a single shadow edge or the divide line discussed :)

widget/list.go
```diff
@@ -0,0 +1,405 @@
...
+
+	item   *listItem
+	layout fyne.Layout
+}
+
+// MinSize calculates the minimum size of a listItem.
+// This is based on the size of the status indicator and the size of the child object.
+func (li *listItemRenderer) MinSize() (size fyne.Size) {
+	li.item.statusIndicator.SetMinSize(fyne.NewSize(theme.Padding(), li.item.child.Size().Height))
```
andydotxyz, line 312: setting minsize in the call to MinSize() seems strange

widget/list.go
```diff
@@ -0,0 +1,405 @@
...
+}
+
+var _ fyne.Layout = (*listLayout)(nil)
+
+func newListLayout(list *List) fyne.Layout {
+	return &listLayout{list: list}
+}
+
+func (l *listLayout) Layout(objects []fyne.CanvasObject, size fyne.Size) {
+	if l.list.offsetY != 0 {
```
andydotxyz, line 346: I am curious about this - why return from layout if it's scrolled?

widget/list.go
```diff
@@ -0,0 +1,405 @@
...
+	}
+}
+
+func (l *listLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
+
+	return fyne.NewSize(l.list.itemMin.Width+theme.Padding(),
+		l.list.itemMin.Height*l.list.Length())
+}
+
+func (l *listLayout) itemAppended(objects []fyne.CanvasObject) {
```
andydotxyz, line 363: It's kindof subtle here but with this being called from appendItem wouldn't appendedItem be better? (and the prepended below)

andydotxyz (changes requested):
A few things internally, some comments, some questions. Seems you've reverted someone elses work unfortunately too.

Thoughts around testing - the pool code seems ripe for standalone testing but I don't think it's covered. It would be good to get some image tests in as well, show that it's rendering correctly.

widget/list.go
okratitan: This is fixed now.  No need to reset things.  Just ensure the proper count of items and update them.

widget/list.go
okratitan: Layout does a full layout/restacking of the items.  If it is a scroll we want to leave the items where they are just append/prepend items above or below rather than a full relayout.

widget/list.go
okratitan: Agreed. Resolved.

widget/list.go
okratitan: This has been Resolved.

widget/list.go
okratitan: This has been resolved.

widget/list.go
okratitan: Agreed - Resolved

cmd/fyne_demo/screens/widget.go
```diff
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"fyne.io/fyne"
-	"fyne.io/fyne/data/validation"
```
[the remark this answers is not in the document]
okratitan: This has now bee resolved

widget/list.go
okratitan: This has been resolved

widget/list.go
okratitan: I moved the length() == 0 quick exit to the top and added a cleanup in case the length of the list has changed from a positive count i.e. 100 to 0 to clean up the items.  As far as the rest... I'm not sure I follow?

widget/list.go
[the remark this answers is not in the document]
okratitan: This has been resolved and now uses base widget background color

widget/list.go
```diff
@@ -0,0 +1,415 @@
...
+	"fyne.io/fyne"
+	"fyne.io/fyne/canvas"
+	"fyne.io/fyne/driver/desktop"
+	"fyne.io/fyne/internal/widget"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/theme"
+)
+
+// List is a widget that caches list items for performance and
+// lays the items out in a vertical direction inside of a scroller
```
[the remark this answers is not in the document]
okratitan: fixed

widget/list.go
andydotxyz: So we have a special case when this widget is first shown?
Will this function correctly if the offset is non-zero before we first show?

okratitan: No. I dont think it would. Though that's not possible currently and would only become possible if we exposed list's internal scroller where offset could be changed before being shown.

andydotxyz: With this in mind do we need to add the ScrollToBottom/ScrollToTop that someone recently added to scroller so that we're not stuck with always-top?

andydotxyz:
I think some min / layout calculations are off...
Is the padding being applied on layout but not added to minsize? when I render min width but expand only height I get this:
list-light

widget/list.go
```diff
@@ -0,0 +1,449 @@
...
+	widget.BaseRenderer
+
+	item *listItem
+}
+
+// MinSize calculates the minimum size of a listItem.
+// This is based on the size of the status indicator and the size of the child object.
+func (li *listItemRenderer) MinSize() (size fyne.Size) {
+	size = fyne.NewSize((theme.Padding()*2)+li.item.child.Size().Width,
+		li.item.child.Size().Height+theme.Padding()*2)
```
stuartmscott, line 341: Is it possible that li.item.child.Size().IsZero(), perhaps this should be li.item.child.MinSize()?

widget/list.go
```diff
@@ -0,0 +1,449 @@
...
+// Layout the components of the listItem widget
+func (li *listItemRenderer) Layout(size fyne.Size) {
+	li.item.statusIndicator.Move(fyne.NewPos(0, 0))
+	s := fyne.NewSize(theme.Padding(), size.Height-1)
+	li.item.statusIndicator.SetMinSize(s)
+	li.item.statusIndicator.Resize(s)
+
+	x := theme.Padding() * 2
+	li.item.child.Move(fyne.NewPos(x, theme.Padding()))
+	li.item.child.Resize(fyne.NewSize(size.Width-x, size.Height-(2*theme.Padding())))
```
stuartmscott, line 354: How much padding does this leave between the trailing edge of the child, and the trailing edge of the list?

widget/list.go
stuartmscott: I don't think this was resolved

okratitan: It was because I committed the suggestion but I forgot to pull locally so it got overwritten again.  Should be good now.

widget/list.go
okratitan: Good question - How much of scroller functionality do we want to expose?  Changing offsets? Scrolling to bottom/top? Callbacks?

widget/list.go
okratitan: I'm not sure I follow... You are suggestion if the item min size is zero that we shouldn't add the padding in and just make it 0?  I don't think that would work as that would not leave room for the indicator or divider in the list item.

widget/list.go
okratitan: 2 * theme.Padding

okratitan: actually none.  This padding is between the child provided by the dev and the trailing edge of our list item.  There is no padding between the trailing edge of our list item and the list.

widget/list.go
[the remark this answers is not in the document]
okratitan: This is now resolved as the item parameter has been removed.

widget/list.go
stuartmscott: Not quite, I'm saying li.item.child.Size() can return 0x0

okratitan: Okay resolved

widget/list.go
```diff
@@ -348,8 +348,12 @@ type listItemRenderer struct {
 // MinSize calculates the minimum size of a listItem.
 // This is based on the size of the status indicator and the size of the child object.
 func (li *listItemRenderer) MinSize() (size fyne.Size) {
-	size = fyne.NewSize(li.item.child.Size().Width+theme.Padding()*2,
-		li.item.child.Size().Height+theme.Padding()*2)
+	itemSize := li.item.child.Size()
+	if itemSize.IsZero() {
+		itemSize = li.item.MinSize()
+	}
+	size = fyne.NewSize(itemSize.Width+theme.Padding()*3,
```
stuartmscott, line 355:
```suggestion
	size = fyne.NewSize(itemSize.Width+theme.Padding()*2,
```

okratitan: No, it is now 3.  The width of the indicator is one theme.Padding, then the theme.Padding() in between the indicator and the item, and then the theme.Padding() after the item.

stuartmscott: Ack

widget/list.go
```diff
@@ -0,0 +1,431 @@
...
+	item *listItem
+}
+
+// MinSize calculates the minimum size of a listItem.
+// This is based on the size of the status indicator and the size of the child object.
+func (li *listItemRenderer) MinSize() (size fyne.Size) {
+	itemSize := li.item.child.Size()
+	if itemSize.IsZero() {
+		itemSize = li.item.MinSize()
+	}
```
stuartmscott, line 354: I'm not sure this is the right fix. If this is called at the start and it gives a MinSize let's call `A`. Then the list is rendered in a large area, and each item is resized to be wider, then this function is called again it gives a MinSize let's call `B`. Then wont `B > A`? So this function is not returning "the minimum size of a listItem", it is returning "the size of a listItem".

Why not just have;
```go
itemSize := li.item.child.MinSize()
```

okratitan: Fair enough. Resolved.

widget/list.go
okratitan: From the chat on Slack it seems the agreement is to hold off on exposing scroller functionality for now

andydotxyz (approved):
Approving it, however the public documentation should really have punctuation (ending eith a full stop) - something for next time perhaps.

stuartmscott (approved):

PR #1262, opened 2020-08-26
Add Material Design "Card" as widget.CardContainer

andydotxyz:
### Description:
A basic card container with header, subheader, content and image - all optional.
We can add more to this later, such as the buttons (style challenges as they are depicted like hyperlink...?)

Deprecate Group in favour of CardContainer and use it in a few places where group was before.
Tests are just image comparison at this time as it's a pretty simple linear layout.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
Current look is as below:

Screenshot 2020-08-26 at 11 13 04

widget/card.go
```diff
@@ -0,0 +1,204 @@
+package widget
```
charlesdaniels, line 1: Considering calling this `cardcontainer.go` for consistency with the other container widgets?

widget/card.go
```diff
@@ -0,0 +1,204 @@
...
+// MinSize calculates the minimum size of a card.
+// This is based on the contained text, image and content.
+func (c *cardRenderer) MinSize() fyne.Size {
+	hasHeader := c.card.Title != ""
+	hasSubHeader := c.card.SubTitle != ""
+	hasImage := c.card.Image != nil
+	hasContent := c.card.Content != nil
+
+	if !hasHeader && !hasSubHeader && !hasContent {
+		return fyne.NewSize(c.card.Image.MinSize().Width+theme.Padding(), cardMediaHeight+theme.Padding())
```
charlesdaniels, line 33: I think this will cause a segfault if `c.card.Image` is nil, since you don't check that in the conditional.

widget/card.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	min := fyne.NewSize(theme.Padding()*5, theme.Padding()*5) // content padding plus 1 pad border
```
charlesdaniels, line 36: Maybe this is a convention used elsewhere in the code that I haven't experienced, but I see a lot of 5s and 3s that seem a little arbitrary? Maybe those should be non-exported `const` variables so they can be documented?

widget/card.go
```diff
@@ -0,0 +1,204 @@
...
+		header, subHeader, c}
+	r.applyTheme()
+	return r
+}
+
+// SetText updates the main title for this card.
+func (c *CardContainer) SetTitle(text string) {
+	c.Title = text
+
+	c.Refresh()
```
charlesdaniels, line 170: It's not necessary to refresh the whole container, and depending on the contents this might be expensive... maybe the labels and image could be children of `CardContainer` rather than it's renderer.

Edit: the same applies to the other `Set*` methods below.

charlesdaniels:
Two other things:

It looks like you have some changes in here that I thought were in other PRs? Maybe I haven't been keeping up, but I thought the list widgets was in #1235. Also it looked like there were some changes to vendor-ed deps. Maybe that's intentional, just thought it was worth pointing out. 

IMO, the borders are too thin/light. Their contrast is so low they are completely invisible on my secondary monitor, a Dell 1908FP. Not exactly a new model, but recent enough that our widgets should be usable on it.

widget/card.go
```diff
@@ -0,0 +1,204 @@
...
+	c.Layout(c.card.Size())
+	canvas.Refresh(c.card.super())
+}
+
+// CardContainer widget groups title, subtitle with content and a header image
+type CardContainer struct {
+	BaseWidget
+	Title, SubTitle string
+	Image           *canvas.Image
+	Content         fyne.CanvasObject
```
stuartmscott, line 136: I can haz two images?

From the spec:
> Media [optional]
> Cards can include a variety of media, including photos, and graphics, such as weather icons.

Also what about the rest of the spec: Thumbnail, Supporting Text, Buttons, Icons? This implementation feels very restrictive.

widget/card.go
stuartmscott, line 1: #827 Please

andydotxyz:
Closed in favour of #1263, will try to incorporate comments there, though questions here I will answer

widget/card.go
andydotxyz: Sorry but I do not understand the question - are you asking if two images can be added, or suggesting that it's strange that they can?

The supporting text is, in this implementation, the "Content" field and it can be anything you like.
Buttons were addressed in the PR opening: "We can add more to this later, such as the buttons (style challenges as they are depicted like hyperlink...?)" - I figured that being part of the button bar we would address icons at the same time.
Thumbnail is presented as an image clipped to a circle - as we cannot do this yet I opted to leave it out.

This is indeed not a "full card" implementation, as noted in the body of the PR.
It is intended as a starting point that is at least better than group, which as a poor API and needs to be deprecated.

widget/card.go
andydotxyz: Fixed - the refactor missed that for some reason - it was originally called Card.

widget/card.go
andydotxyz: Hmm, fair point - I did not think of the case when someone would want a card with no content

andydotxyz:
> This PR looks to include unrelated changes from #1253

Yes, sorry about that - the fixed PR is now #1263

andydotxyz:
> IMO, the borders are too thin/light.

It's just going with the "1dp" elevation like the spec. Perhaps we need to adjust shadow renderer to give more definition?

widget/card.go
andydotxyz: Hmm, I would often agree but in this case it is somewhat arbitrary... This was tweaking layouts to align things nicely for the component. 
I could write it out 5 as borderPadding + horizontalContentPadding*2 having defined a few new variables - is it worth it?

widget/card.go
andydotxyz: Well, It's not that simple - if it were set to "" (or from "") then we do need a complete refresh.
Should that logic be in her to compare old and new values to do the different type of refreshes?

widget/card.go
charlesdaniels: I think it's worth breaking this kind of stuff out at least into consts with comments, since it makes it easier for a future reader to see what you meant / where you got the numbers from. Even if it's just an arbitrary magic number, explaining how you arrived at that value might be useful to future maintainers.

widget/card.go
charlesdaniels: Considering that the content could potentially be very expensive to refresh, I think it makes sense to avoid propagating it down the line needlessly. In particular, I imagine a common use case for cards would be as a container for a custom widget -- downstream developers may not be as careful about avoiding spurious refreshes, so I think for the best developer experience we should do as much of the legwork as possible in that area.

charlesdaniels:
> It's just going with the "1dp" elevation like the spec. Perhaps we need to adjust shadow renderer to give more definition?

Maybe this is a separate issue. But on one of my monitors the borders are completely invisible. It's an older monitor with crappy contrast, but it's not old enough I would be willing to say we shouldn't support it.

andydotxyz:
Let's take a look separately. A change in shadow calculation will change every widget image test with a shadow - if this PR inlcuded that I suspect that there would be a number of questions?

widget/card.go
andydotxyz: should be fixed now

PR #1263, opened 2020-08-26
Feature/card (retry)

andydotxyz:
### Description:
A basic card container with header, subheader, content and image - all optional.
We can add more to this later, such as the buttons (style challenges as they are depicted like hyperlink...?)

Deprecate Group in favour of CardContainer and use it in a few places where group was before.
Tests are just image comparison at this time as it's a pretty simple linear layout.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


cmd/fyne_demo/screens/container.go
```diff
@@ -81,6 +82,17 @@ func makeButtonList(count int) []fyne.CanvasObject {
...
 
+func makeCardTab() fyne.CanvasObject {
+	card1 := widget.NewCardContainer("Book a table", "Which time suits?",
+		widget.NewRadio([]string{"6:30pm", "7:00pm", "7:45pm"}, func(string) {}))
+	card2 := widget.NewCardContainer("With media", "No content, with image", nil)
+	card2.Image = canvas.NewImageFromResource(theme.FyneLogo())
+	card3 := widget.NewCardContainer("Title 3", "Subtitle", widget.NewCheck("Check me", func(bool) {}))
+	card4 := widget.NewCardContainer("Title 4", "Another card", widget.NewLabel("Content"))
+	return fyne.NewContainerWithLayout(layout.NewGridLayout(3), widget.NewVBox(card1, card4),
+		widget.NewVBox(card2), widget.NewVBox(card3))
```
stuartmscott, line 93: Why use VBox with a single child?

internal/painter/image.go
```diff
@@ -22,7 +22,7 @@ import (
 // PaintImage renders a given fyne Image to a Go standard image
 func PaintImage(img *canvas.Image, c fyne.Canvas, width, height int) image.Image {
 	if width <= 0 || height <= 0 {
-		return nil
+		return image.NewNRGBA(image.Rect(0, 0, 0, 0))
```
stuartmscott, line 25: Is this required for cardcontainer? Seem unnecessary to allocate this. Can't the caller do a nil check?

widget/cardcontainer.go
```diff
@@ -0,0 +1,211 @@
...
+	header.TextStyle.Bold = true
+	subHeader := canvas.NewText(c.SubTitle, theme.TextColor())
+
+	objects := []fyne.CanvasObject{header, subHeader}
+	if c.Image != nil {
+		objects = append(objects, c.Image)
+	}
+	if c.Content != nil {
+		objects = append(objects, c.Content)
+	}
```
stuartmscott, line 46: If Image or Content are set after a delay (eg goroutine downloads image from network, goroutine loads content from file), is it possible the widget's renderer is created before the new Image and Content are set and so they will not be in the renderer's objects, thus not rendered?

widget/cardcontainer_test.go
```diff
@@ -0,0 +1,64 @@
...
+func TestCard_Layout(t *testing.T) {
+	test.NewApp()
+	test.ApplyTheme(t, theme.LightTheme())
+
+	for name, tt := range map[string]struct {
+		title, subtitle string
+		icon            *canvas.Image
+		content         fyne.CanvasObject
+	}{
+		"titles": {
```
stuartmscott, line 22: Consider adding:
- Title only
- Subtitle only
- Content only

cmd/fyne_demo/screens/container.go
andydotxyz: So that the border that card draws will collapse to it's content instead of stretching to the whole column

widget/cardcontainer_test.go
andydotxyz: Added. Thanks, exposed minor sizing issue

widget/cardcontainer.go
andydotxyz: Updated and 2 new tests demonstrate it

internal/painter/image.go
andydotxyz: removed

widget/cardcontainer.go
```diff
@@ -0,0 +1,220 @@
...
+	"fyne.io/fyne"
+	"fyne.io/fyne/canvas"
+	"fyne.io/fyne/internal/widget"
+	"fyne.io/fyne/theme"
+)
+
+// CardContainer widget groups title, subtitle with content and a header image
+type CardContainer struct {
+	BaseWidget
+	Title, SubTitle string
```
stuartmscott, line 15: nit: subtitle is one word, not a concatenation
```suggestion
	Title, Subtitle string
```

widget/cardcontainer.go
```diff
@@ -0,0 +1,220 @@
...
+)
+
+func (c *cardRenderer) BackgroundColor() color.Color {
+	return theme.BackgroundColor()
+}
+
+// Layout the components of the card container.
+func (c *cardRenderer) Layout(size fyne.Size) {
+	pos := fyne.NewPos(theme.Padding()/2, theme.Padding()/2)
+	size = size.Subtract(fyne.NewSize(theme.Padding(), theme.Padding()))
```
stuartmscott, line 106: Only half padding? Why not have full theme.Padding() all around?

widget/cardcontainer.go
```diff
@@ -0,0 +1,220 @@
...
+	card *CardContainer
+}
+
+const (
+	cardMediaHeight = 128
+)
+
+func (c *cardRenderer) BackgroundColor() color.Color {
+	return theme.BackgroundColor()
+}
```
stuartmscott, line 101: Unnecessary - already in BaseRenderer

widget/cardcontainer.go
```diff
@@ -0,0 +1,220 @@
...
+type cardRenderer struct {
+	*widget.ShadowingRenderer
+
+	header, subHeader *canvas.Text
+
+	card *CardContainer
+}
+
+const (
+	cardMediaHeight = 128
```
stuartmscott, line 96: Longer term should this be in theme? Like IconInlineSize()

widget/cardcontainer_test.go
stuartmscott: Maybe few more:
- titles w/ image (title, subtitle, image)
- titles w/ content (title, subtitle, content)
- all four (title, subtitle, image, content)

stuartmscott (changes requested):
Good progress!

widget/cardcontainer.go
andydotxyz: Perhaps yes, but as with many other items we cannot do that now

widget/cardcontainer.go
andydotxyz: Consistent with the button shadow. It's inset into the content by half padding, then we pad internally from that

andydotxyz:
Re:
> titles w/ image (title, subtitle, image)
> titles w/ content (title, subtitle, content)
> all four (title, subtitle, image, content)

To what end? If you think every permutation of options should be tested please say so - I added the ones you thought were missing, and this list is additional to that. Do you also need subtitle with content, subtitle with image, what about with just title missing as well?

widget/cardcontainer.go
andydotxyz: Thanks

stuartmscott:
Sorry, my initial list should have been more comprehensive to avoid a back and forth. I don't think we need every permutation, just the main ones a dev is expected to use.

andydotxyz:
OK, I think If I add titles+image and all4 we should have pretty good coverage

andydotxyz:
That should be everything now

stuartmscott (approved):

PR #1267, opened 2020-08-27
Move validators to data/validation

Jacalz:
### Description:
Per discussion with @andydotxyz, we decided that it makes a lot more sense to not have the validators inside the top level fyne package. This will make it easier to expand with more validators in the future.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
How about "validation" so it's slightly broader than just implementations of Validator.
Like how `canvas` package is not called `canvasobjects`.

One question though - should we consider this grouped with other data things, like data/binding and data/validation
and/or would widget/validation be useful?
Feels like perhaps we already have enough top level packages..?

Jacalz:
Yeah I like "validation" too. That fits the bill quite nicely. Like we discussed before, having it inside or below the widget package means that it can be interpreted as an actual widget. I think that `data/validation` perhaps would make the most sense.

Jacalz:
I have now moved it into `data/validation`. Makes a lot more sense imho :)

andydotxyz (approved):

PR #1273, opened 2020-08-29
Test moving travis to Go 1.15 and speeding up travis builds

Jacalz:
### Description:
This PR tries to move Travis to using the latest Go release and also makes sure to apply some changes to make builds faster.
It now uses caching for modules and apt packages while now only running code checks where we tell it to (basically means that we don't run it on each operating system virtual machine).

### Checklist:

- [-] Tests included.
- [-] Lint and formatter run with no errors.
- [ ] Tests all pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,27 +65,27 @@ env:
 
 matrix:
   include:
-    - go: 1.12.x
+    - go: 1.13.x
       os: linux
       env: GO111MODULE=on GOFLAGS=
-    - go: 1.13.x
+    - go: 1.14.x
       os: linux
       env: GO111MODULE=off GOFLAGS=
-    - go: 1.13.x
+    - go: 1.14.x
       os: osx
       env: GO111MODULE=off GOFLAGS=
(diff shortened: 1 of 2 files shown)


Jacalz:
This has improved the run time for travis builds by about 2 minutes. See before and after:
image
image

andydotxyz (changes requested):
Please do t remove 1.12 - we need to continue checking compatibility

Jacalz:
Oh, I see. I honestly thought that people would have dropped of that by now, but I see that Ubuntu 19.10 still uses that. Should we just drop testing for 1.13 and switch that to 1.12?

andydotxyz:
Yeah, going back to the first item being 1.12 works.
At some point we can drop older versions - but that is a separate decision and not something to do accidentally by stopping checking in CI ;).

Jacalz:
Yeah absolutely. Go 1.12.x is now back in again :)

andydotxyz (approved):

PR #1277, opened 2020-09-02
Propose more semantic naming for Button.HideShadow

andydotxyz:
Proposing a more semantic bsaed naming of the ability to remove shadows... The API before doesn't quite fit the widget naming.
MD calls it LowEmphasis, I have added Flat as a synonym for more discoverability...

Thoughts?

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


stuartmscott (dismissed):

andydotxyz:
I'm having second thoughts about two different names. Flat is more discoverable but is it "our way"? I don't know... @stuartmscott @toaster

andydotxyz:
Following discussions we acreed that the duplicate naming here isn;'t great and that something between the two might be right.

Working idea: "Importance" enum with "Primary", "Default" and "Plain" - or "Major", "Default", "Minor" perhaps

andydotxyz:
or "High", "Medium" and "Low" importance?

andydotxyz:
I think I a, currently falling on "Primary", "Medium" and "Low" - they relate to the material usage on https://material.io/components/buttons#text-button whilst being fairly discoverable and clear.

toaster:
I can’t see where the linked MD document uses “primary” and “medium” or “low” for the same thing.
What I can see is:
* They have “contained”, “outlined” and “text” buttons as three distinguished _types_ of buttons which represent different importance.
* Importance is consistently named “high”, “medium” and “low” emphasis AFAICS.
* “primary” is used to describe a button which triggers the primary action. AFAIU it is not forbidden to have multiple buttons of high importance but of course only one of these can be the primary button.

That’s why I vote for “high”, “medium” and “low” importance.

andydotxyz:
Yes That is a great summary thanks - I think I agree too :)

toaster:
This should then also have an effect on the `Button`’s appearance when disabled. Currently, a button without shadow (aka low importance) gets a different background colour when being disabled which makes it look more like a button with medium importance.
This kind of blocks the bugfix for disabled `SelectEntry` :).

toaster:
Plus, I vote for deprecating the old `DefaultButton` and `PrimaryButton` and also remove the `FlatButton`.

andydotxyz:
> Plus, I vote for deprecating the old `DefaultButton` and `PrimaryButton` and also remove the `FlatButton`.

Agreed, yes.
I need to get this wrapped up and include all these suggestions.

Hopefully today.

widget/button.go
```diff
@@ -182,11 +313,51 @@ func (b *buttonRenderer) Refresh() {
 	canvas.Refresh(b.button.super())
 }
 
+func alignedPosition(align ButtonAlign, padding, objectSize, layoutSize fyne.Size) (pos fyne.Position) {
```
stuartmscott, line 316: nit: should be below the renderer methods

andydotxyz: thanks, not surprised I missed one in the reshuffle

stuartmscott (approved):

widget/button.go
```diff
@@ -341,7 +357,7 @@ func (b *buttonRenderer) buttonColor() color.Color {
 	switch {
 	case b.button.Disabled():
 		return theme.DisabledButtonColor()
```
toaster, line 348: A button with low importance has the same background as normal content and no surrounding shadow.
When deactivating it, it suddenly becomes more visible because it has now a background different from the normal content.
However, I think this should be fixed by another PR.

toaster (approved):

widget/button.go
```diff
@@ -13,7 +13,190 @@ import (
...
+// Button widget has a text label and triggers an event func when clicked
+type Button struct {
+	DisableableWidget
+	Text string
+	// Deprecated, use Importance instead, where HighImportance matches PrimaryButton
+	Style ButtonStyle
+	Icon  fyne.Resource
+	// Specify how prominent the button should be, High will highlight the button and Low will remove some decoration.
+	Importance    ButtonImportance
+	disabledIcon  fyne.Resource
```
stuartmscott, line 73: nit: move down to other unexported fields `hovered, tapped bool`

andydotxyz: As discussed before the field order has not been specified in the order chat.
Before asking for it to be moved we should agree on the order.

Issue #1278, opened 2020-09-03
File Dialog: toggle hidden files


kevingentile:
### Is your feature request related to a problem? Please describe:

The current file dialog does not give the option to show hidden files.

### Is it possible to construct a solution with the existing API?

Yes, hidden files are [skipped explicitly](https://github.com/fyne-io/fyne/blob/86d26ebe4d97a525aa5cf1b6720186fc76d3b669/dialog/file.go#L196).

### Describe the solution you'd like to see:

A way to toggle hidden files. Ideally a check box with a configurable default value.

Jacalz:
Ideally, we should also support toggling this with a shortcut such as `ctrl + h` on Linux.

kevingentile:
The draft above implements these changes, but the layout of the checkbox still doesn't feel right
<img width="677" alt="Screen Shot 2020-09-03 at 9 54 20 PM" src="https://user-images.githubusercontent.com/10604101/92193437-5599aa00-ee36-11ea-989d-1f4cd95db57a.png">

andydotxyz:
The development of this was continued and landed for 2.0 release

PR #1282, opened 2020-09-04
implement show/hide hidden files

kevingentile:
### Description:

Addresses #1278

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -139,16 +139,16 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 		}
 	})
 	buttons := widget.NewHBox(f.dismiss, f.open)
+	f.showHiddenCheck = widget.NewCheck("Show Hidden Files", func(_ bool) {
+		f.refreshDir(f.dir)
+	})
 	footer := fyne.NewContainerWithLayout(layout.NewBorderLayout(nil, nil, nil, buttons),
-		buttons, widget.NewHScrollContainer(f.fileName))
+		buttons, widget.NewHScrollContainer(f.fileName), f.showHiddenCheck)
 
 	f.files = fyne.NewContainerWithLayout(layout.NewGridWrapLayout(fyne.NewSize(fileIconCellWidth,
 		fileIconSize+theme.Padding()+fileTextSize)),
 	)
 	f.fileScroll = widget.NewScrollContainer(f.files)
-	f.showHiddenCheck = widget.NewCheck("Show Hidden Files", func(_ bool) {
-		f.refreshDir(f.dir)
-	})
 	verticalExtra := int(float64(fileIconSize) * 0.25)
 	f.fileScroll.SetMinSize(fyne.NewSize(fileIconCellWidth*2+theme.Padding(),
 		(fileIconSize+fileTextSize)+theme.Padding()*2+verticalExtra))
@@ -157,8 +157,7 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 	scrollBread := widget.NewHScrollContainer(f.breadcrumb)
 	body := fyne.NewContainerWithLayout(layout.NewBorderLayout(scrollBread, nil, nil, nil),
 		scrollBread, f.fileScroll)
-	headerLabel := widget.NewLabelWithStyle(label+" File", fyne.TextAlignLeading, fyne.TextStyle{Bold: true})
-	header := widget.NewHBox(headerLabel, f.showHiddenCheck)
+	header := widget.NewLabelWithStyle(label+" File", fyne.TextAlignLeading, fyne.TextStyle{Bold: true})
 	favorites := widget.NewGroup("Favorites", f.loadFavorites()...)
 	return fyne.NewContainerWithLayout(layout.NewBorderLayout(header, footer, favorites, nil),
 		favorites, header, footer, body)
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -8,6 +8,7 @@ import (
 
 	"fyne.io/fyne"
 	"fyne.io/fyne/canvas"
+	"fyne.io/fyne/driver/desktop"
 	"fyne.io/fyne/layout"
 	"fyne.io/fyne/storage"
 	"fyne.io/fyne/theme"
@@ -20,13 +21,14 @@ type textWidget interface {
 }
 
 type fileDialog struct {
-	file       *FileDialog
-	fileName   textWidget
-	dismiss    *widget.Button
-	open       *widget.Button
-	breadcrumb *widget.Box
-	files      *fyne.Container
-	fileScroll *widget.ScrollContainer
+	file            *FileDialog
+	fileName        textWidget
+	dismiss         *widget.Button
+	open            *widget.Button
+	breadcrumb      *widget.Box
+	files           *fyne.Container
+	fileScroll      *widget.ScrollContainer
+	showHiddenCheck *widget.Check
 
 	win      *widget.PopUp
 	selected *fileDialogItem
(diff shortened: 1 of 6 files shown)


Jacalz (commented):
Hi @kevingentile. Thanks for the PR. Pretty much all Pull Requests should target the `develop` branch. Would you mind changing the target for this? Please see [this guide](https://stackoverflow.com/questions/24159036/how-to-modify-a-pull-request-on-github-to-change-target-branch-to-merge-into) if you are unsure how to do that.

Fyi, tests seem to be failing with this error:
```go
--- FAIL: TestShowFileOpen (0.01s)
panic: interface conversion: fyne.CanvasObject is *widget.Box, not *widget.Label [recovered]
	panic: interface conversion: fyne.CanvasObject is *widget.Box, not *widget.Label
```

dialog/file.go
```diff
@@ -193,7 +199,7 @@ func (f *fileDialog) refreshDir(dir string) {
 		icons = append(icons, fi)
 	}
 	for _, file := range files {
-		if isHidden(file.Name(), dir) {
+		if !f.showHiddenCheck.Checked && isHidden(file.Name(), dir) {
```
andydotxyz, line 202: Reaching out of this context to read the value of a UI element seems like poor separation, this function takes a parameter for what directory to show, so it should probably have hidden logic similarly. 

Normally a method called with the same parameters should produce the same outcome

dialog/file.go
```diff
@@ -151,7 +156,8 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 	scrollBread := widget.NewScrollContainer(f.breadcrumb)
 	body := fyne.NewContainerWithLayout(layout.NewBorderLayout(scrollBread, nil, nil, nil),
 		scrollBread, f.fileScroll)
-	header := widget.NewLabelWithStyle(label+" File", fyne.TextAlignLeading, fyne.TextStyle{Bold: true})
+	headerLabel := widget.NewLabelWithStyle(label+" File", fyne.TextAlignLeading, fyne.TextStyle{Bold: true})
```
andydotxyz, line 159: I don't think that adding a non-essential configuration belongs in the header

dialog/file.go
```diff
@@ -292,6 +298,12 @@ func showFile(file *FileDialog) *fileDialog {
 		(fileIconSize+fileTextSize)+theme.Padding()*4))
 
 	d.win = widget.NewModalPopUp(ui, file.parent.Canvas())
+	showHiddenShortcut := &desktop.CustomShortcut{KeyName: fyne.KeyH, Modifier: desktop.ControlModifier}
+	d.win.Canvas.AddShortcut(showHiddenShortcut, func(shortcut fyne.Shortcut) {
```
andydotxyz, line 302: If you are adding a keyboard shortcut into the canvas when this shows it needs to be removed when it is hidden.
Otherwise we are either leaking the UI to satisfy the callback, or it will crash as the UI is destroyed.

andydotxyz (changes requested):
Some thoughts inline.
Also please add tests - you should work through the checklist in the PR template

andydotxyz:
Thanks so much for your contribution - I think the comments from @Jacalz and myself should help steer this towards completion.

dialog/file.go
kevingentile: I agree. Suggestions for an alternative location?

dialog/file.go
kevingentile: I was unable to find a method to remove shortcuts so I added one to the canvas interface. Not sure if this change would be a desired approach. For this ticket it feels like scope creep.

dialog/file.go
kevingentile: moved to footer

andydotxyz:
We talked about this on our community call today and agree it's challenging to find the right place. At some point we're likely to add a toolbar with various items but for now that's overkill.

So we thought - what about an "options" button that pops up a menu with the "Show Hidden Files" checkbox in it?
If similar things come up we can add them there, and if the toolbar arrives then they can easily be moved into that.

dialog/file.go
andydotxyz: Good point. I'd say we can't really use canvas.AddShortcut for temporal shortcuts, so maye move that to another PR along with the code that adds the ability to remove them?

dialog/file.go
andydotxyz: Isn't that where the current file name is shown? if so it takes quite a bit of space away from that.

dialog/file.go
kevingentile: Would you mind evaluating a bit on how you envision temporal shortcuts to work? I'm picking up the code base as I go, so I doubt I would have a well versed initial approach to its implementation.

Shortcuts like copy/paste feel like they would always be available through the system so these feel like maybe "persistent" shortcuts.

Temporary shortcuts like the hide/show hidden files shortcut would only available for a single dialog. Would the canvas not be the ideal place to manage these?

dialog/file.go
kevingentile: Doesn't appear to be, unless I'm missing something

Screen Shot 2020-09-04 at 10 37 36 PM

andydotxyz: I think you've removed it, should look like this: 
Screenshot 2020-09-07 at 22 58 13

dialog/file.go
andydotxyz: Shortcuts for inputs (like Entry) are registered on the entry, so come in to effect when that is focused.
In this case we have not got a focused input so the canvas is the right place - but we are only shown for a short while so the same shortcut should be removed once we hide. The dialog does not have a separate canvas you see.
I'll check out your PR now, I hope this is what you assumed.

dialog/file.go
kevingentile: Yep. I didn't provide the check in the footer layout. Move to top:

Screen Shot 2020-09-08 at 7 39 06 PM

andydotxyz:
It looks like that label is taking a whole row of space off the bottom of the dialog which seems a shame.
Did you not like our idea of using the bottom of the sidebar?

dialog/file.go
```diff
@@ -20,13 +20,14 @@ type textWidget interface {
...
-	files      *fyne.Container
-	fileScroll *widget.ScrollContainer
+	file            *FileDialog
+	fileName        textWidget
+	dismiss         *widget.Button
+	open            *widget.Button
+	breadcrumb      *widget.Box
+	files           *fyne.Container
+	fileScroll      *widget.ScrollContainer
+	showHiddenCheck *widget.Check
```
andydotxyz, line 30: We generally avoid putting types in variable names, Go likes to avoid duplication

dialog/file.go
```diff
@@ -136,8 +137,11 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 		}
 	})
 	buttons := widget.NewHBox(f.dismiss, f.open)
-	footer := fyne.NewContainerWithLayout(layout.NewBorderLayout(nil, nil, nil, buttons),
-		buttons, widget.NewHScrollContainer(f.fileName))
+	f.showHiddenCheck = widget.NewCheck("Show Hidden Files", func(changed bool) {
```
andydotxyz, line 140: I think this is what I saw in my comment in the conversation - this takes a lot of space away from the file panel above... Could it fit in the sidebar instead? (perhaps with the "Options" popout we suggested)

dialog/file_test.go
```diff
@@ -129,6 +129,8 @@ func TestShowFileOpen(t *testing.T) {
 	nameLabel := ui.Objects[2].(*fyne.Container).Objects[1].(*widget.ScrollContainer).Content.(*widget.Label)
 	buttons := ui.Objects[2].(*fyne.Container).Objects[0].(*widget.Box)
 	open := buttons.Children[1].(*widget.Button)
+	hidden := ui.Objects[2].(*fyne.Container).Objects[2].(*widget.Check)
+	assert.Equal(t, hidden.Text, "Show Hidden Files")
```
andydotxyz, line 133: This isn't really testing the functionality, just the presentation.
It would be preferrable to have a hidden file in the testdata directory and show that it isn't visible by default, then tapping the checkbox makes it appear.

andydotxyz (changes requested):
Thanks for this. The functionality is solid, but the test and UI placement could probably be improved upon I think.

andydotxyz:
Thanks for the removeshortcut addition - that is landed now so merging develop in here should make it available for this change

andydotxyz:
Hi @kevingentile Are you able to continue with this PR now that it's unblocked?
We will be doing the last 1.4 planning session this week as we approach release and it would be great to get this in if possible.

andydotxyz:
Moving out of the 1.4 to see if we can finish this off for a point release

andydotxyz:
Replaced by #1689

PR #1286, opened 2020-09-04
Ensure provider is up to date before MinSize

stuartmscott:
### Description:
This ensures the textProvider contains the same text as the label before calling MinSize.

Fixes the issue where Text field is updated and MinSize is called which delegates to textProvider which returns an incorrect size.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


stuartmscott:
Grrr
```
--- FAIL: TestFileWatcher_FileDeleted (0.21s)
    settings_desktop_test.go:92: File watcher callback was not called
FAIL
```

andydotxyz:
> Grrr

I don't know why this is still happening. I especially don't know why it's only some times.
At least it's less often :)

Kicked travis so it'll be back in a sec

andydotxyz (approved):

PR #1292, opened 2020-09-07
Feature/primary color user setting

### LINKED ISSUE #1284: Add option to pick primary color
In fyne_settings - a short list of colors that will match the UI aesthetic.

andydotxyz:
### Description:
Provide user a choice of primary colours that can replace the default blue.
Add this to the fyne_settings app. Move to dynamically generate a preview along with light/dark.
Add settings access from fyne_demo and generally improve the menu features.

Fixes #1284 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


cmd/fyne_settings/settings/appearance.go
```diff
@@ -128,3 +207,74 @@ func (s *Settings) saveToFile(path string) error {
 
 	return ioutil.WriteFile(path, data, 0644)
 }
+
+type colorButton struct {
```
stuartmscott, line 211: The color picker I'm working on also defines a colorbutton, albeit slightly different, perhaps we should converge on a design as this would be a useful exported widget?

cmd/fyne_settings/settings/appearance.go
```diff
@@ -78,15 +109,45 @@ func (s *Settings) chooseTheme(name string) {
...
+		widget.NewTabItemWithIcon("Welcome", theme.HomeIcon(), empty),
+		widget.NewTabItemWithIcon("Graphics", theme.DocumentCreateIcon(), empty),
+		widget.NewTabItemWithIcon("Widgets", theme.CheckButtonCheckedIcon(), makeProgressTab()),
+		widget.NewTabItemWithIcon("Containers", theme.ViewRestoreIcon(), empty),
+		widget.NewTabItemWithIcon("Windows", theme.ViewFullScreenIcon(), empty))
+	tabs.SetTabLocation(widget.TabLocationLeading)
+	tabs.SelectTabIndex(2)
+
+	c.SetContent(tabs)
+	c.Resize(fyne.NewSize(380, 380))
```
stuartmscott, line 143: I understand that this, and makeProgressTab below, are needed to showcase the chosen color, but I'm not a huge fan of the duplication, is there another way?

stuartmscott (changes requested):

cmd/fyne_settings/settings/appearance.go
andydotxyz: I just copied and pasted this to show the feature - I am working on a replacement, it's this that is holding it in Draft state

app/settings.go
```diff
@@ -38,6 +40,21 @@ type settings struct {
...
 
+func (s *settings) PrimaryColor() string {
+	s.propertyLock.RLock()
+	defer s.propertyLock.RUnlock()
+	return s.schema.PrimaryColor
+}
+
+// OverrideTheme allows the settings app to temporarily preview different theme details.
+// Please make sure that you remember the original settings and call this again to revert the change.
+func (s *settings) OverrideTheme(theme fyne.Theme, name string) {
```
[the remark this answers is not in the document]
andydotxyz: I wondered about that but it overrides the theme as well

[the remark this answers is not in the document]
stuartmscott: Would it be possible to override the theme once, and then switch out the chosen color inside that theme? In that case your original naming would work and `name string` param can be dropped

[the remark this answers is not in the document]
andydotxyz: If you have a custom theme that supports mutating the primary colour this can be done.
However we do not currently have an API that supports extending a theme to switch the colour, so it seemed like doing it here was the only choice.
Open to better ideas

cmd/fyne_demo/main.go
```diff
@@ -104,7 +109,19 @@ func main() {
...
+	}),
+		fyne.NewMenuItem("Support", func() {
+			u, _ := url.Parse("https://fyne.io/support/")
+			_ = a.OpenURL(u)
+		}),
+		fyne.NewMenuItemSeparator(),
+		fyne.NewMenuItem("Sponsor", func() {
+			u, _ := url.Parse("https://github.com/sponsors/fyne-io")
+			_ = a.OpenURL(u)
+		}))
```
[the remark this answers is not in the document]
andydotxyz: I have no idea why but this is what goimports wants - I tried less indentation on the support/sponsor or more on help and it keeps putting it back this way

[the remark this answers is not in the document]
stuartmscott: Try it with a newline after `"Help",`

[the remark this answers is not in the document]
andydotxyz: Ah yes I see, thanks

andydotxyz:
I added some API to theme so we can keep colour list in sync on colour picker like you suggested @stuartmscott

cmd/fyne_settings/settings/appearance.go
```diff
@@ -10,20 +12,36 @@ import (
...
 )
 
 const (
 	systemThemeName = "system default"
 )
 
+var (
+	colorNames = []string{theme.ColorBlue, theme.ColorGreen, theme.ColorYellow,
+		theme.ColorOrange, theme.ColorRed, theme.ColorGray}
+	colorValues = map[string]color.Color{
```
stuartmscott, line 29: Add Indigo, Violet/Purple for a full rainbow.
Brown would be a nice addition too.

stuartmscott (changes requested):
> I added some API to theme so we can keep colour list in sync on colour picker like you suggested @stuartmscott

Cool thanks, I can use this for the "basic" colours in Colour Picker to replace this list:
```go
"#f44336", //Red
"#ff9800", //Orange
"#ffeb3b", //Yellow
"#4caf50", //Green
"#2196f3", //Blue
"#3f51b5", //Indigo
"#9c27b0", //Purple
"#795548", //Brown
```
Colour Picker has another list for the greyscale colours, should this also be a public API?
```go
"#ffffff",
"#dddddd",
"#bbbbbb",
"#999999",
"#666666",
"#444444",
"#222222",
"#000000",
```

cmd/fyne_settings/settings/appearance.go
stuartmscott: This is no longer in draft, but I don't see a replacement

andydotxyz: Erm, it's there: https://github.com/fyne-io/fyne/pull/1292/files#diff-fc4b8de1d5fbe30aec3d3d9d22e86237R119

cmd/fyne_settings/settings/appearance.go
andydotxyz: let's not add to this by covering greyscale as well - this is only adding colours that can work as primary and you need to get the shade just right to do that so most greys won't work

andydotxyz:
OK, updating the colors. I left off Indigo as it seemed like light/dark green etc, 8 colours seems like a good number :)
Screenshot 2020-10-05 at 21 45 03

cmd/fyne_settings/settings/appearance.go
andydotxyz: Added briwn and purple too, indigo seemed like it didn't match the overall aim to match [saturation]?

stuartmscott: This is fine for here, but there is a problem using this in colour picker as it has a row of primary colours and a row of greyscale - it will be odd to have grey in the primary colours above the row of greyscale

cmd/fyne_settings/settings/appearance.go
stuartmscott: I don't follow.

> I'm not a huge fan of the duplication

> I am working on a replacement

_marked ready for review_

> I don't see a replacement

_link to the same duplicated code_

cmd/fyne_settings/settings/appearance.go
andydotxyz: You can remove Color.Grey from the list of colours you use surely? It's not an all-or-nothing API

cmd/fyne_settings/settings/appearance.go
andydotxyz: Can you explain exactly what the duplication is? I think I am missing the point.
I coded a whole new tab container with different tabs, content and an overlay...
As per screenshot at the end of this conversation

cmd/fyne_settings/settings/appearance.go
```diff
@@ -128,3 +178,99 @@ func (s *Settings) saveToFile(path string) error {
...
+}
+
+func (c *colorRenderer) Objects() []fyne.CanvasObject {
+	return c.objs
+}
+
+func (c *colorRenderer) Destroy() {
+}
+
+func makeHomeTab() fyne.CanvasObject {
```
stuartmscott, line 253: Does this really need to be a separate function?

stuartmscott (changes requested):

cmd/fyne_settings/settings/appearance.go
stuartmscott: The duplication is this tabber, why not use fyne_demo tabber which already showcases various UI elements?

cmd/fyne_settings/settings/appearance.go
stuartmscott: ok

cmd/fyne_settings/settings/appearance.go
andydotxyz: because they are different apps and serve different purposes?
Fyne_demo is about to be completely rewritten using Tree anyhow so that concern is probably moot.

cmd/fyne_settings/settings/appearance.go
andydotxyz: No, it does not have to be.
For a while it contained more content - and in the future it might be more interesting than the simple label.
I can take it out if this is a concern

cmd/fyne_settings/settings/appearance.go
stuartmscott: Well the goal is to show how the primary colour looks across the UI, and so you'd probably want more than just a current tab and button (the only usage of primary colour in the screenshot below) which is why I was suggesting hooking into demo, but this is probably fine for a first attempt

andydotxyz: Yeah I played with progress bars but it looked awkward. Open to ideas of realistic content that could be dropped in though (under the overlay?)

app/settings.go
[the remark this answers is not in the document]
stuartmscott: I can't think of any better approaches right now, but I'm not sure this is a good addition to the public API. Though I guess `settings` is unexported, can it be `overrideTheme`?

cmd/fyne_settings/settings/appearance.go
stuartmscott: Let's remove for now, we can revisit in future

cmd/fyne_settings/settings/appearance.go
stuartmscott: I had a look through fyne_demo to see what components we could put here, but there aren't many that show primary color when not in focus (which is good in general, but bad for this particular situation). I think what you have here so far is solid, we can always revisit in future.
What if the username entry in the login dialog could put into focus, so there is at least 3 uses of primary color?

stuartmscott:
Apologies for the bad communication earlier. This PR is really close, and could be merged once the two pending items are addressed (overrideTheme and makeHomeTab).

I think this will greatly improve the toolkit by allowing end-users to personalize applications without any effort on the developer's side. Nice work Andy!

Now I need to figure out what colour I'll pick, it'll probably either be Blue or Green...

app/settings.go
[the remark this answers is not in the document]
andydotxyz: If it was unexported it would not be possible to call it using the temporary interface at:
https://github.com/fyne-io/fyne/pull/1292/files#diff-fc4b8de1d5fbe30aec3d3d9d22e86237R102.

As you say this is not exported in Settings API, but it can be used by other developers if they dig deep enough.
This is a special case as we want to apply the theme temporarily without updating the whole application.

The only other approach I can think of is to spwan another process, save the output to an image and load the image file in (becuse we cannot run multiple apps in one runtime, and because the widgets use "current theme" to render).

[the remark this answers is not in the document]
stuartmscott: Ah okay, disregard

andydotxyz:
Nice suggestion on focused...
I wonder if I could sneak a "Downloading <progress bar>" to the bottom of the home screen without it looking too weird?
That's about all the primary I can think of too.

stuartmscott:
Maybe, though I think we should get this in first and iterate later

andydotxyz:
Screenshot 2020-10-07 at 15 50 30

stuartmscott (approved):

Issue #1299, opened 2020-09-09
Software Renderer doesn't draw canvas.Raster


stuartmscott:

stuartmscott:
Fix confirmed

PR #1300, opened 2020-09-09
Add Color Widgets

stuartmscott:
- ColorArea - for displaying a color space and responding to taps
- ColorButton - for displaying a color and responding to taps
- ColorChannel - for displaying and editing a particular channel of a color
- ColorBasicPicker - a picker for basic colors
- ColorGreyscalePicker - a picker for greyscale colors
- ColorRecentPicker - a picker for recent colors
- ColorAdvancedPicker - a picker for custom colors
- ColorPickerDialog - a dialog combining basic, greyscale, recent, and advanced pickers

Pending issues
- Alignment of ColorChannel Label, Entry, and Slider is not right.
- Typing in entry triggers onchange callback which updates UI but also
  resets cursor, perhaps the update should be triggered on focus lost?
- Tapping on ColorArea results in selection slightly down and right,
  possibly due to padding.
- RGB ColorArea isn't implemented yet, not sure of the right approach for
  reducing the 3D color space (RGB) to a 2D image. Maybe a wheel like HSL?
- ColorArea and ColorAdvancedPicker image tests are incorrect because
  software renderer does not draw canvas.Raster (#1299)
- ColorPickerDialog confirm button can be confusing for users as tapping
  a colorbutton (basic, greyscale, recent) will immediately pick the color.
  The confirm button is for the user to choose the color currently in the
  advanced picker. The confirm was added next to the cancel button for
  consistency rather than only having it in the advanced picker but perhaps
  there is a better way to solve this?

### Checklist:

- [x] Tests included. (mainly image tests, but more will be added)
- [x] Tests all pass. (except ColorArea, see above)

#### Where applicable:


stuartmscott:
Screen Shot 2020-09-09 at 3 57 42 PM

andydotxyz:
The test failures are missing master images I think

andydotxyz:
This looks really great, thanks @stuartmscott. Just glancing at the screenshot I wonder - if the solid colour preview was just a horizontal rectangle the height of the alpha slider we'd save a bunch more space.

stuartmscott:
> The test failures are missing master images I think

Correct, will update once #1301 lands.

> I don't know about all of these exported APIs, they are components of the dialog and, until we see other use cases, I would recommend leaving them unexported. As far as I can see the only really needed method is dialog.NewColorPicker() (and possibly the dialog.ShowColorPicker() helper).

I disagree somewhat as the use case I have is a toolbar item displaying a ColorButton that when tapped opens the ColorPickerDialog. To do this the following need to be exported:
- ColorButton
- ColorPickerDialog - the type is exported so dev can call "SetColor(c color.Color)" before calling Show. ShowColorPicker would be a nice helper, will add.

If a dev wants to display a color picker outside of a dialog context then these should also be exported:
- ColorBasicPicker - not actually a type, just a helper method
- ColorGreyscalePicker - not actually a type, just a helper method
- ColorRecentPicker - not actually a type, just a helper method
- ColorAdvancedPicker

The following could be unexported as they aren't typically used standalone:
- ColorArea
- ColorChannel

> This looks really great, thanks @stuartmscott.

Awesome, you're very welcome.

> Just glancing at the screenshot I wonder - if the solid colour preview was just a horizontal rectangle the height of the alpha slider we'd save a bunch more space.

Good point, tho I had ear marked that extra space for a couple of entries for reading/writing the chosen color as Hex/Web formats (#ffffff, rgb(126,254,32), hsl(359,1,1)) for easy copy/paste. But the dialog is rather tall and thin, perhaps a relayout could balance it out.

andydotxyz:
Sorry I meant to include the "ColorPickerDialog" type in the export along with ShowColorPickerDialog.
I'm not sure about ColorButton - this is just a rectangle that is tappable right? 

The other types could be helpful if someone wanted to replicate the functionality outside of the picker. But the same could be said of the fileitem and filelisting code in File dialog but we agreed to land it first, understand usage, polish API then export the right, minimal API over time - should we not take the same approach here?

andydotxyz:
I think travis is erroring on a formatting issue in pool now.
Which is strange as that's not part of this PR...

stuartmscott:
ColorButton is a coloured rectangle that handles hover and tap, which is relatively simple. However if this is unexported I will have to write another one in my app. Similarly I need a different dialog style, which is why the Basic, Greyscale, Recent, and Advanced pickers are exported.

Ah yeah, the pool error is about import ordering - weird that Travis didn't detect this before

Jacalz (commented):
Just a little thought from my side. This seems like a very specialized set of functionality. Would it make more sense to move the widget code into something like `fyne.io/widget/color` instead?

widget/color.go
```diff
@@ -0,0 +1,254 @@
...
+		a = float64(alpha) / 65535.0
+	}
+	return
+}
+
+func cartesianToPolar(x, y, width, height float64) (angle, radius, limit float64) {
+	dx := float64(x - (width / 2))
+	dy := float64(y - (height / 2))
+	angle = math.Atan2(dy, dx)
+	radius = math.Sqrt(dx*dx + dy*dy)
```
Jacalz, line 147: Should we perhaps be using `complex128` instead and just use the `cmplx.Polar()` function from the standard library? https://pkg.go.dev/math/cmplx?tab=doc#Polar

widget/color.go
```diff
@@ -0,0 +1,254 @@
...
+	max := math.Max(r, math.Max(g, b)) //Max. value of RGB
+	del := max - min                   //Delta RGB value
+
+	l := (max + min) / 2.0
+
+	var h, s float64
+
+	if del == 0.0 { // Achromatic
+		h = 0.0
+		s = 0.0
```
Jacalz, line 183: It might be a bit nicer to just do this and avoid the else statement.
```suggestion
		return 0, 0, l
```

Jacalz (commented):
Managed to find some smaller things. Apart from this, code looks good to me :)

stuartmscott:
Screen Shot 2020-09-11 at 5 03 48 PM

stuartmscott:
New layout:
Screen Shot 2020-09-12 at 9 08 47 AM
Screen Shot 2020-09-12 at 9 09 04 AM

widget/color.go
stuartmscott: Fixed

widget/color.go
stuartmscott: It is a little weird to say X is real, and Y is imaginary but in general it is better to use a well-tested standard library

andydotxyz:
The latest layout in advanced shows a lot of sliders and information, I think when it was tabbed it was clearer for the uninitiated. RGB probably a good default for most users.
I wondered... If we keep the wider look (which is cleaner I think than the long thin) what if the sliders / value / name were on one line? (i.e. Hue: ----o---         25%) maybe it could be a smoother flow?

The above is just thoughts, rather subjective. What I consider more important is that the raw values move to the formatted number (% for HSL and 0-255 for RGB).

stuartmscott:
@andydotxyz the channel layout, and range changes have been made.

dialog_expanded_theme_dark

dialog/color.go
```diff
@@ -0,0 +1,334 @@
...
+	return
+}
+
+func colorToHSLA(c color.Color) (int, int, int, int) {
+	r, g, b, a := colorToRGBA(c)
+	h, s, l := rgbToHsl(r, g, b)
+	return h, s, l, a
+}
+
+func colorToRGBA(c color.Color) (r, g, b, a int) {
```
andydotxyz, line 218: This function seems like it could be simplified.
You could resolve your TODO and improve this at the same time.
Just call .RGBA() on any colour.Color and then divide by the alpha value (if non-1 and non-0)

dialog/color_test.go
```diff
@@ -0,0 +1,345 @@
...
+			r, g, b := hslToRgb(tt.h, tt.s, tt.l)
+			assert.Equal(t, tt.r, r)
+			assert.Equal(t, tt.g, g)
+			assert.Equal(t, tt.b, b)
+		})
+	}
+}
+
+func Test_hueToRgb(t *testing.T) {
+	// TODO
```
andydotxyz, line 344: Should this be filled in?

dialog/color_wheel.go
```diff
@@ -0,0 +1,209 @@
...
+	radians := degrees * math.Pi / 180.0
+	c := cmplx.Rect(radius, radians)
+	return int(real(c) + w/2.0), int(imag(c) + h/2.0)
+}
+
+func (a *colorWheel) trigger(pos fyne.Position) {
+	x, y := a.locationForPosition(pos)
+	if c := a.cache; c != nil {
+		b := c.Bounds()
+		if f := a.onChange; f != nil {
```
andydotxyz, line 154: The body of this function is nested in two nil checks, can it be cleaner with corner cases exit early?
The a.Refresh could be in defer if it needs to run last no matter what

cmd/fyne_demo/screens/window.go
```diff
@@ -161,6 +171,18 @@ func loadDialogGroup(win fyne.Window) *widget.Card {
 				fileSaved(writer)
 			}, win)
 		}),
+		widget.NewButton("Simple Color Picker", func() {
```
andydotxyz, line 174: I don't think this needs both - just add one, the type of picker feels more like a config rather than a new widget

dialog/color.go
```diff
@@ -0,0 +1,334 @@
...
+		},
+	}
+	p.dialog.setButtons(newButtonList(p.dialog.dismiss, confirm))
+	return p
+}
+
+// NewSimpleColorPicker creates a simple color dialog and returns the handle.
+// Using the returned type you should call Show() and then set its color through SetColor().
+// The callback is triggered when the user selects a color.
+func NewSimpleColorPicker(title, message string, callback func(c color.Color), parent fyne.Window) *ColorPickerDialog {
```
andydotxyz, line 57: I'm not really seeing the benefit of two APIs here (SimpleColorPicker vs ColorPicker) - for one thing the docs don't even explain the difference.
You are returning a ColorPickerDialog, so I think that "ShowAdvanced" or "HideAdvanced" or some other bool would be appropriate to expose the fact that developers can decide whether or not that option is available.
So `p := dialog.NewColorPicker(); p.HideAdvanced; p.Show()` would remove the need for the simple api variants.

andydotxyz (changes requested):
A few inline comments, some pending TODOs.
But mostly I don't think the API variants (Simple vs otherwise) don't really need double the number of constructors, it's more like an option that can be set on the type returned. Should make for a cleaner API.

andydotxyz:
Should the recent row should be visible all the time perhaps?

dialog/color.go
stuartmscott: Okay cool, that did simplify it, thanks. I did need to add two cases for NRGBA tho which seems odd - I got an error that it was expecting *color.NRGBA but was actually color.NRGBA :/

dialog/color_test.go
stuartmscott: Filled!

dialog/color_wheel.go
stuartmscott: Fixed! Tho I merged the `if`s rather than returning early

stuartmscott:
> Should the recent row should be visible all the time perhaps?

What's the rationale? I don't see the advantage in displaying empty recent buttons.

Currently, when a user opens the picker and picks a colour, the next time they open the picker the first difference they will see is the last colour they chose

stuartmscott:
Screen Shot 2020-09-18 at 5 32 10 PM

cmd/fyne_demo/screens/window.go
stuartmscott: I've added `dialog.Refresh` and `Advanced` bool to `ColorPickerDialog` so devs can toggle the functionality, but it may still be a good idea to demo both so they can see the difference, thoughts?

andydotxyz: Sure that makes sense. Should advanced be off by default? I am torn :)

andydotxyz (commented):
Looking good.
I wonder though is it possible to not require Refresh() after Advanced=true if it's not yet been shown?
Other widgets would only require Refresh if it's changed after display...

dialog/color.go
```diff
@@ -49,6 +49,9 @@ func (p *ColorPickerDialog) SetColor(c color.Color) {
 
 // Refresh causes this dialog to be updated
 func (p *ColorPickerDialog) Refresh() {
+	if w := p.win; w != nil {
+		w.Hide()
```
andydotxyz, line 53: This seems strange... It will hide an existing dialog if you call Refresh() - but I don't see why.
a) there is no Show() at the end of this method
b) can't we change the content of the dialog without hiding it first?

stuartmscott: The problem is that Show doesn't create the popup, setButtons does

andydotxyz: Sure, but setButtons could be called later right?

stuartmscott: Where should it be called?

andydotxyz: My immediate thought would be that it's set the first time Show is called - would that work OK?
After first Show then Refresh would clearly be needed.

stuartmscott: Refresh and Show now both delegate to a helper, does this resolve your concern or have I misunderstood?

cmd/fyne_demo/screens/window.go
stuartmscott: I think the default should be simple, with the option of more - it feels odd to have the default be everything and the dev has to "downgrade"

stuartmscott:
dialog_expanded_theme_light

dialog/color.go
andydotxyz: That seems good

cmd/fyne_demo/screens/window.go
```diff
@@ -161,6 +171,20 @@ func loadDialogGroup(win fyne.Window) *widget.Card {
...
+				colorPicked(c, win)
+			}, win)
+			picker.Show()
+		}),
+		widget.NewButton("Advanced Color Picker", func() {
+			picker := dialog.NewColorPicker("Pick a Color", "What is your favorite color?", func(c color.Color) {
+				colorPicked(c, win)
+			}, win)
+			picker.Advanced = true
+			picker.Refresh()
```
andydotxyz, line 185: Can you take this line out so we know it works without :)

andydotxyz:
It's looking great. Shame about that shadow!

andydotxyz (dismissed):
Mammoth effort!

cmd/fyne_demo/screens/window.go
stuartmscott: Done!

andydotxyz (approved):

PR #1305, opened 2020-09-11
Fixing style of accordion to use dividers

andydotxyz:
Don't use button shadows any more

### Checklist:
- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
Screenshot 2020-09-11 at 23 40 39

widget/accordion.go
```diff
@@ -129,20 +129,26 @@ type accordionRenderer struct {
...
+			div := r.dividers[i-1]
+			div.Move(fyne.NewPos(x, y-1-theme.Padding())) // middle of padding gap between items
+			div.Resize(fyne.NewSize(size.Width, divHeight))
 		}
 		h := r.headers[i]
 		h.Move(fyne.NewPos(x, y))
 		min := h.MinSize().Height
 		h.Resize(fyne.NewSize(size.Width, min))
-		y += min
+		y += min + divHeight
```
toaster, line 151: This even adds `divHeight` to the `padding` space after the divider (which has then `padding-1` before and `padding+divHeight` after). This line should be reverted, especially with the proposed changes from above.

```suggestion
		y += min
```

widget/accordion.go
```diff
@@ -155,9 +161,10 @@ func (r *accordionRenderer) Layout(size fyne.Size) {
 }
 
 func (r *accordionRenderer) MinSize() (size fyne.Size) {
+	divHeight := 1
```
toaster, line 164: There should only be one authoritative answer to the question for `divHeight`. It should be extracted into a constant.

toaster (changes requested):
The `fyne` command update should at least have had its own commit. I assume it’s okay.

For the divider, see my suggestions. The math doesn't seem to be right but the pictures look okay. Either it’s just one unnoticable pixel or something else is wrong.
However, the math should be corrected.

andydotxyz:
> The `fyne` command update should at least have had its own commit. I assume it’s okay.

Ooops! it was not supposed to be in there, sorry - reverted

widget/accordion.go
```diff
@@ -129,20 +129,26 @@ type accordionRenderer struct {
...
 	x := 0
 	y := 0
 	for i, ai := range r.container.Items {
 		if i != 0 {
-			y += theme.Padding()
+			y += theme.Padding() * 2
+
+			div := r.dividers[i-1]
+			div.Move(fyne.NewPos(x, y-1-theme.Padding())) // middle of padding gap between items
+			div.Resize(fyne.NewSize(size.Width, divHeight))
```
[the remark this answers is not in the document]
andydotxyz: I've rearranged the code to be more logical. There was indeed an off-by-one for the detail when i>0 thanks.

widget/accordion.go
andydotxyz: Good catch, fixed in response to above comment

widget/accordion.go
andydotxyz: Indeed, corrected

widget/accordion.go
```diff
@@ -151,13 +158,17 @@ func (r *accordionRenderer) Layout(size fyne.Size) {
...
+			y += theme.Padding()
+		}
 	}
 }
 
 func (r *accordionRenderer) MinSize() (size fyne.Size) {
 	for i, ai := range r.container.Items {
 		if i != 0 {
-			size.Height += theme.Padding()
+			size.Height += theme.Padding()*2 + accordionDividerHeight
```
stuartmscott, line 171: Originally this code only added padding between items. With this change an Accordion with a single item adds no padding - which is correct, but one with two items will have `padding*2+1` - I'm not sure about `*2`, doesn't this put double padding between items, what is the rationale here?

widget/accordion.go
```diff
@@ -151,13 +158,17 @@ func (r *accordionRenderer) Layout(size fyne.Size) {
 			d.Resize(fyne.NewSize(size.Width, min))
 			y += min
 		}
+
+		if i < len(r.container.Items)-1 {
+			y += theme.Padding()
+		}
```
stuartmscott, line 164: This adds `padding` between items, but line 145 already adds `padding + 1` between items

widget/accordion_test.go
```diff
@@ -197,12 +197,12 @@ func TestAccordionContainer_Layout(t *testing.T) {
...
 func TestAccordionContainer_MinSize(t *testing.T) {
 	minSizeA := fyne.MeasureText("A", theme.TextSize(), fyne.TextStyle{})
-	minSizeA.Width += theme.IconInlineSize() + theme.Padding()*7
-	minSizeA.Height = fyne.Max(minSizeA.Height, theme.IconInlineSize()) + theme.Padding()*4
+	minSizeA.Width += theme.IconInlineSize() + theme.Padding()*5
+	minSizeA.Height = fyne.Max(minSizeA.Height, theme.IconInlineSize()) + theme.Padding()*2
 	minSizeB := fyne.MeasureText("B", theme.TextSize(), fyne.TextStyle{})
-	minSizeB.Width += theme.IconInlineSize() + theme.Padding()*7
-	minSizeB.Height = fyne.Max(minSizeB.Height, theme.IconInlineSize()) + theme.Padding()*4
-	minSize1 := fyne.MeasureText("111111", theme.TextSize(), fyne.TextStyle{})
```
stuartmscott, line 205: Why is this string being shortened throughout this test?

stuartmscott (changes requested):

widget/accordion_test.go
andydotxyz: I think there is a rounding issue in string measurement so it was getting off-by-1s with that exact string

widget/accordion.go
andydotxyz: Between each item is pad+divide+pad, hence padding*2+1.
It was originally just padding, which resulted in pad/2+div+pad/2 and we agreed it looked squashed.
See the test images if you're unsure and let me know what yout think.

widget/accordion.go
andydotxyz: Correct, see above for explanation

widget/accordion_test.go
stuartmscott: I made it 6 chars so it was long enough to trigger something - can't quite remember.

Changing test data so the tests pass is not typically the right approach.

andydotxyz: Very true, but the graphica output looked right so I could not figure the issue.
If you could remember what it was testing that would be really helpful. Otherwise it's not clear why the longer string is the magic value that we must test

andydotxyz: Reverting the tests to fix and turns out that having reverted buttons shadow had made the problem go away.

stuartmscott (approved):

toaster (approved):

PR #1307, opened 2020-09-12
Run code checks on newest go, not oldest

Jacalz:
### Description:
This makes sure that we run our code checks on the latest version of Go, not the oldest. Should help us hopefully catch a few more things than what we otherwise would have done.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ matrix:
   include:
     - go: 1.12.x
       os: linux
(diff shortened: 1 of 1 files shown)


stuartmscott (commented):
Could we run on both the oldest and newest?

toaster (approved):

toaster:
> Could we run on both the oldest and newest?

Of course, but what would this be good for?

PR #1308, opened 2020-09-12
Move code checks to Github Actions and enable staticcheck

Jacalz:
### Description:
This makes sure to clean up more code that staticcheck flags as incorrect. My goal is to be able to run staticcheck on the whole codebase as part of CI, but it is currently held up by a few issues. Hopefully those can get ironed out in the future to give us some better code checks :slightly_smiling_face: 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


stuartmscott (approved):
Always good to see PRs that improve the code base

cmd/fyne_settings/settings/bundled.go
```diff
@@ -16,7 +16,9 @@ var appearanceIcon = &fyne.StaticResource{
...
+*/
```
stuartmscott, line 24: See the first line in this file - this change will get reverted the next time the file is generated. Instead remove https://github.com/fyne-io/fyne/blob/master/cmd/fyne_settings/data/gen.go#L60

stuartmscott (changes requested):

cmd/fyne_settings/settings/bundled.go
Jacalz: Yeah good catch. I don't know why I didn't think of that :sweat_smile:

stuartmscott (approved):

stuartmscott (changes requested):
Travis has a failure:
`fatal error: runtime: out of memory`

Jacalz:
Hm, that is a weird one. I guess that `staticcheck` can be a bit memory hungry :(

Jacalz:
Have been working on moving code checks out to Github Actions. I guess that now might be the time :question:

dominikh:
> Hm, that is a weird one. I guess that staticcheck can be a bit memory hungry :(

@Jacalz Were you using the latest release (2020.1.x) of staticcheck, or the master branch? The master branch should use significantly less memory, and if it still OOM'd in Travis then that's probably something I need to investigate.

Jacalz:
@dominikh I was using the latest stable release, I think. That should be what a regular `go get` should give us right?

I know that the stable releases have been quite memory hungry on my own machine (before I upgraded to 16GB of memory from 8GB, my Linux machine used to almost max out memory usage). I moved over to GitHub Actions to try to avoid Travis and that seemed to work fine.

dominikh:
> That should be what a regular go get should give us right?

In module mode, yes. In GOPATH mode, that downloads the master branch.

I'll make a note to check staticcheck's memory usage on this project.

andydotxyz (approved):

stuartmscott (approved):

PR #1312, opened 2020-09-14
WIP: Refactoring of cli to use spf13/cobra

Jacalz:
### Description:
This refractors the fyne cli command to use `spf13/cobra` for building the backbones of the application structure. This simplifies a lot of our homegrown cli code, makes it look a bit nicer and makes it easier to add new functionality in the future.

image

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


andydotxyz:
Just skimming through can you please make sure that it's backwards compatible? On the surface it looks like "fyne version" may not work any more...

Jacalz:
Yeah good catch there. Will change the versioning to behave like it used to, for now. Tried to keep as much of the original code as possible to avoid breakages and make reviewing easier.

Will make sure to test that everything works and is backwards compatible before marking this as ready. :)

Jacalz:
TODO: Remind myself that dependencies need to be vendored. That is the reason for Travis failures.

Jacalz:
Notes to myself:
- Install is currently not working correctly
- Should perhaps avoid printing usage in a few places
- Get is currently not working
-  Some flags need better descriptions

Jacalz:
This was deemed to not meet demands due to the strict posix compliance.

PR #1318, opened 2020-09-16
Bugfix #948: support focus traversal in dialogs

### LINKED ISSUE #948: Support Focus Traversal in Dialogs
Although it is possible for windows, it is not possible to use tab to traverse input elements in a dialog.

Not that I know of.

Similar functionality as seen in windows, ported to dialogs as well.

toaster:
### Description:

This PR introduces `FocusManager`s per overlay and the according handling in the `glCanvas`. Thus it is possible to traverse `Focusable`s in dialogs (and other overlays).

Fixes #948 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


internal/driver/glfw/canvas.go
```diff
@@ -386,6 +389,21 @@ func (c *glCanvas) objectTrees() []fyne.CanvasObject {
...
+	if prevFocusMgr == nil {
+		prevFocusMgr = c.focusMgr
+	}
+	focusMgr := c.overlays.TopFocusManager()
+	if focusMgr == nil {
+		focusMgr = c.focusMgr
+	}
+	c.RUnlock()
+	prevFocusMgr.FocusLost()
+	focusMgr.FocusGained()
```
stuartmscott, line 404: Is it possible for `prevFocusMgr == focusMgr`?

stuartmscott (commented):
Looks good to me!

internal/driver/glfw/canvas.go
toaster: Good catch! It might be possible with concurrent changes to the overlay stack.
I’ll add missing locking to the stack and also change the callback to be safe by design.

internal/driver/glfw/canvas.go
```diff
@@ -386,6 +384,20 @@ func (c *glCanvas) objectTrees() []fyne.CanvasObject {
 	return trees
 }
 
+func (c *glCanvas) overlayChanged(focusMgr, prevFocusMgr *app.FocusManager) {
+	c.Lock()
+	defer c.Unlock()
```
stuartmscott, line 389: nit: It might be better to not use a defer here, and instead move `c.Unlock()` down to after line 396. In case dev is doing something weird in Focus{Lost|Gained} which could cause deadlock

internal/driver/glfw/canvas.go
stuartmscott: Instead of locking, what if you just return early if they are equal?

stuartmscott (approved):

PR #1319, opened 2020-09-16
fix: fyne package targeting windows

garrettcorn:
### Description:
fyne package is grabbing GOARCH from runtime.GOARCH when building for windows causing issues when the target architecture doesn't match local architecture.

Fixes #(issue)
Change fyne package to grab GOARCH from the environment variable GOARCH and fallback to runtime.GOARCH if GOARCH is not set.

### Checklist:

- [x] Tests all pass.


Jacalz (approved):
Thanks for the fix. Looks good to me 👍

andydotxyz (approved):

andydotxyz:
Great thanks!

andydotxyz:
That's on develop now ready for 1.4 release, thanks

PR #1322, opened 2020-09-18
Fix dividers on accordion remove and add

andydotxyz:
### Description:
Just ensuring dividers hide (and show - and that headers re-show)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/accordion.go
```diff
@@ -198,6 +200,7 @@ func (r *accordionRenderer) updateObjects() {
 		var h *Button
 		if i < hs {
 			h = r.headers[i]
+			h.Show()
```
stuartmscott, line 203: nit: is this needed?

stuartmscott (dismissed):

widget/accordion.go
andydotxyz: Some times - a header can be hidden, so if it is to be reused after hidden it must be re-shown.

andydotxyz:
Conflict on test invalidated review :(

stuartmscott (approved):

PR #1328, opened 2020-09-21
 Container add Remove finally

andydotxyz:
### Description:
Add Remove, rename AddObject to Add. 
People keep asking, so let's do this. And the "Object" postfix isn't used elsewhere

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container.go
+++ b/container.go
@@ -14,99 +14,6 @@ type Container struct {
 	Objects []CanvasObject // The set of CanvasObjects this container holds
 }
 
-func (c *Container) layout() {
-	if c.Layout != nil {
-		c.Layout.Layout(c.Objects, c.size)
-		return
-	}
-}
-
-// Size returns the current size of this container.
-func (c *Container) Size() Size {
-	return c.size
-}
-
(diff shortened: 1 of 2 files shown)


container.go
```diff
@@ -14,37 +14,50 @@ type Container struct {
...
-	return c.size
+// NewContainerWithoutLayout returns a new Container instance holding the specified CanvasObjects
+// that are manually arranged.
+func NewContainerWithoutLayout(objects ...CanvasObject) *Container {
+	ret := &Container{
+		Objects: objects,
+	}
+
+	ret.size = ret.MinSize()
+	ret.layout()
```
Jacalz, line 32: Like stated in Slack, can probably be removed ;)

Jacalz (commented):
just a small note on a layout call. Otherwise it looks good to me :+1:

container.go
andydotxyz: Fixed, thanks

stuartmscott (approved):

PR #1330, opened 2020-09-21
List fixes for removing items and clearing the list

### LINKED ISSUE #1326: List widget: Inconsistent layout of children
### Describe the bug:
I found that adding a lot of items to the list before showing it results in a consistent layout of the items, but keeping the list empty and then adding items as we go leads to items being laid out in a very inconsistent way.

### To Reproduce:
1. Create an empty list 
2. Add one or more items, when for example pressing a button
3. Compare to a list with all items already created

#### Items added before:
image

#### Items added after:
image

### Example code:
https://github.com/Jacalz/wormhole-gui/blob/list-widget/bridge/widgets/list.go

 - OS: Linux
 - Version: 5.6.18
 - Go version: 1.15.2
 - Fyne version: 0a2fa00d4f28

okratitan:
### Description:
Fix issues where list wasn't refreshing when clearing contents
Fix issues where list would crash if contents were cleared and scroller offset was not 0
Fix issues where list would crash removing items one by one
Add the ability to clear contents and append list items or remove list items one by one to the demo.

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

#### Where applicable:


stuartmscott:
>  Tests included

I don't see any new tests introduced

Jacalz (changes requested):
Like @stuartmscott  pointed out, it seems like the tests somehow got removed. I see that they are there in earlier commits though. Would you mind adding them back in again? Apart from that, it looks good to me :)

Jacalz (approved):
Looks good to me. From discussion in slack, looks like tests were never part of PR.

cmd/fyne_demo/screens/widget.go
```diff
@@ -305,18 +305,29 @@ func makeListTab() fyne.CanvasObject {
...
 		func(index int, item fyne.CanvasObject) {
-			item.(*fyne.Container).Objects[1].(*widget.Label).SetText(data[index])
+			item.(*fyne.Container).Objects[2].(*widget.Label).SetText(data[index])
 		},
 	)
 	list.OnItemSelected = func(index int) {
 		label.SetText(data[index])
 		icon.SetResource(theme.DocumentIcon())
 	}
+	clearButton := widget.NewButton("Clear Contents",
```
andydotxyz, line 318: It seems like this is really deomnstration that a bug is fixed - so it should be in a unit test, not the demo UI...

andydotxyz (changes requested):
Let's not put tests into the demo UI but write unit tests.

This PR has a strange name, it's not really about layout fixes in general but specifically fixing refresh when the data list is emptied.
In my own tests I find there is still an issue if 1 item is removed, for example down from 3 to 2 items will crash still.

widget/list.go
```diff
@@ -176,6 +177,15 @@ func (l *listRenderer) getItem() fyne.CanvasObject {
 }
 
 func (l *listRenderer) offsetChanged() {
+	if l.list.Length() == 0 {
```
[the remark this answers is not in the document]
okratitan: If the offset of the scroller is not 0 and then the list is cleared and refreshed, the offset of the scroller changes from X to 0.  This is not due to a scroll but an adjustment due to the scroller's child's min size changing.  Makes sense to me.

cmd/fyne_demo/screens/widget.go
okratitan: I added more unit tests should be good to go now

widget/list.go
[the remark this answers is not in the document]
andydotxyz: If it is triggered by other code then why not put it there? 
Perhaps this deserves to be a different method like "resetScroll" and it's called when needed rather than in this scroll callback?

widget/list.go
```diff
@@ -117,6 +117,12 @@ func (l *listRenderer) Layout(size fyne.Size) {
 
 	// Relayout What Is Visible - no scroll change - initial layout or possibly from a resize.
 	l.visibleItemCount = int(math.Ceil(float64(l.scroller.size.Height) / float64(l.list.itemMin.Height)))
+	if len(l.children) > l.list.Length() {
+		for i := len(l.children); i > l.list.Length(); i-- {
+			l.itemPool.Release(l.children[len(l.children)-1])
+			l.children = l.children[:len(l.children)-1]
```
andydotxyz, line 123: Can't this be done once after the loop its in is complete?
Release l.children[i-1] in each, then re-slice at the end

widget/list_test.go
```diff
@@ -161,10 +161,74 @@ func TestList_SmallList(t *testing.T) {
 			item.(*fyne.Container).Objects[1].(*Label).SetText(data[index])
 		},
 	)
-	list.Resize(fyne.NewSize(200, 1000))
+	w := test.NewWindow(list)
+	w.Resize(fyne.NewSize(200, 1000))
```
andydotxyz, line 165: Maybe these images could be smaller? that's a lot of empty space

andydotxyz (changes requested):

cmd/fyne_demo/screens/widget.go
andydotxyz: The new tests are great thanks, but do we need this row of buttons in the demo?

widget/list.go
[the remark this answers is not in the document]
okratitan: I'm not sure why you think this is a scroll callback?  The callback is called offsetChanged - which happens when the scroller alerts us that the offset has changed - which is reasonably the case when the min size of the scroller content changes?  It makes sense for me to reset our offset tracking code to 0 in the offset changing callback when the offset changing callback is called with 0.  Would it make it more understandable if the code was if l.scroller.Offset.Y == 0 instead of l.list.Length() == 0?

widget/list.go
okratitan: Sure

widget/list_test.go
okratitan: sure

widget/list.go

widget/list_test.go
okratitan: Resolved

widget/list.go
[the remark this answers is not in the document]
okratitan: Moved variable resetting to the Length == 0 check inside of renderer's Layout called by Refresh.

cmd/fyne_demo/screens/widget.go
okratitan: I'm not sure.  I think it's nice to exhibit that the list is dynamic but I guess that is the teams call.  I just put it there to show how list could add/remove items dynamically.  (And it is nice to have for when people say okay to replicate add some list items, then remove one, then clear, then add, then remove, boom crash, etc...)

andydotxyz: It's just not how any of the other widgets are presented, so seems strange in comparison

widget/list.go
```diff
@@ -116,17 +123,23 @@ func (l *listRenderer) Layout(size fyne.Size) {
 
 	// Relayout What Is Visible - no scroll change - initial layout or possibly from a resize.
 	l.visibleItemCount = int(math.Ceil(float64(l.scroller.size.Height) / float64(l.list.itemMin.Height)))
+	if len(l.children) > l.list.Length() {
```
andydotxyz, line 126: I did not notice that this code was repeated 3 times sorry, I can't help but think this is not all needed.
For example the first two are doing the same, reducing to the smaller of l.list.Length() and l.visibleItemCount (I think).
So this could be a single loop where the end point was the smaller of these two?

widget/list.go
```diff
@@ -116,17 +123,23 @@ func (l *listRenderer) Layout(size fyne.Size) {
...
+	}
 	if len(l.children) > l.visibleItemCount {
 		for i := len(l.children); i > l.visibleItemCount; i-- {
-			l.itemPool.Release(l.children[len(l.children)-1])
-			l.children = l.children[:len(l.children)-1]
+			l.itemPool.Release(l.children[i-1])
 		}
+		l.children = l.children[:l.visibleItemCount]
 	}
 	if l.visibleItemCount > 0 && l.list.Length() < l.visibleItemCount && len(l.children) > l.visibleItemCount-1 {
```
andydotxyz, line 138: With the fix in place above it looks like this code will no longer be executed - it iterates through the difference of visibleItemCount and list.Length(), but between the code loops above are you not ensuring that those items are already gone?

andydotxyz (changes requested):
Nice thanks.
Just a bit of duplication or redundant code that looks like it could be tidied.

andydotxyz (approved):

PR #1331, opened 2020-09-21
Ensure tab children are resized before being shown

stuartmscott:
### Description:
TabContainer calls Show before calling Resize, resulting in content being laid out with 0x0 size.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


andydotxyz (approved):

andydotxyz:
Nice catch, subtle!

PR #1339, opened 2020-09-25
Move FileIcon to widget and make it more dynamic

Jacalz:
### Description:
This moves over the dialog.NewFileIcon to the widget package, tires to make it more in line with other widgets while also making it more dynamic to make it possible to update the URI after creation.

This is currently a draft due to text scaling issues and tests that need to be added back in.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


Jacalz:
The tests are failing and I just can't figure out why that is the case. I have looked and looked without gaining nay insight to understanding why they are failing. The file dialog is working just like I would expect it to.

andydotxyz:
Internal tests are failing due to the tested state not being set until CreateRenderer is called. So unrendered icons have no state, hence tests fail.

Jacalz:
There might be an issue with text placement and size for smaller text sizes, but apart from that, this should be ready for review.

widget/fileicon.go
```diff
@@ -82,9 +78,11 @@ func (i *FileIcon) CreateRenderer() fyne.WidgetRenderer {
 	i.propertyLock.RLock()
 	defer i.propertyLock.RUnlock()
 
-	i.updateURI(i.URI)
+	i.setURI(i.URI)
+
 	s := &fileIconRenderer{file: i}
 	s.updateObjects()
+	i.cachedURI = i.URI
```
andydotxyz, line 92: isn't setURI above going to do this?

andydotxyz (changes requested):
A tiny duplication, otherwise this looks good.

widget/fileicon.go
Jacalz: I don't think it should. If it did, the updating of new objects in refresh would not be called otherwise.
Refresh alls updates the cachedURI after objects have been updated. A bit like how the icon caching for widget.Icon sets cache to `nil` and then updates after items have been updated.

Jacalz: But thinking about this, it might no be needed here specifically. Because the widget will be calling `Refresh()` before it is shown, right?

andydotxyz: Apolgies, I misread line 65

andydotxyz (dismissed):

Jacalz:
This should be ready for a re-review now. The y-position for the text doesn't work perfect when the height differs a lot from the width, which I can't really figure out why. Tried to get it better, but this is the best I can get it.

andydotxyz (approved):
This looks good to me.
I'm not sure how to replicate the scaling issues you mention, but it looks good on the file dialog.

Jacalz:
Thanks @andydotxyz. I only see the scaling issue when the size.Height is much bigger than the width.

andydotxyz:
The main issue was that the extension height was being adjusted and so it doubled up the offset.
I have tidied up a little bit of the maths and re-introduced the constants again. I think the result is what you want:

```diff
diff --git a/widget/fileicon.go b/widget/fileicon.go
index 66e7819b..0133d7a8 100644
--- a/widget/fileicon.go
+++ b/widget/fileicon.go
@@ -10,6 +10,11 @@ import (
        "fyne.io/fyne/theme"
 )
 
+const (
+       ratioDown = 0.45
+       ratioTextSize = 0.22
+)
+
 // FileIcon is an adaption of widget.Icon for showing file information
 type FileIcon struct {
        BaseWidget
@@ -109,17 +114,19 @@ func (s *fileIconRenderer) MinSize() fyne.Size {
 }
 
 func (s *fileIconRenderer) Layout(size fyne.Size) {
-       isize := min(size.Width, size.Height)
+       isize := fyne.Min(size.Width, size.Height)
 
        xoff := 0
-       yoff := int(float64(size.Height) * 0.40)
+       yoff := (size.Height - isize) / 2
 
        if size.Width > size.Height {
                xoff = (size.Width - isize) / 2
+               yoff = 0
        }
+       yoff += int(float64(isize) * ratioDown)
 
-       s.ext.TextSize = int(float64(isize) * 0.22)
-       s.ext.Resize(fyne.NewSize(isize, yoff))
+       s.ext.TextSize = int(float64(isize) * ratioTextSize)
+       s.ext.Resize(fyne.NewSize(isize, s.ext.MinSize().Height))
        s.ext.Move(fyne.NewPos(xoff, yoff))
 
        s.Objects()[0].Resize(size)
@@ -178,10 +185,3 @@ func splitMimeType(uri fyne.URI) string {
        }
        return mimeTypeSplit[0]
 }
-
-func min(a, b int) int {
-       if a < b {
-               return a
-       }
-       return b
-}
```

andydotxyz (approved):
Nice :)

PR #1344, opened 2020-09-26
Feature/table

andydotxyz:
### Description:
Adding table widget

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


widget/scroller.go
```diff
@@ -279,6 +279,10 @@ func (r *scrollContainerRenderer) MinSize() fyne.Size {
 }
 
 func (r *scrollContainerRenderer) Refresh() {
+	if len(r.BaseRenderer.Objects()) == 0 || r.BaseRenderer.Objects()[0] != r.scroll.Content {
+		// push updated content object to baseRenderer
+		r.BaseRenderer.SetObjects([]fyne.CanvasObject{r.scroll.Content})
+	}
```
stuartmscott, line 285: This bug fix could be merged separately

widget/table.go
```diff
@@ -0,0 +1,299 @@
...
+	"fyne.io/fyne/theme"
+)
+
+// Table widget is a grid of items that can be scrolled and a cell selected.
+// It's performance is provided by caching cell templates created with NewCell and re-using them with UpdateCell.
+// The size of the content rows/columns is returned by the DataSize callback.
+type Table struct {
+	BaseWidget
+
+	DataSize       func() (int, int)
```
stuartmscott, line 18: It maybe longer, but `Dimensions` might be a better name

widget/table.go
```diff
@@ -0,0 +1,299 @@
...
+	NewCell        func() fyne.CanvasObject
+	UpdateCell     func(row int, col int, template fyne.CanvasObject)
+	OnCellSelected func(row int, col int)
+
+	cells                       *tableCells
+	scroll                      *ScrollContainer
+	SelectedRow, SelectedColumn int
```
stuartmscott, line 25: Could the exported fields be moved above the unexported ones?

widget/table.go
```diff
@@ -0,0 +1,299 @@
...
+
+func (r *tableCellsRenderer) Objects() []fyne.CanvasObject {
+	return r.objects
+}
+
+func (r *tableCellsRenderer) Destroy() {
+	r.clearPool()
+}
+
+func (r *tableCellsRenderer) clearPool() {
```
stuartmscott, line 273: clearPool is a misleading name - it is not clearing the pool, rather it is releasing all cells into the pool

widget/table.go
```diff
@@ -0,0 +1,299 @@
...
+type Table struct {
+	BaseWidget
+
+	DataSize       func() (int, int)
+	NewCell        func() fyne.CanvasObject
+	UpdateCell     func(row int, col int, template fyne.CanvasObject)
+	OnCellSelected func(row int, col int)
+
+	cells                       *tableCells
+	scroll                      *ScrollContainer
```
stuartmscott, line 24: Neither List nor Tree keeps a reference to the scroller, only the renderer does. Why is this needed? Most of the usages are `t.t.scroll`

widget/table.go
```diff
@@ -0,0 +1,299 @@
...
+	t *Table
+
+	rowMarker, colMarker *canvas.Rectangle
+	dividers             []fyne.CanvasObject
+
+	objects  []fyne.CanvasObject
+	cellSize fyne.Size
+}
+
+func (t *tableRenderer) moveOverlay() {
```
stuartmscott, line 69: Is this the best name? Overlay already has a specific meaning in Fyne

widget/table_test.go
```diff
@@ -0,0 +1,62 @@
+package widget
+
+import (
+	"testing"
+
+	"fyne.io/fyne"
+	"fyne.io/fyne/test"
+	"fyne.io/fyne/theme"
+	"github.com/stretchr/testify/assert"
+)
```
stuartmscott, line 10: Add a test to ensure `&Table{}` usage doesn't crash

widget/table.go
```diff
@@ -0,0 +1,299 @@
...
+			t.rowMarker.Show()
+
+			top := fyne.Max(theme.Padding(), y1)
+			t.rowMarker.Move(fyne.NewPos(0, top))
+			t.rowMarker.Resize(fyne.NewSize(theme.Padding(), fyne.Min(y2, t.t.size.Height)-top))
+		}
+	}
+	t.rowMarker.Refresh()
+
+	colDivs := int(math.Ceil(float64(t.t.size.Width+1) / float64(t.cellSize.Width+1)))   // +1 for div width
```
stuartmscott, line 106: Extracting `1` to `const tableDividerThickness = 1` would negate the need for these comments, while improving readability

stuartmscott (changes requested):

widget/scroller.go
andydotxyz: It was, as 2ccac85f6, this is just due to the base being old I think?

widget/table.go
andydotxyz: It's used to ask the overlay to update when the content of the scroller is tapped.
It's an indirection to avoid the cells trying to get a reference to the render of the parent component.
I can write the indirection in through another means.

widget/table.go
andydotxyz: -> updateMarkers

stuartmscott: s/Markers/Indicators - consistent with naming used in List/Tree

andydotxyz: But it is indicators plus dividers

widget/table.go
andydotxyz: Yes. Though worth noting that ordering of fields in struct is not something we have standardised.

widget/table.go
stuartmscott: The functionality for ensuring the tapped cell is visible in the scroller is a nice-to-have, but not something List or Tree do. You could drop this to simplify the code for this first attempt

andydotxyz: It's not related to is the cell visible - it's related to how do we ensure the indicator updates when it's tapped

widget/table.go
```diff
@@ -0,0 +1,301 @@
...
+
+			left := fyne.Max(theme.Padding(), x1)
+			t.colMarker.Move(fyne.NewPos(left, 0))
+			t.colMarker.Resize(fyne.NewSize(fyne.Min(x2, t.t.size.Width)-left, theme.Padding()))
+		}
+	}
+	t.colMarker.Refresh()
+
+	if t.t.SelectedRow == -1 {
+		t.colMarker.Hide()
```
stuartmscott, line 91: col by itself makes me wonder if row was accidentally missed

widget/table.go
```diff
@@ -0,0 +1,301 @@
...
+func (t *tableRenderer) BackgroundColor() color.Color {
+	return theme.BackgroundColor()
+}
+
+func (t *tableRenderer) Objects() []fyne.CanvasObject {
+	return t.objects
+}
+
+func (t *tableRenderer) Destroy() {
+}
```
stuartmscott, line 190: Could baseRenderer be embedded to avoid needing these? (would need to call SetObjects rather than objects field)

widget/table.go
```diff
@@ -0,0 +1,301 @@
...
+
+	scroll               *ScrollContainer
+	rowMarker, colMarker *canvas.Rectangle
+	dividers             []fyne.CanvasObject
+
+	objects  []fyne.CanvasObject
+	cellSize fyne.Size
+}
+
+func (t *tableRenderer) moveIndicators() {
```
stuartmscott, line 71: nit: #827: this should come after Refresh

widget/table.go
```diff
@@ -0,0 +1,301 @@
...
+	if t.t.SelectedRow == -1 {
+		t.colMarker.Hide()
+	} else {
+		offY := t.t.SelectedRow*(t.cellSize.Height+1) - t.scroll.Offset.Y
+		y1 := theme.Padding() + offY
+		y2 := y1 + t.cellSize.Height
+		if y2 < theme.Padding() || y1 > t.t.size.Height {
+			t.rowMarker.Hide()
+		} else {
+			t.rowMarker.Show()
```
stuartmscott, line 99: Not sure if it makes a difference, but seems like Show should be called after Resize, wdyt?

widget/table.go
```diff
@@ -0,0 +1,301 @@
...
+
+import (
+	"image/color"
+	"math"
+
+	"fyne.io/fyne"
+	"fyne.io/fyne/canvas"
+	"fyne.io/fyne/theme"
+)
+
```
stuartmscott, line 11: Pleas add conformity assertions to the Widgets and WidgetRenderers in here

widget/table.go
```diff
@@ -0,0 +1,301 @@
...
+	r.Layout(t.Size())
+	return r
+}
+
+type tableRenderer struct {
+	t *Table
+
+	scroll               *ScrollContainer
+	rowMarker, colMarker *canvas.Rectangle
+	dividers             []fyne.CanvasObject
```
stuartmscott, line 65: Is there any advantage to splitting dividers into horizontal and vertical so the driver doesn't have to re-draw if the size is the same?

stuartmscott (changes requested):

widget/table.go
andydotxyz: That is a good point. But would that imply that List should have been Dimension?

I was thinking about the semantics and it's tough, does Dimensions (plural of dimension) refer to many different dimensions, whereas this is sort of 1 dimension that is two dimensional ;).

widget/table.go
andydotxyz: Resolved I think

widget/table.go
andydotxyz: It's still in - I don't think the code is too hard to follow, but will leave it to others to determine.

widget/table.go
andydotxyz: I used moveIndicators as I didn't really mind :)

widget/table.go
stuartmscott: Hmmm, interesting. To flip it around, if this is DataSize, then should list also be DataSize?

I think it would be okay to have List.Length and Table.Dimensions, but I also like the consistency of List.Dimension and Table.Dimensions, @okratitan wdyt?

widget/table.go
stuartmscott: Yup, something we should add to the consistency wiki

widget/table.go
andydotxyz: Good catch thanks :)

widget/table.go
andydotxyz: I don't think so - a simple filled rectangle is essentially a 1x1px uniform that is drawn at whatever size inside the GL code

widget/table.go
```diff
@@ -0,0 +1,328 @@
...
+type tableCellsRenderer struct {
+	cells   *tableCells
+	pool    pool
+	objects []fyne.CanvasObject
+	visible map[cellID]fyne.CanvasObject
+}
+
+func (r *tableCellsRenderer) Layout(_ fyne.Size) {
+	// we deal with cached objects so just refresh instead
+}
```
stuartmscott, line 248: nit: could the layout code be moved here, and have Refresh call Layout?

widget/table.go
```diff
@@ -0,0 +1,328 @@
...
+	offX := r.cells.t.offset.X - (r.cells.t.offset.X % (r.cells.cellSize.Width + tableDividerThickness))
+	minCol := offX / (r.cells.cellSize.Width + tableDividerThickness)
+	offY := r.cells.t.offset.Y - (r.cells.t.offset.Y % (r.cells.cellSize.Height + tableDividerThickness))
+	minRow := offY / (r.cells.cellSize.Height + tableDividerThickness)
+
+	wasVisible := r.visible
+	r.visible = make(map[cellID]fyne.CanvasObject)
+	var cells []fyne.CanvasObject
+	for y := minRow; y < minRow+rows; y++ {
+		for x := minCol; x < minCol+cols; x++ {
```
stuartmscott, line 273: The minXYZ variables ensure top and leading cells outside the viewport are not rendered, but shouldn't there also be maxXYZ variables to ensure bottom and trailing cells outside the viewport are not rendered?

widget/table.go
```diff
@@ -0,0 +1,328 @@
...
+			cells = append(cells, c)
+		}
+	}
+
+	for id, old := range wasVisible {
+		if _, ok := r.visible[id]; !ok {
+			r.pool.Release(old)
+		}
+	}
+	r.objects = cells
```
stuartmscott, line 299: A nice sanity check just while this PR is in development would be add `log.Println(len(cells))` here to ensure only the minimum cells are rendered

stuartmscott (changes requested):

widget/table.go
andydotxyz: Not sure there is any point, maxCol would just be minCol + cols (as rows, cols are assigned above as the visible count)

stuartmscott: Oh nevermind - visibleCount is handling this

widget/table.go
andydotxyz: That would be slower as we would have to iterate all the cells twice.

widget/table.go
andydotxyz: can't hurt :)

widget/table.go
stuartmscott: Yes I see you added those

widget/table.go
```diff
@@ -0,0 +1,365 @@
...
+		t.dividers[divs].Show()
+		divs++
+	}
+
+	for i := divs; i < len(t.dividers); i++ {
+		t.dividers[divs].Hide()
+	}
+}
+
+// Declare conformity with Widget interface.
```
stuartmscott, line 216: Also Tappable

widget/table.go
andydotxyz: Isn't that better for static objects renderer? I figured this was doing a lot of object manipulation.

stuartmscott: `objects` is still a field, but it would be inside the embedded type, and allow 3 methods to be dropped from here. The only difference in the rest of the code would be replacing `r.objects = ...` with `r.SetObjects(...)`

widget/table.go
okratitan: I think Dimension implies spatial undertones and for list that is irrelevant.  When I hear Dimension I think "Okay so with dimension I must be defining if this list should be vertical or horizontal" more so than I think about it as "How many" which it is  For table using dimensions makes sense because you are defining "10 columns" by "10 rows" which is spatial and not the how many... "How many" is 100.  I do think dimensions sounds better than data size for list.

widget/table.go
okratitan: If the OnTapped is inside of the renderer like it is in list then I don't think this is a problem and the reference to the scroll container could be removed as Stuart suggests.  https://github.com/fyne-io/fyne/blob/a022e44470ad884476176dd7a423e9bd66e78281/widget/list.go#L261

andydotxyz: I guess I wanted to avoid creating a new widget type that wrapped all the cells - there can be many hundreds of them visible at any time. So a single Tapped handler in the parent made more sense to me...

andydotxyz: Perhaps someone can remind me why the line in question is a problem?

widget/table.go
andydotxyz: I don't know if Dimension is clearly separated from the concept of widget size enough - it could be deemed ambiguous.
Length for List works in this way, but can we name this in a way that does not suffer that potential confusion?

charlesdaniels: I do agree that it makes sense for both list and table to use similar terminology.

I think that Size() makes more sense than dimension, and would also apply to list as well. 

Could also have two separate calls, a `NumberOfRows()`, and `NumberOfCols()` (or just Rows and Cols). I think `Rows()` also makes sense in the context of list. That said, the short versions (Rows and Cols) may be confused with a method to actually retrieve the data of the rows/cols, rather than the number of rows/cols.

widget/table.go
stuartmscott: Not a problem per se, just uncommon for widgets to hold references to subcomponents - mostly held by widgetrenderers. This isn't a blocker for me, just wanted to explore alternatives

andydotxyz: I guess. I think that popups and scrollers are special cases as they often need to be manipulated when the parent object changes. See Select/SelectEntry.

stuartmscott (approved):

okratitan (approved):

PR #1345, opened 2020-09-27
Change Select so that it doesn't expand to fit

### LINKED ISSUE #1247: select widget grows the enclosing window too wide
### Describe the bug:
I have several select widgets that get options added dduring runtime.  Adding long options to these select widgets causes the width of the enclosing window to grow without bounds. In select.go, only using the PlaceHolder size and not including the maximum option size makes the button stay at it's original size and the display box expands into the window below to show all the options .

### To Reproduce:
create a select widget and add longer and longer options to the  Options [] string.
open the select box after each add.
the enclosing window grow, and grows,....

### Screenshots:
this image if after a homemade patch (see below) to select.go to keep the MinSize at the size of the PlaceHolder:

image

### Example code:
func (s *selectRenderer) MinSize() fyne.Size {
	s.combo.propertyLock.RLock()
	defer s.combo.propertyLock.RUnlock()
	min := fyne.MeasureText(s.combo.PlaceHolder, s.label.TextSize, s.label.TextStyle)
	//	for _, option := range s.combo.Options {
	//		optionMin := fyne.MeasureText(option, s.label.TextSize, s.label.TextStyle)
	//		min = min.Union(optionMin)
	//	}
	min = min.Add(fyne.NewSize(theme.Padding()*4, theme.Padding()*2))
	return min.Add(fyne.NewSize(theme.IconInlineSize()+theme.Padding(), 0))

 - OS: Windows
 - Version: 10
 - Go version:  go1.14.4 windows/amd64
 - Fyne version: 1.3

andydotxyz:
This improves select so it can have long content without expanding the app :).
Test images change as we moved to label to do this - but looking at them the alignment is better with this change!

Biggest change is that the drop-down can be larger than button.
Also button will not grow to fit selected item.

Fixes #1247

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (approved):
Looks good to me 👍

PR #1346, opened 2020-09-28
Simplify validator so we can pass around just function pointers

andydotxyz:
### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/validation/regexp.go
+++ b/data/validation/regexp.go
@@ -8,31 +8,20 @@ import (
 	"fyne.io/fyne"
 )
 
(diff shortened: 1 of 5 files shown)


Jacalz (changes requested):
Sorry, but my validator fix means that this has to be rebased on develop.
Thanks for the simplification. Looks good to me apart from that rebase needing to take place :)

andydotxyz:
Thanks, that should be merged and resolved now

Jacalz (approved):
Thanks. Looks good to me 👍

PR #1347, opened 2020-09-28
Add caching to widget.Icon

Jacalz:
### Description:
This commit adds some simple caching to widget.Icon.
Testing seems to indicate that the calls to updateObjects in fyne_demo shrunk to less than half of what it was before.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/icon.go
+++ b/widget/icon.go
@@ -32,9 +32,12 @@ func (i *iconRenderer) BackgroundColor() color.Color {
 }
 
 func (i *iconRenderer) Refresh() {
-	i.image.propertyLock.RLock()
-	i.updateObjects()
-	i.image.propertyLock.RUnlock()
+	if i.image.Resource != i.image.cachedRes {
+		i.image.propertyLock.RLock()
+		i.updateObjects()
+		i.image.propertyLock.RUnlock()
+	}
+
 	i.Layout(i.image.Size())
 	canvas.Refresh(i.image.super())
 }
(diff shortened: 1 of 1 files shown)


andydotxyz:
This looks cunning. Does it work across theme changes as well?
(In that case the resource is still the same ID but it has different content)

widget/icon.go
```diff
@@ -54,12 +57,14 @@ func (i *iconRenderer) updateObjects() {
...
 
-	Resource fyne.Resource // The resource for this icon
+	Resource  fyne.Resource // The resource for this icon
+	cachedRes fyne.Resource
 }
 
 // SetResource updates the resource rendered in this icon widget
 func (i *Icon) SetResource(res fyne.Resource) {
 	i.Resource = res
+	i.cachedRes = res
```
andydotxyz, line 67: This looks wrong - doesn't it mean that changing the resource will not bother resetting objects because the cache will be deemed the same as the resource...? Surely setting it to nil would be right?

andydotxyz (changes requested):
Pretty sure that setting the resource on an active icon will not result in the cache being reset...

widget/icon.go
Jacalz: Good catch. Fixed that :)

andydotxyz (approved):

PR #1349, opened 2020-09-29
Clarify some Focus() behaviour

### LINKED ISSUE #1329: Focused widget in second window is not actually focused
### Describe the bug:
Launching a new window in 1.3.3 makes the entry in it focus and that is great. However, with develop branch right now it launches the window and the entry looks like it is focused, but actually isn't and doesn't care about taking any inputs from the keyboard. This is very unclear and makes the user think that something isn't working as it should. Just clicking in the entry makes it actually focused.

### To Reproduce:
1. Open a new window with an entry in at the top.
2. Try to type without doing anything.
3. Notice that nothing is typed into the entry, even though it looks to be focused.

### Screenshots:
Hard to get a good gif of this because changing over to the recorder makes it unfocus the window. Both windows that are launched are in focus, just that in 1.3.3 the entry take input and on develop it doesn't.

### Example code:
Was using the sending of text in [wormhole-gui](github.com/Jacalz/wormhole-gui) to replicate this. Building release v1.3.0 with fyne from develop and fyne v1.3.3 should show the differences. Just send some text and it will be apparent :)

 - OS: Linux
 - Version: 5.6.19
 - Go version: 1.15.2
 - Fyne version: https://github.com/fyne-io/fyne/commit/0a2fa00d4f280eb15a2681932b7733381fa8c83e

Jacalz:
### Description:
The issue was that item was focused before the content and that was apparently undefined behaviour that just happened to work with the earlier focusing behaviour. This updates with some comments to signal how it behaves.

Fixes #1329 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


canvas.go
```diff
@@ -7,7 +7,11 @@ import "image"
 type Canvas interface {
 	Content() CanvasObject
 	SetContent(CanvasObject)
+	
 	Refresh(CanvasObject)
+	
+	// Focus makes the provided item focused
```
toaster, line 13:
```suggestion
	// Focus makes the provided item focused.
```

canvas.go
```diff
@@ -7,7 +7,11 @@ import "image"
...
+	// Must be called after the item has been added to the contents of the canvas.
```
toaster, line 14: The “must” looks like I have to call this method. What about:
```suggestion
	// The item has to be added to the contents of the canvas before calling this.
```

toaster (approved):

andydotxyz (approved):

PR #1350, opened 2020-09-30
Add SetValue() to slider

Jacalz:
### Description:
This adds a SetValue() function to be consistent with other widgets.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
This is skipping the range clamping which seems risky. And you’re setting the value after calling OnChanged which is not what others to iirc...

Jacalz:
Seems that radio doesn’t, but I can get that changed if you would like. Should be fixed now :)

andydotxyz:
> Seems that radio doesn’t

The other trigger within Slider might have been a better comparison ;)

But yes it looks like radio might be wrong - the "OnXxx" callbacks signal that something has already happened.

andydotxyz (approved):

Jacalz:
I was the one to add support for that to Radio, so I can claim that as my fault ;)
Will get that fixed :)

PR #1351, opened 2020-10-01
Correct paths for favorites in file dialog

### LINKED ISSUE #1248: Problem with buttons of Save Dialog window on non English Ubuntu
### Describe the bug:
The save file dialog show me the expected buttons (Home, Documents, Downloads, Computer). But, the Documents when clicked cause these errors:

`2020/08/12 08:14:19 Fyne error:  Unable to read path /home/rodrigo/Documents
2020/08/12 08:14:19   Cause: open /home/rodrigo/Documents: no such file or directory
2020/08/12 08:14:19   At: /home/rodrigo/go/pkg/mod/fyne.io/fyne@v1.3.3/dialog/file.go:183`

The cause, I believe, is that my PC uses another language than English (Brazilian Portuguese), and my folder is called **Documentos**. 
Downloads and Home folders are not translated, so unaffected. Computer points to /, also unaffected.

### To Reproduce:
1. Run the app on a non English Ubuntu PC (where Documents folder has its name translated)
2. Save a file using the save dialog window
3. Click the Documents button.
4. See an empty folder and the error above on terminal.

### Device:
 - OS: Ubuntu
 - Version: 20.04
 - Go version: 1.14.7
 - Fyne version: 1.3.3

P.s.: On Windows 10 I've noticed that if Downloads folder has its path changed, the **Downloads** button also fails to reach the new location determined within the folder properties (It may happen to other system folders with altered paths). But I can't give more details, I rarely use Windows...

PucklaMotzer09:
### Description:
Uses correct paths for favorite locations for different languages on Linux.
It uses the ```xdg-user-dir``` command to determine the location.

Fixes #1248

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


PucklaMotzer09:
I want to add a test for the favorites, but they are inside a ```widget.Group```. This struct does not export their children, so to test the favorites correctly, the children need to somehow be exported.

PucklaMotzer09:
I added a test by just using the slice returned by ```loadFavorites```.

andydotxyz:
Thanks for this. We would need a different approach for darwin as it also is not XDG.
More problematic is that my linux computer does not include the xdg-user-dir command (arch linux).
I read that it functions by finding a user-dirs.dirs file, but that does not exist either.

How universal is this feature? If it is fairly common then maybe we can go forward but with sensible fallbacks for other Unix systems?

PucklaMotzer09:
> This will work with any linux distribution that has a DE which follow the freedesktop specification, which means: KDE, Gnome, XFCE and many more(exactly like the solution already posted by the OP).

Accordording to the [first answer](https://stackoverflow.com/questions/13741873/how-to-get-users-desktop-path-in-python-independent-of-language-install-linux).

> sensible fallbacks for other Unix systems

If the command fails an error is returned and the default one (eg. $HOME/Documents) is used.

> We would need a different approach for darwin as it also is not XDG

According to [this post](https://stackoverflow.com/questions/57002295/how-to-find-mac-documents-path-language-independent) these folders are handled similiar to Windows and just show the folder in a different language, but on the file system it is language independent.

andydotxyz:
Apologies I misread the code when it comes to the actual lookup.
When it's chunks of code that are replaced based on OS we usually use different files for each variant (the app package has a better example) it's a little easier to read when the linux specific code is in a linux file, though in this case it's XDG rather than linux specifically...

It is unusual for DEs to install these utility executables on some systems - you can see in Arch that this is not a requirement of any of the desktops you listed: https://www.archlinux.org/packages/extra/x86_64/xdg-user-dirs/

andydotxyz:
A more robust approach might be to use exec.LookPath to see if xdg-user-dirs is installed. If not then printing lots of errors is not really required.

PucklaMotzer09:
> When it's chunks of code that are replaced based on OS we usually use different files for each variant

I actually wanted to do that, but I would have had to copy a lot of functions from ```file_other.go``` to ```file_linux.go```.

andydotxyz:
> but I would have had to copy a lot of functions

I get your point - but there were basically 15 lines of code in 4 small functions.
That file is now 45 linux specific lines to avoid the small duplication...

PucklaMotzer09:
You got a point. I'll figure something out.

andydotxyz (changes requested):
getFavoriteLocations is not defined for non-linux unix systems like BSD.
You might see that in the `app` package we use the _xdg extension and use the build rule:

```
// +build linux openbsd freebsd netbsd
// +build !android
```

alternatively a file_bsd.go would work, if you think the behavior for BSD and Linux is different.

andydotxyz:
Thanks. I just checked on BSD (Not on the travis run) and it's happy now :)

andydotxyz (approved):

andydotxyz:
Please don't force push after a review, it results in the following un-usable change set:

Screenshot 2020-10-06 at 12 05 08

And so PRs must be re-reviewed in full...

dialog/file_test.go
```diff
@@ -282,3 +283,29 @@ func TestFileFilters(t *testing.T) {
...
+	favorites := dlg.dialog.loadFavorites()
+	places := dlg.dialog.loadPlaces()
+	assert.Equal(t, 3+len(places), len(favorites))
+
+	curDir := ""
+	for _, f := range favorites {
+		btn := f.(*widget.Button)
+		t.Log(fmt.Sprint("Changing from ", curDir, " to ", btn.Text))
+		test.Tap(btn)
+		assert.NotEqual(t, curDir, dlg.dialog.dir)
```
andydotxyz, line 306: Would it not be better to assert that the correct directory is now set?

dialog/file_test.go
```diff
@@ -282,3 +283,29 @@ func TestFileFilters(t *testing.T) {
...
+	dlg.Show()
+
+	favorites := dlg.dialog.loadFavorites()
+	places := dlg.dialog.loadPlaces()
+	assert.Equal(t, 3+len(places), len(favorites))
+
+	curDir := ""
+	for _, f := range favorites {
+		btn := f.(*widget.Button)
+		t.Log(fmt.Sprint("Changing from ", curDir, " to ", btn.Text))
```
andydotxyz, line 304: We don't normally print debug in tests - only relevant info is when a test fails

andydotxyz (changes requested):

dialog/file_test.go
PucklaMotzer09: I don't know the exact name of the directory since it depends on the username and language.

dialog/file_test.go
PucklaMotzer09: I think this info is relevant to know why the test fails. There can be other reasons for why the test fails because I don't check against the exact directory name.

dialog/file_test.go
andydotxyz: I would have thought it existed in the map returned by favouriteLocations()?
I understand that it might not be a trivial lookup/comparison - but by not doing it you run the risk that this is simply setting the un-localised path which would be a failure and yet this test would pass...

dialog/file_test.go
andydotxyz: I suppose my comment only makes sense in combination with the one below.

dialog/file_test.go
PucklaMotzer09: Yes, the paths are inside the map, but using those would be like checking strings against themselves. The unlocalised default paths are used if the xdg-user-dir command doesn't exist or if some other error occurs. The default paths are also used if the xdg-user-dir command returns the home directory. This happens when eg. the documents folder doesn't exist. This is the case for the CI tests.

andydotxyz: But this test is testing what happens when you tap the button.
To check that the favourites are created correctly is the remit of another test (one that does not create a file dialog UII).
Testing the dialog's new location against the string representing the location that should have been loaded when a button was pressed seems quite correct to me.
If your concern is that the localisation is not tested if we do that then yes another test can be added. But by ignoring the contents of the string I don't see that anything is being added.

PucklaMotzer09: I changed it so that ```dlg.dialog.dir``` is checked against ```getFavoriteLocations```.

dialog/file_test.go
PucklaMotzer09: I removed the log

dialog/file_test.go
PucklaMotzer09: > To check that the favourites are created correctly is the remit of another test

What test is this exactly? I couldn't find a test which is testing the favorites.

andydotxyz: I did not mean that it existed, I was more hinting that tests should test one thing, so maybe a test is missing?

andydotxyz:
Thanks for this update, a positive tests seems more complete.
Also good catch on the drive buttons, I had forgotten about them!

dialog/file_test.go
andydotxyz: I think that this would be wrapped up nicely if we could add a test of the file_xdg contents - do you think that is possible? (clearly they would only run on linux :) )

PucklaMotzer09: I did my best, but it is difficult to test this since the paths depend on the languange and sometimes don't even exist.

dialog/file_xdg_test.go
```diff
@@ -0,0 +1,44 @@
...
+	assert.Nil(t, err)
+
+	for e := range expected {
+		fav, ok := favoriteLocations[e]
+		if assert.True(t, ok) {
+			if expected[e] == "" {
+				assert.Equal(t, homeDir, fav)
+			} else {
+				assert.Equal(t, homeDir, filepath.Dir(fav))
+				if _, err := os.Stat(fav); os.IsNotExist(err) {
```
andydotxyz, line 36: I don't understand this test sorry.
It seems to iterate expected directories and confirms that they end with the expected string.
But the XDG code exists to make this configurable - so as soon as someone has that set these will fail won't they?

The purpose of testing the favourite location lookup would be to test the wrapper around xdg-user-dir I would think, which is based on environment variables, so can we manipualte them to test the code is working?

andydotxyz (changes requested):
I might have misunderstood the new tests, but hopefully my comments help

andydotxyz:
I probably should have said if the XDG tests are proving difficult then don't worry, I can merge this in and complete it myself.
Platform specifics can be a pain.

dialog/file_xdg_test.go
PucklaMotzer09: > The way it works is that xdg-user-dirs-update is run very early in the login phase. This program reads a configuration file, and a set of default directories. It then creates localized versions of these directories in the users home directory and sets up a config file in $(XDG_CONFIG_HOME)/user-dirs.dirs (XDG_CONFIG_HOME defaults to ~/.config) that applications can read to find these directories. 

This is from [here](https://freedesktop.org/wiki/Software/xdg-user-dirs/).

So it's not really about environment variables, but more about a config file. I think this is too "hacky" for me so I will leave it at that.

> so as soon as someone has that set these will fail won't they?

The paths are only getting checked against those strings when the folders don't exist. So this test is not testing wether ```getFavoriteLocations``` returns the correct paths. It is only checking if the fallback paths are correct.

andydotxyz: Ah I understand thanks. I was struggling to follow the code - I played with it a bit, maybe some of these changes can help? Let me know what you think.

```go
func TestFavoriteLocations(t *testing.T) {
	favoriteLocations, _ := getFavoriteLocations()

	// This is a map containing the base name
	// of every key that should be in favoriteLocations
	expected := map[string]string{
		"Home":      "",
		"Documents": "Documents",
		"Downloads": "Downloads",
	}

	homeDir, err := os.UserHomeDir()
	assert.Nil(t, err)

	for name, subdir := range expected {
		fav, ok := favoriteLocations[name]
		if !ok {
			t.Errorf("missing favourite location", name)
			continue
		}

		if subdir == "" {
			assert.Equal(t, homeDir, fav)
		} else {
//			assert.Equal(t, homeDir, filepath.Dir(fav)) // I don't think this assertion must be true, the user could specify a different parent directory
			_, err := os.Stat(fav)
			if err == nil {
				continue // folder found
			}
			if !os.IsNotExist(err) {
				t.Errorf("failed to read directory %s", fav)
				continue
			}
			fallbackName := filepath.Base(fav)
			if !strings.EqualFold(subdir, fallbackName) {
				t.Errorf("%s should equal fold with %s", subdir, fallbackName)
			}
		}
	}
}
```

PucklaMotzer09: So the only thing that's different here is that
```go
assert.Equal(t, homeDir, filepath.Dir(fav))
```
has been removed. Right?

PucklaMotzer09:
I rewrote most of the code because of #1379. I used ```fyne.ListableURI``` everywhere where it should be used. But this needs a rebase and a force push of course. So just tell me when you are ready for that.

andydotxyz:
Hi. In general we would prefer a merge into the PR with the conflict resolution commit.
However in this instance I can see that the majority of the code has changed, so go ahead.

andydotxyz:
All of the Linux builds are failing so I don't think it's random error @PucklaMotzer09

PucklaMotzer09:
Now that I use ```fyne.ListableURI``` every favorite location has to exist. But the folders can be missing on some systems, so I can't really test them now.

andydotxyz:
Looking at the latest change I see that you've removed an error when the favourite does not exist but also checking that an existing favourite cannnot be loaded.
It seems strange that both would be removed, consider the user experience - the left-panel should show those locations that we have found I think? So that list works from the favourite locations right? Then the end of the test that tries to find a favourite location should surely still pass, because it being in the list means that we found it?

(Apologies if that is not a correct representation of the current code, I was thinking out loud how it maybe should work).
What do you think?

PucklaMotzer09:
Every favorite location that has been found and exists will be added to the list. Every path that doesn't exist does not appear on the left panel. This happens because ```getFavoriteLocations``` now returns ```fyne.ListableURI``` and you can only create a ListableURI from an URI if the path exists. I think this behaviour is reasonable, but this makes it difficult to test since every system can have different paths or no paths at all.

andydotxyz:
I think that sounds great, but if that is the case then why did the second part of the test need to be removed?
If we are only checking against existing favourite locations then we should be able to test for them...

andydotxyz:
I think maybe we are not on the same page. I was asking that we re-instate some of the tests that should still pass, but now you have removed more?

PucklaMotzer09:
If a path doesn't exist then it doesn't get added to the ```favoriteLocations``` map. If I would add it to the map than that value would be nil, since I can't create a ListableURI from a non existing path. So the only favorite location we actually can test is ```Home```. There is no need to check if the paths exist since we already know that they do.

PucklaMotzer09:
> but now you have removed more?

I removed them because they were meaningless.

PucklaMotzer09:
Yes we should think about how to test them properly. The tests just didn't check something meaningful as they were.

andydotxyz:
I thought the tests were confirming that the custom directory locations did in fact exist.
I appreciate that this may not have been meaningful on Travis, but surely they deserved to remain for the plethora of Linux devices that this is tested on during development?

PucklaMotzer09:
Yes you are right. I forgot about the other platforms and the drives. I added a check for the existance of the directories. But ```TestFavoriteLocations``` is pretty meaningless right now.

andydotxyz (approved):

PR #1352, opened 2020-10-01
Fix calling Resize before Show in file dialog

### LINKED ISSUE #1254: File Dialog Resize invalid memory address or nil pointer dereference
### Describe the bug:
If you try to resize a file dialog on the develop branch you get this error:

[signal 0xc0000005 code=0x0 addr=0x40 pc=0x1252b11]

fyne.io/fyne/dialog.(*FileDialog).Resize(0xc000200180, 0x1f4, 0x1f4)
        D:/Programmieren Server/Go/src/fyne.io/fyne/dialog/file.go:321 +0x31
main.main()
        C:/Users/Max/go/src/github.com/Cookie04DE/pqsg/cmd/Mahnungen/main.go:31 +0x1d2
exit status 2
### To Reproduce:
1. Create a window
2. Create new file open dialog
3. Call resize on the dialog
### Example code:
```go
package main

import (
	"fmt"
	"io/ioutil"

	"fyne.io/fyne/storage"

	"fyne.io/fyne/dialog"

	"fyne.io/fyne"

	"fyne.io/fyne/app"
)

var a fyne.App

func main() {
	a = app.New()
	w := a.NewWindow("Choose WinWord.exe")
	winWordOpen := dialog.NewFileOpen(func(readCloser fyne.URIReadCloser, err error) {
		if err != nil {
			Error(err)
			return
		}
		fmt.Println(readCloser.Name())
		readCloser.Close()
	}, w)
	winWordOpen.SetFilter(storage.NewExtensionFileFilter([]string{".exe"}))
	w.SetFixedSize(true)
	w.Resize(fyne.NewSize(500, 500))
	winWordOpen.Resize(fyne.NewSize(500, 500))
	winWordOpen.Show()
	w.ShowAndRun()
}

//Error shows a new window containing the error message
func Error(err error) {
	w := a.NewWindow("")
	w.Resize(fyne.NewSize(100, 100))
	dialog.ShowError(err, w)
	w.Show()
}
```

### Device:
 - OS: Windows 10
 - Go version: 1.15
 - Fyne version: 1.3.3

PucklaMotzer09:
### Description:
Stores size in ```desiredSize``` when ```dialog``` is nil and applies it when calling ```Show```

Fixes #1254

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
Thanks for working on this, but instead of printing errors we should actually store the requested resize and apply it later.
In this instance we know what the user wanted so let's solve it rather than push it back to them.

dialog/file.go
```diff
@@ -313,12 +313,17 @@ func (f *FileDialog) Show() {
...
+		f.Resize(*f.desiredSize)
+		f.desiredSize = nil
+	}
 }
 
 // Resize dialog, call this function after dialog show
 func (f *FileDialog) Resize(size fyne.Size) {
 	if f.dialog == nil {
-		fyne.LogError("FileDialog.Resize", errors.New("is called before Show"))
+		f.desiredSize = new(fyne.Size)
```
andydotxyz, line 325: would it not be simpler to write `f.desiredSize = &size`?

andydotxyz (changes requested):
Much better thanks. Just one question as it looks more complex than should be needed

andydotxyz:
Please don't force-push once a review has been added. It makes it harder to follow on from previous review.
If history really needs to be squashed it can be done so on merge.

Great work though, thanks.

andydotxyz (approved):

Issue #505, opened 2019-10-28
Menu bar initially doesn't respond to mouse input on macOS


richardwilkes:
**Describe the bug**
Playing with the demo (commit 98fc45eeb2f12a3d3d0ad62f485883130e292312) on macOS (catalina), the menu bar won’t respond to mouse clicks when launching it from the command line (admittedly not normal for a macOS app) until you switch to another app and come back.

**To Reproduce**
1. cd cmd/fyne_demo
2. go run main.go
3. Try to click on the menu entries the demo app put up... none respond
4. Switch to another app
5. Switch back to the demo app
6. Menu now responds to mouse input

**Screenshots**

**Device (please complete the following information):**
 - OS: macOS
 - Version  10.15 (Catalina)
 - Go version 1.13.3
 - Fyne version 98fc45eeb2f12a3d3d0ad62f485883130e292312

toaster:
Hi, I'm on Mojave here and cannot reproduce this. Can anyone confirm the steps to reproduce this issue? And if so, on which OS version?

andydotxyz:
It only happens on Catalina. If you open a Fyne app it’s menu does not respond. Switch away and back again and it works ok.

andydotxyz:
I have since found out that if you package the app up the problem does not occur. So this is really only an issue for us developers. I may consider removing "Blocker" now...

richardwilkes:
Yes. I did mention it is only an issue when the executable is launched directly from the command line (i.e. not via the UI nor via the use of `open`, which ultimately behaves as if you opened the app from the UI). This is true whether packaged or not -- but I agree that no one should consider this a blocker.

richardwilkes:
I've found a workaround that you can use to fix this problem. First, you need a way to detect you're being launched without a bundle. The easiest way I could think of to do this is to determine if `os.stdout` was connected to a terminal. If it is, then issue a hide followed immediately by an unhide.

```
var termios syscall.Termios
if _, _, errno := syscall.Syscall6(syscall.SYS_IOCTL, os.Stdout.Fd(), ioctlReadTermios, uintptr(unsafe.Pointer(&termios)), 0, 0, 0); errno == 0 {
    hideApp() // needs to do this in native code: [[NSRunningApplication currentApplication] hide];
    unhideApp() // needs to do this in native code: [[NSRunningApplication currentApplication] unhide];
}

```
I do this right before entering the run loop.

andydotxyz:
[...]

PR #1353, opened 2020-10-01
Hide console on Windows

gonutz:
When not adding -ldflags "-H=windowsgui" to go build/run/install on Windows, a console window shows up when you run the program.

This PR hides the console in an init function. The console window might be visible for a split second but then disappears. This is nice for debugging and might even be used for deployment.

There are some considerations, though:

1. Do we always want to hide the console? It might be useful for programmers during development. Usually you will work in your IDE which shows you the stdout and stderr output but if you really want to click on your .exe and see output, you can using the default console window. We might add an option to not hide the console on Windows.

2. Is this the right place? I just put the code in an init function because I want it to happen right when the program starts and because I do not know where else to put it. What is the code path that every fyne app takes? waiting until ShowAndRun would be too late, settings up things, maybe pulling a database for example, it might take a while for the program to get to show the GUI. Until then we do not want the console to be visible, possibly printing debug output. As said in 1. this might be good for the developer but if this is deployed, it is bad.

This PR was easy for me to make, I do not expect it to be merged without discussion but I wanted to get something out there that people can try before deciding if we really want this to happen and if yes, how to do it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ It should look like this:
   </table>
 </div>
 
-> Note that Windows applications load from a command prompt by default, which means if you click an icon you may see a command window.
+> Note that Windows applications load from a command prompt by default, which means if you click an icon you may see a command window. Fyne hides this for you at the start of the program but it might still be visible for a split second.
 > To fix this add the parameters `-ldflags -H=windowsgui` to your run or build commands.
 
 # Documentation
--- a/app.go
+++ b/app.go
@@ -5,6 +5,10 @@ import (
 	"sync"
 )
 
+func init() {
+	hideConsoleOnWindows()
+}
+
 // An App is the definition of a graphical application.
 // Apps can have multiple windows, it will exit when the first window to be
 // shown is closed. You can also cause the app to exit by calling Quit().
--- a/hide_console_dummy.go
+++ b/hide_console_dummy.go
@@ -1,0 +1,6 @@
+//+build !windows
+
+package fyne
+
+// hideConsoleOnWindows does nothing on non-Windows operating systems.
+func hideConsoleOnWindows() {}
--- a/hide_console_windows.go
+++ b/hide_console_windows.go
@@ -1,0 +1,63 @@
+//+build windows
+
+package fyne
+
+import (
+	"syscall"
+	"unsafe"
+)
+
+// hideConsoleOnWindows hides the associated console window that gets created
+// for Windows applications that are of type console instead of type GUI. When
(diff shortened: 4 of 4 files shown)


andydotxyz:
What if I am running on a console and I just type "go run ."?

gonutz:
If you start from the console, as well as from an IDE, nothing happens. That is why there is a check for whether the thread is the one who created the console.

gonutz:
I have been using this code for years, it works as you would expect, only closing the console if it was created because Go creates exe type console and not GUI:

https://github.com/gonutz/wui/blob/ac034522c879ac42ca5abeb7aa930a6998340d65/window.go#L953

https://github.com/gonutz/prototype/blob/98c5cc2572eea5cd47936ac8bf00f20746a335d4/draw/window_windows.go#L339

andydotxyz:
But when you package the app for installation the console is not displayed though right? 
"fyne package" or "fyne install" (or fyne-cross builds for that matter) all suppress the console already...

gonutz:
If they specify ldflags -H=windowsgui, the console is suppressed anyway. Windows will not try to create one in the first place, thus nothing happens in that case.

andydotxyz:
Yes, our tools do just that.
I'm just not a fan of the workarounds that flicker windows on to screen (there is a similar workaround to a macOS bug that we won't land because it's just ugly).
How about we just push people to use the tools provided? :)

gonutz:
Sure, this is why I am asking feedback. Personally I do not mind flickering during development. Most of the time it is not even noticable (on my machine at least). This is more to not have that ugly window there during development. For a release product I would always build with -H=windowsgui anyway. This is more a feature for the developer while debugging.

gonutz:
What is your similar thing on OS X?

andydotxyz:
> What is your similar thing on OS X?

#505

andydotxyz:
Perhaps you can expand on your "during development" workflow? If the console is not closed from IDE or command line launches what other development types exist that trigger new terminals being opened?

This was initially an issue when installing apps with this not set, which is why the package tool adds it.

gonutz:
Oh well, you are right. It turned up for me when `go build`ing your demo apps. Opening them via double-click on the icon opened the console window. But of course when I went to my IDE to put this in, it did not appear. So really this would actually be for deployment but not a good enough solution for that.

In that case, I will just close this PR and everything stays as is :-)

andydotxyz:
OK. Thanks for helping us to improve - sorry that we'd got something in place already :)

gonutz:
Let me just say, you are doing a really good job, hats off. Keep up the good work :-)

PR #1355, opened 2020-10-02
Refactoring: cleanup

toaster:
### Description:

I found some focus artifacts in `PopUpMenu`.
I removed them and it turned out that `testCanvas` had no proper focus support.
I then moved the overlay focus manager handling into the `internal.OverlayStack`. Thus, the test canvas can make use of it. In addition I copied the simple focus manager logic from `glCanvas` into `testCanvas` which is sufficient to make the tests happy again.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


test/testcanvas.go
```diff
@@ -43,90 +44,109 @@ type testCanvas struct {
...
 
-func (c *testCanvas) SetContent(content fyne.CanvasObject) {
-	c.propertyLock.Lock()
-	c.content = content
-	c.propertyLock.Unlock()
-
-	if content == nil {
-		return
+// LaidOutObjects returns all fyne.CanvasObject starting at the given fyne.CanvasObject which is laid out previously.
+func LaidOutObjects(o fyne.CanvasObject) (objects []fyne.CanvasObject) {
```
andydotxyz, line 57: Seems like a strange place to move this function to

andydotxyz (commented):
Looks good, though the rearranged function order seems strange

test/testcanvas.go
toaster: I thought we agreed to exported package functions before any package type functions, not?

andydotxyz: Pretty sure it’s type -> constructor -> exported type func -> unexported type func.
We didn’t put exported package funcs in the list I don’t think, I would imagine that it’s either at the top or the bottom of the file, but not in between type declarations and type functions.

toaster: Okay, so I propose the top of the file :).

stuartmscott (approved):

PR #1365, opened 2020-10-07
test utility functions to traverse focus

toaster:
### Description:

This PR introduces `test.FocusNext(fyne.Canvas)` and `test.FocusPrevious(fyne.Canvas)` to traverse the given canvas‘ focus.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


test/test_test.go
```diff
@@ -81,6 +155,31 @@ func (d *draggable) DragEnd() {
 	d.wasDragged = true
 }
 
+type focusable struct {
+	widget.BaseWidget
+	focused bool
+}
+
+var _ fyne.Focusable = (*focusable)(nil)
```
stuartmscott, line 163: nit: these conformity assertions are usually above the type

test/test_test.go
```diff
@@ -81,6 +155,31 @@ func (d *draggable) DragEnd() {
...
+
+func (f *focusable) Focused() bool {
```
stuartmscott, line 165: nit: alphabetical ordering would put Focused below FocusLost

stuartmscott (changes requested):
Few nits, otherwise looks good!

test/test_test.go
toaster: Yeah, you’re right.

test/test.go
```diff
@@ -54,6 +54,20 @@ func Drag(c fyne.Canvas, pos fyne.Position, deltaX, deltaY int) {
 	o.(fyne.Draggable).DragEnd()
 }
 
+// FocusNext focuses the next focusable on the canvas.
+func FocusNext(c fyne.Canvas) {
+	if tc, ok := c.(*testCanvas); ok {
```
andydotxyz, line 59: Should we log an error if it's not a testCanvas?
Otherwise people using this test code outside of the test context might be confused about the lack of effect.

test/test_test.go
toaster: Did we agreed to that?

test/test.go
toaster: I thought of it but it’s probably YAGNI.

test/test_test.go
stuartmscott: You're right we did not explicitly agree to that, but in general vars before types

test/test.go
toaster: I now changed it to fail :).

stuartmscott (approved):

test/test.go
andydotxyz: I don't know if this is a nit or not - but that's not going to fail, that's going to panic!

toaster: Ah, okay … for me that’s kind of failing :).

PR #1371, opened 2020-10-07
Fix crash when calling List.Layout before init

lusingander:
### Description:

When calling List.Layout before initializing, panic occurred because of out of range.
This could happens, for example, when switching themes right after startup with `fyne_demo` app.
I suppose that it is due to the fact that the `visibleItemCount` is no longer checked in #1330.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz (approved):

andydotxyz:
Great thanks!

PR #1373, opened 2020-10-08
List length fix

okratitan:
### Description:
This will prevent Length() from being called if the function has not been set.

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

#### Where applicable:

- [N/A] Public APIs match existing style.
- [N/A] Any breaking changes have a deprecation path or have been discussed.

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -91,7 +91,11 @@ func newListRenderer(objects []fyne.CanvasObject, l *List, scroller *ScrollConta
 }
 
 func (l *listRenderer) Layout(size fyne.Size) {
-	if l.list.Length() == 0 {
+	lengthFunc := l.list.Length
+	if lengthFunc == nil {
+		return
+	}
+	if lengthFunc() == 0 {
 		if len(l.children) > 0 {
 			for _, child := range l.children {
 				l.itemPool.Release(child)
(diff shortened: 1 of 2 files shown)


andydotxyz (changes requested):
Can you please add a test that shows this possible crash is now fixed?

widget/list.go
```diff
@@ -91,7 +91,11 @@ func newListRenderer(objects []fyne.CanvasObject, l *List, scroller *ScrollConta
 }
 
 func (l *listRenderer) Layout(size fyne.Size) {
-	if l.list.Length() == 0 {
+	lengthFunc := l.list.Length
+	if lengthFunc == nil {
+		return
+	}
+	if lengthFunc() == 0 {
```
stuartmscott, line 98: nit: Length is called multiple times
```suggestion
	length := 0
	if f == l.list.Length; f != nil {
		length = f()
	}
	if length == 0 {
		return
	}
```

okratitan: resolved

widget/list.go
```diff
@@ -412,8 +423,12 @@ func (l *listLayout) Layout(objects []fyne.CanvasObject, size fyne.Size) {
...
 
 func (l *listLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
-	return fyne.NewSize(l.list.itemMin.Width,
-		l.list.itemMin.Height*l.list.Length())
+	size := fyne.NewSize(0, 0)
+	if f := l.list.Length; f != nil {
+		size = fyne.NewSize(l.list.itemMin.Width,
+			l.list.itemMin.Height*f())
+	}
+	return size
```
stuartmscott, line 431: Using named return values saves you a line:
```go
func (l *listLayout) MinSize(objects []fyne.CanvasObject) (size fyne.Size) {
	if f := l.list.Length; f != nil {
		size.Width = l.list.itemMin.Width
		size.Height = l.list.itemMin.Height*f()
	}
	return
}
```

andydotxyz: Personally I would just invert the orignal logic.
```
if l.list.Length == nil {
    return fyne.Size{}
}

return itemMin etc...
```

That's exiting early and putting the logic at the top level.

okratitan: but not guarding against the Length variable changing which is the point of the whole f :=

andydotxyz: True - are we supporting the functions changing? I guess it makes sense.

You could remove a line of code by not having the size varable and just replacing the assign and return with two returns.

okratitan: resolved

stuartmscott (approved):

andydotxyz (approved):

Issue #1375, opened 2020-10-08
Table Doesn't Show All Cells


stuartmscott:
### Describe the bug:
Table only shows all cells if size exceeds scroller viewport

### Screenshots:
Working:
<img width="712" alt="Screen Shot 2020-10-08 at 4 24 37 PM" src="https://user-images.githubusercontent.com/1117080/95523152-c4f94100-0982-11eb-9d01-70b0b83e1cd4.png">

Not Working:
<img width="712" alt="Screen Shot 2020-10-08 at 4 25 50 PM" src="https://user-images.githubusercontent.com/1117080/95523221-efe39500-0982-11eb-8bba-3f0c369de35c.png">

### Example code:
```go
package main

import (
	"fmt"
	"fyne.io/fyne"
	"fyne.io/fyne/app"
	"fyne.io/fyne/dialog"
	"fyne.io/fyne/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Table Dialog")
	// WORKS
	data := [][]string{
		{"a", "b", "c", "d", "e", "f"},
		{"g", "h", "i", "j", "k", "l"},
		{"m", "n", "o", "p", "q", "r"},
		{"s", "t", "u", "v", "w", "x"},
	}
	/* DOESN'T WORK
	data := [][]string{
		{"a", "b", "c"},
		{"d", "e", "f"},
		{"g", "h", "i"},
	}
	*/
	w.SetContent(widget.NewButton("Show Table Dialog", func() {
		l := widget.NewTable(
			func() (int, int) {
				return len(data), len(data[0])
			},
			func() fyne.CanvasObject {
				return widget.NewLabel("Template")
			},
			func(row, col int, item fyne.CanvasObject) {
				item.(*widget.Label).SetText(fmt.Sprintf("Cell %s", data[row][col]))
			})
		d := dialog.NewCustom("Table Dialog", "OK", l, w)
		d.Show()
		d.Resize(fyne.NewSize(500, 500))
	}))
	w.Resize(fyne.NewSize(600, 600))
	w.ShowAndRun()
}
```

stuartmscott:
Screen recording of resize: https://slack-files.com/T029RQSE6-F01C63KFTTP-7c5815901a

PR #1376, opened 2020-10-09
Fix/1374

### LINKED ISSUE #1374: Table Renders Cells Beyond Limit
The table in fyne_demo has 500 x 150 cells, but is rendered with nearly double that.

Screen Shot 2020-10-08 at 4 22 23 PM

andydotxyz:
### Description:
Various graphical glitches resolved in Table.

Fixes #1374, #1375 

### Checklist:

- [ ] Tests included. <- the redraw code should be tested - except in the test infra it does not happen (different order of events)
- [x] Tests all pass.

### DIFF
--- a/widget/table.go
+++ b/widget/table.go
@@ -290,17 +290,23 @@ func (r *tableCellsRenderer) Refresh() {
 		r.returnAllToPool()
 	}
 
+	dataRows, dataCols := 0, 0
+	if f := r.cells.t.Length; f != nil {
+		dataRows, dataCols = r.cells.t.Length()
+	}
 	rows, cols := r.visibleCount()
 	offX := r.cells.t.offset.X - (r.cells.t.offset.X % (r.cells.cellSize.Width + tableDividerThickness))
 	minCol := offX / (r.cells.cellSize.Width + tableDividerThickness)
+	maxCol := fyne.Min(minCol+cols, dataCols)
 	offY := r.cells.t.offset.Y - (r.cells.t.offset.Y % (r.cells.cellSize.Height + tableDividerThickness))
 	minRow := offY / (r.cells.cellSize.Height + tableDividerThickness)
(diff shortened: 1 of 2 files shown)


widget/table.go
```diff
@@ -226,6 +226,14 @@ func (c *tableCells) CreateRenderer() fyne.WidgetRenderer {
 	return &tableCellsRenderer{cells: c, pool: &syncPool{}, visible: make(map[cellID]fyne.CanvasObject)}
 }
 
+func (c *tableCells) Resize(s fyne.Size) {
+	if s == c.size {
+		return
+	}
+	c.BaseWidget.Resize(s)
```
stuartmscott, line 233: Not sure if this is needed - it checks the size is different (which is done above on line 230), updates the size, and calls Layout (which does nothing in tableCells). Besides locking, I don't see why this call is needed, am I missing something?

widget/table.go
```diff
@@ -35,8 +35,8 @@ type Table struct {
 // The first returns the data size in rows and columns, second parameter is a function that returns cell
 // template objects that can be cached and the third is used to apply data at specified data location to the
 // passed template CanvasObject.
-func NewTable(len func() (int, int), create func() fyne.CanvasObject, update func(int, int, fyne.CanvasObject)) *Table {
-	t := &Table{Length: len, CreateCell: create, UpdateCell: update, SelectedRow: -1, SelectedColumn: -1}
+func NewTable(length func() (int, int), create func() fyne.CanvasObject, update func(int, int, fyne.CanvasObject)) *Table {
```
stuartmscott, line 38: Thanks

widget/table.go
andydotxyz: It was just a best practice I thought - we have overridden BaseWidget so we propagate.
That function could do other things in the future.

stuartmscott (approved):

PR #1379, opened 2020-10-09
File dialog locations

### LINKED ISSUE #821: FileDialog: Set start directory

andydotxyz:
Description:
Replaces PR #1108 

Fixes #821

Checklist:
- [x] Tests included.
- [x] Tests all pass.

Where applicable:
-   [x] Public APIs match existing style.


dialog/file.go
```diff
@@ -78,17 +82,21 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 		if f.file.save {
 			callback := f.file.callback.(func(fyne.URIWriteCloser, error))
 			name := f.fileName.(*widget.Entry).Text
-			path := filepath.Join(f.dir, name)
+			path, _ := storage.Child(f.dir, name)
```
charlesdaniels, line 85: Should rename `path` to `location`.

charlesdaniels: Commentary: we agreed that this was the terminology that should be `location` for URIs. I thought I went back and caught the places where I used `path`. I think we could also abbreviate to `loc` if you like.

dialog/file.go
```diff
@@ -208,13 +277,20 @@ func (f *fileDialog) refreshDir(dir string) {
...
 	f.dir = dir
 
 	f.breadcrumb.Children = nil
-	buildDir := filepath.VolumeName(dir)
-	for i, d := range strings.Split(dir, string(filepath.Separator)) {
+
+	localdir := dir.String()[len(dir.Scheme())+3:]
+
+	buildDir := filepath.VolumeName(localdir)
+	for i, d := range strings.Split(localdir, "/") {
```
charlesdaniels, line 293: We should consider a `.Components()` member for URI (or possibly just a `path` method which does this). We know that splitting on `/` is safe for URIs, but I worry that converting to a normal string and then doing the split later could end up causing confusion down the line. Not really applicable in this particular case, this is more of a general remark.

charlesdaniels (approved):

dialog/file.go
andydotxyz: Yeah, in most cases this is right and I'm updating. In one or two it relates to file specific code that we agreed would be removed in another PR

dialog/file.go
andydotxyz: This code will be stripped out. Generally it shouldn't be done like this on a URI and it is a hold-on from the old file code.
We have Child() and Parent() which supports traversing. Combined with the URIRootError we can figure out the whole path I am pretty sure. Addressing in follow-up PR

andydotxyz:
OK that's all comments addressed.
I also added storage.NewFileURI because I didn't want that windows file path handling in the generic NewURI constructor...

I am not certain now if StartingLocation is right - should we instead use SetLocation (so it's not start specific) or make it a Location field that can be written or read...

stuartmscott (approved):

charlesdaniels (approved):

andydotxyz:
Thanks for your help on this @charlesdaniels and @PucklaMotzer09, I know it's been a long haul but we got there :).
I will spend a little time looking at making it not file-specific which I think will allow a little cleanup too - will open another PR.

Issue #1338, opened 2020-09-25
Deprecate WidgetRenderer.BackgroundColor


stuartmscott:
If a dev wanted to display a background color they can use a layout.MaxLayout with canvas.Rectangle, or if they wanted to display an image they can use a layout.MaxLayout with canvas.Image.

Deprecating WidgetRenderer.BackgroundColor in 1.4 would pave the way for #1297 in 2.0.

stuartmscott:
Deprecation tag added by #1388, this can now be moved to 2.0 milestone

stuartmscott:
On develop, finally :)

PR #1388, opened 2020-10-10
Deprecate WidgetRenderer.BackgroundColor

stuartmscott:
### Description:
This is the first step of #1338, the second step is to remove existing usages in 2.0 release.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget.go
+++ b/widget.go
@@ -18,6 +18,9 @@ type Widget interface {
 // function and should be exactly one instance per widget in memory.
 type WidgetRenderer interface {
 	// BackgroundColor returns the color that should be used to draw the background of this renderer’s widget.
+	// Deprecated: Widgets will no longer have a background to support hover and selection indication in collection widgets.
+	// If a widget requires a background color or image, this can be achieved by using a canvas.Rect or canvas.Image
+	// as the first child of a MaxLayout, followed by the rest of the widget components.
 	BackgroundColor() color.Color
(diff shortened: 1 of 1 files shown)


widget.go
```diff
@@ -18,6 +18,9 @@ type Widget interface {
 // function and should be exactly one instance per widget in memory.
 type WidgetRenderer interface {
 	// BackgroundColor returns the color that should be used to draw the background of this renderer’s widget.
+	// Deprecated: Widgets will no longer have a background to support hover and selection indication in collection widgets.
+	// If a widget requires a background color or image, this can be achieved by using a canvas.Rect or canvas.Image
+	// as the first child of a MaxLayout, followed by the rest of the widget components.
```
toaster, line 23: `Deprecated:` should start a new paragraph and therefore needs to be separated with an empty line.
```suggestion
	//
	// Deprecated: Widgets will no longer have a background to support hover and selection indication in collection widgets.
	// If a widget requires a background color or image, this can be achieved by using a canvas.Rect or canvas.Image
	// as the first child of a MaxLayout, followed by the rest of the widget components.
```

toaster (changes requested):

widget.go
stuartmscott: Danke!

toaster (approved):

PR #1390, opened 2020-10-10
cmd/fyne: Small refactor and templates for packaging

Jacalz:
### Description:
This is some draft work on moving the packaging code to files per operating system and also moving to using templates to more easily changing files in the future. This adds up to making the packaging code much easier to read.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


cmd/fyne/internal/templates/Makefile
```diff
@@ -0,0 +1,14 @@
+# If PREFIX isn't provided, we check for /usr/local and use that if it exists.
+# Otherwice we fall back to using /usr.
+
+LOCAL != test -d /usr/local && echo -n "/local" || echo -n ""
```
andydotxyz, line 4: It ocurred to me that this should probavbly also respect the DESTDIR prefix...

cmd/fyne/internal/templates/gen.go
```diff
@@ -0,0 +1,59 @@
+// +build ignore
```
andydotxyz, line 1: This is quite a bit of duplicated code, can we re-use the internal code or use the "fyne package" tool that has it built as an external app?

cmd/fyne/package-common.go
```diff
@@ -0,0 +1,193 @@
+package main
```
andydotxyz, line 1: for this file I would just call it "package.go" we use "xxx-common" or "xxx-other" when it's a fallback to functionality that is otherwise OS specific, but in this case it's the loader that will access the OS code I think?

cmd/fyne/package-darwin.go
```diff
@@ -0,0 +1,61 @@
+package main
```
andydotxyz, line 1: Is this file supposed to only build on darwin? If so the format would be package_darwin.go

andydotxyz (changes requested):
Mostly open questions but I think we can avoid duplicating code by reusing the tools or re-use internal code instead of copying the whole bundle script?

cmd/fyne/package-darwin.go
andydotxyz: ignore this, I realised now it's just grouping :)

cmd/fyne/package-common.go
Jacalz: Sure. Just thought it looked better with them all being -something. Will change back :)

cmd/fyne/internal/templates/gen.go
Jacalz: For future reference: Using `go:genernate` works like a champ :)

cmd/fyne/internal/templates/Makefile
Jacalz: We do already? If you look at the install part, we already respect that I think.

andydotxyz: Literally in this line. We are checking the existence of a target prefix outside of the DESTDIR context

Jacalz: Sorry, but I am not quite sure I follow.

andydotxyz: We will write to $DESTDIR$PREFIX/bin...
to find out PREFIX we check /usr/local vs /usr.
But I think this should be $DESTDIR/usr/local vs $DESTDIR/usr surely?

Jacalz: Yeah absolutely you are right. Sorry that I didn't get that initially. Makes a lot of sense.

Jacalz: I just implemented the change and verified that local install works and I also set destdir and tested that it worked correctly.

andydotxyz (approved):

cmd/fyne/package-darwin.go
stuartmscott: It is unusual to have a hyphen in the name, typically an underscore is used

andydotxyz: Using an underscore would trigger conditional compilation but this is simply organising code per platform, excluding the methods would break the current implementation.

Issue #1314, opened 2020-09-15
Common Divider


stuartmscott:
Accordion, List, and Tree all use a canvas.Rectangle of 1dp thickness, and theme.ShadowColor() to represent a divider.

Perhaps theme.DividerThickness() and theme.DividerColor() would be useful additions.

andydotxyz:
Whilst we could add this in 1.4 I'm wondering if there is little benefit so pushing back to the theme review in 2.0

stuartmscott:
@Tilo did a great job on #1438, so moving this to 1.4

andydotxyz:
Just to capture part of a conversation from yesterday:

When using a custom layout you may want to resize a Separator widget. Whilst you widen or heighten this widget you should not do adjust both dimensions at the same time, so the recommended resize calls would be:

```go
separator.Resize(fyne.NewSize(newWidth, separator.MinSize().Height)) // make it wider
separator.Resize(fyne.NewSize(separator.MinSize().Width, newHeight)) // make it taller
```

In this way we can ensure that separators maintain a standard thickness and in 2.0 this will be theme controlled.

andydotxyz:
Done, the theme change in 2.0 will make it configurable.

PR #1392, opened 2020-10-11
Bugfix: separators

toaster:
### Description:

This PR adds a new internal separator widget which adjusts its color via a callback and therefore is able to follow theme changes.
It also uses this widget for the menu separator, the list divider and the separator in color dialogs.
The new widget is tested implicitly via the rendering tests of the widgets where it is used.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


internal/widget/separator.go
```diff
@@ -0,0 +1,99 @@
...
+// NewSeparator creates a separator with the given min size and the color delivered by the given callback.
+func NewSeparator(minSize fyne.Size, colorFn func() color.Color) *Separator {
+	return &Separator{minSize: minSize, colorFn: colorFn}
+}
+
+// CreateRenderer returns a new renderer for the separator.
+//
+// Implements: fyne.Widget
+func (s *Separator) CreateRenderer() fyne.WidgetRenderer {
+	bar := canvas.NewRectangle(theme.DisabledTextColor())
```
andydotxyz, line 29: I think the reason the images are failing is that widgets are using ShadowColor() for this at the moment

toaster: Sorry, I don’t know which tests you refer to.

andydotxyz: Quite right they are not failing sorry. The reason you had to replace a lot of images is that the divider is using shadow colour in most places. Isn’t that more logical than the text disabled colour in menu?

andydotxyz:
I don’t follow what bug this PR with bugfix prefix is fixing?

Re: naming isn’t this quite close to the material design divider that is built into tree, table, list? (And also it should probably be used in them too).

[fyi:
We discussed this commonality on a call a few weeks back but decided not to add the Divider widget yet, though adding it internally is a nice first step.]

toaster:
> I don’t follow what bug this PR with bugfix prefix is fixing?

You don’t want me to open an issue for every minor bug I find, don’t you?

> Re: naming isn’t this quite close to the material design divider that is built into tree, table, list? (And also it should probably be used in them too).

Okay, so `Divider` instead of `Separator`.

stuartmscott:
> so Divider instead of Separator

:+1:

stuartmscott:
Here's the initial feature request: #1314

andydotxyz:
> Not read it yet sorry @Stuart Scott - been working this morning, just getting back to the PRs now

Apologies I did not mean that there should be a bug filed before a PR, I meant that it was not clear what bug is being fixed.

toaster:
> Apologies I did not mean that there should be a bug filed before a PR, I meant that it was not clear what bug is being fixed.

Okay, so the bug is, that the current implementation of the dividers and the menu item separator is a simple rectangle which does not follow theme changes. While this is fairly easy fixable for widget renderers where the renderer directly contains the rectangle, it becomes awkward when the divider is just part of a list of items like for the menu.

andydotxyz:
Thanks @toaster, that helps to understand why the wider approach was needed.

andydotxyz:
I think the remaining question is whether or not we have 2 widgets, or if it should become consistent:

* Menu: 2dp, textdisabled color
* Collections etc: 1dp shadow color

internal/widget/divider.go
```diff
@@ -13,20 +13,20 @@ var _ fyne.Widget = (*Divider)(nil)
 // Divider is a widget for displaying a divider with themeable color.
 type Divider struct {
 	Base
-	colorFn func() color.Color
-	minSize fyne.Size
+	ColorFn func() color.Color
+	MinSz   fyne.Size
```
andydotxyz, line 17: I appreciate that you can't use MinSize, but not sure about `Sz` would Min be OK?

widget/menu_item.go
```diff
@@ -12,7 +12,10 @@ import (
 
 // newMenuItemSeparator creates a separator meant to separate MenuItems.
 func newMenuItemSeparator() fyne.CanvasObject {
-	return widget.NewDivider(fyne.NewSize(1, 2), theme.DisabledTextColor)
+	d := widget.NewDivider()
+	d.MinSz = fyne.NewSize(1, 2)
+	d.ColorFn = theme.DisabledTextColor
```
andydotxyz, line 17: Now that menus are on the same background as everything else, should we not be using a consistent divider color?

andydotxyz (changes requested):

internal/widget/divider.go
toaster: I think `Min` is too generic. You cannot understand it without having a look at the type.
I’m still not happy with the exported fields at all.
Maybe it is not that bad to have a `NewMenuItemSeparator` constructor for the `Divider` instead?
And maybe call it `NewMenuItemDivider` to make more clear that it’s just another `Divider`?

andydotxyz: If you really think they should be different then this approach sounnds good.

andydotxyz:
I wanted to explore this a little further, and I suspect you have a point about the contrast for dividers.
Here is current:

Screenshot 2020-10-13 at 10 25 06

Then I explored using the higher contrast dividers, and updated the menu divider size to match, which creates:

Screenshot 2020-10-13 at 10 30 36

I quite like the result, this could work as a single divider a happy medium?

toaster:
Agreed. We could still make it more complex in the future ;).

stuartmscott (approved):
Looks good. Though `theme.DisabledTextColor()` feels wrong as it has nothing to do with text. In future there should be an addition like `theme.DividerColor()`

widget/tree.go
```diff
@@ -502,9 +501,7 @@ func (r *treeContentRenderer) MinSize() (min fyne.Size) {
 
 func (r *treeContentRenderer) Objects() (objects []fyne.CanvasObject) {
 	r.treeContent.propertyLock.RLock()
-	for _, d := range r.dividers {
-		objects = append(objects, d)
-	}
+	objects = append(objects, r.dividers...)
```
andydotxyz, line 504: could just use objects = r.dividers, then return objects at the end

internal/widget/divider.go
```diff
@@ -0,0 +1,95 @@
...
+// Implements: fyne.Widget
+func (d *Divider) Hide() {
+	HideWidget(&d.Base, d)
+}
+
+// MinSize returns the minimal size of the divider.
+//
+// Implements: fyne.Widget
+func (d *Divider) MinSize() fyne.Size {
+	return MinSizeOf(d)
```
andydotxyz, line 45: We could save the renderer lookup here as we already know the answer - (1, 1)

andydotxyz (changes requested):
Looks great.
Minor request to avoid the renderer lookup in MinSize just to return a constant value.
And a suggestion about a line that could be tidied if you like.

andydotxyz:
This looks great but I have a new question - should this be a public widget? seems pretty helpful

andydotxyz (dismissed):

stuartmscott (approved):

PR #1393, opened 2020-10-11
Add Selection methods to list and tree.

okratitan:
### Description:

Fixes #(issue)

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

#### Where applicable:


widget/list.go
```diff
@@ -60,22 +71,43 @@ func (l *List) MinSize() fyne.Size {
...
+	}
+	l.selectedIndex = index
+	if index >= l.firstItemIndex && index <= l.lastItemIndex {
+		l.Refresh()
+		return
+	}
+	if index < l.firstItemIndex {
+		l.scroller.Offset.Y = index * l.itemMin.Height
+	} else {
+		l.scroller.Offset.Y = (index * l.itemMin.Height) - ((l.visibleItemCount - 1) * l.itemMin.Height)
```
andydotxyz, line 91: I can't quite figure out this maths.
Shouldn't it just be ((index+1)*minheight) - scrollerHeight?

widget/list.go
```diff
@@ -19,14 +19,18 @@ var _ fyne.Widget = (*List)(nil)
...
-	selectedItem   *listItem
-	selectedIndex  int
-	itemMin        fyne.Size
-	offsetY        int
+	Length           func() int
+	CreateItem       func() fyne.CanvasObject
+	UpdateItem       func(index int, item fyne.CanvasObject)
+	OnItemSelected   func(index int)
+	visibleItemCount int
+	firstItemIndex   int
```
andydotxyz, line 27: I think you can work without these additional fields cached.
If you calculate the offset position based on index then compare it to the scroll offsets, you'll know if it's visible.

okratitan: You're right - although those fields were already cached - I just move their location from the renderer to the widget.  If we already had that information seems to be more readable and simpler with less calcs.  No strong opinions either way though.

widget/list.go
okratitan: Yes good call

widget/list.go
```diff
@@ -60,22 +76,45 @@ func (l *List) MinSize() fyne.Size {
 	return l.BaseWidget.MinSize()
 }
 
+// SetSelected sets the index of the data that is to be selected, refreshes, and calls the OnSelected callback.
+func (l *List) SetSelected(index int) {
```
stuartmscott, line 80: nit: #1395
```suggestion
func (l *List) SetSelectedItem(index int) {
```

widget/list.go
```diff
@@ -60,22 +76,45 @@ func (l *List) MinSize() fyne.Size {
 	return l.BaseWidget.MinSize()
 }
 
+// SetSelected sets the index of the data that is to be selected, refreshes, and calls the OnSelected callback.
```
stuartmscott, line 79:
```suggestion
// SetSelected sets the index of the data that is to be selected, refreshes, and calls the OnSelectionUpdated callback.
```

stuartmscott (changes requested):

widget/list.go
andydotxyz: Don't you mean `OnSelectionChanged`?

widget/list.go
andydotxyz: I'll go with whatever another reviewer thinks. Am just not a fan of caching render details in the widget if it can be avoided.
Also I have a suspicion that using the approach of rounding to the nearest cell that scrolling an item that is off-bottom into the screen might bring it further into the view than necessary (as it's not checking the actual view dimensions when determining what to show).

andydotxyz:
I tried to test out what happens with the scroll-down code I mentioned above.
First I got this:

```
fyne.io/fyne/widget.(*List).SetSelected(0xc000160fc0, 0x7d)
	/Users/andy/Go/src/fyne.io/fyne/widget/list.go:96 +0xd9
fyne.io/fyne/cmd/fyne_demo/screens.makeListTab(0x44db060, 0xc0005a23f0)
	/Users/andy/Go/src/fyne.io/fyne/cmd/fyne_demo/screens/widget.go:327 +0x3fc
fyne.io/fyne/cmd/fyne_demo/screens.WidgetScreen(0x44db060, 0xc00029c300)
	/Users/andy/Go/src/fyne.io/fyne/cmd/fyne_demo/screens/widget.go:349 +0x4c7
main.main()
	/Users/andy/Go/src/fyne.io/fyne/cmd/fyne_demo/main.go:139 +0xbf7
exit status 2
```

Then I changed it to:
```
go func () {
	time.Sleep(10*time.Second)
	list.SetSelected(125)
}
```

And it worked just fine

widget/list.go
andydotxyz: Tested (as mentioned below) and the maths works out fine, sorry for the noise.

widget/list.go
okratitan: Resolved now

widget/list.go
okratitan: Resolved now

widget/list.go
```diff
@@ -19,24 +19,40 @@ var _ fyne.Widget = (*List)(nil)
...
-	UpdateItem     func(index int, item fyne.CanvasObject)
-	OnItemSelected func(index int)
-	selectedItem   *listItem
-	selectedIndex  int
-	itemMin        fyne.Size
-	offsetY        int
+	Length            func() int
+	CreateItem        func() fyne.CanvasObject
+	UpdateItem        func(index int, item fyne.CanvasObject)
+	OnSelectionChange func(index int)
```
stuartmscott, line 25: nit:
```suggestion
	OnSelectionChanged func(index int)
```

widget/tree.go
```diff
@@ -20,16 +20,16 @@ var _ fyne.Widget = (*Tree)(nil)
...
-	OnNodeSelected func(uid string)                                      // Called when the Node with the given Unique ID is selected.
+	ChildUIDs         func(uid string) (c []string)                         // Return a sorted slice of Children Unique IDs for the given Node Unique ID
+	IsBranch          func(uid string) (ok bool)                            // Return true if the given Unique ID represents a Branch
+	CreateNode        func(branch bool) (o fyne.CanvasObject)               // Return a CanvasObject that can represent a Branch (if branch is true), or a Leaf (if branch is false)
+	UpdateNode        func(uid string, branch bool, node fyne.CanvasObject) // Called to update the given CanvasObject to represent the data at the given Unique ID
+	OnBranchOpened    func(uid string)                                      // Called when a Branch is opened
+	OnBranchClosed    func(uid string)                                      // Called when a Branch is closed
+	OnSelectionChange func(uid string)                                      // Called when the Node with the given Unique ID is selected.
```
stuartmscott, line 32: nit:
```suggestion
	OnSelectionChanged func(uid string)                                      // Called when the Node with the given Unique ID is selected.
```

stuartmscott (changes requested):
Thanks, this is very close

widget/list.go
okratitan: Crash fixed.

widget/list.go

widget/list.go

widget/tree.go
okratitan: Resolved

widget/list.go

widget/list.go
```diff
@@ -60,22 +76,53 @@ func (l *List) MinSize() fyne.Size {
...
+	if f := l.Length; f != nil {
+		length = f()
+	}
+	if index < 0 || index >= length {
+		return
+	}
+	l.selected = index
+	if index >= l.firstItemIndex && index <= l.lastItemIndex {
+		l.Refresh()
+		return
```
stuartmscott, line 96: This return means OnSelectionChanged will not be triggered if the selection is in viewport

stuartmscott (changes requested):

widget/list.go
okratitan: Oopsy great eye. Fixed.

widget/list.go
okratitan: Okay cached values moved back to renderer

widget/list.go
```diff
@@ -60,6 +73,40 @@ func (l *List) MinSize() fyne.Size {
...
+	if l.scroller == nil {
+		if f := l.OnSelectionChanged; f != nil {
+			f(index)
+		}
+		return
+	}
+	y := index * l.itemMin.Height
+	if y < l.scroller.Offset.Y {
+		l.scroller.Offset.Y = y
+	} else if y > l.scroller.Offset.Y+l.scroller.Size().Height {
```
stuartmscott, line 100:
```suggestion
	} else if y+l.itemMin.Height > l.scroller.Offset.Y+l.scroller.Size().Height {
```

stuartmscott (approved):

PR #1394, opened 2020-10-11
Table hover and selection (functions)

andydotxyz:
### Description:
Adding hover effects to table.
Add SetSelected and ClearSelection functions

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/table.go
```diff
@@ -67,6 +89,50 @@ func (t *Table) CreateRenderer() fyne.WidgetRenderer {
...
+// SetSelected will mark the specified cell (at row, col) to be marked as selected.
+func (t *Table) SetSelected(row, col int) {
+	t.SelectedRow = row
+	t.SelectedColumn = col
+
+	t.scrollToVisible(row, col)
+	if t.OnCellSelected != nil {
+		t.OnCellSelected(row, col)
+	}
+
```
okratitan, line 101: I'm not sure we should be calling OnCellSelected when we are quite literally doing the opposite of what the api suggests by unselecting?

widget/table.go
```diff
@@ -67,6 +89,50 @@ func (t *Table) CreateRenderer() fyne.WidgetRenderer {
...
+	if t.moveCallback != nil {
+		t.moveCallback()
```
okratitan, line 103: Surely we shouldn't be moving the scroller on unselection either... but I will have to go back and look at the original code to see what moveCallback does.

okratitan (changes requested):

widget/table.go
andydotxyz: I think you might have commented on the wrong method?

widget/table.go
andydotxyz: it updates the indicators around the border.

widget/table.go
okratitan: yes sorry

widget/table.go
okratitan: Gotcha

widget/table.go
```diff
@@ -41,24 +44,43 @@ func NewTable(length func() (int, int), create func() fyne.CanvasObject, update
 	return t
 }
 
+// ClearSelection will clear any cell selection.
+func (t *Table) ClearSelection() {
+	t.SelectedRow = -1
+	t.SelectedColumn = -1
+
+	if t.OnCellSelected != nil {
+		t.OnCellSelected(-1, -1)
```
okratitan, line 53: I'm not sure we should be calling OnCellSelected when we are quite literally doing the opposite of what the api suggests by unselecting?

okratitan (changes requested):

widget/table.go
andydotxyz: I think that in conversation we agreed that this was consistent, so can you approve this please?

okratitan (dismissed):
Looks good

widget/table.go
```diff
@@ -67,6 +91,50 @@ func (t *Table) CreateRenderer() fyne.WidgetRenderer {
 	return r
 }
 
+// SetSelected will mark the specified cell (at row, col) to be marked as selected.
+func (t *Table) SetSelected(row, col int) {
```
stuartmscott, line 95: nit: #1395
```suggestion
func (t *Table) SetSelectedCell(row, col int) {
```

stuartmscott (changes requested):

widget/table.go
okratitan: So under this suggestion then list will need to be changed from SetSelected to SetSelectedItem?

andydotxyz: This was following the already-in API was it not? I didn't think I was making any choices here.

andydotxyz: List had SelectedItem field and Tree has SetSelectedNode

stuartmscott: > This was following the already-in API was it not?

This API hasn't hit master yet, so now is the time to get consistency

andydotxyz: Indeed, I was just referencing what you guys already had pushed to develop.
I didn't realise that we were adapting these PRs to #1395, I assumed that because it needed more discussion we would be addressing it later. It covers many widgets afterall.

andydotxyz: I think this now matches the updated discussion

widget/table_test.go
```diff
@@ -85,27 +108,91 @@ func TestTable_Selected(t *testing.T) {
...
+func TestTable_SetSelection(t *testing.T) {
+	app := test.NewApp()
+	defer test.NewApp()
+	app.Settings().SetTheme(theme.LightTheme())
+
+	table := NewTable(
+		func() (int, int) { return 5, 5 },
+		func() fyne.CanvasObject {
+			return NewLabel("placeholder")
+		}, func(row, col int, c fyne.CanvasObject) {
```
stuartmscott, line 167:
```suggestion
		},
		func(row, col int, c fyne.CanvasObject) {
```

widget/table_test.go
```diff
@@ -85,27 +108,91 @@ func TestTable_Selected(t *testing.T) {
...
+func TestTable_Selection(t *testing.T) {
+	app := test.NewApp()
+	defer test.NewApp()
+	app.Settings().SetTheme(theme.LightTheme())
+
+	table := NewTable(
+		func() (int, int) { return 5, 5 },
+		func() fyne.CanvasObject {
+			return NewLabel("placeholder")
+		}, func(row, col int, c fyne.CanvasObject) {
```
stuartmscott, line 133:
```suggestion
		},
		func(row, col int, c fyne.CanvasObject) {
```

widget/table_test.go
```diff
@@ -72,7 +72,30 @@ func TestTable_ChangeTheme(t *testing.T) {
...
+func TestTable_ClearSelection(t *testing.T) {
+	app := test.NewApp()
+	defer test.NewApp()
+	app.Settings().SetTheme(theme.LightTheme())
+
+	table := NewTable(
+		func() (int, int) { return 3, 5 },
+		func() fyne.CanvasObject {
+			return NewLabel("placeholder")
+		}, func(row, col int, c fyne.CanvasObject) {
```
stuartmscott, line 84:
```suggestion
		},
		func(row, col int, c fyne.CanvasObject) {
```

widget/table.go
```diff
@@ -236,18 +339,41 @@ func (c *tableCells) Resize(s fyne.Size) {
...
+	c.t.selectedColumn = e.Position.X / (c.cellSize.Width + tableDividerThickness)
+	c.t.selectedRow = e.Position.Y / (c.cellSize.Height + tableDividerThickness)
+
+	if c.t.OnSelectionChanged != nil {
+		c.t.OnSelectionChanged(c.t.selectedRow, c.t.selectedColumn)
+	}
+
+	if c.t.moveCallback != nil {
+		c.t.moveCallback()
+	}
```
stuartmscott, line 357: I wonder if the widget should be up-to-date before calling callbacks, ie call moveCallback before OnSelectionChanged

widget/table.go
```diff
@@ -67,6 +91,55 @@ func (t *Table) CreateRenderer() fyne.WidgetRenderer {
...
+func (t *Table) SetSelection(row, col int) {
+	t.selectedRow = row
+	t.selectedColumn = col
+
+	t.scrollToVisible(row, col)
+	if t.OnSelectionChanged != nil {
+		t.OnSelectionChanged(row, col)
+	}
+
+	if t.moveCallback != nil {
```
stuartmscott, line 109: ditto

widget/table.go
```diff
@@ -20,45 +21,68 @@ var _ fyne.Widget = (*Table)(nil)
...
+// ClearSelection will clear any cell selection.
+func (t *Table) ClearSelection() {
+	t.selectedRow = -1
+	t.selectedColumn = -1
+
+	if t.OnSelectionChanged != nil {
+		t.OnSelectionChanged(-1, -1)
+	}
+
+	if t.moveCallback != nil {
```
stuartmscott, line 58: ditto

stuartmscott (changes requested):
Few minor issues, otherwise looks good

widget/table_test.go
stuartmscott: Another one on line 107 which slipped passed the last review

widget/table.go
```diff
@@ -67,6 +90,55 @@ func (t *Table) CreateRenderer() fyne.WidgetRenderer {
...
+		t.moveCallback()
+	}
+
+	if t.OnSelectionChanged != nil {
+		t.OnSelectionChanged(row, col)
+	}
+}
+
+func (t *Table) scrollToVisible(row, col int) {
```
stuartmscott, line 113: nit: is there a better name than scrollToVisible? It implies that it will scroll to something that is visible, but only objects in the viewport are visible. scrollToShow row col, or scrollToSelection
```suggestion
func (t *Table) scrollToSelection(row, col int) {
```

andydotxyz: ScrollToSelection would be more wrong if we take the row, col parameters.
This was intended to be a utility method to scroll the specified index to be visible.
It could be called from a hypothetical "ScrollToTopLeft" whereby it would make (0, 0) visible.

stuartmscott: Then how about just `scrollTo`?

stuartmscott (approved):

PR #1398, opened 2020-10-12
Get the last bunch of staticchecks working (except for deprecations)

Jacalz:
### Description:
This makes sure to fix the mobile code and also run some checks that couldn't run before due to missing deps. 
There is still a compile error for `fyne.io/mobile` module, so compile errors are still turned off.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz (approved):
This is smart stuff, I had to read carefully some of the code to understand why it was not needed :)

PR #1402, opened 2020-10-13
Refactoring: separate radio item

toaster:
### Description:

This PR extracts the radio item into a separate widget.
This helps implementing the keyboard control and also simplifies the hovering and tapping code.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/radio_extended_test.go
+++ b/widget/radio_extended_test.go
@@ -8,6 +8,7 @@ import (
 	"fyne.io/fyne/internal/cache"
 	"fyne.io/fyne/test"
 	"fyne.io/fyne/theme"
+
 	"github.com/stretchr/testify/assert"
 )
 
@@ -20,6 +21,7 @@ func newExtendedRadio(opts []string, f func(string)) *extendedRadio {
 	ret.Options = opts
 	ret.OnChanged = f
 	ret.ExtendBaseWidget(ret)
+	ret.update() // Not needed for extending Radio but for the tests to be able to access items without creating a renderer first.
 
 	return ret
 }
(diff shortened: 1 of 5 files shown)


widget/radio_extended_test.go
```diff
@@ -36,7 +38,7 @@ func TestRadio_Extended_Selected(t *testing.T) {
 	radio := newExtendedRadio([]string{"Hi"}, func(sel string) {
 		selected = sel
 	})
-	radio.Tapped(&fyne.PointEvent{Position: fyne.NewPos(theme.Padding(), theme.Padding())})
```
andydotxyz, line 39: I think that we need to re-add Tapped to Radio, as this is technically part of the public API.
These tests should not change during a refactor

widget/radio_extended_test.go
```diff
@@ -135,59 +137,48 @@ func TestRadio_Extended_Hovered(t *testing.T) {
...
 			radio.Horizontal = tt.isHorizontal
-			render := cache.Renderer(radio).(*radioRenderer)
+			item1 := radio.items[0]
+			render1 := cache.Renderer(item1).(*radioItemRenderer)
+			render2 := cache.Renderer(radio.items[1]).(*radioItemRenderer)
 
-			assert.Equal(t, false, radio.hovered)
-			assert.Equal(t, theme.BackgroundColor(), render.items[0].focusIndicator.FillColor)
-			assert.Equal(t, theme.BackgroundColor(), render.items[1].focusIndicator.FillColor)
+			assert.False(t, item1.hovered)
```
andydotxyz, line 144: Just for clarity of the above, changing the test code here is OK, as it's internal render detail :)

andydotxyz (changes requested):
Overall good, but the exported methods that were removed need to be re-added just as a wrapper to the internal taps, to maintain compatibility.

PR #1403, opened 2020-10-13
special test theme for visual tests

toaster:
### Description:

This PR introduces a new, less ugly test theme for usage with visual tests.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


widget/hyperlink.go
```diff
@@ -94,7 +94,7 @@ func (hl *Hyperlink) textStyle() fyne.TextStyle {
 
 // textColor tells the rendering textProvider our color
 func (hl *Hyperlink) textColor() color.Color {
-	return theme.HyperlinkColor()
```
andydotxyz, line 97: This is going to break custom themes.
We deprecated the Hyperlink color for a reason, we can't really change it until 2.0

andydotxyz (changes requested):
Adding a new ttheme to do this makes sense - but I don't think we can change widget behaviour to do this.
Just because the builtin theme makes HyperlinkColor the same as Primary colour does not mean that custom themes will have done the same.

theme/icons.go
```diff
@@ -143,7 +143,7 @@ func (res *DisabledResource) Name() string {
 
 // Content returns the disabled style content of the correct resource for the current theme
 func (res *DisabledResource) Content() []byte {
-	clr := fyne.CurrentApp().Settings().Theme().DisabledIconColor()
```
andydotxyz, line 146: Oops, same as hyperlink, dont change the public API use of colour yet

widget/hyperlink.go
toaster: Okay, then I will remove the deprecated color from the test theme and apply the same delegation as for the internal themes.

PR #1404, opened 2020-10-14
Use fyne.URI in isHidden

PucklaMotzer09:
### Description:
This fixes the ```isHidden``` function. Previously there were a lot of error logs when isHidden was called which actually slowed down the execution.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz (approved):

andydotxyz:
Thanks for this. Quick hint, tick the check boxes in the PR template if you have done the task

PR #1407, opened 2020-10-14
fix Select not refreshing appropriate when using SetSelected

toaster:
### Description:

This PR fixes a bug where the `Label` of a `Select` was not refreshed and therefore not displaying the correct value after changing the selection with `SetSelected`.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Great catch thanks

PR #1408, opened 2020-10-14
Add default Tree constructor

stuartmscott:
### Description:
Adds default constructor with four required callbacks

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


widget/tree.go
```diff
@@ -36,6 +36,17 @@ type Tree struct {
 	leafMinSize   fyne.Size
 }
 
+// NewTree returns a new performant tree widget defined by the passed functions.
+// childUIDs returns the child Unique IDs of the given node.
+// isBranch returns true if the given node is a branch, false if it is a leaf.
+// create returns a new template object that can be cached.
+// update is used to apply data at specified data location to the passed template CanvasObject.
+func NewTree(childUIDs func(uid string) (c []string), isBranch func(uid string) (ok bool), create func(branch bool) (o fyne.CanvasObject), update func(uid string, branch bool, node fyne.CanvasObject)) *Tree {
```
andydotxyz, line 44: The name of the return types doesn't seem needed here?
Old habbits right? ;)

andydotxyz (changes requested):

widget/tree.go
stuartmscott: Indeed, resolved

andydotxyz (approved):

PR #1409, opened 2020-10-14
FileTree with Filter and Sorter

stuartmscott:
### Description:
Creates a new type FileTree which embeds Tree and supports Filtering and Sorting

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/filetree.go
+++ b/widget/filetree.go
@@ -9,6 +9,8 @@ import (
 	"fyne.io/fyne/theme"
 )
 
+// FileTree widget displays a hierarchical file system.
+// Each node of the tree must be identified by a Unique ID.
 type FileTree struct {
 	Tree
 	Filter func(string) bool
--- a/widget/filetree.go
+++ b/widget/filetree.go
@@ -13,8 +13,8 @@ import (
 // Each node of the tree must be identified by a Unique ID.
 type FileTree struct {
 	Tree
-	Filter func(fyne.URI) bool
-	Sorter func(fyne.URI, fyne.URI) bool
+	Filter func(fyne.URI) bool           // Returns true if the given URI passes the filter, false if it should be dropped
+	Sorter func(fyne.URI, fyne.URI) bool // Returns true if the first parameter should be sorted above the second
 }
 
 // NewFileTree creates a new tree with the given file system URI.
--- a/widget/filetree.go
+++ b/widget/filetree.go
@@ -1,0 +1,88 @@
+package widget
+
+import (
+	"sort"
+
+	"fyne.io/fyne"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/storage"
+	"fyne.io/fyne/theme"
+)
+
+type FileTree struct {
+	Tree
+	Filter func(string) bool
+	Sorter func(string, string) bool
+}
+
+// NewFileTree creates a new tree with the given file system URI.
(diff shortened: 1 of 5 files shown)


widget/filetree.go
```diff
@@ -0,0 +1,88 @@
...
+	"fyne.io/fyne"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/storage"
+	"fyne.io/fyne/theme"
+)
+
+type FileTree struct {
+	Tree
+	Filter func(string) bool
+	Sorter func(string, string) bool
```
andydotxyz, line 15: I struggled to guess how this method would be used based on the name. Is there any way it could be clearer?
For example when this is passed to `sort.Slice` it's called `less` which helped me to guess.

It's a bit random but how about `SortAbove`?

andydotxyz:
From a PR point of view this looks good, but there is a conversation in Slack just now about whether it should be landed into this repo in this form...

okratitan (changes requested):
I do not believe it is correct to create a new widget that is just loading data into an existing widget.  Tree is designed to have data loaded in.  The developer can load data in.  There is literally nothing here that extends a widget whatsoever.  It sets up a tree and sets up the functions required to work with tree in the exact default manner that tree is designed for, with no extension of tree, no additional widgets, no additional objects, no changes to sizing, layout, etc...  This is quite literally adding an entirely new widget just to load some data in to tree as a helper.

andydotxyz:
Personally I feel there is value in ‘widget.NewFileTree’ as it is a clean, concise API that developers can use without needing to know how it is pulled together. 

The alternative of having a constructor function for tree that takes 3 file related parameters makes less sense from an external developers point of view, I imagine.
Sensible defaults and ways to change them where required has been our approach, and to achieve this another type seems required.

okratitan:
If it did anything at all to extend or change tree or fill a use case tree couldn't handle i could get on board with that idea. As it is providing a shortcut to setting a widgets properties/values/methods/etc... does not make a new widget. It aligns closer to a constructor. "Shortcut to set up this widget". Setting that precedent isn't good. What's to stop us from creating new widgets to set properties and values and methods on other widgets. Over simplified and philosophically how is this different than creating a new widget to add a button with the label Accept or a new widget to add an icon with a folder image or a widget with a radio group for Yes/No/Cancel since all of those are very common use cases too?  Creating widgets to construct another widget a certain way is bad practice and thats the purpose of constructors and helpers.  I think the rule of thumb should be. 1. Can widget A accomplish what what widget B does? 2. Does widget B add anything to widget A? 3. Is widget Bs purpose only to help set values and methods in widget A?  If you answered yes, no, yes it shouldn't be a new widget.

andydotxyz:
I agree with the utility and convenience of this widget.
From an API users point of view it's a file tree. Everything included to show and work with a file tree.
It would be possible to make this less "just a wrapper" by hiding the tree in the renderer and exposing file-specific callbacks (OnSelected(uri) would make sense).

Given the arguments above however it seems unlikely to get the usual level of approval so let's hold off. Following on from @okratitan thoughts it raises questions about the NewTreeWithFiles constructor as well, given that the end result from an API usage point of view is basically the same but with a less clearly-named constructor.

stuartmscott:
Superseded by https://github.com/fyne-io/fyne/pull/1415

PR #1410, opened 2020-10-14
code style and cleanup

toaster:
### Description:

This PR adjusts the code style of the `Radio` widget and removes some deprecated method calls.
It also contains a fix regarding locking in the `Radio` widget.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
From the tests it looks like this broke infinite progressbar

andydotxyz:
(I think there is some sort of test.NewApp loop or buildup?)

stuartmscott (approved):

PR #1412, opened 2020-10-14
added an "uninstall" option to the packaged Makefile

starkers:
### Description:
Just a quick mod as I thought a `make uninstall` option would be handy

I prefix'ed the `rm` commands with a `-` which means that failure is acceptable and it will clean up all effected files (so in the case that someone removed or renamed their `.desktop` file.. the uninstall target would continue..)

image

### DIFF
--- a/cmd/fyne/package.go
+++ b/cmd/fyne/package.go
@@ -148,10 +148,15 @@ func (p *packager) packageLinux() error {
 			"\n"+
 			"default:\n"+
 			`	# Run "sudo make install" to install the application.`+"\n"+
+			`	# Run "sudo make uninstall" to uninstall the application.`+"\n"+
 			"install:\n"+
 			"	install -Dm00644 usr/"+local+"share/applications/"+p.name+`.desktop $(DESTDIR)$(PREFIX)/share/applications/`+p.name+".desktop\n"+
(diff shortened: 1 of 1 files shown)


Jacalz:
Thanks for your PR @starkers. This seems like a great addition :)

Pull Requests should be targeting the `develop` branch. Current changes to that branch sadly make it not possible to merge this in the current form (partly due to a GitHub workflow that is required but not existing in master branch). Could you please rebase this on the `develop` branch and target that branch for merge?

Please do ask if you have any questions. The develop branch has had a big refactor of this code. [This file](https://github.com/fyne-io/fyne/blob/develop/cmd/fyne/internal/templates/data/Makefile) would need to be edited and then you would just need to run `go generate` in [cmd/fyne/internal/templates](https://github.com/fyne-io/fyne/tree/develop/cmd/fyne/internal/templates) to get it updated in the bundled template. Please do ask if you have any questions :)

andydotxyz:
Exactly what @Jacalz said. I would have been tempted to land this on master and have another PR for the develop version, but we expect to release in 2 weeks from develop, so let's just get that version sorted and it will be out soon enough.

andydotxyz:
Picked it over to develop, will see if I can rescue the changes onto the new template system

andydotxyz:
Thanks for this, I saw the conflict on `develop` after switching branch so I merged the conflict in a new PR #1446.
The initial commit from @starkers is still in the history :)

PR #1415, opened 2020-10-15
Remove File Tree

stuartmscott:
Moved to https://github.com/fyne-io/examples

### Checklist:

- [n/a] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne_demo/screens/container.go
+++ b/cmd/fyne_demo/screens/container.go
@@ -4,13 +4,11 @@ import (
 	"fmt"
 	"image/color"
 	"net/url"
-	"os"
 
 	"fyne.io/fyne"
 	"fyne.io/fyne/canvas"
 	"fyne.io/fyne/container"
-	internalwidget "fyne.io/fyne/internal/widget"
-	"fyne.io/fyne/storage"
+	internalWidget "fyne.io/fyne/internal/widget"
 	"fyne.io/fyne/theme"
 	"fyne.io/fyne/widget"
 )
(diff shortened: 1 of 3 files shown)


stuartmscott:
Moved to https://github.com/fyne-io/examples/pull/28

andydotxyz (approved):
It's a shame but I agree we should not publish with controversial APIs.

andydotxyz:
An extract from the call on friday suggested there may be a way forward if this incorporated the favourites etc that file dialog presents...

```
Worried about the precident of a new API with no new widget implemented under?
A more complete selector could be good, including favourites buttons etc.

The data binding will provide shared handling of the main data but probably would not cover the favourites

Could have a FileTree that is the favourites etc in addition to a file structure browse could be a reasonable widget.
- uri handling
- some file handling helpers?
- perhaps some file handling context menus?

- what if the "default" root was Home?
(other roots are the favourites / drives like file dialog left panel)
drives and favourites open other tree branches
```

stuartmscott:
Seems like useful functionality, tho not sure if that needs to be a separate widget as all it does is return home and favourites as the children of the root, something like:

```go
tree.ChildUIDs(uid string) []string {
  var children []string
  if uid == tree.Root {
        children = append(children, homeDir, downloadsDir)
  } else {
        // Normal directory listing...
  }
  return children
}
```

andydotxyz:
Perhaps, except it also needs to manage the possible duplicates by mounting the data tree in different roots

stuartmscott:
I think that is outside the scope for an initial collection widget, but could be considered for future releases, feel free to file a feature request

Issue #1387, opened 2020-10-10
Runtime Release Flag


stuartmscott:
### Is your feature request related to a problem? Please describe:
An app built in development mode is different to an app built in release mode that will be published to the general public. Often time behaviour is changed and/or different APIs are accessed. Ideally an app could lookup at runtime whether it is a development version, or release version.

### Is it possible to construct a solution with the existing API?
not that I can see

### Describe the solution you'd like to see:
The addition of `App.Settings.Release bool` which defaults to false for development builds, but is set to true by the future `fyne release` command.

andydotxyz:
I was thinking more about this and have the following solution proposal:

```go
type BuildType int

const (
	StandardBuild BuildType = iota
	DebugBuild
	TestBuild
	ReleaseBuild
)
```

This way testApp.BuildType() will always return TestBuild.

The fyne "release" command (still in development) will cause the default implementation to change from returning StandardBuild to ReleaseBuild. I figure that at some point we will want debug info so that's in there too.

stuartmscott:
The overall idea sounds good to me.

Though I'm not sure we need `StandardBuild` and `TestBuild` (tests should use `DebugBuild`).

The 0-value would be `DebugBuild`

```go
type BuildType int

const (
	DebugBuild BuildType = iota
	ReleaseBuild
)
```

The `fyne release` command would use a build tag and there would be two files;

```go
// +build !release

func (a *App) BuildType() BuildType {
    return DebugBuild
}
```

```go
// +build release

func (a *App) BuildType() BuildType {
    return ReleaseBuild
}
```

andydotxyz:
Maybe I should add some more background context so we can gather all the previous discussions in this area.

* Debug mode has been requested to include things like outlining graphical components, printing layout sizing info etc
* Test mode is when our test suite is running the application in a non-visible way, but can still do image comparisons so needs to render in a theme-appropriate manner
* Release mode is where we perform additional certification and/or environment variable / flag tweaking for release.

According to the above the debug mode cannot be on by default, whichh is why it is not `0` in my proposal above.
Following the logic of all the above we need something that is neither debug nor test nor release, hence `StandardBuild` which was tough to name but I could not think of better - `DefaultBuild` was the only other that came to mind but we have previously discussed that, for example, `DefaultButton` usage has led to confusion in the past.

[...]

PR #1416, opened 2020-10-16
Feature/release

### LINKED ISSUE #86: Add “release” subcommand to build and package
A shortcut to build and package that sets the appropriate environment variables then call “go build”

andydotxyz:
### Description:
A work in progress for the various release / certification fixes we need.
Currently present:

- [x] Release command
- [x] Android zipalign
- [x] Release / build mode in API (see #1387)

Fixes #86 

### Checklist:

- [x] Tests included. <- needs more
- [x] Tests all pass.


andydotxyz:
I will create a second PR for the iOS portions, that will include:

- [ ] iOS .ipa build
- [ ] iOS certification
- [ ] Update / remove README-mobile.md

andydotxyz:
Moving Android certification to another PR, possibly with the version number update required to build without editing manifest...

app/settings_test.go
```diff
@@ -5,11 +5,21 @@ import (
...
+	"fyne.io/fyne/test"
 	"fyne.io/fyne/theme"
 )
 
+func TestSettingsBuildType(t *testing.T) {
+	set := test.NewApp().Settings()
+	assert.Equal(t, fyne.StandardBuild, set.BuildType()) // during test we should have a normal build
+
+	set = &settings{}
+	assert.Equal(t, buildMode, set.BuildType()) // when testing this package only it could be debug or release
```
stuartmscott, line 20: Not sure I follow the comment here. The assertion seems to just test the getter returns the constant.

cmd/fyne/commands/release.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (r *releaser) PrintHelp(indent string) {
+	fmt.Println(indent, "The release command prepares an application for public distribution.")
+}
+
+func (r *releaser) Run(params []string) {
+	r.packager.release = true
+
+	err := r.beforePackage()
+	if err != nil {
```
stuartmscott, line 40: nit:
```suggestion
	if err := r.beforePackage(); err != nil {
```

cmd/fyne/commands/release.go
```diff
@@ -0,0 +1,83 @@
...
+		fmt.Fprintf(os.Stderr, "%s\n", err.Error())
+		return
+	}
+	r.packager.Run(params)
+	err = r.afterPackage()
+	if err != nil {
```
stuartmscott, line 46: nit:
```suggestion
	if err := r.afterPackage(); err != nil {
```

cmd/fyne/commands/release.go
```diff
@@ -0,0 +1,83 @@
...
+func (r *releaser) beforePackage() error {
+	if util.IsAndroid(r.os) {
+		return util.RequireAndroidSDK()
+	}
+
+	return nil
+}
+
+func (r *releaser) zipAlign(path string) error {
+	buildTools := util.AndroidBuildToolsPath()
```
stuartmscott, line 69: nit: this var is only read once, consider inlining it into line 76

cmd/fyne/internal/util/mobile.go
```diff
@@ -1,8 +1,57 @@
...
+
+	if childDir == "" {
+		return dir
+	}
+	return filepath.Join(dir, childDir)
+}
+
+// IsAndroid returns true if the given os parameter represents one of the Android targets.
+func IsAndroid(os string) bool {
+	return strings.Index(os, "android") == 0
```
stuartmscott, line 41: This uses `strings.Index`, whereas the function below uses `strings.HasPrefix` - is there an advantage to one over the other?
Could they be unified, where `IsMobile` returns `os == "ios" || IsAndroid(os)`

cmd/fyne/internal/util/mobile.go
```diff
@@ -1,8 +1,57 @@
...
 
 // IsMobile returns true if the given os parameter represents a platform handled by gomobile.
 func IsMobile(os string) bool {
 	return os == "ios" || strings.HasPrefix(os, "android")
 }
+
+// RequireAndroidSDK will return an error if it cannot establish the location of a valid Android SDK installation.
+// This is currently deduced using ANDROID_HOME environment variable.
+func RequireAndroidSDK() error {
+	if env, ok := os.LookupEnv("ANDROID_HOME"); !ok || env == "" {
```
stuartmscott, line 52: Should this also check `env` points to a readable directory?

stuartmscott (changes requested):

app/settings_test.go
andydotxyz: It is attempting to demonstrate / document the difference between the BuildType API which must always be StandardBuild for testing, but the tests in this package are testing the internal settings code.
I considered not adding the second line, but it felt like it could be misleading for anyone who wanted to use the buildMode constant during test

cmd/fyne/internal/util/mobile.go
andydotxyz: It's not something we've done elsewhere - when using os.UserHome() we don't check that is a readable directory

stuartmscott: ack

app/settings_test.go
stuartmscott: Hmm so line 17 should always pass as testApp.BuildType() always returns Standard, and line 20 tests that a generic settings should return whatever the package was built as (Standard, Debug, or Release)

stuartmscott (approved):

PR #1422, opened 2020-10-17
Feature: keyboard controllable select

toaster:
### Description:

This PR enables keyboard control for `Select`.
`Up`, `Down` and `Space` open the drop-down.
`Left` and `Right` traverse the options if the drop-down is not open.

The keyboard control for the drop-down will be available as soon as menus have keyboard control.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Great addition.

However I am worried that we're getting an exploding number of images in tests.
In the change-item test (left/right) would checking the Select.Selected value not be sufficient?

stuartmscott:
> Up, Down and Space open the drop-down.
> Left and Right traverse the options if the drop-down is not open.

Not sure about this, feels like switching from vertical (Up/Down) to horizontal (Left/Right) would be unnecessarily confusing for the end user.

toaster:
> Not sure about this, feels like switching from vertical (Up/Down) to horizontal (Left/Right) would be unnecessarily confusing for the end user.

I don’t know how it works on Windows (who uses this anyway? ;)) but on MacOS it works exactly that way.

toaster:
> However I am worried that we're getting an exploding number of images in tests.
> In the change-item test (left/right) would checking the Select.Selected value not be sufficient?

Checking `Select.Selected` is not sufficient because `Select.Selected = "…"` is not.
Checking the rendered outcome is the most obvious and intuitive way and it’s easy, too.
Maybe we need the ASCII renderer which also would allow to have the expectations inline.

andydotxyz:
> Checking Select.Selected is not sufficient because Select.Selected = "…" is not.

I’m sorry but I did not understand this sentence - can you expand on it?

My expectation from the description above is that “press left/right” == selection changed so we could just check that it updated.

toaster:
> > Checking Select.Selected is not sufficient because Select.Selected = "…" is not.
> 
> I’m sorry but I did not understand this sentence - can you expand on it?
> 
> My expectation from the description above is that “press left/right” == selection changed so we could just check that it updated.

The expectation is that “press left/right” not only alters `Select.Selected` but also updates the rendered content. `Select.Selected = …` does not do that because it does not `Refresh()`.

stuartmscott:
Just tried this in fyne_demo, pressing down does indeed display the popup, but pressing down again does not move the focus into the popup so it isn't possible to select something from the popup with the keyboard.

Also, this doesn't seem to work for SelectEntry.

andydotxyz:
I think the need for keyboard handling in menus is another ticket.
Fixing that would work for this, selectentry as well as context menus and manu menus once they are activated.

PR #1425, opened 2020-10-18
Tree Selection

stuartmscott:
### Description:
Triggers OnSelectionChanged in SetSelection, adds ClearSelection, and TreeNoSelection.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/tree.go
+++ b/widget/tree.go
@@ -178,6 +178,11 @@ func (t *Tree) Resize(size fyne.Size) {
 	t.Refresh() // trigger a redraw
 }
 
+// Selection returns the Unique ID of the currently selected node, or TreeNoSelection if nothing is selected.
+func (t *Tree) Selection() string {
+	return t.selected
+}
+
 // SetSelection updates the current selection to the node with the given Unique ID.
 func (t *Tree) SetSelection(uid string) {
 	t.selected = uid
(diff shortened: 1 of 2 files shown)


widget/tree_internal_test.go
```diff
@@ -278,6 +278,35 @@ func TestTree_MinSize(t *testing.T) {
...
+	}
+
+	tree.SetSelection("A")
+	select {
+	case s := <-selection:
+		assert.Equal(t, "A", s)
+	case <-time.After(1 * time.Second):
+		assert.Fail(t, "Selection should have occured")
+	}
+
```
andydotxyz, line 300: Can you check that the Selection() method returns the correct result after the selection is changed as well?

widget/tree.go
```diff
@@ -68,6 +71,11 @@ func NewTreeWithStrings(data map[string][]string) (t *Tree) {
 	return
 }
 
+// ClearSelection clears the current selection.
```
andydotxyz, line 74: ClearSelection and SetSelection are here but no Selection() method

andydotxyz (changes requested):
From the agreed API list I think you're missing `Tree.Selection() string`

widget/tree_internal_test.go
stuartmscott: Done!

widget/tree.go
stuartmscott: Ooops forgot that one. Added now :)

andydotxyz (approved):

PR #1427, opened 2020-10-19
Prevent panic on Text() for an empty TextGrid

jangler:
Text() now returns an empty string instead.

### Description:

`NewTextGrid().Text()` currently passes `-1` as a length argument to `make()`, causing a panic. This change makes sure the argument is at least zero.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


widget/textgrid.go
```diff
@@ -114,7 +114,7 @@ func (t *TextGrid) Text() string {
 		count += len(row.Cells)
 	}
 
-	runes := make([]rune, count)
+	runes := make([]rune, int(math.Max(0, float64(count))))
```
andydotxyz, line 117: It would probably be easier to read and a bit cleaner if it's just:
```go
if count == 0 {
    return ""
}
```
No need for math or to fall through the for loops. We like early exits for corner cases :)

andydotxyz (changes requested):
Great catch, thanks.
Perhaps you can make this requested change to better match our style?

stuartmscott:
The Travis failure appears unrelated to this PR: `gocyclo: command not found`

jangler:
The early exit was my initial fix, but I second-guessed myself \:)

widget/textgrid.go
```diff
@@ -114,6 +114,10 @@ func (t *TextGrid) Text() string {
 		count += len(row.Cells)
 	}
 
+	if count < 0 {
```
stuartmscott, line 117:
```suggestion
	if count <= 0 {
```

stuartmscott (commented):
Thanks @jangler, this looks good to me!

@andydotxyz should we land this on master or develop?

andydotxyz:
Gocyclo error is fixed on develop - which flagged neatly that this should be on develop branch :)
Often we will accept hot fixes on master, but given we are releasing develop in 10 days we're agregating everything there.

jangler:
Ah, that should have been obvious if I had checked the branch histories, but maybe that info is worth including in CONTRIBUTING.md and/or the PR template?

andydotxyz:
Yes, we need to reconcile the various contributor docs. It’s on my list once the release is feature complete.

The updated changes look fine to me but I have to leave . Once the conflict is resolved and builds all pass @stuartmscott can go ahead and merge :)

andydotxyz:
I think the develop rebase did not go well. Might be easier to open a new PR this time...

jangler:
Opened #1428.

PR #1430, opened 2020-10-20
Fix issue where long or tall images could jump on load

### LINKED ISSUE #1266: Image with ImageFillContain seen stretched before correcting for aspect.
### Describe the bug:
 Image first shows as stretched to fill canvas. After a short pause (probably processing time) the image is re-displayed corrected for aspect.

### To Reproduce:
Run this code with local image

package main

import (
	"fmt"
	"imaging"
	"os"

	"fyne.io/fyne"
	"fyne.io/fyne/app"
	"fyne.io/fyne/canvas"
)

func main() {
	myApp := app.New()
	w := myApp.NewWindow("Image")
	w.Resize(fyne.NewSize(600, 600)) // problem for all sizes tried

	// Open JPEG file and get io.Reader

	fileName := os.Args[1]
	file, err := os.Open(fileName)
	if err != nil {
		fmt.Printf("Error opening file %s\n", fileName)
		return
	}

	// Set the AutroOrientation option for decoding.
	// We need to use imaging,Decode because it checks for and corrects for orientation data.
	// If present this is held in an EXIF tag. image.Decode() follows the standard
	// JPEG library and assumes file will be soley JFIF and correctly orientated.
        // Using imaging package to get round this.

	opt := imaging.AutoOrientation(true)  // create option for imaging.Decode()
	img, err := imaging.Decode(file, opt) // decoded file as image.Image

	if err != nil {
		fmt.Printf("Error decoding image %s\nError %v\n", fileName, err)
		return
	}

	image := canvas.NewImageFromImage(img)
	image.FillMode = canvas.ImageFillContain
	w.SetContent(image)
	w.ShowAndRun()
}

 - OS: "Linux Ubuntu"
 - Version: "18.04.5 LTS (Bionic Beaver)"
 - Go version: "go1.15 linux/amd64"
 - Fyne version: "v1.3.3-3-g6c9ed1b5"]

andydotxyz:
Issue was that we were reading aspect before setting it on the first frame.
Fixes #1266

### Checklist:

- [ ] Tests included. <- this was a rendering glitch, you can load the reproduction code to test, but it's not seen in test code
- [x] Tests all pass.

### DIFF
--- a/internal/painter/gl/draw.go
+++ b/internal/painter/gl/draw.go
@@ -56,13 +56,21 @@ func (p *glPainter) rectCoords(size fyne.Size, pos fyne.Position, frame fyne.Siz
 }
 
 func (p *glPainter) drawTextureWithDetails(o fyne.CanvasObject, creator func(canvasObject fyne.CanvasObject) Texture,
-	pos fyne.Position, size, frame fyne.Size, fill canvas.ImageFill, alpha, aspect float32, pad int) {
+	pos fyne.Position, size, frame fyne.Size, fill canvas.ImageFill, alpha float32, pad int) {
 
 	texture := getTexture(o, creator)
 	if texture == NoTexture {
 		return
 	}
 
+	aspect := float32(0)
+	if img, ok := o.(*canvas.Image); ok {
+		aspect = painter.GetAspect(img)
+		if aspect == 0 {
+			fyne.LogError("Missing aspect ratio for loaded image", nil)
+			aspect = 1 // fallback, should not occur
+		}
+	}
 	points := p.rectCoords(size, pos, frame, fill, aspect, pad)
 	vbo := p.glCreateBuffer(points)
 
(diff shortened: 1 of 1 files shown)


internal/painter/gl/draw.go
```diff
@@ -56,13 +56,21 @@ func (p *glPainter) rectCoords(size fyne.Size, pos fyne.Position, frame fyne.Siz
...
 	}
 
+	aspect := float32(0)
+	if img, ok := o.(*canvas.Image); ok {
+		aspect = painter.GetAspect(img)
+		if aspect == 0 {
+			fyne.LogError("Missing aspect ratio for loaded image", nil)
+			aspect = 1 // fallback, should not occur
+		}
+	}
```
stuartmscott, line 73: Should the zero check and fallback be moved outside of the type check and cast?
```suggestion
	if img, ok := o.(*canvas.Image); ok {
		aspect = painter.GetAspect(img)
	}
	if aspect == 0 {
		fyne.LogError("Missing aspect ratio for loaded image", nil)
		aspect = 1 // fallback, should not occur
	}
```

andydotxyz: Not in this case - see how 0 was passed in many times as a valid parameter.
0 is just not OK for images

stuartmscott: Ack

stuartmscott (approved):

PR #1436, opened 2020-10-21
Switch bundled resources to strings

heschik:
### Description:
The gopls team has received reports of high memory usage in the Fyne
project. The problem is the very large byte slice literals in the theme
package, which consume about 200MiB for the AST and another 200-400 for
type information.

While this is obviously a deficiency in gopls, it's not easy to fix
while maintaining full functionality. Changing the byte slices to
strings is straightforward.

As a further optimization, it might be good to change them to be *used*
as strings, since byte slices have to live on the heap and therefore
consume actual memory, whereas strings can be left on disk until
accessed.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
Is it safe to be encoding unknown []byte content as a string?

heschik:
I don't see why not. %q will escape non-printable characters, such as in the PNG at the top of bundled-icons.go. Do you have a specific concern?

andydotxyz:
Not sure why the tests are passing for you, but Travis is reporting failures on every OS...

heschik:
The failures appear to be in the CI setup, not the tests. The windows builds are complaining in setup:

```
/home/travis/.travis/functions: line 111: test: too many arguments
The command "if [ "$TRAVIS_OS_NAME" != "windows" ]
then
fi
" exited with 2.
```

and all of them are failing looking for `gocyclo`:

```
$ gocyclo -over 50 $GO_FILES
gocyclo: command not found
The command "gocyclo -over 50 $GO_FILES" exited with 127.
```

I don't think either are likely to be a problem with my change?

stuartmscott:
> which consume about 200MiB for the AST and another 200-400 for type information.

@heschik how much does this change reduce the footprint?

heschik:
Uh, that would probably have been useful to include. I should be clear that these numbers are estimates, since there's no perfect way to calculate the memory usage of an object graph. But I estimate that it goes from ~684MiB to 8MiB at rest. And if the user is actually modifying a package that `theme` depends on, it will be type checked many times over, creating a ton of garbage and ballooning the heap size.

(Again, I'm not saying any of this is good behavior on gopls' part, but Fyne has come up a couple of times and this was an easy fix.)

stuartmscott:
That is quite a significant delta! Thanks for looking into this @heschik, we appreciate your effort. 

The gocyclo is a known issue on master CI right now - we typically work in develop and merge into master from there. Speaking of which we are currently working on the 1.4 which will get released later on this month. Would it be okay if we held off on this change until after that release so we have time to vet it fully? - to Andy's question earlier about safety, we just want to make sure this doesn't put us in a hole further down the road.

heschik:
Sure, I'm not in any rush. @Jacalz may want to patch this in since they're the one having the problem.

andydotxyz:
Changed the base of this to try and trigger a cleaner CI run

andydotxyz:
@heschik is it possible for you to rebase this on latest develop?
The infrastructure fixes are in but so also are new icons which means this won't merge.

heschik:
Done.

andydotxyz:
Tested locally, it's possible the CPU reduced by 5% at runtime but hard to say. Certainly no worse, so this seems like a positive change

andydotxyz (approved):

PR #1437, opened 2020-10-21
Re-sign with keystore after build

### LINKED ISSUE #1359: fyne package command for packaging Android is signing the apk without using developer's cert
### Describe the bug:
fyne package command that creates apk is signing the apk with an unknown keystore and password. When I use jarsigner to sign my personal certificate and upload to Play Store, it complains that there are more than 1 cert is present.

Same certificate must be used during upgrading of the apk to a new version in Play store, and certificate ensure the originality of the creator of the apk. If fyne uses their own cert, the author of all fyne app will automatically be fyne instead of the individual developer

### To Reproduce:
fyne package -os android -appID com.johnkenedy.fourdanalysis -name "4D Analysis" -icon notes.png

above will produce an apk which is already sign. Fyne should provide options to include developer personal keystore and input password so that the apk versionCode can be upgraded later on in Play Console and make sure the app is owned by the developer.

### Screenshots:
No Screenshot

### Example code:
No Code

 - OS: Windows Machine to generate APK, APK is tested on Play Console
 - Version: 10
 - Go version: 1.13
 - Fyne version: 1.2

andydotxyz:
Ideally we would sign correctly first time but that is *way* out of scope due to the JKS
proprietary format and lack of APIs wrapping it.
We already need the devtools, so let's use them again.

Test plan: "fyne package -os android" to get the developer version then "go release -os android -keyStore ..." etc to get the signed version. compare.

Fixes #1359

### Checklist:

- [ ] Tests included. <- requires inspection extraction of resulting .apk
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/commands/release.go
+++ b/cmd/fyne/commands/release.go
@@ -60,8 +60,7 @@ func (r *releaser) afterPackage() error {
 	if util.IsAndroid(r.os) {
 		target := mobile.AppOutputName(r.os, r.packager.name)
 		apk := filepath.Join(r.dir, target)
-		err := r.zipAlign(apk)
-		if err != nil {
+		if err := r.zipAlign(apk); err != nil {
 			return err
 		}
 		return r.sign(apk)
@@ -72,8 +71,7 @@ func (r *releaser) afterPackage() error {
 
 func (r *releaser) beforePackage() error {
 	if util.IsAndroid(r.os) {
-		err := util.RequireAndroidSDK()
-		if err != nil {
+		if err := util.RequireAndroidSDK(); err != nil {
 			return err
 		}
 	}
--- a/cmd/fyne/commands/release.go
+++ b/cmd/fyne/commands/release.go
@@ -17,6 +17,8 @@ var _ Command = (*releaser)(nil)
 
 type releaser struct {
 	packager
+
+	keyStore string
 }
 
 // NewReleaser returns a command that can adapt app packages for distribution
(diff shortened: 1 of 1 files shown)


cmd/fyne/commands/release.go
```diff
@@ -56,20 +59,37 @@ func (r *releaser) Run(params []string) {
 func (r *releaser) afterPackage() error {
 	if util.IsAndroid(r.os) {
 		target := mobile.AppOutputName(r.os, r.packager.name)
-		return r.zipAlign(filepath.Join(r.dir, target))
+		apk := filepath.Join(r.dir, target)
+		err := r.zipAlign(apk)
+		if err != nil {
```
stuartmscott, line 64: nit:
```suggestion
		if err := r.zipAlign(apk); err != nil {
```

stuartmscott (commented):
At first glance it looks good, tho I'm not sure how to specify the keystore password, key name (keystore can store multiple keys) and key password - is this requested on the command line? How would this be passed through fyne-cross / docker to the fyne command?

cmd/fyne/commands/release.go
stuartmscott: Should this be added the contribution guide as "Declare variables in the smallest scope"?

andydotxyz: We need something as this has tripped up most of my recent committs.
However I don't think we can specify to always make it smallest or you will end up with very long, ugly lines.
Happy enough for you to add it in this context of error checking as that is quite common.

andydotxyz:
> is this requested on the command line?

Yes

> How would this be passed through fyne-cross / docker to the fyne command?

That will be added to fyne-cross before it's 1.0.0, that's why I held it back at 0.9.0 for now

cmd/fyne/commands/release.go
stuartmscott: Added, please take a look

stuartmscott (approved):

cmd/fyne/commands/release.go
andydotxyz: That is clear thanks. However my preference is that we don't add the "no" or "don't do this" versions if possible. If it stands alone with the "like this" version it feels overall a more positive doc.

stuartmscott: Fair enough, I'll make that change

PR #1439, opened 2020-10-21
Don't hover outside data in tree

andydotxyz:
Just don't show hover marks when cursor is outside data.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/table_test.go
```diff
@@ -117,6 +117,25 @@ func TestTable_Hovered(t *testing.T) {
 	defer w.Close()
 	w.Resize(fyne.NewSize(180, 180))
 
+	test.MoveMouse(w.Canvas(), fyne.NewPos(35, 100))
+
+	assert.Equal(t, -1, table.hoveredColumn)
+	assert.Equal(t, -1, table.hoveredRow)
+
+	test.AssertImageMatches(t, "table/nohovered.png", w.Canvas().Capture())
```
stuartmscott, line 125: nit: no_hover or not_hovered

widget/table.go
```diff
@@ -360,13 +360,22 @@ func (c *tableCells) hoverAt(pos fyne.Position) {
...
+	}
+
 	if c.t.moveCallback != nil {
 		c.t.moveCallback()
 	}
 }
 
 func (c *tableCells) hoverOut() {
-	c.t.hoveredRow = -1
+	c.t.hoveredColumn = -1
```
stuartmscott, line 378: Wow how'd I miss that? Did the test catch it?

widget/table_test.go
```diff
@@ -104,7 +104,7 @@ func TestTable_Hovered(t *testing.T) {
 	app.Settings().SetTheme(theme.LightTheme())
 
 	table := NewTable(
-		func() (int, int) { return 3, 5 },
+		func() (int, int) { return 2, 2 },
```
stuartmscott, line 107: What caused this change, was there too many cells for the canvas size, or is there something special about square tables?

widget/table_test.go
```diff
@@ -117,6 +117,25 @@ func TestTable_Hovered(t *testing.T) {
...
+	test.AssertImageMatches(t, "table/nohovered.png", w.Canvas().Capture())
+
+	table = NewTable(
+		func() (int, int) { return 3, 5 },
+		func() fyne.CanvasObject {
+			return NewLabel("placeholder")
+		},
+		func(row, col int, c fyne.CanvasObject) {
+			text := fmt.Sprintf("Cell %d, %d", row, col)
+			c.(*Label).SetText(text)
```
stuartmscott, line 134: nit: could be one line
```suggestion
			c.(*Label).SetText(fmt.Sprintf("Cell %d, %d", row, col))
```

stuartmscott (changes requested):
Couple small things, otherwise looks good!

widget/table_test.go
andydotxyz: It is not a change - this was inserted, I think you are seeing an artifact of how github presents the info

andydotxyz: The additional table (first one) is smaller so we can mouse out of it in the code below

stuartmscott: Ahh I see

stuartmscott (approved):

PR #1450, opened 2020-10-23
Tree scroll to selection

stuartmscott:
### Description:
Updated the scroller offset to ensure the selected node is in the viewport.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/tree.go
+++ b/widget/tree.go
@@ -32,6 +32,7 @@ type Tree struct {
 	OnBranchClosed     func(uid string)                                      // Called when a Branch is closed
 	OnSelectionChanged func(uid string)                                      // Called when the Node with the given Unique ID is selected.
 
+	scroller      *ScrollContainer
 	open          map[string]bool
 	branchMinSize fyne.Size
 	leafMinSize   fyne.Size
@@ -101,6 +102,7 @@ func (t *Tree) CreateRenderer() fyne.WidgetRenderer {
 	t.ExtendBaseWidget(t)
 	c := newTreeContent(t)
 	s := NewScrollContainer(c)
+	t.scroller = s
 	r := &treeRenderer{
 		BaseRenderer: widget.NewBaseRenderer([]fyne.CanvasObject{s}),
 		tree:         t,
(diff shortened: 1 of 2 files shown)


widget/tree.go
```diff
@@ -186,6 +188,40 @@ func (t *Tree) Selection() string {
...
+				y += m.Height
+			}
+		})
+		if y < t.scroller.Offset.Y {
+			t.scroller.Offset.Y = y
+		} else if y+size.Height > t.scroller.Offset.Y+t.scroller.Size().Height {
+			t.scroller.Offset.Y = y + size.Height - t.scroller.Size().Height
+		}
+		t.scroller.onOffsetChanged()
+		// TODO Setting a node as selected should open all parents if they aren't already
```
okratitan, line 223: For the future I think there may be a way that we can do this in the same pass that we are doing walkall - I.E.  Open the parents as we walk into them looking for selection and then close them if the item isn't there - other wise if the item is found you can leave the parents open.

widget/tree_internal_test.go
```diff
@@ -309,6 +309,39 @@ func TestTree_Selection(t *testing.T) {
...
+	tree.Refresh()                                            // Force layout
+	tree.SetSelection("A")
+	// Tree should scroll to the top to show A
+	assert.Equal(t, 0, tree.scroller.Offset.Y)
+
+	// Below
+	tree.scroller.Offset.Y = 0 // Showing "A" & "B"
+	tree.Refresh()             // Force layout
+	tree.SetSelection("F")
+	// Tree should scroll to the top to show F
```
okratitan, line 341: Nit: It should scroll to the bottom correct?

widget/tree.go
```diff
@@ -186,6 +188,40 @@ func (t *Tree) Selection() string {
...
+		var size fyne.Size
+		t.walkAll(func(uid string, branch bool, depth int) {
+			m := t.leafMinSize
+			if branch {
+				m = t.branchMinSize
+			}
+			if uid == t.selected {
+				found = true
+				size = m
+			} else if !found {
```
okratitan, line 203: Can we not return early when uid == t.selected and found is then set true  thus not needing the else if and removing the indention?

okratitan (commented):
Just a few things.  Looks good.

widget/tree_internal_test.go
stuartmscott: lol yup, copy/pasta, will fix

widget/tree.go
stuartmscott: Once we start walking the tree we can't abort once we find the node. The code would become longer;
```suggestion
				return
			}
			if !found {
```

okratitan: You wouldn't need the !found would you because if we made it this far and we returned early then we know the node wasn't found

widget/tree.go
```diff
@@ -186,6 +188,40 @@ func (t *Tree) Selection() string {
...
+					return
+				}
+				// If this is not the first item, add a divider
+				if y > 0 {
+					y += treeDividerHeight
+				}
+
+				y += m.Height
+			}
+		})
```
okratitan, line 216:
```suggestion
				size = m
				return
			}
			// Root node is not rendered unless it has been customized
			if t.Root == "" && uid == "" {
				// This is root node, skip
				return
			}
			// If this is not the first item, add a divider
			if y > 0 {
				y += treeDividerHeight
			}
			y += m.Height
		}
	})
```

widget/tree.go
stuartmscott: It would be possible to make the callback return bool so walk could determine whether to continue, though this PR would get larger

widget/tree.go
stuartmscott: Returning early does not stop walkAll from iterating through the tree

okratitan: I get that - but it makes for cleaner code here does it not?  Especially if it's going to iterate regardless?

widget/tree.go
okratitan: I was missing the part where found had to be persisted.  Gotcha.

okratitan (approved):

PR #1452, opened 2020-10-23
Feature/demo to tree

andydotxyz:
### Description:
Using the tree widget to provide a much nicer UI.
Avoids our nested tabs :).

This will want further work to flesh out content, this just ports over what we have already created.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz (commented):
This is looking great. I just have a couple smaller issues that I am seeing, though I think only one is related to this PR specifically. Would it be possible to make the settings window a bit bigger? It is too small and the theme select drop-down is partly out of view.
image

Jacalz (commented):
I also think it might be a good idea to get the images in the README updated when this lands.

andydotxyz:
I think those thoughts are sorted. I forgot that the settings window size was in fyne_demo since we added it to the menu. I just copied the size from the fyne_settings usage.

Jacalz (approved):
I think this looks great. Looks like there is a conflict that needs to be resolved before this lands though.

Jacalz (approved):
Looks good to me. Nice work :)

Issue #1395, opened 2020-10-11
Consistent Selection API


stuartmscott:
### Is your feature request related to a problem? Please describe:
Selection APIs in widgets and collections are inconsistent.

### Is it possible to construct a solution with the existing API?
n/a

### Describe the solution you'd like to see:
```go
Radio.ClearSelection()
Radio.SetSelectedOption(option string)
Radio.OnChanged(option string)

Select.ClearSelection()
Select.SelectedIndex() int
Select.SetSelectedOption(text string)
Select.SetSelectedIndex(index int)
Select.OnChanged(option string)

TabContainer.SelectedTab() *TabItem
TabContainer.SelectedTabIndex() int
TabContainer.SetSelectedTab(item *TabItem)
TabContainer.SetSelectedIndex(index int)

List.ClearSelection()
List.SelectedItem() int
List.SetSelectedItem(index int)
List.OnSelectionChanged(index int) // -1 == no selection

Table.ClearSelection()
Table.SelectedCell() (int, int)
Table.SetSelectedCell(row, col int)
Table.OnSelectionChanged(row, col int) // -1, -1 == no selection

Tree.ClearSelection()
Tree.SelectedNode() string
Tree.SetSelectedNode(uid string)
Tree.OnSelectionChanged(uid string) // tree.noSelection = "nothing" == no selection
```

Deprecation warning will be added in 1.4 and the APIs will be renamed in 2.0

andydotxyz:
`Radio` and `Select` export the selection as `Selected` hence the setter is currently `SetSelected`. If we Change the setters we might need to change Field/Getter as well.

I think that we might find that using the same API for collection/container and simple input widgets may not be the simplest solution.

stuartmscott:
I think there is value in extending this convention to other widgets;
```go
Radio.ClearSelection()
Radio.Selection() string
Radio.SetSelection(option string)
Radio.OnSelectionChanged(option string)

Select.ClearSelection()
Select.Selection() string
Select.SetSelection(text string)
Select.SetSelectionByIndex(index int)
Select.OnSelectionChanged(option string)

TabContainer.Selection() *TabItem
TabContainer.SelectionIndex() int
TabContainer.SetSelection(item *TabItem)
TabContainer.SetSelectionByIndex(index int)
TabContainer.OnSelectionChanged(item *TabItem)
```

andydotxyz:
Happy with the collection/container stuff.
[...]

PR #1456, opened 2020-10-23
List Selection

stuartmscott:
### Description:
Updates the List Selection API to match #1395

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz (commented):
I'll leave @okratitan to do the review, but I wanted to add some thoughts:

From the demo code II am pleased to see that this reads more cleanly.
However I think it still suffers from the struct-initialising 0 != -1 issue that we discussed?

stuartmscott:
> still suffers from the struct-initialising 0 != -1 issue that we discussed

yes, I'll update to use a slice. In future with multi select it may become a performance issue to iterate the slice for each list item, so we could consider switching to a map, but slice should be fine for now.

stuartmscott:
> From the demo code II am pleased to see that this reads more cleanly.

Me too, tho the code below will never be called as the list has 125 pre-selected, and there is no way to unselect

```go
list.OnUnselected = func(id widget.ListItemID) {
	label.SetText("Select An Item From The List")
	icon.SetResource(nil)
}
```

andydotxyz:
> effect of the code below will never be seen

But unselecting to select a new one is still going to execute the same callback right?

stuartmscott:
It will be called, but the label's text will be replaced soon after by the select callback, thus the user will never see "Select An Item From The List"

andydotxyz:
Kicked travis, looks like random error

stuartmscott:
Looks like another random error:
```
--- FAIL: TestEntry_SelectSnapLeft (0.10s)
    util.go:47: 
        	Error Trace:	util.go:47
        	            				test.go:32
        	            				entry_test.go:1048
        	Error:      	Not equal: 
        	Test:       	TestEntry_SelectSnapLeft
```

okratitan (approved):

PR #1457, opened 2020-10-24
Add folder support to widget.FileIcon

Jacalz:
This commit makes use of the new fyne.ListableURI to be able to show a folder.

### Description:
This commit makes use of the new fyne.ListableURI to be able to show a folder.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/fileicon.go
```diff
@@ -48,23 +48,9 @@ func (i *FileIcon) setURI(uri fyne.URI) {
...
-		i.resource = theme.FileTextIcon()
-	case "video":
-		i.resource = theme.FileVideoIcon()
-	default:
-		i.resource = theme.FileIcon()
-	}
-
 	i.URI = uri
 	i.cachedURI = nil
+	i.resource = i.setIcon(i.URI)
```
andydotxyz, line 53: This is not a setter, how about `lookupIcon`?

andydotxyz (changes requested):

widget/fileicon.go
Jacalz: Right, good point. Will change that. It was a setter initially :)

andydotxyz:
Thanks for the change, looks good.

However it is possible that URIs are passed in place of a ListableURI for a directory I think.
Should we attempt to check the mime type as well? It varies across systems, macOS uses "inode/directory" and I think Linux is "application/x-directory".

Happy to chat on Slack about this. The PR looks good but before accepting let's check if we want to cover the fallback case as well.

stuartmscott (changes requested):

okratitan (approved):

andydotxyz (approved):

PR #1459, opened 2020-10-24
Minor Tree Field Organization

stuartmscott:
### Description:
Move unexported fields below exported

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


widget/tree.go
```diff
@@ -36,6 +36,8 @@ type Tree struct {
 
 	branchMinSize fyne.Size
 	leafMinSize   fyne.Size
+
+	scroller *ScrollContainer
```
andydotxyz, line 40: Merging develop pushed this in out of order, should it be fixed too?

stuartmscott: I moved that as part of the conflict fix. Figured it wasn't related to open/selected state, or min sizes, so kept separated

andydotxyz (approved):

PR #1467, opened 2020-10-25
New Select API on Tree

andydotxyz:
### Description:
Moved to agreed selection API.
Remove the divider constant.
A few tidies as well.

See #1395 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


stuartmscott (changes requested):
Looking good so far. Now that there is TreeNodeID we can avoid the excessive "Unique ID" usage in the comments

stuartmscott (approved):

PR #1472, opened 2020-10-26
Update table to use agreed upon selection api.

### LINKED ISSUE #1395: Consistent Selection API

okratitan:
### Description:
Move table to use the agreed upon selection API

Fixes #1395 

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/table.go
+++ b/widget/table.go
@@ -15,48 +15,39 @@ const tableDividerThickness = 1
 // Declare conformity with Widget interface.
 var _ fyne.Widget = (*Table)(nil)
 
+// CellID is a type that represents a cell's position in a table based on it's row and column location.
+type CellID struct {
+	Row int
+	Col int
+}
+
 // Table widget is a grid of items that can be scrolled and a cell selected.
 // It's performance is provided by caching cell templates created with CreateCell and re-using them with UpdateCell.
 // The size of the content rows/columns is returned by the Length callback.
 type Table struct {
 	BaseWidget
 
-	Length     func() (int, int)
-	CreateCell func() fyne.CanvasObject
-	UpdateCell func(row, col int, template fyne.CanvasObject)
-	// OnSelectionChanged is called whenever the selection changes, such as through user tap.
-	// The row and col values will be -1 if no cell is currently selected.
-	OnSelectionChanged func(row, col int)
+	Length       func() (int, int)
+	CreateCell   func() fyne.CanvasObject
+	UpdateCell   func(id *CellID, template fyne.CanvasObject)
+	OnSelected   func(id *CellID)
+	OnUnselected func(id *CellID)
 
-	selectedRow, selectedColumn int
-	hoveredRow, hoveredColumn   int
-	cells                       *tableCells
-	moveCallback                func()
-	offset                      fyne.Position
-	scroll                      *ScrollContainer
+	selectedCell, hoveredCell *CellID
+	cells                     *tableCells
+	moveCallback              func()
+	offset                    fyne.Position
+	scroll                    *ScrollContainer
 }
 
 // NewTable returns a new performant table widget defined by the passed functions.
 // The first returns the data size in rows and columns, second parameter is a function that returns cell
 // template objects that can be cached and the third is used to apply data at specified data location to the
 // passed template CanvasObject.
-func NewTable(length func() (int, int), create func() fyne.CanvasObject, update func(int, int, fyne.CanvasObject)) *Table {
-	t := &Table{Length: length, CreateCell: create, UpdateCell: update, selectedRow: -1, selectedColumn: -1}
+func NewTable(length func() (int, int), create func() fyne.CanvasObject, update func(*CellID, fyne.CanvasObject)) *Table {
+	t := &Table{Length: length, CreateCell: create, UpdateCell: update, selectedCell: nil, hoveredCell: nil}
 	t.ExtendBaseWidget(t)
 	return t
-}
-
-// ClearSelection will clear any cell selection.
-func (t *Table) ClearSelection() {
-	t.selectedRow = -1
-	t.selectedColumn = -1
-	if t.moveCallback != nil {
-		t.moveCallback()
-	}
-
-	if t.OnSelectionChanged != nil {
-		t.OnSelectionChanged(-1, -1)
-	}
 }
 
 // CreateRenderer returns a new renderer for the table.
@@ -68,8 +59,6 @@ func (t *Table) CreateRenderer() fyne.WidgetRenderer {
 	rowMarker := canvas.NewRectangle(theme.PrimaryColor())
 	colHover := canvas.NewRectangle(theme.HoverColor())
 	rowHover := canvas.NewRectangle(theme.HoverColor())
-	t.hoveredColumn = -1
-	t.hoveredRow = -1
 
 	cellSize := t.templateSize().Add(fyne.NewSize(theme.Padding()*2, theme.Padding()*2))
 	t.cells = newTableCells(t, cellSize)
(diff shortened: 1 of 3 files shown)


widget/table.go
```diff
@@ -15,50 +15,41 @@ const tableDividerThickness = 1
 // Declare conformity with Widget interface.
 var _ fyne.Widget = (*Table)(nil)
 
+// CellID is a type that represents a cell's position in a table based on it's row and column location.
+type CellID struct {
```
stuartmscott, line 19: nit: Missing collection prefix
```suggestion
// TableCellID is a type that represents a cell's position in a table based on it's row and column location.
type TableCellID struct {
```

widget/table.go
```diff
@@ -90,41 +79,52 @@ func (t *Table) CreateRenderer() fyne.WidgetRenderer {
...
 
-// Selection will mark the specified cell (at row, col) to be marked as selected.
-func (t *Table) Selection() (row, col int) {
-	return t.selectedRow, t.selectedColumn
+// Select will mark the specified cell as selected.
+func (t *Table) Select(id *CellID) {
+	if id == nil {
+		return
+	}
+	t.Unselect(t.selectedCell)
```
stuartmscott, line 87: This will result in `moveCallback` being called twice

stuartmscott (commented):
Looking good so far

widget/table.go
```diff
@@ -15,50 +15,41 @@ const tableDividerThickness = 1
...
-
-	selectedRow, selectedColumn int
-	hoveredRow, hoveredColumn   int
-	cells                       *tableCells
-	moveCallback                func()
-	offset                      fyne.Position
-	scroll                      *ScrollContainer
+	Length       func() (int, int)
+	CreateCell   func() fyne.CanvasObject
+	UpdateCell   func(id *CellID, template fyne.CanvasObject)
```
andydotxyz, line 32: ID here is just a value param and cannot be nil, so don't use pointers

widget/table.go
```diff
@@ -15,50 +15,41 @@ const tableDividerThickness = 1
...
+	OnSelected   func(id *CellID)
+	OnUnselected func(id *CellID)
+
+	selectedCell, hoveredCell *CellID
```
andydotxyz, line 36: This works for pointers, so they default to nil well - nice pattern

widget/table.go
andydotxyz: Also it should not be called if the `selectedCell` is nil

andydotxyz (changes requested):
Nice work - a few nice tidies in there.
Mostly let's just not use struct pointers for data values

widget/table.go
okratitan: Resolved

widget/table.go
okratitan: Resolved

widget/table.go
okratitan: Resolved

widget/table.go
okratitan: Resolved

widget/table.go
```diff
@@ -15,50 +15,41 @@ const tableDividerThickness = 1
...
 }
 
 // NewTable returns a new performant table widget defined by the passed functions.
 // The first returns the data size in rows and columns, second parameter is a function that returns cell
 // template objects that can be cached and the third is used to apply data at specified data location to the
 // passed template CanvasObject.
-func NewTable(length func() (int, int), create func() fyne.CanvasObject, update func(int, int, fyne.CanvasObject)) *Table {
-	t := &Table{Length: length, CreateCell: create, UpdateCell: update, selectedRow: -1, selectedColumn: -1}
+func NewTable(length func() (int, int), create func() fyne.CanvasObject, update func(TableCellID, fyne.CanvasObject)) *Table {
+	t := &Table{Length: length, CreateCell: create, UpdateCell: update, selectedCell: nil, hoveredCell: nil}
```
stuartmscott, line 48: nit: selectedCell and hoveredCell will already be nil, no need to explicitly assign them to nil

widget/table.go
```diff
@@ -90,41 +79,51 @@ func (t *Table) CreateRenderer() fyne.WidgetRenderer {
 	return r
 }
 
-// Selection will mark the specified cell (at row, col) to be marked as selected.
-func (t *Table) Selection() (row, col int) {
-	return t.selectedRow, t.selectedColumn
+// Select will mark the specified cell as selected.
+func (t *Table) Select(id TableCellID) {
+	if t.OnUnselected != nil && t.selectedCell != nil {
+		t.OnUnselected(*t.selectedCell)
```
stuartmscott, line 85: nit: assign callback to local var for safety
```suggestion
	if f := t.OnUnselected; f != nil && t.selectedCell != nil {
		f(*t.selectedCell)
```

widget/table.go
```diff
@@ -90,41 +79,51 @@ func (t *Table) CreateRenderer() fyne.WidgetRenderer {
...
+	}
+	t.selectedCell = &id
+
+	t.scrollTo(t.selectedCell)
+	if t.moveCallback != nil {
+		t.moveCallback()
+	}
+
+	if t.OnSelected != nil {
+		t.OnSelected(id)
```
stuartmscott, line 95: nit: ditto

widget/table.go
```diff
@@ -90,41 +79,51 @@ func (t *Table) CreateRenderer() fyne.WidgetRenderer {
...
 
-	t.scrollTo(row, col)
 	if t.moveCallback != nil {
 		t.moveCallback()
 	}
 
-	if t.OnSelectionChanged != nil {
-		t.OnSelectionChanged(row, col)
+	if t.OnUnselected != nil {
+		t.OnUnselected(id)
```
stuartmscott, line 108: nit: ditto

stuartmscott (changes requested):

widget/table.go
okratitan: Yeah I thought about doing this too but just went with the pattern that already existed.  I'll update it.

stuartmscott: Thank you

okratitan: Resolved

widget/table.go
okratitan: Resolved

widget/table.go
okratitan: Resolved

widget/table.go
okratitan: Resolved

okratitan:
This also fixed the scrollTo typo where col was being used as the row scrollTo number for calculating both horizontal/vertical distance.  This also fixes safety for all cases of calling the exported methods.

widget/table.go
```diff
@@ -90,41 +79,51 @@ func (t *Table) CreateRenderer() fyne.WidgetRenderer {
 	return r
 }
 
-// Selection will mark the specified cell (at row, col) to be marked as selected.
-func (t *Table) Selection() (row, col int) {
-	return t.selectedRow, t.selectedColumn
+// Select will mark the specified cell as selected.
+func (t *Table) Select(id TableCellID) {
```
stuartmscott, line 83: The is a small optimization you could add here to check if id == selectedCell

widget/table.go
```diff
@@ -90,41 +79,51 @@ func (t *Table) CreateRenderer() fyne.WidgetRenderer {
...
+		f(id)
+	}
 }
 
-// SetSelection will mark the specified cell (at row, col) to be marked as selected.
-func (t *Table) SetSelection(row, col int) {
-	t.selectedRow = row
-	t.selectedColumn = col
+// Unselect will mark the cell provided by id as unselected.
+func (t *Table) Unselect(id TableCellID) {
```
stuartmscott, line 100: Similar optimization here, if t.selectedCell is already nil return early

widget/table.go
```diff
@@ -98,7 +101,9 @@ func (t *Table) Select(id TableCellID) {
 
 // Unselect will mark the cell provided by id as unselected.
 func (t *Table) Unselect(id TableCellID) {
-	t.selectedCell = nil
```
stuartmscott: This shouldn't have been removed

widget/table.go
okratitan: Resolved

widget/table.go
okratitan: Resolved

widget/table.go
okratitan: Resolved

widget/table.go
andydotxyz: It would be safer to just write `t.selectedCell = &id` - no need to dereference any data and no need to nil check

okratitan: Resolved

andydotxyz: I think I misread this first time round - I thought it was assigning.
As a comparison I might have been wrong... This equality check will only work when the pointers are the same - and given it's an exposed API this will not always be the case.
Sorry!

stuartmscott (dismissed):

andydotxyz (approved):

stuartmscott (approved):

PR #1475, opened 2020-10-27
Fix list to use dividers the same way tree/table do.

okratitan:
### Description:
Make list reuse dividers and place them correctly rather than using a border on list items.

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

#### Where applicable:


andydotxyz (changes requested):
I think this breaks some size calculations, I can no longer access the bottom of the list.

andydotxyz:
Image for the comment - the demo should have 1000 items, but alas:

Screenshot 2020-10-27 at 09 09 46

andydotxyz:
Also I noticed that list items are smaller than tree and table items again - we had this lined up nicely a while back

okratitan:
Fyne Demo accessing the last item in the list:
image

List and tree items being the same size  not smaller:
image


Are you sure you are up today with the PR?  Is something else breaking it?  It passes all test and works and looks as it should here

andydotxyz:
This is up to date - it appears that it's harder to repliate than I thought - but the issue is in there somewhere.
Scroll related it seems. Steps to replicate:

* Scroll to end of list bu dragging the scrollbar
* Using scroll wheel drag up around half a screen
* Scroll down again
* Notice that it no longer reaches the end of the list
* Repeat and you will see the error is cumulative.

andydotxyz:
> Also I noticed that list items are smaller than tree and table items again - we had this lined up nicely a while back

Apologies, this is not an issue - I must have been comparing develop where it is still broken.

widget/list.go
```diff
@@ -458,25 +457,32 @@ func (l *listLayout) Layout(objects []fyne.CanvasObject, size fyne.Size) {
...
 	}
 	l.layoutEndY = y
+	l.updateDividers()
 }
 
 func (l *listLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
 	if f := l.list.Length; f != nil {
 		return fyne.NewSize(l.list.itemMin.Width,
-			l.list.itemMin.Height*f())
+			l.list.itemMin.Height*f()+(f()-1)*listDividerHeight)
```
andydotxyz, line 472: You could avoid calling f() twice, and remove the second multiplication using:

```suggestion
			(l.list.itemMin.Height+listDividerHeight)*f()-listDividerHeight
```

andydotxyz (changes requested):
With the height fixed it looks like the indicators are now too short...

Screenshot 2020-10-27 at 14 43 01

stuartmscott (approved):

stuartmscott (approved):

andydotxyz (approved):

PR #1492, opened 2020-11-01
Feature/animation

### LINKED ISSUE #36: Add animation context for state transition
This API should allow simple method calls to be translated to animated changes (such as making a widget taller expand over a period)

andydotxyz:
### Description:
Start work on animations.
This is "generic" animations - i.e. not controlling a specific object or field.
They are registered on the driver which begins to tick through once some animations are running and will stop when they lapse.

Further smarter animations could be added, which may control objects directly.
The Animation.Start() is designed to encapsulate the complexities of starting them.

Fixes #36

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


canvas/animate.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"fyne.io/fyne"
+)
+
+const (
+	// DurationStandard is the time a standard interface animation will run.
+	DurationStandard = time.Millisecond * 300
+)
+
+func NewColorAnimator(start, stop color.Color, d time.Duration, fn func(color.Color)) *fyne.Animation {
```
Markcial, line 15: This way to expose the animation API is hard to scale, more properties more different methods to use.

Maybe it can be abstracted with some mapping using consts

i.e.:

```
type Property int
const (
  Unknown = iota
  Color
  X
  Y
  Width
  Height
 // Newer properties 
)

// A Property to value mapping table near the defined values

func NewAnimator(start, stop float64, prop Property, d time.Duration, fn func(v float64)) *fyne.Animation {
```

the logic of the function will make the mapping to the corresponding attribute and make it change over the time

Markcial:
Really wishing to see that feature on the library. I tested some libraries for a project we are working on and this one was the easiest one to compile on a win7 x86 capable binary, so congratulations for an amazing job, thanks

canvas/animate.go
andydotxyz: This sort of API should be added as well, though this relates to the object-specific animation so you need to pass the `CanvasObject` as well. In this situation it may require that the animation is tied to the canvas, so likely separate / additional to this API which allows more generic / flexible API.

andydotxyz: Just for clarity these methods will not multiply out with each property, there are only multiple methods right now for the different types to avoid animations having to do the calcuations of "completion" -> "colour value" for example.

andydotxyz: It is also possible that property-based animators might relate to the DataBinding API also landing in 2.0.

animation.go
```diff
@@ -0,0 +1,24 @@
...
+// rendered frame between time.Now() and the specified duration. The callback values start at 0.0 and
+// will be 1.0 when the animation completes.
+func NewAnimation(d time.Duration, fn func(float32)) *Animation {
+	return &Animation{Duration: d, Tick: fn}
+}
+
+// Start registers the animation with the application run-loop and starts its execution.
+func (a *Animation) Start() {
+	d := CurrentApp().Driver()
+	d.StartAnimation(a)
```
Jacalz, line 23: Just a small nitpick, but I think this could probably be on just one line without hurting readability. Please ignore this comment if you don't agree.

canvas/animate.go
```diff
@@ -0,0 +1,62 @@
+package canvas
+
+import (
+	"image/color"
+	"time"
+
+	"fyne.io/fyne"
+)
+
+const (
```
Jacalz, line 10: Just a nitpick again. If we only have one `const`, we should have it on one line. Please ignore this comment if you intend to add more constants with future work.

Jacalz (changes requested):
Just a few smaller nitpicks code wise. I don't think I have fiddled around with animations in other stuff to give any useful response in API structure, sorry.

I think the tab animation does look good, but is perhaps a bit too slow. It does, in my opinion, make the tab switching feel slow. It would also be great if the slide animation could handle being reversed when pressed for the second time (right now it starts the animation from the star position again).

Apart from those comments, I certainly think that the animations does look great and you seem to have done an amazing job. This is an amazing feature :)

canvas/animation.go
```diff
@@ -0,0 +1,60 @@
...
+	"fyne.io/fyne"
+)
+
+// DurationStandard is the time a standard interface animation will run.
+const DurationStandard = time.Millisecond * 300
+
+// NewColorAnimation sets up a new animation that will transition from the start to stop Color over
+// the specified Duration. The content of fn should apply the color values to an object and refresh it.
+// You should call Start() on the returned animation to start it.
+func NewColorAnimation(start, stop color.Color, d time.Duration, fn func(color.Color)) *fyne.Animation {
```
stuartmscott, line 16: Animating through the RGBA colour space gives a different result than through the HSLA colour space.
```suggestion
func NewRGBAColorAnimation(start, stop color.Color, d time.Duration, fn func(color.Color)) *fyne.Animation {
```

animation.go
```diff
@@ -0,0 +1,34 @@
+package fyne
+
+import "time"
+
+// AnimationCurve represents the different animation algorithms for animation timeline.
+type AnimationCurve int
+
+const (
+	// AnimationEaseInOut is the default easing, it starts slowly, accelerates to the middle and slows to the end.
+	AnimationEaseInOut AnimationCurve = iota
```
stuartmscott, line 10: What ease in & out is this? There are many (Sine, Cosine, Cubic, Quadratic) so the naming should be more explicit, this maybe helpful: https://easings.net/

internal/driver/glfw/animation.go
```diff
@@ -0,0 +1,77 @@
...
+			d.animationMutex.Unlock()
+			var newList []*anim
+			for _, a := range oldList {
+				if d.tickAnimation(a) {
+					newList = append(newList, a)
+				}
+			}
+			d.animationMutex.Lock()
+			d.animations = append(newList, d.animations...)
+			done = len(newList) == 0
```
stuartmscott, line 35: Is it possible newList is empty, but another animation was added in between? Or would that be executed by another goroutine? In which case an animation could end up being ticked by multiple threads
```suggestion
			done = len(d.animations) == 0
```

stuartmscott (commented):
One of the great things about Fyne is the ability to test a UI. Animations are extremely hard to get right so I would love to see more on the tooling, testing, and debugging side, either here or https://github.com/fyne-io/fyne/wiki/Animation

canvas/animation.go
andydotxyz: Good catch thanks. However I'm not sure about the name. How would you feel about `NewColorAnimationRGB` or  `NewColorRGBAnimation`? (I think we can omit the `A` as that is not part of the colour model).

animation.go
andydotxyz: I'm not certain that we need to provide all the different types, or include it in the naming. For portions of a second is it worth the technical understanding and/or the far slower maths behind the others?
If one type of EaseInOut is enough for iOS it could be good enough for us ;)

internal/driver/glfw/animation.go
andydotxyz: We only have 1 thread running - this loop is handled by the driver.
New animations cannot be added to this list during a tick because of the lock - this loop takes a snapshot of `d.animations` and works with it. Line 34 could probably just be `d.animations = newList` now that I think about it - it just left us open to letting things drop through if the locking changed.

canvas/animation.go
stuartmscott: I think `A` should be kept to be consistent with Color Picker and image.RGBA. `NewRGBAAnimation` & `NewHSLAAnimation` would be fine with me

animation.go
stuartmscott: There are many things I disagree with iOS on, so "good enough for them, good enough for us" doesn't fly with me. If there is no way for developers to set their own curves then we should provide the common ones, which don't all need to be added now, but if they will be added in future then the current naming should be explicit enough so it doesn't need to renamed when the others land.
It could be a better design to have the curve be a field of `Animation` so devs can provide their own if desired, or use a prebuilt one we provide.

internal/driver/glfw/animation.go
stuartmscott: > New animations cannot be added to this list during a tick because of the lock

I disagree, the lock is unlocked on line 26 and relocked on line 33, so an animation could be added between them.

andydotxyz: Hah, I missed that. I really should just work on one feature at a time

animation.go
andydotxyz: Thanks, this triggerd a thought about how this could be done better.
Update provides the same API but allows for custom animation curves - let me know what you think.

canvas/animation.go
andydotxyz: The point of putting `Color` earlier in the name as I illustrated was to make it more searchable and the methods would be grouped in ordering. Was your providing of two different names objecting to the suggestion?

canvas/animation.go
```diff
@@ -0,0 +1,60 @@
...
+func NewSizeAnimation(start, stop fyne.Size, d time.Duration, fn func(fyne.Size)) *fyne.Animation {
+	return &fyne.Animation{
+		Duration: d,
+		Tick: func(done float32) {
+			fn(fyne.NewSize(scaleInt(start.Width, stop.Width, done), scaleInt(start.Height, stop.Height, done)))
+		}}
+}
+
+func scaleColor(start uint32, end uint32, done float32) uint8 {
+	diff := int(end) - int(start)
```
stuartmscott, line 51: Not sure these casts are needed
```suggestion
	diff := end - start
```

animation.go
```diff
@@ -0,0 +1,61 @@
...
+	AnimationEaseIn = animationEaseIn
+	// AnimationEaseOut starts at speed and slows to the end.
+	AnimationEaseOut = animationEaseOut
+	// AnimationLinear is a linear mapping for animations that progress uniformly through their duration.
+	AnimationLinear = animationLinear
+)
+
+// Animation represents an animated element within a Fyne canvas.
+// These animations may control individual objects or entire scenes.
+type Animation struct {
```
stuartmscott, line 24: Random thought; it could be useful to be able to specify a delay before starting the animation, and (either another field or the same field) before repeating the animation

canvas/animation.go
stuartmscott: As per the call let's go with `NewColorRGBAAnimation` & `NewColorHSLAAnimation`

animation.go
stuartmscott, line 24: Another random thought; it could be useful to have `AnimationGroup` for running multiple animations in serial, or in parallel

animation_test.go
```diff
@@ -0,0 +1,17 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAnimationLinear(t *testing.T) {
+	assert.Equal(t, float32(0.25), animationLinear(0.25))
+}
+
+func TestAnimationEaseInOut(t *testing.T) {
```
stuartmscott, line 13: Missing `TestAnimationEaseIn` and `TestAnimationEaseOut`

cmd/fyne_demo/tutorials/animate.go
```diff
@@ -0,0 +1,79 @@
...
+	return fyne.NewContainerWithoutLayout(label1, label2, label3, label4, box1, box2, box3, box4, start)
+}
+
+func makeAnimationCurveItem(label string, curve fyne.AnimationCurve, yOff int) (
+	text *widget.Label, box fyne.CanvasObject, anim *fyne.Animation) {
+	text = widget.NewLabel(label)
+	text.Alignment = fyne.TextAlignCenter
+	text.Resize(fyne.NewSize(380, 30))
+	text.Move(fyne.NewPos(0, yOff))
+	box = canvas.NewRectangle(theme.TextColor())
```
stuartmscott, line 69: Consider using a circle with a radius slightly smaller than the gauge - it may look nicer :)

internal/driver/glfw/driver.go
```diff
@@ -71,6 +75,17 @@ func (d *gLDriver) Run() {
 	d.runGL()
 }
 
+func (d *gLDriver) StartAnimation(a *fyne.Animation) {
+	d.animationMutex.Lock()
+	defer d.animationMutex.Unlock()
+	wasStopped := len(d.animations) == 0
+
+	d.animations = append(d.animations, &anim{a, time.Now(), time.Now().Add(a.Duration)})
+	if wasStopped {
```
stuartmscott, line 84: Could save a var & line here :)
```suggestion
	if len(d.animations) == 1 {
```

canvas/animation.go
andydotxyz: it is, because it coule be negative but a uint32 cannot be

animation.go
andydotxyz: The animation will not start until you call Start() - if the developer wants to delay they should probably delay calling Start().
Delay on repeating probably relates to enhanced repeat like AutoReverse and other ideas I would like to work on once basics are in.

andydotxyz: That's a nice idea - Can we open this in a ticket to track, I'd like to delay grouping until we have explored individual for a while

internal/driver/glfw/driver.go
andydotxyz: True. I just implemented this then felt the need to write "// was Stopped" at the end, so I'm not convinced it's clearer.
The `== 0` is clearer and the variable name marks it's significance in my opinion.

cmd/fyne_demo/tutorials/animate.go
andydotxyz: I found that the boxes square edges worked more clearly to highlight the difference in position because the corners are a clearer contrast than the circles.
Also not sure I want them to look like radios or sliders ;)

animation_test.go
andydotxyz: thanks

cmd/fyne_demo/tutorials/animate.go
stuartmscott: Fair enough

canvas/animate.go
andydotxyz: If you're happy can you mark this conversation resolved please @markcial?

stuartmscott (dismissed):

Jacalz (approved):
Looks good to me 🙂

andydotxyz:
> I would love to see more on the tooling, testing, and debugging side

Started gathering notes on what remains to be done on the wiki Animation page.
Also with the requests that came in via Twitter

Jacalz:
I updated the wiki page and added a note about animating the select widget as well. https://github.com/fyne-io/fyne/search?q=todo+animation&type=

andydotxyz:
Good catch. Select and menus etc were discussed later on the call yesterday and probably require we implement object clipping first. We can't just lay out at every different size from 0 to size - it could be slow and it will be jarring.
We need to draw at the correct size, clipped, then expose the clipped area until complete. (I think)

Jacalz (changes requested):
Hmm, just a final thought here. Could the tab animations perhaps be changed to fade in and out the primary colour instead of moving it? I think the animation makes the interface feel painfully slow.
slow-tabs

andydotxyz:
> Could the tab animations perhaps be changed to fade in and out

Why would we not use the same animations as Material Design show?
If it's too slow then we can make it a shorter duration...
https://material.io/components/tabs

Jacalz:
> > Could the tab animations perhaps be changed to fade in and out
> 
> Why would we not use the same animations as Material Design show?
> If it's too slow then we can make it a shorter duration...
> https://material.io/components/tabs

Ah, I didn't know that it was part of the MD spec. Then we should probably make the duration shorter.

widget/tabcontainer.go
```diff
@@ -380,8 +381,25 @@ func (r *tabContainerRenderer) moveSelection() {
...
+		r.animation = canvas.NewPositionAnimation(r.underline.Position(), underlinePos, canvas.DurationStandard, func(p fyne.Position) {
+			r.underline.Move(p)
+			canvas.Refresh(r.underline)
+			if p == underlinePos {
+				r.animation = nil
+			}
+		})
+		r.animation.Start()
+
+		canvas.NewSizeAnimation(r.underline.Size(), underlineSize, canvas.DurationStandard, func(s fyne.Size) {
```
Jacalz, line 398:
```suggestion
		canvas.NewSizeAnimation(r.underline.Size(), underlineSize, time.Millisecond * 100, func(s fyne.Size) {
```
I did some testing and 100 milliseconds is what I would consider the best.

Jacalz (changes requested):

Jacalz (approved):
The shortened transition works for me. Thanks :)

Jacalz (approved):

animation.go
```diff
@@ -0,0 +1,61 @@
...
+import "time"
+
+// AnimationCurve represents an animation algorithm for calculating the progress through a timeline.
+// Custom animations can be provided by implementing the "func(float32) float32" definition.
+// The input parameter will start at 0.0 when an animation starts and travel up to 1.0 at which point it will end.
+// A linear animation would return the same output value as is passed in.
+type AnimationCurve func(float32) float32
+
+var (
+	// AnimationEaseInOut is the default easing, it starts slowly, accelerates to the middle and slows to the end.
```
toaster, line 12: AFAICS, the default is `nil` which is translated to `AnimationLinear` by the GLFW driver. Therefore, this seems to be wrong.

toaster (commented):
Please see my comments.

Also: Why is the animation handling driver specific? What would be the difference on mobile?

animation.go
andydotxyz: Great catch thanks

PR #1495, opened 2020-11-02
Use file ops for file:// and document ops for content://

stuartmscott:
### Description:
This enables apps to access app-specific directories using the file:// scheme, whilst still allowing apps to access other directories using the Document API.

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/gomobile/android.c
+++ b/internal/driver/gomobile/android.c
@@ -123,7 +123,6 @@ void* openStream(uintptr_t jni_env, uintptr_t ctx, char* uriCstr) {
 	jthrowable loadErr = (*env)->ExceptionOccurred(env);
 
 	if (loadErr != NULL) {
-		(*env)->ExceptionDescribe(loadErr);
 		(*env)->ExceptionClear(env);
 		return NULL;
 	}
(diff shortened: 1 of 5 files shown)


internal/driver/gomobile/android.c
```diff
@@ -155,93 +158,183 @@ void closeStream(uintptr_t jni_env, uintptr_t ctx, void* stream) {
...
 
-	char *str = getString(jni_env, ctx, type);
-	return strcmp(str, "vnd.android.document/directory") == 0;
+		if (type == NULL) {
+			return false;
+		}
+
+		char *str = getString(jni_env, ctx, type);
+		return strcmp(str, "vnd.android.document/directory") == 0;
+	}
```
andydotxyz, line 223: I think we should log here that the URI scheme was not known - there will be no other errors logged to help figure such a corner case out.

internal/driver/gomobile/android.c
```diff
@@ -155,93 +158,183 @@ void closeStream(uintptr_t jni_env, uintptr_t ctx, void* stream) {
...
 }
 
 char* uriList(uintptr_t jni_env, uintptr_t ctx, char* uriCstr) {
-	JNIEnv *env = (JNIEnv*)jni_env;
-	jobject resolver = getContentResolver(jni_env, ctx);
-	jstring uriStr = (*env)->NewStringUTF(env, uriCstr);
-	jobject uri = parseURI(jni_env, ctx, uriCstr);
+	if (hasPrefix(uriCstr, "file://")) {
+
+		// Get file path from URI
```
andydotxyz, line 230: I think it would be easier to read if this was now split to two sub-methods, `listFileURI` and `listContentURI` to avoid this new 120-line method

andydotxyz (changes requested):
Just structural stuff, this is a great addition

internal/driver/gomobile/android.c
stuartmscott: Yup good idea, should this be log_fatal, or just log_info?

internal/driver/gomobile/android.c
stuartmscott: Done

andydotxyz (approved):

PR #1496, opened 2020-11-03
Test: markup renderer

toaster:
This PR is my proposal on how to test rendered output without images.
It introduces a markup renderer which creates a markup representation of the render tree.
Common icons and colours are displayed by name instead by filename or colour values.

This is a draft for now because there are a lot of tests that still need to be converted and maybe there might be adjustments to the markup renderer during this process.


andydotxyz:
This is totally cool :).
I wonder though if it could be done without the `heredoc.Doc` dependency?
If we could just pass raw strings it would be much neater, perhaps this calls for another test helper?

benet37 (approved):

toaster:
> I wonder though if it could be done without the `heredoc.Doc` dependency?
> If we could just pass raw strings it would be much neater, perhaps this calls for another test helper?

It should be possible. You’re thinking of a helper along the lines:
```go
func AssertMarkup(t *testing.T, expected string, actual string)
```
or even:
```go
func AssertMarkup(t *testing.T, expected string, c fyne.Canvas)
```

andydotxyz:
> It should be possible. You’re thinking of a helper along the lines:

Yeah looks good

toaster:
This PR is ready to land, I think.
It contains a markup renderer for tests and an assertion method to access it.
I switched all public widget specs and some other to markup based instead of image based.

toaster:
This PR now included the (stripped down version of) #1403.
This was needed to ensure distinguishable colours, which became necessary because I had to switch to map based lookup to reduce cyclomatic complexity.

test/testtheme.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 var (
-	red   = &color.RGBA{R: 255, G: 0, B: 0, A: 255}
```
andydotxyz, line 11: Is this change required? I'm not sure what it imacts

test/theme.go
```diff
@@ -0,0 +1,167 @@
...
+// Theme returns a theme useful for image based tests.
+func Theme() fyne.Theme {
+	if defaulTheme == nil {
+		defaulTheme = &configurableTheme{
+			background:         color.NRGBA{R: 0x44, G: 0x44, B: 0x44, A: 0xff},
+			bold:               bold,
+			boldItalic:         bolditalic,
+			button:             color.NRGBA{R: 0x33, G: 0x33, B: 0x33, A: 0xff},
+			disabledButton:     color.NRGBA{R: 0x22, G: 0x22, B: 0x22, A: 0xff},
+			disabledIcon:       color.NRGBA{R: 0xff, G: 0x00, B: 0x00, A: 0xff}, // deprecated: outstanding red variant to make it visible
```
andydotxyz, line 49: I think maybe "bright" not "outstanding"

test/markup_renderer.go
```diff
@@ -0,0 +1,501 @@
...
+	r.writeIndent()
+	r.writeCloseTag("canvas")
+}
+
+func (r *markupRenderer) writeCanvasObject(obj fyne.CanvasObject, _, clipPos fyne.Position, clipSize fyne.Size) bool {
+	attrs := map[string]*string{}
+	r.setPosAttr(attrs, "pos", obj.Position())
+	r.setSizeAttr(attrs, "size", obj.Size())
+	if clipPos != fyne.NewPos(0, 0) || clipSize != fyne.NewSize(math.MaxInt32, math.MaxInt32) {
+		r.setRectAttr(attrs, "clip", clipPos, clipSize)
```
andydotxyz, line 229: I don't think this should be rendered - it is a property of the driver / renderer not the widget.
Also it gets in the way of this markup being writable manually...

andydotxyz (changes requested):

test/testtheme.go
toaster: With the former colour the background icon of dialogs was not visible at all. With this it is at least faintly visible.

andydotxyz:
It's failing static analysis:
```
/home/runner/work/fyne/fyne/test/markup_renderer.go
  (129, 26)  U1000  func (*markupRenderer).setRectAttr is unused
```

andydotxyz (dismissed):
This resolves my comments, thanks. I would like to get @stuartmscott feedback as well, I think he returns next week.

test/markup_renderer.go
```diff
@@ -0,0 +1,497 @@
...
+
+func (r *markupRenderer) writeIndent() {
+	for i := 0; i < r.indentation; i++ {
+		r.w.WriteRune('\t')
+	}
+}
+
+func (r *markupRenderer) writeImage(i *canvas.Image, attrs map[string]*string) {
+	r.setStringAttr(attrs, "file", i.File)
+	r.setResourceAttr(attrs, "rsc", i.Resource)
```
stuartmscott, line 297: It took me a while to figure this out, I think we can afford the storage space of using full names; `resource`

test/markup_renderer.go
```diff
@@ -0,0 +1,497 @@
...
+	r.setBoolAttr(attrs, "img", i.Image != nil)
```
stuartmscott, line 298: Ditto; `image`

test/markup_renderer.go
```diff
@@ -0,0 +1,497 @@
...
+		r.writeCloseTag("overlay")
+	}
+	r.indentation--
+	r.writeIndent()
+	r.writeCloseTag("canvas")
+}
+
+func (r *markupRenderer) writeCanvasObject(obj fyne.CanvasObject, _, _ fyne.Position, _ fyne.Size) bool {
+	attrs := map[string]*string{}
+	r.setPosAttr(attrs, "pos", obj.Position())
```
stuartmscott, line 225: Ditto; `position`

dialog/base_test.go
```diff
@@ -7,23 +7,24 @@ import (
...
 
 	d := NewCustom("Title", "OK", label, w)
 
 	d.Show()
-	test.AssertImageMatches(t, "dialog-custom-dark.png", w.Canvas().Capture())
+	test.AssertImageMatches(t, "dialog-custom-default.png", w.Canvas().Capture())
 
-	test.ApplyTheme(t, theme.LightTheme())
-	test.AssertImageMatches(t, "dialog-custom-light.png", w.Canvas().Capture())
+	test.ApplyTheme(t, test.NewTheme())
```
stuartmscott, line 28: Maybe we should go all in and call this `test.UglyTheme()`

stuartmscott (commented):
Few naming nits, otherwise looks good!

Have you considered how this can be used in the reverse? This would allow developers to define a UI in markup and then convert it into a canvas object. The main challenges I foresee here are mapping resource name to bundled resource, and defining the callbacks eg button.OnTapped, list.Length, list.CreateItem, et al.

dialog/base_test.go
toaster: > Have you considered how this can be used in the reverse? This would allow developers to define a UI in markup and then convert it into a canvas object. The main challenges I foresee here are mapping resource name to bundled resource, and defining the callbacks eg button.OnTapped, list.Length, list.CreateItem, et al.

No, I didn’t. And, honestly, I am not a fan of defining UI via markup. I prefer to write the code.

toaster:
> Have you considered how this can be used in the reverse? This would allow developers to define a UI in markup and then convert it into a canvas object. The main challenges I foresee here are mapping resource name to bundled resource, and defining the callbacks eg button.OnTapped, list.Length, list.CreateItem, et al.

No, I didn’t. And, honestly, I am not a fan of defining UI via markup. I prefer to write the code.

dialog/base_test.go
toaster: We might, but not in this PR :).

test/markup_renderer.go
andydotxyz: I think the #fyne-votes channel is showing that "res" might be more liked

test/markup_renderer.go
andydotxyz: We use "pos" already - i.e. fyne.NewPos

test/markup_renderer.go
andydotxyz: I think that img is a pretty reasonable abbreviation. It only removes 2 letters and "file" is 4 so I don't have strong feelings either way.

toaster: The vote has clearly favoured the abbreviations (no one voted for full names).

test/markup_renderer.go
toaster: The vote has clearly favoured the abbreviations (no one voted for full names).

test/markup_renderer.go
toaster: `rsc` and `res` both got 2 votes. Therefore, I won’t change this 😜.

test/test.go
```diff
@@ -32,6 +34,18 @@ func AssertImageMatches(t *testing.T, masterFilename string, img image.Image, ms
...
 
+// AssertRendersToMarkup asserts that the given canvas renders the expected markup.
+// The expected markup is stripped by leading common whitespace retaining the relative indentation
+// and thus matching the indentation of the output of the used markup renderer.
+//
+// Be aware, that the indentation has to use tab characters ('\t') instead of spaces.
+// Every element starts on a new line indented one more than its parent.
+// Closing elements stand on their own line, too, using the same indentation as the opening element.
+// The only exception to this are text elements which do not contain line breaks unless the text includes them.
+func AssertRendersToMarkup(t *testing.T, expected string, c fyne.Canvas, msgAndArgs ...interface{}) bool {
```
andydotxyz, line 45: Can you please add
```
//
// Since: 2.0.0
```

andydotxyz (changes requested):
Merge looks good.
However since this was opened we added the "Since:" tag - if you could add that to any new public APIs (I think it's just one) then we're good to go

test/test.go
toaster: And If you pick this for a 1.4.x or a 1.5.0 you’ll adjust it?

andydotxyz: Sure thing :)

andydotxyz (approved):

Issue #1498, opened 2020-11-03
Icons aren't generated with the appropriate dimensions causing failure do distribute app


MagicalTux:
### Describe the bug:

When uploading the generated ipa file to Apple, the following errors appear:

    [2020-11-03 22:52:08 JST] <main> ERROR: ERROR ITMS-90023: "Missing required icon file. The bundle does not contain an app icon for iPad of exactly '76x76' pixels, in .png format for iOS versions >= 7.0. To support older operating systems, the icon may be required in the bundle outside of an asset catalog. Make sure the Info.plist file includes appropriate entries referencing the file. See https://developer.apple.com/documentation/bundleresources/information_property_list/user_interface"
    [2020-11-03 22:52:08 JST] <main> ERROR: ERROR ITMS-90023: "Missing required icon file. The bundle does not contain an app icon for iPad of exactly '152x152' pixels, in .png format for iOS versions >= 7.0. To support older operating systems, the icon may be required in the bundle outside of an asset catalog. Make sure the Info.plist file includes appropriate entries referencing the file. See https://developer.apple.com/documentation/bundleresources/information_property_list/user_interface"

The bundle indeed contains files matching the required sized (Icon_76.png, Icon_152.png) but those files are actually 120x120 pixels images.

The issue comes from `cmd/fyne/commands/package-mobile.go` on line 37, where `sips` is invoked with a -Z value fixed to 120 while it should actually be `size`.

### To Reproduce:
Steps to reproduce the behaviour:
1. Build ipa file
2. Upload with `xcrun altool --upload-app`

### Device (please complete the following information):
 - OS: MacOS
 - Version: 10.15.7
 - Go version: 1.14.10
 - Fyne version: 1.4.0

andydotxyz:
Thanks for the fix. It's on `develop` now and will be in the 1.4.1 release

PR #1500, opened 2020-11-03
Fixes #1497 #1498 Skip build of darwin/arm and pass correct parameter to sips

### LINKED ISSUE #1497: Go1.15 does not support darwin/arm anymore for iOS
### Describe the bug:
Building for iOS with go1.15+ will fail with:

go [-e -json -compiled=true -test=false -export=false -deps=false -find=true -tags=ios -- .]: exit status 2: cmd/go: unsupported GOOS/GOARCH pair darwin/arm

### To Reproduce:
1. build with `-os ios` with go1.15+

 - OS: MacOS
 - Version: 10.15.7
 - Go version: 1.15.3
 - Fyne version: 1.4.0

### Solving the issue
In file `cmd/fyne/internal/mobile/build.go` on line 155:

    targetArchs = []string{"arm", "arm64"}

Could be replaced with:

    targetArchs = []string{"arm64"}

If 32bit support is still required, adding a command line flag to enable it or define the target archs would be nice.

MagicalTux:
### Description:

Fixes #1497 #1498: Skip build of darwin/arm and pass correct parameter to sips

Note that I haven't been able to test this due to the way go modules work. I'm not sure how to get this to work locally, and I am not sure if removing `darwin/arm` has any impact on anything else.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/commands/package-mobile.go
+++ b/cmd/fyne/commands/package-mobile.go
@@ -34,6 +34,6 @@ func (p *packager) packageIOS() error {
 }
 
 func copyResizeIcon(size, dir string) error {
-	cmd := exec.Command("sips", "-o", dir+"/Icon_"+size+".png", "-Z", "120", dir+"/Icon.png")
+	cmd := exec.Command("sips", "-o", dir+"/Icon_"+size+".png", "-Z", size, dir+"/Icon.png")
 	return cmd.Run()
 }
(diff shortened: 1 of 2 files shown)


andydotxyz:
Thanks so much for looking into this. Unfortunately the arm fix needs a bigger change as the `allArchs` is iterated to get the architecture list. Additionally "arm" is used in `bind.go` to look up package configuration.

If you don't mind I would suggest you strip this back to the sips fix which is great and I can open a bigger PR for the arch fix which you could then help us test?

PR #1502, opened 2020-11-03
Remove 32bit from iOS apps, Go 1.15 does not allow it

### LINKED ISSUE #1497: Go1.15 does not support darwin/arm anymore for iOS

andydotxyz:
Also this was stopped by Apple many years ago (iPhone 5s)

Fixes #1497

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/mobile/bind.go
+++ b/cmd/fyne/internal/mobile/bind.go
@@ -65,14 +65,14 @@ func writeFile(filename string, generate func(io.Writer) error) error {
 	return generate(f)
 }
 
-func packagesConfig(targetOS string) *packages.Config {
+func packagesConfig(targetOS, targetArch string) *packages.Config {
 	config := &packages.Config{}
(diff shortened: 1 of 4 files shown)


andydotxyz:
Fixed manifest, upload validated correctly at this end now

stuartmscott (approved):

PR #1506, opened 2020-11-04
Add missing LoadResourceFromURI

andydotxyz:
Missed this with the other storage / uri work that was going on.

Needed to be in storage rather than root, but I expect more such things to move here

More to come in the future once we figure out http/https standard providers etc

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/storage/resource.go
+++ b/storage/resource.go
@@ -1,0 +1,25 @@
+package storage
+
+import (
+	"io/ioutil"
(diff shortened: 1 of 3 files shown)


Jacalz (approved):
Looks good to me. This is a very useful addition :+1:

PR #1508, opened 2020-11-05
Update list.go

ruanwenfeng:
This min value is equal to -1 when minimized.  and line 192 panic: runtime error: index out of range [-1]

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -185,7 +185,7 @@ func (l *listRenderer) Layout(size fyne.Size) {
 	if l.visibleItemCount == 0 {
 		return
 	}
-	min := int(math.Min(float64(length), float64(l.visibleItemCount)))
+	min := int(math.Max(math.Min(float64(length), float64(l.visibleItemCount)), 1))
 	if len(l.children) > min {
 		for i := len(l.children); i >= min; i-- {
 			l.itemPool.Release(l.children[i-1])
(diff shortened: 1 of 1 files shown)


andydotxyz:
It is not clear which value is -1 and when. Is it possible that your data callback `Length()` is returning -1 incorrectly?

ruanwenfeng:
> It is not clear which value is -1 and when. Is it possible that your data callback `Length()` is returning -1 incorrectly?

@andydotxyz the line I modified, the min variable is qual to -1  when the software is minimized. The list widget needs to be displayed on the interface.

andydotxyz:
Yes, but as that is a result of a Min() calculation it might be good to know the cause - which of those values was incorrect?
It will influence what makes a good fix for the issue.

ruanwenfeng:
> Yes, but as that is a result of a Min() calculation it might be good to know the cause - which of those values was incorrect?
> It will influence what makes a good fix for the issue.

@andydotxyz Yes, I agree with you. It seems to be caused by the l.visibleItemCount variable.  we cang change the == in line 185 to <=

andydotxyz:
> change the == in line 185 to <=

Perfect :)

andydotxyz (approved):

andydotxyz:
Thanks for the contribution, it will be in 1.4.1 release

PR #1516, opened 2020-11-09
Enable mobile tests on Travis and fix any issues

toaster:
The mobile tests did not run properly on Travis because the `-tags ci` of the script overrides the `GOFLAGS=-tags=mobile` from the env.


andydotxyz (approved):
Thanks for tidying this up.
The settings code could do with a little love - there are separate issues for that I think.

PR #1517, opened 2020-11-09
Maintain object slice

stuartmscott:
### Description:
Maintain a slice of canvas objects to ensure consistent ordering.

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/tree.go
+++ b/widget/tree.go
@@ -514,8 +514,6 @@ func (r *treeContentRenderer) MinSize() (min fyne.Size) {
 }
 
 func (r *treeContentRenderer) Objects() []fyne.CanvasObject {
-	r.treeContent.propertyLock.RLock()
-	defer r.treeContent.propertyLock.RUnlock()
 	return r.objects
 }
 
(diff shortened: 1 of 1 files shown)


widget/tree.go
```diff
@@ -508,17 +513,10 @@ func (r *treeContentRenderer) MinSize() (min fyne.Size) {
...
-	objects = r.dividers
-	for _, b := range r.branches {
-		objects = append(objects, b)
-	}
-	for _, l := range r.leaves {
-		objects = append(objects, l)
-	}
-	r.treeContent.propertyLock.RUnlock()
-	return
+	defer r.treeContent.propertyLock.RUnlock()
```
toaster, line 518: I’d suspect that we do no longer need the lock on tree content here, don’t we?

toaster (changes requested):

toaster (approved):

widget/tree.go
pymq: @toaster @stuartmscott I think we still need read lock here because it is access to a referenced value which is not atomic. Or you meant that `Objects()` and `Layout()` in practice are not called in parallel?

stuartmscott: That's a good point @pymq, thanks for raising it. I think in practice they aren't called in parallel, and even if they were `Objects()` would be called again after `Layout()` so any inconsistency would be resolved. Having said that I would not be opposed to adding the read locks to be safe. @toaster what do you think?

Issue #135, opened 2019-02-22
Add a data driven / MVVM capability


andydotxyz:
This is needed to drive proper list / table capabilities.
Also possible data binding like more advanced forms...

This will require a fair bit of discussion and desigm (document to be linked)

andydotxyz:
Tagged with design but this is an architecture type design requirement.

andydotxyz:
This is, at least in basic form, landed on develop branch

andydotxyz:
All the major features in now

PR #1520, opened 2020-11-10
Add the data binding for data primitives

andydotxyz:
### Description:
This adds the new APIs for data binding of `DataItem`, supporting the primitive data types.
There is still list/map to come - these are being explored on th e binding branch that would incorporate this once landed.

Also nice is that I realised we can make these additions without breaking widgets in most cases :).

A significant step forward for #135

### Checklist:
- [x] Tests included.
- [x] Tests all pass.


Jacalz (changes requested):
Just a quick comment. I think we agreed to have the binding package in `data/binding` (when we moved validation to `data/validation`)?

andydotxyz:
Thanks @Jacalz - fixed

data/binding/binding.go
```diff
@@ -0,0 +1,58 @@
...
+type listener struct {
+	callback func(DataItem)
+}
+
+func (l *listener) DataChanged(i DataItem) {
+	l.callback(i)
+}
+
+type base struct {
+	listeners []DataItemListener
```
stuartmscott, line 31: Reads and writes to this slice should be locked

data/binding/binding.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// AddListener allows a data listener to be informed of changes to this item.
+func (b *base) AddListener(l DataItemListener) {
+	b.listeners = append(b.listeners, l)
```
stuartmscott, line 36: New listeners should be triggered with the current state: https://github.com/fyne-io/fyne/pull/804/files#diff-3f981e7a9d32b5147a37811648cc9f675d6f4519a1abd33797711994659da8d2R55

data/binding/binditems.go
```diff
@@ -0,0 +1,244 @@
...
+// BindBool returns a new bindable value that controls the contents of the provided bool variable.
+func BindBool(v *bool) Bool {
+	if v == nil {
+		return NewBool() // never allow a nil value pointer
+	}
+
+	return &bindBool{val: v}
+}
+
+type bindBool struct {
```
stuartmscott, line 28: Consider a more noun-ish naming convention;
```suggestion
type boundBool struct {
```

widget/check.go
```diff
@@ -200,6 +201,21 @@ func NewCheck(label string, changed func(bool)) *Check {
 	return c
 }
 
+// NewCheckWithData returns a check widget connected with the specified data source.
+func NewCheckWithData(label string, data binding.Bool) *Check {
+	check := NewCheck(label, func(b bool) {
+		data.Set(b)
+	})
+	check.Checked = data.Get()
```
stuartmscott, line 209: Not necessary if bindings notify new listeners of current state

data/binding/binding.go
```diff
@@ -0,0 +1,58 @@
...
+			b.listeners = b.listeners[:len(b.listeners)-1]
+		} else {
+			b.listeners = append(b.listeners[:i], b.listeners[i+1:]...)
+		}
+	}
+}
+
+func (b *base) trigger(i DataItem) {
+	for _, listen := range b.listeners {
+		listen.DataChanged(i)
```
stuartmscott, line 56: Do we not want to execute listeners in separate goroutines?

stuartmscott (commented):
With the current approach widgets must be bound to a binding at creation time and cannot be unbound or changed later. While I agree the 'NewABCWithData' constructors are useful, they should be convenience methods for calling `BindXYZ` methods, which are accompanied by `UnbindXYZ` methods as in #804

data/binding/binding.go
andydotxyz: That is cool for sure, but is it the expectation of a developer?
Normally listeners are triggers of change - when we are starting with a provided value (i.e. BindFloat(myVal) ) the developer already knows.
Not disagreeing, just wanted to make sure we considered this angle.

data/binding/binding.go
andydotxyz: yup thanks

data/binding/binditems.go
andydotxyz: Thanks, good catch.

data/binding/binding.go
andydotxyz: This is an interesting question. I know we went round it before - but did we come to a conclusion?
It can't be new goroutines due to ordering requirements but it could be a background thread like the svg cache janitor etc.
One more consideration is whether we expect third party bindings - if so they may have to hook into the same (or at least be able to, i.e. BaseBinding) to ensure that their code does not break our design.

andydotxyz:
Thanks, that's the changes made outside goroutine and firing listeners on connect. Let's discuss these on Slack @stuartmscott

data/binding/binding.go
stuartmscott: Are we agreed that it cannot be as it currently is, as (some/most?) data binding changes will come from the UI ie GLFW event thread?

andydotxyz: If we've built things currently then no developer code can run on the main thread (GLFW events are on main).
I don't think that currently it is possible that the binding will be called from main thread.
However it is likely that it may become possible if app properties or window management makes use of this so it should probably change before that happens.

stuartmscott: I assumed things like `OnTapped`, `TypedKey`, and `TypedRune` were run on the event thread

andydotxyz: "the event thread" is our event thread not GLFW's main event thread.

data/binding/binding.go
andydotxyz: After chat on Slack it was agree that adding this with some clear docs that explain it would be sufficient.
Resolved.

widget/check.go
andydotxyz: Resolved

data/binding/binding.go
andydotxyz: Added a data item thread

andydotxyz:
I think that is all the feedback addressed now.
As this PR is focusing on getting the databinding work in and not completing the widgets I have not included the suggested `Bind` and `Unbind` the helper constructors could delegate to them in a later PR with no breaks so I'm not concerned.
What I do think is that we should probably have explored Bind and Unbind in the documenting exploration phase because it brings up some complex questions about the undefined states of data that is being replaced by a bind (who's data wins) and what happens on unbind. Let's discuss that on a separate thread.

data/binding/binding.go
```diff
@@ -0,0 +1,77 @@
...
+	AddListener(DataItemListener)
+	// RemoveListener will detach the specified change listener from the DataItem.
+	// Disconnected listener will no longer be triggered when changes occur.
+	RemoveListener(DataItemListener)
+}
+
+// DataItemListener is any object that can register for changes in a bindable DataItem.
+// See NewDataItemListener to define a new listener using just an inline function.
+type DataItemListener interface {
+	DataChanged(DataItem)
```
stuartmscott, line 21: Could this parameter be removed?

andydotxyz: It looks like it can!

andydotxyz: Updated, thanks

Jacalz (approved):
Looks good to me 🙂

PR #1524, opened 2020-11-12
Add support for setting column width on table

andydotxyz:
This overrides the (template defined) fallback.
We use similar units and refer to the content size rather than the full-with-padding width.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


widget/table.go
```diff
@@ -154,6 +180,43 @@ func (t *Table) templateSize() fyne.Size {
...
+		if w, ok := t.columnWidths[i]; ok {
+			width = w
+		}
+
+		if colOffset <= t.offset.X-width-tableDividerThickness {
+			// before scroll
+		} else if colOffset <= t.offset.X {
+			minCol = i
+			offX = colOffset
+			isVisible = true
```
stuartmscott, line 204: Should this get reset to false if the column is after the scroll viewport, or are you just relying on the break on 209?

andydotxyz: Relying on break - we are starting from 0 so it should not be possible that we pass the window without realising

stuartmscott (approved):

PR #1527, opened 2020-11-12
[fyne-cli] Android: allow to specify an inline password for the keystore

lucor:
### Description:

This PR adds the possibility to specify the keystore password inline.

It allow to specify the password via the `keyStorePass` flag and defaults to the current behaviour that reads from stdin.

Note:  the `apksigner` tool supports the following formats:
-  pass:<password> – Password provided inline with the rest of the apksigner sign command.
-  env:<name> – Password is stored in the given environment variable.
-  file:<filename> – Password is stored as a single line in the given file.
-  stdin – Password is provided as a single line in the standard input stream. This is the default behavior for --ks-pass.

If we plan to support the env and/or file some possible ways are:
a) support them with additional flags like keyStorePassFromEnv and keyStorePassFromFile
b) update the implementation to accept the apksigner format

Fixes fyne-io/fyne-cross/pull/3

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


stuartmscott:
Note: a keystore is a collection of keys and so the keystore password and key password may be different

lucor:
@stuartmscott good catch, thanks.
There are other additional flags along with `--key-pass` that we'd add in order to support a fyne release command without user action (i.e. specify values from stdin)?

andydotxyz (approved):

PR #1534, opened 2020-11-14
Support multiple mouse buttons combined

### LINKED ISSUE #1533: desktop.MouseHover Button state is not reliable
### Describe the bug:

When using desktop.MouseHover to capture `MouseMoved(MouseEvent)` events in a widget, the button status may not work if both buttons are pressed at the same time.

### To Reproduce:
1. Move mouse, Button=0
2. Press left mouse button, Button=1
3. Press right mouse button too, Button=2 (expecting 3)
4. Release any one of the two mouse buttons, Button=0
5. Moving the mouse shows Button=0 despite one of the mouse buttons being still pressed
6. Release last button, a MouseUp event will be generated as expected (with the right value for Button)

Right now mouse buttons are handled as values (iota+1), I believe these should be handled as flags (1<<iota) in order to keep track of multiple buttons used at the same time.

Multiple buttons used at the same time might not be typical in desktop environment, but could happen in games (left button to move forward, right button to fire kind of things for example).

Extra: I'm not sure if it's expected or not, but enabling drag events (Dragged/DragEnd) causes MouseMoved events to not be fired while Button is not zero.

 - OS: Linux
 - Go version: 1.15.3
 - Fyne version: 1.4.0

andydotxyz:
Buttons should be additive rather than override.

Also remove old comment - that was supposed to reserve the iota, but
it got broken some time ago
Fixes #1533

### Checklist:

- [ ] Tests included. <- we still have not figured GLFW hardware based test mocks :(
- [x] Tests all pass.


MagicalTux:
Tested and confirmed fixing issue #1533

Jacalz (approved):
Looks good to me :+1:

PR #1535, opened 2020-11-14
Reword some popup deprecations for clarity

Jacalz:
### Description:
This rewords the deprecation comments for some of the popup deprecations.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/popup_menu.go
+++ b/widget/popup_menu.go
@@ -107,7 +107,7 @@ func (p *PopUpMenu) adjustedPosition(pos fyne.Position, size fyne.Size) fyne.Pos
 // NewPopUpMenuAtPosition creates a PopUp widget populated with menu items from the passed menu structure.
 // It will automatically be positioned at the provided location and shown as an overlay on the specified canvas.
 //
-// Deprecated: This will no longer show the pop-up in 2.0. Use ShowPopUpMenuAtPosition() instead.
+// Deprecated: Use ShowPopUpMenuAtPosition() instead.
 func NewPopUpMenuAtPosition(menu *fyne.Menu, c fyne.Canvas, pos fyne.Position) *PopUp {
 	options := NewVBox()
 	for _, option := range menu.Items {
@@ -134,7 +134,7 @@ func NewPopUpMenuAtPosition(menu *fyne.Menu, c fyne.Canvas, pos fyne.Position) *
 // NewPopUpMenu creates a PopUp widget populated with menu items from the passed menu structure.
 // It will automatically be shown as an overlay on the specified canvas.
 //
-// Deprecated: Use ShowPopUpMenuAtPosition() with an empty position instead.
+// Deprecated: Use ShowPopUpMenuAtPosition() with a zero position instead.
 func NewPopUpMenu(menu *fyne.Menu, c fyne.Canvas) *PopUp {
 	return NewPopUpMenuAtPosition(menu, c, fyne.NewPos(0, 0))
 }
--- a/widget/popup_menu.go
+++ b/widget/popup_menu.go
@@ -134,7 +134,7 @@ func NewPopUpMenuAtPosition(menu *fyne.Menu, c fyne.Canvas, pos fyne.Position) *
 // NewPopUpMenu creates a PopUp widget populated with menu items from the passed menu structure.
 // It will automatically be shown as an overlay on the specified canvas.
 //
-// Deprecated: This will no longer show the pop-up in 2.0. Use ShowPopUpMenuAtPosition() instead.
+// Deprecated: Use ShowPopUpMenuAtPosition() with an empty position instead.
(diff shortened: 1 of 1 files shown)


widget/popup_menu.go
```diff
@@ -134,7 +134,7 @@ func NewPopUpMenuAtPosition(menu *fyne.Menu, c fyne.Canvas, pos fyne.Position) *
 // NewPopUpMenu creates a PopUp widget populated with menu items from the passed menu structure.
 // It will automatically be shown as an overlay on the specified canvas.
 //
-// Deprecated: Use ShowPopUpMenuAtPosition instead.
+// Deprecated: This will no longer show the pop-up in 2.0. Use ShowPopUpMenuAtPosition() instead.
```
Jacalz, line 137: There is not any `ShowPopUpMenu()`. Should we add that or should this be deprecated in favour of `NewPopUpMenuAtPosition()` with an empty position?

toaster: I assume, you mean “`ShowPopUpMenuAtPosition`with an empty position”.
IMO, we don’t need convenience for this one.
Why would we need it?
What is special about showing a (non modal) pop-up implicitly at the top-left corner of the screen?

I think, if someone needs this, he can simply be explicit.

toaster (approved):

widget/popup_menu.go
Jacalz: You are right. I added that text. I figured it would be good as we have `ShowPopup()` and that isn't deprecated. Should that be deprecated?

widget/popup_menu.go
```diff
@@ -107,7 +107,7 @@ func (p *PopUpMenu) adjustedPosition(pos fyne.Position, size fyne.Size) fyne.Pos
 // NewPopUpMenuAtPosition creates a PopUp widget populated with menu items from the passed menu structure.
 // It will automatically be positioned at the provided location and shown as an overlay on the specified canvas.
 //
-// Deprecated: Use ShowPopUpMenuAtPosition instead.
+// Deprecated: This will no longer show the pop-up in 2.0. Use ShowPopUpMenuAtPosition() instead.
```
andydotxyz, line 110: I'm not too sure this particular line is correct - the is an old constructor and may not be kept with different semantics. That's why it was shown as a simple use-a-different-method sort of warning

widget/popup_menu.go
```diff
@@ -134,7 +134,7 @@ func NewPopUpMenuAtPosition(menu *fyne.Menu, c fyne.Canvas, pos fyne.Position) *
 // NewPopUpMenu creates a PopUp widget populated with menu items from the passed menu structure.
 // It will automatically be shown as an overlay on the specified canvas.
 //
-// Deprecated: Use ShowPopUpMenuAtPosition instead.
+// Deprecated: Use ShowPopUpMenuAtPosition() with an empty position instead.
```
andydotxyz, line 137: Do you mean a zero position?

andydotxyz (changes requested):

widget/popup_menu.go
Jacalz: Indeed. That is probably better.

widget/popup_menu.go
Jacalz: I must say that I am very confused then. Are we supposed to keep NewPopUpMenu() then or not?

Jacalz: If we are deprecating both this and NewPopUpMenu() then we have no constructors that actually returns a PopUpMenu instance.

andydotxyz: Those both return *PopUp... the new functions work on *PopUpMenu - we need to remove the old ones and then re-add a new one, and so what exactly will be the transition is undefined - hence why we are pushing off these.

andydotxyz:
I just realised that this is the wrong base branch - `develop` is where these will be removed for the next release.
Try using `release/v1.4.x` instead if you want these changes to be visible in any releases

widget/popup_menu.go
Jacalz: Ah, that explains things. Thank you. I was very confused why we were removing functions that seemed to have no real alternative :)

Jacalz:
Changing the base branch to 1.4.x created a whole world of problems. Closing this.

widget/popup_menu.go
toaster: Well, I would say yes but I don’t want to be the one to decide :).

andydotxyz: ShowPopUp and ShowModalPopUp were not deprecated as their return types are not expose and so will not break with the upcoming changes.
If you think we should remove the functionality because it's inadvisable that's probably another PR / issue :)
That said it's too late to add a new deprecation before 2.0 so if you mark deprecated now it's probably slated for removal in 3.0 - maybe we should hang back and see how this release goes?

andydotxyz (approved):

PR #1552, opened 2020-11-15
Adding validating form dialog

drognisep:
### Description:
This is a validating dialog factory function that takes `[]FormItem`s, places them in a form layout, and validates any `Validatable` widgets before allowing the dialog to be confirmed. An invalid form can still be cancelled.

Added an entry in the Dialog demo to show how this is expected to be used (very much like any other Dialog).

### Checklist:
- [x] Tests included.
  - Relies on the same functionality and tests as provided by the base dialog.
  - If that's not enough I'm happy to add more to keep up the quality of the repo.
  - `go vet` `gofmt` `goimport` ran from Goland
- [x] Tests all pass. 
  - **Not quite**, but I don't think I broke it.

This test is failing, but I'm not sure how I could have possibly broke it (other than running on Windows :P ). Is this leftover from someone else's merge, or is this a Windows bug?
```
--- FAIL: Test_ListableURI (0.00s)
    file_test.go:100: stringify URI file://C:/Users/josep/AppData/Local/Temp/file_test908998623/aaa to file://C:/Users/josep/AppData/Local/Temp/file_test908998623/aaa
    file_test.go:100: stringify URI file://C:/Users/josep/AppData/Local/Temp/file_test908998623/bbb to file://C:/Users/josep/AppData/Local/Temp/file_test908998623/bbb
    file_test.go:100: stringify URI file://C:/Users/josep/AppData/Local/Temp/file_test908998623/ccc to file://C:/Users/josep/AppData/Local/Temp/file_test908998623/ccc
    file_test.go:110:
                Error Trace:    file_test.go:110
                Error:          element file://C:/Users/josep/AppData/Local/Temp/file_test908998623/aaa appears more times in [file://C:/Users/josep/AppData/Local/Temp/file_test908998623/aaa file://C:/Users/josep/AppData/Local/Temp/file_test908998623/bbb file://C:/Users/josep/AppData/Local/Temp/file_test908998623/ccc] than in [file://C:\Users\josep\AppData\Local\Temp\file_test908998623\aaa file://C:\Users\josep\AppData\Local\Temp\file_test908998623\bbb file://C:\Users\josep\AppData\Local\Temp\file_test908998623\ccc]
                Test:           Test_ListableURI
2020/11/15 12:30:36 Fyne error:  Main menu must have at least one child menu
2020/11/15 12:30:36   At: C:/Users/josep/git/fyne/internal/driver/glfw/menu.go:9
FAIL
FAIL    fyne.io/fyne/internal/driver/glfw       10.027s
```

#### Where applicable:

  - This needs to be discussed more here. There was a little conversation over slack, but I want to make sure I'm not tainting such a well designed API.
  - No breaking changes, and deprecated references copied from `NewCustomDialog` removed.


Jacalz (changes requested):
Before I give my full review tomorrow, I would greatly appreciate if you could move out the changes from `dialog/base.go` and move them to a `dialog/form.go`. Whilst there, it would be great if it could be created as a new Form struct (I suggest looking at how `dialog/confirm.go` is made).

drognisep:
Thanks for the feedback, @Jacalz! That made refactoring for testing a lot less gross.

dialog/base.go
```diff
@@ -12,8 +12,9 @@ import (
 )
 
 const (
-	padWidth  = 32
-	padHeight = 16
+	padWidth              = 32
+	padHeight             = 16
+	validationErrorPrefix = "Validation error:"
```
drognisep, line 17: Forgot about this const. I'll remove this.

drognisep:
I think this PR is in good shape now, let me know if any further changes are needed. :)

dialog/form.go
```diff
@@ -0,0 +1,86 @@
...
+				}
+			}
+		}
+		confirmBtn.Enable()
+	}
+	validateItems()
+
+	for _, item := range items {
+		if validatable, canValidate := item.Widget.(fyne.Validatable); canValidate {
+			validatable.SetOnValidationChanged(func(error) {
```
Jacalz, line 63: Once the comment about moving the validation error into a method is done, you should just be able to plug it in directly here.

dialog/form.go
```diff
@@ -0,0 +1,86 @@
...
+// NewFormDialog creates and returns a dialog over the specified application using
+// the provided FormItems. The cancel button will have the dismiss text set and the confirm button will
+// use the confirm text. The response callback is called on user action after validation passes.
+// If any Validatable widget reports that validation has failed, then the confirm
+// button will be disabled. The initial state of the confirm button will reflect the initial
+// validation state of the items added to the form dialog.
+// The MinSize() of the CanvasObject passed will be used to set the size of the window.
+func NewFormDialog(title, confirm, dismiss string, items []*widget.FormItem, callback func(bool),
+	parent fyne.Window) *FormDialog {
+	var itemObjects = make([]fyne.CanvasObject, 0, len(items)*2)
```
Jacalz, line 27: This makes a new slice with zero length and `len(items)*2` capacity. I think it would be a nice little performance improvement if we instead set the length to `len(items)*2`. We would then need to change each item individually. Such as something like this:
```suggestion
	var itemObjects = make([]fyne.CanvasObject, len(items)*2)
	for i := 0; i < len(items)*2-2; i +=2 {
	    itemObjects[i] = widget.NewLabel(ii.Text)
	    itemObjects[i+1] = ii.Widget
	}
```

Jacalz (changes requested):
Here is my first review with some suggested changes. I have not yet looked at the tests.

dialog/form.go
Jacalz: Just realized that the suggestion above wouldn't work. Ignore this one.

dialog/form.go
```diff
@@ -0,0 +1,86 @@
...
+	confirmBtn := &widget.Button{Text: confirm, Icon: nil, Importance: widget.HighImportance,
+		OnTapped: func() {
+			d.hideWithResponse(true)
+		}}
+	// Mitigation for issue #1553
+	confirmBtn.SetIcon(theme.ConfirmIcon())
+	validateItems := func() {
+		for _, item := range items {
+			if validatable, canValidate := item.Widget.(fyne.Validatable); canValidate {
+				if err := validatable.Validate(); err != nil {
```
Jacalz, line 51: The FormItem has a ValidationError field. Let’s updated that as well in case someone wants to use it.

Jacalz: Sorry. I thought it was public. Ignore this one.

dialog/form.go
drognisep: Not by much, but more than 0. Definitely worth the extra code. :)

drognisep: Not as typed, but I can use this.
```
	var itemObjects = make([]fyne.CanvasObject, len(items)*2)
	for ii, item := range items {
		itemObjects[ii*2] = widget.NewLabel(item.Text)
		itemObjects[ii*2+1] = item.Widget
	}
```

cmd/fyne_demo/tutorials/dialog.go
```diff
@@ -143,6 +144,34 @@ func dialogScreen(win fyne.Window) fyne.CanvasObject {
 				log.Println("Please Authenticate", username.Text, password.Text)
 			}, win)
 		}),
+		widget.NewButton("Form Dialog (Login Form)", func() {
+			userRx := validation.NewRegexp(`^[A-Za-z0-9_-]+$`, "username can only contain letters, numbers, '_', and '-'")
```
Jacalz, line 148: Very minor comment. You could just set this directly instead of creating a new variable.

cmd/fyne_demo/tutorials/dialog.go
```diff
@@ -143,6 +144,34 @@ func dialogScreen(win fyne.Window) fyne.CanvasObject {
...
+			username := widget.NewEntry()
+			username.Validator = userRx
+			passRx := validation.NewRegexp(`^[A-Za-z0-9_-]+$`, "password can only contain letters, numbers, '_', and '-'")
```
Jacalz, line 151: Very minor comment. You could just set this directly instead of creating a new variable.

dialog/form.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"fyne.io/fyne"
+	"fyne.io/fyne/container"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/theme"
+	"fyne.io/fyne/widget"
+)
+
+// FormDialog is like the standard Dialog but with confirm and cancel buttons, as well as the FormItems in the dialog.
```
Jacalz, line 11:
```suggestion
// FormDialog is a simple dialog window for displaying FormItems inside a form.
```

dialog/form.go
```diff
@@ -0,0 +1,98 @@
...
+
+// validateItems acts as a validation edge state handler that will respond to an individual widget's validation
+// state before checking all others to determine the net validation state. If the error passed is not nil, then the
+// confirm button will be disabled. If the error parameter is nil, then all other Validatable widgets in items are
+// checked as well to determine whether the confirm button should be disabled.
+// This method is passed to each Validatable widget's SetOnValidationChanged method in items by NewFormDialog.
+func (d *FormDialog) validateItems(err error) {
+	if err != nil {
+		d.confirm.Disable()
+	} else {
```
Jacalz, line 27: Minor comment. It is cleaner to just return early and avoid the else statement.

dialog/form.go
```diff
@@ -0,0 +1,98 @@
...
+	confirmBtn.SetIcon(theme.ConfirmIcon())
+	formDialog := &FormDialog{
+		dialog:  d,
+		items:   items,
+		confirm: confirmBtn,
+		cancel:  d.dismiss,
+	}
+	formDialog.validateItems(nil)
+
+	for _, item := range items {
```
Jacalz, line 79: The operation within this loop could probably be moved into the one at the top of the function.

Jacalz (changes requested):
This is looking really polished. Just a few more minor comments. Apart from that, it is looking really good. Still need to have a look at the tests and test it myself locally though (I am a bit short of time at the moment).

dialog/form.go
Jacalz: That's genius. Nice work :)

dialog/form.go
drognisep: That would be kind of tricky to put together since the items are being operated on by a struct that needs the items to exist before they can be operated upon, and the dialog also needs them to fill out its content, which is needed by the struct. Seems like a lot of sacrifice to readability to eliminate a one-time loop. How would you recommend this be done?

Jacalz: Good point. Did not think about the contents withing that function that is set.

Jacalz: Let's just keep it as is.

cmd/fyne_demo/tutorials/dialog.go
```diff
@@ -143,6 +144,32 @@ func dialogScreen(win fyne.Window) fyne.CanvasObject {
 				log.Println("Please Authenticate", username.Text, password.Text)
```
Jacalz, line 144: I just tested this locally. I think you could remove the "Custom Dialog (Login Form)". Your new dialog is basically doing the same thing, but better :)

dialog/form_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+import (
+	"errors"
+	"testing"
+	"time"
+
+	"fyne.io/fyne"
+	"fyne.io/fyne/test"
+	"fyne.io/fyne/widget"
+	testify "github.com/stretchr/testify/assert"
```
Jacalz, line 11: We generally just import it as `assert` (without changing import name) and then use the regular assert functions instead of the `.New(t)` and then calling methods on that. That applies in more parts of the code, just mentioning it here.

Jacalz (changes requested):
Looking really good. Just a few last comments. I will approve this after these are fixed :)

dialog/form.go
```diff
@@ -0,0 +1,98 @@
...
+// the provided FormItems. The cancel button will have the dismiss text set and the confirm button will
+// use the confirm text. The response callback is called on user action after validation passes.
+// If any Validatable widget reports that validation has failed, then the confirm
+// button will be disabled. The initial state of the confirm button will reflect the initial
+// validation state of the items added to the form dialog.
+// The MinSize() of the CanvasObject passed will be used to set the size of the window.
+func NewFormDialog(title, confirm, dismiss string, items []*widget.FormItem, callback func(bool),
+	parent fyne.Window) *FormDialog {
+	var itemObjects = make([]fyne.CanvasObject, len(items)*2)
+	for ii, item := range items {
```
Jacalz, line 50: Minor comment. We can use `i` instead of `ii`. It is what is most common.

Jacalz (changes requested):

dialog/form.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"fyne.io/fyne"
+	"fyne.io/fyne/container"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/theme"
+	"fyne.io/fyne/widget"
+)
+
+// FormDialog is a simple dialog window for displaying FormItems inside a form.
+type FormDialog struct {
```
andydotxyz, line 12: Exporting this new type provides no functionality - did you consider making it a private type and having the constructors return `fyne.Dialog` instead?

dialog/form.go
```diff
@@ -0,0 +1,98 @@
...
+	d.confirm.Enable()
+}
+
+// NewFormDialog creates and returns a dialog over the specified application using
+// the provided FormItems. The cancel button will have the dismiss text set and the confirm button will
+// use the confirm text. The response callback is called on user action after validation passes.
+// If any Validatable widget reports that validation has failed, then the confirm
+// button will be disabled. The initial state of the confirm button will reflect the initial
+// validation state of the items added to the form dialog.
+// The MinSize() of the CanvasObject passed will be used to set the size of the window.
```
andydotxyz, line 46: I don't think this line is needed as that is how all of Fyne layout works

dialog/form_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+func TestFormDialog_InvalidCannotSubmit(t *testing.T) {
+	assert := testify.New(t)
+	ch := make(chan bool)
+	fd := validatingFormDialog(ch, test.NewWindow(nil))
+	fd.Show()
+
+	assert.False(fd.win.Hidden)
+	assert.True(fd.confirm.Disabled(), "Confirm button should be disabled due to validation state")
+	go test.Tap(fd.confirm)
```
andydotxyz, line 37: without this goroutine you don't need the timeout below

andydotxyz (changes requested):
I think you can remove the goroutine for button taps and the timeouts that compensate for them.
I imagine the code could have been copied from elsewhere, but we moved away from that model for new tests.

dialog/form.go
drognisep: Yeah that makes more sense. :)

cmd/fyne_demo/tutorials/dialog.go
drognisep: Okay, will do.

dialog/form_test.go
drognisep: I can change that. Is there a reason `require` isn't used?

dialog/form.go
drognisep: Okidoke

dialog/form.go
drognisep: Removed

dialog/form.go
Jacalz: Hmm, it looks like this change makes golint complain.

dialog/form.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// NewFormDialog creates and returns a dialog over the specified application using
+// the provided FormItems. The cancel button will have the dismiss text set and the confirm button will
+// use the confirm text. The response callback is called on user action after validation passes.
+// If any Validatable widget reports that validation has failed, then the confirm
+// button will be disabled. The initial state of the confirm button will reflect the initial
+// validation state of the items added to the form dialog.
+func NewFormDialog(title, confirm, dismiss string, items []*widget.FormItem, callback func(bool),
+	parent fyne.Window) *formDialog {
```
Jacalz, line 47: The tests are failing because golint is complaining about this being unexported. I think you should return fyne.Dialog instead, like @andydotxyz mentioned.

Jacalz (changes requested):

dialog/form.go
drognisep: Yeah, I just realized that wouldn't work at all. Sorry about that. Just pushed a fix.

drognisep:
@Jacalz & @andydotxyz 
I think I have everything resolved now, but it looks like CI is stuck.

Jacalz:
Yes. Travis has sadly cut down on the open source builders which is hurting our build times massively.

andydotxyz:
It looks like there were some infrastructure errors in the build run, I have kicked it to see if they will resolve.

Jacalz (approved):
Nice work! Congratulations on your first contribution to Fyne 🥳

andydotxyz (approved):
Great work, thanks!

drognisep:
Wonderful! Thanks for all the great feedback, guys. :)

andydotxyz:
Since this was landed I realised the naming was off - it should be `dialog.ShowForm` to avoid the duplication.

I updated this on develop and added the "Since: 2.0.0" as well.

PR #1554, opened 2020-11-15
Add blinking cursor in Entry

andydotxyz:
This needs Animation.AutoReverse and Animation.Stop
demo these in fyne_demo.

ezgif com-gif-maker

### Checklist:

- [ ] Tests included. <- existing widget tests all passing, the animation lacks solid test now, so use fyne_demo
- [x] Tests all pass.

### DIFF
--- a/canvas/animation.go
+++ b/canvas/animation.go
@@ -22,17 +22,20 @@ func NewColorRGBAAnimation(start, stop color.Color, d time.Duration, fn func(col
 	r1, g1, b1, a1 := start.RGBA()
 	r2, g2, b2, a2 := stop.RGBA()
 
-	rDelta := int(r2) - int(r1)
-	gDelta := int(g2) - int(g1)
-	bDelta := int(b2) - int(b1)
-	aDelta := int(a2) - int(a1)
+	rStart := uint8(r1 >> 8)
+	gStart := uint8(g1 >> 8)
+	bStart := uint8(b1 >> 8)
+	aStart := uint8(a1 >> 8)
+	rDelta := float32(int(r2>>8) - int(rStart))
+	gDelta := float32(int(g2>>8) - int(gStart))
+	bDelta := float32(int(b2>>8) - int(bStart))
+	aDelta := float32(int(a2>>8) - int(aStart))
 
 	return &fyne.Animation{
 		Duration: d,
 		Tick: func(done float32) {
-
-			fn(color.NRGBA{R: scaleChannel(r1, rDelta, done), G: scaleChannel(g1, gDelta, done),
-				B: scaleChannel(b1, bDelta, done), A: scaleChannel(a1, aDelta, done)})
+			fn(color.RGBA{R: scaleChannel(rStart, rDelta, done), G: scaleChannel(gStart, gDelta, done),
+				B: scaleChannel(bStart, bDelta, done), A: scaleChannel(aStart, aDelta, done)})
 		}}
 }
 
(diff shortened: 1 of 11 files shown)


andydotxyz:
This now includes optimisations missed in the feedback from first animation PR

canvas/animation.go
```diff
@@ -19,47 +19,58 @@ const (
 // The content of fn should apply the color values to an object and refresh it.
 // You should call Start() on the returned animation to start it.
 func NewColorRGBAAnimation(start, stop color.Color, d time.Duration, fn func(color.Color)) *fyne.Animation {
+	r1, g1, b1, a1 := start.RGBA()
+	r2, g2, b2, a2 := stop.RGBA()
+
+	rDelta := int(r2) - int(r1)
+	gDelta := int(g2) - int(g1)
+	bDelta := int(b2) - int(b1)
+	aDelta := int(a2) - int(a1)
```
toaster, line 28: I suggest to store these (and all other deltas) as `float32` because it is used in a `float32` computation.
Colour channels should also be shifted 8 bit to the right before converted into `float32` because this saves the shift in the `scaleChannel` function.

toaster (changes requested):

canvas/animation.go
andydotxyz: Unfortunately I don't think we can pre-shift, because the delta can be a negative number...

AlbinoGeek:
As a 90% Blind individual, please give us ways to customize this blink.

E.g: 

- "Fade" (the default you've shown)
- "Blink" (only on/off states)
- "Solid" (only on)
- "None" (? should this be included for completeness)

canvas/animation.go
```diff
@@ -22,17 +22,20 @@ func NewColorRGBAAnimation(start, stop color.Color, d time.Duration, fn func(col
...
-	bDelta := int(b2) - int(b1)
-	aDelta := int(a2) - int(a1)
+	rStart := uint8(r1 >> 8)
+	gStart := uint8(g1 >> 8)
+	bStart := uint8(b1 >> 8)
+	aStart := uint8(a1 >> 8)
+	rDelta := float32(int(r2>>8) - int(rStart))
+	gDelta := float32(int(g2>>8) - int(gStart))
+	bDelta := float32(int(b2>>8) - int(bStart))
+	aDelta := float32(int(a2>>8) - int(aStart))
```
toaster, line 32: The start values are always converted to `int` before being used. Therefore, they should be stored as `int` already.

And regarding the delta values for animation: Storing them as `float32` saves conversion during animation and does not use up more space than `int`.

toaster (changes requested):
One last step to go.

andydotxyz:
> * "Solid" (only on)

I think we could add this as a user preference - should be sufficient right?

AlbinoGeek:
@andydotxyz You are correct, Accessibility should be a user preference v.s. developer preference.

I was mainly thinking about those who want a specific "feel" to their application, e.g:
- A console with fade cursor would be distracting

As per why I suggested those 3-4 options:
- Blink is a pretty standard OS cursor (on, off, etc.)
- Solid means "always on" (never goes off), traditional "I Beam"

AlbinoGeek:
However, when it comes to blink cursor, some OS actually have options for this, the question is; would Fyne want to integrate with the DE when it is present and has this setting? (to pick a friendly default)

Gnome -> Settings -> Accessibility -> Typing -> Cursor Blink

image

On Gnome at least,  you would query these preferences using `gconf-tool2` or `gsettings` in Gnome Shell 3+

```bash
$ gsettings get org.gnome.desktop.interface cursor-blink
true
$ gsettings get org.gnome.desktop.interface cursor-blink-time
1200
$ gsettings get org.gnome.desktop.interface cursor-blink-timeout
10
```

---

I haven't used Windows for some time, but I know they have similar settings:

image
image

andydotxyz:
Some systems offer such customisation and some don't (as far as I can see mac/iOS/Android don't).
Over time we could do more to match - but that really belongs in a feature request I would think, given that accessibility is still being figured out for future releases.

AlbinoGeek:
Of course, like we discussed in chat, will do!
I'll make a proposal feature before touching PR.

toaster (approved):

andydotxyz:
I have merged this PR - can you open tickets for the items that were not covered in this addition @AlbinoGeek if we have not already covered them elsewhere?

PR #1557, opened 2020-11-16
Avoid unnecessary validation check on refresh in widget.Form

Jacalz:
### Description:
The updateButtons() function is called each time the widget is refreshed and the function included the check for validations earlier. This meant that each time the form was refreshed, it would re-iterate through all widgets updating the validation. This was entirely unnecessary as each widget does this themselves when the status changes. 

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ More detailed release notes can be found on the [releases page](https://github.c
 * Fyne package creating invalid windows packages (#1521)
 * Menu bar initially doesn't respond to mouse input on macOS (#505) 
 * CenterOnScreen causes crash on MacOS when called from goroutine (#1539)
+* Avoid unnecessary validation check on Refresh in widget.Form
 
 
 ## 1.4 - 1 November 2020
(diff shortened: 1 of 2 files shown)


andydotxyz:
Doesn't this remove the initial display - CreateRenderer used to trigger this indirectly through updateButtons

Jacalz:
@andydotxyz I don't think it would make much of a difference. The validation errors are only updated on validation changed in the widgets meaning that calling it in updateButtons on CreateRenderer would not have made much of a difference. We might want to call `.Validate()` at first some time in the future, but I think that might be something for a different PR.

andydotxyz:
If the form loads with initially invalid content will the buttons not now have the wrong state?

Jacalz:
Oh yes indeed, but not as a consequence of this PR. It was something I think was over-looked when implementing this. Looks like I have time to fix it now so I will go ahead and do so :)

andydotxyz (approved):

PR #1561, opened 2020-11-16
Data binding values to preferences

andydotxyz:
Wanted to share this early work on getting data binding connected to preferences

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/binding.go
+++ b/data/binding/binding.go
@@ -4,6 +4,9 @@ package binding
 
 import "sync"
 
+// Because there is no preference listener yet we connect any listeners asking for the same key.
+var prefBinds = make(map[string]DataItem)
+
 // DataItem is the base interface for all bindable data items.
 type DataItem interface {
 	// AddListener attaches a new change listener to this DataItem.
--- a/data/binding/binding.go
+++ b/data/binding/binding.go
@@ -12,20 +12,20 @@ type DataItem interface {
 	// AddListener attaches a new change listener to this DataItem.
 	// Listeners are called each time the data inside this DataItem changes.
 	// Additionally the listener will be triggered upon successful connection to get the current value.
-	AddListener(DataItemListener)
+	AddListener(DataListener)
 	// RemoveListener will detach the specified change listener from the DataItem.
 	// Disconnected listener will no longer be triggered when changes occur.
-	RemoveListener(DataItemListener)
+	RemoveListener(DataListener)
 }
 
-// DataItemListener is any object that can register for changes in a bindable DataItem.
-// See NewDataItemListener to define a new listener using just an inline function.
-type DataItemListener interface {
+// DataListener is any object that can register for changes in a bindable DataItem.
+// See NewDataListener to define a new listener using just an inline function.
+type DataListener interface {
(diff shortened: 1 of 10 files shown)


andydotxyz:
Tests and some API improvements added, this is now complete.
I realised that `DataItemListener` defining `DataChanged()` instead of `DataChanged(DataItem)` meant that we could use this for list and map as well, so renamed to `DataListener` and updated the convenience constructor as well.

data/binding/preference_test.go
```diff
@@ -0,0 +1,107 @@
...
+	p := a.Preferences()
+	key := "test-bool"
+
+	p.SetBool(key, true)
+	bind := BindPreferenceBool(key, p)
+	assert.True(t, bind.Get())
+
+	bind.Set(false)
+	assert.False(t, bind.Get())
+	assert.False(t, p.Bool(key))
```
stuartmscott, line 22: Ensure that calling `p.SetBool(key, true)` again will trigger the binding

stuartmscott: And ditto for other types

data/binding/binding.go
```diff
@@ -4,25 +4,28 @@ package binding
 
 import "sync"
 
+// Because there is no preference listener yet we connect any listeners asking for the same key.
+var prefBinds = make(map[string]DataItem)
```
stuartmscott, line 8: Could this be moved to data/binding/preference.go?

data/binding/preference_test.go
andydotxyz: Sadly it won't yet - as we don't have any preference listener (see comment in your other feedback).
We could have the properties code depend on dataBinding, and have a specific "PreferenceUpdated()" called, not sure if this is a good approach or a workaround...?

I'd be happy to open a new issue "Preferences set directly do not trigger binding changes" so we can implement it properly? I think it would also need to include code so that the "preferencesLoad" validatess which items have changed so we don't just blindly call back every preference value if the file is touched...

data/binding/preference.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// BindPreferenceBool returns a bindable bool value that is managed by the application preferences.
+// Changes to this value will be saved to application storage and when the app starts the previous values will be read.
+func BindPreferenceBool(key string, p fyne.Preferences) Bool {
+	if listen, ok := prefBinds[key]; ok {
+		if l, ok := listen.(Bool); ok {
+			return l
+		}
+		fyne.LogError("A previous preference binding exists with different type for key: "+key, nil)
```
stuartmscott, line 24: Could this be extracted to a helper `logKeyTypeError` defined at the top of this file?

data/binding/preference_test.go
stuartmscott: Yes let's open an issue to track this, though I wouldn't worry about blindly calling back - data bindings should be smart enough to only trigger their listeners when their value changes.

stuartmscott: #1577

stuartmscott (approved):

PR #1568, opened 2020-11-17
add PresetFileName

biberino:
### Description:

Allows to preset the filename in a FileSaveDialog.

### Checklist:

- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -30,6 +30,8 @@ type fileDialog struct {
 	win      *widget.PopUp
 	selected *fileDialogItem
 	dir      fyne.ListableURI
+	// this will the initial filename in a FileSave dialog
+	initialFileName string
 }
 
 // FileDialog is a dialog containing a file picker for use in opening or saving files.
@@ -44,6 +46,8 @@ type FileDialog struct {
 	desiredSize      *fyne.Size
 	// this will be applied to dialog.dir when it's loaded
 	startingLocation fyne.ListableURI
+	// this will the initial filename in a FileSave dialog
+	initialFileName string
 }
 
 // Declare conformity to Dialog interface
@@ -131,6 +135,9 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 	})
 	f.open.Style = widget.PrimaryButton
 	f.open.Disable()
+	if f.file.save {
+		f.fileName.SetText(f.initialFileName)
+	}
 	dismissLabel := "Cancel"
 	if f.file.dismissText != "" {
 		dismissLabel = f.file.dismissText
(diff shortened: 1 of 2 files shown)


dialog/file.go
```diff
@@ -506,6 +506,14 @@ func (f *FileDialog) SetFilter(filter storage.FileFilter) {
 	}
 }
 
+// PresetFileName sets the filename in a FileSaveDialog.
+// This is normally called before the dialog is shown.
+func (f *FileDialog) PresetFileName(fileName string) {
```
andydotxyz, line 511: `Preset` here does not seem required - it is not happening before anything in particular - why did you decide not to use `SetFileName`?

andydotxyz (changes requested):

dialog/file.go
biberino: Because this will usually be called before Show() so it "presets" the filename before revealing the dialog. However SetFileName seems fine as well

biberino: Want me to rename to `SetFileName`?

andydotxyz: Yes please - "Preset" asserts usage expectations that are not met in what the code does. Plus we don't use that word elsewhere.

P.S. Can you please add tests? I would think 1 showing that it works called before show, and another that shows after?

dialog/file.go
```diff
@@ -311,8 +318,12 @@ func (f *fileDialog) setSelected(file *fileDialogItem) {
 	f.selected = file
 
 	if file == nil || file.location.String()[len(file.location.Scheme())+3:] == "" {
-		f.fileName.SetText("")
-		f.open.Disable()
+		// keep user input while navigating
+		// in a FileSave dialog
+		if !f.file.save {
```
andydotxyz, line 323: When I tap a file in save mode it should fill this with that file name...
Doesn't this code stop that functionality?

dialog/file.go
```diff
@@ -506,11 +517,15 @@ func (f *FileDialog) SetFilter(filter storage.FileFilter) {
 	}
 }
 
-// PresetFileName sets the filename in a FileSaveDialog.
+// SetFileName sets the filename in a FileSaveDialog.
```
andydotxyz, line 520: There is no type FileSaveDialog - so probably should be "FileDialog in save mode" or similar?

dialog/file.go
```diff
@@ -30,6 +30,8 @@ type fileDialog struct {
 	win      *widget.PopUp
 	selected *fileDialogItem
 	dir      fyne.ListableURI
+	// this will the initial filename in a FileSave dialog
```
andydotxyz, line 33: This (and line 49) is not a complete sentence. Did you mean "this will be"?

andydotxyz (changes requested):

dialog/file.go
biberino: No this will only execute if file == nil, eg when nothing is selected. If you tap a file the filename will be entered. If you navigate around (only clicking folders) the last tapped filename will remain in the entry.

andydotxyz (approved):

andydotxyz:
Great, thanks for the addition

PR #1569, opened 2020-11-17
Optimize CreateRenderer call of widget.Form

Jacalz:
### Description:
This small optimization makes the renderer creation between 0.5 and 3 times faster (depending on the amount of items). The difference might be even bigger with larger amounts of items added.

It also fixes an issue where the validation could be triggered before the renderer was created and thus cause a panic when trying to disable (or possibly enable) a button that did not yet exist.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
Looks like a good direction. I'd merge latest develop in so you can test the form dialog as well

Jacalz:
The optimization is already added to to the form dialog and this issue can not happen there as the validation function is set after the button is created :)

I forgot to say this on the call today, but the panic issue is really strange. I know that it probably could happen before, but it only surfaced after the optimization was added and the form was suddenly panicking all the time. Could it be timing issue that made it surface now that it is faster?

widget/form_test.go
```diff
@@ -132,3 +133,35 @@ func TestForm_ChangeTheme(t *testing.T) {
...
+	test.Type(entry2, "not-")
+	entry1.SetText("incorrect")
+	w = test.NewWindow(form)
+
+	test.AssertImageMatches(t, "form/validation_invalid.png", w.Canvas().Capture())
+
+	entry1.SetText("15-true")
+	w = test.NewWindow(form)
+
+	test.AssertImageMatches(t, "form/validation_valid.png", w.Canvas().Capture())
```
Jacalz, line 166: I don't quite get why this image has the submit button icon not showing.

andydotxyz: It may relate to an issue regarding disabled buttons.
It may just relate to it.
Can you try the same code based on develop and see if you have the same issue?

Jacalz: Same issue sadly. I have no idea as to what is causing it.

andydotxyz:
The code looks good, but I think we need to solve the button test failures.
Worth noting that "validation_initial.png" is wrong as well - it has the enabled icon with disabled text...

andydotxyz (changes requested):
We can't really commit broken test images (the initial, and valid states for the form submit button seem wrong)

CHANGELOG.md
```diff
@@ -23,10 +23,12 @@ These changes likely break some apps, please read the
 
 * Change bundled resources to use more efficient string storage
 * Desktop left and right mouse buttons renamed to MouseButtonPrimary and MouseButtonSecondary
+* Many optimisations and widget performance enhancements
 
 ### Fixed
 
-*
+* Validating a widget in widget.Form before renderer was created could cause a panic
+* Appended items in widget.Form had validation set up twice
```
andydotxyz, line 31: This line seems like a duplicate of line 26 (the widget optiomisations comment)
Is the line above a bug that developers could encounter, or was it just our internal code?

andydotxyz:
Glad this is working now. Code and tests look good.
I'm just not sure it needs 3 lines in the CHANGELOG.
Just for clarity - the change log is a public facing document to inform developers using our toolkit what changes they can expect. So if it's not visible to them or won't imact their work we should probably omit.

CHANGELOG.md
Jacalz: Good point. Changed it.

andydotxyz (approved):

PR #1586, opened 2020-11-20
add: icons MediaStop and Upload

AlbinoGeek:
## Description:
Added some useful icons that were missing from the existing groups:
- Media Stop
- Upload

### Checklist:

- [X] Tests all pass.


theme/gen.go
```diff
@@ -183,7 +186,9 @@ func main() {
 
 	bundleIcon("download", f)
 	bundleIcon("computer", f)
+	bundleIcon("refresh", f)
```
stuartmscott, line 189: Is this different from view-refresh?

AlbinoGeek:
Ahh, I missed where they were referenced as public functions -- fixing that next.

stuartmscott (commented):
Thanks for taking the time to make this contribution @AlbinoGeek

The `media-stop` and `upload` icons make perfect sense and I would happily approve their addition to this project.

I have seen the `check-box-indeterminate` icon used in the MD spec for checkboxes with a parent-child relationship, however this is not something Fyne checkboxes support at this time. Therefore I'd be tempted to hold off on adding this icon until that support is added, unless there is another compelling reason to add it now.

I do not think the new `document-rename` icon and the existing `document-create` icon go together well.

I'm not sure how the new `refresh` icon differs from the existing `view-refresh` icon.

andydotxyz:
I agree with @stuartmscott comments, thanks.
Also could you add the icons to the "Theme Icons" section of `cmd/fyne_demo` please?

theme/gen.go
AlbinoGeek: Correct -- sorry about that.

AlbinoGeek:
The issue I ran into is that `file-` in fyne currently references `content type metadata` and not... literal file operations, which has been a little confusing when it comes to selecting matching icons from Material Design.

There were a lot of icons I really wanted but simply could not suggest adding because they were only available in outline format, such as `cloud_folder_upload` -- this was one of those issues, where I don't think we have any semantic representation of a `rename` operation at the moment, but could not find a matching icon.

`drive-file-rename-outline` being the only other semantic rename I could find, but it was also a non-matching icon.

Perhaps that should wait until the move to outline icons then.

---

Good points on `check-box-indeterminate`, as this is not a feature currently implemented in fyne. The other use-case to markdown specification is `CheckGroup` logic, in the sense that you have a UI containing multiple check-boxes, and a header which has a matching check box which implements "Select/Deselect All" logic -- which would then have the indeterminate state if "some but not all" items were selected.

---

Fixing that all now.

stuartmscott (dismissed):
Great work @AlbinoGeek, thanks for you contribution!

Its worth noting that if you have a need for an icon which doesn't make sense to bundle in the Fyne toolkit, you can bundle it into you app's code base instead

andydotxyz:
I think there is a copy/paste bug:

Screenshot 2020-11-20 at 20 40 34

cmd/fyne_demo/tutorials/icons.go
```diff
@@ -154,6 +155,7 @@ var icons = []struct {
 	{"MediaFastRewind", theme.MediaFastRewindIcon()},
 	{"MediaPause", theme.MediaPauseIcon()},
 	{"MediaPlay", theme.MediaPlayIcon()},
+	{"MediaStop", theme.MediaPlayIcon()},
```
andydotxyz, line 158: Wrong icon

andydotxyz (changes requested):

cmd/fyne_demo/tutorials/icons.go
AlbinoGeek: Good catch! Bad copy/paste for sure.

AlbinoGeek:
* rebased onto `develop` after `release/v1.4.x` merge

Jacalz (approved):

AlbinoGeek:
Is there still a requested change? I can't see it.

andydotxyz (approved):

andydotxyz:
For the indeterminate checkbox there will be an example in the upcoming Fyne book :)

PR #1592, opened 2020-11-20
Some pre-allocation optimizations for accordion, radio_group and fyne_demo

Jacalz:
### Description:

This applies the per-allocation optimization to a few more places. There are probably a few more that could be added but this is a good step forward in the meantime, fixing the more straight-forward cases. In summary, it boils down to:
- Setting the objects in accordion went from (a total time over three calls) `10.022µs` to `1.567µs` (6.4x faster).
- Adding all objects in CreateRenderer of radio_group went from (a total time over four calls) `2.327µs` to `1.486µs` (1.5x faster).
- Creating the icon strings in fyne_demo went from `11.444µs` to `3.566µs` (3.2x faster). 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/accordion.go
+++ b/widget/accordion.go
@@ -194,6 +194,7 @@ func (r *accordionRenderer) Refresh() {
 func (r *accordionRenderer) updateObjects() {
 	is := len(r.container.Items)
 	hs := len(r.headers)
+	ds := len(r.dividers)
 	i := 0
 	for ; i < is; i++ {
 		ai := r.container.Items[i]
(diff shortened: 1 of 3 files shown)


widget/accordion.go
```diff
@@ -232,12 +232,12 @@ func (r *accordionRenderer) updateObjects() {
...
-	for _, h := range r.headers {
-		objects = append(objects, h)
+	objects := make([]fyne.CanvasObject, len(r.headers)+len(r.container.Items))
+	for i, header := range r.headers {
+		objects[i] = header
 	}
-	for _, i := range r.container.Items {
-		objects = append(objects, i.Detail)
+	for i, item := range r.container.Items {
+		objects[len(r.headers)+i] = item.Detail
```
stuartmscott, line 240: `len(r.headers)` is recalculated for each iteration of the loop, and is calculated on line 235. Instead, consider calculating it once and storing in a variable

Jacalz: Good point. Will do that :+1:

Jacalz: I fixed it and managed to drop a few more calls to append.

stuartmscott (approved):

PR #1597, opened 2020-11-22
Expose additional method when binding to variables

andydotxyz:
Manually updating requires that we trigger the change to be read

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/binding.go
+++ b/data/binding/binding.go
@@ -15,6 +15,10 @@ type DataItem interface {
 	RemoveListener(DataListener)
 }
 
+type DataItemFromVariable interface {
+	UpdateFromVariable()
+}
+
 // DataListener is any object that can register for changes in a bindable DataItem.
 // See NewDataListener to define a new listener using just an inline function.
 type DataListener interface {
--- a/data/binding/binding.go
+++ b/data/binding/binding.go
@@ -15,10 +15,10 @@ type DataItem interface {
 	RemoveListener(DataListener)
 }
 
-// DataItemFromVariable is used by any binding that is backed by a source variable.
-type DataItemFromVariable interface {
-	// UpdateFromVariable should be called if this binding's value has been directly edited outside of the binding API.
-	UpdateFromVariable()
+// DataPointer is used by any binding that is backed by a source variable.
+type DataPointer interface {
+	// Reload should be called if this binding's value has been directly edited and the value should be reloaded.
+	Reload()
 }
 
 // DataListener is any object that can register for changes in a bindable DataItem.
--- a/data/binding/binding.go
+++ b/data/binding/binding.go
@@ -15,12 +15,6 @@ type DataItem interface {
 	RemoveListener(DataListener)
 }
 
-// DataPointer is used by any binding that is backed by a source variable.
-type DataPointer interface {
-	// Reload should be called if this binding's value has been directly edited and the value should be reloaded.
-	Reload()
-}
-
 // DataListener is any object that can register for changes in a bindable DataItem.
 // See NewDataListener to define a new listener using just an inline function.
 type DataListener interface {
--- a/data/binding/binding.go
+++ b/data/binding/binding.go
@@ -15,7 +15,9 @@ type DataItem interface {
 	RemoveListener(DataListener)
 }
 
+// DataItemFromVariable is used by any binding that is backed by a source variable.
 type DataItemFromVariable interface {
+	// UpdateFromVariable should be called if this binding's value has been directly edited outside of the binding API.
 	UpdateFromVariable()
 }
 
--- a/data/binding/binditems.go
+++ b/data/binding/binditems.go
@@ -10,10 +10,10 @@ type Bool interface {
 	Set(bool)
 }
 
-// BoolFromVariable supports binding a bool value in a Fyne application
-type BoolFromVariable interface {
+// BoolPointer supports binding a bool value in a Fyne application
+type BoolPointer interface {
 	Bool
-	DataItemFromVariable
+	DataPointer
 }
 
 // NewBool returns a bindable bool value that is managed internally.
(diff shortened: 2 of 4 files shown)


data/binding/binding.go
```diff
@@ -15,6 +15,12 @@ type DataItem interface {
 	RemoveListener(DataListener)
 }
 
+// DataItemFromVariable is used by any binding that is backed by a source variable.
```
stuartmscott, line 18: I think we could come up with a better name, maybe DataPointer, or DataAddress?

stuartmscott: Similarly; UpdateFromPointer, UpdateFromAddress?

stuartmscott: Or maybe we should follow the 'On' prefix; OnValueUpdated

andydotxyz: I will think more about the top two.
On item 3 certainly not - "On..." is for user registered callbacks, this however is equivalent of `Refresh()` as the developer is telling the toolkit something...

andydotxyz: > DataPointer, DataReference, or DataAddress

I'm not keen on pointer or reference as they can mean specific things which may or may not technically hold through the implementation. Another name in the ring, `DataItemFromSource` or `DataItemWithSource`?

data/binding/binding.go
```diff
@@ -15,6 +15,12 @@ type DataItem interface {
 	RemoveListener(DataListener)
 }
 
+// DataPointer is used by any binding that is backed by a source variable.
+type DataPointer interface {
+	// Reload should be called if this binding's value has been directly edited and the value should be reloaded.
```
stuartmscott, line 20:
```suggestion
	// Reload should be called if the value this binding points to has been directly edited and should be reloaded.
```

data/binding/binditems.go
```diff
@@ -39,35 +45,50 @@ func (b *boundBool) Get() bool {
...
+		}
+
 		*b.val = val
 	}
 
 	b.trigger()
 }
 
+func (b *boundBool) Reload() {
+	if b.val == nil {
```
stuartmscott, line 62: Can this ever be nil?

data/binding/binditems.go
```diff
@@ -39,35 +45,50 @@ func (b *boundBool) Get() bool {
 }
 
 func (b *boundBool) Set(val bool) {
-	if *b.val == val {
-		return
-	}
 	if b.val == nil { // was not initialized with a blank value, recover
```
stuartmscott, line 48: Can this ever be nil?

data/binding/binditems.go
```diff
@@ -10,16 +10,22 @@ type Bool interface {
...
 }
 
+// BoolPointer supports binding a bool value in a Fyne application
+type BoolPointer interface {
+	Bool
+	DataPointer
+}
+
 // NewBool returns a bindable bool value that is managed internally.
 func NewBool() Bool {
```
stuartmscott, line 20: If this returned `{{ .Name }}Pointer` then the cast on line 28 wouldn't be needed

data/binding/binditems.go
```diff
@@ -10,16 +10,22 @@ type Bool interface {
...
 	blank := false
 	return &boundBool{val: &blank}
 }
 
 // BindBool returns a new bindable value that controls the contents of the provided bool variable.
-func BindBool(v *bool) Bool {
+func BindBool(v *bool) BoolPointer {
```
stuartmscott, line 26: I wonder if this should be `NewBoolPointer` for consistency

data/binding/binditems.go
andydotxyz: Yes. As soon as you allow developers to pass in the pointer it can be nil

data/binding/binditems.go
andydotxyz: I don't think so personally. As you get into more complex bindings it get's really wordy - i.e. `NewStringListPointer`.
This is, after all, not returning a pointer it is returning a `Bool` binding with the additional method attached.
Revising so it is less inconsistent.

data/binding/binditems.go
andydotxyz: Hmm, maybe I added more safety since this

andydotxyz: Yeah good catch thanks. At some point nil was possible but that loophole was closed and the code not removed.

andydotxyz:
Fixed naming to avoid implying that `XxxPointer` is a whole new type

data/binding/binding.go
stuartmscott: I'm not sure about `<Data>FromPointer`, I thought we'd agreed on `<Data>Pointer`?

data/binding/binditems.go
stuartmscott: I don't think this is true as the pointer was checked by the constructor

stuartmscott:
I don't think `<Data>FromPointer` is the way to go

data/binding/binditems.go
stuartmscott: How about `func NewBound<Data>(v *<data>) <Data>Pointer`?

andydotxyz: It should not have the `New` prefix as it is not creating a new. `NewBool` creates a new bool, `BindBool` binds to an existing one.

data/binding/binditems.go
andydotxyz: Doing this would expose `Reload` on the standard `Bool` type which I am working very hard to avoid.
The pointer is internal and auto-generated, not sure we need to worry.

data/binding/binditems.go
andydotxyz: Yeah. There was another way to construct at some point, and I missed that it could no longer happen

andydotxyz:
I was attempting to overcome the raised concern of `BoolPointer` implying a new type - which it shouldn't - the output is still just a `Bool`. I would much prefer to go back to `BindBool` returning `BoolPointer`, but somewhere along the lines we're no longer in agreement on this naming.

data/binding/binding.go
andydotxyz: Yes, I added that to address the concerns about needing a `NewDataPointer` which didn't seem desirable. Rolled that back as-per Slack conversation.

toaster (approved):
LGTM

andydotxyz:
Closing this because a better way has been suggested

PR #1600, opened 2020-11-24
Bugfix: split container content width computation

toaster:
### Description:
This PR fixes two issues with the split container’s content width computation:
- the offset was applied using the container size not the available size (which excludes the divider) and therefore led to wrong results
- when the container size is not sufficient (for whatever reason) the computation led to negative sizes for the content which could trigger a crash if the content assumes a non-negative size (which is reasonable)

### Checklist:
- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Great catch thanks!

PR #1601, opened 2020-11-24
Proposal: Custom cursors in Fyne

### LINKED ISSUE #1537: Feature request: expose custom cursors

Fyne currently has a [cursorable desktop api](https://github.com/fyne-io/fyne/blob/master/driver/desktop/cursor.go) that allows overriding the cursor for a given object, using standard cursors created via `glfw.CreateStandardCursor`.

glfw has a `glfw.CreateCursor` api that takes a `image.Image` plus hotspot coordinates, which could allow widgets to expose a custom cursor. Fyne could cache cursors based on either the image.Image address, or using a value that the developer would need to provide (cache id?).

MagicalTux:
This proposal acts by changing desktop.Cursor to be an interface allowing implementation by user so that custom cursors can be used, solving issue #1537 

### Description:

There might be times when the standard cursors aren't enough to convey an action, and custom cursors may be needed. For example a hand cursor (for links) and others are likely missing from the list, and implementing every single possible cursor is likely not an option.

This works by replacing `desktop.Cursor` from its current type int to an interface with a method. The original `desktop.Cursor` becomes `desktop.StandardCursor` which implements a dummy method returning a nil value. Widgets implementing `Cursorable` can now return a `desktop.Cursor` object should they wish.

As far as I can see the change should not break any existing code, considering values such as `desktop.DefaultCursor` still fit into type `desktop.Cursor`.

Minimum implementation for a custom cursor (desktop could have a NewCursorFromImage(img image.Image, x, y int) method, not sure if it makes sense):

```go
import "image"

type fyneCursor struct {
        img  image.Image
        x, y int
}

func (f *fyneCursor) Image() (image.Image, int, int) {
        return f.img, f.x, f.y
}
```

Fixes #1537

### Checklist:

- [ ] Tests included
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -52,12 +52,12 @@ type window struct {
 	decorate   bool
 	fixedSize  bool
 
-	cursor    desktop.Cursor
-	rawCursor *glfw.Cursor // rawCursor is only filled for custom cursors in order to be destroyed when becoming un-needed
-	canvas    *glCanvas
-	title     string
-	icon      fyne.Resource
-	mainmenu  *fyne.MainMenu
+	cursor       desktop.Cursor
+	customCursor *glfw.Cursor
+	canvas       *glCanvas
+	title        string
+	icon         fyne.Resource
+	mainmenu     *fyne.MainMenu
 
 	clipboard fyne.Clipboard
 
@@ -601,12 +601,12 @@ func (w *window) mouseMoved(viewport *glfw.Window, xpos float64, ypos float64) {
 			viewport.SetInputMode(glfw.CursorMode, glfw.CursorNormal)
 			viewport.SetCursor(rawCursor)
 		}
-		if w.rawCursor != nil {
-			w.rawCursor.Destroy()
-			w.rawCursor = nil
+		if w.customCursor != nil {
+			w.customCursor.Destroy()
+			w.customCursor = nil
 		}
 		if isCustomCursor {
-			w.rawCursor = rawCursor
+			w.customCursor = rawCursor
 		}
 	}
 	if obj != nil && !w.objIsDragged(obj) {
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -564,15 +564,15 @@ func fyneToNativeCursor(cursor desktop.Cursor) (*glfw.Cursor, bool) {
 	case desktop.StandardCursor:
 		ret, ok := cursorMap[v]
 		if !ok {
-			return cursorMap[desktop.DefaultCursor], true
+			return cursorMap[desktop.DefaultCursor], false
 		}
-		return ret, true
+		return ret, false
 	default:
 		img, x, y := cursor.Image()
 		if img == nil {
 			return nil, true
 		}
-		return glfw.CreateCursor(img, x, y), false
+		return glfw.CreateCursor(img, x, y), true
 	}
 }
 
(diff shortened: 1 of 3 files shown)


internal/driver/glfw/window.go
```diff
@@ -52,11 +52,12 @@ type window struct {
 	decorate   bool
 	fixedSize  bool
 
-	cursor   *glfw.Cursor
-	canvas   *glCanvas
-	title    string
-	icon     fyne.Resource
-	mainmenu *fyne.MainMenu
+	cursor    desktop.Cursor
+	rawCursor *glfw.Cursor // rawCursor is only filled for custom cursors in order to be destroyed when becoming un-needed
```
toaster, line 56: I’d prefer a name that delivers the intent of this property. That would save the comment and also make the code in `mouseMoved` easier to understand.
AFAICS a fairly complete name would be `customGLFWCursorToBeFreed` but this is too long, of course.

I can see 2 (and a half) things that make this “cursor” different from the `cursor` property:
1. it is a GLFW cursor (I guess `raw` should mediate this)
2a. it is custom
2b. and should therefore be freed after use

Probably, we can focus on 2. Maybe `customCursor` is already enough?
Or `customRawCursor` which would contain a hint on 1, too?

internal/driver/glfw/window.go
```diff
@@ -582,12 +592,25 @@ func (w *window) mouseMoved(viewport *glfw.Window, xpos float64, ypos float64) {
...
+		if newCursor == nil {
 			viewport.SetInputMode(glfw.CursorMode, glfw.CursorHidden)
 		} else {
 			viewport.SetInputMode(glfw.CursorMode, glfw.CursorNormal)
-			viewport.SetCursor(cursor)
+			viewport.SetCursor(newCursor)
+		}
+		if oldCursor != nil {
+			// destroy old cursor now as to avoid viewport reverting to default cursor even for an instant
+			oldCursor.Destroy()
```
toaster, line 613: I guess, this could be simplified to this:
```go
		rawCursor, isStdCursor := fyneToNativeCursor(cursor)
		if rawCursor == nil {
			viewport.SetInputMode(glfw.CursorMode, glfw.CursorHidden)
		} else {
			viewport.SetInputMode(glfw.CursorMode, glfw.CursorNormal)
			viewport.SetCursor(cursor)
			viewport.SetCursor(rawCursor)
		}

		if c.rawCursor != nil {
			// destroy old cursor now as to avoid viewport reverting to default cursor even for an instant
			c.rawCursor.Destroy()
			w.rawCursor = nil
		}
		if !isStdCursor {
			w.rawCursor = rawCursor
		}
```

internal/driver/glfw/window.go
```diff
@@ -582,12 +592,25 @@ func (w *window) mouseMoved(viewport *glfw.Window, xpos float64, ypos float64) {
 
 	if w.cursor != cursor {
 		// cursor has changed, store new cursor and apply change via glfw
+		newCursor, isStdCursor := fyneToNativeCursor(cursor)
```
toaster, line 595: I suggest to invert the meaning of the `bool` returned by `fyneToNativeCursor(…)` to `isCustomCursor`.
It makes `false` the more common value and emphasises that there is something special in case it is `true`.

toaster (changes requested):
This looks very promising and I have only a couple of smaller things:

- [x] invert bool return value of `fyneToNativeCursor`
- [ ] improve name of `rawCursor` (and remove comment on it)
- [x] simplify the “set cursor” implementation
- [ ] add
    ```
    //
    // Since: 2.0.0
    ```
    to the documentation of all newly exported types or methods.

internal/driver/glfw/window.go
MagicalTux: That makes sense, thanks!

toaster (approved):

andydotxyz:
I was about to merge this in, but then I saw `desktop.Cursor(standard)` which is quite a big break in usage.
Is it possible to add this functionality without breaking developer usage of this?

On the surface the `Cursorable` interface looks the same but it's usage has changed - if avoidable it would be interesting to discuss?

andydotxyz (changes requested):
Just holding this back until we cover whether or not this should be a breaking change to the usage? i.e.
```go
func Cursor() { return desktop.TextCursor }
!=
func Cursor() { return desktop.Cursor(desktop.TextCursor) }
```

MagicalTux:
> Just holding this back until we cover whether or not this should be a breaking change to the usage? i.e.
> 
> ```go
> func Cursor() { return desktop.TextCursor }
> !=
> func Cursor() { return desktop.Cursor(desktop.TextCursor) }
> ```

Both of these would work since desktop.TextCursor satisfies the desktop.Cursor interface. Actually there is no need to add `desktop.Cursor()`.

andydotxyz:
> Both of these would work since desktop.TextCursor satisfies the desktop.Cursor interface. Actually there is no need to add `desktop.Cursor()`.

Ah yes I see your point. I was confused by the casting inside window.go and missed that it was not needed elsewhere.

andydotxyz (approved):

PR #1602, opened 2020-11-24
[fyne-cli] Add support for passing custom build tags

### LINKED ISSUE #1538: Add support for passing custom build tags
### Describe the bug:
Fyne offers a package command, however package is also building. In my case, I need to compile my binary with TAGLIB, I use the following:
`go build --tags "darwin libtag" -v`

Fyne package should be responsible for the certificate, versioning, packaging, and the building should be optional, or provide a way to allow additional tags to be configured.

lucor:
### Description:

Fixes #1538

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ More detailed release notes can be found on the [releases page](https://github.c
 
 ## 1.4.2 - Ongoing
 
+### Added
+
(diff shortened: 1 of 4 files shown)


andydotxyz (approved):
Nice :)

PR #1606, opened 2020-11-25
fixed a couple of focus manager bugs/misbehaviours

toaster:
### Description:

This PR
- ensures that disabled objects are not focusable (was already the case -> test only)
- implements the same for hidden objects
- fixes the focus manager to not focus objects that are not part of its content

This is part of the preparations for keyboard controllable menus.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/app/focus_manager.go
+++ b/internal/app/focus_manager.go
@@ -25,7 +25,10 @@ func (f *FocusManager) Focus(obj fyne.Focusable) {
 	f.Lock()
 	defer f.Unlock()
 	if obj != nil {
-		found := driver.WalkCompleteObjectTree(
+		if dis, ok := obj.(fyne.Disableable); ok && dis.Disabled() {
+			return
+		}
+		found := driver.WalkVisibleObjectTree(
 			f.content,
 			func(object fyne.CanvasObject, _, _ fyne.Position, _ fyne.Size) bool {
 				return object == obj.(fyne.CanvasObject)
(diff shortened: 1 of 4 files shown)


internal/app/focus_manager.go
```diff
@@ -73,6 +85,10 @@ func (f *FocusManager) focus(obj fyne.Focusable) {
 		obj = nil
 	}
 
+	if hid, ok := obj.(fyne.CanvasObject); ok && !hid.Visible() {
```
andydotxyz, line 88: Is this catching a specific case or a general concern that we were focusing hidden objects?
I ask because `Visible()` is not returning "is currently visible" but more like "has not been asked to hide"...
i.e. a tab container will hide tabs that are not current - but the contents of that tab will still return `true` for `Visible()`...

toaster: This was only a general concern. You’re right about the container stuff and I did not think of it.
But after doing so, I’d say that implementing a walk, to see if this object might be hidden because a parent is, is too much effort just to calm a general concern.

andydotxyz: Would it be worth using `driver.WalkVisibleObjectTree` instead of `driver.WalkCompleteObjectTree`?
That is pretty much what it is for.

toaster: Hell, yes! Why I didn’t see this?

andydotxyz (approved):

PR #1607, opened 2020-11-25
fix driver utils to apply clipping to any scrollable

toaster:
### Description:

This PR fixes the driver utils to apply clipping for all `fyne.Scrollable`s instead of scrollable `fyne.Widget` only.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/util.go
+++ b/internal/driver/util.go
@@ -155,11 +155,11 @@ func walkObjectTree(
 		children = co.Objects
 	case fyne.Widget:
 		children = cache.Renderer(co).Objects()
+	}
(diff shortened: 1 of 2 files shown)


stuartmscott (commented):
I would prefer clipping be applied to a new type `fyne.Clippable`

toaster:
> I would prefer clipping be applied to a new type `fyne.Clippable`

That might be useful but does not change the necessity of this bugfix :).

andydotxyz (approved):

PR #1608, opened 2020-11-26
New Theme API!

andydotxyz:
This PR introduces the much more flexible new theme system.
Theme API 2 (as discussed on the wiki page) allows much more flexible future for themes 👍 

Included in this PR is also a `theme.FromLegacy` which allows old themes to be loaded into the new API with a simple wrapper.
Additionally I have added a handy `theme.ExtendDefaultTheme()` which allows you to implement a partial theme and delegate the rest to standard theme. This one is kindof helpful as-is but will be facilitate the partial-JSON once we add JSON support.

This is the start of 2.0 breaking changes.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
Just an FYI, here is what I had to change to get a custom theme working when updating 1.4 to 2.0 for a third party app (beebui):
```
--- a/main.go
+++ b/main.go
@@ -11,6 +11,7 @@ import (
        "fyne.io/fyne"
        "fyne.io/fyne/canvas"
        "fyne.io/fyne/driver/desktop"
+       "fyne.io/fyne/theme"
 
        "github.com/skx/gobasic/eval"
        "github.com/skx/gobasic/tokenizer"
@@ -256,7 +257,7 @@ func (b *beeb) runProg(prog string) {
 // Show starts a new beeb computer simulator
 func Show(app fyne.App) {
        b := beeb{}
-       app.Settings().SetTheme(&beebTheme{})
+       app.Settings().SetTheme(theme.FromLegacy(&beebTheme{}))
 
        window := app.NewWindow("BBC Emulator")
        window.SetContent(b.loadUI())
```

app/settings.go
```diff
@@ -31,6 +31,7 @@ type settings struct {
 	propertyLock   sync.RWMutex
 	theme          fyne.Theme
 	themeSpecified bool
+	mode           fyne.ThemeVariant
```
stuartmscott, line 34:
```suggestion
	variant           fyne.ThemeVariant
```

cmd/fyne_demo/data/icons.go
```diff
@@ -11,7 +12,7 @@ type ThemedResource struct {
 }
 
 func isLight() bool {
-	r, g, b, _ := fyne.CurrentApp().Settings().Theme().TextColor().RGBA()
+	r, g, b, _ := theme.TextColor().RGBA()
```
stuartmscott, line 15: If Text and Icon use the same colour, perhaps it should be called `Foreground` colour (consistent with `Background` colour)?

theme/legacy.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"image/color"
+
+	"fyne.io/fyne"
+)
+
+// FromLegacy returns a 2.0 Theme created from the given LegacyTheme data.
+//
+// Since: 2.0.0
+func FromLegacy(t fyne.LegacyTheme) fyne.Theme {
```
stuartmscott, line 12: Should this have a deprecation path?

theme/theme.go
```diff
@@ -11,28 +11,105 @@ import (
...
+
+	// Variants lists each of the known theme variants that can apply to a color lookup.
+	//
+	// Since 2.0.0
+	Variants = struct {
+		Dark  fyne.ThemeVariant
+		Light fyne.ThemeVariant
+		// potential for adding theme types such as high visibility or monochrome...
+	}{
+		1,
```
stuartmscott, line 102: If this starts at 1, what does 0 represent?

app/settings.go
andydotxyz: I was trying to avoid duplication...

theme/legacy.go
andydotxyz: I don't know what you mean - this is the deprecation path.
If we mark this as deprecated in 2.0 then people won't want to use it.
The only normal alternative seemed like a TODO which didn't read well.
I could add a ticket to the 3.0 milestone perhaps? or to 2.0 with a "Not until later" mark?

theme/theme.go
andydotxyz: Good catch, there was an "unspecified" concept for a while but I removed it - revising now

theme/legacy.go
stuartmscott: Right right, my comment wasn't clear (and also in the wrong place). `fyne.LegacyTheme` is the deprecation path for the old `fyne.Theme`, but will `fyne.LegacyTheme` eventually be deprecated and removed too? If so a ticket on the 3.0 milestone to do this would fine, but we should also warn devs in the usual way

app/settings.go
stuartmscott: Hmm, I think adding this another, similar term confuses more that the duplication would cause stutter, besides there is already `theme fyne.Theme` above

theme/theme.go
```diff
@@ -11,28 +11,105 @@ import (
...
 	ColorGray = "gray"
 )
 
+var (
+	// Colors specifies each of the known colour names that a theme can contain.
+	//
+	// Since 2.0.0
+	Colors = struct {
+		Background     fyne.ThemeColorName
+		Foreground     fyne.ThemeColorName
```
stuartmscott, line 53: nit: let's keep the alphabetical ordering

theme/theme.go
```diff
@@ -11,28 +11,105 @@ import (
...
+
+	// Sizes specifies each of the known size names that a theme can contain.
+	//
+	// Since 2.0.0
+	Sizes = struct {
+		Padding        fyne.ThemeSizeName
+		ScrollBar      fyne.ThemeSizeName
+		ScrollBarSmall fyne.ThemeSizeName
+		Text           fyne.ThemeSizeName
+		InlineIcon     fyne.ThemeSizeName
```
stuartmscott, line 85: nit: ditto - it isn't a public API, but the alphabetical ordering would be nice :)

stuartmscott (commented):
Few minor nits, otherwise looks good

test/markup_renderer.go
```diff
@@ -390,14 +390,14 @@ func knownColor(c color.Color) string {
...
-		rgbaColor(theme.DisabledTextColor()):   "disabled text",
+		rgbaColor(theme.DisabledColor()):       "disabled",
 		rgbaColor(theme.FocusColor()):          "focus",
 		rgbaColor(theme.HoverColor()):          "hover",
 		rgbaColor(theme.PlaceHolderColor()):    "placeholder",
 		rgbaColor(theme.PrimaryColor()):        "primary",
 		rgbaColor(theme.ScrollBarColor()):      "scrollbar",
 		rgbaColor(theme.ShadowColor()):         "shadow",
-		rgbaColor(theme.TextColor()):           "text",
+		rgbaColor(theme.ForegroundColor()):     "foreground",
```
toaster, line 400: That should be inserted at the correct place regarding order.

theme/extend.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"image/color"
+
+	"fyne.io/fyne"
+)
+
+// ExtendDefaultTheme allows a theme to specify alterations to the default theme.
+// It will delegate all colour, theme or font lookups ignored to the built-in theme.
+func ExtendDefaultTheme(t fyne.Theme) fyne.Theme {
```
toaster, line 13: What is the advantage of using

```go
func newMyTheme() fyne.Theme {
    return ExtendDefaultTheme(&myTheme{})
}

type myTheme struct {}

func (t *myTheme)Color(n fyne.ThemeColorName, v fyne.ThemeVariant) color.Color {
    if whatever {
        return color.Black
    }
    return nil
}
```

instead of

```go
func newMyTheme() fyne.Theme {
    return &myTheme{}
}

type myTheme struct {}

func (t *myTheme)Color(n fyne.ThemeColorName, v fyne.ThemeVariant) color.Color {
    if whatever {
        return color.Black
    }
    return theme.Current().Color(n, v)
}
```

The only thing latter would need is to give access to the current theme or define `Color()`, `Size()` and `Font()` on `theme` package.
I think the latter would be easier to understand.

toaster (changes requested):
A couple of questions and suggestions, see comments.

cmd/fyne_demo/data/icons.go
andydotxyz: thanks yes, done

theme/legacy.go
andydotxyz: I'll open the ticket - but adding deprecation is something to do later. We had a previous discussion about how we want to avoid releasing deprecated APIs.

app/settings.go
andydotxyz: fixed

theme/legacy.go
andydotxyz: I added a note, I think this is the best we can do for 2.0

stuartmscott (approved):

theme/extend.go
toaster: Okay, maybe we should export `theme.Default()` instead of creating a new type, then?

toaster: Agreed. However, for me it feels weird to have to call a special method instead of simply wrapping up. Maybe we should not deprecate `theme.DefaultTheme()` but instead let it cache and create the _one_ default theme instance?

andydotxyz: I'm not sure which deprecation you are referring to, I can't find `theme.DefaultTheme()` in 1.4.
Can you expand on what you mean by "simply wrapping up" - I would have thought "ExtendDefaultTheme" was doing wrapping, so I can't tell the difference.

toaster: Sorry, my fault. I misread it.
So, I suggest to let `theme.DefaultTheme()` cache the _one_ default theme instance.
Then one could simply do:
```go
func newMyTheme() fyne.Theme {
    return &myTheme{}
}

type myTheme struct {}

func (t *myTheme)Color(n fyne.ThemeColorName, v fyne.ThemeVariant) color.Color {
    if whatever {
        return color.Black
    }
    return theme.DefaultTheme().Color(n, v)
}
```

andydotxyz: Hmm. So you propose adding `DefaultTheme` in preference to `ExtendDefaultTheme` and require that all themes be complete through delegation rather than allowing partial implementations?

I don't feel particularly that one API is stronger than the other.

toaster: I suggest that we use the mechanisms that are built into Go for embedding instead of implementing other patterns for the same goal.
My suggestion saves the whole `themeWrapper` and the `ExtendDefaultTheme()` method and does not require a single line of extra code.
It even does not _need_ the cache for the default theme but of course its performance would be bad if the default theme would be initialized on every call to `DefaultTheme()`. Even if you still disagree to my suggestion, the `DefaultTheme()` method should not create a new default theme every time.

toaster (approved):

stuartmscott (approved):

PR #1612, opened 2020-11-27
Apply min macOS version as 10.11 like the app store packaging

### LINKED ISSUE #886: fyne package macOS missing mmacosx-version-min
My binary runs in mac but the terminal window opens and I want to distribute it with an icon and maybe a nice .dmg image disk installer.

This feels like a whole other part of the fyne echo system if it doesn't already exist?

A whole documented system for how to release your final fyne binary on mac, windows and linux.

mac

when I right click on a normal MacOS app I get "Show Package Contents" menu option and then all this inside the .app file ^

What's the process for distributing apps created with fyne to the Mac App Store?

andydotxyz:
This suppports most computers sold in the last 12 years
and goes beyond Apple support plans.

Fixes #886

Can test with:
`otool -l MyApp.app/Contents/MacOS/MyApp | grep -b4 VERSION_MIN_MACOSX`

### Checklist:

- [ ] Tests included. <- manual inspection of the output binary
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/commands/build.go
+++ b/cmd/fyne/commands/build.go
@@ -20,18 +20,23 @@ func (b *builder) build() error {
 		goos = targetOS()
 	}
 
-	args := []string{}
+	args := []string{"build"}
+	env := os.Environ()
+	env = append(env, "CGO_ENABLED=1") // in case someone is trying to cross-compile...
+
 	if goos == "windows" {
 		if b.release {
-			args = append(args, "build", "-ldflags", "-s -w -H=windowsgui")
+			args = append(args, "-ldflags", "-s -w -H=windowsgui")
 		} else {
-			args = append(args, "build", "-ldflags", "-H=windowsgui")
+			args = append(args, "-ldflags", "-H=windowsgui")
 		}
 	} else {
+		if goos == "darwin" {
+			env = append(env, "CGO_CFLAGS=-mmacosx-version-min=10.11")
+			env = append(env, "CGO_LDFLAGS=-mmacosx-version-min=10.11")
+		}
 		if b.release {
-			args = append(args, "build", "-ldflags", "-s -w")
-		} else {
-			args = append(args, "build")
+			args = append(args, "-ldflags", "-s -w")
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


Jacalz (commented):
This is a really nice change. Go 1.15 dropped support for 10.11, should we perhaps instead have 10.12 as the least supported?

andydotxyz:
Thanks, but I'm not sure I agree - We still support Go 1.12 so we cannot only support target platforms of Go 1.15.
The 32bit support was a little different as we litereally could not build the apps any more - but for OS versioning we should be OK to support older.

cmd/fyne/commands/build.go
```diff
@@ -20,18 +20,23 @@ func (b *builder) build() error {
...
 		if b.release {
-			args = append(args, "build", "-ldflags", "-s -w -H=windowsgui")
+			args = append(args, "-ldflags", "-s -w -H=windowsgui")
 		} else {
-			args = append(args, "build", "-ldflags", "-H=windowsgui")
+			args = append(args, "-ldflags", "-H=windowsgui")
 		}
 	} else {
+		if goos == "darwin" {
+			env = append(env, "CGO_CFLAGS=-mmacosx-version-min=10.11")
```
Jacalz, line 35: Should we perhaps move out the minimum version to a string contestant at the top of the file? I imagine something like `const minimumSupportedOSX = “10.11”?

Jacalz (changes requested):
Oh, I see. The removal of 1.15 was why I asked. Did not know that we literally couldn’t build them. Left a little suggestion, otherwise I think it is fine 🙂

cmd/fyne/commands/build.go
andydotxyz: I wasn't sure. It also appears in the release packaging file, so a single const just here might imply something it shouldn't? (i.e. that changing just once changes all)

Jacalz: Good point. We can change it in the future in case it seems relevant then.

Jacalz (approved):

cmd/fyne/commands/build.go
andydotxyz: Yes. We could consider it being a configuration - but we don't currently do things at that level yet.
Maybe once we have an (optional) metadata file we could do more...

PR #1613, opened 2020-11-27
Fix issue where dialog shadows may not resize

### LINKED ISSUE #1370: Dialog shadow does not resize after Refresh
### Describe the bug:
If the contents of a dialog changes, the shadow is not resized

### Screenshots:
https://user-images.githubusercontent.com/1117080/95370444-9ea5a980-088d-11eb-9609-f3943a53c305.png

andydotxyz:
The inner size was not updated when our content forced it to.
We cannot override it or it won't shrink again, so this fixes the calculation for shadow size.

Fixes #1370

### Checklist:

- [x] Tests included. <- previoous tests had committed broken images, now resolved
- [x] Tests all pass.


stuartmscott (approved):
Thanks!

PR #1615, opened 2020-11-27
Allow glCanvas.Focus() to focus every focusable object of its content

toaster:
### Description:

This PR gives `glCanvas.Focus()` the ability to focus an object regardless of its content layer.
Prior it was only possible to focus an object of the top layer (menu, top-overlay, content in this order).
Also an error is logged if `glCanvas.Focus()` is called with an object which is not part of the canvas’ content.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


internal/app/focus_manager.go
```diff
@@ -21,25 +21,42 @@ func NewFocusManager(c fyne.CanvasObject) *FocusManager {
 }
 
 // Focus focuses the given obj.
-func (f *FocusManager) Focus(obj fyne.Focusable) {
+func (f *FocusManager) Focus(obj fyne.Focusable) bool {
 	f.Lock()
 	defer f.Unlock()
 	if obj != nil {
-		if dis, ok := obj.(fyne.Disableable); ok && dis.Disabled() {
```
andydotxyz, line 28: Can you say why this early exit was dropped to the end?

toaster: Because we should return `false` if the object is not part of the content but `true` if it is, regardless if we actually changed the focus.

andydotxyz:
As this is a change to behaviour, can we get the equivalent change in the mobile driver?

toaster:
> As this is a change to behaviour, can we get the equivalent change in the mobile driver?

I think this should be a separate PR because `mobileCanvas` diverged from `glCanvas` long before this PR when it did not use `FocusManager`.

andydotxyz:
> I think this should be a separate PR

Can this please be captured somewhere? An issue or some note that we have diverged behavior in the drivers that needs to be resolved?

toaster:
Here you go: #1625

andydotxyz (approved):

PR #1616, opened 2020-11-27
Multiple fixes around entry.SetValidationError()

Jacalz:
### Description:
This fixes `entry.SetValidationError()` that broke when moving validation to `.Refresh()`. We also stop validating on `.Refresh()` as that can be quite expensive with complex validators and a lot of text. It also makes sure to call `onValidationChanged()` if the validation did change.

@stuartmscott Please check that this works for you as you were the one to suggest moving validation to `.Refresh()`. I think calling `entry.SetValidationError()` should suffice. If not, we can easily make `.Validate()` update the status in the widget.

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@ More detailed release notes can be found on the [releases page](https://github.c
 ### Changed
 
 * Run validation on content change instead of on each Refresh in widget.Entry
-* Run onValidationChanged() when calling entry.SetValidationError()
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ More detailed release notes can be found on the [releases page](https://github.c
 
 * [fyne-cli] Add support for passing custom build tags (#1538)
 
+### Changed
+
+* Run validation on content change instead of on each Refresh in widget.Entry
+* Run onValidationChanged() when calling entry.SetValidationError()
+
 ### Fixed
 
 * [fyne-cli] Android: allow to specify an inline password for the keystore
@@ -19,6 +24,7 @@ More detailed release notes can be found on the [releases page](https://github.c
 * Support older macOS by default - back to 10.11 (#886)
 * Complete certification of macOS App Store releases (#1443)
 * Fix compilation errors for early stage Wayland testing
+* Fix entry.SetValidationError() not working correctly
 
 
 ## 1.4.1 - 20 November 2020
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -1000,15 +1000,7 @@ func (e *Entry) updateText(text string) {
 	})
 
 	if validate := e.Validator; validate != nil {
-		if err := validate(text); err != e.validationError {
-			e.validationError = err
-
-			if f := e.onValidationChanged; f != nil {
-				f(err)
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ More detailed release notes can be found on the [releases page](https://github.c
 
 * [fyne-cli] Add support for passing custom build tags (#1538)
 
+### Changed
+
+* Run validation on content change instead of on each Refresh in widget.Entry
+* Run onValidationChanged() when calling entry.SetValidationError()
```
andydotxyz, line 15: I don't think we should list internal details in a changelog.

andydotxyz (changes requested):
I don't know quite how to put this but I don't think 3 lines in the changelog reflects the impact of this change.

Adding it to the changed section should only be needed if this is going to impact the resulting app or how the API is used IMHO.

CHANGELOG.md
Jacalz: Good point. It also kind of ties in with it not working correctly (in widget.Form that is) :)

widget/entry.go
```diff
@@ -999,6 +999,18 @@ func (e *Entry) updateText(text string) {
...
+	if validate := e.Validator; validate != nil {
+		if err := validate(text); err != e.validationError {
+			e.validationError = err
+
+			if f := e.onValidationChanged; f != nil {
+				f(err)
+			}
+		}
+
+		e.validationStatus.Refresh()
```
stuartmscott, line 1011: Would this be simpler?
```suggestion
			e.setValidationError(err)
		} else {
			e.validationStatus.Refresh()
```

widget/entry_validation.go
```diff
@@ -29,10 +29,15 @@ func (e *Entry) SetOnValidationChanged(callback func(error)) {
 // SetValidationError manually updates the validation status until the next input change
 func (e *Entry) SetValidationError(err error) {
 	e.validationError = err
+	if e.Validator == nil {
+		return
+	}
 
-	if e.Validator != nil {
-		e.validationStatus.Refresh()
+	if err != e.validationError && e.onValidationChanged != nil {
```
stuartmscott, line 36: `err != e.validationError` will never be true because of line 31

widget/entry.go
Jacalz: Thanks for the suggestion. I slimmed things down even more :)

widget/entry_validation.go
Jacalz: Good point. I don't know why that slipped in.

CHANGELOG.md
andydotxyz, line 15: The way this is expressed it's an internal detail. Either it does not belong here, or it should describe why it is a developer facing change. Maybe it's part of the line above?
Not sure that having it as a change as well as a fix is required? We don't list changes that are bug fixes.

andydotxyz (changes requested):

CHANGELOG.md
Jacalz: Oh, sorry. Was pretty sure that I had fixed this.

andydotxyz (approved):

PR #1617, opened 2020-11-28
Horizontal scrolling Select widget

mrjrieke:
### Description:
The popup ticker widget.  This is an endlessly scrolling selection widget allowing scrolling from left to right.  It's a bit like a selection drop down box, but horiziontal and without bar separators.  It's been a challenge to implement, but it's been fun.  ;)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


mrjrieke:
This is just a new widget I've been toying with in my own free time.  Not sure if there is interest in merging it into the common repo or not.  I've had to toy with some changes to lower levels in order to get it recognized.  So really, just submitting in order to get feedback from the community.  I have no testing in place and static analysis is failing, so this is not even close to being in a merged in any way.  If there is interest, I can make further changes to make it more acceptable...  fix static checks, add testing, etc...

andydotxyz:
Thanks for sharing this. Is it possible to share a screenshot of the widget, and maybe a description of when you expect this would be useful?

I noticed what you have said was required internal changes but they should not have been needed. First of all the new `EventConsumer` type should not be needed - indeed I can't see where it is used.
You also added Entry.HideSeparator but I cannot see where that is set either.

mrjrieke:
I thought you might notice those other changes.  Didn't mean for those to slip into this pr.  I actually need those changes for something else I'm working on.  I need to be able to listen to events from the entry widget, but they aren't needed for this widget.  I'll work on removing those.  I'll try to get a snapshot of the widget by making a tiny example app.  When I had a friend try it out on a Mac, it mostly worked, but there was some odd behavior there that I'm going to have to look into.  It would help if I had someone else with a Mac to help me fix my Mac bug - if they exist.  :)

Also, what is expected of the unit test if I were to build one out for this project?

andydotxyz:
> I need to be able to listen to events from the entry widget

You shouldn't need to do it this way. If you extend the Entry widget you can listen to any events and delegate them to the original if you want.

For unit testing you can check out our other code. We have behaviour based unit tests, API based tests as well as image validation for the rendering.

mrjrieke:
> > I need to be able to listen to events from the entry widget
> 
> You shouldn't need to do it this way. If you extend the Entry widget you can listen to any events and delegate them to the original if you want.
-- This mostly works... Unfortunately, I can't delegate to entryRenderer due to it's private nature.  Would it be possible to make that public?  Otherwise, to customize that, I'll have to write a *lot* of code there.
> For unit testing you can check out our other code. We have behaviour based unit tests, API based tests as well as image validation for the rendering.
-- ok thanks for the high level view.  I'll take a look.

mrjrieke:
> > -- This mostly works... Unfortunately, I can't delegate to entryRenderer, text provider, and placeholder due to their private nature.  Would it be possible to make that and its fields more public?  Otherwise, to customize that, I'll have to write a _lot_ of code there.  But maybe that's to be expected...  Thoughts?
-- FYI -- just for fun, I started to try to do this... and it's a deeep rabbit hole.  :)  I'm going to slowly back away from this approach for now... Probably best for me to just branch these minimal changes to widget entry so I can back them out of this pr.

mrjrieke:
One more thought -- maybe an easier way would be to style the different elements in the widget so you don't have to mess with the existing flow/layout logic.  I know I could change the theme defaults, but that would mess all other widgets that might rely on the default theme styles.

andydotxyz:
We don't support having individual standard element themed specifically - the widget package is designed to be uniform.
Removing an underline form one entry would remove the hint that it's an entry.

> Unfortunately, I can't delegate to entryRenderer, text provider, and placeholder

Yes. these are render details, this is separate to the widget itself which supports overriding in the way I described.

mrjrieke:
> We don't support having individual standard element themed specifically - the widget package is designed to be uniform.
> Removing an underline form one entry would remove the hint that it's an entry.
As an end user, I appreciate being able to customize widgets individually.  Styling is a pattern used in the web world that I find quite nice.  For now, I'm backing the change out since it doesn't have to do with the ticker popup.

> > Unfortunately, I can't delegate to entryRenderer, text provider, and placeholder
> 
> Yes. these are render details, this is separate to the widget itself which supports overriding in the way I described.

andydotxyz:
> As an end user, I appreciate being able to customize widgets individually. Styling is a pattern used in the web world that I find quite nice

We are taking a different design approach to web (which started as a document delivery system).
As you can see from the widget API we focus on behavior and functionality, with the design for our built-in components defined by the theme.

internal/driver/glfw/canvas.go
```diff
@@ -160,7 +161,7 @@ func (c *glCanvas) Resize(size fyne.Size) {
 			// TODO: remove this when #707 is being addressed.
 			// “Notifies” the PopUp of the canvas size change.
 			p.Resize(p.Content.Size().Add(fyne.NewSize(theme.Padding()*2, theme.Padding()*2)))
-		} else if p, ok := overlay.(*widget.TickerPopUp); ok {
+		} else if p, ok := overlay.(*dialog.TickerPopUp); ok {
```
andydotxyz, line 164: This special case should not be required, the size management should be possible within the code.
See note above how special cases are being removed

mrjrieke: > This special case should not be required, the size management should be possible within the code.
> See note above how special cases are being removed
I'd love to be able to remove this but haven't figured out how yet.  So for now it stays.  I'll try to figure it out before I publish this pr.

andydotxyz:
We have now created a new extensions repository for development of community contributions.
I think it would be best if you could open a new PR in `fyne-x` instead https://github.com/fyne-io/fyne-x/pulls.
This repository is imported as `fyne.io/x/fyne` and you could add the widget at `fyne.io/x/fyne/widget`.

The testing infrastructure is set up so it should provide a test bed to get the component ready to merge.
Over time community widgets may be promoted to core :).

Once moved over we will close this PR.

andydotxyz:
Hi @mrjrieke are you still working on this? Will you be able to move it to the `fyne-io/fyne-x` repo?
I'm just working to tidy our PR backlog and realised this was still here.

Jacalz:
Hi @mrjrieke. Are you prepared to close this and move it to `fyne-x` any time soon? :)

internal/driver/glfw/canvas.go
mrjrieke: Still haven't found way to deal with this elegantly.

mrjrieke:
Sorry, don't have time to move it these days.  Closing it.  I rebuilt my local fork because it was in complete chaos. So even if I wanted to fix this, I couldn't.  Maybe some day if I can spend all my time on my pet projects instead of working a normal 9-5 job, I'll have time to work this into fyne-x.

PR #1619, opened 2020-11-29
Expose MiddleMouseButton on desktop

MagicalTux:
making it the third option so RightMouseButton stays the same, also use glfw values for left/right instead of Button1 and Button2

### Description:
This adds MiddleMouseButton, and changes `glfw.MouseButton1` and 2 to `glfw.MouseButtonLeft` and `glfw.MouseButtonRight` which are easier to read.

No matching issue, but the middle mouse button was commented with TODO.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/driver/desktop/mouse.go
+++ b/driver/desktop/mouse.go
@@ -11,6 +11,9 @@ const (
 
 	// RightMouseButton is the secondary button on most mouse input devices.
 	RightMouseButton
+
+	// MiddleMouseButton is the middle button on the mouse, assuming it has one.
+	MiddleMouseButton
 )
 
 // MouseEvent contains data relating to desktop mouse events
--- a/driver/desktop/mouse.go
+++ b/driver/desktop/mouse.go
@@ -12,7 +12,9 @@ const (
 	// RightMouseButton is the secondary button on most mouse input devices.
 	RightMouseButton
 
-	// MiddleMouseButton is the middle button on the mouse, assuming it has one. Since: 2.0.0
+	// MiddleMouseButton is the middle button on the mouse, assuming it has one.
+	//
+	// Since: 2.0.0
 	MiddleMouseButton
 )
 
--- a/driver/desktop/mouse.go
+++ b/driver/desktop/mouse.go
@@ -12,7 +12,7 @@ const (
 	// RightMouseButton is the secondary button on most mouse input devices.
 	RightMouseButton
 
-	// MiddleMouseButton is the middle button on the mouse, assuming it has one.
+	// MiddleMouseButton is the middle button on the mouse, assuming it has one. Since: 2.0.0
 	MiddleMouseButton
 )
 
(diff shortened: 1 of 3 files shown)


driver/desktop/mouse.go
```diff
@@ -11,6 +11,9 @@ const (
 
 	// RightMouseButton is the secondary button on most mouse input devices.
 	RightMouseButton
+
+	// MiddleMouseButton is the middle button on the mouse, assuming it has one.
+	MiddleMouseButton
```
andydotxyz, line 16: Please add Since: 2.0.0

andydotxyz (changes requested):
Nice addition, just missing the flag for new API

toaster (changes requested):
While “left” and “right” are easier to read than “1” and “2”, the name is misleading because for left-handed people it’s the other way around.
Maybe “first”, “second” and “third” is a better choice?
Also, what if we want to support additional mouse buttons? My MX Anywhere has two additional buttons (if I don’t count left/right of the scroll wheel).

AlbinoGeek:
I have a Flight stick with 38 buttons that Linux detects as a mouse ❌

andydotxyz:
>  for left-handed people it’s the other way around

This is a good point. GLFW just hard-codes Left to 1

driver/desktop/mouse.go
andydotxyz: Sorry I should have been more specific.
Like the Deprecated: lines specified by Go, this needs to be on a newline

andydotxyz:
Apologies I did not flag this first time @MagicalTux but a small test would be good - even just to check the convert function

toaster:
> > for left-handed people it’s the other way around
> 
> This is a good point. GLFW just hard-codes Left to 1

They probably did so because of @AlbinoGeek 's 38 button mouse and the OS changing the meaning all the time ;).

andydotxyz:
@stuartmscott pointed out that "MouseButtonPrimary" would be in-keeping with our naming elsewhere, but this PR is not setting up our API for left/right etc, but making use of the GLFW terms, so I don't think it's problematic to leave it as is (1, 2 etc) and review left vs right in a future PR that considers how we better support left handed mice.

toaster:
> @stuartmscott pointed out that "MouseButtonPrimary" would be in-keeping with our naming […].

I thought about this, too, but gave up when I didn’t know the word for the third button let alone the other 35 buttons of @MagicalTux mouse.

toaster (approved):

stuartmscott:
> I thought about this, too, but gave up when I didn’t know the word for the third button let alone the other 35 buttons of @MagicalTux mouse.

Seems an outlier, most mice have fewer than 10 buttons, and this PR is not trying to support that many buttons anyway. I'd say it will suffice to just add 3; `primary, secondary, tertiary` corresponding to `left, right, middle` for a right-handed user.

Longer term we could use names for the first 10;
`primary, secondary, tertiary, quaternary, quinary, senary, septenary, octonary, nonary denary`
and then introduce another type `MouseButtonExtended` that can hold any larger button ID.

andydotxyz:
Not sure that it makese sense naming the buttons > 3... Primary is a clear function, secondary another and in some products there is a tertiary button. But beyond that it really is button mappings to a number.

andydotxyz (approved):

PR #1620, opened 2020-11-30
Ensure Collections calculate MinSize consistently

stuartmscott:
### Description:
All three collections will not return the larger of a element's minsize and the scroller's minsize.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -213,7 +213,7 @@ func (l *listRenderer) Layout(size fyne.Size) {
 }
 
 func (l *listRenderer) MinSize() fyne.Size {
-	return l.scroller.MinSize()
+	return l.scroller.MinSize().Max(l.list.itemMin)
 }
 
 func (l *listRenderer) Refresh() {
--- a/widget/table.go
+++ b/widget/table.go
@@ -244,7 +244,7 @@ func (t *tableRenderer) Layout(s fyne.Size) {
 }
 
 func (t *tableRenderer) MinSize() fyne.Size {
-	return t.t.scroll.MinSize().Max(t.cellSize)
+	return t.t.scroll.MinSize().Max(t.cellSize.Add(fyne.NewSize(theme.Padding(), 0)))
 }
 
 func (t *tableRenderer) Refresh() {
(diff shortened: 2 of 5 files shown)


okratitan (approved):

andydotxyz (changes requested):
Good change, but I think there is a test missing for this.
Possibly just Table?

stuartmscott:
Tests added

okratitan (dismissed):

widget/table.go
```diff
@@ -244,7 +244,7 @@ func (t *tableRenderer) Layout(s fyne.Size) {
 }
 
 func (t *tableRenderer) MinSize() fyne.Size {
-	return t.cellSize
+	return t.t.scroll.MinSize().Max(t.cellSize.Add(fyne.NewSize(theme.Padding(), 0)))
```
andydotxyz, line 247: If we are leaving space for the indicators then table needs to have the same on the vertical direction

widget/list.go
```diff
@@ -213,7 +213,7 @@ func (l *listRenderer) Layout(size fyne.Size) {
 }
 
 func (l *listRenderer) MinSize() fyne.Size {
-	return l.scroller.MinSize()
+	return l.scroller.MinSize().Max(l.list.itemMin)
```
andydotxyz, line 216: Does itemMin include the padding amount for the indicator?

andydotxyz (changes requested):
Following discussions in Slack I don't think we're leaving enough space for each of them to show indicators and a full cell/item/node on screen at minimum size.

The one I am sure about I commented inline - table needs to have padding at the top as well

widget/list.go
okratitan: Yes

stuartmscott: Yup

widget/table.go
stuartmscott: Done

stuartmscott:
Added padding for column selection indicator as well

okratitan (approved):

andydotxyz (approved):

Issue #761, opened 2020-03-23
Support for tab character


andydotxyz:
### Is your feature request related to a problem? Please describe:
Strings with a tab character display a block instead of the correct spaces.

### Is it possible to construct a solution with the existing API?
No, I can search for \t and replace with spaces but this is not the same (no alignment to tab stops).

### Describe the solution you'd like to see:
In label at least (tab in entry is not possible I guess?) and the new textgrid, pressing tab should move to the next “tab stop” I.e. multiple of a number of m-spaces width (8 by default?) and continue render from that position.

andydotxyz:
Minimamal is required for 2.0.
Could just be 4 spaces **FOR NOW**.

adrianre12:
I have a POC of how to do it written, just need to rebase it to 2.0 then I will share.

PR #1626, opened 2020-12-04
emit KeyDown/KeyUp for Tab key

MagicalTux:
### Description:

KeyDown/KeyUp are not triggered for the Tab key. This fixes the issue by moving the processing of tab key to after KeyDown/KeyUp are triggered.

Note that the previous `if action == glfw.Release` (in KeyDown/KeyUp processing) will cause a return meaning the `if action != glfw.Release` in the Tab key handling will never be false, so I removed it and added a comment instead so the intent isn't lost.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -977,19 +977,6 @@ func (w *window) keyPressed(_ *glfw.Window, key glfw.Key, scancode int, action g
 	keyEvent := &fyne.KeyEvent{Name: keyName}
 	keyDesktopModifier := desktopModifier(mods)
 
-	if keyName == fyne.KeyTab {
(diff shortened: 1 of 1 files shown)


andydotxyz:
Possibly relates to #761

MagicalTux:
This would emit KeyDown/KeyUp for Tab but no `TypedKey` since it still returns prior to reaching that event trigger, which keeps existing behavior (no `TypedKey` or `TypedRune` event generated for tab), so it would not impact #761, which I think is more about handling the display of tab character (which the display engine should replace with spacing up to the next stop point).

andydotxyz (approved):

PR #1627, opened 2020-12-04
create new image scaling modes ImageScaleLinear and ImageScaleNearest…

MagicalTux:
### Description:
create new image scaling modes `ImageScaleLinear` and `ImageScaleNearest` allowing to disable software scaling and improve performances where beauty is not required

Fixes performance issues on large canvas being updated often.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
I had a quick chat with @MagicalTux and it looks like maybe pixel and nearest are potentially duplicating, so maybe we can reduce this to 3 modes in total :)

MagicalTux:
Made some changes for this, I did keep the `case canvas.ImageScalePixels` down in the function since it won't impact performances and can be useful as a fallback should we be able to add checks for hardware support (ie. if the target is not opengl and doesn't support scaling, perform software scaling anyway)

internal/painter/software/draw.go
```diff
@@ -82,10 +82,16 @@ func drawImage(c fyne.Canvas, img *canvas.Image, pos fyne.Position, base *image.
...
+		// do not scale or duplicate image since not needed, draw directly
+		drawTex(x, y, width, height, base, origImg, clip)
+		return
+	}
+
 	scaledBounds := image.Rect(0, 0, width, height)
 	scaledImg := image.NewNRGBA(scaledBounds)
 	switch mode {
-	case canvas.ImageScalePixels:
+	case canvas.ImageScalePixels, canvas.ImageScaleFastest:
```
andydotxyz, line 94: Thanks for changing this method, it does need to scale.
However I think the addition of Fastest here will give a pixelated output - was that your intent?

MagicalTux: That's true. I'm guessing `BiLinear` would be too slow, so maybe `ApproxBiLinear`?

andydotxyz (approved):

PR #1634, opened 2020-12-06
Adding DataList to binding and add binding support to widget.List

andydotxyz:
Adding list type bindings and bind them to widget.List.

Still TODO:

- [x] Bind/UnBind on widget to be cleaner
- [x] Add all the missing Since: 2.0.0 to all of data binding!
- [x] Figure out how list can more cleanly de-register when items go away <- Bind/Unbind did it :)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


cmd/fyne_demo/tutorials/bind.go
```diff
@@ -37,6 +37,31 @@ func bindingScreen(_ fyne.Window) fyne.CanvasObject {
...
+			text := obj.(*fyne.Container).Objects[0].(*widget.Label)
+			text.Bind(binding.FloatToStringWithFormat(item.(binding.Float), "item %0.1f"))
+
+			btn := obj.(*fyne.Container).Objects[1].(*widget.Button)
+			btn.OnTapped = func() {
+				item.(binding.Float).Set(item.(binding.Float).Get() + 1)
+			}
+		}))
+
+	return container.NewBorder(item, nil, nil, nil, list)
```
stuartmscott, line 66: Could these two border layouts be merged?
```suggestion
	list := widget.NewListWithData(dataList, ...
	return container.NewBorder(item, button, nil, nil, list)
```

cmd/fyne_demo/tutorials/bind.go
```diff
@@ -37,6 +37,31 @@ func bindingScreen(_ fyne.Window) fyne.CanvasObject {
...
+			return container.NewBorder(nil, nil, nil, widget.NewButton("+", nil),
+				widget.NewLabel("item x.y"))
+		},
+		func(item binding.DataItem, obj fyne.CanvasObject) {
+			text := obj.(*fyne.Container).Objects[0].(*widget.Label)
+			text.Bind(binding.FloatToStringWithFormat(item.(binding.Float), "item %0.1f"))
+
+			btn := obj.(*fyne.Container).Objects[1].(*widget.Button)
+			btn.OnTapped = func() {
+				item.(binding.Float).Set(item.(binding.Float).Get() + 1)
```
stuartmscott, line 62: The cast `item.(binding.Float)` appears thrice in this function, consider pulling out into a variable

data/binding/listbinding.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+type listBase struct {
+	base
+	val []DataItem
+}
+
+// GetItem returns the DataItem at the specified index.
+func (b *listBase) GetItem(i int) DataItem {
+	if i >= len(b.val) {
```
stuartmscott, line 19: Might also need a check that `i` isn't negative

widget/check.go
```diff
@@ -100,6 +100,30 @@ type Check struct {
...
+	checkSource   binding.Bool
+	checkListener binding.DataListener
+}
+
+// Bind connects the specified data source to this Check.
+// The current value will be displayed and any changes in the data will cause the widget to update.
+// User interactions with this Check will set the value into the data source.
+//
+// Since: 2.0.0
+func (c *Check) Bind(data binding.Bool) {
```
stuartmscott, line 113: `Bind` should first call `Unbind` to avoid the widget getting updated by a previously added listener, see #804

widget/check.go
```diff
@@ -257,3 +272,17 @@ func (c *Check) TypedRune(r rune) {
...
+//
+// Since: 2.0.0
+func (c *Check) Unbind() {
+	c.OnChanged = nil
+	if c.checkSource == nil || c.checkListener == nil {
+		return
+	}
+
+	c.checkSource.RemoveListener(c.checkListener)
+	c.checkListener = nil
```
stuartmscott, line 287: Add `c.checkSource = nil`

widget/list.go
```diff
@@ -46,6 +47,24 @@ func NewList(length func() int, createItem func() fyne.CanvasObject, updateItem
...
+		func() int {
+			return data.Length()
+		},
+		createItem,
+		func(i ListItemID, o fyne.CanvasObject) {
+			item := data.GetItem(i)
+			updateItem(item, o)
+		})
+
+	data.AddListener(binding.NewDataListener(l.Refresh))
```
stuartmscott, line 64: This is really nice and clean :)

data/binding/binding.go
```diff
@@ -44,6 +50,12 @@ func (b *base) AddListener(l DataListener) {
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
+	for _, listen := range b.listeners { // TODO maybe fix? workaround for list not being able to de-register
+		if listen == l {
+			return
+		}
+	}
```
stuartmscott, line 57: Can you elaborate on this issue? What breaks if this is removed?

data/binding/bindlists.go
```diff
@@ -0,0 +1,239 @@
...
+	Append(string)
+	Get(int) string
+	Prepend(string)
+	Set(int, string)
+}
+
+// NewStringList returns a bindable list of string values.
+//
+// Since: 2.0.0
+func NewStringList() StringList {
```
stuartmscott, line 209: We can't bind to an existing slice
```go
strings := []string{"a", "b", "c"}
bind := NewStringList(strings)
```
As this is something we want to support longer term, let's articulate the major hurdles; I would guess the StringList cannot determine if the slice size (1), or ordering (2) has changed, or if individual elements have changed (3);
```go
strings = append(strings, "d") // (1)
strings[0], strings[2] = strings[2], strings[1] // (2)
strings[0]+="aa" // (3)
``` 
Are there any I'm missing?

data/binding/binding.go
andydotxyz: Nothing, it was a hangover from earlier revision, will remove

cmd/fyne_demo/tutorials/bind.go
andydotxyz: Yes, good catch thanks

data/binding/bindlists.go
andydotxyz: Opened and added other thoughts alongside this to #1645

andydotxyz:
I think that's all the comments addressed.
Also added a (does not reflect external changes, like the primitive BindXxx calls) BindXxxList functionality.

data/binding/bindlists.go
```diff
@@ -0,0 +1,404 @@
...
+func BindBoolList(v *[]bool) BoolList {
+	if v == nil {
+		return NewBoolList()
+	}
+
+	b := &boundBoolList{val: v}
+
+	for _, i := range *v {
+		b.Append(i)
+	}
```
stuartmscott, line 37: Not sure about this, wouldn't calling `BindBoolList(&[]bool{true})` initialize `boundBoolList` with the slice pointer, and then this `for` would append the data again, resulting in `&[]bool{true, true}`?

andydotxyz: Oh darn, this was inintially split into value and binding init. 
Let me go see

andydotxyz: Fixed, good catch.
The tests missed it because the binding count was correct, but our data was duplicated

stuartmscott: Not sure about the fix, is `val` ever used now?

andydotxyz: The get looks it up as an optimisation.
Additionally we need to retain a reference to determine what changes when we get into reload requests - whatever api approach we take.

stuartmscott: > The get looks it up as an optimisation.

It will always be nil tho

andydotxyz: Oh I see. good catch.
Better fix in place - what I should have done first time.
Also tests to validate this time!

widget/entry_test.go
```diff
@@ -3211,3 +3247,7 @@ func teardownImageTest(w fyne.Window) {
 	w.Close()
 	test.NewApp()
 }
+
+func waitForBinding() {
+	time.Sleep(time.Millisecond * 100) // data resolves on background thread
+}
```
stuartmscott, line 3253: I noticed this is duplicated in the `widget` package, shouldn't all binding tests be in `widget_test`?

widget/entry.go
```diff
@@ -114,6 +118,25 @@ func NewPasswordEntry() *Entry {
...
+	e.textSource = data
+
+	e.textListener = binding.NewDataListener(func() {
+		e.SetText(data.Get())
+	})
+	data.AddListener(e.textListener)
+
+	e.OnChanged = func(s string) {
+		data.Set(s)
+	}
```
stuartmscott, line 137:
```suggestion
	e.OnChanged = data.Set
```

widget/entry.go
```diff
@@ -83,6 +85,8 @@ func NewEntry() *Entry {
 }
 
 // NewEntryWithData returns an Entry widget connected to the specified data source.
+//
+// Since: 2.0.0
```
stuartmscott, line 89: Commenting here because I can't comment on unchanged lines.

Lines 99-101
```go
entry.OnChanged = func(s string) {
	data.Set(s)
}
```
Could be simplified to;
```go
entry.OnChanged = data.Set
```

widget/check.go
```diff
@@ -100,6 +100,31 @@ type Check struct {
...
+		c.Checked = data.Get()
+		if cache.IsRendered(c) {
+			c.Refresh()
+		}
+	})
+	data.AddListener(c.checkListener)
+
+	c.OnChanged = func(b bool) {
+		data.Set(b)
+	}
```
stuartmscott, line 127:
```suggestion
	c.OnChanged = data.Set
```

widget/check.go
```diff
@@ -100,6 +100,31 @@ type Check struct {
...
+	checkSource   binding.Bool
+	checkListener binding.DataListener
+}
+
+// Bind connects the specified data source to this Check.
+// The current value will be displayed and any changes in the data will cause the widget to update.
+// User interactions with this Check will set the value into the data source.
+//
+// Since: 2.0.0
+func (c *Check) Bind(data binding.Bool) {
```
stuartmscott, line 113: Should be defined lower down, with the rest of the methods, not above the constructors.

andydotxyz: I think that it is not above constructors so much as this file has not been updated to reflect the organisation - SetChecked, Hide, MouseIn, MouseOut, MouseMoved, Tapped, MinSize and CreateRenderer are all above the constructor.

widget/check.go
andydotxyz: Thanks, updated and I realised there was an issue with constructors differing from Bind() code - simplified.

widget/entry_test.go
andydotxyz: Many of the widgets did not yet have widget_test variants.
The exact setup of testing is still not finalised in our document so I didn't want to introduce the change in this PR.
We certainly could get it clarified and applied later.

andydotxyz:
Other than the issue that some widget files have not been organised to the current scheme I think that's the comments addressed.

stuartmscott (approved):

PR #1637, opened 2020-12-08
RFC: New container.New() and container.NewWithoutLayout() functions

Jacalz:
### Description:

This is a proposal about adding `container.New()` and `container.NewWithoutLayout()` to the container package, like discussed.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/container.go
+++ b/container.go
@@ -15,13 +15,13 @@ type Container struct {
 }
 
 // NewContainer returns a new Container instance holding the specified CanvasObjects.
-// Deprecated: Use container.NewManual() to create a container that uses manual layout.
+// Deprecated: Use container.NewWithoutLayout() to create a container that uses manual layout.
 func NewContainer(objects ...CanvasObject) *Container {
 	return NewContainerWithoutLayout(objects...)
 }
 
 // NewContainerWithoutLayout returns a new Container instance holding the specified CanvasObjects that are manually arranged.
-// Deprecated: Use container.NewManual() instead.
+// Deprecated: Use container.NewWithoutLayout() instead
 func NewContainerWithoutLayout(objects ...CanvasObject) *Container {
 	ret := &Container{
 		Objects: objects,
@@ -32,7 +32,7 @@ func NewContainerWithoutLayout(objects ...CanvasObject) *Container {
 }
 
 // NewContainerWithLayout returns a new Container instance holding the specified CanvasObjects which will be laid out according to the specified Layout.
-// Deprecated: Use container.New() instead.
+// Deprecated: Use container.New() instead
 func NewContainerWithLayout(layout Layout, objects ...CanvasObject) *Container {
 	ret := &Container{
 		Objects: objects,
--- a/container.go
+++ b/container.go
@@ -21,7 +21,8 @@ func NewContainer(objects ...CanvasObject) *Container {
 	return NewContainerWithoutLayout(objects...)
 }
 
-// NewContainerWithoutLayout returns a new Container instance holding the specified CanvasObjects that are manually arranged.
+// NewContainerWithoutLayout returns a new Container instance holding the specified
+// CanvasObjects that are manually arranged.
 //
 // Deprecated: Use container.NewWithoutLayout() instead
 func NewContainerWithoutLayout(objects ...CanvasObject) *Container {
--- a/container.go
+++ b/container.go
@@ -33,7 +33,8 @@ func NewContainerWithoutLayout(objects ...CanvasObject) *Container {
 	return ret
 }
 
-// NewContainerWithLayout returns a new Container instance holding the specified CanvasObjects which will be laid out according to the specified Layout.
+// NewContainerWithLayout returns a new Container instance holding the specified
+// CanvasObjects which will be laid out according to the specified Layout.
 //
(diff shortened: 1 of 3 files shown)


andydotxyz:
Thanks for this, seems like a good idea.
However I'm having second thoughts about the `NewWithoutLayout` naming - without the `NewWithLayout` alongside it doesn't seem quite as logical - could a shorter version be found?

stuartmscott:
I thought the same, perhaps `New` and `NewWithLayout` would be better, though it implies the default is to use a container without a layout :/

andydotxyz:
Nah, the default certainly needs to be with-layout.
I guess `NewManual` or something like that would fit the naming in this package?

stuartmscott:
Maybe we're conflating two separate things. I agree container with layout should be the default. Do we need to support container without a layout, what are the use cases?

andydotxyz:
Yes we need to support a container without layout. It's called "Manual layout" and is a way for people to position items in a container individually - whereas a layout would work on the anonymous slice of elements.

stuartmscott:
How is that different from `fyne.NewContainerWithLayout(MyCustomLayout(), items...)`?

andydotxyz:
> How is that different from `fyne.NewContainerWithLayout(MyCustomLayout(), items...)`?

Well, like I thought I said - manual layout operates on individual named elements, a layout operates on an anonymous slice of elements.

Maybe think of `WithoutLayout` like the Swing `nil` layout? It's a useful way to control small areas or drawing elements.

Jacalz:
I think that I would prefer going with `.New()` and `.NewManual()`. It is the best names we have, in my opinion at least.

andydotxyz:
> I think that I would prefer going with `.New()` and `.NewManual()`

I think that's pretty clear and concise, happy to go with that.

stuartmscott:
What value does it provide other than being a struct containing a slice of fyne.CanvasObject?

andydotxyz:
No other value that I can think of. The `fyne.Container` exists to include multiple elements where otherwise only one would be possible.
Usually this contains a layout to help arrange these, but it is not required.

Jacalz:
Updated this PR with new naming.

stuartmscott (changes requested):
I disagree with the 'manual' naming. Manual implies the alternative is Automatic, and as all of this is code nothing happens manually ie there is no human work to lay these objects out.

Perhaps container is conflating two things, one which just groups a set of canvasobjects together, and another which lays out a group of canvasobjects.

andydotxyz:
Hmm, not sure. In a way using a `Layout` is automatic. it adapts to size and reconfigures appropriately.
The "manual" cannot so does require additional work to get a similar outcome...

stuartmscott:
Both constructors provide something that can hold multiple canvasobjects, so that is the core functionality. The ability to provide a layout for arranging said canvasobjects is an optional extra, which brings me back to `New` and `NewWithLayout`

andydotxyz:
We reversed the default from  `NewContainer` and `NewContainerWithLayout` in the `fyne` package to `NewContainer` and `NewContainerWithoutLayout` because it was causing confusion.
People were using the default, shorter, constructor and then complaining that layouts didn't work and that containers were difficult to use.
I don't want to revert to that state. Whilst the layout is optional it should not be default omitted.

Jacalz:
I have one suggestion that might kind of work. Could we not have just one `container.New()` function that takes and the objects. We can then and let manual layout happen with passing a `nil` layout just like we now pass `nil` to icons if we don't want them in for example a `widget.Card` or `widget.Icon`. I know that passing `nil` isn't the most intuitive way, but it is the best solution that I can fink of and it seems sensible in my opinion.

andydotxyz:
It would probably be an acceptable solution - though it doesn't feel great.
If you're deprecating the old methods we would end up with a deprecation for `NewContainerWIthoutLayout` recommending `container.New(nil)` which feels a bit of a backwards step...

Jacalz:
Reverted back to the old naming. It seemed pretty clear on the call today that `container.NewWithoutLayout()` might be the best we have at the moment.

stuartmscott (approved):

container.go
```diff
@@ -15,14 +15,13 @@ type Container struct {
 }
 
 // NewContainer returns a new Container instance holding the specified CanvasObjects.
-//
```
andydotxyz, line 18: This bank line is required.
Also please add one to the other deprecated notices - it needs to begin a new paragraph.

andydotxyz (changes requested):

container/container.go
```diff
@@ -0,0 +1,16 @@
+// Package container provides container widgets that are used to lay out and organise applications
+package container
+
+import (
+	"fyne.io/fyne"
+)
+
+// New returns a new Container instance holding the specified CanvasObjects which will be laid out according to the specified Layout.
```
stuartmscott, line 8: Needs "Since: 2.0.0" annotations

stuartmscott (changes requested):

Jacalz:
Should be fixed now :)

stuartmscott (approved):

andydotxyz (approved):

PR #1638, opened 2020-12-08
cmd/fyne: Switch to using strings for bundled resources

Jacalz:
### Description:

This moves on with moving to strings over []byte for bundled items. Should decrease memory usage in gopls even more and for applications using bundled resources. Hopefully this can get fynedesk staticcheck to not die from memory usage.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


lucor:
Wondering if an ignore directive for staticcheck could make sense since these are autogenerated files. I don't know if exists something similar for gopls that could help.

andydotxyz:
> Wondering if an ignore directive for staticcheck could make sense since these are autogenerated files

Not rreally - we are generating the files, so should generate them correctly. That is what @Jacalz change does I think.

Jacalz:
It looks like staticcheck doesn’t have support for that yet. You can turn off specific checks for a file but not exclude them fully. https://github.com/dominikh/go-tools/issues/429

lucor:
I see. I was thinking only some types of checks was causing problem, hence the idea to have the generator to add a [linter ignore directive](https://staticcheck.io/docs/#ignoring-problems).

andydotxyz:
> we are generating the files, so should generate them correctly

I think this still holds?

lucor:
> > we are generating the files, so should generate them correctly
> 
> I think this still holds?
Agreed generated files should be generated correcly. Just wondering if an accurate and manual check on the generated files could be good enough for now, since we probably won't change the generator often and the static analisys on the generated files is causing trouble :)

Jacalz:
Given that staticcheck doesn't support disabling for whole files (going through and testing the individual checks seems time consuming) and we already apply `// auto-generated` to the file, I think we should be fine.

andydotxyz:
> static analisys on the generated files is causing trouble :)

I don't think this is the case... this PR changed the code that generates the files, and the output was not correct. 
By adjusting it further the output will be correct again.

andydotxyz (dismissed):

Jacalz:
Sorry @andydotxyz. You will need to re-approve now that I have rebased on latest develop (release 1.4.x merge related).

andydotxyz (approved):

PR #1639, opened 2020-12-09
Fix/formimage

andydotxyz:
Fix glitch in icon appearance. Was due to a fault being triggered in aspect ratio lookup.

### Checklist:

- [x] Tests included. <- using the tests on @Jacalz form validation branch confirmed the fix
- [x] Tests all pass.


internal/painter/svg_cache.go
```diff
@@ -32,7 +32,13 @@ func init() {
 
 // GetAspect looks up an aspect ratio of an image
 func GetAspect(img *canvas.Image) float32 {
-	aspect := aspects[img.Resource]
+	aspect := float32(0.0)
```
stuartmscott, line 35: A named return variable would avoid this line ;)

stuartmscott (approved):
Looks good

internal/painter/svg_cache.go
andydotxyz: True - but something about comparing the value of a not-visibly-initialised variable 8 lines lower made me feel uneasy. Possibly the C developer in me?

Jacalz: Both points are very valid, yet I think a named return value might look cleaner. Do as you please :)

Jacalz (approved):
This does fix form even to the point where I can stop using the constructors. Thanks :+1:

PR #1641, opened 2020-12-09
Add support for overriding standard icons in the new theme interface

andydotxyz:
Makes for more flexibility as we start using specific icons in our more advanced widgets (like FileDialog for example).

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/theme/icons.go
+++ b/theme/icons.go
@@ -9,6 +9,142 @@ import (
 	"fyne.io/fyne"
 )
 
+var (
+	// Icons specifies each of the known icon names that a theme can contain.
+	//
+	// Since 2.0.0
+	Icons = struct {
+		Cancel, Confirm, Delete, Search, SearchReplace, Menu, MenuExpand                fyne.ThemeIconName
+		CheckButtonChecked, CheckButton, RadioButton, RadioButtonChecked                fyne.ThemeIconName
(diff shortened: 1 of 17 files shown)


theme/icons.go
```diff
@@ -9,6 +9,142 @@ import (
 	"fyne.io/fyne"
 )
 
+var (
+	// Icons specifies each of the known icon names that a theme can contain.
+	//
+	// Since 2.0.0
+	Icons = struct {
```
toaster, line 16: What is the purpose of this struct?
If it’s just for providing constants, we should actually _use_ constants instead.

AFAICS, the struct approach allows to
```go
theme.Icons.Cancel = "confirm"
```
and that might not be intended.

andydotxyz: You're right we certainly don't want those remapped.
It was an attempt to avoid the explosion of const names without much namespace, but perhaps it was not the right approach.
As it was following the colour and size naming I will need to go back and revise them as well.

toaster (approved):

andydotxyz:
@stuartmscott I just realised you may want to quickly review this as well as you approved the previous theme API which @toaster has pointed out had a flaw, fixed in this change.

stuartmscott (approved):

PR #1643, opened 2020-12-09
Deprecations for Progress, ProgressInfinite and Entry dialogs

Jacalz:
### Description:

Deprecates a couple dialogs as discussed and makes Entry dialog use Form dialog.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (changes requested):
Let's not have fyne_demo be showing off deprecated components.
Either remove them, or update to recommended replacements.

andydotxyz (changes requested):
Not sure why I didn't notice before, but the Deprecated line needs to have an empty line before it. From the Go docs:

> add a paragraph to its doc comment that begins with Deprecated:

Jacalz:
Updated this to be rebased on the form changes done in develop. Should  be ready for a re-review now.

andydotxyz (approved):

PR #1644, opened 2020-12-10
Use position to pixel coordinate mapping for gl Capture

edaniels:
### Description:

On macOS retina screens, scaling is done at the texture level which can be 2.0. This change makes sure that ReadPixels ultimately captures the right amount of pixels from a canvas. I tested that this continues to work as a window is dragged back and forth between a retina and non-retina screen.

Fixes #(issue)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/internal/driver/glfw/canvas.go
+++ b/internal/driver/glfw/canvas.go
@@ -198,12 +198,6 @@ func (c *glCanvas) Scale() float32 {
 	return c.scale
 }
 
-func (c *glCanvas) TextureScale() float32 {
-	c.RLock()
-	defer c.RUnlock()
-	return c.texScale
-}
-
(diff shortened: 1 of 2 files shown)


andydotxyz (changes requested):
Great catch thanks.
I think it could be done without exposing the texture scale, which is currently an internal detail.

Perhaps
```go
pos := fyne.NewPos(c.Size().Width, c.Size().Height)
width, height := c.PixelCoordinateForPosition(pos)
```

edaniels:
Worked like a charm. Thanks :)

andydotxyz (approved):

andydotxyz:
Great catch, thanks for the revision.
I'll cherry-pick this onto the 1.4.x branch as well for 1.4.3

PR #1652, opened 2020-12-13
Reorder some struct fields for memory savings

Jacalz:
### Description:

This uses [structslop](https://github.com/orijtech/structslop) for specifying how structs should be laid out for best allocation efficiency. The animation struct is optimized for 33.3% memory savings and the fileitems in file dialog for 11.1% savings (the latter will be more noticeable with many files).

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/animation.go
+++ b/animation.go
@@ -22,11 +22,11 @@ var (
 // Animation represents an animated element within a Fyne canvas.
 // These animations may control individual objects or entire scenes.
 type Animation struct {
-	AutoReverse bool
 	Curve       AnimationCurve
 	Duration    time.Duration
-	Repeat      bool
 	Tick        func(float32)
+	AutoReverse bool
+	Repeat      bool
 }
 
 // NewAnimation creates a very basic animation where the callback function will be called for every
--- a/animation.go
+++ b/animation.go
@@ -21,7 +21,7 @@ var (
 
 // Animation represents an animated element within a Fyne canvas.
 // These animations may control individual objects or entire scenes.
-type Animation struct {
+type Animation struct { // NOTE: These structs are not in alphabetical order. They are structured using structslop to reduce padding and memory usage.
 	Curve       AnimationCurve
 	Duration    time.Duration
 	Tick        func(float32)
--- a/dialog/fileitem.go
+++ b/dialog/fileitem.go
@@ -18,12 +18,11 @@ const (
 
 type fileDialogItem struct {
 	widget.BaseWidget
+	name      string
+	location  fyne.URI
 	picker    *fileDialog
 	isCurrent bool
-
-	name     string
-	location fyne.URI
-	dir      bool
+	dir       bool
 }
 
 func (i *fileDialogItem) Tapped(_ *fyne.PointEvent) {
--- a/dialog/fileitem.go
+++ b/dialog/fileitem.go
@@ -16,6 +16,7 @@ const (
 	fileIconCellWidth = fileIconSize * 1.25
 )
 
+// NOTE: These structs are not in alphabetical order. They are structured using structslop to reduce padding and memory usage.
 type fileDialogItem struct {
 	widget.BaseWidget
 	name      string


Jacalz:
There were a few more recommendations, but I avoided them as I figured that it wasn't worth it for readability. If you think that any of these would be worth applying, please say so.
```
./fyne/internal/driver/glfw/canvas.go:22:15: struct has size 256 (size class 256), could be 240 (size class 240), you'll save 6.25% if you rearrange it to:
struct {
	shortcut        fyne.ShortcutHandler
	size            fyne.Size
	content         fyne.CanvasObject
	menu            fyne.CanvasObject
	context         driver.WithContext
	painter         gl.Painter
	onTypedRune     func(rune)
	menuTree        *renderCacheTree
	overlays        *overlayStack
	onTypedKey      func(*fyne.KeyEvent)
	onKeyDown       func(*fyne.KeyEvent)
	menuFocusMgr    *app.FocusManager
	contentFocusMgr *app.FocusManager
	contentTree     *renderCacheTree
	refreshQueue    chan fyne.CanvasObject
	dirtyMutex      *sync.Mutex
	onKeyUp         func(*fyne.KeyEvent)
	sync.RWMutex
	texScale      float32
	detectedScale float32
	scale         float32
	dirty         bool
	padded        bool
}

./fyne/test/testcanvas.go:30:17: struct has size 168 (size class 176), could be 160 (size class 160), you'll save 9.09% if you rearrange it to:
struct {
	fyne.ShortcutHandler
	size         fyne.Size
	content      fyne.CanvasObject
	painter      SoftwarePainter
	hovered      desktop.Hoverable
	onTypedRune  func(rune)
	onTypedKey   func(*fyne.KeyEvent)
	overlays     *internal.OverlayStack
	focusMgr     *app.FocusManager
	propertyLock sync.RWMutex
	scale        float32
	padded       bool
}
./fyne/internal/driver/gomobile/canvas.go:20:19: struct has size 264 (size class 288), could be 248 (size class 256), you'll save 11.11% if you rearrange it to:
struct {
	shortcut        fyne.ShortcutHandler
	content         fyne.CanvasObject
	windowHead      fyne.CanvasObject
	menu            fyne.CanvasObject
	dragging        fyne.Draggable
	painter         gl.Painter
	size            fyne.Size
	focused         fyne.Focusable
	touchLastTapped fyne.CanvasObject
	onTypedKey      func(event *fyne.KeyEvent)
	onTypedRune     func(rune)
	touched         map[int]mobile.Touchable
	touchCancelFunc context.CancelFunc
	touchTapCount   int
	lastTapDown     map[int]time.Time
	lastTapDownPos  map[int]fyne.Position
	overlays        *internal.OverlayStack
	refreshQueue    chan fyne.CanvasObject
	minSizeCache    map[fyne.CanvasObject]fyne.Size
	scale           float32
	inited          bool
	padded          bool
}
```

animation.go
```diff
@@ -22,11 +22,11 @@ var (
 // Animation represents an animated element within a Fyne canvas.
 // These animations may control individual objects or entire scenes.
 type Animation struct {
-	AutoReverse bool
```
andydotxyz, line 25: This is a change to a public API and will break any instances of &Animation{} without field indexes used.
Do we want to do that? I know the API is not final yet - but still worth noting the break.

andydotxyz:
Bear in mind that reordering exported fields in a public struct is a breaking change as anyone using strict initialisation without field indexes will stop compiling.

animation.go
Jacalz: Well, technically adding a new field as a feature down the line will also break it (as you need to give a value to each field that way). I don't personally think it is a huge issue with this breakage, but I will let you decide :)

andydotxyz: Quite right, I hadn't thought of that.
I still think we would need a way to mark this as "not alphabetical for x reason" otherwise later refactors and/or reviews could reorder or reject.

PR #1654, opened 2020-12-14
Proper dot escape in vedoring regexp and performance improvement

Jacalz:
### Description:

This fixes a potential security vulnerability where the vendor command could accept a maliciously crafted path.
It also simplifies and improves the performance by only compiling regexp once not for each iteration in the for-loop and uses MatchString() instead of converting to []byte for each match.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### To reviewer:
- Please test to make sure that the `fyne vendor` command works on your machine.

### DIFF
--- a/cmd/fyne/vendor.go
+++ b/cmd/fyne/vendor.go
@@ -22,12 +22,15 @@ const (
 	goModFile = "go.mod"
 	// goModVendorFile is the vendor module file
 	goModVendorFile = "vendor/modules.txt"
-	// glfwModNew is the glfw module regexp for 3.3+
-	glfwModNew = `github\.com/go-gl/glfw/(v(\d+\.)?(\*|\d)?(\*|\d+))/glfw`
-	// glfwModOld is the glfw module regexp for 3.2
-	glfwModOld = "github.com/go-gl/glfw"
 	// glfwModSrcDir is the glfw dir containing the c source code to copy
 	glfwModSrcDir = "glfw"
+)
+
+var (
+	// glfwModNew is the glfw module regexp for 3.3+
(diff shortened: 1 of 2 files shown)


cmd/fyne/vendor.go
```diff
@@ -22,14 +22,17 @@ const (
...
-	glfwModOld = "github.com/go-gl/glfw"
 	// glfwModSrcDir is the glfw dir containing the c source code to copy
 	glfwModSrcDir = "glfw"
 )
 
+var (
+	// glfwModNew is the glfw module regexp for 3.3+
+	glfwModNew = regexp.MustCompile(`github\.com/go-gl/glfw/(v(\d+\.)?(\*|\d)?(\*|\d+))/glfw`)
+	// glfwModOld is the glfw module regexp for 3.2
+	glfwModOld = regexp.MustCompile(`github\.com/go-gl/glfw`)
```
stuartmscott, line 33: Does this need to be a regexp? Seems like a simple string comparison would work as there is no number matching

Jacalz: It isn't a simple string comparison. It is checking the anywhere in the input because of not having start and end characters. I am just fixing up the obvious things here 🙂

andydotxyz:
I don't think that we need `fyne vendor` any more, as glfw was fixed upstream (IIRC).
If this is the case perhaps we should remove this code, just outputting instead a warning about it being deprecated and then launch `go mod tidy && go mod vendor` directly?

cmd/fyne/vendor.go
stuartmscott: Ah okay, well `strings.Contains` may still be more efficient, tho I appreciate that you're trying to minimize the size of this code change

PR #1656, opened 2020-12-15
Remove BackgroundColor

### LINKED ISSUE #1338: Deprecate WidgetRenderer.BackgroundColor

stuartmscott:
### Description:
Removes fyne.WidgetRenderer.BackgroundColor and replaces instances where a background is needed with a suitable canvas.Rectangle.

Fixes #1338

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.


widget/fileicon.go
```diff
@@ -73,7 +72,12 @@ func (i *FileIcon) CreateRenderer() fyne.WidgetRenderer {
 
 	i.setURI(i.URI)
 
-	s := &fileIconRenderer{file: i}
+	// TODO should FileIcon render a background representing selection, or should it be in the collection?
+	// TODO file dialog currently uses a container with NewGridWrapLayout, but if this changes to List, or Table then the primary color background would be rendered twice.
+	bg := canvas.NewRectangle(theme.PrimaryColor())
```
Jacalz, line 77: Hmm, I am not so sure if I think the base fileicon should have a background at all. We probably need a better way to visually indicate the selection, I guess.

stuartmscott: Agreed, I added the background so the behaviour is consistent and it made more sense to add it here as FileIcon has SetSelected than to add it to the grid. Do you think `SetSelected` and the general selection code should be deprecated and removed from FileIcon? A lot depends on FileDialog. If we're agreed that the right approach is to use List & Table (would be nice to have a toggle switch so users can pick how they want the files laid out) to display the files/directories and the collections will have selection indication then the background can be removed from FileIcon. However we may still want the functionality to invert the colours, but if we do it for one widget we may have to do it for all.

Jacalz: That sounds like a splendid idea. Switching to table and list would be a big win performance wise and then let them handle the selection indication. Yeah I can see colour inversion might need to be done for that.

toaster: Note: I dont’t think this is limited to list or table but every container that allows selection or other indications should handle this by itself instead of delegating it to the widget.

toaster:
LGTM
The move of selection indication into containers should probably be a separate PR.

widget/fileicon.go
stuartmscott: Agreed, tho it should be noted that changing the background to primary colour to indicate selection can sometimes make child canvasobject less readable, hence the colour invertion.

Perhaps the preferred approach should be the primary colour shadow "halo" effect? The downside with this is that the halo is rendered outside the bounds of the canvasobject.

stuartmscott:
In the case of showing hover, I'm using a canvas.Rectangle. When the item is not hovered, is it cheaper to use color.Transparent, or Hide/Show?

stuartmscott:
@toaster the remaining failures revolve around child menus, whereby the hover background is guarded by `if !fyne.CurrentDevice().IsMobile() && (r.i.hovered || (r.i.child != nil && r.i.child.Visible())) {` however it seems this is not sufficient, and the menuitem with the child continues to be displayed as hovered. Can you advise a solution?

andydotxyz:
> is it cheaper to use color.Transparent, or Hide/Show? I assumed Hide/Show

Yes, Hide/show is cheaper, assuming that your MinSize does not change when the child item is hidden.

stuartmscott (commented):
There is still an issue with menus.

Additionally;
 1. internal/driver/glfw/testdata/menu_bar_active_file.png is wrong
 2. internal/painter/software/testdata/draw_text_clipped.png has text that is hard to see
 3. test/testcanvas.go needs a sanity check
 4. widget/fileicon.go needs consensus on the right approach - if collections handle selection then FileIcon.SetSelected should be deprecated, but how will FileIcon (or any other widget) invert their resources?
 5. widget/testdata/form/theme_changed.png needs a sanity check

andydotxyz:
> There is still an issue with menus.

For the record I have a fix for this but will apply it over the top of this due to the size of the merges resolved here.

> 4\. widget/fileicon.go needs consensus on the right approach - if collections handle selection then FileIcon.SetSelected should be deprecated, but how will FileIcon (or any other widget) invert their resources?

We still have not picked a selection color (we just use Primary) If we pick one that is less vibrant then we can avoid inverting colours. This seems desirable if we will be using background in many places.

Jacalz (commented):
Nice work. This seems to reduce memory usage a bit and does feel slightly more snappy. The only thing I notice is that we need a better solution for selection in the file manager. It looks a little strange with the selection only covering the icon, and not the text. Can we put the rectangle in the fileItem code instead?

andydotxyz:
On my system the file selection looks even further off...
Screenshot 2021-01-14 at 21 06 46

andydotxyz:
If `dialog/fileitem.go` line 96 were changed to:
```go
	s.text.Move(fyne.NewPos(0, size.Height - fileTextSize-theme.Padding()*2))
```
Then it would be OK to use for now.

stuartmscott:
Yeah I think the text is being rendered outside of bounds. Do you want that change in this PR?

stuartmscott:
Yup, looks better now;
Screen Shot 2021-01-14 at 13 21 09

andydotxyz (approved):

PR #1657, opened 2020-12-15
Capture on draw thread to avoid gl race

edaniels:
### Description:
Right now the glfw canvas Capture will run on the main thread which can contend with the draw thread in its draw loop which can cause spurious sig aborts. Moving Capture over to the draw thread as well resolves this issue.

### Checklist:

- [x ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
This is a great catch thanks

andydotxyz:
Cherry picked onto the release branch as well

PR #1662, opened 2020-12-16
Remove deprecated popup and dialog components

### LINKED ISSUE #852: Dialogs are shown even from New...Dialog calls without .Show()
### Describe the bug:
Dialogs are all shown immediately despite the fact that New...Dialog is not intended to show.

### To Reproduce:
1. Call `dialog.NewConfirm(...)`
2. See that it is visible immediately

### Example code:
```go
dialog.NewConfirm("Confirmation", "Are you enjoying this demo?", confirmCallback, win)
```

 - OS: Mac, Linux
 - Version: latest
 - Go version: 1.12, 1.13
 - Fyne version: 1.2.x, develop

This is due to the underlying `NewModalPopUp` showing immediately.
Therefore we should not create the popup until Show() is called. This makes sense anyhow.
There is too much display code in the setup methods.
The attached patch would fix the issue but is ugly, I think that the whole `dialog` package needs work so that it's more semantic based.

andydotxyz:
Fix it so that dialogs now don't show until Show or ShowXxx is called
Fixes #852

### Checklist:

- [x] Tests included. <- existing tests, this is removing old code
- [x] Tests all pass.


Jacalz (commented):
Looks to be mostly good to me. Was it intended to add the `cmd/fyne_demo/Icon.png` file?
I guess the changes here should probably also be added to the changelog (including the removal of dialog.FileIcon)?

andydotxyz:
Thanks for those hints @Jacalz - the icon was not supposed to be in.
Resolved and CHANGELOG updated.

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ These changes likely break some apps, please read the
 * A new, more flexible, Theme interface has been created that we encourage developers to use
 * The desktop.Cursor definition was renamed desktop.StandardCursor to make way for custom cursors
 
+* Dialogs no longer show when created, unless using the ShowXxx convenience methods
+* Removed deprecated types including dialog.FileIcon
```
Jacalz, line 19: You probably should mention that they should use widget.FileIcon instead?

Jacalz (changes requested):
Sorry, just one little comment on the changelog.

CHANGELOG.md
andydotxyz: Fixed

Jacalz (approved):

toaster (approved):
LGTM

Issue #1509, opened 2020-11-05
Delete/Rename URIs


squash:
### Is your feature request related to a problem? Please describe:
Currently there is no straightforward way to remove or rename a file by URI.

### Is it possible to construct a solution with the existing API?
I am implementing a hacky solution that breaks the URI apart into a normal file path, does the required operation, and then (for renames) parses a new URI. 

### Describe the solution you'd like to see:
It would be nice if this functionality were included as part of the URI type

andydotxyz:
Indeed. Let's get these in. We could not add them easily in 1.4 as it would be a breaking change in the URI type - so we will pop it in 2.0

charlesdaniels:
I have rolled this into #1665. I think adding delete functionality via `URI.Destroy()` is a good idea, but I'm not sold on `Rename()`, as this merely composes several existing operations (create, write, delete). On the other hand, exposing this to the underlying implementation as a `Rename()` operation may allow for optimizations (e.g. doing the rename on the remote end, rather than pulling the data down and writing it back over the wire for a remote fs).

Thoughts?

andydotxyz:
All good points I guess.
Worth noting that if we don't add Rename now it cannot appear until 3.0 as it will break the URI API.
Doing that inside the app is indeed possible - but could produce bad intermediate states (like create, write, delete where delete fails leaves duplicates...).

Not sure that I prefer Destroy() over Delete() though (we use it internally to free memory which I think is different.
Usually the choice is between Delete and Remove.

andydotxyz:
Landed, great job thanks @charlesdaniels and the team.

PR #1665, opened 2020-12-17
WiP: URI Rework

charlesdaniels:
### Description:

This PR tracks work being done to re-work URIs based on prior discussion (#1233) for the 1.4 release, and also addresses #1509. In particular, it moves several functions into the URI interface that were previously stand-alone functions.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


charlesdaniels:
Would like some input from @andydotxyz / @stuartmscott relating to the proposed changes before I do the legwork to implement. See commends in the diff.

uri.go
```diff
@@ -27,10 +27,116 @@ type URIWriteCloser interface {
 // This resource may be a file or another data source such as an app or file sharing system.
 type URI interface {
 	fmt.Stringer
+
+	// Extension should return the file extension of the resource
+	// referenced by the URI. For example, the Extension() of
+	// 'file://foo/bar.baz' is 'baz'. May return an empty string if no the
```
stuartmscott, line 33:
```suggestion
	// 'file://foo/bar.baz' is 'baz'. May return an empty string if the
```

uri.go
```diff
@@ -27,10 +27,116 @@ type URIWriteCloser interface {
...
+	// referenced resource has none.
 	Extension() string
+
+	// Name should return the base name of the item referenced by the URI.
+	// For example, the Name() of 'file://foo/bar.baz' is 'bar.baz'.
 	Name() string
+
+	// MimeType should return the content type of the resource referenced
+	// by the URI. The returned string should be in the format described
+	// by section 5 of RFC2045 ("Content-Type Header Field").
```
stuartmscott, line 43:
```suggestion
	// by Section 5 of RFC2045 ("Content-Type Header Field").
```

uri.go
```diff
@@ -27,10 +27,116 @@ type URIWriteCloser interface {
...
+	//   the parent of a file:// URI requires reading information from
+	//   the filesystem, it could fail with a permission error.
+	//
+	// NOTE: To the extent possible, Parent() should not modify, create,
+	// or interact with referenced resources. For example, it should
+	// usually be possible to use Parent() to create a reference to
+	// a resource which does not exist, though future operations on this
+	// resource may fail.
+	//
+	// Since: 2.0
```
stuartmscott, line 79:
```suggestion
	// Since: 2.0.0
```

uri.go
```diff
@@ -27,10 +27,116 @@ type URIWriteCloser interface {
...
+	// MimeType should return the content type of the resource referenced
+	// by the URI. The returned string should be in the format described
+	// by section 5 of RFC2045 ("Content-Type Header Field").
 	MimeType() string
+
+	// Scheme should return the URI scheme of the URI. For example,
+	// the Scheme() of 'file://foo/bar.baz` is 'file'.
 	Scheme() string
+
+	// Parent should return a URI referencing the parent resource of the
```
stuartmscott, line 50: Also note that the returned URI should be a ListableURI

storage/uri.go
```diff
@@ -123,6 +123,8 @@ func parentGeneric(location string) (string, error) {
 // removing the last item.
 //
 // Since: 1.4
+//
+// Deprecated: Parent() is a member of the URI interface now.
```
stuartmscott, line 127: Consistency :)
```suggestion
// Deprecated: Parent() is now a member of the URI interface.
```

charlesdaniels:
For the record, I agree with Stuart's changes, but am waiting to put more commits into this until everybody is agreed on [the relevant proposal](https://github.com/fyne-io/fyne/wiki/Proposal:-The-Fyne-URI-Philosophy).

charlesdaniels:
I have sketched out more or less what my current thinking is on the API in a87a2058fdca184543e74f850df3395daf3699d1 and also wrote a commit message with some more thoughts.

If people have objections, now is the time to raise them. We are getting down to the wire on 2.X. I have done my very best to balance what I believe will be a good API design, what will be backwards compatible with the existing 1.4.X API, and what we have discussed in previous meetings and in the proposal.

One thing I would like explicit comments on is whether folks feel we should move `Repository` into it's own package, or keep it in storage. I am on the fence, but leaning towards moving it into `storage/repository`, to reduce the clutter in the storage package - most developers will not need to interact with repositories directly, so hiding it makes sense.

charlesdaniels:
Specifically paging @andydotxyz, @toaster, @stuartmscott, and @AlbinoGeek w.r.t the above comment.

AlbinoGeek (commented):
While I'm currently having a hard time (mentally) putting together an implementation for the Repository, after reviewing the code I feel it should be straight-forward to use. However, an example would go a long way in helping that.

I may have missed this somewhere, should `file://` scheme be implemented as a Repository, to make sure it conforms with the interface, etc? (this would also be a great way to tell if any major / important features were missing from the API.)

That would serve as a great example. Does it belong in another PR or this one?

---

Overall I like the direction and can't wait to see what others think too.

storage/repository.go
```diff
@@ -0,0 +1,152 @@
...
+	//
+	// Since 2.0.0
+	ParseURI() func(string) (fyne.URI, error)
+
+	// Init will be called while the repository is being registered.  This
+	// is the appropriate place to register any optional handlers relevant
+	// to a particular URI scheme.
+	//
+	// Since 2.0.0
+	Init() func() error
```
andydotxyz, line 71: I'm not sure about including `Init` to be run during register - if we are supporting a single `Repository` backing multiple schemes then this could be called multiple times, which the name does not imply.
What is this needed for that cannot be expected of the implementation before calling `Register`?

storage/repository.go
```diff
@@ -0,0 +1,152 @@
...
+func RegisterRename(scheme string, renameImplementation func(fyne.URI, fyne.URI) error) error {
+}
+
+// RegisterRepository registers a storage repository so that operations on URIs
+// of the registered scheme will use methods implemented by the relevant
+// repository implementation. This method will call repository.Init(), and may
+// error if it does.
+//
+// Since 2.0.0
+func RegisterRepository(scheme string, repository Repository) error {
```
andydotxyz, line 151: Given that `Repository` is the "main" interface, I would suggest `Register(string, Repository)` so it is more clearly differentiated from the optional registrations above.

storage/uri.go
```diff
@@ -202,3 +266,184 @@ func Exists(u fyne.URI) (bool, error) {
...
+//
+// * Checking for listability depended on a lower level operation
+//   such as network or filesystem access that has failed in some way.
+//
+// Listable is backed by the repository system - this function either calls
+// into a scheme-specific implementation from a registered repository, or fails
+// with a URIOperationNotSupported error.
+//
+// Since 2.0.0
+func Listable(u fyne.URI) (bool, error) {
```
andydotxyz, line 422: `Listable` feels almost an assertion, would `CanList` be OK? It feels less ambiguous.

storage/uri_operation_impossible.go
```diff
@@ -0,0 +1,16 @@
+package storage
+
+// Declare conformance with Error interface
+var _ error = URIOperationImpossible
```
andydotxyz, line 4: I'm sure it would be OK to have all declared errors in a single error.go file :)

uri.go
```diff
@@ -23,19 +23,70 @@ type URIWriteCloser interface {
...
+	// RFC3986.
+	//
+	// NOTE: the RFC3986 can be obtained by combining the User and Host
+	// Fields of net/url's URL structure. Consult IETF RFC3986, section
+	// 3.2, pp. 17.
+	//
+	// Since: 2.0.0
+	Authority() string
+
+	// Path should return the URI path, as defined by IETF RFC3986.
```
andydotxyz, line 67: We need to add to this doc to make it clear that one must not assume a file path.
As we use URI to replace file handling I have heard many assume that "IF you would just implement `Path()` then I could send it to my file managent code...

andydotxyz (changes requested):
This looks really solid.
A few changes requested and thoughts inline but I think this fulfils the requirements.

Some other thoughts back on the URI philosophy document so it's consistent.

storage/repository.go
stuartmscott: I am also not sure this needs to be included. I envisioned the initialization going the other way, where the developer initializes a repository, most likely by importing a third party library, which runs its `func init()`, which registers itself with the storage API (like when importing "image/png")

storage/repository.go
```diff
@@ -0,0 +1,152 @@
...
+//
+// If no Copy implementation is registered for a particular scheme, then copies
+// will be implemented as if the URIs were of different types, that is the
+// source URI will be read using ReaderFrom(), and the destination written
+// using WriterTo(). This function is an opportunity to implement a more
+// optimal approach, such as leveraging remote operations for a network-backed
+// repository.
+//
+// Since 2.0.0
+func RegisterCopy(scheme string, copyImplementation func(fyne.URI, fyne.URI) error) error {
```
stuartmscott, line 124: nit:
```suggestion
func RegisterCopy(scheme string, copy func(fyne.URI, fyne.URI) error) error {
```

storage/uri.go
stuartmscott: Or `IsListable`

storage/uri_operation_impossible.go
stuartmscott: I agree, I also don't think the separate types is very go idiomatic. https://golang.org/src/io/io.go has some exported errors that I have used before;
```go
var EOF = errors.New("EOF")
```

storage/repository.go
stuartmscott, line 151: Will this will also silently replace any previously registered Repository? Will the previously registered Repository get any notification that it is no longer in use?

storage/uri_operation_impossible.go
andydotxyz: Discussing idomatic errors in Go is tricky as they are mid-transition ;) We need to go with what was best in 1.12 I think as some of the newer things we can't opt in to without upgrading our version required.
That said I don't remember what best practice was 3 versions ago.

storage/uri.go
andydotxyz: That crossed my mind to, though generally we prefer to drop "Get" and "Is" prefixes.

charlesdaniels:
@andydotxyz @stuartmscott @AlbinoGeek 

Per our meeting today, I have written a revised proposal ([link](https://github.com/fyne-io/fyne/wiki/Proposal:-The-Fyne-URI-Philosophy#revised-proposal)). I think this represents our current consensus based on what we discussed. 

1. If anyone has final comments on the URI proposal, Monday the 11th is the deadline to speak or forever hold your peace. Please use the new discussion section I added for the updated proposal. I will keep editing it in-place to track what we want the final API to be.
2. We went over the comments on this PR that are currently open in our meeting, and I think the revised proposal addresses them. If so, please mark the conversations as resolved.

I plan to resume work on this on Monday (or maybe a little on Sunday).

storage/repository.go
stuartmscott: From https://github.com/fyne-io/fyne/wiki/Proposal:-The-Fyne-URI-Philosophy#revised-proposal; yes and yes via `repository.Destoy()`

PR #1667, opened 2020-12-18
call UpdateNode during Refresh

### LINKED ISSUE #1663: Tree failing to call UpdateNode on manual refresh
Hi,
is it possible to change a node label text in a tree widget?

thank you)

andydotxyz:
Resolve issue where tree did not refresh from data using UpdateNode

Fixes #1663

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (approved):
Looks good to me, but it is probably best if @stuartmscott have a look as well :)

widget/tree.go
```diff
@@ -694,7 +695,9 @@ func (r *treeNodeRenderer) Objects() (objects []fyne.CanvasObject) {
 
 func (r *treeNodeRenderer) Refresh() {
 	if c := r.treeNode.content; c != nil {
-		c.Refresh()
+		if f := r.treeNode.tree.UpdateNode; f != nil {
+			f(r.treeNode.uid, r.treeNode.isLeaf, c)
```
stuartmscott, line 699: The second parameter of `UpdateNode` is `isBranch`

widget/tree.go
```diff
@@ -782,6 +786,7 @@ func newLeaf(tree *Tree, content fyne.CanvasObject) (l *leaf) {
 		&treeNode{
 			tree:    tree,
 			content: content,
+			isLeaf:  false,
```
stuartmscott, line 789: Copy/Pasta error - should be true

widget/tree.go
```diff
@@ -586,6 +586,7 @@ type treeNode struct {
 	depth   int
 	hovered bool
 	icon    fyne.CanvasObject
+	isLeaf  bool
```
stuartmscott, line 589: Let's keep consistent with the rest of Tree and use `isBranch`

stuartmscott (changes requested):

widget/tree.go
andydotxyz: good catch thanks, will fix the logic

widget/tree.go
andydotxyz: yeah, as above, good catch

widget/button_test.go
```diff
@@ -65,6 +65,10 @@ func TestButton_Tapped(t *testing.T) {
 }
 
 func TestButton_Disable(t *testing.T) {
+	app := test.NewApp()
+	defer test.NewApp()
+	app.Settings().SetTheme(theme.LightTheme())
```
stuartmscott, line 70: This is not related to fixing #1663, please remove

andydotxyz: Are you sure? It seemed to start failing when I added the tree test, hence the commit message "Resolve flakey button test impacted by this tree change".

andydotxyz: Seems like this was broken on release branch and we didn't notice.
Rebase on top of the fix, so it's no longer part of this PR diff.

stuartmscott (approved):

PR #1670, opened 2020-12-18
#1668 tab container panic when removing tabs

### LINKED ISSUE #1668: TabContainer Panic when removing tabs and selected tab is becomes of range
### Describe the bug:
When removing tabs from a TabContainer this panic is thrown
`fyne.io/fyne/widget.(*TabContainer).RemoveIndex(...)
	C:/Users/Adrian/go/pkg/mod/fyne.io/fyne@v1.4.2/widget/tabcontainer.go:125
main.main.func3()
	C:/Users/Adrian/go/src/Test/main.go:16 +0x18f
fyne.io/fyne/widget.(*Button).Tapped(0xc0000ca160, 0xc00008c0e0)
	C:/Users/Adrian/go/pkg/mod/fyne.io/fyne@v1.4.2/widget/button.go:196 +0xaa
fyne.io/fyne/internal/driver/glfw.(*window).mouseClicked.func5()
	C:/Users/Adrian/go/pkg/mod/fyne.io/fyne@v1.4.2/internal/driver/glfw/window.go:738 +0x3f
fyne.io/fyne/internal/driver/glfw.(*window).runEventQueue(0xc0000a8000)
	C:/Users/Adrian/go/pkg/mod/fyne.io/fyne@v1.4.2/internal/driver/glfw/window.go:1156 +0x83
created by fyne.io/fyne/internal/driver/glfw.(*gLDriver).createWindow.func1
	C:/Users/Adrian/go/pkg/mod/fyne.io/fyne@v1.4.2/internal/driver/glfw/window.go:1180 +0x10d`

It happens when the selected tab becomes out of range for the []items i.e. TabContainer.current > len(TabContainer.Items) - 1
I.E tabs A B C, select C, then remove C results in panic
I have also noticed that if the selected tab is removed and it is not out of range then the selected tab remains at the same value. Not sure if this is desirable behaviour. I.E tabs A B C, select B, then remove B. tabs become A B, and B is still selected.

### Example code:
With this example follow the above instructions
`package main

import (
	"fyne.io/fyne"
	"fyne.io/fyne/app"
	"fyne.io/fyne/widget"
)

func main() {
	myApp := app.New()
	myWindow := myApp.NewWindow("TabContainer Widget")
	tabs := widget.NewTabContainer()
	tabs.Append(widget.NewTabItem("Tab A", widget.NewButton("Remove A", func() { tabs.RemoveIndex(0) })))
	tabs.Append(widget.NewTabItem("Tab B", widget.NewButton("Remove B", func() { tabs.RemoveIndex(1) })))
	tabs.Append(widget.NewTabItem("Tab C", widget.NewButton("Remove C", func() { tabs.RemoveIndex(2) })))
	tabs.SetTabLocation(widget.TabLocationTop)
	myWindow.SetContent(tabs)
	myWindow.Resize(fyne.NewSize(600, 600))
	myWindow.ShowAndRun()
}
`

 - OS: Windows
 - Version: 10
 - Go version: go1.15.5 windows/amd64
 - Fyne version: 1.4.2

adrianre12:
### Description:
As requested by Andrew Williams this is based on the v1.4.2 branch
To fix the panic after tab deletion when the selected tab is out of range of the TabContainer.items slice. I have changed the code so that TabContaner.current value is updated. 
If the selected tab index is higher than that of the deleted tab, the current value is decreased by one. E.G. tabs A B C with C selected, current=2. Delete B, tabs become A C and current=1 so that it still points to C.
If the selected tab index is equal to that of the deleted tab. The value of current is set to -1 which represents no selection
E.G. tabs A B C with B selected current=1. Delete B and current=-1.

There was also an issue where the underline was remaining if the last tab was deleted. To fix this TabContainer.movSelection now hides the underline when current=-1

Finally an unrelated issue, .gitignore is not ignoring windows executables. I saw that it has been setup to ignore linux executables so have added the .exe files too.

As far as I can tell all the TabContainer tests pass, but I can't get TestEntry_Select to pass but it looks like that fails for me on the v1.4.2 branch too.

Fixes #1668


Jacalz (changes requested):
I think there was a misunderstanding. The PR should be based on and have the `release/1.4.x` branch as the target to merge into. Everything will be added to master when the release happens.

adrianre12:
It is my first time using github PR. I am not sure if I understand.
I forked fyne-io and checked out release/1.4.x  branch, did my changes and created the PR 
The reason I mentioned v1.4.2 is in the PR template is says to use branch develop. Did I do it wrong?

Jacalz:
Ah, then I really must congratulate you. Congratulations for both creating your first PR and your first contribution to Fyne 🎉

The target branch is where you want your PR to end up when merged. 9/10 times this will be develop, but when doing small bug fixes that are intended for a bug fix release, it should target the latest release branch (`release/1.4.x` in this case). You have based the PR on that branch, so you are off to a really good start. Now we just need to make it target that branch too 🙂

You can change the target branch of the PR by pressing edit where you usually edit the title and then change the little select button to say that you want to merge it into the name of the branch (`release/1.4.x` in this case). I did this for you now, but it looks like GitHub didn’t change much. Strange.

andydotxyz:
I think all the code / git work was correct, when creating a PR it defaults to master - if you'd picked `release/v1.4.x` when opening the PR it would have been flawless ;)

andydotxyz (changes requested):
Thanks so much for your contribution.

Can you please add a test indicating what breakage has been fixed?

widget/tabcontainer.go
```diff
@@ -122,12 +122,21 @@ func (c *TabContainer) Remove(item *TabItem) {
 
 // RemoveIndex removes tab by index
 func (c *TabContainer) RemoveIndex(index int) {
+	if c.current == index {
+		c.setTabIndex(-1)
+	}
+	if c.current > index {
+		c.setTabIndex(c.current - 1)
+	}
```
stuartmscott, line 130: If either of these branches are taken, the TabContainer will be refreshed, and then when the items are set below the TabContainer will be refreshed again. It might be better to assign the field directly;
```suggestion
	if c.current == index {
		c.current = -1
	}
	if c.current > index {
		c.current = c.current - 1
	}
```

widget/tabcontainer.go
```diff
@@ -122,12 +122,21 @@ func (c *TabContainer) Remove(item *TabItem) {
...
 	c.SetItems(append(c.Items[:index], c.Items[index+1:]...))
 }
 
 // SetItems sets the container’s items and refreshes.
 func (c *TabContainer) SetItems(items []*TabItem) {
 	c.Items = items
+	if c.current >= len(c.Items) {
+		c.setTabIndex(-1)
```
stuartmscott, line 138: ditto

.gitignore
```diff
@@ -3,6 +3,10 @@ cmd/fyne/fyne
 cmd/fyne_demo/fyne_demo
 cmd/fyne_settings/fyne_settings
 cmd/hello/hello
+cmd/fyne/fyne.exe
+cmd/fyne_demo/fyne_demo.exe
+cmd/fyne_settings/fyne_settings.exe
+cmd/hello/hello.exe
```
stuartmscott, line 9: This change is unrelated

stuartmscott (changes requested):
Nice work @adrianre12, thanks for the fix! I added a few comments, but otherwise looks good

widget/tabcontainer.go
adrianre12: I was considering doing that but in the original SelectTabIndex after the Refresh it does
`	if c.OnChanged != nil {
		c.OnChanged(c.Items[c.current])
	}`
I am unsure what that does, so thought it safer to include it at the cost of an extra refresh.  I could have just cut/pasted it everywhere but did not think that was the correct way. So I split SelectTabIndex to create setTabIndex and called that to only have the functionality in one place. 
I am more than happy to make your suggested change if you think the lines above are not needed when current=-1

.gitignore
adrianre12: Do you want a separate issue raised for the .gitignore?

andydotxyz: Looks to me like it was cleanly added in a separate commit, I am sure that we can merge the PR with this in it.
The overhead for a PR process is sometimes just not worth it for single commits like this metadata.
As far as I am concerned as long as each commit has clear purpose and they are not distinct discussion points for unrelated work then it's not a problem.

stuartmscott: ack

widget/tabcontainer.go
stuartmscott: OnChanged is a field to which the developer can assign a function if they want to respond when the current tab changes.

You bring up a good point. @andydotxyz should devs expect a change notification after calling RemoveIndex, or SetItems? Would they expect `OnChange(nil)` to be called when there is no longer a selection ie. `current == -1`? Perhaps it would be better if TabContainer used the same OnSelected, OnUnselected pattern as the collections.

One downside I can see in RemoveIndex is that the change notification will be triggered _before_ the index is removed, so OnChange will be called with the wrong TabItem.

Another issues is that setTabIndex doesn't have the bounds checking so it could crash when indexing the Items slice with -1

andydotxyz: I would expect the OnChange to fire when the selected item changes. If the item removed after current tab then not firing, if the removal was before then it's tough to say, perhaps not, as semantically it's still the same content?

If we set the index to -1 (and it wasn't before) then we do need to fire yes.
Moving to OnSelected could make sense, I think it's on (or should be on) the consistent-selection-api ticket #1395.

adrianre12:
I have been looking at the tests and saw I had missed that TabContainer_DynamicTabs is now failing. I now realise I have made changes to the TagContainer by altering its behaviour when a tab is selected after tab removal. To me with my old QA manager hat on non backwardly compatible changes in a point release would be unacceptable, but it all depends on what your views are.
I am quite willing to undo the breaking changes and limit this to just fixing the panic, This should allow the TabContainer_DynamicTabs test to pass. Then raise another PR based on whatever branch you wish with the other changes if you still feel they are applicable.  In summary, they are;

- If selected_index = removed_Index deselect tab (current=-1) . Currently, the selected_index does not change which means the selected tab effectively shifts to the next tab. If this now points to an out of range Item we get a Panic. The fix for the panic should be if selected_index >= len(items) deselect tab.
- If selected_index > removed_Index decrease selected_index by one so that it remains pointing to the same tab. Currently, the selected_index does not change which means the selected tab effectively shifts to the next tab.

stuartmscott:
You're absolutely right, we try to avoid non-backwardly compatible changes in point releases. Minimizing this PR to just the panic and moving the other behaviour to another PR (on develop branch) feels like the right approach.

I think it would suffice for the change to be;
```go
if c.current >= removed {
    c.current--
}
```
This way closing a tab after the current tab has no effect, closing a tab after the current tab shift the index down so the same tab is still selected, and closing the current tab selects the previous tab, if any, else selects -1

adrianre12:
Scott, that would still be a breaking change. Currently, if we have tabs A B with A selected, remove A and B is selected. With that change, no tab would be selected. I am pretty sure this would make TabContainer_DynamicTabs fail.
I tried this yesterday and it worked.
`if current >= len(Items) {
   current = -1
}`
This does not change behaviour as at the moment it would panic.

adrianre12:
I have removed all the feature changes so only the panic fix remains and added a test for the panic, should be good to go now

widget/tabcontainer.go
```diff
@@ -128,6 +128,9 @@ func (c *TabContainer) RemoveIndex(index int) {
 // SetItems sets the container’s items and refreshes.
 func (c *TabContainer) SetItems(items []*TabItem) {
 	c.Items = items
+	if c.current >= len(c.Items) {
+		c.current = -1
+	}
```
stuartmscott, line 133: Perhaps we should select the last item, so we only fallback to -1 if there are no items.
```suggestion
	if l := len(c.Items); c.current >= l {
		c.current = l - 1
	}
```

stuartmscott (approved):

stuartmscott:
Thanks for this @adrianre12

widget/tabcontainer.go
```diff
@@ -128,6 +128,9 @@ func (c *TabContainer) RemoveIndex(index int) {
 // SetItems sets the container’s items and refreshes.
 func (c *TabContainer) SetItems(items []*TabItem) {
 	c.Items = items
+	if l := len(c.Items); c.current >= l {
```
andydotxyz, line 131: Can you add a test for this as well please?
Just something simple that shows the selection change.

andydotxyz (changes requested):
Looks great, but I think it's missing a test for the case when len(c.Items) is > 0.
i.e. if it goes from 3 to 2 items (just to differentiate from the -1 case tested)

widget/tabcontainer.go
adrianre12: Yeup, done

andydotxyz (approved):
Sweet, thanks

andydotxyz:
Pending change requests expired. Approved.

PR #1671, opened 2020-12-18
Move the binding and preferences to 64bit int for consistency

andydotxyz:
This PR moves the data binding, and preferences API from `int` to `int64`.
After discovering that `int` could be 32 bit on some systems this makes it more consistent.

However the Go tour recommends to use `int` unless there is a reason not to.
This is a breaking change for Preferences, but not for the data binding which is not yet public.

I am considering that this change is not desirable after all... Thoughts?

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/preferences_test.go
+++ b/app/preferences_test.go
@@ -43,7 +43,7 @@ func TestPreferences_Load(t *testing.T) {
 	p.loadFromFile(filepath.Join("testdata", "preferences.json"))
 
 	assert.Equal(t, "value", p.String("keyString"))
-	assert.Equal(t, 4, p.Int("keyInt"))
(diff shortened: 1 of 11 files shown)


Jacalz (commented):
My view on the matter is that it doesn’t make much sense. 32 bit operating systems are more or less come quite far in on the deprecation phase and those are the cases where the length will be different. As almost all machines nowadays are 64 bit, I think we should just keep int as is.

andydotxyz:
On the whole this doesn't seem to be something that is required just now.
It can be added later if we find that some new use-case emerges.

PR #1676, opened 2020-12-21
Support error reporting in binding

andydotxyz:
Get/Set etc will report or propagate errors.
List/map item lookup will return ok flag.

Update Entry so that we can hook this into the validation

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne_demo/tutorials/bind.go
+++ b/cmd/fyne_demo/tutorials/bind.go
@@ -85,7 +85,7 @@ func newFormWithData(data binding.DataMap) *widget.Form {
 	for i, k := range keys {
 		data, ok := data.GetItem(k)
 		if !ok {
-			continue
+			items[i] = widget.NewFormItem(k, widget.NewLabel("item removed"))
 		}
 		items[i] = widget.NewFormItem(k, createBoundItem(data))
 	}
--- a/cmd/fyne_demo/tutorials/bind.go
+++ b/cmd/fyne_demo/tutorials/bind.go
@@ -59,7 +59,8 @@ func bindingScreen(_ fyne.Window) fyne.CanvasObject {
 
 			btn := obj.(*fyne.Container).Objects[1].(*widget.Button)
 			btn.OnTapped = func() {
-				f.Set(f.Get() + 1)
+				val, _ := f.Get()
+				_ = f.Set(val + 1)
 			}
 		})
 
@@ -82,7 +83,11 @@ func newFormWithData(data binding.DataMap) *widget.Form {
 	keys := data.Keys()
 	items := make([]*widget.FormItem, len(keys))
 	for i, k := range keys {
-		items[i] = widget.NewFormItem(k, createBoundItem(data.GetItem(k)))
+		data, ok := data.GetItem(k)
+		if !ok {
+			continue
+		}
+		items[i] = widget.NewFormItem(k, createBoundItem(data))
 	}
 
 	return widget.NewForm(items...)
--- a/cmd/fyne_demo/tutorials/bind.go
+++ b/cmd/fyne_demo/tutorials/bind.go
@@ -83,9 +83,9 @@ func newFormWithData(data binding.DataMap) *widget.Form {
 	keys := data.Keys()
 	items := make([]*widget.FormItem, len(keys))
 	for i, k := range keys {
-		data, ok := data.GetItem(k)
-		if !ok {
-			items[i] = widget.NewFormItem(k, widget.NewLabel("item removed"))
+		data, err := data.GetItem(k)
+		if err != nil {
+			items[i] = widget.NewFormItem(k, widget.NewLabel(err.Error()))
 		}
 		items[i] = widget.NewFormItem(k, createBoundItem(data))
 	}
--- a/data/binding/convert.go
+++ b/data/binding/convert.go
@@ -56,12 +56,12 @@ func (s *stringFromBool) Set(str string) error {
 	if val == old {
 		return nil
 	}
-	err = s.from.Set(val)
-
-	if err == nil {
-		s.trigger()
+	if err = s.from.Set(val); err != nil {
+		return err
 	}
-	return err
+
+	s.trigger()
+	return nil
 }
 
 func (s *stringFromBool) DataChanged() {
@@ -117,12 +117,12 @@ func (s *stringFromFloat) Set(str string) error {
 	if val == old {
 		return nil
 	}
-	err = s.from.Set(val)
-
-	if err == nil {
-		s.trigger()
+	if err = s.from.Set(val); err != nil {
+		return err
 	}
-	return err
+
+	s.trigger()
+	return nil
 }
 
 func (s *stringFromFloat) DataChanged() {
@@ -178,12 +178,12 @@ func (s *stringFromInt) Set(str string) error {
 	if val == old {
 		return nil
 	}
-	err = s.from.Set(val)
-
-	if err == nil {
-		s.trigger()
+	if err = s.from.Set(val); err != nil {
+		return err
 	}
-	return err
+
+	s.trigger()
+	return nil
 }
 
 func (s *stringFromInt) DataChanged() {
@@ -240,9 +240,12 @@ func (s *stringToBool) Set(val bool) error {
 		return err
 	}
 
-	err = s.from.Set(str)
+	if err = s.from.Set(str); err != nil {
+		return err
+	}
+
 	s.trigger()
-	return err
+	return nil
 }
 
 func (s *stringToBool) DataChanged() {
@@ -299,9 +302,12 @@ func (s *stringToFloat) Set(val float64) error {
 		return err
 	}
 
-	err = s.from.Set(str)
+	if err = s.from.Set(str); err != nil {
+		return err
+	}
+
 	s.trigger()
-	return err
+	return nil
 }
 
 func (s *stringToFloat) DataChanged() {
@@ -358,9 +364,12 @@ func (s *stringToInt) Set(val int) error {
 		return err
 	}
 
-	err = s.from.Set(str)
+	if err = s.from.Set(str); err != nil {
+		return err
+	}
+
 	s.trigger()
-	return err
+	return nil
 }
 
 func (s *stringToInt) DataChanged() {
--- a/data/binding/convert.go
+++ b/data/binding/convert.go
@@ -36,25 +36,26 @@ func BoolToStringWithFormat(v Bool, format string) String {
 	return str
 }
 
-func (s *stringFromBool) Get() string {
-	val := s.from.Get()
+func (s *stringFromBool) Get() (string, error) {
+	val, err := s.from.Get()
 
-	return fmt.Sprintf(s.format, val)
+	return fmt.Sprintf(s.format, val), err
 }
 
-func (s *stringFromBool) Set(str string) {
+func (s *stringFromBool) Set(str string) error {
 	var val bool
 	n, err := fmt.Sscanf(str, s.format, &val)
 	if err != nil || n != 1 {
-		fyne.LogError("bool parse error", err)
-		return
+		return err
 	}
-	if val == s.from.Get() {
-		return
+	old, err := s.from.Get()
+	if val == old {
+		return err
 	}
-	s.from.Set(val)
+	err = s.from.Set(val)
 
 	s.trigger()
+	return err
 }
 
 func (s *stringFromBool) DataChanged() {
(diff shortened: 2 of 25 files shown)


andydotxyz:
I think this needs some more tests for the corner cases, but wanted to share it now as the shape of it looks solid.

data/binding/gen.go
```diff
@@ -130,25 +132,26 @@ func {{ .Name }}ToStringWithFormat(v {{ .Name }}, format string) String {
 	return str
 }
 
-func (s *stringFrom{{ .Name }}) Get() string {
-	val := s.from.Get()
+func (s *stringFrom{{ .Name }}) Get() (string, error) {
+	val, err := s.from.Get()
 
-	return fmt.Sprintf(s.format, val)
+	return fmt.Sprintf(s.format, val), err
```
stuartmscott, line 138: Only format the value if err is nil, we don't want to be returning errors and non-empty results at the same time;
```go
val, err := s.from.Get()
if err != nil {
    return "", err
}
return fmt.Sprintf(s.format, val), nil
```

data/binding/gen.go
```diff
@@ -130,25 +132,26 @@ func {{ .Name }}ToStringWithFormat(v {{ .Name }}, format string) String {
...
 	if err != nil || n != 1 {
-		fyne.LogError("{{ .Type }} parse error", err)
-		return
+		return err
 	}
-	if val == s.from.Get() {
-		return
+	old, err := s.from.Get()
+	if val == old {
+		return err
```
stuartmscott, line 149: This is combining two cases, and it would be better to be more explicit;
```go
old, err := s.from.Get()
if err != nil {
    return err
}
if val == old {
    return nil
}
```

data/binding/gen.go
```diff
@@ -130,25 +132,26 @@ func {{ .Name }}ToStringWithFormat(v {{ .Name }}, format string) String {
...
 	}
-	s.from.Set(val)
+	err = s.from.Set(val)
```
stuartmscott, line 151: Should we trigger if err != nil?

data/binding/gen.go
```diff
@@ -185,30 +188,32 @@ func StringTo{{ .Name }}WithFormat(str String, format string) {{ .Name }} {
 	return v
 }
 
-func (s *stringTo{{ .Name }}) Get() {{ .Type }} {
-	str := s.from.Get()
-	if str == "" {
-		return {{ .Default }}
+func (s *stringTo{{ .Name }}) Get() ({{ .Type }}, error) {
+	str, err := s.from.Get()
```
stuartmscott, line 192: Shouldn't this err get returned if not nil?

data/binding/mapbinding.go
```diff
@@ -89,12 +89,12 @@ type mapBase struct {
 // It will return nil if the key was not found.
 //
 // Since: 2.0.0
-func (b *mapBase) GetItem(key string) DataItem {
+func (b *mapBase) GetItem(key string) (DataItem, bool) {
```
stuartmscott, line 92: Could this be simplified to `return b.items[key]`?

widget/entry.go
```diff
@@ -118,15 +118,30 @@ func (e *Entry) Bind(data binding.String) {
 	e.Unbind()
 	e.textSource = data
 
+	var convertErr error
+	e.Validator = func(string) error {
```
stuartmscott, line 122: In case the dev has already attached Validator we should call that first;
```go
v := e.Validator
e.Validator = func(text string) error {
    if v != nil {
        if err := v(text); err != nil {
            return err
        }
    }
    return convertErr
}
```

stuartmscott: Unbind presents a complication, if we reset the Validator we lose the binding validator and any previous one the dev set. If we don't reset it we could end up chaining binding validators :/

Maybe we flip it around and say devs should only set the Validator after binding?

andydotxyz: I don't think we can do that, as the developer could set e.Validator directly and break this.
I think the only solution would be to support multiple validations (internally?)..?

data/binding/gen.go
andydotxyz: good catch, thanks

data/binding/mapbinding.go
andydotxyz: That doesn't work, believe it or not: `Not enough parameters to return` - I think the compiler does not see the optional second parameter so assumes it's missing and returns just the value...

andydotxyz:
Related to boolean parsing: https://github.com/golang/go/issues/43306

data/binding/mapbinding.go
stuartmscott: Huh, odd

widget/entry.go
stuartmscott: As in instead of `Validator func(string)` it would be `Validators []func(string)`?

cmd/fyne_demo/tutorials/bind.go
```diff
@@ -82,7 +83,11 @@ func newFormWithData(data binding.DataMap) *widget.Form {
 	keys := data.Keys()
 	items := make([]*widget.FormItem, len(keys))
 	for i, k := range keys {
-		items[i] = widget.NewFormItem(k, createBoundItem(data.GetItem(k)))
+		data, ok := data.GetItem(k)
+		if !ok {
+			continue
```
stuartmscott, line 88: Can form deal with one of the items being nil?

data/binding/gen.go
stuartmscott: I'm not sure this is the right pattern, don't we prefer early exits?
```go
if err := s.from.Set(val); err != nil {
    return err
}
s.trigger()
return nil
```

stuartmscott: Also line 221-223

data/binding/listbinding.go
```diff
@@ -1,41 +1,45 @@
...
 	base
-	val []DataItem
+	items []DataItem
 }
 
 // GetItem returns the DataItem at the specified index.
-func (b *listBase) GetItem(i int) DataItem {
-	if i < 0 || i >= len(b.val) {
-		return nil
+func (b *listBase) GetItem(i int) (DataItem, bool) {
```
stuartmscott, line 22: I wonder if `GetItem(i int) (DataItem, error)` would be better so it can return `errOutOfBounds`. Although this means DataList would deviate from DataMap, it seems the least-surprising approach

data/binding/mapbinding.go
```diff
@@ -192,26 +194,31 @@ type boundReflect struct {
...
 
-func (b *boundReflect) get() interface{} {
-	return b.val.Interface()
+func (b *boundReflect) get() (interface{}, error) {
+	return b.val.Interface(), nil
 }
 
-func (b *boundReflect) set(val interface{}) {
+func (b *boundReflect) set(val interface{}) error {
+	// TODO catch the panic and return as error
```
stuartmscott, line 202: Can we get these resolved before merge?

data/binding/convert.go
```diff
@@ -36,25 +34,26 @@ func BoolToStringWithFormat(v Bool, format string) String {
 	return str
 }
 
-func (s *stringFromBool) Get() string {
-	val := s.from.Get()
+func (s *stringFromBool) Get() (string, error) {
+	val, err := s.from.Get()
```
stuartmscott, line 38: This doesn't seem to be the latest from the generator

widget/entry.go
andydotxyz: I don't know. I think we discussed previously `NavigatorSet` or similar, that could contain multiple and check them all as it could implement the same interface...

cmd/fyne_demo/tutorials/bind.go
andydotxyz: no :(

data/binding/listbinding.go
andydotxyz: It depends if we want to match the primitive types behaviour or a consistent higher level API design.
As you say currently Get/Set return an error, whereas GetItem for list and map are both bools.
The only error GetItem could return is the out of bounds, so is a generic error that then has to be checked for the reason the most helpful approach?

data/binding/mapbinding.go
andydotxyz: I forgot, sorry. Not sure but I will look into it.

data/binding/listbinding.go
stuartmscott: Yes, the only error GetItem could return from these in-memory list bindings is out-of-bounds. But if someone implemented a list binding to the lines in a file, or the rows in a database, they could generate a plethora of errors.

widget/entry.go
stuartmscott: Assuming you meant `ValidatorSet`, that would be fine with me, but `Bind` would have to check if the entry's current validator type is `ValidatorSet`, then append the new validator. `Unbind` would also need to remove it.

Perhaps we're overthinking/overcomplicating this :S

data/binding/mapbinding.go
andydotxyz: Added :)

data/binding/listbinding.go
andydotxyz: Are you sure? This is the lookup that returns a binding to a value - so it doesn't do anything with the values...
I think you could make the same argument that map lookup should be `error` instead of `bool` - so I still think they should be consistent, whichever we use.

widget/entry.go
andydotxyz: I wasn't imagining that Bind would check the type at all, it would set a `ValidatorSet` with itself and the `e.Validator` content. Then `Unbind` would unpack it.

I do think that we may be overthining because, as much as this could work, developers could still just stomp all over it because the field is designed to be publicly settable.
I wonder if the only safe solution is to have a second, private, validator field that is also checked and put the logic in Entry...
It's the same complication with `Check.OnChanged`.

data/binding/mapbinding.go
```diff
@@ -1,17 +1,23 @@
...
 import (
+	"errors"
 	"reflect"
 
 	"fyne.io/fyne"
 )
 
+var (
+	errKeyNotFound = errors.New("key not found")
+	errParseFailed = errors.New("format did not match 1 value")
```
stuartmscott, line 12: nit: this error isn't used by this file, consider moving to convert.go

widget/label.go
```diff
@@ -55,7 +56,10 @@ func (l *Label) Bind(data binding.String) {
 	l.Unbind()
 	l.textSource = data
 	l.textListener = binding.NewDataListener(func() {
-		l.SetText(l.textSource.Get())
+		val, err := l.textSource.Get()
+		if err == nil {
+			l.SetText(val)
```
stuartmscott, line 61: As this would create the renderer, could it effect widgets embedding Label?

stuartmscott (commented):
Good work Andy!

I wonder if widgets that don't use the errors returned from bindings (ie. all except Entry) should at least log it with `fyne.LogError`

data/binding/mapbinding.go
andydotxyz: I was trying to keep useful consts like this out of generated files. In this case there was no equivalent non-generated convert file. I guess `binding.go` is another option.

stuartmscott: Yeah maybe we should put all the errors together in the package-named file

widget/label.go
```diff
@@ -30,7 +31,8 @@ func NewLabel(text string) *Label {
 //
 // Since: 2.0.0
 func NewLabelWithData(data binding.String) *Label {
-	label := NewLabel(data.Get())
+	val, _ := data.Get()
+	label := NewLabel(val)
```
stuartmscott, line 35: I don't think this is needed as listener will be notified of current state when added

andydotxyz: The label needs a value in it’s constructor, surely the known current value is better than a zero that will be changed immediately?

stuartmscott: Well given it will be changed immediately the value doesn't really matter, but if the options are `""` (which is essentially free), and `data.Get()` (which could be a disk read, or a network call) then I'd opt for the former.

andydotxyz: If it is "" it will have to change when the new value comes in. If it is set to the current value we avoid the need to re-render the widget.
Of course mostly it won't have rendered yet so maybe it makes less difference than I thought.

andydotxyz: Pushed. 
I had to change the tests to add a wait for binding to resolve as we can no longer guarantee the start state.

stuartmscott (approved):
Nice work!

PR #1680, opened 2020-12-23
sticking plaster fix for clipboard test failure

### LINKED ISSUE #1679: Intermittent clipboard test failure on Windows
### Describe the bug:
I am having trouble reliably running tests on windows. It seems to go through phases of passing multiple times then failing multiple times (like 10 times in a row)
I run 
$ go clean -testcache ./...
$ go test ./...

When it fails I get the error
`2020/12/22 17:36:47 Fyne error:  Failed to focus object which is not part of the canvas’ content, menu or overlays.
2020/12/22 17:36:47   At: C:/Users/Adrian/go/src/github.com/adrianre12/fyne/internal/driver/glfw/canvas.go:85
2020/12/22 17:36:48 Fyne error:  Main menu must have at least one child menu
2020/12/22 17:36:48   At: C:/Users/Adrian/go/src/github.com/adrianre12/fyne/internal/driver/glfw/menu.go:9
2020/12/22 17:36:54 PlatformError: Win32: Failed to open clipboard: Access is denied.
--- FAIL: TestWindow_Clipboard (0.05s)
    window_test.go:864:
                Error Trace:    window_test.go:864
                Error:          Not equal:
                                expected: "My content from test window"
                                actual  : ""

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -My content from test window
                                +
                Test:           TestWindow_Clipboard
FAIL
FAIL    fyne.io/fyne/internal/driver/glfw       14.288s`

I suspect the Access Denied is to do with resource locking, but so far I have not been able to identify a reliable repeatable success or failure scenario.

 - OS: Windows
 - Version: 10
 - Go version: 1.15.5
 - Fyne version: v1.4.2 and develop

adrianre12:
### Description:
This is a sticking plaster fix for the Clipboard Access Denied error when running tests on windows.
Really the glfw should expose the error for correct retry handling, but this retry means at least the Fyne test will pass on windows. Most of the time it now passes with one retry but I have seen two.
I had to run the test 14 times before I could get an example output where the test passed without a retry.
`$ go clean -testcache ./... ; go test ./... -v | grep "Clipboard"
=== RUN   TestWindow_Clipboard
    window_test.go:868: Clipboard Retry
--- PASS: TestWindow_Clipboard (0.55s)

$ go clean -testcache ./... ; go test ./... -v | grep "Clipboard"
=== RUN   TestWindow_Clipboard
--- PASS: TestWindow_Clipboard (0.07s)`

Fixes #1679

### Checklist:
- [ N/A ] Tests included.
- [X] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window_test.go
+++ b/internal/driver/glfw/window_test.go
@@ -851,7 +851,7 @@ func TestWindow_Clipboard(t *testing.T) {
 	text := "My content from test window"
 	cb := w.Clipboard()
 
-	cliboardContent := cb.Content() // this sometimes fails but does not break the test, just means the CB is not restored at the end.
+	cliboardContent := cb.Content()
 	if cliboardContent != "" {
 		// Current environment has some content stored in clipboard,
 		// set temporary to an empty string to allow test and restore later.
@@ -860,19 +860,8 @@ func TestWindow_Clipboard(t *testing.T) {
 
 	assert.Empty(t, cb.Content())
 
-	cb.SetContent(text) // if this fails it will get caught by the retry
-
-	cbText := cb.Content() // if this fails it will get caught by the retry
-	i := 0
-	for cbText == "" && i < 3 { // retry sticking plaster fix for CB Access Denied error
-		t.Log("Clipboard Retry")
-		i++
-		time.Sleep(time.Duration(i*500) * time.Millisecond)
-		cb.SetContent(text)
-		cbText = cb.Content()
-	}
-
-	assert.Equal(t, text, cbText)
+	cb.SetContent(text)
+	assert.Equal(t, text, cb.Content())
 
 	// Restore clipboardContent, if any
 	cb.SetContent(cliboardContent) // this can fail but does not break the test, just means the CB is not restored.
--- a/internal/driver/glfw/window_test.go
+++ b/internal/driver/glfw/window_test.go
@@ -851,14 +851,14 @@ func TestWindow_Clipboard(t *testing.T) {
 	text := "My content from test window"
 	cb := w.Clipboard()
 
-	cliboardContent := cb.Content() // this fails but does not break the test
+	cliboardContent := cb.Content() // this sometimes fails but does not break the test, just means the CB is not restored at the end.
 	if cliboardContent != "" {
 		// Current environment has some content stored in clipboard,
 		// set temporary to an empty string to allow test and restore later.
-		cb.SetContent("") // this fails but does not break the test
+		cb.SetContent("")
 	}
 
-	assert.Empty(t, cb.Content()) // this fails but does not break the test
+	assert.Empty(t, cb.Content())
 
 	cb.SetContent(text) // if this fails it will get caught by the retry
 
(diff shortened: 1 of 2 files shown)


andydotxyz:
Thanks for this.
With reference to the comments `// this fails but does not break the test` should we expect it to fail in the tests? (if so we should check for a failure) Or is it one of the *fails sometimes* areas, (in which case the comment could be clearer).

adrianre12:
having a nightmare with goimports as it is not compatible with CRLF

andydotxyz:
Sorry I did not notice that this was updated.
Looking at the updated comment "// this sometimes fails but does not break the test, just means the CB is not restored at the end." And then seeing that it branches based on "" value immediately after it is not clear what the failure is?
If there is nothing in the clipboard and we get an empty string it does not need to be restored. Is this line a distraction from the real problem?

I see the attempt to retry in the test, but cannot help but think this means that developers will also have to retry until we fix the underlying issue. Why not move the retry into the actuall setter? You would be able to verify that the content is set correctly I would think, and the test should then pass as expected?

adrianre12:
The underlying problem is that GLFW is masking the "Access Denied" error, it catches it and prints it out, but upstream just gets "".  The user can not tell if for write it worked, or for read if the clipboard is empty or if the clipboard is locked. So it is incorrect to put the retry in the setter.  It's up to you guys to decide which way you want to go, accept windows is flaky and remove the test for windows, or try and get GLFW fixed.

andydotxyz:
Thanks for the explanation, but can you expand on why it is wrong to put the retry in the setter?
Is it not possible for the set to check that contents were updated using get?

adrianre12:
Of course, it is possible to check set with a get, but it is fixing a symptom not the cause. I guess it would also be possible to have retry on an empty cb to try and work around the get issue. Definitely a pragmatic solution

adrianre12:
I have moved the retry into the getter/setters, they only run for windows. 
I ran the test over 100 times and it passed 100% 
Does not slow it down too much, 

`=== RUN   TestWindow_Clipboard
2021/01/06 15:34:57 PlatformError: Win32: Failed to open clipboard: Access is denied.
--- PASS: TestWindow_Clipboard (0.27s)
=== RUN   TestWindow_Clipboard
2021/01/06 15:34:57 PlatformError: Win32: Failed to open clipboard: Access is denied.
2021/01/06 15:34:57 PlatformError: Win32: Failed to open clipboard: Access is denied.
--- PASS: TestWindow_Clipboard (0.28s)
=== RUN   TestWindow_Clipboard
--- PASS: TestWindow_Clipboard (0.23s)
=== RUN   TestWindow_Clipboard
2021/01/06 15:34:57 PlatformError: Win32: Failed to open clipboard: Access is denied.
--- PASS: TestWindow_Clipboard (0.28s)
PASS`

internal/driver/glfw/window_test.go
```diff
@@ -864,7 +864,7 @@ func TestWindow_Clipboard(t *testing.T) {
 	assert.Equal(t, text, cb.Content())
 
 	// Restore clipboardContent, if any
-	cb.SetContent(cliboardContent)
+	cb.SetContent(cliboardContent) // this can fail but does not break the test, just means the CB is not restored.
```
andydotxyz, line 867: Is this comment still correct? It seems like it might have been addressed by the rest of the code...

andydotxyz (approved):
Thanks, I think this will help stick over the windows clipboard issues until we can solve them upstream

andydotxyz:
If you don't mind I will call for another opinion on this. I am aware it's a bit of a workaround so want to someone to sanity check what we have agreed.

stuartmscott (approved):

PR #1684, opened 2020-12-24
Fix the release build flag for mobile

andydotxyz:
Release tag had been missed off mobile builds, oops

### Checklist:

- [ ] Tests included. <- nothing around the mobile output package can be/is validated :(
- [x] Tests all pass.

#### Where applicable:


stuartmscott (approved):
Verified locally

PR #1689, opened 2020-12-27
Toggle hidden files

### LINKED ISSUE #1278: File Dialog: toggle hidden files

adrianre12:
### Description:
This is a continuation of kgentile's PR #1282
I have attempted to address the outstanding requested changes.

Fixes #1278

### Checklist:
- [X] Tests included.
- [X] Tests all pass.

This is the dialog/testdata folder not showing hidden files
image

Clicking Options button pops up a list, at the moment only one entry Show Hidden Files. Clicking this refreshes the file dialog and shows the hidden files. I have included the file .hidden in the testdata folder and ensured it works on windows and Linux.

image


dialog/file.go
```diff
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"fyne.io/fyne"
+
```
Jacalz, line 10: I don’t think this new line belongs here.

dialog/file.go
```diff
@@ -181,10 +184,36 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
...
-	favoritesGroup := widget.NewGroup("Favorites", favorites...)
-	return fyne.NewContainerWithLayout(layout.NewBorderLayout(header, footer, favoritesGroup, nil),
-		favoritesGroup, header, footer, body)
+	favoritesGroup := widget.NewVScrollContainer(widget.NewGroup("Favorites", favorites...))
+	var optionsButton *widget.Button
+	optionsButton = widget.NewButton("Options", func() {
+		f.optionsMenu(fyne.CurrentApp().Driver().AbsolutePositionForObject(optionsButton), optionsButton.Size())
+	})
+
+	left := fyne.NewContainerWithLayout(layout.NewBorderLayout(nil, optionsButton, nil, nil), favoritesGroup, optionsButton)
```
Jacalz, line 193: For future reference, `fyne.NewContainerWithLayout()` has been deprecated in favor of `container.New()`. However, you can use friendly container.NewBorder() constructor instead.

dialog/file.go
```diff
@@ -181,10 +184,36 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
...
+
+	return fyne.NewContainerWithLayout(layout.NewBorderLayout(header, footer, left, nil),
```
Jacalz, line 195: Same as above.

dialog/file.go
```diff
@@ -181,10 +184,36 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
...
+}
+
+func (f *fileDialog) optionsMenu(position fyne.Position, buttonSize fyne.Size) {
+	hiddenFiles := widget.NewCheck("Show Hidden Files", func(changed bool) {
+		f.showHidden = changed
+		f.refreshDir(f.dir)
+	})
+	hiddenFiles.SetChecked(f.showHidden)
+
+	title := widget.NewLabel("Options")
```
Jacalz, line 207: I don’t think this options label is needed. The user would have pressed a button with the same name to bring up the menu. I think it can safely be removed from the menu.

dialog/file.go
```diff
@@ -181,10 +184,36 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
...
+	title.Alignment = fyne.TextAlignCenter
+	title.TextStyle = fyne.TextStyle{Bold: true}
+
+	content := widget.NewVBox(widget.NewCard("", "", title),
```
Jacalz, line 211: You should use `container.NewVBox` instead, `widget.NewVBox` is deprecated.
As we probably can ignore the label, you can remove the card here.

dialog/file.go
```diff
@@ -181,10 +184,36 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 
 	favorites := f.loadFavorites()
 
-	favoritesGroup := widget.NewGroup("Favorites", favorites...)
-	return fyne.NewContainerWithLayout(layout.NewBorderLayout(header, footer, favoritesGroup, nil),
-		favoritesGroup, header, footer, body)
+	favoritesGroup := widget.NewVScrollContainer(widget.NewGroup("Favorites", favorites...))
+	var optionsButton *widget.Button
+	optionsButton = widget.NewButton("Options", func() {
```
Jacalz, line 189:
```suggestion
	optionsButton := widget.NewButton("Options", func() {
```
You should just be able to create the button in one step without declaring the var first.

dialog/file.go
```diff
@@ -181,10 +184,36 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 
 	favorites := f.loadFavorites()
 
-	favoritesGroup := widget.NewGroup("Favorites", favorites...)
-	return fyne.NewContainerWithLayout(layout.NewBorderLayout(header, footer, favoritesGroup, nil),
-		favoritesGroup, header, footer, body)
+	favoritesGroup := widget.NewVScrollContainer(widget.NewGroup("Favorites", favorites...))
```
Jacalz, line 187: It might be outside the scope of this PR, but given that we now are relying on scrolling, it probably makes sense to use a `widget.List` instead, for better performance.

Jacalz (changes requested):
Thank you very I have for the PR and the great work you have done in effort of continuing the work of the other PR. I just had a quick look and suggested some changes. This is all I have time with, but I hope that the suggestions makes sense :)

dialog/file.go
adrianre12: Updated to container.NewBorder()

dialog/file.go
adrianre12: Dito

dialog/file.go
adrianre12: Nope, the var is needed, as the func needs to reference the options button.

Jacalz: Oh, right. Sorry. Don’t know how I could miss that.

dialog/file.go
adrianre12: Need more experienced brains than mine to answer that properly, but I would have thought the favourites list is not a huge performance hit as I would expect it to be short.

dialog/file.go
adrianre12: Done

dialog/file.go
adrianre12: A matter of taste I guess, any other views? Here it is without the title
image

Jacalz: Thanks for sharing a picture with it removed. I certainly prefer it that way. With the title label, it mostly looks like another options button to press, IMHO.

dialog/file.go
Jacalz: Sure. You can probably just ignore this comment. It is definitely outside the scope of this PR.

andydotxyz: I don’t think we should start using List in every time that there is a scrolling VBox, the semantics are different. It would mean introducing selection. 
And unless you’re scrolling 100s of items it likely makes little difference. Here we will have somewhere between 5 and 30 elements only.

dialog/file.go
andydotxyz: Erm, the usage of “:=“ will result in the same variable being available for reference...

dialog/file.go
andydotxyz: Perhaps if the menu was above or below, instead of to the right of, the button the header is less needed?

stuartmscott: Or options could be an Accordion

adrianre12:
Doesn't an accordion have to have space to expand into? Isn't a Popup cleaner?

dialog/file.go
stuartmscott: > Doesn't an accordion have to have space to expand into?

Sure, but when Options expands, Favourites collapses so there would be space available

> Isn't a Popup cleaner?

Maybe, we'll only know if explore the different options and compare the results :)

adrianre12: I think that would have to go hand in hand with a complete overhaul. Is what I have made acceptable for release?

dialog/file.go
adrianre12: Yes the use of ":=" creates teh same variable as the var line, but it is created after the inline anonymous function is evaluated which results in the compiler error "undefined: optionsButton"

dialog/file_test.go
```diff
@@ -231,6 +235,70 @@ func TestShowFileOpen(t *testing.T) {
...
+	// assert.False(t, open.Disabled())
+
+	// test.Tap(open)
+	// assert.Nil(t, win.Canvas().Overlays().Top())
+	// assert.Nil(t, openErr)
+
+	// assert.Equal(t, target.location.String(), chosen.URI().String())
+
+	// err = chosen.Close()
+	// assert.Nil(t, err)
```
stuartmscott, line 299: Please uncomment or remove

charlesdaniels:
This might need to be a separate PR, but we should consider also having a keyboard shortcut/shortcuts. I believe that on macOS, this is command+shift+n, and on Linux it's normally control+h. We might want to either support both and/or toggle it by OS.

I don't think this is a blocker for merging this PR, but if it's going to be separated, we should move this into an issue.

dialog/file.go
```diff
@@ -181,10 +184,28 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
...
+	var optionsButton *widget.Button
+	optionsButton = widget.NewButton("Options", func() {
```
stuartmscott, line 189: Consider adding the gear icon;
```suggestion
	optionsButton = widget.NewButtonWithIcon("Options", theme.SettingsIcon(), func() {
```

dialog/file_test.go
adrianre12: Opps my bad, they are gone now

Jacalz:
The tests are failing on animations. I think you might need to rebase on develop to get that fixed (if I remember correctly).

Jacalz:
@adrianre12 Could you please rebase on develop?

adrianre12:
Sry, was having a couple of days off. Merged now and tests passed

Jacalz (approved):
This looks good now. Thank you :+1:

Issue #1297, opened 2020-09-09
Transparent Background


stuartmscott:
### Is your feature request related to a problem? Please describe:
Collection widgets such as List, Table, and Tree should use background colour to indicate cell state such as hovered, or selected, however most widget backgrounds default to an image of uniform colour, which overdraws the collection cell background.

### Is it possible to construct a solution with the existing API?
It is possible to set the theme background colour to transparent, but this is a visual API break.

### Describe the solution you'd like to see:
The default option being no background, with the ability for the dev to set a custom background colour or image.

andydotxyz:
All collections now moved

PR #1690, opened 2020-12-27
Transparent Window

### LINKED ISSUE #181: Transparent background
Thank contributor for awesome repo and sorry for my English.

I want to make an app with a transparent background.
I found the theme color:  color.RGBA{} but param is unit32, A must type float ( example  0.2) to make a transparent background. 

Can you give me some suggestion?

Thank again.

gcstr:
Proof of concept implementing transparent window option.

Fixes #181 and #1297 

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/driver/desktop/driver.go
+++ b/driver/desktop/driver.go
@@ -7,4 +7,7 @@ import "fyne.io/fyne"
 type Driver interface {
 	// Create a new borderless window that is centered on screen
 	CreateSplashWindow() fyne.Window
+
+	// Create a new window that can be transparent and borderless
+	CreateAttributedWindow(title string, decorate bool, transparent bool, centered bool) fyne.Window
 }
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -46,11 +46,12 @@ func initCursors() {
 var _ fyne.Window = (*window)(nil)
 
 type window struct {
-	viewport   *glfw.Window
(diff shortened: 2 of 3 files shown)


andydotxyz:
We cannot break App.NewWindow to be able to add this functionality. Also I doubt very much this is supported on all platforms (certainly is not on mobile) so it should probably be a platform specific extension (like the SplashWindow in driver/desktop).

p.S. you seem to have a lot of changes in window.go that breaks bordered/fixed size as the values are all set to `1` for some reason.

gcstr:
Yes, absolutely. I refactored it to have a similar implementation for what [was done for the splash window](https://github.com/fyne-io/fyne/commit/ac53a579d1f6ca3c67149bc31688ac990c8a9161), putting it under the desktop driver.

But now I see that the whole implementation is useless since there is no support for flexible theming [before v2.0](https://github.com/fyne-io/fyne/wiki/Roadmap).

I'm using it locally with a hard-coded transparent theme, for now. I believe this would not be a priority for the project at the moment, but thanks for taking the time anyway. :)

andydotxyz:
I am not sure about the most recent comment here. Even in Theme v2 there will be a BackgroundColor() - and this colour will be used to draw the window background.
I don't think that this patch is any more or less useful with v2.0.
It may relate more to the deprecation of widget BackgroundColor (#1656), but I'm not sure.

PR #1694, opened 2020-12-29
Reset dir to startingLocation on dismiss or save in file dialog

Jacalz:
### Description:

Saving a reference to the dialog and then reusing it be calling .Show() led to it being at the last used location.
This change makes sure that we return to the start directory on calling .Show().

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -29,7 +29,10 @@ type fileDialog struct {
 
 	win      *widget.PopUp
 	selected *fileDialogItem
-	dir      fyne.ListableURI
+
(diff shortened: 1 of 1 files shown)


andydotxyz:
Are you sure we should do this?
Other widgets that are hidden and then re-shown do not reset state in this way.

Jacalz:
You are right, we certainly don't want to do it this way. Resetting to startingLocation should ideally only be done when the dialog has either been dismissed or saved.

PR #1695, opened 2020-12-29
Support clipping child clips, avoid overflow of nested scroll containers

### LINKED ISSUE #1682: Incorrect clipping behavior when scroll containers are nested
### Describe the bug:

When a scroll container is nested within another scroll container, the clipping behavior is incorrect.

According to @andydotxyz:

> There is a known issue where a scrollContainer inside a ScrollContainer does not clip correctly

> I know what the issue is (child clip should intersect parent rather than replace it - but I have not found time to resolve it

The attached video shows a list (which includes a scroll container), where the list items are cards which have scroll containers in their content. Notice how as the outer container is scrolled, the content is not clipped properly and is displayed inside of the menu bar.

https://user-images.githubusercontent.com/11738895/103025754-74749700-4520-11eb-9eef-075d51b109a8.mp4

I have tested this with Fyne v 1.4.2 on Linux with Go 1.15.6.

andydotxyz:
Fixes issue where scroll container inside scroll container would not correctly clip visuals.
Now we work how how much should be visible by intersecting with parent.

Fixes #1682

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ More detailed release notes can be found on the [releases page](https://github.c
 * Fix possible race conditions for canvas capture
 * Improvements to `fyne get` command downloader
 * Fix tree, so it refreshes visible nodes on Refresh()
+* Incorrect clipping behaviour with nested scroll containers (#1682)
 
 
 ## 1.4.2 - 9 December 2020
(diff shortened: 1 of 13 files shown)


stuartmscott (approved):
Looks good to me

PR #1696, opened 2020-12-29
Improve detection of SVG from resources in case name was wrong

andydotxyz:
Fixes a common case when someone names their svg resource "myImage".
We can't really do this for files as they are streamed off the filesystem. However in those cases it is likely named correctly.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/painter/image.go
+++ b/internal/painter/image.go
@@ -30,17 +30,20 @@ func PaintImage(img *canvas.Image, c fyne.Canvas, width, height int) image.Image
 	case img.File != "" || img.Resource != nil:
 		var file io.Reader
 		var name string
+		isSVG := false
 		if img.Resource != nil {
 			name = img.Resource.Name()
 			file = bytes.NewReader(img.Resource.Content())
+			isSVG = isResourceSVG(img.Resource)
 		} else {
 			name = img.File
 			handle, _ := os.Open(img.File)
 			defer handle.Close()
 			file = handle
+			isSVG = isFileSVG(img.File)
 		}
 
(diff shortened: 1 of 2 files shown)


internal/painter/image.go
```diff
@@ -163,3 +166,28 @@ func checkImageMinSize(img *canvas.Image, c fyne.Canvas, pixX, pixY int) bool {
...
+
+	if strings.ToUpper(string(res.Content()[:5])) == "<!DOC" {
+		return true
+	}
+	if strings.ToLower(string(res.Content()[:5])) == "<?xml" {
+		return true
+	}
+	if strings.ToLower(string(res.Content()[:5])) == "<svg " {
+		return true
+	}
```
stuartmscott, line 191: Could be simplified;
```suggestion
	switch strings.ToUpper(string(res.Content()[:5])) {
		case "<!DOC", "<?xml", "<svg ":
			return true
	}
```

stuartmscott (approved):

internal/painter/image.go
```diff
@@ -30,17 +30,20 @@ func PaintImage(img *canvas.Image, c fyne.Canvas, width, height int) image.Image
 	case img.File != "" || img.Resource != nil:
 		var file io.Reader
 		var name string
+		isSVG := false
```
stuartmscott, line 33: if anymore vars are added here we could move to;
```suggestion
		var (
			file io.Reader
			name string
			isSVG bool
		)
```

PR #1700, opened 2020-12-30
Use unique identifier for mac notifications instead of incremental integer

### LINKED ISSUE #1699: MacOS Notifications are not shown on subsequent app runs after first run due to duplicate identifier
### Describe the bug:
On MacOS, an incremental counter is used for notification identifier. So, app shows notifications on first run, but in subsequent runs notifications are assigned same identifier from `0`. If notifications with these identifiers are already present in the notification center, mac doesn't show these notifications as popup again and replaces old notification content with a new one inside the notification center itself.

### To Reproduce:
1. Run the mac app with notifications
2. Don't interact with notifications, let them go in the notification center
3. Quit app and restart
4. Notifications won't popup on second run

### Device:
 - OS: MacOS
 - Version: 11.1
 - Go version: 1.15.5
 - Fyne version: 1.4.2

AdityaHirapara:
### Description:
On macOS, an incremental counter is used for notification identifier. So, notifications are not shown on subsequent app runs after the first run due to duplicate identifiers. Use unique identifiers instead of a counter to solve this.

Fixes #1699 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/app_darwin.m
+++ b/app/app_darwin.m
@@ -22,8 +22,6 @@ extern void themeChanged();
 
 @end
 
-static int notifyNum = 0;
-
 void sendNSUserNotification(const char *, const char *);
 
 bool isBundled() {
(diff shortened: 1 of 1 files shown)


andydotxyz (approved):
Sweet fix thanks

PR #1710, opened 2021-01-02
Align size and position to closest pixel for crisp output

andydotxyz:
This is most noticable on low resolution screens where the new floating-point position could look fuzzy (most notably text, but also lines etc).
Not only does it now look as crisp as when we used integers I think that on a HiDPI this now looks exceptional :)

### Checklist:

- [ ] Tests included. <- GL only, just look at any app and see that the text is clear
- [x] Tests all pass.

### DIFF
--- a/internal/painter/gl/draw.go
+++ b/internal/painter/gl/draw.go
@@ -2,6 +2,7 @@ package gl
 
 import (
 	"image/color"
+	"math"
 
 	"fyne.io/fyne"
 	"fyne.io/fyne/canvas"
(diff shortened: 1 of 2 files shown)


internal/painter/gl/draw.go
```diff
@@ -158,3 +113,67 @@ func (p *glPainter) drawObject(o fyne.CanvasObject, pos fyne.Position, frame fyn
...
+		x1, y1, 0, 0.0, 0.0, // bottom left
+		x2, y2, 0, 1.0, 1.0, // top right
+		x2, y1, 0, 1.0, 0.0, // bottom right
+	}
+}
+
+func rectInnerCoords(size fyne.Size, pos fyne.Position, fill canvas.ImageFill, aspect float32) (fyne.Size, fyne.Position) {
+	if fill == canvas.ImageFillContain || fill == canvas.ImageFillOriginal {
+		// change pos and size accordingly
+
```
Jacalz, line 145: Is this newline intended to be here?

Jacalz (approved):
Thank you very much. Applications are now razor sharp again (almost feel better than when we had the integer positions). :+1: 

I added a comment on a newline that might be unintended, but I ma going approve anyway :)

internal/painter/gl/draw.go
andydotxyz: This was just moving the code around to fix ordering, it was a separate commit so you can see it was just that ;)

PR #1711, opened 2021-01-02
Fix various issues with how List and Map advertised Get/Set

andydotxyz:
Issues raised during chat about re-loading data in binding

a) Get and Set were not possible for the underlying list or struct, so the current Get/Set was renamed GetValue/SetValue
b) New Get and Set were added to provide access to the core data, checking for changes where possible for set
c) trigger the container on size/key change and otherwise trigger value listeners if the data item changed.
d) export the `Untyped` type because sometimes we will need it just like `Int`

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


AlbinoGeek (commented):
Love the direction :) This allows us to (`ini` `json` `yaml`) `Unmarshal` and `Set()` as discussed in Slack, without having to create and re-assign the binding variable.

AlbinoGeek (approved):
[ not a reviewer, but ]

Considering the discussion in slack, I think this implementation looks well, and would handle (for example) `Unmarshal` correctly now. The only missing bit belongs in another PR anyways (concerning the `Reload` method.)

This feels more consistent with the other binding types. It looks good to me.

Jacalz (approved):
Looks like a sane approach to me.

Issue #1647, opened 2020-12-11
Menu and PopUp keyboard control


andydotxyz:
Part of the large list at #1515, but menus are the biggest unresolved component.

toaster:
I think this is done with #2283.

andydotxyz:
Agreed :)

Issue #1515, opened 2020-11-08
Make keyboard controls a first class citizen


aksdb:
I hereby want to propose to make keyboard controls a first class citizen in the design of Fyne and especially its widgets.

Some other tickets already talk about accessibility (#1093 and #1285), so I think thinking about inputs is a very important step in that direction. There a multiple disabilities that make it hard to properly navigate a mouse or touchpad, and even with voice control (or joysticks) it is easier to give directional commands (up, down, left, right, next, back) than moving a cursor around.

A *widget* is already the base for every active component of the application. It can therefore be assumed, that each widget will in some way deal with user input (clicking, scrolling, entering content, etc.).

My proposal would therefore be, to implement the `Keyable` interface on the base `Widget`. (Bonus: expose `OnKeyDown`, `OnKeyUp`, `OnKeyTyped` etc. callbacks for custom code *outside* the widget (for the user of the widget; not for a sub-"class").)

The *FocusManager* (#1294) can likely deal with the tab order and deal with switching focus.

## Examples of key behavior:

* **Global**: *tab* focuses the next component, *shift+tab* the previous
* **List**: up/down moves the currently selected item.
* **Accordion**: up/down moves to the previous/next item, left/right folds/expands the current item
* **Tabs**: depending on the orientation left/right or up/down moves between tabs
* **Select**: up/down changes the current selection
* **Slider**: space/enter toggles the state
* **Table**: left/right/up/down navigates between cells
* **Tree**: up/down changes selected node, left/right expands/folds a node with children
* **Check**: space/enter toggles the state
* **Radio**: up/down selects the previous/next item

### Special scenario: focus trap

For example a text control with formatting might want to use tab and shift+tab to actually indent or unindent text. For these cases, it must be possible to differentiate between "edit" and "navigate". Giving such a textbox focus, does *not* enter edit mode. Pressing *enter* would enter edit mode. Then tab/shift+tab are no longer used for navigation, but are propagated to the widget for use. Hitting *escape* will leave edit mode and switch back to navigiation mode. The widget is still highlighted/focused (so pressing *enter* would go back into edit mode, but tab/shift+tab would now navigate between controls).

Therefore it should be possible to mark a widget as *FocusTrappable* (or something like that). In this case, the *FocusManager* allows hitting "enter" to enter said trap. Afterwards *tab* and *shift+tab* are propagated to this widget (while it has focus). *Esc*(ape) however is handled by the *FocusManager* to take that exclusive/trap mode away again.

[...]

PR #1712, opened 2021-01-02
Feature/keyboard controllable main menu

toaster:
### Description:

This makes the main menu keyboard controllable. It is the first (and biggest) part of #1647.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
A quick question here... You have added `FocusNext` and `FocusPrevious to the `Canvas` API - but only used it internally to the drivers as far as I can see... could it not just be local to the glCanvas and mobileCanvas?

Jacalz:
I think I might be able to answer that question @andydotxyz. I am one of the people that really want those methods. It is going to make changing focus a lot easier then the current way of keeping references to the things that want to be focused. It would clean up the code in Sparta that currently looks like this: https://github.com/Jacalz/sparta/blob/develop/internal/gui/widgets/entry_advanced.go.

andydotxyz:
I think the right fix in your situation @Jacalz is that we implement "move on when enter pressed in form if single line", which is a separate issue.
I guess there are happy side-effects of exposing the methods here, but is that the reason that they were added? :)

Jacalz:
> I think the right fix in your situation @Jacalz is that we implement "move on when enter pressed in form if single line", which is a separate issue.
> I guess there are happy side-effects of exposing the methods here, but is that the reason that they were added? :)
I am not talking about widgets inside of a form and neither am I talking about pressing enter. The on Enter code is only about executing a function. I am talking about pressing up or down arrow to move between widgets in a VBox. Adding this is a good addition in my opinion 🙂

andydotxyz:
> I am talking about pressing up or down arrow to move between widgets in a VBox

Ah OK thanks, was hard to tell from just reading the code. As above I don't mean to say it isn't useful, but was asking if it is required for the PR...

Jacalz:
> Ah OK thanks, was hard to tell from just reading the code.
Hehe, I can see that. That code is in serious need for a rewrite 🙃

andydotxyz (changes requested):
It looks like `FocusNext` and `FocusPrevious` on the `Canvas` API are not required for this change.

Jacalz:
> It looks like `FocusNext` and `FocusPrevious` on the `Canvas` API are not required for this change.

Can we perhaps have that moved out to a separate PR then? I still think it is a very good and useful addition 🙂

toaster:
> > It looks like `FocusNext` and `FocusPrevious` on the `Canvas` API are not required for this change.
> 
> Can we perhaps have that moved out to a separate PR then? I still think it is a very good and useful addition 🙂

I will do that.

Jacalz (commented):
Just a heads up. Looks like tests are failing importunately.

andydotxyz:
I got a crash if I pressed Enter on a folder:

```

andydotxyz (changes requested):
Can you look into the crash I put in the channel? (selecting a folder menu).

Also in the original #1515 it was added that `Esc` should close the menu - is that in scope for this change as well?

andydotxyz (approved):
This is great, thanks

PR #1717, opened 2021-01-04
Provide a way to inform the external variable based bindings that they should refresh state.

### LINKED ISSUE #1645: Data Binding refresh from external bound variables
The calls `data.binding.BindXxx[Yyy]` calls connect an existing variable into a data binding.
At this time external changes to that data are not immediately reflected in the data, (though `Get()` will function correctly).
We want to ensure that external changes trigger the internal refresh including value adjustments and callbacks.

# Scope
## Affects

BindXxx, BindXxxList, BindUntypedMap and BindStruct

## Does not affect

NewXxx, BindPreferenceXxx, NewXxxList, NewUntypedMap, external implementations of the Xxx and XxxList, DataItem, DataList or DataMap type interfaces.

Developers can implement their own data bindings, doing so does not require this additional consideration as they will trigger data binding when their data source has changed.

# Requirements

## Primitive

This is not a problem when the data source is internal (NewXxx) or in some other programmatic source (such as a conversion or widget binding).

We should inform of changes in the variable, rather than simply returning from Getter when requested.
```
f := 5.0
b := BindFloat(&f)
f = 6.2 // does not trigger change
```

## Lists 

This is not a problem when the list is managed intermally such as NewXxxList.

BindXxxList should ensure changes of the slice size (1), or ordering (2), or individual elements changed (3);
```
strings := []string{"hi"}
b := BindStringList(&strings)

strings = append(strings, "d") // (1)
strings[0], strings[2] = strings[2], strings[1] // (2)
strings[0]+="aa" // (3)
```

## Maps 

This is not a problem when the map is managed internally such as NewUntypedMap.

BindUntypedMap (which connects binding to a `map[string]interface{}` type) should check if maps items have been added (1) or removed (2), or individual elements changed (3);
```
vals := map[string]interface{}{
	"foo": "bar",
	"val": 5,
	"bas": 0.2,
}
b := BindUntypedMap(&vals)

vals["key"] = 4.3 // (1)
delete(vals, "key") //  (2)
vals["foo"] = "hi" // (3)
```

## Structs

BindStruct (which connects a struct to the data binding system) should check which of the items have changed (1);
```
s := struct {
	Foo string
	Val int
	Bas float64
}{
	"bar",
	5,
	0.2,
}
b := BindStruct(&s)

s.Foo = "Hello" // (1)
s.Val = 3 // (1)
```

# Proposals

The aim of these possible approaches is to enable the data to be updated when it changes from an external source (specifically for the BindXxxType calls - not needed for NewXxxType or third party implementations). The solution should ensure that the API can trigger change event propagation but it should not introduce potentially confusion in what is a minimal API.

## Adding function to base DataItem

Alongside the `AddListener` and `RemoveListener` we could have a method `CallListeners` (or similar).
When the external data changes, the `CallListeners` is called to trigger any listeners of this data.

**Pros**
The simplest implementation, no additional types and remains consistent.

**Cons**
This adds the method to all type bindings, so `NewInt` would have additional `CallListeners` despte it not being needed. This approach does not imply any internal logic, so listners may be called when nothing changed.

## Export new type extending the `Type` interfaces that exports `Reload`

With this approach we see `BindXxx` return `ExternalInt` (or similar) in place of current `Int`, that indicates the return has different semantics.
Calling `Reload` on the new type can check the data, and make appropriate set / trigger calls.

**Pros**
This approach minimises the potential confusion of the new method, it is only exported on types that might need to have it called.

**Cons**
It introduces many new types, people may not understand how `ExternalInt` relates to `Int`.

## Monitor for changes internally

With this approach there is no new API. We check regularly (*very* regularly) for each data to have changed, and update bindings if that is the cae.

**Pros**
No additional work for the developers, no additional APIs

**Cons** 
Performance, to check regularly enough to be transparent we need to do so many times a second, at least 10, up to 60. Unfortunately as we implemented lists of structs we realised that this could mean thousands, or higher, pointer/value comparisons on every single change.
Consider this inside a table or list data source and you see how slow it can get.

andydotxyz:
OK, primarily this is about adding `Reload` to external data types where that is required.
This is BindXxx (not NewXxx) BindXxxList (not NewXxxList), BindUntypedMap (not NewUntypedMap) and BindStruct.
These interfaces can be ignored by anybody not using external variable binding, so to widgets, converters, validators etc this completely transparent.

During this development I also found that our set tests were a little lacking in map and list - re-using the same variable had unintended consequences. These are now fixed as well and all the tests pass.

To be finished
- [x] Calling `Reload` when the value has not changed should not trigger on primitive types
- [x] Calling `Reload` on a list/map/struct should not trigger the child element callbacks if the value has not changed

Fixes #1645

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/gen.go
+++ b/data/binding/gen.go
@@ -308,10 +308,10 @@ func Bind{{ .Name }}List(v *[]{{ .Type }}) External{{ .Name }}List {
 		return New{{ .Name }}List().(External{{ .Name }}List)
 	}
 
-	b := &bound{{ .Name }}List{val: v}
+	b := &bound{{ .Name }}List{val: v, updateExternal: true}
 
 	for i := range *v {
-		b.appendItem(bind{{ .Name }}ListItem(v, i))
+		b.appendItem(bind{{ .Name }}ListItem(v, i, b.updateExternal))
 	}
 
 	return b
@@ -320,7 +320,8 @@ func Bind{{ .Name }}List(v *[]{{ .Type }}) External{{ .Name }}List {
 type bound{{ .Name }}List struct {
 	listBase
 
-	val *[]{{ .Type }}
+	updateExternal bool
+	val            *[]{{ .Type }}
 }
 
 func (l *bound{{ .Name }}List) Append(val {{ .Type }}) error {
@@ -382,7 +383,7 @@ func (l *bound{{ .Name }}List) doReload() (retErr error) {
 		l.trigger()
 	} else if oldLen < newLen {
 		for i := oldLen; i < newLen; i++ {
-			l.appendItem(bind{{ .Name }}ListItem(l.val, i))
+			l.appendItem(bind{{ .Name }}ListItem(l.val, i, l.updateExternal))
 		}
 		l.trigger()
 	}
@@ -392,7 +393,12 @@ func (l *bound{{ .Name }}List) doReload() (retErr error) {
 			break
 		}
 
-		err := item.(*bound{{ .Name }}ListItem).setIfChanged((*l.val)[i])
+		var err error
+		if l.updateExternal {
+			err = item.(*boundExternal{{ .Name }}ListItem).setIfChanged((*l.val)[i])
+		} else {
+			err = item.(*bound{{ .Name }}ListItem).doSet((*l.val)[i])
+		}
 		if err != nil {
 			retErr = err
 		}
@@ -416,8 +422,15 @@ func (l *bound{{ .Name }}List) SetValue(i int, v {{ .Type }}) error {
 	return item.({{ .Name }}).Set(v)
 }
 
-func bind{{ .Name }}ListItem(v *[]{{ .Type }}, i int) {{ .Name }} {
-	return &bound{{ .Name }}ListItem{val: v, index: i, old: (*v)[i]}
+func bind{{ .Name }}ListItem(v *[]{{ .Type }}, i int, external bool) {{ .Name }} {
+	if external {
+		ret := &boundExternal{{ .Name }}ListItem{old: (*v)[i]}
+		ret.val = v
+		ret.index = i
+		return ret
+	}
+
+	return &bound{{ .Name }}ListItem{val: v, index: i}
 }
 
 type bound{{ .Name }}ListItem struct {
@@ -425,7 +438,6 @@ type bound{{ .Name }}ListItem struct {
 
 	val   *[]{{ .Type }}
 	index int
-	old   {{ .Type }}
 }
 
 func (b *bound{{ .Name }}ListItem) Get() ({{ .Type }}, error) {
@@ -439,13 +451,23 @@ func (b *bound{{ .Name }}ListItem) Set(val {{ .Type }}) error {
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
+	return b.doSet(val)
+}
+
+func (b *bound{{ .Name }}ListItem) doSet(val {{ .Type }}) error {
 	(*b.val)[b.index] = val
 
 	b.trigger()
 	return nil
 }
 
-func (b *bound{{ .Name }}ListItem) setIfChanged(val {{ .Type }}) error {
+type boundExternal{{ .Name }}ListItem struct {
+	bound{{ .Name }}ListItem
+
+	old {{ .Type }}
+}
+
+func (b *boundExternal{{ .Name }}ListItem) setIfChanged(val {{ .Type }}) error {
 	if val == b.old {
 		return nil
 	}
--- a/data/binding/gen.go
+++ b/data/binding/gen.go
@@ -270,7 +270,7 @@ type {{ .Name }}List interface {
 //
 // Since: 2.0.0
 func New{{ .Name }}List() {{ .Name }}List {
-	return &bound{{ .Name }}List{}
+	return &bound{{ .Name }}List{val: &[]{{ .Type }}{}}
 }
 
 // Bind{{ .Name }}List returns a bound list of {{ .Type }} values, based on the contents of the passed slice.
@@ -297,19 +297,13 @@ type bound{{ .Name }}List struct {
 }
 
 func (l *bound{{ .Name }}List) Append(val {{ .Type }}) error {
-	if l.val != nil {
-		*l.val = append(*l.val, val)
-	}
+	*l.val = append(*l.val, val)
 
 	l.appendItem(Bind{{ .Name }}(&val))
 	return nil
 }
 
 func (l *bound{{ .Name }}List) Get() ([]{{ .Type }}, error) {
-	if l.val == nil {
-		return []{{ .Type }}{}, nil
-	}
-
 	return *l.val, nil
 }
 
@@ -317,33 +311,17 @@ func (l *bound{{ .Name }}List) GetValue(i int) ({{ .Type }}, error) {
 	if i < 0 || i >= l.Length() {
 		return {{ .Default }}, errOutOfBounds
 	}
-	if l.val != nil {
-		return (*l.val)[i], nil
-	}
-
-	item, err := l.GetItem(i)
-	if err != nil {
-		return {{ .Default }}, err
-	}
-	return item.({{ .Name }}).Get()
+	return (*l.val)[i], nil
 }
 
 func (l *bound{{ .Name }}List) Prepend(val {{ .Type }}) error {
-	if l.val != nil {
-		*l.val = append([]{{ .Type }}{val}, *l.val...)
-	}
+	*l.val = append([]{{ .Type }}{val}, *l.val...)
 
 	l.prependItem(Bind{{ .Name }}(&val))
 	return nil
 }
 
 func (l *bound{{ .Name }}List) Set(v []{{ .Type }}) (retErr error) {
-	if l.val == nil { // was not initialized with a blank value, recover
-		l.val = &v
-		l.trigger()
-		return nil
-	}
-
 	oldLen := len(l.items)
 	*l.val = v
 	newLen := len(v)
@@ -380,9 +358,7 @@ func (l *bound{{ .Name }}List) SetValue(i int, v {{ .Type }}) error {
 	if i < 0 || i >= l.Length() {
 		return errOutOfBounds
 	}
-	if l.val != nil {
-		(*l.val)[i] = v
-	}
+	(*l.val)[i] = v
 
 	item, err := l.GetItem(i)
 	if err != nil {
--- a/data/binding/gen.go
+++ b/data/binding/gen.go
@@ -67,34 +67,13 @@ func (b *bound{{ .Name }}) Get() ({{ .Type }}, error) {
 }
 
 func (b *bound{{ .Name }}) Reload() error {
-	b.lock.Lock()
-	defer b.lock.Unlock()
-
-	return b.setIfChanged(*b.val)
+	return b.Set(*b.val)
 }
 
 func (b *bound{{ .Name }}) Set(val {{ .Type }}) error {
 	b.lock.Lock()
 	defer b.lock.Unlock()
-
-	if *b.val == val {
-		return nil
-	}
-	if b.val == nil { // was not initialized with a blank value, recover
-		b.val = &val
-	} else {
-		*b.val = val
-	}
-
-	b.trigger()
-	return nil
-}
-
-func (b *bound{{ .Name }}) setIfChanged(val {{ .Type }}) error {
-	if val == b.old {
-		return nil
-	}
-	b.old = val
+	*b.val = val
 
 	b.trigger()
 	return nil
--- a/data/binding/gen.go
+++ b/data/binding/gen.go
@@ -395,9 +395,13 @@ func (l *bound{{ .Name }}List) doReload() (retErr error) {
 
 		var err error
 		if l.updateExternal {
+			item.(*boundExternal{{ .Name }}ListItem).lock.Lock()
 			err = item.(*boundExternal{{ .Name }}ListItem).setIfChanged((*l.val)[i])
+			item.(*boundExternal{{ .Name }}ListItem).lock.Unlock()
 		} else {
+			item.(*bound{{ .Name }}ListItem).lock.Lock()
 			err = item.(*bound{{ .Name }}ListItem).doSet((*l.val)[i])
+			item.(*bound{{ .Name }}ListItem).lock.Unlock()
 		}
 		if err != nil {
 			retErr = err
--- a/data/binding/gen.go
+++ b/data/binding/gen.go
@@ -57,6 +57,9 @@ type bound{{ .Name }} struct {
 }
 
 func (b *bound{{ .Name }}) Get() ({{ .Type }}, error) {
+	b.lock.RLock()
+	defer b.lock.RUnlock()
+
 	if b.val == nil {
 		return {{ .Default }}, nil
 	}
@@ -64,10 +67,16 @@ func (b *bound{{ .Name }}) Get() ({{ .Type }}, error) {
 }
 
 func (b *bound{{ .Name }}) Reload() error {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
 	return b.setIfChanged(*b.val)
 }
 
 func (b *bound{{ .Name }}) Set(val {{ .Type }}) error {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
 	if *b.val == val {
 		return nil
 	}
@@ -128,6 +137,8 @@ func (b *prefBound{{ .Name }}) Get() ({{ .Type }}, error) {
 func (b *prefBound{{ .Name }}) Set(v {{ .Type }}) error {
 	b.p.Set{{ .Name }}(b.key, v)
 
+	b.lock.RLock()
+	defer b.lock.RUnlock()
 	b.trigger()
 	return nil
 }
@@ -199,11 +210,13 @@ func (s *stringFrom{{ .Name }}) Set(str string) error {
 		return err
 	}
 
-	s.trigger()
+	s.DataChanged()
 	return nil
 }
 
 func (s *stringFrom{{ .Name }}) DataChanged() {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
 	s.trigger()
 }
 `
@@ -266,11 +279,13 @@ func (s *stringTo{{ .Name }}) Set(val {{ .Type }}) error {
 		return err
 	}
 
-	s.trigger()
+	s.DataChanged()
 	return nil
 }
 
 func (s *stringTo{{ .Name }}) DataChanged() {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
 	s.trigger()
 }
 `
@@ -331,12 +346,18 @@ type bound{{ .Name }}List struct {
 }
 
 func (l *bound{{ .Name }}List) Append(val {{ .Type }}) error {
+	l.lock.Lock()
+	defer l.lock.Unlock()
+
 	*l.val = append(*l.val, val)
 
 	return l.doReload()
 }
 
 func (l *bound{{ .Name }}List) Get() ([]{{ .Type }}, error) {
+	l.lock.RLock()
+	defer l.lock.RUnlock()
+
 	return *l.val, nil
 }
 
@@ -344,20 +365,30 @@ func (l *bound{{ .Name }}List) GetValue(i int) ({{ .Type }}, error) {
 	if i < 0 || i >= l.Length() {
 		return {{ .Default }}, errOutOfBounds
 	}
+	l.lock.RLock()
+	defer l.lock.RUnlock()
+
 	return (*l.val)[i], nil
 }
 
 func (l *bound{{ .Name }}List) Prepend(val {{ .Type }}) error {
+	l.lock.Lock()
+	defer l.lock.Unlock()
 	*l.val = append([]{{ .Type }}{val}, *l.val...)
 
 	return l.doReload()
 }
 
 func (l *bound{{ .Name }}List) Reload() error {
+	l.lock.Lock()
+	defer l.lock.Unlock()
+
 	return l.doReload()
 }
 
 func (l *bound{{ .Name }}List) Set(v []{{ .Type }}) error {
+	l.lock.Lock()
+	defer l.lock.Unlock()
 	*l.val = v
 
 	return l.doReload()
@@ -395,7 +426,10 @@ func (l *bound{{ .Name }}List) SetValue(i int, v {{ .Type }}) error {
 	if i < 0 || i >= l.Length() {
 		return errOutOfBounds
 	}
+
+	l.lock.Lock()
 	(*l.val)[i] = v
+	l.lock.Unlock()
 
 	item, err := l.GetItem(i)
 	if err != nil {
@@ -417,10 +451,16 @@ type bound{{ .Name }}ListItem struct {
 }
 
 func (b *bound{{ .Name }}ListItem) Get() ({{ .Type }}, error) {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
 	return (*b.val)[b.index], nil
 }
 
 func (b *bound{{ .Name }}ListItem) Set(val {{ .Type }}) error {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
 	(*b.val)[b.index] = val
 
 	b.trigger()
--- a/data/binding/gen.go
+++ b/data/binding/gen.go
@@ -323,8 +323,7 @@ type bound{{ .Name }}List struct {
 func (l *bound{{ .Name }}List) Append(val {{ .Type }}) error {
 	*l.val = append(*l.val, val)
 
-	l.appendItem(Bind{{ .Name }}(&val))
-	return nil
+	return l.doReload()
 }
 
 func (l *bound{{ .Name }}List) Get() ([]{{ .Type }}, error) {
@@ -341,14 +340,22 @@ func (l *bound{{ .Name }}List) GetValue(i int) ({{ .Type }}, error) {
 func (l *bound{{ .Name }}List) Prepend(val {{ .Type }}) error {
 	*l.val = append([]{{ .Type }}{val}, *l.val...)
 
-	l.prependItem(Bind{{ .Name }}(&val))
-	return nil
+	return l.doReload()
 }
 
-func (l *bound{{ .Name }}List) Set(v []{{ .Type }}) (retErr error) {
-	oldLen := len(l.items)
+func (l *bound{{ .Name }}List) Reload() error {
+	return l.doReload()
+}
+
+func (l *bound{{ .Name }}List) Set(v []{{ .Type }}) error {
 	*l.val = v
-	newLen := len(v)
+
+	return l.doReload()
+}
+
+func (l *bound{{ .Name }}List) doReload() (retErr error) {
+	oldLen := len(l.items)
+	newLen := len(*l.val)
 	if oldLen > newLen {
 		for i := oldLen - 1; i >= newLen; i-- {
 			l.deleteItem(i)
@@ -366,14 +373,16 @@ func (l *bound{{ .Name }}List) Set(v []{{ .Type }}) (retErr error) {
 			break
 		}
 
-		old, err := l.items[i].({{ .Name }}).Get()
-		val := (*(l.val))[i]
-		if err != nil || (*(l.val))[i] != old {
-			err = item.(*bound{{ .Name }}).Set(val)
-			if err != nil {
-				retErr = err
-			}
-		}
+// TODO cache values and do comparison - for now we just always trigger child elements
+//		old, err := l.items[i].({{ .Name }}).Get()
+//		val := (*(l.val))[i]
+//		if err != nil || (*(l.val))[i] != old {
+//			err = item.(*bound{{ .Name }}).Set(val)
+//			if err != nil {
+//				retErr = err
+//			}
+//		}
+		item.(*bound{{ .Name }}ListItem).trigger()
 	}
 	return
 }
--- a/data/binding/gen.go
+++ b/data/binding/gen.go
@@ -34,7 +34,7 @@ type External{{ .Name }} interface {
 // Since: 2.0.0
 func New{{ .Name }}() {{ .Name }} {
 	blank := {{ .Default }}
-	return &bound{{ .Name }}{val: &blank, old: blank}
+	return &bound{{ .Name }}{val: &blank}
 }
 
 // Bind{{ .Name }} returns a new bindable value that controls the contents of the provided {{ .Type }} variable.
@@ -53,7 +53,6 @@ type bound{{ .Name }} struct {
 	base
 
 	val *{{ .Type }}
-	old {{ .Type }}
 }
 
 func (b *bound{{ .Name }}) Get() ({{ .Type }}, error) {
--- a/data/binding/gen.go
+++ b/data/binding/gen.go
@@ -34,7 +34,7 @@ type External{{ .Name }} interface {
 // Since: 2.0.0
 func New{{ .Name }}() {{ .Name }} {
 	blank := {{ .Default }}
-	return &bound{{ .Name }}{val: &blank}
+	return &bound{{ .Name }}{val: &blank, old: blank}
 }
 
 // Bind{{ .Name }} returns a new bindable value that controls the contents of the provided {{ .Type }} variable.
@@ -53,6 +53,7 @@ type bound{{ .Name }} struct {
 	base
 
 	val *{{ .Type }}
+	old {{ .Type }} 
 }
 
 func (b *bound{{ .Name }}) Get() ({{ .Type }}, error) {
@@ -63,8 +64,7 @@ func (b *bound{{ .Name }}) Get() ({{ .Type }}, error) {
 }
 
 func (b *bound{{ .Name }}) Reload() error {
-	b.trigger() // TODO we should cache the old value and compare
-	return nil
+	return b.setIfChanged(*b.val)
 }
 
 func (b *bound{{ .Name }}) Set(val {{ .Type }}) error {
@@ -76,6 +76,16 @@ func (b *bound{{ .Name }}) Set(val {{ .Type }}) error {
 	} else {
 		*b.val = val
 	}
+
+	b.trigger()
+	return nil
+}
+
+func (b *bound{{ .Name }}) setIfChanged(val {{ .Type }}) error {
+	if val == b.old {
+		return nil
+	}
+	b.old = val
 
 	b.trigger()
 	return nil
@@ -373,16 +383,10 @@ func (l *bound{{ .Name }}List) doReload() (retErr error) {
 			break
 		}
 
-		// TODO cache values and do comparison - for now we just always trigger child elements
-		//		old, err := l.items[i].({{ .Name }}).Get()
-		//		val := (*(l.val))[i]
-		//		if err != nil || (*(l.val))[i] != old {
-		//			err = item.(*bound{{ .Name }}).Set(val)
-		//			if err != nil {
-		//				retErr = err
-		//			}
-		//		}
-		item.(*bound{{ .Name }}ListItem).trigger()
+		err := item.(*bound{{ .Name }}ListItem).setIfChanged((*l.val)[i])
+		if err != nil {
+			retErr = err
+		}
 	}
 	return
 }
@@ -401,7 +405,7 @@ func (l *bound{{ .Name }}List) SetValue(i int, v {{ .Type }}) error {
 }
 
 func bind{{ .Name }}ListItem(v *[]{{ .Type }}, i int) {{ .Name }} {
-	return &bound{{ .Name }}ListItem{val: v, index: i}
+	return &bound{{ .Name }}ListItem{val: v, index: i, old: (*v)[i]}
 }
 
 type bound{{ .Name }}ListItem struct {
@@ -409,6 +413,7 @@ type bound{{ .Name }}ListItem struct {
 
 	val   *[]{{ .Type }}
 	index int
+	old   {{ .Type }}
 }
 
 func (b *bound{{ .Name }}ListItem) Get() ({{ .Type }}, error) {
@@ -417,6 +422,17 @@ func (b *bound{{ .Name }}ListItem) Get() ({{ .Type }}, error) {
 
 func (b *bound{{ .Name }}ListItem) Set(val {{ .Type }}) error {
 	(*b.val)[b.index] = val
+
+	b.trigger()
+	return nil
+}
+
+func (b *bound{{ .Name }}ListItem) setIfChanged(val {{ .Type }}) error {
+	if val == b.old {
+		return nil
+	}
+	(*b.val)[b.index] = val
+	b.old = val
 
 	b.trigger()
 	return nil
--- a/data/binding/gen.go
+++ b/data/binding/gen.go
@@ -53,7 +53,7 @@ type bound{{ .Name }} struct {
 	base
 
 	val *{{ .Type }}
-	old {{ .Type }} 
+	old {{ .Type }}
 }
 
 func (b *bound{{ .Name }}) Get() ({{ .Type }}, error) {
--- a/data/binding/gen.go
+++ b/data/binding/gen.go
@@ -308,7 +308,7 @@ func Bind{{ .Name }}List(v *[]{{ .Type }}) External{{ .Name }}List {
 	b := &bound{{ .Name }}List{val: v}
 
 	for i := range *v {
-		b.appendItem(Bind{{ .Name }}(&((*v)[i])))
+		b.appendItem(bind{{ .Name }}ListItem(v, i))
 	}
 
 	return b
@@ -363,7 +363,7 @@ func (l *bound{{ .Name }}List) doReload() (retErr error) {
 		l.trigger()
 	} else if oldLen < newLen {
 		for i := oldLen; i < newLen; i++ {
-			l.appendItem(Bind{{ .Name }}(&((*l.val)[i])))
+			l.appendItem(bind{{ .Name }}ListItem(l.val, i))
 		}
 		l.trigger()
 	}
@@ -398,6 +398,28 @@ func (l *bound{{ .Name }}List) SetValue(i int, v {{ .Type }}) error {
 		return err
 	}
 	return item.({{ .Name }}).Set(v)
+}
+
+func bind{{ .Name }}ListItem(v *[]{{ .Type }}, i int) {{ .Name }} {
+	return &bound{{ .Name }}ListItem{val: v, index: i}
+}
+
+type bound{{ .Name }}ListItem struct {
+	base
+
+	val   *[]{{ .Type }}
+    index int
+}
+
+func (b *bound{{ .Name }}ListItem) Get() ({{ .Type }}, error) {
+	return (*b.val)[b.index], nil
+}
+
+func (b *bound{{ .Name }}ListItem) Set(val {{ .Type }}) error {
+	(*b.val)[b.index] = val
+
+	b.trigger()
+	return nil
 }
 `
 
--- a/data/binding/gen.go
+++ b/data/binding/gen.go
@@ -21,6 +21,14 @@ type {{ .Name }} interface {
 	Set({{ .Type }}) error
 }
 
+// External{{ .Name }} supports binding a {{ .Type }} value to an external value.
+//
+// Since: 2.0.0
+type External{{ .Name }} interface {
+	{{ .Name }}
+	Reload() error
+}
+
 // New{{ .Name }} returns a bindable {{ .Type }} value that is managed internally.
 //
 // Since: 2.0.0
@@ -30,11 +38,12 @@ func New{{ .Name }}() {{ .Name }} {
 }
 
 // Bind{{ .Name }} returns a new bindable value that controls the contents of the provided {{ .Type }} variable.
+// If your code changes the content of the variable this refers to you should call Reload() to inform the bindings.
 //
 // Since: 2.0.0
-func Bind{{ .Name }}(v *{{ .Type }}) {{ .Name }} {
+func Bind{{ .Name }}(v *{{ .Type }}) External{{ .Name }} {
 	if v == nil {
-		return New{{ .Name }}() // never allow a nil value pointer
+		return New{{ .Name }}().(External{{ .Name }}) // never allow a nil value pointer
 	}
 
 	return &bound{{ .Name }}{val: v}
@@ -51,6 +60,11 @@ func (b *bound{{ .Name }}) Get() ({{ .Type }}, error) {
 		return {{ .Default }}, nil
 	}
 	return *b.val, nil
+}
+
+func (b *bound{{ .Name }}) Reload() error {
+	b.trigger() // TODO we should cache the old value and compare
+	return nil
 }
 
 func (b *bound{{ .Name }}) Set(val {{ .Type }}) error {
@@ -266,6 +280,15 @@ type {{ .Name }}List interface {
 	SetValue(int, {{ .Type }}) error
 }
 
+// External{{ .Name }}List supports binding a list of {{ .Type }} values from an external variable.
+//
+// Since: 2.0.0
+type External{{ .Name }}List interface {
+	{{ .Name }}List
+
+	Reload() error
+}
+
 // New{{ .Name }}List returns a bindable list of {{ .Type }} values.
 //
 // Since: 2.0.0
@@ -274,11 +297,12 @@ func New{{ .Name }}List() {{ .Name }}List {
 }
 
 // Bind{{ .Name }}List returns a bound list of {{ .Type }} values, based on the contents of the passed slice.
+// If your code changes the content of the slice this refers to you should call Reload() to inform the bindings.
 //
 // Since: 2.0.0
-func Bind{{ .Name }}List(v *[]{{ .Type }}) {{ .Name }}List {
+func Bind{{ .Name }}List(v *[]{{ .Type }}) External{{ .Name }}List {
 	if v == nil {
-		return New{{ .Name }}List()
+		return New{{ .Name }}List().(External{{ .Name }}List)
 	}
 
 	b := &bound{{ .Name }}List{val: v}
--- a/data/binding/gen.go
+++ b/data/binding/gen.go
@@ -373,15 +373,15 @@ func (l *bound{{ .Name }}List) doReload() (retErr error) {
 			break
 		}
 
-// TODO cache values and do comparison - for now we just always trigger child elements
-//		old, err := l.items[i].({{ .Name }}).Get()
-//		val := (*(l.val))[i]
-//		if err != nil || (*(l.val))[i] != old {
-//			err = item.(*bound{{ .Name }}).Set(val)
-//			if err != nil {
-//				retErr = err
-//			}
-//		}
+		// TODO cache values and do comparison - for now we just always trigger child elements
+		//		old, err := l.items[i].({{ .Name }}).Get()
+		//		val := (*(l.val))[i]
+		//		if err != nil || (*(l.val))[i] != old {
+		//			err = item.(*bound{{ .Name }}).Set(val)
+		//			if err != nil {
+		//				retErr = err
+		//			}
+		//		}
 		item.(*bound{{ .Name }}ListItem).trigger()
 	}
 	return
@@ -408,7 +408,7 @@ type bound{{ .Name }}ListItem struct {
 	base
 
 	val   *[]{{ .Type }}
-    index int
+	index int
 }
 
 func (b *bound{{ .Name }}ListItem) Get() ({{ .Type }}, error) {
--- a/data/binding/mapbinding.go
+++ b/data/binding/mapbinding.go
@@ -116,10 +116,6 @@ type mapBase struct {
 	val   *map[string]interface{}
 }
 
-// GetItem returns the DataItem at the specified key.
-// It will return nil if the key was not found.
-//
-// Since: 2.0.0
 func (b *mapBase) GetItem(key string) (DataItem, error) {
 	if v, ok := b.items[key]; ok {
 		return v, nil
@@ -128,9 +124,6 @@ func (b *mapBase) GetItem(key string) (DataItem, error) {
 	return nil, errKeyNotFound
 }
 
-// Keys returns a list of all the keys in this data map.
-//
-// Since: 2.0.0
 func (b *mapBase) Keys() []string {
 	ret := make([]string, len(b.items))
 	// TODO lock
@@ -143,9 +136,6 @@ func (b *mapBase) Keys() []string {
 	return ret
 }
 
-// Delete removes the specified key and tha value associated with it.
-//
-// Since: 2.0.0
 func (b *mapBase) Delete(key string) {
 	delete(b.items, key)
 
@@ -160,9 +150,6 @@ func (b *mapBase) Get() (map[string]interface{}, error) {
 	return *b.val, nil
 }
 
-// Get returns the value stored at the specified key.
-//
-// Since: 2.0.0
 func (b *mapBase) GetValue(key string) (interface{}, error) {
 	if i, ok := b.items[key]; ok {
 		return i.(Untyped).get()
@@ -235,10 +222,6 @@ func (b *mapBase) doReload() (retErr error) {
 	return
 }
 
-// Set stores the value d at the specified key.
-// If the key is not present it will create a new binding internally.
-//
-// Since: 2.0.0
 func (b *mapBase) SetValue(key string, d interface{}) error {
 	if i, ok := b.items[key]; ok {
 		i.(Untyped).set(d)
--- a/data/binding/mapbinding.go
+++ b/data/binding/mapbinding.go
@@ -207,17 +207,11 @@ func (b *mapBase) doReload() (retErr error) {
 		b.trigger()
 	}
 
-	for _, item := range b.items {
-		// TODO cache values and do comparison - for now we just always trigger child elements
-		//		old, err := b.items[k].(Untyped).get()
-		//		val := (*(b.val))[k]
-		//		if err != nil || (*(b.val))[k] != old {
-		//			err = item.(Untyped).set(val)
-		//			if err != nil {
-		//				retErr = err
-		//			}
-		//		}
-		item.(*boundUntyped).trigger()
+	for k, item := range b.items {
+		err := item.(*boundUntyped).setIfChanged((*b.val)[k])
+		if err != nil {
+			retErr = err
+		}
 	}
 	return
 }
@@ -241,7 +235,7 @@ func (b *mapBase) setItem(key string, d DataItem) {
 }
 
 func bindUntypedMapValue(m *map[string]interface{}, k string) Untyped {
-	return &boundUntyped{val: m, key: k}
+	return &boundUntyped{val: m, key: k, old: (*m)[k]}
 }
 
 type boundUntyped struct {
@@ -249,6 +243,7 @@ type boundUntyped struct {
 
 	val *map[string]interface{}
 	key string
+	old interface{}
 }
 
 func (b *boundUntyped) get() (interface{}, error) {
@@ -261,6 +256,17 @@ func (b *boundUntyped) get() (interface{}, error) {
 
 func (b *boundUntyped) set(val interface{}) error {
 	(*b.val)[b.key] = val
+
+	b.trigger()
+	return nil
+}
+
+func (b *boundUntyped) setIfChanged(val interface{}) error {
+	if val == b.old {
+		return nil
+	}
+	(*b.val)[b.key] = val
+	b.old = val
 
 	b.trigger()
 	return nil
--- a/data/binding/mapbinding.go
+++ b/data/binding/mapbinding.go
@@ -122,6 +122,9 @@ type mapBase struct {
 }
 
 func (b *mapBase) GetItem(key string) (DataItem, error) {
+	b.lock.RLock()
+	defer b.lock.RUnlock()
+
 	if v, ok := b.items[key]; ok {
 		return v, nil
 	}
@@ -130,8 +133,10 @@ func (b *mapBase) GetItem(key string) (DataItem, error) {
 }
 
 func (b *mapBase) Keys() []string {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
 	ret := make([]string, len(b.items))
-	// TODO lock
 	i := 0
 	for k := range b.items {
 		ret[i] = k
@@ -142,12 +147,18 @@ func (b *mapBase) Keys() []string {
 }
 
 func (b *mapBase) Delete(key string) {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
 	delete(b.items, key)
 
 	b.trigger()
 }
 
 func (b *mapBase) Get() (map[string]interface{}, error) {
+	b.lock.RLock()
+	defer b.lock.RUnlock()
+
 	if b.val == nil {
 		return map[string]interface{}{}, nil
 	}
@@ -156,6 +167,9 @@ func (b *mapBase) Get() (map[string]interface{}, error) {
 }
 
 func (b *mapBase) GetValue(key string) (interface{}, error) {
+	b.lock.RLock()
+	defer b.lock.RUnlock()
+
 	if i, ok := b.items[key]; ok {
 		return i.(Untyped).get()
 	}
@@ -164,10 +178,16 @@ func (b *mapBase) GetValue(key string) (interface{}, error) {
 }
 
 func (b *mapBase) Reload() error {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
 	return b.doReload()
 }
 
 func (b *mapBase) Set(v map[string]interface{}) error {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
 	if b.val == nil { // was not initialized with a blank value, recover
 		b.val = &v
 		b.trigger()
@@ -179,6 +199,9 @@ func (b *mapBase) Set(v map[string]interface{}) error {
 }
 
 func (b *mapBase) SetValue(key string, d interface{}) error {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
 	if i, ok := b.items[key]; ok {
 		i.(Untyped).set(d)
 		return nil
@@ -216,7 +239,7 @@ func (b *mapBase) doReload() (retErr error) {
 			}
 		}
 		if !found {
-			b.Delete(key)
+			delete(b.items, key)
 			changed = true
 		}
 	}
@@ -246,6 +269,9 @@ type boundStruct struct {
 }
 
 func (b *boundStruct) Reload() (retErr error) {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
 	v := reflect.ValueOf(b.orig).Elem()
 	t := v.Type()
 	for j := 0; j < v.NumField(); j++ {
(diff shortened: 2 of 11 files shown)


stuartmscott (changes requested):
Data bindings should not maintain an internal copy of external data just to avoid triggering listeners when nothing changes. This PR doubles the memory usage of data bindings just to save a few CPU/GPU cycles. As soon as the data size exceeds the L1-L3 cache sizes, any cycles saved by this approach will be blown away when the CPU has to fetch from main memory.

Worse still, this PR doubles the memory usage of non-external primitive bindings which in the best case (binding to a single datum) is wasteful, and in the worst case (binding to a slice of 1 million datums) is not scalable.

The non-collection widgets that support bindings (Check, Entry, Label, ProgressBar, Slider) already store an internal copy of the data they are displaying so their listeners should instead compare the value from the binding with the value currently inside the widget and only update and refresh the widget if the data is new.

A lot of work went into the collection widgets to avoid as much as possible maintaining internal copies of large data sets, this PR in comparison feels like a step backwards.

The notion that a UI toolkit knows more about the data than the developer is harmful. The "optimizations" in this PR are devised to avoid doing work that the developer has explicitly requested.

andydotxyz:
> The notion that a UI toolkit knows more about the data than the developer is harmful. The "optimizations" in this PR are devised to avoid doing work that the developer has explicitly requested.

The alternative would be to add lots of methods such as to inform the binding "I updated item 5 in that list" or "field 'foo' of my struct was changed" if we want to allow developers to specify that something small-in-scope has changed. If we do not want to do that (and I can't imagine we want such a complex API) then the only alternative I can see is that the bindings remain completely dumb. If we take that route and I have a 100'000 long []string that is then bound, and I then want to change 1 element we go from 1 callback to 100'001 callbacks. I don't see how that is desirable...

> Worse still, this PR doubles the memory usage of non-external primitive bindings which in the best case (binding to a single datum) is wasteful, and in the worst case (binding to a slice of 1 million datums) is not scalable.

Fair point, I was trying to reduce the number of new types as that seemed to be a concern of some earlier discssions.
I can certainly split out the in-memory from the external-track versions of the primitives to avoid these excess bytes.

andydotxyz:
Another use-case that was discussed in the channel was that of JSON un-marshalling. Commonly a long list of maps or structs we may want to re-load to discover a single change. If we remove the cached values for this comparison then it will call back on every single node of that tree, plus every list/map/struct branch as well. Surely it is desirable to remove this overhead from the developers code?

okratitan:
> > The notion that a UI toolkit knows more about the data than the developer is harmful. The "optimizations" in this PR are devised to avoid doing work that the developer has explicitly requested.
> 
> The alternative would be to add lots of methods such as to inform the binding "I updated item 5 in that list" or "field 'foo' of my struct was changed" if we want to allow developers to specify that something small-in-scope has changed. If we do not want to do that (and I can't imagine we want such a complex API) then the only alternative I can see is that the bindings remain completely dumb. If we take that route and I have a 100'000 long []string that is then bound, and I then want to change 1 element we go from 1 callback to 100'001 callbacks. I don't see how that is desirable...
> 
> > Worse still, this PR doubles the memory usage of non-external primitive bindings which in the best case (binding to a single datum) is wasteful, and in the worst case (binding to a slice of 1 million datums) is not scalable.
> 
> Fair point, I was trying to reduce the number of new types as that seemed to be a concern of some earlier discssions.
> I can certainly split out the in-memory from the external-track versions of the primitives to avoid these excess bytes.

Wouldn't the optimization be better from the collections widget side? Only have listeners for the visible items? Then if you want to change one item you may have to go through 20 to 30 callbacks not 100,001 ... as the updates to any other item would be gotten through the visual changes such as scrolling

stuartmscott:
> a 100'000 long []string that is then bound, and I then want to change 1 element we go from 1 callback to 100'001 callbacks

Yes changing one item will generate 100001 function calls, at 2 ns/function call (https://billglover.me/2018/09/17/how-expensive-is-a-go-function-call/) thats 200002 nanoseconds so ~0.2 milliseconds.

But let's compare apples to apples here, from this PR:
```go
func (l *boundBoolList) doReload() (retErr error) {
/* in the example given an element in a slice is updated, but the slice itself hasn't changed length so we can ignore this
	oldLen := len(l.items)
	newLen := len(*l.val)
	if oldLen > newLen {
		for i := oldLen - 1; i >= newLen; i-- {
			l.deleteItem(i)
		}
		l.trigger()
	} else if oldLen < newLen {
		for i := oldLen; i < newLen; i++ {
			l.appendItem(bindBoolListItem(l.val, i))
		}
		l.trigger()
	}
*/
	for i, item := range l.items {
		if i > oldLen || i > newLen {
			break
		}

		// So this is called 100000 times and in the function there will be a read from main memory, so ~100 ns (https://gist.github.com/jboner/2841832)
		err := item.(*boundBoolListItem).setIfChanged((*l.val)[i])
		if err != nil {
			retErr = err
		}
	}
	return
}
```
So you are trading 100001 function calls (0.2ms) for 100000 function calls + 100000 main memory references (0.2ms + 10ms = 10.2ms) so this PR with all its "optimizations" is actually 100x slower as well as taking up 2x more memory.

Furthermore, as the collection widgets unbind items when they are reused and bound to a new one most of the bindings in the slice will not have any listeners so apart from acquiring the mutex, base.trigger() is essentially a noop.

Which reminds me, this PR has no locking and thus prone to race conditions if two threads call Reload.

> JSON un-marshalling

Why couldn't a JSON-specific binding be introduced? As it stands the entire string of JSON will have to get un-marshalled into structs (lots of memory allocations) and compared with copies internal to bindings (lots of main-memory references) to end up with a few listener callbacks (function calls - super cheap in the grand scheme of things). Whereas if bindings didn't maintain an internal copy then we trade the expensive main-memory references for cheap function calls, which as I said above, will mostly be noops.

stuartmscott:
I was interested in getting more accurate numbers so I made a benchmark:

```go
package main

import (
	"fyne.io/fyne/data/binding"
	"testing"
)

func BenchmarkBinding(b *testing.B) {
	count := b.N * 100000

	// Create a slice full of 0.0s
	data := make([]float64, count)

	// Create binding from slice
	bind := binding.BindFloatList(&data)

	for i := 0; i < count; i++ {
		// Update value in slice
		data[i] = float64(i)
		// Notify binding
		bind.Reload()
	}
}
```

andydotxyz/fix/1645:
```
$ go test -bench=BenchmarkBinding -benchmem
goos: darwin
goarch: amd64
pkg: github.com/stuartmscott/bindingbench
BenchmarkBinding-4   	       1	89653953023 ns/op	18072032 B/op	  100047 allocs/op
PASS
ok  	github.com/stuartmscott/bindingbench	89.831s
```

Control:
```
diff --git a/data/binding/binding.go b/data/binding/binding.go
index cb1a9e36..6cabc55a 100644
--- a/data/binding/binding.go
+++ b/data/binding/binding.go
@@ -30,6 +30,9 @@ type DataItem interface {
        // RemoveListener will detach the specified change listener from the DataItem.
        // Disconnected listener will no longer be triggered when changes occur.
        RemoveListener(DataListener)
+       // Reload notifies the listeners of a binding that its data has changed.
+       // Functions like `Set(float64) error` will call this automatically.
+       Reload()
 }
 
 // DataListener is any object that can register for changes in a bindable DataItem.
@@ -87,6 +90,10 @@ func (b *base) RemoveListener(l DataListener) {
        }
 }
 
+func (b *base) Reload() {
+       b.trigger()
+}
+
 func (b *base) trigger() {
        b.lock.RLock()
        defer b.lock.RUnlock()
```

```
$ go test -bench=BenchmarkBinding -benchmem
goos: darwin
goarch: amd64
pkg: github.com/stuartmscott/bindingbench
BenchmarkBinding-4   	      52	  28341861 ns/op	15344164 B/op	  100001 allocs/op
PASS
ok  	github.com/stuartmscott/bindingbench	2.027s
```

89653953023 ns/op vs 28341861 ns/op makes this PR ~3163x slower than the control.

stuartmscott:
I realized the control was not fair as it was not triggering the listeners of the elements of a list or map.

New Control:
```
diff --git a/data/binding/binding.go b/data/binding/binding.go
index cb1a9e36..6cabc55a 100644
--- a/data/binding/binding.go
+++ b/data/binding/binding.go
@@ -30,6 +30,9 @@ type DataItem interface {
        // RemoveListener will detach the specified change listener from the DataItem.
        // Disconnected listener will no longer be triggered when changes occur.
        RemoveListener(DataListener)
+       // Reload notifies the listeners of a binding that its data has changed.
+       // Functions like `Set(float64) error` will call this automatically.
+       Reload()
 }
 
 // DataListener is any object that can register for changes in a bindable DataItem.
@@ -87,6 +90,10 @@ func (b *base) RemoveListener(l DataListener) {
        }
 }
 
+func (b *base) Reload() {
+       b.trigger()
+}
+
 func (b *base) trigger() {
        b.lock.RLock()
        defer b.lock.RUnlock()
diff --git a/data/binding/listbinding.go b/data/binding/listbinding.go
index bba87ed0..1f298f87 100644
--- a/data/binding/listbinding.go
+++ b/data/binding/listbinding.go
@@ -45,3 +45,10 @@ func (b *listBase) prependItem(i DataItem) {
 
        b.trigger()
 }
+
+func (b *listBase) Reload() {
+       for _, i := range b.items {
+               i.Reload()
+       }
+       b.trigger()
+}
diff --git a/data/binding/mapbinding.go b/data/binding/mapbinding.go
index c7f48684..e8132ec8 100644
--- a/data/binding/mapbinding.go
+++ b/data/binding/mapbinding.go
@@ -230,6 +230,13 @@ func (b *mapBase) SetValue(key string, d interface{}) error {
        return nil
 }
 
+func (b *mapBase) Reload() {
+       for _, i := range b.items {
+               i.Reload()
+       }
+       b.trigger()
+}
+
 func (b *mapBase) setItem(key string, d DataItem) {
        b.items[key] = d
```

```
$ go test -bench=BenchmarkBinding -benchmem -memprofile mem.out -cpuprofile cpu.out
goos: darwin
goarch: amd64
pkg: github.com/stuartmscott/bindingbench
BenchmarkBinding-4   	       1	213667783648 ns/op	16472032 B/op	  100047 allocs/op
PASS
ok  	github.com/stuartmscott/bindingbench	214.081s
```

The result above is much larger and means the control is ~2.38x slower than this PR (213667783648 ns/op vs 89653953023 ns/op).

What can I say? Benchmarks are hard.

Looking at the profile it is apparent that the Mutex is main cause of slowness;
```
$ go tool pprof control.cpu.out 
Type: cpu
Time: Jan 4, 2021 at 6:35pm (PST)
Duration: 3.56mins, Total samples = 3.44mins (96.57%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 205.50s, 99.50% of 206.53s total
Dropped 49 nodes (cum <= 1.03s)
      flat  flat%   sum%        cum   cum%
    75.08s 36.35% 36.35%     75.63s 36.62%  sync.(*RWMutex).RUnlock
    73.25s 35.47% 71.82%     73.27s 35.48%  sync.(*RWMutex).RLock (inline)
    24.33s 11.78% 83.60%    173.69s 84.10%  fyne.io/fyne/data/binding.(*base).trigger
    22.95s 11.11% 94.71%    205.41s 99.46%  fyne.io/fyne/data/binding.(*listBase).Reload
     8.48s  4.11% 98.82%    182.21s 88.22%  fyne.io/fyne/data/binding.(*base).Reload
     1.32s  0.64% 99.46%      1.32s  0.64%  runtime.asyncPreempt
     0.09s 0.044% 99.50%    205.51s 99.51%  github.com/stuartmscott/bindingbench.BenchmarkBinding
         0     0% 99.50%    205.51s 99.51%  testing.(*B).run1.func1
         0     0% 99.50%    205.51s 99.51%  testing.(*B).runN
```

These benchmarks are still not a fair comparison and should be re-run once this PR has the appropriate locking.

---

As for memory usage, the new control uses 9978.08kB total and this PR uses 11874.42kB, making this PR ~1.19x larger. The 1986kB difference is ~2.5x the amount needed to store the 100000 float64s (781.25kB).

Another consideration as I mentioned above is cache size. If the data can fit in the caches it can be accessed more quickly than from main memory. 100000 float64s can easily fit in my Intel i7 L3 Cache (4 MB). If we increase replace the float64s with strings the numbers get even more interesting;

benchmark:
```go
package main

import (
	"fmt"
	"fyne.io/fyne/data/binding"
	"testing"
)

const (
	formatA = "%d: Lorem ipsum dolor sit amet, consectetur adipiscing elit."
	formatB = "Lorem ipsum dolor sit amet, consectetur adipiscing elit.: %d"
)

func BenchmarkBinding(b *testing.B) {
	count := b.N * 100000

	// Create a slice full of ""s
	data := make([]string, count)

	// Create binding from slice
	bind := binding.BindStringList(&data)

	// Format A
	for i := 0; i < count; i++ {
		// Update value in slice
		data[i] = fmt.Sprintf(formatA, i)
		// Notify binding
		bind.Reload()
	}

	// Format B
	for i := 0; i < count; i++ {
		// Update value in slice
		data[i] = fmt.Sprintf(formatB, i)
		// Notify binding
		bind.Reload()
	}
}
```

andydotxyz:fix/1645:
```
$ go test -bench=BenchmarkBinding -benchmem -memprofile 1717.mem.out -cpuprofile 1717.cpu.out -timeout 1h
goos: darwin
goarch: amd64
pkg: github.com/stuartmscott/bindingbench
BenchmarkBinding-4   	       1	221340101286 ns/op	33314816 B/op	  499567 allocs/op
PASS
ok  	github.com/stuartmscott/bindingbench	221.715s

$ go tool pprof 1717.cpu.out 
Type: cpu
Time: Jan 4, 2021 at 8:35pm (PST)
Duration: 3.69mins, Total samples = 3.45mins (93.57%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 203.11s, 98.01% of 207.23s total
Dropped 91 nodes (cum <= 1.04s)
      flat  flat%   sum%        cum   cum%
   115.73s 55.85% 55.85%    129.24s 62.37%  fyne.io/fyne/data/binding.(*boundStringListItem).setIfChanged
    74.60s 36.00% 91.84%    204.14s 98.51%  fyne.io/fyne/data/binding.(*boundStringList).doReload
    12.71s  6.13% 97.98%     12.71s  6.13%  runtime.memequal
     0.05s 0.024% 98.00%    205.32s 99.08%  github.com/stuartmscott/bindingbench.BenchmarkBinding
     0.02s 0.0097% 98.01%    204.16s 98.52%  fyne.io/fyne/data/binding.(*boundStringList).Reload
         0     0% 98.01%      1.66s   0.8%  runtime.mstart
         0     0% 98.01%      1.18s  0.57%  runtime.mstart1
         0     0% 98.01%      1.18s  0.57%  runtime.sysmon
         0     0% 98.01%    205.32s 99.08%  testing.(*B).run1.func1
         0     0% 98.01%    205.32s 99.08%  testing.(*B).runN

$ go tool pprof 1717.mem.out 
Type: alloc_space
Time: Jan 4, 2021 at 8:39pm (PST)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 26226.31kB, 100% of 26226.31kB total
Showing top 10 nodes out of 14
      flat  flat%   sum%        cum   cum%
12288.75kB 46.86% 46.86% 12288.75kB 46.86%  fmt.Sprintf
 6886.29kB 26.26% 73.11%  6886.29kB 26.26%  fyne.io/fyne/data/binding.(*listBase).appendItem (inline)
 4608.35kB 17.57% 90.69%  4608.35kB 17.57%  fyne.io/fyne/data/binding.bindStringListItem (inline)
 1028.32kB  3.92% 94.61%  1930.90kB  7.36%  runtime/pprof.(*profileBuilder).emitLocation
  902.59kB  3.44% 98.05%   902.59kB  3.44%  compress/flate.NewWriter
  512.01kB  1.95%   100% 24295.40kB 92.64%  github.com/stuartmscott/bindingbench.BenchmarkBinding
         0     0%   100%   902.59kB  3.44%  compress/gzip.(*Writer).Write
         0     0%   100% 11494.65kB 43.83%  fyne.io/fyne/data/binding.BindStringList
         0     0%   100%  1930.90kB  7.36%  runtime/pprof.(*profileBuilder).appendLocsForStack
         0     0%   100%  1930.90kB  7.36%  runtime/pprof.(*profileBuilder).build
```

New Control:
```
$ go test -bench=BenchmarkBinding -benchmem -memprofile control.mem.out -cpuprofile control.cpu.out -timeout 1h
goos: darwin
goarch: amd64
pkg: github.com/stuartmscott/bindingbench
BenchmarkBinding-4   	       1	438277421920 ns/op	31707920 B/op	  499573 allocs/op
PASS
ok  	github.com/stuartmscott/bindingbench	438.668s

$ go tool pprof control.cpu.out 
Type: cpu
Time: Jan 4, 2021 at 8:25pm (PST)
Duration: 7.31mins, Total samples = 6.88mins (94.22%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 407.45s, 98.63% of 413.10s total
Dropped 91 nodes (cum <= 2.07s)
      flat  flat%   sum%        cum   cum%
   151.15s 36.59% 36.59%    151.19s 36.60%  sync.(*RWMutex).RLock
   147.59s 35.73% 72.32%    148.42s 35.93%  sync.(*RWMutex).RUnlock
    46.31s 11.21% 83.53%    346.77s 83.94%  fyne.io/fyne/data/binding.(*base).trigger
    45.38s 10.99% 94.51%    409.40s 99.10%  fyne.io/fyne/data/binding.(*listBase).Reload
    16.97s  4.11% 98.62%    363.77s 88.06%  fyne.io/fyne/data/binding.(*base).Reload
     0.05s 0.012% 98.63%    410.55s 99.38%  github.com/stuartmscott/bindingbench.BenchmarkBinding
         0     0% 98.63%    410.55s 99.38%  testing.(*B).run1.func1
         0     0% 98.63%    410.55s 99.38%  testing.(*B).runN

$ go tool pprof control.mem.out 
Type: alloc_space
Time: Jan 4, 2021 at 8:32pm (PST)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 21421.24kB, 100% of 21421.24kB total
Showing top 10 nodes out of 23
      flat  flat%   sum%        cum   cum%
 9728.59kB 45.42% 45.42%  9728.59kB 45.42%  fmt.Sprintf
 4787.96kB 22.35% 67.77%  4787.96kB 22.35%  fyne.io/fyne/data/binding.(*listBase).appendItem (inline)
 2048.03kB  9.56% 77.33% 17588.65kB 82.11%  github.com/stuartmscott/bindingbench.BenchmarkBinding
 1184.27kB  5.53% 82.86%  1184.27kB  5.53%  runtime/pprof.StartCPUProfile
 1024.06kB  4.78% 87.64%  1024.06kB  4.78%  fyne.io/fyne/data/binding.BindString (inline)
  902.59kB  4.21% 91.85%  2136.21kB  9.97%  compress/flate.NewWriter
  650.62kB  3.04% 94.89%  1233.63kB  5.76%  compress/flate.(*compressor).init
  583.01kB  2.72% 97.61%   583.01kB  2.72%  compress/flate.newDeflateFast (inline)
  512.10kB  2.39%   100%   512.10kB  2.39%  runtime/pprof.allFrames
         0     0%   100%  2136.21kB  9.97%  compress/gzip.(*Writer).Write
```

| |Control | This PR | Delta |
| --- | --- | --- |
| CPU | 438277421920 ns/op | 221340101286 ns/op | 1.98x |
| Memory | 21421.24kB | 26226.31kB | 0.82x |

---

> The alternative would be to add lots of methods such as to inform the binding "I updated item 5 in that list" or "field 'foo' of my struct was changed" if we want to allow developers to specify that something small-in-scope has changed.

If the memory usage of this PR, and the cpu usage of the control are unsatisfactory then this maybe the only remaining avenue and when I explored it with the code below, the performance was indeed the best (2568x faster than this PR).

```
$ go test -bench=BenchmarkBinding -benchmem -memprofile extra.mem.out -cpuprofile extra.cpu.out -timeout 1h
goos: darwin
goarch: amd64
pkg: github.com/stuartmscott/bindingbench
BenchmarkBinding-4   	      15	  86184948 ns/op	32700947 B/op	  499970 allocs/op
PASS
ok  	github.com/stuartmscott/bindingbench	1.898s
```

andydotxyz:
The locking belongs in the main binding code - there was a TODO 3 commits back that was missed. The code introduced by this PR does not increase the chance of this being an issue, I have opened #1719 to cover this.

It is a little unfortunate that so much code and benchmarking was pasted into this issue, when as you say it is an unfair representation.

I cannot see how to respond to them all, not knowing which are still relevant. I guess the crux of the matter is whether or not we want to be smart about which triggers we fire. I will attempt to create some simple benchmarks and then extend them as you have done to indicate string formatting etc.

andydotxyz:
Apologies for the long post, TL;DR there seem to be misunderstandings of how `Reload` is intended to be used, and indeed how to test performance of bindings that spin on a background thread. If we actually read the data in a callback (otherwise trigger is almost no-op, as it does not actually read the data) and then also *wait* for binding to process then this PR is at worst no better than "control" and in a slightly chained more real-world usage it may be twice as fast.

---

It is possible that you have misunderstood how this API is intended to be used, allow me to illustrate:

```go
func BenchmarkBinding(b *testing.B) {
	count := 100000

	// Create a slice full of 0.0s
	data := make([]float64, count)

	// Create binding from slice
	bind := binding.BindFloatList(&data)

	// Update value in slice
	data[5] = 2.4

	// Notify binding
	bind.Reload()

	// and then wait until all the propagation is complete for a realistic answer
	binding.WaitForItems()
}
```

And the results of this run are:

```
$ go test -bench=BenchmarkBinding
BenchmarkBinding-8   	1000000000	         0.00893 ns/op
PASS
ok  	fyne.io/fyne/cmd/hello	1.259s
```

If we edit one element (or an unknown element), and then call `Reload` as intended that is the outcome. It's not instant, but it is fast. Locking would be done on a single entry point and so will not cause any real change in the numbers. As in the TL;DR above, I note here that there is no real calculations or data read going on, this is just the bare bones, absolute minimum trigger code which is not likely useful but is a baseline to refute the allegations of slowness in previous posts.

Please see the `binding.WaitForItems()` - for this text I exported that internal method so that we could see how long it took for the binding to _take effect_, otherwise the tests are simply seeing how fast they can queue goroutines before exiting.

Note that I have removed the "b.N *" value - it's purpose is to vary as the benchmark runs to "flex" the code. In this instance we are interested only in the 100'000 long slice - when I used the same multiplication as your examples it was heading north of 5 million instead, so the numbers were rather confusing.

Next we will compare this to the illustration that @stuartmscott was making, the code above remains the same but we remove logic from the float list `doReload` so that it blindly fires all listeners:

```go
func (l *boundFloatList) doReload() (retErr error) {
	//oldLen := len(l.items)
	//newLen := len(*l.val)
	//if oldLen > newLen {
	//	for i := oldLen - 1; i >= newLen; i-- {
	//		l.deleteItem(i)
	//	}
	//	l.trigger()
	//} else if oldLen < newLen {
	//	for i := oldLen; i < newLen; i++ {
	//		l.appendItem(bindFloatListItem(l.val, i))
	//	}
	//	l.trigger()
	//}

	for _, item := range l.items {
		//if i > oldLen || i > newLen {
		//	break
		//}

		//err := item.(*boundFloatListItem).setIfChanged((*l.val)[i])
		//if err != nil {
		//	retErr = err
		//}
		item.(*boundFloatListItem).trigger()
	}
	return
}
```

Running again creates the following output:

```
$ go test -bench=BenchmarkBinding
BenchmarkBinding-8   	1000000000	         0.00870 ns/op
PASS
ok  	fyne.io/fyne/cmd/hello	1.339s
```

And so we can see that in the situation whe are making minimal changes and have no load down the chain the proposed PR equivalent in speed - **not** ~3163x slower.

Next we will expand the example to do a string conversion as you suggested. Rolling back the change made so that we have the binding code as-per this PR, I then update the example test to be as follows:

```go
package main

import (
	"fmt"
	"testing"
	"time"

	"fyne.io/fyne/data/binding"
)

const formatA = "%f: Lorem ipsum dolor sit amet, consectetur adipiscing elit."

func BenchmarkBinding(b *testing.B) {
	count := 100000

	// Create a slice full of 0.0s
	data := make([]float64, count)

	// Create binding from slice
	bind := binding.BindFloatList(&data)

	// Format A
	for i := 0; i < count; i++ {
		item, _ := bind.GetItem(i)
		str := binding.FloatToStringWithFormat(item.(binding.Float), formatA)

		str.AddListener(binding.NewDataListener(func() {
			_, _ = str.Get()
			// here we would consume the string. This is simulating a label
		}))
	}

	t := time.Now()
	// Update value in slice
	data[5] = 2.4

	// Notify binding
	bind.Reload()

	// and then wait until all the propagation is complete for a realistic answer
	binding.WaitForItems()

	fmt.Println("Reload took:", time.Since(t))
}
```

The benchmark is now setting up the binding, attaching the string conversion to every node and consuming the result (I did not want to add a `fmt.Println` or anything else that could skew the results).

Bbecause of the way we added "Notify on connection" some of this code will be running twice as it gets the 0 value before we start `Reload`. For this reason I added the Println at the end so that we see the result of the Reload run, not the initial setup code.

```
$ go test -bench=BenchmarkBinding
BenchmarkBinding-8   	Reload took: 23.845457ms
Reload took: 17.531423ms
Reload took: 17.460288ms
Reload took: 18.897462ms
Reload took: 15.963794ms
Reload took: 15.59572ms
Reload took: 16.121742ms
Reload took: 16.939664ms
1000000000	         0.0798 ns/op
PASS
ok  	fyne.io/fyne/cmd/hello	2.337s
```

And then with the same change as before, removing all of the trigger-on-change logic from float list, we get:

```
BenchmarkBinding-8   	Reload took: 45.627102ms
Reload took: 34.897061ms
Reload took: 36.229652ms
Reload took: 36.671954ms
Reload took: 36.53713ms
Reload took: 37.031529ms
Reload took: 39.594245ms
Reload took: 40.053061ms
Reload took: 38.139808ms
1000000000	         0.0990 ns/op
PASS
ok  	fyne.io/fyne/cmd/hello	2.988s
```

From these measurements we see that without the logic introduced in this PR the triggering takes _twice the time_.
This was of course a fairly basic example. Imagine next what the imact would be if there were more complex downstream binding conversions or validations... I hope this illustrates that it is better not to call triggers on data that has not changed.
Or at least it will have dispelled the myth that this PR is hugely much slower than the naieve case.

I sincerely hope that the numbers used in previous messages on this PR were the result of a misunderstanding. Otherwise it feels rather peculiar that you could clasify this as "~3163x slower than the control".

data/binding/gen.go
```diff
@@ -66,6 +80,16 @@ func (b *bound{{ .Name }}) Set(val {{ .Type }}) error {
 	b.trigger()
 	return nil
 }
+
+func (b *bound{{ .Name }}) setIfChanged(val {{ .Type }}) error {
+	if val == b.old {
+		return nil
+	}
+	b.old = val
```
stuartmscott, line 88: A mutex is needed here

data/binding/gen.go
```diff
@@ -297,64 +331,49 @@ type bound{{ .Name }}List struct {
 }
 
 func (l *bound{{ .Name }}List) Append(val {{ .Type }}) error {
-	if l.val != nil {
-		*l.val = append(*l.val, val)
-	}
+	*l.val = append(*l.val, val)
```
stuartmscott, line 334: all access to `val` and `items` needs to be locked, ditto for map bindings.

stuartmscott (changes requested):
Filing #1719 as way to defer what a reviewer has requested is not acceptable in this situation - locking needs to be included here so the performance impacts can be considered before merging.

Your results seems to be similar to mine; the control uses more cpu, this PR uses more memory (although you carefully neglected to report the memory impact in your analysis).

Good call on `binding.WaitForItems()`, I will include that in the next benchmark to be run once this PR has the requested locking

andydotxyz:
> Wouldn't the optimization be better from the collections widget side? Only have listeners for the visible items? Then if you want to change one item you may have to go through 20 to 30 callbacks not 100,001 ... as the updates to any other item would be gotten through the visual changes such as scrolling

I thought more about this @okratitan and I wonder if we need to consider how this could be used outside of the standard widgets as well. I agree that normally large data sets would pair well with collection widgets, but I don't think we can rely on them being the only place where large data from bindings are consumed. If we can avoid unneccessary conversions / triggers etc this seems desirable irrespective of how smart the widgets it is attached to could be.

andydotxyz:
Running again with locking added. Including "-benchmem", apologies for the oversight.
Back to the the code as exists in this PR, with the simple example:

```go
package main

import (
	"testing"

	"fyne.io/fyne/data/binding"
)

func BenchmarkBinding(b *testing.B) {
	count := 100000

	// Create a slice full of 0.0s
	data := make([]float64, count)

	// Create binding from slice
	bind := binding.BindFloatList(&data)

	// Update value in slice
	data[5] = 2.4

	// Notify binding
	bind.Reload()

	// and then wait until all the propagation is complete for a realistic answer
	binding.WaitForItems()
}
```

Running:
```
$ go test -bench=BenchmarkBinding -benchmem
BenchmarkBinding-8   	1000000000	         0.0137 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	fyne.io/fyne/cmd/hello	0.333s
```

And then running again with the modifications to Float List binding. This time we need to add appropriate locking around the call into b.trigger, as that is a private call so elsewhere is already within a lock. So this time the code change is like this:

```go

```
func (l *boundFloatList) doReload() (retErr error) {
	//oldLen := len(l.items)
	//newLen := len(*l.val)
	//if oldLen > newLen {
	//	for i := oldLen - 1; i >= newLen; i-- {
	//		l.deleteItem(i)
	//	}
	//	l.trigger()
	//} else if oldLen < newLen {
	//	for i := oldLen; i < newLen; i++ {
	//		l.appendItem(bindFloatListItem(l.val, i))
	//	}
	//	l.trigger()
	//}

	for _, item := range l.items {
		//if i > oldLen || i > newLen {
		//	break
		//}

		//err := item.(*boundFloatListItem).setIfChanged((*l.val)[i])
		//if err != nil {
		//	retErr = err
		//}
		item.(*boundFloatListItem).lock.Lock()
		item.(*boundFloatListItem).trigger()
		item.(*boundFloatListItem).lock.Unlock()
	}
	return
}
```
$ go test -bench=BenchmarkBinding -benchmem
BenchmarkBinding-8   	1000000000	         0.0147 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	fyne.io/fyne/cmd/hello	0.268s
```

Then we return to the updated example that adds the formatter string binding (as a reminder, the code is as follows)...

```go
package main

import (
	"fmt"
	"testing"
	"time"

	"fyne.io/fyne/data/binding"
)

const formatA = "%f: Lorem ipsum dolor sit amet, consectetur adipiscing elit."

func BenchmarkBinding(b *testing.B) {
	count := 100000

	// Create a slice full of 0.0s
	data := make([]float64, count)

	// Create binding from slice
	bind := binding.BindFloatList(&data)

	// Format A
	for i := 0; i < count; i++ {
		item, _ := bind.GetItem(i)
		str := binding.FloatToStringWithFormat(item.(binding.Float), formatA)

		str.AddListener(binding.NewDataListener(func() {
			_, _ = str.Get()
			// here we would consume the string. This is simulating a label
		}))
	}

	t := time.Now()
	// Update value in slice
	data[5] = 2.4

	// Notify binding
	bind.Reload()

	// and then wait until all the propagation is complete for a realistic answer
	binding.WaitForItems()

	fmt.Println("Reload took:", time.Since(t))
}
```

Again with the code in this PR, subject to the locking addition:

```
$ go test -bench=BenchmarkBinding -benchmem
BenchmarkBinding-8   	Reload took: 30.381702ms
Reload took: 26.609376ms
Reload took: 30.71734ms
Reload took: 24.176746ms
Reload took: 25.598455ms
Reload took: 28.468021ms
Reload took: 26.597492ms
Reload took: 24.725493ms
Reload took: 27.47986ms
1000000000	         0.0931 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	fyne.io/fyne/cmd/hello	1.205s
```

And again without the logic to minimise triggers on unchanged data:

```
$ go test -bench=BenchmarkBinding -benchmem
BenchmarkBinding-8   	Reload took: 55.247078ms
Reload took: 54.384607ms
Reload took: 55.754413ms
Reload took: 57.00343ms
Reload took: 58.055836ms
Reload took: 57.799591ms
Reload took: 58.96864ms
Reload took: 59.91763ms
Reload took: 59.279976ms
Reload took: 58.960851ms
1000000000	         0.121 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	fyne.io/fyne/cmd/hello	1.657s
```

What I think the above illustrates is that there is not much of a difference. In fact the existance of more locking has impacted the "trigger everything" version possibly due to the fact that `trigger()` locks on the way in.

P.S. sorry it looks like the benchmem is not working above, I will look into why and update this accordingly...

data/binding/gen.go
andydotxyz: fixed

andydotxyz:
Corrrecting the locking was a rather larger change than perhaps this PR implied - but if it is required to do a benchmark then fair enough.
The changes are now in and close #1719.

stuartmscott:
> P.S. sorry it looks like the benchmem is not working above, I will look into why and update this accordingly...

Try  `-benchmem -memprofile mem.out` then you can use `go tool pprof mem.out` to analyze it with the `top` command. Remember to divide the total by the number of times the benchmark was run - this has caught me out before :)

andydotxyz:
Thanks @stuartmscott that worked. I have not done any analysis or converting of the number, just dumped them from "pprof top" and pasted the data inline.
As you can see by far the biggest consumer of memory is the string conversion, which results in an almost 50% increase if we call trigger on all the nodes compared to the data required to avoid performing such calls.

stuartmscott:
For a fair memory usage comparison, the `old {{ .Type }}` must be moved from the `bound{{ .Name }}` and `bound{{ .Name }}ListItem` structs in the Control.

Jacalz:
> As you can see by far the biggest consumer of memory is the string conversion, which results in an almost 50% increase if we call trigger on all the nodes compared to the data required to avoid performing such calls.

I think I might have a solution to that. We should avoid `fmt.Sprintf()` and all other related functions within that packages if we can and instead use those from the `strconv` package. They are much faster and better at what they are doing. That should make conversions much faster. Side note, Go 1.16 also features a new string to float algorithm that should be around twice as fast (for anyone interested) :)

https://groups.google.com/d/topic/golang-nuts/pN1AtedMX3g

andydotxyz:
OK, updated the primitive types to no longer duplucate. This caused a slight change in that calling `Set` or `Reload` on a primitive will always trigger. That seemed  a reasonable compromise because if you are calling in at that level you probably know what's happenning. The caching and change detection is now only in list and map, which makes sense as this is where the large data and therefore the need to reduce callbacks, occurs.

Updating all the test runs above and I see:

PR, as is, and a single float value change in 100000 elements:
```
$ go test -bench=BenchmarkBinding -benchmem -memprofile mem.out
BenchmarkBinding-8   	1000000000	         0.0115 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	fyne.io/fyne/cmd/hello	0.298s
$ (pprof) top
Showing nodes accounting for 115.96MB, 99.57% of 116.46MB total
Dropped 1 node (cum <= 0.58MB)
      flat  flat%   sum%        cum   cum%
   56.26MB 48.31% 48.31%    56.26MB 48.31%  fyne.io/fyne/data/binding.(*listBase).appendItem (inline)
   41.50MB 35.64% 83.94%    41.50MB 35.64%  fyne.io/fyne/data/binding.bindFloatListItem (inline)
   15.27MB 13.11% 97.05%    15.27MB 13.11%  fyne.io/fyne/data/binding.init
    2.93MB  2.52% 99.57%   100.69MB 86.46%  fyne.io/fyne/cmd/hello.BenchmarkBinding
         0     0% 99.57%    97.76MB 83.94%  fyne.io/fyne/data/binding.BindFloatList
         0     0% 99.57%    15.77MB 13.54%  runtime.doInit
         0     0% 99.57%    15.77MB 13.54%  runtime.main
         0     0% 99.57%    92.01MB 79.01%  testing.(*B).launch
         0     0% 99.57%     8.68MB  7.45%  testing.(*B).run1.func1
         0     0% 99.57%   100.69MB 86.46%  testing.(*B).runN
```

compared to the trigger-everything version:

```
$ go test -bench=BenchmarkBinding -benchmem -memprofile mem.out
BenchmarkBinding-8   	1000000000	         0.0133 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	fyne.io/fyne/cmd/hello	0.395s
$ (pprof) top
Showing nodes accounting for 122.79MB, 100% of 122.79MB total
      flat  flat%   sum%        cum   cum%
   59.13MB 48.16% 48.16%    59.13MB 48.16%  fyne.io/fyne/data/binding.(*listBase).appendItem (inline)
   43.50MB 35.43% 83.59%    43.50MB 35.43%  fyne.io/fyne/data/binding.bindFloatListItem (inline)
   15.27MB 12.43% 96.02%    15.27MB 12.43%  fyne.io/fyne/data/binding.init
    4.88MB  3.98%   100%   107.52MB 87.57%  fyne.io/fyne/cmd/hello.BenchmarkBinding
         0     0%   100%   102.64MB 83.59%  fyne.io/fyne/data/binding.BindFloatList
         0     0%   100%    15.27MB 12.43%  runtime.doInit
         0     0%   100%    15.27MB 12.43%  runtime.main
         0     0%   100%    88.70MB 72.24%  testing.(*B).launch
         0     0%   100%    18.82MB 15.33%  testing.(*B).run1.func1
         0     0%   100%   107.52MB 87.57%  testing.(*B).runN
```

Then we do the second test that adds string conversion to each element, just as an example of possible chaining of non-trivial data translation.

```
$ go test -bench=BenchmarkBinding -benchmem -memprofile mem.out
BenchmarkBinding-8   	Reload took: 35.753081ms
Reload took: 27.098362ms
Reload took: 24.416182ms
Reload took: 27.273059ms
Reload took: 20.351447ms
Reload took: 25.439767ms
Reload took: 29.669641ms
Reload took: 24.867241ms
Reload took: 27.782294ms
1000000000	         0.0902 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	fyne.io/fyne/cmd/hello	1.268s
$ (pprof) top
Showing nodes accounting for 569.38MB, 100% of 569.38MB total
Showing top 10 nodes out of 20
      flat  flat%   sum%        cum   cum%
  158.51MB 27.84% 27.84%   158.51MB 27.84%  fmt.Sprintf
      99MB 17.39% 45.23%       99MB 17.39%  fyne.io/fyne/data/binding.(*base).AddListener
   82.29MB 14.45% 59.68%    82.29MB 14.45%  fyne.io/fyne/data/binding.(*listBase).appendItem (inline)
   69.51MB 12.21% 71.89%   117.51MB 20.64%  fyne.io/fyne/data/binding.FloatToStringWithFormat (inline)
   67.01MB 11.77% 83.66%    67.01MB 11.77%  fyne.io/fyne/data/binding.bindFloatListItem (inline)
   45.29MB  7.95% 91.61%   371.10MB 65.18%  fyne.io/fyne/cmd/hello.BenchmarkBinding
   24.50MB  4.30% 95.91%    24.50MB  4.30%  fyne.io/fyne/data/binding.(*base).trigger
   15.27MB  2.68% 98.59%    15.27MB  2.68%  fyne.io/fyne/data/binding.init
       8MB  1.41%   100%        8MB  1.41%  fyne.io/fyne/data/binding.NewDataListener (inline)
         0     0%   100%   158.51MB 27.84%  fyne.io/fyne/cmd/hello.BenchmarkBinding.func1
```

and without the list/map logic:

```
$ go test -bench=BenchmarkBinding -benchmem -memprofile mem.out
BenchmarkBinding-8   	Reload took: 50.637444ms
Reload took: 57.532066ms
Reload took: 62.812049ms
Reload took: 58.053015ms
Reload took: 60.400339ms
Reload took: 62.876917ms
Reload took: 64.469241ms
Reload took: 60.992411ms
Reload took: 61.872191ms
Reload took: 59.648361ms
1000000000	         0.127 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	fyne.io/fyne/cmd/hello	1.803s
$ (pprof) top
Showing nodes accounting for 759.87MB, 100% of 759.87MB total
Showing top 10 nodes out of 22
      flat  flat%   sum%        cum   cum%
  228.52MB 30.07% 30.07%   228.52MB 30.07%  fmt.Sprintf
   99.28MB 13.07% 43.14%    99.28MB 13.07%  fyne.io/fyne/data/binding.(*listBase).appendItem (inline)
   95.01MB 12.50% 55.64%    95.01MB 12.50%  fyne.io/fyne/data/binding.bindFloatListItem (inline)
   92.50MB 12.17% 67.81%    92.50MB 12.17%  fyne.io/fyne/data/binding.(*base).AddListener
   92.01MB 12.11% 79.92%   143.51MB 18.89%  fyne.io/fyne/data/binding.FloatToStringWithFormat (inline)
      92MB 12.11% 92.03%       92MB 12.11%  fyne.io/fyne/data/binding.(*base).trigger
   38.79MB  5.11% 97.14%   455.09MB 59.89%  fyne.io/fyne/cmd/hello.BenchmarkBinding
   15.27MB  2.01% 99.14%    15.27MB  2.01%  fyne.io/fyne/data/binding.init
    6.50MB  0.86%   100%     6.50MB  0.86%  fyne.io/fyne/data/binding.NewDataListener (inline)
         0     0%   100%   228.52MB 30.07%  fyne.io/fyne/cmd/hello.BenchmarkBinding.func1
```

And after all of those tests we can see... Not a whole lot of difference. But at least the small cases are now optimal, and the big ones still performant.

andydotxyz:
> I think I might have a solution to that. We should avoid `fmt.Sprintf()` and all other related functions within that packages if we can and instead use those from the `strconv` package.

This is a good point. Perhaps our converters can be smarter to use the `strconv` where possible. I don't think it would help in this example as we are formatting a larger string.

stuartmscott:
Weird, when I fetch this PR, export `binding.WaitForItems()`, and run the most recently posted string formatting benchmark, it times out after 11 minutes;

```
$ git status
HEAD detached at andydotxyz/fix/1645
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   data/binding/queue.go

no changes added to commit (use "git add" and/or "git commit -a")

$ git diff
diff --git a/data/binding/queue.go b/data/binding/queue.go
index 7587a889..177dc189 100644
--- a/data/binding/queue.go
+++ b/data/binding/queue.go
@@ -27,7 +27,7 @@ func processItems() {
        }
 }
 
-func waitForItems() {
+func WaitForItems() {
        done := make(chan interface{})
        itemQueue <- itemData{done: done}
        <-done

$ go test -bench=BenchmarkBinding -benchmem -memprofile control.mem.out -cpuprofile control.cpu.out
...
*** Test killed with quit: ran too long (11m0s).
```

Can anyone else run it?

Jacalz:
> This is a good point. Perhaps our converters can be smarter to use the `strconv` where possible. I don't think it would help in this example as we are formatting a larger string.

Sure. Want me to file an issue for that?

andydotxyz:
@stuartmscott did you up the queue length like I suggested?

> P.P.S. with the wait in place you may also need to increase the size of the binding queue. We should probably revisit this before release. In data/binding/queue.go:
> ```
> var itemQueue = make(chan itemData, 1024)
> ```

stuartmscott:
> (add a few zeros)

Ah yes that was the issue, thanks

andydotxyz:
> Sure. Want me to file an issue for that?

Totally, thanks.
Or a PR ;)

andydotxyz:
I am pretty happy with this PR now, thanks for all the feedback. It feels like a good balance of smarts, simple public API and CPU/RAM performance.

As pointed out it could be optimised for memory further (without a CPU hit) by splitting out internal only map and list types from the external binding types, that need to track old values to monitor for changes. This would allow a reduced memory footprint for internal-only large data volumes.
However it will mean duplicating many internal types to make this happen - I think 7 more types, some of which are generated and one not. So I wanted to post here just to flag that it will increase the complexity a reasonable amount to get that outcome. Do we want to?

stuartmscott:
Here are the results I get from the benchmark:

This PR:

```
$ go test -bench=BenchmarkBinding -benchmem -memprofile 1717.mem.out -cpuprofile 1717.cpu.out
Reload took: 26.892184ms
goos: darwin
goarch: amd64
pkg: github.com/stuartmscott/bindingbench
BenchmarkBinding-4   	Reload took: 29.440679ms
Reload took: 27.190843ms
Reload took: 52.732777ms
Reload took: 24.358871ms
Reload took: 64.873018ms
Reload took: 70.155573ms
Reload took: 55.417237ms
Reload took: 54.589399ms
Reload took: 97.316033ms
Reload took: 66.788521ms
Reload took: 17.912953ms
Reload took: 91.549516ms
Reload took: 45.925223ms
1000000000	         0.358 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/stuartmscott/bindingbench	5.234s

(pprof) top
Showing nodes accounting for 943.04MB, 99.65% of 946.36MB total
Dropped 26 nodes (cum <= 4.73MB)
Showing top 10 nodes out of 20
      flat  flat%   sum%        cum   cum%
  215.52MB 22.77% 22.77%   215.52MB 22.77%  fmt.Sprintf
  156.25MB 16.51% 39.28%   156.25MB 16.51%  fyne.io/fyne/data/binding.init
  129.99MB 13.74% 53.02%   129.99MB 13.74%  fyne.io/fyne/data/binding.(*listBase).appendItem (inline)
     126MB 13.31% 66.33%      126MB 13.31%  fyne.io/fyne/data/binding.(*base).AddListener
  108.01MB 11.41% 77.75%   177.51MB 18.76%  fyne.io/fyne/data/binding.FloatToStringWithFormat (inline)
   99.01MB 10.46% 88.21%    99.01MB 10.46%  fyne.io/fyne/data/binding.bindFloatListItem (inline)
   60.27MB  6.37% 94.58%   530.78MB 56.09%  github.com/stuartmscott/bindingbench.BenchmarkBinding
   40.50MB  4.28% 98.86%    40.50MB  4.28%  fyne.io/fyne/data/binding.(*base).trigger
    7.50MB  0.79% 99.65%     7.50MB  0.79%  fyne.io/fyne/data/binding.NewDataListener (inline)
         0     0% 99.65%   215.52MB 22.77%  fyne.io/fyne/data/binding.(*listener).DataChanged
```

Control:
```
$ go test -bench=BenchmarkBinding -benchmem -memprofile control.mem.out -cpuprofile control.cpu.out
Reload took: 74.211641ms
goos: darwin
goarch: amd64
pkg: github.com/stuartmscott/bindingbench
BenchmarkBinding-4   	Reload took: 46.544647ms
Reload took: 58.273008ms
Reload took: 58.638638ms
Reload took: 74.651833ms
Reload took: 103.079449ms
Reload took: 108.03206ms
Reload took: 138.534753ms
Reload took: 103.442923ms
Reload took: 125.621081ms
Reload took: 87.525151ms
Reload took: 49.336927ms
Reload took: 59.399604ms
Reload took: 55.051924ms
Reload took: 57.077595ms
1000000000	         0.171 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/stuartmscott/bindingbench	4.849s

(pprof) top
Showing nodes accounting for 1050.19MB, 99.76% of 1052.73MB total
Dropped 13 nodes (cum <= 5.26MB)
Showing top 10 nodes out of 22
      flat  flat%   sum%        cum   cum%
  357.53MB 33.96% 33.96%   357.53MB 33.96%  fmt.Sprintf
  139.50MB 13.25% 47.21%   139.50MB 13.25%  fyne.io/fyne/data/binding.(*base).AddListener
  129.50MB 12.30% 59.52%   129.50MB 12.30%  fyne.io/fyne/data/binding.(*base).trigger
  124.77MB 11.85% 71.37%   124.77MB 11.85%  fyne.io/fyne/data/binding.(*listBase).appendItem (inline)
  116.01MB 11.02% 82.39%   183.01MB 17.38%  fyne.io/fyne/data/binding.FloatToStringWithFormat (inline)
   91.01MB  8.64% 91.03%    91.01MB  8.64%  fyne.io/fyne/data/binding.BindFloat (inline)
   64.25MB  6.10% 97.13%   586.03MB 55.67%  github.com/stuartmscott/bindingbench.BenchmarkBinding
   15.62MB  1.48% 98.62%    15.62MB  1.48%  fyne.io/fyne/data/binding.init
      12MB  1.14% 99.76%       12MB  1.14%  fyne.io/fyne/data/binding.NewDataListener (inline)
         0     0% 99.76%    38.50MB  3.66%  fyne.io/fyne/data/binding.(*base).Reload
```

So Control is ~2x faster (0.171 ns/op vs 0.358 ns/op) but uses 1.1x memory (1052.73MB vs 946.36MB). Which is likely due to the Control triggering the `fmt.Sprintf` unnecessarily.

Good call on switching to strconv @Jacalz, tho we still need to figure out if this is a realistic use case - is that what you would want to use bindings for? (Yes I'm aware I originally presented the benchmark, though it is not how I envision I will use the API.)

There has been talk of binding to a file, or database, or remote server where we wont know what specifically has changed, just that there could be change somewhere.

If we can narrow down the set of situations we want to support, we will be able to create better benchmarks that stress the implementation in a realistic way.

andydotxyz:
> So Control is ~2x faster (0.171 ns/op vs 0.358 ns/op)

That is how fast operations occurring... not how long the process took.
For that purpose I added the "Reload took: 46.544647ms" lines, that measures the cumulative amount that takes into account the number of operations, not just how long each one took.

stuartmscott:
From (https://golang.org/pkg/testing/):

> The output
> `BenchmarkRandInt-8   	68453040	        17.8 ns/op`
> means that the loop ran 68453040 times at a speed of 17.8 ns per loop.

So it is measuring the total time to allocate the slice, create the bindings, setup the listeners, update a value in the slice and call Reload. If we just want to measure the Reload time then we can add `b.ResetTimer()` just before `bind.Reload()`

andydotxyz:
> From (https://golang.org/pkg/testing/):
> 
> > The output
> > `BenchmarkRandInt-8   	68453040	        17.8 ns/op`
> > means that the loop ran 68453040 times at a speed of 17.8 ns per loop.
> 
> So it is measuring the total time to allocate the slice, create the bindings, setup the listeners, update a value in the slice and call Reload. If we just want to measure the Reload time then we can add `b.ResetTimer()` just before `bind.Reload()`

If the code we were testing operated in the current goroutine that would hold.
As most of data binding operates on a background thread the gap between loop iterations and `WaitForItems()` needs to be included in the timing. That is why I added the printouts. You can see the benchmark runs around 10 times, and the print shows how long it took to resolve all the callbacks.

But you don't need to take my word for it. multiply the count up and run the examples against a stop watch. The version that calls all the triggers takes longer, so I don't understand how you can say it is 2x faster?

andydotxyz:
I now have the de-duplication for the internal-only cases where we know that the external data cannot have been changed. The results of the benchmark you have as-is, as a reminder:

```
$ go test -bench=BenchmarkBinding -benchmem -memprofile mem.out
BenchmarkBinding_External_FloatToString-8   	1000000000	         0.244 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	fyne.io/fyne/cmd/hello	12.852s
$ (pprof) top
Showing nodes accounting for 7947.23MB, 100% of 7947.23MB total
Showing top 10 nodes out of 20
      flat  flat%   sum%        cum   cum%
 2085.66MB 26.24% 26.24%  2085.66MB 26.24%  fmt.Sprintf
 1276.53MB 16.06% 42.31%  1276.53MB 16.06%  fyne.io/fyne/data/binding.(*base).AddListener
 1184.53MB 14.90% 57.21%  1184.53MB 14.90%  fyne.io/fyne/data/binding.(*listBase).appendItem (inline)
 1080.58MB 13.60% 70.81%  1080.58MB 13.60%  fyne.io/fyne/data/binding.bindFloatListItem (inline)
 1080.08MB 13.59% 84.40%  1726.10MB 21.72%  fyne.io/fyne/data/binding.FloatToStringWithFormat (inline)
  555.74MB  6.99% 91.39%  5281.47MB 66.46%  fyne.io/fyne/cmd/hello.BenchmarkBinding_External_FloatToString
  427.51MB  5.38% 96.77%   427.51MB  5.38%  fyne.io/fyne/data/binding.(*base).trigger
  152.59MB  1.92% 98.69%   152.59MB  1.92%  fyne.io/fyne/data/binding.init
     104MB  1.31%   100%      104MB  1.31%  fyne.io/fyne/data/binding.NewDataListener (inline)
         0     0%   100%  2085.66MB 26.24%  fyne.io/fyne/cmd/hello.BenchmarkBinding_External_FloatToString.func1
```

and then replacing the `BindFloatList` with `NewFloatList`, as follows (we change how it is constructed, and then how it is updated):
```go
	// Create binding internally and expand it to large data
	bind := binding.NewFloatList()
	bind.Set(data)

...

	// Update value in binding
	item, _ := bind.GetItem(5)
	item.(binding.Float).Set(2.5)
```

Running the benchmark with this change.
```
$ go test -bench=BenchmarkBinding -benchmem -memprofile mem.out
BenchmarkBinding_Internal_FloatToString-8   	1000000000	         0.235 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	fyne.io/fyne/cmd/hello	12.694s
$ (pprof) top
Showing nodes accounting for 7771.06MB, 100% of 7771.06MB total
Showing top 10 nodes out of 21
      flat  flat%   sum%        cum   cum%
 2151.16MB 27.68% 27.68%  2151.16MB 27.68%  fmt.Sprintf
 1260.03MB 16.21% 43.90%  1260.03MB 16.21%  fyne.io/fyne/data/binding.(*base).AddListener
 1183.38MB 15.23% 59.12%  1183.38MB 15.23%  fyne.io/fyne/data/binding.(*listBase).appendItem (inline)
 1038.08MB 13.36% 72.48%  1643.59MB 21.15%  fyne.io/fyne/data/binding.FloatToStringWithFormat (inline)
  881.55MB 11.34% 83.83%   881.55MB 11.34%  fyne.io/fyne/data/binding.bindFloatListItem (inline)
  552.74MB  7.11% 90.94%  5018.79MB 64.58%  fyne.io/fyne/cmd/hello.BenchmarkBinding_Internal_FloatToString
  448.51MB  5.77% 96.71%   448.51MB  5.77%  fyne.io/fyne/data/binding.(*base).trigger
  152.59MB  1.96% 98.67%   152.59MB  1.96%  fyne.io/fyne/data/binding.init
     103MB  1.33%   100%      103MB  1.33%  fyne.io/fyne/data/binding.NewDataListener (inline)
         0     0%   100%  2151.16MB 27.68%  fyne.io/fyne/cmd/hello.BenchmarkBinding_Internal_FloatToString.func1
```

What I think this shows is that the two are almost identical in CPU us age but by using only the internal storage we save around 10% of the overall memory usage. If we look at the "bindFloatListItem" that is where the saving is, as expected. We have saved 20% on that memory structure.
Which I suppose summarises this approach - provide the ability to avoid duplicate listener callbacks at the cost of some more memory usage. Adding with the data illustrated when comparing with calling all listeners what we have saved is up to 50% of the CPU usage for a 10% increase in the RAM used to store the data. Compared to the application RAM usage this increase will barely be noticed, but the CPU delay could be seen by end users.

I am happy with this tradeoff and think that it is an efficient implementation of the "calling `Reload()` proagates events for changed data" design approach.

data/binding/gen.go
```diff
@@ -380,16 +426,57 @@ func (l *bound{{ .Name }}List) SetValue(i int, v {{ .Type }}) error {
...
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	(*b.val)[b.index] = val
+
+	b.trigger()
+	return nil
+}
+
+func (b *bound{{ .Name }}ListItem) setIfChanged(val {{ .Type }}) error {
```
stuartmscott: When this is called by list it is holding the listbinding mutex, but it is not holding the base binding mutex. There is a race condition between `item, err := listBinding.GetItem(i); item.Set(value)` and `listBinding.Set(i, value)`

andydotxyz: Thanks, fixed

andydotxyz:
With all the locking, same runs as above.
BindFloatList:

```
$ go test -bench=BenchmarkBinding -benchmem -memprofile mem.out
BenchmarkBinding_External_FloatToString-8   	1000000000	         0.294 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	fyne.io/fyne/cmd/hello	15.985s
$ (pprof) top
Showing nodes accounting for 9154.65MB, 100% of 9154.65MB total
Showing top 10 nodes out of 20
      flat  flat%   sum%        cum   cum%
 2442.19MB 26.68% 26.68%  2442.19MB 26.68%  fmt.Sprintf
 1491.04MB 16.29% 42.96%  1491.04MB 16.29%  fyne.io/fyne/data/binding.(*base).AddListener
 1346.12MB 14.70% 57.67%  1346.12MB 14.70%  fyne.io/fyne/data/binding.(*listBase).appendItem (inline)
 1238.59MB 13.53% 71.20%  1971.11MB 21.53%  fyne.io/fyne/data/binding.FloatToStringWithFormat (inline)
 1227.09MB 13.40% 84.60%  1227.09MB 13.40%  fyne.io/fyne/data/binding.bindFloatListItem (inline)
  629.51MB  6.88% 91.48%  6066.86MB 66.27%  fyne.io/fyne/cmd/hello.BenchmarkBinding_External_FloatToString
  493.02MB  5.39% 96.86%   493.02MB  5.39%  fyne.io/fyne/data/binding.(*base).trigger
  152.59MB  1.67% 98.53%   152.59MB  1.67%  fyne.io/fyne/data/binding.init
  134.50MB  1.47%   100%   134.50MB  1.47%  fyne.io/fyne/data/binding.NewDataListener (inline)
         0     0%   100%  2442.19MB 26.68%  fyne.io/fyne/cmd/hello.BenchmarkBinding_External_FloatToString.func1
```

and using NewFloatList:

```
$ go test -bench=BenchmarkBinding -benchmem -memprofile mem.out
BenchmarkBinding_External_FloatToString-8   	1000000000	         0.268 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	fyne.io/fyne/cmd/hello	16.952s
$ (pprof) top
Showing nodes accounting for 8768.61MB, 100% of 8768.61MB total
Showing top 10 nodes out of 21
      flat  flat%   sum%        cum   cum%
 2413.68MB 27.53% 27.53%  2413.68MB 27.53%  fmt.Sprintf
 1455.54MB 16.60% 44.13%  1455.54MB 16.60%  fyne.io/fyne/data/binding.(*base).AddListener
 1359.49MB 15.50% 59.63%  1359.49MB 15.50%  fyne.io/fyne/data/binding.(*listBase).appendItem (inline)
 1188.09MB 13.55% 73.18%  1948.61MB 22.22%  fyne.io/fyne/data/binding.FloatToStringWithFormat (inline)
  979.56MB 11.17% 84.35%   979.56MB 11.17%  fyne.io/fyne/data/binding.bindFloatListItem (inline)
  610.14MB  6.96% 91.31%  5712.32MB 65.15%  fyne.io/fyne/cmd/hello.BenchmarkBinding_External_FloatToString
  490.01MB  5.59% 96.90%   490.01MB  5.59%  fyne.io/fyne/data/binding.(*base).trigger
  152.59MB  1.74% 98.64%   152.59MB  1.74%  fyne.io/fyne/data/binding.init
  119.50MB  1.36%   100%   119.50MB  1.36%  fyne.io/fyne/data/binding.NewDataListener (inline)
         0     0%   100%  2413.68MB 27.53%  fyne.io/fyne/cmd/hello.BenchmarkBinding_External_FloatToString.func1
```

So, with the individual element locking we see a small slowdown evenly spread across both of the API usage types.

AlbinoGeek:
@stuartmscott Your last benchmark code on my machine runs as follows on the fix branch:

```
$ /usr/bin/go test -benchmem -run=^$ -bench ^(BenchmarkBinding_External_.*)
BenchmarkBinding_External_FloatToString-8   	1000000000	         0.387 ns/op	       0 B/op	       0 allocs/op
BenchmarkBinding_External_GetItemSet-8      	1000000000	         0.000006 ns/op	       0 B/op	       0 allocs/op
ok  	liquid	25.491s
```

```
CPU: i7-2600k @ 3.084Ghz
RAM: 32GB DDR3-1600 MHz
```

---

Which is the latest benchmark?

andydotxyz:
Sorry, for avoidance of doubt after applying the changes it takes the benchmarks to:

```go
package main

import (
	"testing"

	"fyne.io/fyne/data/binding"
)

const (
	count = 1000000 // 1 Million so the size of data is sufficiently larger than the go runtime
	format = "Lorem ipsum dolor sit amet, %f consectetur adipiscing elit." // Long-ish string with the format in the middle to be more realistic - most string comparisons cannot be determined by looking at the first byte
)

// Benchmarks binding to large data set where each item has a converter to compare the cost of calling all listeners vs comparing the data with an internal copy and only call the listeners of the data that changed.
func BenchmarkBinding_External_FloatToString(b *testing.B) {
	// Create a slice full of 0.0s
	data := make([]float64, count)

	// Create binding from slice
	bind := binding.BindFloatList(&data)

	// Attach formatter to each element
	for i := 0; i < count; i++ {
		item, _ := bind.GetItem(i)
		str := binding.FloatToStringWithFormat(item.(binding.Float), format)

		str.AddListener(binding.NewDataListener(func() {
			_, _ = str.Get()
			// here we would consume the string. This is simulating a label
		}))
	}

	// Update value in slice
	data[5] = 2.4

	// We only want to time how long it takes to reload
	b.ResetTimer()

	// Notify binding
	bind.Reload()

	// and then wait until all the propagation is complete for a realistic answer
	binding.WaitForItems()
}

// Benchmark managing a large data set and updating a single element.
func BenchmarkBinding_Internal_FloatToString(b *testing.B) {
	// Create binding internally and expand it to large data
	bind := binding.NewFloatList()
	bind.Set(make([]float64, count))

	// Attach formatter to each element
	for i := 0; i < count; i++ {
		item, _ := bind.GetItem(i)
		str := binding.FloatToStringWithFormat(item.(binding.Float), format)

		str.AddListener(binding.NewDataListener(func() {
			_, _ = str.Get()
			// here we would consume the string. This is simulating a label
		}))
	}

	// We only want to time how long it takes to reload
	b.ResetTimer()

	// Update value in binding
	item, _ := bind.GetItem(5)
	item.(binding.Float).Set(2.5)

	// and then wait until all the propagation is complete for a realistic answer
	binding.WaitForItems()
}
```

AlbinoGeek:
### External

```
$ go test -bench=BenchmarkBinding_External -benchmem -memprofile memExternal.out
BenchmarkBinding_External_FloatToString-8       1000000000               0.342 ns/op           0 B/op          0 allocs/op
ok      liquid  23.703s

(pprof) top
Showing nodes accounting for 10644.66MB, 100% of 10644.66MB total
Showing top 10 nodes out of 20
      flat  flat%   sum%        cum   cum%
 2800.21MB 26.31% 26.31%  2800.21MB 26.31%  fmt.Sprintf
 1723.04MB 16.19% 42.49%  1723.04MB 16.19%  fyne.io/fyne/data/binding.(*base).AddListener
 1606.37MB 15.09% 57.58%  1606.37MB 15.09%  fyne.io/fyne/data/binding.(*listBase).appendItem (inline)
 1449.11MB 13.61% 71.20%  2333.63MB 21.92%  fyne.io/fyne/data/binding.FloatToStringWithFormat (inline)
 1423.11MB 13.37% 84.57%  1423.11MB 13.37%  fyne.io/fyne/data/binding.bindFloatListItem (inline)
  725.04MB  6.81% 91.38%  7074.68MB 66.46%  liquid.BenchmarkBinding_External_FloatToString
  613.52MB  5.76% 97.14%   613.52MB  5.76%  fyne.io/fyne/data/binding.(*base).trigger
  156.25MB  1.47% 98.61%   156.25MB  1.47%  fyne.io/fyne/data/binding.init
  147.50MB  1.39%   100%   147.50MB  1.39%  fyne.io/fyne/data/binding.NewDataListener (inline)
    0.50MB 0.0047%   100%  3029.98MB 28.46%  fyne.io/fyne/data/binding.BindFloatList
```

### Internal

 ```
$ go test -bench=BenchmarkBinding_Internal -benchmem -memprofile memInternal.out
BenchmarkBinding_Internal_FloatToString-8       1000000000               0.349 ns/op           0 B/op          0 allocs/op
ok      liquid  22.661s

(pprof) top
Showing nodes accounting for 9959.72MB, 100% of 9959.72MB total
Showing top 10 nodes out of 21
      flat  flat%   sum%        cum   cum%
 2782.21MB 27.93% 27.93%  2782.21MB 27.93%  fmt.Sprintf
 1659.54MB 16.66% 44.60%  1659.54MB 16.66%  fyne.io/fyne/data/binding.(*base).AddListener
 1516.61MB 15.23% 59.82%  1516.61MB 15.23%  fyne.io/fyne/data/binding.(*listBase).appendItem (inline)
 1368.60MB 13.74% 73.57%  2203.13MB 22.12%  fyne.io/fyne/data/binding.FloatToStringWithFormat (inline)
 1118.07MB 11.23% 84.79%  1118.07MB 11.23%  fyne.io/fyne/data/binding.bindFloatListItem (inline)
  655.91MB  6.59% 91.38%  6476.24MB 65.02%  liquid.BenchmarkBinding_Internal_FloatToString
  545.02MB  5.47% 96.85%   545.02MB  5.47%  fyne.io/fyne/data/binding.(*base).trigger
  157.50MB  1.58% 98.43%   157.50MB  1.58%  fyne.io/fyne/data/binding.NewDataListener (inline)
  156.25MB  1.57%   100%   156.25MB  1.57%  fyne.io/fyne/data/binding.init
         0     0%   100%  2634.68MB 26.45%  fyne.io/fyne/data/binding.(*boundFloatList).Set
```

### Together

`... still running`

**Comment by andydotxyz** [comment]
Remember @stuartmscott that this supports both. If you want to control the data directly use `NewXxxList/Map` and use the API. If you are open to sync with unknown external changges then use `BindXxxList/Map` I am happy we can support both here and will work to progress the docuementation to reflect both usages.

**Review by AlbinoGeek — APPROVED** [review]
After discussion in slack I feel this is a good enough implementation of minimal updates given the wide variety of use-cases possible with binding, given the option to use `NewXXX` instead where you do not wish for the memory overhead we discovered (as high as 19% and as low as 5% in the last benchmarks.)

I think a separate PR should address a tutorial or guide (documentation) in regards to handling large data sets, just as you had suggested in the discussion.

File: data/binding/mapbinding.go [hunk 4]
```diff
@@ -236,27 +272,97 @@ func (b *mapBase) setItem(key string, d DataItem) {
...
 }
 
+type boundExternalMapValue struct {
+	boundMapValue
+
+	old interface{}
+}
+
+func (b *boundExternalMapValue) setIfChanged(val interface{}) error {
+	if val == b.old {
```
charlesdaniels, line 358: Is this guaranteed to be safe for any possible `interface{}`? My understanding is that an `interface{}` could include pointers to things, in which case simply comparing the pointer values may not be desired. We may want to use a "deep equals" here - there are libraries that can do this.

charlesdaniels (approved):
I know this PR has been the point of some contention. @andydotxyz requested me to look at it so we can get it merged.

I will not claim to be well informed on either the controversy around this PR, nor with the underlying bindings infrastructure. My ability to comment on the quality of the approach presented here as opposed to any other is limited, so I will make no claim on this front.

Aside from the one noted nit, I will vouch that the code I see presented here appears to follow proper style conventions, is reasonably well structured, and does not exhibit any obviously dangerous or incorrect behaviors.

I do not want to stir the pot here, just trying to keep the process moving along

data/binding/mapbinding.go
andydotxyz: That is a good point. Do we want to support that sort of data though?
If it turns out to be useful we could also check for pointer types and do clever things - ot just call without trying to do expensive deep check...

PR #1728, opened 2021-01-06
Resource bundler: Go-idiomatic comment for generated code

joonas-fi:
### Description:

There's a "mini-spec" here: https://golang.org/s/generatedcode (also here: https://golang.org/cmd/go/#hdr-Generate_Go_files_by_processing_source )

The pattern is agreed on so that automatic tooling can recognize generated code files.

The `'$ fyne bundle'` part is not required, but I thought it'd be easier for fresh eyes to understand.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
Unfortunately that is what we had but it blew up some static analysis tools - so we moved to this machine readable version.

andydotxyz:
Would it work if we included both like: https://github.com/fyne-io/fyne/blob/f58de04a068aa8fafad55b831d8d575f46c660dd/theme/gen.go#L17

Jacalz:
Fyi, this PR should ideally target `develop` branch due to the changes done there.

AlbinoGeek:
This may break the documentation on:
https://github.com/fyne-io/fyne/wiki/GitHooks

andydotxyz:
I wonder if goreportcard have updated their systems to support both?

Jacalz:
> I wonder if goreportcard have updated their systems to support both?

I highly doubt it. The upstream project looks to be more or less dead. Last time I looked, there were issues pointing to that their servers are almost full and no new repos can be added (existing ones work though).

andydotxyz:
Maybe we need an alternative then we can move to the "mini standard" and fix our scripts to match?

Jacalz:
Just double checked by creating a new report card and it seems like they have gotten their stuff sorted again. It is not actively maintained by any means, but it works at least. I also checked and it looks like they don't support this way of indicated autogenerated code.

Jacalz:
I am debating on if this is something we want to add or not. We already have multiple comments for this and we don't have the slightest idea if there are anything out there that uses this tag specifically.

AlbinoGeek:
I don't agree that logic around the lines of: "It's not widely used"

Is valid reason to avoid using a standard/idiomatic feature/syntax.
The decision only serves to perpetuate the "evidence"/reasoning.

Jacalz:
Yeah, I was certainly wrong. I think we should replace the `// **** THIS FILE IS AUTO-GENERATED, PLEASE DO NOT EDIT IT **** //`` line with this one. However, this PR still needs to be updated and rebased on `develop` branch to be able to be merged.
@joonas-fi  Do you think that you could do that? :)

joonas-fi:
I'm not good with Git but I think I managed to rebase it to develop. If I understood correctly, for goreportcard there still needs to be the `// auto-generated` comment? I put it as the second line, so the goreportcard's one is "secondary" and hopefully it can be removed if goreportcard starts to support the idiomatic one.

Jacalz:
Thanks. I think you'd have to swap them around. If I remember correctly, the goreportcard one needs to be at the top of the file. The one from Go states that it can be anywhere in the file.

joonas-fi:
JFC :facepalm: :smile: 

Fixed, great catch!

andydotxyz (approved):

Jacalz (approved):

andydotxyz:
Thanks for getting this all sorted. I will re-run our bundlers with this change.

Jacalz:
Thanks @joonas-fi. Congratulations on your first PR in fyne 🎉🍾

PR #1729, opened 2021-01-06
Add animation repeat count

### LINKED ISSUE #1578: Add support for limited repeating in animations
We should change `Animation.Repeat` to `Animation.RepeatCount` - this will count down as items are repeated.
A (default) value of `0` means no repeat.
A `RepeatInfinite` value will mark this as repeating forever (until Stopped or  hidden) - probably a `-1` const.

andydotxyz:
Moves `Repeat bool` to `RepeatCount int`. Adds special `AnimationRepeatForever` for un-ending repeat.

If an animation has `Reverse` set to `true` then the repeat count will be decremented after the reverse is complete.

Fixes #1578

### Checklist:

- [ ] Tests included. <- I have included it in fyne_demo, the logic is in the internal runner and we have not figured out how to unit test that yet.
- [x] Tests all pass.

### DIFF
--- a/animation.go
+++ b/animation.go
@@ -8,6 +8,9 @@ import "time"
 // A linear animation would return the same output value as is passed in.
 type AnimationCurve func(float32) float32
 
+// AnimationRepeatForever is an AnimationCount value that indicates it should not stop looping.
+const AnimationRepeatForever = -1
+
 var (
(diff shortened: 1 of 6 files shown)


AlbinoGeek (approved):
Other than the question of testing, I like what I see.

Jacalz (approved):
Looks good to me. Good job 🙂

PR #1723, opened 2021-01-06
Add OnSubmitted to Entry

stuartmscott:
Fixes #1683

### Checklist:
- [x] Tests included.
- [x] Lint and formatter run with no errors.
- [x] Tests all pass.

#### Where applicable:
- [x] Public APIs match existing style.

PR #1730, opened 2021-01-06
Make setupSelection release shift

stuartmscott:
### Description:
Previously the setupSelection helper in the widget.Entry tests would not release the shift key and expect tests that don't need it to release it, this change flips it so Shift is more opt-in than opt-out

Fixes no issue, but relevant to #1723

### Checklist:
- [x] Tests all pass.

### DIFF
--- a/widget/entry_test.go
+++ b/widget/entry_test.go
@@ -648,7 +648,7 @@ func TestEntry_MultilineSelect(t *testing.T) {
 	c := window.Canvas()
 
 	// Extend the selection down one row
-	typeKeys(e, fyne.KeyDown)
+	typeKeys(e, keyShiftLeftDown, fyne.KeyDown)
 	test.AssertRendersToMarkup(t, `
 		<canvas padded size="150x200">
 			<content>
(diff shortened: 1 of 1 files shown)


andydotxyz:
This seems right, as the shift key down was never released.
As you said in the original PR it would be good to get @toaster opinion as well.

andydotxyz:
One thought... rather than all the shift key down in 'TestEntry_Select' would it be cleaner if the runner function issued the key down (and probably an up after as well?).

Jacalz (approved):
This looks good to me. The suggestion from Andy might be a cleaner solution, but this works in the meantime if we want this in before the 2.0 feature freeze.

stuartmscott:
The `TestEntry_Select` runner could type `keyShiftLeftDown` before, and `keyShiftLeftUp` afterwards, but this would reintroduce the `keyShiftLeftUp` in the `enter` and `enter_reverse` tests that was objected to in #1723.

If it is all the same to you @andydotxyz and @Jacalz I'll land this PR as is.

Jacalz:
It’s all the same to me. I still think this PR is fine merging as is.

andydotxyz:
Fine by me too

andydotxyz (approved):

PR #1734, opened 2021-01-07
Add scrolling to Entry

### LINKED ISSUE #471: Scrollbars for multiline input(entry) widget

andydotxyz:
Single line scrolls horizontally
Multiline scrolls whichever is required

This is drafft until I figure out why right click stopped working.

Fixes #471

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


CHANGELOG.md
```diff
@@ -25,12 +25,12 @@ These changes likely break some apps, please read the
 * The second parameter of `theme.NewThemedResource` was removed, it was previously ignored
 * The desktop.Cursor definition was renamed desktop.StandardCursor to make way for custom cursors
 
+* iOS apps preferences will be lost in this upgrade as we move to more advanced storage
```
stuartmscott, line 28: Is this relevant to this PR?

CHANGELOG.md
```diff
@@ -25,12 +25,12 @@ These changes likely break some apps, please read the
...
 * Dialogs no longer show when created, unless using the ShowXxx convenience methods
+* Entry widget now contains scrolling so should no longer be wrapped in a scroll container
+
 * Removed deprecated types including:
   - dialog.FileIcon (now widget.FileIcon)
-  - canvas.SetScale()
-  - device.SystemScale()
-* iOS apps preferences will be lost in this upgrade as we move to more advanced storage
```
stuartmscott, line 33: Perhaps this change should be its own PR, so we keep this one focused

dialog/color_picker.go
```diff
@@ -186,7 +186,7 @@ func (p *colorAdvancedPicker) CreateRenderer() fyne.WidgetRenderer {
 					preview,
 				),
 			),
-			fyne.NewContainerWithLayout(layout.NewCenterLayout(), hex),
+			hex,
```
stuartmscott, line 189: I was unsure about this change, but when I looked at the golden images it made sense and looked good (I would not have been able to make this judgement from reading an XML string).

Perhaps the entry, and the colour channel entries, should become right aligned so they all match up?

internal/driver/gomobile/canvas.go
```diff
@@ -337,7 +337,7 @@ func (c *mobileCanvas) tapDown(pos fyne.Position, tapID int) {
 
 	co, objPos, layer := c.findObjectAtPositionMatching(pos, func(object fyne.CanvasObject) bool {
 		switch object.(type) {
-		case fyne.Tappable, mobile.Touchable, fyne.Focusable:
+		case mobile.Touchable, fyne.Focusable:
```
stuartmscott, line 340: What is the rationale behind this change? On the face of it, tapDown should be searching for a Tappable object, what am I missing?

stuartmscott (commented):
I just did a quick review, will have a deeper dive later time permitting.

Overall the approach of moving widget functionality into entryContent so Entry can wrap it in a scroller looks good and a step in the direction of collection-ifying (definitely made that word up) it as List, Table, and Tree all follow this pattern.

CHANGELOG.md
andydotxyz: I was updating the changelog and realised it would get too long if we listed every change, so removed the methods to leave just actual type removal.
So in a way it's related, yes.

CHANGELOG.md
andydotxyz: as above

dialog/color_picker.go
andydotxyz: A separate PR to remove the center layout? I think that's going a bit far.
This change causes the entry to shrink further than it did before, so the field did not look good with the defined layout.

internal/driver/gomobile/canvas.go
andydotxyz: I can see the confusion here yes. `Tappable` happens on tapUp, the tapDown does not trigger those events, it's inclusion in this list was incorrect because it could find a tappable object above a focusable object, and then not action any event as a result. And so focusable would not get the event it should because it thought tappable overrode.

CHANGELOG.md
andydotxyz: I have extracted the changed not strictly caused by this PR. Should be cleaner now.

andydotxyz:
> the approach of moving widget functionality into entryContent so Entry can wrap it in a scroller looks good and a step in the direction

Thanks. The changes were less than it might appear from the overall diff. A lot of methods were mostly just moved (with slightly different scope to access the object properties). To make it easier to review I tried to split up the commits so first was the code change and second was the move within the file. Otherwise it is hard to see what changed.

andydotxyz:
> in the direction of collection-ifying

Yeah. With this breaking change done the internal work to make it happen should be possible in upcoming releases much more easily (and I forsee that work taking some time!).

dialog/color_picker.go
stuartmscott: I didn't request a separate PR to remove the center layout, I had wondered about it, and guessed that either approach would break the test image. As I said, this change made sense and looked good, so it can stay. The comment was more thinking out aloud, sorry if that wasn't clear.

internal/driver/gomobile/canvas.go
stuartmscott: Makes sense, thanks for the explanation

stuartmscott (commented):
Overall I think this is a good approach, though I am concerned this changes the public API without following the appropriate deprecation path

andydotxyz:
Yes I guess that's true, I will replace the ones that were removed as wrappers

stuartmscott (approved):

PR #1738, opened 2021-01-08
Preferences improvements, fix locking and watch file

### LINKED ISSUE #1718: fatal error: concurrent map iteration and map write
### Describe the bug:
When updating to the latest develop version, I noticed that dragging the slider in wormhole-gui's settings caused a panic due to a fatal error. This did not occur a week or two back on develop before I updated.

### To Reproduce:
I have yet to narrow it down in a smaller code example, but it seems to be related to the change listeners.
The best way is to just draw the slider **fast** under the settings tab in [wormhole-gui](https://github.com/Jacalz/wormhole-gui/tree/develop), on latest develop.
fatal-error

 - OS: Linux
 - Version: 5.10.4
 - Go version: 1.15.6
 - Fyne version: 5fc52fcc1b31fd8dbbbe06bf58a65faab221f208

I have successfully managed to track down that it probably is caused by https://github.com/fyne-io/fyne/commit/77a198529eed3100e070b7c6b66c351c6a2e12e8.

andydotxyz:
Adds monitoring preference files for changes (re-using the settings code thankfully).
It hit the map concurrency issue so I resolved that as well.
The external looking APIs are all internal so the change is non-breaking.

Fixes #1718

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -113,8 +113,8 @@ func newAppWithDriver(d fyne.Driver, id string) fyne.App {
 	fyne.SetCurrentApp(newApp)
 
 	newApp.prefs = newPreferences(newApp)
-	if pref, ok := newApp.prefs.(interface{ load(string) }); ok && id != "" {
-		pref.load(id)
+	if pref, ok := newApp.prefs.(interface{ load() }); ok && id != "" {
+		pref.load()
 	}
 	newApp.settings = loadSettings()
 	newApp.storage = &store{a: newApp}
--- a/app/preferences.go
+++ b/app/preferences.go
@@ -62,7 +62,6 @@ func (p *preferences) load() {
 
 func (p *preferences) loadFromFile(path string) error {
 	file, err := os.Open(path) // #nosec
-	defer file.Close()
 	if err != nil {
 		if os.IsNotExist(err) {
 			err := os.MkdirAll(filepath.Dir(path), 0700)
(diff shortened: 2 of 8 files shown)


Jacalz:
@andydotxyz Looks like staticcheck failed.

Jacalz (changes requested):
Staticcheck is complaining about an error that isn't handled. This does fix the crash, but I am getting a lot of these errors when dragging the slider in wormhole-gui:
image

andydotxyz:
> This does fix the crash, but I am getting a lot of these errors when dragging the slider in wormhole-gui:

Oh no! This is strange. Has the file got corrupted somehow?
I tested against a few preference based apps here and it was quite happy.

Jacalz:
> > This does fix the crash, but I am getting a lot of these errors when dragging the slider in wormhole-gui:
> 
> Oh no! This is strange. Has the file got corrupted somehow?

I thought about that to, but removed it and tried again with the same result. It seemed to be more prominent when changing it fast.

andydotxyz:
> It seemed to be more prominent when changing it fast.

Ah, I think I have not tested lots of changes in a row. Will see if I can replicate.

Jacalz (approved):
Works great now. Thanks for the fix :+1:

PR #1746, opened 2021-01-08
Editor support for Tab

### LINKED ISSUE #761: Support for tab character

andydotxyz:
As agreed in a planning call, let's do something (simple) to get tab supported in our editors.
This will indent any tab by 4 spaces. Not ideal but it works.
Move / select / copy / paste etc all happy.

As it's a bit of a low-tech approach I wanted to share earlier days to make sure it was acceptable.

Still to do:
- [x] TextGrid support

Fixes #761

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


stuartmscott (commented):
This seems like a reasonable stop-gap approach to address the developer request 👍

Jacalz (approved):
Sens like a reasonable approach until we have proper tab support.

stuartmscott (approved):

Issue #1669, opened 2020-12-18
Move Test Markup to Files


stuartmscott:
### Is your feature request related to a problem? Please describe:
Updating XML markup tests is tedious as deltas need to be copied from the test output to a specific place inside the test files.

### Is it possible to construct a solution with the existing API?
No. Golden image tests are simpler to update as they only require copying the test output file to the golden image file.

### Describe the solution you'd like to see:
Move XML markup to files under `testdata/`, and write test output markup under `testdata/failed` so that updating a test is as simple as copying a file.

andydotxyz:
This would make it much easier to make large changes if required.

stuartmscott:
Assigned to the subject matter expert :)

andydotxyz:
On develop for testing :) (thanks @toaster)

PR #1747, opened 2021-01-09
Feature Filled Entry with narrow underline 

fpabl0:
### Description:

Background color and narrow underline for widget.Entry to be more inline with Material Design Specs.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/theme/theme.go
+++ b/theme/theme.go
@@ -103,6 +103,11 @@ const (
 	// Since 2.0.0
 	ColorNameShadow fyne.ThemeColorName = "shadow"
 
+	// ColorNameEntryBackground is the name of theme lookup for entry background color.
+	// TODO or ColorNameEntryFill?
+	// Since 2.0.0
+	ColorNameEntryBackground fyne.ThemeColorName = "entryBackground"
+
 	// SizeNameInlineIcon is the name of theme lookup for inline icons size.
 	//
 	// Since 2.0.0
@@ -132,6 +137,11 @@ const (
 	//
 	// Since 2.0.0
 	SizeNameText fyne.ThemeSizeName = "text"
+
+	// SizeNameEntryUnderline is the name of theme lookup for widget.Entry underline size.
+	//
+	// Since 2.0.0
+	SizeNameEntryUnderline fyne.ThemeSizeName = "entryUnderline"
 
 	// VariantDark is the version of a theme that satisfies a user preference for a light look.
 	//
@@ -175,6 +185,8 @@ var (
 		ColorNamePlaceHolder:    color.NRGBA{0xb2, 0xb2, 0xb2, 0xff},
 		ColorNameScrollBar:      color.NRGBA{0x0, 0x0, 0x0, 0x99},
 		ColorNameShadow:         color.NRGBA{0x0, 0x0, 0x0, 0x66},
+		// TODO define the color (why the map doesn't use the constants created??)
+		ColorNameEntryBackground: color.NRGBA{0xff, 0xff, 0xff, 0x0f},
 	}
 
 	lightPalette = map[fyne.ThemeColorName]color.Color{
@@ -187,7 +199,10 @@ var (
 		ColorNameHover:          color.NRGBA{0x0, 0x0, 0x0, 0x0f},
 		ColorNamePlaceHolder:    color.NRGBA{0x88, 0x88, 0x88, 0xff},
 		ColorNameScrollBar:      color.NRGBA{0x0, 0x0, 0x0, 0x99},
-		ColorNameShadow:         color.NRGBA{0x0, 0x0, 0x0, 0x33}}
+		ColorNameShadow:         color.NRGBA{0x0, 0x0, 0x0, 0x33},
+		// TODO define the color (why the map doesn't use the constants created??)
+		ColorNameEntryBackground: color.NRGBA{0x0, 0x0, 0x0, 0x0f},
+	}
 )
 
 type builtinTheme struct {
@@ -286,6 +301,8 @@ func (t *builtinTheme) Size(s fyne.ThemeSizeName) float32 {
 		return 3
 	case SizeNameText:
 		return 14
+	case SizeNameEntryUnderline:
+		return 2
 	default:
 		return 0
 	}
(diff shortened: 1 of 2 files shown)


theme/theme.go
```diff
@@ -163,6 +173,8 @@ var (
 		"placeholder":    color.NRGBA{0xb2, 0xb2, 0xb2, 0xff},
 		"scrollBar":      color.NRGBA{0x0, 0x0, 0x0, 0x99},
 		"shadow":         color.NRGBA{0x0, 0x0, 0x0, 0x66},
+		// TODO define the color (why the map doesn't use the constants created??)
```
andydotxyz, line 176: This was a good question - there is a PR open to fix this already ;)

theme/theme.go
```diff
@@ -123,6 +128,11 @@ const (
 	// Since 2.0.0
 	SizeNameText fyne.ThemeSizeName = "text"
 
+	// SizeNameEntryUnderline is the name of theme lookup for widget.Entry underline size.
+	//
+	// Since 2.0.0
+	SizeNameEntryUnderline fyne.ThemeSizeName = "entryUnderline"
```
andydotxyz, line 134: I could not think of a more generic way to describe this. Perhaps someone else can?
Or would we be better off hard coding it in Entry as `Padding()/2` like some other widgets do?

theme/theme.go
```diff
@@ -98,6 +98,11 @@ const (
 	// Since 2.0.0
 	ColorNameShadow fyne.ThemeColorName = "shadow"
 
+	// ColorNameEntryBackground is the name of theme lookup for entry background color.
+	// TODO or ColorNameEntryFill?
+	// Since 2.0.0
+	ColorNameEntryBackground fyne.ThemeColorName = "entryBackground"
```
andydotxyz, line 104: We try to go for the generic naming from Material, rather than being widget specific.
I think on the call yesterday was saw that MD uses FocusColor for entry background, so my vote gets that.
We already defined this name in some areas, but it was not fully, or correctly, implemented. So we can fix that here.

widget/entry.go
```diff
@@ -1067,8 +1068,11 @@ func (r *entryRenderer) Destroy() {
 }
 
 func (r *entryRenderer) Layout(size fyne.Size) {
-	r.line.Resize(fyne.NewSize(size.Width, theme.Padding()))
-	r.line.Move(fyne.NewPos(0, size.Height-theme.Padding()))
+	r.line.Resize(fyne.NewSize(size.Width, theme.EntryUnderlineSize()))
+	r.line.Move(fyne.NewPos(0, size.Height-theme.EntryUnderlineSize()))
+	// Substracting "r.box.StrokeWidth/2" in order to be ready to support OutlineBorder later in future.
```
andydotxyz, line 1073: :)
I'm happy with this approach - but should the line perhaps come in on the left/right in the same manner?

andydotxyz (changes requested):
Thanks for this. A small discussion about naming but it's probably just the thing.

Please note that there are test failures now that need to be addressed.
When we change graphical elements some rendering test will fail, it outputs the new version in a "failed" folder and prints the path - you just drag that over the old image and commit :). GitHub will then allow us to see the change inline.

Run `go test ./...` at the project root to see the list of issues.

fpabl0:
@andydotxyz 
> Thanks for this. A small discussion about naming but it's probably just the thing.

You're welcome.

> When we change graphical elements some rendering test will fail, it outputs the new version in a "failed" folder and prints the path. 

I have just done that (I have moved images from driver/software/testdata/failed to driver/software/testdata/, is it right?). However, they are still tests failed. And not only for entry, for widget.Form too (I think because it uses Entry too?). I wanted to solve these problems, but I am not sure how these tests work (specially test.AssertRendersToMarkup(...)), is there a guide that show how to use Fyne test suite?

widget/entry.go
fpabl0: You mean substracting "r.box.StrokeWidth/2" in "line" too, right?

andydotxyz:
> specially test.AssertRendersToMarkup(...)

Oh I am sorry I forgot about this (there is an ongoing discussion about how to make it as easy as the image file changes you did).
At this time you need to see the output of the test failure, and copy the change back into the embedded XML in the test code. Apologies for this. There is an open ticket to make it easier that may land tomorrow... #1669

andydotxyz:
I just realised you are probably best holding off for now. There is another big Entry PR and it will conflict with yours.
Hopefully we can land that today and you can rebase.
Otherwise the merge conflict on the test files will be huge.
Sorry I forgot about that before! @fpabl0

fpabl0:
@andydotxyz 
> At this time you need to see the output of the test failure, and copy the change back into the embedded XML in the test code.

Ok, no problems.

> There is another big Entry PR and it will conflict with yours.
Hopefully we can land that today and you can rebase.

Oh, it's ok, I will wait then. Could you tag that PR, just to be aware when it is possible to rebase this?

theme/theme.go
fpabl0: Oh yes, I forgot about it. However the obvious first step must be update the FocusColor with right value, isn't it? Otherwise, the entry will not have any form (just a blue rectangle).

andydotxyz:
Apologies I forgot to ping. All the pending Entry and Scroll changes that I am aware about are now landed.

Also, we prefer people not to use force-push once a review has been added, because it makes incremental reviewing very difficult.

theme/theme.go
andydotxyz: This fix was landed

widget/entry.go
andydotxyz: I imagined this would be required for the line and background to align, yes. Just on the x/width dimension.

theme/theme.go
fpabl0: I don't like the idea of hardcoding the size with padding as those are completely different sizes (they are not related).

widget/entry.go
fpabl0: Ok, I will do that change

fpabl0:
> Apologies I forgot to ping. All the pending Entry and Scroll changes that I am aware about are now landed.

No problems, great.

> Also, we prefer people not to use force-push once a review has been added, because it makes incremental reviewing very difficult.

Sorry about that, I don't know how to do it without forcing, I was following this [guide](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request). Could you guide how to do without `push force` or a link that can guide me? Thanks, and sorry for inconveniences.

andydotxyz:
> I don't know how to do it without forcing,

Actually if you are rebasing there is probably no way to not force. Once the reviews have started then `git merge develop` allows you to preserve history.
It's worth noting that once conflicts arise a rebase may not even be possible.
Some times it's just easier to open a new PR :(

fpabl0:
@andydotxyz 
> Not really, it just I don't know if the new value for FocusColor is ok. I am using a color a bit darker than current hover color, is that ok?

I see, that's ok. I will close this then.

andydotxyz:
FocusColor now merged, so this work should be unblocked.

PR #1750, opened 2021-01-09
Move scroll and split containers to their new locations

andydotxyz:
Removing old deprecated access.

Scroll needed to be internal as it's required in both widget and container.
It still has the same public API as before.

### Checklist:

- [ ] Tests included. <- just refactor, same code, same tests
- [x] Tests all pass.


container/split.go
```diff
@@ -1,30 +1,316 @@
...
+	Offset     float64
+	Horizontal bool
+	Leading    fyne.CanvasObject
+	Trailing   fyne.CanvasObject
+}
 
 // NewHSplit creates a horizontally arranged container with the specified leading and trailing elements.
 // A vertical split bar that can be dragged will be added between the elements.
-//
-// Since: 1.4
```
stuartmscott, line 16: Why remove these annotations?

cmd/fyne_demo/tutorials/collection.go
```diff
@@ -52,7 +52,7 @@ func makeListTab(_ fyne.Window) fyne.CanvasObject {
 	}
 	list.Select(125)
 
-	return widget.NewHSplitContainer(list, fyne.NewContainerWithLayout(layout.NewCenterLayout(), hbox))
+	return container.NewHSplit(list, fyne.NewContainerWithLayout(layout.NewCenterLayout(), hbox))
```
stuartmscott, line 55: Optionally
```suggestion
	return container.NewHSplit(list, container.NewCenter(hbox))
```

internal/widget/scroller.go
```diff
@@ -479,26 +574,26 @@ func computeOffset(start, delta, outerWidth, innerWidth float32) float32 {
 	return offset
 }
 
+// AddScrollOffsetChangedListener allows Fyne internal code to get an inside notification when a scroller moves.
+// TODO find a way to make this generally available
+func AddScrollOffsetChangedListener(s *ScrollContainer, f func()) {
```
stuartmscott, line 579: Could this be a method in ScrollContainer?
```suggestion
func (s *ScrollContainer) SetOffsetChanged(callback func()) {
```
Or maybe onOffsetChanged could be exported? (exports from `internal/` are only accessible inside the project right?)

container/split.go
andydotxyz: Apologies, that was a mistake, good catch

internal/widget/scroller.go
andydotxyz: That is why I left the TODO in there.
I wanted to get the refactor done with the understanding we knew we were special.
I would like to come back to it later, so that the refactoring PRs are not adding new public APIs.

stuartmscott (approved):

stuartmscott (changes requested):
I think this needs a rebase after your other Entry PR landed

andydotxyz:
> I think this needs a rebase after your other Entry PR landed

Oops, yes of course I should have realised!

andydotxyz:
I realised it was still using the old name as well

stuartmscott (approved):

PR #1754, opened 2021-01-09
New color value for theme.FocusColor

fpabl0:
### Description:

Set a new color value for theme.FocusColor() to be more inline with Material Design Specs.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


widget/entry_test.go
```diff
@@ -413,12 +413,12 @@ func TestEntry_FocusWithPopUp(t *testing.T) {
 	c := window.Canvas()
 
 	test.TapSecondaryAt(entry, fyne.NewPos(1, 1))
-	fmt.Println(theme.FocusColor())
+	fmt.Println(theme.FocusColor()) // TODO is this print needed?
```
andydotxyz, line 416: That should probably have been removed

andydotxyz (commented):
Thanks. Sorry that there was the massive change to tests - I guess the dummy theme used for testing did not use the same focus and primary colour.

One question - what do you mean by "ColorNameFocus: ... // TODO adjust" - is there still work to be done?

fpabl0:
> Sorry that there was the massive change to tests - I guess the dummy theme used for testing did not use the same focus and primary colour. 

No problem 👍

> One question - what do you mean by "ColorNameFocus: ... // TODO adjust" - is there still work to be done?

Not really, it just I don't know if the new value for FocusColor is ok. I am using a color a bit darker than current HoverColor, is that ok?

andydotxyz:
> Not really, it just I don't know if the new value for FocusColor is ok. I am using a color a bit darker than current HoverColor, is that ok?

If you could post a screenshot here it would be easier to discuss. One for the light and one for the dark themes.

fpabl0:
@andydotxyz 
> If you could post a screenshot here it would be easier to discuss. One for the light and one for the dark themes.

Ok, I will do that in the new PR.

Issue #1753, opened 2021-01-09
MenuItem support Selection Indication


stuartmscott:
### Is your feature request related to a problem? Please describe:
MenuItem cannot indicate that it is currently selected.

### Is it possible to construct a solution with the existing API?
no

### Describe the solution you'd like to see:
A way to indicate that a MenuItem within a Menu or PopUp is selected.

The main use case I have for this is showing which option within a PopUp is currently selected eg. the currently selected option of a Select or SelectEntry, or the current tab in an AppTab overflow PopUp or DocTab all tabs PopUp.

andydotxyz:
On `develop` for testing :)

Issue #1752, opened 2021-01-09
MenuItem support Icon


stuartmscott:
### Is your feature request related to a problem? Please describe:
MenuItem supports text, but does not support Icon.

### Is it possible to construct a solution with the existing API?
no

### Describe the solution you'd like to see:
The addition of `MenuItem.Icon fyne.Resource` to be displayed next to the existing `MenuItem.Label string`

andydotxyz:
This was merged for v2.2.0

Issue #1751, opened 2021-01-09
Change Scroll.Direction


stuartmscott:
### Describe the bug:
`ScrollContainer.Direction` is a public API however changing it from one restriction (eg. `ScrollHorizontalOnly`) to another (eg. `ScrollVerticalOnly`) will crash as the ScrollBarAreas are only created once in `ScrollContainer.CreateRenderer`

### Stack:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x42c5f89]

goroutine 13 [running]:
fyne.io/fyne/widget.(*scrollContainerRenderer).layoutBars(0xc0002042a0, 0x4200000043d18000)
	/Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/widget/scroller.go:252 +0x1a9
fyne.io/fyne/widget.(*scrollContainerRenderer).Layout(0xc0002042a0, 0x4200000043d18000)
	/Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/widget/scroller.go:274 +0x117
fyne.io/fyne/widget.(*scrollContainerRenderer).Refresh(0xc0002042a0)
	/Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/widget/scroller.go:293 +0xea
fyne.io/fyne/widget.(*BaseWidget).Refresh(0xc0002444d0)
	/Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/widget/widget.go:137 +0x5c
fyne.io/fyne/widget.(*ScrollContainer).refreshWithoutOffsetUpdate(...)
	/Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/widget/scroller.go:432
fyne.io/fyne/widget.(*ScrollContainer).Refresh(0xc0002444d0)
	/Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/widget/scroller.go:428 +0x4a
```

andydotxyz:
Fixed on `develop`

Issue #1310, opened 2020-09-13
Split TabContainer


stuartmscott:
### Is your feature request related to a problem? Please describe:
There are two main use cases for tabs:
- app navigation
  - limited tabs (3-7)
  - static data
  - bottom or leading placement
- content
  - unlimited tabs
  - dynamic data
  - top, bottom, leading, or trailing placement

### Is it possible to construct a solution with the existing API?
TabContainer is used for both use cases.

### Describe the solution you'd like to see:
Splitting TabContainer into two widgets each focusing on a specific use case:
- NavigationBar
  - displays NavigationItems
  - combination of [Navigation Rail](https://material.io/components/navigation-rail) and [Bottom Navigation](https://material.io/components/bottom-navigation)
  - tab overflow popup
- TabBar
  - displays TabItems
  - implementation of [Tabs](https://material.io/components/tabs)
  - all tabs popup

stuartmscott:
It was proposed in a Slack discussion that Navigation and Tabs would be names

stuartmscott:
Tabs could have in addition to a button which displays all tabs in a dropdown, a new tab button which triggers a callback

stuartmscott:
Relevant prior issue: #1166 

container.Router is a shorter name than container.Navigation

andydotxyz:
Router was an unfortunate name IMHO as it referred to the manner that you could pass a string/URL style component to navigate to a certain screen and they were all "mounted". Given that in some places it is used for that specific meaning it could be misleading.

stuartmscott:
Updates:

1. Names changed to AppTabs, and DocTabs.

2. AppTabs will be for static data such as the pages/sections/areas/features of an app. TabItems within AppTabs will be rendered with icon (if any) above the text (if any), and will appear in the bar running along either the bottom, or leading edge of the container.

3. DocTabs will be for dynamic data such as open files/urls. TabItems within DocTabs will be rendered with icon (if any) leading the text (if any), and will appear in the bar running along any border. Tabs will include a close button if DocTabs.OnTabClosed != nil, where OnTabClosed is a `func(*TabItem) bool` callback provided by the developer. Returning false from this callback will abort the close, else the TabItem is removed from the container.

andydotxyz:
Summary there is great.
[...]

PR #1755, opened 2021-01-10
New DocTabs and tab menu

### LINKED ISSUE #371: Implement overflow on tab container

stuartmscott:
### Description:
This PR splits TabContainer into AppTabs and DocTabs so the two main use cases of a Tabber can be addressed fully.

AppTabs is targeted towards the use case of splitting an application into different screens, and will show as many tabs as can fit on the screen, up to a maximum of 7. All other tabs will appear in an Overflow Menu. On mobile devices where screen space is limited, AppTabs ensures tabs may only be placed along the smaller edges (Top & Bottom in Portrait, Leading and Trailing in Landscape).

DocTabs is targeted towards the use case of showing one of many open documents, and like AppTabs will show as many tabs as can fit on the screen, which no maximum. All tabs appear in an All Tabs Menu.

Fixes #371 #1310

During development several issues and/or limitations were discovered and have been filed as #1751, #1752, #1753.

### Demo:
https://user-images.githubusercontent.com/1117080/104111588-78324880-5298-11eb-902d-470dff43cefa.mov

### Remaining work:
- [x] Tests still need to be written, but I felt it important to get the current progress up so others can chime in earlier.
- [x] Once #1669 is complete, TabContainer will be removed and its existing tests will be copied over.
- [x] The use of Scroll in DocTabs was debated but eventually chosen as no better option could be found. There are still numerous issues with this approach; 1) the selection indicator does not move with with the scroller as widget.ScrollContainer.onOffsetChanged is not accessible from the container package, tho this may soon be addressed by #1750 or a follow up PR. 2) The scroller bar can get in the way of selecting a tab. 3) A new scroller is created each refresh due to #1751 and as a result the scroll offset gets reset.
- [ ] Both of these tabbers need to be thoroughly verified on mobile

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


AlbinoGeek:
I'm **really** happy with how the DocTabs looks [very close to the Material Tabs]!

AppTabs also seems like a more stable-feeling interface on Desktop.

I'm exciting to be able to use these, hoping they land in 2.x

container/tabsapps.go
```diff
@@ -0,0 +1,220 @@
...
+	"fyne.io/fyne"
+	"fyne.io/fyne/canvas"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/theme"
+	"fyne.io/fyne/widget"
+)
+
+// MaxAppTabs is the maximum number of visible tabs in an AppTab.
+// Any addition tabs are moved to the overflow menu.
+const MaxAppTabs = 7
```
andydotxyz, line 13: The number of tabs displayed should not be static - it makes more sense to display all that fit, and leave the rest in the overflow.

container/tabsapps.go
```diff
@@ -0,0 +1,220 @@
...
+
+// Declare conformity with Widget interface.
+var _ fyne.Widget = (*AppTabs)(nil)
+
+// AppTabs container is used to split your application into various different areas identified by tabs.
+// The tabs contain text and/or an icon and allow the user to switch between the content specified in each TabItem.
+// Each item is represented by a button at the edge of the container.
+//
+// Since: 2.0.0
```
andydotxyz, line 22: This was "Since: 1.4" before

container/tabsapps.go
```diff
@@ -0,0 +1,220 @@
...
+type AppTabs struct {
+	baseTabs
+}
+
+// NewAppTabs creates a new tab container that allows the user to choose between different areas of an app.
+//
+// Since: 2.0.0
```
andydotxyz, line 29: As above, not a new API

theme/icons.go
```diff
@@ -160,6 +160,16 @@ const (
 	// Since 2.0.0
 	IconNameDocumentSave fyne.ThemeIconName = "documentSave"
 
+	// IconNameDotsHorizontal is the name of theme lookup for three horizontal dots.
+	//
+	// Since 2.0.0
+	IconNameDotsHorizontal fyne.ThemeIconName = "dotsHorizontal"
```
andydotxyz, line 166: "Dots" seems rather literal, a theme could choose to replace "More" with an icon that is not dots...

andydotxyz (changes requested):
Great work. It would be nice if you could show the WiP of DocTabs in the PR thread so people can discuss more easily.

container/tabsapps.go
stuartmscott: I am following what was outlined in the two bugs this PR is addressing.

theme/icons.go
stuartmscott: Dots is the MD name, but could be changed to More

andydotxyz: Do you mean it's the name of the file, or the name of it's usage?
Maybe there is a better name than "More" as well, perhaps someone else has thoughts.

container/tabsapps.go
andydotxyz: It was indeed written rather specifically - I would focus more on the "UI tends to overfit" and less on the "7". In mobile, portrait mode 5 is pretty much the limit.

andydotxyz: Sorry - It is possible I misread this the first time, I thought it was cutting the apps at 7, not limiting it.
With that in mind is limiting it to 7 the right thing to do even on desktop when the window could be much larger?

andydotxyz:
For doctabs I think we discussed close options too. Do you think that's a post-2.0 feature?
It looks really great, and should make a big difference for the usefulness in larger apps.

container/tabsapps.go
stuartmscott: Yes the behaviour is that it will display as many tabs as can fit, but no more than 7. All other tabs will be be in the overflow. I don't mind having separate limits for mobile vs desktop as the screen sizes are different, but what would be a sensible limit, or should there not be a limit at all?

container/tabsapps.go
stuartmscott: Will fix

container/tabsapps.go
stuartmscott: Will fix

theme/icons.go
stuartmscott: I agree "More" is a more semantic name for the AppTabs use case (overflow tabs), but not for DocTabs use case (all tabs)

stuartmscott:
Due to the large amount of remaining work, this will unfortunately not make the 2.0 release

container/tabsapps.go
andydotxyz: I wonder if the limit is adding anything in hindsight...
It just seems like it might lead to some empty space when not required.

theme/icons.go
andydotxyz: Yes I guess. Are there other instances that we can see the icon being used that might help us?

andydotxyz:
@stuartmscott is it possible to split out the new DocTabs work so that the overflow can be landed for AppTabs?
From what you have showed it seems like we could get that hald prepped in time.

stuartmscott:
I think it is best to keep them together as they share a common base. During development of DocTabs, I found that many changes were required in baseTabs, and thus AppTabs. I assume this will hold true as more features are added to DocTaps, therefore I would rather not rush to release a PR with code that may hinder, or require large changes to support, a future PR containing DocTabs.

andydotxyz:
> therefore I would rather not rush to release a PR with code that may hinder, or require large changes to support, a future PR containing DocTabs.

Sure, makese sense.

AlbinoGeek:
Could this hit 2.1 maybe?

andydotxyz:
> Could this hit 2.1 maybe?

I don't see why not, Let's see if we can start planning at the call on Friday.

stuartmscott:
> Could this hit 2.1 maybe?

Yup should be possible if the work to migrate to the consistent selection API (#1395) is deferred to other PRs (will require new APIs, deprecating old ones, and finally removing old ones in next major release)

container/tabsapps.go
stuartmscott: Maybe this limit should only apply on mobile?

theme/icons.go
stuartmscott: I couldn't find any, renamed to "More"

stuartmscott:
Demo:

https://user-images.githubusercontent.com/1117080/106040612-8d2c1b80-608f-11eb-9166-80a22be35efe.mov

Remaining work;
- [ ] Scroll to Selection
- [ ] Scroller seems to stick and not refresh when the selection is scrolled out of view - as there is no general purpose clipping widget I adjusted the indicator position&size so it doesn't break the bounds of that tab bar, this mostly works except for the aforementioned stickiness, I would appreciate another set of eyes on this code
- [ ] Scroller triggers OnScroll when offset hasn't changed (#1868)
- [ ] MenuItem doesn't support icons (#1752)
- [ ] MenuItem can't show if it is the currently selected tab (#1753)

Questions
- Should the close button become red on hover?
- Should the create button change selection to be the new tab?

fpabl0:
> Should the close button become red on hover?

Hmm I think the current color is ok. Maybe it should be better if the Close button only appears on Hover event?

> Should the create button change selection to be the new tab?

Yes, it should, because at some point, the user would not see if the new doc tab was added or not (because it will appear in the not visible part of the scroll container). If the create event move to the new created tab is less confused.

stuartmscott:
I agree with your reasoning on both points, thanks for your answers

container/tabsapps.go
stuartmscott: (a grid layout is used on mobile so there wont be any empty space - tab buttons will be expanded to fill)

AlbinoGeek (dismissed):
Looks good to me.

Please give me an option to enable/disable the behaviour of "Select Newly Created Tab" as a developer.

Most straight-forward example: Every browser has a setting as to whether or not you "Open new tabs when they are created."

---

The only other things I would ask for might belong in their own PRs:

- Some mechanism to prevent de-selection/navigation away from or closing of an AppTab (such as for saving the user settings on navigation, or displaying a "Hey, there are unsaved changes" dialog)

- Then... as per `DocTabs`, would doing something like this require exposing `OnScrolled`? Note how it focuses a different code sample on the right as you scroll down.

https://www.twilio.com/docs/voice/make-calls

stuartmscott:
> Please give me an option to enable/disable the behaviour of "Select Newly Created Tab" as a developer.

Not sure how we'd expose this as a developer option - Fyne is typically more opinionated

> Some mechanism to prevent de-selection/navigation away from or closing of an AppTab (such as for saving the user settings on navigation, or displaying a "Hey, there are unsaved changes" dialog)

AppTabs can't be closed, only DocTabs and it has OnCloseIntercepted for you to display a "Hey, there are unsaved changes, would you like to save, discard, or cancel?" dialog, or OnClosed for you to perform an action after the close such as saving a preference

> Then... as per DocTabs, would doing something like this require exposing OnScrolled? Note how it focuses a different code sample on the right as you scroll down.

The scrolling of the tab bar in DocTabs is not exposed, but it would not be needed to replicate the functionality presented, Instead your content could be in a scroller which you can attach a function to OnScrolled that could update other parts of the UI. In general tabbers don't control their child content, only allow the user to pick one from a set

AlbinoGeek:
> Not sure how we'd expose this as a developer option - Fyne is typically more opinionated

Agreed, however, I think this is something warranting exposing a toggle all the way down to the user, similar to the way that we would expose an #1813 `Disable Animations` option or preference somewhere in the future.

Consider this hell: https://answers.microsoft.com/en-us/edge/forum/edge_other-edge_win10/how-do-i-get-focus-on-a-new-tab-immediaely-after/c332d7fd-89c4-401d-939e-a61ed33b380a

But also, even the subtle difference between how a browser works with this setting on:

- Clicking on a link that opens a new tab results in the tab receiving focus by default.
- Right-clicking "Open in new tab" on a link DOES NOT FOCUS THE NEW TAB

etc.

> AppTabs can't be closed, only DocTabs and it has OnCloseIntercepted for you to display a "Hey, there are unsaved changes, would you like to save, discard, or cancel?" dialog, or OnClosed for you to perform an action after the close such as saving a preference

Also sounds good, we discussed this in Slack.

> The scrolling of the tab bar in DocTabs is not exposed, but it would not be needed to replicate the functionality presented, Instead your content could be in a scroller which you can attach a function to OnScrolled that could update other parts of the UI. In general tabbers don't control their child content, only allow the user to pick one from a set

Sounds good, I was not sure if the child content has an embedded scroller now that more widgets do this.

stuartmscott:
Putting browsers aside for a minute, in `container.DocTab` the user has explicitly tapped the "Create Tab" button, I think it is more intuitive that the newly created tab becomes selected. I don't understand the use case where the user has asked to create a new tab, but doesn't want to move to that tab, worse I think it will cause user-confusion especially if the newly created tab is clipped by the scroller - the user may think the button is broken, because they cannot see the result of triggering it.

andydotxyz:
Nice work.
Running the demo I couldn't help but notice two strange things.
1) When you load the doctabs the underline seems to change size - I think it loads to small and then adjusts

https://user-images.githubusercontent.com/294436/106361353-5ff09f00-6315-11eb-8a7d-fb769403804f.mov

2) The tabs screen in the demo seems to have a totally differentt style. I think this screen is not really needed - we could just have `AppTabs` and `DocTabs` as children of the `Containers` item.

Screenshot 2021-01-30 at 16 09 34

cmd/fyne_demo/tutorials/container.go
```diff
@@ -76,6 +96,21 @@ func makeCenterLayout(_ fyne.Window) fyne.CanvasObject {
 	return container.NewCenter(middle)
 }
 
+func makeDocTabsTab(_ fyne.Window) fyne.CanvasObject {
+	tabs := container.NewDocTabs(
+		container.NewTabItem("Tab 1", widget.NewLabel("Content of tab 1")),
```
andydotxyz, line 101: How about "Doc 1" instead of "Tab 1"? helps to compare with AppTabs demo.

cmd/fyne_demo/tutorials/data.go
```diff
@@ -60,6 +56,18 @@ var (
 			"A container that provides scrolling for it's content.",
 			makeScrollTab,
 		},
+		"apptabs": {"AppTabs",
```
andydotxyz, line 59: This was in alphabetical order within the section, so if keeping the tabs subgroup I think this and doctabs belong below it, if it's being removed then they belong further up in the list (in previous position)

cmd/fyne_demo/tutorials/data.go
```diff
@@ -140,7 +148,8 @@ var (
 	TutorialIndex = map[string][]string{
 		"":            {"welcome", "canvas", "animations", "icons", "widgets", "collections", "containers", "dialogs", "windows", "binding", "advanced"},
 		"collections": {"list", "table", "tree"},
-		"containers":  {"apptabs", "border", "box", "center", "grid", "split", "scroll"},
+		"containers":  {"border", "box", "center", "grid", "split", "scroll", "tabs"},
+		"tabs":        {"apptabs", "doctabs"},
```
andydotxyz, line 152: See comment in the PR conversation about whether this extra hierarchy is needed

container/apptabs.go
```diff
@@ -3,678 +3,312 @@ package container
...
-	"fyne.io/fyne/v2/driver/desktop"
-	"fyne.io/fyne/v2/internal"
 	"fyne.io/fyne/v2/layout"
 	"fyne.io/fyne/v2/theme"
 	"fyne.io/fyne/v2/widget"
 )
 
+// MaxAppTabs is the maximum number of visible tabs in an AppTab.
+// Any addition tabs are moved to the overflow menu.
+const MaxAppTabs = 7
```
andydotxyz, line 13: I'm not sure about this constant, but if it is needed can it be an unexported constant?

container/apptabs.go
```diff
@@ -3,678 +3,312 @@ package container
...
+// Declare conformity with Widget interface.
+var _ fyne.Widget = (*AppTabs)(nil)
+
 // AppTabs container is used to split your application into various different areas identified by tabs.
 // The tabs contain text and/or an icon and allow the user to switch between the content specified in each TabItem.
 // Each item is represented by a button at the edge of the container.
 //
 // Since: 1.4
 type AppTabs struct {
-	widget.BaseWidget
```
andydotxyz, line 19: This change breaks `&AppTabs{Items: ...}` constructor - can we keep compatibility?

container/apptabs.go
```diff
@@ -3,678 +3,312 @@ package container
...
+	r.updateTabs(1)
+	r.updateIndicator()
 	return r
 }
 
 // CurrentTab returns the currently selected TabItem.
-func (c *AppTabs) CurrentTab() *TabItem {
-	if c.current < 0 || c.current >= len(c.Items) {
+//
+// Deprecated: Use `AppTabs.Selection() *TabItem` instead.
```
andydotxyz, line 65: I thought that selection API change was to be a different PR?

container/apptabs.go
```diff
@@ -3,678 +3,312 @@ package container
...
-	case TabLocationTrailing:
-		buttonWidth := tabBarMinSize.Width
-		tabBarPos = fyne.NewPos(size.Width-buttonWidth, 0)
-		tabBarSize = fyne.NewSize(buttonWidth, size.Height)
-		barWidth := buttonWidth + theme.Padding()
-		linePos = fyne.NewPos(size.Width-barWidth, 0)
-		lineSize = fyne.NewSize(theme.Padding(), size.Height)
-		childPos = fyne.NewPos(0, 0)
-		childSize = fyne.NewSize(size.Width-barWidth, size.Height)
+	// Try render as many tabs as will fit, others will appear in the overflow
```
andydotxyz, line 139: With `MaxAppTabs` this comment is slightly misleading as it will be capped so it's not really as many as will fit.

container/doctabs.go
```diff
@@ -0,0 +1,298 @@
...
+// DocTabs container is used to display various pieces of content identified by tabs.
+// The tabs contain text and/or an icon and allow the user to switch between the content specified in each TabItem.
+// Each item is represented by a button at the edge of the container.
+//
+// Since: 2.0
+type DocTabs struct {
+	baseTabs
+	CreateTab          func() *TabItem
+	OnClosed           func(*TabItem)
+	OnCloseIntercepted func(*TabItem)
```
andydotxyz, line 23: Intercepting is not an after-the-fact event. The window interface calls it `SetCloseIntercept(func())` - can we better match that? i.e. `SetCloseIntercept()` func or `CloseIntercept` field?

container/doctabs.go
```diff
@@ -0,0 +1,298 @@
...
+)
+
+// Declare conformity with Widget interface.
+var _ fyne.Widget = (*DocTabs)(nil)
+
+// DocTabs container is used to display various pieces of content identified by tabs.
+// The tabs contain text and/or an icon and allow the user to switch between the content specified in each TabItem.
+// Each item is represented by a button at the edge of the container.
+//
+// Since: 2.0
```
andydotxyz, line 18: It will be 2.1 won't it?

container/doctabs.go
```diff
@@ -0,0 +1,298 @@
...
+func (r *docTabsRenderer) Objects() []fyne.CanvasObject {
+	objects := r.baseTabsRenderer.Objects()
+	if i, is := r.docTabs.current, r.docTabs.Items; i >= 0 && i < len(is) {
+		objects = append(objects, is[i].Content)
+	}
+	return objects
+}
+
+func (r *docTabsRenderer) Refresh() {
+	// TODO Offset Scroller so current tab is always visible
```
andydotxyz, line 112: Agreed, I was hoping that the current tab would be scrolled into view

container/doctabs.go
```diff
@@ -0,0 +1,298 @@
...
+	} else {
+		buttons.Layout = layout.NewHBoxLayout()
+		iconPos = buttonIconInline
+	}
+
+	for i := 0; i < count; i++ {
+		item := r.docTabs.Items[i]
+		button, ok := r.buttonCache[item]
+		if !ok {
+			button = &tabButton{
```
andydotxyz, line 178: I know this was inline in AppTabs, but this is quite a bit more complex now, perhaps a button constructor func will help reduce the complexity of this function?

container/tabs.go
```diff
@@ -0,0 +1,694 @@
...
+//
+// Since: 1.4
+func NewTabItemWithIcon(text string, icon fyne.Resource, content fyne.CanvasObject) *TabItem {
+	return &TabItem{Text: text, Icon: icon, Content: content}
+}
+
+type baseTabs struct {
+	widget.BaseWidget
+
+	Items        []*TabItem
```
andydotxyz, line 52: I think I mentioned it above but moving `Items` out of `AppTabs` (and `TabContainer` before it) is a breaking change... can it be avoided?

container/tabs.go
```diff
@@ -0,0 +1,694 @@
...
+	}
+	r.label.TextSize = theme.TextSize()
+	if r.button.text == "" {
+		r.label.Hide()
+	} else {
+		r.label.Show()
+	}
+
+	if r.icon != nil && r.icon.Resource != nil {
+		switch res := r.icon.Resource.(type) {
```
andydotxyz, line 554: This code seems unnecessary as the icons are always shown with the foreground colour...

andydotxyz: Though now I type it maybe the issue that the button icons are not being changed to primary when they should?

andydotxyz (changes requested):
This is looking great, particular with the hover effects on desktop :).

A few things inline but 2 big items:
1) This breaks `AppTabs` API contract regarding `Items` field. Perhaps the `Item` could be moved back into the widgets and private methods / interface could be used for the commonality and shared code?
2) To match list/tree/table scrolling it would be great to make the selected tab visible each time it is set.

andydotxyz:
One more thought - the add and more buttons could probably be in a horizontal row when Leadding/Trailing edge is used to take up less space...

container/tabs.go
stuartmscott: Yup, I didn't change this code, it is copied from https://github.com/fyne-io/fyne/blob/develop/container/apptabs.go#L648

container/doctabs.go
stuartmscott: Simplified by moving closing code to method

container/doctabs.go
stuartmscott: Done

container/doctabs.go
stuartmscott: Done

container/apptabs.go
stuartmscott: Done

container/apptabs.go
stuartmscott: yes, removing the deprecations will be a separate PR

container/apptabs.go
stuartmscott: Done

cmd/fyne_demo/tutorials/data.go
stuartmscott: Done

cmd/fyne_demo/tutorials/container.go
stuartmscott: Done

cmd/fyne_demo/tutorials/data.go
stuartmscott: Done

stuartmscott:
> When you load the doctabs the underline seems to change size - I think it loads to small and then adjusts

Hmm weird, I don't see this locally

> One more thought - the add and more buttons could probably be in a horizontal row when Leadding/Trailing edge is used to take up less space...

Done, tho I don't think it looks nice

stuartmscott:
Moving `Items` out of `baseTabs` and into `AppTabs` and `DocTabs` is tricky. I can't see a nice way of doing it other that duplicating the functionality which defeats the point of having `baseTabs` - 7 out of the 11 exported methods use `Items`

container/tabs.go
andydotxyz: Well, it's not being called now as far as I can see - the icons are the same colour on selected and not-selected tabs

andydotxyz:
> I can't see a nice way of doing it

Maybe a private `items()` method that both provide (there is already a setter if that was needed too).
It may be a pain, but if possible then surely worth it to maintain API compatibility.

container/tabs.go
stuartmscott: Where are you looking? The demo doesn't use icons, but when I add one it works as expected

https://user-images.githubusercontent.com/1117080/106517086-dd8aeb00-648c-11eb-9e73-ea8b6b196e86.mov

stuartmscott:
I'm not following, can you elaborate?

container/tabs.go
andydotxyz: Apologies, I forgot there were two - I meant the close icon

andydotxyz:
I think you could pass the implementation into the common code constructor to provide the data owning type. A simple private interface would define that the items can be returned from a method and that could be used to access the data.

stuartmscott:
We've spoken before about how brevity in PR comments is counter-productive. Two sentences is not much of an elaboration :(

I don't see how passing the implementation to the common code constructor will work for the `&AppTabs{Items: ...}` use case as it does not call any constructor.

container/tabs.go
stuartmscott: You want the close icon to be coloured differently if the tab is selected? What colour should it be?

andydotxyz:
Sorry I am doing a bad job explaining.
I think something that could be looked at is `internal/widget/base.go` where the base functions allow you to pass in the implementing object - this supports code-reuse but also separate types to hold the data.
It's not exactly the same but it should be more helpful than my messages can convey.

container/tabs.go
andydotxyz: What I think I was expecting was the close icon to be the same colour as the tab icon.
Don't you think so?

stuartmscott: When selected the tab's text and icon become the primary colour, which I feel has an important conveyance about what the main action a user is expected to do. Eg button uses primary colour to convey that it has high importance. Maybe tab item should use selected colour when that is introduced, but that's a separate discussion.

Close is a separate action, and a destructive one at that, so my first thought was that it should be a warning colour like red, that would be used all the time and not depending on whether the tab item is selected, but @fpabl0 didn't feel it was necessary. My main worry is that in the icon-only use case, two icons next to each other with the same colour where one changes the selected tab and the other removes tabs, it would be easy for the user to accidentally close tho I tried to avoid this somewhat by making the close button smaller and inset.

fpabl0: > Close is a separate action, and a destructive one at that, so my first thought was that it should be a warning colour like red, that would be used all the time and not depending on whether the tab item is selected

IMHO, it should not be colored, as the intention of the tabs is to be selected not be closed. Usually in ui design, as far as I know, you want to contrast things that user usually want to use. Closing a tab is not the main purpose so it should not have a contrast color in my opinion.

stuartmscott:
Thanks. This approach requires a private interface for `baseTabs` to get and set the `Items` field, and would mean the exported methods in `baseTabs` would need to be copied out to _both_ `AppTabs` and `DocTabs` which will call into unexported methods in `baseTabs` passing themselves as implementors of the interface. While it would solve the issue, I'm not sure it is the best approach for `AppTabs` and `DocTabs` to own the slice of `TabItem`s long term.

I'm considering another approach where the exported methods in `baseTabs` would be copied to only `AppTabs` which will copy `AppTabs.Items` into `baseTabs.Items` and then call the methods in `baseTabs`. I could then mark `AppTabs.Items` as deprecated and remove it and the copies in the next major. I like this approach more because it would not affect `DocTabs` and all the commonality could remain in `baseTabs` which I think will be easiest to maintain long term. Most users of `AppTabs` would not be affected when migrating to the next major as they could still assign to `Items` field and call the same public APIs, but they would be in `baseTabs` instead of `AppTabs`, the only devs affected would be ones using struct init (`a := &AppTabs{Items: ...}`) who'll need to change to (`a := &AppTabs{}; a.Items = ...`)

container/tabs.go
fpabl0: For example, online stores have a button to delete an item from the cart, but although it is a destructive action is not red, because the main purpose of the cart is not delete items, but buy them. From amazon: (Delete is "Eliminar" (spanish)):

Captura de Pantalla 2021-02-07 a la(s) 13 50 47

andydotxyz: > so my first thought was that it should be a warning colour like red

This sounds like it could be helpful - we have `theme.ErrorColor` and appropriate resource helper.

> My main worry is that in the icon-only use case, two icons next to each other with the same colour where one changes the selected tab and the other removes tabs, it would be easy for the user to accidentally close

Not sure I fully follow this - but if the close button is white on a tab that has a blue icon (because it is selected) and the icon next to it is white (because it is not selected) then is it not easier to mistake the close icon for belonging to the white icon to the right?

andydotxyz: (not the best example, but this is VS Code)
Screenshot 2021-02-08 at 20 50 03

fpabl0: > When I look at the tabs on my computer where the colour of the text changes for active/inactive the majority of the close icons match the colour of the text.

Yes, you are right. But they don't have the primary color on it. The primary color is the background not the text color.

fpabl0: > (not the best example, but this is VS Code)

Indeed, it is a good example. That looks good, because the text color is white and not red, it would look weird if it uses the primary color.

stuartmscott: I thought about using `theme.ErrorColor` but the semantics of using error to convey danger seemed off.

I agree with Pablo that the way the selection is shown is different. Where in our tabbers change the foreground colour but that VS code example change the background colour. Maybe it would be better if tabbers followed the new collections which use background to indicate selection?

fpabl0: >  Maybe it would be better if tabbers followed the new collections which use background to indicate selection?

I think it would look better with the background to indicate selection :)

stuartmscott: It would get rid of the current underline indicator and associated animation

fpabl0: Yes, another way could be this:

Captura de Pantalla 2021-02-08 a la(s) 16 52 13

Captura de Pantalla 2021-02-08 a la(s) 16 56 31

However, a new theme color would be needed, I think.

andydotxyz: > that VS code example change the background colour.

If you look carefully it is changing both

> Maybe it would be better if tabbers followed the new collections which use background to indicate selection?

We used to do that - people did not like it and it is not what material uses for tabs

What I don't follow is why any of these thoughts provide a compelling reason to break from the MD looks.

stuartmscott: Fair enough, the Tabs spec (https://material.io/components/tabs) doesn't mention anything about closing tabs so I'm inclined to leave this design as is until we know better and can iterate in future.

fpabl0: > We used to do that - people did not like it and it is not what material uses for tabs

I see, that's fair. As @stuartmscott said, Material does not mention anything about Doc Tabs (with close button), because it does not exist there, as far as I know. However, after looking the Google Cloud Shell terminal, @andydotxyz suggestion maybe its the better choice?

Captura de Pantalla 2021-02-09 a la(s) 11 29 34

Looking at how this is implemented in other apps based on Material Design Specs, I found this:

image

container/tabs.go
stuartmscott: Fixed

container/apptabs.go
stuartmscott: Fixed

container/apptabs.go
```diff
@@ -22,6 +22,10 @@ var _ fyne.Widget = (*AppTabs)(nil)
 // Since: 1.4
 type AppTabs struct {
 	baseTabs
+	// Deprecated: 'Items' has been moved into an embedded type.
```
andydotxyz, line 25: I don't think this is wise. If you remove this then it will no longer be possible to do:

```go
&AppTabs{
    Items: ....
}
```

It's not just a case that the field will have moved, but that the shared type is private so it cannot be initialised externally.

container/apptabs.go
```diff
@@ -60,6 +66,15 @@ func (t *AppTabs) CreateRenderer() fyne.WidgetRenderer {
 	return r
 }
 
+// Append adds a new TabItem to the end of the tab bar.
+//
+// Deprecated: Moved into an embedded type.
```
andydotxyz, line 71: These don't need to be deprecated as far as I can tell, moving a function to an embedded type in the future is a transparent operation.

container/apptabs.go
```diff
@@ -35,6 +39,7 @@ func NewAppTabs(items ...*TabItem) *AppTabs {
 			BaseWidget: widget.BaseWidget{},
 			current:    -1,
 		},
+		Items: items,
```
andydotxyz, line 42: I don't agree with this duplication of data if it can be avoided.
If it cannot be avoided then you are missing setters in `Refresh` where the data could have changed due to external updates.

andydotxyz (changes requested):
I don't like the idea that we plan to break the struct initialisation of AppTabs in the future

container/apptabs.go
stuartmscott: Yes it will no longer be possible to set `Items` in the struct initialization, devs will need to set it afterwards as the comment on line 27 details the release notes will also need to mention this.
It can be initialized externally as the baseTabs is embedded so `a.Items = ...` will work.
It does not seem unwise to me to trade off setting `Items` in the struct initialization for simpler and more maintainable code.
I don't see the issue with moving `Item` initialization onto a separate line (`a.Items = ...` or `a.SetItems(...)`) as struct initialization alone is not sufficient to initialize AppTabs anyway - the dev will still need to call `a.ExtendBaseWidget(a)`

container/apptabs.go
stuartmscott: Fixed

container/apptabs.go
stuartmscott: Yeah I wasn't sure how to articulate this, it should not affect any devs, but at the same time I want to tag this method for removal. Maybe `//TODO @stuartmscott Remove in next major`?

container/tabs.go
stuartmscott: Close buttons now become primary coloured if tab is selected;

https://user-images.githubusercontent.com/1117080/107537202-78638380-6b77-11eb-84f1-e5e7a9b12c4e.mov

container/apptabs.go
andydotxyz: > the dev will still need to call a.ExtendBaseWidget(a)

This should not be required, the `CreateRenderer` or `MinSize` will take care of this like other widgets won't it?
That will only be needed if they want to extend further with their own code.

andydotxyz:
> the only devs affected would be ones using struct init (a := &AppTabs{Items: ...}) who'll need to change to (a := &AppTabs{}; a.Items = ...)

You will also notice that the fields that belong to a private struct will disappear from the documentation as they are not exported on the type.

I think this change is setting a precident that I am not comfortable with - altering API to accomodate internal implementation details.

stuartmscott:
I don't think the interface approach is the right way to go, but I've done it anyway - I'll just assign future bugs to you 😛

container/apptabs.go
stuartmscott: Moot

container/apptabs.go
stuartmscott: Moot

container/doctabs.go
stuartmscott: Done

andydotxyz:
> I don't think the interface approach is the right way to go, but I've done it anyway - I'll just assign future bugs to you 😛

Wow - it went the other end of the spectrum. I was meaning that you could use an interface for defining access to the items (and other exported fields I guess, if required).
Was it not beneficial to have the internal state and shared methods on a single type - I don't think there was any harm deriving from a struct, it was just moving public init fields that did not belong there.

stuartmscott:
Latest DocTabs demo;

https://user-images.githubusercontent.com/1117080/107585044-5e936200-6bb2-11eb-9ef1-6f9a1000deca.mov

container/apptabs.go
andydotxyz: What I meant is that moving to a new selection API should be a separate PR, as it is not core to getting this working.
Additionally the `Selection` was illustrated for multi-select towards the end of the discussion on this.
`Selected` seems more appropriate for single item selection, and matches the API elsewhere of `OnSelected`.
Our other widgets still use `Selected` until we agree on the larger item of work to move them all to a new API in preparation for multiselect.

container/apptabs.go
```diff
@@ -18,670 +23,396 @@ import (
...
 	}
-	return c.Items[c.current]
+	return t.Items[t.current]
 }
 
 // CurrentTabIndex returns the index of the currently selected TabItem.
-func (c *AppTabs) CurrentTabIndex() int {
-	return c.current
+//
+// Deprecated: Use `AppTabs.SelectionIndex() int` instead.
```
andydotxyz, line 86: `SelectedIndex` is the current state-of-play regarding single selections.

container/apptabs.go
```diff
@@ -18,670 +23,396 @@ import (
...
+		t.OnChanged(t.Items[t.current])
 	}
 }
 
-// SetTabLocation sets the location of the tab bar
-func (c *AppTabs) SetTabLocation(l TabLocation) {
-	c.tabLocation = l
-	c.Refresh()
+// Selection returns the currently selected TabItem.
+func (t *AppTabs) Selection() *TabItem {
```
andydotxyz, line 167: It should be `Selected` - we have not implemented (nor finalised) the multi-select enabling future API for selection

container/apptabs.go
```diff
@@ -18,670 +23,396 @@ import (
...
+	return selection(t)
 }
 
-// Show this widget, if it was previously hidden
-func (c *AppTabs) Show() {
-	c.BaseWidget.Show()
-	c.SelectTabIndex(c.current)
-	c.Refresh()
+// SelectionIndex returns the index of the currently selected TabItem.
+func (t *AppTabs) SelectionIndex() int {
```
andydotxyz, line 172: `SelectedIndex`

container/doctabs.go
```diff
@@ -0,0 +1,431 @@
...
+	t.Refresh()
+}
+
+// SelectIndex sets the TabItem at the specific index to be selected and its content visible.
+func (t *DocTabs) SelectIndex(index int) {
+	selectIndex(t, index)
+	t.Refresh()
+}
+
+// Selection returns the currently selected TabItem.
```
andydotxyz, line 126: Let's match existing widgets and not jump to the nearly-agreed multiselect version of the api.
`Selected` is what is used elsewhere and matches `OnSelected` callback on this widget.

container/doctabs.go
```diff
@@ -0,0 +1,431 @@
...
+func (t *DocTabs) Selection() *TabItem {
+	return selection(t)
+}
+
+// SelectionIndex returns the index of the currently selected TabItem.
+func (t *DocTabs) SelectionIndex() int {
```
andydotxyz, line 132: `SelectedIndex`

andydotxyz (changes requested):
It looks like this is not using the current Selection method names but possibly some future version that we have not begun work on?
I raised this in a previous review, but the comment was dismissed

container/doctabs.go
```diff
@@ -0,0 +1,431 @@
...
+	r.updateCreateTab()
+	r.updateTabs()
+	r.updateIndicator(false)
+	r.applyTheme(t)
+	return r
+}
+
+// Hide hides the widget.
+//
+// Implements: fyne.Widget
```
andydotxyz, line 87: I think this should be `fyne.CanvasObject`

container/tabs.go
```diff
@@ -0,0 +1,675 @@
...
+		return fyne.NewSize(theme.Padding()*4, theme.Padding()*2)
+	}
+	return fyne.NewSize(theme.Padding()*2, theme.Padding()*2)
+}
+
+var _ fyne.Widget = (*tabCloseButton)(nil)
+var _ fyne.Tappable = (*tabCloseButton)(nil)
+var _ desktop.Hoverable = (*tabCloseButton)(nil)
+
+type tabCloseButton struct {
```
andydotxyz, line 570: Did you consider a button with an icon and Importance = LowImportance?

container/apptabs.go
stuartmscott: By "our other widgets" do you mean `widget.Select`? It does not implement the Selection API from (#1395) so I don't think it should not be used as a guiding principle. Actually the only widgets implementing #1395 are `widget.List`, `widget.Table`, and `widget.Tree` and neither #1395 nor any of the implementations have a `Selected` method.

I could deprecate `CurrentTab` and `CurrentTabIndex` and not provide an alternative, would that be better?

What selection model should DocTabs use? It would be bad to introduce an API in a new widget that will be deprecated. It would also be bad if the two tabbers had different selection APIs.

container/apptabs.go
stuartmscott: That is not what #1395 says

container/apptabs.go
stuartmscott: That is not what #1395 says, and tabbers will never support multi-select

container/apptabs.go
stuartmscott: I don't know where you're getting this from, I'm following #1395

stuartmscott:
> Can you remind me the purpose of maxAppTabs? I can't recall and in reading through the code it seems to be needed in documentation as a warning - is it really what developers will expect?

It is a limit defined in the requesting ticket: "limited tabs (3-7)" (#1310) and the MD documentation: "display three to five destinations" (https://material.io/components/bottom-navigation#usage). If you had an issue with this limit you really should have brought it up earlier on the ticket, not 5 months later after it had been implemented

container/doctabs.go
stuartmscott: Widget encapsulates CanvasObject, prior art;

- internal/driver/glfw/menu_bar.go#64
- internal/driver/glfw/menu_bar_item.go#73
- internal/widget/overlay_container.go#35
- internal/widget/scroller.go#110,241,499
- internal/widget/shadow.go#63
- widget/entry.go#291
- widget/menu.go#147
- widget/menu_item.go#61
- widget/popup_menu.go#47
- widget/select.go#97
- widget/separator.go#41

container/tabs.go
stuartmscott: When `tabButton` is hovered the `tabCloseButton` becomes visible, but even after `tabButton` receives `MouseOut` the `tabCloseButton` must remain visible if it is hovered. `widget.Button.hovered` is not accessible from the `container` package.

stuartmscott: (container/tabs.go#540)

container/apptabs.go
andydotxyz: Scroll all the way to the end and you will see `Selected` is singular and `Selection` is the (proposed) version for multi. Note also that there are 2 proposals. One includes `Selection` and one does not.

Either way the multi-select portion of that has a) not been implemented and b) does not apply to singular selection widgets.
Can this please match `Radio`, `Select` and `FileIcon` in the current iteration so we do not have an inconsistent API?

container/apptabs.go
andydotxyz: `Selection` in the (not yet implemented) proposal is for multiselect - so see above for recommendation

container/apptabs.go
andydotxyz: Likewise

container/doctabs.go
andydotxyz: Goodness it looks like we have a lot to tidy up, thanks.
I think RadioGroup may be the only one that references the root interface.

container/tabs.go
andydotxyz: Thanks

andydotxyz:
> If you had an issue with this limit you really should have brought it up earlier on the ticket, not 5 months later after it had been implemented

I brought it up many times during development, hence my use of "remind me" in the request.

container/apptabs.go
stuartmscott: I don't see `func Selected()` at the end of #1395, can you link to the message you are referring to?

> Can this please match Radio, Select and FileIcon in the current iteration so we do not have an inconsistent API?

So not a func, but an exported `Selected` field?

stuartmscott:
As discussed on the call the maxAppTabs limit will be dropped

container/doctabs.go
stuartmscott: Fixed annotation in AppTabs and DocTabs, other instances will need to be addressed in separate PRs

container/doctabs.go
stuartmscott: Done

container/apptabs.go
stuartmscott: Done

container/apptabs.go
stuartmscott: Done

container/doctabs.go
stuartmscott: Done

container/apptabs.go
stuartmscott: Implemented `func Selected()`

container/apptabs.go
stuartmscott: Implemented `func Selected()`

stuartmscott:
Flakey form tests failing on Ubuntu: https://github.com/fyne-io/fyne/runs/1889563531

andydotxyz (approved):

andydotxyz:
Very excited for this one, great work thanks.

Wilm0rien:
Hi @stuartmscott, this feature looks really great and I'm looking forward to get my fingers on it as soon it is on master!

PR #1760, opened 2021-01-10
Fix incorrect progressbar infinite documentation

Jacalz:
### Description:

It was incorrectly referencing a value and max value while also commenting on internal API details.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/progressbarinfinite.go
+++ b/widget/progressbarinfinite.go
@@ -113,7 +113,7 @@ func (p *infProgressRenderer) start() {
 	}
 }
 
-// Stop the infinite progress goroutine and sets value to the Max
+// Stop the infinite progress bar background thread from to updating it
 func (p *infProgressRenderer) stop() {
 	p.progress.propertyLock.Lock()
 	defer p.progress.propertyLock.Unlock()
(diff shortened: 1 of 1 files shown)


widget/progressbarinfinite.go
```diff
@@ -113,7 +113,7 @@ func (p *infProgressRenderer) start() {
 	}
 }
 
-// Stop the infinite progress goroutine and sets value to the Max
+// Stop the infinite progress bar background thread from to updating it
```
stuartmscott, line 116:
```suggestion
// Stop the infinite progress bar background thread from updating it
```

Or better yet:
```suggestion
// Stop the background thread from updating the infinite progress bar
```

stuartmscott: Note: this in an unexported function, so the doc is not developer-facing

stuartmscott (commented):
Good call on using animation over goroutine/background thread

andydotxyz:
Looks good, though I agree with Stuart about the changes.

widget/progressbarinfinite.go
Jacalz: Oh, good suggestion. Indeed, that’s why I made sure to keep the internal details here 🙂

stuartmscott (approved):

PR #1763, opened 2021-01-10
Feature Filled narrow underline entry

fpabl0:
### Description:

Background color and narrow underline for widget.Entry to be more inline with Material Design Specs.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


fpabl0:
There are some tests failed, however I want to know if it is safe to replace just the images, or editing the "mock theme" used for tests??

This image I got from failed folder:
theme_initial

And the original one is:
theme_initial

As you can see, the one on failed folder, you cannot see the underline, maybe because of the colors.

fpabl0:
Should I edit this part to make PrimaryColor different from FocusColor?

testtheme.go
```
func NewTheme() fyne.Theme {
	return &configurableTheme{
		colors: map[fyne.ThemeColorName]color.Color{
			theme.ColorNameBackground:     red,
			theme.ColorNameButton:         color.Black,
			theme.ColorNameDisabled:       color.Black,
			theme.ColorNameDisabledButton: color.White,
			theme.ColorNameError:          blue,
			theme.ColorNameFocus:          green,
			theme.ColorNameForeground:     color.White,
			theme.ColorNameHover:          green,
			theme.ColorNamePlaceHolder:    blue,
			theme.ColorNamePrimary:        green,
			theme.ColorNameScrollBar:      blue,
			theme.ColorNameShadow:         blue,
		},
```

andydotxyz:
> Should I edit this part to make PrimaryColor different from FocusColor and add SizeNameEntryUnderline?

You should update test/testtheme.go so they are different - but you should change FocusColor, not PrimaryColor :)

andydotxyz:
Great addition, thanks!

widget/entry.go
```diff
@@ -1039,8 +1040,12 @@ func (r *entryRenderer) Destroy() {
 }
 
 func (r *entryRenderer) Layout(size fyne.Size) {
-	r.line.Resize(fyne.NewSize(size.Width, theme.Padding()))
-	r.line.Move(fyne.NewPos(0, size.Height-theme.Padding()))
+	// Used to be ready to support OutlineBorder later in future.
+	remBoxBorder := r.box.StrokeWidth / 2
```
andydotxyz, line 1044: I didn't realise this before, but `remBoxBorder` will always be 0.
Did you mean to set it as `theme.EntryUnderlineSize()`?

andydotxyz (changes requested):
Just a quick question about the maths, either something was missed or it can be simplified

widget/entry.go
fpabl0: Yes, it is 0, because we are not using border, but if we add border in the future (if that is the case), it won't be 0 anymore

andydotxyz: Ah I misunderstood. There is no point having code in there that does nothing.
I thought you were adjusting it so that if border were added later it would not change the sizes.

fpabl0: > I thought you were adjusting it so that if border were added later it would not change the sizes.

I am doing this. Now I am lost.
canvas.Rectangle size only applies to its content and half-border, the other half is outside its size. If later we add border and we do not have that line, we have this look:

Captura de Pantalla 2021-01-11 a la(s) 12 36 08

But if we keep that line, the entry will look good:

Captura de Pantalla 2021-01-11 a la(s) 12 36 56

andydotxyz: If box.StrokeWith is currently 0 then this code has no impact and the background reaches the edges I think.
If you then set the strokewidth higher it covers the part of the entry, changing the usable space.
I had thought this was to ensure we didn't change that space and/or the outside size.

If the alteration is to move the rectangle in from the edge, and have no impact on anything else, then the code will do nothing until a future release when we add outlines - so it's just confusing to leave it in at this time.

fpabl0: > the code will do nothing until a future release when we add outlines

In fact, that was the idea, that's why I wrote "// Used to be ready to support OutlineBorder later in future".

So I will delete it now to avoid confusion.

fpabl0: @andydotxyz I have just removed the unused code.

fpabl0:
I don't have test errors locally 🧐

andydotxyz (approved):

andydotxyz:
Looks good thanks.
I think you have caught up in the develop issues, so you need to rebase again - sorry!

PR #1764, opened 2021-01-10
Button/select tap animation

### LINKED ISSUE #178: Buttons does not have a click animation
Is there something in progress or a way to implement it?

andydotxyz:
Button tap (and select button) animation.
It's not exactly what MD specifies, but it might be good enough?

As this is design am looking to get feedback as well, what more does this need to be OK for initial version?

Fixes #178 

### Checklist:

- [ ] Tests included. <- we have not got animation test framework figured out yet, just run an app and tap a button
- [x] Tests all pass.


Jacalz (changes requested):
This is a lot better than no tap animation, but I think I would prefer to have the animation go out like an ellipsis or circle shape (possibly depending on button size of course). It is hard to explain, but it looks a little bit too "hard". Perhaps a slightly brighter colour for the animation would work better?

I really like how this looks (as an example):
button-tap

andydotxyz:
>  I would prefer to have the animation go out like an ellipsis or circle shape

I would prefer that as well, but until we have widgets allowed to do arbitrary cliping thhat can't really be done.
Rectangles are the only  route available at the moment as they can stop at the widget edge.

Jacalz:
How about a light colour for the Tao animation then? I think that might look slightly better.

andydotxyz:
Thanks. Updated with light tap for the dark theme. Turns out that's what material wanted too ;)

https://user-images.githubusercontent.com/294436/104300712-bdd05a80-54be-11eb-94c7-c11c938f5929.mov

andydotxyz:
Updated and fixed an issue where the animation could overlap button shadow

Jacalz:
Thanks for sharing a video. That looks really good 👍 Will do a review later today (don’t have access to my development machine currently) 🙃

Just a heads up. It seems that the tests are failing ❌

andydotxyz:
> Just a heads up. It seems that the tests are failing ❌

Thanks. Seems like the doubletapped tests have an instability in them, I  don't think it's this ticket's fault but let's see on a re-run.

andydotxyz:
I think it was not a new issue, but this exacerbated it. I have a fix in now so it's stable again.
Turns out the test was not one of our most accurate.

andydotxyz:
We are green now :) ✅

Jacalz (approved):
Nice work! This looks great visually and code looks good as well :+1: 

We probably want to make sure that this applies to tabs, tree and more in the future, but this is a great stop-gap. Might even want to make the animation go out from the cursor tap, but that is probably also a different change down the line.

Issue #1737, opened 2021-01-08
Cursor does not go down to next line with wrapping


Areturn:
### Describe the bug:

What's happening here?

### Screenshots:

https://user-images.githubusercontent.com/25275823/103982939-6335b680-51bf-11eb-8c3b-91a1ca7a24e7.mp4

### Example code:

	entryLoremIpsum := widget.NewMultiLineEntry()
	entryLoremIpsum.SetText("11111111111111111111111")
	entryLoremIpsumScroller := container.NewVScroll(entryLoremIpsum)

	entryLoremIpsum.Wrapping = fyne.TextWrapWord
	w.SetContent(entryLoremIpsumScroller)
	w.ShowAndRun()

### Device (please complete the following information):
 - OS: kali
 - Version: 2.0
 - Go version: 1.15.2
 - Fyne version: 1.14.2

andydotxyz:
Well found.
I think we must be missing tests for Entry with word wrap turned on!

andydotxyz:
This is being worked on by @aayush420

aayush420:
@Jacalz I'm using this test to verify the bug. 
```
func TestEntry_CursorColumn_Wrap2(t *testing.T) {
	entry := widget.NewMultiLineEntry()
	entry.Wrapping = fyne.TextWrapWord
	entry.Resize(fyne.NewSize(64, 64))
	entry.SetText("1234")
	entry.CursorColumn = 3
	test.Type(entry, "a")
	test.Type(entry, "b")
	test.Type(entry, "c")
	assert.Equal(t, 1, entry.CursorColumn)
	assert.Equal(t, 1, entry.CursorRow)
	w := test.NewWindow(entry)
	w.Resize(fyne.NewSize(70, 70))
	test.AssertImageMatches(t, "dummyfile.png", w.Canvas().Capture())
}
```
Here's the `dummyfile.png` which shows the required output.
![dummyfile](https://user-images.githubusercontent.com/16898783/104158169-cff4bf00-5412-11eb-9638-c0c967a4b38f.png)

aayush420:
PR at https://github.com/fyne-io/fyne/pull/1766

andydotxyz:
Merged, thanks for the help on this @aayush420 :)

PR #1766, opened 2021-01-11
Fix and test case for 1737

aayush420:
### Description:
Moves the cursor down to next line with wrapping after the line is full

Fixes #(issue)
[1737](https://github.com/fyne-io/fyne/issues/1737)

### Checklist:
- [x] Tests included.
- [x] Tests all pass.


andydotxyz (changes requested):
The fix looks good but can you use a more sensible filename for the test artifact?

andydotxyz (approved):

andydotxyz:
Can we get 1 more set of eyes please? I was quite close to the implementation of this.

widget/entry.go
```diff
@@ -820,6 +820,9 @@ func (e *Entry) rowColFromTextPos(pos int) (row int, col int) {
 				row++
 			}
 			col = pos - b[0]
+			if e.Wrapping != fyne.TextWrapOff && b[0] == pos && col == 0 && pos != 0 {
```
stuartmscott, line 823: This check is guarding with TextWrapOff, presumably so that the cursor moves down in the case of TextWrapBreak & TextWrapWord, but what about TextTruncate?

stuartmscott (changes requested):
Please add tests to ensure the fix works for each of the 3 types of wrapping; on, off, truncate. So far it appears only the first two are tested.

widget/entry.go
andydotxyz: I thought we did not support truncating an entry?

andydotxyz:
```go
func (e *Entry) textWrap() fyne.TextWrap {
	if e.Wrapping == fyne.TextTruncate {
		fyne.LogError("Entry does not allow Truncation", nil)
```

stuartmscott: Oh yes, quite right

stuartmscott (approved):

Issue #1701, opened 2020-12-30
widget.Entry and widget.Form redesign


fpabl0:
### Is your feature request related to a problem? Please describe:
Yes, the current widget.Entry does not show the error messages, just notify to the user that there is an error. Current widget.Form only allows us to create 1 column form.

### Is it possible to construct a solution with the existing API?
Yes, it is possible but there will be break changes (at least for the widget.Form, the other ones can be implemented as new widgets)

### Describe the solution you'd like to see:
I have designed some apps with Flutter framework and I could say that its API is great. So maybe Fyne could adopt somethings from it. Based on it and to make things easier, the Entry widget could be redesign to this appearance:

https://user-images.githubusercontent.com/12239342/103375215-9c777380-4aa7-11eb-9574-e929e23bce15.mp4

And then create something like a widget.TextFormField that will add the label on the top of rectangle input as you can see in the video.
widget.Form could be completely redesign to accept variable columns. 
This is how the above video was done:
```
input := driverservice.CreateInput{}
vtOpts := []string{"Motorcycle","Car"}
form := xwidget.Form{}
form.Inputs = []xwidget.FormInput{
    xwidget.NewTextFormField("First name:", "", xwidget.InputFieldText).
	SetValidator(xwidget.NewFormValidator().NotEmpty().Build()).
	SetOnSaved(func(s string) { input.FirstName = s }),
   xwidget.NewTextFormField("Last name:", "", xwidget.InputFieldText).
	SetValidator(xwidget.NewFormValidator().NotEmpty().Build()).
	SetOnSaved(func(s string) { input.LastName = s }),
   xwidget.NewTextFormField("ID Card:", "", xwidget.InputFieldText).
	SetValidator(xwidget.NewFormValidator().NotEmpty().Build()).
	SetOnSaved(func(s string) { input.IDCard = s }),
   xwidget.NewTextFormField("Driver License:", "", xwidget.InputFieldText).
	SetValidator(xwidget.NewFormValidator().NotEmpty().Build()).
	SetOnSaved(func(s string) { input.DriverLicense = s }),
   xwidget.NewTextFormField("Email:", "", xwidget.InputFieldEmail).
	SetValidator(xwidget.NewFormValidator().NotEmpty().Email().Build()).
	SetOnSaved(func(s string) { input.Email = s }),
  xwidget.NewTextFormField("Phone:", "", xwidget.InputFieldText).
[...]

PR #1767, opened 2021-01-11
Initial form hint and error text shown

### LINKED ISSUE #1269: TODO: Validation error reason should notify user (HelperText?)
The input validation has support for including a reason to the user as to why their input isn't valid. We need to have some good way of showing this to the user.

Should be possible yes. How we do it is another issue entirely. Do we want HelperText per Material Design or something simpler?

I think that I prefer HelperText, but it needs more thought.

andydotxyz:
Show errors when they are reported for any Validatable input.
Add Hint text to display in the same space.
Compact to original height when neither are present.

Could discuss how to make the error and input validation error warning sync - but it didn't seem quite right, so perhaps somewhere in between?

Fixes #1269
Relates to #1701

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/form_test.go
```diff
@@ -121,7 +121,7 @@ func TestForm_ChangeTheme(t *testing.T) {
 		},
 		OnSubmit: func() {}, OnCancel: func() {}}
 	w := test.NewWindow(form)
-	w.Resize(fyne.NewSize(340, 240))
+	w.Resize(fyne.NewSize(340, 280))
```
stuartmscott, line 124: Could this use `form.MinSize()` so the golden images are as small as possible?

andydotxyz: To my knowledge in the theme change tests we have not done so, because the min-size changes mid-test.
The number there is actually smaller than the minimum on the ugly test, as can be seen by the thin bottom edge.

cmd/fyne_demo/tutorials/widget.go
```diff
@@ -273,8 +273,8 @@ func makeFormTab(_ fyne.Window) fyne.CanvasObject {
 
 	form := &widget.Form{
 		Items: []*widget.FormItem{
-			{Text: "Name", Widget: name},
-			{Text: "Email", Widget: email},
+			{Text: "Name", Widget: name, HintText: "Your full name"},
+			{Text: "Email", Widget: email, HintText: "A valid email address"},
```
fpabl0, line 277: Don't you think should be a great idea of having NewEntryFormField or NewEntryFormInput (just as shorthands), that encapsulate `{Text: "Name", Widget: name, HintText: "Your full name"}`, then user does not have to apparently create two objects? Something like:
```
e := widget.NewEntryFormField("Label", "HintText")
[]*widget.FormItem{
   e, ...
}
```

andydotxyz: Well, it is an interesting question, but I don't think the code illustration works - as it seemingly moves the fields into Entry and would require a new constructor for every type of input/widget, as the current design can apply to any widget.

Normally our constructors cover only required elements, which is why HintText would not really fit.

fpabl0: > Well, it is an interesting question, but I don't think the code illustration works - as it seemingly moves the fields into Entry

Not really, there could be a file called form_inputs.go and move all formInputs related to that file and define there these new constructors, maybe??

> Normally our constructors cover only required elements, which is why HintText would not really fit.

Yes, you are right, maybe it could be:

```
e := widget.NewEntryFormField("Label").SetHintText("HintText")
[]*widget.FormItem{
   e, ...
}
```

widget/form_test.go
stuartmscott: The window will resize to accommodate the content's min size tho, and the default and ugly themes are compared against separate images so their sizes don't need to match. I bring it up because it there was concern image tests were too large, this is one way to reduce them

Jacalz (changes requested):
One possible in a typo, otherwise I think this looks good.

cmd/fyne_demo/tutorials/widget.go
andydotxyz: That style of code ("chaining") is something we discussed but have not yet decided upon.
As far as I can see your code is introducing a `FormEntry` type to avoid using the `FormItem` metadata struct.
I'm not following why this is better, sorry.

widget/form_test.go
andydotxyz: Yeah fair point. Maybe we should make a mental note to fix each as we see them.
I've updated these ones now.

theme/theme.go
```diff
@@ -312,6 +319,11 @@ func ButtonColor() color.Color {
 	return current().Color(ColorNameButton, currentVariant())
 }
 
+// CaptionTextSize returns the standard text size
```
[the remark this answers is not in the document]
andydotxyz: Fixed, also not sure how some test images disappearerd, re-added

cmd/fyne_demo/tutorials/widget.go
fpabl0: > That style of code ("chaining") is something we discussed but have not yet decided upon.

I see, sorry. Although I think would be a nice idea that kind of style specially if the constructors only have required parameters. Or maybe create a method SetOpts(&WidgetOpts{}), so options can be unified.

> As far as I can see your code is introducing a FormEntry type to avoid using the FormItem metadata struct.

Actually, I was thinking on upgrade FormItem to an actual widget that holds hint/error texts and label.

> I'm not following why this is better, sorry. 

I think in this way, label and entry is more compact, as label is defined at the same time as the entry and not later in the form. Also this will let us to use that entry variant with hint/error text outside Form. However, it is just style decision, maybe I am wrong.

andydotxyz: > Actually, I was thinking on upgrade FormItem to an actual widget that holds hint/error texts and label.

That approach precludes the ability for labels to be sometimes-left and sometimes-above for different form layouts.

> Also this will let us to use that entry variant with hint/error text outside Form

This is a good point, and it came up over the last few days. I think the agreement was that we land in form, see how it works out, and if we need to extrat the behaviour to another widget we could do so later.

fpabl0: > That approach precludes the ability for labels to be sometimes-left and sometimes-above for different form layouts.

That's true, however I think we can manage a private property, in order to reserve or not space for label, then if horizontal alignment is used, there is no need to reserve space for label, and it will just look as now. 

> This is a good point, and it came up over the last few days. I think the agreement was that we land in form, see how it works out, and if we need to extrat the behaviour to another widget we could do so later.

Looks good to me.

andydotxyz: >  in order to reserve or not space for label, 

I don't think the issue would be that it would not reserve the space so much as how do we tell this new widget not to draw the label because we want it on the left instead?

fpabl0: > I don't think the issue would be that it would not reserve the space so much as how do we tell this new widget not to draw the label because we want it on the left instead?

Yes, you are right, something like that can be done. After looking at the available widgets, I wonder what other widgets (apart from Entry) are benefited with helper/error text decoration. It seems Entry is the only one, am I missing something?

andydotxyz:
Current status I think this is all I plan to add functionality wise:

Screenshot 2021-01-11 at 22 25 55

andydotxyz:
Looks like moving the label was not the right thing to do:

Screenshot 2021-01-11 at 23 22 16

andydotxyz:
Better this way I think, also fixes the entry content to look central 👍

andydotxyz:
Alternative approach, adjust form but don't edit other widgets, looks like smaller things (slider, check) are not well aligned but the rest is
Screenshot 2021-01-12 at 00 03 54

andydotxyz:
OK, I have made the fairly sweeping change that label, radio, check needed to be larger. This fixes alignment pretty much across the board: 
Screenshot 2021-01-12 at 13 04 17

I am sure there are further tweaks we will find, but this feels like the sort of break that needed to be in before 2.0 hits the shelves

andydotxyz:
Outside of form context:

data/binding/queue.go
```diff
@@ -1,6 +1,6 @@
 package binding
 
-var itemQueue = make(chan itemData, 1024)
+var itemQueue = make(chan itemData, 10000024)
```
stuartmscott, line 3: Is this change related to form hint and error text?

widget/form.go
```diff
@@ -128,16 +132,50 @@ func (f *Form) checkValidation(err error) {
...
+			return item.Widget
+		}
+	}
+
+	text := canvas.NewText(item.HintText, theme.PlaceHolderColor())
+	text.TextSize = theme.CaptionTextSize()
+	text.Move(fyne.NewPos(theme.Padding()*2, theme.Padding()*-0.5))
+	item.helperOutput = text
+	f.updateHelperText(item)
+	return fyne.NewContainerWithLayout(layout.NewVBoxLayout(), item.Widget, fyne.NewContainerWithoutLayout(text))
```
stuartmscott, line 151: Why does text need to be wrapped in a container without a layout?

widget/form.go
```diff
@@ -128,16 +132,50 @@ func (f *Form) checkValidation(err error) {
 	f.submitButton.Enable()
 }
 
+func (f *Form) setUpHints(item *FormItem) fyne.CanvasObject {
```
stuartmscott, line 135: I'm not sure `setUpHints` fully conveys what this function does. It is returning either the form item widget, or a vertical box containing the form item widget and hint text. Essentially returning the input side of the form item. Perhaps `createInput` to match `createLabel` above?

widget/form_test.go
```diff
@@ -121,17 +121,39 @@ func TestForm_ChangeTheme(t *testing.T) {
...
 		OnSubmit: func() {}, OnCancel: func() {}}
 	w := test.NewWindow(form)
-	w.Resize(fyne.NewSize(340, 240))
 	defer w.Close()
 
 	test.AssertImageMatches(t, "form/theme_initial.png", w.Canvas().Capture())
 
 	test.WithTestTheme(t, func() {
 		form.Refresh()
+		w.Resize(form.MinSize().Add(fyne.NewSize(theme.Padding()*2, theme.Padding()*2)))
```
stuartmscott, line 130: Would it be better to use `w := test.NewWindow(layout.NewPaddedLayout(form))`?

stuartmscott (changes requested):

widget/form.go
andydotxyz: It was requested that it be moved closer to the element above than the standard padding amount. This was the easiest way to do it.

data/binding/queue.go
andydotxyz: Good catch sorry.

widget/form_test.go
andydotxyz: Unfortunately not - this is accounting for window padding. If you add more content it's still off by padding amount.

widget/form.go
stuartmscott: So the core issue is that VBox adds padding between subelements? I'm confused how making one of the elements a container without a layout removes this padding

widget/form_test.go
stuartmscott: Hmm, windows don't include this padding when ensuring they are big enough for their content?

widget/form.go
andydotxyz: Well, basically. All layouts add padding between elements. By using a container without layout I can control the position manually.

widget/form_test.go
andydotxyz: This is the crux of the matter with tests and internally-size-adjusting content. There is no "ensure the window is big enough for new content" outside of requesting it manually. Previously it was manual numbers, now it is calculated.

Jacalz (approved):
This is a big step. We seriously need a feature to turn off the validationErrors. Sparta is massive, but I think that it technically isn't a problem due to this PR just how I have things set up.
image

widget/form_test.go
stuartmscott: Ack, something for another PR

widget/form.go
```diff
@@ -134,10 +157,25 @@ func (f *Form) setUpValidation(widget fyne.CanvasObject, i int) {
 		w.SetOnValidationChanged(func(err error) {
```
fpabl0, line 157: @andydotxyz comment `f.Items[i].validationError = w.Validate()` to avoid showing error in the same time user enters to the form.

widget/form.go
stuartmscott: So Box adds padding between item.Widget and the container, but then the container positions its child at (0,-padding)?

widget/form.go
andydotxyz: I tried that but unfortunately I cannot. The value is only updated when the error changes, and that error will never change until the validation passes and then fails again.
So if we leave it off initially then it is off past the point that it is useful

andydotxyz: I would love a better solution to this - I just can't think of one!

widget/form.go
andydotxyz: yes

andydotxyz: The other approach is to write a custom layout which felt overkill for this 1-liner

widget/form.go
fpabl0: I am not understanding

fpabl0: https://user-images.githubusercontent.com/12239342/104376222-b5921280-54f2-11eb-89ae-781b62a45781.mp4

fpabl0: or are you talking about that the error is shown even when the underline is not red?

andydotxyz: Hold on maybe you have hit something I failed to solve (the red I can live with for now - that is harder to solve)

andydotxyz: Ah no I remember now. If you do that then the Submit button is enabled initially when it should not be

fpabl0: Hmm what if by default Submit is enabled when everything is empty. When user press Submit, Validate() on every form item is called, so Submit will be disabled at that time, and then it will work as always.

andydotxyz: That feels like a step backwards.

Can you share the diff for the code running above? I cannot figure out what you did to have it change on first key press

fpabl0: I just commented this line `f.Items[i].validationError = w.Validate()`

andydotxyz: When I do that it does not update until I have entered a valid email and then backed out again

fpabl0: That's weird.
Captura de Pantalla 2021-01-12 a la(s) 16 42 06

andydotxyz: Hmm, there is magic in `w.Validate()` I did not see before

fpabl0: Yes, w.Validate() launch errors in Validatable, then it should not be called at first time.

andydotxyz: I have a solution that works for the current button behaviour and hiding initially, just a sec

fpabl0: Yeah me too
Captura de Pantalla 2021-01-12 a la(s) 16 52 21

fpabl0: https://user-images.githubusercontent.com/12239342/104378994-ce042c00-54f6-11eb-9fbc-8bee8784d446.mp4

andydotxyz: I think your one could still be wrong if there were two validators and you just fixed the first one...
I am about to push a fix that should handle all cases - essentially the same trick though :)

fpabl0: needs optimization as that loop logic can be done in the previous f.Items loop, but that's the idea

fpabl0: > I think your one could still be wrong if there were two validators and you just fixed the first one...

I don't understand you 😅

> I am about to push a fix that should handle all cases - essentially the same trick though :)

Great! 😄

andydotxyz: (the problem with only calling SetValidationError(nil) and not caching the previous result was that the callback would set it to nil - with your workaround line at the end the initial state is correct, but further checks will get the wrong nil values. This meant that it was OK when you were editing the only validated field but would give a false positive if you fixed the first of many).

Anyhow - happy with this result?

fpabl0: > but further checks will get the wrong nil values. 

I don't think so, maybe I am wrong. When I initially validate and set error to nil, nothing changes (and it is called just once when Renderer is created) as validation is trigger on text changed, right?

fpabl0: Ignore my last comment, yes it is more flexible, the way you do 👍
I think now it looks good, I don't think it needs more changes, do it?

andydotxyz: I think we are good, thanks :)

fpabl0: You're welcome. I'm glad to know that I can help with something. 👍

stuartmscott (approved):

Issue #1028, opened 2020-05-23
Race conditions related to WidgetRenderer and cache


andydotxyz:
We need to review how renderers are invoked and executed.
It is currently racey, but the design does not require it.
A single thread or lock could be used to control access to renderers

andydotxyz:
Dropped into 1.3.x but this may be a 2.0 issue

Bluebugs:
After recent discussion about race between the rendering code and the developer use of Refresh, I think the best strategy is to:
- Add an internal state that is a copy of the public field
- Add a lock
- Take the lock and copy the public state in the internal state on Refresh
- Take that when rendering

This stop race condition between rendering and call to Refresh. It won't prevent race condition in the developer code changing fields in multiple goroutine and calling Refresh in parallel, but there is no way for the toolkit to know which of those racing goroutine should actually win. It should be the duty of the developer to synchronize those goroutine an example of such pattern would be how data binding does all its change in one goroutine and this should be documented.

This proposal means no API break and can be rolled out in 2.2.

dweymouth:
I was reading through the discussions on thread safety and I think that with the current API design it is impossible to make access to some public fields free of race conditions - though the chance that such race conditions can occur can be made "negligible". All public string fields for instance cannot even be copied in a race-free way since a string is two machine words long (a pointer and a size). However for the race to occur a user goroutine would need to assign to the field in between when the length and pointer values of the string were copied.

It may be worth thinking about deprecating public fields (at least string fields) on widgets/CanvasObjects and forcing all user access to go through getters/setters. Unfortunately it makes the API a little verbose and is a large breaking change (since the way extended widgets are initialized would need to change). Another option would be to add getters and setters for all the public fields but leave them exported, with a comment specifying that they should only be directly accessed on widget construction. Though we are definitely OK with deferring these considerations for quite awhile as the chance that this kind of data race would occur is very very small.

andydotxyz:
No races with the new render thread model.

PR #1773, opened 2021-01-12
Fix Select appearance

fpabl0:
### Description:

Fix Select widget appearance to be similar to SelectEntry.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


fpabl0:
Before updating tests, would like to know if this is the desired appearance:
Captura de Pantalla 2021-01-12 a la(s) 15 29 43
Captura de Pantalla 2021-01-12 a la(s) 15 29 51

andydotxyz:
Oh I missed this sorry.
Direction seems good - it is an input and not a button as you say.
With the new colours (pending in #1782) it will be more distinct from a large primary button as well

fpabl0:
So should I wait until #1782 is landed?

andydotxyz:
> So should I wait until #1782 is landed?

I don't think you need to wait. There will be a merge conflict either way and we will get them both in at some point in the next few days

fpabl0:
I have just merged develop into fix/select branch, and I found widget.Entry size changed a lot, isn't its height too short?
Captura de Pantalla 2021-01-14 a la(s) 02 10 09

andydotxyz:
> I have just merged develop into fix/select branch, and I found entry widgets size changed a lot, isn't its height too short?

Wow, good catch, it was off by 2pt. The top edge is inset to match the bottom line thickness so that the text appears centrally - is that wrong? It was requested when we were reviewing the Entry design. (but I accidentally intended by Padding size instead of Underline size)

AlbinoGeek:
> > I have just merged develop into fix/select branch, and I found entry widgets size changed a lot, isn't its height too short?
> 
> Wow, good catch, it was off by 2pt. The top edge is inset to match the bottom line thickness so that the text appears centrally - is that wrong? It was requested when we were reviewing the Entry design. (but I accidentally intended by Padding size instead of Underline size)

If this was my request, it was actually that the text move up by that amount (2dp for example) -- So, to use fpabl0's screenshot as example, the word "Option" would be in the vertical middle if it were moved up 1dp.

image

Where the Label is drawn in the vertical middle as per the green line marked `c`

[ignoring the dp dimensions in the spec graphic, because we chose to adapt those to fyne.]

andydotxyz:
Unfortunately "just move it up" impacts every single widget now that we have consistent alignment.
I am not keen to change the baseline for all widgets so that Entry can be taller TBH.

AlbinoGeek:
Gotcha, so the direction this PR should go is making the Select smaller to match the new Entry then, yeah?

andydotxyz:
> Gotcha, so the direction this PR should go is making the Select smaller to match the new Entry then, yeah?

Indeed. The existing select is much smaller than the one in that screenshot I think. I am trying to get all our single line widgets to the same height so that alignment "just works".

fpabl0:
@andydotxyz 
> The top edge is inset to match the bottom line thickness so that the text appears centrally - is that wrong?

Centrally Text looks good 👍, however I think it is too thin, but for now I think is ok.

> how is your Select 74px tall, and Entry 62px !?
Do you have scaling other than 1.0 or a hidpi monitor?

@AlbinoGeek Hmm I am not sure, maybe I am implementing the MinSize function wrongly? I am using `s.label.MinSize()` instead of `fyne.MeasureText(s.combo.PlaceHolder, theme.TextSize(), s.combo.textStyle())`, is that wrong? BTW, now I am curious, and I really don't understand why there are 2 MinSize methods to implement (one of base widget and one of the renderer), how are they different? @andydotxyz 

> The existing select is much smaller than the one in that screenshot I think. I am trying to get all our single line widgets to the same height so that alignment "just works".

Got it, I will work on that tonight.

andydotxyz:
> Got it, I will work on that tonight.

Cool. Be sure to pull latest develop before starting as we had 2 or 3 big merges again today.

fpabl0:
> Cool. Be sure to pull latest develop before starting as we had 2 or 3 big merges again today.

👍. After looking carefully both Select and SelectEntry, they actually are the same height, it is just optical illusion. In DarkMode it looks that they are different size, but that is because the underline has a color similar to the background and makes the optical illusion it is smaller. If we see the screenshot in light mode, it is a little easier to see they are the same height:

Captura de Pantalla 2021-01-15 a la(s) 03 54 29

I have compared Select and SelectEntry MinSize values and both of them give me the same height (37).

andydotxyz:
> Where the Label is drawn in the vertical middle as per the green line marked `c`

Out of curiosity I just measured their 'c' line and it is not dead-centre, it is off by 1 just as our Entry is 😎 .
Also our entry padding looks to be the same, so we can probably use Entry as a good guide now when comparing other elements.

andydotxyz:
> After looking carefully both Select and SelectEntry, they actually are the same height, it is just optical illusion

@fpabl0 would it be better if the background was brought in on the top and bottom by the underline thickness? 
That might look less confusing to the casual observer.

andydotxyz:
Focus colour change is now in as well, so one last rebase onto develop and you should not have to worry further about theme changes.
Sorry again! This has been a fast moving branch for a couple of weeks now.

fpabl0:
> Out of curiosity I just measured their 'c' line and it is not dead-centre, it is off by 1 just as our Entry is 😎 .

Great!! 😄. So we are on the right track !

> would it be better if the background was brought in on the top and bottom by the underline thickness?

Maybe, but that will lead to some problems, because it will look good in a vertical alignment, but in a horizontal alignment, Select will look thinner than the SelectEntry (as it actually will be). Btw, I am going to try and post the screenshots here to see how it looks.

> Focus colour change is now in as well, so one last rebase onto develop and you should not have to worry further about theme changes.
Sorry again! This has been a fast moving branch for a couple of weeks now.

No worries, I know it has been busy days for you all, but it has great results. Thanks for this toolkit! 👍

fpabl0:
So after those changes, I got these images:
1. This image shows the Select padding and comparing with SelectEntry (Horizontal alignment):
Captura de Pantalla 2021-01-16 a la(s) 02 38 20

2. Light and dark mode (Horizontal alignment):
Captura de Pantalla 2021-01-16 a la(s) 02 40 19
Captura de Pantalla 2021-01-16 a la(s) 02 39 29

3. Light and dark mode (Vertical alignment):
Captura de Pantalla 2021-01-16 a la(s) 02 41 44
Captura de Pantalla 2021-01-16 a la(s) 02 41 57

andydotxyz:
Thanks @fpabl0 I see what you mean.
I guess the other option is that we avoid the insetting - the select goes back to full height and the entry is, with the underline drawn just inside.
This could lead to the text appearing off-center by a small amount but it's worth looking at as an option.
I think @AlbinoGeek first noted off-center issue, but that was probably when the underline was twice the size - perhaps he has thoughts here too?

AlbinoGeek:
I think this is one of those moments we just "do what we can."

In MD Spec, the Text Field is the only one with the underline,

Which leads to this inevitable sizing issue, because `fyne`'s layouts are based on `MinSize`, and we chose not to do the "everything must have `48dp` Minimum Height" as described in the MD Text Field. As a result, elements end up smaller (such as `Select`, etc.)

So... whatever makes them align the best, should be what we go with.

andydotxyz:
@AlbinoGeek I don't see how using a different "same height for all" number makes this more problematic?

If select and selecteentry are to be the same height (as MD would agree) and yet one of them has the underline, then one is going to look off-center if you wish them to have the same alignment.
Surely that is unavoidable?

AlbinoGeek:
> Surely that is unavoidable?

Yeah, this is what we'll have to work with.

I like what we're seeing in the last screen shots, and as long as they look okay next to a button, I feel is good to go with.

fpabl0:
@andydotxyz @AlbinoGeek I have setup a flutter project, just to see how they manage this part and this is what I got:

Captura de Pantalla 2021-01-16 a la(s) 17 31 35

Select and Entry look the same, the only difference is the dropdown arrow. 

https://user-images.githubusercontent.com/12239342/104824575-e9d23f80-5820-11eb-9ceb-e58f4642fcca.mp4

Select does not get focus color in flutter, what do you think?

AlbinoGeek:
We could follow Material Design Lite's interpretation and use the underline anyways (just like you saw in Flutter.)

fpabl0:
Yes, I think that should be the best choice, what do you think @andydotxyz ?

fpabl0:
Just verify, Flutter's select do have focus color (it just that the dropdown menu cover it all):

Captura de Pantalla 2021-01-16 a la(s) 17 48 40

andydotxyz:
Sounds fine to me, if you are happy that the editable and non-editable selects look identical
I wonder if we should use the focus colour as a background if a select gets "input focus" like it does at the moment, so when interacting it becomes clear that they are differen?

fpabl0:
> Sounds fine to me, if you are happy that the editable and non-editable selects look identical
> I wonder if we should use the focus colour as a background if a select gets "input focus" like it does at the moment, so when interacting it becomes clear that they are differen?

Hmm tbh, I think non-editable select is much more common than an editable one (i would prefer a filtered one instead of an editable). Yeah, we can use focus color as background however what would happen with the underline color? Should it be primary color? So, when Select is focus its background should be focus color and the underline primary color? I will look weird, I think. Or both background and underline get the same color then it will look as a solid body (just like before) only when it is focused? I think this last option could work

andydotxyz:
> Or both background and underline get the same color then it will look as a solid body (just like before) only when it is focused? I think this last option could work

this sounds like a plan - let’s see how that looks 👍

fpabl0:
@andydotxyz Here are all the options:

1. Bg = FocusColor & Line = FocusColor

https://user-images.githubusercontent.com/12239342/104827607-6e2abf80-582d-11eb-8b17-74d27f19a1f3.mp4

2. Bg = FocusColor & Line = PrimaryColor

https://user-images.githubusercontent.com/12239342/104827615-94e8f600-582d-11eb-95bd-497a4667c56d.mp4

3. Bg = FocusColor & Line = ShadowColor

https://user-images.githubusercontent.com/12239342/104827618-a6ca9900-582d-11eb-8657-0850cffc4873.mp4

4. Bg = InputBackgroundColor & Line = PrimaryColor

https://user-images.githubusercontent.com/12239342/104827622-b649e200-582d-11eb-90d0-8b76307cde37.mp4

AlbinoGeek:
Personally I like (4) -- would love to know what others feel.

fpabl0:
Should it be discussed on Slack to have more feedback?

andydotxyz:
After looking at the illustrations above I realise the underline thickness offset at the top looks uneven in somelayouts, so I'll back that out of entry and we can mirror it here.
Given our timeline I think that we either pick a soltuion and go with it today, or move the decision out for a future release.

fpabl0:
@andydotxyz  Ok, so I think we would go either with 1 or 2, as they are similar to the current look. Later, we can discuss the final appearance.

Which one is better 1 or 2? @AlbinoGeek 

> move the decision out for a future release.

I think it is not viable, because current Select looks weird comparing with others input widgets.

andydotxyz:
> I think it is not viable,

It was agreed on the community call that this is not a blocking issue, so it's down to timing I guess. 
Code is scheduled to freeze in 1 day.

fpabl0:
@andydotxyz Ok, so we should wait to 2.1.0 then

andydotxyz:
> @andydotxyz Ok, so we should wait to 2.1.0 then

Following chat on Slack we agreed it might still be possible for 2.0...

widget/select.go
```diff
@@ -333,17 +336,20 @@ func (s *selectRenderer) MinSize() fyne.Size {
 	s.combo.propertyLock.RLock()
 	defer s.combo.propertyLock.RUnlock()
 
-	min := fyne.MeasureText(s.combo.PlaceHolder, theme.TextSize(), s.combo.textStyle())
-
-	min = min.Add(fyne.NewSize(theme.Padding()*6, theme.Padding()*4))
-	return min.Add(fyne.NewSize(theme.IconInlineSize()+theme.Padding(), 0))
+	// x1			      x4 (reserved)          x1    x1     x2
+	// pad  -     label (placeholder)  -  pad - icon - pad*2
```
andydotxyz, line 340: I’m not sure what these comments mean - should they be removed?

widget/select.go
```diff
@@ -295,36 +295,39 @@ func (s *Select) updateSelected(text string) {
...
 
-	s.background.Move(fyne.NewPos(theme.Padding(), theme.Padding()))
-	s.background.Resize(size.Subtract(fyne.NewSize(doublePad, doublePad)))
-
-	offset := fyne.NewSize(theme.IconInlineSize(), 0)
-	labelSize := inner.Subtract(offset)
+	iconPos := fyne.NewPos(size.Width-theme.IconInlineSize()-theme.Padding()*2, (size.Height-theme.IconInlineSize())/2)
+	labelSize := fyne.NewSize(iconPos.X-theme.Padding(), s.label.MinSize().Height)
 
+	// TODO are these needed?
```
andydotxyz, line 322: Is there something additional to do here?

andydotxyz (changes requested):

widget/select.go
fpabl0: I can removed, that comment is just to show how the MinSize is build, but it does not have any importance

widget/select.go
fpabl0: Yes, I would like to know why is needed to use mutex lock in this part (line 323):

```
         s.combo.propertyLock.RLock()
	defer s.combo.propertyLock.RUnlock()

	s.label.Resize(labelSize)
	s.label.Move(fyne.NewPos(doublePad, theme.Padding()))
	s.label.Move(fyne.NewPos(theme.Padding(), (size.Height-labelSize.Height)/2))
```

andydotxyz: We went through a process of hunting race conditions a while back. I cannot immediately see what this was protecting, perhaps “test -race” will show something extra if they are removed? If not then probably it is historic.

fpabl0: I have just run `go test -race ./...` and there are a lot of race conditions, is that expected?

fpabl0: Btw, I have run now `go test -race ./... | grep Select` to filter just Select related tests (after removing the mutex lock) and everything seems to be ok, so I think it is save to remove that piece of code.

andydotxyz: > I have just run `go test -race ./...` and there are a lot of race conditions, is that expected?

yes, the widget renderer access had to be disabled (in 2.0) before I can fix another class of race condition (there is a ticket to track these #1028)

andydotxyz (approved):
This makes form look lots tidier thanks!

widget/select.go
fpabl0: Oh ok great! :)

PR #1779, opened 2021-01-13
Fix Cursor Animation on widget.Entry

### LINKED ISSUE #1778: Cursor anim. sometimes distorts the text
### Describe the bug:

This bug seems to be sporadic and related  to the cursor animation, because it appeared when animation was added.

### To Reproduce:
1. Run fyne_demo.
2. Go to any tab with an Entry.
3. Write and delete quickly.

### Screenshots:

https://user-images.githubusercontent.com/12239342/104426037-662dff80-554f-11eb-9db5-c5b18b26a7a3.mp4

https://user-images.githubusercontent.com/12239342/104426055-6b8b4a00-554f-11eb-92b3-04957106917f.mp4

 - **OS:** MacOS 
 - **Go version:** 1.15.6 
 - **Fyne version:** develop @ ee882d72

fpabl0:
### Description:

Fixes #1778 

It seems that #1778 appears, because fade animation is running when user is typing. 
Cursor should be static when user is interacting with the Entry (as normally cursor animation).
Implementation in this PR could be improved, but that's the idea basically.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


fpabl0:
This is how the cursor is shown with this PR:

https://user-images.githubusercontent.com/12239342/104494896-11b56f00-55a5-11eb-8af3-9e50aa2dbf9d.mp4

AlbinoGeek:
This is a pleasing change, as this makes the cursor consistent with the OS cursor in Gnome (which will not flash or fade while the user is actively typing, although it has a much shorter pause than your video does.) (Also, it's a little jumpy, but again, I get what you're trying to do, and like the direction.)

fpabl0:
@AlbinoGeek 
> although it has a much shorter pause than your video does.

Currently, it pauses 500 ms, should it be less?

> it's a little jumpy

What do you mean?

fpabl0:
@AlbinoGeek I have decreased the pause time, what do you think now?

https://user-images.githubusercontent.com/12239342/104698285-1e3de280-56df-11eb-9082-5941bc6cdf37.mp4

Testing this feature, I think there is a bug in cursor animation, even without the changes in this draft:

https://user-images.githubusercontent.com/12239342/104698349-3d3c7480-56df-11eb-87dc-0e1ebe696e2f.mp4

It seems that the problem is in the fyne.Animation implementation (it seems that the animation time is shared maybe, and when cursor animation start, sometimes its defined interval is almost done, so first 'animation' is very short). What do you think @andydotxyz ?

andydotxyz:
> sometimes its defined interval is almost done, so first 'animation' is very short). What do you think @andydotxyz ?

It's possible depending on how long the scheduler takes to kick it off.
The end time of the duration is calculated when the animation starts, maybe there is a delay somehow.

You can see in `internal/animation/animation.go` line 19, each animation sets it's own end time and ticks towards it on a standard FPS refresh.

fpabl0:
@andydotxyz I have seen the animation implementation, and I think first `start` and `end` parameter values should be set just before tickAnimation (maybe checking if start and end are equal to 0, then set their values), otherwise there is a chance that user requests animation the tick has just passed, so it's animation will run after 1/60 seconds (but Fyne has already set the start and end fields), so there is 1/60 seconds less for the first animation step. What do you think? Am I missing something?

andydotxyz:
> What do you think? Am I missing something?

Your summary is correct, we could miss the first frame of animation. But is that going to lead to visual issues such as you have reported? The cursor should fade out over half a second and back in over another half, meaning that the first frame is a 60th of the animation, or 30th of the fade. Maybe there is another issue at play?

fpabl0:
> But is that going to lead to visual issues such as you have reported?

You are right. It isn't the root of the problem. In fact, that doesn't even make it better. Not sure, what it is.
Btw, do you like this change? I have inverted the cursor animation to make a smooth transition between a pause and the animation, however that breaks some tests - some of them expects the cursor is "visible" but now the cursor is not, I think that's because I inverted the animation:

Captura de Pantalla 2021-01-17 a la(s) 00 26 27

Captura de Pantalla 2021-01-17 a la(s) 00 26 54

Is it safe to replace them?

andydotxyz:
>  I have inverted the cursor animation to make a smooth transition between a pause and the animation

It was coded originally to fade it in. With the inversion you added it appears immediately then fades out.
The test driver just runs a frame of animation at 1.0, i.e. the last frame, so that is why the images changed.

I'm not sure why this improves the pause code, perhaps there is something else at play?
Instead of writing lots of code in entry would it make more sense for us to add `Animation.[Un]Pause()`?
It sounds like we might not get to a complete solution for tomorrow's code freeze, but as it's not a blocker that should not be a problem.

andydotxyz:
I have fixed the initial state of the cursor, and removed some code that was accidentally overriding the colour on refresh.
This should fix the issue you saw with the animation being sort. On `develop` branch now.

fpabl0:
> This should fix the issue you saw with the animation being sort.

Yes, that fixed it. Thanks :)

> I'm not sure why this improves the pause code, perhaps there is something else at play?

The problem is that if I don't invert it, it looks like a jumpy transition, see:

https://user-images.githubusercontent.com/12239342/104853063-70902680-58cc-11eb-80fd-3f3f7b46db43.mp4

Inverted animation looks smooth instead:

https://user-images.githubusercontent.com/12239342/104853069-81d93300-58cc-11eb-9948-b36859f465f9.mp4

> Instead of writing lots of code in entry would it make more sense for us to add Animation.[Un]Pause()?

Hmm I am not sure, the pause in this PR is temporary so after some time the animation starts without any action. Animation.pause will do the same as the current Animation.stop, won't it?

andydotxyz:
> Animation.pause would do the same as the current Animation.stop, wouldn't it?

Not really. Stop resets the animation so start runs from the beginning. Pause would remember the current state so it could be resumed (at least that’s what was in my head when I wrote the message).

It seems that the animation you want to start is from 1.0 but on entering we want it to start from 0.0. Maybe stopping the enter animation and starting a new one that is the opposite upon the timeout lapsing would solve this without changing the current animation / tests?

I think that the current code could be simplified by doing this - and we should be able to avoid the specific ticking code as well I would think.

andydotxyz:
> I don't understand well how is the addition of Animation.pause would simplify current code, as we need to restart the animation from the beginning.

I assumed that the complicated code in this PR was to manage the pause / resume? It’s almost a custom animation ticker in entry_cursor.go

fpabl0:
> I assumed that the complicated code in this PR was to manage the pause / resume? It’s almost a custom animation ticker in entry_cursor_anim.go

Yes, but it is not a common pause, it is a temporary pause that will auto-start the animation from the beginning after some time (not from where it was when it was paused). That's why I think Pause/Resume functions in fyne.Animation would not simplify this code as it is different? Although I think Pause/Resume function would be a great addition :), a function to know check the animation status would be great too!

andydotxyz:
> Yes, but it is not a common pause, it is a temporary pause that will auto-start the animation from the beginning after some time (not from where it was when it was paused). That's why I think Pause/Resume functions in fyne.Animation would not simplify this code as it is different?

I understand this - but surely something like pausing it when typing starts, then after the time has lapsed unpause would help? The only custom code you should need is determine when to call `UnPause()`...

fpabl0:
@andydotxyz 
> I understand this - but surely something like pausing it when typing starts, then after the time has lapsed unpause would help? The only custom code you should need is determine when to call UnPause()...

To be honest, I don't understand what do you mean. 
This is the function `TemporaryPause`:

```
func (a *entryCursorAnimation) TemporaryPause() {
	a.counter.Reset()
	a.mu.Lock()
	defer a.mu.Unlock()
	if a.anim == nil {
		return
	}
	a.anim.Stop() // 👉 here, you say it could be replaced by a.anim.Pause(), but I am not sure what is the difference?
	if !a.inverted {
		a.anim = a.createAnim(true)
	}
	if a.paused {
		return
	}
	a.paused = true
	a.cursor.FillColor = theme.PrimaryColor()
	a.cursor.Refresh()
}
```
And then, the animation is restarted here:

```
a.counter.Inc(1)
if a.counter.Value() == cursorPauseTimex10ms {
	// >>>> lock
	a.mu.Lock()
	if a.anim != nil {
		a.anim.Start() // 👉 here, you say could be replaced by  a.anim.Unpause(), but it just a different name, isn't it?
	}
	a.paused = false
	a.mu.Unlock()
	// <<<< unlock
}
```

andydotxyz:
Difference between stop and pause:

Animation is running from 0.0 to 1.0, at 0.5 we want to interrupt it:
* Calling Stop at this point will halt the animation and reset it's values
* Calling Pause would (when implemented) halt the animation but have no further effect

Then when we want to restart it, we call the counterpart:
* Start will begin the animation exactly as before, starting at 0.0
* Unpause would resume the animation where it had paused, at 0.5 (or the frame right after it)

This is, in effect, the same as the difference between stop and pause when watchng a movie or listening to music - pause is a temporary interruption that does not reset your progress.

fpabl0:
@andydotxyz  Yes, that's right, but here we want to restart always from 0.0 (not from it was paused). Maybe `TemporaryPause` is not a good name, it should be `TemporaryStop` as it effectively stops the animation and reset its values, what do you think?

widget/entry_cursor_anim.go
```diff
@@ -0,0 +1,156 @@
...
+		a.cursor.FillColor = c
+		a.cursor.Refresh()
+	})
+	anim.RepeatCount = fyne.AnimationRepeatForever
+	anim.AutoReverse = true
+	return anim
+}
+
+// Start starts cursor animation.
+func (a *entryCursorAnimation) Start() {
```
andydotxyz, line 83: If we're not exporting the type it's best to not look like the methods are exported.

widget/entry_cursor_anim.go
```diff
@@ -0,0 +1,156 @@
...
+	a.state = cursorStateRunning
+	go func() {
+		defer func() {
+			a.mu.Lock()
+			a.state = cursorStateStopped
+			a.mu.Unlock()
+		}()
+		for {
+			a.sleepFn(10 * time.Millisecond)
+			// >>>> lock
```
andydotxyz, line 100: When we see comments inline it can indicate that the code is too complex and should be broken out to new methods.
However in this case it annotates a line that has the same words in it, so seems redundant

widget/entry_cursor_anim.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	a.anim = a.createAnim(false)
+	a.state = cursorStateRunning
+	go func() {
+		defer func() {
+			a.mu.Lock()
+			a.state = cursorStateStopped
+			a.mu.Unlock()
+		}()
+		for {
```
andydotxyz, line 98: Looking at the counter and the loop and I wondered if an alternative approach might be easier to read (and less background processing)... this is just thinking out loud.

Could you wait until you suspect the cursor would wake up (i.e. simple delay), then at that event see if the time has lapsed, and if not (because a new delay was added) then wait the difference between the first delay and the latter, eventually seeing there is no additional delay and return to the running animation? 

This would also mean that the animation wakes up after the specified time, without needing to check every 10ms

widget/entry_cursor_anim.go
```diff
@@ -0,0 +1,156 @@
...
+			a.mu.RUnlock()
+			// <<<< unlock
+			if cancel {
+				return
+			}
+			if !tempStop {
+				continue
+			}
+			a.counter.Inc(1)
+			if a.counter.Value() == cursorStopTimex10ms {
```
andydotxyz, line 113: You can reduce complexity by "exiting early" so the main business of a function is less indented. i.e.

```go
if a.counter.Value() != cursorStopTimex10ms {
    continue
}
```

widget/entry_cursor_anim.go
```diff
@@ -0,0 +1,156 @@
...
+	c.cnt = 0
+}
+
+const cursorStopTimex10ms = 35 // stop time in multiple of 10 ms
+
+type cursorState int
+
+const (
+	cursorStateRunning cursorState = iota
+	cursorStateTemporarilyStopped
```
andydotxyz, line 42: It could be shorter as `cursorStatePaused` :)

widget/entry_cursor_anim.go
```diff
@@ -0,0 +1,156 @@
...
+	"time"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/canvas"
+	"fyne.io/fyne/v2/theme"
+)
+
+type safeCounter struct {
+	mu  sync.RWMutex
+	cnt int
```
andydotxyz, line 15: Doesn't `cnt` sound bad?

andydotxyz (changes requested):

widget/entry_cursor_anim.go
fpabl0: Oh ok, you are right, I will rename it

widget/entry_cursor_anim.go
fpabl0: Yes, should be. However, it isn't really paused, it is stopped. I had that name before (paused) , but since the discussion about `TemporaryPause` function, I think it is better to call it `TemporarilyStopped` (to avoid confusions) because when the animation start again, it will not be restarted from the previous state, it will be restarted from the beginning.

widget/entry_cursor_anim.go
fpabl0: Yes, that was the first thing I had in mind, but the problem is the reset part, if we do not check every "x" time then we will introduce some delays in the animation, or maybe I am wrong?

widget/entry_cursor_anim.go
fpabl0: Yes, my fault, I will remove them.

widget/entry_cursor_anim.go
fpabl0: Yes, could be, I will try :)

widget/entry_cursor_anim.go
fpabl0: Yeah, I will rename it to `val`, is it better?

widget/entry_cursor_anim.go
andydotxyz: Fair point. perhaps "Interrupted" or something that is not Paused but also is not re-using Stopped?

widget/entry_cursor_anim.go
andydotxyz: Perhaps I don't follow but my understanding was that it would always pause at least the delay time, which may get longer if the user keeps typing. So what is the reset event that we need to check sooner?

widget/entry_cursor_anim.go
fpabl0: Yes, that's a good name, I will update!

fpabl0: Should I also rename `TemporaryStop` to `Interrupt`?

widget/entry_cursor_anim.go
fpabl0: Hmm ok, I will implement the logic with a ticker instead maybe it is more clear with it?

andydotxyz: I would have thought it would be. Worth exploring.
If it is cleaner without let me know and I will do a more thorough review of this code

fpabl0: @andydotxyz  Ok, I have updated the code using now time.Ticker. I am just worry if calling ticker.Reset() so quickly would lead to a bad performance.

fpabl0: However, now I realized that time.Ticker does not have Reset() method in Go 1.12 😞

fpabl0: @andydotxyz  I have rewritten the solution with time.Timer instead of time.Ticker to ensure backwards compatibility with Go. Tests are ok now! :)

andydotxyz (changes requested):
Thanks for looking into this further.
I can't quite understand this code unfortunately - why does it need 2 new types, cursorTicker and realTicker (why is `time.After` not sufficient)?
Also, I don't think you have to keep replacing `a.anim`, just `Start()` and `Stop()` should be enough.

fpabl0:
@andydotxyz 

> why does it need 2 new types, cursorTicker and realTicker

There aren't really needed, they just have sense for testing, otherwise it would be difficult to test (as you can see in entry_cursor_anim_test.go there is a type `mockTicker` that implements cursorTicker for testing).

> why is time.After not sufficient

Because if I use it, it will not be efficient, as the timer should be stopped when user types, and time.After does not give us that functionality. Quoting golang time package documentation:

`"The underlying Timer is not recovered by the garbage collector until the timer fires. If efficiency is a concern, use NewTimer instead and call Timer.Stop if the timer is no longer needed."`

> Also, I don't think you have to keep replacing a.anim, just Start() and Stop() should be enough.

Yes, I though it too, but in this way is how the cursor animation is implemented currently in entry.go: 

```
	if focused {
		r.cursor.Show()
		if r.cursorAnim == nil {
			r.cursorAnim = makeCursorAnimation(r.cursor)
			r.cursorAnim.Start()
		}
	} else {
		if r.cursorAnim != nil {
			r.cursorAnim.Stop()
			r.cursorAnim = nil
		}
		r.cursor.Hide()
	}
```

And that's maybe because Focus event is triggered on every click and that could lead to multiple calls to Start when animation is already started.

andydotxyz:
> Yes, I though it too, but in this way is how the cursor animation is implemented currently in entry.go:

That is to handle the situation when the widget loses focus so we don't want the animation any more.

> Also anyway, I have to replace anim to ensure that the cursor fade starts as it is now

Calling `Start()` on an animation will always start it from the beginning, it does not have to have been newly created.

fpabl0:
@andydotxyz 
> That is to handle the situation when the widget loses focus so we don't want the animation any more.

I know and that is what I am doing. If you look `entry.go`, you will see this:
```
	if focused {
		r.cursor.Show()
		r.content.entry.cursorAnim.start()
	} else {
		r.content.entry.cursorAnim.stop()
		r.cursor.Hide()
	}
```
That is equivalent to the current implementation. I don't understand what difference you are trying to tell me.

> Calling Start() on an animation will always start it from the beginning, it does not have to have been newly created.

I know, I am not creating it again and again, I am just creating the first one (beginning from dim to opaque) and then when user starts typing create a second one (beginning from opaque to dim). The second one is kept until user unfocus the Entry. The first one is needed to avoid changing the test images. And the second one is needed because if I use the first one (dim to opaque) the cursor will look jumpy in the transition.

andydotxyz:
> That is equivalent to the current implementation. Sorry, but I don't understand what difference you are trying to tell me.

Yes I think I misread it sorry, it is creating only as many animations as are needed.

> There aren't really needed, they just have sense for testing, otherwise it would be difficult to test (as you can see in entry_cursor_anim_test.go there is a type mockTicker that implements cursorTicker for testing).

The way that animations work in test at the moment is that they always tick the final element of the animation - you can tick them at other positions if required for testing. The more fine grained animation test approach you have made probably suits better in our overall test support if it's needed.

i.e. `Animation.Tick(0.5)` marks it's half-time point.

fpabl0:
> The way that animations work in test at the moment is that they always tick the final element of the animation - you can tick them at other positions if required for testing. The more fine grained animation test approach you have made probably suits better in our overall test support if it's needed.

I see, but I am not testing specifically the fyne.Animation, I am testing the cursor animation that has its own custom go routine.

andydotxyz:
> I am testing the cursor animation that has its own custom go routine.

Maybe this is where I am lost again. I thought it was a regular animation but with a delay so it's not always running.
It seems like this could be tested without the new types?

fpabl0:
>  I thought it was a regular animation but with a delay so it's not always running.

Yes, it is. entryCursorAnim is only managing the fyne.Animation, when it should be in a running, interrupted or stopped state. For Interrupted state it is necessary something like a timeout to restart the animation automatically after some time. The new types are useful to test this situation (InterruptedState). 

> It seems like this could be tested without the new types?

Yes, it is possible, however it will introduce delays in tests, and I think it is undesired, am I wrong?

andydotxyz:
> Yes, it is possible, however it will introduce delays in tests, and I think it is undesired, am I wrong?

I don't really understand why it has to add delays if you remove the abstraction - but I think this relates to your tests wishing to test every tick of animation frame, which is why I suggested it may relate to the wider how-do-we-test-animations question. In the main toolkit when animations are concerned we test the start (before it is started) and end (what happens immediately after start) because that is how the test driver handles animations at this time.

I am not sure what your tests are trying to solve that cannot take the same simplified approach.

fpabl0:
> but I think this relates to your tests wishing to test every tick of animation frame

Sorry, I don't understand how my tests do what you say. My tests are just testing that the internal fyne.Animation is effectively stopped, when `TemporaryStop` is called, and automatically restarted after `cursorInterruptTime`. 

I have updated the way tests are performed. cursorTicker interface and mockTicker types were removed. cursorTicker now is the previous `realTicker`. CursorTicker is needed because time.Ticker does not support Reset() function before Go 1.15.

andydotxyz (changes requested):

widget/entry_cursor_anim.go
```diff
@@ -0,0 +1,197 @@
...
+	"fyne.io/fyne/v2/canvas"
+	"fyne.io/fyne/v2/theme"
+)
+
+// ===============================================================
+// Cursor ticker
+// ===============================================================
+
+type cursorTicker struct {
+	tim      *time.Timer
```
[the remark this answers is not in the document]
fpabl0: Got it.

widget/entry_cursor_anim.go
```diff
@@ -0,0 +1,197 @@
...
+	rstCh    chan struct{}
+	duration time.Duration
+	mockWait func() (reset bool)
+}
+
+func (t *cursorTicker) Start() {
```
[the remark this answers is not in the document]
fpabl0: Yes, my mistake, I will fix it.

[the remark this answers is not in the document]
fpabl0: > Also the ordering - Start should come after Reset to keep it alphabetical

Oh ok, I thought it would be better to order the functions according to the order they are intended to be called.

widget/entry_cursor_anim.go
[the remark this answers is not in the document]
fpabl0: done

widget/entry_cursor_anim.go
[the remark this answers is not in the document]
fpabl0: done

widget/entry_cursor_anim.go
```diff
@@ -0,0 +1,197 @@
...
+	if !t.tim.Stop() {
+		<-t.tim.C
+	}
+	t.tim = nil // TODO is it safe?
+	t.rstCh = nil
+}
+
+func (t *cursorTicker) WaitTick() (reset bool) {
+	if !t.Started() {
+		reset = true // TODO what to do here?
```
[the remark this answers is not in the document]
fpabl0: done

widget/entry_cursor_anim.go
```diff
@@ -0,0 +1,197 @@
...
+		reset = true
+		if !t.tim.Stop() {
+			<-t.tim.C
+		}
+	}
+	t.tim.Reset(t.duration)
+	return
+}
+
+func (t *cursorTicker) Started() bool { return t.tim != nil }
```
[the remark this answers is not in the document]
fpabl0: done

stuartmscott:
Would I be right in condensing this PR into the following?;

`Only blink the cursor if the Entry has focus, but the user has not interacted with it for X ms`

fpabl0:
Yes, basically that is the goal of this PR.

stuartmscott:
Hmmm, then I think it could be simplified;

1. Cursor is visible and Animation is started when Focus is gained
2. Cursor is hidden and Animation is stopped when Focus is lost
3. Whenever user inputs data into Entry (eg. TypedRune) record `time.Now()` as `lastInputTime`
4. On each Animation tick compare `time.Now()` with `lastInputTime`; if less that `threshold` (X ms) make Cursor fully opaque, else calculate Cursor colour as before.

fpabl0:
> Hmmm, then I think it could be simplified;

Yes, that's a possibility (Andy suggested a similar idea), the only problem with it, is that it will restart the blink in an unknown animation frame that will cause a jumpy animation. However, I think it is possible to nullify it by waiting for the expected initial frame before letting the animation calculate Cursor color as before. I will try it to see if it's possible. Thanks :)

andydotxyz:
> it will restart the blink in an unknown animation frame

I don't think this is true. Animation has no pause, each time you call `Start()` it will be at the beginning.

stuartmscott:
No Pablo is right in the suggestion I made where the animation is only started when the Entry gains focus and runs continuously but gets essentially "ignored" if there has been input in the last Xms. When the last input time is greater than the threshold the animation will no longer be "ignored" and this could cause the cursor to go from fully opaque to anywhere along the animation curve.

Perhaps the solution to this is the addition of `animation.Reset()` which could be called whenever the cursor is forced opaque to ensure when the animation resumes it does so from the start.

andydotxyz:
Oh, I missed the “ignored” bit.
If you leave it running but set the cursor to opaque won’t you be continually fighting the current animation ticks?

If you call `Stop()` when it is to be ignored and `Start()` where you proposed a Resume call you don’t need a new API

stuartmscott:
> If you leave it running but set the cursor to opaque won’t you be continually fighting the current animation ticks?

Not fighting, the animation tick would have an `if` comparing current time to `lastInputTime` and _either_ setting the colour to the animation colour _or_ fully opaque

> If you call Stop() when it is to be ignored and Start() where you proposed a Resume call you don’t need a new API

Well then you need a timer to know when to resume the animation, I was trying to avoid this for simplicity and just using the animation ticks as the timer

fpabl0:
> Perhaps the solution to this is the addition of animation.Reset() which could be called whenever the cursor is forced opaque to ensure when the animation resumes it does so from the start.

That would be a nice addition :)

> If you call Stop() when it is to be ignored and Start() where you proposed a Resume call you don’t need a new API

I don't think it is possible to call anim.Stop() inside the animation tick function, is it?

> Not fighting, the animation tick would have an if comparing current time to lastInputTime and either setting the colour to the animation colour or fully opaque and reseting the animation to frame 0.

Agreed, however it would not be that simple, because when the Entry initially get focus,  the current animation (the one in `develop` branch)  run, but when the animation is restarted after been interrupted on `TypedRune`, it needs to be inverted (swap start and end colors) to avoid jumpy animation. So it means we have to create two animations at least once, then we can just reset the inverted one until the Entry lose its focus.

stuartmscott:
I think instead of the "inverted" complication we might just want to change the animation to always be a fade out, so the cursor immediately appears then fades out, and when the animation reverses it fades back in, and then repeats. I think users would expect the cursor to immediately appear when they tap the Entry so it feels responsive.

fpabl0:
> I think instead of the "inverted" complication we might just want to change the animation to always be a fade out, so the cursor immediately appears then fades out, and when the animation reverses it fades back in, and then repeats. I think users would expect the cursor to immediately appear when they tap the Entry so it feels responsive.

Yes, that was the first possibility, unfortunately it would mean a lot of failed tests, because now the cursor would not be visible based on the way how animations are tested (`Tick(1.0)` when `anim.Start`).

andydotxyz:
> Not fighting, the animation tick would have an if comparing current time to lastInputTime and either setting the colour to the animation colour or fully opaque and reseting the animation to frame 0.

Oh, sorry I thought you were talking about stock color animations. It could indeed be done if we code up the color animation code with an if as you suggest. Cunning.

> I don't think it is possible to call anim.Stop() inside the animation tick function, is it?

You should be able to call `Stop()` from anywhere you like.

stuartmscott:
> unfortunately it would mean a lot of failed tests

The tests can manually tick the animation to whatever they need

fpabl0:
Closed in favor of #1968.

PR #1782, opened 2021-01-13
Move back to a more intuitive definition of Focus and primary-like color

andydotxyz:
Move the input background to InputBackgroundColor

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (commented):
The focus indicator for widget.Tree looks to still use primary. Is that expected?

Jacalz (dismissed):
Never mind. I see why and I think this makes sense.

Jacalz (approved):

PR #1784, opened 2021-01-13
Improve mobile tap handling

andydotxyz:
Once dragging starts be smooth.
When taps happen offset them up the screen to compensate for fat fingers. This is what iOS and others have done too.

Overall this makes mobile feel smoother and easier to handle.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/gomobile/driver.go
+++ b/internal/driver/gomobile/driver.go
@@ -23,7 +23,10 @@ import (
 	"fyne.io/fyne/theme"
 )
 
-const tapSecondaryDelay = 300 * time.Millisecond
+const (
+	tapMoveThreshold  = 4.0                    // how far can we move before it is a drag
+	tapSecondaryDelay = 300 * time.Millisecond // how long before secondary tap
+)
 
 type mobileDriver struct {
 	app   app.App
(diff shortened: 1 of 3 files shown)


AlbinoGeek:
I'm glad I caught this issue, it'll be hell for me later with absolute positioning if I didn't know this y compensation was there (to account for / offset / disable it's effect when I don't want it, heh.)

andydotxyz:
@AlbinoGeek I admit it looks strange - but it is what Apple etc have in their code as well.
Because folk tap lower than they think they do on touch screens.

What is the situation that you expect to need to disable this behaviour?

AlbinoGeek:
When a calibrated stylus or touch controller with an accuracy below 2mm (current industry standard) is in use (think: wacom-type display), and when the device's is levelled horizontally (drawing/drafting table), or when an android virtual device (or other "more precise" tap sources are introduced) are my specific use cases.

andydotxyz:
Good point. We currently don't support fully external hardware keyboard or pointers on Mobile.
There is more work to do here once we are able to ascertain the type of input device.

internal/driver/gomobile/driver.go
```diff
@@ -23,7 +23,11 @@ import (
 	"fyne.io/fyne/theme"
 )
 
-const tapSecondaryDelay = 300 * time.Millisecond
+const (
+	tapMoveThreshold  = 4.0                    // how far can we move before it is a drag
```
stuartmscott, line 27: Is this and tapYOffset adjusted by scale? 4 pixels on a low end screen is much larger than 4 pixels on a high end screen

andydotxyz: Yes, this is a point value - it will scale per-screen. On my iPhone it is about 15 pixels.
I tweaked it until I was reliably able to tap a button when my finger felt like it was in the right place.

stuartmscott: Perfect

stuartmscott (approved):

PR #1788, opened 2021-01-13
Fix/slider

andydotxyz:
Fix "only Draggable" break after Entry work.
This seems to fix it!

But... the colour dialog sliders are offset!?

After this release I think some work is required on taking the canvas logic out of drivers where possible, or at least refactoring the GLFW code as it is not very neat.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
@stuartmscott any chance that you have a bright idea on why those dialog sliders would be working differently to a generic slider?

PR #1789, opened 2021-01-13
move drag init into mouse moved

andydotxyz:
Fix "only Draggable" break after Entry work.
This seems to fix it!

But... the colour dialog sliders are offset!?

After this release I think some work is required on taking the canvas logic out of drivers where possible, or at least refactoring the GLFW code as it is not very neat.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -590,6 +590,9 @@ func (w *window) mouseMoved(viewport *glfw.Window, xpos float64, ypos float64) {
 		if cursorable, ok := object.(desktop.Cursorable); ok {
 			cursor = cursorable.Cursor()
 		}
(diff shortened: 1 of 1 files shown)


andydotxyz:
@stuartmscott any chance that you have a bright idea on why those dialog sliders would be working differently to a generic slider?

(sorry for the new PR)

stuartmscott:
Hmm nothing jumps out - it is using a vanilla Slider

stuartmscott:
On second look, could be a missing `ExtendBaseWidget` call

andydotxyz:
Thanks, that wasn't the issue but with fresh eyes in that area I saw what was wrong with the position parameters, ready for review now.

stuartmscott (approved):

PR #1792, opened 2021-01-13
Fix location of mobile popups by fixing the absolute position lookup

### LINKED ISSUE #1771: Mobile dropdowns are too low...
Looks like a bug in the `AbsolutePositionForObject` logic doubling the `SafeArea` code.

andydotxyz:
We just were not adjusting for the mobile canvas details.

We need to check the InteractiveArea for the top window (see screenshot of menu).
For child windows we also need to check there is a window head that is more than just the menu button (second screenshot).

Fixes #1771

### Checklist:

- [ ] Tests included. <- images attached until we improve mobile test infra
- [x] Tests all pass.


andydotxyz:
IMG_0086
IMG_0085

stuartmscott (approved):

PR #1799, opened 2021-01-14
Support wrap modes now Entry can handle them again

### LINKED ISSUE #1793: Entry no longer supports TextWrapOff
### Describe the bug:
Entry content is now wrapped in a scroller meaning it will no longer expand to fit the text content and instead hides it which is more like TextTruncate.

Full discussion: https://gophers.slack.com/archives/CK5U4BU87/p1610561504440100

### Screenshots:
https://gophers.slack.com/archives/CK5U4BU87/p1610563330454500

 - **OS:** MacOS
 - **Version:** Big Sur
 - **Go version:** 1.15
 - **Fyne version:** f2c3fb0c

andydotxyz:
Truncate means scrollable (as we cannot just hide editable content)
Single line cannot be wrapped as before

Fixes #1793

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


stuartmscott (approved):
Does this also fix #1751?

andydotxyz:
> Does this also fix #1751?

It does indeed, thanks for linking :)

PR #1804, opened 2021-01-14
As themes can override these methods now we add some safety

andydotxyz:
Be safe so bad themes cannot crash apps!

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


theme/theme.go
```diff
@@ -545,6 +545,36 @@ func loadCustomFont(env, variant string, fallback fyne.Resource) fyne.Resource {
 	return res
 }
 
+func safeColorLookup(n fyne.ThemeColorName, v fyne.ThemeVariant) color.Color {
+	col := current().Color(n, v)
+	if col == nil {
+		return &color.NRGBA{}
```
stuartmscott, line 551: Would this be black? We may want to default to an outrageous color (magenta?) to draw the dev's attention to a missing definition

andydotxyz: It is transparent, that felt right. When we add new colours this could happen.
Should we indicate that clearly or fall back to no color?

(note this is not normal fallback, just if they did not build their theme completely)

stuartmscott: I'm just thinking that if a dev built an empty theme, nothing would be visible

andydotxyz: Is nothing visible better or worse than a bug ugly coloured box?

I guess it's the intermediate I am more concerned about - as the toolkit should move on the flexible theme API was designed with the idea that a theme code could lag without significant ramifications. That's why I think transparent makes a better fallback in this case.

Ideally they would delegate to `theme.DefaultTheme().lookuptype(...)` but if they have chosen not to that is where this code gets called.

stuartmscott: I'm not sure what is better, any designers care to chime in?

Whatever the result, it should be publicly documented, and maybe we should return a constant rather than a fresh allocation, ie. theme.ColorFallback or color.Transparent

andydotxyz:
Implemented suggestions from call - we log on finding a broken theme and show it vividly with "-tags hints"

theme/theme.go
```diff
@@ -545,6 +545,36 @@ func loadCustomFont(env, variant string, fallback fyne.Resource) fyne.Resource {
...
+	}
+	return col
+}
+
+func safeFontLookup(s fyne.TextStyle) fyne.Resource {
+	col := current().Font(s)
```
toaster, line 558: Copy’n’paste mistake: `col` should be `font` or alike.

theme/theme.go
```diff
@@ -545,6 +545,36 @@ func loadCustomFont(env, variant string, fallback fyne.Resource) fyne.Resource {
...
+
+	if s.Monospace {
+		return DefaultTextMonospaceFont()
+	}
+	if s.Bold {
+		if s.Italic {
+			return DefaultTextBoldItalicFont()
+		}
+		return DefaultTextBoldFont()
+	}
```
toaster, line 572: I’d prefer
```go
if s.Bold && s.Italic {
	return DefaultTextBoldItalicFont()
}
if s.Bold {
	return DefaultTextBoldFont()
}
```
But I’ll leave the decision to you.

theme/theme_test.go
```diff
@@ -1,17 +1,13 @@
...
-	"fyne.io/fyne"
 	"github.com/stretchr/testify/assert"
-)
 
-// TODO figure how to test the default theme if we are messing with it...
-//func TestThemeDefault(t *testing.T) {
-//	loaded := fyne.GlobalSettings().Theme()
-//	assert.NotEqual(t, lightBackground, loaded.BackgroundColor())
-//}
+	"fyne.io/fyne"
```
toaster, line 9: I thought we wanted to have Fyne imports before 3rd party imports, didn’t we?

toaster (changes requested):
some nits

theme/theme_test.go
andydotxyz: I can't remember. Since switching to goimports the IDE wants it the other way.
Reverting this for the PR, but probably need to find a way for lint to ensure we are consistent.

theme/theme.go
andydotxyz: I was matching the way it's done in all the other font style comparisons

toaster (approved):

PR #1807, opened 2021-01-14
Fix/menutests

andydotxyz:
This is the menu colours working without the `Widget.BackgroundColor`

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


internal/driver/glfw/testdata/menu_bar_kbdctrl_traverse_menu_down_2.xml
```diff
@@ -16,7 +16,7 @@
 				<rectangle size="0x0"/>
 				<text bold pos="12,8" size="53x21">Button</text>
 			</widget>
-			<widget size="300x29" type="*glfw.MenuBar">
+			<widget backgroundColor="button" size="300x29" type="*glfw.MenuBar">
```
stuartmscott, line 19: this shouldn't be here - backgroundColor was removed

internal/driver/glfw/testdata/menu_bar_kbdctrl_traverse_menu_down_2.xml
```diff
@@ -55,13 +53,11 @@
...
 					<widget size="85x103" type="*widget.Scroll">
 						<widget size="85x103" type="*widget.menuBox">
-							<rectangle fillColor="background" size="85x111"/>
 							<container pos="0,4" size="85x111">
 								<widget size="85x29" type="*widget.menuItem">
-									<rectangle fillColor="focus" size="85x29"/>
 									<text pos="8,4" size="33x21">New</text>
 								</widget>
-								<widget pos="0,33" size="85x29" type="*widget.menuItem">
+								<widget backgroundColor="focus" pos="0,33" size="85x29" type="*widget.menuItem">
```
stuartmscott, line 60: ditto

internal/driver/glfw/testdata/menu_bar_kbdctrl_traverse_menu_down_2.xml
```diff
@@ -27,12 +27,10 @@
 					<radialGradient centerOffset="0.5,-0.5" pos="-4,29" size="4x4" startColor="shadow"/>
 					<linearGradient angle="270" endColor="shadow" pos="-4,0" size="4x29"/>
 				</widget>
-				<widget size="300x300" type="*glfw.menuBarUnderlay">
+				<widget size="300x300" type="*glfw.menuBarBackground">
```
stuartmscott, line 30: this type no longer exists

widget/menu.go
```diff
@@ -18,6 +18,7 @@ type Menu struct {
 	OnDismiss   func()
 	activeItem  *menuItem
 	customSized bool
+	hiding      bool
```
stuartmscott, line 21: I'm not sure how this is better than the bool in basewidget. Was locking the issue?

andydotxyz: That is a good question indeed. The internal bool was being used before and it didn't work. This line had to change to avoid it:

```go
	return i.Child() != nil && i.Child().Visible()
	return i.Child() != nil && !i.Child().hiding
```

internal/driver/glfw/testdata/menu_bar_kbdctrl_traverse_menu_down_2.xml
andydotxyz: This is very strange. I will try syncing again

internal/driver/glfw/testdata/menu_bar_kbdctrl_traverse_menu_down_2.xml
andydotxyz: A test run was missed somehow, should be fixed now

internal/driver/glfw/testdata/menu_bar_kbdctrl_traverse_menu_down_2.xml
andydotxyz: also fixed I think

andydotxyz:
All PRs have been blocked now for 12 hours and this is the fix.
I am going to merge it and we can review a better fix once more people are around.

Issue #1819, opened 2021-01-16
develop: TestGlCanvas_walkTree fails


toaster:
Since f5332ba57a95c042f95257f539a87bccfc030829 (maybe 10a773dc008c7961136ecf32c86717fd6a08bdfb) #1801 TestGlCanvas_walkTree fails with:

```
=== RUN   TestGlCanvas_walkTree
    canvas_test.go:574: 
        	Error Trace:	canvas_test.go:574
        	Error:      	Not equal: 
        	            	expected: []glfw.beforeCall{glfw.beforeCall{obj:(*fyne.Container)(0xc000020cc0), parent:fyne.CanvasObject(nil), pos:fyne.Position{X:17, Y:42}}, glfw.beforeCall{obj:(*fyne.Container)(0xc000020c40), parent:(*fyne.Container)(0xc000020cc0), pos:fyne.Position{X:317, Y:442}}, glfw.beforeCall{obj:(*canvas.Rectangle)(0xc000306840), parent:(*fyne.Container)(0xc000020c40), pos:fyne.Position{X:318, Y:444}}, glfw.beforeCall{obj:(*canvas.Rectangle)(0xc0003068a0), parent:(*fyne.Container)(0xc000020c40), pos:fyne.Position{X:337, Y:472}}, glfw.beforeCall{obj:(*fyne.Container)(0xc000020c80), parent:(*fyne.Container)(0xc000020cc0), pos:fyne.Position{X:24, Y:50}}, glfw.beforeCall{obj:(*canvas.Rectangle)(0xc000306900), parent:(*fyne.Container)(0xc000020c80), pos:fyne.Position{X:524, Y:650}}, glfw.beforeCall{obj:(*canvas.Rectangle)(0xc000306960), parent:(*fyne.Container)(0xc000020c80), pos:fyne.Position{X:84, Y:120}}}
[...]

Issue #1818, opened 2021-01-16
develop: TestWindow_DragObjectThatMoves fail


toaster:
This was extracted out of #1814 because the failures have different origins.

When running the GLFW window_test.go, TestWindow_DragObjectThatMoves fails.

This failure was
- originally introduced by e5bfd3c4d53ba6c770664c7bdbd5181fa02dc543 of #1783
- modified by bb4c70a554d62478e6897e8fbc2368eb47c3d834 and 4759e3a5f2a32f10baf519e556853d4503ba0e3a of #1789

The current outcome is:

```
=== RUN   TestWindow_DragObjectThatMoves
    window_test.go:392: 
        	Error Trace:	window_test.go:392
        	Error:      	Not equal: 
        	            	expected: &fyne.DragEvent{PointEvent:fyne.PointEvent{AbsolutePosition:fyne.Position{X:10, Y:10}, Position:fyne.Position{X:7, Y:7}}, Dragged:fyne.Delta{DX:2, DY:2}}
        	            	actual  : &fyne.DragEvent{PointEvent:fyne.PointEvent{AbsolutePosition:fyne.Position{X:10, Y:10}, Position:fyne.Position{X:6, Y:6}}, Dragged:fyne.Delta{DX:2, DY:2}}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -7,4 +7,4 @@
        	            	   Position: (fyne.Position) {
        	            	-   X: (float32) 7,
        	            	-   Y: (float32) 7
        	            	+   X: (float32) 6,
        	            	+   Y: (float32) 6
        	            	   }
        	Test:       	TestWindow_DragObjectThatMoves
--- FAIL: TestWindow_DragObjectThatMoves (0.21s)
```

**Note**: it might be necessary to disable the loop’s window closed handling (lines 103-148 of glfw/loop.go) to ensure that the test runs do not exit prematurely.

andydotxyz:
> Note: it might be necessary to disable the loop’s window closed handling (lines 103-148 of glfw/loop.go) to ensure that the test runs do not exit prematurely.

I think you can just comment out line 146 (the d.Quit)

andydotxyz:
Fixed, apoloiges for having caused this in the original commit

Issue #1815, opened 2021-01-16
develop: TestWindow_DragIntoNewObjectKeepingFocus & TestWindow_HoverableOnDragging fail


adrianre12:
### Describe the bug:
Windows10 and a linux Mint 20 VM are both failing with the following errors
--- FAIL: TestWindow_DragIntoNewObjectKeepingFocus (0.88s)
    window_test.go:306: 
        	Error Trace:	window_test.go:306
        	Error:      	Not equal: 
        	            	expected: *desktop.MouseEvent(&desktop.MouseEvent{PointEvent:fyne.PointEvent{AbsolutePosition:fyne.Position{X:19, Y:9}, Position:fyne.Position{X:15, Y:5}}, Button:1, Modifier:0})
        	            	actual  : <nil>(<nil>)
        	Test:       	TestWindow_DragIntoNewObjectKeepingFocus
    window_test.go:316: 
        	Error Trace:	window_test.go:316
        	Error:      	Expected nil, but got: &desktop.MouseEvent{PointEvent:fyne.PointEvent{AbsolutePosition:fyne.Position{X:19, Y:9}, Position:fyne.Position{X:1, Y:5}}, Button:1, Modifier:0}
        	Test:       	TestWindow_DragIntoNewObjectKeepingFocus
FAIL
--- FAIL: TestWindow_HoverableOnDragging (0.96s)
    window_test.go:377: 
        	Error Trace:	window_test.go:377
        	Error:      	Expected nil, but got: &desktop.MouseEvent{PointEvent:fyne.PointEvent{AbsolutePosition:fyne.Position{X:8, Y:8}, Position:fyne.Position{X:4, Y:4}}, Button:1, Modifier:0}
        	Test:       	TestWindow_HoverableOnDragging
FAIL

Started with commit
commit bb4c70a554d62478e6897e8fbc2368eb47c3d834
Author: Andy Williams <andy@andywilliams.me>
Date:   Wed Jan 13 20:21:17 2021 +0000

    move drag init into mouse moved

### To Reproduce:
go test ./... -count=1

### Device (please complete the following information):
 - **OS:** Linux and Windows
 - **Version:** Mint 20 and Windows 10
 - **Go version:** 1.15.6
 - **Fyne version:** develop HEAD

toaster:
Note: it might be necessary to disable the loop’s window closed handling (lines 103-148 of glfw/loop.go) to ensure that the test runs do not exit prematurely.

[...]

Issue #1814, opened 2021-01-16
develop: Test_gLDriver_AbsolutePositionForObject fail


adrianre12:
### Describe the bug:
Windows10 and a linux Mint 20 VM are both failing with the following error
--- FAIL: Test_gLDriver_AbsolutePositionForObject (0.98s)
    --- FAIL: Test_gLDriver_AbsolutePositionForObject/a_row (0.00s)
        driver_test.go:103: 
            	Error Trace:	driver_test.go:103
            	Error:      	Not equal: 
            	            	expected: fyne.Position{X:4, Y:66}
            	            	actual  : fyne.Position{X:4, Y:74}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -2,3 +2,3 @@
            	            	  X: (float32) 4,
            	            	- Y: (float32) 66
            	            	+ Y: (float32) 74
            	            	 }
            	Test:       	Test_gLDriver_AbsolutePositionForObject/a_row
    --- FAIL: Test_gLDriver_AbsolutePositionForObject/an_overlay_item (0.00s)
        driver_test.go:103: 
            	Error Trace:	driver_test.go:103
            	Error:      	Not equal: 
            	            	expected: fyne.Position{X:79, Y:85}
            	            	actual  : fyne.Position{X:79, Y:81}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -2,3 +2,3 @@
            	            	  X: (float32) 79,
            	            	- Y: (float32) 85
            	            	+ Y: (float32) 81
            	            	 }
            	Test:       	Test_gLDriver_AbsolutePositionForObject/an_overlay_item
    --- FAIL: Test_gLDriver_AbsolutePositionForObject/the_overlay_content (0.00s)
        driver_test.go:103: 
            	Error Trace:	driver_test.go:103
            	Error:      	Not equal: 
            	            	expected: fyne.Position{X:79, Y:52}
            	            	actual  : fyne.Position{X:79, Y:40}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -2,3 +2,3 @@
            	            	  X: (float32) 79,
            	            	- Y: (float32) 52
            	            	+ Y: (float32) 40
            	            	 }
            	Test:       	Test_gLDriver_AbsolutePositionForObject/the_overlay_content
FAIL
[...]

PR #1820, opened 2021-01-16
fix menu background issues

toaster:
### Description:

This PR fixes issues with the menu item background which let a lot of tests fail.
It first reverts prior attempts to fix the issue, then fixes some test files and finally applies the fix.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass. (except of non-CI tests, see #1811, #1814, #1815, #1818 and #1819)


andydotxyz:
The coveralls failure must be because our storage branch has upped the test coverage - this PR does not reduce it noticeably

andydotxyz (approved):
Approving as it resolves the test issues.
But noting that this hides the menu hover colours because active overrides hover...

PR #1826, opened 2021-01-16
remove hover from menus

toaster:
### Description:
This removes the “hovered” state from `MenuItem` and `menuBarItem`.
Since the keyboard control implementation this state was mostly invisible because it was ruled out by the “active” state.
The hover emphasis was only visible when a menu was activated via mouse and then the active menu item was changed by keyboard.
Thus, the “hovered” state is not really useful and was removed. No tests were affected because the described behaviour is not a real use-case :).

### Checklist:
- [x] Tests included (the existing ones).
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/menu_bar_item.go
+++ b/internal/driver/glfw/menu_bar_item.go
@@ -19,9 +19,8 @@ type menuBarItem struct {
 	Menu   *fyne.Menu
 	Parent *MenuBar
 
-	active  bool
-	child   *publicWidget.Menu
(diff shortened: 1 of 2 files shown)


andydotxyz:
Looks great, but I think some old test images are still in... I ran 
```
go test -tags no_native_menus ../../internal/driver/glfw
```
and got the same image errors as you can see in some of the CI runs.

andydotxyz (approved):

PR #1827, opened 2021-01-16
If we want to be relative to a moving object we really need to track its movements

andydotxyz:
I thought we had nailed it, but this is more logical than the previous

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -729,11 +729,9 @@ func (w *window) mouseClicked(_ *glfw.Window, btn glfw.MouseButton, action glfw.
 			if w.mouseDragged == nil {
 				w.queueEvent(func() { wid.MouseUp(mev) })
 			} else {
-				dragged := w.mouseDragged.(interface{}).(fyne.CanvasObject)
-				_, tapable := dragged.(desktop.Mouseable)
-				if dragged != nil && tapable {
+				if dragged, ok := w.mouseDragged.(desktop.Mouseable); ok {
 					mev.Position = w.mousePos.Subtract(w.mouseDraggedOffset)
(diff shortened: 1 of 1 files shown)


internal/driver/glfw/window.go
```diff
@@ -724,13 +726,17 @@ func (w *window) mouseClicked(_ *glfw.Window, btn glfw.MouseButton, action glfw.
...
 				w.queueEvent(func() { wid.MouseUp(mev) })
+			} else {
+				dragged := w.mouseDragged.(interface{}).(fyne.CanvasObject)
+				_, tapable := dragged.(desktop.Mouseable)
+				if dragged != nil && tapable {
+					mev.Position = w.mousePos.Subtract(w.mouseDraggedOffset)
+					w.queueEvent(func() { dragged.(desktop.Mouseable).MouseUp(mev) })
+				} else {
+					w.queueEvent(func() { wid.MouseUp(mev) })
+				}
```
toaster, line 739:
```suggestion
				if dragged, ok := w.mouseDragged.(desktop.Mouseable); ok {
					mev.Position = w.mousePos.Subtract(w.mouseDraggedOffset)
					w.queueEvent(func() { dragged.MouseUp(mev) })
				} else {
					w.queueEvent(func() { wid.MouseUp(mev) })
				}
```

toaster (changes requested):
LGTM but mostly I trust you here :).

One suggestion to simplify the mouse-up event.

internal/driver/glfw/window.go
andydotxyz: Thanks. For some reason I could not get it to work with that casting last night, no idea what I was doing wrong!

toaster (approved):

PR #1836, opened 2021-01-20
Tabs support

### LINKED ISSUE #1791: Tab indents should align correctly
In #761 we made a rough approximation.
In full implementation we need to align to the next tab-location  so that "  \tfoo" and "\tfoo" align.

adrianre12:
### Description:
This PR adds tab support to Entry and Label via the text Canvas. It is done by replacing font.Drawer with painter.Drawer. painter.
Drawer embeds font.Drawer and implements custom DrawString() and MeasureString() that handle the tab rune.
When a tab rune is encountered the draw dot is advanced to the next tab position. The tab positions are whole multiples of 
`tabWidth * spaceWidth` where tab width is in characters and the space width is in dots.
The tab width is set in the theme and defaults to 4. The space width is found from the width of the current fonts space glyph.

I have reverted the tab approximations done in #761

Fixes #1791

Monospaced=true
image

MonoSpaced=false
image

### Checklist:

- [X ] Tests included.
- [X ] Lint and formatter run with no errors.
- [X ] Tests all pass.

#### Where applicable:
- [X ] Public APIs match existing style.


go.mod
```diff
@@ -3,9 +3,11 @@ module fyne.io/fyne/v2
 go 1.12
 
 require (
+	fyne.io/fyne v1.4.3
```
Jacalz, line 6: This seems wrong, I think. We shouldn’t need to require an old version of fyne in the go.mod. There is probably a stray import somewhere that is using the old import path. Might be worth running `go mod tidy` a second time.

go.sum
```diff
@@ -1,5 +1,8 @@
+fyne.io/fyne v1.4.3 h1:356CnXCiYrrfaLGsB7qLK3c6ktzyh8WR05v/2RBu51I=
+fyne.io/fyne v1.4.3/go.mod h1:8kiPBNSDmuplxs9WnKCkaWYqbcXFy0DeAzwa6PBO9Z8=
```
Jacalz, line 2: I think this relates to my comment above.

go.sum
```diff
@@ -1,5 +1,8 @@
...
 github.com/Kodeworks/golang-image-ico v0.0.0-20141118225523-73f0f4cfade9 h1:1ltqoej5GtaWF8jaiA49HwsZD459jqm9YFz9ZtMFpQA=
 github.com/Kodeworks/golang-image-ico v0.0.0-20141118225523-73f0f4cfade9/go.mod h1:7uhhqiBaR4CpN0k9rMjOtjpcfGd6DG2m04zQxKnWQ0I=
+github.com/adrianre12/fyne v1.4.2 h1:Zoiqls4+3e3Vz5gEK6wJZh4cMQDkZ44cMd6gGVlvOx0=
```
Jacalz, line 5: This also seems a little strange. Why is this included in the modules?

Jacalz (changes requested):

go.sum
adrianre12: Good catch, havn't a clue why it is in there. Go tidy sorted it out.

theme/theme.go
```diff
@@ -327,6 +332,8 @@ func (t *builtinTheme) Size(s fyne.ThemeSizeName) float32 {
 		return 11
 	case SizeNameInputBorder:
 		return 2
+	case SizeNameTabWidth:
```
andydotxyz, line 335: I'm not sure about this name, or maybe even if this is the right place?
All the other sizes are in "dp" so 4 is really small (=Padding).
And then I started thinking, should it maybe be a user preference instead of a theme preference?
I'm not convinced by either right now, should we set '4' as a constant for "TabSpaceWidth" or something in the painter package for now?

widget/textgrid.go
```diff
@@ -15,6 +15,7 @@ const (
 	textAreaSpaceSymbol   = '·'
 	textAreaTabSymbol     = '→'
 	textAreaNewLineSymbol = '↵'
+	textTabIndent         = "    "
```
andydotxyz, line 18: I don't think this is needed

andydotxyz (changes requested):
This is a ggreat apprach, sorry it got left for a while.
One or two change requests and an open question about where we really should store the "4 spaces" option.

widget/textgrid.go
adrianre12: Hmm, I agree that the constant should be replaced by a variable based on the TabWidth setting, but the whole tab in a textgrid has to to be looked at. 
In `SetText()` every '\t' is replaced by 4 cells of ' '
Then in `textGridRenderer.refreshGrid()` if ShowWhitespace is true and the cell is '\t' it is replaced with the textAreaTabSymbol which can never happen as there are no '\t' in the cells.

To make the textGrid handle tabs the same as the rendered text will take a bit more work. It would have to handle the "12\t567" scenario and making tabs visible would be interesting!? I think it would involve inverting the white space handling. Ie replace all whitespace with the visible characters, which would expand the '\t' to the relevant number of textAreaTabSymbol. Then in refreshGrid() convert them back to whitespace if ShowWhiteSpace=false.

Assuming of course you still want visible whitespace and tab handling in textGrid?

theme/theme.go
adrianre12: I was not sure about it when I did it. First, I wanted it in the TextStyle, but that does not make it down to the level of the font drawing. I then chose the Theme as you seem to want to control all the aspects of the appearance via that, and handily it is globally available.
The argument of tab width and if a tab is a user or developer setting is another can of worms. i.e if I was writing a GO editor, as a developer I would want to set the tab width to 8. But if I was writing a node.js app I would want it set to 2. If I was writing a text editor I would want it as a user preference. I might even want to change it depending on what I have open and even have multiple different sizes. 

I don't think it should be a constant and should at least be changeable by the developer. I think you should decide where you want it as you have a better understanding of Fyne than me.

internal/painter/drawer.go
```diff
@@ -0,0 +1,86 @@
...
+	for _, c := range s {
+		if prevC >= 0 {
+			d.Dot.X += d.Face.Kern(prevC, c)
+		}
+		if c == '\t' {
+			d.Dot.X = tabStop(d.Face, d.Dot.X)
+		} else {
+			dr, mask, maskp, a, ok := d.Face.Glyph(d.Dot, c)
+			if !ok {
+				// TODO: is falling back on the U+FFFD glyph the responsibility of
```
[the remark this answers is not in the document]
adrianre12: I assume you mean the TODO?
I haven't a clue, I was quite surprised when I opened [font.go](https://github.com/golang/image/blob/master/font/font.go) for the first time and saw the number of TODOs. It has been like it for 6 years and it does not seem to be an issue so left it well alone.
It does allude to the question of multiple font handling. i.e. the other week there was a question in Fyne about displaying the check box characters U+2610 and U+2611. They are not in the NotoSans or NotoMono fonts but have been put in Noto-emoji. Should the font drawer handle this?

theme/theme.go
andydotxyz: I agree with your summary. The API would probably be needed in the Entry/Label so it can be exposed to the developer.
To do this it may be required to set it on Text - I will leave it to see what you think.

widget/textgrid.go
andydotxyz: > if ShowWhitespace is true and the cell is '\t' it is replaced with the textAreaTabSymbol which can never happen as there are no '\t' in the cells.

That should not be the case - we compare a `\t` in the source `string`, it is not subject to the same string rendering that you have updated for the `canvas.Text`.

TextGrid does not render a literal `\t` character - it puts the `\t` in the `Rune` field, but shows either a space or a symbol based on the whitespace mode.
It leaves a certain number of characters to emulate the same indentation as you have coded into Text.
To complete it here the number it moves should be %numberOfSpaces.

The reason I commented on this line addition is that I thought it was not needed.
Either we leave this in and make another PR to update TextGrid calculations, or we can get this fixed up here...

andydotxyz: The `TextGrid` `SetText` method could be updated without this `textTabIntent`, something like this:
```go
func (t *TextGrid) SetText(text string) {
	lines := strings.Split(text, "\n")
	rows := make([]TextGridRow, len(lines))
	for i, line := range lines {
		lineLen := len(line)
		for _, r := range line {
			if r == '\t' {
				len += tabSpaceCount-1
			}
		}

		cells := make([]TextGridCell, lineLen)
		extras := 0
		for j, r := range line {
			cells[j+extras] = TextGridCell{Rune: r}

			if r == '\t' {
				for k := j + extras + 1; k < j+extras+tabSpaceCount; k++ {
					cells[k] = TextGridCell{Rune: ' '}
				}
				extras += tabSpaceCount - 1
			}
		}

		rows[i] = TextGridRow{Cells: cells}
	}

	t.Rows = rows
	t.Refresh()
}
```

theme/theme.go
adrianre12: I had forgotton about Fyne.MeasureText() It is what made me chose the theme before. As it has nothing to do with Label or Entry setting up tabWidth in textProvider wont work.
Options:

- Stick with theme.
- Use user preferences but hard code a default of 4. I don't see this gives any advantages vs the theme.
- Modify the fyne.MeasureText() call to also require tabWidth. Breaks the API
- Add tabWidth to textStyle. This would be my choice.

andydotxyz: 1 vs 2 - it depends if we think this is user preference or developer preference. Users cannot customise theme beyone primary colour and light/dark and even that is optional for themes.
Following 3 seems like a breaking change that is not needed.
4 could be a smart move, seems consistent with the use of that to configure text in other ways.

internal/painter/drawer.go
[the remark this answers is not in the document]
andydotxyz: We need to support whatever the user specifies in FYNE_FONT, or in the theme font handling.
Does that answer the question?

theme/theme.go
adrianre12: TabWidth is now in the TextStyle

widget/textgrid.go
adrianre12: I have implemented tab stops in TextGrid. Tab characters are now padded with spaces to the next tab stop instead of being padded with 3 spaces. This now matches the behaviour of the canvas.Text().
One thing I am unsure about is how I have set the tabWidth. I have added it to `TextGrid` and not to the `TextGridStyle` this is because of how TextGrid works.  When `SetText()` is called the tabs are padded with spaces and rune by rune written to the grid cells. Conversely when `Text()` is called the runes are read from the grid cell by cell and tab padding spaces are removed. This means tab padding spaces are 'hard coded' into the cells it is not possible to update the tabWidth after the text has been set. As the style can be changed at any time it is not a suitable place to set the tabWidth. 
I did look at the possibility of updating the text with every style change but this gets complicated because of the ability to change the cell contents without using `SetText()`. i.e it is possible to insert unpadded tab characters or replace padding spaces with other characters.

andydotxyz: Hmm, this is tricky.
How about caching a copy of the current tab width value - then we can see if it changed on refresh.
If it did then we could iterate the rows of cells. If we use the tab indicator rune to show each tab location it could be re-arranged in the line I think? (possibly a lot more work?)

adrianre12: I thought about using a special rune perhaps the HT U+2409 but no matter what symbol you use you end up with the problem "what if someone wants to display symbol X"
I considered keeping a text version of the rows and refreshing them on each `SetTabWidth()` but then you have the problem of two sources of truth. Which is the master value, the text line or cell value?
To do it properly really means separating the data and display. Then only edit the data and refresh the display on each update. Which is really "let's write an editor".
This TextGrid is a strange beast, neither fish nor fowl. Is it looks like a spreadsheet but is used as a text field. If it was a spreadsheet it would not need to support tabs, if it was a text field it would not need to support cell updates.

internal/painter/drawer.go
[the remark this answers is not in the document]
andydotxyz: Do we need to follow up on this message? Can you confirm the code works as expected when FYNE_FONT is specified on the command line?

text.go
```diff
@@ -30,12 +30,33 @@ const (
...
+	DefaultTabWidth = 4
+)
+
 // TextStyle represents the styles that can be applied to a text canvas object
 // or text based widget.
 type TextStyle struct {
 	Bold      bool // Should text be bold
 	Italic    bool // Should text be italic
 	Monospace bool // Use the system monospace font instead of regular
+	tabWidth  int  // Width of tabs in spaces
```
andydotxyz, line 44: Can't we just used a public field like the other items? the check for 0 can be done inside the text code instead?

widget/textgrid.go
andydotxyz: I think this text const could be removed now?

widget/textgrid.go
```diff
@@ -84,34 +86,48 @@ func (t *TextGrid) Resize(size fyne.Size) {
...
-
 		rows[i] = TextGridRow{Cells: cells}
 	}
 
 	t.Rows = rows
 	t.Refresh()
 }
 
+// TabWidth either returns the set tab width or if not set the returns the DefaultTabWidth
+func (t *TextGrid) TabWidth() int {
```
andydotxyz, line 113: As above, we could just use the field instead of two new methods.

widget/textgrid.go
```diff
@@ -161,6 +171,7 @@ func (t *TextGrid) Row(row int) TextGridRow {
 
 // RowText returns a string representation of the content at the row specified.
 // If the index is out of bounds it returns an empty string.
+// Tab characters do not have padding spaces removed.
```
andydotxyz, line 174: It seems like we can (and should) fix this rather than documenting the exception.

widget/textgrid.go
```diff
@@ -175,6 +186,7 @@ func (t *TextGrid) RowText(row int) string {
 
 // SetRow updates the specified row of the grid's contents using the specified content and style and then refreshes.
 // If the row is beyond the end of the current buffer it will be expanded.
+// Tab characters are not padded with spaces.
```
andydotxyz, line 189: I'd consider rewording this slightly so it does not sound like a warning.
Maybe `The cells content will be set directly and no interpretation of tabs or spaces will occur`?

andydotxyz (changes requested):

widget/textgrid.go
adrianre12: You want `RowText()` to support tabs, shouldnt `SetRow()` also support tabs?

text.go
adrianre12: done

widget/textgrid.go
adrianre12: done

widget/textgrid.go
andydotxyz: Hmm, well `SetRow` is not `SetRowText` the text helpers do tab conversion, the code that accesses cells directly does not

widget/textgrid.go
```diff
@@ -109,20 +108,12 @@ func (t *TextGrid) SetText(text string) {
 	t.Refresh()
 }
 
-// TabWidth either returns the set tab width or if not set the returns the DefaultTabWidth
-func (t *TextGrid) TabWidth() int {
-	if t.tabWidth == 0 {
+// tabWidth either returns the set tab width or if not set the returns the DefaultTabWidth
+func (t *TextGrid) tabWidth() int {
```
andydotxyz, line 112: Sorry to be picky, but now this is unexported it belongs lower in the file as we group exported alphabetically followed by unexported alphabetically (for each tupe) - more at https://github.com/fyne-io/fyne/wiki/Contributing.

andydotxyz (changes requested):

widget/textgrid.go
adrianre12: Ah my bad, didnt read the method. Odd that you don't have a SetRowText

widget/textgrid.go
adrianre12: No worries

internal/painter/drawer.go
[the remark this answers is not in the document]
adrianre12: No, we do not need to follow up on the comment, it was copied from the original GO code just in case it was changed in the original code and someone was hunting for why Fyne is not behaving the same. 
Yes it does work with FYNE_FONT.

widget/textgrid.go
adrianre12: Done

internal/painter/drawer.go
```diff
@@ -0,0 +1,86 @@
...
+	"golang.org/x/image/font"
+	"golang.org/x/image/math/fixed"
+)
+
+// Drawer extends "golang.org/x/image/font" to add support for tabs
+// Drawer draws text on a destination image.
+//
+// A Drawer is not safe for concurrent use by multiple goroutines, since its
+// Face is not.
+type Drawer struct {
```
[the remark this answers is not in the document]
adrianre12: Done

text.go
```diff
@@ -30,12 +30,33 @@ const (
 	TextWrapWord
 )
 
+const (
+	// DefaultTabWidth is the default width in spaces
```
andydotxyz, line 34: API introduced needs a since header:
```
//
// Since: 2.1
```

However - I don't think this API is necessary. If the defaulting was handled in the renderer and widget packages we could hide it in an `internal` package until we decide if this detail should be exposed...

text.go
andydotxyz: I don't think this is done. The default code below is what causes the exposed public API above - if we just use a public field as suggested then the default handling would be in painter / widget and would avoid the additional API requirements.

text.go
```diff
@@ -30,12 +30,33 @@ const (
...
 	Bold      bool // Should text be bold
 	Italic    bool // Should text be italic
 	Monospace bool // Use the system monospace font instead of regular
+	tabWidth  int  // Width of tabs in spaces
+}
+
+// TabWidth either returns the set tab width or if not set the returns the DefaultTabWidth
+func (ts *TextStyle) TabWidth() int {
+	if ts.tabWidth == 0 {
+		return DefaultTabWidth
```
andydotxyz, line 50: Neither this nor the setter are required if the field is public

andydotxyz (changes requested):
Not sure we were on the same page regarding the benefits of `TextStyle.TabWidth`.
Exposing the field instead of the methods used would reduce the API footprint of `TextStyle` as well as remove the need for the public `DefaultTabWidth`.

andydotxyz:
Hey @Jacalz would you be OK if I landed this and made the changes I requested, or did you have additional requests?
I feel bad that some PRs are hanging on so long when it's just simple differences holding it back.
Folk get busy but I'd like to help land some of these older changes.

Jacalz:
Go ahead. It's fine for me 🙂

andydotxyz:
Thanks so much for all your work on this @adrianre12, I have now landed it with the changes requested above.
When you have time feel free to look at `develop` branch if you want to check over my work.

PR #1837, opened 2021-01-20
Optimize the cost of calling on the main/draw threads

changkun:
### Description:

Calling a function on the main thread is a common operation
in Go's GUI programming. For instance, the runOnMain must
be called at each render loop.

Previously, the runOnMain allocates a done channel to wait
for a scheduled call on the main thread to complete and thus
can produce the amount of work and awake the garbage collector
quite often. Instead of allocating a channel for each call
on the main thread, this change uses a channel pool and
optimizes this particular case.

Aside, the calls on the draw thread can also be optimized and
use the same channel pool.

Benchmark results (benchstat):

```
$ go test -run=^$ -bench=. -count=10
```

```
goos: linux
goarch: amd64
pkg: fyne.io/fyne/internal/driver/glfw
cpu: Intel(R) Core(TM) i9-9900K CPU @ 3.60GH

name          old time/op     new time/op      delta
RunOnMain-16   5.26µs ±1%      5.02µs ±2%     -4.64%  (p=0.000 n=10+10)
RunOnDraw-16  44.5µs ±12%     43.2µs ±10%        ~    (p=0.393 n=10+10)

name          old alloc/op   new alloc/op      delta
RunOnMain-16    96.0B ±0%       0.0B        -100.00%  (p=0.000 n=10+10)
RunOnDraw-16     107B ±3%        10B ±33%    -90.15%  (p=0.000 n=10+10)

name          old allocs/op new allocs/op      delta
RunOnMain-16     1.00 ±0%      0.00         -100.00%  (p=0.000 n=10+10)
RunOnDraw-16     2.00 ±0%      1.00 ±0%      -50.00%  (p=0.000 n=10+10)
```

Trace visualization:

```
$ cd internal/driver/glfw
& go test -run=none -bench=Main -trace trace.out
```

Before

image

After:

image

(explanation: no garbage!)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/loop.go
+++ b/internal/driver/glfw/loop.go
@@ -31,6 +31,9 @@ var drawFuncQueue = make(chan drawData)
 var runFlag = false
 var runMutex = &sync.Mutex{}
 var initOnce = &sync.Once{}
+var donePool = &sync.Pool{New: func() interface{} {
+	return make(chan struct{})
+}}
 
 // Arrange that main.main runs on main thread.
 func init() {
@@ -50,7 +53,8 @@ func runOnMain(f func()) {
 	if !running() {
 		f()
 	} else {
-		done := make(chan struct{})
+		done := donePool.Get().(chan struct{})
+		defer donePool.Put(done)
 
 		funcQueue <- funcData{f: f, done: done}
 		<-done
@@ -59,7 +63,8 @@ func runOnMain(f func()) {
 
 // force a function f to run on the draw thread
 func runOnDraw(w *window, f func()) {
-	done := make(chan struct{})
+	done := donePool.Get().(chan struct{})
+	defer donePool.Put(done)
 
 	drawFuncQueue <- drawData{f: f, win: w, done: done}
 	<-done
--- a/internal/driver/glfw/loop_test.go
+++ b/internal/driver/glfw/loop_test.go
@@ -1,0 +1,32 @@
+// +build !ci
+// +build !mobile
+
+package glfw
+
+import "testing"
+
+// BenchmarkRunOnMain measures the cost of calling a function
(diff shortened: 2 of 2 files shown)


Jacalz (approved):
Thanks! Looks good to me. Always nice to see performance improvements 🙂

andydotxyz:
I was about to approve this (well, after I have tested some of our larger apps) but wanted to check:
Will the pool grow indefinitely if we get huge numbers of simultaneous requests?

changkun:
> Will the pool grow indefinitely if we get huge numbers of simultaneous requests?

Pool allocated objects will be garbage collected every two GC cycles (after 1.13). In particular, after the first GC cycle, allocated objects are victims and will be reused if there are any needs to reuse them, otherwise, they will be released.

The [GC pacer](https://docs.google.com/document/d/1wmjrocXIWTr1JxU-3EQBI6BK6KgtiFArkG47XK73xIQ/edit#) has a [soft and hard limit](https://github.com/golang/proposal/blob/master/design/14951-soft-heap-limit.md) that provides the progress guarantee, which prevents the growth of the heap goes to infinite.

changkun:
@andydotxyz ping?

andydotxyz:
I'm sorry I forgot. We had a discussion on the community call but I didn't get back to you.

changkun:
Never mind, what's the conclusion out of the discussion?

andydotxyz:
Ah yes, sorry for the ambiguity. This was given the go-ahead. I was waiting until we had forked the bugfix release as this wants to be in main develop for significant testing before release :)

andydotxyz:
That's on develop now :)

PR #1858, opened 2021-01-24
Port cmd/fyne to using urfave/cli

Jacalz:
### Description:

This PR moves ahead with the task of porting over the entire `cmd/fyne` cli application to using https://github.com/urfave/cli. It currently needs more testing and I would like if people take a look at my proposed flag aliases and names along with any usages or descriptions that need to be updated with better wording. Of note is that the `cmd/fyne/commands` package has a couple breakages, but it should only affect the commands themselves (I don't think the intention ever was to let other people create cli applications with it).  

The work on porting to cobra was abandoned due to the fact that flags had to be posix compliant.
Supersedes https://github.com/fyne-io/fyne/pull/1312

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


AlbinoGeek:
I really don't like single dash command arguments `-key=val` that Golang does, they specifically chose not to listen to the standardized way of doing things, with no discernible benefit whatsoever (except typing one less dash? but arguments are auto-completed by any shell written in the last 10 years... so who cares?)

In fact, it makes single dash arguments "not chainable", which is pretty dumb. (see: any *nix command ever: `cmd -ancRQ`)

Which brings me to the point of that rant:

- Does urfive support auto-complete? `fyne r<tab>elease` etc
- Does urfive support levenshtein distance-based auto-correction? `fyne rlesae` -> `fyne release`

Those were the only real real features that cobra provided over any other cli impl. (Well, that and viper integration.)

cmd/fyne/commands/deprecated.go
```diff
@@ -0,0 +1,45 @@
...
+			cmd := exec.Command("go", "mod", "vendor")
+			cmd.Stdout, cmd.Stderr = os.Stdout, os.Stderr
+			return cmd.Run()
+		},
+	}
+}
+
+// Version returns the cli command for the program version
+//
+// Deprecated: Use "go mod vendor" instead
```
AlbinoGeek, line 30: This deprecation comment is inconsistent with the usage instructions for this command.

```suggestion
// Deprecated: Use "fyne --version" or "fyne -v" instead.
```

cmd/fyne/commands/deprecated.go
```diff
@@ -0,0 +1,45 @@
...
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"runtime/debug"
+
+	"github.com/urfave/cli/v2"
+)
+
+// Vendor returns the vendor cli command
```
AlbinoGeek, line 13: Nit:

I'm not going to blag every occurrence of this, just the first in each package.

Apparently code comments are supposed to end with a period if they are not to start a heading.

```suggestion
// Vendor returns the vendor cli command.
```

From go blog: https://blog.golang.org/godoc

> Notice this comment is a **complete sentence** that begins with the name of the element it describes. This important convention allows us to generate documentation in a variety of formats, from plain text to HTML to UNIX man pages, and makes it read better when tools truncate it for brevity, such as when they extract the first line or sentence.

From godoctricks: https://godoc.org/github.com/fluhus/godoc-tricks#Paragraphs

Searching through official documentation on the standard library we can also see this standard in practice:

https://golang.org/src/archive/tar/common.go

cmd/fyne/commands/get.go
```diff
@@ -1,19 +1,33 @@
...
 
 	"github.com/pkg/errors"
+	"github.com/urfave/cli/v2"
 
 	"fyne.io/fyne/v2/cmd/fyne/internal/util"
 )
 
-// Declare conformity to Command interface
-var _ Command = (*Getter)(nil)
+// Get returns the command for getting a fyne application downloaded
```
AlbinoGeek, line 15: Consider:

```suggestion
// Get returns the command which downloads and installs fyne applications.
```

cmd/fyne/commands/get.go
```diff
@@ -26,58 +40,33 @@ func NewGetter() *Getter {
...
-
 // SetIcon allows you to set the app icon path that will be used for the next Get operation.
 func (g *Getter) SetIcon(path string) {
 	g.icon = path
 }
 
-func get(pkg, icon string) error {
+// Get automates the download and install of a named GUI app package.
+func (g *Getter) Get(pkg string) error {
 	path := filepath.Join(goPath(), "src", pkg)
```
AlbinoGeek, line 50: Could move to L61, where it is first used.

cmd/fyne/commands/install.go
```diff
@@ -2,19 +2,75 @@ package commands
...
-	"flag"
-	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
 
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/cmd/fyne/internal/mobile"
+	"github.com/urfave/cli/v2"
```
AlbinoGeek, line 12: inconsistent import ordering within the same package

cmd/fyne/commands/get.go
```diff
@@ -26,58 +40,33 @@ func NewGetter() *Getter {
...
 
-	install := &installer{srcDir: path, icon: icon, release: true}
-	err = install.validate()
-	if err != nil {
+	install := &installer{srcDir: path, icon: g.icon, release: true}
+	if err := install.validate(); err != nil {
 		return errors.Wrap(err, "Failed to set up installer")
 	}
+
 	return install.install()
```
AlbinoGeek, line 70: If `Get` installs, why is `Install` a different command in content? Perhaps `Install` needs a new name?

cmd/fyne/commands/install.go
```diff
@@ -2,19 +2,75 @@ package commands
...
 )
 
-// Declare conformity to Command interface
-var _ Command = (*installer)(nil)
+// Install returns the cli command for installing fyne applications
+func Install() *cli.Command {
```
AlbinoGeek, line 16: This sounds like what used to be `fyne package`? Except that it also Installs....
Humm... Now I don't know what to think about `Get` v.s. `Install`

cmd/fyne/commands/install.go
```diff
@@ -23,43 +79,10 @@ type installer struct {
 }
 
 // NewInstaller returns an install command that can install locally built Fyne apps.
-func NewInstaller() Command {
+func NewInstaller() *installer {
```
AlbinoGeek, line 82: Is this func called from other packages, does it need to be exported or even exist?

cmd/fyne/commands/package.go
```diff
@@ -15,15 +14,84 @@ import (
 	"fyne.io/fyne/v2/cmd/fyne/internal/util"
 
 	"github.com/pkg/errors"
+	"github.com/urfave/cli/v2"
```
AlbinoGeek, line 17: Nit: inconsistent import ordering

cmd/fyne/commands/package.go
```diff
@@ -15,15 +14,84 @@ import (
...
+			&cli.StringFlag{
+				Name:        "icon",
+				Usage:       "The name of the application icon file.",
+				Value:       "Icon.png",
+				Destination: &p.icon,
+			},
+			&cli.StringFlag{
+				Name:        "appiD",
+				Aliases:     []string{"id"},
+				Usage:       "For ios or darwin targets an appID is required, for ios this must match a valid provisioning profile",
```
AlbinoGeek, line 81:
```suggestion
				Usage:       "For most platforms an appID in the form of a reversed domain names is required, for ios this must match a valid provisioning profile",
```

* perhaps I'm entirely wrong about iOS package names, please ignore this if that is the case.

Android specifically (because Java) used package names in the form of reversed domain names.

Microsoft also (on the Store) recommends package names following a similar form.

cmd/fyne/commands/release.go
```diff
@@ -13,21 +11,131 @@ import (
 	"fyne.io/fyne/v2/cmd/fyne/internal/mobile"
 	"fyne.io/fyne/v2/cmd/fyne/internal/templates"
 	"fyne.io/fyne/v2/cmd/fyne/internal/util"
+	"github.com/urfave/cli/v2"
```
AlbinoGeek, line 14: Nit: inconsistent import order

cmd/fyne/commands/release.go
```diff
@@ -13,21 +11,131 @@ import (
...
-	macAppStoreCategories = []string{
-		"business", "developer-tools", "education", "entertainment", "finance", "games", "action-games",
-		"adventure-games", "arcade-games", "board-games", "card-games", "casino-games", "dice-games",
-		"educational-games", "family-games", "kids-games", "music-games", "puzzle-games", "racing-games",
-		"role-playing-games", "simulation-games", "sports-games", "strategy-games", "trivia-games", "word-games",
-		"graphics-design", "healthcare-fitness", "lifestyle", "medical", "music", "news", "photography",
-		"productivity", "reference", "social-networking", "sports", "travel", "utilities", "video", "weather",
-	}
-)
+var macAppStoreCategories = []string{
```
AlbinoGeek, line 17: These are no longer current as of February 5th, 2020 with the universal platform:
https://developer.apple.com/news/?id=02052020a

cmd/fyne/commands/release.go
```diff
@@ -13,21 +11,131 @@ import (
...
+	r := &releaser{}
+
+	return &cli.Command{
+		Name:  "release",
+		Usage: "Prepares an application for public distribution.",
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:        "target",
+				Aliases:     []string{"os"},
+				Usage:       "The operating system to target (android, android/arm, android/arm64, android/amd64, android/386, darwin, freebsd, ios, linux, netbsd, openbsd, windows)",
```
AlbinoGeek, line 37: What is `android` v.s `android/386` ?  Is that practically `android/arm386`?

Also, where did 32-bit BSD, Linux and Windows go?

cmd/fyne/commands/release.go
```diff
@@ -13,21 +11,131 @@ import (
...
+			},
+			&cli.IntFlag{
+				Name:        "appBuild",
+				Usage:       "Build number, should be greater than 0 and incremented for each build",
+				Destination: &r.appBuild,
+			},
+			&cli.StringFlag{
+				Name:        "appiD",
+				Aliases:     []string{"id"},
+				Usage:       "For ios or darwin targets an appID is required, for ios this must match a valid provisioning profile",
```
AlbinoGeek, line 78: [ same comment as other appID usage ]

cmd/fyne/main.go
```diff
@@ -2,116 +2,41 @@
...
 
 import (
-	"flag"
 	"fmt"
 	"os"
+	"runtime/debug"
 
-	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/cmd/fyne/commands"
+	"github.com/urfave/cli/v2"
```
AlbinoGeek, line 10: Nit: inconsistent import ordering

cmd/fyne/main.go
```diff
@@ -2,116 +2,41 @@
...
-	loadCommands()
-	flag.Usage = printUsage
-	// first let's extract the first "command"
-	args := os.Args[1:]
-	if len(args) < 1 {
-		help()
+	app := &cli.App{
+		Name:        "fyne",
+		Usage:       "A command line helper for various Fyne tools",
+		Description: "The fyne command provides tooling for fyne applications and their development",
```
AlbinoGeek, line 17:
```suggestion
		Description: "The fyne command provides tooling for fyne applications to assist in their development",
```

go.mod
```diff
@@ -20,6 +21,8 @@ require (
 	github.com/srwiley/oksvg v0.0.0-20200311192757-870daf9aa564
 	github.com/srwiley/rasterx v0.0.0-20200120212402-85cb7272f5e9
 	github.com/stretchr/testify v1.5.1
+	github.com/urfave/cli v1.22.5 // indirect
```
AlbinoGeek, line 24: Uh-oh, 1.x and 2.x both imported.

andydotxyz:
Is there any way we can test that this change is backward compatible?
I like the idea, as discussed on the call, but it's important that the changes are all aliased/deprecated to not break any usage.

cmd/fyne/commands/get.go
```diff
@@ -26,58 +40,33 @@ func NewGetter() *Getter {
 	return &Getter{}
 }
 
-// Get automates the download and install of a named GUI app package.
-func (g *Getter) Get(pkg string) error {
```
andydotxyz, line 30: This is part of a public API and cannot be removed - see usage at https://github.com/fyne-io/apps/blob/master/welcome.go#L164

andydotxyz (changes requested):

cmd/fyne/commands/get.go
Jacalz: I was pretty sure that I didn't remove this. Will have to see what's up with that.

Jacalz: It isn't removed. Look further down in the file.

cmd/fyne/commands/deprecated.go
Jacalz: Sure. A pretty picky thing given that a bunch of comments within Fyne doesn't do that, but I will get that updated.

cmd/fyne/commands/get.go
Jacalz: I think that's outside of the scope of this PR.

cmd/fyne/commands/install.go
Jacalz: Outside of the scope of this PR. This has not changed since before.

cmd/fyne/commands/release.go
Jacalz: Pretty sure that that isn't in the scope of this PR. Better to open up an issue or PR for that in the future.

cmd/fyne/commands/release.go
Jacalz: We probably shouldn't list all of possible values in the usage. The list we have is way too long.

cmd/fyne/commands/deprecated.go
```diff
@@ -25,21 +25,19 @@ func Vendor() *cli.Command {
...
 		Action: func(_ *cli.Context) error {
-			if info, ok := debug.ReadBuildInfo(); !ok {
-				return errors.New("could not retrieve version information (ensure module support is activated and build again)")
-			} else {
+			if info, ok := debug.ReadBuildInfo(); ok {
 				fmt.Println("fyne cli version:", info.Main.Version)
 			}
 
-			return nil
+			return errors.New("could not retrieve version information (ensure module support is activated and build again)")
```
AlbinoGeek, line 40: Wouldn't this error be returned even if a `fyne cli version` was successfully returned?

AlbinoGeek (commented):
Love what I'm seeing :) New changes look great.

Also good to know I'm now the only one using `spew.Dump()` to debug :P

cmd/fyne/commands/get.go
andydotxyz: Oh I see. Can you keep the methods in the correct order so things don't get mixed up please?
Also the new static functions should appear after the type functions IIRC.

cmd/fyne/commands/release.go
AlbinoGeek: Makes sense. If extended helps comes up somewhere, we could link an FAQ or developer.fyne page that has information on building, or lists architectures.

cmd/fyne/commands/deprecated.go
Jacalz: Good point. I had missed a return.

Jacalz:
> Love what I'm seeing :) New changes look great.
> 
> Also good to know I'm now the only one using `spew.Dump()` to debug :P

Thanks. I don't use `spew.Dump()`. Where did you get that from?

AlbinoGeek:
> > Love what I'm seeing :) New changes look great.
> > Also good to know I'm now the only one using `spew.Dump()` to debug :P
> 
> Thanks. I don't use `spew.Dump()`. Where did you get that from?

I got that from the `go.sum` removal of a lingering import of `spew`.

cmd/fyne/commands/bundle.go
```diff
@@ -1,7 +1,7 @@
-// Package commands provides functionality for managing fyne packages and the build process
```
andydotxyz, line 1: Should this have been removed?

cmd/fyne/commands/bundle.go
```diff
@@ -12,25 +12,157 @@ import (
...
+	"github.com/urfave/cli/v2"
 )
 
 const fileHeader = "// auto-generated\n" + // to exclude this file in goreportcard (it has to be first)
 	"// Code generated by '$ fyne bundle'. DO NOT EDIT." // idiomatic mark, see https://golang.org/s/generatedcode
 
-// Declare conformity to Command interface
-var _ Command = (*bundler)(nil)
+// Bundle returns the vendor cli command
+func Bundle() *cli.Command {
```
andydotxyz, line 22: Isn't this adding new CLI specific exported commands?
It may seem very similar to before, but this time we are exposing. third party library in our public API which goes against our design.

cmd/fyne/commands/bundle.go
```diff
@@ -12,25 +12,157 @@ import (
...
+				Usage:       "Append an existing go file (don't output headers).",
+				Destination: &b.noheader,
+			},
+		},
+		Action: func(ctx *cli.Context) error {
+			if ctx.Args().Len() != 1 {
+				return errors.New("missing required file or directory parameter after flags")
+			}
+
+			outFile := os.Stdout
```
andydotxyz, line 66: Can this go into a separate function perhaps? The number of indentations in here is hard to read.
And harder to test.

cmd/fyne/commands/bundle.go
```diff
@@ -12,25 +12,157 @@ import (
...
+	}
+}
 
 type bundler struct {
 	name, pkg, out string
 	prefix         string
 	noheader       bool
 }
 
+func (b *bundler) dirBundle(dirpath string, out *os.File) error {
```
andydotxyz, line 118: unexported methods belong after the public ones (where they were before).

cmd/fyne/commands/bundle.go
andydotxyz: Can you help me also to see where this code moved from? The GitHub diff is not showing the source

andydotxyz (changes requested):
I think a few changes like re-ordering and doc removal need to be reversed.

The main problem here is that we have a new public API that is exposing 3rd party CLI interface types instead of our own. This is something we try to avoid, the solution may need some more thinking.

cmd/fyne/commands/bundle.go
Jacalz: No, sorry. Got wiped out when I moved around a few files.

cmd/fyne/commands/bundle.go
Jacalz: It is. Will be fixed soon.

cmd/fyne/commands/bundle.go
Jacalz: I have moved most of these out to separate functions now. It is taken from the Run() function.

cmd/fyne/internal/commands/deprecated.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+// Version returns the cli command for the program version.
+//
+// Deprecated: Use "fyne --version" or "fyne -v" instead.
+func Version() *cli.Command {
+	return &cli.Command{
+		Name:  "version",
+		Usage: "Deprecated: Use \"fyne --version\" or \"fyne -v\" instead.",
```
andydotxyz, line 34: Is there any particular reason to deprecate "fyne version" when it follows the pattern of other subcommands?

andydotxyz (changes requested):
Overall this is looking good - sorry it has been so much work!

There is a problem - as far as I can see `NewGetter` has been removed from the public API.
The struct is there, but not the constructor or methods.

cmd/fyne/internal/commands/deprecated.go
Jacalz: It is because `--version` (which is built into urfave/cli) basically is the standard way of showing versions on UNIX. It also cleans up our code a slight bit to not add one extra command 🙂

andydotxyz: But it removes the expected location for the version command for anyone working with go tools...

Jacalz: That is a very good point. Didn't think about that. Will see what I can do 🙂

Jacalz: Moved back to what we had.

andydotxyz (changes requested):
So close.

However when I try to use this the `Get(string)` and `SetIcon(string)` methods are not found on `Getter`. We may need public definitions that wrap...

Or you could use `type Getter = commands.Getter` to expose all of the internally-public methods.

andydotxyz (dismissed):
Looks good to me. Might be good to get another 👍 if possible, but it won't block merging.

go.sum
```diff
@@ -1,7 +1,12 @@
+fyne.io/fyne v1.4.3 h1:356CnXCiYrrfaLGsB7qLK3c6ktzyh8WR05v/2RBu51I=
+github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
+github.com/Jacalz/wormhole-gui v1.3.0 h1:rZrYxtNp8LS8XCvlB13oeeJMIGvuhqcsNDqmKDrt6zE=
```
AlbinoGeek, line 3: Just a few extra `go.sum` entries leaked in :)

AlbinoGeek (approved):
Looks like everything I'd commented on has been addressed.

go.sum
andydotxyz: Good catch!

Jacalz:
Sorry for the force push. Had to rebase on develop and git wouldn't let me push normally.

go.sum
Jacalz: No idea how those got there (can't remember getting wormhole-william), but should be fixed now. Nice catch!

cmd/hello/main.go
```diff
@@ -11,6 +11,10 @@ func main() {
 	a := app.New()
 	w := a.NewWindow("Hello")
 
+	w2 := a.NewWindow("Second")
```
andydotxyz, line 14: Not sure when this slipped in?

andydotxyz (changes requested):
Looks like a random code addition snuck into `cmd/hello/main.go`

cmd/hello/main.go
Jacalz: Me neither. Strange. Will get that removed.

andydotxyz (approved):

PR #1867, opened 2021-01-26
Adds Prepend to *Container

donatj:
### Description:

Adds Prepend method to container.

VBox and HBox in 1.0 had a Prepend and I have a lot of code making heavy use of the functionality. It's really handy and would be nice to have back now that the Box's are containers.

Fixes #(issue)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container.go
+++ b/container.go
@@ -57,6 +57,12 @@ func (c *Container) Add(add CanvasObject) {
 	c.layout()
 }
 
+// Prepend prepends the specified object to the items this container manages.
+func (c *Container) Prepend(add CanvasObject) {
+	c.Objects = append([]CanvasObject{add}, c.Objects...)
+	c.layout()
+}
+
 // AddObject adds another CanvasObject to the set this Container holds.
 //
 // Deprecated: Use replacement Add() function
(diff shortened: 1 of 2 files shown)


andydotxyz:
This was not added to container API because ordering of container children is non-sensical.
VBox and HBox are ordered first-to-last but border, max, padded etc are not.

I agree that this was a helpful method, but I don't think that this is the right solution.

donatj:
I suspect then that differentiating Container from OrdersContainer might be in… order?

andydotxyz:
Perhaps, I could see that being beneficial so that Box and Grid can have this extra funcitonality.

andydotxyz:
@donatj Is looking how we add ordering to container something that this PR will track, or should we close this in preference of a future PR that takes that approach?

andydotxyz:
I just had an idea @donatj - without having to deal with the semantics of `Append`/`Prepend` how about `Insert`?
That would take an object and an index, I would think so calling `Insert(obj, 0)` would have the same semantics as your `Prepend` call...
What do you think?

andydotxyz:
p.s. You may need to pull in `develop` changes as we have updated tests that are required, but not present on the commit this PR was based from.

Jacalz:
Hi @donatj. Are you interested in continuing working on this PR?

donatj:
I rewrote my code to work around it, so meh.

PR #1875, opened 2021-01-27
Use full background to indicate hover and selection

### LINKED ISSUE #1297: Transparent Background

stuartmscott:
### Description:
Replaces hover and selection indicators in List and Tree with full size background

Fixes #1297

### Checklist:
- [ ] Tests included. // previous tests repurposed
- [x] Tests all pass.


AlbinoGeek (commented):
I like it! It's very jarring that this colour is so... high intensity, but we don't have "shades of primary" yet, so this is definitely a push in the right direction.

What I did notice, is there's no XML test for "Hover" state on these widgets.

andydotxyz:
> I like it! It's very jarring that this colour is so... high intensity, but we don't have "shades of primary" yet,

Select is using `theme.FocusColor` which whould probably work better here as well.
It's the best option until we have a proper selected colour across the toolkit.

stuartmscott:
I can switch the background color to `theme.FocusColor`, but what is preventing us having `theme.SelectedColor` / what's the ETA?

stuartmscott:
> What I did notice, is there's no XML test for "Hover" state on these widgets.

Correct, tho it the behaviour is tested with regular unit tests, and I'm not sure what XML tests will give over that.

stuartmscott:
Using focus color;

https://user-images.githubusercontent.com/1117080/106174442-e6588580-6149-11eb-900a-0f0f2d85d57e.mov

AlbinoGeek:
That feels way less jarring.

I must ask however, should the background be restricted to the separators in contexts where they have them, or no? (This actually brings into question why the separators in that list respect border left/right, but the selection background doesn't?)

stuartmscott:
It's a good point. The separators were inset so that the selection indicator could appear in the gutter, now that selection indication is full background should;

A. Separators become full width?
b. Background become insert?

AlbinoGeek:
I agree, (A) seems like the most logical approach, at least within `List` and `Tree`. This better matches the styling of `Table`.

stuartmscott:
Full width separators:

https://user-images.githubusercontent.com/1117080/106222075-a4543180-6193-11eb-9b52-e37b1de1c2a7.mov

AlbinoGeek:
Crisp!

andydotxyz (approved):

PR #1885, opened 2021-01-30
Add support for building on DragonflyBSD

Jacalz:
### Description:

My idea with this is to have it be merged before #1858  to be able to backport to the 2.0.1 release. This isn't really as much of a new feature as it is just the adding necessary build flags to make compilation work on DragonflyBSD.

NOTE: The GOOS value is `dragonfly` and not `dragonflybsd`.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/app_other.go
+++ b/app/app_other.go
@@ -1,4 +1,4 @@
-// +build ci !linux,!darwin,!windows,!freebsd,!openbsd,!netbsd
+// +build ci !linux,!darwin,!windows,!freebsd,!openbsd,!netbsd,!dragonfly
 
 package app
 
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -1,6 +1,6 @@
 // +build !ci
 
-// +build linux openbsd freebsd netbsd
+// +build linux openbsd freebsd netbsd dragonfly
 // +build !android
 
 package app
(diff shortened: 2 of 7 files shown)


Jacalz:
@andydotxyz Do you think that you could have this backported to the 2.0.1 release please?

andydotxyz:
Sure thing, we have not branched the 2.0.x branch yet.

mode you manage to test the fyne cmdline as well as the demo on dragonfly @Jacalz ?

Jacalz:
> did you manage to test the fyne cmdline as well as the demo on dragonfly @Jacalz ?

I tried, but I could not get my dragonfly bsd install to work. There is some sort of issue with the package manager not working at all and has no repositories set up. It is just a fork of FreeBSD, so given that it works on FreeBSD, it should work here as well. This should at least make it possible for anyone using DragonflyBSD to get it up and running.

andydotxyz:
I may be overly cautious but I think we should at least be able to run the changes before merging - especially if it’s to be in a bug fix release soon

Jacalz:
Given that the other bsd variants were added the same way, I guess this is half-fine, but rest assured I will have a second attempt (in the week) at getting it working :)

andydotxyz:
I agree that the code looks ok - but given the variance in distributions it’s no guarantee that it will work. Particularly the configuration location and file monitoring.

Jacalz:
The Go documentation mentions that the dragonflybsd port isn't very stable. Given that in combination that I can't get the package manager to work in the installed os, I will close this.

PR #1898, opened 2021-02-03
Remove custom bundler scripts and rebundle with latest fyne bundle

Jacalz:
### Description:

This removes the `gen.go` custom bundle scripts that were used in `cmd/fyne_demo` and `cmd/fyne_demo` in favor of using `go:generate` in combination with `fyne bundle`.
The only remaining custom bundle script is the one in the theme package, but that requires a lot more work due to the fonts and the large amount of icons.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Neat, thanks

PR #1903, opened 2021-02-04
Fix issue with Focus call crashing

### LINKED ISSUE #1893: Canvas.Focus() before window visible causes application to crash
Here is a part of the related stack trace:
```
2021/02/02 11:36:42 panic was raised: runtime error: invalid memory address or nil pointer dereference, stack: goroutine 27 [running]:
panic(0x46fac20, 0x536b050)
	/usr/local/Cellar/go/1.14.1/libexec/src/runtime/panic.go:967 +0x166
fyne.io/fyne/v2/internal/app.(*FocusManager).Focus(0x0, 0x4c62cc0, 0xc0007f2900, 0x0)
	/Users/ababo/go/pkg/mod/fyne.io/fyne/v2@v2.0.0/internal/app/focus_manager.go:25 +0x4f
fyne.io/fyne/v2/internal/driver/glfw.(*glCanvas).Focus(0xc0002a8100, 0x4c62cc0, 0xc0007f2900)
	/Users/ababo/go/pkg/mod/fyne.io/fyne/v2@v2.0.0/internal/driver/glfw/canvas.go:88 +0x1a8
```
This happens when trying to focus on `widget.Entry` or a custom control based on `widget.List`.

Platform: `OS X 11.1`
Go version: `go1.14.1 darwin/amd64`
Fyne version: `v2.0.0`

andydotxyz:
Remove crash when calling Focus before canvas is visible.
Also fix issue where item focused before visible was not remembered (worked < 2.0.0).

Fixes #1893

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/canvas.go
+++ b/internal/driver/glfw/canvas.go
@@ -94,7 +94,7 @@ func (c *glCanvas) Focus(obj fyne.Focusable) {
 		}
 	}
 
-	c.contentFocusMgr.FocusBeforeAdded(obj) // not found yet assume we are preparing the UI
+	fyne.LogError("Failed to focus object which is not part of the canvas’ content, menu or overlays.", nil)
 }
 
 func (c *glCanvas) Focused() fyne.Focusable {
--- a/internal/driver/glfw/canvas.go
+++ b/internal/driver/glfw/canvas.go
@@ -84,13 +84,17 @@ func (c *glCanvas) Focus(obj fyne.Focusable) {
 	c.RUnlock()
 
 	for _, mgr := range focusMgrs {
+		if mgr == nil {
+			continue
+		}
 		if focusMgr != mgr {
 			if mgr.Focus(obj) {
 				return
 			}
 		}
 	}
-	fyne.LogError("Failed to focus object which is not part of the canvas’ content, menu or overlays.", nil)
+
+	c.contentFocusMgr.FocusBeforeAdded(obj) // not found yet assume we are preparing the UI
 }
 
 func (c *glCanvas) Focused() fyne.Focusable {
@@ -452,7 +456,14 @@ func (c *glCanvas) paint(size fyne.Size) {
 func (c *glCanvas) setContent(content fyne.CanvasObject) {
 	c.content = content
 	c.contentTree = &renderCacheTree{root: &renderCacheNode{obj: c.content}}
+	var focused fyne.Focusable
+	if c.contentFocusMgr != nil {
+		focused = c.contentFocusMgr.Focused() // keep old focus if possible
+	}
 	c.contentFocusMgr = app.NewFocusManager(c.content)
+	if focused != nil {
+		c.contentFocusMgr.Focus(focused)
+	}
 }
 
 func (c *glCanvas) setDirty(dirty bool) {
--- a/internal/driver/glfw/canvas_test.go
+++ b/internal/driver/glfw/canvas_test.go
@@ -216,12 +216,12 @@ func TestGlCanvas_Focus_SetContent(t *testing.T) {
 	w := createWindow("Test")
 	w.SetPadded(false)
 	e := widget.NewEntry()
-	w.SetContent(e)
+	w.SetContent(container.NewHBox(e))
 	c := w.Canvas().(*glCanvas)
 	c.Focus(e)
 	assert.Equal(t, e, c.Focused())
 
-	w.SetContent(e)
+	w.SetContent(container.NewVBox(e))
 	assert.Equal(t, e, c.Focused())
 }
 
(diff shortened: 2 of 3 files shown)


toaster:
> Whether or not this should be supported is an interesting question. However what is more important is:
> "This used to work, did we mean to break it in 2.0.0"
> to which the answer is clearly, No - this was an accident.
> 
> So do we want to proagate the break, and put it in the notes for all breaking changes in 2.0 upgrade notes, or should we fix it beacause it has been supported in the past.

I propose the former. Also, disabling the ability to focus an out-of-tree object was a conscious decision and therefore not an accident. I just did not think that this “feature” would be in use.
The bug in the implementation which led to the crash was an accident, of course.

> 
> Without this feature a develop that wishes an element to be focused will have to get a callback from the `main()` of their app to say that it has set the window, so the focusing can happen. This feels like it is bleeding different concerns of the app.

As soon as a developer has added the object to the content tree, it can be focused. I still cannot see where this is a problem.

internal/driver/glfw/canvas.go
```diff
@@ -452,7 +456,14 @@ func (c *glCanvas) paint(size fyne.Size) {
 func (c *glCanvas) setContent(content fyne.CanvasObject) {
 	c.content = content
 	c.contentTree = &renderCacheTree{root: &renderCacheNode{obj: c.content}}
+	var focused fyne.Focusable
+	if c.contentFocusMgr != nil {
+		focused = c.contentFocusMgr.Focused() // keep old focus if possible
+	}
 	c.contentFocusMgr = app.NewFocusManager(c.content)
+	if focused != nil {
+		c.contentFocusMgr.Focus(focused)
```
toaster, line 465: Reasonable feature, but where is the test?

toaster (changes requested):

internal/driver/glfw/canvas.go
toaster: Also, This could be a different PR :).

andydotxyz:
> As soon as a developer has added the object to the content tree, it can be focused. I still cannot see where this is a problem.

We have discussed this many times now - a complex app probably does not want to expose a widget reference all the way up the tree when creating and/or add a new "you're added" callback for main to call.

Let me propose another example (beyond what the original bug reporter showed us).
With data binding the developer no longer needs a reference to any widget to access values, and so they create the object tree and launch the app. Data binding handles all the data transfer so the application no longer needs any widget references. But the developer needs to somehow reach into the heirarchy to access a specific widget after it is set to the window so that it can be focused.

andydotxyz:
> Also, disabling the ability to focus an out-of-tree object was a conscious

In the future we need to get better at documenting the intentional breaks - it would have been ideal to include this in the upgrade notes so it does not surprise more people.

internal/driver/glfw/canvas.go
andydotxyz: Test added

toaster:
> Let me propose another example (beyond what the original bug reporter showed us).
> With data binding the developer no longer needs a reference to any widget to access values, and so they create the object tree and launch the app. Data binding handles all the data transfer so the application no longer needs any widget references. But the developer needs to somehow reach into the heirarchy to access a specific widget after it is set to the window so that it can be focused.

Understood. Seems we need some elegant solution without pitfalls for this.

toaster:
> > Also, disabling the ability to focus an out-of-tree object was a conscious
> 
> In the future we need to get better at documenting the intentional breaks - it would have been ideal to include this in the upgrade notes so it does not surprise more people.

Agreed.

internal/driver/glfw/canvas_test.go
```diff
@@ -204,6 +204,27 @@ func TestGlCanvas_Focus(t *testing.T) {
...
+func TestGlCanvas_Focus_SetContent(t *testing.T) {
+	w := createWindow("Test")
+	w.SetPadded(false)
+	e := widget.NewEntry()
+	w.SetContent(e)
+	c := w.Canvas().(*glCanvas)
+	c.Focus(e)
+	assert.Equal(t, e, c.Focused())
+
+	w.SetContent(e)
```
toaster, line 224: Because this test will be false positive as soon as someone optimises `SetContent` with `if c == canvas.content { return }`, I propose the following:

```suggestion
	w.SetContent(container.NewHBox(e))
	c := w.Canvas().(*glCanvas)
	c.Focus(e)
	assert.Equal(t, e, c.Focused())

	w.SetContent(container.NewVBox(e))
```

Also, it would be helpful to have a test which checks that `SetContent` resets the focus if the current focused object is not part of the new content.

toaster (changes requested):
Almost there :).

toaster (approved):

PR #1908, opened 2021-02-06
Deprecate the extending of containers

andydotxyz:
This shouldn't really have been supported.
Relates to the accidental breaking of Scroll extending - discussed on call 5th Feb.

### Checklist:

- [ ] Tests included. <- just deprecations
- [x] Tests all pass.

### DIFF
--- a/container/apptabs.go
+++ b/container/apptabs.go
@@ -56,7 +56,7 @@ func NewAppTabs(items ...*TabItem) *AppTabs {
 		// Current is first tab item
 		tabs.current = 0
 	}
-	tabs.ExtendBaseWidget(tabs)
+	tabs.BaseWidget.ExtendBaseWidget(tabs)
 
 	if tabs.mismatchedContent() {
 		internal.LogHint("AppTabs items should all have the same type of content (text, icons or both)")
@@ -86,7 +86,7 @@ func (c *AppTabs) Append(item *TabItem) {
 
 // CreateRenderer is a private method to Fyne which links this widget to its renderer
 func (c *AppTabs) CreateRenderer() fyne.WidgetRenderer {
-	c.ExtendBaseWidget(c)
+	c.BaseWidget.ExtendBaseWidget(c)
 	r := &appTabsRenderer{line: canvas.NewRectangle(theme.ShadowColor()),
 		underline: canvas.NewRectangle(theme.PrimaryColor()), container: c}
 	r.updateTabs()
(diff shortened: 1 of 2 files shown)


container/apptabs.go
```diff
@@ -106,9 +106,16 @@ func (c *AppTabs) CurrentTabIndex() int {
 	return c.current
 }
 
+// ExtendBaseWidget is used by an extending widget to make use of BaseWidget functionality.
+//
+// Deprecated: Support for extending containers is being removed
+func (c *AppTabs) ExtendBaseWidget(wid fyne.Widget) {
```
Jacalz, line 112: I thought we said, on the call yesterday, that we wouldn't add "new" deprecated methods?

container/split.go
```diff
@@ -60,6 +60,13 @@ func (s *Split) CreateRenderer() fyne.WidgetRenderer {
 	}
 }
 
+// ExtendBaseWidget is used by an extending widget to make use of BaseWidget functionality.
+//
+// Deprecated: Support for extending containers is being removed
+func (s *Split) ExtendBaseWidget(wid fyne.Widget) {
```
Jacalz, line 66: I thought we said, on the call yesterday, that we wouldn't add "new" deprecated methods?

Jacalz (commented):
Just a question regarding the implementation which I thought we said something else about on the call yesterday. I am not technically against it, just asking. Aparart from my question, it looks good 👍

container/apptabs.go
andydotxyz: Yes we did. I have not added the new no-op `ExtendBaseWidget` in `Scroll` where it would be required.
These methods are not "new" they are wrappers so that I can mark Deprecated.
Obviously we don't want to set it Deprecated on `widget.BaseWidget` which is what it overrides now.

Jacalz: Ah, I see. Thanks for the clarification.

Jacalz (approved):

container/apptabs.go
```diff
@@ -56,7 +56,7 @@ func NewAppTabs(items ...*TabItem) *AppTabs {
 		// Current is first tab item
 		tabs.current = 0
 	}
-	tabs.ExtendBaseWidget(tabs)
+	tabs.BaseWidget.ExtendBaseWidget(tabs)
```
stuartmscott, line 59: As the Deprecated ExtendBaseWidget is not a noop, are these changes needed?

andydotxyz: With it this change you have deprecation warnings... that seemed undesirable

stuartmscott: Ah fair enough

stuartmscott (approved):

PR #1909, opened 2021-02-06
Implement  control of tab entry

adrianre12:
### Description:
This adds the ability for Widgets to control if the Tab key is either passed to the Widget via the `Focusable ` interface or is used to switch focus.
I have implemented the `Taggable ` interface which requires the hook function `AcceptTags()` which is called by the driver to determine if Tab key events should be sent tot he Widget. There is also the function `SetAcceptTags()` which is used to set the internal state for the Widget.
I have implemented this on `widget.Entry` with the following behaviour;

- Default: `SetAcceptTags()` has not been called. Tabs will only be accepted if `Multiline` is set. This is the same behaviour as now.
- `SetAcceptTags()` has been called. Acceptance of Tabs follows what is set, the state of `Multiline` is ignored. This allows single line Entries to accept Tabs and for Multiline Entries to reject Tabs.

### Checklist:
- [X] Tests included.
- [X] Tests all pass.

#### Where applicable:


widget/entry.go
```diff
@@ -403,6 +415,16 @@ func (e *Entry) SelectedText() string {
 	return string(e.textProvider().buffer[start:stop])
 }
 
+// SetAcceptTabs sets if Entry will accept tabs for text entry.
+// This is overrides the default behaviour of Multiline Entries accepting tabs
+func (e *Entry) SetAcceptTabs(acceptTabs bool) {
+	if acceptTabs {
+		e.acceptTabs = 1
```
Jacalz, line 422: You can add a `return` here and remove the else statement.

Jacalz (changes requested):
Just a small recommendation. Apart from that, it looks fine to me.

widget/entry.go
adrianre12: Done

Jacalz (approved):
Looks good to me, but it is probably good if someone else has a look as well before merging.

canvasobject.go
```diff
@@ -96,3 +96,12 @@ type Focusable interface {
 type Shortcutable interface {
 	TypedShortcut(Shortcut)
 }
+
+// Tabable describes any object that needs to accept the Tab key presses.
+type Tabable interface {
+	// AcceptTabs() is a hook called by the key press handling logic.
+	// If it returns true then the Tab key events will be sent TypedKey and TypedRun.
```
andydotxyz, line 103: Typo of Rune?

canvasobject.go
```diff
@@ -96,3 +96,12 @@ type Focusable interface {
...
+	AcceptTabs() bool
+	// SetAcceptTabs() sets the value returned by AcceptTabs().
+	SetAcceptTabs(bool)
```
andydotxyz, line 106: Is this method required? These interfaces are implemented by widgets, should they not be in control of whether tabs are accepted?

andydotxyz (changes requested):
One change and one more open question

andydotxyz:
Regarding my comment about `SetAcceptsTab` if it is normally not something to be customised we encourage not exposing setter because people can extend Entry and override the return. This would avoid the three-state boolean

canvasobject.go
adrianre12: Do you mean the SetAcceptTabs()? I included it as part of the interface so any widget would have a uniform implementation of turning it on and off. Without it, we would be in a similar state to now, where a widget design decision limits the use of the widget. Having it encourages widget writers to at least consider the tab/no-tab question even if in the end they just implement it as a null function with a suitable comment.
The default state of AcceptTabs() is up to the widget.

canvasobject.go
adrianre12: Yeup, well spotted

canvasobject.go
andydotxyz: > Without it, we would be in a similar state to now, where a widget design decision limits the use of the widget.

I don't really follow / agree with this statement. Widget writers are not compelled to use this interface, it is opt-in.
When they decide they want to handle tap then they should implement so that `AcceptsTabs` is covered. exposing `SetAcceptTabs` is in fact leaving the decision open to /app developers/ which is what I don't think we should encourage. If they truly want to override the behaviour they could do so by extension, as I mentioned in the other comment.

adrianre12: The tri-state is to allow Entry to maintain its current functionality, but allow it to be overridden and avoid a breaking change. It would be a lot simpler to decouple the tabs from MultiLine and just give Entry a field called say EnableTabs. 
Re SetAcceptTabs, I was looking at it from the contributor's point of view, not an end-user. Not a problem to drop it from the interface now I see your point.

andydotxyz: > The tri-state is to allow Entry to maintain its current functionality, but allow it to be overridden

Yes I see how it is needed with the current code, but if you remove `SetAcceptTabs` then it is no longer needed becase a developer wishing to change behaviour will override it to make their change.

adrianre12: Huh? do I read you right? You want to make it so users have to extend widget.Entry just to turn on/off tab handling?  Isn't that rather tortuous?

andydotxyz: Our API design is that common use-cases get core API and others are acheived through extending the widget types.

So it hinges on the question - do we want to encourage developers to change how tabs function in standard widgets?

You can compare this to `Cursorable` - it sets a good default and allows developers to change through extension (as it's not super common).

andydotxyz: Bear in mind that this simple "turn off tabs" is basically changing how the built-in focus handling works and will therefore impact accessibility controls (when it appears) and the general keyboard based control of an app.

andydotxyz:
Just to clarify (as it has been a while with a few changes) I don't think we should expose `Entry.SetAcceptTabs`.
Developers who want to override the response from `Entry.AcceptsTabs` can do so through extension, this is the pattern we follow for not-common use-cases.

adrianre12:
`Entry.SetAcceptTabs` removed

canvasobject.go
```diff
@@ -96,3 +96,10 @@ type Focusable interface {
 type Shortcutable interface {
 	TypedShortcut(Shortcut)
 }
+
+// Tabable describes any object that needs to accept the Tab key presses.
+type Tabable interface {
```
andydotxyz, line 101: This needs to have the API version added, sorry I missed this before. Add the following to the doc:
```go
\\
\\ Since: 2.1
```

widget/entry.go
```diff
@@ -114,6 +115,13 @@ func NewPasswordEntry() *Entry {
 	return e
 }
 
+// AcceptTabs returns if Entry accepts Tabs or not.
+//
+// Implements: fyne.Tabable
+func (e *Entry) AcceptTabs() bool {
```
andydotxyz, line 121: This should also show the since line as above

andydotxyz (changes requested):
Looks good, but I missed the need for API version in previous comments, sorry.

widget/entry.go
andydotxyz: Just a ping. The PR looks good except this is missing the since still.

andydotxyz:
I can add the final `Since:` tag for @adrianre12, I guess he is busy.

andydotxyz (approved):

adrianre12:
yeh sorry busy with other things, but other PRs are still in my todo.

andydotxyz:
> yeh sorry busy with other things, but other PRs are still in my todo.

Not a problem, some of the others require a bit more work, but this one I figured I could land for you :).

Issue #1910, opened 2021-02-06
`fyne` doesn't remove temporary binary from `src`


stuartmscott:
### Describe the bug:
After running `fyne get <package>` a binary is left in the `src` directory.

### To Reproduce:
Steps to reproduce the behaviour:
1. `fyne get aletheiaware.com/spacefynego/cmd/spacefyne`
2. Go on $GOPATH/src/aletheiaware/spacefynego
3. Either run `git status` or `ls cmd/spacefyne`
4. See the `spacefyne` binary

### Device (please complete the following information):
 - **OS:** Ubuntu & MacOS
 - **Version:** 20.04 Focal & Big Sur
 - **Go version:** go1.15.7 linux/amd64 & go1.15 darwin/amd64
 - **Fyne version:** 9a0f87ce

Jacalz:
Seems to apply to the whole build command in itself when  trying to package an application. Compiling with `--release` does not use a release compiled binary unless the one there is removed first.

stuartmscott:
Is `--release` different from `fyne release`?

Jacalz:
Yes `fyne package —release` basically strips the debug symbols form the binaries and not much more, I think.

stuartmscott:
Oh thanks. It would be great if `fyne release` stripped debug symbols too. We also probably don't need two commands for creating releases. Though both of these should be separate tickets.

Jacalz:
Yeah. We might want to have it as `--strip-debug` instead, or something along those lines. Like you say though, something for another ticket.

andydotxyz:
`fyne release` also strips symbols. 
The difference between `fyne release` and `fyne package —release` is only the build artefact, in all other ways they are the same. The release mode is expected for distribution through official stores, package is for local instal or sharing on ad-hoc systems.

andydotxyz:
On `release/v2.0.x` for testing

PR #1918, opened 2021-02-07
Fix release builds not removing debug symbols

Jacalz:
### Description:

I noticed on my linux computer that packaging and adding `--release` has no effect.
This change tidies up the code a bit and makes sure that the `-ldflags` call get the correct information.

I guess we might want this change for 2.0.1 as well, but that might not be possible to backport as is due to the changes after the cli rework.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
This seems to  have been caused by #1910. Will close this. Will have to punt the cleanups to some other PR.

PR #1919, opened 2021-02-07
Proposal: Switch from gz to xz compression for unix packages

Jacalz:
### Description:

Nowadays, `xz` compression should be available on default on most Linux and BSD systems.
Switching to `xz` gives, from my testing, roughly 20-30% smaller files at the expense of slightly longer compression time.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/package-unix.go
+++ b/cmd/fyne/internal/commands/package-unix.go
@@ -66,9 +66,8 @@ func (p *Packager) packageUNIX() error {
 			return errors.Wrap(err, "Failed to write Makefile string")
 		}
 
-		tarCmd := exec.Command("tar", "zcf", p.name+".tar.gz", "-C", tempDir, "usr", "Makefile")
-		err = tarCmd.Run()
-		if err != nil {
+		tarCmd := exec.Command("tar", "-Jcf", p.name+".tar.xz", "-C", tempDir, "usr", "Makefile")
+		if err = tarCmd.Run(); err != nil {
 			return errors.Wrap(err, "Failed to create archive with tar")
 		}
 	}


adrianre12:
How sure are you that xz is supported on all the Linux flavors Fyne is used on? Might it be pragmatic to make they use of xz or gz an option?

Jacalz:
I guess you can't really be 100% sure given that Fyne basically can be compiled and run in any Linux distro, but the availability is pretty good from what I can gather. Xz compression is the default in Gnome Nautilus (file manager) and I can't remember when I last used a Linux distro without support for it. Either way, I will do some digging and check out the major players like Ubuntu, Debian, Fedora and alike to gather some more information.

Jacalz:
I have done quite an elaborate test against various distributions (checked means that it has xz installed).
The systems were tested with different desktops to be more sure:

- [x] Fedora Workstation using Gnome
- [x] Ubuntu using Gnome
- [x] Debian using Gnome
- [ ] Void Linux using both Enlightment and Cinnamon
- [x] Linux Mint (Ubuntu Based) using CInnamon
- [x] Endevour OS (Arch Based) using Xfce
- [x] Sparky Linux (Debian Based) using Lxqt
- [x] Manjaro (Arch based) using KDE Plasma 5
- [x] Solus using Budgie
- [x] GhostBSD (FreeBSD based) using Mate
- [x] OpenMandriva using KDE Plasma 5
- [x] MX Linux (Debian based) using Xfce

From all 12 distributions tested, Void Linux was the only one that I could find to not include `xz` by default. I think that is looking really good, especially as installing `xz` on Void only is one terminal command away.

andydotxyz:
I'm assuming those are latest distros?
Is the situation the same if you take releases from 2 or 3 years back?
The oldest version that supports Go 1.12 would be a reasonable target I'd guess as we have no other upgrad requirement.

adrianre12:
Is the reason to change to XZ to reduce the size? could running GZ at a higher compression give similar size benefits?

Jacalz:
> Is the reason to change to XZ to reduce the size? could running GZ at a higher compression give similar size benefits?

Yes that is the reason. I had already tried higher gz compression and it didn’t even make a 0.1 megabyte difference in my tests. That was why I opened this PR. Will dig into some older LTS Linux releases to make sure that availability is good.

Jacalz:
I gave the following versions a try (checked is again marked as having `xz` installed:

- [x] OpenSuse Leap 15.2 using KDE and Gnome (initially from 2018)
- [x] Ubuntu 18.04.5 using Gnome (initially from 2018)
- [x] Mageia 7.1 using Xfce (from 2019)

All this testing of various distributions in a VM really reminds me of the good old days when I was distro-hopping. Either way, I don't really know what more to tests here. All of these (except Mageia) are older than Go 1.12 and does not support compiling Fyne, yet they have `xz` installed. I'd say that the availability is looking really good.

Jacalz:
Some more interesting data that I have found:
- MacOS supports archiving (through `tar`) and extracting `.tar.xz` files (from Finder and using `tar`).
- All the compressed images from FreeBSD are compressed using `xz`.

Again, I think this looks really good but I have added it to the agenda for today's call to make sure that everyone is okay with it :slightly_smiling_face:

stuartmscott:
I'm no expert but from https://en.wikipedia.org/wiki/XZ_Utils;

> In most cases, xz achieves higher compression rates than alternatives like gzip and bzip2. Decompression speed is faster than bzip2, but slower than gzip. Compression can be much slower than gzip, and is slower than bzip2 for high levels of compression, and is most useful when a compressed file will be used many times.

So if we're optimizing for memory then this PR makes sense, but not if we're optimizing for CPU. Yup let's discuss on the call ☎️

andydotxyz (approved):

Jacalz:
This was agreed as a good idea on todays call. Merging 🙂

PR #1920, opened 2021-02-07
The onConfirm callback should be called only when user click on 'Ok' of a EntryDialog dialog.

zdima:
### Description:

The onConfirm callback is getting called whether user click on Ok or Cancel button of the EntryDialog.
Such behavior doesn't allow to distinguish between Ok and Cancel actions.

The expected behavior is when onConfirm called only when use click on Ok button.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/entry.go
+++ b/dialog/entry.go
@@ -49,9 +49,9 @@ func (i *EntryDialog) SetOnClosed(callback func()) {
 func NewEntryDialog(title, message string, onConfirm func(string), parent fyne.Window) *EntryDialog {
 	i := &EntryDialog{entry: widget.NewEntry()}
 	items := []*widget.FormItem{widget.NewFormItem(message, i.entry)}
-	i.formDialog = NewForm(title, "Ok", "Cancel", items, func(_ bool) {
+	i.formDialog = NewForm(title, "Ok", "Cancel", items, func(ok bool) {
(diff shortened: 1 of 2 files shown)


Jacalz (approved):
Good catch. Looks to be my fault from when I reworked the Entry Dialog to use form when deprecating it. Given that this is a bugfix to a slight behavior change, I think this should be backported to 2.0.1.

stuartmscott (approved):
Unfortunately the dev has no way to handle the case where the dialog was cancelled. Perhaps an `onCancelled func()` should be added, or `onConfirm func(string)` should become `callback func(string, bool)`, but either way is an API break.

However this limitation was not introduced by this PR, so approved.

Jacalz:
> Unfortunately the dev has no way to handle the case where the dialog was cancelled. Perhaps an `onCancelled func()` should be added, or `onConfirm func(string)` should become `callback func(string, bool)`, but either way is an API break.
> 

That is true, but I am not entirely sure if we should add new features to a deprecated dialog.

andydotxyz:
> Unfortunately the dev has no way to handle the case where the dialog was cancelled

I don’t think this is the case. All dialogs have `OnClosed` if you want to be informed that it went away.

PR #1922, opened 2021-02-07
fix: making gl thing to compile and work on apple silicon

### LINKED ISSUE #1739: Error compiling fyne on Apple M1 arm64
Cannot compile fyne_demo on Apple M1 (if it matters).

I am following the procedure described at the [tour](https://developer.fyne.io/started/#run-demo), got an error message:

```
GO111MODULE=on go get fyne.io/fyne/cmd/fyne_demo
go: downloading github.com/go-gl/glfw/v3.3/glfw v0.0.0-20200625191551-73d3c3675aa3
# github.com/go-gl/gl/v3.1/gles2
go/pkg/mod/github.com/go-gl/gl@v0.0.0-20190320180904-bf2b1f2f34d7/v3.1/gles2/package.go:38:11: fatal error: 'KHR/khrplatform.h' file not found
 #include <KHR/khrplatform.h>
          ^~~~~~~~~~~~~~~~~~~
1 error generated.
```

 - OS: MacOS
 - Version: Big Sur 11.1
 - Go version: go version go1.16beta1 darwin/arm64
 - Fyne version: v1.4.3

sirkon:
### Description:

Changes to make fyne to compile and work on Apple Silicon. It is based on recent changes in Go, where GOOS for iPhones and iPads was changed to ios and GOOS=darwin now only refers to MacOS. https://github.com/golang/go/issues/42100

Fixes #1739

PS This is not a final solution, rather show a way of workable configuration to build for Apple Silicon.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/glfw_es.go
+++ b/internal/driver/glfw/glfw_es.go
@@ -1,5 +1,5 @@
 // +build gles arm arm64
-// +build !arm64,!darwin
+// +build !arm64,darwin arm64,!darwin
 
 package glfw
 
--- a/internal/driver/glfw/glfw_es.go
+++ b/internal/driver/glfw/glfw_es.go
@@ -1,5 +1,5 @@
 // +build gles arm arm64
-// +build !arm64,darwin arm64,!darwin
+// +build !darwin
 
 package glfw
 
--- a/internal/driver/glfw/glfw_es.go
+++ b/internal/driver/glfw/glfw_es.go
@@ -1,4 +1,5 @@
 // +build gles arm arm64
+// +build !arm64,!darwin
 
 package glfw
 
--- a/internal/painter/gl/gl_core.go
+++ b/internal/painter/gl/gl_core.go
@@ -1,4 +1,4 @@
-// +build !gles,!arm,!arm64,!android,!ios,!mobile arm64,darwin
+// +build !gles,!arm,!arm64,!android,!ios,!mobile darwin
 
 package gl
 
--- a/internal/painter/gl/gl_core.go
+++ b/internal/painter/gl/gl_core.go
@@ -1,4 +1,4 @@
-// +build !gles,!arm,!android,!ios,!mobile amr64,darwin
+// +build !gles,!arm,!arm64,!android,!ios,!mobile arm64,darwin
 
 package gl
 
--- a/internal/painter/gl/gl_core.go
+++ b/internal/painter/gl/gl_core.go
@@ -1,4 +1,4 @@
-// +build !gles,!arm,!arm64,!android,!ios,!mobile
+// +build !gles,!arm,!android,!ios,!mobile amr64,darwin
 
 package gl
 
--- a/internal/painter/gl/gl_es.go
+++ b/internal/painter/gl/gl_es.go
@@ -1,6 +1,6 @@
 // +build gles arm arm64
 // +build !android,!ios,!mobile
-// +build !arm64,darwin arm64,!darwin
+// +build !darwin
 
 package gl
 
--- a/internal/painter/gl/gl_es.go
+++ b/internal/painter/gl/gl_es.go
@@ -1,4 +1,4 @@
-// +build gles arm arm64
+// +build gles arm arm64,!darwin
 // +build !android,!ios,!mobile
 
 package gl
--- a/internal/painter/gl/gl_es.go
+++ b/internal/painter/gl/gl_es.go
@@ -1,5 +1,6 @@
-// +build gles arm arm64,!darwin
+// +build gles arm arm64
 // +build !android,!ios,!mobile
+// +build !arm64,darwin arm64,!darwin
 
 package gl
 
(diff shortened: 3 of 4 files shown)


internal/painter/gl/gl_core.go
```diff
@@ -1,4 +1,4 @@
-// +build !gles,!arm,!arm64,!android,!ios,!mobile
```
andydotxyz, line 1: I think this change is not quite right.
If I have a linux arm64 this file should not be activated, but with that removed it will

andydotxyz (changes requested):
This is a great fix, thanks - but one of the build rules looks off (mentioned inline).

internal/painter/gl/gl_core.go
sirkon: Yes, know this. Some options were just off. Rewrite all things with the following ideas in mind:

1. `*_core.go` files all getting `x OR (arm64 AND darwin)`, this means they are to be used with MacOS on Apple M1.
2. `*_es.go` files all getting `x AND (!arm64 AND darwin OR arm64 !darwin). This means they are excluded when compiling for MacOS on Apple M1.

Where `x` represents the existing set conditions.

BTW, `!arm64 AND darwin OR arm64 !darwin` == `!(arms64 AND darwin)`, this is a rule of boolean algebra. 
In fact I could simplify the negation in _es files for Apple M1 if keeping existing conditions in mind, but decided to stay it this way to keep things cleaner. You will be able to put another pairs if needed in inclusive conditions and append lines with exclusive conditions with just copy-pasting and minor editing.

andydotxyz: I think this has overcomplicated it now, sorry. the ES stuff should never run on darwin, and the core always should (as iOS is a different flag). All of the others were OK as far as I could see, just this one not.
Re-adding the "!arm64" to this line should and adding "arm64,darwin" at the end should have done it.

andydotxyz: I think the new change for this file is correct - but all of the other updates were not needed.
I'm not certain that your summary of boolean algebra is correct - you seem to have described exclusive or, which is not `!(x and y)`.

andydotxyz (changes requested):
I think it was closer before (see inline)

internal/painter/gl/gl_es.go
```diff
@@ -1,5 +1,6 @@
-// +build gles arm arm64,!darwin
+// +build gles arm arm64
```
zdima, line 1: the previous change in this file was good.
it is compiled on M1 with previous changes.

internal/driver/glfw/glfw_es.go
```diff
@@ -1,5 +1,5 @@
 // +build gles arm arm64
-// +build !arm64,!darwin
+// +build !arm64,darwin arm64,!darwin
```
zdima: based on previous comments, I think this should be just one line:
// +build gles arm arm64,!darwin
i.e. build on any arm64 but not Darwin.

zdima (commented):
I still having issue with "service port" when application start.
GLFW: An uncaught error has occurred: PlatformError: Cocoa: Failed to find service port for display. Do I have to pull other versions of GL*?

internal/painter/gl/gl_core.go
sirkon: > which is not !(x and y)

shit, indeed. Turned all conditions into `x AND !darwin` for _es and `x OR darwin` for _core

sirkon: Keep in mind though, this will work for Go > 1.16 only.

andydotxyz: I cannot see which elements of this change are only supporting Go 1.16?
We support back to 1.12, we will upgrade to 1.14 or 1.15 once Debian releases `testing`.

sirkon: `GOOS=darwin` for iOS prior Go 1.16. It is `GOOS=ios` since 1.16.

andydotxyz: Hmm, we embed the `gomobile` project for building `ios` and that sets the tag.
Does that mean that there is no problem?

sirkon: Looks like this indeed

andydotxyz:
Could we get @ipfans to test this as well please?

@zdima some of the messages above are not totally clear. Can you confirm whether or not this latest version works for you please?

zdima:
> Could we get @ipfans to test this as well please?
> 
> @zdima some of the messages above are not totally clear. Can you confirm whether or not this latest version works for you please?

I adde to the go.mod in github.com/fyne-io/examples:
`replace fyne.io/fyne/v2 => github.com/sirkon/fyne/v2 v2.0.1-0.20210209074931-f93d4e04b627`

Run:
`
go clean -I -r;
go mod tidy;
go build
./examples
`
Project compiled without issues and program runs.
There is output on the console:
```
$ ./examples 
GLFW: An uncaught error has occurred: PlatformError: Cocoa: Failed to find service port for display
GLFW: Please report this bug in the Go package immediately.
2021/02/09 11:12:35 PlatformError: Cocoa: Failed to find service port for display
```

andydotxyz (approved):

PR #1924, opened 2021-02-08
Fix FileIcon flicker/disappear

andydotxyz:
Re-using objects instead of recreating fixed some glitches when calling SetURI on small icons (in a tree).

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/fileicon.go
+++ b/widget/fileicon.go
@@ -78,7 +78,12 @@ func (i *FileIcon) CreateRenderer() fyne.WidgetRenderer {
 	background.Hide()
 
 	s := &fileIconRenderer{file: i, background: background}
-	s.updateObjects()
+	s.img = canvas.NewImageFromResource(s.file.resource)
+	s.img.FillMode = canvas.ImageFillContain
+	s.ext = canvas.NewText(s.file.extension, theme.BackgroundColor())
+	s.ext.Alignment = fyne.TextAlignCenter
+
+	s.SetObjects([]fyne.CanvasObject{s.background, s.img, s.ext})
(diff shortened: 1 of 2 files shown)


widget/fileicon.go
```diff
@@ -161,9 +166,11 @@ func (s *fileIconRenderer) Refresh() {
 	if s.file.URI != s.file.cachedURI {
 		s.file.propertyLock.RLock()
 		s.file.setURI(s.file.URI)
-		s.updateObjects()
 		s.file.cachedURI = s.file.URI
 		s.file.propertyLock.RUnlock()
+
+		s.img.Resource = s.file.resource
+		s.ext.Text = s.file.extension
```
stuartmscott, line 173: Should these not be read with the read lock?

widget/fileicon.go
```diff
@@ -161,9 +166,11 @@ func (s *fileIconRenderer) Refresh() {
 	if s.file.URI != s.file.cachedURI {
 		s.file.propertyLock.RLock()
 		s.file.setURI(s.file.URI)
```
stuartmscott, line 168: I know this is outside of this PR but I wonder if setURI should be called with the write lock as it changing fields, also cachedURI is read without a lock and written with a read lock

widget/fileicon.go
andydotxyz: Good catch thanks

stuartmscott (approved):

Jacalz (approved):
Looks good to me. Thanks for the fix :+1:

PR #1931, opened 2021-02-09
Refresh dialogs correctly, fixes #1866

### LINKED ISSUE #1866: Dialogs are not refreshed correctly on .Show()
### Describe the bug:

The dialogs are not correctly refreshed when calling .Show() on a dialog that has already been created. This can be indicated by saving a reference to a dialog, hiding that dialog and then either changing the theme or resizing the window causes it to have the wrong colour or being at the wrong screen position.

### To Reproduce:
1. Save a reference to a dialog
2. Open the dialog
3. Resize the window and/or change theme
4. Notice incorrect size, position and/or wrong theme

### Screenshots:

dialog-issues

### Example code:

```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/dialog"
	"fyne.io/fyne/v2/theme"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Dialog resize")
	a.Settings().SetTheme(theme.LightTheme())

	themes := "light"
	ctheme := widget.NewButton("Change theme", func() {
		if themes == "dark" {
			a.Settings().SetTheme(theme.LightTheme())
			themes = "light"
			return
		}

		a.Settings().SetTheme(theme.DarkTheme())
		themes = "dark"
	})

	dialogs := dialog.NewInformation("Close me", "Close me and then resize window and/or change theme", w)

		ctheme,
		widget.NewButton("Open dialog", func() {
			dialogs.Show()
		}),
	))

	w.ShowAndRun()
}
```

 - **OS:** Linux
 - **Version:** 5.10.9
 - **Go version:** 1.15.7
 - **Fyne version:** 2.0.0

fpabl0:
### Description:

Fixes #1866 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.
- [x] Move `win.Refresh` from `dialog` to `widget.PopUp`.
- [x] Add popup tests.


fpabl0:
I have not updated the test images, because I am not sure why those images were correctly before:

Before: 
dialog-custom-ugly

After:
dialog-custom-ugly

dialog/base.go
```diff
@@ -104,6 +104,7 @@ func (d *dialog) Show() {
 		d.win.Resize(d.desiredSize)
 	}
 	d.win.Show()
+	d.Refresh()
```
andydotxyz, line 107: Refresh would surely be better placed before show to avoid possible flicker?

dialog/base.go
```diff
@@ -245,27 +246,28 @@ func newButtonList(buttons ...*widget.Button) fyne.CanvasObject {
 	return list
 }
 
-// dialogTitle is really just a normal title but we use the Refresh() hook to update the background rectangle.
-type dialogTitle struct {
-	widget.Label
+// TODO I don't think it is needed
```
andydotxyz, line 249: I agree that this is not needed.
But please don't commit commented out code.

widget/popup.go
```diff
@@ -203,6 +203,7 @@ func (r *popUpRenderer) Refresh() {
 	if r.background.Size() != r.popUp.innerSize || r.background.Position() != r.popUp.innerPos {
 		r.Layout(r.popUp.Size())
 	}
+	// TODO should fix this refresh too?
```
andydotxyz, line 206: I don't know what this comment means, what is broken about the background refresh?

widget/popup.go
```diff
@@ -241,5 +242,9 @@ func (r *modalPopUpRenderer) Refresh() {
 	if r.background.Size() != r.popUp.innerSize {
 		r.Layout(r.popUp.Size())
 	}
+	if !r.popUp.Canvas.Size().Subtract(r.popUp.BaseWidget.Size()).IsZero() {
+		r.popUp.BaseWidget.Resize(r.popUp.Canvas.Size())
+	}
+	r.popUp.Content.Refresh()
```
andydotxyz, line 248: Why is this refresh needed?
That invalidates all the caches for the whole dialog content which could be quite expensive.

andydotxyz (changes requested):

dialog/base.go
fpabl0: Agreed. I will change it.

dialog/base.go
fpabl0: Sorry, I will remove it.

widget/popup.go
fpabl0: I don't write it well, but it means if we should implemented in the same way as modalPopupRenderer.Refresh.

widget/popup.go
fpabl0: In my understanding, if something in a parent widget change, it should updates its children. This refresh guarantees the popup updates all its Content if theme changes.

dialog/base.go
fpabl0: done

dialog/base.go
fpabl0: done

widget/popup.go
andydotxyz: The theme application is done at the driver level and it forces all widgets to refresh.
A widget should only be refreshed if you know that its content must be re-drawn.

If the only change is the resize then it will cause a refresh on the elements that need to refresh on resize will do so themselves.

andydotxyz: I see, this was not visible when theme applied so was not updated.
OK

toaster (changes requested):
The change in the theme test images is expected due to the bugfix.
Nevertheless, I’d prefer to have a new test which checks the particular situation #1866 describes.
The test should:
1. render the shown dialogue with theme A and check the image
2. hide the dialogue
3. switch the theme
4. show the dialogue and check the image
5. hide the dialogue again
6. switch the theme back
7. show the dialogue and check the image

toaster:
@andydotxyz:
After examining this, I found that the `dialog` rendering looks good by chance.
I want to refactor it (for 2.1) to create a render subtree that reflects the expected output better. Also, I’d like to separate `dialog` from its layout. Any objections?

widget/popup.go
toaster: @fpabl0 if you can think of a test which produces output that will change if you apply additional refresh here, then the answer to your question is “yes”.
Alternatively, it might be sufficient if the test which I requested with my review detects a change if you apply the refresh here.

fpabl0:
> I’d prefer to have a new test which checks the particular situation #1866 describes.

Got it. I will do it.
After looking carefully, I think the popup refresh should occur in `popup.Show` and not in `dialog.Show` because it is more a problem with widget.Popup than with dialog.

Basically it would mean to have this in dialog:
```go
func (d *dialog) Show() {
	if !d.desiredSize.IsZero() {
		d.win.Resize(d.desiredSize)
	}
	d.applyTheme() // <<< instead of d.Refresh()
	d.win.Show()
}
```
And this in widget.Popup:
```go
// Show this pop-up as overlay if not already shown.
func (p *PopUp) Show() {
	if !p.overlayShown {
		if p.Size().IsZero() {
			p.Resize(p.MinSize())
		}
		p.Canvas.Overlays().Add(p)
		p.overlayShown = true
	}
	p.Refresh() // <<< added
	p.BaseWidget.Show()
}
```
What do you think? @toaster 

> Also, I’d like to separate dialog from its layout.

I think this part was already done by @andydotxyz in one of the latest commits.

toaster:
> After looking carefully, I think the popup refresh should occur in `popup.Show` and not in `dialog.Show` because it is more a problem of the widget.Popup than the dialog.
> 
[…]
> 
> What do you think? @toaster

I think this whole thing is more a problem of all outside tree widgets and theme changes. Currently, we don’t have a public interface to apply theme changes and `Refresh` is not necessarily recursive.
As a workaround™ (probably in place forever ;)), I suggest to call `internal.app.ApplyThemeTo()` within `BaseWidget`’s `Show()` method.
@andydotxyz WDYT?

andydotxyz:
> Currently, we don’t have a public interface to apply theme changes and Refresh is not necessarily recursive.

We removed that because it was confusing/duplication.
The framework should ensure that `Refresh()` is called on all widgets through ApplyThemeTo.
See the code in `internal/app/theme.go` when the settings change the theme should already be applied.

If the dialog is not in the overlay list then it probably should ensure the same state by using 
`ApplyThemeTo(overlay, window.Canvas())`. I don't think this should be put in `BaseWidget` as it only applies to overlay layers.

andydotxyz:
> > Also, I’d like to separate dialog from its layout.

> I think this part was already done by @andydotxyz in one of the latest commits.

Yes this has already been done.

dialog/base.go
```diff
@@ -260,7 +268,7 @@ func (l *dialogLayout) Layout(obj []fyne.CanvasObject, size fyne.Size) {
 	l.d.bg.Move(fyne.NewPos(0, 0))
 	l.d.bg.Resize(size)
 
-	btnMin := obj[3].MinSize().Max(obj[3].Size())
+	btnMin := obj[3].MinSize()
```
fpabl0, line 271: I have removed the part `.Max(obj[3].Size())` because it makes the dialog button ignore the theme change (sizing variables). @andydotxyz do you know if there was a reason to have that part? 
By removing it, a test image from the color picker changed, however it seems that before this change, that test image was wrong:
Left side is after this change and right side is before this change:

Captura de Pantalla 2021-02-16 a la(s) 01 36 49

dialog/base.go
```diff
@@ -40,7 +40,7 @@ type dialog struct {
 	desiredSize fyne.Size
 
 	win            *widget.PopUp
-	bg             *canvas.Rectangle
+	bg             *themedBackground
```
fpabl0, line 43: @andydotxyz I finally realized why `dialogTitle` was important (although it was clearly written in comments what it does, I didn't understand before). `dialogTitle` had the mission to update the background based on the theme. So, should I keep `dialogTitle` or is it better to have a widget called `themedBackground` as the one added here?

Personally, I don't like `dialogTitle` because it is a bit confusing (it is a child that updates its parent), although clearly it is the shortest solution.

widget/popup.go
fpabl0: Ok, after some tests, popUpRenderer needs the same fix otherwise we are going to have the same issue as in dialog. So I will add popup tests too.

toaster:
> If the dialog is not in the overlay list then it probably should ensure the same state by using
> `ApplyThemeTo(overlay, window.Canvas())`. I don't think this should be put in `BaseWidget` as it only applies to overlay layers.

Yes, as long as a dev does not remove an object from the tree, keeps it and adds it again later on. However, this is quite unlikely.

dialog/base.go
```diff
@@ -228,6 +221,45 @@ func newButtonList(buttons ...*widget.Button) fyne.CanvasObject {
...
+type themedBackgroundRenderer struct {
+	rect    *canvas.Rectangle
+	objects []fyne.CanvasObject
+}
+
+func (renderer *themedBackgroundRenderer) Objects() []fyne.CanvasObject { return renderer.objects }
+func (renderer *themedBackgroundRenderer) Destroy()                     {}
+func (renderer *themedBackgroundRenderer) Layout(size fyne.Size)        { renderer.rect.Resize(size) }
+func (renderer *themedBackgroundRenderer) MinSize() fyne.Size           { return renderer.rect.MinSize() }
+func (renderer *themedBackgroundRenderer) Refresh() {
```
toaster, line 253: AFAIK we do not have one-liner methods up to now. I’m not happy with those, probably we should come to a core team decision and add it to the style guidelines, @andydotxyz?

Also, the methods should be ordered alphabetically.

toaster (changes requested):

dialog/base.go
fpabl0: > AFAIK we do not have one-liner methods up to now. I’m not happy with those

I see, not a problem, I can update them :). I used that style to avoid using too much space, but maybe the other form is more clear.

> Also, the methods should be ordered alphabetically.

Got it. I forgot to sort them.

andydotxyz:
> Yes, as long as a dev does not remove an object from the tree, keeps it and adds it again later on. However, this is quite unlikely.

If required we could add that if an element has been hidden for a while it should have `Refresh()` called?
If we start removing renderers from hidden widgets after a time this may even happen automatically.

dialog/base.go
andydotxyz: Now that `BackgroundColor` is removed from `WidgetRenderer` is this `bg` needed at all?

dialog/base.go
andydotxyz: I don't recall - but this is the change https://github.com/fyne-io/fyne/commit/72cac28414e59cd6d6dd09fddfeb52caade7c4d2

dialog/base.go
andydotxyz: I added:

```
* Functions and other scope blocks (if, for etc) should not be arranged on one line - a scope body belongs on a new line.
```

dialog/base.go
fpabl0: Hmm I see, that change breaks `dialog.NewProgress...`, however:
`btnMin := obj[3].MinSize().Max(obj[3].Size())`
Isn't the above code breaking the principle that objects inside a Layout cannot manipulate manually their size neither their position? @andydotxyz

dialog/base.go
fpabl0: I would say it is still needed, because that background is used to add transparency to the 'background' icon:

With background:
Captura de Pantalla 2021-02-18 a la(s) 16 08 19

Without background:
Captura de Pantalla 2021-02-18 a la(s) 16 07 14

dialog/base.go
fpabl0: done

dialog/base.go
fpabl0: Also, I have seen that both `dialog.NewProgress` and `dialog.NewProgressInfinite` are deprecated and the message says:
```go
// Deprecated: Create a new custom dialog with a widget.ProgressBarInfinite() inside.
```
So using `dialog.NewCustom` the widget would be inside dialog content and not as a button as it is done currently inside `dialog.NewProgress`. 
However it is still possible to make the dialog.NewProgress looks like before, doing this:
```go
c := container.NewVBox(widget.NewProgressBarInfinite())
d := dialog.NewCustom("Title", "", c, w)
d.Resize(fyne.NewSize(200, d.MinSize().Height))
d.Show()
```
But NewCustom always render a dismiss button, I think there should be a way to tell dialog.NewCustom to not render that button if user does not want it.

fpabl0:
It seems that there is a flaky test related to entry widget:
https://github.com/fyne-io/fyne/pull/1931/checks?check_run_id=1930590506#step:5:101

dialog/base.go
toaster: Couldn’t we change the background image colour instead?

toaster (approved):

fpabl0:
I have added a little layout fix for popup that would benefit `dialog` too, please kindly check.

dialog/base.go
fpabl0: Yes, that is a possibility, however it would mean to have new functions `xyzThemedResource` to deal with this new color 🤔

widget/popup.go
fpabl0: done

dialog/base.go
andydotxyz: Unfortunately we can’t - the renderer does not currently support a faded image with an alpha channel. This BG was a workaround for that, I remember now

dialog/base.go
andydotxyz: I think you are describing `widget.NewModalPopUp`

fpabl0: Sorry, I didn't understand, what do you mean?

dialog/base.go
fpabl0: I see, then I will resolve this conversation.

toaster (approved):
LGTM

dialog/base.go
andydotxyz: Sorry, what I was meaning is: isn't a dialog with no buttons a modal popup?
I.e. dialogs are dismissable using buttons, to do something that isn't you can show a modal popup.

andydotxyz: Either way, for now perhaps the right thing to do is place a comment next to that line saying "// TODO remove once progress dialogs are gone" - if there is no sensible way to fix the issue with this PR.

andydotxyz: Maybe the comment in the deprecation should be fixed.
Anyhow, as I am sure you agree, we can't break the feature even if it's deprecated.

fpabl0: > dialogs are dismissable using buttons, to do something that isn't you can show a modal popup.

Oh I see, yes you are right.

> for now perhaps the right thing to do is place a comment next to that line saying "// TODO remove once progress dialogs are gone"

By doing it, a wrong test image needs to be committed, is that ok?

> if there is no sensible way to fix the issue with this PR.

We could update dialog.NewProgress with the above code, then we don't need to commit wrong test images, while ensuring the same behavior.

> Maybe the comment in the deprecation should be fixed.

Yes, I think.

> Anyhow, as I am sure you agree, we can't break the feature even if it's deprecated.

Completely, agreed.

andydotxyz:
Can we just clarify please, is the size / layout of the deprecated progress dialog still broken with this proposed change?

fpabl0:
Yes, it is broken, if we re-write the progress dialog to:
```go
c := container.NewVBox(widget.NewProgressBarInfinite())
d.setButtons(c)
d.Resize(fyne.NewSize(200, d.MinSize().Height))
```
then it works as before.

fpabl0:
Well, my assumption was wrong the above code does not fix broken progress dialogs, but this does:
```go
bar := widget.NewProgressBarInfinite()
rect := canvas.NewRectangle(color.Transparent)
rect.SetMinSize(fyne.NewSize(200, 0))

d.setButtons(container.NewMax(rect, bar))
```
Is that fine? It is just a workaround to a deprecated feature, so I think it would be ok, what do you think? @andydotxyz

andydotxyz:
> Is that fine? It is just a workaround to a deprecated feature, so I think it would be ok, what do you think? @andydotxyz

Yes, like you say it's a workaround but the code is on the way out.

andydotxyz (approved):

andydotxyz:
Looks great - are you still happy @toaster ?

toaster (approved):

toaster:
> Looks great - are you still happy @toaster ?

Yes, I am :).

Issue #1382, opened 2020-10-09
Dialog not returning to requested size


stuartmscott:
A dialog with size set by Resize, but shrunk due to window resize, does not return to set size when window resizes.

https://slack-files.com/T029RQSE6-F01CAMDFG3U-5b40a55576

Recommended resolution is to save the size given to dialog.Resize, and regrow up to that size if window permits

andydotxyz:
Fixed on `develop` and `release/v2.0.x`

PR #1932, opened 2021-02-09
Ensure Popups are resized in newly created window and when window is resized, fixes #1692

### LINKED ISSUE #1692: Dialog not resizing in newly created window
### Describe the bug:

I'm trying to create a new window and show a FolderOpen dialog that pretty much takes the full window.
However, it is not resizing as expected and instead stays at a small (default?) size.

### Example code
```golang
w := fyne.CurrentApp().NewWindow("Select destination...")
w.Resize(fyne.NewSize(1000, 600))
w.Show()
dialog := dialog.NewFolderOpen(func(list fyne.ListableURI, err error) {
	fmt.Println(list, err)
	w.Close()
}, w)
dialog.Show()
dialog.Resize(fyne.NewSize(975, 575))
```
### Screenshot of the too small dialog
screenshot_2020-12-28_at_18 55 20

### Workaround
@andydotxyz pointed out that a sleep might be needed before showing the dialog.
Modifying the code snippet like this, results in the expected full-window dialog:
```golang
...
dialog.Show()
time.Sleep(100 * time.Millisecond) // YMMV, 100ms worked for me
dialog.Resize(fyne.NewSize(975, 575))
```

 - OS: [e.g. Linux, MacOS or iOS] macOS
 - Version: [e.g. 22] 10.15.7
 - Go version: [e.g. 1.12.3] 1.15
 - Fyne version: [e.g. 1.2.0 or git SHA] 1.4.2

fpabl0:
### Description:

Fixes #1692 and #1382 when #1931 is landed.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.
- [x] #1931 landed
- [x] Add these changes to mobile driver.
- [x] Copy fixes in FileDialog Resize method to other dialogs that use widget.PopUp too.
- [x] Decision about #707 TODO code comments. (#707 is a design proposal, so it is out of scope of this PR)


andydotxyz:
This seems to break resizing with visible dialogs:

Screenshot 2021-02-10 at 11 42 01

dialog/file.go
```diff
@@ -476,21 +476,23 @@ func (f *FileDialog) Resize(size fyne.Size) {
...
-	}
-	newHeight := size.Height
-	if size.Height > maxSize.Height {
-		newHeight = maxSize.Height
-	} else if size.Height < minSize.Height {
-		newHeight = minSize.Height
-	}
-	f.dialog.win.Resize(fyne.NewSize(newWidth, newHeight))
+	f.dialog.win.Resize(minSize.Max(size))
+	// maxSize := f.dialog.win.Size()
```
andydotxyz, line 481: Please don't commit commented out code.

Additionally I don't think that removing the checks that the dialog fits within the window bounds is the right fix here.
You'll see that with this the file dialog no longer shrinks down to minsize if the window is smaller than requested size.

andydotxyz (changes requested):
The changes here seem to break sizing in other areas.
Also it only has changed file dialog, not all dialogs.

fpabl0:
> This seems to break resizing with visible dialogs:

How do you test it? This solution can only be seen after #1931 is landed, as I said in the first comment.

dialog/file.go
fpabl0: > Please don't commit commented out code.

Sorry about it.

> Additionally I don't think that removing the checks that the dialog fits within the window bounds is the right fix here.
You'll see that with this the file dialog no longer shrinks down to minsize if the window is smaller than requested size.

They are fixed with #1931.

fpabl0:
> Also it only has changed file dialog, not all dialogs.

Yes, other dialogs need the same fix. That's why I put it in the Check List in the first comment.
I will mark this as a draft to avoid confusing, sorry.

andydotxyz:
> Yes, other dialogs need the same fix. That's why I put it in the Check List in the first comment.

Ah of course, I missed that sorry. Draft seems like a good idea.

andydotxyz:
> How do you test it? This solution can only be seen after #1931 is landed, as I said in the first comment.

Open fyne_demo go to dialogs and open a file dialog.
Then resized the window.

andydotxyz:
If you can merge in latest `develop` this should work now...

fpabl0:
@andydotxyz  I have updated, check if now you can see the fix :), I still need to do the same change to other dialogs, but want feedback first :)

andydotxyz:
Thanks @fpabl0 now that it is overlayed on the other fix I don't see the issue I raised before, sorry for the confusion.
After quick testing it looks like a great fix. I'll do a proper review once you're happy with the code :)

fpabl0:
> sorry for the confusion

No worries, it was my fault, because I opened this as PR instead of a draft 😅

> I'll do a proper review once you're happy with the code :)

Great! 👍 I would like to know what is exactly the issue #707, how do I test it? There are comments in the code referencing that issue that this PR will overwrite so want to know if this overwrite still satisfies that workaround.

By the way, do you know what happened with the tests? It says that it was cancelled, but no more information 🧐

fpabl0:
Ok this is almost finished. I am not sure if I can add tests to this fix, because the real fix is actually on canvas implementation, that is overwritten in test environment 🤔

I would appreciate if @toaster can review this too, as he opened #707 and this PR is going to overwrite some things around that issue.

andydotxyz:
> Ok this is almost finished. I am not sure if I can add tests to this fix, because the real fix is actually on canvas implementation, that is overwritten in test environment 🤔

For these things the tests go in `internal/driver/glfw` and `internal/driver/gomobile` - those use driver-specific canvases instead of the test objects.

andydotxyz:
> It says that it was cancelled, but no more information 🧐

Sorry I don't know what happened - tests normally cancel if a newer commit was pushed or one of the other tests failed.
Looks like it was not a permanent failure.

test/testcanvas.go
```diff
@@ -160,8 +160,19 @@ func (c *testCanvas) Resize(size fyne.Size) {
...
 
+	// Ensure testcanvas mimics real canvas.Resize behavior
 	for _, overlay := range overlays.List() {
-		overlay.Resize(size)
+		type popupWidget interface {
+			fyne.CanvasObject
+			ShowAtPosition(fyne.Position)
+		}
+		if p, ok := overlay.(popupWidget); ok {
+			// TODO: remove this when #707 is being addressed.
```
andydotxyz, line 170: I know this is copied from GLFW but I wonder, is this impacted/fixed by the PR?

fpabl0: The latest tests I have added to `popup_test.go` give different results (wrong ones) compared with real implementation.

andydotxyz: Yes I see why the code is added - I was remarking about the bugfix mentioned in the code you coped.
In other areas of the PR you suspected it might be resolved and have removed some of the mentions.

fpabl0: Oh I see.  The code I removed isn't needed for the test as it already guarantee that behavior (it maybe should have been removed before ??).  I found # 707 a bit confusing, it seems to be a design proposal, so don't know what decision was taken around it

fpabl0: @andydotxyz  Based on the Slack discussion, is it ok to keep this addition to `testcanvas.go`?

andydotxyz: Yes I think so

andydotxyz (approved):

PR #1933, opened 2021-02-10
Fix widget.List layout when the items length is lower than the current visible ones, fixes #1864

### LINKED ISSUE #1864: List widget panic when refreshing after changing content through combobox changed.
### Describe the bug:
If I display the list with 20 items but did not scroll it at all, changed the value of list and refresh. the list will display 1 item successfully. Whereas if I display 20 items and scrolled down, change the value of list and refresh will panic immediately with that error stack.

### To Reproduce:
1. Run the example code.
2. Change combo box value to 20.
3. Scroll down for a bit.
4. Change combo box value to 1.
5. panic. 

*If step 3 is ignored, it will run successfully.*

### Example code:
```go
package main

import (
	"strconv"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/layout"
	"fyne.io/fyne/v2/widget"
)

func main() {
	myApp := app.New()
	myWindow := myApp.NewWindow("List example")

	userTab, _ := renderUsersTab(myWindow)

	tabs := container.NewAppTabs(
		container.NewTabItem("Users", userTab),
	)
	tabs.SetTabLocation(container.TabLocationTop)

	myWindow.SetContent(tabs)
	myWindow.CenterOnScreen()
	myWindow.Resize(fyne.Size{Width: 700, Height: 500})
	myWindow.ShowAndRun()
}

func renderUsersTab(w fyne.Window) (fyne.CanvasObject, error) {
	groups := []string{"1", "10", "20"}
	members := generateEmails(1)

	l := widget.NewList(func() int {
		return len(members)
	}, func() fyne.CanvasObject {
		return widget.NewLabel("PLACEHOLDERS@DOMAIN.COM")
	}, func(lii widget.ListItemID, co fyne.CanvasObject) {
		if lii >= len(members) {
			return
		}
		val := members[lii]
		label := co.(*widget.Label)
		label.SetText(val)
	})

	groupLabel := widget.NewLabel("Groups")
	groupCombo := widget.NewSelect(groups, func(value string) {
		num, _ := strconv.Atoi(value)
		members = generateEmails(num)
		l.Refresh()
	})

	grid := container.New(layout.NewFormLayout(), groupLabel, groupCombo)
	tab := container.New(layout.NewBorderLayout(grid, nil, nil, nil), grid, l)
	return tab, nil
}

func generateEmails(num int) []string {
	emails := []string{}

	for i := 0; i < num; i++ {
		emails = append(emails, "testuser"+strconv.Itoa(i)+"@gmail.com")
	}

	return emails
}
```

### Device:
 - **OS:** Windows
 - **Version:** 10 Pro
 - **Go version:** go version go1.15.7 windows/amd64
 - **Fyne version:** 2.0.0

fpabl0:
### Description:

Fixes #1864 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/list.go
```diff
@@ -207,6 +207,13 @@ func (l *listRenderer) Layout(size fyne.Size) {
 		return
 	}
 	min := int(fyne.Min(float32(length), float32(l.visibleItemCount)))
+	if length < len(l.children) {
+		l.firstItemIndex = 0
```
andydotxyz, line 211: I think this code runs when the dataset has shrunk right?
I don't think that this means we should scroll to top.
The only scrolling that should be required is if our visible viewport is beyond the space now required for the list.

andydotxyz (changes requested):
I think you've found why this is happening, but I don't think that scrolling to the top will always be desirable in this case.

widget/list.go
fpabl0: Fair enough. I will remove the scroll to top.

fpabl0: done

andydotxyz:
Thanks.
I have tested this more and it is a good fix to the crash - but there are still related issues, i.e. it seems to not correctly calculate how many items are visible:

Screenshot 2021-02-12 at 10 09 42

andydotxyz:
I think what I illustrated above is a separate issue, happy with this crash fix.

andydotxyz (approved):

fpabl0:
> I have tested this more and it is a good fix to the crash - but there are still related issues, i.e. it seems to not correctly calculate how many items are visible:

Oh I see, is there a way to reproduce it? I have just remembered that issue but after this fix, I couldn't be able to reproduce again, so I though it was fixed too 🤔

fpabl0:
@andydotxyz I have just realized you opened an issue about it.

> Code to replicate is at #1864 - run it, choose 20, scroll to bottom then choose 10.

Exactly that was how I reproduced that issue before this fix. But after that, I am not able to reproduce 🤔

fpabl0:
Update: I have been able to reproduce it again, I will comment on issue instead of here, to keep this organized.

PR #1935, opened 2021-02-10
Fix entry widget mouse drag selection when scrolled, fixes #1916

### LINKED ISSUE #1916: Entry widget mouse drag selection offset when scrolled
### Describe the bug:

Text selection not working properly if entry widget scroll(horizontal or vertical). Selection not appeared under cursor.

### To Reproduce:
1. Input long text in entry widget
2. Scroll and try to select last parts of text

### Screenshots:

https://user-images.githubusercontent.com/56516965/107147337-8b4f3b80-695e-11eb-8a76-7ef7e569e7b5.mp4

### Example code:
```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Hello")

		widget.NewEntry(),
	))

	w.ShowAndRun()
}

```

 - **OS:** Windows
 - **Version:** 10 Pro
 - **Go version:** 1.15.2 windows/amd64
 - **Fyne version:** 2.0.0

fpabl0:
### Description:

Fixes #1916

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
This looks like a good fix, thanks - but can you please add a test?
There are related tests in `widget/entry_internal_test.go`

widget/entry.go
```diff
@@ -237,12 +237,15 @@ func (e *Entry) DragEnd() {
 //
 // Implements: fyne.Draggable
 func (e *Entry) Dragged(d *fyne.DragEvent) {
+	pevt := d.PointEvent
+	// substract scroll offset because we need visible column and row numbers to render
```
andydotxyz, line 241: I'm not sure that a comment is needed as the code is pretty clear.
If you do think it should remain perhaps it should be about correcting coordinates rather than rendering selection, as that is not what this code is doing.

andydotxyz (changes requested):
Nice thanks. Question inline about code comments.
But mostly this just needs a test.

widget/entry.go
fpabl0: At least for me, looking at the code it is not clear what is happening. A better comment should be this?
"subtract scroll offset because we need visible column and row numbers to correct coordinates, not the total ones."

fpabl0:
> but can you please add a test?

Yes, I will try to add it later today.

widget/entry.go
andydotxyz: Isn't that the opposite of what is happening?
The event is relative to the top-left of the widget (Entry) so to know how far into the scroll content we should go it is required to adjust by `Offset` amount (offset will be negative, as the content is left of the widget it is in).

fpabl0: > The event is relative to the top-left of the widget (Entry) so to know how far into the scroll content we should go it is required to adjust by Offset amount

Yes, but in the case of selection you need column number relative to what is visible.

> (offset will be negative, as the content is left of the widget it is in)

Hmm I will do some tests. Although I thought at the most-left part of the widget, the offset is 0, but I will investigate.

fpabl0: > The event is relative to the top-left of the widget (Entry) so to know how far into the scroll content we should go it is required to adjust by Offset amount (offset will be negative, as the content is left of the widget it is in).

You are right, the confusing part is maybe that `getRowCol` function (called in `updateMousePointer`) makes the rows and columns relative to the whole widget instead of the top-left visible, to mitigate that effect I subtract the scroll offset that is added inside `getRowCol`.

fpabl0:
I have added the test, but I am not sure if it is right. @andydotxyz could you check it, please?

widget/entry.go
```diff
@@ -238,8 +238,12 @@ func (e *Entry) DragEnd() {
 // Implements: fyne.Draggable
 func (e *Entry) Dragged(d *fyne.DragEvent) {
 	pevt := d.PointEvent
-	// substract scroll offset because we need visible column and row numbers to render
-	// the selection rectangles, not the total ones.
+	// For selection text, rows and columns numbers relative to the most-left
```
andydotxyz, line 241: So much text is very distracting and not really our style. Perhaps the following explains it?:

```
Convert the relative drag position from our Entry coordinates to be relative for Scroll.Content.
```

widget/entry_internal_test.go
```diff
@@ -93,6 +93,29 @@ func TestEntry_DragSelect(t *testing.T) {
 	assert.Equal(t, "r the laz", entry.SelectedText())
 }
 
+func TestEntry_DragSelectWithScroll(t *testing.T) {
+	entry := NewEntry()
+	entry.SetText("The quick brown fox jumped over and over the lazy dog.")
+	entry.Resize(entry.MinSize())
```
andydotxyz, line 99: You shouldn't need to resize to minsize

widget/entry_internal_test.go
```diff
@@ -93,6 +93,29 @@ func TestEntry_DragSelect(t *testing.T) {
...
+
+	// get position after the letter 'a'.
+	ev1 := getClickPosition("The quick brown fox jumped over and over the la", 0)
+	// get position after the letter 'u'
+	ev2 := getClickPosition("The qu", 0)
+
+	// mouse down and drag from 'a' to 'i'
+	me := &desktop.MouseEvent{PointEvent: *ev1, Button: desktop.MouseButtonPrimary}
+	entry.MouseDown(me)
+	for ; ev1.Position.X > ev2.Position.X; ev1.Position.X-- {
```
andydotxyz, line 109: You don't have to drag all increments - just mouse down at `ev1`, drag to `ev2` and mouse up at `ev2`.

andydotxyz (changes requested):
The test looks good, though it could be simplified.
A couple more notes inline.

widget/entry_internal_test.go
fpabl0: That's what I think, but I was unsure because previous test does it. I will remove it :)

widget/entry.go
fpabl0: > So much text is very distracting and not really our style.

Yes, I think it too, but not found less words to describe the situation.

> Convert the relative drag position from our Entry coordinates to be relative for Scroll.Content.

That looks great! Thanks :)

fpabl0: done

widget/entry_internal_test.go
fpabl0: done

widget/entry_internal_test.go
fpabl0: Got it. Done.

widget/entry.go
fpabl0: Comment updated. Done.

andydotxyz (approved):

andydotxyz:
Great thanks!

PR #1938, opened 2021-02-10
Fix/form tests

andydotxyz:
Fixing initial state of entry should setting the somewhat unstable form test on Windows.

(Tested on Windows and macOS)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (approved):
Looks good to me. Thanks :+1:

PR #1943, opened 2021-02-11
Add a MinSize() to dialog

andydotxyz:
### Checklist:

- [x] Tests included. <- see how an existing test now confirms minsize through the output size
- [x] Tests all pass.


stuartmscott (commented):
Overall this change looks good. I like that the layout/minsize methods are split out into a layout object, and that any changes in minsize can still be detected by the tests. My only concern is that the dialog shadow and semi-transparent overlay are no longer in the golden images so any changes to them could slip past the tests - could the window be resized to the dialog minsize + some extra?

andydotxyz:
> could the window be resized to the dialog minsize + some extra?

Good plan. It also uncovered an inconsistency in the tests. I think this should fix it, though as you can see it is to be improved - I think one of the other open PRs addresses this, but let's keep an eye on the overall consistency as it develops

stuartmscott (approved):

PR #1944, opened 2021-02-11
Keep track of windows closing without removing context

### LINKED ISSUE #1604: SIGSEGV Sometimes When Closing a Program by Clicking a Button
### Describe the bug:
I've noticed that sometimes my program faults with "[signal SIGSEGV: segmentation violation code=0x1 addr=0xf8 pc=0x7fdae2aa6fc4]". It doesn't happen every time, though, so I'm not sure exactly where it is going wrong, full error is here: https://hastebin.com/efiyuxofag.go

### To Reproduce:
* Run any Fyne program on my system which closes the window by clicking a button widget, it will sometimes need to be ran lots of times until I can reproduce the exact error message.

### Example Code

```
package main

import (
	"fyne.io/fyne/app"
	"fyne.io/fyne/container"
	"fyne.io/fyne/layout"
	"fyne.io/fyne/widget"
	"fmt"
)


func showConfirm(title, question string, callback func(bool)) {
	a := app.New()
	w := a.NewWindow(title)
	yes := widget.NewButton("Yes", func() {
		callback(true)
		w.Close()
	})
	yes.Importance = widget.HighImportance
		widget.NewLabel(question),
		container.NewHBox(layout.NewSpacer(),
			widget.NewButton("No", func() {
				callback(false)
				w.Close()
			}), yes,
			layout.NewSpacer())))
	w.ShowAndRun()
}

func main() {
	showConfirm("MyTitle", "Are you sure?", func(b bool) {
		if b {
			fmt.Println("Confirmed")
		}
	})
}
```

 - OS: Ubuntu Linux
 - Version: 20.04
 - Go version: 1.12.1
 - Fyne version: 0ca05dc12c03b8873d2f22caa6ce4c28988b3886

andydotxyz:
Fixes #1604

### Checklist:

- [x] Tests included. <- existing tests OK, could not find a way to replicate in tests as it was random / racey
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -274,7 +274,7 @@ func (w *window) fitContent() {
 		return
 	}
 
-	if w.viewport == nil || w.closing {
+	if w.isClosing() {
 		return
 	}
 
@@ -406,7 +406,7 @@ func (w *window) doShow() {
 }
 
 func (w *window) Hide() {
-	if w.closing {
+	if w.isClosing() {
 		return
 	}
 
@@ -424,7 +424,7 @@ func (w *window) Hide() {
 }
 
 func (w *window) Close() {
-	if w.closing {
+	if w.isClosing() {
 		return
 	}
 
@@ -1178,7 +1178,7 @@ func (w *window) focused(_ *glfw.Window, isFocused bool) {
 }
 
 func (w *window) RunWithContext(f func()) {
-	if w.closing {
+	if w.isClosing() {
 		return
 	}
 	w.viewport.MakeContextCurrent()
@@ -1195,7 +1195,7 @@ func (w *window) RescaleContext() {
 }
 
 func (w *window) rescaleOnMain() {
-	if w.viewport == nil || w.closing {
+	if w.isClosing() {
 		return
 	}
 	w.fitContent()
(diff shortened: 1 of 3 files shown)


internal/driver/glfw/window.go
```diff
@@ -423,10 +424,11 @@ func (w *window) Hide() {
 }
 
 func (w *window) Close() {
-	if w.viewport == nil {
+	if w.closing {
 		return
 	}
 
+	w.closing = true
 	w.viewport.SetShouldClose(true)
```
stuartmscott, line 432: Should this still be protected by a nil check?
```suggestion
	if v := w.viewport; v != nil {
		v.SetShouldClose(true)
	}
```

internal/driver/glfw/window.go
```diff
@@ -273,7 +274,7 @@ func (w *window) fitContent() {
 		return
 	}
 
-	if w.viewport == nil {
+	if w.viewport == nil || w.closing {
```
stuartmscott, line 277: consider adding `w.isClosing()` which could encapsulate these two tests and be used in several call sites in this file

internal/driver/glfw/window.go
andydotxyz: I don't think so - previously viewport was set to nil to show it was closed, now it is only nil before the window appears.
I guess it couldn't hurt to use the suggested helper at the top instead though.

stuartmscott (approved):

PR #1954, opened 2021-02-13
TabsContainer: do not restart animation if there is already one executing with the same end value, fixes #1952

fpabl0:
### Description:

Fixes #1952 

### Checklist:

- [ ] Tests included.
- [x] Lint and formatter run with no errors.
- [x] Tests all pass.

Issue #1952, opened 2021-02-12
AppTab crashes on Android


stuartmscott:
### Describe the bug:
Nil animation pointer

### To Reproduce:
Steps to reproduce the behaviour:
1. Open fyne_demo
2. Go to Containers -> AppTabs
3. Click various tabs

### Stacktrace
```
02-12 15:15:59.871 25633     0 E Go      : panic: runtime error: invalid memory address or nil pointer dereference
02-12 15:15:59.871 25633     0 E Go      : [signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x6e4c363178]
02-12 15:15:59.871 25633     0 E Go      : 
02-12 15:15:59.871 25633     0 E Go      : goroutine 25 [running]:
02-12 15:15:59.871 25633     0 E Go      : fyne.io/fyne/v2/internal/animation.newAnim(0x0, 0x6e40800000)
02-12 15:15:59.871 25633     0 E Go      : 	/Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/internal/animation/animation.go:19 +0x38
02-12 15:15:59.871 25633     0 E Go      : fyne.io/fyne/v2/internal/animation.(*Runner).Start(0x4000308000, 0x0)
02-12 15:15:59.871 25633     0 E Go      : 	/Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/internal/animation/runner.go:22 +0x60
02-12 15:15:59.871 25633     0 E Go      : fyne.io/fyne/v2/internal/driver/gomobile.(*mobileDriver).StartAnimation(0x4000300050, 0x0)
02-12 15:15:59.871 25633     0 E Go      : 	/Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/internal/driver/gomobile/animation.go:6 +0x2c
02-12 15:15:59.871 25633     0 E Go      : fyne.io/fyne/v2.(*Animation).Start(0x0)
02-12 15:15:59.871 25633     0 E Go      : 	/Users/stuartscott/Documents/Projects/Go/src/fyne.io/fyne/animation.go:58 +0x48
02-12 15:15:59.871 25633     0 E Go      : fyne.io/fyne/v2/container.(*baseTabsRenderer).moveIndicator(0x40000ea6c0, 0x41e8000000000000, 0x4080000042d00000, 0x4042040001)
[...]

PR #1956, opened 2021-02-14
Fixes animation runner memory leak

fpabl0:
### Description:

This PR will allow us to do the following actions without memory leak or crash:
1. Stop animation immediately after Start.
2. Stop animation inside animation tick function.

This PR also fixes the crash in AppTabs (#1952), however the performance (slow animation) can only be fixed by #1954.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/animation/animation.go
+++ b/internal/animation/animation.go
@@ -20,14 +20,13 @@ type anim struct {
 }
 
 func newAnim(a *fyne.Animation) *anim {
-	// TODO should we add a nil check here? (to avoid panic)
 	animate := &anim{a: a, start: time.Now(), end: time.Now().Add(a.Duration)}
 	animate.total = animate.end.Sub(animate.start).Nanoseconds() / 1000000 // TODO change this to Milliseconds() when we drop Go 1.12
 	animate.repeatsLeft = a.RepeatCount
 	return animate
 }
 
-func (a *anim) setStopFlag() {
+func (a *anim) setStopped() {
 	a.mu.Lock()
 	a.stopped = true
 	a.mu.Unlock()
--- a/internal/animation/animation.go
+++ b/internal/animation/animation.go
@@ -1,6 +1,7 @@
 package animation
 
 import (
+	"sync"
 	"time"
 
 	"fyne.io/fyne/v2"
@@ -13,11 +14,28 @@ type anim struct {
 	reverse     bool
 	start       time.Time
 	total       int64
+
+	mu      sync.RWMutex
+	stopped bool
 }
 
 func newAnim(a *fyne.Animation) *anim {
+	// TODO should we add a nil check here? (to avoid panic)
 	animate := &anim{a: a, start: time.Now(), end: time.Now().Add(a.Duration)}
 	animate.total = animate.end.Sub(animate.start).Nanoseconds() / 1000000 // TODO change this to Milliseconds() when we drop Go 1.12
 	animate.repeatsLeft = a.RepeatCount
 	return animate
+}
+
+func (a *anim) setStopFlag() {
+	a.mu.Lock()
+	a.stopped = true
+	a.mu.Unlock()
+}
+
+func (a *anim) isStopped() bool {
+	a.mu.RLock()
+	ret := a.stopped
+	a.mu.RUnlock()
+	return ret
 }
--- a/internal/animation/runner.go
+++ b/internal/animation/runner.go
@@ -9,19 +9,24 @@ import (
 
 // Runner is the main driver for animations package
 type Runner struct {
-	animationMutex sync.RWMutex
-	animations     []*anim
+	animationMutex    sync.RWMutex
+	animations        []*anim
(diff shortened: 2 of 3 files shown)


internal/animation/animation.go
```diff
@@ -13,11 +14,28 @@ type anim struct {
 	reverse     bool
 	start       time.Time
 	total       int64
+
+	mu      sync.RWMutex
+	stopped bool
 }
 
 func newAnim(a *fyne.Animation) *anim {
+	// TODO should we add a nil check here? (to avoid panic)
```
andydotxyz, line 23: Always good to be safe. Normally we would only do nil checks for public APIs, but maybe this one is called blindly?

internal/animation/animation.go
```diff
@@ -13,11 +14,28 @@ type anim struct {
...
 	animate := &anim{a: a, start: time.Now(), end: time.Now().Add(a.Duration)}
 	animate.total = animate.end.Sub(animate.start).Nanoseconds() / 1000000 // TODO change this to Milliseconds() when we drop Go 1.12
 	animate.repeatsLeft = a.RepeatCount
 	return animate
 }
+
+func (a *anim) setStopFlag() {
```
andydotxyz, line 30: It would be more symmetrical as `setStopped`

internal/animation/runner.go
```diff
@@ -47,19 +69,22 @@ func (r *Runner) runAnimations() {
 			<-draw.C
 			r.animationMutex.Lock()
 			oldList := r.animations
-			r.animations = nil // clear the list so we can append any new ones after processing
 			r.animationMutex.Unlock()
-			var newList []*anim
+			newList := make([]*anim, 0, len(oldList))
 			for _, a := range oldList {
-				if r.tickAnimation(a) {
+				if r.tickAnimation(a) && !a.isStopped() {
```
andydotxyz, line 75: Shouldn't we check if it is stopped before ticking it? (i.e. these arguments the other way round).

andydotxyz (changes requested):

internal/animation/animation.go
fpabl0: Yes, you are right, good catch! That situation is impossible to happen. I will remove that TODO.

internal/animation/animation.go
fpabl0: Agreed. I will update the name.

internal/animation/runner.go
fpabl0: Hmm I don't think so, the idea of writing it after the tick is that if inside the tick, the animation is stopped, we don't append it to the new list.

andydotxyz: But as animations will more commonly be stopped from outside is it not sensible to avoid the tick? If stop is set during tick it will still be cancelled, just before the next tick rather than after the current?

fpabl0: Yes, you are right, we can move it before or added before. I mean:
```go
if !a.isStopped() && r.tickAnimation(a) && !a.isStopped() {
```
Then animations are always stopped immediately and `isStopped` is not an expensive function, is it?

fpabl0:
@andydotxyz in the function `runner.tickAnimation` couldn't it possible to have an early return below this line?

https://github.com/fyne-io/fyne/blob/c29a0624ed96ba1b8f45d903b6941824d50e0502/internal/animation/runner.go#L92

It seems that the below code will not be very important (when the animation expired) because `delta` in theory will be always 0 and in consequence `val` too, what do you think?

andydotxyz:
> @andydotxyz in the function `runner.tickAnimation` couldn't it possible to have an early return below this line?

Yes I think you are right. I wondered if we were missing some `return true` and I think you found the right place to add it.

internal/animation/runner.go
andydotxyz: I guess it's not expensive, but checking in both places seems like it could be overkill as we will find out in the next iteration anyway.
I'd be wary of calling mutexes in tight loops if it can be avoided - but maybe you know more about the specifics herer and how it performs.

fpabl0: Yeah, you are right, I am just worried about appending stopped animations in a slice will increase its capacity (that it is not needed). But, maybe that's something we can leave as it and check if we have problems later. I will update :)

fpabl0:
> Yes I think you are right. I wondered if we were missing some return true and I think you found the right place to add it.

Great, will update! :)

internal/animation/animation.go
fpabl0: done

internal/animation/animation.go
fpabl0: done

internal/animation/runner.go
fpabl0: done

andydotxyz (approved):

andydotxyz:
Nice thanks, sorry for the delay

PR #1958, opened 2021-02-15
Fix the position of elements after SetColumnWidth called

### LINKED ISSUE #1951: Cells are not getting draw in correct location after column resize.
### Describe the bug:

After calling SetColumnWidth the cells are not getting draw in correct location.
The table.Refresh() doesn't help.
The column's border is properly displayed and selection highlight correct.
This is tested with development branch due to Apple M1 (github.com/sirkon/fyne/v2 v2.0.1-0.20210209074931-f93d4e04b627).

### To Reproduce:
Run provided test program.
After 5 second delay the columns border is changed.
The Cells widgets are not moved to correct location.
Try select individual cells, the selection marks are displayed properly.

### Screenshots:

Screen Shot 2021-02-12 at 1 14 59 PM

Screen Shot 2021-02-12 at 1 15 05 PM

### Example code:

```go
package main

import (
	"fmt"
	"time"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Examples")
	w.SetContent(newTable())
	w.Resize(fyne.Size{Width: 800, Height: 300})
	w.ShowAndRun()
}

func newTable() fyne.CanvasObject {
	var data map[int][]string = make(map[int][]string, 1)
	var colWidth []float32 = make([]float32, 3)
	data[0] = []string{"Header 1", "Header 2", "Header 3"}
	var table *widget.Table
	table = widget.NewTable(func() (int, int) {
		return len(data), 3
	}, func() fyne.CanvasObject {
		l := widget.NewLabel("")
		l.Resize(fyne.Size{Width: 200, Height: 20})
		return l
	}, func(tci widget.TableCellID, co fyne.CanvasObject) {
		if label, ok := co.(*widget.Label); ok {
			col := tci.Col
			label.Text = data[tci.Row][col]
			label.Refresh()
			width := label.MinSize().Width
			if colWidth[col] < width {
				colWidth[col] = width
				// Options A
				// go func() {
				// 	table.SetColumnWidth(col, colWidth[col])
				// 	table.Refresh()
				// }()
			}
		}
	})

	table.SetColumnWidth(0, 200)
	table.SetColumnWidth(1, 200)
	table.SetColumnWidth(2, 200)

	table.OnSelected = func(id widget.TableCellID) {
		fmt.Printf("Select col %d row %d\n", id.Col, id.Row)
	}

	// Options B
	go func() {
		time.Sleep(time.Second * 5)
		table.SetColumnWidth(0, 100)
		table.SetColumnWidth(1, 100)
		table.SetColumnWidth(2, 100)
		table.Refresh()
	}()

	return table
}
```

go.mod for Apple M1:
```
module table-test

go 1.16

replace fyne.io/fyne/v2 => github.com/sirkon/fyne/v2 v2.0.1-0.20210209074931-f93d4e04b627

require fyne.io/fyne/v2 v2.0.0-00010101000000-000000000000
```

 - **MacOS:** 
 - **11.2:** 
 - **go1.16rc1:** 
 - **2.0 (dev):**

andydotxyz:
Also fix the minsize for the scroller which was ignoring the value
Fixes #1951

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/table.go
```diff
@@ -508,7 +508,20 @@ func (r *tableCellsRenderer) MinSize() fyne.Size {
 	} else {
 		fyne.LogError("Missing Length callback required for Table", nil)
 	}
-	return fyne.NewSize(r.cells.cellSize.Width*float32(cols)+float32(cols-1), r.cells.cellSize.Height*float32(rows)+float32(rows-1))
+
+	width := float32(0)
```
stuartmscott, line 512: nit: not sure if this makes a difference to the compiler, but declaring a var seems nicer than casting 0
```suggestion
	var width float32
```

widget/table.go
```diff
@@ -556,12 +569,12 @@ func (r *tableCellsRenderer) Refresh() {
...
 				}
-
-				c.Move(fyne.NewPos(theme.Padding()+cellOffset,
-					theme.Padding()+float32(row)*(r.cells.cellSize.Height+separatorThickness)))
-				c.Resize(fyne.NewSize(colWidth-theme.Padding()*2, r.cells.cellSize.Height-theme.Padding()*2))
 			}
 
+			c.Move(fyne.NewPos(theme.Padding()+cellOffset,
+				theme.Padding()+float32(row)*(r.cells.cellSize.Height+separatorThickness)))
+			c.Resize(fyne.NewSize(colWidth-theme.Padding()*2, r.cells.cellSize.Height-theme.Padding()*2))
```
stuartmscott, line 576: `Resize` can call `Layout`, so I wonder if it would be better to have this after the `updateCell` call, WDYT?

andydotxyz: Is there a difference to moving around widgets that contain template data compared to real data?
In a way it felt better to call update with the dimensions set, in case they matter

widget/table.go
andydotxyz: If we think there should be a rule for doing it one way or another then probably should also be in the style guide.

widget/table.go
stuartmscott: The prior art here is tree.go#465, so whichever approach is picked should be consistent across the collections

widget/table.go
stuartmscott: In the absence of a style guide, reviewers' style is adopted, no?

andydotxyz: That doesn't really make sense as we can have many reviewers - so there then would be conflicts.
If a PR does not go against our style or standards it should be approved surely?

widget/table.go
andydotxyz: That doesn't answer my questions, sorry. I would rather do the right thing rather than copy what is elsewhere.
If they both need to update I could do that.

Jacalz (approved):

widget/table.go
stuartmscott: The dev-defined-canvasobject should not need to be resized before being updated, but there certainly isn't any promise in the public API either way so it shouldn't hold up the review

widget/table.go
stuartmscott: The var declaration would be shorter, and doesn't have a type cast, but whatever

Issue #1960, opened 2021-02-15
Calling Select on List before draw can crash


andydotxyz:
```go
l := widget.NewList(...)
l.Select(0)
```

will crash

```
panic: runtime error: index out of range [-1] [recovered]
        panic: runtime error: index out of range [-1]

goroutine 9 [running]:
testing.tRunner.func1.1(0x43dcec0, 0xc000031160)
        /usr/local/Cellar/go/1.15.7/libexec/src/testing/testing.go:1072 +0x30d
testing.tRunner.func1(0xc0002c4780)
        /usr/local/Cellar/go/1.15.7/libexec/src/testing/testing.go:1075 +0x41a
panic(0x43dcec0, 0xc000031160)
        /usr/local/Cellar/go/1.15.7/libexec/src/runtime/panic.go:969 +0x1b9
fyne.io/fyne/v2/widget.(*listRenderer).scrollDown(0xc0001cc380, 0x42340000)
        /Users/andy/Go/src/fyne.io/fyne/v2/widget/list.go:291 +0x318
fyne.io/fyne/v2/widget.(*listRenderer).offsetChanged(0xc0001cc380)
        /Users/andy/Go/src/fyne.io/fyne/v2/widget/list.go:271 +0xbd
fyne.io/fyne/v2/widget.(*listRenderer).offsetUpdated(...)
        /Users/andy/Go/src/fyne.io/fyne/v2/widget/list.go:363
fyne.io/fyne/v2/widget.(*List).Select(0xc00031e140, 0x0)
        /Users/andy/Go/src/fyne.io/fyne/v2/widget/list.go:128 +0x1b8
...
```

andydotxyz:
On `develop` and prepping for 2.0.1 release as well

PR #1961, opened 2021-02-15
Refactoring List code to remove lots of internal state that caused some render issues

### LINKED ISSUE #1948: List widget does not render all visible content after content data gets shorter
In addition to #1933 there are issues when the data size shrinks and the List is scrolled down from the top.

107755218-6ec65100-6d1a-11eb-9cca-ce97f92a5806

Resizing the widget fixes the rendering, though not immediately - first the content fixes, then later the dividers reappear.

Code to replicate is at #1864 - run it, choose 20, scroll to bottom then choose 10.

andydotxyz:
I could not track down the issues directly, so imported Tree/Table style code to render with simpler code.

TODO: 
- [x] Solve why some test are unstable...?

Fixes #1948, #1960

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/list.go
```diff
@@ -512,24 +316,96 @@ func (l *listLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
...
-		objects[len(objects)-1].Move(fyne.NewPos(0, objects[len(objects)-2].Position().Y+l.list.itemMin.Height+theme.SeparatorThicknessSize()))
-	} else {
-		objects[len(objects)-1].Move(fyne.NewPos(0, 0))
+func (l *listLayout) setupListItem(item fyne.CanvasObject, id ListItemID) {
+	li := item.(*listItem)
+	previousIndicator := li.selected
+	li.selected = false
+	for _, s := range l.list.selected {
+		if id == s {
+			li.selected = true
```
Jacalz, line 342: I think this could be simplified to `li.selected = id == s` while also removing the if statement.

widget/list.go
```diff
@@ -512,24 +316,96 @@ func (l *listLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
...
+	length := 0
+	if f := l.list.Length; f != nil {
+		length = f()
+	}
+	visibleItemCount := int(math.Ceil(float64(l.list.scroller.Size().Height)/float64(l.list.itemMin.Height+theme.SeparatorThicknessSize()))) + 1
+	offY := l.list.offsetY - float32(int(l.list.offsetY)%int(l.list.itemMin.Height+separatorThickness))
+	minRow := ListItemID(offY / (l.list.itemMin.Height + separatorThickness))
+	maxRow := ListItemID(fyne.Min(float32(minRow+visibleItemCount), float32(length)))
+
+	updateItem := l.list.UpdateItem
```
Jacalz, line 370: I think you could just use this in the if statement below without assigning a variable for it.

Jacalz (changes requested):
Left a couple comments for possible code improvements. Apart from that, it seems to be working fine.

widget/list.go
andydotxyz: If there are (in the future?) more than one elements in the selected list then that will lead to only the last item being marked as selected.

widget/list.go
andydotxyz: Good point, fixed

widget/list.go
Jacalz: Good point. I thought the variable was part of the loop range. Never mind.

Jacalz (dismissed):
Looks good to me but probably good if someone more has a look as well.

widget/list.go
```diff
@@ -512,24 +316,95 @@ func (l *listLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
...
+	l.updateList()
 }
 
-func (l *listLayout) appendedItem(objects []fyne.CanvasObject) {
-	if len(objects) > 1 {
-		objects[len(objects)-1].Move(fyne.NewPos(0, objects[len(objects)-2].Position().Y+l.list.itemMin.Height+theme.SeparatorThicknessSize()))
-	} else {
-		objects[len(objects)-1].Move(fyne.NewPos(0, 0))
+func (l *listLayout) setupListItem(item fyne.CanvasObject, id ListItemID) {
+	li := item.(*listItem)
```
fpabl0, line 337: Is there any particular reason to receive a fyne.CanvasObject instead of directly a `*listItem`? 🤔

widget/list.go
```diff
@@ -512,24 +316,95 @@ func (l *listLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
...
+	previousIndicator := li.selected
+	li.selected = false
+	for _, s := range l.list.selected {
+		if id == s {
+			li.selected = true
```
fpabl0, line 342: If this is setting up a single list item, wouldn't it be more efficient to break the loop when the list item selected was found?

widget/list.go
andydotxyz: No, a hangover from old code - thanks

Jacalz (approved):

PR #1962, opened 2021-02-15
Fix tab button layout on mobile

stuartmscott:
### Description:
Tab buttons were always rendered in a grid of columns, this PR uses a grid of rows when the tab location is leading or trailing

### Checklist:
- [x] Tests included.
- [x] Tests all pass.


andydotxyz (changes requested):
Thanks, but not quite the complete solution - 2 cases seem wrong.

Mobile, Portrait, Leading - tabs appear at bottom (should be top)
Mobile, Landscape, Bottom - tabs appear at left (leading), but should be right

stuartmscott:
Where are you getting these expectations from?

From the ticket (#1310), AppTabs is a combination of two Material Design components;
- Bottom Navigation - which, as the name, suggests doesn't appear at the top.
- Navigation Rail - "The navigation rail is placed on the left side of the screen for left-to-right languages, and on the right side of the screen for right-to-left languages."

andydotxyz:
> Where are you getting these expectations from?

From how it worked before?

stuartmscott:
> From how it worked before?

I don't think that is true, Fyne v2.0.0 AppTab on Mobile always moved Leading & Trailing to Bottom (https://github.com/fyne-io/fyne/blob/master/container/apptabs.go#L323):
```go
func (r *appTabsRenderer) adaptedLocation() TabLocation {
	tabLocation := r.container.tabLocation
	if fyne.CurrentDevice().IsMobile() && (tabLocation == TabLocationLeading || tabLocation == TabLocationTrailing) {
		return TabLocationBottom
	}

	return r.container.tabLocation
}
```
And does not have different behaviour for Mobile Portrait vs Mobile Landscape

andydotxyz:
> Fyne v2.0.0 AppTab on Mobile always moved Leading & Trailing to Bottom

Strange, seems I misremembered how it was previously working.

Issue #682, opened 2020-02-16
Complete Native Menu Implementation


Jonconradt:
Currently missing from native menus, see [darwin implementation](https://github.com/fyne-io/fyne/blob/master/internal/driver/glfw/menu_darwin.m):

- [Enable](https://developer.apple.com/documentation/appkit/nsmenuitem/1514863-isenabled)/Disable a menu item 
- Separators
- KeyEquivalents
- Indentation levels

The Apple Documentation for MenuItem: https://developer.apple.com/documentation/appkit/nsmenuitem

While I realize that these seem like small things compared to the larger work that needs to be done, menus are a primary UX element for users. These small features would help the apps feel equivalent to other UX frameworks.

andydotxyz:
This would be great to add - but we will have to add support for those same things inside the Fyne canvas menus as well (for systems other than macOS).

Separator and disabled could be done relatively easily, indentation is tracked by another ticket (#395) but the key equivalents will take more work overall in the toolkit first.

toaster:
Separators were added to develop branch and will be part of 1.3.
Disabled menu items are planned but probably won’t be part of 1.3.
The same applies to KeyEquivalents (aka Shortcuts).

Regarding indentation: I don’t think that this is useful. Apples design documentation (https://developer.apple.com/design/human-interface-guidelines/macos/menus/menu-anatomy/) says:

> Use a submenu instead of indenting menu items. Indentation results in an inconsistent interface and doesn’t express relationships between menu items. If you need to indent, then a submenu is a better choice.

Submenus probably will land in 1.3 (#905).

toaster:
#905 now contains the support for native submenus, too

vmcrash:
See also https://github.com/fyne-io/fyne/issues/1372

andydotxyz:
I think the key equivalent / shortcut feature is the only remaining item

andydotxyz:
We think this is complete now in Bowmore release :)

PR #1975, opened 2021-02-18
main menu: add disabled and checked state for menu items

### LINKED ISSUE #1753: MenuItem support Selection Indication

tehsphinx:
### Description:
This draft / PR is intended to add enabling/disabling of menu items as well as adding the ability to add a check before a menu item.

Fixes #1753
Addresses Enable/Disable part of #682

### Checklist:

- [ ] Tests included. (objectiveC tests missing)
- [ ] Tests all pass.
- [ ] Public APIs match existing style. (to be discussed)

### DIFF
--- a/widget/menu_item.go
+++ b/widget/menu_item.go
@@ -274,8 +274,11 @@ func (r *menuItemRenderer) Refresh() {
 
 	if r.i.Item.Disabled {
 		r.text.Color = theme.DisabledColor()
+		// TODO: enable after fixing translucence and alpha channel (https://github.com/fyne-io/fyne/issues/1977)
+		// r.checkIcon.Translucency = 0.7
 	} else {
 		r.text.Color = theme.ForegroundColor()
+		r.checkIcon.Translucency = 0
 	}
 	r.text.Refresh()
 
--- a/widget/menu_item.go
+++ b/widget/menu_item.go
@@ -47,10 +47,14 @@ func (i *menuItem) CreateRenderer() fyne.WidgetRenderer {
 		icon = canvas.NewImageFromResource(theme.MenuExpandIcon())
 		objects = append(objects, icon)
 	}
+	var checkIcon *canvas.Image
+	checkIcon = canvas.NewImageFromResource(theme.ConfirmIcon())
+	objects = append(objects, checkIcon)
 	return &menuItemRenderer{
 		BaseRenderer: widget.NewBaseRenderer(objects),
 		i:            i,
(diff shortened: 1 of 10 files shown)


tehsphinx:
Would be happy about some help with tests, especially the objectiveC tests as I don't know objectiveC at all.

widget/menu_item.go
```diff
@@ -245,6 +271,24 @@ func (r *menuItemRenderer) Refresh() {
 		r.background.Hide()
 	}
 	r.background.Refresh()
+
+	if r.i.Item.Disabled {
+		r.text.Color = theme.DisabledColor()
+		// TODO: enable after fixing translucence and alpha channel (https://github.com/fyne-io/fyne/issues/1977)
```
andydotxyz, line 277: This is not the right way - it should be changed to a DisabledResource and the colour will update appropriately.
You could look at the existing `Button` widget that disables icons as appropriate.

widget/menu_item.go
```diff
@@ -245,6 +271,24 @@ func (r *menuItemRenderer) Refresh() {
...
+		// r.checkIcon.Translucency = 0.7
+	} else {
+		r.text.Color = theme.ForegroundColor()
+		r.checkIcon.Translucency = 0
```
andydotxyz, line 281: Shouldn't be needed - see above, it will be toggling disabled type for the resource instead.

andydotxyz (changes requested):
This looks really great thanks.
Please merge develop in to fix the conflict and address the TODO which has a solution available already.

prifre:
I wrote on slack: 
"I am looking for how to disable/enable a MenuItem...
I looked at https://github.com/fyne-io/fyne/pull/1975 but I cannot implementation...
What is the status?"
I got reply from tehsphinx  13:32
"fyne has made some decisions that make it unsuitable for our needs, so we moved on to other alternatives."

andydotxyz:
I will try to pick this up myself to follow the PR in to landing.

andydotxyz:
Thanks for your hard work on this @tehsphinx, I have resolved the merge conflicts and fixed a couple of issues noted above to hopefully land this cool feature - #2261.

PR #1978, opened 2021-02-19
Improved static ci

Jacalz:
### Description:

This updates the static analysis to use Go 1.16 for the improvements with using `go install` with a version number (only used for staticcheck for now as it receives a lot of updates and just pulling latest might break us). It also allows it to be used without modifying `go.mod` (thus not needing to turn off modules each time). Go 1.16 also has a slightly faster linker so the installation and build of the ci tools should be a bit faster.

It also switches to a faster go setup action which is about three times faster while also allowing us to see inline staticcheck reports in PR diffs (cool stuff). Only doing this for static analysis for now as I can't get it to past for the platform tests just yet.
I also made sure to tell staticcheck what Go version we are targeting with potential newer checks not being possible to apply for us.

I plan on getting the platform tests updated for Go 1.16 as well, but I want one or two bug fix releases to land first just in case.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -13,9 +13,9 @@ jobs:
     - name: Get dependencies
       run: |
         sudo apt-get update && sudo apt-get install gcc libgl1-mesa-dev libegl1-mesa-dev libgles2-mesa-dev libx11-dev xorg-dev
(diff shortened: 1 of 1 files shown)


andydotxyz (approved):

andydotxyz:
Slick :)

PR #1982, opened 2021-02-19
Allow custom Quit menu item

tehsphinx:
### Description:
fyne automatically adds a Quit menu item if the last item in the first menu is not called "Quit". This does not work for e.g. other languages.

This PR adds an option to suppress that without breaking the previous behaviour.

It does change the behaviour slightly if no action is provided on the "Quit" item. Then it adds the quit action. Can be removed if you don't agree.

Thought I saw an issue for this, but can't find it any more.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/menu.go
+++ b/internal/driver/glfw/menu.go
@@ -9,16 +9,29 @@ func buildMenuOverlay(menus *fyne.MainMenu, c fyne.Canvas) fyne.CanvasObject {
 		fyne.LogError("Main menu must have at least one child menu", nil)
 		return nil
 	}
+
+	menus = addMissingQuit(menus)
+	return NewMenuBar(menus, c)
+}
+
+func addMissingQuit(menus *fyne.MainMenu) *fyne.MainMenu {
 	var firstItem *fyne.MenuItem
 	if len(menus.Items[0].Items) > 0 {
 		firstItem = menus.Items[0].Items[len(menus.Items[0].Items)-1]
+		if firstItem.Label == "Quit" {
+			firstItem.IsQuit = true
+		}
 	}
-	if firstItem == nil || firstItem.Label != "Quit" { // make sure the first menu always has a quit option
-		quitItem := fyne.NewMenuItem("Quit", func() {
-			fyne.CurrentApp().Quit()
-		})
+	if firstItem == nil || !firstItem.IsQuit { // make sure the first menu always has a quit option
+		quitItem := fyne.NewMenuItem("Quit", nil)
 		menus.Items[0].Items = append(menus.Items[0].Items, fyne.NewMenuItemSeparator(), quitItem)
 	}
-
-	return NewMenuBar(menus, c)
+	for _, item := range menus.Items[0].Items {
+		if item.IsQuit && item.Action == nil {
+			item.Action = func() {
+				fyne.CurrentApp().Quit()
+			}
+		}
+	}
+	return menus
 }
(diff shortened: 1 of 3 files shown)


AlbinoGeek:
I had made an issue about this, but apparently Mac OS X enforces a Quit option in the menu? [but again... I don't see why we couldn't suppress it for every other OS.)

tehsphinx:
@AlbinoGeek: The code in question is never run for MacOS native menu. We are not enforcing a Quit for native menus on MacOS. The OS creates that an extra menu entry containing a Quit item itself, anyway.

For `-tags=no_native_menus` MacOS will still add a native menu to the application containing a quit entry.

Not sure how enforcing a Quit entry in the non-native menu is helping with the situation. I think it makes it worse as we are enforcing a duplicate Quit on MacOS. (if using the `no_native_menus`.)

On other OSes it might have made sense to enforce a Quit item. But enforcing it via the label text is a problem.

internal/driver/glfw/menu.go
```diff
@@ -9,16 +9,29 @@ func buildMenuOverlay(menus *fyne.MainMenu, c fyne.Canvas) fyne.CanvasObject {
 		fyne.LogError("Main menu must have at least one child menu", nil)
 		return nil
 	}
+
+	menus = addMissingQuit(menus)
+	return NewMenuBar(menus, c)
+}
+
+func addMissingQuit(menus *fyne.MainMenu) *fyne.MainMenu {
 	var firstItem *fyne.MenuItem
```
andydotxyz, line 18: I know it's not strictly in your code, but I was confused by this variable - could it be renamed to `lastItem`?

menu.go
```diff
@@ -15,8 +15,9 @@ func NewMenu(label string, items ...*MenuItem) *Menu {
 // MenuItem is a single item within any menu, it contains a display Label and Action function that is called when tapped.
 type MenuItem struct {
 	ChildMenu   *Menu
-	IsSeparator bool
 	Label       string
+	IsSeparator bool
+	IsQuit      bool
```
andydotxyz, line 20: New API should have the documentation:
```go
// Since: 2.1
```

menu.go
```diff
@@ -15,8 +15,9 @@ func NewMenu(label string, items ...*MenuItem) *Menu {
 // MenuItem is a single item within any menu, it contains a display Label and Action function that is called when tapped.
 type MenuItem struct {
 	ChildMenu   *Menu
-	IsSeparator bool
```
andydotxyz, line 18: Is the re-ordering required/correct here?

andydotxyz (changes requested):
Good approach. The full fix is probably the localisation suppot, but this seeks like a good approach that would be compatible with that when we have it.

Perhaps the title should be "Allow custom Quit menu item", as the current one isn't quite right?

andydotxyz:
It is my understanding that with this change we are still enforcing a quit item (for OS that support quitting) but that in this way your app can now provide an alternative.
I am happy wih this approach, as it's not (like the title suggests) allowing apps with no ability to quit...

internal/driver/glfw/menu.go
tehsphinx: Yes, gladly. I tripped a few times over it, too 😄

menu.go
tehsphinx: It optimizes the struct for memory alignment having the two bool values after each other. `Label` seemed more central to a menu item than the 2 boolean values, so I decided to move `Label` before `IsSeparator`.

If that is not something you care about (memory alignment), I can undo the move.

tehsphinx: I moved the `IsQuit` before `Label` now, undoing the reorder but keeping the two boolean values together.

andydotxyz (approved):
Looks good to me

PR #1986, opened 2021-02-20
mobileDriver.currentWindow should not return hidden windows, fixes #1835

### LINKED ISSUE #1835: Creating a window shows it before calling .Show() and .Hide() does not work
### Describe the bug:

Creating a second window and building using the mobile driver shows the second window without every calling `.Show()` on the second window. The only way to avoid showing the window is to explicitly close it. Calling `.Hide()` leaves the first window not rendering at all.

### To Reproduce:
1. Run the example code using `go run -tags mobile main.go`
2. Close the second window by adding `w2.Close()`
3. Run the modified example code using `go run -tags mobile main.go`
4. Change the `w2.Close()` call to `w2.Hide()`
5. Run the modified example code using `go run -tags mobile main.go`

### Screenshots:

image

### Example code:

```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Hello")

	w2 := a.NewWindow("Second")
	w2.SetContent(widget.NewLabel("Second window"))

		hello,
			hello.SetText("Welcome :)")
		}),
	))

	w.ShowAndRun()
}
```

 - **OS:** Linux (-tags mobile) and Android have been tested
 - **Version:** 5.10.6
 - **Go version:** 1.15.6 and 1.13.8
 - **Fyne version:** 2.0.0-rc5

fpabl0:
### Description:

Fixes #1835 

### Checklist:

- [ ] Tests included. (not sure how to test mobile driver)
- [x] Tests all pass.

### DIFF
--- a/internal/driver/gomobile/driver.go
+++ b/internal/driver/gomobile/driver.go
@@ -63,7 +63,15 @@ func (d *mobileDriver) currentWindow() fyne.Window {
 		return nil
 	}
 
-	return d.windows[len(d.windows)-1]
+	var last fyne.Window
+	for i := len(d.windows) - 1; i >= 0; i-- {
+		last = d.windows[i]
+		if last.(*window).visible {
+			return last
+		}
+	}
+
+	return last
 }
 
 func (d *mobileDriver) RenderedTextSize(text string, size float32, style fyne.TextStyle) fyne.Size {


internal/driver/gomobile/driver.go
```diff
@@ -63,7 +63,15 @@ func (d *mobileDriver) currentWindow() fyne.Window {
...
-	return d.windows[len(d.windows)-1]
+	var last fyne.Window
+	for i := len(d.windows) - 1; i >= 0; i-- {
+		last = d.windows[i]
+		if last.(*window).visible {
+			return last
+		}
+	}
+
+	return last
```
andydotxyz, line 74: This code will return the last hidden window if they are all hidden - is that intended?
Above we return nil if there are no windows, I think we should do that here too

andydotxyz (changes requested):

internal/driver/gomobile/driver.go
fpabl0: > This code will return the last hidden window if they are all hidden - is that intended?

Yes, that was intended.

> Above we return nil if there are no windows, I think we should do that here too

That was something that I was not sure. In mobile, if all the windows are hidden, what does it mean? 🤔

andydotxyz: It should not be done - app will be unusable. But not sure that the most recently added window is the right response in this situation... It should never happen, but there is specifically code in their to accommodate it so I figured we should be sure that is what should occur if it can.

fpabl0: > But not sure that the most recently added window is the right response in this situation...

Indeed this change is returning the parent window (not the most recently added) because I am looping in reverse. Now I realized I have misread your first comment. It is not the last hidden window but the first one (parent one).

andydotxyz: Oh yes I see, sorry for misreading the first time

andydotxyz:
Looks good to me, as the original reporter @Jacalz could you please test this as well?

Jacalz (approved):

PR #1991, opened 2021-02-20
Add github workflow for `golangci-lint`

tehsphinx:
### Description:
Adds a github workflow to run [golangci-lint](https://golangci-lint.run/). This can replace parts of the `checks` workflow.

Currently `staticcheck` in the `checks` workflow has the Go version set to `1.12`. This option is not available for `golangci-lint`, so `staticcheck` for golangci-lint is disabled. All other linters of the `checks` workflow are set up the same as in the current `develop` branch.

### Checklist:
- [ ] Duplicate linters removed from `checks` workflow.
- [ ] More possible linters discussed and enabled.


tehsphinx:
Ok... something went seriously wrong here...

PR #1992, opened 2021-02-20
Add github workflow for `golangci-lint`

tehsphinx:
### Description:
Adds a github workflow to run [golangci-lint](https://golangci-lint.run/). This can replace parts of the `checks` workflow.

Currently `staticcheck` in the `checks` workflow has the Go version set to `1.12`. This option is not available for `golangci-lint`, so `staticcheck` for golangci-lint is disabled. All other linters of the `checks` workflow are set up the same as in the current `develop` branch.

### Checklist:
- [ ] Duplicate linters removed from `checks` workflow.
- [ ] More possible linters discussed and enabled.


.golangci.yml
```diff
@@ -0,0 +1,91 @@
...
+#    statements: 40
+#  lll:
+#    line-length: 150
+#  nakedret:
+#    max-func-lines: 10
+
+linters:
+  disable-all: true
+  enable:
+#    - asciicheck
```
tehsphinx, line 31: > Simple linter to check that your code does not contain non-ASCII identifiers

Not really important, but doesn't hurt either.

.golangci.yml
```diff
@@ -0,0 +1,91 @@
...
+#    - bodyclose
```
tehsphinx, line 32: > checks whether HTTP response body is closed successfully

Enabling this immediately finds a problem. I'd definitely enable this one as connections cannot be closed (they time out instead) if the body is not closed e.g. of a `http.Get` like in `LoadResourceFromURLString` function.

.golangci.yml
```diff
@@ -0,0 +1,91 @@
...
+#    - deadcode
```
tehsphinx, line 33: > Finds unused code

I'd enable it to help keep the code base clean.

.golangci.yml
```diff
@@ -0,0 +1,91 @@
...
+#    - depguard
```
tehsphinx, line 34: > Go linter that checks if package imports are in a list of acceptable packages

I use this in combination with a blacklist of packages I do not want to be used. Mainly helpful if you e.g. decided to use a certain test package and want to warn developers about using another package.

.golangci.yml
```diff
@@ -0,0 +1,91 @@
...
+#    - dogsled
+#    - dupl
+#    - errcheck
```
tehsphinx, line 37: > Errcheck is a program for checking for unchecked errors in go programs. These unchecked errors can be critical bugs in some cases

I would enable this. It currently finds around 20 to 30 places where the error is not handled. The linter can be silenced easily by using the `_ = someFunc()` notation to explicitly ignore an error. Then it is clear the developer realised there was an error and decided to ignore it, rather than not realising the function could fail.

.golangci.yml
```diff
@@ -0,0 +1,91 @@
...
+#    - exhaustive
+#    - funlen
+#    - gochecknoglobals
+#    - gochecknoinits
+#    - goconst
+#    - gocritic
```
tehsphinx, line 43: > The most opinionated Go source code linter

In my opinion one of the best linters for Go. Finds a variety of things from "suspicious" `if` statements (potential bug), replacing `i = i + 1` with `i++`, using switch statements instead of long if .. else if .. else cases to missing spaces after `//` in comments. Different checks can be disabled per config.

Jacalz (changes requested):
Sorry, but I am not going to accept this. Golangci-lint is a big bloated mess in my opinion. More linters is not always the answer and is just going to slow down the ci. I want our static analysis infrastructure to stay lean and fast while letting us have the control over it.

.golangci.yml
```diff
@@ -0,0 +1,91 @@
...
+    - gocyclo
+#    - gofmt
+    - goimports
+    - golint
+#    - gomnd
+#    - goprintffuncname
+#    - gosec
```
tehsphinx, line 50: > Inspects source code for security problems

Might not be right fit for fyne as it is not used to create web servers.

.golangci.yml
```diff
@@ -0,0 +1,91 @@
...
+#    - gosimple
+    - govet
+#    - ineffassign
```
tehsphinx, line 53: > Detects when assignments to existing variables are not used

Another linter being able to detect some potentially serious bugs. e.g. errors that are assigned to an already existing `err` variable but never checked.

.golangci.yml
```diff
@@ -0,0 +1,91 @@
...
+#    - interfacer
+#    - lll
+#    - maligned
+#    - misspell
+#    - nakedret
```
tehsphinx, line 58: > Finds naked returns in functions greater than a specified function length

Want to ban naked returns from the code? This ones for you. And yes, there are naked returns in fyne in functions with 35 lines of code and more.

.golangci.yml
```diff
@@ -0,0 +1,91 @@
...
+#    - noctx
+#    - nolintlint
+#    - prealloc
+#    - rowserrcheck
+#    - scopelint
```
tehsphinx, line 63: > Scopelint checks for unpinned variables in go programs

A linter for the dangerous bug of using the scope variable of a for loop in a goroutine without passing or pinning it.

.golangci.yml
```diff
@@ -0,0 +1,91 @@
...
+    - staticcheck
+#    - structcheck
+#    - stylecheck
+#    - typecheck
+#    - unconvert
```
tehsphinx, line 68: > Remove unnecessary type conversions

which there are quite a few in `internal/mobile/binres/table.go`

Jacalz (changes requested):
Can we please move the `golangci/golangci-lint-action@v2.4.0` action into into the static analysis file that we already have please? No need to run things twice.

I would also prefer if you enable `skip-go-installation` so we use that Go version that we have defined in that file. This way we can control what go version we run with. I think you can remove the other commented out optional settings as we don't really have any reason to enable them.

.github/workflows/static_analysis.yml
```diff
@@ -20,17 +20,13 @@ jobs:
...
-
-    - name: Golint
-      run: golint -set_exit_status $(go list -tags ci ./...)
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v2.4.0
+      with:
+        version: v1.37
+        skip-go-installation: true
+        # skip-pkg-cache: true
+        # skip-build-cache: true
```
Jacalz, line 29: Not sure if we really need to keep these. We can always add them in if needed sometime in the future.

.github/workflows/static_analysis.yml
```diff
@@ -20,17 +20,13 @@ jobs:
...
 
     - name: Staticcheck
       run: staticcheck -go 1.12 -f stylish ./...
```
Jacalz, line 32: I know it's outside the scope of the PR, but could you please remove `-f stylish` from here?
The creator mentioned that it does not work with inline mentions in PRs. https://github.com/fyne-io/fynedesk/pull/174#issuecomment-782888549

.github/workflows/static_analysis.yml
```diff
@@ -20,17 +20,13 @@ jobs:
...
-      run: go vet -tags ci ./...
-
-    - name: Goimports
-      run: test -z $(goimports -e -d . | tee /dev/stderr)
-
-    - name: Gocyclo
-      run: gocyclo -over 50 .
-
-    - name: Golint
-      run: golint -set_exit_status $(go list -tags ci ./...)
```
Jacalz, line 33: You can remove the installation of these tools (except for staticcheck) further up in the file.

Jacalz (changes requested):
Looking much better. A few comments made inline.

.github/workflows/static_analysis.yml
tehsphinx, line 29:
```suggestion
```

.github/workflows/static_analysis.yml
tehsphinx, line 30:
```suggestion
      run: staticcheck -go 1.12 -f stylish ./...
```
```suggestion
      run: staticcheck -go 1.12 ./...
```

.github/workflows/static_analysis.yml
tehsphinx: right, I forgot.

.golangci.yml
```diff
@@ -0,0 +1,18 @@
+# See https://golangci-lint.run/usage/configuration/#config-file for configuration options.
+
+linters-settings:
+  golint:
+    min-confidence: 0.8
+  gocyclo:
+    min-complexity: 50
```
tehsphinx, line 7: BTW: setting complexity to 50 you can as well take the linter out. If you don't want to change some of the complex functions or ignore some linter warnings that are not really applicable, you can always put a comment in the code above (or behind) the line that a specific linter complains about:

```go
// notlint: gocyclo
func someVeryComplexFunc() {}
```

tehsphinx: There even is a linter to complain about such comments that are not needed any more, in case that becomes a problem some day.

andydotxyz: > BTW: setting complexity to 50 you can as well take the linter out. Default is already 30 (which is high!). Recommendation is between 10-20.

We had some complex methods when this was added. We did not want to ignore them - instead we wanted to fix them over time.
So we will do so. Each time we make the complexity lower we can lower this value.

tehsphinx: well... now you are ignoring all functions. Adding a comment to ignore them and a TODO or Issue would better, I think.

andydotxyz: > now you are ignoring all functions

I don't understand your point. We are slowly reducing complexity.
If we support writing exclusions from these checks then they diminish the requirement.

tehsphinx: Too much discussion. Moving on.

Jacalz (approved):
I think this is looking good now. Probably want some more eyes before we land it though.

andydotxyz (approved):

PR #1995, opened 2021-02-21
Refactor: fixing (potential) bugs around closing

tehsphinx:
### Description:
There was one http body close missing.

The other fixes resolve around ignoring the error on writable file handler closing (which should not be done). An error might be returned from `Close` that indicates a write failure. For more details see https://www.joeshaw.org/dont-defer-close-on-writable-files/

### Checklist:
- [x] Tests all pass.

### DIFF
--- a/app/preferences.go
+++ b/app/preferences.go
@@ -64,19 +64,22 @@ func (p *preferences) load() {
 	}
 }
 
-func (p *preferences) loadFromFile(path string) error {
+func (p *preferences) loadFromFile(path string) (err error) {
 	file, err := os.Open(path) // #nosec
 	if err != nil {
 		if os.IsNotExist(err) {
-			err := os.MkdirAll(filepath.Dir(path), 0700)
-			if err != nil {
+			if err := os.MkdirAll(filepath.Dir(path), 0700); err != nil {
 				return err
 			}
 			return nil
 		}
 		return err
 	}
-	defer file.Close()
+	defer func() {
+		if r := file.Close(); r != nil {
+			err = r
+		}
+	}()
 	decode := json.NewDecoder(file)
 
 	p.InMemoryPreferences.WriteValues(func(values map[string]interface{}) {
--- a/app/preferences.go
+++ b/app/preferences.go
@@ -76,7 +76,7 @@ func (p *preferences) loadFromFile(path string) (err error) {
 		return err
 	}
 	defer func() {
-		if r := file.Close(); r != nil {
+		if r := file.Close(); r != nil && err == nil {
 			err = r
 		}
 	}()
--- a/cmd/fyne/internal/commands/bundle.go
+++ b/cmd/fyne/internal/commands/bundle.go
@@ -138,35 +138,23 @@ func (b *Bundler) Run(args []string) {
 	}
 }
 
-func (b *Bundler) bundleAction(ctx *cli.Context) error {
+func (b *Bundler) bundleAction(ctx *cli.Context) (err error) {
 	if ctx.Args().Len() != 1 {
 		return errors.New("missing required file or directory parameter after flags")
 	}
 
 	outFile := os.Stdout
 	if b.out != "" {
-		fileModes := os.O_RDWR | os.O_CREATE | os.O_TRUNC
-		if b.noheader {
-			fileModes = os.O_RDWR | os.O_APPEND
+		file, closeFile, err := getOutputFile(b.out, b.noheader)
+		if err != nil {
+			return err
 		}
-
-		f, err := os.OpenFile(b.out, fileModes, 0666)
-		if err == nil {
-			outFile = f
-		} else {
-			if os.IsNotExist(err) {
-				f, err = os.Create(b.out)
-				if err == nil {
-					outFile = f
-				} else {
-					fyne.LogError("Unable to read, or create, output file : "+b.out, err)
-					return err
-				}
-			} else {
-				fyne.LogError("Unable to open output file", err)
-				return err
+		defer func() {
+			if r := closeFile(); r != nil {
+				err = r
 			}
-		}
+		}()
+		outFile = file
 	}
 
 	arg := ctx.Args().First()
@@ -225,6 +213,32 @@ func (b *Bundler) doBundle(filepath string, out *os.File) {
 		b.name = sanitiseName(path.Base(filepath), b.prefix)
 	}
 	writeResource(filepath, b.name, out)
+}
+
+func getOutputFile(filePath string, noheader bool) (file *os.File, close func() error, err error) {
+	fileModes := os.O_RDWR | os.O_CREATE | os.O_TRUNC
+	if noheader {
+		fileModes = os.O_RDWR | os.O_APPEND
+	}
+
+	f, err := os.OpenFile(filePath, fileModes, 0666)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			fyne.LogError("Unable to open output file", err)
+			return nil, nil, err
+		}
+
+		// try creating the file
+		f, err = os.Create(filePath)
+		if err != nil {
+			fyne.LogError("Unable to read, or create, output file : "+filePath, err)
+			return nil, nil, err
+		}
+	}
+
+	return f, func() error {
+		return f.Close()
+	}, nil
 }
 
 func sanitiseName(file, prefix string) string {
--- a/cmd/fyne/internal/commands/bundle.go
+++ b/cmd/fyne/internal/commands/bundle.go
@@ -145,7 +145,7 @@ func (b *Bundler) bundleAction(ctx *cli.Context) (err error) {
 
 	outFile := os.Stdout
 	if b.out != "" {
-		file, closeFile, err := getOutputFile(b.out, b.noheader)
+		file, closeFile, err := openOutputFile(b.out, b.noheader)
 		if err != nil {
 			return err
 		}
@@ -215,7 +215,7 @@ func (b *Bundler) doBundle(filepath string, out *os.File) {
 	writeResource(filepath, b.name, out)
 }
 
-func getOutputFile(filePath string, noheader bool) (file *os.File, close func() error, err error) {
+func openOutputFile(filePath string, noheader bool) (file *os.File, close func() error, err error) {
 	fileModes := os.O_RDWR | os.O_CREATE | os.O_TRUNC
 	if noheader {
 		fileModes = os.O_RDWR | os.O_APPEND
--- a/cmd/fyne/internal/commands/bundle.go
+++ b/cmd/fyne/internal/commands/bundle.go
@@ -236,7 +236,7 @@ func getOutputFile(filePath string, noheader bool) (file *os.File, close func() 
 		}
 	}
 
-	return f, f.Close(), nil
+	return f, f.Close, nil
 }
 
 func sanitiseName(file, prefix string) string {
--- a/cmd/fyne/internal/commands/bundle.go
+++ b/cmd/fyne/internal/commands/bundle.go
@@ -236,9 +236,7 @@ func getOutputFile(filePath string, noheader bool) (file *os.File, close func() 
 		}
 	}
 
-	return f, func() error {
-		return f.Close()
-	}, nil
+	return f, f.Close(), nil
 }
 
 func sanitiseName(file, prefix string) string {
--- a/cmd/fyne/internal/commands/package-darwin.go
+++ b/cmd/fyne/internal/commands/package-darwin.go
@@ -31,7 +31,7 @@ func (p *Packager) packageDarwin() (err error) {
 		return errors.Wrap(err, "Failed to write plist template")
 	}
 	defer func() {
-		if r := infoFile.Close(); r != nil {
+		if r := infoFile.Close(); r != nil && err == nil {
 			err = r
 		}
 	}()
@@ -65,7 +65,7 @@ func (p *Packager) packageDarwin() (err error) {
 		return errors.Wrap(err, "Failed to open destination file")
 	}
 	defer func() {
-		if r := dest.Close(); r != nil {
+		if r := dest.Close(); r != nil && err == nil {
 			err = r
 		}
 	}()
--- a/cmd/fyne/internal/commands/package-darwin.go
+++ b/cmd/fyne/internal/commands/package-darwin.go
@@ -20,25 +20,31 @@ type darwinData struct {
 	Category      string
 }
 
-func (p *Packager) packageDarwin() error {
+func (p *Packager) packageDarwin() (err error) {
 	appDir := util.EnsureSubDir(p.dir, p.name+".app")
 	exeName := filepath.Base(p.exe)
 
 	contentsDir := util.EnsureSubDir(appDir, "Contents")
 	info := filepath.Join(contentsDir, "Info.plist")
-	infoFile, _ := os.Create(info)
+	infoFile, err := os.Create(info)
+	if err != nil {
+		return errors.Wrap(err, "Failed to write plist template")
+	}
+	defer func() {
+		if r := infoFile.Close(); r != nil {
+			err = r
+		}
+	}()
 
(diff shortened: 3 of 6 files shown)


Jacalz:
I have been meaning to do this for a while now. Thank you very much for getting this sorted :+1:

cmd/fyne/internal/commands/package-darwin.go
```diff
@@ -20,25 +20,31 @@ type darwinData struct {
...
 	appDir := util.EnsureSubDir(p.dir, p.name+".app")
 	exeName := filepath.Base(p.exe)
 
 	contentsDir := util.EnsureSubDir(appDir, "Contents")
 	info := filepath.Join(contentsDir, "Info.plist")
-	infoFile, _ := os.Create(info)
+	infoFile, err := os.Create(info)
+	if err != nil {
+		return errors.Wrap(err, "Failed to write plist template")
+	}
```
tehsphinx, line 32: There is a change in behaviour here. Before a failure in opening/creating the file was ignored. However it would have failed below anyway when trying to write the template to the file. So it should behave the same.

cmd/fyne/internal/commands/package-darwin.go
```diff
@@ -20,25 +20,31 @@ type darwinData struct {
...
+	defer func() {
+		if r := infoFile.Close(); r != nil {
+			err = r
+		}
+	}()
```
tehsphinx, line 37: This is new. before the infoFile was never closed.

cmd/fyne/internal/commands/release.go
```diff
@@ -260,21 +261,18 @@ func (r *Releaser) packageIOSRelease() error {
...
+	if _, err := r.writeEntitlements(templates.EntitlementsDarwinMobile, struct{ TeamID, AppID string }{
 		TeamID: team,
 		AppID:  r.appID,
-	}
-	err = templates.EntitlementsDarwinMobile.Execute(entitlements, entitlementData)
-	entitlements.Close()
-	if err != nil {
+	}); err != nil {
 		return errors.New("failed to write entitlements plist template")
 	}
```
tehsphinx, line 277: If this should be the same as in `packageMacOSRelease`, then `cleanup` needs to be executed here, too.

tehsphinx:
> I have been meaning to do this for a while now. Thank you very much for getting this sorted 👍

I was merely following some linter issues. `gosec` was complaining about file closing without error checks. Then I started to stumble over more stuff. I don't think this is complete. There probably is more.

Jacalz:
> > I have been meaning to do this for a while now. Thank you very much for getting this sorted 👍
> 
> I was merely following some linter issues. `gosec` was complaining about file closing without error checks. Then I started to stumble over more stuff. I don't think this is complete. There probably is more.

I know. That’s exactly what I have Ben intending to do for quite some time 🙂

andydotxyz:
@Jacalz from your comments I assumed you would be best placed to review this.
Just say if you would prefer that I do?

Jacalz:
> @Jacalz from your comments I assumed you would be best placed to review this.
> Just say if you would prefer that I do?

Please go ahead. I have a lot to do at the moment.

app/preferences.go
```diff
@@ -64,19 +64,22 @@ func (p *preferences) load() {
 	}
 }
 
-func (p *preferences) loadFromFile(path string) error {
+func (p *preferences) loadFromFile(path string) (err error) {
```
andydotxyz, line 67: Naming the return in this case is slightly confusing as it is shadowed completely by the next line

cmd/fyne/internal/commands/bundle.go
```diff
@@ -138,35 +138,23 @@ func (b *Bundler) Run(args []string) {
...
-				} else {
-					fyne.LogError("Unable to read, or create, output file : "+b.out, err)
-					return err
-				}
-			} else {
-				fyne.LogError("Unable to open output file", err)
-				return err
+		defer func() {
+			if r := closeFile(); r != nil {
+				err = r
```
andydotxyz, line 154: This is not setting the return `err` but the one a few lines above that shadows it.

cmd/fyne/internal/commands/bundle.go
```diff
@@ -227,6 +215,32 @@ func (b *Bundler) doBundle(filepath string, out *os.File) {
...
+
+		// try creating the file
+		f, err = os.Create(filePath)
+		if err != nil {
+			fyne.LogError("Unable to read, or create, output file : "+filePath, err)
+			return nil, nil, err
+		}
+	}
+
+	return f, func() error {
```
andydotxyz, line 239: surely `return f, f.Close, nil` is sufficient?

cmd/fyne/internal/commands/bundle.go
```diff
@@ -227,6 +215,32 @@ func (b *Bundler) doBundle(filepath string, out *os.File) {
 	writeResource(filepath, b.name, out)
 }
 
+func getOutputFile(filePath string, noheader bool) (file *os.File, close func() error, err error) {
```
andydotxyz, line 218: Naming here could be improved - "get" methods should have no side effects, but this one has (it will create files). Not inspired, but maybe "createOrAppendFile" is more descriptive?

cmd/fyne/internal/commands/package-darwin.go
```diff
@@ -20,25 +20,31 @@ type darwinData struct {
...
 	contentsDir := util.EnsureSubDir(appDir, "Contents")
 	info := filepath.Join(contentsDir, "Info.plist")
-	infoFile, _ := os.Create(info)
+	infoFile, err := os.Create(info)
+	if err != nil {
+		return errors.Wrap(err, "Failed to write plist template")
+	}
+	defer func() {
+		if r := infoFile.Close(); r != nil {
+			err = r
```
andydotxyz, line 35: This is also setting a shadow error instead of the return value...

cmd/fyne/internal/commands/package-darwin.go
```diff
@@ -58,7 +64,11 @@ func (p *Packager) packageDarwin() error {
 	if err != nil {
 		return errors.Wrap(err, "Failed to open destination file")
 	}
-	defer dest.Close()
+	defer func() {
+		if r := dest.Close(); r != nil {
+			err = r
```
andydotxyz, line 69: Another shadow error I think

cmd/fyne/internal/commands/release.go
andydotxyz: Yes it should, thanks

cmd/fyne/internal/util/file.go
```diff
@@ -55,7 +54,11 @@ func copyFileMode(src, tgt string, perm os.FileMode) error {
 	if err != nil {
 		return err
 	}
-	defer target.Close()
+	defer func() {
+		if r := target.Close(); r != nil {
+			err = r
```
andydotxyz, line 59: This is shadowed as well

andydotxyz (changes requested):
A lot of good catches here, thanks.
However many of the "defer returned" errors are not returned as their values are shadowed by local variables.

cmd/fyne/internal/commands/bundle.go
tehsphinx: No it doesn't. That is not how named returns and `defer` works.

app/preferences.go
tehsphinx: It doesn't work if you remove the named return

cmd/fyne/internal/commands/package-darwin.go
tehsphinx: Nope. That's not how named returns and 'defer' works

cmd/fyne/internal/commands/package-darwin.go
tehsphinx: same again: No

cmd/fyne/internal/util/file.go
tehsphinx: again: No

cmd/fyne/internal/commands/bundle.go
tehsphinx: How about:
```suggestion
func openOutputFile(filePath string, noheader bool) (file *os.File, close func() error, err error) {
```

cmd/fyne/internal/commands/bundle.go
tehsphinx: yes it is. Got refactored and not simplified...

cmd/fyne/internal/commands/bundle.go
andydotxyz: Sure

cmd/fyne/internal/commands/bundle.go
andydotxyz: It seems from a little more digging that this is not special for `defer` but for named returns.
Newly assigned variables don't shadow return names like I thought they did.
Apologies for the confusion.

andydotxyz (dismissed):

cmd/fyne/internal/util/file.go
```diff
@@ -55,7 +54,11 @@ func copyFileMode(src, tgt string, perm os.FileMode) error {
 	if err != nil {
 		return err
 	}
-	defer target.Close()
+	defer func() {
+		if r := target.Close(); r != nil {
```
fpabl0, line 58: same here, could be simplified by doing just:
```go
err = target.Close()
return
```

cmd/fyne/internal/commands/release.go
```diff
@@ -447,6 +443,26 @@ func (r *Releaser) validate() error {
...
+		if r := entitlements.Close(); r != nil {
+			err = r
+		}
+	}()
+
+	if err := tmpl.Execute(entitlements, entitlementData); err != nil {
+		return nil, err
+	}
+	return func() {
+		_ = os.Remove(entitlementPath)
```
fpabl0, line 464: If error is not handled why not just:
```go
os.Remove(entitlementPath)
```

fpabl0 (commented):
Some comments about named returns and simplification.

cmd/fyne/internal/util/file.go
```diff
@@ -39,9 +39,8 @@ func EnsureSubDir(parent, name string) string {
 	return path
 }
 
-func copyFileMode(src, tgt string, perm os.FileMode) error {
-	_, err := os.Stat(src)
-	if err != nil {
+func copyFileMode(src, tgt string, perm os.FileMode) (err error) {
+	if _, err := os.Stat(src); err != nil {
```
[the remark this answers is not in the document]
tehsphinx: That does not matter, no.

And definitely not use a `naked return`.

cmd/fyne/internal/util/file.go
tehsphinx: no, that would not work.

cmd/fyne/internal/commands/release.go
tehsphinx: because that is a sign to your future self, other developers (and linters, if you use them) that you are well aware of the fact that you are ignoring the error. If you use `os.Remove(entitlementPath)` nobody knows if you just forgot to handle the error or you purposefully ignore it.

cmd/fyne/internal/util/file.go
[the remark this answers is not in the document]
tehsphinx: There even is a proposal to entirely remove `naked returns` because of the extra overhead it has for the reader of the code:  https://github.com/golang/go/issues/21291

cmd/fyne/internal/util/file.go
tehsphinx: You'd not return the error then, but just return from the deferred function, ignoring the error.

cmd/fyne/internal/util/file.go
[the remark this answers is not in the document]
fpabl0: Oh got it, that's a good point.

cmd/fyne/internal/commands/release.go
fpabl0: Good point :)

cmd/fyne/internal/util/file.go
fpabl0: You are right, I didn't realize it, sorry for the confusion. However, now I wonder if it could be rewritten to:
```go
if r := target.Close(); r != nil && err == nil {
     err = r
}
```
Otherwise, the error in the defer function will overwrite any previous error that maybe it is more helpful because it is the root cause of the problem?

tehsphinx: That is correct.

```suggestion
		if r := target.Close(); r != nil && err == nil {
```

app/preferences.go
```diff
@@ -64,19 +64,22 @@ func (p *preferences) load() {
...
+			if err := os.MkdirAll(filepath.Dir(path), 0700); err != nil {
 				return err
 			}
 			return nil
 		}
 		return err
 	}
-	defer file.Close()
+	defer func() {
+		if r := file.Close(); r != nil {
```
fpabl0, line 79: add `&& err == nil` to avoid ignoring the first error.

cmd/fyne/internal/commands/package-darwin.go
```diff
@@ -20,25 +20,31 @@ type darwinData struct {
...
 
 	contentsDir := util.EnsureSubDir(appDir, "Contents")
 	info := filepath.Join(contentsDir, "Info.plist")
-	infoFile, _ := os.Create(info)
+	infoFile, err := os.Create(info)
+	if err != nil {
+		return errors.Wrap(err, "Failed to write plist template")
+	}
+	defer func() {
+		if r := infoFile.Close(); r != nil {
```
fpabl0, line 34: same here

cmd/fyne/internal/commands/package-darwin.go
```diff
@@ -58,7 +64,11 @@ func (p *Packager) packageDarwin() error {
 	if err != nil {
 		return errors.Wrap(err, "Failed to open destination file")
 	}
-	defer dest.Close()
+	defer func() {
+		if r := dest.Close(); r != nil {
```
fpabl0, line 68: same here

cmd/fyne/internal/commands/release.go
```diff
@@ -447,6 +445,26 @@ func (r *Releaser) validate() error {
...
 }
 
+func (r *Releaser) writeEntitlements(tmpl *template.Template, entitlementData interface{}) (cleanup func(), err error) {
+	entitlementPath := filepath.Join(r.dir, "entitlements.plist")
+	entitlements, err := os.Create(entitlementPath)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if r := entitlements.Close(); r != nil {
```
fpabl0, line 455: same here

fpabl0 (changes requested):
Based on the last commit, I think that change should be applied to the remaining defer functions too.

app/preferences.go
tehsphinx:
```suggestion
		if r := file.Close(); r != nil && err == nil {
```

cmd/fyne/internal/commands/package-darwin.go
tehsphinx:
```suggestion
		if r := infoFile.Close(); r != nil && err == nil {
```

cmd/fyne/internal/commands/package-darwin.go
tehsphinx:
```suggestion
		if r := dest.Close(); r != nil && err == nil {
```

cmd/fyne/internal/commands/release.go
tehsphinx:
```suggestion
		if r := entitlements.Close(); r != nil && err == nil {
```

fpabl0 (approved):
LGTM. Thanks :)

andydotxyz (approved):
Thanks.
And thanks for educating us on some of the defer and return code.
Glad to hear they are considering removing naked returns :)

PR #1996, opened 2021-02-21
move to an infinite queue for binding callbacks

### LINKED ISSUE #1838: Data binding limited to 1024 simultaneous operations
I forgot to commit a fix to allow higher number of elements to be called in the same loop item.
Not often a problem except when we bind a large structure as they can all hit the "initialise on first connection" at once.

andydotxyz:
Fixes #1838

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/queue.go
+++ b/data/binding/queue.go
@@ -19,7 +19,7 @@ func makeInfiniteQueue() (chan<- *itemData, <-chan *itemData) {
 	in := make(chan *itemData)
 	out := make(chan *itemData)
 	go func() {
-		var queued []*itemData
+		queued := make([]*itemData, 0, 1024)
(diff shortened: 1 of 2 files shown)


andydotxyz:
I can’t quite make sense of the benchmark results because the way the code changed impacts how it is reported, but it looks like CPU-wise a similar speed. However it requires more memory than the previous version, though of course it scales based on how many items you simultaneously trigger

fpabl0 (approved):
LGTM

andydotxyz:
Thanks for the 👍 @fpabl0. This has now been open for comment for over 7 days so we'll merge it and test through release process. Better solutions may be found for future releases.

PR #1998, opened 2021-02-21
replacing http.Get with context call

tehsphinx:
### Description:
http.Get without context is best not used any more.

### Checklist:
- [x] Tests all pass.

### DIFF
--- a/resource.go
+++ b/resource.go
@@ -69,6 +69,8 @@ func LoadResourceFromURLString(urlStr string) (Resource, error) {
 	return LoadResourceFromURLStringWithCtx(ctx, urlStr)
 }
 
+// LoadResourceFromURLStringWithCtx creates a new StaticResource in memory using the body of the specified URL.
+// The provided context will be passed the the http request.
 func LoadResourceFromURLStringWithCtx(ctx context.Context, urlStr string) (Resource, error) {
 	req, err := http.NewRequest(http.MethodGet, urlStr, nil)
 	if err != nil {
--- a/resource.go
+++ b/resource.go
@@ -1,9 +1,11 @@
 package fyne
 
 import (
+	"context"
 	"io/ioutil"
 	"net/http"
 	"path/filepath"
+	"time"
 )
 
 // Resource represents a single binary resource, such as an image or font.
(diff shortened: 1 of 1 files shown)


resource.go
```diff
@@ -59,7 +61,14 @@ func LoadResourceFromPath(path string) (Resource, error) {
 
 // LoadResourceFromURLString creates a new StaticResource in memory using the body of the specified URL.
 func LoadResourceFromURLString(urlStr string) (Resource, error) {
-	res, err := http.Get(urlStr)
+	req, err := http.NewRequest(http.MethodGet, urlStr, nil)
+	if err != nil {
+		return nil, err
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
```
tehsphinx, line 68: timeout ok?

AlbinoGeek: I have to wonder what other things would benefit from adding context, and if the timeout should be stored somewhere where it can be used across these places?

tehsphinx: There probably are more places, yes. If you can think of any, feel free to point me there.

I changed improved the code so users can also set the context themselves.

tehsphinx:
Is there a way to restart the failed test? It failed downloading a `go mod` dependency...

resource.go
```diff
@@ -59,7 +63,21 @@ func LoadResourceFromPath(path string) (Resource, error) {
...
-	res, err := http.Get(urlStr)
+	ctx, cancel := context.WithTimeout(context.Background(), httpReqTimeout)
+	defer cancel()
+
+	return LoadResourceFromURLStringWithCtx(ctx, urlStr)
+}
+
+// LoadResourceFromURLStringWithCtx creates a new StaticResource in memory using the body of the specified URL.
+// The provided context will be passed the the http request.
+func LoadResourceFromURLStringWithCtx(ctx context.Context, urlStr string) (Resource, error) {
```
andydotxyz, line 74: I don't think that this new function adds much utility, if someone understand `Context` then they can probably manage the connection as well.
Just FYI the `LoadResourceFromURLString` will be deprecated some time soon as we move it to `fyne.URI` and the `storage` package (once http and https are in the core repository code).

resource.go
andydotxyz: It might be simpler to just add `Timeout` to client - does the context add more functionality than the timeout?
```
client := http.Client{
    Timeout: time.Second * 60,
}
res, err := client.Get(url)
```

andydotxyz:
Moving to context seems like a good thing to do, but I don't think we should mix new APIs with fixing code health.

resource.go
tehsphinx: > then they can probably manage the connection as well

Don't see how they should do that as there is no access exposed to the http connection.

resource.go
tehsphinx: It is equivalent if we don't add the 'LoadResourceFromURLStringWithCtx' function.

If we do, the context is superior as it does not only allow timeouts but can also e.g. be cancelled early.

resource.go
andydotxyz: I was not proposing re-using an http connection, just that reading bytes from a URL is relatively straight forward - this helper was for people who don't want the hassle.
I guess when I meant "the connetion" I was meaning conceptually not the specific variable we use.

PR #2000, opened 2021-02-21
interface generalization

tehsphinx:
### Description:
A more general use of existing functionality is possible by using the appropriate interfaces. This merge request applies that in 2 places.

### Checklist:
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/bundle.go
+++ b/cmd/fyne/internal/commands/bundle.go
@@ -187,7 +187,7 @@ func (b *Bundler) bundleAction(ctx *cli.Context) error {
 	return nil
 }
 
-func (b *Bundler) dirBundle(dirpath string, out *os.File) error {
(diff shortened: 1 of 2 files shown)


andydotxyz:
In the draw code we want to know that images are `NRGBA` Because that is the pixel format used lower down. Using the abstraction interface could lead to converting colour models during paint, which is going to add overhead.

Jacalz:
I agree with @andydotxyz on this. At the same time, I also don't really see the need for the other interface generalizations. Using interfaces is slower than just using the pointers and if we don't use something else than those specific types, I don't see why we would switch to the interfaces.

andydotxyz:
I guess I could see the argument if these were public methods...

tehsphinx:
We can close this is you don't agree. It's just a suggestion. And yes, it definitely would make a lot more sense if the methods/functions were public.

PR #2016, opened 2021-02-23
Have something like "issues_playground" folder added to .gitignore

fpabl0:
### Description:

Most of the time, when there are issues to be solved, it is desired to run the code the user provide on the issue template, so I think it should be great to have a folder (ignored by git) to store these issues. In this way, it would be easier to test if the fix works or not along the solution progress.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,9 @@ cmd/hello/hello.apk
 cmd/hello/hello.app
 cmd/hello/hello.exe
 
+### Issues playground
+cmd/issues_playground/*
+
 ### Tests
 **/testdata/failed
 


Jacalz (commented):
Hmm, I'm not so sure about this. Why do we need to store them in the fyne repository at all? I think everyone has different ways of doing it. You could easily have a folder somewhere else on the folder where you keep them.

fpabl0:
> Hmm, I'm not so sure about this. Why do we need to store them in the fyne repository at all?

Hmm, they won't be really stored, the folder will be ignored. And I think it is needed to guarantee you are testing the fyne repo you are editing and not the one in GOPATH or in the modules' cache.

> I think everyone has different ways of doing it.

Yes, you are right, however I thought it would be great, because you usually open your editor in the fyne repository folder to start solving issues (at least for me, sorry if it is just me), and if you can have that example in the same workspace, you don't need to open anything else. But yes, this is a subjective decision.

> You could easily have a folder somewhere else on the folder where you keep them.

I use that approach, however I always have to move them to the workspace and remove them before committing (I know it is possible to commit specific files, but it is easier to do `git add .` 😅).

andydotxyz:
There is no need to be testing Fyne apps inside the fyne repository.
If you want to (and sometimes I do) just use `main.go` inside `cmd/hello` ;)

fpabl0:
> There is no need to be testing Fyne apps inside the fyne repository.

Hmm how? I have multiple copies of fyne 😅 one in the GOPATH, other in modules' cache and in folders. For my projects use modules' cache that has the latest stable release, and I have the fork in a folder, how do I tell go to use the fyne repo inside that folder instead of the one in GOPATH or in modules' cache? 🤔

> If you want to (and sometimes I do) just use main.go inside cmd/hello ;)

Me too, but at the end I need to undo those changes to avoid changing the hello world when committing :(

stuartmscott:
Ideally the code becomes a test case so the issue can never re-emerge

fpabl0:
> Ideally the code becomes a test case so the issue can never re-emerge

Yes, that's a good point. It is just convenient proposal. It is not needed, but should be great in my opinion.

andydotxyz:
> how do I tell go to use the fyne repo inside that folder instead of the one in GOPATH or in modules' cache? 🤔

Use module replace directive.
I'm going to close this PR because I think it relates more to questions about how to set up the environment and should not require changes to the repository.

fpabl0:
> Use module replace directive.

Oh got it, thanks. Sorry for open this.

andydotxyz:
Don't apologise, it was an interesting question.
It might have been easier in this instance to ask in the Slack.
Perhaps we should improve our contributor documentation to help with some of these questions.

PR #2025, opened 2021-02-24
fyne_settings: Do not allow scale 0 that causes infinite text size

fpabl0:
### Description:

Do not allow scale 0 that causes infinite text size.

### DIFF
--- a/cmd/fyne_settings/settings/appearance.go
+++ b/cmd/fyne_settings/settings/appearance.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"time"
 
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/app"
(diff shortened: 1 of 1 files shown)


andydotxyz:
Nice catch, but I wonder - would it be better to put the check earlier so that it shows the default text size selected

fpabl0:
> Nice catch, but I wonder - would it be better to put the check earlier so that it shows the default text size selected

Oh I see what you mean, you are right, will update :)

fpabl0:
@andydotxyz updated :)

cmd/fyne_settings/settings/appearance.go
```diff
@@ -128,6 +129,8 @@ func (s *Settings) createPreview() image.Image {
 
 	c.SetContent(tabs)
 	c.Resize(fyne.NewSize(380, 380))
+	// wait for indicator animation
+	time.Sleep(150 * time.Millisecond)
```
andydotxyz, line 133: I think it would be better if this referenced the constant instead - `canvas.DurationShort`

andydotxyz (changes requested):

cmd/fyne_settings/settings/appearance.go
fpabl0: You are right, updated! :)

andydotxyz (approved):

Issue #1887, opened 2021-02-01
Table widget crash when scrolling rapidly


wenhsinhsieh:
### Describe the bug:
The table widget goes panic when scrolling up and down rapidly

### To Reproduce:
Steps to reproduce the behaviour:
1. Run fyne_demo
2. switch to Collections-> Table
3. maximize window
4. mouse hold on vertical scroll bar can scroll up and down rapidly
5. eventually the program goes panic

Note: It seems doesn't happen (or maybe hard to reproduce) if the window is not maximize. It also doesn't happen if scrolling slowly.

### Screenshots:
![圖片](https://user-images.githubusercontent.com/3656867/106438443-77648f00-64b1-11eb-9346-87d539a60e9d.png)

### Device (please complete the following information):
 - **OS:** Windows
 - **Version:** 10
 - **Go version:** 1.14.6
 - **Fyne version:** 2.0

wenhsinhsieh:
Additional info: Linux also happens too.

### Device (please complete the following information):
 - **OS:** Ubuntu
 - **Version:** 18.04
 - **Go version:** 1.14.7
 - **Fyne version:** 2.0

### Callstack:

```
panic: runtime error: index out of range [31] with length 31

goroutine 15 [running, locked to thread]:
fyne.io/fyne/v2/widget.(*tableRenderer).moveIndicators(0xc0003096c0)
	/home/user/go/src/fyne.io/fyne/v2/widget/table.go:350 +0xc65
fyne.io/fyne/v2/widget.(*tableRenderer).Layout(0xc0003096c0, 0x4443400044b7b333)
	/home/user/go/src/fyne.io/fyne/v2/widget/table.go:240 +0x2f
fyne.io/fyne/v2/internal/driver/glfw.updateLayout(0xb5eb20, 0xc000a3a780)
	/home/user/go/src/fyne.io/fyne/v2/internal/driver/glfw/canvas.go:638 +0x122
fyne.io/fyne/v2/internal/driver/glfw.(*glCanvas).ensureMinSize.func1(0xc000c18f80)
	/home/user/go/src/fyne.io/fyne/v2/internal/driver/glfw/canvas.go:349 +0x18d
fyne.io/fyne/v2/internal/driver/glfw.(*glCanvas).walkTree.func2(0xb5d860, 0xc000a29b60, 0xb5ea60, 0xc000028ab0)
[...]

PR #2026, opened 2021-02-25
Ensure we never access out of range table dividers, fixes #2018

### LINKED ISSUE #2018: fyne_demo crash after selecting custom Theme and table
### Describe the bug:
Fyne_demo crash after selecting custom theme on advanced and enter on Table 

### To Reproduce:
open fyne_demo
go to advanced 
select custom theme 
open collection tree 
select table (when custome theme on)

### Example code:
fyne_demo code 

 - **OS:** windows 10
 - **Go version:** 1.16
 - **Fyne version:** 2

fpabl0:
### Description:

Fixes #2018 and possibly #1887.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/table_test.go
+++ b/widget/table_test.go
@@ -303,3 +303,42 @@ func TestTable_ShowVisible(t *testing.T) {
 	cellRenderer.Refresh()
 	assert.Equal(t, 8, len(cellRenderer.Objects()))
 }
+
+type separatorThicknessZeroTheme struct{}
+
+func (separatorThicknessZeroTheme) Color(c fyne.ThemeColorName, v fyne.ThemeVariant) color.Color {
+	return theme.DefaultTheme().Color(c, v)
+}
+
+func (separatorThicknessZeroTheme) Font(style fyne.TextStyle) fyne.Resource {
+	return theme.DefaultTheme().Font(style)
+}
+
+func (separatorThicknessZeroTheme) Icon(n fyne.ThemeIconName) fyne.Resource {
+	return theme.DefaultTheme().Icon(n)
+}
+
+func (separatorThicknessZeroTheme) Size(s fyne.ThemeSizeName) float32 {
+	if s == theme.SizeNameSeparatorThickness {
+		return 0
+	}
+	return theme.DefaultTheme().Size(s)
+}
+
+func TestTable_SeparatorThicknessZero_NotPanics(t *testing.T) {
+	test.NewApp()
+	defer test.NewApp()
+
+	test.ApplyTheme(t, &separatorThicknessZeroTheme{})
+
+	table := NewTable(
+		func() (int, int) { return 500, 150 },
+		func() fyne.CanvasObject {
+			return NewLabel("placeholder")
+		},
+		func(TableCellID, fyne.CanvasObject) {})
+
+	assert.NotPanics(t, func() {
+		table.Resize(fyne.NewSize(400, 644))
+	})
+}
--- a/widget/table_test.go
+++ b/widget/table_test.go
@@ -304,7 +304,26 @@ func TestTable_ShowVisible(t *testing.T) {
 	assert.Equal(t, 8, len(cellRenderer.Objects()))
 }
 
-type separatorThicknessZeroTheme struct{}
+func TestTable_SeparatorThicknessZero_NotPanics(t *testing.T) {
+	test.NewApp()
+	defer test.NewApp()
+
+	test.ApplyTheme(t, &separatorThicknessZeroTheme{})
+
+	table := NewTable(
+		func() (int, int) { return 500, 150 },
+		func() fyne.CanvasObject {
+			return NewLabel("placeholder")
+		},
+		func(TableCellID, fyne.CanvasObject) {})
+
+	assert.NotPanics(t, func() {
+		table.Resize(fyne.NewSize(400, 644))
+	})
+}
+
+type separatorThicknessZeroTheme struct {
+}
 
 func (separatorThicknessZeroTheme) Color(c fyne.ThemeColorName, v fyne.ThemeVariant) color.Color {
 	return theme.DefaultTheme().Color(c, v)
@@ -323,22 +342,4 @@ func (separatorThicknessZeroTheme) Size(s fyne.ThemeSizeName) float32 {
 		return 0
 	}
 	return theme.DefaultTheme().Size(s)
-}
-
-func TestTable_SeparatorThicknessZero_NotPanics(t *testing.T) {
-	test.NewApp()
-	defer test.NewApp()
-
-	test.ApplyTheme(t, &separatorThicknessZeroTheme{})
-
-	table := NewTable(
-		func() (int, int) { return 500, 150 },
-		func() fyne.CanvasObject {
-			return NewLabel("placeholder")
-		},
-		func(TableCellID, fyne.CanvasObject) {})
-
-	assert.NotPanics(t, func() {
-		table.Resize(fyne.NewSize(400, 644))
-	})
 }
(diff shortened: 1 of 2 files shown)


andydotxyz (commented):
Good catch on the fix - but I don't understand why the 0 width theme is required to test it

fpabl0:
> Good catch on the fix - but I don't understand why the 0 width theme is required to test it

Theme with separator thickness equals to zero is a special case that causes the current code accesses out of range dividers, indeed that is the root cause of #2018, so I thought it would be useful if there is a test for that situation, doesn't it?

andydotxyz:
If that is the root cause, it should of course be fixed.
What I am unsure about is that the fix does not reference size - are you certain that is the root cause?

andydotxyz:
Just FYI we just got the following in another PR tests - is this the same bug? The widget tests don't have 0 width separators.

```
panic: runtime error: index out of range [1] with length 1

goroutine 309 [running]:
fyne.io/fyne/v2/widget.(*tableRenderer).moveIndicators(0xc001d8b650)
	/Users/runner/work/fyne/fyne/widget/table.go:350 +0xc30
fyne.io/fyne/v2/widget.(*Table).CreateRenderer.func1(0x0)
	/Users/runner/work/fyne/fyne/widget/table.go:78 +0x65
fyne.io/fyne/v2/internal/widget.(*Scroll).updateOffset(0xc0010599e0, 0x0, 0x143f120)
	/Users/runner/work/fyne/fyne/internal/widget/scroller.go:580 +0x1dd
fyne.io/fyne/v2/internal/widget.(*Scroll).Refresh(0xc0010599e0)
	/Users/runner/work/fyne/fyne/internal/widget/scroller.go:534 +0x3a
fyne.io/fyne/v2/internal/widget.(*Scroll).Resize(...)
	/Users/runner/work/fyne/fyne/internal/widget/scroller.go:546
fyne.io/fyne/v2/widget.(*tableRenderer).Layout(0xc001d8b650, 0x0)
	/Users/runner/work/fyne/fyne/widget/table.go:243 +0x145
fyne.io/fyne/v2/widget.(*Table).CreateRenderer(0xc000dd1680, 0x14555e0, 0xc000dd1680)
	/Users/runner/work/fyne/fyne/widget/table.go:81 +0x59c
fyne.io/fyne/v2/internal/cache.Renderer(0x14f3dc0, 0xc000dd1680, 0xc000dd1680, 0x14f3dc0)
	/Users/runner/work/fyne/fyne/internal/cache/widget.go:29 +0x119
fyne.io/fyne/v2/internal/app.ApplyThemeTo(0x14f2820, 0xc000dd1680, 0x14f5080, 0xc000dd1720)
	/Users/runner/work/fyne/fyne/internal/app/theme.go:17 +0x1bf
fyne.io/fyne/v2/internal/app.ApplySettings(0x14f0ea0, 0xc001bc9e00, 0x14f4a40, 0xc001d578f0)
	/Users/runner/work/fyne/fyne/internal/app/theme.go:36 +0x105
fyne.io/fyne/v2/test.NewApp.func1(0xc001f68180, 0xc001d578f0)
	/Users/runner/work/fyne/fyne/test/testapp.go:106 +0x79
created by fyne.io/fyne/v2/test.NewApp
	/Users/runner/work/fyne/fyne/test/testapp.go:102 +0x1f8
```

fpabl0:
> What I am unsure about is that the fix does not reference size - are you certain that is the root cause?

Well the real problem is the initial state (like the case when windows is created and it reports zero size and then the real size), in that case almost all related variables to the renderer are 0 and the value of the theme.SeparatorThickness is decisive:
```go
for y := theme.Padding() + t.scroll.Offset.Y - float32(count) - separatorThickness; y < t.scroll.Offset.Y+t.t.size.Height && i < rows-1 && divs < len(t.dividers); y += t.cellSize.Height + separatorThickness {
	if y < theme.Padding()+t.scroll.Offset.Y {
		continue
	}
        ....
}
```
If you run that test in the current develop branch it will fail 100% of the times with index out of range.

> Just FYI we just got the following in another PR tests - is this the same bug? The widget tests don't have 0 width separators.

It is the same bug (index out of range), but different situation for sure. This PR would solve that bug too as now it is impossible to go out of range. For example #1887 is a different situation but with the same bug, however that situation is harder to replicate and so doing a test for that case is so difficult.

The test with SeparatorThicknessZero theme is consistent (and a special case), but certainly is not the best way to test this in a general form, should I remove it?

andydotxyz:
Thanks for the summary. It would be a shame not to test so you're probably right to leave it.
Might be able to reduce the line count by embedding a theme, such as `test.Theme()` and override just the Size method...

Oh, and the private type should be at the end of the file, below the exported test methods.

fpabl0:
> Might be able to reduce the line count by embedding a theme, such as test.Theme() and override just the Size method...

Yes, that would be great, but currently it is not possible because there isn't a public theme object available to be embedded, is it?

> Oh, and the private type should be at the end of the file, below the exported test methods.

You are right, I forgot it, sorry, will update :)

andydotxyz:
> Yes, that would be great, but currently it is not possible because there isn't a public theme object available to be embedded, is it?

This example works - could be adapted to reduce your line count I am sure:
```go
type bigTextTheme struct {
	fyne.Theme
}

func (b *bigTextTheme) Size(n fyne.ThemeSizeName) float32 {
	if n == theme.SizeNameText {
		return 22
	}

	return b.Theme.Size(n)
}
```

That can be set like follows:
```go
	a.Settings().SetTheme(&bigTextTheme{test.Theme()})
```

fpabl0:
Oh that's great, I didn't know that it is possible to embed an interface 😅 into a struct.

fpabl0:
@andydotxyz  updated :)

widget/table_test.go
```diff
@@ -303,3 +303,32 @@ func TestTable_ShowVisible(t *testing.T) {
 	cellRenderer.Refresh()
 	assert.Equal(t, 8, len(cellRenderer.Objects()))
 }
+
+func TestTable_SeparatorThicknessZero_NotPanics(t *testing.T) {
+	test.NewApp()
+	defer test.NewApp()
+
+	test.ApplyTheme(t, &separatorThicknessZeroTheme{theme.DefaultTheme()})
```
andydotxyz, line 311: For tests we should use `test.Theme()`

widget/table_test.go
```diff
@@ -303,3 +303,32 @@ func TestTable_ShowVisible(t *testing.T) {
...
+
+type separatorThicknessZeroTheme struct {
+	fyne.Theme
+}
+
+func (separatorThicknessZeroTheme) Size(s fyne.ThemeSizeName) float32 {
+	if s == theme.SizeNameSeparatorThickness {
+		return 0
+	}
+	return theme.DefaultTheme().Size(s)
```
andydotxyz, line 333: If you delegate to the embedded theme instead you'll get the correct fallback instead of accessing a new theme.

andydotxyz (changes requested):

widget/table_test.go
fpabl0: Oh yes, good catch. I only removed the extra code and didn't realize it, will update! :)

widget/table_test.go
fpabl0: Got it.

fpabl0: done

widget/table_test.go
fpabl0: done

andydotxyz (approved):

andydotxyz:
Thanks. I guess this should be added to the `v2.0.1` work as well?

fpabl0:
> Thanks.

You're welcome.

> I guess this should be added to the v2.0.1 work as well?

Yes, it should. In fact, any bug fixes that don't involve new methods or breaking changes should be added IMHO :)

PR #2030, opened 2021-02-25
Remove separator between items for the List widget

lucor:
### Description:

This PR removes the separator between items for the List widget to improve the compatibility with the [material design spec](https://material.io/components/lists)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
I don’t agree that simply removing separators makes it closer to spec.

For example “When lists don’t have an anchoring element (such as an icon), spacing may not be enough to separate content. Full-bleed dividers can help separate individual tiles.” https://material.io/components/dividers#types

andydotxyz:
Closing as discussed. Maybe making this optional would be a possible approach?

PR #2032, opened 2021-02-26
mobile tests on CI

toaster:
### Description:

This PR enables mobile tests on CI.
It adjusts the tests and masters (images/markups) to make them pass.
There are known issues with doc tab rendering which are addressed in #1962.

### Checklist:

- ~~[ ] Tests included.~~
- [x] Tests all pass.


andydotxyz:
This add 6 * mobile test runs (doubling the number we currently check), is that what we want?
`release/v2.0.x` adds just 1 more test run to check mobile which seems like enough for the verification

Jacalz:
I think the approach from @andydotxyz might be the best one going forward. It allows us to more easily expanding the mobile tests in the future. This adds a little bit too much complexity to the regular platform tests in my opinion.

andydotxyz:
I landed my file to develop. If that was merged in to this PR we should see the checks pass.
They are currently marked as not-required because they fail on develop :(.

toaster:
How about I’ll simply extract the test fixes out of this PR into a separate one?

andydotxyz:
> How about I’ll simply extract the test fixes out of this PR into a separate one?

Probably easier, go ahead :)

PR #2041, opened 2021-02-27
Fix data race condition in fyne preferences

fpabl0:
### Description:

Fix data race condition in fyne preferences, reported by doing:
`go test -race fyne.io/fyne/v2/app`

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):

andydotxyz:
This looks good,  yet you will need to merge in develop so the mobile tests pass - I have made mobile tests required now as they should be working.

PR #2042, opened 2021-02-28
Fix race conditions in preference bindings

fpabl0:
### Description:

Fixes race conditions in preference bindings.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


data/binding/binding_test.go
```diff
@@ -22,7 +22,6 @@ func TestBase_AddListener(t *testing.T) {
 	assert.Equal(t, 1, len(data.listeners))
 
 	waitForItems()
-	data.trigger()
```
andydotxyz, line 25: I’m curious about this change, can you help?

fpabl0: Yes, `data.AddListener(fn)` already enqueues DataChanged callback, so trigger it again is not needed, is it?

andydotxyz: That looks to be the case

andydotxyz (approved):

andydotxyz:
I felt like I hit some hangs whilst testing this, but not certain this was the cause

andydotxyz (changes requested):
Freezes now reproducible it seems.
Open fyne_demo on non-data tab, then switch to data binding.
Sometimes hangs

fpabl0:
> Open fyne_demo on non-data tab, then switch to data binding.
Sometimes hangs

Oh that is bad, I will investigate.

fpabl0:
> Sometimes hangs

I am not able to reproduce it. does it happen frequently?

fpabl0:
With this last commit, do you still experiment freeze?

andydotxyz:
> With this last commit, do you still experiment freeze?

I can't seem to replicate it any more.

andydotxyz (approved):

PR #2044, opened 2021-02-28
CI: mobile

toaster:
This PR adjusts the tests and masters (images/markups) of mobile tests to make them pass.
This allows to require mobile specs for pull requests.
There are known issues with doc tab rendering which are addressed in #1962.


andydotxyz (approved):
Not a fan of commenting out code, but in this case it should not be long lasted and it gets our builds running on mobile in the meantime

Issue #1957, opened 2021-02-14
race detector: fatal error: checkptr: pointer arithmetic computed bad pointer value


tehsphinx:
### Describe the bug:
Starting an application with `go run --race .` crashes with the following error at startup:

```
fatal error: checkptr: pointer arithmetic computed bad pointer value

goroutine 40 [running, locked to thread]:
runtime.throw(0x4601533, 0x37)
/usr/local/go/src/runtime/panic.go:1116 +0x72 fp=0xc000629538 sp=0xc000629508 pc=0x403b9b2
runtime.checkptrArithmetic(0xc, 0x0, 0x0, 0x0)
/usr/local/go/src/runtime/checkptr.go:26 +0xd7 fp=0xc000629568 sp=0xc000629538 pc=0x400d1f7
github.com/go-gl/gl/v3.2-core/gl.PtrOffset(...)
/Users/user/go/pkg/mod/github.com/go-gl/gl@v0.0.0-20190320180904-bf2b1f2f34d7/v3.2-core/gl/conversions.go:55
fyne.io/fyne/v2/internal/painter/gl.(*glPainter).glCreateBuffer(0xc000606000, 0xc000026050, 0x14, 0x14, 0x0)
/Users/user/go/pkg/mod/fyne.io/fyne/v2@v2.0.0/internal/painter/gl/gl_core.go:219 +0x409 fp=0xc000629608 sp=0xc000629568 pc=0x43f2f29
fyne.io/fyne/v2/internal/painter/gl.(*glPainter).drawTextureWithDetails(0xc000606000, 0x46726c0, 0xc0002e42a0, 0xc0006296e8, 0x4140000041000000, 0x41a8000042ee0000, 0x43fa000044480000, 0x0, 0x3f800000)
/Users/user/go/pkg/mod/fyne.io/fyne/v2@v2.0.0/internal/painter/gl/draw.go:27 +0x152 fp=0xc000629668 sp=0xc000629608 pc=0x43ee272
fyne.io/fyne/v2/internal/painter/gl.(*glPainter).drawText(0xc000606000, 0xc0002e42a0, 0x4140000041000000, 0x43fa000044480000)
/Users/user/go/pkg/mod/fyne.io/fyne/v2@v2.0.0/internal/painter/gl/draw.go:78 +0x357 fp=0xc000629708 sp=0xc000629668 pc=0x43ef057
[...]

PR #2048, opened 2021-03-01
Fix several race conditions in internal/driver/glfw package

fpabl0:
### Description:

Fix several race conditions in internal/driver/glfw package.
To test this fix, run:
```
go test -race -gcflags=all=-d=checkptr=0 fyne.io/fyne/v2/internal/driver/glfw
```

The flag `-gcflags=all=-d=checkptr=0` is to temporarily ignore the crash related to gl (#1957).

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


fpabl0:
After testing a lot the test called `TestWindow_Focus` (have written a script to run the test again and again until it fails haha) , I finally found where is the data race and in consequence why this test occasionally fails. The race occurs here:
```go
w.charInput(w.viewport, 'd')
w.keyPressed(w.viewport, glfw.KeyTab, 0, glfw.Press, 0)
```
`charInput(...)` always enqueue the event to be called in a go routine later.
`w.keyPressed` with `glfw.KeyTab` event runs immediately in the main loop. Internally, it does `w.canvas.FocusNext()` that triggers the `FocusLost` method on widget.Entry and so refreshing the entry.Text and ignoring the `charInput` events because of the data race.

So, the question is should we also enqueue `w.canvas.FocusNext()` event from keyPressed to ensure all these operations are running sequentially in the same go routine?

andydotxyz:
> So, the question is should we also enqueue w.canvas.FocusNext() event from keyPressed to ensure all these operations are running sequentially in the same go routine?

Yes I think that sounds correct

andydotxyz (approved):
Seems like good improvements thanks

PR #2049, opened 2021-03-01
Allow custom themes to react theme variant changes, fixes #2006

### LINKED ISSUE #2006: Custom theme does not refresh when variant changes
### Describe the bug:

A custom theme loaded that supports light and dark modes not seem to switch correctly at runtime when settings change.

### To Reproduce:
1. Run an app with custom theme that supports variant (like github.com/andydotxyz/notes)
2. see current theme (light or dark)
3. load fyne_settings and change theme variant
4. Notice that the app does not update

 - **OS:** macOS
 - **Version:** Catalina
 - **Go version:** 1.15
 - **Fyne version:** develop

fpabl0:
### Description:

Fixes #2006 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/settings.go
+++ b/app/settings.go
@@ -14,7 +14,7 @@ import (
 // SettingsSchema is used for loading and storing global settings
 type SettingsSchema struct {
 	// these items are used for global settings load
-	ThemeName    string  `json:"theme"`
+	ThemeVariant string  `json:"theme"`
 	Scale        float32 `json:"scale"`
 	PrimaryColor string  `json:"primary_color"`
 }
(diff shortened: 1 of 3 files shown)


app/settings.go
```diff
@@ -14,7 +14,7 @@ import (
 // SettingsSchema is used for loading and storing global settings
 type SettingsSchema struct {
 	// these items are used for global settings load
-	ThemeName    string  `json:"theme"`
+	ThemeVariant string  `json:"theme"`
```
andydotxyz, line 17: This is part of our public API and cannot be changed outside a deprecation cycle and breaking release

cmd/fyne_settings/settings/appearance.go
```diff
@@ -89,11 +89,11 @@ func (s *Settings) LoadAppearanceScreen(w fyne.Window) fyne.CanvasObject {
 	return container.NewBorder(box, bottom, nil, nil, s.preview)
 }
 
-func (s *Settings) chooseTheme(name string) {
-	if name == systemThemeName {
-		name = ""
+func (s *Settings) chooseThemeVariant(variant string) {
```
andydotxyz, line 92: Not sure I would have renamed the method as it wasn't wrong in the way that variable names were

andydotxyz (changes requested):
This is a great catch thanks.
Unfortunately 1 blocker - we cannot change public API outside of a x.0 release and deprecation cycle.
Also a note on naming that's less important.

cmd/fyne_settings/settings/appearance.go
fpabl0: Yes, I was unsure about that change 😅, will revert that renaming :)

app/settings.go
fpabl0: Yes, I forgot it. Should we keep `ThemeName` instead of `ThemeVariant`? I renamed it, because I think the new name is more clear based on the usage

andydotxyz: To get this in a bugfix we have to keep it as-is. We could look to change it in the future

fpabl0: Got it, will update :)

app/settings_test.go
```diff
@@ -76,16 +76,68 @@ func TestOverrideTheme_IgnoresSettingsChange(t *testing.T) {
 	}
 	set.setupTheme()
 	assert.Equal(t, theme.LightTheme(), set.Theme())
+
```
fpabl0, line 79: I have edited this test, by removing the behavior that ignores forever the changes in the application once FYNE_THEME env var is set at least once, is that ok?

app/settings.go
fpabl0: done

cmd/fyne_settings/settings/appearance.go
fpabl0: done

app/settings_test.go
andydotxyz: FYNE_THEME is an override, once set it should not be changed

andydotxyz: I don't think we had considered the env changing during runtime. That could go either way.
What I meant was that if env is set nothing else should influence the setting

andydotxyz (approved):
Seems good. Clearing an env at runtime was not supported before - we can look at the details for that later

PR #2054, opened 2021-03-02
Ensure mobile window content enqueues a refresh event on show

fpabl0:
### Description:

Fixes issue running fyne_demo with mobile tag, reported on Slack.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/widget/scroller.go
+++ b/internal/widget/scroller.go
@@ -402,6 +402,9 @@ func (r *scrollContainerRenderer) handleShadowVisibility(offset, contentSize, sc
 }
 
 func (r *scrollContainerRenderer) updatePosition() {
+	if r.scroll.Content == nil {
+		return
+	}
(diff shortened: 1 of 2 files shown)


internal/widget/scroller.go
```diff
@@ -402,6 +402,9 @@ func (r *scrollContainerRenderer) handleShadowVisibility(offset, contentSize, sc
 }
 
 func (r *scrollContainerRenderer) updatePosition() {
+	if r.scroll.Content == nil {
```
fpabl0, line 405: While trying to fix the bug about `window.Show`, I found this other bug, that causes a crash in fyne_demo with mobile tag, not sure why it worked before. But it seems to be unrelated to the PR #1986

andydotxyz (approved):

andydotxyz:
Thanks, on develop and release branches

PR #2055, opened 2021-03-02
Track draggable object position in mobile, fixes #1061

### LINKED ISSUE #1061: Dragging out of widget drops event with mobile driver
### Describe the bug:
Open fyne_demo -> Widgets -> input and drag the slider.
If your finger goes above or below the widget it resets to 0 - but it should track the finger movement.
Most probably a bug in tracking motion events in mobile driver.

### To Reproduce:
1. Go to fyne_demo
2. Click on Widgets then input
3. Slide the slider then move your finger up
4. Norice it resets to 0 instead of following finger horizontal mottion

 - OS: iOS, Android, -tags mobile
 - Version: any
 - Go version: 1.14
 - Fyne version: develop

fpabl0:
### Description:

Track draggable object position to avoid moving it to (0,0) when the drag event goes out of the widget.

Fixes #1061 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz (changes requested):
Sorry, the geometry here is off somehow - it breaks the sliders in colour picker.
This is a tricky corner but my guess would be that yourr new variables are initialised without respecting the widget offset from parent.

fpabl0:
> Sorry, the geometry here is off somehow - it breaks the sliders in colour picker.

Oh good catch, will investigate.

fpabl0:
@andydotxyz I think, that issue could be a different one, on `develop` branch, colour picker does not work really well either:

https://user-images.githubusercontent.com/12239342/110345552-b334cc00-7ffc-11eb-9b4c-48234cf5ac02.mp4

Same thing occurs in `master` branch.

fpabl0:
> that issue could be a different one

I was wrong, yes it is something bad with the geometry due to this change, as text selection on entry is broken too. Nevertheless, above issues are real ones and maybe unrelated to this.

andydotxyz:
> @andydotxyz I think, that issue could be a different one, on develop branch, colour picker does not work really well either:

This is not the same - in your vidoes it may be a little laggy. On this PR if you tap the sliders they go to 100% and become immovable.

fpabl0:
> On this PR if you tap the sliders they go to 100% and become immovable.

Yes, I have just realized it in my previous comment where I noticed that text selection is broken too due to this PR.

>  in your vidoes it may be a little laggy

I don't think it it just laggy, the whole colour picker gets crazy (it keeps changing forever) but it is a different issue, as you pointed before.

fpabl0:
> if you tap the sliders they go to 100% and become immovable.

I think, now this is fixed, could you confirm?

andydotxyz (approved):
This looks like a great fix thanks.
Please open a new issue with the videos of the strange continuing-update for colour picker as there must be another issue to resolve somewhere

PR #2068, opened 2021-03-06
Fixes on widget.Entry

### LINKED ISSUE #1787: Text selection - Shift+Tab bug
### Describe the bug:

Using Shift+Tab to focus widgets breaks text selection.

### To Reproduce (Updated):
1. Run fyne_demo.
2. Go to any tab with more than one Entry.
3. Focus an Entry.
4. Press Shift + Tab to go to the previous Entry.
5. Return back again with Tab.
6. Try to use arrow keys to move in the text or tap on it to remove selection.
7. See the error

### Screenshots:

https://user-images.githubusercontent.com/12239342/104859892-4c493f80-58f6-11eb-8ce9-ff0b7b1c42aa.mp4

 - **OS:** MacOS 
 - **Go version:** 1.15.6 
 - **Fyne version:** develop @ 442a50c8

### LINKED ISSUE #1880: Cursor in Disabled Entry
### Describe the bug:

It is possible to place the cursor within a disable entry, further, it is possible sometimes to duplicate the cursor between both a disabled entry and an enabled entry.

### To Reproduce:

1. Use attached code
2. Select "one-way flight"
3. Click in the first Entry
4. Click in the second Entry a few times (only takes a couple)
5. Click back into the first entry

### Screenshots:
https://user-images.githubusercontent.com/1910461/106277221-93d0a500-61ed-11eb-9a49-1988d561fcb7.mp4
[ 0:17-0:20 and 0:26-0:31 ]

### Example code:

```go
package main

import (
	"time"

	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

var ukDateWithDecimals = "02.01.2006"

func main() {
	a := app.New()
	w := a.NewWindow("Flight Booker")

	d1 := binding.NewString()
	d2 := binding.NewString()

	t1 := widget.NewEntryWithData(d1)
	t2 := widget.NewEntryWithData(d2)

	options := []string{"one-way flight", "return flight"}
	combo := widget.NewSelect(options, func(s string) {
		if s == options[0] {
			t2.Disable()
		} else {
			t2.Enable()
		}
	})

	button := widget.NewButton("Book", func() {})

	w.SetContent(container.NewVBox(combo, t1, t2, button))
	w.ShowAndRun()
}
```

 - **OS:** Fedora Workstation Linux
 - **Version:** 33 (kernel 5.10.10-200)
 - **Go version:** `go1.15.6 linux/amd64`
 - **Fyne version:** `2.0.0`

fpabl0:
### Description:

Fixes:
1. Allow copy content from a disabled widget.Entry using the keyboard.
2. Hide text selection on a disabled widget.Entry when it loses focus. 
3. Do not show validation state if widget.Entry is disabled.
4. If CursorColumn is equal to selectColumn and CursorRow is equal selectRow in DragEnd event, reset selection (with [2] fixes #1880).
5. Reset selectKeyDown variable when widget.Entry loses its focus (fixes #1787).

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
Lots of great fixes, thanks.

I'm not sure about this one:
> Allow copy content from a disabled widget.Entry using the keyboard.

As that requires we move disabled widgets back into the focus cycle... This seems like an anti-pattern as you should not be able to focus things that cannot be interacted - should you?

fpabl0:
> This seems like an anti-pattern as you should not be able to focus things that cannot be interacted - should you?

I agree with you, however widget.Entry is a special because it has the disabled state mixed with readonly state. Disabled entry can be interacted (text selection), and if we don't put disabled Entry into the focus cycle, text selection can't be erased on it (one of the current problems).

andydotxyz:
> text selection can't be erased on it (one of the current problems).

This seems like a tap handler issue not a keyboard one. Tapping should clear selection.

fpabl0:
> This seems like a tap handler issue not a keyboard one. Tapping should clear selection.

Tapping indeed clear selection, but it only happens if you tap on the disabled entry, if you tap outside of it, you can't clear the selection (current issue). As far as I can see, the only way to clear it, is when Entry loses focus, that is how it works when entry is enabled.

andydotxyz:
It's hard to be objective here as I don't really understand the difference between a ReadOnly / Disabled Entry and a Label widget.

IMHO it's not terrible that you have to tap the entry to deselect when disabled, as you have to make the selection through taps as well...

andydotxyz:
I don't think that the difference between read-only and disabled are meaningful.
An entry can either be edited (hence get focus and manage keystrokes) or it cannot (and so should not be in the tab-focus-list).

> > IMHO it's not terrible that you have to tap the entry to deselect when disabled, as you have to make the selection through taps as well...
>
> I don't follow that, sorry. 

Apart from using the mouse/touch gestures, how do you select text for a disabled entry?
My understanding is that you can't - so removing selection through tap is consistent.

fpabl0:
> An entry can either be edited (hence get focus and manage keystrokes) or it cannot (and so should not be in the tab-focus-list).

I agree with you. This PR does not change the tab-focus-list. It only uses the focus to be able to copy the content and unselect the text when you tap outside, all other characteristics remain the same.

> Apart from using the mouse/touch gestures, how do you select text for a disabled entry?
My understanding is that you can't - so removing selection through tap is consistent.

Hmm, maybe I don't write it well, I mean this is not good:

https://user-images.githubusercontent.com/12239342/110495581-a5497e80-80c2-11eb-8706-dd48d4d9b91a.mp4

A good example of how it should work, is this github conversation, every comment block is readonly and you can select the text, copy it with the keyboard and unselect it when tapping outside. I think that is the behavior final users would expect, don't they?

andydotxyz:
> I agree with you. This PR does not change the tab-focus-list. It only uses the focus to be able to copy the content and unselect the text when you tap outside, all other characteristics remain the same.

I am confused then - how does the top line "Allow copy content from a disabled widget.Entry using the keyboard" work if you are not allowing the entry to be part of the focus list? keystrokes are only sent to the currently focused element.

> A good example of how it should work, is this github conversation, every comment block is readonly

I disagree - the comment blocks are simple HTML text. On a web page all text is selectable - unlike in an application. I would consider these labels, not read-only entries. Web pages are documents transitioned into an application, whereas fyne apps are apps not documents. I don't think that it's a direct comparison.

fpabl0:
> I am confused then - how does the top line "Allow copy content from a disabled widget.Entry using the keyboard" work if you are not allowing the entry to be part of the focus list? keystrokes are only sent to the currently focused element.

Because the function that allow us to go to the next focusable widget, check again if it is disabled or not:
https://github.com/fyne-io/fyne/blob/ec5bf01f65040907a0e60646fa0f389f34584f34/internal/app/focus_manager.go#L117-L125

> I disagree - the comment blocks are simple HTML text. On a web page all text is selectable - unlike in an application.

Oh yes, you are right, sorry for the confusion (I hadn't realized the difference before). After testing some native apps, disabled inputs can't even be selected (I mean the text). However, I wonder if we should go so rigid in that direction. 

Considering that most of the modern apps are web based (VSCode, Postman, Slack, Discord, MicrosoftTeams, MongoDB Compass...) and final users are familiarized with that kind of applications, don't you think that should be good to adopt part of this behavior to avoid final users get confused?

MongoDB Compass example:

https://user-images.githubusercontent.com/12239342/110653695-7d721d80-818b-11eb-8381-1a2712b7c4a3.mp4

andydotxyz:
So to clarify what I think I am beginning to understand:
You're updating this so that disabled items can be focused if tapped, but they cannot be by tabbing in to?

And, if that's correct, then by extension selecting inside a disabled entry will de-focus whatever was focused before?

andydotxyz:
Thanks for the clarification. I'd like to get @toaster opinion on this as he knows focus really well.
What you describe sounds like a reasonable trade-off, but somehow I feel uneasy. Maybe it's checking for `SelectedText()`, or maybe it's because this will enable developers to focus items that were previously not possible.

fpabl0:
> Thanks for the clarification.

You're welcome.

>  I'd like to get @toaster opinion on this as he knows focus really well.

Great! :)

> What you describe sounds like a reasonable trade-off, but somehow I feel uneasy.

Yes, that's right, I have the same feeling at code level, but at final user level I am satisfied 😅. 

>  Maybe it's checking for SelectedText(), or maybe it's because this will enable developers to focus items that were previously not possible.

Initially I tried checking directly for *widget.Entry, but it results in a cyclic import. Then I thought that maybe checking for SelectedText() should be enough, because if a widget has this method it would probably need the same behavior.

andydotxyz:
I have added this to today's agenda to chat it through

andydotxyz (approved):
Ok I think we’re good. Just needs develop pulled in for the new tests to pass

PR #2071, opened 2021-03-08
When window init fails we should quit the app

### LINKED ISSUE #1593: Driver errors don't crash the program or tell the developer
I updated the following Fedora 33 Workstation packages:

```
avahi
cmake
environment-modules
flatpak
flatpak-libs
gjs
glib2
gnome-shell
libhandy1
linux-firmware-whence
micro
mingw32-libxml2
mingw64-libxml2
mutter
pam
steam
systemtap-sdt-devel
totem
```

And now my `fyne` application gives this error, and shows no windows:

```
2020/11/20 18:54:05 Fyne error:  window creation error
2020/11/20 18:54:05   Cause: VersionUnavailable: GLX: Failed to create context: BadValue (integer parameter out of range for operation)
2020/11/20 18:54:05   At: /home/damon/go/pkg/mod/fyne.io/fyne@v1.4.1/internal/driver/glfw/window.go:1214
```

`glxinfo` confirmed this was an issue outside of `fyne` too.

### So why the bug report?

My application didn't crash, but no windows were shown either.

My code was not made privy to the error itself, as it was thrown by driver code.

So, the user experience was a process that "does nothing, but is still running."

andydotxyz:
IIf the driver is running already then quit the app.
If not we can exit before it starts.

Fixes #1593 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/driver.go
+++ b/internal/driver/glfw/driver.go
@@ -3,6 +3,7 @@
 package glfw
 
 import (
+	"os"
 	"runtime"
 	"strconv"
 	"strings"
@@ -111,6 +112,16 @@ func (d *gLDriver) windowList() []fyne.Window {
 	d.windowLock.RLock()
 	defer d.windowLock.RUnlock()
 	return d.windows
+}
+
+func (d *gLDriver) initFailed(msg string, err error) {
+	fyne.LogError(msg, err)
+
+	if running() {
+		d.Quit()
+	} else {
+		os.Exit(1)
+	}
 }
 
 func goroutineID() int {
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -57,6 +57,7 @@ type window struct {
 	cursor       desktop.Cursor
 	customCursor *glfw.Cursor
 	canvas       *glCanvas
+	driver       *gLDriver
 	title        string
 	icon         fyne.Resource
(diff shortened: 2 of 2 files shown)


internal/driver/glfw/driver.go
```diff
@@ -113,6 +114,16 @@ func (d *gLDriver) windowList() []fyne.Window {
 	return d.windows
 }
 
+func (d *gLDriver) initFailed(msg string, err error) {
+	fyne.LogError(msg, err)
+
+	if running() {
+		d.Quit()
+	} else {
+		os.Exit(1)
```
fpabl0, line 123: It would mean that if the creation of any window in the application runtime fails, the whole app will quit? Is that intended?

fpabl0: > When a window cannot be opened it basically means the graphics driver connection has failed or we are out of memory - both are pretty much irrecoverable.

That's a good point, I don't know what other things could fail. 
I was just thinking in the case that an application could try to create a new window, and it failed but not with a severe error, the application could continue working and just ignore that error and don't create the window. In that sense, the driver should quit the app if the creation of the `master` window fails.

However, if an error in window creation is always a severe error, then I think it is fine :)

fpabl0 (approved):

PR #2072, opened 2021-03-08
Faster ci

Jacalz:
### Description:

The setup of each workflow run should now be about 3x faster thanks to using https://github.com/WillAbides/setup-go-faster.
This includes a workaround until WillAbides/setup-go-faster#11 is fixed.

This also makes sure that we run tests on the latest stable Go release (Go 1.16). It has the side-effect that the CI should be even faster: "Linking is 20-25% faster than 1.15 and requires 5-15% less memory on average for linux/amd64, with larger improvements for other architectures and OSes.".

### Checklist:

- ~[] Tests included.~
- [x] Tests all pass.

### DIFF
--- a/.github/workflows/mobile_tests.yml
+++ b/.github/workflows/mobile_tests.yml
@@ -11,7 +11,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - uses: actions/setup-go@v2
+    - uses: WillAbides/setup-go-faster@v1.5.0
       with:
         go-version: ${{ matrix.go-version }}
 
--- a/.github/workflows/mobile_tests.yml
+++ b/.github/workflows/mobile_tests.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


andydotxyz:
It seems like WillAbides/setup-go-faster#11 might be addressed, should we get that fix in here so the workaround is not needed?

Jacalz:
> It seems like [WillAbides/setup-go-faster#11](https://github.com/WillAbides/setup-go-faster/issues/11) might be addressed, should we get that fix in here so the workaround is not needed?

From what I can tell, it isn't fixed. The only commits referenced there are his fork of `Jacalz/fyne` where he made the workaround and my commits from this PR.

andydotxyz (approved):

Jacalz:
It looks like the required task are specific to the versions and thus it wants the 1.15 versions to complete. Probably needs to be changed int he repo settings before this can be merged.

andydotxyz:
Chicken and egg... I will juggle the config and land this

andydotxyz:
That's in and the new set of required all passing

PR #2073, opened 2021-03-08
Fix multiline entry cursor position when pasting content (with text wrap)

fpabl0:
### Description:

Fixes the multiline entry cursor position when pasting content (with text wrap).

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Great catch, I love it when the fix is to remove code ;)

PR #2078, opened 2021-03-11
Small improvements on widget.BaseWidget

fpabl0:
### Description:

Small improvements on widget.BaseWidget (mainly just removing some defers).

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


fpabl0:
Based on these small changes, should we remove `getImpl` function in favor of `super()` (or viceversa)? Because now they are exactly the same.

I have finally understood the importance of `super()` for extended widgets. And now I have a doubt, by making `super()` a private method, isn't it will limit the custom widgets to be fully extended? (taking the example of the widget.Entry and the passwordRevealer)

PR #2082, opened 2021-03-11
Small improvements BaseWidget

fpabl0:
### Description:

Small improvements in BaseWidget.

### Checklist:

~~- [ ] Tests included.~~
- [x] Tests all pass.


fpabl0:
Based on these small changes, should we remove getImpl function in favor of super() (or viceversa)? Because now they are exactly the same.

I have finally understood the importance of super() for extended widgets. And now I have a doubt, by making super() a private method, isn't it will limit the custom widgets to be fully extended? (taking the example of the widget.Entry and the passwordRevealer)

widget/widget.go
```diff
@@ -160,23 +158,15 @@ func (w *BaseWidget) setFieldsAndRefresh(f func()) {
 
 	impl := w.getImpl()
 	if impl == nil {
-		w.Refresh()
```
andydotxyz, line 163: I am unsure about this removal.
It means that a widget with no base defined will not refresh...

widget/widget.go
```diff
@@ -160,23 +158,15 @@ func (w *BaseWidget) setFieldsAndRefresh(f func()) {
...
 	}
+	impl.Refresh()
 }
 
 // super will return the actual object that this represents.
 // If extended then this is the extending widget, otherwise it is self.
 func (w *BaseWidget) super() fyne.Widget {
-	impl := w.getImpl()
-
-	if impl == nil {
```
andydotxyz, line 174: Removing this nil check breaks the method signature. The nil return from `getImpl` is ignored.
This means "otherwise it is self" is no longer true.

andydotxyz (changes requested):
Mostly agreed but I don't follow the idea that `getImpl` and `super` are the same - as noted inline.

andydotxyz:
> I have finally understood the importance of super() for extended widgets. And now I have a doubt, by making super() a private method, isn't it will limit the custom widgets to be fully extended?

The `super` function works for anything extending `BaseWidget`, you don't have to be in this package because the provided base handles it.

If you want to write a completely bespoke widget that avoids this helper then you will need to determine your own way to match `WidgetRenderer` with `Widget` in your  custom-extended custom-widget.
This is all part of the "how do we map renderer to widget" question that feels like it could be improved.

widget/widget.go
fpabl0: Hmm if a widget has no base defined, then it won't use this, will it? Also `w.Refresh` would not do anything if `impl` is nil, as it is defined as:
```go
// Refresh causes this widget to be redrawn in it's current state
func (w *BaseWidget) Refresh() {
	impl := w.getImpl()
	if impl == nil {
		return
	}
	render := cache.Renderer(impl)
	render.Refresh()
}
```

widget/widget.go
fpabl0: > This means "otherwise it is self" is no longer true.

You are right, but I think, that part of the description should be removed instead. The reason is that `BaseWidget` is not a widget (it does not have CreateRenderer method), so it would not be able to return itself as it will always panics with interface conversion error.

fpabl0:
> The super function works for anything extending BaseWidget, you don't have to be in this package because the provided base handles it.

That's true, maybe it should be fixed internally to always use the `super()` widget and avoid to have this case:
https://github.com/fyne-io/fyne/blob/c18fd156da5d1b9208414a3396320a07c6c9415b/widget/entry_password.go#L45-L50

> This is all part of the "how do we map renderer to widget" question that feels like it could be improved.

Totally agreed :)

andydotxyz:
> That's true, maybe it should be fixed internally to always use the super() widget and avoid to have this case:

Tidying up cases like that would help - and avoid bugs :)

widget/widget.go
andydotxyz: Hmm, interesting. This is another case of "I thought it worked" but looking at it now I am less sure.

widget/widget.go
andydotxyz: Yes, the combination of the two probably makes this unneeded, sorry.

fpabl0: Indeed, I think it should be replaced to just w.Refresh() to avoid repeating code, what do you think?

widget/widget.go
fpabl0: I see, so should I mark this as resolved?

fpabl0: Also with this change `super()` and `getImpl()` are exactly the same, so I think we should remove one of them, should I remove `getImpl()`?

fpabl0:
> Tidying up cases like that would help - and avoid bugs :)

I have an idea. Maybe we could move BaseWidget definition to internal cache package (or move cache renderer and BaseWidget to internal widget package). Then to expose that type to users we can do (inside widget public package):
```go
type BaseWidget = cache.BaseWidget
// or
type BaseWidget = widget.BaseWidget
```
Then inside internal cache or widget package we could define a `Super()` function to use it along all the internal implementation:

```go
type extendedWidget interface {
    ExtendBaseWidget(wid fyne.Widget)
    super() fyne.Widget
}
func Super(wid fyne.Widget) fyne.Widget {
   if ext, ok := wid.(extendedWidget); ok {
       return ext.super()
   }
   return wid
}
```
In that way, we could use `Super(...)` in other places like:
```go
cache.Super(wid)
// or
widget.Super(wid)
```
And the user will not see this details as they are inside the internal package 😄

fpabl0:
Hmm however it will hide all the private properties and functions helpers 🤔. Another idea could be have a type in the internal package called something like "ExtendedWidget" or other better name:
```go
type ExtendedWidget struct {
	impl     fyne.Widget
	implLock sync.RWMutex
}
func (w *ExtendedWidget) ExtendBaseWidget(wid fyne.Widget) { ... }
func (w *ExtendedWidget) super() fyne.Widget { ... }
```
And this could be embedded in BaseWidget:
```go
type BaseWidget struct {
    cache.ExtendedWidget
    ...
}
```
That could work :), but maybe I am missing something 😅. Nevertheless, I create a simple test of how it should work, you can see the file `widget/widget_extended_test.go` in:
https://github.com/fpabl0/fyne/tree/extended-idea

andydotxyz:
> I have an idea. Maybe we could move BaseWidget definition to internal cache package (or move cache renderer and BaseWidget to internal widget package)

>  Another idea could be a type in the internal package called something like "ExtendedWidget" or other better name:

There are a number of ways that this code could be changed - but are we on the same page about the problem to be solved?
I think this discussion should be separated from the currrent PR as it is certainly not a "small improvement".

fpabl0:
> I think this discussion should be separated from the currrent PR as it is certainly not a "small improvement".

@andydotxyz Yes, of course, I was just thinking a loud. I never thought to add this to the current PR. I just didn't want to lose the idea, sorry to write it here.

widget/widget.go
fpabl0: @andydotxyz any thoughts about this?

widget/widget.go
andydotxyz: What happens if a widget has overridden `Refresh()`?

widget/widget.go
andydotxyz: I would lose `getImpl` yes, as that was added more recently.

widget/widget.go
fpabl0: Oh, I did not take in account that, then it should be simplified to:
```go
	impl := w.getImpl()
	if impl == nil {
		return
	}
	impl.Refresh()
```

andydotxyz: Yes I think that would probably do it.

widget/widget.go
fpabl0: done in 31dcbd5

widget/widget.go
fpabl0: done in 31dcbd5

andydotxyz (approved):
Nice, thanks

andydotxyz:
Thanks for this.
Once it is landed I will try to reconcile the internal and external BaseWidgets so we get more consistency again.

fpabl0:
Maybe it is better to move this to internal widget package, and then make it public using:
```go
package widget // (the public one)

import (
     "fyne.io/fyne/v2/internal/widget"
)

type BaseWidget = widget.BaseWidget
```

In that way, IMHO we don't need to repeat the implementation in two places.

PR #2085, opened 2021-03-12
Reduce cyclo complexity to a better starting point of 30

andydotxyz:
From here we can improve further, but this is a better starting point

### Checklist:

- [ ] Tests included. <- just refactor in this PR
- [x] Tests all pass.


andydotxyz:
No reviews for > 7 days, auto-accepting

PR #2096, opened 2021-03-17
Mobile (Android): get filename of content:// URI

howeyc:
### Description:
When we open content on android, sometimes we don't get nice "file://path/to/file/here.txt" URI, we get "content://34234" instead.
In cases where we want to show the name of the file, this is not ideal. This change overrides the Name() and Ext() functions of the URI interface to instead get the actual filename represented by the content:// URI.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/internal/driver/gomobile/file.go
+++ b/internal/driver/gomobile/file.go
@@ -9,12 +9,6 @@ import (
 	"fyne.io/fyne/v2/storage"
 )
 
-// Override Name on android for content://
-type mobileURI struct {
-	systemURI string
-	fyne.URI
-}
-
 type fileOpen struct {
 	io.ReadCloser
 	uri  fyne.URI
--- a/internal/driver/gomobile/file.go
+++ b/internal/driver/gomobile/file.go
@@ -9,6 +9,12 @@ import (
 	"fyne.io/fyne/v2/storage"
 )
 
+// Override Name on android for content://
+type mobileURI struct {
+	systemURI string
+	fyne.URI
+}
+
 type fileOpen struct {
 	io.ReadCloser
 	uri  fyne.URI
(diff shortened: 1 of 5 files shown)


internal/driver/gomobile/file_android.go
```diff
@@ -12,18 +12,34 @@ char* readStream(uintptr_t jni_env, uintptr_t ctx, void* stream, int len, int* t
...
 	"io"
 	"os"
+	"path/filepath"
 	"unsafe"
 
 	"fyne.io/fyne/v2/storage/repository"
 	"github.com/fyne-io/mobile/app"
 )
 
+func (m mobileURI) Name() string {
```
andydotxyz, line 29: We try to keep things ordered alphabetically (type, then method) - more info at https://github.com/fyne-io/fyne/wiki/Contributing.
So I think mobileURI should come after javaStream.
I wondered also if there was a reason for not using a pointer receiver here, when you create the type as a pointer?

internal/driver/gomobile/file.go
```diff
@@ -56,7 +62,10 @@ func ShowFileOpenPicker(callback func(fyne.URIReadCloser, error), filter storage
 				callback(nil, nil)
 				return
 			}
-			f, err := fileReaderForURI(storage.NewURI(uri))
+			f, err := fileReaderForURI(&mobileURI{
```
andydotxyz, line 65: Only Android provides this extension, but you're returning it for iOS as well - is that intentional? 
Maybe this init should change - or maybe the code should move to `repository.go` out from the android code?

internal/driver/gomobile/file_android.go
andydotxyz: Or (from note above) perhaps it goes at the end of `repository.go`?
It depends on whether iOS should use this functionality or not I think.

andydotxyz (changes requested):
Nice enhancement, just some thoughts around location of the code and/or whether this should impact iOS.

internal/driver/gomobile/file.go
howeyc: I figured if iOS needs to use it, it can Override Name() like I'm doing, otherwise just does nothing and embedded URI is used.

howeyc:
The latest changes I made are probably better, this way the uri stuff is android only.

andydotxyz (approved):
Nice, that looks cleaner, thanks.

andydotxyz:
Hmm, the static analysis is now failing - due to a timeout.
Not sure if this can have triggered it, or if the service is down?

andydotxyz (approved):

Issue #2069, opened 2021-03-08
Memory leak in widget renderer and texture caches


fpabl0:
### Describe the bug:

Fyne widgets use internally a renderer cache to improve performance, however when the widget goes out of the scope, Fyne does not have any way to free that memory from the cache. 

https://github.com/fyne-io/fyne/blob/8fa5af426bbfa7349d2b0c57ea06345b640babec/internal/cache/widget.go#L9

The same thing occurs with canvas primitives and the texture cache:

https://github.com/fyne-io/fyne/blob/8fa5af426bbfa7349d2b0c57ea06345b640babec/internal/painter/gl/gl_common.go#L17

This memory leak occurs commonly when user replaces objects in containers:

```go
c.Objects[2] = widget.NewLabel("Hello1")
c.Objects[3] = canvas.NewText("Text1", color.Black)
...
c.Objects[2] = widget.NewLabel("Hello2") //  🔴 "Hello1" renderer is kept in renderer cache and have no way to be freed
c.Objects[3] = canvas.NewText("Text2", color.Black) // 🔴 "Text1" is kept in texture cache and have no way to be freed
```
This could be the root cause of #348.
This is also an extension from #735.

### To Reproduce:
This memory can be seen in `fyne_demo`, when the tutorials contents are set: 

https://github.com/fyne-io/fyne/blob/8fa5af426bbfa7349d2b0c57ea06345b640babec/cmd/fyne_demo/main.go#L111-L112

1. Run `fyne_demo`.
2. Change tutorial views.
3. Use the Activity Monitor and see how the memory used is increased every time you change the view.

### Device (this issue is platform independent):
 - **OS:** MacOS
 - **Version:** 10.15.4 Catalina
 - **Go version:** 1.16
 - **Fyne version:** `develop` @ 8fa5af4

fpabl0:
This introduce also a severe memory leak when selecting text in `widget.Entry`, because selection rectangles are regenerated:

https://github.com/fyne-io/fyne/blob/8fa5af426bbfa7349d2b0c57ea06345b640babec/widget/entry.go#L1464-L1467

And the entry's objects are generated dynamically:

https://github.com/fyne-io/fyne/blob/8fa5af426bbfa7349d2b0c57ea06345b640babec/widget/entry.go#L1358-L1363

if you keep selecting and deselecting text, you will see that the memory used will increase.

fpabl0:
Same memory leak issue has the widget.List:

[...]

Issue #735, opened 2020-03-09
Memory leaks in renderers and canvases cache maps


pymq:
### Describe the bug:
Once created `CanvasObject` and its renderer are never freed. Two caches are stored here:
`internal/driver/glfw/driver.go` - canvases
`internal/cache/widget.go` - renderers
As I can see from code, they are only cleared on window close and only objects that are still in use by that window.
The current workaround is to reuse all widgets but it's hard to do in dynamic interfaces.

### To Reproduce:
Continuously create new widgets, for instance in a long running applications.

### Example code:
```go
package main

import (
	"time"

	"fyne.io/fyne/app"
	"fyne.io/fyne/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("test")

	setContent := func() {
		box := widget.NewVBox(
			widget.NewEntry(), widget.NewEntry(),
			widget.NewButton("submit", nil),
		)
		w.SetContent(widget.NewGroup("group", box))
	}

	setContent()
	go func() {
		for range time.NewTicker(100 * time.Millisecond).C {
			setContent()
		}
	}()

	w.CenterOnScreen()
	w.ShowAndRun()
}
```

### Device:
 - Go version: 1.13.8
 - Fyne version: develop, 84bca6e4cfaa

fpabl0:
Fixed on `develop`.

PR #2101, opened 2021-03-19
Fix memory leaks in some cache maps

### LINKED ISSUE #348: Memory leak in unused cached images
Just a video of me going through the fyne demo with task manager on the side , it seems to go well until the graphics tab as thats where memory usage makes massive spikes and is then never freed. 

https://youtu.be/b4Yh9dGj9e4?t=70

go1.12.4 windows/amd64

fpabl0:
### Description:

Potentially fix #348, #735, #2069.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.
- [x] ~~Call `cache.Clean` on mobile too.~~ Update mobile driver, canvas and windows to have the same desktop changes.
- [x] Find a better way to clean textures cache.
- [x] Move svg_cache to cache package.
- [x] Decision about minSize cache on mobile driver (needs refactor).
- [x] Remove the mutex (added in this PR) for texture cache as it is not needed (texture cache is only manipulated in one goroutine).
- [x] Solve texture cache cleaning for multiple windows.


fpabl0:
Oh I forgot that `ticker.Reset` does not exist in Go 1.12, I will update later.

Jacalz:
> Oh I forgot that `ticker.Reset` does not exist in Go 1.12, I will update later.

Just add a comment about it so that we can change it when we update the minimum Go version 🙂

fpabl0:
@andydotxyz @Jacalz it is not finished, but I would like to know your thoughts/suggestions about the current changes :)

Basically, in the latest changes I have removed the go routine used to clean the cache in favor of using the paint events as the ticker. 

Some of my doubts are:
1. Should we destroy expired renderers independently of a canvas refresh action? (this is the current behavior with this PR)
2. Texture cache map does not need a mutex because its functions are always called from the same go routine, however as you can see now the texture cache map saves a reference for the canvas which the primitive belongs to, by doing it we could use that map (only for primitives) inside `CanvasForObject` instead of the canvases cache map, this action would allow us to reduce the canvases cache map (because it would not contain primitives anymore) and delete more quickly them when they are destroyed (if we use canvases cache map we would need to wait them for expiration), however this implies to add a mutex to the textures cache map (`CanvasForObject` can be called from anywhere), what do you think? Should we keep all objects in canvases cache or just the ones that are not primitives?

andydotxyz:
1. This is a good question. I think it was put that way for thinking it was faster. But other approaches probably OK :) 
2. I'm not sure about the underlying thought here. Many of our primitives don't translate to graphics primitives (yet?). We need to rasterise some elements (like rectangles with borders, and lines) so that's why the cache exists. Does that impact your thoughts?

fpabl0:
> This is a good question. I think it was put that way for thinking it was faster. But other approaches probably OK :)

Indeed it is faster. Current PR use both of them (destroy on refresh and destroy when expired). However destroy on refresh will not destroy objects that are out of scope (because they no longer belonged to the object tree).

> I'm not sure about the underlying thought here. Many of our primitives don't translate to graphics primitives (yet?). We need to rasterise some elements (like rectangles with borders, and lines) so that's why the cache exists. Does that impact your thoughts?

Hmm for primitives I mean the objects created with `canvas.New...` (Rectangle, Image, Text and others) they all are textures for `gl` and so they are saved in the textures cache map. Basically this is the functionality of every cache map:
- rendereres map => widgets associated with their renderers
- textures map => `canvas.New...` objects associated with their texture.
- canvases map => all objects (widgets, containers and `canvas.New...` objects)

Textures are created and destroyed really fast. The destroy action will delete immediately the texture from the textures cache map but not from the canvases map (we would need to wait them to expire to release that memory).

So if we change the map data to:
- rendereres map => widgets associated with their renderers
- textures map => `canvas.New...` objects associated with their texture **and canvas.**
- canvases map => all objects (**only** widgets and containers)

So canvases map will not longer hold texture objects, by doing so, when we just need to delete the texture from the textures cache map.

Then `CanvasForObject` should only check the type of the object. If it is a widget or container, it will use the `canvases map`, otherwise it will use the `textures map` to get its canvas.

andydotxyz:
>  However destroy on refresh will not destroy objects that are out of scope (because they no longer belonged to the object tree).

Good point!

> So if we change the map data to:
>
>   * rendereres map => widgets associated with their renderers
>   * textures map => canvas.New... objects associated with their texture and canvas.
>   * canvases map => all objects (only widgets and containers)
>
> So canvases map will not longer hold texture objects, and by doing so, we would just need to delete the texture from the textures cache map.

I don't think this is wise from what I have read - the texture cache refers to the *current rendering* of an object. If this object is changed and `Refresh()` called then the object should remain in the canvases map, but it's texture needs to be invalidated.

fpabl0:
> I don't think this is wise from what I have read - the texture cache refers to the current rendering of an object. If this object is changed and Refresh() called then the object should remain in the canvases map, but it's texture needs to be invalidated.

On yes, you are right, I missed that fact. Thanks :)

fpabl0:
@andydotxyz I have moved the svg_cache to the cache package and removed the 10 ms delay in mobile driver. Could you a do quick review in `internal/driver/gomobile/driver.go` if you are agreed with the changes? I think that change probably solve the theme change issue on mobile.

fpabl0:
@andydotxyz Ok this is almost done. Desktop side is ready (maybe better naming for the added cache functions is needed), however on mobile side I think a refactorization is needed specially to rid off the minSizeCache and use renderCacheNode-canvasesCache solution used in desktop (or viceversa, I am not sure what is the better solution, maybe was it discussed before?)

So should I add TODO comments and leave this PR as it is now? Or should this PR wait until mobile driver is refactored (mainly just add the renderCacheNode and queueEvent method)?

andydotxyz:
> Or should this PR wait until mobile driver is refactored (mainly just add the renderCacheNode and queueEvent method)?

It would make it a lot easier to apply this in a shared-code manner if the mobile driver had the equivalent of https://github.com/fyne-io/fyne/commit/7461574f15a036bbb156f6fe53839f3db968aac5 applied.

I am nervous that the drivers are moving further apart instead of coming together, so let's not make the differences and performance gap greater yet.

fpabl0:
> It would make it a lot easier to apply this in a shared-code manner if the mobile driver had the equivalent of 7461574 applied.

@andydotxyz Totally agreed. So should I open another PR as the initial step to try to share code between the two drivers? Then this PR should wait until that PR is landed. Is that ok?

andydotxyz:
> @andydotxyz Totally agreed. So should I open another PR as the initial step to try sharing code between the two drivers? Then this PR should wait until that PR is landed. Is that ok?

If you're OK doing that I think it sounds like a great idea

internal/cache/base.go
```diff
@@ -0,0 +1,206 @@
...
+	select {
+	case <-canvasRefreshCh:
+		canvasRefresh = true
+	default:
+		if now.Sub(lastClean) < cleanTaskInterval {
+			return
+		}
+	}
+	destroyExpiredRenderers(now)
+	destroyExpiredSvgs(now)
```
fpabl0, line 44: Should expired renderers and svgs be removed with no synchronization with canvas.Refresh? Basically occurs two things:
1. In the way it is now, after some time (with no user interaction) widgets renderers will be removed so when user starts interaction again he would look like a blink because all the needed renderers will be created at that time.
2. If we remove expired renderers only when a canvas refresh is executed, then user would not see anything strange, but if no user interaction, the expired renderers will not be removed until a canvas.Refresh.

So both of them have disadvantages and I am not sure which choose. Any suggestions?

internal/cache/widget.go
```diff
@@ -24,29 +25,46 @@ func Renderer(wid fyne.Widget) fyne.WidgetRenderer {
...
 // DestroyRenderer frees a render implementation for a widget.
 // This is typically for internal use only.
 func DestroyRenderer(wid fyne.Widget) {
 	Renderer(wid).Destroy()
 
-	renderers.Delete(wid)
+	renderersLock.Lock()
+	delete(renderers, wid)
+	renderersLock.Unlock()
 }
```
fpabl0, line 56: Should this function also deletes child renderers? 🤔 In that way, if a big widget (which renderer has other widgets) is deleted, all its children will do it too at the same time. Am I missing something?

internal/cache/base.go
fpabl0: Or we can delete renderers immediately when we detect they are not in the objects tree anymore, what do you think?

internal/cache/widget.go
andydotxyz: That is an interesting point. It should either be done here or in the driver where they call `DestroyRenderer` I think.

internal/cache/base.go
andydotxyz: Caching is a careful balance. We need to remove things that are not needed - and quickly - but keep things still used that may not have actually been refreshed recently.

If there is a way to do this without a timer that might be interesting, but if we need a timer then I don't think the problem above can be avoided.

internal/cache/widget.go
fpabl0: Maybe here in `DestroyRenderer()` is the appropriate and easiest way for adding it.

internal/cache/base.go
fpabl0: Hmm are you talking about option 1 or 2?

internal/cache/widget.go
andydotxyz: Yes, probably true

internal/cache/base.go
andydotxyz: Neither, I was commenting on how there is no right answer.
I guess 2 is a slightly better compromise as it is unlikely that the cache would grow without items being refreshed :).

fpabl0: Agree :)

fpabl0: done in 2a3abfb

internal/cache/widget.go
fpabl0: After thinking carefully about this, maybe this isn't necessary as cache.DestroyRenderer is only called when the window is closed, but in that part currently we are cleaning all the objects from the canvas using `cache.ForceCleanFor`. So perhaps this can be revisited later if we find this is really needed

andydotxyz: That's possible too. We don't make much use of `Destroy` yet. I think this will change in 2.1 if we can get the render / extend smoothed out.

fpabl0: Great! So I will mark this as resolved.

Jacalz:
Is this ready for a re-review? @fpabl0

fpabl0:
Yes, it is :)

Jacalz (changes requested):
This is looking really good. Just have a few suggestions for changes :)

internal/cache/base.go
```diff
@@ -0,0 +1,207 @@
...
+	renderersLock.Unlock()
+}
+
+// NotifyCanvasRefresh notifies to the caches that a canvas refresh was triggered.
+func NotifyCanvasRefresh() {
+	select {
+	case canvasRefreshCh <- struct{}{}:
+	default:
+		return
+	}
```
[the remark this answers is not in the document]
Jacalz: Never mind. I guess the intent is to be non-blocking if the channel already contains something? You might be able to ignore the comment for this ;)

DiskGetor:
]


---Original---
From: ***@***.***&gt;
Date: Thu, Jun 3, 2021 02:40 AM
To: ***@***.***&gt;;
Cc: ***@***.***&gt;;
Subject: Re: [fyne-io/fyne] Fix memory leaks in some cache maps (#2101)




 
Is this ready for a re-review? @fpabl0
 
&mdash;

internal/cache/base.go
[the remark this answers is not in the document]
fpabl0: > I guess the intent is to be non-blocking if the channel already contains something?

Yes, that is the purpose :)

internal/cache/base.go
```diff
@@ -0,0 +1,207 @@
...
+import (
+	"os"
+	"sync"
+	"time"
+
+	"fyne.io/fyne/v2"
+)
+
+var (
+	cacheDuration     = 1 * time.Minute
```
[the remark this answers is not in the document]
fpabl0: I can remove it, but not sure which one is better 🤔

internal/cache/base.go
```diff
@@ -0,0 +1,207 @@
...
+	cleanTaskInterval = cacheDuration / 2
+
+	canvasRefreshCh = make(chan struct{}, 1)
```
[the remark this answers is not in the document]
fpabl0: It is necessary, because we need to guarantee that we can send the event even if there is no receiver channel listening at that moment.

internal/cache/base.go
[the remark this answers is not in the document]
Jacalz: No worries. I was just wondering wether or not this was a leftover from a longer timer or for readability. It certainly does improve readability, so I’m fine with keeping it.

internal/cache/base.go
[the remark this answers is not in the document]
Jacalz: My bad. It is zero that is an unbuffered channel, not one.

Jacalz (dismissed):
Thanks for this. It is certainly a good improvement. I checked the memory usage before and after and and it certainly does seem to be much better at releasing memory. I still see the memory slowly climbing without me doing anything inside the application, but I guess that's another issue :)

fpabl0:
>  I still see the memory slowly climbing without me doing anything inside the application

Hmm the memory (after some time) should be constant (without increasing) 🤔. Could you share the code you are using to test this?

Jacalz:
> >  I still see the memory slowly climbing without me doing anything inside the application
> 
> 
> 
> Hmm the memory (after some time) should be constant (without increasing) 🤔. Could you share the code you are using to test this? 

I was just using fyne_demo inside the branch that this PR is based on (on Linux). I think it started at 22Mb and then it slowly climbed to 29Mb. I don't know if this was before or after the time it should have stabilised in.

internal/cache/base.go
```diff
@@ -0,0 +1,207 @@
...
+		// canvases cache should be invalidated only on canvas refresh, otherwise there wouldn't
+		// be a way to recover them later
+		destroyExpiredCanvases(now)
+	}
+	lastClean = time.Now()
+}
+
+// ForceCleanFor forces a complete remove of all the objects that belong to the specified
+// canvas. Usually used to free all objects from a closing windows.
+func ForceCleanFor(canvas fyne.Canvas) {
```
andydotxyz, line 56: Is `Force` needed? perhaps `cache.CleanCanvas` would work?

internal/cache/base.go
```diff
@@ -0,0 +1,207 @@
...
+		if !ok {
+			continue
+		}
+		winfo.renderer.Destroy()
+		delete(renderers, wid)
+	}
+	renderersLock.Unlock()
+}
+
+// NotifyCanvasRefresh notifies to the caches that a canvas refresh was triggered.
```
andydotxyz, line 96: This isn't really a notification - would `Request` work better than `Notify`?

internal/cache/base.go
```diff
@@ -0,0 +1,207 @@
...
+func NotifyCanvasRefresh() {
+	select {
+	case canvasRefreshCh <- struct{}{}:
+	default:
+		return
+	}
+}
+
+// ===============================================================
```
andydotxyz, line 105: We don't use big separator comments

internal/cache/base.go
```diff
@@ -0,0 +1,207 @@
...
+	if len(expiredSvgs) > 0 {
+		svgLock.Lock()
+		for _, exp := range expiredSvgs {
+			delete(svgs, exp)
+		}
+		svgLock.Unlock()
+	}
+}
+
+// ===============================================================
```
andydotxyz, line 170: As above, these comments don't add meaning, if it's required for readability maybe we need a new file.

internal/driver/glfw/window.go
```diff
@@ -430,17 +430,29 @@ func (w *window) Close() {
...
+		w.viewLock.Lock()
+		w.closing = true
+		w.viewLock.Unlock()
+		w.viewport.SetShouldClose(true)
+		cache.RangeTexturesFor(w.canvas, func(obj fyne.CanvasObject) {
+			w.canvas.Painter().Free(obj)
+		})
+	})
 
+	// destroy current renderers
```
andydotxyz, line 444: I'm not sure this (and other comments) are required.
Comments generally show that the code could be clearer - so we refactor.
The in-method comments should be kept for those lines that will need an explanation.

internal/driver/glfw/window.go
```diff
@@ -430,17 +430,29 @@ func (w *window) Close() {
...
 	w.canvas.WalkTrees(nil, func(node *common.RenderCacheNode) {
 		if wid, ok := node.Obj().(fyne.Widget); ok {
 			cache.DestroyRenderer(wid)
 		}
 	})
 
+	// remove all canvas objects from the cache
+	w.QueueEvent(func() {
+		cache.ForceCleanFor(w.canvas)
```
andydotxyz, line 453: This is not an input event, why is it being queued with them?

andydotxyz (changes requested):
Great work, but it's not quite in-keeping with our code style.
You've included a lot of comments, some that are blocks to separate methods, and some that are telling what a line of code does - both seem superfluous.
If code needs a comment then it generally should be more cleanly written.

Additionally `cache/base.go` adds a lot of new code but I don't think that any tests cover it.

internal/cache/base.go
```diff
@@ -0,0 +1,207 @@
...
+
+	"fyne.io/fyne/v2"
+)
+
+var (
+	cacheDuration     = 1 * time.Minute
+	cleanTaskInterval = cacheDuration / 2
+
+	canvasRefreshCh = make(chan struct{}, 1)
+	expiredObjects  = make([]fyne.CanvasObject, 0, 50)
```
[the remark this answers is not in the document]
fpabl0: I made it global only to avoid having to recreate the slice on every clean task, in that way the same slice is reused on every task without the need to allocate memory again. Should I make it local?

internal/cache/base.go
```diff
@@ -0,0 +1,207 @@
...
+	lastClean       time.Time
+)
+
+func init() {
+	if t, err := time.ParseDuration(os.Getenv("FYNE_CACHE")); err == nil {
```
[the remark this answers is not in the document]
fpabl0: I agree with this proposed name, however `FYNE_CACHE` environment variable existed before this PR, if we rename it, isn't it going to be a breaking change?

internal/cache/base.go
fpabl0: I don't think `Request` is the ideal replacement here because this function is not requesting a canvas refresh, it is just a way to notify to the cache that a `canvas.Refresh` was requested.

internal/driver/glfw/window.go
fpabl0: Got it, sorry for them, I forgot to remove them. I will make the changes :)

internal/driver/glfw/window.go
fpabl0: Because we need to ensure that the objects are removed when there is no more input events that depend on them running, and a way I think we could accomplish it is adding it to the event queue

fpabl0:
> I was just using fyne_demo inside the branch that this PR is based on (on Linux). I think it started at 22Mb and then it slowly climbed to 29Mb. I don't know if this was before or after the time it should have stabilised in.

Hmm in fyne_demo, is a bit difficult to test this, because if you go to the screens that already have memory leaks, then the memory will never be stable

fpabl0:
> Additionally cache/base.go adds a lot of new code but I don't think that any tests cover it.

Yes, you are right, I will try to add them later today.

internal/driver/glfw/window.go
andydotxyz: Perhaps this is too soon then? As the window is closing we will at some point know that there are no pending events.
Maybe in the actual destroy instead of the `Close()` API call?

internal/driver/glfw/window.go
```diff
@@ -430,17 +430,29 @@ func (w *window) Close() {
...
-	w.closing = true
-	w.viewLock.Unlock()
-	w.viewport.SetShouldClose(true)
+	// set w.closing flag inside draw thread to ensure we can free textures
+	runOnDraw(w, func() {
+		w.viewLock.Lock()
+		w.closing = true
+		w.viewLock.Unlock()
+		w.viewport.SetShouldClose(true)
+		cache.RangeTexturesFor(w.canvas, func(obj fyne.CanvasObject) {
```
andydotxyz, line 439: This is probably not the right place - `Close()` is for programmatic closures, it is not called when user closes the window using the OS drawn accessories - see `closed()` for that.

internal/cache/base.go
andydotxyz: I think it is unclear here the purpose of the API. A canvas callback would exist in the canvas code, not the cache.
Also this is not passing the canvas so it's a global flag - is that the right way to indicate a canvas is dirty?
With multiple windows we will be doing a lot of unnecessary work.

If you can outline what this is doing and who calls it for what purpose perhaps we come up with a better name and/or see if this is the right way to implement that requirement.

internal/cache/base.go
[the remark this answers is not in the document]
andydotxyz: I did wonder but I could not see the removal in the PR. If it was there before lets keep it.

internal/cache/base.go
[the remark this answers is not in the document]
andydotxyz: Making it global requires locking, surely being able to avoid that outweighs any time required in allocating memory?

internal/cache/base.go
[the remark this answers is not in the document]
fpabl0: Got it! Should I put a TODO comment to rename it later?

internal/cache/base.go
fpabl0: done in 6fc2bfc

internal/cache/base.go
```diff
@@ -0,0 +1,207 @@
...
+		cacheDuration = t
+		cleanTaskInterval = cacheDuration / 2
+	}
+}
+
+// CleanTask run cache clean task, it should be called on paint events.
+func CleanTask() {
```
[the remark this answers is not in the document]
fpabl0: done in 6fc2bfc

internal/cache/base.go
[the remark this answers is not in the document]
fpabl0: It does not require locking because cache.Clean won't run asynchronously

internal/cache/base.go
fpabl0: > With multiple windows we will be doing a lot of unnecessary work.

cache.Clean is not calling when multiple windows are looping, so it won't do unnecessary work.

> If you can outline what this is doing and who calls it for what purpose perhaps we come up with a better name and/or see if this is the right way to implement that requirement.

Maybe I could remove that function by destroying the svgs on Canvas refresh too, previously it was needed for renderers but since renderers are going to be removed when canvas refresh, then this function should not be necessary. I will remove it, then :)

internal/driver/glfw/window.go
fpabl0: done in ad7f974

internal/cache/base.go
fpabl0: done in ad7f974

internal/cache/base.go
fpabl0: done in ad7f974

internal/driver/glfw/window.go
fpabl0: > Maybe in the actual destroy instead of the Close() API call?

Yes that is a possibility, however we are going to do cleaning tasks in two different functions and it will not be consistent with gomobile driver that does not have that function. IMHO, calling this here does not represent any problem, it keeps consistency between the two drivers and the code for clean the cache on Close is centralized.

internal/driver/glfw/window.go
fpabl0: Not sure what you mean here. `closed()` is defined as:
```go
func (w *window) closed(viewport *glfw.Window) {
	viewport.SetShouldClose(false)

	if w.onCloseIntercepted != nil {
		w.QueueEvent(w.onCloseIntercepted)
		return
	}

	w.Close()
}
```
So `Close()` is always executed when `closed` is called unless there is a closeIntercepted callback, however inside that callback (correct me if I am wrong) users should use `Close` if they really want to close the window. At the end, `Close()` is always called 🤔

internal/cache/base.go
fpabl0: done in 342712a

fpabl0:
@andydotxyz I have worked in some of your suggestions, waiting for your review now. If everything goes fine, I will start writing the tests.

internal/driver/glfw/window.go
andydotxyz: Apologies, I forgot that we changed the order.
However it remains that a window is not truly closed until the driver removes it - `closed` or `Close` just request that it be removed, that may be why you are getting errors in the order of events.

andydotxyz: However, it may be that I have got my threads crossed with another area. If so apologies and this is fine.

internal/driver/glfw/window.go
andydotxyz: Calling it here does not represent a problem. Adding it to the user input event queue is confusing.
If it can be called without this workaround in another place (such as the GLFW driver's code that actually destroys windows) it will be cleaner within this driver.
I don't think that keeping the drivers "consistent" in code placement necessarily makes sense with regards to windows as they are not actually following the same lifecycle internally.

internal/cache/base.go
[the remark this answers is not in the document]
fpabl0: Is this ok?

internal/driver/glfw/window.go
fpabl0: > However it remains that a window is not truly closed until the driver removes it - closed or Close just request that it be removed

Yes, that's true, however it should be called here, because to free texture cache effectively we need to delete them using the viewport context and it is only available when the window is not closed, so calling it just before the window is closed seems to be only option.

internal/driver/glfw/window.go
fpabl0: Got it. Done in 569a354

andydotxyz:
Looking good thanks. If tests can be added things should be good.

fpabl0:
> Looking good thanks. If tests can be added things should be good.

Great :), I'll try to add them tonight.

internal/cache/base.go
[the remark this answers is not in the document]
fpabl0: @andydotxyz Any comment about this?

Jacalz (approved):
There are most likely more things we can do to improve the memory usage but this is a great step in the right direction. Nice work! :+1:

andydotxyz (approved):

andydotxyz:
Yes, nice work thanks.

PR #2107, opened 2021-03-22
Mobile storage writer

howeyc:
### Description:
Adds the ability to write/save files using fyne storage interface on android.

There is two parts:

The default root on android is already unique (data/app/id/path/files) without needing to go to a further subdirectory (data/app/id/path/files/app.id.path/).

The second commit is for handling the case where any URI path is sent to writer. In that case, we need to create any necessary directories, and create the file.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/internal/driver/gomobile/repository.go
+++ b/internal/driver/gomobile/repository.go
@@ -1,6 +1,10 @@
 package gomobile
 
 import (
+	"errors"
+	"os"
+	"path/filepath"
+
 	"fyne.io/fyne/v2"
 
 	"fyne.io/fyne/v2/storage/repository"
@@ -69,5 +73,20 @@ func (m *mobileFileRepo) Reader(u fyne.URI) (fyne.URIReadCloser, error) {
 }
 
 func (m *mobileFileRepo) Writer(u fyne.URI) (fyne.URIWriteCloser, error) {
+	if u == nil || u.Scheme() != "file" {
+		return nil, repository.ErrOperationNotSupported
+	}
+
+	path := u.Path()
+	_, serr := os.Stat(path)
+	if errors.Is(serr, os.ErrNotExist) {
+		dirpath := filepath.Dir(path)
+		os.MkdirAll(dirpath, 0755)
+		f, ce := os.Create(path)
+		if ce != nil {
+			return nil, ce
+		}
+		f.Close()
+	}
 	return fileWriterForURI(u)
 }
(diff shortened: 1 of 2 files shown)


internal/driver/gomobile/repository.go
```diff
@@ -69,5 +73,20 @@ func (m *mobileFileRepo) Reader(u fyne.URI) (fyne.URIReadCloser, error) {
 }
 
 func (m *mobileFileRepo) Writer(u fyne.URI) (fyne.URIWriteCloser, error) {
+	if u == nil || u.Scheme() != "file" {
+		return nil, repository.ErrOperationNotSupported
+	}
+
+	path := u.Path()
+	_, serr := os.Stat(path)
+	if errors.Is(serr, os.ErrNotExist) {
```
Jacalz, line 82: This is not going to work. We support Go 1.12 as the minimum version. You'd have to do `os.IsNotExists` instead.

internal/driver/gomobile/repository.go
```diff
@@ -69,5 +73,20 @@ func (m *mobileFileRepo) Reader(u fyne.URI) (fyne.URIReadCloser, error) {
...
+		dirpath := filepath.Dir(path)
+		os.MkdirAll(dirpath, 0755)
```
Jacalz, line 84: You should always handle errors for file operations.

internal/driver/gomobile/repository.go
andydotxyz, line 82: This was added in Go 1.13, we still support 1.12

Jacalz (changes requested):

internal/driver/gomobile/repository.go
```diff
@@ -69,5 +73,20 @@ func (m *mobileFileRepo) Reader(u fyne.URI) (fyne.URIReadCloser, error) {
 }
 
 func (m *mobileFileRepo) Writer(u fyne.URI) (fyne.URIWriteCloser, error) {
+	if u == nil || u.Scheme() != "file" {
+		return nil, repository.ErrOperationNotSupported
```
andydotxyz, line 77: We can write to URIs that are not file, so this check isn't right.
The workaround we need is for when a file write fails with file not found (or equivalent)

andydotxyz (changes requested):
Thanks, close but I think this code is too agressive on things it will reject (i.e. we could be writing to an existing content://)

howeyc:
After some internal discussion, going to handle this slightly differently.

Closing this.

PR #2109, opened 2021-03-23
Do not double update color channels when their associated entry text is changed programmatically, fixes #2075

### LINKED ISSUE #2075: [mobile] Advanced Color picker keeps updating values forever after modifying it with sliders
### Describe the bug:

Advanced Color picker dialog keeps updating values forever after modifying it with sliders.

### To Reproduce:
1. Run fyne_demo with mobile tag.
2. Click on Dialogs and then click on Advance Color Picker.
3. Try to change color channels values by moving the sliders.
4. See the error.

### Screenshots:

https://user-images.githubusercontent.com/12239342/110345552-b334cc00-7ffc-11eb-9b4c-48234cf5ac02.mp4

 - **OS:** MacOS
 - **Go version:** 1.16
 - **Fyne version:** develop @ 912d8d80

fpabl0:
### Description:

Fixes #2075

While this PR fixes #2075, color picker still has some problems on mobile, when the user move the slider in one of the R G B channels, the remaining RGB channels are moved too sometimes 😞

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/color_channel.go
+++ b/dialog/color_channel.go
@@ -126,9 +126,6 @@ func (r *colorChannelRenderer) updateObjects() {
 
 type colorChannelEntry struct {
 	userChangeEntry
-
-	lock      sync.RWMutex
-	userTyped bool
 }
 
 func newColorChannelEntry(c *colorChannel) *colorChannelEntry {
--- a/dialog/color_channel.go
+++ b/dialog/color_channel.go
@@ -2,6 +2,7 @@ package dialog
 
 import (
 	"strconv"
+	"sync"
 
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/canvas"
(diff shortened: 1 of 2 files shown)


andydotxyz:
Thanks for this fix.
I wonder though - do you have a thought on why this happens on mobile apps but not desktop?
Perhaps there is a bug elsewhere that we are working around here?

fpabl0:
> I wonder though - do you have a thought on why this happens on mobile apps but not desktop?

My thought is that it is caused due to the performance differences between mobile and desktop drivers. If you look carefully color picker seems to refresh things in a cyclic way: 
- It updates the slider value when entry text change.
- It updates the entry text when the slider value change. 

The Refresh by other hand set the entry text value and set the slider value, that results in multiple updates due to points above. If the two above points are executed enough fast we don't see the problem because it is the same value (desktop case) - although it is not performant. In mobile, these points are not executed so fast, so they overlaps each other with the next value updates. That cause that OnChanged callback in Entry tries to update colorChannel to different value comparing to the OnChanged callback in the slider. So it keeps changing forever because they can't be agreed which value is the right one.

fpabl0:
> Perhaps there is a bug elsewhere that we are working around here?

I would say the bug is in desktop and mobile, but desktop is enough fast to mitigate it, while mobile cannot.

andydotxyz (approved):

andydotxyz:
Let me know when are happy to landed this @fpabl0 an I will cherry pick it to the release branch

fpabl0:
@andydotxyz  Landed :)

andydotxyz:
Thanks, picked over

PR #2111, opened 2021-03-23
Reduce duplicate Slider events when using float64 Step value

andydotxyz:
Add an approximate comparison for floats which leads to fewer repeated events.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/slider.go
+++ b/widget/slider.go
@@ -104,7 +104,7 @@ func (s *Slider) Dragged(e *fyne.DragEvent) {
 
 	s.updateValue(ratio)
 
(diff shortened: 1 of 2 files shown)


fpabl0:
Looks good, but I don't think it is good to have `almostEqual` function, as the addition of it seems like a workaround for a problem instead of the real one that is update correctly the slider.Value.

fpabl0:
Ohh I see now what you mean, sorry. You are right :)

fpabl0 (approved):

PR #2117, opened 2021-03-25
Fix tab button layout on mobile

andydotxyz:
### Description:
Tab buttons were always rendered in a grid of columns, this PR uses a grid of rows when the tab location is leading or trailing

This is a re-application of "Fix tab button layout on mobile #1962" on top of the current `develop` code.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (commented):
Hmm, I don't see any changes compared to develop (visually). The top and bottom locations seem to work as expected, but trailing and leading seem to just be the same as bottom.
mobile-tabs

andydotxyz:
Maybe it was just doctabs that was the problem:

doctab-lead

fpabl0:
> Maybe it was just doctabs that was the problem:
I don't think so, the problem is in AppTabs too, I have faced that problem just a couple of days ago, but it seems to be very random :(

andydotxyz:
> I don't think so, the problem is in AppTabs too, I have faced that problem just a couple of days ago, but it seems to be very random :(

Does it work with this PR applied?

fpabl0:
> Does it work with this PR applied?

It is hard to say, because it was a random bug 😕

andydotxyz:
Is there anything I can do to move this forward @Jacalz? It's been a couple of weeks since the last comment.

Jacalz:
> Is there anything I can do to move this forward @Jacalz? It's been a couple of weeks since the last comment.

Sorry, I don’t know. I could not see any difference before and after so I can’t comment on it much unfortunately.

andydotxyz:
> I could not see any difference before and after

You don’t see the same as the screenshot I posted for doctabs?

fpabl0:
@andydotxyz 
I think @Jacalz is trying to say this (from his first comment):
> trailing and leading seem to just be the same as bottom.

Doc tabs respect leading and trailing position, but AppTaps no (it is always at bottom), is that intended?

andydotxyz:
You are quite right, something got messed up in transition.
The mobile tabs should only be positioned on the shorter edges.
This is now fixed and consistent between app and doc tabs.

stuartmscott:
The different tab location restrictions were in the requesting ticket (#1310), though it is not a hill worth dying on - if someone really wanted tabs on the long edges of mobile then a custom drawer widget that slides out to display the tabs would probably give a better UX.

andydotxyz:
Yeah this is more tidying up doc tabs to be consistent, but also resolving a mistake that somehow crept in to AppTabs too.
In summary, AppTabs have never drawn on the long edge for mobile, this commit flips leading position to top when in portrait or top to leading in landscape (I don’t know how/when this stopped being the case but you can see from the code comment that was intended). Then it updates doctabs to have the same layout constraint on mobile.
A navbar approach would indeed be a good approach if someone wants the left style, which is typically a richer component so I’m happy for them to be different. I think someone has a work in progress component heading to fyne-x at some point.

andydotxyz:
Any further thoughts @Jacalz or @fpabl0 ?

fpabl0 (approved):
LGTM

PR #2119, opened 2021-03-25
Synchronize ApplySettings with the draw thread

fpabl0:
### Description:

Fixes some issues on mobile where sometimes the objects are not repainted when the theme changes.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


internal/driver/gomobile/driver.go
```diff
@@ -113,100 +113,114 @@ func (d *mobileDriver) Quit() {
...
-						size := fyne.NewSize(float32(currentSize.WidthPx)/canvas.scale, float32(currentSize.HeightPx)/canvas.scale)
-						d.paintWindow(current, size)
-						a.Publish()
-					}
-				}
-			case size.Event:
-				if e.WidthPx <= 0 {
+		settingsChange := make(chan fyne.Settings)
+		fyne.CurrentApp().Settings().AddChangeListener(settingsChange)
+		draw := time.NewTicker(time.Second / 70)
```
andydotxyz, line 120: I don't think we should be driving the mobile at a constant rate - FPS and throughput varies so much.
To remove the issue of "wait 10ms then paint" blocking other events, would it be easier to use something like `time.AfterFunc` to move that block to a separate routine?

I tried the constant timer earlier in development and it caused issues. If you have managed to test with new and old iPhones, iPads and Android devices then fair enough, but if not I think we need to be cautious of the ramifications.

andydotxyz (changes requested):
This is a good idea, cleans things up.
But the mobile FPS based painting makes me nervous, as detailed inline.

internal/driver/gomobile/driver.go
fpabl0: > I don't think we should be driving the mobile at a constant rate

I see, but in this case it will only paint at constant rate if the previous paint has finished. Doesn't it solve the kind of issues you experienced before? I mean this will not overflow the mobile events as we always wait for the previous paint in order to trigger a new one.

> I tried the constant timer earlier in development and it caused issues.

🤔 What kind of issues did you have?

andydotxyz: > we always wait for the previous paint in order to trigger a new one.

Technically yes because we can only process 1 event at a time. But if 1 draw takes 4 frames in duration then as a result you will have 5 painting back to back, wasting CPU/GPU won’t you?

> What kind of issues did you have?

it was a long time back now (1.2) but I think we got a mix of hangs and phones getting hot.

fpabl0: @andydotxyz 
> Technically yes because we can only process 1 event at a time. But if 1 draw takes 4 frames in duration then as a result you will have 5 painting back to back, wasting CPU/GPU won’t you?

I think, I didn't understand you. Paintings are not queued, if 1 draw takes 4 frames, then only one more paint will happen at most, because I am not sending painting events on every tick, I first checked if the last one was finished or not:
```go
select {
case <-draw.C:
    if !drawFinish {
        continue
    }
    drawFinish = false
    a.Send(paint.Event{})
case paint.Event:
    ...
    drawFinish = true
}
```
Maybe you are telling me that paint.Event just delegate its job and so it doesn't actually wait until it is finished?, if that is the case then indeed it would do be a problem.

> it was a long time back now (1.2) but I think we got a mix of hangs and phones getting hot.

Wow, that's really bad. I will try to do it with a timer but I would like to know the above doubt first.

andydotxyz: Ah, I see - I had not understood what `drawFinish` did.
You are right this should stop the queue from building, however it has the undesired side-effect of ignoring pending changes.
If I have a slow drawing frame then updates will be ignored. My frame then renders but the UI is a few frames behind. Is this a problem? I'm not sure from looking.

fpabl0: > If I have a slow drawing frame then updates will be ignored.

As far as I can see, only the updates from the tick would be ignored, other ones would happen as usual because I am not limiting the usage of `a.Send(paint.Event{})`, I am preventing the ticker to do send paint events without finished the previous one. 

Nevertheless, it does have little problems in terms of performance:
1. As it is a ticker it will consume 1 loop cycle even if it cannot send a paint event.
2. Two consecutive paint events could be triggered if a paint event is sent from other place and it finished just an instance before the tick.

Using a timer will give us more granularity to solve the above problems, so I will try to implement it with it.

fpabl0: I have replaced the ticker by a timer, please check :)

fpabl0: Not sure what does this mean 🤔:
```
cyclomatic complexity 32 of func `(*mobileDriver).Run` is high (> 30) (gocyclo)
```

fpabl0: @andydotxyz Updated according to the Friday's call.

andydotxyz: I think the 70 should be a 60?
I know we had an additional 10ms delay before, but that was without a timer...

fpabl0: You are right, I was not sure about this, because unlike the desktop driver the paintEvent does not occur immediately on every tick, it has to wait its turn inside the channel selector, so I thought that 70 maybe mitigate this situation?

andydotxyz: I don't understand. If you send something to a channel every 60th sec, then (assuming nothing is busy) it will be read at the other end every 60th... How does slowing the input speed help?

fpabl0: Oh you are right, I thought that `a.Events()` always had events, but that is not case, I will update.

fpabl0: @andydotxyz Updated :)

andydotxyz (approved):
This feels really slick now, thanks

andydotxyz (approved):

Issue #2122, opened 2021-03-26
widget.Check component callback adds id mark


dosgo:
widget.Check component callback adds id mark

I have added the NewCheckById and SetId methods, which are very useful in using the tree+check combination. I hope they can be combined.

https://github.com/dosgo/fyne/tree/master/widget

<pre>
func example1() {
	type GroupCheck struct {
		GroupName string
		GroupId int
		Items []string;
	}
	var listCheck []GroupCheck
	listCheck=make([]GroupCheck,0)
	listCheck=append(listCheck,GroupCheck{"group1",1,[]string{"Items1","Items2","Items3"}})
	listCheck=append(listCheck,GroupCheck{"group2",2,[]string{"Items3","Items4","Items5"}})
	var allData sync.Map
	a := app.New()
	mainWindow := a.NewWindow("test")
	//bottom
	var checkGroup []fyne.CanvasObject
	for _,v:=range listCheck{
		allCheck:=widget.NewCheckById("all Check",v.GroupId, func(id int, b bool) {
			items,ok:=allData.Load(id)
			if ok {
				for _, v := range items.([]binding.Bool) {
					v.Set(b)
				}
			}
		})
		_group := container.NewHBox(container.NewMax(widget.NewLabel(v.GroupName)),allCheck)
		_item := container.NewVBox(_group)
		tempBool:=make([]binding.Bool,0)
		for _,v1:=range v.Items {
			boolData := binding.NewBool()
			tempBool=append(tempBool,boolData)
			_item.Add(widget.NewCheckWithData(v1, boolData))
		}
		allData.Store(v.GroupId,tempBool)
		checkGroup=append(checkGroup,_item);
	}
	bottom:=container.NewVBox(checkGroup...)
	mainWindow.SetContent(bottom);
	mainWindow.SetMaster()
	mainWindow.ShowAndRun();
}
</pre>

<pre>
tree + check
func example2() {
	var checkds =make(map[int]bool,0)
	type Tutorial struct {
		Title string
	}
	var _Tutorials = map[string]Tutorial{
		"group1": {"group1"},
		"group2": {"group2"},
		"item1": {"item1"},
		"item2": {"item2"},
		"item3": {"item3"},
		"item4": {"item4"},
	};
	var TutorialIndex = map[string][]string{
		"":            {"group1", "group2"},
		"group1":  {"item1", "item3"},
		"group2":     {"item2", "item4"},
	}

	var uidentifier=map[string]int{
		"item1":1,
[...]

PR #2124, opened 2021-03-27
widget.Check component callback adds id mark.

dosgo:
widget.Check component callback adds id mark

I have added the NewCheckById and SetId methods, which are very useful in using the tree+check combination.
https://github.com/fyne-io/fyne/issues/2122

### DIFF
--- a/widget/check.go
+++ b/widget/check.go
@@ -96,15 +96,25 @@ func (c *checkRenderer) updateFocusIndicator() {
 type Check struct {
 	DisableableWidget
 	Text    string
+	Id string
 	Checked bool
 
 	OnChanged func(bool) `json:"-"`
+	OnChangedById func(string,bool) `json:"-"`
 
 	focused bool
 	hovered bool
 
 	checkSource   binding.Bool
 	checkListener binding.DataListener
+}
+// Bind connects the specified data source to this Check.
+// The current value will be displayed and any changes in the data will cause the widget to update.
+// User interactions with this Check will set the value into the data source.
+//
+// Since: 2.0
+func (c *Check) SetId(Id string) {
+	c.Id=Id;
 }
 
 // Bind connects the specified data source to this Check.
@@ -147,6 +157,9 @@ func (c *Check) SetChecked(checked bool) {
 
 	if c.OnChanged != nil {
 		c.OnChanged(c.Checked)
+	}
+	if c.OnChangedById!=nil {
+		c.OnChangedById(c.Id,c.Checked)
 	}
 
 	c.Refresh()
@@ -228,11 +241,27 @@ func NewCheck(label string, changed func(bool)) *Check {
 		DisableableWidget: DisableableWidget{},
 		Text:              label,
 		OnChanged:         changed,
+		OnChangedById: nil,
 	}
 
 	c.ExtendBaseWidget(c)
 	return c
 }
+
+// NewCheck creates a new check widget with the set label and change handler
+func NewCheckById(label string,Id string, changed func(string,bool)) *Check {
+	c := &Check{
+		DisableableWidget: DisableableWidget{},
+		Text:              label,
+		Id:                Id,
+		OnChanged:         nil,
+		OnChangedById: changed,
+	}
+	c.ExtendBaseWidget(c)
+	return c
+}
+
+
 
 // NewCheckWithData returns a check widget connected with the specified data source.
 //
--- a/widget/check.go
+++ b/widget/check.go
@@ -108,9 +108,9 @@ type Check struct {
 	checkSource   binding.Bool
 	checkListener binding.DataListener
 }
-// Bind connects the specified data source to this Check.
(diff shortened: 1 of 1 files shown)


andydotxyz:
Before you put too much effort into these ID based PRs I wanted to comment that it is not the “Fyne way” to use strings to identify widgets in a tree. It is preferable to use simple Go variables so that type safety is provided and the compiler can verify what your code is doing.

dosgo:
Sorry I didn't find a way to identify widgets in fyne. Strings are based on slices. Using int is indeed a good way.

andydotxyz:
It depends on why you need to maintain a reference really, but:
```go
myCheck := widget.NewCheck(...)
```
Then `myCheck` can be used later, and is of type `*widget.Check`.

But like I said it depends on usage and if you’re in `Tree` things may be less clear.  Because of caching internally you cannot reference one and be sure of which data it represents, as they will be re-used. So the `UpdateNode` function is responsible for setting an `OnChanged` or similar which is attached to data instead of looking up a reusable UI component.

dosgo:
I need to get the widget.check checked by the child nodes of the tree. I don't know how the existing controls are implemented, and I haven't found a similar method, and my tree is dynamically generated based on the configuration file.
Because the tree will be reused, it is difficult to know which node's check is selected using OnChanged or NewCheckWithData.

dosgo:
I need a function like this, but my tree is dynamic and I find it difficult to implement fyne.
image

andydotxyz:
Tree (and list and table) are designed to be backed by an application data structure (see  https://developer.fyne.io/tour/widget/list.html) so you should not access inside components to get their state.
Instead apply the `Checked` state inside the `UpdateNode` callback, and also set an `OnChanged` that updates the data if it is changed (the `TreeNodeID` will tell you the node that is changing).

dosgo:
Such?  thank you! I don't know if this will bring a burden to gc. Then how to deal with this in example 1, the parent node selects all the children, and the parent node cancels all cancels.
<pre>
func example3() {
	var checkds =make(map[string]bool,0)
	type Tutorial struct {
		Title string
	}
	var _Tutorials = map[string]Tutorial{
		"group1": {"group1"},
		"group2": {"group2"},
		"item1": {"item1"},
		"item2": {"item2"},
		"item3": {"item3"},
		"item4": {"item4"},
	};
	var TutorialIndex = map[string][]string{
		"":            {"group1", "group2"},
		"group1":  {"item1", "item3"},
		"group2":     {"item2", "item4"},
	}

	var uidentifier=map[string]int{
		"item1":1,
		"item2":2,
		"item3":3,
		"item4":4,
	}

	a := app.New()
	mainWindow := a.NewWindow("test")
	tree := &widget.Tree{
		ChildUIDs: func(uid string) []string {
			return TutorialIndex[uid]
		},
		IsBranch: func(uid string) bool {
			children, ok := TutorialIndex[uid]
			return ok && len(children) > 0
		},
		CreateNode: func(branch bool) fyne.CanvasObject {
			return container.NewHBox(widget.NewLabel(""),widget.NewCheck("", func(b bool) {
			}));
		},
		UpdateNode: func(uid string, branch bool, obj fyne.CanvasObject) {
			t, ok := _Tutorials[uid]
			if !ok {
				fyne.LogError("Missing tutorial panel: "+uid, nil)
				return
			}
			obj.(*fyne.Container).Objects[0].(*widget.Label).SetText(t.Title);
			if branch {
				obj.(*fyne.Container).Objects[1].(*widget.Check).Hide()
			} else  {
				if _,ok:=uidentifier[uid]; ok {
					obj.(*fyne.Container).Objects[1].(*widget.Check).OnChanged= func(b bool) {
						checkds[uid]=b;
						fmt.Printf("checkds:%+v\r\n",checkds)
					}
				}
				obj.(*fyne.Container).Objects[1].(*widget.Check).Show()
			}
			_Tutorials[uid]=t;
		},
		OnSelected: func(uid string) {
		},
	}
	mainWindow.SetContent(tree);
	mainWindow.SetMaster()
	mainWindow.ShowAndRun();
}
</pre>

andydotxyz:
That looks roughly correct. The hide/show for check should not be needed, as the leaf/branch nodes have different caches / so you can use different templates.

what is your concern about GC?

dosgo:
The CreateNode method seems to have no distinction between parent node and child node. This dynamic binding method will create multiple anonymous functions. Only one named function needs to be bound to the id method, but I don’t know how golang gc works.

andydotxyz:
> Only one named function needs to be bound to the id method, 

if you set a single method using an ID that bound to the CanvasObject it will be invalid upon re-use, we need to keep minimal graphical elements in memory, so with 10000 items of data you’ll only have around 10 Checks created.

dosgo:
In the UpdateNode method, you can re-setid according to the TreeNodeID, or it can be reused, just like SetText. I found no problems in my test.

PR #2126, opened 2021-03-28
Call mouseOut function when the mouse is out of the current hoverable object

fpabl0:
### Description:

Fix this behavior:
Before this PR:

https://user-images.githubusercontent.com/12239342/112759939-daa70500-8fba-11eb-91e0-d30c5c1f286f.mp4

After:

https://user-images.githubusercontent.com/12239342/112759951-e5619a00-8fba-11eb-82e0-eebe00306c1d.mp4

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/testdata/windows_hover_object.xml
+++ b/internal/driver/glfw/testdata/windows_hover_object.xml
@@ -1,0 +1,49 @@
+<canvas size="200x300">
+	<content>
+		<widget pos="4,4" size="192x292" type="*widget.List">
+			<widget size="192x292" type="*widget.Scroll">
+				<container size="192x292">
+					<widget size="192x45" type="*widget.listItem">
+						<widget pos="8,4" size="180x37" type="*widget.Entry">
+							<rectangle fillColor="rgba(25,25,25,25)" pos="0,2" size="180x33"/>
+							<rectangle fillColor="shadow" pos="0,35" size="180x2"/>
+							<widget pos="0,2" size="180x33" type="*widget.Scroll">
+								<widget size="180x33" type="*widget.entryContent">
+									<widget size="180x33" type="*widget.textProvider">
+										<text color="placeholder" pos="8,6" size="168x21"></text>
+									</widget>
+									<widget size="180x33" type="*widget.textProvider">
+										<text pos="8,6" size="168x21"></text>
+									</widget>
+								</widget>
+							</widget>
+						</widget>
+					</widget>
+					<widget pos="0,46" size="192x45" type="*widget.listItem">
+						<rectangle fillColor="hover" size="192x45"/>
+						<widget pos="8,4" size="180x37" type="*widget.Entry">
+							<rectangle fillColor="rgba(25,25,25,25)" pos="0,2" size="180x33"/>
+							<rectangle fillColor="shadow" pos="0,35" size="180x2"/>
+							<widget pos="0,2" size="180x33" type="*widget.Scroll">
+								<widget size="180x33" type="*widget.entryContent">
+									<widget size="180x33" type="*widget.textProvider">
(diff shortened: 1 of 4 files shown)


andydotxyz (changes requested):
I don't think this is the correct fix.
You are looking for any object - not limiting it to those that support hoverable. The general contract is that we send these events based on the heirarcy so a hoverable parent will remain so when a child is hovered, if the child does not grab the state.

The bug here is that the List widget (which is hoverable) does not "unhover" child elements when the emptu background space is hovered. It assumes I guess that a child will always be under the mouse if the mouse is inside the List (incorrectly).

fpabl0:
@andydotxyz 
> You are looking for any object - not limiting it to those that support hoverable. 

Hmm, I think the normal behavior of "hover" is "unhover" it when the mouse is not over the object (in other words, the mouse is over any other object different from the hoverable one).

> The general contract is that we send these events based on the heirarcy so a hoverable parent will remain so when a child is hovered, if the child does not grab the state.

I am not sure, what you mean here 🤔. That behavior was not modified in this PR.

> The bug here is that the List widget (which is hoverable) does not "unhover" child elements when the emptu background space is hovered.

List widget is not hoverable, its children are.

andydotxyz:
> Hmm, I think the normal behavior of "hover" is "unhover" it when the mouse is not over the object (in other words, the mouse is over any other object different from the hoverable one).

What you have described here does not match the code as far as I can see - that is traversing to see if the element is a child of the current hovered.
Something doesn't seem to quite add up here - it seems that only List exhibits this. Take a look at Tabs and they work as expected, with very similar code. Can you explain why the List is different, and so why this fix is the right thing to do?
The fix seems quite broad but the problem very specific.

fpabl0:
@andydotxyz 
> What you have described here does not match the code as far as I can see - that is traversing to see if the element is a child of the current hovered.

I mean other other objects that do not belong to the hoverable object.

> Take a look at Tabs and they work as expected, with very similar code.

Indeed, Tabs have the same problem, take a look:

https://user-images.githubusercontent.com/12239342/112895868-46649d00-90a3-11eb-8a59-a0195c5526ee.mp4

> Can you explain why the List is different, and so why this fix is the right thing to do?

Yes, I can. This problem will appear every time you have a hoverable object inside a draggable object with some empty space. This is the case for DocTabs and List because they have internally a scroll container (that is a draggable object), and both of them resize the scroll container to fill the space, so if their children are not enough to fill the scroll container, then there would be a gap, and so you would experience the problem.

fpabl0:
This PR solves this current problem too (scroll bar keeps its hover state while user is typing and they can't move the cursor easily)

https://user-images.githubusercontent.com/12239342/113219862-d4798880-9247-11eb-8def-8f2e51881b9b.mp4

fpabl0:
This bug would also be fixed:

https://user-images.githubusercontent.com/12239342/113222552-e873b900-924c-11eb-882e-d6c3017d7500.mp4

andydotxyz:
> Indeed, Tabs have the same problem, take a look:

Apologies, I was testing on an old version - it seems the issue is relatively new.
I will try to understand this better and how it came about.

fpabl0:
@andydotxyz  You are right, this seems to be introduced in #1789. In 1.4 (I think), there was an assumption that draggable objects should be hoverable too. But that assumption was changed in 2.0 (I would say it was fixed), so objects can now be "only Draggable" instead of be Hoverable too. The specific line that introduce this is:
https://github.com/fyne-io/fyne/blob/73e3659e2479d3f3445438623c104b8b95cf87e2/internal/driver/glfw/window.go#L597-L599

andydotxyz (approved):

PR #2131, opened 2021-04-02
Fix crash in label with binding

### LINKED ISSUE #2125: Data binding with widget.List sometimes crash while scrolling
### Describe the bug:

Using data binding for a `widget.List` sometimes crash while scrolling. It seems to be a race condition. The error is:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x427a206]

goroutine 8 [running]:
fyne.io/fyne/v2/data/binding.(*stringFromFloat).Get(0x0, 0xc0002be508, 0xc000380290, 0xc0002be5b8, 0xc00033c800)
        .../fyne/data/binding/convert.go:109 +0x26
fyne.io/fyne/v2/widget.(*Label).Bind.func1()
        .../fyne/widget/label.go:59 +0x3c
fyne.io/fyne/v2/data/binding.(*listener).DataChanged(0xc001616728)
        .../fyne/data/binding/binding.go:55 +0x28
fyne.io/fyne/v2/data/binding.processItems()
        .../fyne/data/binding/queue.go:56 +0x94
created by fyne.io/fyne/v2/data/binding.init.0
        .../fyne/data/binding/queue.go:15 +0x35
exit status 2
```
### To Reproduce:
1. Run fyne_demo
2. Click on the binding tutorial.
3. Try appending items and scrolling (sometimes it crashes)

 - **OS:** MacOS
 - **Version:** 10.15.7 Catalina
 - **Go version:** 1.16.2
 - **Fyne version:** develop @ 23e0ed1

andydotxyz:
Moved to a different setup model where we don't recreate listeners all the time.
Fixes #2125

If this is a desirable change we could move other Bind/Unbind widget to match.

### Checklist:

- [ ] Tests included. <- race
- [x] Tests all pass.

### DIFF
--- a/widget/label.go
+++ b/widget/label.go
@@ -98,11 +98,12 @@ func (l *Label) SetText(text string) {
 //
 // Since: 2.0
 func (l *Label) Unbind() {
-	if l.textSource == nil || l.textListener == nil {
+	src := l.textSource
+	if src == nil {
 		return
 	}
 
-	l.textSource.RemoveListener(l.textListener)
+	src.RemoveListener(l.textListener)
 	l.textSource = nil
 }
 
@@ -112,10 +113,11 @@ func (l *Label) createListener() {
 	}
 
 	l.textListener = binding.NewDataListener(func() {
-		if l.textSource == nil {
+		src := l.textSource
+		if src == nil {
 			return
 		}
-		val, err := l.textSource.Get()
+		val, err := src.Get()
 		if err != nil {
 			fyne.LogError("Error getting current data value", err)
 			return
--- a/widget/label.go
+++ b/widget/label.go
@@ -55,18 +55,7 @@ func NewLabelWithStyle(text string, alignment fyne.TextAlign, style fyne.TextSty
 func (l *Label) Bind(data binding.String) {
 	l.Unbind()
 	l.textSource = data
-	l.textListener = binding.NewDataListener(func() {
-		val, err := l.textSource.Get()
-		if err != nil {
-			fyne.LogError("Error getting current data value", err)
-			return
-		}
-
-		l.Text = val
-		if cache.IsRendered(l) {
-			l.Refresh()
-		}
-	})
+	l.createListener()
 	data.AddListener(l.textListener)
 }
 
@@ -114,8 +103,29 @@ func (l *Label) Unbind() {
 	}
 
 	l.textSource.RemoveListener(l.textListener)
-	l.textListener = nil
 	l.textSource = nil
+}
+
+func (l *Label) createListener() {
+	if l.textListener != nil {
+		return
+	}
+
+	l.textListener = binding.NewDataListener(func() {
+		if l.textSource == nil {
+			return
+		}
+		val, err := l.textSource.Get()
+		if err != nil {
+			fyne.LogError("Error getting current data value", err)
+			return
+		}
+
+		l.Text = val
+		if cache.IsRendered(l) {
+			l.Refresh()
+		}
+	})
 }
 
 // textAlign tells the rendering textProvider our alignment


fpabl0 (approved):
LGTM. I will try to explore the connected flag in other PR to see if it is viable :)

fpabl0:
It seems that there is something more we have to do here, sometimes I got an error inside convert.go:

```go
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x42541c6]

goroutine 8 [running]:
fyne.io/fyne/v2/data/binding.(*stringFromFloat).Get(0x0, 0xc000048f78, 0x42545fa, 0xc0003bcdd8, 0x40638e0)
        .../fyne/data/binding/convert.go:109 +0x26
        .../fyne/widget/label.go:71 +0x45
fyne.io/fyne/v2/data/binding.(*listener).DataChanged(0xc00018e9c0)
        .../fyne/data/binding/binding.go:63 +0x2e
fyne.io/fyne/v2/data/binding.processItems()
        .../fyne/data/binding/queue.go:56 +0x94
created by fyne.io/fyne/v2/data/binding.init.0
        .../fyne/data/binding/queue.go:15 +0x35
exit status 2
```

andydotxyz:
> It seems that there is something more we have to do here, sometimes I got an error inside convert.go:

Those line numbers don't mat this PR... do you have local changes?

fpabl0:
Yes, but they are commented. I copied your changes and paste in a local branch. I will try to reproduce it, by cloning directly your branch, although I think it would be the same 🤔. I will investigate. However it seems that this PR is ok and this can be solved in a different one?

fpabl0:
Ok, I have tested your branch without any change, this is the error trace:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x4254c46]

goroutine 8 [running]:
fyne.io/fyne/v2/data/binding.(*stringFromFloat).Get(0x0, 0xc0002b0508, 0xc00009a470, 0xc0002b05b8, 0x40643a0)
        /Users/pablofuentes/Desktop/fyne/data/binding/convert.go:109 +0x26
        /Users/pablofuentes/Desktop/fyne/widget/label.go:118 +0x45
fyne.io/fyne/v2/data/binding.(*listener).DataChanged(0xc000338008)
        /Users/pablofuentes/Desktop/fyne/data/binding/binding.go:55 +0x28
fyne.io/fyne/v2/data/binding.processItems()
        /Users/pablofuentes/Desktop/fyne/data/binding/queue.go:56 +0x94
created by fyne.io/fyne/v2/data/binding.init.0
        /Users/pablofuentes/Desktop/fyne/data/binding/queue.go:15 +0x35
```

fpabl0:
This is really weird, it seems that there is a chance that `s` could be nil here:
```go
func (s *stringFromFloat) Get() (string, error) {
	val, err := s.from.Get()
	if err != nil {
		return "", err
	}

	return fmt.Sprintf(s.format, val), nil
}
```

widget/label.go
```diff
@@ -114,10 +103,31 @@ func (l *Label) Unbind() {
...
 	l.textSource = nil
 }
 
+func (l *Label) createListener() {
+	if l.textListener != nil {
+		return
+	}
+
+	l.textListener = binding.NewDataListener(func() {
+		if l.textSource == nil {
```
fpabl0, line 115: It seems that this part should be written as:
```go
src := l.textSource
if src == nil {
    return
}
val, err := src.Get()
```
To nullify any chance to have a nil textSource when we actually call `Get()`.
This is because it seems that there is a little probability that textSource is nil when Go (internally) is converting the interface to its actual type? (I am supposing this, as that could be the only reason to explain this). Take a look this code:
https://play.golang.org/p/bqEsHGUYdfw

fpabl0 (changes requested):
I have found something interesting that I didn't know before (probably it was logical but I hadn't realized it). Nil objects can call its methods without actually crashing the whole program:
https://play.golang.org/p/tOzpRaZTt3m

widget/label.go
```diff
@@ -114,10 +103,31 @@ func (l *Label) Unbind() {
 	}
```
fpabl0, line 103: `Unbind` method should be re-written to:
```go
	src := l.textSource
	listen := l.textListener
	if src == nil || listen == nil {
		return
	}

	src.RemoveListener(listen)
	l.textSource = nil
```

fpabl0 (changes requested):
Same change should be done on `Unbind` method.

fpabl0 (approved):

PR #2132, opened 2021-04-03
DataItems should not trigger their listeners if their underlying value has not changed

fpabl0:
### Description:

DataItems should not trigger their listeners if their underlying value has not changed.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


fpabl0:
I have intentionally exclude boundURI from this change, as I am not sure how it is need to be compared

data/binding/gen.go
```diff
@@ -66,12 +66,19 @@ func (b *bound{{ .Name }}) Get() ({{ .Type }}, error) {
 }
 
 func (b *bound{{ .Name }}) Reload() error {
-	return b.Set(*b.val)
+	b.lock.Lock()
+	defer b.lock.Unlock()
+	b.trigger()
```
andydotxyz, line 71: This seems to do the opposite of your PR title - it will call the trigger irrespective of the change. 
Looking more does it even set the value?

data/binding/gen.go
```diff
@@ -521,6 +528,7 @@ type bindValues struct {
 	Format, Since        string
 	SupportsPreferences  bool
 	FromString, ToString string // function names...
+	Custom               bool   // check whether it is a custom type or a built-in one
```
andydotxyz, line 531: "Custom" is pretty vague... perhaps comparable?
It would be better to provide a comparator so that URI (and other future types) are consistent with the rest.

andydotxyz (changes requested):
Thanks for tackling this area.
This was quite contentious during develpment but I agree it makes more sense.
However I think that the implementation can be improved, noted inline.

data/binding/gen.go
fpabl0: > it will call the trigger irrespective of the change.

@andydotxyz Yes, I am aware of it. But `Reload` is part only of `ExternalXxx`, then directly trigger would be the only thing we can do, as it would already have the `b.Val` set by reference? 🤔. From my perspective `Reload` is like a force trigger, isn't it?

data/binding/gen.go
fpabl0: > "Custom" is pretty vague... perhaps comparable?

Agree, but I am not sure `comparable` is the right replacement, because as you can see at the end all the types should be comparable, perhaps `BuiltinEquality`?

> It would be better to provide a comparator so that URI (and other future types) are consistent with the rest.

Yes, it was to be my next proposal 😅. Perhaps there could a new interface called `fyne.Equatable`:
```go
type Equatable interface {
     Equal(v Equatable) bool
}
```

andydotxyz: No need for new types - instead of passing a `bool` then pass in an optional `func(from, to interface{}` so it is handled in this package. URL is easy as you just compare the `String()` versions

data/binding/gen.go
andydotxyz: `Reload` is not like forcing trigger. It is ensuring data is up to date. 
If you consider a `[]string` bind with 10000 items we only want to trigger refresh on the items that have changed - to save 99997 callbacks.
We should be consistent and having internal and external data behave differently is not good.
One reason this was so contentious was that to correctly handle "if-changed" for all types would requiring storing a duplicate for comparison on the types where the data is external.

fpabl0: > Reload is not like forcing trigger. It is ensuring data is up to date.

But `ExternalXxx` versions would always ensure data is up to date, as you are saving its reference. So the only thing `Reload` can do is trigger the listener's callback to report that change. If user does not want to trigger unnecessarily all the listeners (assuming the change of the external variable is the same as before), then they can use `Set` function.

> If you consider a []string bind

`DataLists` work different internally I think, this change is only for `DataItems`.

> We should be consistent and having internal and external data behave differently is not good.

Well considering `Reload` is only present in `ExternalXxx` variants, I don't think they behave different from internal ones with this change, do they?

data/binding/gen.go
fpabl0: > No need for new types - instead of passing a bool then pass in an optional func(from, to interface{} so it is handled in this package.

I am bit confused 😅. Is this related to the `Equatable` interface I mentioned or the `Custom` flag?

andydotxyz: It replaces both

andydotxyz:
You mgiht want to check out when I coded this feature originally https://github.com/fyne-io/fyne/commit/1c0259ccc8d59834f85b6ea14a3fdd3cfdce9d3d and then the commit where it was backed out due to the additional memory required https://github.com/fyne-io/fyne/commit/8e1ecb0689a4bf954c52eae8e144d3fb4577f8bb (though the "old" field was backed out 2 commits later as I had missed it).

fpabl0:
@andydotxyz  Thanks for pointed me out these commits. I already checked them before along with the PR that you coded (https://github.com/fyne-io/fyne/pull/1717).

With that background, maybe DataItems should be implemented just like the data list items, I mean having two different types one for "internals" and other for "externals" one, then only the external variant would have the `old` cache value. 

Nevertheless, I don't think it is a bad idea having `Reload` method as in this PR, because at the end, if user uses `Reload`, it is because the user is sure that the changes should be "reloaded", did you see my last comment on your review?

data/binding/gen.go
andydotxyz: Comparator works, no need for tthis to be used for builtin I wouldn't think - nil makes sense.

andydotxyz:
I am thinking about this and it's not clear. You say that someone calling `Reload` means that the developer knows it has changed, and yet the purpose of this commit is to check for change when `Set` is called. It's slightly ironic because you could state the same assumption for `Set`...

fpabl0:
> It's slightly ironic because you could state the same assumption for Set...

Yes, this is contradictory 😞. But at least now the developer has the choice, developer could use `Set` to ensure listeners are only triggered when the value effectively changes, or could use `Reload` if it doesn't matter (and he/she updates it from external value).

Perhaps my previous comment is a better alternative?:
> maybe DataItems should be implemented just like the data list items, I mean having two different types one for "internals" and other for "externals" one, then only the external variant would have the old cache value.

data/binding/gen.go
fpabl0: Got it, I will update :)

andydotxyz:
The option that you provide does not make sense to me. The `Set` is the only way to set data when using internal representations. The contract of `Reload` for external is to look at the data and sync it up (this is where list and map are already doing smart things).
This PR feels like it is essentially reversing that contract.

fpabl0:
> The contract of Reload for external is to look at the data and sync it up

For external values, the data is already synced up (at least for dataItem's values that are not pointers). Previous `Reload` implementation does the same thing as in this PR. 

As far as I can see, the only usage of `Reload` (for DataItems no DataListItems) is trigger the listeners. 

If we really want this to work only if it data effectively changes, then I think there should be new types for external ones like `boundExternalXxx` that embeds `boundXxx` and adds the old cache value.

```go
type boundExternalFloat {
      boundFloat
      old float64
}
```
Extra memory for old value in `boundExternalXxx` is better than having things triggered when they are not needed, because we don't have control of what listener callback could the developer add to it, this callback could be expensive and trigger it lot of times, does not a good deal.

data/binding/gen.go
fpabl0: I am not sure how to implement this, you mean `Comparator` should be part of the `bindValues` struct but that struct is only used for template execution. Am I missing something? Maybe I misunderstood you.

andydotxyz:
> Extra memory for old value in boundExternalXxx is better than having things triggered when they are not needed, because we don't have control of what listener callback could the developer add to it. This callback could be expensive and trigger it lot of times, does not a good deal.

I completely agree with this, but not everybody does. It is how the list/map functionality works and I think that consistency is important as always :)

fpabl0:
> but not everybody does

I see, but I think splitting types for internal and external binds, was accepted in https://github.com/fyne-io/fyne/pull/1717, doesn't it?

> It is how the list/map functionality works and I think that consistency is important as always :)

Yes :), so should I update to have different types for internal and external ones, just like in DataListItems?

andydotxyz:
> > but not everybody does
> 
> I see, but I think splitting types for internal and external binds, was accepted in #1717, doesn't it?
> 
> > It is how the list/map functionality works and I think that consistency is important as always :)
> 
> Yes :), so should I update to have different types for internal and external ones, just like in DataListItems?

I think you got the measure of it there :) If you read the chat log on that issue you can see what some of the concerns were. However as you say it was landed and being consistent is a good aim.

fpabl0:
@andydotxyz Updated :)

andydotxyz (approved):
Thanks. It's good to be consistent across the different types of data again.
And it will remove the infinite loop that one of the 7guis examples got into!

PR #2137, opened 2021-04-04
DataLists should notify only once when change

fpabl0:
### Description:

DataLists should notify only once when change.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


data/binding/gen.go
```diff
@@ -308,12 +308,12 @@ const listBindTemplate = `
...
-	Append({{ .Type }}) error
+	Append(val {{ .Type }}) error
 	Get() ([]{{ .Type }}, error)
-	GetValue(int) ({{ .Type }}, error)
-	Prepend({{ .Type }}) error
-	Set([]{{ .Type }}) error
-	SetValue(int, {{ .Type }}) error
+	GetValue(index int) ({{ .Type }}, error)
+	Prepend(val {{ .Type }}) error
+	Set(v []{{ .Type }}) error
```
andydotxyz, line 315: If this change is for readability I'd suggest that the `v` should be `val` to match the other additions?

andydotxyz (changes requested):
This is a good catch in many ways thanks, I had not noticed that it would trigger so many times on construction!

Just a note on naming.

data/binding/gen.go
fpabl0: Agree, initially I chose `v` for `Set(...)`, because I wanted to difference singular values from composite ones (slices), but maybe this is less clear? Or should we use `l` (**l**ist)?

andydotxyz: What is wrong with using `list`? or `values` something that make it clear there is a difference.

fpabl0: Agree :) I will update.

fpabl0: @andydotxyz Updated :)

andydotxyz (approved):
nice thanks

PR #2140, opened 2021-04-05
Do not show virtual keyboard if the canvas object (widget) is disabled, fixes 2139

### LINKED ISSUE #2139: No keyboard popup on Android in disabled Entry Widget
My app shows a long help text in a new window, containing a MultiLineEntry widget within a scroll container. Because the text shall not be changeable, the entry widget is disabled.
On an Android phone it is almost impossible to grab the scrollbar and scroll the help text. Instead, with every touch (try) to get the narrow scrollbar, the keyboard pops up.
There is no need to let the keyboard pop up if the widget is disabled.

not that I'm aware of

If possible on Android, popup of the keyboard should be prevented (or possible to prevent) for a disabled Entry widget.

fpabl0:
### Description:

Fixes #2139 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/gomobile/canvas.go
+++ b/internal/driver/gomobile/canvas.go
@@ -253,7 +253,11 @@ func (c *mobileCanvas) focusManager() *app.FocusManager {
 }
 
 func (c *mobileCanvas) handleKeyboard(obj fyne.Focusable) {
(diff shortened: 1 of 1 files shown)


andydotxyz (approved):
Good catch

PR #2143, opened 2021-04-06
Adding a few new svg icons

xines:
### Description:

I found these icons missing for my own project so others might aswell find them useful.

The svg icons are found on https://materialdesignicons.com

If i made any mistakes please let me know as this is my first pull request here.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (commented):
Thank you very much for the PR. I don’t mean to have this review be negative, I just try to provided my thoughts on the icons being added 😅

I’m not quite sure how much of these we should add in. I personally think we should keep the amount of icons low and mostly sordid to those that are most likely to be used. I don’t really have anything against account icon, but the chart icon I think shouldn’t be added in until we have some kind of chart widget.

As for the Twitter icon, I think it and other social media icons are up to other users to integrate themselves. Adding one social media icons means that we more or less would need to add in all the other sooner or later.

That’s just my personal take on this. Feel free to disagree.

xines:
> I don’t really have anything against account icon, but the chart icon I think shouldn’t be added in until we have some kind of chart widget.
- I agree, the account icon is the one i orginally wanted to only pull but had in mind that charting would be nice for the future but aren't really much usage for now.

> As for the Twitter icon, I think it and other social media icons are up to other users to integrate themselves. Adding one social media icons means that we more or less would need to add in all the other sooner or later.
- Yes social media icons was just a idea, and i'll have to agree and disagree due to having a few "big" social media platform icons added doesn't seem like a problem but i dislike the idea of adding smaller platforms as it's obvious an endless task and as you mentioned they can add the specific ones themselves.

- Also Ty for taking the time to review, criticism is all fine no worries 😄

Jacalz:
Also, I think that adding social media icons might cause license issues as the companies own the copyright on their logos. I'm not entirely sure given that these are adaptions of those logos, but it might be worth thinking about.

xines:
> Also, I think that adding social media icons might cause license issues as the companies own the copyright on their logos. I'm not entirely sure given that these are adaptions of those logos, but it might be worth thinking about.

- Yeah indeed would have to remove the twitter untill clear information. I think i naivly thought that since https://materialdesignicons.com was hosting them they was already approved for use but i see that might be wrong to assume.

andydotxyz:
There was a chat in our slack group yesterday about this and we were not sure if the chart icon is useful to enough use-cases to include it in the main repo.
We did like the addition of account, but wondered if it should be paired with a log-out as those are often used together.
Does that sound sensible?
The reason we try to keep the icons to a small number is that each additional one grows all applications a little, but any developer can include their icons - so the ones we provide are mostly convenience rather than features.

xines:
> There was a chat in our slack group yesterday about this and we were not sure if the chart icon is useful to enough use-cases to include it in the main repo.

- Agree since theres no chart module yet it seems a bit empty to add.

> We did like the addition of account, but wondered if it should be paired with a log-out as those are often used together.
> Does that sound sensible?

- Totally, but wouldn't that also kinda require a log-in variant then to match, either way i found 2 different logout svgs i need opinions on them. 😄 

1. https://materialdesignicons.com/icon/logout-variant
2. https://fonts.google.com/icons?selected=Material%20Icons%3Alogout

@andydotxyz
@Jacalz

andydotxyz:
The second link is from a font so the size/aspect/corners don't match the other icons, making the first of the two more in-keeping.
I guess login could be useful as well, for some reason I'd associated that with the account icon or a full dialog instead of a button / menu item.

stuartmscott:
I like these;
- https://materialdesignicons.com/icon/account
- https://materialdesignicons.com/icon/login
- https://materialdesignicons.com/icon/logout

However there are also variants with another dev may prefer;
- https://materialdesignicons.com/icon/account-box
- https://materialdesignicons.com/icon/account-circle
- https://materialdesignicons.com/icon/home-account
- https://materialdesignicons.com/icon/shield-account

Or depending on what the app does;
- https://materialdesignicons.com/icon/book-account
- https://materialdesignicons.com/icon/briefcase-account
- https://materialdesignicons.com/icon/calendar-account
- https://materialdesignicons.com/icon/camera-account

Or for an enterprise app;
- https://materialdesignicons.com/icon/account-hard-hat
- https://materialdesignicons.com/icon/account-tie

So it seems to depend a lot on the app and the type of account. Since Fyne cannot please everyone, perhaps it is better not to include these icons and instead leave the decision to the dev.

xines:
Alright, so 1 or 2? @andydotxyz 

GitHub Logo

Jacalz:
I would personally vote for 2.

andydotxyz:
> I would personally vote for 2.

I agree that 2 matches better, I'm not super sure that the login icon is clear in it's meaning, but in the context of an account menu etc it should be OK.

xines:
> I'm not super sure that the login icon is clear in it's meaning, but in the context of an account menu etc it should be OK.

- Yeah totally same opinion, but guess it will do for now.

andydotxyz (approved):
Great, thanks

andydotxyz (changes requested):
Oh no I forgot - please change "Since: 2.0.2" to "Since: 2.1".
For one thing 2.0.2 has shipped, and also we don't add new API in patch releases so it will have to wait on develop until 2.1

xines:
Done 😄

andydotxyz (approved):
Great, thanks

PR #2153, opened 2021-04-10
[fyne_demo] AnimationCurveItem boxes should be recolored when the theme changes

fpabl0:
### Description:

AnimationCurveItem boxes should be recolored when the theme changes.

### Checklist:

~~- [ ] Tests included.~~
- [x] Tests all pass.


andydotxyz (approved):
The ExtendBaseWidget is probably not required in CreateRenderer as this is a private widget and we always call the constructor. But no harm including it just to be safe :)

PR #2155, opened 2021-04-10
Unbreak CI by getting rid of golangci-lint

Jacalz:
### Description:

We have recently encountered golangci-lint being unreliable. This PR moves us back to the old style of CI we had and also incorporates a few more improvements on top of it (see commits). It also removes the old `.travis.yml` file that was laying around and removes the commented out modules check because Go 1.16 complains about incorrect modules either way (we got that for free) :slightly_smiling_face:

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


.github/workflows/static_analysis.yml
```diff
@@ -6,22 +6,32 @@ jobs:
...
-        version: v1.37
-        skip-go-installation: true
+    - name: Vet
+      run: go vet -tags ci ./...
+
+    - name: Goimports
+      run: test -z $(goimports -e -d . | tee /dev/stderr)
+
+    - name: Gocyclo
+      run: gocyclo -over 50 .
```
fpabl0, line 31: I don't know anything about CI, but shouldn't this be 30 to match the current behavior?

Jacalz: You are absolutely correct. I copied over the CI from FyneDesk thinking it would be updated 😅

andydotxyz:
Probably should remove the .golangci file as well?

Jacalz:
> Probably should remove the .golangci file as well?

Indeed. I missed that.

Jacalz:
I have held off upgrading the platform tests as version 1.6.09 seems to fail, see https://github.com/WillAbides/setup-go-faster/issues/13.

andydotxyz (approved):
Thanks, back to (slightly slower) stability :)

Issue #2152, opened 2021-04-09
Clicking cancel on the ShowFolderOpen dialog gets segfault


lazyhacker:
### Describe the bug:

Clicking the "cancel" button on the ShowFolderOpen dialog results in the app crashing with:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x81daef]
```
Selecting a folder and click "Open" works with no crashes.

### To Reproduce:
Steps to reproduce the behaviour:
1.  Click on folder icon to bring up the folder dialog.
2.  Click cancel.

### Example code:
The code is at:

https://github.com/lazyhacker/getgo/blob/d09a81254db9e9a6725ccac3cdb1d2855b0ccb83/internal/guimain/fyne.go#L115

### Device (please complete the following information):
 - **OS:** Linux and Windows
 - **Version:** Fedora 32, Windows 10
 - **Go version:** 1.16.3
 - **Fyne version:** 2.0.2

Jacalz:
I am pretty sure that this is a duplicate of https://github.com/fyne-io/fyne/issues/2145. You will find the solution to your crash in https://github.com/fyne-io/fyne/issues/2145#issuecomment-814850779.

However, two very similar issues in a row signals to me that we might have to update the API docs with a note about this.

andydotxyz:
Maybe we need to review this "feature"?
It was designed before `OnClosed` was added so there was no way for it to be known.
However consider the confirmation callback `ffunc(bool)` - would then that either get called with `true` or not called?
Considering that maybe improving the docs is the best solution...

lazyhacker:
Thanks!   It was the same reason.

I agree that clarifying the documentation would help.  What I missed was that both "cancel" and "open" button on the dialog will trigger the callback except that "cancel" will pass nil values.  I wasn't expecting this and assumed that "cancel" simply closes the dialog.

I should've check for nil regardless and it could've helped me realize what was happening even if I didn't know that cancel was triggering the callback, but having the docs explain this behavior would help newbies to Fyne (such as myself) who are still trying to learn the different behaviors of things.

Jacalz:
Documentation was added to make this more clear. Closing.

PR #2156, opened 2021-04-11
Added additional documentation and example code snippet.

lazyhacker:
### Description:

Adding additional comment that the callback function is triggered for both Open and Cancel.

Fixes #(issue)

#2152

### Checklist:

- [X] Tests included.
- [X ] Lint and formatter run with no errors.
- [X] Tests all pass.


dialog/folder.go
```diff
@@ -19,9 +19,27 @@ func NewFolderOpen(callback func(fyne.ListableURI, error), parent fyne.Window) *
...
+//               func (uri fyne.ListableURI, err error) {
+//                   // Cancel will pass a nil value.
+//                   if uri == nil || err != nil {
+//                       return
+//                   }
+//                   fmt.Println("URI", uri)
+//               }, w)
+//       }
+//    )
+//
```
Jacalz, line 42: I really appreciate the effort, but you are not supposed to add code examples to the API docs in this way. See this link for an example https://blog.golang.org/examples on how it should be done for interactive examples.

However, we do have the `fyne_demo` application so I don't know if we really should add this kind of example in (someone else will have to comment on that) because adding it to one function would mean that we open up the possibility of adding to much more functions in the future.

dialog/folder.go
```diff
@@ -19,9 +19,27 @@ func NewFolderOpen(callback func(fyne.ListableURI, error), parent fyne.Window) *
 	return dialog
 }
 
-// ShowFolderOpen creates and shows a file dialog allowing the user to choose a folder to open.
+// ShowFolderOpen creates and shows a file dialog allowing the user to choose a
+// folder to open. It takes a callback function that is called whenever the open
+// or cancel buttons are tapped.
```
Jacalz, line 24: Hmm, I think we might want to word this a bit differently. Technically it takes more than just a callback function.
```suggestion
// folder to open. The callback function will run when the dialog closes.
// The URI will be nil when nothing is selected.
```
I'm not entirely sure about my own suggestion here, but it's just an example (and I couldn't pay attention to wrapping as I wrote it).

Jacalz (changes requested):
I very much appreciate the PR. There are a few changes that I would suggest, but thanks for taking on this effort 🙂

This kind of documentation should be added to the other file dialogs as fell (Save and Open) as they will be called with nil values as well.

dialog/folder.go
lazyhacker: The interactive examples are normally able to be executed as a test and in the browser and I'm not sure how to get something like that to work in this case where it's rendering a GUI dialog box.  That's why I just put in in a code snippet as an example similar to what Go's doc do for similar cases (e.g. https://golang.org/pkg/net/).

It's perfectly alright with me if this is not the right fit for the project so please feel free to reject the PR. 

If the `fyne_demo` has the illustrative snippet of code, my suggestion would be to add a direct reference to it somehow in the doc to make it easy to find.  A demo app's source code (especially one as complete as the demo) isn't as easily navigable IMHO so help to get to the right part would help.

dialog/folder.go
lazyhacker: I'll hold off on any changes until it's decided whether this is the right way to document this, but I feel that to say "close" might still be ambiguous.  

Although "cancel" does close the dialog, I wonder if someone might still miss that "cancel" will still trigger the callback.  Also, from the dialog itself, it doesn't seem like nothing is selected.  The UI makes it looks like if home directory is the default folder.

That's why I tried to be explicit, but it does come with the cost of being more verbose.

dialog/folder.go
andydotxyz: Fine code can be rendered in the playground in theory, but there is a 5sec timeout that I am sure means we can’t get the app compiled (you could check out the tools/playground package though).
With WASM aiming for 2.1 this may get much better :)

dialog/folder.go
andydotxyz: I think that @Jacalz suggestion here is good, it is quite explicit about the nil which seems to be the main cause for confusion. Of course it should be possible to combine both comments :).

dialog/folder.go
andydotxyz: I am nervous about adding code examples like this (i.e. entire applications) because it includes so much outside the scope of this function. Linking to the fyne_demo project could be a nice way to do it, though those links may go stale as well.
I liked what was done with the clarification in the paragraph of text above the example, but feel that large examples like this would be better placed in the tutorials at https://developer.fyne.io.

andydotxyz (changes requested):
The clarification of usage is a great addition, but let's not go down the path of what could be an application per-function

dialog/folder.go
```diff
@@ -19,7 +19,10 @@ func NewFolderOpen(callback func(fyne.ListableURI, error), parent fyne.Window) *
 	return dialog
 }
 
-// ShowFolderOpen creates and shows a file dialog allowing the user to choose a folder to open.
+// ShowFolderOpen creates and shows a file dialog allowing the user to choose a
+// folder to open. The callback function will run when the dialog closes.  The
+// URI will be nil when nothing is selected or cancelled.
```
andydotxyz, line 24: The latest change made this slightly ambiguous sorry - "nothing is selected or cancelled".
I think "the user cancelled or nothing is selected" is clearer.

andydotxyz (changes requested):

Jacalz (changes requested):
I agree with @andydotxyz on the wording. Also please update the other file/folder dialogs with the same note.

andydotxyz (approved):
Thank you so much for this clarification

andydotxyz:
I'm so sorry that this dropped off my watch list.
It seems that GitHub stopped auto-running CI for new contributors and so I didn't get the usual notifications.

andydotxyz:
I think this is waiting for your re-review @Jacalz if you could check it out :)

dialog/file.go
```diff
@@ -559,21 +559,31 @@ func (f *FileDialog) SetFileName(fileName string) {
...
+//
 // The dialog will appear over the window specified when Show() is called.
 func NewFileSave(callback func(fyne.URIWriteCloser, error), parent fyne.Window) *FileDialog {
 	dialog := &FileDialog{callback: callback, parent: parent, save: true}
 	return dialog
 }
 
-// ShowFileOpen creates and shows a file dialog allowing the user to choose a file to open.
+// ShowFileOpen creates and shows a file dialog allowing the user to choose a
+// file to open. The callback function will run when the dialog closes.  The URI
```
Jacalz, line 584:
```suggestion
// file to open. The callback function will run when the dialog closes. The URI
```

It's grammatically incorrect with double spaces. There are a few more places where this occurs but I just made one comment about it.

Jacalz (changes requested):
Sorry, I had forgotten about this. I had a minor thing that I found when re-reviewing this. Please remove the double spaces 🙂

dialog/file.go
andydotxyz: There is actually a school of thinking that says double space is grammatically correct. However I agree that we have settled on single spacing.

Jacalz: I did not know that. It is very, very forbidden in Swedish at least 😅

Jacalz (approved):
Looks good to me. Sorry for the long wait, but I never got a request for a re-review.

Issue #776, opened 2020-03-27
Can maximize window?


mondy:
I want to maximize the window from the program.
And I want to know if the window is maximized.
Are these possible?
Or do you have any plans to implement it?

andydotxyz:
Maximize has proved complicated, but it could be considered - mostly window based operations are left up to the user and/or desktop environment. i.e. your desktop could remember maximised state, so should it be overridden?

Window.SetFullScreen() is an option at the moment - is that any help?

mondy:
Thank you for the reply.
I want to maximize window on start application. 
But I don't remember my desktop.

I think that maximized window should be remembered by each application.

> Window.SetFullScreen() is an option at the moment - is that any help?

Thanks.
But I know Window.SetFullScreen. However, it is different from the maximize window.

My environment is as follows.

* OS: Ubuntu 18.04
* go version 1.14
* fyne.io version v1.2.3

andydotxyz:
Remembering position and maximised state is typically the job of your desktop environment (though it has more detail in #554). Modern desktops seeem to be moving to "fullscreen" for truly maximized and then various different types of "zoom" (used to be called maximize) such as "left half" etc.

So it becomes difficult to see what a user expectation is here. Can you expand on your use-case?

mondy:
I see... I understand that it's a desktop environment job.
But what I'm looking for is not a "remember feature" but a "getter/setter feature".
For example, you may want to always maximize and start regardless of the state of the last time you quit.
For example, you might want to maximize it using a button with a widget.
"remember feature" is not necessary for fyne, but I think it should be possible to get and set the maximization state so that developers using fyne can implement it.
And it's also true that some environments don't remember position and maximised state such as my desktop environment (Ubuntu/GNOME).
I looked for a way to remember the location and maximized state in Ubuntu/GMOME, but could not find it.
I don't want to give up Ubuntu/GNOME for this.

The topic is different, but there is a possible use case that requires a getter/setter for the window position.
Think of a painting tool like GIMP, for example. Tools, layers, etc. can be displayed as a separate window. You will need to finely determine the window size and window position and align the windows. For this, we need a window position getter/setter.

In summary, I hope that the following features will be added
* window maximization setter/getter (what I need)
* window minimization setter/getter (implemented as above)
* window position setter/getter (possible use cases that need it)

andydotxyz:
I think it would be a very, very bad idea for a window to be able to start iconified. 
[...]

PR #2160, opened 2021-04-11
Provide glfw.Window as context

softkot:
That small change make possible to restore the fyne window from minimized state with this code snippet:

```
type WithContext interface {
	RunWithContext(f func())
	RescaleContext()
	Context() interface{}
}

func ShowWindow(wnd fyne.Window) {
	if ctx, ok := wnd.(WithContext); ok {
		contextWnd := ctx.Context()
		if gwnd, ok := contextWnd.(*glfw.Window); ok && gwnd != nil {
			gwnd.Restore()
		}
	}
	wnd.Show()
	wnd.RequestFocus()
}
```

At first attempt i was try to acheive the same with RunWithContext and glfw.GetCurrentContext().Restore() but it did not works  and sometimes causeing and error that context is already active.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -1231,7 +1231,7 @@ func (w *window) rescaleOnMain() {
 }
 
 func (w *window) Context() interface{} {
-	return nil
+	return w.viewport
 }
 
 // Use this method to queue up a callback that handles an event. This ensures


andydotxyz:
I don’t think that WithContext is part of our public API.
We don’t expose the internal dependencies (glfw, and others) and would not recommend you rely on internal state as there is no guarantee that a window is backed by GLFW.

softkot:
You are right, that will work only for glfw backed windows, and the code snipped knows that.
Is there any other way to fine control window attributes on the known backend?

andydotxyz:
As we do not expose these handles, nor promise a “known backend” it is not possible to gain the handles. If there is functionality missing then it should be added to the public API, assuming it meets with a good cross-platform app use-case

softkot:
i think the discussion is also related to #776

andydotxyz:
There was a longer discussion in the issue you linked, though I have now closed it.
Windows that the OS/User have iconified should remain there until they decide that it should return (same for maximized, in the linked ticket).
As we move to expose more of the application lifecycle which has been popularised in mobile apps we will be focusing on the user being in control, not the applications.

PR #2161, opened 2021-04-11
add go1.16 embed file

jinxing3114:
### Description:

Add go1.16 embed file function, compatible with existing functions
```go
package main

import (
    "embed"
    "fyne.io/fyne/v2"
    "fyne.io/fyne/v2/app"
)

//go:embed logo.png
var res embed.FS

func init() {
    fyne.SetEmbedResource(res)
}

func main() {
    a := app.New()
    logo, err := fyne.LoadResourceFromPath("-logo.png") // - Indicates that the file is an embed file
    if err != nil {
        log.Fatalln(err)
    }
    a.SetIcon(logo)
}
```

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module fyne.io/fyne/v2
 
-go 1.12
+go 1.16
 
 require (
 	github.com/Kodeworks/golang-image-ico v0.0.0-20141118225523-73f0f4cfade9
--- a/resource.go
+++ b/resource.go
@@ -1,10 +1,14 @@
 package fyne
 
 import (
+	"embed"
+	"errors"
 	"io/ioutil"
 	"net/http"
 	"path/filepath"
 )
+
+var embedRes embed.FS
 
 // Resource represents a single binary resource, such as an image or font.
 // A resource has an identifying name and byte array content.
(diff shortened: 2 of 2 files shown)


andydotxyz:
Sorry, we currently support back to 1.12, hence why the tests are now failing.
We can’t rely on 1.16 until that is our lowest supported version.

DiskGetor:
i have same problem,resource file is bigger for show in goland. if develop branch creat some api for this,i it will be better.

Jacalz:
You should be able to do it manually. Just assign the embedded file to a []byte variable and set that variable as the resource for the StaticContent field in a fyne.StaticResource. Don’t forget tot set the name as well.

andydotxyz:
Not even the `develop` branch will be moving to 1.16 as it's support will not be wide enough at the time of our next release. We are currently exploring 1.14 or 1.15 as possible lowest required version for 2.1, it depends on what operating systems support by that time (Debian stable is a good yard stick).

Jacalz:
Indeed. I'd suggest using the method that I posted, in the meantime, if you need to use the embed package. For most use cases the fyne bundle command should be useful enough.

PR #2162, opened 2021-04-11
Save dialog with filename for Android

howeyc:
### Description:
Save dialog on android can have the filename populated. This pull request uses the *initialFileName* which gets set when calling **SetFileName()** on the fyne dialog.

Note: The changes against mobile repository need to be merged first, and then the go.mod here needs to be bumped, and vendor needs to be updated. The GoNativeActivity.java is here as well, as that's the file I used to run gendex.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
Thanks, you can update this PR for `650a313` version of mobile in the go.mod

howeyc:
I updated to use latest mobile version, should be good now.

andydotxyz (approved):
Great, thanks

PR #2168, opened 2021-04-14
Add UnselectAll to collection widgets

andydotxyz:
Add UnselectAll to the collection widgets.
The documentation elements aimed to match the wording of the original author hence the variations.

See #1395

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -144,6 +144,8 @@ func (l *List) Unselect(id ListItemID) {
 }
 
 // UnselectAll removes all items from the selection.
+//
+// Since: 2.1
 func (l *List) UnselectAll() {
 	if len(l.selected) == 0 {
 		return
--- a/widget/list.go
+++ b/widget/list.go
@@ -132,7 +132,7 @@ func (l *List) Select(id ListItemID) {
 
 // Unselect removes the item identified by the given ID from the selection.
 func (l *List) Unselect(id ListItemID) {
-	if len(l.selected) == 0 {
+	if len(l.selected) == 0 || l.selected[0] != id {
 		return
 	}
 
--- a/widget/list.go
+++ b/widget/list.go
@@ -143,6 +143,22 @@ func (l *List) Unselect(id ListItemID) {
 	}
 }
 
+// UnselectAll removes all items from the selection.
+func (l *List) UnselectAll() {
+	if len(l.selected) == 0 {
+		return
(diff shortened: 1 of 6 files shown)


fpabl0:
Should this PR fix the Unselect method?
```go
func (l *List) Unselect(id ListItemID) {
	if len(l.selected) == 0 {
		return
	}
	l.selected = nil
	l.Refresh()
	if f := l.OnUnselected; f != nil {
		f(id)
	}
}
```
Because it already unselects all no matters what ID is passed.

andydotxyz:
> Should this PR fix the Unselect method?

A good catch thanks, I had not noticed that list was implemented this way.

fpabl0 (commented):
> A good catch thanks, I had not noticed that list was implemented this way.

You're welcome. 

I know that currently only one item is possible to be selected, but shouldn't the Unselect function be implemented as if it were multiple selection? Just like UnselectAll that is aware of multiple selected options.

Also, I wonder if this change in `Unselect` will be considered as a breaking change. Because there could be some users, that used this function as it were `UnselectAll` 🤔

andydotxyz:
> shouldn't the Unselect function be implemented as if it were multiple selection?

I don't think so. If you call `Select(4)` then `Select(5)` 4 must be unselected.

> Also, I wonder if this change in Unselect will be considered as a breaking change

Only in the way that every bug fix is a breaking change. Nobody should have relied on `Unselect` ignoring it's parametter

fpabl0:
> I don't think so. If you call Select(4) then Select(5) 4 must be unselected.

Yes, but I am talking about `Unselect(..)` passing a valid `id` will still unselect all. This is correct now because there is single selection, but it would not be right when multiple selection were landed.

> Only in the way that every bug fix is a breaking change. Nobody should have relied on Unselect ignoring it's parameter.

Fair enough :)

andydotxyz:
> Yes, but I am talking about Unselect(..) passing a valid id will still unselect all. This is correct now because there is single selection, but it would not be right when multiple selection were landed.

"This is correct now" may be all we need at this time?
In the future, when we add a "Multiselect" mode then yes the code will need to change, but so will a lot more.

Either way I don't think the `Select` functions should be changed by adding `UnselectAll`.

fpabl0 (approved):

PR #2169, opened 2021-04-14
Some smaller optimizations of SelectEntry

Jacalz:
### Description:

This does two smaller things. Firstly, it avoids creating and calculating a padding size on each iteration of the loop. This is not a huge speedup be any means, but good to add given how often MinSize seems to be called. Secondly, SetupOptions now preallocates the slice of menu items beforehand. My testing in fyne_demo indicates up to 30-40% improvement in speed there (from about 1200ns to 770ns). It was pretty fast already, but this should, especially for larger amounts of items, result in a good improvement.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/select_entry.go
+++ b/widget/select_entry.go
@@ -80,10 +80,10 @@ func (e *SelectEntry) Resize(size fyne.Size) {
 // SetOptions sets the options the user might select from.
 func (e *SelectEntry) SetOptions(options []string) {
 	e.options = options
-	var items []*fyne.MenuItem
-	for _, option := range options {
+	items := make([]*fyne.MenuItem, len(options))
+	for i, option := range options {
 		option := option // capture
-		items = append(items, fyne.NewMenuItem(option, func() { e.SetText(option) }))
+		items[i] = fyne.NewMenuItem(option, func() { e.SetText(option) })
 	}
 	e.dropDown = fyne.NewMenu("", items...)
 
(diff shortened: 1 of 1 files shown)


widget/select_entry.go
```diff
@@ -80,10 +80,10 @@ func (e *SelectEntry) Resize(size fyne.Size) {
 // SetOptions sets the options the user might select from.
 func (e *SelectEntry) SetOptions(options []string) {
 	e.options = options
-	var items []*fyne.MenuItem
-	for _, option := range options {
+	items := make([]*fyne.MenuItem, len(options))
+	for i, option := range options {
 		option := option // capture
```
andydotxyz, line 85: With the addition of `i` I think you can remove the capture and use `options[i]` in place of `option`

Jacalz: Hmm, not quite. It looks like the captures are still needed when doing it that way as test are failing otherwise.

andydotxyz: Did you change the second occurrence of `option`? the first one won't matter but the one in `SetText` is what I meant.

Jacalz: Yes. I removed the item that we get from the range and replaced both occurrences with options[i]. This caused the tests to fail.

andydotxyz: Oh yes I see now, the `i` would have to have been captured. Apologies for the noise!

andydotxyz (approved):

widget/select_entry.go
Jacalz: No worries.

PR #2171, opened 2021-04-14
Attempt to share common code between desktop and mobile drivers

fpabl0:
### Description:

Attempt to share common code between desktop and mobile drivers.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/internal/driver/common/driver.go
+++ b/internal/driver/common/driver.go
@@ -1,0 +1,19 @@
+package common
+
+import (
+	"sync"
+
+	"fyne.io/fyne/v2"
+)
+
+var (
+	canvasMutex sync.RWMutex
+	canvases    = make(map[fyne.CanvasObject]fyne.Canvas)
+)
+
+// CanvasForObject returns the canvas for the specified object.
+func CanvasForObject(obj fyne.CanvasObject) fyne.Canvas {
+	canvasMutex.RLock()
+	defer canvasMutex.RUnlock()
+	return canvases[obj]
+}
--- a/internal/driver/common/canvas.go
+++ b/internal/driver/common/canvas.go
@@ -1,0 +1,513 @@
+package common
+
+import (
+	"sync"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/internal"
+	"fyne.io/fyne/v2/internal/app"
+	"fyne.io/fyne/v2/internal/cache"
+	"fyne.io/fyne/v2/internal/driver"
+	"fyne.io/fyne/v2/internal/painter/gl"
+)
+
+// SizeableCanvas defines a canvas with size related functions.
+type SizeableCanvas interface {
+	fyne.Canvas
+	Size() fyne.Size
+	Resize(fyne.Size)
+	MinSize() fyne.Size
+}
+
+// Canvas defines common canvas implementation.
+type Canvas struct {
+	sync.RWMutex
+
+	OnFocus   func(obj fyne.Focusable)
+	OnUnfocus func()
+
+	impl SizeableCanvas
+
+	contentFocusMgr *app.FocusManager
+	menuFocusMgr    *app.FocusManager
+	overlays        *overlayStack
+
+	shortcut fyne.ShortcutHandler
+
+	painter gl.Painter
+
+	dirty                                  bool
+	dirtyMutex                             sync.Mutex
+	refreshQueue                           chan fyne.CanvasObject
+	mWindowHeadTree, contentTree, menuTree *renderCacheTree
+}
+
+// Initialize initializes the canvas.
+func (c *Canvas) Initialize(impl SizeableCanvas, refreshQueueLen int, onOverlayChanged func()) {
+	c.impl = impl
+	c.refreshQueue = make(chan fyne.CanvasObject, refreshQueueLen)
+	c.overlays = &overlayStack{
+		OverlayStack: internal.OverlayStack{
+			OnChange: onOverlayChanged,
+			Canvas:   impl,
+		},
+	}
+}
+
+// ===============================================================
+// Shortcut related
+// ===============================================================
+
+// AddShortcut adds a shortcut to the canvas.
+func (c *Canvas) AddShortcut(shortcut fyne.Shortcut, handler func(shortcut fyne.Shortcut)) {
+	c.shortcut.AddShortcut(shortcut, handler)
+}
+
+// RemoveShortcut removes a shortcut from the canvas.
+func (c *Canvas) RemoveShortcut(shortcut fyne.Shortcut) {
+	c.shortcut.RemoveShortcut(shortcut)
+}
+
+// ===============================================================
+// Object tree related
+// ===============================================================
+
+// EnsureMinSize ensure canvas min size.
+//
+// This function uses lock.
+func (c *Canvas) EnsureMinSize() bool {
+	var lastParent fyne.CanvasObject
+
+	windowNeedsMinSizeUpdate := false
+	ensureMinSize := func(node *renderCacheNode) {
+		obj := node.obj
+		canvasMutex.Lock()
+		canvases[obj] = c.impl
+		canvasMutex.Unlock()
+
+		if !obj.Visible() {
+			return
+		}
+		minSize := obj.MinSize()
+		minSizeChanged := node.minSize != minSize
+		if minSizeChanged {
+			objToLayout := obj
+			node.minSize = minSize
+			if node.parent != nil {
+				objToLayout = node.parent.obj
+			} else {
+				windowNeedsMinSizeUpdate = true
+				size := obj.Size()
+				expectedSize := minSize.Max(size)
+				csize := c.impl.Size()
+				if expectedSize != size && size != csize {
+					objToLayout = nil
+					obj.Resize(expectedSize)
+				}
+			}
+
+			if objToLayout != lastParent {
+				updateLayout(lastParent)
+				lastParent = objToLayout
+			}
+		}
+	}
+	c.WalkTrees(nil, ensureMinSize)
+
+	min := c.impl.MinSize()
+	csize := c.impl.Size()
+	shouldResize := windowNeedsMinSizeUpdate && (csize.Width < min.Width || csize.Height < min.Height)
+	if shouldResize {
+		c.impl.Resize(csize.Max(min))
+	}
+
+	if lastParent != nil {
+		c.RLock()
+		updateLayout(lastParent)
+		c.RUnlock()
+	}
+	return windowNeedsMinSizeUpdate
+}
+
+// ObjectTrees return canvas object trees.
+//
+// This function uses lock.
+func (c *Canvas) ObjectTrees() []fyne.CanvasObject {
+	c.RLock()
+	content := c.contentTree.root.obj
+	menu := c.menuTree.root.obj
+	c.RUnlock()
+	trees := make([]fyne.CanvasObject, 0, len(c.Overlays().List())+2)
+	trees = append(trees, content)
+	if menu != nil {
+		trees = append(trees, menu)
+	}
+	trees = append(trees, c.Overlays().List()...)
+	return trees
+}
+
+// SetContentTreeAndFocusMgr sets content tree and focus manager.
+//
+// This function does not use the canvas lock.
+func (c *Canvas) SetContentTreeAndFocusMgr(content fyne.CanvasObject) {
+	c.contentTree = &renderCacheTree{root: &renderCacheNode{obj: content}}
+	var focused fyne.Focusable
+	if c.contentFocusMgr != nil {
+		focused = c.contentFocusMgr.Focused() // keep old focus if possible
+	}
+	c.contentFocusMgr = app.NewFocusManager(content)
+	if focused != nil {
+		c.contentFocusMgr.Focus(focused)
+	}
+}
+
+// SetMenuTreeAndFocusMgr sets menu tree and focus manager.
+//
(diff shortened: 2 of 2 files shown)


fpabl0:
@andydotxyz could you check if you are agree with these changes? I only need to add queueEvent and runOnDraw methods, but first would like to have feedback.

andydotxyz:
Yes it seems like this is going in the right direction, thanks. Be careful with `runOnDraw` because mobile doesn't have a dedicated draw goroutine.

When it comes to the new code, however, please stay within our guidelines for ordering methods alphabetically instead of grouped by functionality (many times when you have a large comment block it breaks this order).

fpabl0:
> Yes it seems like this is going in the right direction, thanks. Be careful with runOnDraw because mobile doesn't have a dedicated draw goroutine.

Yes, I am aware of it.

> When it comes to the new code, however, please stay within our guidelines for ordering methods alphabetically instead of grouped by functionality (many times when you have a large comment block it breaks this order).

Yes, I have taken it in account, I will order the functions when the PR is ready, it is just like this for convenience now :)

internal/driver/glfw/window.go
```diff
@@ -435,16 +435,16 @@ func (w *window) Close() {
 	w.viewLock.Unlock()
 	w.viewport.SetShouldClose(true)
 
-	w.canvas.walkTrees(nil, func(node *renderCacheNode) {
-		switch co := node.obj.(type) {
+	w.canvas.WalkTrees(nil, func(node *common.RenderCacheNode) {
+		switch co := node.Obj().(type) {
```
Jacalz, line 439: I think you'd be better off with a simple if here.

internal/driver/glfw/window.go
```diff
@@ -1107,10 +1095,10 @@ func (w *window) capturesTab(keyName fyne.KeyName, modifier desktop.Modifier) bo
 		if ent, ok := w.canvas.Focused().(fyne.Tabbable); ok && !ent.AcceptsTab() {
 			switch modifier {
 			case 0:
-				w.queueEvent(func() { w.canvas.FocusNext() })
+				w.QueueEvent(func() { w.canvas.FocusNext() })
```
Jacalz, line 1098: You can pass the function directly instead of wrapping it.

internal/driver/glfw/window.go
```diff
@@ -1107,10 +1095,10 @@ func (w *window) capturesTab(keyName fyne.KeyName, modifier desktop.Modifier) bo
...
 				return false
 			case desktop.ShiftModifier:
-				w.queueEvent(func() { w.canvas.FocusPrevious() })
+				w.QueueEvent(func() { w.canvas.FocusPrevious() })
```
Jacalz, line 1101: You can pass the function directly instead of wrapping it.

internal/driver/gomobile/driver.go
```diff
@@ -253,58 +255,58 @@ func (d *mobileDriver) paintWindow(window fyne.Window, size fyne.Size) {
...
+		w.QueueEvent(func() { wid.Tapped(ev) })
 	}, func(wid fyne.SecondaryTappable, ev *fyne.PointEvent) {
-		go wid.TappedSecondary(ev)
+		w.QueueEvent(func() { wid.TappedSecondary(ev) })
 	}, func(wid fyne.DoubleTappable, ev *fyne.PointEvent) {
-		go wid.DoubleTapped(ev)
+		w.QueueEvent(func() { wid.DoubleTapped(ev) })
 	}, func(wid fyne.Draggable) {
-		go wid.DragEnd()
+		w.QueueEvent(func() { wid.DragEnd() })
```
Jacalz, line 309: You can pass the function directly instead of wrapping it.

internal/driver/gomobile/window.go
```diff
@@ -155,13 +157,19 @@ func (w *window) Close() {
 		d.windows = append(d.windows[:pos], d.windows[pos+1:]...)
 	}
 
-	w.canvas.walkTree(nil, func(obj, _ fyne.CanvasObject) {
-		switch co := obj.(type) {
+	w.canvas.WalkTrees(nil, func(node *common.RenderCacheNode) {
+		switch co := node.Obj().(type) {
```
Jacalz, line 161: I think you'd be better of with an if here.

Jacalz (changes requested):

internal/driver/glfw/window.go
fpabl0: Absolutely :) thank you. I wasn't sure if I should change it, because it seems like it expected to handle other kind of CanvasObjects 🤔

internal/driver/glfw/window.go
fpabl0: You are right. I will update :)

fpabl0: Done! :)

internal/driver/glfw/window.go
fpabl0: Updated! :)

internal/driver/common/canvas_test.go
```diff
@@ -0,0 +1,375 @@
...
+		if pd.obj != node.obj {
+			panic("node cache does not match node obj - nodes should not be reused for different objects")
+		}
+		pd.lastAfterCallIndex = index
+		node.painterData = pd
+	}
+
+	//
+	// test that first walk calls the hooks correctly
+	//
```
Jacalz, line 63: Suspicious comment formatting?

internal/driver/common/driver.go
```diff
@@ -0,0 +1,19 @@
...
+var (
+	canvasMutex sync.RWMutex
+	canvases    = make(map[fyne.CanvasObject]fyne.Canvas)
+)
+
+// CanvasForObject returns the canvas for the specified object.
+func CanvasForObject(obj fyne.CanvasObject) fyne.Canvas {
+	canvasMutex.RLock()
+	defer canvasMutex.RUnlock()
+	return canvases[obj]
```
Jacalz, line 18: I am no master when it comes to mutexes, but are these locks really going to work? Wouldn't we have to assign `canvases[obj]` to a new variable in-between?

Jacalz (commented):
Looks pretty good in my opinion. Very nice to see a common codebase for both driver. Should make a lot of new features and optimizations easier in the future. Just had two small comments.

internal/driver/common/canvas_test.go
fpabl0: Not sure, it seems that the comment was written like that to clearly divide the tests?

internal/driver/common/driver.go
fpabl0: Hmm, not sure what you mean 🤔. The mutex is there, to ensure we read atomically the canvases map (ensuring we are not writing and reading it at the same time). I think, assigning it to a variable does not make any difference (when you are returning the value, you are already creating a new variable to assign it) 🤔

Jacalz: I don't know. I thought defers ran before returning the value, but I might be very wrong about that.

fpabl0: Oh I see, `defers` ran after returning the value 😄:
https://play.golang.org/p/AhQBXxb1HAH

Jacalz: Ah, sorry then 😅

fpabl0: No worries :)

Jacalz (dismissed):
@andydotxyz Probably has more things to say, but this looks good from my perspective.

fpabl0:
@andydotxyz is this ok to be merged?

andydotxyz:
I think last week we agreed to hold off until the release was merged due to the potential conflicts.
However I may be confusing this with another PR discussion.

Jacalz:
I remember seeing the same thing in the meeting notes afterwards. Just popped my approve as it was fine from my end.

fpabl0:
> I think last week we agreed to hold off until the release was merged due to the potential conflicts.

Oh yes, sorry I forgot it completely

internal/driver/gomobile/driver.go
```diff
@@ -200,15 +204,13 @@ func (d *mobileDriver) Run() {
 					}
 					if !canvas.inited {
 						canvas.inited = true
-						canvas.painter.Init() // we cannot init until the context is set above
+						canvas.Painter().Init() // we cannot init until the context is set above
 					}
 
-					if d.freeDirtyTextures(canvas) {
+					if canvas.FreeDirtyTextures() {
```
fpabl0, line 210: Is it ok if I change this part to:
```go
if canvas.IsDirty() {
    canvas.FreeDirtyTextures()
    ....
}
```
?
It would be useful if we consider the dirty flag to solve #2205 and also it would be consistent with the way how this is handled in the desktop version.

internal/driver/gomobile/driver.go
```diff
@@ -85,6 +88,7 @@ func (d *mobileDriver) CanvasForObject(fyne.CanvasObject) fyne.Canvas {
 	}
 
 	// TODO figure out how we handle multiple windows...
+	// TODO should use common.CanvasForObject as in desktop?
 	return d.currentWindow().Canvas()
```
fpabl0, line 92: Any suggestion about this?

internal/driver/gomobile/driver.go
andydotxyz: I would think that would be fine. The `freeDiretyTextures` returning a bool was introduced in this PR anyway.

internal/driver/gomobile/driver.go
andydotxyz: I'm not sure what you mean. I agree with that the TODO states, it would fix an issue in mobile where the assumption about always being the top window is false.

internal/driver/gomobile/driver.go
fpabl0: > I would think that would be fine.

Great! I will update.

> The freeDiretyTextures returning a bool was introduced in this PR anyway.

Not really, that was the mobile driver implementation, indeed, you can see it in the diff.

internal/driver/gomobile/driver.go
fpabl0: > I agree with that the TODO states, it would fix an issue in mobile where the assumption about always being the top window is false.

Got it.

internal/driver/gomobile/driver.go
fpabl0: done

internal/driver/gomobile/driver.go
fpabl0: Unfortunately, I can't replace it with common.CanvasForObject as in desktop, because it causes a strange bug when you switch between two windows too fast (there is a possibility that the window is not correctly refreshed so you would see the previous window ui although you are in another window; resizing solves it, but the point is that it is not refreshed at time)

fpabl0:
staticcheck error is really weird:
```
Run staticcheck -go 1.12 ./...
Error: internal/driver/glfw/window_test.go:586:4: w.waitForEvents undefined (type *window has no field or method waitForEvents, but does have WaitForEvents) (compile)
Error: internal/driver/glfw/window_test.go:1029:4: w.waitForEvents undefined (type *window has no field or method waitForEvents, but does have WaitForEvents) (compile)
Error: Process completed with exit code 1.
```
`w.waitForEvents` is not called anywhere in the code.

fpabl0:
> I hate to disagree, but grepping the current state of this PR and I find:

Ok this is super strange. Could you confirm what do you see in those lines by looking at the "Files Changed" tab in this PR?
This is what I see:

Captura de Pantalla 2021-05-06 a la(s) 12 20 03

Captura de Pantalla 2021-05-06 a la(s) 12 20 44

Finally, if those errors were true, all the tests should fail but that doesn't happen :S

fpabl0:
Any idea what is happening here?

andydotxyz:
It looks like GitHub may be showing the wrong diff.
Have you tried checking out your branch clean and grepping for "waitForEvents"?

andydotxyz:
> Finally, if those errors were true, all the tests should fail but that doesn't happen :S

It's because the file that is failing has `!ci` at the top so it's skipped on the automated tests.

fpabl0:
> Have you tried checking out your branch clean and grepping for "waitForEvents"?

Yes, I have and I see the same as in the "Files changed" tabs (no "waitForEvents" anywhere).

Captura de Pantalla 2021-05-08 a la(s) 12 14 22

> It's because the file that is failing has !ci at the top so it's skipped on the automated tests.

However, the previous commit (that already had these changes) passed all the checks :S. Look: https://github.com/fyne-io/fyne/pull/2171/commits/4d5b4046fc68e869a3133bc62684fff872653c71

andydotxyz:
I would say that it's likely caused by merging your latest changes and develop.
We should start all getting the same results if you can merge develop into your branch and run the tests again.

fpabl0:
Oh that is a possibility, I will try :)

internal/driver/glfw/window_test.go
```diff
@@ -583,7 +583,6 @@ func TestWindow_Scrolled(t *testing.T) {
 
 	w.mousePos = fyne.NewPos(50, 60)
 	w.mouseScrolled(w.viewport, 10, 10)
-	w.waitForEvents()
```
fpabl0, line 586: I removed this, because `mouseScrolled` is a sync function, so `waitForEvents` does not do anything, IMHO.

fpabl0:
> merging your latest changes and develop.

This solved the issue, thanks :)

fpabl0:
Not sure what release comes now: 2.0.4 or 2.1. So should this be merged now or should I wait?

Issue #2164, opened 2021-04-12
Drop-down menu with many items is slow


Aquarian-Age:
### Describe the bug:

Select the drop-down menu year, the corresponding time is too long

### Screenshots:

![Peek 2021-04-14 06-44](https://user-images.githubusercontent.com/36611556/114630457-3bd61480-9ced-11eb-8d54-a520517b41f1.gif)

### Example code:

```go
package main

import (
        "fmt"
        "time"

        "fyne.io/fyne/v2"
        "fyne.io/fyne/v2/app"
        "fyne.io/fyne/v2/container"
        "fyne.io/fyne/v2/layout"
        "fyne.io/fyne/v2/theme"
        "fyne.io/fyne/v2/widget"
)

var t = time.Now().Local()

func main() {
        a := app.New()
        w := a.NewWindow("hello")
        a.Settings().SetTheme(theme.LightTheme())

        //select options
        years := func() []string {
                var s []string
                for i := 1601; i <= 3499; i++ {
                        is := fmt.Sprintf("%dyear", i)
                        s = append(s, is)
                }
                return s
        }

        selecty := widget.NewSelect(years(), func(s string) {
        })
        selecty.PlaceHolder = "year"

        btnOK := widget.NewButton("OK", func() {
                if selecty.SelectedIndex() == -1 {
                        selecty.SetSelectedIndex(t.Year() - 1601)
                }

                fmt.Printf("%s year index=%d\n", selecty.Selected, selecty.SelectedIndex())
        })

        hbox := container.New(layout.NewHBoxLayout(), selecty, btnOK)

        w.SetContent(container.New(layout.NewVBoxLayout(), hbox))
        w.Resize(fyne.Size{600, 600})
        w.ShowAndRun()

}
```

### Device
 - **OS: Linux** 
 - **Version: VERSION_ID="20210401" PRETTY_NAME="openSUSE Tumbleweed"**
 - **Go version: go1.16.2 linux/amd64**
 - **Fyne version: v2.0.2**

Jacalz:
Thanks for opening this bug report. Could you please fill out the description and add notes on how to reproduce. I would also appreciate if you could trim down the code example to be abort and concise. It makes it a lot easier for us to triage and review all the issues that we get. Thank you 🙂

Aquarian-Age:
[...]

PR #2180, opened 2021-04-17
Some simple optimizations for widget.Select

Jacalz:
### Description:

This makes the creation of the menu items about 30% faster when we have 500 objects.
Updates #2164

I don't judge this to be a final fix as it is still very slow. Mostly because the menu items are recreated each time they are shown, but this is at least a step in the right direction in the meantime. I also optimized selection quite a bit. It used the slow path fore selection, SetSelected(), which loops through all options.

We know that our item is in the list and can thus use the fast path.

```go
package main

import (
	"strconv"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Example")

	s := make([]string, 500)
	for i := range s {
		s[i] = strconv.Itoa(i)
	}

	selecty := &widget.Select{Options: s, PlaceHolder: "Numbers"}

	w.SetContent(container.NewVBox(selecty))
	w.Resize(fyne.Size{600, 600})
	w.ShowAndRun()
}
```

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/select.go
+++ b/widget/select.go
@@ -236,9 +236,9 @@ func (s *Select) popUpPos() fyne.Position {
 
 func (s *Select) showPopUp() {
 	items := make([]*fyne.MenuItem, len(s.Options))
-	for i, option := range s.Options {
-		text := option // capture
-		items[i] = fyne.NewMenuItem(option, func() {
+	for i := range s.Options {
+		text := s.Options[i] // capture
+		items[i] = fyne.NewMenuItem(text, func() {
 			s.optionTapped(text)
 		})
 	}
--- a/widget/select.go
+++ b/widget/select.go
@@ -224,11 +224,6 @@ func (s *Select) object() fyne.Widget {
 	return nil
 }
 
-func (s *Select) optionTapped(text string) {
-	s.SetSelected(text)
-	s.popUp = nil
-}
-
 func (s *Select) popUpPos() fyne.Position {
 	buttonPos := fyne.CurrentApp().Driver().AbsolutePositionForObject(s.super())
 	return buttonPos.Add(fyne.NewPos(0, s.Size().Height-theme.InputBorderSize()))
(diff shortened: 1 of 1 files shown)


widget/select.go
```diff
@@ -235,13 +235,12 @@ func (s *Select) popUpPos() fyne.Position {
 }
 
 func (s *Select) showPopUp() {
-	var items []*fyne.MenuItem
-	for _, option := range s.Options {
+	items := make([]*fyne.MenuItem, len(s.Options))
+	for i, option := range s.Options {
```
fpabl0, line 239: Maybe this could be re-written? 🤔:
```go
for i := range s.Options {
     text := s.Options[i] // capture
     items[i] = fyne.NewMenuItem(text, func() {
	 s.optionTapped(text)
      })
}
```

Jacalz: It could, but is there any benefit to doing so?

fpabl0: yes but it is almost imperceptible I think. The benefit is that `option` variable will not be created once and will not be assigned on every iteration without any actual need

Jacalz: I thought that the compiler wouldn't be smart enough to optimize that away (I thought it would update the range value each time anyway), but you are right. The compiler is smart enough and there is a very small difference 🙂

fpabl0: To avoid any doubt about range 😄:
```go
for i := 0; i < len(s.Options); i++ {
     text := s.Options[i] // capture
     items[i] = fyne.NewMenuItem(text, func() {
	 s.optionTapped(text)
      })
}
```

Jacalz: Yeah. I will stick with the first for now. https://github.com/fyne-io/fyne/pull/2181 will drop this code out of existence.

Jacalz:
I also managed to squeeze in a large performance boost for selecting an item when there are many items.

fpabl0 (approved):

Jacalz:
@andydotxyz Would it be fine for you to have this and https://github.com/fyne-io/fyne/pull/2169/files be backported to 2.0.3. Then perhaps https://github.com/fyne-io/fyne/pull/2181 can land for 2.1.0.

andydotxyz:
That is now on `release/v2.0.x` ready for 2.0.3 release

PR #2183, opened 2021-04-18
Fix movement of select entry popup

Jacalz:
### Description:

The popup was not being moved when open. Now the move calls propagate property down to the popup as well.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
I think we ought to have this in for 2.0.3.

andydotxyz (changes requested):
Good catch, I agree we should get this into 2.0.3, I will start picking the fixes and preparing the release this week.
In the meantime could you add a test to this, it just needs to verify that the popup has moved when the entry moves, no images needed.

Jacalz:
> Good catch, I agree we should get this into 2.0.3, I will start picking the fixes and preparing the release this week.
> In the meantime could you add a test to this, it just needs to verify that the popup has moved when the entry moves, no images needed.

The reason that I did not add a test was unfortunately because I am unusure how to write a good test for this without using images or markup assertions.

Jacalz:
Any suggestions would be appreciated. Would be nice to get this wrapped up.

fpabl0:
@Jacalz I think this should be enough:
```go
func TestSelectEntry_MoveWithPopUp(t *testing.T) {
	test.NewApp()
	defer test.NewApp()
	e := NewSelectEntry([]string{"one"})
	w := test.NewWindow(e)
	defer w.Close()
	entrySize := e.MinSize()
	w.Resize(entrySize.Add(fyne.NewSize(100, 100)))
	e.Resize(entrySize)

	// open the popup
	test.Tap(e.ActionItem.(fyne.Tappable))

	// first movement
	e.Move(fyne.NewPos(10, 10))
	assert.Equal(t, fyne.NewPos(10, 10), e.Entry.Position())
	assert.Equal(t,
		fyne.NewPos(10, 10+entrySize.Height-theme.InputBorderSize()),
		e.popUp.Position(),
	)

	// second movement
	e.Move(fyne.NewPos(30, 27))
	assert.Equal(t, fyne.NewPos(30, 27), e.Entry.Position())
	assert.Equal(t,
		fyne.NewPos(30, 27+entrySize.Height-theme.InputBorderSize()),
		e.popUp.Position(),
	)
}
```

andydotxyz:
Thanks for getting there before me @fpabl0 - I was still working through a backlog of notifications.
Perhaps for correctness the `Entry` should be inside `container.NewWithoutLayout`?

Jacalz:
Many thanks @fpabl0. I appreciate the help. I had to update the test to use the same package though so we could get access to the unexported methods.

> Thanks for getting there before me @fpabl0 - I was still working through a backlog of notifications.
> Perhaps for correctness the `Entry` should be inside `container.NewWithoutLayout`?

The other tests in that test file did not, so I did not add it in.

fpabl0:
> I had to update the test to use the same package though so we could get access to the unexported methods.

Maybe the you should only need to add a new file called `select_entry_internal_test.go` as other widgets do? 🤔

> Thanks for getting there before me @fpabl0 - I was still working through a backlog of notifications.
Perhaps for correctness the Entry should be inside container.NewWithoutLayout?

I don't this is needed as it is redundant 🤔. By doing `SetContent(entry)` you are already saying to not use a layout so it is just the same, isn't it?

andydotxyz:
> The other tests in that test file did not, so I did not add it in.

But the other tests are not moving it around are they?
I was just meaning that if we want to override layout it may be better to specify that there isn't one.

andydotxyz:
> By doing SetContent(entry) you are already saying to not use a layout so it is just the same, isn't it?

That's not the case - the root will behave like a Max container, automatically resizing all content to fill.

fpabl0:
> That's not the case - the root will behave like a Max container, automatically resizing all content to fill.

Oh you are right. I didn't realize about that before! Then, I agree with your suggestion.

Jacalz:
I initially thought to add another test file for internal tests, but decided against it given that it was just one test. Should I change that?

fpabl0:
> I initially thought to add another test file for internal tests, but decided against it given that it was just one test. Should I change that?

Yes, agree. Maybe renaming the existing one to end with *_internal_test.go should be enough to keep current file test names consistency? Any thoughts @andydotxyz ?

Jacalz:
Good idea @fpabl0. I have updated with that solution for now.

Jacalz:
Weird, there seems to have been some random test failures. Re-run seems to clear things up.

Jacalz:
Never mind. Platform test are still failing on entry validation.

fpabl0:
Hmm it only fails in windows platform 🤔

Jacalz:
Yeah. I think I have seen it on other PRs as well. Does not seem to be related to this PR though.

andydotxyz:
It also does not happen reliably on Windows desktop! If you kick the builds it should fix

Jacalz:
I had already kicked the tests once, but now they seem to work. Everything is green now 🙂

andydotxyz (approved):

andydotxyz:
Thanks. This is picked onto release/v2.0.x as well

PR #2185, opened 2021-04-19
Fix: filter out shortcut triggers where key-modifier is same

### LINKED ISSUE #2148: Don't create empty shortcuts
### Describe the bug:
When typing out a keyboard shortcut like "Ctr-A", first an empty "-Ctrl" shortcut is sent, which should not be

### To Reproduce:
1. Implement `TypedShortcut`
2. Print the content of any `desktop.CustomShortcut`
3. Notice that the first is always listing `LeftControl` or `RightControl` as the `KeyName`

 - **OS:** darwin
 - **Version:** Catalina
 - **Go version:** 1.16
 - **Fyne version:** 2.0.2

AnkushJadhav:
### Description:
Shortcuts triggered an event to the queue for the shortcut's modifier key press. Added a filter to prevent the modifier key press from triggering the shortcut callback.

Fixes #2148

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -1201,6 +1201,20 @@ func desktopModifier(mods glfw.ModifierKey) desktop.Modifier {
 	return m
 }
 
+func isKeyModifierPair(keyName fyne.KeyName, modifier desktop.Modifier) bool {
+	switch modifier {
+	case desktop.ShiftModifier:
+		return (keyName == desktop.KeyShiftLeft || keyName == desktop.KeyShiftRight)
+	case desktop.ControlModifier:
+		return (keyName == desktop.KeyControlLeft || keyName == desktop.KeyControlRight)
+	case desktop.AltModifier:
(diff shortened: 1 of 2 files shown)


andydotxyz (changes requested):
This is a great fix thanks, would it be possible to add a test that shows the shortcut is no longer sent to an Entry?

andydotxyz:
Oh one other thought - we try to keep method names in alphabetical order (grouped by type) so the new method probably belongs further down the file.

AnkushJadhav:
Thanks. Will make the changes and update the PR asap. Since the fix is applicable for any component on the window that implements `TypedShortcut`, should the test for this be on `window` or on the `entry` widget?

AnkushJadhav:
@andydotxyz Added tests to a generic implementation of `Shortcutable` instead of entry, as the code change impacts all implementations of `Shortcutable`

andydotxyz (approved):
Great thank you very much :)

PR #2187, opened 2021-04-20
allow tabs to store transitioning state to enable transition animations

### LINKED ISSUE #2135: container: Tab animation is too slow vertically and starts too high
### Describe the bug:

I initially noted this in wormhole-gui when using the mobile branch. There I see that the tab animation starts way too high on the screen an then  slowly falls down to where it should be. That happens a bit too fast in the code example below to be noticeable enough, but it can be seen when resizing the window (the animation var does not stay with the tab).

### To Reproduce:
1. Resize window with tab bar at the bottom
2. Or: Start application with tab bar at the bottom.

### Screenshots:

The refresh rate on the gif is a bit too slow to see the full effect:
falling-tabs

The example code resizing:
falling-tabs-example

### Example code:
```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Hello")

	tabs := container.NewAppTabs(container.NewTabItem("Test", hello))
	tabs.SetTabLocation(container.TabLocationBottom)

	w.SetContent(tabs)
	w.Resize(fyne.NewSize(200, 600))

	w.ShowAndRun()
}
```

 - **OS:** Linux
 - **Version:** 5.11.9
 - **Go version:** 1.16.2
 - **Fyne version:** 2.0.2

AnkushJadhav:
### Description:
Modified the `tabs` interface to enforce 2 new methods: `transitioning() bool` and `setTransitioning(bool)` which allows the implementations of `tabs` (currently: `AppTabs` and `DocTabs`) to capture the state while a tab change is happening.
This allows us to enable tab transitioning animation on the tab indicator only during a valid tab change and not every time the tab indicator is rendered.

ezgif com-gif-maker

Fixes #2135 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container/apptabs.go
+++ b/container/apptabs.go
@@ -26,8 +26,9 @@ type AppTabs struct {
 	OnSelected   func(*TabItem)
 	OnUnselected func(*TabItem)
 
-	current  int
-	location TabLocation
+	current         int
+	location        TabLocation
+	isTransitioning bool
 
 	popUpMenu *widget.PopUpMenu
 }
@@ -59,7 +60,7 @@ func (t *AppTabs) CreateRenderer() fyne.WidgetRenderer {
 	// Initially setup the tab bar to only show one tab, all others will be in overflow.
 	// When the widget is laid out, and we know the size, the tab bar will be updated to show as many as can fit.
 	r.updateTabs(1)
-	r.updateIndicator()
+	r.updateIndicator(false)
 	r.applyTheme(t)
 	return r
 }
@@ -237,8 +238,16 @@ func (t *AppTabs) setSelected(selected int) {
 	t.current = selected
 }
 
+func (t *AppTabs) setTransitioning(transitioning bool) {
+	t.isTransitioning = transitioning
+}
+
 func (t *AppTabs) tabLocation() TabLocation {
 	return t.location
+}
+
+func (t *AppTabs) transitioning() bool {
+	return t.isTransitioning
 }
 
 // Declare conformity with WidgetRenderer interface.
@@ -268,7 +277,10 @@ func (r *appTabsRenderer) Layout(size fyne.Size) {
 	}
 
 	r.layout(r.appTabs, size)
-	r.updateIndicator()
+	r.updateIndicator(r.appTabs.transitioning())
+	if r.appTabs.transitioning() {
+		r.appTabs.setTransitioning(false)
+	}
 }
 
 func (r *appTabsRenderer) MinSize() fyne.Size {
@@ -354,7 +366,7 @@ func (r *appTabsRenderer) buildTabButtons(count int) *fyne.Container {
 	return buttons
 }
 
-func (r *appTabsRenderer) updateIndicator() {
+func (r *appTabsRenderer) updateIndicator(animate bool) {
 	if r.appTabs.current < 0 {
 		r.indicator.Hide()
 		return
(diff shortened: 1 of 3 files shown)


container/apptabs.go
```diff
@@ -237,10 +238,18 @@ func (t *AppTabs) setSelected(selected int) {
...
+}
+
 func (t *AppTabs) tabLocation() TabLocation {
 	return t.location
 }
 
+func (t *AppTabs) transitioning() bool {
+	return t.isTransitioning
+}
+
```
Jacalz, line 252: Do we really need functions for this if they just return the value? I know that it kind of already is like that in the file though.

container/doctabs.go
```diff
@@ -193,10 +194,18 @@ func (t *DocTabs) setSelected(selected int) {
...
+}
+
 func (t *DocTabs) tabLocation() TabLocation {
 	return t.location
 }
 
+func (t *DocTabs) transitioning() bool {
+	return t.isTransitioning
+}
+
```
Jacalz, line 208: Same thoughts here.

Jacalz (commented):
Thanks for the PR. Just had some thoughts 💭

AnkushJadhav:
@Jacalz Added the `transitioning` function just to match the existing design pattern and to complement the `setTransitioning` function. Doesn't really provide any value other than that. Your thoughts?

Jacalz:
Indeed. I see why you did it :)

I personally feel that we would be better off without the functions, but it'd be interested to see what more people have to say about it.

andydotxyz:
> I personally feel that we would be better off without the functions, but it'd be interested to see what more people have to say about it.

It is there because this is common code shared by two concrete implementations and they (for struct init and backwards compatibility) expose their own structs.

andydotxyz:
This is an excellent fix to have in thanks @AnkushJadhav, however I wonder if we can do it without the extra flags.
Thinking about when this occurs is it not always from a resize? If so then the code should already know whether or not it is "transitioning" or whether it is "resizing". For example in Layout if the size requested != the current size then we could turn the animation off?

AnkushJadhav:
@andydotxyz Thought about this approach, but on debugging found out that the AppTabs size is already updated with the new size before the Layout function is called with the new size. So there is no way to check the previous size of the AppTabs using the Size() function. Please let me know if I was looking for the current size at the wrong place?

Aside: I also thought that capturing this data would be handy, in case something like an `OnTabChange` kind of event needs to emitted in the future.

andydotxyz:
> but on debugging found out that the AppTabs size is already updated with the new size before the Layout function is called with the new size

That does make it a little less convenient - but we could cache the last size that Layout saw perhaps? But maybe that adds the same problem of fields and methods required to access?

andydotxyz:
> Aside: I also thought that capturing this data would be handy, in case something like an `OnTabChange` kind of event needs to emitted in the future.

We already have `OnChanged`, is that not what you meant?

AnkushJadhav:
@andydotxyz yes, caching the last updated size would also require a field on the AppTab's renderer, which get's us back to the same place of having to maintain a field. Not that it would make a huge difference, but caching a bool would be more efficient than a size object?

And yes, `OnChanged` is exactly what I meant. Missed that when I went through the documentation. My bad.

andydotxyz:
That is a pretty solid explanation I think - what about @Jacalz ?

I was looking at this for 2.0.3 but the code on `develop` is quite different.
Should we get another PR for that release? (I was hoping to open the release vote today but it could delay a little I guess)?

Jacalz (approved):

stuartmscott (approved):

PR #2192, opened 2021-04-24
fix: allow apptabs to store state while transitioning tabs

### LINKED ISSUE #2135: container: Tab animation is too slow vertically and starts too high

AnkushJadhav:
### Description:
Added an internal field to AppTabs that stores the state when a tab change is happening, thereby allowing animations only on valid tab changes.
This is a port of #2187 for release 2.0.3.

Fixes #2135 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container/apptabs.go
+++ b/container/apptabs.go
@@ -425,30 +425,29 @@ func (r *appTabsRenderer) moveSelection(withAnimation bool) {
 		return
 	}
 
-	if withAnimation {
-		if r.animation != nil {
-			r.animation.Stop()
-		}
-		r.animation = canvas.NewPositionAnimation(r.underline.Position(), underlinePos, canvas.DurationShort, func(p fyne.Position) {
-			r.underline.Move(p)
-			canvas.Refresh(r.underline)
-			if p == underlinePos {
-				r.animation = nil
-			}
-		})
-		r.animation.Start()
-
-		canvas.NewSizeAnimation(r.underline.Size(), underlineSize, canvas.DurationShort, func(s fyne.Size) {
-			r.underline.Resize(s)
-			canvas.Refresh(r.underline)
-		}).Start()
-	} else {
-		if r.animation == nil {
-			r.underline.Move(underlinePos)
-			r.underline.Resize(underlineSize)
-			canvas.Refresh(r.underline)
-		}
+	if !withAnimation && r.animation == nil {
+		r.underline.Move(underlinePos)
+		r.underline.Resize(underlineSize)
+		canvas.Refresh(r.underline)
+		return
 	}
+
+	if r.animation != nil {
+		r.animation.Stop()
+	}
+	r.animation = canvas.NewPositionAnimation(r.underline.Position(), underlinePos, canvas.DurationShort, func(p fyne.Position) {
+		r.underline.Move(p)
+		canvas.Refresh(r.underline)
+		if p == underlinePos {
+			r.animation = nil
+		}
+	})
+	r.animation.Start()
+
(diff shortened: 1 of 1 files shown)


AnkushJadhav:
@andydotxyz The flicker on tab switching was due to #1952 , which has been fixed on develop. I have implemented a lesser elegant solution in this PR, which avoids the flicker.

container/apptabs.go
```diff
@@ -89,7 +90,7 @@ func (c *AppTabs) CreateRenderer() fyne.WidgetRenderer {
 	c.BaseWidget.ExtendBaseWidget(c)
 	r := &appTabsRenderer{line: canvas.NewRectangle(theme.ShadowColor()),
 		underline: canvas.NewRectangle(theme.PrimaryColor()), container: c}
-	r.updateTabs()
+	r.updateTabs(c.isTransitioning)
```
andydotxyz, line 93: When creating a renderer isn't this always false?

container/apptabs.go
```diff
@@ -422,22 +425,30 @@ func (r *appTabsRenderer) moveSelection() {
...
 
-	if r.animation != nil {
-		r.animation.Stop()
-	}
-	r.animation = canvas.NewPositionAnimation(r.underline.Position(), underlinePos, canvas.DurationShort, func(p fyne.Position) {
-		r.underline.Move(p)
-		canvas.Refresh(r.underline)
-		if p == underlinePos {
-			r.animation = nil
+	if withAnimation {
```
andydotxyz, line 428: If this was !withAnimation then we could handle the non-animated case and return, returning the main code to the least indentation.
This is our style but also makes it easier to merge back to develop after release.

andydotxyz (changes requested):
Thanks so much, you found the problem I could not!
Just a couple of small things inline if that's OK and I think we're ready.

container/apptabs.go
AnkushJadhav: Yes, `c.isTransitioning` would ideally always be `false` at this point. Would it be better if I explicitly set it as false?

container/apptabs.go
AnkushJadhav: That makes sense, updated the PR.

Jacalz (approved):
Nice work. Thank you very much for the fixes :+1:

andydotxyz (approved):
Great thanks

PR #2194, opened 2021-04-25
Remove unnecessary usage of fmt.Sprintf

Jacalz:
### Description:

This removes a couple uses of `fmt.Sprintf` that were unnecessary and replaces it with string appends or `strconv` in places where it does not noticeably hurt the code readability. Some of the changes should be more noticable than others, giving performance improvements all over the place. I think the changes to the `theme` package, `widget.ProgressBar` and `widget.TextGrid` should be the most noticeable cases, performance wise. 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -4,8 +4,8 @@
 package app // import "fyne.io/fyne/v2/app"
 
 import (
-	"fmt"
 	"os/exec"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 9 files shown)


app/app.go
```diff
@@ -43,7 +43,7 @@ func (app *fyneApp) UniqueID() string {
 	}
 
 	fyne.LogError("Preferences API requires a unique ID, use app.NewWithID()", nil)
-	app.uniqueID = fmt.Sprintf("missing-id-%d", time.Now().Unix()) // This is a fake unique - it just has to not be reused...
+	app.uniqueID = "missing-id-" + strconv.FormatInt(time.Now().Unix(), 10) // This is a fake unique - it just has to not be reused...
```
andydotxyz, line 46: Is there a reason you didn’t use `ItoA` here?

Jacalz: Because `time.Now().Unix()` is of type `int64` and that is what `strconv.FormatInt()` takes. Using `strconv.Itoa()` would mean that we would have converted it from `int64` to `int` and then back into `int64` (Itoa is a wrapper around FormatInt).

andydotxyz (approved):

PR #2195, opened 2021-04-25
ScrollTo API on List widget

### LINKED ISSUE #1892: Need to reset the scroll position of table widget

No

### LINKED ISSUE #2105: Scrolling API for List Widget

As new items appended to the List widget until it has the scroll, it probably better for UX when she see that newly created item at the bottom of the list. 

Probably. From what I see, the `List` widget has `scroller` API in it but just not exposed yet.

```
list.ScrollToBottom()
```

or event better
```
list.ScrollToIndex(elementIndexToBeShown)
```

AnkushJadhav:
### Description:
Exposed a `ScrollTo` API that scrolls a List widget and sets the focus on the ListItem represented by the id given to the API.

Fixes #2105 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -147,6 +147,22 @@ func (l *List) ScrollTo(id ListItemID) {
 	l.Refresh()
 }
 
+// ScrollToBottom scrolls to the end of the list
+func (l *List) ScrollToBottom() {
+	length := 0
+	if f := l.Length; f != nil {
+		length = f()
+	}
+	l.scrollTo(length)
+	l.Refresh()
+}
+
+// ScrollToTop scrolls to the start of the list
+func (l *List) ScrollToTop() {
+	l.scrollTo(0)
+	l.Refresh()
+}
+
 // Unselect removes the item identified by the given ID from the selection.
 func (l *List) Unselect(id ListItemID) {
 	if len(l.selected) == 0 || l.selected[0] != id {
--- a/widget/list.go
+++ b/widget/list.go
@@ -99,7 +99,7 @@ func (l *List) scrollTo(id ListItemID) {
 	if l.scroller == nil {
 		return
 	}
-	y := float32(id) * (l.itemMin.Height + theme.SeparatorThicknessSize())
+	y := (float32(id) * l.itemMin.Height) + (float32(id) * theme.SeparatorThicknessSize())
 	if y < l.scroller.Offset.Y {
 		l.scroller.Offset.Y = y
 	} else if y+l.itemMin.Height > l.scroller.Offset.Y+l.scroller.Size().Height {
@@ -153,7 +153,7 @@ func (l *List) ScrollToBottom() {
 	if f := l.Length; f != nil {
 		length = f()
 	}
-	l.scrollTo(length)
+	l.scrollTo(length - 1)
 	l.Refresh()
 }
 
--- a/widget/list.go
+++ b/widget/list.go
@@ -135,6 +135,8 @@ func (l *List) Select(id ListItemID) {
 }
 
 // ScrollTo scrolls to the item represented by id
+//
+// Since: 2.1
 func (l *List) ScrollTo(id ListItemID) {
 	length := 0
 	if f := l.Length; f != nil {
@@ -148,6 +150,8 @@ func (l *List) ScrollTo(id ListItemID) {
 }
 
 // ScrollToBottom scrolls to the end of the list
+//
+// Since: 2.1
 func (l *List) ScrollToBottom() {
 	length := 0
 	if f := l.Length; f != nil {
@@ -161,6 +165,8 @@ func (l *List) ScrollToBottom() {
 }
 
 // ScrollToTop scrolls to the start of the list
+//
+// Since: 2.1
 func (l *List) ScrollToTop() {
 	l.scrollTo(0)
 	l.Refresh()
--- a/widget/list.go
+++ b/widget/list.go
@@ -99,7 +99,7 @@ func (l *List) moveFocusTo(id ListItemID) {
 	if l.scroller == nil {
 		return
 	}
-	y := (float32(id) * l.itemMin.Height) + (float32(id) * theme.SeparatorThicknessSize())
+	y := float32(id) * (l.itemMin.Height + theme.SeparatorThicknessSize())
 	if y < l.scroller.Offset.Y {
 		l.scroller.Offset.Y = y
 	} else if y+l.itemMin.Height > l.scroller.Offset.Y+l.scroller.Size().Height {
--- a/widget/list.go
+++ b/widget/list.go
@@ -153,7 +153,10 @@ func (l *List) ScrollToBottom() {
 	if f := l.Length; f != nil {
 		length = f()
 	}
-	l.scrollTo(length - 1)
+	if length > 0 {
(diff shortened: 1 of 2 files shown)


widget/list.go
```diff
@@ -95,6 +95,19 @@ func (l *List) MinSize() fyne.Size {
 	return l.BaseWidget.MinSize()
 }
 
+func (l *List) moveFocusTo(id ListItemID) {
+	if l.scroller == nil {
+		return
+	}
+	y := (float32(id) * l.itemMin.Height) + (float32(id) * theme.SeparatorThicknessSize())
```
Jacalz, line 102: I know that it was like this before, but this can be simplified. You can rewrite it as the following:
```suggestion
	y := float32(id) * (l.itemMin.Height) + theme.SeparatorThicknessSize())
```

Jacalz (changes requested):
Nick work. Just a little suggestion to simplify a math operation.

widget/list.go
Jacalz: Out of interest, I double checked to see if the resulting assembly were different or if the compiler would be smart enough to optimize it away. It does look like this change indeed does save us a multiplication, as I expected 🙂 https://go.godbolt.org/z/1556hfGnr

AnkushJadhav: That's a great catch! Thanks for the suggestion!

Jacalz: You're welcome. Feels a bit strange though that the compiler wouldn't be able to optimize it away. Opened an issue over at golang/go, https://github.com/golang/go/issues/45760.

Jacalz (dismissed):
Thanks. This looks good to me 👍 

Just one little note. This seems to incorrectly mention fixing https://github.com/fyne-io/fyne/issues/1892 which is specific to the table widget. Did you plan to add in a fix for that as well?

widget/list.go
Jacalz: Comments in that issue mentioned that rounding will make this create different values. https://play.golang.org/p/ziUoFO0OhNT

I think you might want to review this change. Sorry 😬

andydotxyz (changes requested):
An api “ScrollTo” should not change selection, there is a separate API that does that already. 

Also the initial issue requests ScrollToBottom, would that and ScrollToTop be useful too?

AnkushJadhav:
> Thanks. This looks good to me 👍
> 
> Just one little note. This seems to incorrectly mention fixing #1892 which is specific to the table widget. Did you plan to add in a fix for that as well?

@Jacalz No, I think I added that one incorrectly. Not sure what I was thinking. Updated the PR description.

AnkushJadhav:
> An api “ScrollTo” should not change selection, there is a separate API that does that already.
> 
> Also the initial issue requests ScrollToBottom, would that and ScrollToTop be useful too?

@andydotxyz The "ScrollTo" API does not change the selection. It simply scrolls to the item provided to it. Any selection made on the list remains intact.
A ScrollToTop and ScrollToBottom seemed more like utility functions to me, that don't add much value. The developer can easily implement these functions themselves by using the ScrollTo API. Your thoughts?

AnkushJadhav:
@andydotxyz The existing `Select` API modified the selection and also scrolled to the selection. While the new `ScrollTo` API ONLY scrolls to the item. Since both these APIs have a "scroll to item" kind of logic in them, so extracted that logic into a common helper function.

widget/list.go
```diff
@@ -95,6 +95,19 @@ func (l *List) MinSize() fyne.Size {
 	return l.BaseWidget.MinSize()
 }
 
+func (l *List) moveFocusTo(id ListItemID) {
```
andydotxyz, line 98: This name confused me. Please don't use the word "Focus" when you are referring to visibility

andydotxyz (changes requested):
I understand now - apologies that I misread the code.
I was confused by the use of the word "Focus" when you were just meaning scroll to make visible. Can you please update the method name so it does not imply focus change.

andydotxyz:
> @andydotxyz The existing `Select` API modified the selection and also scrolled to the selection. While the new `ScrollTo` API ONLY scrolls to the item. Since both these APIs have a "scroll to item" kind of logic in them, so extracted that logic into a common helper function.

You're quite right, apologies. The method name tripped me up, can you find a better name?

andydotxyz:
> A ScrollToTop and ScrollToBottom seemed more like utility functions to me, that don't add much value

I was just suggesting it because those were added to `container.Scroll` at the request of previous users.

AnkushJadhav:
> > A ScrollToTop and ScrollToBottom seemed more like utility functions to me, that don't add much value
> 
> I was just suggesting it because those were added to `container.Scroll` at the request of previous users.

Oh, then it makes sense to have them for consistency. Updated the PR.

widget/list.go
```diff
@@ -147,6 +147,22 @@ func (l *List) ScrollTo(id ListItemID) {
 	l.Refresh()
 }
 
+// ScrollToBottom scrolls to the end of the list
+func (l *List) ScrollToBottom() {
+	length := 0
+	if f := l.Length; f != nil {
+		length = f()
+	}
+	l.scrollTo(length)
```
andydotxyz, line 156: should this not be `length - 1` (as long as length is > 0)?

andydotxyz (changes requested):
I think there is an off-by-one in there.

widget/list.go
Jacalz: Oh, sorry looks like there was a typo in that message. I meant that you might need to revert the change. It can cause issues with rounding and be less precise that what it was before.

andydotxyz (approved):
Great, thanks

andydotxyz:
It seems that GitHub has a policy or blocking first-time contributor CI. I have asked it to progress now.
As a result there are some staticcheck errors to fix, sorry!

widget/list.go
Jacalz: Thanks for reverting it back to how it was. You can remove the parenthesis if you like, but now exactly needed.

andydotxyz (changes requested):
I am so sorry for forgetting about this (my head was in bug fixing) but each of the new methods needs the following comment added:
```go
//
// Since: 2.1
```

widget/list.go
AnkushJadhav: I prefer parenthesis from a code readability perspective. Can remove them if it goes against the project design guidelines.

AnkushJadhav:
@andydotxyz Thanks for pointing that out. Sorry, this is my first time on a big open source project, so still learning all the nitty-gritty.

widget/list.go
Jacalz: I don’t think it goes against any guidelines, no. I just feel that they are unnecessary given that order of operation means that they have no meaning. This is not really hard to read code in the first place, but if you want to keep them, feel free to do so. I’m not going to stop you 🙂

andydotxyz:
> @andydotxyz Thanks for pointing that out. Sorry, this is my first time on a big open source project, so still learning all the nitty-gritty.

Don't apologise, we have a lot of processes to keep the API need and consistent. There is a load written down at https://github.com/fyne-io/fyne/wiki/Contributing

andydotxyz (approved):

PR #2196, opened 2021-04-25
data/validation: add time format validator

Jacalz:
### Description:

Adds a validator that I used in [Sparta](https://github.com/Jacalz/sparta) for verifying that dates and times were correct.
Can probably be useful for a lot of other use cases as well :)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/validation/time.go
+++ b/data/validation/time.go
@@ -8,7 +8,8 @@ import (
 
 // NewTime creates a new validator that verifies times using time.Parse.
 // The validator will return nil if valid, otherwise returns an error with a reason text.
-// See time.Parse() for reference time format: https://pkg.go.dev/time#Parse
+// The reference time for the format: Mon Jan 2 15:04:05 -0700 MST 2006.
+// See time.Parse() for more information about the reference time: https://pkg.go.dev/time#Parse
 //
 // Since: 2.1
 func NewTime(format string) fyne.StringValidator {
--- a/data/validation/time.go
+++ b/data/validation/time.go
@@ -13,10 +13,7 @@ import (
 // Since: 2.1
 func NewTime(format string) fyne.StringValidator {
 	return func(text string) error {
-		if _, err := time.Parse(format, text); err != nil {
-			return err
-		}
-
-		return nil
+		_, err := time.Parse(format, text)
+		return err
 	}
 }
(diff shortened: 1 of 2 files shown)


data/validation/time.go
```diff
@@ -0,0 +1,22 @@
...
+// The validator will return nil if valid, otherwise returns an error with a reason text.
+//
+// Since: 2.1
+func NewTime(format string) fyne.StringValidator {
+	return func(text string) error {
+		if _, err := time.Parse(format, text); err != nil {
+			return err
+		}
+
+		return nil
```
stuartmscott, line 20:
```suggestion
		_, err := time.Parse(format, text)
		return err
```

stuartmscott (changes requested):
A useful addition Jacob!

data/validation/time.go
Jacalz: Good idea. Thanks 🙂

stuartmscott (approved):

andydotxyz (commented):
Nice new Validator. 
mid suggest linking to the date format docs though, the Go team made a really good job of describing various formats.

stuartmscott (dismissed):

andydotxyz (approved):

fpabl0:
Although it is good to have the link to the go docs. I would prefer to keep this line too:
```
The reference time for the format: Mon Jan 2 15:04:05 -0700 MST 2006.
```
Sometimes, you already know how it works, but you do not remember the reference time, and opening a link just to see it does not seem to be a good experience, does it?

andydotxyz:
> Although it is good to have the link to the go docs. I would prefer to keep this line too:

That is a good point, I agree

andydotxyz (changes requested):
I like @fpabl0 comment about re-adding the reference line for quicker doc lookup for the familiar.

Jacalz:
Good point. Added it back :slightly_smiling_face:

andydotxyz (approved):
Great thanks

data/validation/time_test.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"testing"
+
+	"fyne.io/fyne/v2/data/validation"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTime(t *testing.T) {
+	time := validation.NewTime("2006-01-02")
```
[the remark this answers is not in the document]
Jacalz: Good point. I renamed it to `validate`.

fpabl0 (approved):
LGTM :)

PR #2201, opened 2021-04-26
Optimizations for AppTabs and DocTabs

Jacalz:
### Description:

This basically does two things. Firstly, we pre-allocate the slice of menu items that we are shoring in the drop-downs.
Secondly, we make sure to use the index of the tab instead of the tab item, for selection.
This avoids ranging through all the items to find the correct index.

There seems to be a panic happening sometimes when pressing the drop down button when it is already open, but does not seem to be related to this PR.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container/apptabs.go
+++ b/container/apptabs.go
@@ -288,7 +288,7 @@ func (r *appTabsRenderer) Refresh() {
 }
 
 func (r *appTabsRenderer) buildOverflowTabsButton() (overflow *widget.Button) {
-	return &widget.Button{Icon: moreIcon(r.appTabs), Importance: widget.LowImportance, OnTapped: func() {
+	overflow = &widget.Button{Icon: moreIcon(r.appTabs), Importance: widget.LowImportance, OnTapped: func() {
 		// Show pop up containing all tabs which did not fit in the tab bar
 
 		itemLen, objLen := len(r.appTabs.Items), len(r.bar.Objects[0].(*fyne.Container).Objects)
@@ -309,6 +309,8 @@ func (r *appTabsRenderer) buildOverflowTabsButton() (overflow *widget.Button) {
 
 		r.appTabs.popUpMenu = buildPopUpMenu(r.appTabs, overflow, items)
 	}}
+
+	return overflow
 }
 
 func (r *appTabsRenderer) buildTabButtons(count int) *fyne.Container {
--- a/container/apptabs.go
+++ b/container/apptabs.go
@@ -294,11 +294,10 @@ func (r *appTabsRenderer) buildOverflowTabsButton() (overflow *widget.Button) {
 		itemLen, objLen := len(r.appTabs.Items), len(r.bar.Objects[0].(*fyne.Container).Objects)
 		items := make([]*fyne.MenuItem, 0, itemLen-objLen)
 		for i := objLen; i < itemLen; i++ {
-			item := r.appTabs.Items[i]
 			index := i // capture
 			// FIXME MenuItem doesn't support icons (#1752)
 			// FIXME MenuItem can't show if it is the currently selected tab (#1753)
-			items = append(items, fyne.NewMenuItem(item.Text, func() {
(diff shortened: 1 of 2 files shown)


stuartmscott (approved):
Nice work Jacob!

Hopefully all the actions pass

container/apptabs.go
```diff
@@ -306,9 +308,7 @@ func (r *appTabsRenderer) buildOverflowTabsButton() (overflow *widget.Button) {
 		}
 
 		r.appTabs.popUpMenu = buildPopUpMenu(r.appTabs, overflow, items)
```
andydotxyz, line 310: I'm not certain but it looks like the overflow variable is no longer set. 
Is this part of named returns magic?

Jacalz: I'd say that it's mostly an oversight. I don't know why it didn't cause my issues as far as I could see.

fpabl0: > Is this part of named returns magic?

Yes, it is, named returns are hard to understand IMHO 😕

> I don't know why it didn't cause my issues as far as I could see.

Because `overflow` is set when the function `returns`. It is certainly not very obvious but it is right. The overflow is set.
https://play.golang.org/p/5tTC9YSiwpm

Jacalz: So, should we a) keep it as is, b) add a comment or c) remove the named return revert to how it was prior?

Jacalz: I'll go with c) then. I think it is a bit unclear as well :slightly_smiling_face:

Jacalz: I had to keep the named return, but removed the magic part about it. Should be good now.

andydotxyz:
Note that the `overflow` happened again with an `all` variable

Jacalz:
Good catch. Fixed that now.

container/apptabs.go
fpabl0: Optionally, you could remove the named return too and just have `buildOverflowTabsButton() *widget.Button`

Jacalz: Nope. Can't access the variable inside the anonymous function by doing it that way (without modifying it in more ways). I think this works good as is now.

container/apptabs.go
```diff
@@ -288,16 +288,18 @@ func (r *appTabsRenderer) Refresh() {
...
-	overflow = widget.NewButtonWithIcon("", moreIcon(r.appTabs), func() {
+	overflow = &widget.Button{Icon: moreIcon(r.appTabs), Importance: widget.LowImportance, OnTapped: func() {
 		// Show pop up containing all tabs which did not fit in the tab bar
 
-		var items []*fyne.MenuItem
-		for i := len(r.bar.Objects[0].(*fyne.Container).Objects); i < len(r.appTabs.Items); i++ {
+		itemLen, objLen := len(r.appTabs.Items), len(r.bar.Objects[0].(*fyne.Container).Objects)
+		items := make([]*fyne.MenuItem, 0, itemLen-objLen)
+		for i := objLen; i < itemLen; i++ {
 			item := r.appTabs.Items[i]
```
fpabl0, line 297: This is optional but I think now we can avoid creating this variable as we now capture the index?
```go
items = append(items, fyne.NewMenuItem(r.appTabs.Items[index].Text, func() {
	r.appTabs.SelectIndex(index)
	if r.appTabs.popUpMenu != nil {
		r.appTabs.popUpMenu.Hide()
		r.appTabs.popUpMenu = nil
	}
}))
```

fpabl0 (dismissed):
LGTM :) just one thought, but feel free to ignore it as I don't think it would have really impact on performance.

container/apptabs.go
Jacalz: Good point. Will fix that tomorrow if I have the time 🙂

andydotxyz (dismissed):

container/apptabs.go
Jacalz: Fixed now.

fpabl0 (approved):
LGTM!

PR #2203, opened 2021-04-27
Remove stale builds

### LINKED ISSUE #2154: requirements for target platforms
I compiled an app on macOS Big Sur to use on another MacBook. The resulting app could not run on that machine because `You can't use this version of the application "MyApp" with this version of macOS. You have macOS 10.15.7. The application requires macOS 11.0 or later."`.

I tried to lookup here and in the docs what minimum OS versions are supported for each platform, but could not find anything. It would be useful if that was more clear. If I missed it, I looked for 'target', 'platform' and 'requirements'.

In case this depends on certain libraries, meaning that if omitted a lower OS version would be supported, then that info would also be very helpful.

andydotxyz:
### Description:
If no executable was specified, but one was in place,
assume it is stale and remove

Fixes #2154

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/commands/package.go
+++ b/cmd/fyne/commands/package.go
@@ -157,6 +157,10 @@ func (p *packager) validate() error {
 
 	if p.exe == "" {
 		p.exe = filepath.Join(p.srcDir, exeName)
+
+		if util.Exists(p.exe) { // the exe was not specified, assume stale
+			p.removeBuild()
+		}
 	} else if p.os == "ios" || p.os == "android" {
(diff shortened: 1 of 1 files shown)


fpabl0 (approved):
This change is good, but I think maybe the help usage should be improved (for the -exe flag), until now I didn't understand what did that flag do 🤔

andydotxyz:
> This change is good, but I think maybe the help usage should be improved (for the -exe flag), until now I didn't understand what did that flag do 🤔

Yes it was out of date. I have updated to:
"Specify an existing binary instead of building before package"

PR #2204, opened 2021-04-27
Add missing position info in ScrollEvent

### LINKED ISSUE #2199: ScrollEvent.Position and ScrollEvent.AbsolutePosition is 0,0
I create Scrolled(*fyne.ScrollEvent) method on my widget. It gets called when I scroll the mouse wheel, but the Position and AbsolutePosition fields of the ScrollEvent seem to be all zeroes.
I was expecting to find mouse position there.

OS: Linux (Ubuntu 20.04)
 Fyne version: v2.0.2

andydotxyz:
Fixes #2199

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -825,7 +825,7 @@ func (w *window) waitForDoubleTap(co fyne.CanvasObject, ev *fyne.PointEvent) {
 }
 
 func (w *window) mouseScrolled(viewport *glfw.Window, xoff float64, yoff float64) {
-	co, _, _ := w.findObjectAtPositionMatching(w.canvas, w.mousePos, func(object fyne.CanvasObject) bool {
+	co, pos, _ := w.findObjectAtPositionMatching(w.canvas, w.mousePos, func(object fyne.CanvasObject) bool {
 		_, ok := object.(fyne.Scrollable)
 		return ok
 	})
@@ -838,6 +838,8 @@ func (w *window) mouseScrolled(viewport *glfw.Window, xoff float64, yoff float64
 		}
 		ev := &fyne.ScrollEvent{}
 		ev.Scrolled = fyne.NewDelta(float32(xoff)*scrollSpeed, float32(yoff)*scrollSpeed)
+		ev.Position = pos
+		ev.AbsolutePosition = w.mousePos
 		wid.Scrolled(ev)
 	}
 }
--- a/internal/driver/glfw/window_test.go
+++ b/internal/driver/glfw/window_test.go
@@ -544,7 +544,7 @@ func TestWindow_Scrolled(t *testing.T) {
 	w.mouseScrolled(w.viewport, 10, 10)
 	w.waitForEvents()
 
-	if e, _ := o.popScrollEvent().(*fyne.PointEvent); assert.NotNil(t, e, "scroll event") {
+	if e, _ := o.popScrollEvent().(*fyne.ScrollEvent); assert.NotNil(t, e, "scroll event") {
 		assert.Equal(t, fyne.NewPos(50, 60), e.AbsolutePosition)
 		assert.Equal(t, fyne.NewPos(46, 56), e.Position)
 	}
--- a/internal/driver/glfw/window_test.go
+++ b/internal/driver/glfw/window_test.go
@@ -534,6 +534,22 @@ func TestWindow_HoverableOnDragging(t *testing.T) {
 	assert.NotNil(t, dh.popMouseOutEvent())
 }
 
+func TestWindow_Scrolled(t *testing.T) {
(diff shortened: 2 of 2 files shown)


internal/driver/glfw/window.go
```diff
@@ -838,6 +838,8 @@ func (w *window) mouseScrolled(viewport *glfw.Window, xoff float64, yoff float64
 		}
 		ev := &fyne.ScrollEvent{}
 		ev.Scrolled = fyne.NewDelta(float32(xoff)*scrollSpeed, float32(yoff)*scrollSpeed)
+		ev.Position = pos
+		ev.AbsolutePosition = w.mousePos
```
fpabl0, line 842: Could you add a TODO comment here? Just to be aware that when this is merged with the develop branch, it needs to be updated with locks. 🤔

fpabl0 (changes requested):

fpabl0:
I am not sure why, but I cannot make the tests pass locally, I always get:
```
=== RUN   TestWindow_Scrolled
    .../fyne/internal/driver/glfw/window_test.go:547:
        	Error Trace:	window_test.go:547
        	Error:      	Expected value not to be nil.
        	Test:       	TestWindow_Scrolled
        	Messages:   	scroll event
```

internal/driver/glfw/window_test.go
```diff
@@ -534,6 +534,22 @@ func TestWindow_HoverableOnDragging(t *testing.T) {
...
+	w := createWindow("Test").(*window)
+	o := &scrollable{Rectangle: canvas.NewRectangle(color.White)}
+	o.SetMinSize(fyne.NewSize(100, 100))
+	w.SetContent(o)
+
+	w.mousePos = fyne.NewPos(50, 60)
+	w.mouseScrolled(w.viewport, 10, 10)
+	w.waitForEvents()
+
+	if e, _ := o.popScrollEvent().(*fyne.PointEvent); assert.NotNil(t, e, "scroll event") {
```
fpabl0, line 547: Should be `fyne.ScrollEvent` instead of `fyne.PointEvent`. Not sure, how the tests passed before 🤔

fpabl0 (changes requested):

internal/driver/glfw/window.go
andydotxyz: I don't like to add TODO in PR if it can be avoided. I have written it on my merge list so it's addressed as soon as the release merges

internal/driver/glfw/window_test.go
andydotxyz: Indeed it should have been. However it works just fine because ScrollEvent is a PointEvent as well

andydotxyz:
> I am not sure why, but I cannot make the tests pass locally, I always get:
> 
> ```
> === RUN   TestWindow_Scrolled
>     .../fyne/internal/driver/glfw/window_test.go:547:
>         	Error Trace:	window_test.go:547
>         	Error:      	Expected value not to be nil.
>         	Test:       	TestWindow_Scrolled
>         	Messages:   	scroll event
> ```

Strange. Does it work if you apply this change?

```diff
--- a/internal/driver/glfw/window_test.go
+++ b/internal/driver/glfw/window_test.go
@@ -537,14 +537,14 @@ func TestWindow_HoverableOnDragging(t *testing.T) {
 func TestWindow_Scrolled(t *testing.T) {
        w := createWindow("Test").(*window)
        o := &scrollable{Rectangle: canvas.NewRectangle(color.White)}
-       o.SetMinSize(fyne.NewSize(100, 100))
        w.SetContent(o)
+       w.Resize(fyne.NewSize(100, 100))
```

fpabl0:
@andydotxyz 
> However it works just fine because ScrollEvent is a PointEvent as well

Hmm, not sure, as far as I know, that is not possible in any go version. When you assert a type by pointer, it should be exactly that type. Partial asserts is only possible when you are asserting behaviors (using interfaces) not values.

That test would fail in go playground too:
https://play.golang.org/p/-oldmGnFedA

fpabl0:
> Strange. Does it work if you apply this change?

The problem was only the type assertion, using `ScrollEvent` instead of `PointEvent` is enough :)

internal/driver/glfw/window.go
fpabl0: Great! :)

fpabl0 (approved):
Tests work now :). However I am still worried about the fact that it works before, it should not.

andydotxyz:
> However I am still worried about the fact that it works before, it should not.

Yes, I agree. It seems that test file is not running on CI - so the logical solution is that my git hooks did not fire AND that I did not run the test on its final version. Which is sadly possible.

Glad it's all running now :)

PR #2210, opened 2021-05-01
More extensible binding through public Untyped 

andydotxyz:
### Description:
Various updates to data binding that both enable and demonstrate binding custom types, 3 main areas:

* Custom type based on an existing basic type (`time.Time` using `Int`)
* Add `Untyped` binding so that developers could use `interface{}` bind directly
* Demonstration and test of binding `user` type by extending the new `Untyped` binding.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


data/binding/gen.go
```diff
@@ -675,6 +677,7 @@ import "fyne.io/fyne/v2"
 		bindValues{Name: "Int", Type: "int", Default: "0", Format: "%d", SupportsPreferences: true},
 		bindValues{Name: "Rune", Type: "rune", Default: "rune(0)"},
 		bindValues{Name: "String", Type: "string", Default: "\"\"", SupportsPreferences: true},
+		bindValues{Name: "Untyped", Type: "interface{}", Default: "struct{}{}", Since: "2.1"},
```
fpabl0, line 680: Any particular reason to have the default value set to "struct{}{}" instead of "nil"?

andydotxyz: No, good catch. It mattered in earlier versions of this code but should have been fixed for PR.
Updated now

data/binding/gen.go
```diff
@@ -33,8 +33,7 @@ type External{{ .Name }} interface {
 //
 // Since: {{ .Since }}
 func New{{ .Name }}() {{ .Name }} {
-	var blank {{ .Type }}
-	blank = {{ .Default }}
+	var blank {{ .Type }} = {{ .Default }}
```
fpabl0, line 36: I think this assignment is unneeded, should be enough this:
```go
var blank {{ .Type }}
```

data/binding/gen.go
```diff
@@ -44,8 +43,7 @@ func New{{ .Name }}() {{ .Name }} {
 // Since: {{ .Since }}
 func Bind{{ .Name }}(v *{{ .Type }}) External{{ .Name }} {
 	if v == nil {
-		var blank {{ .Type }}
-		blank = {{ .Default }}
+		var blank {{ .Type }} = {{ .Default }}
```
fpabl0, line 46: Same here

fpabl0 (changes requested):

data/binding/gen.go
andydotxyz: It depends on what the value of `.Default` is - working without it may be coincidence.
We'd be ignoring the default field, which doesn't seem right.

andydotxyz: I guess what I was meaning is whether that would be a different change. This PR was just adding a new type.

fpabl0: > I guess what I was meaning is whether that would be a different change. This PR was just adding a new type.

Oh sorry, I just thought you wanted to fix this too, because you modified it. Then it is ok, sorry for confusion.

fpabl0 (approved):

PR #2212, opened 2021-05-03
Optimisation: perform row-first scan when generating a raster image

AnkushJadhav:
### Description:
The current implementation of `NewRasterWithPixels` does a column-first scan to set the pixels of the raster image. This works fine for smaller rasters (less than approx 10000 pixels), but starts showing performance issues as the raster size increases. This is because large memory jumps have to be made to traverse the 2D array in a column-first manner.

As horizontally-adjacent data is stored closer in memory, a row-first scan takes advantage of cache coherence and provides a 3x better performance while scanning a 2D array (the raster image in this case).

Column first scan rendering:
column-first

Row first scan rendering:
row-first

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/canvas/raster.go
+++ b/canvas/raster.go
@@ -98,8 +98,8 @@ func NewRasterWithPixels(pixelColor func(x, y, w, h int) color.Color) *Raster {
 				pix.img = dst
 			}
 
-			for x := 0; x < w; x++ {
-				for y := 0; y < h; y++ {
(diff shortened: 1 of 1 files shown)


Jacalz (approved):
Great stuff. Thanks for the contribution 👍

andydotxyz (approved):
A most excellent catch :)

PR #2216, opened 2021-05-04
Add a windows 10 workaround to force dark mode

### LINKED ISSUE #2184: How to theme window title bar?
Select dark theme, but window title bar is not themed. My OS (win 10) system theme is dark mode.

### To Reproduce:

 - **OS:** Windows 10
 - **Version:** Enterprise build 19042
 - **Go version:** 1.16
 - **Fyne version:** 2.0.2

andydotxyz:
This seems to work well for my machines, looking for more testing

Fixes #2184 

### Checklist:

- [ ] Tests included. <- this is low level window manager overrides :(
- [x] Tests all pass.


internal/driver/glfw/window.go
```diff
@@ -1533,3 +1535,17 @@ func isKeyModifier(keyName fyne.KeyName) bool {
 		keyName == desktop.KeyAltLeft || keyName == desktop.KeyAltRight ||
 		keyName == desktop.KeySuperLeft || keyName == desktop.KeySuperRight
 }
+
+//lint:ignore U1000 This is fine, just a helper function for some platforms
```
Jacalz, line 1539: I'm wondering if there perhaps would be a better way to do this, without needing the ignore.

Jacalz (commented):
I don't know if fyne-cross didn't pick up my vendor folder or if it was this code, but I could not see dark title bars. Sorry for a less than optimal test of this.

andydotxyz:
> I don't know if fyne-cross didn't pick up my vendor folder or if it was this code, but I could not see dark title bars

I don't know how your vendor is working, but you could try compiling the hello (left in image) or fyne_demo instead?
I also used a replace directive on the terminal project (right in image) to show this working outside the fyne tree...

dark-windows

internal/driver/glfw/window.go
andydotxyz: Yes, the other possibility was to create a new file for this function. I have done that now

andydotxyz:
I realised that it was possible for FYNE_THEME=light to override the OS detected dark theme. This is now fixed and it might fix your test results @Jacalz .

Jacalz:
Thanks. I'll have a look. I was changing the theme back and forth using .SetTheme() in my app so that might have been the issue, yes.

andydotxyz:
Ah, yes the title bar should match OS setting

andydotxyz:
You marked this as request changes but its not clear what I can change.
Perhaps you can provide reproduction steps or a go mod so that I can replicate the issue?

I am confused about your comment for vendor and fyne_demo, this code is not in the Fyne vendor folder...?

Jacalz:
I just marked it as requesting changes because it didn't work. Sorry if I should't have done so.

As for the reproduce, it's not anything special. I just copied fyne_demo to a new folder. Replaced fyne.io/fyne/v2 with this PR and vendored it so that fyne-cross would use this code for fyne. Then I just ran it on Windows. Not much more to it.

I can't install all the build environments on my Windows machine as it is my school laptop.

andydotxyz:
Can't you just test fyne_demo in place (`fyne-cross windows ./cmd/fyne_demo`) rather than going through move/mod/vendor changes?
Trying to eliminate issues here because I cannot replicate it not working.

P.S. marking it changes requested is fine, but I could not progress it so was looking for more info.

Jacalz:
I don't think it works to do it that way. Without vendor information, it just creates a new module file in the directory and uses the latest stable release if I remember correctly. I put in a print line in the vendor folder (with -console passed to fyne-cross) to verify that it was being used, and it was.

andydotxyz:
I am still confused about what this has to do with vendor? cross-compiling the fyne_demo from cmd should use the top level go mod file and all of the code in the currently checked out version..?

Jacalz:
Sorry for a late reply. It doesn't quite have anything to do with vendor. I was just bringing it up because I couldn't get this PR to work during testing.

andydotxyz:
But it would be helpful if you could try `fyne_demo` or `hello` so that vendor / module lookup is eliminated from your troubles.

Jacalz:
I tried this once more, in the way that you suggest and there is no luck getting it working for me when using fyne_demo. The bar is till white no matter the theme. That it my school laptop though, so it might be a few versions behind on Windows 10. Will give it a try on my Gaming computer to see if there is any more luck there.

Jacalz (approved):
This did not initially work on my other computer either, but with an update to the latest major Windows 10 version, it worked flawlessly. Seems to be something odd with the older versions of Windows 10 then, but I would call that Microsoft's fault and not ours 🙂 Nice work 👍

PR #2221, opened 2021-05-05
Fix for #2220 except for widget/button.go's buttonColor()

### LINKED ISSUE #2220: Themed resources are drawn with the wrong colors if a theme's Color() method returns certain types of Colors
### Describe the bug:
I've written a fix for this, and a bunch of tests. (I'm also going to make a pull request after posting this)

Fyne displays incorrect colors in PrimaryThemedResources (such as checkboxes) which are SVGs, when a theme's Color method returns an RGBA64, NRGBA64, RGBA with 0<a<0xff, NRGBA with 0<a<0xff, Alpha16, Gray16, or any custom Color implementation where the least significant byte of each component returned by RGBA() is not equal to the most significant byte. The correct colors are only shown with RGBA with a=0 or a=0xff, Alpha, or Gray, and those only appear correct because their RGBA() method simply clones each 8 bit component of the color into both the high and low byte of the 16 bit output components (Go's (color.RGBA) RGBA() returns r<<8|r, g<<8|g, b<<8|b, a<<8|a, where r is uint32(c.R), etc). The reason it gives incorrect colors when 0<a<0xff for RGBA and NRGBA is because of the alpha-premultiplication (RGBA has it built into the components, unless an RGBA is incorrectly constructed without it, and NRGBA's RGBA() multiplies the alpha into the r, g, and b return values), and theme/svg.go's colorToHexAndOpacity's current code is written as if it expects RGBA() to return non-alpha-premultiplied values (it doesn't remove the alpha-premultiplication) whose most significant byte matches their least significant byte (it doesn't >>8 the return values from RGBA() before casting them to bytes).

I also found the same problem in other places in Fyne's code:
- dialog/base.go's themedBackgroundRenderer's Refresh()
- test/markup_renderer.go's markupRenderer's setColorAttrWithDefault
- test/markup_renderer.go's rgbaColor
- widget/button.go's buttonColor() (which appears to have another issue as well, which I'll have to make a separate issue about)

I noticed this because I was returning an (opaque) NRGBA64 from a custom Theme's Color method. I have a video of that below.

(When I looked into this, I found out that svgs can't support 64 bit colors anyways, but this still seems like a bug worth fixing, especially since not all resources are SVGs, and there's other code with the same problem that doesn't use SVGs.)

### To Reproduce:
1. Colorize a resource with NRGBA64{R: 0xff00, G: 0xffff, B: 0x00ff, A: 0xffff}.
2. Draw it.
3. It looks cyan when it should look yellow.
4. Do it again but give it an alpha value of 0x7fff.
5. It looks like a translucent light green color when it should look like a translucent yellow color.

### Video:
In this video, I change the value of an NRGBA64 color returned by a theme by dragging a hue slider (which goes from 0 to 360). The color created is opaque, and generated from converting HSL to RGB. You can see everything in the window change at the correct rate except the checkbox, which changes much more quickly because colorToHexAndOpacity is calling RGBA() and using the least significant byte of R, G, and B, instead of the most significant byte.
https://user-images.githubusercontent.com/6283593/116925385-20dd2d00-ac27-11eb-92d5-7a2c74528cfd.mp4

### Example code:
These are two of the subtests I added to theme/icons_test.go's TestColorizeResource. If colorToHexAndOpacity was working correctly, the first would be #ffff00, and the second would be #ffff00 with 50% alpha. Instead, the first is #00FFFF, and the second is #7fff7f with 50% alpha.
```go
		"NRGBA64": {
			svgFile: "circles.svg",
			color:     color.NRGBA64{R: 0xff00, G: 0xffff, B: 0x00ff, A: 0xffff},
			wantImage: "colorized/circles_yellow.png",
		},
		"translucent NRGBA64": {
			svgFile:   "circles.svg",
			color:     color.NRGBA64{R: 0xff00, G: 0xffff, B: 0x00ff, A: 0x7fff},
			wantImage: "colorized/circles_yellow_translucent.png",
		},
```

 - **OS:** Windows 10 Pro
 - **Version:** 20H2
 - **Go version:** 1.16.3 windows/amd64
 - **Fyne version:** 083f0344089df647ed3a2e33b8b77e0972b75517 (which is the last commit on the develop branch at time of writing)

AmanitaVerna:
### Description:
- Moves colorToRGBA and unmultiplyRGBA into theme/theme.go and renames colorToRGBA to ColorToNRGBA so everything can access it (and so the name is accurate).
- Makes improvements to both functions to make them handle all stock Go Color types without needing to un-premultiply most of them, and to make unmultiplyAlpha more accurate and efficient. 
- Corrects the following methods or functions:
- theme/svg.go's colorToHexAndOpacity()
- dialog/base.go's themedBackgroundRenderer's Refresh()
- test/markup_renderer.go's markupRenderer's setColorAttrWithDefault()
- test/markup_renderer.go's rgbaColor(), and also renames it to nrgbaColor
- Adds a bunch of test cases and four test images to theme/icons_test.go's TestColorizeResource().
- Moves Test_UnmultiplyAlpha so it can still access unmultiplyAlpha. 

Fixes #2220 except for widget/button.go's buttonColor(), which has an additional issue

### Checklist:
- [x] Tests included.
- [x] Lint and formatter run with no errors. (lint, anyways. I have auto-formatting on through vs code's go extension. I tried manually running gofmt on a single file, piping the output to a new file instead of letting it overwrite the existing one, and got what github desktop sees as a "binary file" (wtf?), so I figured maybe I shouldn't tell gofmt to change anything)
- [x] Tests all pass.

#### Where applicable:
- [ ] Public APIs match existing style? (I have no idea if ColorToNRGBA matches the existing API style; I hadn't noticed there was one)


AmanitaVerna:
I attempted to sign the CLA as mentioned in https://github.com/fyne-io/fyne/wiki/Contributing, but it said "This account has not been added to the early access program."

theme/theme.go
```diff
@@ -545,6 +545,98 @@ func PrimaryColorNamed(name string) color.Color {
 	return col
 }
 
+//ColorToNRGBA converts a color to RGBA values which are not premultiplied, unlike color.RGBA().
+func ColorToNRGBA(c color.Color) (r, g, b, a int) {
```
andydotxyz, line 549: I'm not sure this warrants a new public API - perhaps it could be somewhere in the `internal/` space?

andydotxyz:
Thanks so much for tidying this all up. Don't worry about the CLA right now - we only require it for members of the core teams. Of course if you would like to sign it anyway just say and I can add you to the list of accounts that can access that page ;).

I'm not too sure why but this PR seems to include changes outside your work - I don't think you should have merged master in to develop.

AmanitaVerna:
I don't see any changes from that - Did it make some and I'm not seeing them? How that happened is that I'm still new to git and github (and I'm using github desktop to try to make it easier), and I was attempting to figure out how to make sure my forked develop branch was up to date with fyne's, so that I could keep it up to date (I figured it out right after). I think that was right after I had made the fork, though I'm not certain.

theme/theme.go
AmanitaVerna: Is there anywhere in particular in internal/ that you would like it to go? Also, unmultiplyAlpha and Test_UnmultiplyAlpha will have to go with it, of course.

andydotxyz: Possibly an `internal/color.go` file? we need to tidy that area up a little so other suggestions welcome.

andydotxyz:
Don't worry there is a lot to learn. I think you created a branch from develop but then updated it from master which is where things got confused. See these items:

Screenshot 2021-05-06 at 09 45 38

It will probably be too complicated to unwind that at this stage, I think we can make it work. Just watch out for future PRs.

theme/theme.go
AmanitaVerna: Since `internal/color.go` results in an import cycle error, how about `internal/col/color.go`? That way there's no possibility of one. Alternately, `internal/color/color.go`, since 'col' is not a super obvious name, in which case everything importing it would have to import it like `import col internal/color` because they all also import `image/color`.

AmanitaVerna:
Ah, okay, I see.
It looks like I have some failing tests to look at, as well.

AmanitaVerna:
Okay, these failing tests were all creating color.RGBA structs directly, and expecting that to work fine (it was working before only because setColorAttrWithDefault just directly used the return value from RGBA() with no attempt at un-premultiplying). 

I changed them all to make NRGBAs instead, and the tests all pass on my end now. I've sent it up to github, along with moving ColorToNRGBA and unmultiplyAlpha into internal/color/color.go. (Github says it's waiting for status to be reported on the tests atm)

theme/theme.go
AmanitaVerna: I've put it in internal/color/color.go since 'col' could mean anything.

andydotxyz (approved):
This looks like a great improvement - who knew we messed up so many colour conversions without realising!
Fantastic to have someone who really knows this stuff :).
I'm happy with the code, would love to get one other opinion before merging though.

andydotxyz:
Looks like staticcheck has a naming issue - I agree with it's recommendation

andydotxyz (approved):

Jacalz (changes requested):
Comments are incorrectly formatted. Should have a space after `//`.

AmanitaVerna:
Okay, I've added a space after the //s that I had added. I think I got all of them.

theme/icons_test.go
```diff
@@ -175,44 +175,114 @@ func TestColorizeResource(t *testing.T) {
...
 		},
 		"group of rects": {
 			svgFile:   "info_GroupRects.svg",
-			color:     color.RGBA{R: 100, G: 100, B: 100, A: 100},
+			color:     color.NRGBA{R: 100, G: 100, B: 100, A: 100},
 			wantImage: "colorized/group_rects.png",
 		},
+		"NRGBA64": {
+			svgFile: "circles.svg",
+			//If the low 8 bits of each component were used, this would look cyan instead of yellow.
```
Jacalz, line 218: There are still incorrectly formatted comments in this file.

Jacalz (changes requested):

Jacalz (approved):
Looks good to me now. Nice work.

AmanitaVerna:
Thanks. Huh, that's odd, why is that failing now? It was passing before, and the only thing that I've changed since it last passed is the comments.

andydotxyz:
Unfortunately there are one or two temperamental tests, I have restarted them which usually helps

andydotxyz (approved):

stuartmscott (approved):

PR #2223, opened 2021-05-06
Now we have standard padded widgets we don't need padding in collections

andydotxyz:
I think the 2.0 larger widgets meant that collections are too padded.
This is now addressed.

I re-added a leading padding to tree so the icons were not (almost) flush to edge.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (commented):
Hmm, I am not so sure about this. Seems like quite a breaking change. While I think it probably makes https://github.com/fyne-io/fyne/pull/2181 work better, it makes wormhole-gui behave significantly worse.
image

andydotxyz:
I’m sorry but I can’t see what is broken in the screenshot

stuartmscott (approved):
As agreed on Slack, this should not land in a patch release, and instead should target the next minor release (most likely 2.1.0).

andydotxyz:
Thanks @stuartmscott that was my intent. I guess the original message was not that clear. Thanks

PR #2234, opened 2021-05-10
Fix for #2222, and #2220 for buttons

### LINKED ISSUE #2220: Themed resources are drawn with the wrong colors if a theme's Color() method returns certain types of Colors

### LINKED ISSUE #2222: buttonRenderer's buttonColor can generate out of range alpha values when hovered (might affect RGB too?)
### Describe the bug:
The math in widget/button.go's buttonRenderer's buttonColor() for color calculations while hovering appears to be very wrong (in addition to being affected by #2220), but I'm not sure what it is actually trying to do.

At present it is doing this when a button is hovered:
```go
		bg := theme.ButtonColor()
		if r.button.Importance == HighImportance {
			bg = theme.PrimaryColor()
		}
		dstR, dstG, dstB, dstA := bg.RGBA()
		srcR, srcG, srcB, srcA := theme.HoverColor().RGBA()
		srcAlpha := float32(srcA) / 0xFFFF
		dstAlpha := float32(dstA) / 0xFFFF
		targetAlpha := 1 - srcAlpha*dstAlpha
		outAlpha := srcAlpha + targetAlpha
		outR := (srcAlpha*float32(srcR) + targetAlpha*float32(dstR)) / outAlpha
		outG := (srcAlpha*float32(srcG) + targetAlpha*float32(dstG)) / outAlpha
		outB := (srcAlpha*float32(srcB) + targetAlpha*float32(dstB)) / outAlpha
		return color.RGBA{R: uint8(uint32(outR) >> 8), G: uint8(uint32(outG) >> 8), B: uint8(uint32(outB) >> 8), A: uint8(outAlpha * 0xFF)}
```

The problem here is that the math doesn't make any sense, even if it's accounting for premultiplication (I'm not sure if it is or not). You can see that if you just look at what's going on with the alpha values:
Let's assume srcA=51, dstA=204.
srcAlpha will be 0.2, and dstAlpha will be 0.8.
targetAlpha = 1 - 0.2 \* 0.8 = 0.84
outAlpha = 0.2 + 0.84 = 1.04
The alpha value that is used in the new color.RGBA will be uint8(1.04\*0xFF).
1.04\*0xFF is 265.2, or 0x109, and so when it's cut off to just 8 bits, the color.RGBA will have an alpha value of 0x09, which is equivalent to if outAlpha had been 0.035.

So that seems like an obvious bug, but less obvious is what the code is *supposed* to be doing.
I checked all the other things that can be hovered, but they pretty much all just set their color to HoverColor. This seems to be something new.

(This doesn't even get into how the alpha calculations may be screwing up the colors too - they'd be screwed up anyways from just having alpha due to #2220, which I haven't fixed for this method)

### To Reproduce:
1. If you want to see it visually, try creating a theme whose background color is 0xff0000, button color is 0x00ff00 with alpha 204, and hover color is 0x0000ff with alpha 51, and a window with a button on it that you can move the mouse over to see what happens. Given that #2220 also applies, the results will probably be confusing. Also you can't actually see the alpha values this way, you'd have to try to infer them, which would be difficult given that the color values won't necessarily make any sense either...

 - **OS:** Windows 10 Pro
 - **Version:** 20H2
 - **Go version:** 1.16.3 windows/amd64
 - **Fyne version:** 083f0344089df647ed3a2e33b8b77e0972b75517

AmanitaVerna:
### Description:
- Replaces the math used for blending the hover color in button.go's buttonRenderer's buttonColor() to fix #2222 and also to fix #2220 for buttonColor().
- Fixes #2220 in button.go's newButtonTapAnimation() as well.
- Adds a math-based test which uses an alternate way of calculating the same blending operation to test buttonRenderer's hover code against (TestButton_Hover_Math in button_internal_test.go)
- Updates the hover test image used by button_test.go's TestButton_Hover()

Fixes #2222, and also fixes the last cases of #2220 that I've found.

### Checklist:
- [x] Tests included.
- [x] Tests all pass. (except when they randomly fail, but that's unrelated to this code. I saw it fail on table and list tests several times, fail to render a set of lines on tables once, and render the button icon in the wrong place once, and I have no idea why any of that is sometimes happening.)


AmanitaVerna:
I missed some tests which weren't in widget or widget_test, which are expecting wrong RGB values (calculated the previous, incorrect, way) or images, and I'll have to fix those.

andydotxyz:
Sorry but the history of this PR is more messed up than the last one. Are you branching from `master` by mistake?

AmanitaVerna:
I took the develop branch I already had, merged upstream/develop into it, and then merged the button changes into it.

I should just delete the branch and recreate it, probably.

andydotxyz:
It looks like your develop may not line up with our develop then.

I would recommend removing it and checking it out from the upstream.
You can then make a branch for your commits and open a PR from there.
For future PRs then just merge upstream into develop and branch off again.

PR #2236, opened 2021-05-11
Feature/lifecycle

### LINKED ISSUE #1053: Add lifecycle support to applications
To support better handling of low-power background operations, or simply pausing games etc when focus is lost we should add some lifecycle handling.

I think that this should handle:
* OnBackground (or OnWillBackground?)
* OnForground (or OnDidForeground?)

I feel like more should be added, but cannot think just now. Apps can already handle DidStart and WillStop by doing:
```go
func main() {
    a := app.New()
    DidStart()
    a.ShowAndRun()
    WillStop()
}
```

andydotxyz:
### Description:
Simple lifecycle API for start/stop and focus gained/lost.

Question: Should it be OnForegrounded & OnBackgrounded instead (just felt too wordy...)

Fixes #1053

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
Maybe OnEnteredBackground and OnEnteredForeground?

stuartmscott (approved):
LGTM

In future we may want to add `SetOnLowMemory` so devs can free some resources, fyne could also purge the texture cache.

andydotxyz:
Agreed. I am thinking about something for timezone/DST changes as well

PR #2237, opened 2021-05-11
Support Dynamic Split Contents

stuartmscott:
### Description:
No longer caches object slice so devs can dynamically update Leading and Trailing contents.

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/container/split.go
+++ b/container/split.go
@@ -151,7 +151,7 @@ func (r *splitContainerRenderer) Objects() []fyne.CanvasObject {
 func (r *splitContainerRenderer) Refresh() {
 	r.objects[0] = r.split.Leading
 	// [1] is divider which doesn't change
-	r.objects[2] = r.split.Leading
+	r.objects[2] = r.split.Trailing
 	r.Layout(r.split.Size())
 	canvas.Refresh(r.split)
 }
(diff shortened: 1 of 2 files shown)


andydotxyz (commented):
This is a great change, thanks.
However I have (and it's not specific to this commit) started to worry about allocating slices in `Objects()` call. I wonder if, because this is static in size, whether it would be worth allocating objects as `fyne.CanvasObject[3]` and setting the leading and trailing elements into this on `Refresh`.
Maybe I'm being overly concerned about the overhead, but thought I would ask.

stuartmscott:
That's a fair point. `Objects()` does get called quite frequently so it would be a win to only allocate once.

I noticed scroller does this;
```go
if len(r.BaseRenderer.Objects()) == 0 || r.BaseRenderer.Objects()[0] != r.scroll.Content {
	// push updated content object to baseRenderer
	r.BaseRenderer.SetObjects([]fyne.CanvasObject{r.scroll.Content})
}
```

But I think we could just set the indices in `Refresh()` as you suggested.

andydotxyz:
Yes, the internal BaseRenderer is problematic. Something I intend to review since realising BaseWidget and the internal Base Widget are not compatible (we cannot extend internal widgets). Something that I think needs tidying up.

andydotxyz (commented):
Sweet.
Would it be worth adding a quick test to demonstrate the fix?

andydotxyz (approved):

PR #2239, opened 2021-05-13
Support for fetching images from remote server

### LINKED ISSUE #144: Add support for loading images from a url
Add appropriate caching so that images can work offline and load faster

AnkushJadhav:
### Description:
Adds `HTTPRepository` to supported storage repositories for using http and https as valid schemes for accessing resources.
Currently a ReadOnly implementation has been done for this repository with the assumption that the remote resource is accessible using the `GET` http method.

Adds support to `canvas.NewImageFromURI` to load images from remote servers.

Fixes #144 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


canvas/image.go
```diff
@@ -107,6 +109,17 @@ func NewImageFromURI(uri fyne.URI) *Image {
 		}
 	}
 
+	if uri.Scheme() == "http" || uri.Scheme() == "https" {
```
andydotxyz, line 112: Do we need to do this?
I know that HEAD is there to help us avoid downloading data, but in this case the developer probably knows it exists already. So how about we do the Go thing and just make the GET request and handle error if isn't there?

andydotxyz (changes requested):
This looks great, but I think it can work well without the extra http specific code in `NewImageFromURI`

canvas/image.go
AnkushJadhav: That makes sense. I was thinking more from a HTTP perspective.

andydotxyz (approved):

andydotxyz:
Nice one thanks :)

PR #2244, opened 2021-05-15
Fix broken test in glfw/window_test.go

### LINKED ISSUE #2243: TestWindow_HandleOutsideHoverableObject is failing
### Describe the bug:
When I run package tests in internal/driver/glfw (or directly run this test), TestWindow_HandleOutsideHoverableObject fails, saying "Expected value not to be nil" on line 291 of window_test.go. I debugged it, and while walking the tree of objects FindObjectAtPositionMatching in internal/driver/util.go finds the hoverable ListItem and saves it to found and foundPos in a very side-effecty way, and then later it finds a non-hoverable (but draggable?) Entry and saves that to found and foundPos, overwriting the saved ListItem, with the result that it ends up setting no mouseOver. This is with no changes to the code which is currently on fyne-io/develop (I was getting it with the code I wrote to fix #2222 too, and was trying to figure out how I could have caused it, but then when I ran the test with just the code on fyne-io/develop, it still happened). 

I have no idea why this is happening for me when I run the tests, but didn't happen during the tests that were run on the last commit to develop, though. I did just upgrade to Go 1.16.4 today, but from following the code the test runs, it looks like it's just a bug in fyne?

### To Reproduce:
1. Run the test TestWindow_HandleOutsideHoverableObject in internal/driver/glfw/window_test.go.

 - **OS:** Windows 10 Pro
 - **Version:** 20H2
 - **Go version:** 1.16.4
 - **Fyne version:** 6b7d4bf6f0ed3864d3f650fcda23e06c24be3a38

fpabl0:
### Description:

Fixes #2243

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/testdata/windows_hover_object.xml
+++ b/internal/driver/glfw/testdata/windows_hover_object.xml
@@ -3,34 +3,34 @@
 		<widget pos="4,4" size="192x292" type="*widget.List">
 			<widget size="192x292" type="*widget.Scroll">
 				<container size="192x292">
-					<widget size="192x45" type="*widget.listItem">
-						<widget pos="8,4" size="180x37" type="*widget.Entry">
-							<rectangle fillColor="rgba(255,255,255,25)" pos="0,2" size="180x33"/>
-							<rectangle fillColor="shadow" pos="0,35" size="180x2"/>
-							<widget pos="0,2" size="180x33" type="*widget.Scroll">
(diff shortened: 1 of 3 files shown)


andydotxyz (approved):
Thanks for this. Very strange that it didn’t get caught in the actual padding change.

PR #2245, opened 2021-05-15
Fix Entry with NotEmpty Validator and Form Hints

### LINKED ISSUE #2179: Entry validation does not work for empty field (?)
Hi, 

I'm using fyne 2.0.2, go 1.14 on macos. Sorry, if the question is studip, I am new to fyne and go:

I'd like to have an Entry where the entered value needs to have at least 2 characters. The validation works ok, if one inputs at least one character, but in case the Entry is left completely empty, no validation is done.

What I am doing wrong?
```
package main

import (
	"errors"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/layout"
	"fyne.io/fyne/v2/widget"
)

func main() {
	app := app.New()

	w := app.NewWindow("Hello")
	e := widget.NewEntry()
	e.Validator = func (str string) error {
		if len(e.Text)<2 {
			return errors.New("Too short")
		}
		return nil
	}

	content := container.New(layout.NewHBoxLayout(),  e)
	w.SetContent(content)

	w.ShowAndRun()
}
```

fpabl0:
### Description:

Fixes #2179. And also a previous bug with the form hints:

before:
before

after this PR:
after

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/entry.go
```diff
@@ -59,12 +59,17 @@ type Entry struct {
...
 	cursorAnim *entryCursorAnimation
 
+	dirty       bool
 	focused     bool
 	text        *textProvider
 	placeholder *textProvider
 	content     *entryContent
 	scroll      *widget.Scroll
 
+	// useful for Form validation (as the error text should only be shown when
```
andydotxyz, line 69: This is a change in behaviour - before now we would show the details of an error if there was any.
We hid the error status if the user had not yet edited the field.

Is this the desired change?

fpabl0: > now we would show the details of an error if there was any.

Not sure what you mean here, that is already the entry behavior.

> We hid the error status if the user had not yet edited the field.

The only behavior change is that now Entry is going to show the error status (if there is any error) after the user has visited the input by the first time. In that way we can ensure that "not empty" validators are shown as soon as the user focused the input and passed it without solving the error.

andydotxyz (approved):
Thanks so much. Tapping through this works nicely.

andydotxyz:
Merged into `release/v2.0.x` as well :)

PR #2246, opened 2021-05-15
Wait for goroutines to finish when firing a change to an app preferences 

### LINKED ISSUE #2241: Preferences file gets unexpectedly cleared
### Describe the bug:
When reading preferences, sometimes the `preferences.json` file gets suddenly cleared. This does not seem to happen if the app window is displayed as in most cases it would be. I stumbled on this when trying to add simple CLI flags that would show stored preferences without displaying normal GUI. What's weird is that it often works as expected (prefs are read successfully) and only once in a couple of runs the preferences file gets destroyed. 

### To Reproduce:
1. Read preferences without showing app window
2. Execute the app a couple of times
3. See the preferences file

### Screenshots:
Here I'm running the attached sample code a couple of times before the prefs file gets emptied:
fyne-prefs-bug

### Example code:
This code assumes that preferences file is already available at `~/.config/fyne/com.example.fyne.prefs/preferences.json` with content like: `{"FOO":"BAR"}`

```go
package main

import (
	"fmt"

	"fyne.io/fyne/v2/app"
)

func main() {
	a := app.NewWithID("com.example.fyne.prefs")
	p := a.Preferences()

	if val := p.String("FOO"); val != "" {
		fmt.Println("setting found:", val)
	}
}
```

 - **OS:** Linux
 - **Version:** Ubuntu 20.10 
 - **Go version:** 1.16.3
 - **Fyne version:** 2.0.3

s77rt:
### Description:
Mainly there was two problems:
- the first and probably the main bug is that sometimes the app closes too early without waiting for the goroutines that were called in [fireChange()](https://github.com/fyne-io/fyne/blob/3814a66f75ff5d24be6fa107f1dfb9c4faa859b1/internal/preferences.go#L69) and one of those goroutines is the one that calls [save()](https://github.com/fyne-io/fyne/blob/3814a66f75ff5d24be6fa107f1dfb9c4faa859b1/app/preferences.go#L35) an easy fix was to implement the sync.WaitGroup functionality which i expected to work just fine, yet it didn't due to
- the second issue which was a misuse of the lockers, in lines [70 and 71](https://github.com/fyne-io/fyne/blob/3814a66f75ff5d24be6fa107f1dfb9c4faa859b1/internal/preferences.go#L70-L71) are using Lock() and after that there are goroutines that are expected to read the same struct, well they won't be able to because Lock() only allows one goroutine (read/write) at a time and in fact the goroutines functions that were thought are being executed were actually not, they were just in stuck position waiting for the fireChange() to return (and to trigger it's defer function to Unlock()) so all those go l() were only waiting for fireChange() to return then get executed, well after the fix (1) the function will never return because it's waiting for the goroutines functions which will never work either because they are waiting for fireChange() ,,, deadlock ; the fix was pretty clear is to use RLock() instead of Lock() that way, where multiple goroutines can read(but not write) the struct and the fireChange() will only returns after all the goroutines are done and only after that the app will close

Fixes #2241

### Checklist:

- [ ] Tests included.
- [X] Tests all pass. (except  TestFileIcon(?))

### DIFF
--- a/internal/preferences.go
+++ b/internal/preferences.go
@@ -11,17 +11,22 @@ type InMemoryPreferences struct {
 	values          map[string]interface{}
 	lock            sync.RWMutex
 	changeListeners []func()
+	waitGroup       *sync.WaitGroup
 }
 
 // Declare conformity with Preferences interface
 var _ fyne.Preferences = (*InMemoryPreferences)(nil)
 
 // AddChangeListener allows code to be notified when some preferences change. This will fire on any update.
+// listener should not try to write values
 func (p *InMemoryPreferences) AddChangeListener(listener func()) {
 	p.lock.Lock()
 	defer p.lock.Unlock()
 
-	p.changeListeners = append(p.changeListeners, listener)
+	p.changeListeners = append(p.changeListeners, func() {
+		defer p.waitGroup.Done()
+		listener()
+	})
 }
 
 // ReadValues provides read access to the underlying value map - for internal use only...
@@ -67,12 +72,15 @@ func (p *InMemoryPreferences) remove(key string) {
 }
 
 func (p *InMemoryPreferences) fireChange() {
-	p.lock.Lock()
-	defer p.lock.Unlock()
+	p.lock.RLock()
+	defer p.lock.RUnlock()
 
 	for _, l := range p.changeListeners {
+		p.waitGroup.Add(1)
 		go l()
 	}
+
+	p.waitGroup.Wait()
 }
 
 // Bool looks up a boolean value for the key
@@ -180,5 +188,6 @@ func (p *InMemoryPreferences) RemoveValue(key string) {
 func NewInMemoryPreferences() *InMemoryPreferences {
 	p := &InMemoryPreferences{}
 	p.values = make(map[string]interface{})
+	p.waitGroup = &sync.WaitGroup{}
 	return p
 }
--- a/internal/preferences.go
+++ b/internal/preferences.go
@@ -23,10 +23,7 @@ func (p *InMemoryPreferences) AddChangeListener(listener func()) {
 	p.lock.Lock()
 	defer p.lock.Unlock()
 
-	p.changeListeners = append(p.changeListeners, func() {
-		defer p.wg.Done()
-		listener()
-	})
+	p.changeListeners = append(p.changeListeners, listener)
 }
 
 // ReadValues provides read access to the underlying value map - for internal use only...
@@ -77,7 +74,10 @@ func (p *InMemoryPreferences) fireChange() {
 
 	for _, l := range p.changeListeners {
 		p.wg.Add(1)
-		go l()
+		go func() {
+			defer p.wg.Done()
+			l()
+		}()
 	}
 
 	p.wg.Wait()
--- a/internal/preferences.go
+++ b/internal/preferences.go
@@ -11,7 +11,7 @@ type InMemoryPreferences struct {
 	values          map[string]interface{}
 	lock            sync.RWMutex
 	changeListeners []func()
-	waitGroup       *sync.WaitGroup
+	wg              *sync.WaitGroup
 }
 
 // Declare conformity with Preferences interface
@@ -24,7 +24,7 @@ func (p *InMemoryPreferences) AddChangeListener(listener func()) {
 	defer p.lock.Unlock()
 
 	p.changeListeners = append(p.changeListeners, func() {
-		defer p.waitGroup.Done()
+		defer p.wg.Done()
 		listener()
 	})
 }
@@ -76,11 +76,11 @@ func (p *InMemoryPreferences) fireChange() {
 	defer p.lock.RUnlock()
 
 	for _, l := range p.changeListeners {
-		p.waitGroup.Add(1)
+		p.wg.Add(1)
 		go l()
 	}
 
-	p.waitGroup.Wait()
+	p.wg.Wait()
 }
 
 // Bool looks up a boolean value for the key
@@ -187,7 +187,7 @@ func (p *InMemoryPreferences) RemoveValue(key string) {
 // NewInMemoryPreferences creates a new preferences implementation stored in memory
 func NewInMemoryPreferences() *InMemoryPreferences {
 	return &InMemoryPreferences{
-		values:    make(map[string]interface{}),
-		waitGroup: &sync.WaitGroup{},
+		values: make(map[string]interface{}),
+		wg:     &sync.WaitGroup{},
 	}
 }
--- a/internal/preferences.go
+++ b/internal/preferences.go
@@ -186,8 +186,9 @@ func (p *InMemoryPreferences) RemoveValue(key string) {
 
 // NewInMemoryPreferences creates a new preferences implementation stored in memory
 func NewInMemoryPreferences() *InMemoryPreferences {
-	p := &InMemoryPreferences{}
-	p.values = make(map[string]interface{})
-	p.waitGroup = &sync.WaitGroup{}
+	p := &InMemoryPreferences{
+		values: make(map[string]interface{}),
+		waitGroup: &sync.WaitGroup{},
+	}
 	return p
 }
--- a/internal/preferences.go
+++ b/internal/preferences.go
@@ -74,10 +74,10 @@ func (p *InMemoryPreferences) fireChange() {
 
 	for _, l := range p.changeListeners {
 		p.wg.Add(1)
-		go func() {
+		go func(listener func()) {
 			defer p.wg.Done()
-			l()
-		}()
+			listener()
+		}(l)
 	}
 
 	p.wg.Wait()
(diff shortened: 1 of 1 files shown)


Jacalz:
Just a minor thought not related to the contents. Can you please name the title of the PR (this and in the future) to what it does and not just what it fixes? It is very hard to glance over numbers and remember what it does.

internal/preferences.go
```diff
@@ -180,5 +188,6 @@ func (p *InMemoryPreferences) RemoveValue(key string) {
 func NewInMemoryPreferences() *InMemoryPreferences {
 	p := &InMemoryPreferences{}
 	p.values = make(map[string]interface{})
+	p.waitGroup = &sync.WaitGroup{}
 	return p
```
Jacalz, line 183:
```go
 	return &InMemoryPreferences{
		values: make(map[string]interface{}),
		waitGroup: &sync.WaitGroup{},
	}
```
I think this would be a bit cleaner while you're at it.

internal/preferences.go
```diff
@@ -11,17 +11,22 @@ type InMemoryPreferences struct {
...
 
 // AddChangeListener allows code to be notified when some preferences change. This will fire on any update.
+// listener should not try to write values
 func (p *InMemoryPreferences) AddChangeListener(listener func()) {
 	p.lock.Lock()
 	defer p.lock.Unlock()
 
-	p.changeListeners = append(p.changeListeners, listener)
+	p.changeListeners = append(p.changeListeners, func() {
+		defer p.waitGroup.Done()
```
Jacalz, line 27: Can't we just place it after the other function instead of deferring?

Jacalz (commented):
Just some small suggestions.

s77rt:
> Just a minor thought not related to the contents. Can you please name the title of the PR (this and in the future) to what it does and not just what it fixes? It is very hard to glance over numbers and remember what it does.

ah sorry for that, the title has been changed

internal/preferences.go
s77rt: No, it's safer that way
a deferred function will always be evaluated now and executed later, while if I placed it after the call to `listener()` it may not get executed if the called function raised a panic

internal/preferences.go
s77rt: sure

internal/preferences.go
Jacalz: Ah, I see. Thanks 👍

internal/preferences.go
Jacalz: It would be cleaner if you return it directly instead of storing it to a variable first.

Jacalz: Goimports also complains about the formatting.

s77rt: updated

Jacalz (approved):
This looks good to me. I'll let someone else look on it as well though because I'm away from my computer and can't test it locally.

internal/preferences.go
andydotxyz, line 27: I think this lacks symettery or proximity of paired operations.
If `p.waitGroup` is ever written, or if `p.changeListeners` ever has a different element added then `fireChange` will hang. I suggested an alternative below.

internal/preferences.go
```diff
@@ -67,12 +72,15 @@ func (p *InMemoryPreferences) remove(key string) {
 }
 
 func (p *InMemoryPreferences) fireChange() {
-	p.lock.Lock()
-	defer p.lock.Unlock()
+	p.lock.RLock()
+	defer p.lock.RUnlock()
 
 	for _, l := range p.changeListeners {
+		p.waitGroup.Add(1)
```
andydotxyz, line 79: It would seem cleaner (assuming it's not to big an overhead) to put the Done inside an anonymous goroutine that then calls `l()` here so that we have code symmetry and safety against the potential issues I mentioned above.

internal/preferences.go
```diff
@@ -178,7 +186,8 @@ func (p *InMemoryPreferences) RemoveValue(key string) {
 
 // NewInMemoryPreferences creates a new preferences implementation stored in memory
 func NewInMemoryPreferences() *InMemoryPreferences {
-	p := &InMemoryPreferences{}
-	p.values = make(map[string]interface{})
-	return p
+	return &InMemoryPreferences{
+		values:    make(map[string]interface{}),
+		waitGroup: &sync.WaitGroup{},
```
andydotxyz, line 191: Go recommends not using a variable name that is the same as the type - can you think of any shorter alternatives?

andydotxyz (changes requested):

andydotxyz:
When running this I get deadlocks at runtime in an app that leans heavily on preferences. You can check out https://github.com/fynelabs/notes, when running (after a few edits) I see:

```
❯ go run .
2021/05/17 09:17:44 Fyne error:  Preferences load error:
2021/05/17 09:17:44   Cause: EOF
2021/05/17 09:17:44   At: /Users/andy/Code/Fyne/fyne/app/preferences.go:70
2021/05/17 09:17:49 Fyne error:  Preferences load error:
2021/05/17 09:17:49   Cause: EOF
2021/05/17 09:17:49   At: /Users/andy/Code/Fyne/fyne/app/preferences.go:70
2021/05/17 09:17:57 Fyne error:  Preferences load error:
2021/05/17 09:17:57   Cause: EOF
2021/05/17 09:17:57   At: /Users/andy/Code/Fyne/fyne/app/preferences.go:70
```

internal/preferences.go
s77rt: I understood how this will make the code in fact much clearer, but i didn't get what was the issue exactly, can you please elaborate?

andydotxyz: When operations happen in pairs (`Add`, `Done` or `Open` and `Close`) it is desirable to keep them together so that it is clear that resources are freed. By separating these it is easy for bugs to be introduced or for debugging to take longer.

internal/preferences.go
s77rt: Would using `wg` be okay?

andydotxyz: Yes, I think we have done that elsewhere

internal/preferences.go
s77rt: Got it, thanks
I will make the changes asap

s77rt:
> When running this I get deadlocks at runtime in an app that leans heavily on preferences. You can check out https://github.com/fynelabs/notes, when running (after a few edits) I see:
> 
> ```
> ❯ go run .
> 2021/05/17 09:17:44 Fyne error:  Preferences load error:
> 2021/05/17 09:17:44   Cause: EOF
> 2021/05/17 09:17:44   At: /Users/andy/Code/Fyne/fyne/app/preferences.go:70
> 2021/05/17 09:17:49 Fyne error:  Preferences load error:
> 2021/05/17 09:17:49   Cause: EOF
> 2021/05/17 09:17:49   At: /Users/andy/Code/Fyne/fyne/app/preferences.go:70
> 2021/05/17 09:17:57 Fyne error:  Preferences load error:
> 2021/05/17 09:17:57   Cause: EOF
> 2021/05/17 09:17:57   At: /Users/andy/Code/Fyne/fyne/app/preferences.go:70
> ```
> 
> That was when trying to quit the app, the window disappeared but the process did not exit.
> Could it be related? It does not happen every time but I am pretty sure I have not seen it before.
> 
> Oh, apologies I should have said earlier, thanks for this - I had wondered if something was incorrect as I realised that sometimes this notes app did not save the final character when making notes and exiting!

I have tried your app and i can't replicate the issue, though sometimes the app does not save the whole string (as you said some chars are left) (some saves are not being saved)

internal/preferences.go
```diff
@@ -11,12 +11,14 @@ type InMemoryPreferences struct {
...
 	lock            sync.RWMutex
 	changeListeners []func()
+	wg              *sync.WaitGroup
 }
 
 // Declare conformity with Preferences interface
 var _ fyne.Preferences = (*InMemoryPreferences)(nil)
 
 // AddChangeListener allows code to be notified when some preferences change. This will fire on any update.
+// listener should not try to write values
```
andydotxyz, line 21: This line added to the docs is not formatted like a complete sentence.

andydotxyz (changes requested):
Thanks the code looks good now, but I missed that line of documentation isn't particularly clear.

internal/preferences.go
s77rt: what would be a better sentence?
the line is meant to indicate that the function `listener` should not try to write to `InMemoryPreferences`

andydotxyz: Maybe I mis-read it, in which case it's just missing the '.'.
I would imagine that it could be slightly easier to read by starting with "The passed 'listener'".

andydotxyz (approved):
Nice thanks

Jacalz:
I will have a look at this again later today. Want to make sure that it works for my use cases before we merge it :)

Jacalz (approved):
Works perfectly fine in my testing. Nice work!

PR #2249, opened 2021-05-17
Initialize bindings queue lazily

fpabl0:
### Description:

Initialize bindings queue lazily. In this way, projects that do not use binding, don't need to deal with a 1024 queue length.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/queue.go
+++ b/data/binding/queue.go
@@ -5,9 +5,10 @@ import (
 )
 
 var (
-	itemQueueIn  chan<- *itemData
-	itemQueueOut <-chan *itemData
-	processOnce  sync.Once
+	itemQueueIn      chan<- *itemData
+	itemQueueOut     <-chan *itemData
+	processOnce      sync.Once
+	queueInitialized = make(chan int)
 )
 
 type itemData struct {
@@ -20,8 +21,10 @@ func queueItem(f func()) {
 		processOnce.Do(func() {
 			itemQueueIn, itemQueueOut = makeInfiniteQueue()
 			go processItems()
+			close(queueInitialized)
 		})
 	}
+	<-queueInitialized
 	itemQueueIn <- &itemData{fn: f}
 }
 
@@ -72,6 +75,7 @@ func processItems() {
 }
 
 func waitForItems() {
+	<-queueInitialized
 	done := make(chan interface{})
 	itemQueueIn <- &itemData{done: done}
 	<-done
--- a/data/binding/queue.go
+++ b/data/binding/queue.go
@@ -1,6 +1,14 @@
 package binding
 
-var itemQueueIn, itemQueueOut = makeInfiniteQueue()
+import (
+	"sync"
+)
+
+var (
+	itemQueueIn  chan<- *itemData
+	itemQueueOut <-chan *itemData
+	processOnce  sync.Once
+)
 
 type itemData struct {
 	fn   func()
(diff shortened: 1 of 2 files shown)


Jacalz (commented):
Really nice stuff. This makes a lot of sense for use cases where, like you say, bindings aren't used 👍

fpabl0:
Maybe I need to improve the test I added, it seems that is not working properly sometimes 😞

fpabl0:
Fixed :)

andydotxyz:
I don’t really get it - if the package isn’t imported then the variable won’t exist and so won’t take up memory...
As I understood it anyhow go compiler includes works at the package level not module level.

fpabl0:
> if the package isn’t imported then the variable won’t exist and so won’t take up memory...

Binding package is imported in widget package so I think it will take up memory...

andydotxyz:
Ah, I forgot that the interfaces were in that package as well, sorry.

fpabl0:
No worries.

data/binding/queue.go
```diff
@@ -1,20 +1,33 @@
...
-var itemQueueIn, itemQueueOut = makeInfiniteQueue()
+import (
+	"sync"
+)
+
+var (
+	itemQueueIn      chan<- *itemData
+	itemQueueOut     <-chan *itemData
+	processOnce      sync.Once
+	queueInitialized = make(chan int)
```
Jacalz, line 11: I think it would be better to use a signalling channel (the int will use up memory) https://github.com/fyne-io/fyne/blob/a25b08480d742f016ee24c1b37aa02c3bb433bd5/internal/driver/glfw/loop.go#L19 or perhaps even just a regular bool and an if condition. I think the latter would be cleaner than trying to read from a closed channel.

Jacalz (changes requested):

data/binding/queue.go
fpabl0: > I think it would be better to use a signalling channel (the int will use up memory)

Yes, you are right, I will update :)

> or perhaps even just a regular bool and an if condition. I think the latter would be cleaner than trying to read from a closed channel.

Hmm the idea of the channel is wait for queue initialization, how could that behavior be accomplished by using just a boolean variable? 🤔

Jacalz: My idea was that you would be able to send the first request within the init function, but I realized that it probably doesn't work.

fpabl0: > the first request within the init function

Not sure what you mean here 🤔

Jacalz: I meant that the queue only would be created once, so the channel is basically a one time use. Every other time, the channel will be closed. Thus you could make the first call to the `queueItem` run ` itemQueueIn <- &itemData{fn: f}` right after it has set up the queue and rely on a book and if outside. That was my idea anyway.

fpabl0: I don't understand sorry. `queueInitialized` is there just for ensure that if` queueItem` is called in different go routines at the same time, all of them will wait until the first one that takes the `sync.Once` initialize the queue, so others don't block trying to send over a nil channel.

If you mean calling queueItem in the init package function, then I don't think that is desirable as it is going to initialize the queue at the very beginning just like before, am I missing something?

Maybe you could show me in code what you mean, I don't understand you very well 😞

andydotxyz (commented):
It is a little complex, but I see that it works.
Is it worth it to save the (I think) 16k of RAM?

Jacalz:
It's not only about the memory. The benefit is also in the time needed to set up the queue. I personally think it's a good idea.

andydotxyz:
Yes I think you're right. On reflection 16K unused is probably a big enough deal anyway.

andydotxyz (approved):

Jacalz (approved):

PR #2251, opened 2021-05-18
FileDialog: Added list view

### LINKED ISSUE #2165: NewFolderOpen: incomplete filenames
Sorry, if this is a double. I still do not know how to search efiiciently in open issues :-(
Also, it is not a bug. But behaviour could be improved.

Tested with Fyne develop branch (12.04.21) on Linux (screen is big enough)

Use dialog.NewFolderOpen resp. dialog.ShowFolderOpen. 
If filenames are a bit longer, you cannot distinguish the files (directory icons and beginning of filenames are identical). 
See screenshot

No. Increasing the dialog window does not help.

Proposal: 
Add another option to show "filenames only" instead of "icon plus filename". If this option is used, a  list shows the filenames under each other (with a scrollbar if there are many). This gives more space for the filenames (almost the width of the dialog).

The user can switch the view via the options if the filenames are too long for the current view.

Screenshot_incomplete_filenames

s77rt:
### Description:

FileDialog: 3 Changes:
- Added list view (with a toggle button to switch between views (grid or list)
- Items are now hoverable
- Adjusted filenames texts positions

Fixes #2165 (partially)

### Checklist:

- [X] Tests included.
- [X] Tests all pass.


dialog/file.go
```diff
@@ -365,6 +382,30 @@ func (f *fileDialog) setSelected(file *fileDialogItem) {
...
+	f.filesWrapper.Objects = []fyne.CanvasObject{f.files}
+	f.filesWrapper.Refresh()
+}
+
+func (f *fileDialog) toggleView() {
+	if f.view == gridView {
+		f.setView(listView)
+	} else {
+		f.setView(gridView)
+	}
```
Jacalz, line 406: Isn't this logic reversed?

dialog/fileitem.go
```diff
@@ -88,22 +105,41 @@ type fileItemRenderer struct {
...
 
 func (s fileItemRenderer) MinSize() fyne.Size {
-	return fyne.NewSize(fileIconSize, fileIconSize+fileTextSize+theme.Padding())
+	var minsize fyne.Size
+	if s.item.picker.view == gridView {
+		minsize = fyne.NewSize(fileIconSize, fileIconSize+fileTextSize+theme.Padding())
+	} else {
+		minsize = fyne.NewSize(fileInlineIconSize+s.text.MinSize().Width, fileTextSize+theme.Padding()*4)
+	}
+	return minsize
```
Jacalz, line 133:
```suggestion
	if s.item.picker.view == gridView {
		return fyne.NewSize(fileIconSize, fileIconSize+fileTextSize+theme.Padding())
	}

	return fyne.NewSize(fileInlineIconSize+s.text.MinSize().Width, fileTextSize+theme.Padding()*4)
```
I think this is a preferred and cleaner way to do it.

Jacalz (changes requested):
Thanks for this. I had some initial requests to improve the code in some places.

dialog/file.go
s77rt: No, it's meant to toggle (switch) 
If we have A then B, if we have B then A.
If it's meant to be the other way around we won't need the if statement in the first place

Jacalz: That seems a little bit contie

dialog/fileitem.go
s77rt: I wanted to do it that way, just it felt like we are making a slight separation by excluding the else statement.
So what do you?
Should i just make the change anyway as it's not really a problem?

Jacalz: It is much cleaner the way I suggested. It is generally the preferred way to do it if looking through other parts of code here.

andydotxyz:
It can be helpful for graphical changes like this to include an image so that you can get feedback from people not currently able to compile your branch.

s77rt:
https://user-images.githubusercontent.com/16493223/118841003-ccd97600-b8bf-11eb-956a-c502fb92bf12.mp4

Jacalz:
Thinking about two ways that this might be improved regarding the implementation of this.

a) I think it would be much nicer with specific icons, looking like it usually does in a file manager (a grid icon when using list and a list icon when using grid). I would prefer it with no text, but it probably doesn't look to good where it is currently.

b) Should the view be inside the options instead?

In short, think that the button is taking up a bit too much space where it is located. The whole layout of the FileDialog would need a big reworks but that is probably outside the scope of this PR.

s77rt:
@Jacalz 
a) that what i had in mind, but i didn't find the desired icons so i had to use a text
b) for the ux i think it's better to have this button as a "dedicated" button, but where this button should be located at?

Jacalz:
> b) for the ux i think it's better to have this button as a "dedicated" button, but where this button should be located at?

I just don't like that it takes up such a huge amount of space. Almost nothing is left for the favourites, basically meaning that I have to scroll each time I want to use it. I still think putting the button inside the menu that pops up when pressing the Options menu, is a good idea.

s77rt:
I meant to have the dedication button in case the button is just an icon, and it should be placed in a different position than it's now, maybe next to the current path or another place that fits better

s77rt:
slight modification


https://user-images.githubusercontent.com/16493223/119202814-88013b00-ba89-11eb-90ca-da43ce026f27.mp4

Jacalz:
Hmm, it does look a lot better. I think it currently is filled though and I would prefer to see the save entry and button be at the bottom. With the options up at the top.

Jacalz:
If you search for "Grid" and "List" you can find sensible icons for the button. I think that woulda me it even better. https://materialdesignicons.com/

andydotxyz:
This is a great feature but please don’t break the standard dialog flow. Action buttons belong at the bottom. The options should not take up too much space, especially in the small sidebar. Maybe a popup menu would help, or it may be time to add a toolbar?

s77rt:
@Jacalz ok thanks i will check. 
@andydotxyz the layout is taken from my linux file manager where the buttons are at the top, i will make the changes to make them at the bottom again, but not sure where the options buttons should be at, the sidebar is best to be reserved for favorites only.

Jacalz:
I have never seen a file dialog with the entry and save button at the top. 

From my point of view, keep the bottom section as it was before this change (cancel button left to the right of the entry). Then just move the options button and toggle button to the top bar (basically moving what is at the bottom in your latest video to the top). That would look good, I think.

Jacalz:
I realise that what I wrote might go a bit against what @andydotxyz  said, but let's at least try and see how it looks before dismissing the idea.

Jacalz:
Something like this is what I had in mind.
image

Jacalz:
Sorry. Sent the wrong screenshot. This is what I meant that I had in mind.
image

s77rt:
Looks good, but what about the options button?
I feel that we need to make a change to it too?
(remove it?, keep only the icon? reposition it?)

Jacalz:
Keeping the icon only, perhaps. I also think you also would have to deselect the list as soon as you go into a different folder, as a side-note.
list-deselect

s77rt:
This has been already implemented;
a favorite will be selected automatically if you enter to that location (directly or indirectly) and it will be also unselected automatically if you are no longer in that location. 
I will push the changes once we agree on a design.

For the buttons: here are my thoughts 
The change layout button should have a low importance (it may be better to not have a border on it) (can you upload how it looks,. I'm using a phone rn) 
The options button icon should be changed to the tree dots icon since it only invokes a popup

Jacalz:
You could probably use a toolbar for that. I don't have any code on my end, it's just copy-paste mock-ups with screenshots from your videos. Nothing fancy.

s77rt:
Isnt a toobar too much for that

andydotxyz:
> @andydotxyz the layout is taken from my linux file manager where the buttons are at the top,

I guess that is Gnome. They have done some very questionable things :(

s77rt:
what do you think?
fyne_dialog_list_3

Jacalz:
It is starting to look really good. I have a few points though:
- It looks off with the buttons having different importance. Don't know which one would look the best.
- Also don't know if I'm too keen on the three dots. The settings icon might still be a better choice in my opinion. I would also prefer to have it to the right of the layout toggle button.
- It looks wrong to have the drops down show outside of the dialog. It should probably drop down below the button and still be inside the icon.

s77rt:
fyne_dialog_list_4

s77rt:
slight change to fit 4 items per line in default view
fyne_dialog

Jacalz:
Nice work. I think it is looking really nice now. I think it perhaps could do with a separator between the path display (that shows the current path) and the items being displayed though.

s77rt:
I didn't understand what you are referring to, can you please elaborate?

Jacalz:
I think it might benefit from having a separator here (at the red line).
CE82B642-0058-4000-9F44-5364D268E3EE

s77rt:
ah i think it's better not to have that currently, because when you get deeper in folders the breadcrumb will have more buttons and a scrollbar will appear, in which case the separator line won't have any point and it will only make the view ugly by having two lines undereach other (one from the separator and the other from the scroller)

andydotxyz:
> slight change to fit 4 items per line in default view

Thanks for this. Probably worth reflecting on how this looked before the PR started so we can better compare (i.e. 4 wide was already default)

Screenshot 2021-05-24 at 09 29 34

andydotxyz (changes requested):
filesWrapper seems to be redundant - you can just set the `Content` of a `container.Scroll` directly.

s77rt:
@andydotxyz thanks for the note,
There is still more commits i didn't push yet as i didn't write the tests
Some tests rely on the ui, so what do you think about it? any requested changes?

andydotxyz:
I'm not sure when the `widget.List` on the left was added - that does not seem to be on the list of changes.
The toggle button at the top left looks good, having moved options up there is fine too.
I don't think that the breadcrumb buttons should have their borders removed personally - it hides them being buttons and led to the request above of a new separator bar to help with visual divide.

Apologies if I have missed the main questions - it is very difficult to comment on the changes when the screenshots don't match the content of the PR.

s77rt:
Sorry for that, i will push the changes asap

s77rt:
update, (this does not include the tests yet)

Jacalz:
Is this ready for a re-review? @s77rt

s77rt:
**Long:**
No, but only one test is missing, i have fixed the old tests (locally)
i was going to make a new commit and mark this as ready to review but i found an issue during the process which made me think if it's a bug or if i have done something wrong: #2268, now it's seems as a fyne's bug, so i will just make the remaining test and this should be ready for review in a day or two.

**Short:**
Hopefully this will be ready for review within 48H

andydotxyz (changes requested):
This works really well, thanks.
However I think the list view could be tweaked a little to fit with the rest of the layout:
1) The list items are taller than the favourites list (I think this was changed recently, so you probably need to reduce by theme.Padding()*2)
2) The icon looks very squashed to the left (divider) - perhaps shifting the layout by theme.Padding() would look more even?

Otherwise great work :)

andydotxyz:
Screenshot 2021-06-03 at 09 45 41

(For comparison - notice that the rows don't take the same space and that the icons feel a bit crushed left.

Jacalz (changes requested):
I agree with @andydotxyz on the visual changes. Otherwise, I think this is looking really solid. The file dialog looks much, much better now :)

s77rt:
> * The list items are taller than the favourites list (I think this was changed recently, so you probably need to reduce by theme.Padding()*2)

they don't have to have the same height, and i can only get them to have the same height by using a fraction in the theme.Padding() (=3.5)

s77rt:
converting this to draft again, i think it would be nice to add a compactview option (worknig on it now...)

Jacalz:
> converting this to draft again, i think it would be nice to add a compactview option (worknig on it now...)

What do you medan with "compactview"? It might be good to take this one step at the time to avoid having to review huge amounts of code.

s77rt:
It's just a few more lines, not an overhead to include in the same pr.
A compact view is meant to not keep much empty space comparing to a normal view.
A compact view differ than a normal view by just the padding value.
I will upload a gif in few minutes

Jacalz:
I still think we might want to be a bit conservative with adding too many new layouts. I think list and grid should be good enough for now.

s77rt:
it's not a new layout, it's a layout parameter

s77rt:
i see that the name compactView looks like it belongs to gridView, listView which lead to the confusion
maybe we should find a better name for the field compactView
i will be pushing the code soon

Jacalz:
The point that I am trying to make is that too many layouts, layout parameters or  whatever is going to clutter up the user interface. If it does not provide enough of meaningful change, I think we should consider if it is actually something that we want/need.

s77rt:
fyne_dialog_compactview

s77rt:
**TODO:**
Add Compact View Test(s)

s77rt:
> The point that I am trying to make is that too many layouts, layout parameters or whatever is going to clutter up the user interface. If it does not provide enough of meaningful change, I think we should consider if it is actually something that we want/need.

Ah i got it, i have pushed the new code and a gif, what do you think?

andydotxyz:
> they don't have to have the same height, and i can only get them to have the same height by using a fraction in the theme.Padding() (=3.5)

Perhaps this is more due to the list view usage than the layout in the main view? Was moving the favourite buttons from VBox to List important for this change? Rolling it back may make them line up again. It’s not essential as you say, but maybe we consider the pros/cons?

Jacalz:
> Ah i got it, i have pushed the new code and a gif, what do you think?

Hmm, I'm a bit torn. Don't know if it's quite worth it. Wdyt @andydotxyz?

> Was moving the favourite buttons from VBox to List important for this change? Rolling it back may make them line up again. It’s not essential as you say, but maybe we consider the pros/cons?

I personally think the list makes it look a lot better. It also makes it easier to navigate with more favourites in the list, allowing us to add more (Picture, Videos, Desktop) in the future and potentially allowing to add custom ones by users (like we have discussed before).

s77rt:
the reason that we should use a list is because list items are selectable while buttons are not.

s77rt:
for the compactview option, i have tested it on using a bigger file dialog, and it's quite useful for navigating files faster
in the gif above it, compactview only makes one more column to fit (gridview), and one more line (listview), but on bigger file dialogs, more can be fitted and more files can be navigated faster...
this is not a really fancy feature, but since it's there for no overhead cost, i think it's okay to keep it
@andydotxyz any thoughts on this?

andydotxyz:
> > Ah i got it, i have pushed the new code and a gif, what do you think?
> 
> Hmm, I'm a bit torn. Don't know if it's quite worth it. Wdyt @andydotxyz?

The theme sets standard padding in all UI. This means that things line up and are consistent across the app.
I don't think adding this feature to alter the padding for a single UI area is a good idea.
If we could keep this PR focused on adding list view we can get it closed sooner, new features should be discussed in separate issues/PRs.

s77rt:
alright, i think now the pr is ready to be merged
excluding the point that the files items and the favorites items must have the same height

dialog/file.go
```diff
@@ -164,75 +182,108 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
...
+			toggleViewButton.SetIcon(theme.ListIcon())
+		}
+	})
+
+	optionsbuttons := container.NewHBox(
+		toggleViewButton,
+		optionsButton,
+	)
+
+	header := fyne.NewContainerWithLayout(layout.NewBorderLayout(nil, nil, nil, optionsbuttons),
```
Jacalz, line 237: This should be using `container.NewBorder()`, the above function is deprecated.

dialog/file.go
```diff
@@ -164,75 +182,108 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
...
+		optionsbuttons, widget.NewLabelWithStyle(title, fyne.TextAlignLeading, fyne.TextStyle{Bold: true}))
+
+	footer := fyne.NewContainerWithLayout(layout.NewBorderLayout(nil, nil, nil, buttons),
```
Jacalz, line 240: Same as above.

dialog/file.go
```diff
@@ -365,6 +435,23 @@ func (f *fileDialog) setSelected(file *fileDialogItem) {
 	}
 }
 
+func (f *fileDialog) setView(view viewLayout) {
+	f.view = view
+	if f.view == gridView {
+		padding := fyne.NewSize(fileIconCellWidth-fileIconSize, theme.Padding())
+		f.files = fyne.NewContainerWithLayout(layout.NewGridWrapLayout(
```
Jacalz, line 442: This should use `container.NewGridWrap()`, the function used above is deprecated.

dialog/file.go
```diff
@@ -365,6 +435,23 @@ func (f *fileDialog) setSelected(file *fileDialogItem) {
...
+			fyne.NewSize(fileIconSize, fileIconSize+fileTextSize).Add(padding),
+		))
+	} else {
+		f.files = fyne.NewContainerWithLayout(layout.NewVBoxLayout())
```
Jacalz, line 446: This should use `container.NewVBox()`.

dialog/file.go
```diff
@@ -164,75 +182,108 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
...
 
 	p := position.Add(buttonSize)
-	pos := fyne.NewPos(p.X, p.Y-content.MinSize().Height-theme.Padding()*2)
+	pos := fyne.NewPos(p.X-content.MinSize().Width-theme.Padding()*2, p.Y+theme.Padding()*2)
 	widget.ShowPopUpAtPosition(content, f.win.Canvas, pos)
 }
 
-func (f *fileDialog) loadFavorites() []fyne.CanvasObject {
+func (f *fileDialog) loadFavorites() {
+	f.favorites = []favoriteItem{}
```
Jacalz, line 265: I think it could be good to move this down to where the `places` variable was used before as it is more clear to how it is being used. As far as I can see, there is no reason that it is set this far up in the method.

Jacalz (changes requested):
This is looking very good. I had a few requested changes for code style.

dialog/file.go
s77rt: I think it's better to use the new style
the methods are clearer, loadFavorites is to load the favorites field with the favorites, which can be used later in the favoriteslist
and getPlaces to get the places and return them as a variable (thus the use of get and not load in the method name)
if you are referring to something else, let me know, maybe i didn't get your point exactly

Jacalz: No, I just mean that you should move this variable down to just before the for-loop where items are added to it.

dialog/file.go
s77rt: parts of the code was already using NewContainerWithLayout and a copy paste led to this
it's hard to keep track of deprecated functions, arent deprecated functions supposed to print a warning?

dialog/file.go
s77rt: ah okay, but why it's better to move this down?
the name of method is loadFavorites and it's normal to see the concerned variable get set to an empty slice at first

s77rt: alright, i see your point

dialog/file.go
Jacalz: I realize that it's hard to keep track of it. With time, you remember which ones are deprecated and which are not. In your own fyne-projects, running staticcheck will print an error about it. 

However, we have turned off the check because manually turning off the check for each use case where we need to use our deprecated functions until a suitable replacement is ready.

I will go through all deprecations manually some time in the future to clean up our deprecation usages and hopefully we can enable the check then. Right now there are too many small uses left over that need to go away. Thankfully code-review can catch new issues and correct old ones in the meantime.

s77rt: okay, thanks for the clarification

andydotxyz (approved):
Really great work, thanks :)

Jacalz (approved):
This is looking great. Nice work :+1:

PR #2261, opened 2021-05-24
Feature/disabled checked menu

### LINKED ISSUE #1753: MenuItem support Selection Indication

andydotxyz:
### Description:

Continuation of #1975 from @tehsphinx

This PR adds enabling/disabling of menu items as well as adding the ability to add a check before a menu item.

Fixes #1753
Addresses Enable/Disable part of #682

### Checklist:

- [ ] Tests included. (objectiveC tests missing)
- [x] Tests all pass.


internal/driver/glfw/menu_darwin.go
```diff
@@ -35,6 +35,8 @@ const char* test_NSMenuItem_title(const void*);
 import "C"
 
 var callbacks []func()
+var menuEnableCallbacks []func() bool
+var menuCheckedCallbacks []func() bool
```
Jacalz, line 39: Hmm, I don't know if it's just me but there is something with function pointer slices that feels wrong. I wonder if there is a more "sane" approach?

menu.go
```diff
@@ -16,10 +16,26 @@ func NewMenu(label string, items ...*MenuItem) *Menu {
...
+func (i *MenuItem) SetItem(menuItem CanvasObject) {
+	i.item = menuItem
+}
+
+// Refresh refreshes the menu item.
+func (i *MenuItem) Refresh() {
+	if i.item == nil {
+		return
+	}
+	i.item.Refresh()
```
Jacalz, line 38: Not a big deal, but wouldn't this be cleaner?
```suggestion
	if i.item != nil {
		i.item.Refresh()
	}
```

widget/menu.go
```diff
@@ -167,9 +170,33 @@ func (m *Menu) Move(pos fyne.Position) {
...
+func (m *Menu) getContainsCheck() bool {
+	for _, item := range m.Items {
+		mi, ok := item.(*menuItem)
+		if !ok {
+			continue
+		}
+		if mi.Item.HasCheckmark {
+			return true
+		}
+	}
```
Jacalz, line 196: I think this might be cleaner. Haven't tested it, but I think it should work.
```suggestion
	for _, item := range m.Items {
		if mi, ok := item.(*menuItem); ok && mi.Item.HasCheckmark {
			return true
		}
	}
```

Jacalz (commented):
Just a few comments on the implementation. Thanks for continuing this.

internal/driver/glfw/menu_darwin.go
andydotxyz: There was a discussion on the original PR I think, but a better approach was not found.
I could convert it to a map of id->func() if you like, not sure if that adds anything though,

menu.go
andydotxyz: It is one line less code, but is counter-indicated from the way we exit early and keep main code least-indented.

Jacalz: Sure. It's no biggie.

internal/driver/glfw/menu_darwin.go
Jacalz: Hmm, I don't think a map would gain us anything. Could be a good idea to see what others think about this and/or if they have any other solutions.

andydotxyz: I would agree but this has been open 24 days now, so not sure who else is available to help.

andydotxyz: We have not had any better ideas, given that this was following the pattern above I think we can progress and come back to it if there are great suggestions later.

Jacalz: I thought @fpabl0 had an idea about it? https://gophers.slack.com/archives/CK5U4BU87/p1623952193313600

andydotxyz: Thanks for the reminder, I think that I have applied the solution now.

Jacalz (commented):
This is looking fine to me, just wondered if perhaps fyne_demo should be updated to show this off?

andydotxyz:
Added now. I also moved `HasCheck` to `Checked` which I realised is more consistent with our other naming.

Jacalz (changes requested):
Sorry that this has taken a while to get to. My last point here is just that it would be a nice demo to have the checked item toggle when pressed.

andydotxyz:
I don’t get what you mean, this PR is not for a toggle checked state - just application controlled check state? I could add another menu item that is called toggle that does this, but it won’t be demonstrating a feature.

Jacalz:
Hmm. I Sorry for asking, but what is the use case for a checked menu that is not toggled by the user? I just can't remember last time I saw menu items that can be checked, so I'm just unsure of how they are used in applications.

andydotxyz:
My point was mostly that it could be doing any number of things when an item is tapped. An item that represents on/off would likely toggle - but it could also represent one-of-many, and so tapping current would likely do nothing.
Either way it’s in application code, so I’m not sure if that code should also be in the demo?

Jacalz (approved):

PR #2263, opened 2021-05-25
Feature/buttonkeys

andydotxyz:
### Description:
Add focus state to `Button` and `Hyperlink` and enable space-bar to action them.
This added a new focus rect around hyperlink as is seen in web browsers:

focus
initial

Relates to #1515

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


fpabl0:
Could buttons and hyperlinks be actioned by tapping `Enter` too?

andydotxyz:
> Could buttons and hyperlinks be actioned by tapping `Enter` too?

So far we have just used space for activation.
The idea of Enter was that it would be used for the primary action in forms / dialogs - so if we wire it into default buttons then we will get inconsistencies as we develop it into more complex scenarios.

widget/entry.go
```diff
@@ -418,6 +419,7 @@ func (e *Entry) MouseDown(m *desktop.MouseEvent) {
 	}
 	e.propertyLock.Unlock()
 
+	fyne.CurrentApp().Driver().CanvasForObject(e).Focus(e)
```
fpabl0, line 422: I think here we should use .super() for extended widgets

widget/entry.go
```diff
@@ -488,6 +490,9 @@ func (e *Entry) SetText(text string) {
 //
 // Implements: fyne.Tappable
 func (e *Entry) Tapped(ev *fyne.PointEvent) {
+	if !e.Disabled() {
+		fyne.CurrentApp().Driver().CanvasForObject(e).Focus(e)
```
fpabl0, line 494: same here

widget/entry.go
```diff
@@ -504,6 +509,7 @@ func (e *Entry) TappedSecondary(pe *fyne.PointEvent) {
 		return // no popup options for a disabled concealed field
 	}
 
+	fyne.CurrentApp().Driver().CanvasForObject(e).Focus(e)
```
fpabl0, line 512: same here

widget/entry.go
```diff
@@ -536,6 +542,33 @@ func (e *Entry) TappedSecondary(pe *fyne.PointEvent) {
 	e.popUp.ShowAtPosition(popUpPos)
 }
 
+// TouchDown is called when this entry gets a touch down event on mobile device, we ensure we have focus.
+//
+// Since: 2.1
+//
+// Implements: mobile.Touchable
+func (e *Entry) TouchDown(*mobile.TouchEvent) {
+	if c := fyne.CurrentApp().Driver().CanvasForObject(e); c != nil {
```
fpabl0, line 551: Although here has been used `super()` in c.Focus. I wonder if we need to use `super()` in `CanvasForObject()` too instead of just the entry. In my understanding when the entry widget is extending, the entry receiver as it does not exist in the canvas (only the new extended widget would exist), am I missing something? 🤔

CHANGELOG.md
```diff
@@ -3,6 +3,17 @@
 This file lists the main changes with each version of the Fyne toolkit.
 More detailed release notes can be found on the [releases page](https://github.com/fyne-io/fyne/releases). 
 
+## 2.1 - Ongoing
+
+### Added
+
+### Updated
+
+* Focusable widgets are no longer focused on tap, add canvas.Focus(obj) in Tapped handler if required
```
fpabl0, line 12: This would also mean that custom widgets that are focused when tapped cannot be extended? (Because .super() is private? 🤔)

fpabl0 (changes requested):

fpabl0:
> so if we wire it into default buttons then we will get inconsistencies as we develop it into more complex scenarios.

I don't understand how Enter-key action would lead to inconsistencies as it would only work when the button is focused. 🤔

andydotxyz:
> I don't understand how Enter-key action would lead to inconsistencies as it would only work when the button is focused. 🤔

Because depending on what widget is focused the Enter button would change from actioning an item to submitting the form.
Space has no such idea so always triggers what is focused.

CHANGELOG.md
andydotxyz: Yes, thanks - I missed some elements here and have tidied it all up

widget/entry.go
andydotxyz: Yes you're right, fixed up.

internal/driver/glfw/menu_bar_test.go
```diff
@@ -54,7 +54,7 @@ func TestMenuBar(t *testing.T) {
 
 		menuBar := glfw.NewMenuBar(menu, c)
 		themeCounter := 0
-		button := widget.NewButton("Button", func() {
+		button := newNotFocusableButton("Button", func() {
```
fpabl0, line 57: I think now this is unneeded and we can use widget.NewButton again IMHO.

internal/driver/glfw/menu_bar_test.go
```diff
@@ -327,7 +327,7 @@ func TestMenuBar(t *testing.T) {
...
```
fpabl0, line 330: same here

internal/driver/glfw/menu_bar_test.go
```diff
@@ -526,3 +526,18 @@ func TestMenuBar_Toggle(t *testing.T) {
...
+
+func newNotFocusableButton(l string, f func()) *notFocusableButton {
+	n := &notFocusableButton{f: f}
+	n.Label.Text = l
+	return n
+}
+
+func (n *notFocusableButton) Tapped(e *fyne.PointEvent) {
+	n.f()
+}
```
fpabl0, line 543: This shouldn't be needed now 🤔

widget/button.go
```diff
@@ -308,6 +333,8 @@ func (r *buttonRenderer) buttonColor() color.Color {
 	switch {
 	case r.button.Disabled():
 		return theme.DisabledButtonColor()
+	case r.button.focused:
+		return theme.FocusColor()
```
fpabl0, line 337: I am unsure if this is the best way to indicate the button is focused 🤔

widget/entry.go
```diff
@@ -892,6 +925,12 @@ func (e *Entry) registerShortcut() {
 	})
 }
 
+func (e *Entry) requestFocus() {
+	if c := fyne.CurrentApp().Driver().CanvasForObject(e.super()); c != nil {
+		c.Focus(e.super().(fyne.Focusable))
```
fpabl0, line 930: Shouldn't is it better to get first the super() in a variable to avoid having to deal internal locks inside BaseWidget?
```go
impl := e.super()
...
c.Focus(impl.(fyne.Focusable))
```

internal/driver/glfw/menu_bar_test.go
andydotxyz: Can you expand on this reasoning please? I added it because it was needed, so a little more help as to why you think it is not needed would be appreciated.

internal/driver/glfw/menu_bar_test.go
andydotxyz: As above, can you help me understand why?

internal/driver/glfw/menu_bar_test.go
fpabl0: > Can you expand on this reasoning please? I added it because it was needed, so a little more help as to why you think it is not needed would be appreciated.

I replaced it locally, and it just look like before, nothing seems to be broken 🤔. I think you tried that solution because previously widgets that are tappable and focusable, were focused when tapped, but now it isn't the case, right?

andydotxyz: No, it breaks the keyboard focus cycle - the menu and button are on the same focus manager in this test and tapping "left, left" should wrap to the right menu, then the middle one, but instead it now hits the button first then the right hand menu. So I made it not-a-focusable-button instead.

fpabl0: Oh I see, you are right, sorry

widget/button.go
andydotxyz: Well, it matches how we have done it for menus - what alternative would be more suitable?

fpabl0: I think material design specs are suitable for this case:
image

However as it was discussed before, those state colors are confusing among all the material design components. But definitely I don't think buttons should change their background color when they are focused (an overlay color is ok but a full color replacement does not seem great).

Another possibility is to have a focused border in buttons (like the hyperlinks), just like github buttons:

https://user-images.githubusercontent.com/12239342/119899651-0128e400-bf09-11eb-81ed-16188176b50a.mp4

fpabl0: > Maybe the addition of SelectionColor and FocusColor as in Material Design could be enough?

This should be right way to go IMHO.

andydotxyz: > Maybe the addition of SelectionColor and FocusColor as in Material Design could be enough?
> But definitely I don't think buttons should change their background color when they are focused

This usage is consistent with Menu, Check, Radio and others. I am using `FocusColor` to indicate focus - I don't see how adding `SelectionColor` helps. We went with a not-grey colour because keyboard is an important part of accessibility and a really subtle indicator is not appropriate.

At this point I think that a request to change how Focus colour is handled should be a separate issue - I don't want to have to change it across all for this ticket and I also don't want to introduce inconsistencies with this PR.

fpabl0: > This usage is consistent with Menu, Check, Radio and others.

Menu, check and radio don't have a configurable specific background through the theme iirc. But buttons do, so the focus color is going to change the whole appearance of a button with no transparent background (would look like a completely different one). This problem would be bigger if primary, secondary, success, danger, etc. colored variants for buttons are added.

> We went with a not-grey colour because keyboard is an important part of accessibility and a really subtle indicator is not appropriate.

That's fine, but it doesn't feel good when you have a widget with a custom background, IMO. Maybe a focused border color should be used for widgets that have background colors different from canvas color? 🤔

> At this point I think that a request to change how Focus colour is handled should be a separate issue - I don't want to have to change it across all for this ticket and I also don't want to introduce inconsistencies with this PR.

Oh sorry, that was not my intention (proposing the Focus color change in this PR), I was just exposing two possible solutions to avoid having to change the whole button background color when it is focused. The first one (changing the FocusColor) I agree should be done in a different PR (if it is decided) and the second one is having a focused border just like the hyperlink (that could be done in this PR). Personally, I think changing completely the background color of the widgets to represent a focus state does not feel natural. 

Maybe this is a subjective area and more opinions are needed?

andydotxyz: Ah yes, it should be blended - like the hover colour - sorry I missed that.
Should be fixed now.

widget/button.go
```diff
@@ -309,25 +334,15 @@ func (r *buttonRenderer) buttonColor() color.Color {
 	switch {
 	case r.button.Disabled():
 		return theme.DisabledButtonColor()
+	case r.button.focused:
+		return blendColor(theme.ButtonColor(), theme.FocusColor())
```
fpabl0, line 338: This should be the corresponding bg color depending on the importance, however I agree that the color resulted after blending with the primary color does not have any appreciable difference, maybe this could be improved if we decide in the future to change the Focus color (to something similar to MaterialDesign), this should be needed if we want different colored buttons (primary, secondary, success, danger, etc). However, for now I think it is ok :)

fpabl0 (approved):
LGTM, just some comments.

widget/button.go
andydotxyz: Yes I wondered that as well, but it just didn't seem to look good. Maybe we can revisit it somehow, like if the focus was not a directly less-vibrant version of primary.

PR #2264, opened 2021-05-26
Fix/consistentbasewidget

andydotxyz:
### Description:
Move our internal Base widget (that could not be extended) to be consistent with the public BaseWidget.
Now we have just 1 API and the ability to extend both.
It is still 2 different classes due to the cycle issues. One cannot be an alias of the other due to certain unexported APIs.
This will be further tidied before 2.1 by hopefully replacing the `ExtendBaseWidget` idea - but for now it should unblock some keyboard operation items.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/menu_bar.go
+++ b/internal/driver/glfw/menu_bar.go
@@ -25,8 +25,11 @@ type MenuBar struct {
 func NewMenuBar(mainMenu *fyne.MainMenu, canvas fyne.Canvas) *MenuBar {
 	items := make([]fyne.CanvasObject, len(mainMenu.Items))
 	b := &MenuBar{Items: items, canvas: canvas}
+	b.ExtendBaseWidget(b)
 	for i, menu := range mainMenu.Items {
-		items[i] = &menuBarItem{Menu: menu, Parent: b}
+		barItem := &menuBarItem{Menu: menu, Parent: b}
+		barItem.ExtendBaseWidget(barItem)
+		items[i] = barItem
 	}
 	return b
 }
@@ -38,6 +41,7 @@ func (b *MenuBar) CreateRenderer() fyne.WidgetRenderer {
 	cont := fyne.NewContainerWithLayout(layout.NewHBoxLayout(), b.Items...)
 	background := canvas.NewRectangle(theme.ButtonColor())
 	underlay := &menuBarUnderlay{action: b.deactivate}
+	underlay.ExtendBaseWidget(underlay)
 	objects := []fyne.CanvasObject{underlay, background, cont}
 	for _, item := range b.Items {
 		objects = append(objects, item.(*menuBarItem).Child())
@@ -55,49 +59,6 @@ func (b *MenuBar) CreateRenderer() fyne.WidgetRenderer {
 // An active menu bar shows the current selected menu and should have the focus.
 func (b *MenuBar) IsActive() bool {
 	return b.active
-}
-
-// Hide hides the menu bar.
-//
-// Implements: fyne.Widget
-func (b *MenuBar) Hide() {
-	widget.HideWidget(&b.Base, b)
-}
-
-// MinSize returns the minimal size of the menu bar.
-//
-// Implements: fyne.Widget
-func (b *MenuBar) MinSize() fyne.Size {
-	return widget.MinSizeOf(b)
-}
-
-// Move sets the position of the widget relative to its parent.
-//
-// Implements: fyne.Widget
-func (b *MenuBar) Move(pos fyne.Position) {
-	widget.MoveWidget(&b.Base, b, pos)
-}
-
-// Refresh triggers a redraw of the menu bar.
-//
-// Implements: fyne.Widget
-func (b *MenuBar) Refresh() {
-	widget.RefreshWidget(b)
-}
-
-// Resize resizes the menu bar.
-// It only affects the width because menu bars are always displayed with their minimal height.
-//
-// Implements: fyne.Widget
-func (b *MenuBar) Resize(size fyne.Size) {
-	widget.ResizeWidget(&b.Base, b, size)
-}
-
-// Show makes the menu bar visible.
-//
-// Implements: fyne.Widget
-func (b *MenuBar) Show() {
-	widget.ShowWidget(&b.Base, b)
 }
 
 // Toggle changes the activation state of the menu bar.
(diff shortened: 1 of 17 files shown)


internal/widget/base.go
```diff
@@ -8,115 +8,153 @@ import (
...
 
-// Base is a simple base widget for internal use.
+// Base provides a helper that handles basic widget behaviours.
 type Base struct {
-	hidden bool
-	pos    fyne.Position
-	size   fyne.Size
+	size     fyne.Size
+	position fyne.Position
+	hidden   bool
```
toaster, line 15: FMI: why did you remove the alphabetical order?

internal/widget/scroller.go
```diff
@@ -538,16 +442,13 @@ func (s *Scroll) Refresh() {
 	s.refreshWithoutOffsetUpdate()
 }
 
-func (s *Scroll) refreshWithoutOffsetUpdate() {
-	RefreshWidget(s)
+func (s *Scroll) Resize(sz fyne.Size) {
+	s.Base.Resize(sz)
+	s.Refresh()
```
toaster, line 448: Shouldn’t we keep the refreshes limited to actual changes of the size (`size != s.size`)?

internal/driver/glfw/menu_bar.go
```diff
@@ -199,6 +160,7 @@ func (r *menuBarRenderer) Refresh() {
 	r.Layout(r.b.Size())
 	r.background.FillColor = theme.ButtonColor()
 	r.background.Refresh()
+	r.ShadowingRenderer.RefreshShadow()
```
toaster, line 163: FYI: I guess, this is not really related to the base widget change and therefore should have been in a separate commit, right?

widget/tree.go
```diff
@@ -643,6 +643,12 @@ func (n *treeNode) Tapped(*fyne.PointEvent) {
 	n.tree.Select(n.uid)
 }
 
+func (n *treeNode) Resize(s fyne.Size) {
+	n.BaseWidget.Resize(s)
+	// TODO find out if the indentation changed. If not, and size is the same, then we can skip this as normal
```
toaster, line 648: TBH, I think that a caller of `Resize` should never rely on the refresh. If the indentation gets changed at some point and depends on the refresh of a `Resize` call with the same size, this is wrong. Instead the change of the indentation should trigger a refresh.

internal/driver/glfw/menu_bar_item.go
```diff
@@ -113,6 +113,11 @@ func (i *menuBarItem) MouseOut() {
 // Implements: fyne.Tappable
 func (i *menuBarItem) Tapped(*fyne.PointEvent) {
 	i.Parent.toggle(i)
+
+	if i.Parent.active {
+		i.Parent.canvas.Focus(i)
+	}
+	i.Refresh()
```
toaster, line 120: There is no test for this, therefore I don’t know which bug you’re trying to fix with it.
But I think this is partly wrong. The parent’s `toggle` already focuses the item on activation and unfocuses it on deactivation.
So, if the menu is active after the toggle, the item should already be focused and the added focus code does nothing.
If the additional `Refresh` fixes some bug, we should try to add a test case for it and also move it to `MenuBar`’s `toggle()`. Currently, the item gets refreshed before it is being focused. Maybe that’s to early?

toaster (changes requested):
A couple of questions/suggestions as normal comments (not needed for my approval).
And one objection to the menu bar refresh fix.

internal/driver/glfw/menu_bar.go
andydotxyz: I had assumed it was something that I broke in the refactoring. If it was already broken then yes it should be a separate commit.
I don't think it's worth unpicking and force pushing a better version, but I will try to keep an eye out in future commits.
If you disagree say so and I'll rework the history

internal/driver/glfw/menu_bar_item.go
andydotxyz: Apologies it was not a bug fix, I had misread the code in the refactor and you are right it should not have been needed. What has changed is that on tapped the focus in is not triggering so we don't get the same refresh request cycle. I tracked down that `activateChild` does not actually refresh the button - only the bar. The bar's refresh does not cascade to the list of items it would seem. I have put in place what feels like a better fix in `activateChild`, what do you think?

internal/widget/base.go
andydotxyz: A mistake, sorry

internal/widget/scroller.go
andydotxyz: Yes, good catch thanks.

widget/tree.go
andydotxyz: It's not relying on a Refresh - it's the Layout that we care about.
Normally we can ignore Resize if the size has not changed - but in the case of Tree it may be the same size but need to have the content moved when it's reused (because increased indent does not change size).
Maybe there is a better solution, but I cannot think of one right now.

andydotxyz (commented):
Thanks for the notes, I hope these comments and the commit help :)

widget/tree.go
toaster: I still don’t fully understand why a `Resize` of a tree node has to handle indentation change. AFAICS an indentation change is implemented by the node’s `update` method which should then do the appropriate things.
Also, AFAIK, the `BaseWidget`'s `Resize` doesn’t do anything if the size didn’t change (apart from heating the world).

andydotxyz: Quite right, I managed to track down the root cause. The `leaf` and `branch` widgets were having the `treeNode` widget inside accessed directly, which was part of the issue. The rest was that the `partialRefresh` called when depth changed was not updating the node layout.

With this in place it should be fixed much more satisfactorily.

Jacalz (approved):
This looks good and sensible to me, though I would have preferred to get rid of `.ExtendBaseWidget()` entirely ;)

andydotxyz:
> This looks good and sensible to me, though I would have preferred to get rid of `.ExtendBaseWidget()` entirely ;)

Yes, that is the plan. But it will be easier to get rid of from a consistent place :).
Plus I don't yet know how we will do it, but I do know that this was blocking.

toaster:
> This looks good and sensible to me, though I would have preferred to get rid of `.ExtendBaseWidget()` entirely ;)

Yeah, that’s something @andydotxyz and me discussed already. Sadly, it’s not trivial.

toaster (approved):

Issue #149, opened 2019-03-13
Basic HTML rendering from Markdown


boussou:
I need a control that will allow to display basic HTML generated from markdown. So very simplistic HTML: bold,italic,ol,ul,p,br. including IMG tag
I would like to see it happen, and I am willing to pay for it.
Any volonteer?

jadr2ddude:
If the goal is to render markdown, it would probably be easier to render the markdown itself rather than the HTML. HTML has lots of formatting quirks, whereas markdown is more minimalist. The [blackfriday](https://godoc.org/gopkg.in/russross/blackfriday.v2) package exports a syntax tree that could be used to set up some object tree in fyne.

andydotxyz:
I also would like to see this happen - and agree with @jadr2ddude. In fact I had already bookmarked blackfriday to explore once the 1.0 release was out :)

The only problem with ignoring the HTML approach is that embedded markup won’t work correctly - but we could find a way to deal with that in the short term I am sure.

boussou:
Then the next problem is to be able to calculate the widget size after render. You have to understand the fyne internal for that

andydotxyz:
That should not be problematic at all - by implementing the layout it will automatically know the correct widget size?
Either it will cause a full space to render all the components, or it can be put into a ScrollContainer which could be set to a more suitable size if you want to scroll.

Unless I misunderstood your concern about sizing?

boussou:
Ok, that is usually the faced problem in such case. So this lib appear to be well designed ;-)

andydotxyz:
Let's hope so! 
We will see where this comes in the planning for 1.1 next week

andydotxyz:
I think we can see our way to add this to the 1.1 milestone. It will, however, require additional work before it can be implemented:

* Add text wrap support to widget.textProvider
* Text markup or style support in text blocks - so that sub-sections can be styled or linked, etc.

This is quite a bit of work and so I don't imagine it would come in early - I imagine that 1.1 could take 3 months to complete, but it has not been fully planned yet.

I was considering setting up a bounty source / donation system to support Fyne development, if that were in place would you (@boussou ) consider donating with the specific aim of accelerating this feature? Looking at the complexity it is a pretty large item of work...

andydotxyz:
Oh, also - can we rename this "Markdown preview widget"? I'd like to steer clear of thinking about HTML rendering for a long time ;)

boussou:
[...]

Issue #21, opened 2018-10-05
Add full rich text to text entry widget


rasteric:
I know it's a very complex control but I'd still suggest, for the future, a multiline text entry field with the following *rich text* features:

- text styles bold, underline, italic; sticky and non-sticky
- text color
- different fonts in one field
- paragraph alignment
- simple inline images, for many purposes an option to add an image in a line is enough, where the surrounding text is aligned bottom, center, or top
- "standard" import/export format: html, xml, rtf, markdown, or whatever else you think is best, but not something totally non-standard
- custom data ranges with an option to extend them when writing text (like a sticky style) and an option for mouse callbacks/clickbacks

The last one requires explanation. An entry field should allow some way to add custom data to ranges of glyphs and some way of retrieving it by cursor position or mouse position. This is hard to implement later by a user of the library without access to the underlying implementation, because that would require one to maintain some parallel data structure and capture all edit events. If you miss one way the text might be changed, then the data structures get our of sync, so it's much better to have a way to add more functionality directly with the styled text. I'm thinking about hyperlinks, tooltips, etc.

*Background: The reason I add this very general feature request is that I've seen dozens of handmade GUI libraries for various languages, and they all lack in the above functionality. So people turn to Qt or browser-based overblown solutions instead.*

ajwillia-ms:
This is a really great suggestion and we will work toward it slowly.
The canvas text object will only ever support 1 set of styles per block of text (to keep the backends simple) but we can do lots of cool stuff in Go.

The widget.Label has recently added multi line so you can see how this is started - we add (with lots of tests) the layout code and the application of styles (through an API or a markup) which will then apply to label and entry.

I agree that this is something many toolkits lack, but it is very hard to get right and keep efficient. Let's make it happen :)

ajwillia-ms:
Work has also begun on the cursor handling in widget.Entry. Once that is complete we should be able to start work on how rich / marked up text be supported...

stuartmscott:
@andydotxyz I'm not sure https://github.com/fyne-io/fyne/commit/263a14d18b4bd12ab8644a61bd5b9de9840162f7 fixes this issue since it is concerned with select box and not entry/label. Should this issue be reopened?

andydotxyz:
Oops, that was a typo in the commit message I guess

carwyn:
[...]

PR #2269, opened 2021-05-30
Text refactor and RichText addition

andydotxyz:
### Description:
Implementation of the proposal https://github.com/fyne-io/proposals/pull/3 is now under way.

Tasks:
- [x] Refactor away old object types
- [x] Complete implementation of segment parsing and layout
- [x] implement discussed segment types

Remaining issues:
- [x] Deleting does not work across segments
- [x] Alignment of link when it does not have leading/trailing text for positioning
- [ ] Baseline alignment for text of different sizes on the same line (potentially a different PR)

Progresses #21 and #149

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/text.go
```diff
@@ -16,212 +16,365 @@ const (
...
-		return // not yet shown
-	}
-	obj := t.presenter.object()
-	if obj == nil {
-		obj = t
-	}
-
-	obj.Refresh()
+	t.Refresh()
+	cache.Renderer(t).Layout(size)
```
fpabl0, line 226: this will create a phantom renderer that is not drawn in the canvas. this should be
```
cache.Renderer(t.super()).Layout(size)
```

Or maybe this could be removed, because this does not do anything before (because it created a new renderer for textProvider instead of using the one for the presenter).

fpabl0 (commented):
I know this is still a draft, but I just wanted to point this out that I accidentally discovered when working on PR #2101

widget/text.go
andydotxyz: Good catch thanks, fixed

andydotxyz:
Now that it's review-ready here is a screenshot of the RichText in fyne_demo:

Screenshot 2021-06-18 at 19 22 30

charlesdaniels:
I got a crash while resizing the window:

```
$ ./main 
2021/06/18 16:43:20 Lifecycle: Started
2021/06/18 16:43:20 Lifecycle: Entered Foreground
2021/06/18 16:43:22 Lifecycle: Exited Foreground
2021/06/18 16:43:23 Lifecycle: Entered Foreground
2021/06/18 16:43:23 Lifecycle: Exited Foreground
2021/06/18 16:43:24 Lifecycle: Entered Foreground
2021/06/18 16:43:32 Lifecycle: Exited Foreground
2021/06/18 16:43:34 Cancelled
2021/06/18 16:44:03 Lifecycle: Entered Foreground
panic: runtime error: index out of range [25] with length 25

goroutine 38 [running, locked to thread]:
fyne.io/fyne/v2/widget.(*textRenderer).MinSize(0xc000f33cc0, 0xc000f48500)
        /home/cad/f/tmp/fyne/widget/text.go:573 +0x585
fyne.io/fyne/v2/widget.(*BaseWidget).MinSize(0xc000f48500, 0x0)
        /home/cad/f/tmp/fyne/widget/widget.go:89 +0x62
fyne.io/fyne/v2/internal/driver/common.(*Canvas).EnsureMinSize.func1(0xc0010ae180)
        /home/cad/f/tmp/fyne/internal/driver/common/canvas.go:71 +0x172
fyne.io/fyne/v2/internal/driver/common.(*Canvas).walkTree.func2(0xa4a758, 0xc000f48500, 0xa4ab78, 0xc0011a4a50)
        /home/cad/f/tmp/fyne/internal/driver/common/canvas.go:423 +0xe3
fyne.io/fyne/v2/internal/driver.walkObjectTree(0xa4a758, 0xc000f48500, 0x44004000433d0000, 0xa4ab78, 0xc0011a4a50, 0x44004000433d0000, 0x44004000433d0000, 0x436f000043e58000, 0xc00025fcf8, 0xc00025fcd0, ...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:197 +0x31a
fyne.io/fyne/v2/internal/driver.walkObjectTree.func1(...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:176
fyne.io/fyne/v2/internal/driver.walkObjectTree(0xa4ab78, 0xc0011a4a50, 0x44004000433d0000, 0xa49f18, 0xc000e09880, 0x44004000433d0000, 0x44004000433d0000, 0x436f000043e58000, 0xc00025fcf8, 0xc00025fcd0, ...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:190 +0x3ff
fyne.io/fyne/v2/internal/driver.walkObjectTree.func1(...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:176
fyne.io/fyne/v2/internal/driver.walkObjectTree(0xa49f18, 0xc000e09880, 0x43ff8000433d0000, 0xa4a2d8, 0xc000320780, 0x43ff8000433d0000, 0x0, 0x4f0000004f000000, 0xc00025fcf8, 0xc00025fcd0, ...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:190 +0x3ff
fyne.io/fyne/v2/internal/driver.walkObjectTree.func1(...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:176
fyne.io/fyne/v2/internal/driver.walkObjectTree(0xa4a2d8, 0xc000320780, 0x42f00000433d0000, 0xa491f8, 0xc000f2bb80, 0x42f00000433d0000, 0x0, 0x4f0000004f000000, 0xc00025fcf8, 0xc00025fcd0, ...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:190 +0x3ff
fyne.io/fyne/v2/internal/driver.walkObjectTree.func1(...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:176
fyne.io/fyne/v2/internal/driver.walkObjectTree(0xa491f8, 0xc000f2bb80, 0x42f00000433d0000, 0xa491f8, 0xc00046a080, 0x42f00000433d0000, 0x0, 0x4f0000004f000000, 0xc00025fcf8, 0xc00025fcd0, ...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:190 +0x3ff
fyne.io/fyne/v2/internal/driver.walkObjectTree.func1(...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:176
fyne.io/fyne/v2/internal/driver.walkObjectTree(0xa491f8, 0xc00046a080, 0x42040000433d0000, 0xa491f8, 0xc00046a2c0, 0x42040000433d0000, 0x0, 0x4f0000004f000000, 0xc00025fcf8, 0xc00025fcd0, ...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:190 +0x3ff
fyne.io/fyne/v2/internal/driver.walkObjectTree.func1(...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:176
fyne.io/fyne/v2/internal/driver.walkObjectTree(0xa491f8, 0xc00046a2c0, 0x4204000040800000, 0xa49858, 0xc0001641c0, 0x4204000040800000, 0x0, 0x4f0000004f000000, 0xc00025fcf8, 0xc00025fcd0, ...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:190 +0x3ff
fyne.io/fyne/v2/internal/driver.walkObjectTree.func1(...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:176
fyne.io/fyne/v2/internal/driver.walkObjectTree(0xa49858, 0xc0001641c0, 0xc000444000, 0x0, 0x0, 0x0, 0x0, 0x4f0000004f000000, 0xc00025fcf8, 0xc00025fcd0, ...)
        /home/cad/f/tmp/fyne/internal/driver/util.go:190 +0x3ff
fyne.io/fyne/v2/internal/driver.WalkVisibleObjectTree(0xa49858, 0xc0001641c0, 0xc00025fcf8, 0xc00025fcd0, 0xc000444000)
        /home/cad/f/tmp/fyne/internal/driver/util.go:134 +0x94
fyne.io/fyne/v2/internal/driver/common.(*Canvas).walkTree(0xc000436240, 0xc000448400, 0x0, 0xc00025fdd8)
        /home/cad/f/tmp/fyne/internal/driver/common/canvas.go:429 +0x15c
fyne.io/fyne/v2/internal/driver/common.(*Canvas).WalkTrees(0xc000436240, 0x0, 0xc00025fdd8)
        /home/cad/f/tmp/fyne/internal/driver/common/canvas.go:341 +0x4f
fyne.io/fyne/v2/internal/driver/common.(*Canvas).EnsureMinSize(0xc000436240, 0xc00025fe50)
        /home/cad/f/tmp/fyne/internal/driver/common/canvas.go:94 +0x132
fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).repaintWindow.func1()
        /home/cad/f/tmp/fyne/internal/driver/glfw/loop.go:161 +0x45
fyne.io/fyne/v2/internal/driver/glfw.(*window).RunWithContext(0xc000456000, 0xc00025fe90)
        /home/cad/f/tmp/fyne/internal/driver/glfw/window.go:1311 +0x58
fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).repaintWindow(0xc000108730, 0xc000456000)
        /home/cad/f/tmp/fyne/internal/driver/glfw/loop.go:160 +0x65
fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).startDrawThread.func1(0xc000108730, 0xc00044c180, 0xc0004400a0)
        /home/cad/f/tmp/fyne/internal/driver/glfw/loop.go:222 +0x2e9
created by fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).startDrawThread
        /home/cad/f/tmp/fyne/internal/driver/glfw/loop.go:187 +0xbd
```

```
$ cd f/tmp/fyne/
$ git log --format=oneline | head
0dac41336ff75959a6dc89f9c11e59436409b933 Improve alignment code and take care of Hyperlink as well as Text
62fad2294b435470f2868ad661c57e32812a0673 Fix issue with width measurement
6fd307757fccf654d25b2207f03e2e3cbbbae557 Fix removal across segment boundaries
8803c4aa7b1f247a2fdfb70bb2bd7e11a1697fb7 Remove unneeded cache lookup
2b689811c6914e47931236f7d94efc81f562f76a Fix staticcheck
4fd33d24fe995976f68d064ea3f8df0ab21cf085 Small simplification
09f27d2d5269e90e7fd083055a4ffe6153dd27c9 Fix the rowBoundary definition - one per row, many segments
00e70e1d9255ae1025882cc721ffb1400d562c5f Move password to a new style
4db3a33ee10bc3515986491ed9908b4123c1ff52 Add inline hyperlinks
1955a128f69f5e085ecc071dfdef9798afbe3083 Add emphasis and strong types as well
```

```
$ cat /etc/os-release 
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
LOGO=archlinux
$ go version
go version go1.16.5 linux/amd64
```

charlesdaniels:
X-post from Slack for record keeping. Attached is a video of the issue occurring.

https://user-images.githubusercontent.com/11738895/122617974-5022ee80-d05b-11eb-95c7-6f70560be803.mp4

andydotxyz:
> I got a crash while resizing the window:

Thanks, fixed

charlesdaniels:
Confirmed, the issue I reported previously no longer occurs on my machine.

widget/label.go
```diff
@@ -45,6 +42,7 @@ func NewLabelWithStyle(text string, alignment fyne.TextAlign, style fyne.TextSty
 		TextStyle: style,
 	}
 
+	l.ExtendBaseWidget(l.super())
```
fpabl0, line 45: Why is the `l.super()` needed now? 🤔

widget/label.go
```diff
@@ -59,29 +57,59 @@ func (l *Label) Bind(data binding.String) {
 	data.AddListener(l.textListener)
 }
 
-// Refresh checks if the text content should be updated then refreshes the graphical context
+// CreateRenderer is a private method to Fyne which links this widget to its renderer
+func (l *Label) CreateRenderer() fyne.WidgetRenderer {
+	l.provider = NewRichTextWithText(l.Text)
+	l.ExtendBaseWidget(l.super())
```
fpabl0, line 63: same doubt here

widget/label.go
```diff
@@ -59,29 +57,59 @@ func (l *Label) Bind(data binding.String) {
...
 		return
 	}
-
-	if l.Text != string(l.provider.buffer) {
-		l.provider.setText(l.Text)
-	} else {
-		l.provider.updateRowBounds() // if truncate/wrap has changed
-	}
+	l.syncSegments()
+	l.provider.Refresh()
```
fpabl0, line 99: maybe this is not needed, as `l.BaseWidget.Refresh()` will already refresh it, am I missing something?

widget/richtext.go
```diff
@@ -0,0 +1,734 @@
...
+
+// NewRichText returns a new RichText widget that renders the given text and segments.
+// If no segments are specified it will be converted to a single segment using the default text settings.
+//
+// Since: 2.1
+func NewRichText(segments ...RichTextSegment) *RichText {
+	t := &RichText{Segments: segments}
+	t.updateRowBounds()
+	return t
+}
```
fpabl0, line 43: If `ExtendBaseWidget` is not called here (I understand why it is not desired here), then `MinSize()` should be reimplemented to call it, as there are sometimes where `MinSize` is called before `CreateRenderer`. In that case min size will be zero because there is no widget implementation.

widget/richtext_objects.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+// RichTextStyle describes the details of a text object inside a RichText widget.
+//
+// Since: 2.1
+type RichTextStyle struct {
+	Alignment fyne.TextAlign
+	ColorName fyne.ThemeColorName
+	Inline    bool
+	SizeName  fyne.ThemeSizeName
```
fpabl0, line 172: So if users still want to use a custom size for the text, they need to use `canvas.Text` with no text wrapping and other benefits of the RichText? ☹️

theme/theme.go
```diff
@@ -275,6 +285,9 @@ func (t *builtinTheme) initFonts() {
 }
 
 func (t *builtinTheme) Color(n fyne.ThemeColorName, v fyne.ThemeVariant) color.Color {
+	if t.variant != variantNameUserPreference {
+		v = t.variant
+	}
```
fpabl0, line 290: This does not seem to be related to the text refactor, and it would change the current behavior of Light and Dark themes (extending using Light or Dark themes is equal to extending DefaultTheme, after this change this will not true anymore). `currentVariant ()` function would not have much sense after this change. I understand this change is the right way to go, but maybe this should be in another PR? 
Also, I wonder if the theme API could decide if the a theme is desired to handle the variants or the variant handles the themes, because now it is a bit confusing. In application examples, the theme handle the variants, but in the core with the use of the Light and Dark themes works vice versa.

widget/label.go
andydotxyz: It's a good question but the extend tests fail if you use `l.ExtendBaseWidget(l)` instead

andydotxyz: The answer is probably in how the Label's replacement ExtendBaseWidget works

widget/richtext.go
andydotxyz: Excellent, thanks for this - that fixes the flicker that @charlesdaniels reported :)

theme/theme.go
andydotxyz: Without this change the theme lookups do not work as expected.
Ultimately I got the implementation wrong the first time round.
I don't see what has been broken, what this does is ensure that `LightTheme()` and `DarkTheme()` ignore the current user preference.

A theme can feel free to ignore the variant, if that is desired. I don't understand what you mean about core themes having this backwards. LightTheme and DarkTheme are (historical) conveniences for apps that want to turn off the automatic switching. We could deprecate them and force developers to implement custom themes, but it seems overly harsh.

widget/richtext.go
andydotxyz: Fixed :)

widget/label.go
fpabl0: I see, however it seems that the right thing to do is using it inside ExtendBaseWidget reimplementation. By doing this:
```go
func (l *Label) ExtendBaseWidget(w fyne.Widget) {
	if w == nil {
		w = l
	}
	l.BaseWidget.ExtendBaseWidget(w)
	if l.provider != nil {
		l.provider.ExtendBaseWidget(l.super()) // 👈
	}
}
```
In that way we could remove all the `l.ExtendBaseWidget(l.super())` :)
```diff
diff --git a/widget/label.go b/widget/label.go
index b0d70842..d93c484a 100644
--- a/widget/label.go
+++ b/widget/label.go
@@ -42,7 +42,7 @@ func NewLabelWithStyle(text string, alignment fyne.TextAlign, style fyne.TextSty
                TextStyle: style,
        }
 
-       l.ExtendBaseWidget(l.super())
+       l.ExtendBaseWidget(l)
        return l
 }
 
@@ -60,7 +60,7 @@ func (l *Label) Bind(data binding.String) {
 // CreateRenderer is a private method to Fyne which links this widget to its renderer
 func (l *Label) CreateRenderer() fyne.WidgetRenderer {
        l.provider = NewRichTextWithText(l.Text)
-       l.ExtendBaseWidget(l.super())
+       l.ExtendBaseWidget(l)
        l.syncSegments()
 
        return l.provider.CreateRenderer()
@@ -73,7 +73,7 @@ func (l *Label) ExtendBaseWidget(w fyne.Widget) {
        }
        l.BaseWidget.ExtendBaseWidget(w)
        if l.provider != nil {
-               l.provider.ExtendBaseWidget(w)
+               l.provider.ExtendBaseWidget(l.super())
        }
 }
```

widget/label.go
fpabl0: I see, in that case `l.BaseWidget.Refresh()` should be removed, currently it refreshes twice. Because the label renderer is the rich text renderer. This could be just:
```go
func (l *Label) Refresh() {
	if l.provider == nil { // not created until visible
		return
	}
	l.syncSegments()
	l.provider.Refresh()
}
```
And the tests will pass :).

widget/richtext_objects.go
fpabl0: > No that is not required, they can add a new size to the theme

😯  didn't know that was possible, how could I add new sizes to the theme?

widget/label.go
```diff
@@ -59,29 +57,59 @@ func (l *Label) Bind(data binding.String) {
...
+	}
+}
+
+// MinSize returns the size that this label should not shrink below.
+//
+// Implements: fyne.Widget
+func (l *Label) MinSize() fyne.Size {
+	if l.provider == nil {
+		l.CreateRenderer()
+	}
```
fpabl0, line 86: Although this apparently fixes the problem before, this now creates a "phantom" renderer, this should be re-written to:
```go
	if l.provider == nil {
		l.ExtendBaseWidget(l)
		cache.Renderer(l.super())
	}
```

fpabl0 (changes requested):

theme/theme.go
fpabl0: > I don't see what has been broken, what this does is ensure that LightTheme() and DarkTheme() ignore the current user preference.

Yes, that is right, it is just that before if you created a custom theme based on the DarkTheme or LightTheme it didn't matter, because both of them acted like the DefaultTheme.

> A theme can feel free to ignore the variant, if that is desired. I don't understand what you mean about core themes having this backwards. LightTheme and DarkTheme are (historical) conveniences for apps that want to turn off the automatic switching. We could deprecate them and force developers to implement custom themes, but it seems overly harsh.

Completely agreed. But what I mean is that usually you use a new theme for each variant (just like the DarkTheme and LightTheme), but the current Theme API encourages you to define one single theme that handles the variants. That leads to have functions like `theme.Dark (). Color (name, theme.VariantLight)` that evidently has no much sense. So, IMHO, we should rather change the theme interface Color method to accept only the color name without using the variant parameter (and encourage to always define a theme per variant) or remove DarkTheme and LightTheme and use the DefaultTheme and force the variant selection ( in that way we are always encouraging to define a single theme that handles all the variants). The first option should be the right direction as that is how usually others toolkits handle this usecase.

andydotxyz: > But what I mean is that usually you use a new theme for each variant

This is not the case. In 2.0 a theme defines how the app will look and the variant represents the user's preference (often derived from the OS settings). Apps should certainly not be encouraged to decide if they want to look light or dark, them should support both (with one theme).

If there is confusion I can deprecate `LightTheme()` and `DarkTheme()` as really they should not be needed by modern theme creation.

widget/richtext_objects.go
andydotxyz: Inside the `Size` method if your theme look for another `ThemeSizeName` that is defined in your app/theme.
If you see it then return your app/theme specific value.

fpabl0: Oh you are right. I didn't realize it before. That's really great, thanks :)

theme/theme.go
fpabl0: As you said before, "LightTheme and DarkTheme are (historical) conveniences for apps that want to turn off the automatic switching.", so deprecating them without offer a wait to force a specific variant seems not to be good (final users may want to use DarkTheme only in that application instead of activating it in the whole system). So if they are deprecated we should consider adding a method to override the default variant, just like Android, iOS, Flutter and other toolkits allow the developers to do:
https://developer.android.com/guide/topics/ui/look-and-feel/darktheme#change-themes
https://developer.apple.com/documentation/uikit/appearance_customization/supporting_dark_mode_in_your_interface/choosing_a_specific_interface_style_for_your_ios_app

But obviously this should be discussed in detail in other PR related to the theme.

andydotxyz: I am confused here sorry - you seem to be suggesting end user configuration, but they don't have access to the APIs. 
The theme API allows app developers to ignore the user preference of variant by adding their own theme - I don't want to add this to the app API as it is not something we should encourage.

fpabl0: > I am confused here sorry - you seem to be suggesting end user configuration, but they don't have access to the APIs.

End user configuration is the goal but to accomplish it developers would need to access to the APIs. I am aware that this is entirely possible at current state, however if I define a theme that handles Light and Dark variants and I want to let end users choose if they want to follow their system configuration or force to a specific variant, then I have to define two different themes to handle it (or add some logic to my current theme to handle the special case), instead of simply override that variant behavior as Android, iOS and others allow.

andydotxyz: We must be talking at crossed purposes - this is making no sense to me.
Current state:
* Default is system configuration (light or dark)
* User can choose (through fyne_settings or FYNE_THEME) to override this and specify light or dark (or fall back to system default)
* All themes are told which the current variant is in the `Color` call - and they can choose to ignore it.

An app that calls `Settings.SetTheme(DarkTheme())` is overriding all of these system and user settings, because the app wants to force a specific look.

On which part are we not managing to agree on?

fpabl0 (approved):
Looks great! :)

andydotxyz:
Just to follow up on the theme discussion above, I have decided that `LightTheme()` and `DarkTheme()` will be deprecated.
Having installed a few apps it is a real jarring experience when an app decides to override my user preferences for theme variant.

It will continue to be possible to force your theme variant inside an app, but it will require a little custom theme that overrides `Color` method and delegates to `DefaultTheme` passing the variant you want.
This feels correct, as developers will have to work harder to ignore user preference.

fpabl0:
Sounds good :). Perhaps we could introduce a function called ForceTheme or OverridenTheme (or something related) that just facilitates that part to the developer. The best thing is that this would work with custom themes too :). It could be something like:
```go
func ForceTheme(t fyne.Theme, variant fyne.ThemeVariant) fyne.Theme {
  return &forcedTheme{
    Theme: t,
    variant: variant,
  }
}

type forcedTheme struct {
   fyne.Theme
   variant fyne.ThemeVariant
}

func (t *forcedTheme) Color(n ..., v fyne.ThemeVariant) color.Color {
   return t.Theme.Color(n, t.variant)
}
```

andydotxyz:
Hmm, I'm not too sure that should be in core. The code will operate quite effectively from the developers repo, so it does not need to be in the library. But overall it encourages the same thing that I just deprecated.
I guess it could be considered for fyne-x if you feel that lots of apps will make use of such a utility?

fpabl0:
> The code will operate quite effectively from the developers repo, so it does not need to be in the library.

Agree, but this addition would be great to avoid defining every time the above code for every app (only if you need it, of course).

> But overall it encourages the same thing that I just deprecated.

Not the same. LightTheme and DarkTheme encourages you to define two themes to handle light and dark variants and make the developer think it is ok to define a theme just for one variant. However `ForceTheme(theme, variant)` tells you that you have to define a theme that handles the two variants, but you can force it to a specified variant if you want.

> I guess it could be considered for fyne-x if you feel that lots of apps will make use of such a utility?

I don't think it should be in `fyne-x`, the code is so simple that getting all the `fyne-x` repository just for that function does not feel good.

andydotxyz:
> Agree, but this addition would be great to avoid defining every time the above code for every app (only if you need it, of course).

This is not a good reason for including it in core. Apps should not be encouraged to ignore user preferences.

fpabl0:
> Apps should not be encouraged to ignore user preferences.

I don't think the addition would do that. The addition would encourage to use app user preferences instead of the global user preferences (that is annoying in some cases for some people). Using `ForceTheme` still needs that the developer creates a theme that handle both variants (then the developer is aware of what he is doing), while LightTheme and DarkTheme are "plug & play" themes that developers can use them without thinking about handling variants according to the user preferences. 
As I said before, it would be just a helper for the developers, because everything is possible to do with the current API, so it is ok to not have it in the core until we are sure that it is valuable for other developers :)

andydotxyz:
> The addition would encourage to use app user preferences instead of the global user preferences (that is annoying in some cases for some people).

As a toolkit that strives for a consistent approach I have deprecated the `LightTheme` and `DarkTheme` methods and will not add new features encouraging apps to force light or dark versions. As we agree it is possible if apps want to provide custom theming solutions, so there is no require to add new functionality here.

PR #2279, opened 2021-06-06
Add a new SimpleRenderer to handle simple cases.

### LINKED ISSUE #709: Add a "SimpleWidget" or "StaticRenderer" so I can make canvas primitives interactive

andydotxyz:
### Description:
Easiest usage is to return NewSimpleRenderer(myObj) from CreateRenderer

Fixes #709

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


internal/widget/simple_renderer.go
```diff
@@ -0,0 +1,64 @@
...
+	return r.object.MinSize()
+}
+
+// Objects returns the objects that should be rendered.
+//
+// Implements: fyne.WidgetRenderer
+//
+// Since: 2.1
+func (r *SimpleRenderer) Objects() []fyne.CanvasObject {
+	return []fyne.CanvasObject{r.object}
```
fpabl0, line 54: I know that maybe this has not a big impact in performance, but we could get a little improvement by simply defining SimpleRenderer as:
```go
type SimpleRenderer struct {
	objects []fyne.CanvasObject
}

func NewSimpleRenderer(object fyne.CanvasObject) *SimpleRenderer {
	return &SimpleRenderer{objects: []fyne.CanvasObject{object}}
}

...

func (r *SimpleRenderer) Layout(s fyne.Size) {
	r.objects[0].Resize(s)
}

...

func (r *SimpleRenderer) Objects() []fyne.CanvasObject {
	return r.objects
}
```

fpabl0: Benchmarks:

```
BenchmarkRecreatingSlice
BenchmarkRecreatingSlice-4   	989818282	         1.217 ns/op	       0 B/op	       0 allocs/op
BenchmarkReusingSlice
BenchmarkReusingSlice-4      	1000000000	         1.175 ns/op	       0 B/op	       0 allocs/op
```

andydotxyz: Yes that is a great suggestion thanks

dialog/color_picker.go
```diff
@@ -268,19 +268,9 @@ type colorPickerRenderer struct {
...
-	r.contents.Resize(size)
-}
-
-func (r *colorPickerRenderer) MinSize() fyne.Size {
-	return r.contents.MinSize()
-}
-
 func (r *colorPickerRenderer) Refresh() {
 	r.updateObjects()
-	r.Layout(r.picker.Size())
```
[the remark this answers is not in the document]
andydotxyz: This removal was intentional, there is no logic in the Layout call that would be affected by the state of the widget, so Refresh should not have needed to call it.

widget/entry_password.go
```diff
@@ -52,24 +51,16 @@ func (r *passwordRevealer) Tapped(*fyne.PointEvent) {
...
-	return color.Transparent
-}
-
 func (r *passwordRevealerRenderer) Layout(size fyne.Size) {
 	r.icon.Resize(fyne.NewSize(theme.IconInlineSize(), theme.IconInlineSize()))
 	r.icon.Move(fyne.NewPos((size.Width-theme.IconInlineSize())/2, (size.Height-theme.IconInlineSize())/2))
 }
 
-func (r *passwordRevealerRenderer) MinSize() fyne.Size {
-	return fyne.NewSize(theme.IconInlineSize(), theme.IconInlineSize())
```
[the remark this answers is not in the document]
andydotxyz: Good point, thanks

widget/widget.go
```diff
@@ -198,25 +199,8 @@ func (w *DisableableWidget) Disabled() bool {
...
-}
-
-func (s *simpleRenderer) Refresh() {
-	s.content.Refresh()
-}
+// SimpleRenderer is a basic renderer that satisfies widget.Renderer interface by wrapping
+// a single fyne.CanvasObject.
+//
+// Since: 2.1
+type SimpleRenderer = internalWidget.SimpleRenderer
```
fpabl0, line 206: I think this should be changed to:
```go
func NewSimpleRenderer(object fyne.CanvasObject) *internalWidget.SimpleRenderer {
	return internalWidget.NewSimpleRenderer(object)
}
```
Otherwise, there is no way to use `SimpleRenderer` outside the Fyne repo. 
Also I think that with this change every usage of this in this package (`widget`) should use `NewSimpleRenderer` instead of `internalWidget.NewSimpleRenderer` (to keep consistency).

fpabl0 (changes requested):

widget/widget.go
andydotxyz: Thanks, I missed the `NewSimpleRenderer`, and I will update the usages.
However the alias should remain, as we should not return `internal` objects from our public API.

fpabl0: > Thanks, I missed the NewSimpleRenderer, and I will update the usages.

You're welcome :)

> However the alias should remain, as we should not return internal objects from our public API.

Agree 👍

fpabl0:
@andydotxyz I wonder if we should use `fyne.WidgetRenderer` directly instead of introducing a new type to the public API, in that way `SimpleRenderer` could be a private type `simpleRenderer`. By doing this we ensure that nobody tries to use the simple renderer as `&SimpleRenderer{}`. What do you think?

andydotxyz:
Good question. Initially I was following the internal renderers etc and they have exposed methods (like SetObjects).
As this one doesn't we could return `fyne.WidgetRenderer` just now and revise it later if we need to add methods.

widget/form.go
```diff
@@ -261,7 +262,7 @@ func (f *Form) CreateRenderer() fyne.WidgetRenderer {
 	}
 	f.itemGrid = fyne.NewContainerWithLayout(layout.NewFormLayout(), objects...)
 
-	renderer := &simpleRenderer{content: fyne.NewContainerWithLayout(layout.NewVBoxLayout(), f.itemGrid, f.buttonBox)}
+	renderer := internalWidget.NewSimpleRenderer(fyne.NewContainerWithLayout(layout.NewVBoxLayout(), f.itemGrid, f.buttonBox))
```
fpabl0, line 265: Should be just NewSimpleRenderer()

dialog/color_picker.go
```diff
@@ -192,7 +192,7 @@ func (p *colorAdvancedPicker) CreateRenderer() fyne.WidgetRenderer {
 	))
 
 	r := &colorPickerRenderer{
-		BaseRenderer:      internalwidget.NewBaseRenderer([]fyne.CanvasObject{contents}),
+		SimpleRenderer:    internalwidget.NewSimpleRenderer(contents),
```
fpabl0, line 195: I think you could use the renderer from the public widget package directly here 🤔, so we don't need to import the internal widget package.

fpabl0 (commented):
Looks good, just some comments :)

dialog/color_picker.go
andydotxyz: Not exactly, there is no type exported following the previous review. But I can make it a generic WidgetRenderer

fpabl0 (approved):
LGTM! :)

PR #2282, opened 2021-06-14
patch go-gl

toaster:
### Description:

For unknown reason the go-gl/gl core package misses a header file.
This might be caused by the package build process of go-gl.


andydotxyz (approved):
Thanks. It seems that somehow this has been caught up in go's non-inclusion of directories if there is no go files

andydotxyz:
I opened an issue upstream that should fix the issue https://github.com/go-gl/gl/issues/140

PR #2294, opened 2021-06-21
Make the platform specific scancode available

### LINKED ISSUE #1523: Request: raw keyboard events
A widget should be able to capture raw keyboard events, including keydown/keyup events.

For example the GameBoy emulator at `github.com/andydotxyz/fynegameboy` will [check if a key repeats within 200ms](https://github.com/andydotxyz/fynegameboy/blob/master/driver/controller.go#L40) to know if it's being held, which is not a good thing.

Instead there should be a way to access raw events where the Device allows it, and this process should be nothing more than a fallback.

andydotxyz:
This is a complex topic and the int will vary based on platform (macOS is strange).
Some applications need a reference to the source and we can't expose GLFW abstraction

Fixes #1523 

### Checklist:

- [ ] Tests included. <- we don't have any way to test glfw generated events
- [x] Tests all pass.

### DIFF
--- a/event.go
+++ b/event.go
@@ -3,6 +3,8 @@ package fyne
 // KeyEvent describes a keyboard input event.
 type KeyEvent struct {
 	Name KeyName
+	// ScanCode is a platform specific field that reports the hardware ID of the key event
+	ScanCode int
 }
 
 // PointEvent describes a pointer input event. The position is relative to the
--- a/event.go
+++ b/event.go
@@ -1,15 +1,18 @@
 package fyne
 
-// ScanCode represents a hardware ID for (normally desktop) keyboard events.
+// HardwareKey contains information associated with physical key events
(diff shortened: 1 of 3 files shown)


MagicalTux:
We still have some issues, see https://github.com/andydotxyz/fyne/blob/fix/1523/internal/driver/glfw/window.go#L1169

* If the key name is empty (key unknown to fyne), the event won't fire. I do not think it's an issue if the event fires with `keyName = ""` because it just won't do anything, and it's unlikely to be any issue in terms of performance/etc. Might be good to add in the documentation a note about it.
* I was about to talk about the tab key but it seems it's solved in fyne 2.1+ with `fyne.Tabbable`

andydotxyz:
I think the last update should fix your point about unrecognised events

andydotxyz:
Review period lapsed

PR #2299, opened 2021-06-25
Fix single pixel lines

### LINKED ISSUE #2298: 1px thin stroke line
**Is your feature request related to a problem? Please describe:**

Can we use canvas.NewLine() to paint a thin line which is just 1px? 

> line := canvas.NewLine(color.Red)
	line.StrokeWidth = 1
	line.StrokeColor = color.Red
	line.Position1 = fyne.NewPos(0, 10)
	line.Position2 = fyne.NewPos(300, 10)

I use export FYNE_SCALE=0.5 and export FYNE_SCALE=1 to run a demo. But the lines looks like the same strokeWidth. The text looks scaled, but lines seem not scaled as expected.

I'm using MBP. but I don't think it's because the HiDPI. 

In fact we can see thin line if we use widget.Card, we can see it draws a thin border. But Card uses widget.NewShadowingRenderer which is internal.

EA7561FD-B387-4612-869C-7312260E1D4F

**Is it possible to construct a solution with the existing API?**
I don't know how to draw a thin line with fyne-toolkit. 

Thanks

andydotxyz:
### Description:
Better support single px lines even on hidpi screens

Fixes #2298

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (commented):
Are the iOS build changes supposed to be part of this PR? It looks to be a good change, but it seems like it might have accidentally slipped in.

Jacalz (commented):
The lines on the colour wheels in the test images looks like they are almost too thin. Is that expected? If yes, should we perhaps make them a bit thicker?

andydotxyz:
> The lines on the colour wheels in the test images looks like they are almost too thin. Is that expected? If yes, should we perhaps make them a bit thicker?

They were coded as 0px thick, so drawing them as thin as possible is probably what was intended?

andydotxyz:
> Are the iOS build changes supposed to be part of this PR? It looks to be a good change, but it seems like it might have accidentally slipped in.

I think it's some weird merge / rebase artefact - the changes don't appear in the diff.

Jacalz:
> > The lines on the colour wheels in the test images looks like they are almost too thin. Is that expected? If yes, should we perhaps make them a bit thicker?
> 
> 
> 
> They were coded as 0px thick, so drawing them as thin as possible is probably what was intended?

Hmm, maybe.I personally think that they would look better a little bit thicker, closer to before this change.

andydotxyz:
> Hmm, maybe.I personally think that they would look better a little bit thicker, closer to before this change.

This is a subjective matter. I far prefer hairlines like that to be as thin as possible.
The old version was 2px thick, the new is 1px - there isn't really an in-between.
If it needs to be adjusted I'd say we do so separately to this.

Jacalz (approved):
> > Hmm, maybe.I personally think that they would look better a little bit thicker, closer to before this change.
> 
> 
> 
> This is a subjective matter. I far prefer hairlines like that to be as thin as possible.
> 
> The old version was 2px thick, the new is 1px - there isn't really an in-between.
> 
> If it needs to be adjusted I'd say we do so separately to this.

Alright then. I'd have preferred keeping them as thick as before, but I'm going to approve this.

andydotxyz:
> Alright then. I'd have preferred keeping them as thick as before, but I'm going to approve this.

Maybe we should review it with @stuartmscott when he gets back, as he was the original author?

PR #2300, opened 2021-06-25
Fix possible issue with setting floats that have format precision

andydotxyz:
### Description:
The fmt.Scanf does not support precision for float verbs, so this strips that out for the `Set` call.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/convert_helper.go
+++ b/data/binding/convert_helper.go
@@ -50,7 +50,7 @@ func stripFormatPrecision(in string) string {
 	if sizeRunes[0] == '.' { // formats like %.2f
 		return fmt.Sprintf("%s%s", string(runes[:start+1]), string(runes[end:]))
 	}
-	return fmt.Sprintf("%s%d%s", string(runes[:start+1]), int(width), string(runes[end:]))
+	return string(runes[:start+1]) + strconv.Itoa(int(width)) + string(runes[end:])
 }
 
 func uriFromString(in string) (fyne.URI, error) {
--- a/data/binding/convert_helper.go
+++ b/data/binding/convert_helper.go
@@ -48,7 +48,7 @@ func stripFormatPrecision(in string) string {
 	}
 
 	if sizeRunes[0] == '.' { // formats like %.2f
-		return fmt.Sprintf("%s%s", string(runes[:start+1]), string(runes[end:]))
+		return string(runes[:start+1]) + string(runes[end:])
 	}
 	return string(runes[:start+1]) + strconv.Itoa(int(width)) + string(runes[end:])
 }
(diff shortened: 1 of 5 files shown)


data/binding/convert_helper.go
```diff
@@ -1,12 +1,58 @@
...
+	}
+
+	sizeRunes := runes[start+1 : end]
+	width, err := parseFloat(string(sizeRunes))
+	if err != nil {
+		return string(runes[:start+1]) + string(runes[:end])
+	}
+
+	if sizeRunes[0] == '.' { // formats like %.2f
+		return fmt.Sprintf("%s%s", string(runes[:start+1]), string(runes[end:]))
```
Jacalz, line 51: Wouldn't it be better to just do as above?
```suggestion
		return string(runes[:start+1]) + string(runes[end:])
```
If so, you might even be able to move it into the error check above.

data/binding/convert_helper.go
```diff
@@ -1,12 +1,58 @@
...
+	}
+	return fmt.Sprintf("%s%d%s", string(runes[:start+1]), int(width), string(runes[end:]))
```
Jacalz, line 53: I think this should be more performant and not much header to read in my opinion.
```suggestion
	return string(runes[:start+1]) + strconv.Itoa(int(width)) + string(runes[end:])
```

data/binding/convert_helper.go
```diff
@@ -1,12 +1,58 @@
...
+func stripFormatPrecision(in string) string {
+	// quick exit if certainly not float
+	if !strings.ContainsAny(in, "f") {
+		return in
+	}
+
+	start := -1
+	end := -1
+	runes := []rune(in)
+	for i, r := range runes {
```
Jacalz, line 21: Shouldn't you just be able to use the string directly here and below? Ranging over the string will give you runes and the rune slice is converted to strings everywhere either way. I

Jacalz (changes requested):

data/binding/convert_helper.go
andydotxyz: Hmm, maybe? I read recently a warning about ranging on string but I don’t remember what it was so was being cautious. We need `[]rune` anyway for the slicing so using the raw string is just re-converting therefor overhead.

andydotxyz: I think we came to the conclusion on Slack that `[]rune` is the right thing. Can this be unblocked @Jacalz?

Jacalz: Oh yeah, sorry. I forgot about that part.

Jacalz (approved):
Looks good to me. Sorry for the holdup.

PR #2310, opened 2021-07-07
Fix: hidden table seperators being rendered incorrectly

### LINKED ISSUE #2266: Extra row dividers rendered on using SetColumnWidth to update a table post initial render
### Describe the bug:
When `SetColumnWidth` is called on a table after it has been initially rendered, a few extra row dividers are rendered.
On playing around with the amount of underlying data that was being updated, I noticed that this issue generally occurs when the underlying data set has a large number of columns (upwards of 10 columns as per my tests).
I also noticed a trend as such:
* Less than 10 columns did not reproduce the issue
* ~10-15 columns caused the extra lines to be rendered, but the lines gradually disappeared on scrolling the table or resizing the window
* Greater than 15 columns caused the lines to be persistent, and they did not disappear even on scrolling or resizing

### To Reproduce:
1. Run the example code provided below
2. Notice the extra row dividers being rendered of the lower half of the window

### Screenshots:
fyne-table-render-issue

### Example code:
```go
import (
	"fmt"
	"time"

	"fyne.io/fyne/v2"
	fyneApp "fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

const (
	appID    = "com.github.ankushjadhav.dynamoist"
	appTitle = "Dynamoist"
)


func getColWidths(data [][]string) []float32 {
	res := make([]float32, 0)
	for _, row := range data {
		for i, col := range row {
			cur := float32(len(fmt.Sprint(col)) * 10)
			if len(res) <= i {
				res = append(res, cur)
			} else {
				if res[i] <= cur {
					res[i] = cur
				}
			}
		}
	}
	return res
}

func main() {
	gui := fyneApp.NewWithID(appID)
	w := gui.NewWindow(appTitle)

	data := make([][]string, 0)

	t := widget.NewTable(func() (int, int) {
		if len(data) > 0 {
			return len(data), len(data[0])
		}
		return 0, 0
	}, func() fyne.CanvasObject {
		return widget.NewLabel("")
	}, func(tci widget.TableCellID, co fyne.CanvasObject) {
		co.(*widget.Label).SetText(data[tci.Row][tci.Col])
	})

	w.SetContent(t)

	go func() {
		time.Sleep(2 * time.Second)
		for i := 0; i < 200; i++ {
			row := make([]string, 0)
			for j := 0; j < 20; j++ {
				row = append(row, fmt.Sprintf("                       "))
			}
			data = append(data, row)
		}
		wi := getColWidths(data)
		for i, v := range wi {
			if v > 200 {
				v = 200
			}
			t.SetColumnWidth(i, v)
		}
		t.Refresh()
	}()

	w.Resize(fyne.Size{700, 700})
	w.CenterOnScreen()
	w.ShowAndRun()
}
```

 - **OS:** MacOS Big Sur
 - **Version:** 11.1
 - **Go version:** 1.16.3
 - **Fyne version:** 2.0.2

AnkushJadhav:
### Description:
The table separators (that are not in view) that should be hidden were still being rendered due to an incorrect variable being used.
Fixed the issue and the incorrect test templates.

Fixes #2266 

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/table.go
+++ b/widget/table.go
@@ -373,7 +373,7 @@ func (t *tableRenderer) moveIndicators() {
 	}
 
 	for i := divs; i < len(t.dividers); i++ {
-		t.dividers[divs].Hide()
+		t.dividers[i].Hide()
 	}
 	canvas.Refresh(t.t)
 }
--- a/widget/testdata/table/desktop/hovered.xml
+++ b/widget/testdata/table/desktop/hovered.xml
@@ -42,12 +42,6 @@
 			<widget pos="4,79" size="168x1" type="*widget.Separator">
 				<rectangle fillColor="disabled" size="168x1"/>
 			</widget>
-			<widget size="0x0" type="*widget.Separator">
-				<rectangle fillColor="disabled" size="0x0"/>
-			</widget>
-			<widget size="0x0" type="*widget.Separator">
-				<rectangle fillColor="disabled" size="0x0"/>
-			</widget>
 		</widget>
 	</content>
 </canvas>
--- a/widget/testdata/table/desktop/hovered_out.xml
+++ b/widget/testdata/table/desktop/hovered_out.xml
@@ -31,15 +31,6 @@
 			<widget pos="4,41" size="168x1" type="*widget.Separator">
 				<rectangle fillColor="disabled" size="168x1"/>
 			</widget>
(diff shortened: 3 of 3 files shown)


andydotxyz (approved):
Excellent thanks :)

PR #2312, opened 2021-07-07
add api to open popup menu on secondaty tap on collection

zdima:
### Description:
Add API to List, Tree and Table to show a popup menu when user secondary click on collection item.

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -32,6 +32,7 @@ type List struct {
 	UpdateItem   func(id ListItemID, item fyne.CanvasObject)
 	OnSelected   func(id ListItemID)
 	OnUnselected func(id ListItemID)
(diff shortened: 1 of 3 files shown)


andydotxyz:
I'm not sure that this is the right place for the API. 
A table can hold different widgets (currently by using containers, and in the future by having per-column types) and the pop-up menu should be specific to the widget context not it's container.

Jacalz (commented):
I agree with Andrew on this. I think this should be up to the user to implement by, for example, extending the collection or it being part of the widget that is being used.

Jacalz:
Thanks for your hard work. I'll close this given the above reasoning that this perhaps shouldn't be part of the API.

PR #2315, opened 2021-07-08
bugfix #2013: addAttributes return parameter is always nil, and not used

### LINKED ISSUE #2013: function never returns a value
### Describe the bug:

Function `addAttributes` in `cmd/fyne/internal/mobile/binres/binres.go` always returns nil, never the promised attributes xml.

nealmcc:
### Description:
Removed unused return values, to avoid ambiguity.  These functions alter their input parameters.

Fixes #2013 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
Thanks for looking at this. I think that a lot of the test code could be removed if you use the stretchr/testify package that we use elsewhere for assert etc

Jacalz:
Can you please use PR titles what they are fixing, instead of what issue they fix? Makes it much easier to glance at an issue an know what it does. Keeping track of a lot of numbers is hard :)

andydotxyz (approved):
Great thanks :)

andydotxyz (changes requested):
Oops, looks like there is another check error:
Error: cmd/fyne/internal/mobile/binres/binres_test.go:60:6: func compareBytes is unused (U1000)

nealmcc:
Sorry about the spam - I meant to wait and push once I'd done this refactoring

nealmcc:
> Oops, looks like there is another check error:
> Error: cmd/fyne/internal/mobile/binres/binres_test.go:60:6: func compareBytes is unused (U1000)

I've removed this helper function since the testify/require library achieves the same thing

nealmcc:
I realised after I pushed the refactoring that you'd already nearly approved the PR without it, and the additional refactoring is likely more time consuming to review than  you would like.

My intent was to clarify (for myself as much as anything) the purpose of each function (particularly the return values vs modified parameters). I don't believe I've changed the behaviour at all and the tests that I added from upstream still pass.  However, if you'd prefer to simplify the PR, I'm happy revert 8461c79 and  7ab8f58

andydotxyz (approved):

PR #2317, opened 2021-07-10
Export table scrolling methods

### LINKED ISSUE #1892: Need to reset the scroll position of table widget

nealmcc:
### Description:
This PR adds the following public methods to widget.Table:
- ScrollTo(id)
- ScrollToEnd()
- ScrollToStart()

Fixes #1892 

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

I'm not certain about the public API style.  The issue mentioned creating ScrollToTop() and ScrollToBottom() but a table is 2D, so I went with ScrollToStart() and ScrollToEnd().   Any thoughts on this?

I figured I'd get some feedback on this before working on scrolling for widget.Tree

### DIFF
--- a/widget/table.go
+++ b/widget/table.go
@@ -92,10 +92,7 @@ func (t *Table) Select(id TableCellID) {
 	}
 	t.selectedCell = &id
 
-	t.scrollTo(id)
-	if t.moveCallback != nil {
-		t.moveCallback()
-	}
+	t.ScrollTo(id)
 
 	if f := t.OnSelected; f != nil {
 		f(id)
@@ -161,6 +158,10 @@ func (t *Table) ScrollTo(id TableCellID) {
 		return
 	}
 
+	if t.scroll == nil {
+		return
+	}
+
 	rows, cols := t.Length()
 	if id.Row >= rows {
 		id.Row = rows - 1
@@ -170,19 +171,42 @@ func (t *Table) ScrollTo(id TableCellID) {
 		id.Col = cols - 1
 	}
 
-	t.scrollTo(id)
+	scrollPos := t.offset
+
+	cellX, cellWidth := t.findX(id.Col)
+	if cellX < scrollPos.X {
+		scrollPos.X = cellX
+	} else if cellX+cellWidth > scrollPos.X+t.scroll.Size().Width {
+		scrollPos.X = cellX + cellWidth - t.scroll.Size().Width
+	}
+
+	cellY, cellHeight := t.findY(id.Row)
+	if cellY < scrollPos.Y {
+		scrollPos.Y = cellY
+	} else if cellY+cellHeight > scrollPos.Y+t.scroll.Size().Height {
+		scrollPos.Y = cellY + cellHeight - t.scroll.Size().Height
+	}
+
+	t.scroll.Offset = scrollPos
+	t.offset = scrollPos
+	t.finishScroll()
 }
 
 // ScrollToBottom scrolls to the last row in the table
 //
 // Since: 2.1
 func (t *Table) ScrollToBottom() {
-	if t.Length == nil {
+	if t.Length == nil || t.scroll == nil {
 		return
 	}
 
 	rows, _ := t.Length()
-	t.scrollTo(TableCellID{Row: rows - 1})
+	cellY, cellHeight := t.findY(rows - 1)
+	y := cellY + cellHeight - t.scroll.Size().Height
+
+	t.scroll.Offset.Y = y
+	t.offset.Y = y
+	t.finishScroll()
 }
 
 // ScrollToLeading scrolls horizontally to the leading edge of the table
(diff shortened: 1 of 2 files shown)


andydotxyz:
Is End and Start best? It seems unlikely that bottom-right is more useful than bottom left? How about top and bottom like list with possible leading and trailing as well?

nealmcc:
So, what would leading and trailing do? scroll all the way left and all the way right, without changing the vertical offset?

andydotxyz:
Yes I think so, if that makes sense?

widget/table.go
```diff
@@ -151,6 +151,101 @@ func (t *Table) UnselectAll() {
...
+// ScrollToBottom scrolls to the last row in the table
+//
+// Since: 2.1
+func (t *Table) ScrollToBottom() {
+	if t.Length == nil {
+		return
+	}
+
+	rows, _ := t.Length()
+	t.scrollTo(TableCellID{Row: rows - 1})
```
andydotxyz, line 185: This is also going to scroll to column 0 - shouldn't it respect the current column if you're only scrolling vertically?

widget/table.go
```diff
@@ -151,6 +151,101 @@ func (t *Table) UnselectAll() {
...
+		t.moveCallback()
+	}
+	t.scroll.Refresh()
+	t.cells.Refresh()
+}
+
+// ScrollToTop scrolls to the first row in the table
+//
+// Since: 2.1
+func (t *Table) ScrollToTop() {
```
andydotxyz, line 208: This will also scroll to the first column, is that intended?

widget/table.go
```diff
@@ -151,6 +151,101 @@ func (t *Table) UnselectAll() {
...
+		return
+	}
+
+	_, cols := t.Length()
+	lastCol := cols - 1
+
+	cellSize := t.templateSize()
+	cellX := float32(0)
+	cellWidth := float32(0)
+	for i := 0; i <= lastCol; i++ {
```
andydotxyz, line 226: I think this is duplicating code in `scrollTo` - if so can you extract it to another method to avoid this?

widget/table_test.go
```diff
@@ -189,6 +189,239 @@ func TestTable_Refresh(t *testing.T) {
...
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			table.ScrollTo(tc.in)
+			assert.Equal(t, tc.want, table.offset)
+			assert.Equal(t, tc.want, table.scroll.Offset)
+		})
+	}
+}
+
+func TestTable_ScrollToBottom(t *testing.T) {
```
andydotxyz, line 289: This function (and the 3 others that test a 1D scroll) currently test if the non-scrolling direction is 0.
I think it should also check that the position is preserved when scrolling, it would have caught some issues mentioned above.

widget/table_test.go
```diff
@@ -189,6 +189,239 @@ func TestTable_Refresh(t *testing.T) {
...
+		func(id TableCellID, c fyne.CanvasObject) {
+			text := fmt.Sprintf("Cell %d, %d", id.Row, id.Col)
+			c.(*Label).SetText(text)
+		})
+
+	w := test.NewWindow(table)
+	defer w.Close()
+
+	var (
+		padding float32 = th.Size(theme.SizeNamePadding)
```
andydotxyz, line 404: Isn't this just `theme.Padding()`?

widget/table_test.go
```diff
@@ -189,6 +189,239 @@ func TestTable_Refresh(t *testing.T) {
...
+		sep             = th.Size(theme.SizeNameSeparatorThickness)
+		sz1             = 4 * (cellWidth + sep) // the inner scroll area
```
andydotxyz, line 406: Comments after variable names normally show that they could have been named better. Please use a name that is clear so the comment can be removed.

widget/table_test.go
```diff
@@ -189,6 +189,239 @@ func TestTable_Refresh(t *testing.T) {
...
+func TestTable_ScrollToTrailing(t *testing.T) {
+	test.NewApp()
+	defer test.NewApp()
+	th := test.NewTheme()
+	test.ApplyTheme(t, th)
+
+	const (
+		maxRows    int     = 20
+		maxCols    int     = 8
+		cellWidth  float32 = 109 // observed from actual output
```
andydotxyz, line 386: Don't worry about commenting how a variable was set up. Tests often have "magic numbers" as long as they are well named it's clear enough. You could put comments like that in the commit or PR if you like.

andydotxyz (changes requested):
Great addition, just a few comments on the exact functionality and comment/variable style

widget/table.go
nealmcc: It was, based on your previous comment about scrolling to the beginning of the last row being more useful than to the end of the last row.

However, with the addition of ScrollToLeading (and Trailing) then I guess this should just scroll vertically.

andydotxyz: Apologies for the implication, I had not meant that we should reset the X axis value.

widget/table_test.go
nealmcc: That was intentional, but I'll change it so they are only 1D scrolls.

widget/table_test.go
nealmcc: I didn't realise that was a thing :) Now I know

widget/table_test.go
nealmcc: Guilty.

andydotxyz (approved):
Good stuff, thanks again :)

PR #2319, opened 2021-07-11
Feature/scroll tree

### LINKED ISSUE #1892: Need to reset the scroll position of table widget

nealmcc:
### Description:
Adds the following exported methods to widget.Tree:

 - ScrollToTop()
 - ScrollToBottom()
 - ScrollTo(TreeNodeID)

Fixes #1892

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


widget/tree.go
```diff
@@ -174,6 +174,52 @@ func (t *Tree) Resize(size fyne.Size) {
 	t.Refresh() // trigger a redraw
 }
 
+// ScrollToBottom scrolls to the bottom of the tree
```
andydotxyz, line 177: Please use full sentences in method comments (probably just needs a ".")
Each of these methods is missing the "Since: 2.1" line

andydotxyz (changes requested):
Nice, just notes about comments and Since.

Also are the test images required? All we are testing is that the offset changes I think.
If you do want to serialise the output please use the Markup style instead of image - it is much smaller on the repo and easier to track changes. We are trying to move to using image tests just for validating render (i.e initial look and theme change)

nealmcc:
> Nice, just notes about comments and Since.
> 
> Also are the test images required? All we are testing is that the offset changes I think.
> If you do want to serialise the output please use the Markup style instead of image - it is much smaller on the repo and easier to track changes. We are trying to move to using image tests just for validating render (i.e initial look and theme change)

Makes sense - in large part I was using the image assertion to see what a tree widget looked like, and how it behaved - I hadn't used them before :)

nealmcc:
Should be good to go now :)

andydotxyz (approved):
Sweet, thanks

PR #2320, opened 2021-07-12
Add baseline to font measurement

### LINKED ISSUE #1859: [Feature] MeasureTextDistanceToBaseline function

Not really, but it would be interesting to have a way to measure the distance from the top of the canvas.Text to its baseline (for alignment purposes in the future), as the current function `fyne.MeasureText` only returns the ideal size to be drawn and it has some "padding" on top and bottom.

Yes, it is possible, although it means introducing a new function.

Maybe there could be an internal function called `RenderedTextDistanceToBaseline` or something like it, it could be implemented with the functions used to implement `RenderedTextSize`:

```
func RenderedTextDistanceToBaseline(size float32, style fyne.TextStyle) float32 {
	var opts truetype.Options
	opts.Size = float64(size)
	opts.DPI = TextDPI
	face := CachedFontFace(style, &opts)
	return float32(face.Metrics().Ascent.Ceil())
}
```

Then there could be a public function called `fyne.MeasureTextDistanceToBaseline` that could be implemented just like `fyne.MeasureText`:

```
func MeasureTextDistanceToBaseline(size float32, style TextStyle) Size {
	return CurrentApp().Driver().RenderedTextDistanceToBaseline(size, style)
}
```

andydotxyz:
### Description:
Add font information to (driver) text measurement.
The `MeasureText` call remains consistent so no public API breakage (except driver, as discussed elsewhere).
There could be a new version of that to expose baseline info, but I don't yet see the need.

What this also did was fix precision of text calculations so a *lot* of tests/images have changed slightly.

Fixes #1859

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


driver.go
```diff
@@ -10,7 +10,7 @@ type Driver interface {
 
 	// RenderedTextSize returns the size required to render the given string of specified
 	// font size and style.
-	RenderedTextSize(string, float32, TextStyle) Size
+	RenderedTextSize(string, float32, TextStyle) (Size, float32)
```
fpabl0, line 13: I wonder if we should name the input and output parameters here, as it is confusing what kind of data it needs to receive and what kind of data it will return. Also, I think the comment should be updated too.

internal/driver/glfw/driver.go
```diff
@@ -39,7 +39,7 @@ type gLDriver struct {
 	animation *animation.Runner
 }
 
-func (d *gLDriver) RenderedTextSize(text string, size float32, style fyne.TextStyle) fyne.Size {
+func (d *gLDriver) RenderedTextSize(text string, size float32, style fyne.TextStyle) (fyne.Size, float32) {
```
fpabl0, line 42: using named return parameters may improve readability here :)

internal/driver/glfw/driver_test.go
```diff
@@ -64,47 +64,57 @@ func Test_gLDriver_AbsolutePositionForObject(t *testing.T) {
 	m2 := mbarCont.Objects[1]
 
 	tests := map[string]struct {
-		object fyne.CanvasObject
-		want   fyne.Position
+		object       fyne.CanvasObject
+		wantX, wantY int
```
fpabl0, line 68: any particular reason about this change?

internal/driver/gomobile/driver.go
```diff
@@ -98,7 +98,7 @@ func (d *mobileDriver) currentWindow() *window {
 	return last
 }
 
-func (d *mobileDriver) RenderedTextSize(text string, size float32, style fyne.TextStyle) fyne.Size {
+func (d *mobileDriver) RenderedTextSize(text string, size float32, style fyne.TextStyle) (fyne.Size, float32) {
```
fpabl0, line 101: same as in gl driver

widget/richtext.go
```diff
@@ -549,14 +549,47 @@ func (r *textRenderer) layoutRow(texts []fyne.CanvasObject, align fyne.TextAlign
...
-
+		var size fyne.Size
+		if txt, ok := text.(*canvas.Text); ok {
+			s, base := fyne.CurrentApp().Driver().RenderedTextSize(txt.Text, txt.TextSize, txt.TextStyle)
+			if tallestBaseline == 0 {
+				tallestBaseline = base
+			} else if base != tallestBaseline {
+				tallestBaseline = fyne.Max(tallestBaseline, base)
+				realign = true
+			}
```
fpabl0, line 563: I think this could be improved to:
```go
if base > tallestBaseline {
    realign = realign || tallestBaseline > 0
    tallestBaseline = base
}
```

widget/richtext.go
```diff
@@ -549,14 +549,47 @@ func (r *textRenderer) layoutRow(texts []fyne.CanvasObject, align fyne.TextAlign
...
+		text.Move(fyne.NewPos(xPos, yPos))
+
 		xPos += size.Width
-		height = fyne.Max(height, size.Height)
+		if height == 0 {
+			height = size.Height
+		} else if height != size.Height {
+			height = fyne.Max(height, size.Height)
+			realign = true
+		}
```
fpabl0, line 577: same here:
```go
if size.Height > height {
    realign = realign || height > 0
    height = size.Height
}
```

widget/richtext.go
```diff
@@ -549,14 +549,47 @@ func (r *textRenderer) layoutRow(texts []fyne.CanvasObject, align fyne.TextAlign
...
 	}
+
+	if realign {
+		for _, text := range texts {
+			if _, ok := text.(*canvas.Text); !ok {
+				continue
+			}
+
+			txt := text.(*canvas.Text)
```
fpabl0, line 586: here if text is a canvas.Text we are asserting it twice. It should be:
```go
txt, ok := text.(*canvas.Text)
if !ok {
   continue
}
```

widget/richtext.go
```diff
@@ -549,14 +549,47 @@ func (r *textRenderer) layoutRow(texts []fyne.CanvasObject, align fyne.TextAlign
...
+			_, base := fyne.CurrentApp().Driver().RenderedTextSize(txt.Text, txt.TextSize, txt.TextStyle)
```
fpabl0, line 587: Hmm I wonder if this has performance impact as `RenderedTextSize` is measuring the text each time even we only need the baseline, maybe the text baseline measurement should have its own function separated from the text size?

internal/driver/glfw/driver_test.go
andydotxyz: Yes, the positions are no longer round whole numbers, so comparing the int form is easier

widget/richtext.go
andydotxyz: Yes that is fewer lines of code but it took me a while to figure hot how it works - do you think it is clearer than the original PR?

widget/richtext.go
andydotxyz: The expensive part of measuring is the text lookup, parse and so forth.
The best solution is probably to cache text measurement, which I was going to add later.

internal/driver/glfw/driver_test.go
fpabl0: I see, got it :)

widget/richtext.go
fpabl0: Hmm, no in that form I think is not very clear, maybe this is clearer? 
```go
if base > tallestBaseline {
    if tallestBaseline > 0 {
        realign = true
    }
    tallestBaseline = base
}
```

widget/richtext.go
fpabl0: Yes, cache could help, however I think it would be great, if apart from the RenderedTextSize, an additional baseline function separated from this, because as far as I know to get the baseline we just need the text style, text size, and one random character 🧐

widget/richtext.go
```diff
@@ -562,7 +562,20 @@ func (r *textRenderer) layoutRow(texts []fyne.CanvasObject, align fyne.TextAlign
...
+				s, base := fyne.CurrentApp().Driver().RenderedTextSize(link.Text, theme.TextSize(), link.TextStyle)
+				if tallestBaseline == 0 {
+					tallestBaseline = base
+				} else if base != tallestBaseline {
+					tallestBaseline = fyne.Max(tallestBaseline, base)
+					realign = true
+				}
+				size = s
+			}
+		}
```
fpabl0, line 577: Hmm it is desired to support more kind of "texts", I think maybe it is better to reorganize this code to something like this:
```go
var (
  textStr string
  textSize float32
  textStyle TextStyle
  size fyne.Size
)
switch v := text.(type) {
   case *canvas.Text:
       textStr = v.Text
       textSize = v.TextSize
       textStyle = v.TextStyle
   case *fyne.Container:
       wid := c.Objects[0]
       if link, ok := wid.(*Hyperlink); ok {
           textStr = link.Text
           textSize = theme.TextSize()
           textStyle = link.TextStyle
       }
}
if textSize > 0 {
    	s, base := fyne.CurrentApp().Driver().RenderedTextSize(txt.Text, txt.TextSize, txt.TextStyle)
	if tallestBaseline == 0 {
		tallestBaseline = base
	} else if base != tallestBaseline {
	         tallestBaseline = fyne.Max(tallestBaseline, base)
		realign = true
	}
        s = size
} else {
        size = text.MinSize()
}
```
But I am not totally sure if this is clearer or not, what do you think?

widget/richtext.go
```diff
@@ -579,11 +592,19 @@ func (r *textRenderer) layoutRow(texts []fyne.CanvasObject, align fyne.TextAlign
...
+						_, base := fyne.CurrentApp().Driver().RenderedTextSize(link.Text, theme.TextSize(), link.TextStyle)
+						delta := tallestBaseline - base
+						text.Move(fyne.NewPos(c.Position().X, yPos+delta))
+					}
+				}
 				continue
 			}
 
-			txt := text.(*canvas.Text)
 			_, base := fyne.CurrentApp().Driver().RenderedTextSize(txt.Text, txt.TextSize, txt.TextStyle)
```
fpabl0: We could avoid doing the same if we save the baselines we got before :), in that case this part could be simply rewritten to:
```go
if realign {
    for i, text := range texts {
        delta := tallestBaseline - baselines[i]
        text.Move(fyne.NewPos(text.Position().X, yPos+delta))
    }
}
...
```

widget/richtext.go
andydotxyz: The cache is in. I would like to avoid adding new methods if possible. Should we need to expose it later we can - but probably by replacing `MeasureText` with `TextMetrics` instead of adding another `TextBaseline` and how many others may be added in the future

widget/richtext.go
```diff
@@ -562,7 +562,20 @@ func (r *textRenderer) layoutRow(texts []fyne.CanvasObject, align fyne.TextAlign
...
+				} else if base != tallestBaseline {
+					tallestBaseline = fyne.Max(tallestBaseline, base)
+					realign = true
+				}
+				size = s
+			}
+		}
+		if size.IsZero() {
 			size = text.MinSize()
 		}
```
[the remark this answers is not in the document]
andydotxyz: That is quite a broad question. If they have a custom text segment using the style based API we created then it will align. If they implement a whole new thing returning a hierarchy of objects that somewhere within holds some text, no it may not align correctly.
One step at a time ;)

widget/richtext.go
andydotxyz: In the future it may be desirable to find a way to make a more generalised alignable API for a RichText element - but I think it is too early to know how, so I just added these two known types for now.

widget/richtext.go
andydotxyz: I'm not sure if that tidy solution is possible with HyperLink in here too is it?

widget/richtext.go
fpabl0: Sounds good

widget/richtext.go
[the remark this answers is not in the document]
fpabl0: Got it. Maybe it could be easier if `Visual()` method in `RichTextSegment` interface returns a custom interface like `RichTextVisual`:
```go
type RichTextVisual interface {
    fyne.CanvasObject
    TextMetrics() (size Size, baseline float32)
}
```
So developers could create custom segments. This addition would facilitate a lot the internal processes too as now we would not need to assert the type to find the data needed to get the text metrics. But, maybe that could be a later improvement.

widget/richtext.go
fpabl0: Got it, my above suggestion may improve this too :), but agree that could be a future improvement, although I think it is relatively easy to add it now 🧐

widget/richtext.go
fpabl0: Yes, it is. This code is used for HyperLink:
```go
delta := tallestBaseline - base
text.Move(fyne.NewPos(c.Position().X, yPos+delta))
```
And `c` is the container that was got from `c, ok := text.(*fyne.Container)`, so `c == text`, then the above code is compatible :)

driver.go
```diff
@@ -10,7 +10,7 @@ type Driver interface {
 
 	// RenderedTextSize returns the size required to render the given string of specified
 	// font size and style.
-	RenderedTextSize(string, float32, TextStyle) (Size, float32)
+	RenderedTextSize(text string, fontSize float32, style TextStyle) (bounds Size, baseline float32)
```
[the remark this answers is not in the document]
andydotxyz: I was trying to avoid the stutter, happy to revert

widget/richtext.go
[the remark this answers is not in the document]
andydotxyz: This assumes that the `TextSegment` is not customisable enough. At this time it is too early to say that it is not IMHO.

widget/richtext.go
andydotxyz: Ah yes I see, updated

widget/richtext.go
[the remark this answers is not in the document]
fpabl0: Sounds good :)

internal/cache/svg.go
```diff
@@ -20,11 +20,15 @@ func GetSvg(name string, w int, h int) *image.NRGBA {
 	return sinfo.pix
 }
 
-// ResetSvg clears all the svg cache map
-func ResetSvg() {
+// ResetThemeCaches clears all the svg and text size cache maps
+func ResetThemeCaches() {
```
fpabl0, line 24: Hmm, maybe it should be better to move this function to base.go? To keep here only the svg related functions.

internal/cache/text.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"sync"
+
+	"fyne.io/fyne/v2"
+)
+
+var (
+	sizeCache = map[sizeEntry]fontMetric{}
+	sizeLock  = sync.RWMutex{}
```
fpabl0, line 11: I think this should be renamed to `fontSizeXxx`.

internal/cache/text.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+type fontMetric struct {
+	bound    fyne.Size
```
fpabl0, line 15: should be `size` as this was changed in other places?

internal/cache/svg.go
andydotxyz: That was my first choice, but that file relates to a graphical cache used for painting and renderers etc so it didn't seem to fit. I will try looking at it again.

andydotxyz: Seems like it wasn't the clean cut I thought. I moved it in and moved some svg stuff out of base.go so it made more sense.

internal/cache/text.go
```diff
@@ -7,37 +7,37 @@ import (
...
-	sizeLock.RUnlock()
+	ent := fontSizeEntry{text, size, style}
+	fontSizeLock.RLock()
+	ret, ok := fontSizeCache[ent]
+	fontSizeLock.RUnlock()
 	if !ok {
 		return fyne.Size{Width: 0, Height: 0}, 0
 	}
-	return ret.bound, ret.baseLine
+	return ret.size, ret.baseLine
```
fpabl0: I hadn't seen carefully this function before. Should be better to rename this to GetFontMetrics? and add named return parameters for clarity?

internal/cache/text.go
```diff
@@ -7,37 +7,37 @@ import (
...
 }
 
 // SetFontSize stores a calculated font size for parameters that were missing from the cache.
-func SetFontSize(text string, size float32, style fyne.TextStyle, bound fyne.Size, base float32) {
-	ent := sizeEntry{text, size, style}
-	sizeLock.Lock()
-	sizeCache[ent] = fontMetric{bound: bound, baseLine: base}
-	sizeLock.Unlock()
+func SetFontSize(text string, fontSize float32, style fyne.TextStyle, size fyne.Size, base float32) {
```
fpabl0, line 38: Same comment here

internal/cache/text.go
andydotxyz: Sure, done

internal/cache/text.go
```diff
@@ -22,9 +22,9 @@ type fontSizeEntry struct {
 	style fyne.TextStyle
 }
 
-// GetFontSize looks up a calculated size required for the specified text parameters.
-func GetFontSize(text string, size float32, style fyne.TextStyle) (fyne.Size, float32) {
-	ent := fontSizeEntry{text, size, style}
+// GetFontMetrics looks up a calculated size and baseline required for the specified text parameters.
+func GetFontMetrics(text string, fontSize float32, style fyne.TextStyle) (size fyne.Size, base float32) {
+	ent := fontSizeEntry{text, fontSize, style}
 	fontSizeLock.RLock()
```
fpabl0, line 28: `fontSizeEntry` should be renamed to `fontMetricEntry` too?

internal/painter/font.go
```diff
@@ -33,13 +33,13 @@ func loadFont(data fyne.Resource) *truetype.Font {
...
+	bound, base := cache.GetFontMetrics(text, size, style)
 	if base != 0 {
 		return bound, base
 	}
 
 	bound, base = measureText(text, size, style)
-	cache.SetFontSize(text, size, style, bound, base)
+	cache.SetFontMetrics(text, size, style, bound, base)
 	return bound, base
 }
```
fpabl0: To keep everything clear, I think the `size float32` should be renamed to `fontSize float32` and add named returns here too.

andydotxyz:
Sorry I keep coming to this and fixing a naming then we find some more.
Can you please let me know when the review is complete and I will fix all the items found?

fpabl0:
Sorry, it is quite difficult to navigate through 537 files. I thought that when FontMetrics rename was accepted, fontSizeCache would be renamed as it seems to be the more logical change. Same in `RenderedTextSize` as `fontSize` parameter was selected in cache functions it should use the same name, but that is only for clarity, other than that everything looks good :).

internal/cache/text.go
andydotxyz: I don't think so - the map is from font size metadata to font metric cache. The one you comment on here is on the left of that equation

fpabl0: Oh I see, got it.

andydotxyz:
Not a problem, the font painter names fixed but the cache keys are size/style not metric so I think that one is OK.

fpabl0 (approved):
LGTM! :) Sorry any inconveniences and delays.

andydotxyz:
> Sorry any inconveniences and delays

Not at all, thanks for your help getting this over the line.

PR #2336, opened 2021-07-22
Add Alignment to Select

### LINKED ISSUE #2329: Feature Request: widget.NewSelect with TextAlignment for options

There's no way to set the text alignment for options for a `widget.Select` similar to a label. It seems to me centering the options would be a common use case.

I'm not sure. I might be but if only with considerable effort? (for something which should trivial)

[widget.Select](https://developer.fyne.io/api/v2.0/widget/select.html) with a `Alignment fyne.TextAlign` field similar to [widget.Label](https://developer.fyne.io/api/v2.0/widget/label.html).

(feature request as recommended on [twitter](https://twitter.com/Fyne_io/status/1412052734791008257))

andydotxyz:
Need to pass it all the way down to menu item. Do so internally for now,
as we don't need to expose alignment of menu items publicly.

Fixes #2329

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/popup_menu.go
```diff
@@ -67,6 +67,15 @@ func (p *PopUpMenu) Move(pos fyne.Position) {
 	p.Base.Move(p.adjustedPosition(pos, p.Size()))
 }
 
+// Refresh ensures the menu updates to reflect data changes.
+//
+// Implements: fyne.Widget
+func (p *PopUpMenu) Refresh() {
+	p.updateItems()
```
fpabl0, line 74: I think this is unnecessary as p.Menu.Refresh() will already update their items?

andydotxyz: Ah yes good point. Not quite that simple, but duplication nonetheless. Fix inbound.

fpabl0 (dismissed):
LGTM :)

andydotxyz:
Sorry @fpabl0 - merge conflicts had to be resolved.

internal/driver/glfw/window_test.go
```diff
@@ -1370,7 +1370,7 @@ func TestWindow_SetPadded(t *testing.T) {
 	if hasNativeMenu() {
 		menuHeight = 0
 	} else {
-		menuHeight = widget.NewToolbar(widget.NewToolbarAction(theme.ContentCutIcon(), func() {})).MinSize().Height
```
fpabl0, line 1373: if the canvas text is empty, will it return a height different from 0?

fpabl0 (approved):
Everything looks good :), just one question mostly for curiosity.

internal/driver/glfw/window_test.go
andydotxyz: Yes, the height of text is defined by its parameters - the width is what changes based on content

PR #2338, opened 2021-07-22
Disable form labels for disabled widgets

### LINKED ISSUE #1530: Disable `Form` labels when the element it applys to is disabled

When creating forms (which currently, I can't use the `Form` type properly, covered in another issue); I would like to be able to "Disable" (grey out, just like with `Entry` fields) Labels corresponding to values which can no longer be modified.

#### Example:
image
(Notice how the `Label` visually seems very out of place when its corresponding `Entry` is disabled?)

Yes

Make `Label` `Disablable` like `Check` and `Entry` already are.
The text would go grey, just like `Check` and `Entry` already do.

andydotxyz:
In 2.1 we could return to using Label as RichText would allow colours.
But for now using Text works :).
In the issue they suggest adding Lable.Disable as well, but this would require a 2.1 release and it seemed like we could get this fix in earlier.

Fixes #1530

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (changes requested):
Nice work! Consider this partially approved.
Let's just get `fyne_demo` updated before landing it.

Jacalz (approved):

PR #2339, opened 2021-07-24
Populate the file dialog with more favorites

Jacalz:
### Description:

I have felt for quite some time that the file dialog has been missing a few very useful favorite locations, but I didn't want to add them because it would take up more space in the list. Now with the rework of the interface, I finally think it is worth it to add music, pictures andvideos as favorites. 
image

This is marked as draft for the time being as I want to test it on Windows and macOS and possibly add a few tests as well before marking as ready for review. Feel free to look at it if you want through.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -704,6 +704,9 @@ func getFavoriteIcons() map[string]fyne.Resource {
 		"Home":      theme.HomeIcon(),
 		"Documents": theme.DocumentIcon(),
 		"Downloads": theme.DownloadIcon(),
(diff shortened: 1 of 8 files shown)


andydotxyz:
Great idea, thanks. On macOS we use "Movies" instead of "Videos", and I agree we should get that plus Windows support in too.

andydotxyz (approved):
Nice :)

PR #2345, opened 2021-08-01
Delivery mouse events by interface

zdima:
### Description:
This change will allow more precision on delivering a mouse event to objects implementing such event  interfaces.
The existing implementation picks the topmost object and try to apply the mouse event to it. The event will be lost when object doesn't implement the interface. For example in List, when a transparent object added to canvas object returned by create function to intercept right-click or double click, the single click on such item will not select the list item as event not delivered to listItem.

### Checklist:
- [X] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -817,7 +817,7 @@ func (w *window) mouseClicked(_ *glfw.Window, btn glfw.MouseButton, action glfw.
 	}
 
 	// when click outside currently Focusable object, remove focus.
-	if w.canvas.Focused() != nil && focusableO.(fyne.Focusable) != w.canvas.Focused() {
+	if focusableO, ok := focusableO.(fyne.Focusable); ok && w.canvas.Focused() != nil && focusableO != w.canvas.Focused() {
 		w.canvas.Unfocus()
 	}
 
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -607,6 +607,9 @@ func (w *window) mouseMoved(viewport *glfw.Window, xpos float64, ypos float64) {
 		if cursorable, ok := object.(desktop.Cursorable); ok {
 			cursor = cursorable.Cursor()
 		}
+		if _, ok := object.(fyne.Draggable); ok {
+			return true
+		}
 
 		_, hover := object.(desktop.Hoverable)
 		return hover
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -71,19 +71,20 @@ type window struct {
 	centered   bool
 	visible    bool
 
-	mouseLock            sync.RWMutex
-	mousePos             fyne.Position
-	mouseDragged         fyne.Draggable
-	mouseDraggedObjStart fyne.Position
-	mouseDraggedOffset   fyne.Position
-	mouseDragPos         fyne.Position
-	mouseDragStarted     bool
-	mouseButton          desktop.MouseButton
-	mouseOver            desktop.Hoverable
-	mouseLastClick       fyne.CanvasObject
-	mousePressed         fyne.CanvasObject
-	mouseClickCount      int
-	mouseCancelFunc      context.CancelFunc
+	mouseLock             sync.RWMutex
+	mousePos              fyne.Position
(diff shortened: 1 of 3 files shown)


Jacalz (commented):
Hmm. Sorry, but I'm not sure that I quite see the benefit of this. The code seems to be harder to read and now has a lot of type assertions.

zdima:
The benefit is that since most of the Fyne's **complex** widgets don't provide interfaces to handle mouse clicks (hm, who [was against the one that give popup callback](https://github.com/fyne-io/fyne/pull/2312)? :)  ), double clicks, in order to create perfect interface end-user will like to work with. The parent widgets like List, Tree and Table start loosing expected behavior when you add components with mouse interfaces.
For developers using Fyne it is become a process of coping the entire widget code and reimplement. Plus, since those widget relay on many internal structures, that become less and less attractive to work with.
This change at least give Fyne a fighting chance to give developers tools to receive mouse events without loosing paren't widget functionality and create fine applications with Fyne.

Jacalz:
> (hm, who [was against the one that give popup callback](https://github.com/fyne-io/fyne/pull/2312)? :)

That API just simply didn't make sense in my option. I don't have anything against your work. I'm just unsure about the implementation here and what it's trying to achieve ;)

andydotxyz:
I think there is an area of improvement here, but the issue summary doesn’t seem quite right. The code is designed to find the top most widget that implements the correct interface for the action.

I had a plan for improving this code that I haven’t been able to get to. We discussed on a previous call briefly that we could lookup the widgets once at the position and find the topmost for all mouse interface matches in one go to save multiple checks and avoid lots of variables set (I.e. a single map instead).

I think you are working in this area which is great - but from reading the ticket I’m not sure if you’re trying the same improvement?

zdima:
I have started with map as we discussed, but changed to variables for number of reasons.
The code become harder to read with all map access, the map will take cpu to search for interface.
I have tried to switch to array, but get the same cluttered feeling.
With variables you can easy check for nil and notation to access is shorter. The access to variable is faster than map and an array (no extra offset operation).
Not sure what you mean and how by "the topmost for all mouse interface matches". I think the method of  findObjectAtPositionMatching call in the PR  does exact this by assigning the topmost object of the interface to a variable.

andydotxyz:
> Not sure what you mean and how by "the topmost for all mouse interface matches". I think the method of findObjectAtPositionMatching call in the PR does exact this

Apologies I was not clear, you are quite right that what you are doing within the method does this.
I had considered doing this once for the current position and caching the result - so a mouse down then up does not require two tree walks if the mouse does not move. This is where map could help as it's a single variable to cache the results. But it was just a thought I never looked into the code.

p.s. Don't add "O" to the end of variables to hint that it is an object type - we have a fully typed language so don't need these hints.

zdima:
I see your point now. It could eliminate the tree-walk when position is the same.
This could help on desktop since the mouse can be controlled with more precision. I can look into this
This won't help much on mobile and tablets with finger control.
What driver is used on chromebook device? Is it mobile or desktop?

andydotxyz:
> This won't help much on mobile and tablets with finger control.

This is true, but if we can do it in some shared code then it doesn't matter that mobile doesn't get the benefits, we benefit from just 1 code path.

> What driver is used on chromebook device? Is it mobile or desktop?

I believe that you would be installing an Android app for chromebook, so it would be the mobile driver.

Jacalz:
> > What driver is used on chromebook device? Is it mobile or desktop?
> 
> 
> 
> I believe that you would be installing an Android app for chromebook, so it would be the mobile driver.

The latest releases have support for running desktop Linux apps in a container. I think it theoretically could use both :)

zdima:
Hi @Jacalz, was busy and didn't had time to look at it.
Can't say when I can deliver this. Looks like I would have to fix merge conflict, and more time will needed to redesign for additional requirement that might or might not improve performance.

andydotxyz:
Overall it seems like the cached version may be slightly simpler.
However I would not use a list of consts and a map - a single struct would hold the objects by name much more cleanly.

The one thing that worried me about the diffs was that mouse tests had been changed. That should not be done - they are the definition of how it should function. Any change required would have to be internal details of storage and I think they are written to not be sensitive to such changes.

PR #2354, opened 2021-08-05
Add support for finding directory MimeTypes

Jacalz:
### Description:

This updates `uri.MimeType()` to try and get a MimeType for folders. This can be done by trying to read the URI and if that works with nothing returned, we can state it as a directory by returning the "inode/directory" MimeType that is part of the freedesktop specification.

It also updates `widget.FileIcon` to be able to take advantage of this and thus trying to avoid calling `uri.CanList()` in cases where we find the directory using mimes. It also updates the slow directory check to only run if we can't find a suitable MimeType, thus making known types use the fast path.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/storage/repository/uri.go
+++ b/storage/repository/uri.go
@@ -48,8 +48,11 @@ func (u *uri) MimeType() string {
 		if err == nil {
 			defer readCloser.Close()
 			scanner := bufio.NewScanner(readCloser)
-			if scanner.Scan() && !utf8.Valid(scanner.Bytes()) {
+			bytes := scanner.Bytes()
+			if scanner.Scan() && !utf8.Valid(bytes) {
 				mimeTypeFull = "application/octet-stream"
+			} else if bytes == nil { // A readable uri without content is a directory.
+				mimeTypeFull = "inode/directory"
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


andydotxyz:
I'm not sure about tying to the freedesktop spec. The const "application/x-directory" was chosen earlier as it seemed generally applicable and more readable.

andydotxyz:
It would also be interesting to see what is faster, on average, `CanList` vs `MimeType` - both have some slow corner cases.

Jacalz:
> It would also be interesting to see what is faster, on average, `CanList` vs `MimeType` - both have some slow corner cases.

Benchmarking may or may not be necessary yet (if it works) as my current solution doesn't use any extra computation. All the directories were running through the https://github.com/fyne-io/fyne/blob/751cf951f8f5cbf398f9998569b212522314d029/storage/repository/uri.go#L42-L54 entire code already, at least on my computer.

Jacalz:
Tests seem to be failing but I will dig deeper to see if we can make them pass. Works great in fyne_demo when testing things locally. I think the tests relying on local lookup from the mimetypes in Go (for files that do not exist) might be causing the test failures.

Jacalz:
This looks to have been a poor solution to solving the mime entries for folders. Closing this.

PR #2358, opened 2021-08-07
Fix issue fileicon not updating if uri mimetype changes

Jacalz:
### Description:

The caching caused this issue because it compared the new uri with the cached uri. However, several parts of the URI are methods and not part of the struct, thus the comparison can be false when it really should be true. The fix here was unfortunatelly to remove the caching, but I have added a small change to make sure that we can avoid calling `storage.CanList` multiple times on the same uri.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/fileicon.go
+++ b/widget/fileicon.go
@@ -119,11 +119,13 @@ func (i *FileIcon) isDir(uri fyne.URI) bool {
 		return true
 	}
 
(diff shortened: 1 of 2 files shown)


andydotxyz (approved):
Good catch, and a clever optimisation thanks :)

PR #2368, opened 2021-08-12
Add in a missing break in for-loop

Jacalz:
### Description:

This adds in a missing break in the mapbinding code. This should improve the performance in some cases as we don't need to check all of them.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
nice catch

PR #2415, opened 2021-08-30
Improve error message while packaging mobile

changkun:
(sorry for opening this once more, accidentally deleted the previous corresponding branch, see previous discussion #2403, code review comments are now fixed.)

### Description:

While packaging a demo application, the command terminates with an exit code 1 without further indicators:

```
$ fyne package -os ios -appID com.example.app -icon app.png
exit status 1
```

The `fyne package` command does not further process flags such as `-x`, `-work`, etc. This makes UX a little bit confusing for users, and I have to clone the repo and enable these flags (in cmd/fyne/internal/mobile), then rebuild the fyne command to identify where is the error.

After some debugging, the core issue is that the `exec.Command` does not report execution results from stdout and stderr. This change improves this situation a little bit and outputs error messages as follows:

```
$ fyne package -os ios -appID com.example.app -icon app.png
/* com.apple.actool.errors */
/Users/changkun/dev/wild/fyne-ex/Images.xcassets: error: The stickers icon set or app icon set named "AppIcon" did not have any applicable content.
/* com.apple.actool.compilation-results */
/dev/null

: exit status 1
```

For the `execabs` usage, we are packaging iOS platform, it requires a macOS, and does not make sense to use `execabs` for Windows security. Hence, use `os/exec` instead.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/package-mobile.go
+++ b/cmd/fyne/internal/commands/package-mobile.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"bytes"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


andydotxyz (approved):
Yeah, looks good thanks.

PR #2420, opened 2021-08-31
Speed up scrolling when not precision

### LINKED ISSUE #775: widget.ScrollContainer scroll speed is slow
I used widget.ScrollContainer.
But scroll speed is slow.
Is there an option to increase the amount of movement on the screen for a single mouse wheel movement.
For example, if you specify 2 as an option, I want the movement on the screen to double.

andydotxyz:
a simple approach to try and adjust for expectatios for faster-than-hardware scrolling.
Improves #775

### Checklist:

- [ ] Tests included. <- glfw input tests is still a black hole :(
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"image"
 	_ "image/png" // for the icon
+	"math"
 	"runtime"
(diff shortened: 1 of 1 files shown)


andydotxyz:
Review time lapsed

PR #2425, opened 2021-09-01
Support graphics switching on macOS

### LINKED ISSUE #2423: Graphics (GPU) switching on macOS
On a MacBook Pro Fyne uses the discrete rather than integrated GPU. This can lead to higher battery usage. Ideally, Fine would by default utilize the automatic switching by adding a hint to the GLFW window.

As is, when running a simple Fyne application I see usage of the discrete GPU, AMD Radeon Pro (as observed in gfxCardStatus and Activity Monitor GPU utilization). Adding `glfw.WindowHint(glfw.CocoaGraphicsSwitching, 1)` to `glfw_core.go` and running the same Fyne application shows usage of the integrated GPU, Intel HD Graphics.

Documentation for the hint is at https://www.glfw.org/docs/latest/window_guide.html 

```
GLFW_COCOA_GRAPHICS_SWITCHING specifies whether to in Automatic Graphics Switching, i.e. to allow the system to choose the integrated GPU for the OpenGL context and move it between GPUs if necessary or whether to force it to always run on the discrete GPU. This only affects systems with both integrated and discrete GPUs. Possible values are GLFW_TRUE and GLFW_FALSE. This is ignored on other platforms.

Simpler programs and tools may want to enable this to save power, while games and other applications performing advanced rendering will want to leave it disabled.

A bundled application that wishes to participate in Automatic Graphics Switching should also declare this in its Info.plist by setting the NSSupportsAutomaticGraphicsSwitching key to true.
```

I believe that providing the hint is a better default as most applications will likely not need the performance of the discrete GPU and be better off saving power. It should also be possible to optionally (though I haven't considered the API) choose to use the discrete GPU (not providing the hint).

andydotxyz:
Fixes #2423

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


internal/driver/glfw/window.go
```diff
@@ -1399,6 +1399,7 @@ func (w *window) create() {
 		} else {
 			glfw.WindowHint(glfw.Resizable, 1)
 		}
+		glfw.WindowHint(glfw.CocoaGraphicsSwitching, glfw.True)
```
Jacalz, line 1402: Wouldn't it make sense to move this into the `initWindowHints()` function?

Jacalz (approved):
I can't test this given that I have no Mac with dual graphics, but it looks sensible :)

PR #2428, opened 2021-09-03
Remove dependency fyne-io/mobile (prototype)

changkun:
### Description:

fyne uses a forked version of gomobile and had branched out for a quite long time. The initial purpose seems to give credits to the Go's mobile project, but the forked version had moved apart from the original golang/mobile project, which is already quite different from the latest gomobile release.

Moreover, fyne-io/mobile supports the mobile platform which is an essential driver, if there is an issue in the driver, manage two different dependencies is actually painful for the debugging process.

Besides, fyne is using a vendor folder to cache all dependent files. Moving fyne-io/mobile into internal/driver/mobile seems to be no problem at all. 

Hence, this PR merges fyne-io/mobile into the fyne project for the greater good, so that we will have a cleaner dependency tree to thrive the development process of fyne.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


app/app_mobile.go
```diff
@@ -1,22 +1,22 @@
+//go:build !ci && (android || ios)
 // +build !ci
-
-// +build android ios mobile
```
andydotxyz, line 3: Why have you removed the mobile build flag? it is super helpful for testing mobile driver on a desktop computer.

cmd/fyne/internal/mobile/env.go
```diff
@@ -326,6 +326,12 @@ func archNDK() string {
 		arch = "x86"
 	case "amd64":
 		arch = "x86_64"
+	case "arm64":
```
andydotxyz, line 329: What is this change for? it looks like it could break iOS build prior to Go 1.16

internal/driver/mobile/app/app.go
```diff
@@ -1,16 +1,13 @@
-// Copyright 2014 The Go Authors. All rights reserved.
```
andydotxyz, line 1: I don't think it's OK to just strip copyright... perhaps add us to it rather than removing all?

internal/driver/mobile/app/darwin_ios.go
```diff
@@ -224,14 +211,15 @@ func drawloop() {
...
+	workAvailable := theApp.worker.WorkAvailable()
+	for {
 		select {
 		case <-workAvailable:
 			theApp.worker.DoWork()
 		case <-theApp.publish:
 			theApp.publishResult <- PublishResult{}
 			return
-		case <-time.After(100 * time.Millisecond): // incase the method blocked!!
+		case <-time.After(time.Millisecond): // incase the method blocked!!
```
andydotxyz, line 222: Why moving to wait only a millisecond? Won't this mean we drop paint frames as they often take >= 10ms

internal/driver/mobile/mobileinit/mobileinit_ios.m
```diff
@@ -1,5 +1,5 @@
-// +build darwin
-// +build arm arm64
+//go:build ios
```
andydotxyz, line 1: This is using newer Go build tags - not supported all the way back to 1.14

andydotxyz (changes requested):
This is a good idea - but the PR contains too many unrelated changes.
There are code alterations, changes to the build process and other things that are hard to review in the middle of a refactor.

Please change so that the large refactor is done with no code changes so it can be accepted and then moved forward after.
Also I am pretty sure it is not OK to remove copyright notices from peoples work, maybe we can add our own notice instead?

app/app_mobile.go
changkun: This is a little bit tricky here. Testing a mobile driver on a desktop that relying on a different implementation is actually less useful. On desktop, we are currently depending on OpenGL 3.2 whereas mobile depends on ES. The application lifetime also manages differently. I agree that the current facility of testing with a mobile compatible size, but most of the runtime issues actually never detected. Hence I removed it. It would be more useful if we test directly on a mobile simulator.

cmd/fyne/internal/mobile/env.go
changkun: This is for M1 macs, see https://golang.org/cl/346153.

internal/driver/mobile/app/app.go
changkun: Removing the header is to make it clear that it has largely deviated from the original distribution. Adding a README in this folder is optimal to keep the attribution so that we don't violate the BSD license. Such as: https://github.com/fyne-io/fyne/blob/master/cmd/fyne/internal/mobile/README.txt

internal/driver/mobile/app/darwin_ios.go
changkun: On iOS it is less responsive when interacting with the UI. Mainly because of this block. If there are no draw events, this function returns after 100 ms and causes laggy experiences.

Reducing the number here does not affect drawing events if `workAvailable` is selected and starts execution, then the function will only return if `worker.DoWork()` is returned.

internal/driver/mobile/mobileinit/mobileinit_ios.m
changkun: Hmm. Good point, my editor did auto-fixes for these tags. Is there a particular reason we have to support 1.14 builds? Especially Go 1.14 is not officially supported by the Go team anymore.

Furthermore, 1.14 and 1.16 have a significant difference that 1.14 ios are building via // +build darwin/arm64, whereas 1.16 builds ios via // +build ios.

changkun:
Thanks for the review! Yes, you are right that this PR contains too many changes that are not ideally separated. I should write the purpose of the PR more clear in the description that it is a proof of concept and for feedbacks.

The purpose of this PR is basically a demonstration of how many changes we might need in order to merge fyne-io/mobile and discuss a feasible plan that can make a consensus to proceed in many separate PRs.

As far as I identified some major changes and steps to the PR:

1. Build tags migration
2. Merge fyne-io/mobile to the repository without any further changes (except import path)
3. Modify the code and further remove necessary dependencies
4. Further remove desktop relevant code from the mobile driver
5. A new testing flow for mobile tests on emulators/simulators

app/app_mobile.go
andydotxyz: This was useful to test apps using the mobile driver. It reported as `true` if you `fyne.CurrentDevice().IsMobile()` and other features for example.
The render pipeline may indeed be GL instead of GLES - but graphics is not set for mobile devices specifically - painting is separate from the driver anyway.
If in the future you think we should remove functionality, please don't bury the change in a refactor PR!

internal/driver/mobile/app/app.go
andydotxyz: But copyright is not the same as licensing - when we fork the code we did not change it's origination or revoke the copyright of it's original holders...

internal/driver/mobile/app/darwin_ios.go
andydotxyz: Has this been verified on old iOS devices? - foreground and background?
The change it was added was titled "Fix background mobile hangs"...

internal/driver/mobile/mobileinit/mobileinit_ios.m
andydotxyz: Yes, we have to support Go 1.14. It is the newest version available on some OS and the last to support building iOS 32-bit

andydotxyz: The old format of build flags is still supported by Go 1.17 so there is no need to move away from the old type.

andydotxyz:
2. and 3. sound great, and fit within the topic if this PR IMHO.
I don't think we should do 4. as the mobile testing on develop is helpful and **super** fast compared to simulators (and so does not need 5).
As noted in other PRs we have to hold off on 1 because we support Go 1.14 and will for some time.

internal/driver/mobile/mobileinit/mobileinit_ios.m
Jacalz: The formatting in `gofmt` should (by default with older versions in go.mod, I think) add both without breaking for older Go versions. I think it could be a stop gap, but might be better to do that all in a separate PR to not add complexity here.

changkun: I can send a prototype to show a global change on the build tags :)

internal/driver/mobile/app/darwin_ios.go
changkun: Yes, the previous change was about restoring the correct OpenGL context. It is quite mysterious why there is a 100 ms for a timeout. I think the block should be restored to 1 ms approx. equal to 1s/960, in this way, the user input events can be handled timely.

changkun: But sadly I don't have an old iOS device...

Jacalz:
Should this be closed given that other PRs?

changkun:
Yes, let's close the PR as the other PRs already replaces this one.

PR #2431, opened 2021-09-04
Use Go 1.17 for static analysis

Jacalz:
### Description:

Also upgrades staticcheck to v0.2.1 for Go 1.17 support and some false positive fixes.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -8,7 +8,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: WillAbides/setup-go-faster@v1.6.0
       with:
-        go-version: '1.16.x'     
+        go-version: '1.17.x'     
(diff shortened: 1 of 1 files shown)


Jacalz:
Hmm, this requires that the code is formatted using Go 1.17 to add the new `go:build` stuff. I don't currently have access to Go 1.17 on my machine :I

andydotxyz:
> Hmm, this requires that the code is formatted using Go 1.17 to add the new `go:build` stuff. I don't currently have access to Go 1.17 on my machine :I

Does this mean we should hold it back, or is the need for 1.17 just temporary for a first pass?

Jacalz:
I think we can wait. This isn’t gaining us a whole lot at the moment, I don’t think.

PR #2436, opened 2021-09-07
internal/driver: move fyne-io/mobile to internal/driver/mobile

changkun:
### Description:

This is the 2nd pull request of a series of pull requests intended to remove the fyne-io/mobile dependency, as per discussed in #2428.

In this change, it moves the current vendored gomobile's fork fyne-io/mobile to the `internal/driver/mobile`.
To prevent build failure, it also fixes the relevant import paths in other dependent packages.

There are a few non-trivial path fixes, and here is a list of them:

```
app/app_mobile_and.go
cmd/fyne/internal/mobile/build_androidapp.go
cmd/fyne/internal/mobile/build.go
cmd/fyne/internal/mobile/binres/genarsc.go
cmd/fyne/internal/mobile/gendex/gendex.go
```

To get the licensing correct, it follows the style of `cmd/fyne/internal/mobile` that adds a README.txt file in the `internal/driver/mobile` folder.

Furthermore, to guarantee static analysis to pass, there are a few non-trivial changes:

1. Vet: Remove GetString() from internal/driver/mobile/gl because it is a clear violation of unsafe.Pointer/uintptr conversion rules, and it is actually not used at all.  See c48cbc1.
2. Gocyclo: The original mobile package use switch statement to assert input conditions, which does not meet the fyne's code complexity checker. Use map structure instead. See 88b0b21.
3. Golint: Code style fixes so that the linter will not complain about the code. See 2de3b44, 29ce593 (by deleting a package).
4. Golint: Removes an unused package `f32` from the original fyne-io/mobile, because "A little copy is better than dependency" -- Rob Pike, Go Proverbs. See 29ce593. 
5. Golint: The original naming style from fyne-io/mobile does not meet camel case style in fyne, rename most of the constants and remove all unused constants from the gl package, see 38e4696.
6. Static check: Do not use 0 != f() Yoda style. See 7c3c526.

@andydotxyz @Jacalz 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


Jacalz:
Looks like the cyclomatic complexity is a bit high, which causes the static analysis to fail.

andydotxyz:
Thanks for this, and for being draft. 
I think merging this in before v2.1 is high risk (as we are in final bug fixing).
Let's get it ready to land once the release has passed.

andydotxyz:
Thanks for this, I see how this is beneficial for our flow.
To get the tests passing I think some docs and capitalisation need to be fixed too - standards are higher in this repo than the previous :)

changkun:
Tests are happy now :) -- and I believe this is a minimum change to move the directory to the fyne main repository.

The change literally does not do any logic changes, but simply fixes the code styles, naming, ect.

cmd/fyne/internal/mobile/gendex/gendex.go
```diff
@@ -59,12 +59,12 @@ func gendex() error {
 	if err := os.MkdirAll(tmpdir+"/work/org/golang/app", 0775); err != nil {
 		return err
 	}
-	javaFiles, err := filepath.Glob("../../../../vendor/github.com/fyne-io/mobile/app/*.java")
+	javaFiles, err := filepath.Glob("../../../../../internal/driver/mobile/app/*.java")
```
andydotxyz, line 62: Changing the path traversal depth does not seem right (internal is next to vendor) - are you sure?

andydotxyz (commented):
Thanks. Just one question about path changes.

andydotxyz:
> I think merging this in before v2.1 is high risk (as we are in final bug fixing).

Having said this I think that we may have to merge it now. It causes a change in how the `fyne package` command works and so should not really be pushed into a smaller release. What do you think @Jacalz ?

cmd/fyne/internal/mobile/gendex/gendex.go
andydotxyz: The generator is run manually, from the directory above.
This is required as that is where the generated file resides - it was probably put there to avoid a `main` package inside the `app` package.
The path traversal was correct, or at least it functions correctly before this change.

changkun: Ah, you are right. Sorry for my naive thoughts. Indeed the `go:generate` is in a parent folder. I pushed a new commit which should fix the issue here. Thank you for catching this :)

Jacalz:
> > I think merging this in before v2.1 is high risk (as we are in final bug fixing).
> 
> 
> 
> Having said this I think that we may have to merge it now. It causes a change in how the `fyne package` command works and so should not really be pushed into a smaller release. What do you think @Jacalz ?

I agree that it is a bit too much of a high risk. It is a bit unfortunate, but punting to 2.2 seems like the most sensible solution at this point in time.

andydotxyz:
As discussed on Slack we should bring this in prior to 2.1 so the ongoing fixes don't cause maintenance overheads.

andydotxyz (approved):

PR #2441, opened 2021-09-07
all: compatible with go:build

changkun:
### Description:

This should be the last pull request of a series of pull requests intended to remove the fyne-io/mobile dependency, as per discussed in #2428.

This change migrates the build tag so the entire repository is now compatible with Go 1.17, Go 1.18 (dev, yet), and the future.

It is a boring change using `go fmt ./...`, it does not break any behavior as to how it works today, the new `go:build` comments are useless for older versions but recognizable and preferred when Go 1.17 or 1.18 is used.

(Mark it as a draft until it the previous PRs are merged so that diff can display nicely)

@andydotxyz @Jacalz

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
> This change migrates the build tag so the entire repository is now compatible with Go 1.17, Go 1.18 (dev, yet), and the future.

I think this text is a little misleading - there is no incompatibility with Go 1.17 - or any public released versions.
This change is not required until Go drops support for the old build tag format.

changkun:
> I think this text is a little misleading

Good point. I just rephreased the description.

Jacalz:
One more point from my side. Given that we aren't fixing anything here, isn't it just more work for developers without Go 1.17+ to maintain both build constraints at the same time?

changkun:
@Jacalz I think there should be very limited or even no more work for maintaining two styles. Note that the two versions are identical except for the grammar. They can be automatically translated. The benefit is that //go:build is more readable using logical operations. 

For developers who have Go <=1.16 can keep just writing // +build tag. The `//go:build` tag will be automatically appended when the file is saved (if editor configures, presumably `gopls`?)

Jacalz:
> @Jacalz I think there should be very limited or even no more work for maintaining two styles. Note that the two versions are identical except for the grammar. They can be automatically translated. The benefit is that //go:build is more readable using logical operations. 
> 
> 
> 
> For developers who have Go <=1.16 can keep just writing // +build tag (and vise versa). The `//go:build` tag (or // +build) will be automatically appended when the file is saved (if editor configures, presumably `gopls`?)

The problem is for those that have Go 1.16 and earlier. The maintainer ship for them is higher because the tags won't be automatically translated and not everyone uses gopls.

changkun:
> The problem is for those that have Go 1.16 and earlier. 

Hmm. I am not entirely sure how difficult for them to switch to a newer version of Go. It is entirely common for them to switch Go versions during the development phase. 

Since I am also not yet familiar with the porting policy defined in fyne. Especially, when covering such a huge spectrum of Go versions. When Go 1.18 is out if fyne is targeting on 1.14+, then //go:build tag must be added otherwise the codebase will not work with Go 1.18 at all (it yet still works for go1.18-dev because the breaking change hasn't merged into the master, but scheduled to happen either in Sep or Oct.)

Thereby, it may have some maintenance cost, but won't be higher than either: 
1) locally switch to a different go version then run `go fmt ./...`, everything is fixed without further thinking;
2) set up a regular go dev environment that does the translation on save.

Jacalz:
Ah, I weren't aware that it was going to break in Go 1.18. Ignore my previous comments then.

andydotxyz:
We have never mandated that people run the latest version of Go for development.
In v2.1 we update from Go 1.12 to 1.14 as the minimum - we should assume that some developers are using all versions we support.
Any other policy makes the barrier to contributions higher.

Jacalz:
Unfortunately, a lot of developers are stuck on older versions wether they like it or not. One example is stable  Linux distro a such as Debian, where there is a two year period between major new releases. The previous stable Debian release had Go 1.11 and the one released this summer has Go 1.15. While not as extreme, I'm personally stuck on Go 1.16 for a month or two until Fedora 35 is released.

The second issue is mostly around macOS support. The cycle with dropping support one version every other Go release is, in my opinion, slightly faster than the rate at which those macOS computers drop out of use. We have to support a few Go releases back in order for users to be able to run the applications. My macOS computer, an my parents' computer, is running 10.13 which means that Go 1.19 will drop support for it but those are still very capable 4 core i5 so I'll most likely want to stick to 1.18 a while in that case (only when developing for the Mac though).

I know that this may be a bit unfortunate for cases, but there is not a whole lot we can do about it. We try to stick to the release that is in the latest stable Debian release. Which would mean Go 1.15 but we are on 1.14 due to macOS support.

changkun:
> Unfortunately, a lot of developers are stuck on older versions wether they like it or not. One example is stable Linux distros such as Debian, where there is a two year period between major new releases. The previous stable Debian release had Go 1.11 and the one released this summer has Go 1.15. While not as extreme, I'm personally stuck on Go 1.16 for a month or two until Fedora 35 is released.

I see. Now I understand a bit why they are using an old version of Go. Then the key issue is about how to use the latest distribution. From experices, Go offers nice distribution where everyone should be able to just download the release folder from golang.org, configure the bin path, then everything should be good to go. Relying on a system pkg distribution manager to update their versions seems to be a different story.

> The second issue is mostly around macOS support. The cycle with dropping support one version every other Go release is, in my opinion, slightly faster than the rate at which those macOS computers drop out of use. We have to support a few Go releases back in order for users to be able to run the applications. My macOS computer, and my parents' computer, is running 10.13 which means that Go 1.19 will drop support for it but those are still very capable 4 core i5 so I'll most likely want to stick to 1.18 a while in that case (only when developing for the Mac though).

This is a good example, however, taking a different approach could also workaround. The current practice for darwin, an OS might be unmaintained, but still receive security updates. This means a old fyne version may drop active maintains, but can still get (security, minor bug) fixes for users who use older systems.

andydotxyz:
> Then the key issue is about how to use the latest distribution.

We test latest in CI. We do not require developers to use latest.
Irrespective of how easy it may or may not be that is how we currently work.
Your suggestion about fixes for old versions is a nice idea for the future but at this time it is already too much work to develop a new feature release whilst adding bug fixes to the current main release, adding another release branch to support is not going to be possible until we are a much larger community.

andydotxyz:
Didn't we agree that this does not need to be actioned until Go decides the old format will no longer be supported?
If we add this in then maintaining it is extra work for anyone developing with Go < 1.17.

changkun:
Yes, we can keep this open until // +build is not supported. But that time won't be long, as the CL is sent and may be merged at any time, see https://go-review.googlesource.com/c/go/+/344955. Arguably the "extra work" is small. Here is a small shell script command for switching different versions of Go :)

```
function switchgo() {
	version=$1
	if [ -z $version ]; then
		echo "Usage: switchgo [version]"
		return
	fi

	if ! command -v "go$version" > /dev/null 2>&1; then
		echo "Go ${version} doesn't exist, start downloading..."
		go get golang.org/dl/go${version} # or go install golang.org/dl/go${version}@latest
		go${version} download
	fi

	ln -sf "$(command -v "go$version")" "$GOBIN/go"

	echo "Switched to Go ${version}"
}
```

andydotxyz:
It is worrying, don't you think, that the Go team are happy to cause backwards incompatible changes to the metadata of the language when there is a very strong guarantee for the actual content of the language to not break before Go 2?

changkun:
> It is worrying, don't you think, that the Go team are happy to cause backwards incompatible changes to the metadata of the language when there is a very strong guarantee for the actual content of the language to not break before Go 2?

I do. Actually, I had some discussions with others about the build tag as a breaking change. It turns out that build tag change is not part of the Go 1 compatibility promise. They belongs to the toolchain changes. See https://golang.org/doc/go1compat#tools

andydotxyz:
Yeah I figured they might say that. I think we can agree it's a pretty weak argument.

PR #2475, opened 2021-09-16
Add SetIcon method on ToolbarAction

### LINKED ISSUE #2040: widget.ToolbarItem SetIcon
At the moment ToolBarItem does not have a SetIcon function like widget buttons even tho toolbars are buttons under the hood.
So as a workaround to set the icon it's required to store the fyne.resource icon into variable and re-update content accordingly.

Yes

```
toolBarItem = widget.NewToolbarAction(theme.FyneLogo(), func() {})
toolBarItem.SetIcon(theme.QuestionIcon())
```

adzo261:
### Description:
- Adds `SetIcon` method on `ToolbarAction`
- Constructors for ToolbarItems - `ToolbarAction`, `ToolbarSpacer`, and `ToolbarSeparator` now return their concrete type instead of generic `ToolbarItem` interface. This would allow us to add more methods if needed in future to those structs.

Fixes #2040 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### Example

```
toolbarIActionItem = widget.NewToolbarAction(theme.FyneLogo(), func() {})
toolbarIActionItem.SetIcon(theme.QuestionIcon())
```


widget/toolbar_test.go
```diff
@@ -74,6 +74,15 @@ func TestToolbar_ItemPositioning(t *testing.T) {
 	}
 }
 
+func TestToolbar_SetIcon(t *testing.T) {
+	oldIcon := theme.FyneLogo()
+	toolbarItem := NewToolbarAction(oldIcon, func() {})
```
Jacalz, line 79: While not a deal-breaker, I just want to point out that you can just pass `nil` instead of an empty function.

Jacalz (approved):
Looks good to me. Thanks for the PR and your first contribution to Fyne :)

widget/toolbar.go
```diff
@@ -27,8 +27,14 @@ func (t *ToolbarAction) ToolbarObject() fyne.CanvasObject {
 	return button
 }
 
+// SetIcon updates the icon on a ToolbarItem
```
andydotxyz, line 30: As this is a new API we need to have a since line. Please add (including the line to separate):
```go
//
// Since: 2.2
```

The next feature release will probably be 2.2, so we pick this version in anticipation of the next larger release.

andydotxyz (changes requested):
Great work, just one thing missed here, sorry!

andydotxyz (approved):
Great thanks for this. I will merge it shortly, just wrangling some v2.1.1 fixes first which this needs to come after

andydotxyz:
Merged on develop, the first feature addition for Bowmore release :)

PR #2477, opened 2021-09-16
Re-fix #2449

### LINKED ISSUE #2449: sequential writes to preferences does not save to file
### Describe the bug:

Setting a string value with the Preferences API more than once in a row causes it to not write the latter ones to the file. unless a gap is introduced.

### To Reproduce:
here is a sample program
```go
a := app.NewWithID("TESTING")
fmt.Println("Before setting")
fmt.Println(a.Preferences().String("a"))
fmt.Println(a.Preferences().String("b"))
a.Preferences().SetString("a", "aaaaaaaaaaaaa")
//<-time.After(100*time.Millisecond)
a.Preferences().SetString("b", "bbbbbbbbbbbbb")
fmt.Println("After")
fmt.Println(a.Preferences().String("a"))
fmt.Println(a.Preferences().String("b"))
```
take this program and run it twice.
you will see that on the second run only `a` is set while `b` is still nothing. 
now uncomment the line in the middle and run it twice more.
now both `a` and `b` are set.

 - **OS:** Windows
 - **Version:** 10
 - **Go version:** 1.16.7
 - **Fyne version:** 2.0.4

nullst:
### Description:

The issue #2449 is still not fixed by the pull request #2450 for me since concurrent writes are not the only problem. There is another problem:
* `preferences.saveToFile()` sets ignoreChanges to `true` for 100 ms (using resetIgnore()).
* A default change listener added in newPreferences() just ignores all changes made during that time window.
* If no changes in preferences occur after the 100ms window, the ignored changes are not saved to disk since the change listener responsible for that will never be invoked.

Suggested solution implemented in this pull request:
* Add a counter `numIgnoredChanges int` to `preferences`. The access is guarded by the same `.prefLock` as `.ignoreChanges`.
* Let the default change listener count how many times it was invoked while ignoreChanges is true (guarded by `p.prefLock.Lock()` instead of `p.prefLock.RLock()` since we modify the contents).
* Change `resetIgnore()` to reset `p.numIgnoredChanges` as well, and if the number was not zero, call `p.InMemoryPreferences.fireChange()`. This is not possible as `fireChange()` is not an exported method of the `internal` module, so it was made public.

Fixes #2449

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/app/preferences.go
+++ b/app/preferences.go
@@ -14,17 +14,9 @@ import (
 type preferences struct {
 	*internal.InMemoryPreferences
 
-	prefLock sync.RWMutex
-	// Normally, any update of preferences is immediately written to disk,
-	// but the initial operation of loading the preferences file performs many separate
-	// changes. To avoid "load->changes!->write" pattern (rewriting preferences file
-	// after each loading), saving file to disk is disabled
-	// during the loading progress. Access guarded by prefLock.
-	loadingInProgress bool
-	// If an application changes its preferences 1000 times per second, we don't want to
-	// rewrite the preferences file after every update. Instead, a time-out mechanism is
-	// implemented in resetSuspend(), limiting the number of file operations per second.
-	suspendChange       bool
+	prefLock            sync.RWMutex
+	loadingInProgress   bool
+	savedRecently       bool
 	numSuspendedChanges int
 
 	app *fyneApp
@@ -33,11 +25,11 @@ type preferences struct {
 // Declare conformity with Preferences interface
 var _ fyne.Preferences = (*preferences)(nil)
 
-func (p *preferences) resetSuspend() {
+func (p *preferences) resetSavedRecently() {
 	go func() {
 		time.Sleep(time.Millisecond * 100) // writes are not always atomic. 10ms worked, 100 is safer.
 		p.prefLock.Lock()
-		p.suspendChange = false
+		p.savedRecently = false
 		changes := p.numSuspendedChanges
 		p.numSuspendedChanges = 0
 		p.prefLock.Unlock()
@@ -54,9 +46,9 @@ func (p *preferences) save() error {
 
 func (p *preferences) saveToFile(path string) error {
 	p.prefLock.Lock()
-	p.suspendChange = true
+	p.savedRecently = true
 	p.prefLock.Unlock()
-	defer p.resetSuspend()
+	defer p.resetSavedRecently()
 	err := os.MkdirAll(filepath.Dir(path), 0700)
 	if err != nil { // this is not an exists error according to docs
 		return err
@@ -138,8 +130,8 @@ func newPreferences(app *fyneApp) *preferences {
 
 	p.AddChangeListener(func() {
 		p.prefLock.Lock()
-		shouldIgnoreChange := p.suspendChange || p.loadingInProgress
-		if p.suspendChange {
+		shouldIgnoreChange := p.savedRecently || p.loadingInProgress
+		if p.savedRecently && !p.loadingInProgress {
 			p.numSuspendedChanges++
 		}
 		p.prefLock.Unlock()
--- a/app/preferences.go
+++ b/app/preferences.go
@@ -14,9 +14,10 @@ import (
 type preferences struct {
 	*internal.InMemoryPreferences
 
-	prefLock          sync.RWMutex
-	ignoreChange      bool
-	numIgnoredChanges int
+	prefLock            sync.RWMutex
+	loadingInProgress   bool
+	suspendChange       bool
+	numSuspendedChanges int
 
 	app *fyneApp
 }
@@ -24,13 +25,13 @@ type preferences struct {
 // Declare conformity with Preferences interface
 var _ fyne.Preferences = (*preferences)(nil)
 
-func (p *preferences) resetIgnore() {
+func (p *preferences) resetSuspend() {
 	go func() {
 		time.Sleep(time.Millisecond * 100) // writes are not always atomic. 10ms worked, 100 is safer.
 		p.prefLock.Lock()
-		p.ignoreChange = false
-		changes := p.numIgnoredChanges
-		p.numIgnoredChanges = 0
+		p.suspendChange = false
+		changes := p.numSuspendedChanges
+		p.numSuspendedChanges = 0
 		p.prefLock.Unlock()
 
 		if changes > 0 {
@@ -45,9 +46,9 @@ func (p *preferences) save() error {
 
 func (p *preferences) saveToFile(path string) error {
 	p.prefLock.Lock()
-	p.ignoreChange = true
+	p.suspendChange = true
 	p.prefLock.Unlock()
-	defer p.resetIgnore()
+	defer p.resetSuspend()
 	err := os.MkdirAll(filepath.Dir(path), 0700)
 	if err != nil { // this is not an exists error according to docs
 		return err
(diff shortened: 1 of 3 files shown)


Jacalz (changes requested):
Thanks for opening this PR. I just want to give a heads up that the tests seem to be failing.

nullst:
I've fixed the formatting issue in the commit above.

Note that I have absolutely no idea what caused the test `testQueueLazyInit` from `data/binding/queue_test.go` to fail for the original pull request, and exclusively during ubuntu platform tests (not other platforms). My changes are not related to this code. Should we just ignore this as a fluke?

Jacalz:
I don’t know. Will see what the latest rounds of tests here give us. Something seems to be wrong with your two latest commits. They give me 404 when trying to look at them.

nullst:
That's weird, the links from this pull request give 404 for me as well. Not sure what happened and how to fix it. The links in https://github.com/nullst/fyne/commits/nullst-patch-1 work fine. Part of that is due to a mismatch between e-mail address I used for github and e-mail address configured with git locally, but it should be fixed now, sorry. If nothing else works, maybe I should just make a new pull request with the same content?

andydotxyz:
Thanks for working on this, I agree there is still a problem.
However I'm not sure that this is the right fix. The `ignoreChange` is meant to avoid a changedfile->load->fire->write loop, so the additional work here to un-ignore internal changes should not be needed.
I think the root of the problem is the following code, around line 115 of app/preferences.go
```go
		p.prefLock.RLock()
		shouldIgnoreChange := p.ignoreChange
		p.prefLock.RUnlock()
		if shouldIgnoreChange { // callback after loading, no need to save
			return
		}
```

Here you see that we have a problem on load as well - we read the file, it sees the change and can then force a write, we don't want this.
I wondered if a more correct change might be to stop the change from triggering a save during load, then this code would not have an impact in the rest of the app, which I think would resolve the issue.

What do you think @nullst?

andydotxyz (changes requested):
As described on the conversation, I think that this is not the simplest fix - it works around a mistaken use of the `ignoreChange` instead of resolving the root problem.

nullst:
I have implemented the strategy suggested by @andydotxyz above: since loading the preferences file is conceptually a different operation than just an application updating multiple preferences very quickly, I added a new flag, "loadingInProgress" to the `preferences` structure. The existing `.ignoreChanges` flag was renamed to '.suspendChanges' and repurposed to avoid rewriting the preferences file too fast.

app/preferences.go
```diff
@@ -14,21 +14,37 @@ import (
 type preferences struct {
 	*internal.InMemoryPreferences
 
-	prefLock     sync.RWMutex
-	ignoreChange bool
+	prefLock sync.RWMutex
+	// Normally, any update of preferences is immediately written to disk,
```
andydotxyz, line 18: We don't really go for large comments in code, in fact any code comments normally indicate that the code is not clear.
We have version control to learn more about the change if looking up the history is required.
In this case the variable names seem clear, can we remove the comments?

andydotxyz (changes requested):
Iim not sure about `numSuspendedChanges`... This counts how many times we write the file in quick succession.
But after writing the file we should never have to re-read it, that is what `ignoreChanges` was about in the first place.
Is there some reason you think we need to read the file shortly after we have finished writing it?

nullst:
Your comment helped me understand the mechanism better, thank you. Mostly, my original changes were not relevant to loading the preferences file at all, I'm concerned with the situation where the application just updates many different preferences in a row. This is kind of similar to what happens during `load()`, but it should still result in a file save after the last change, and this wasn't working properly due to the measures implemented to avoid triggering save() after load().

In general, as I see it now, there are three situations we want to avoid:

1. "Load causes save": loading preferences from file changes in-memory preferences, and normally those changes should be immediately written back to disk, but in this case this behavior is undesirable.
2. "Save causes load": when an application updates preferences, the preferences file will be modified on disk. But we are also watching the file for changes (in `app/preferences_other.go`), to reload preferences when modified. The "file changed" callback is invoked at some unknown time slightly after saving is finished. So we want to neuter the callback for some time after saving. (And if a file is modified by an external program during that small time period, too bad for the external program -- I don't think there are easy workarounds.)
3. "Save during saving": if another save is triggered (by the application updating preferences again) during the time period where the "file changed" callback is disabled because of a previous change, as implemented currently, the timer responsible for flipping `ignoreChanges` flag that started for the first save could disable the flag very shortly after the second save since they do not coordinate, causing situation (2). To be honest, this is not a big deal: at worst, one spurious full reload of the preferences file will happen if you are unlucky. But we may still want to prevent this scenario.

Based on this description, I've renamed the "ignoreChanges" variable to "savedRecently" since it better reflects the role. I've also removed the comments, as you suggested. Originally the same variable was used to protect against both (1) and (2), and I think it does not lead to the correct logic.

andydotxyz:
Thanks, that is a great summary @nullst.
I agree with your points, but I'm still not sure that the counter is needed. It is read only here:
```go
		if changes > 0 {
			p.InMemoryPreferences.FireChange()
		}
```
inside the callback that resets the ignore status, which is called after we save. So my question is - why do we need to ensure that we re-load the file after we have written it out a number of times?

nullst:
Sorry, I'm not sure I understand the question. `FireChange()` does not reload the file, it saves it to disk. This part is mostly about point (3) of the summary above: suppose an application calls `Preferences().SetString()` two times in a row, let's say with 50ms passed between them:

1. The first time, `savedRecently` (previously called `ignoreChanges`) flag is set, the preferences file is updated on disk, and we run a timer to disable `savedRecently` after 100ms.
2. The second time an update happens while `savedRecently` flag is still on. There are now two approaches. Approach **A** (not implemented): we immediately rewrite the file on disk again and then we want to keep `savedRecently` flag on for another 100ms. To do this correctly, we would need to communicate with the timer started during the first save to extend the time-out. Approach **B** (implemented here): we delay the re-writing of the file on disk until the flag `savedRecently` is flipped by the timer started during the first save, and we remember to call `.save()` after that. This is what `numSuspendedChanges` is for: it is nonzero if we registered that `save()` should be called when the timer expires. It does not have to be a counter, it can just be `bool` if you prefer.

For approach **A** a more complicated synchronization pattern would be required than just firing up a goroutine with a timer in `ignoreChanges`. Approach **B** seems simpler and more robust. Moreover, for free we get an IO-limiting mechanism: even if the application updates 5000 preferences per second, we only overwrite the file 10 times per second, each time storing the current state of the preferences. This limiting mechanism is not particularly valuable, but getting it naturally is probably a nice bonus.

andydotxyz:
Ah, great explanation thanks. Should the callback perhaps call save instead of feeding it back through the FireChanges loop again (I think this is what confused me).
Perhaps a bool may make it a little cleaner too, as we don’t actually care what number it is after 1

nullst:
This makes sense, it's actually easier. Now `resetSavedRecently` calls `.save()` directly and `bool` is used instead of a counter. I also made `FireChange()` an internal function again (i.e., renamed to `fireChange()`).

internal/preferences.go
```diff
@@ -68,6 +68,7 @@ func (p *InMemoryPreferences) remove(key string) {
 	delete(p.values, key)
 }
 
+// FireChange causes InMemoryPreferences to activate all its .changeListeners
```
andydotxyz, line 71: The case mis-match here due to it not being exported any more will (now or in the future?) trip the linter...

andydotxyz (changes requested):
Thanks, much clearer now I think - just a small casing issue in the doc

internal/preferences.go
nullst: After the latest round of changes, this added comment is the only thing changed in this file. I'll remove it to return the file back to the original state (besides, this particular function is simple and does not require an extra comment).

andydotxyz (approved):
Thanks, apologies for the delay on this - was getting a bugfix release out first.

andydotxyz:
I think we need @Jacalz to OK the changes too

PR #2483, opened 2021-09-19
Fix drawing rectangle with stroke only

Jacalz:
### Description:

A typo prevented rectangles with nil fill color to be drawn with a stroke.

### Checklist:

- [ ] Tests included.
- Not quite sure here. The canvas primitives didn't seem to contain many visual tests and the commit that broke this didn't add any tests either.
- [x] Tests all pass.


andydotxyz (approved):
Nice catch

Jacalz:
Could you please backport this to a 2.1.x branch?

andydotxyz:
I don't think we have branched that off yet. I don't normally do that until we see an API change that won't be part of a bugfix release

PR #2486, opened 2021-09-20
internal/driver/mobile: clear touchIDs at every touch end event

### LINKED ISSUE #2407: Runtime panic because out of touchID
### Describe the bug:

While debugging #2406 , a runtime panic crashes the app. Here is a stack trace:

```
panic: out of touchIDs

github.com/fyne-io/mobile/app.sendTouch(0x130127b98, 0x10404a2a0, 0x30127b98, 0x1)
	/Users/changkun/go/pkg/mod/github.com/fyne-io/mobile@v0.1.3-0.20210811181917-d62fb16c28ec/app/darwin_ios.go:182 +0x184
github.com/fyne-io/mobile/app._Cfunc_runApp()
	_cgo_gotypes.go:201 +0x38
github.com/fyne-io/mobile/app.main(0x130114760)
	/Users/changkun/go/pkg/mod/github.com/fyne-io/mobile@v0.1.3-0.20210811181917-d62fb16c28ec/app/darwin_ios.go:75 +0x58
github.com/fyne-io/mobile/app.Main(...)
	/Users/changkun/go/pkg/mod/github.com/fyne-io/mobile@v0.1.3-0.20210811181917-d62fb16c28ec/app/app.go:21
fyne.io/fyne/v2/internal/driver/gomobile.(*mobileDriver).Run(0x13011b450)
	/Users/changkun/dev/changkun.de/fyne/internal/driver/gomobile/driver.go:138 +0x64
fyne.io/fyne/v2/internal/driver/gomobile.(*window).ShowAndRun(0x104329ae0)
	/Users/changkun/dev/changkun.de/fyne/internal/driver/gomobile/window.go:187 +0x3c
main.main()
	/Users/changkun/dev/changkun.de/fyne/iosrend/main.go:46 +0x2a0
```

It looks like sendTouch may happen concurrently. Adding an RWMutex would be a quick fix.

### To Reproduce:

1. run the reproducer in #2406 
2. keep clicking the button
3. if lucky, the problem occurs

 - **OS:** iOS 15
 - **Version:** beta 6
 - **Go version:** gotip based on CL 344969
 - **Fyne version:**  develop 83ca35d98be7cf401579c2f78b53dabd7aea250a

changkun:
### Description:

The pointer addresses of UITouch* are used as touchIDs, but they are only unique during a round of touch begin and touch end events. Only clear one touchID when touch.TypeEnd arrives can cause out of touchID panic. So, clear the entire touchIDs at every touch ends.

Fixes #2407 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/mobile/app/darwin_ios.go
+++ b/internal/driver/mobile/app/darwin_ios.go
@@ -154,12 +154,7 @@ func updateConfig(width, height, orientation int32) {
 //
 // It is widely reported that the iPhone can handle up to 5 simultaneous
 // touch events, while the iPad can handle 11.
-var (
-	// touchIDs may arrive concurrently, use a lock for safety.
-	// See https://github.com/fyne-io/fyne/issues/2407.
-	touchMu  sync.RWMutex
-	touchIDs [11]uintptr
-)
+var touchIDs [11]uintptr
 
 var touchEvents struct {
 	sync.Mutex
@@ -169,17 +164,12 @@ var touchEvents struct {
 //export sendTouch
 func sendTouch(cTouch, cTouchType uintptr, x, y float32) {
 	id := -1
-
-	touchMu.RLock()
 	for i, val := range touchIDs {
 		if val == cTouch {
 			id = i
 			break
 		}
 	}
-	touchMu.RUnlock()
-
-	touchMu.Lock()
 	if id == -1 {
 		for i, val := range touchIDs {
 			if val == 0 {
@@ -195,9 +185,14 @@ func sendTouch(cTouch, cTouchType uintptr, x, y float32) {
 
 	t := touch.Type(cTouchType)
 	if t == touch.TypeEnd {
-		touchIDs[id] = 0
+		// Clear all touchIDs when touch ends. The UITouch pointers are unique
+		// at every multi-touch event. See:
+		// https://github.com/fyne-io/fyne/issues/2407
+		// https://developer.apple.com/documentation/uikit/touches_presses_and_gestures?language=objc
+		for idx := range touchIDs {
+			touchIDs[idx] = 0
+		}
 	}
-	touchMu.Unlock()
 
 	theApp.eventsIn <- touch.Event{
 		X:        x,


internal/driver/mobile/app/darwin_ios.go
```diff
@@ -195,9 +185,14 @@ func sendTouch(cTouch, cTouchType uintptr, x, y float32) {
 
 	t := touch.Type(cTouchType)
 	if t == touch.TypeEnd {
-		touchIDs[id] = 0
+		// Clear all touchIDs when touch ends. The UITouch pointers are unique
+		// at every multi-touch event. See:
+		// https://github.com/fyne-io/fyne/issues/2407
+		// https://developer.apple.com/documentation/uikit/touches_presses_and_gestures?language=objc
+		for idx := range touchIDs {
```
andydotxyz, line 192: Wouldn't it be quicker to just set `touchIDs` to `nil`?

andydotxyz (commented):
I wonder if you're certain it is safe to remove the mutex? It seems from an observer that these could still be called concurrently?

internal/driver/mobile/app/darwin_ios.go
changkun: - This involves compiler optimization (see https://github.com/golang/go/issues/5373).
- Set it to nil will cause memory allocation.

changkun:
Yes, it is

> I wonder if you're certain it is safe to remove the mutex?

Yes, it is certain. the sendTouch is called from the OS's main thread, which does not occur concurrently.

> It seems from an observer that these could still be called concurrently?

The previous observer was wrong, as it did not fix the issue. The current fix suggests the correct fix to the cause of the panic.

andydotxyz:
> Yes, it is certain. the sendTouch is called from the OS's main thread, which does not occur concurrently.

Ok, thanks. In future PR can you separate removal of unneeded code and the bug fix itself in different commits?

andydotxyz (approved):
Great thanks

changkun:
>  In future PR can you separate removal of unneeded code and the bug fix itself in different commits?

No problem 👍

PR #2492, opened 2021-09-20
Update windows.go to fix miss-clicks issues/2484

alexballas:
From the code above I understand that there is some logic to alter the w.mouseDragStarted state. However in the last if, we use the mouseDragStarted value which we only set once at the beginning of the function. As part of this change, I've updated mouseDragStarted back to false to properly reflect the correct state

 

### Description:

Fixes #(2484)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -813,6 +813,7 @@ func (w *window) mouseClicked(_ *glfw.Window, btn glfw.MouseButton, action glfw.
 			w.mouseLock.Lock()
 			w.mouseDragStarted = false
 			w.mouseLock.Unlock()
+			mouseDragStarted = false
 		}
 		if shouldMouseOut {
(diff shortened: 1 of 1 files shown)


Jacalz:
Thank you for the PR. Could you please target the `develop` branch instead? We are using that branch for development, per the contributor guidelines: https://github.com/fyne-io/fyne/blob/master/CONTRIBUTING.md.

alexballas:
> Thank you for the PR. Could you please target the `develop` branch instead? We are using that branch for development, per the contributor guidelines: https://github.com/fyne-io/fyne/blob/master/CONTRIBUTING.md.

Hello!
alexballas changed the base branch from master to develop 1 hour ago

andydotxyz (approved):
Amazing, thanks!

PR #2487, opened 2021-09-20
internal/driver/mobile: support darwin/arm64

changkun:
### Description:

This PR allows applications can use the mobile driver on darwin/arm64.

Before the change:

```
$ go run -tags mobile cmd/fyne_demo/main.go
# fyne.io/fyne/v2/internal/driver/mobile/gl
In file included from internal/driver/mobile/gl/work.go:32:
./work.h:20:10: fatal error: 'OpenGLES/ES2/glext.h' file not found
#include <OpenGLES/ES2/glext.h>
         ^~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

After the change:

```
$ go run -tags mobile cmd/fyne_demo/main.go
```

Screenshot 2021-09-20 at 13 15 19

### Checklist:

- [ ] Tests included.
- [x] Lint and formatter run with no errors.
- [x] Tests all pass.

PR #2495, opened 2021-09-21
cmd/fyne: support -os iossimulator

### LINKED ISSUE #1917: Apps cannot launch in iOS simulator
It seems to relate to invalid codesign signature

changkun:
### Description:

This PR adds `-os iossimulator` to the fyne command, and fixes the iOS simulator issue on darwin/arm64.

Fixes #1917

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/mobile/build.go
+++ b/cmd/fyne/internal/mobile/build.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"os"
 	"regexp"
+	"runtime"
 	"strings"
 
 	"fyne.io/fyne/v2/cmd/fyne/internal/util"
@@ -302,8 +303,8 @@ func addBuildFlags(cmd *command) {
 func addBuildFlagsNVXWork(cmd *command) {
 	cmd.Flag.BoolVar(&buildN, "n", false, "")
 	cmd.Flag.BoolVar(&buildV, "v", false, "")
-	cmd.Flag.BoolVar(&buildX, "x", false, "")
-	cmd.Flag.BoolVar(&buildWork, "work", false, "")
+	cmd.Flag.BoolVar(&buildX, "x", true, "")
+	cmd.Flag.BoolVar(&buildWork, "work", true, "")
 }
 
 func init() {
@@ -423,6 +424,19 @@ func parseBuildTarget(buildTarget string) (os string, archs []string, _ error) {
 	if os == "ios" {
 		targetOS = "darwin"
 	}
+
+	if buildTarget == "iossimulator" {
+		if before116 {
+			// If the build target is iossimulator, and the go distribution also before
+			// 1.16, then we can only build amd64 arch app for simulators. Because
+			// arm64 simulators is only supported after go 1.16.
+			allArchs[os] = []string{"amd64"}
+		} else {
+			// Otherwise, the iossimulator arch is depending on the host arch.
+			allArchs[os] = []string{runtime.GOARCH}
+		}
+	}
+
 	if all {
 		return targetOS, allArchs[os], nil
 	}
--- a/cmd/fyne/internal/mobile/build.go
+++ b/cmd/fyne/internal/mobile/build.go
@@ -15,6 +15,8 @@ import (
 	"regexp"
 	"strings"
 
+	"fyne.io/fyne/v2/cmd/fyne/internal/util"
+
 	"golang.org/x/sys/execabs"
 	"golang.org/x/tools/go/packages"
 )
@@ -372,8 +374,11 @@ func parseBuildTarget(buildTarget string) (os string, archs []string, _ error) {
 	archNames := []string{}
 	for i, p := range strings.Split(buildTarget, ",") {
 		osarch := strings.SplitN(p, "/", 2) // len(osarch) > 0
-		if osarch[0] != "android" && osarch[0] != "ios" {
+		if !util.IsAndroid(osarch[0]) && !util.IsIOS(osarch[0]) {
 			return "", nil, fmt.Errorf(`unsupported os`)
+		}
+		if osarch[0] == "iossimulator" {
+			osarch[0] = "ios"
 		}
 
 		if i == 0 {
--- a/cmd/fyne/internal/mobile/build.go
+++ b/cmd/fyne/internal/mobile/build.go
@@ -303,8 +303,8 @@ func addBuildFlags(cmd *command) {
 func addBuildFlagsNVXWork(cmd *command) {
 	cmd.Flag.BoolVar(&buildN, "n", false, "")
 	cmd.Flag.BoolVar(&buildV, "v", false, "")
-	cmd.Flag.BoolVar(&buildX, "x", true, "")
-	cmd.Flag.BoolVar(&buildWork, "work", true, "")
+	cmd.Flag.BoolVar(&buildX, "x", false, "")
+	cmd.Flag.BoolVar(&buildWork, "work", false, "")
 }
 
 func init() {
--- a/cmd/fyne/internal/mobile/build.go
+++ b/cmd/fyne/internal/mobile/build.go
@@ -82,7 +82,7 @@ func AppOutputName(os, name string) string {
 	switch os {
 	case "android":
 		return androidPkgName(name) + ".apk"
-	case "ios":
+	case "ios", "iossimulator":
 		return rfc1034Label(name) + ".app"
 	}
 
--- a/cmd/fyne/internal/mobile/build_iosapp.go
+++ b/cmd/fyne/internal/mobile/build_iosapp.go
@@ -154,9 +154,13 @@ func goIOSBuild(pkg *packages.Package, bundleID string, archs []string,
 		}
 	}
 
-	// Use codesign to sign the application so that it can be used in iOS simulator.
-	if err := execabs.Command("codesign", "--force", "--sign", "-", buildO).Run(); err != nil {
-		return nil, err
+	// Use codesign to remove the codesign certificate for the built application
+	// so that it can run in iOS simulator.
+	if buildTarget == "iossimulator" {
+		if out, err := execabs.Command("codesign", "--force", "--sign", "-", buildO).CombinedOutput(); err != nil {
+			printcmd("codesign --force --sign --keychain %s\n%s", buildO, out)
+			return nil, err
+		}
 	}
 
 	return nmpkgs, nil
--- a/cmd/fyne/internal/mobile/build_iosapp.go
+++ b/cmd/fyne/internal/mobile/build_iosapp.go
@@ -153,6 +153,12 @@ func goIOSBuild(pkg *packages.Package, bundleID string, archs []string,
 			return nil, err
 		}
 	}
+
+	// Use codesign to sign the application so that it can be used in iOS simulator.
+	if err := execabs.Command("codesign", "--force", "--sign", "-", buildO).Run(); err != nil {
+		return nil, err
+	}
+
 	return nmpkgs, nil
 }
 
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ single codebase.
 Version 2.1 is the current release of the Fyne API, it introduced RichText
 and the DocTabs container, as well as the document storage API and FyneApp.toml
 metadata support.
-We are now working towards the next big release, codenamed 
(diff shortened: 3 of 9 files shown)


andydotxyz:
Do we need another target system to support this? the iOS package already includes both amd64 and arm64...
Unfortunately on my system iOS builds with or without this patch do not run on the simulator and I have not been able to figure out why.

changkun:
> Do we need another target system to support this? the iOS package already includes both amd64 and arm64...

Yes, but only works for darwin/amd64 systems. The iOS simulator fails on M1 macs.
Packing darwin/arm64 for device and darwin/arm64 for iOS seems impossible when using `xcrun lipo`. As they have the same architecture. Adding a iossimulator better isolates ios targets and iossimulator targets.

andydotxyz:
Good point. In that case should we remove amd64 from the iOS builds?

changkun:
> In that case should we remove amd64 from the iOS builds?

Sure, the latest commit removes amd64 for iOS builds.

cmd/fyne/internal/mobile/build_iosapp.go
```diff
@@ -153,6 +153,12 @@ func goIOSBuild(pkg *packages.Package, bundleID string, archs []string,
 			return nil, err
 		}
 	}
+
+	// Use codesign to sign the application so that it can be used in iOS simulator.
```
andydotxyz, line 157: Isn't this handled in the "fyne package" command?

changkun: I figured. The `fyne package` was for codesign in a real device, if we are installing to a simulator, then the certificate should be removed.

See changes in a new commit.

andydotxyz:
This is looking like a great idea. However when I tried to package on my darwin/amd64 I get this error:
```
 fyne package -os iossimulator
open .
go build -tags ios -ldflags=-w -o=/var/folders/rk/wps4hdmx73ldvbq2dbprd_3c0000gn/T/fyne-work-147869268/arm fyne.io/fyne/v2/cmd/fyne_demo failed: exit status 2
# fyne.io/fyne/v2/internal/driver/mobile/gl
In file included from ../../internal/driver/mobile/gl/work.go:32:
./work.h:20:10: fatal error: 'OpenGLES/ES2/glext.h' file not found
#include <OpenGLES/ES2/glext.h>
         ^~~~~~~~~~~~~~~~~~~~~~
1 error generated.

```

changkun:
> when I tried to package on my darwin/amd64 I get this error:

Hmm... that's weird, probably because the Go command was set for an amd64 which does not exist in the work header. Will have to find an amd64 device later to verify this hypothesis.

changkun:
Does this change work for you?

```diff
diff --git a/internal/driver/mobile/gl/work.go b/internal/driver/mobile/gl/work.go
index 3e5338df..67b30a7d 100644
--- a/internal/driver/mobile/gl/work.go
+++ b/internal/driver/mobile/gl/work.go
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
 // +build darwin linux openbsd freebsd
 
 package gl
@@ -17,6 +18,7 @@ package gl
 
 #cgo android            CFLAGS: -Dos_android
 #cgo ios                CFLAGS: -Dos_ios
+#cgo ios,amd64          CFLAGS: -Dos_ios
 #cgo darwin,amd64,!ios  CFLAGS: -Dos_osx
 #cgo darwin,arm         CFLAGS: -Dos_ios
```

andydotxyz:
> Does this change work for you?

No, I didn't think it would, it's just reinforcing the rule above isn't it?

I tried changing
```
#cgo ios                LDFLAGS: -framework OpenGLES
```
to
```
#cgo ios,arm64                LDFLAGS: -framework OpenGLES
#cgo ios,amd64                LDFLAGS: -framework OpenGL
```

But then I got the error:
```
ld: framework not found OpenGL
```

changkun:
ios,amd64 should use OpenGLES in a simulator, isn't it?

```diff
-#cgo ios,amd64                LDFLAGS: -framework OpenGL
+#cgo ios,amd64                LDFLAGS: -framework OpenGLES
```

andydotxyz:
> ios,amd64 should use OpenGLES in a simulator, isn't it?

That is effectively where we started, and macOS does not provide openGLES headers. Maybe something in the compile process is not using XCode correctly on amd64 with the iossimulator tag?

andydotxyz:
One thing that may help: I think you can install Go amd64 on an M1 which is not quite the same but might enable you to test the code path for that architecture?

changkun:
Strage, I am having a darwin/amd64 now. But I could not reproduce your issue..

changkun:
I recall I have had some system upgrades lately (macOS 11.16 Big Sur), this might be the only reason why I cannot reproduce the issue there. What's your system version and Xcode toolchain version?

```
$ uname -a
Darwin changkun-lmu.local 20.6.0 Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64 x86_64

$ go version
go version devel go1.18-527609d47b Wed Aug 25 17:07:58 2021 +0200 darwin/amd64

$ xcode-select -version
xcode-select version 2384.
```

changkun:
The error of GLES also seems similar to what's reported in https://github.com/fyne-io/fyne/pull/2487

changkun:
I just rebased the PR on top of the latest develop branch, which includes the changes that were in #2487. Hopping this could resolve the issue of not finding OpenGLES..

andydotxyz:
Sadly not, this is still showing the same issue.
I suspect that somewhere the new iossimulator type is not hitting some condition that ios turns on...
(just to confirm that ios build still works with this PR applied)

changkun:
OK, I finally get the build issue reproduced. Go after 1.16 is fine, but Go 1.14 and 1.15 got failed with the mentioned GL error. Let's see what's wrong with the versions...

changkun:
@andydotxyz I just pushed a new commit and hope it will resolve the issue for you :)

andydotxyz:
> @andydotxyz I just pushed a new commit and hope it will resolve the issue for you :)

Excellent, thanks that worked !

cmd/fyne/internal/mobile/build.go
```diff
@@ -159,7 +159,7 @@ func runBuildImpl(cmd *command) (*packages.Package, error) {
 		}
 		if buildRelease {
 			if len(allArchs["ios"]) > 2 {
-				targetArchs = []string{"arm", "arm64"}
+				targetArchs = []string{"arm64"}
```
andydotxyz, line 162: Sadly this (and the change below) removes iOS 32bit, which we currently support.
Both iOS and iOSSimulator targets build arm64 and x86_64, would it not be better if the simulator built arm64 and x86_64, but the ios targeted arm and arm64?

andydotxyz (changes requested):
There must be a way to support this new feature without removing 32bit iOS support?

cmd/fyne/internal/mobile/build.go
changkun: I propose to drop the iOS 32bit support in the next release, as we have already done #2487 already.

andydotxyz: > I propose to drop the iOS 32bit support in the next release, as we have done #2487 already.

I'm not sure about your assertion here - `develop` still builds for iOS 32bit when compiled with Go 1.14.
We should not drop support for a platform as a side effect of another feature.
Additionally whilst we support Go 1.14 we should support not drop support for some of the platforms it builds for.

changkun: > still builds for iOS 32bit when compiled with Go 1.14.

I don't really understand why iOS 32bit is still targeted. When Apple (as the system manufacturer) is not supporting these devices (i.e. iPhone 5/5c or earlier, iPad 3 or earlier). People may still have these devices (I do, I still have an iPhone 5), but are people still frequently using them today? Building software that targeting for such a device is like fantasy -- meaning the developer must have: a compatible OS X, a compatible Xcode toolchain, and an appropriate physical device -- otherwise, testing can also be an issue.

Claiming the support of these systems may sound quite cool and responsible. But the support can strap the development process and future moving, such as no testing environment, too many version checks, etc.

> We should not drop support for a platform as a side effect of another feature.

I won't framing the argument of "dropping the support" as a side effect of this feature. I'd propose it become a general release goal.

> Additionally whilst we support Go 1.14 we should support not drop support for some of the platforms it builds for.

Honestly, if possible I'd also propose to bump to Go 1.16 for several reasons:

1. 1.14 is not supported by Go team
2. Installing a new go version is not a problem if one follow the Go's official distribution, such as `go install golang.org/dl/go1.16@latest`; `go1.16 download`, then `go1.16` is already an usable `go` command.

changkun: One more thing: "still builds" is far away from "quite usable". Given the device's actual computation capability, it is reasonable to say that the device may not run a rendering extensive application smoothly and swiftly.

andydotxyz: > I don't really understand why iOS 32bit is still targeted.

It relates to my previous concerns that Go team are removing support for things too aggressively.

> Building software that targeting for such a device is like fantasy

That seems a little harsh - it is working for some of our developers.

> I won't framing the argument of "dropping the support" as a side effect of this feature. I'd propose it become a general release goal.

Ok, so open a new ticket to hold the discussion, this thread is probably not the right place.

> Honestly, if possible I'd also propose to bump to Go 1.16 for several reasons:

Sorry this is not possible. Our process is to support Go versions that are delivered as standard in up to date operating systems. Debian, for example, packages 1.15 in the current release - their next major release is not for 18 months or more, so we won't update beyond 1.15 until then.

I am sorry that this may seem frustrating - but if the first thing that developers looking to work with Fyne see is some cryptic compile error because we require new APIs then we are not doing a great job for onboarding.

> it is reasonable to say that the device may not run a rendering extensive application smoothly and swiftly.

In my testing the apps work reasonably for many usages. I would worry that if we require the full-on modern CPUs to render well, then there is not much processor left for apps to do their own work ;)

changkun:
Concert this to draft as I am still waiting for my darwin/amd64 repairment. Until I could further build and test 32bit devices, nothing seems able to proceed here...

andydotxyz:
> Concert this to draft as I am still waiting for my darwin/amd64 repairment. Until I could further build and test 32bit devices, nothing seems able to proceed here...

I'm a little lost here sorry - the previous comment was about testing on M1, did I miss the point?
What resources can I provide that will help to progress?

changkun:
> I'm a little lost here sorry - the previous comment was about testing on M1, did I miss the point? What resources can I provide that will help to progress?

The previous discussion eventually ends on removing 32-bit iOS devices as the PR can work both for amd64/arm64 Macs by [removing `arm` target from the `fyne` command](https://github.com/fyne-io/fyne/pull/2495/files#diff-1ba63994477c4c6f48eb87f403429996b420495ff0cb14ec058cda322222cd97L160). However, as discussed, removing 32-bit iOS devices is not the goal at this point. The previous impression was there might be a build tag not set properly, but further investigation somehow failed at least for quite some time. Later, my amd64 Mac is broken, and I have no device that can compile a 32 bit application. Hence this PR was paused.

I don't know what resources I need except having real devices to really compile a 32-bit iOS application at least once.

andydotxyz:
> I don't know what resources I need except having real devices to really compile a 32-bit iOS application at least once.

Unfortunately I do not have one that I can provide access to, though I am able to test on it myself. If I can be of assistance you can let me know what commands on what branch would be helpful :).

changkun:
I think I've figured the problem. As I tested the latest status of this PR on two of my Macs, it has fixed the previous concerned issue. Now the code works both for `ios/arm`, `ios/arm64`, `ios/amd64`, `iossimulator/amd64`, `iossimulator/arm64` on either `amd64` Macs or `arm64` Macs.

PTAL.

andydotxyz (approved):
Tested and working well thanks.
It is a shame about the duplicate file, but I guess with the import issues it is not possible to resolve another way?
We can remove the old file once we stop supporting Go 1.14 ;)

PR #2499, opened 2021-09-22
Don't use the toolchain name for nm, they were removed

### LINKED ISSUE #2498: Android -nm tool not found with NDK 23
As reported by a few users the latest Android NDK cannot build. The current fix is to downgrade to 21.
Error:

```
$ ANDROID_NDK_HOME=~/Android/Sdk/ndk/23.0.7123448/ fyne package -os android -appID com.example.myapp -icon ./../fyne_settings/Icon.png 
/home/###/Android/Sdk/ndk/23.0.7123448/toolchains/llvm/prebuilt/linux-x86_64/bin/arm-linux-androideabi-nm /tmp/gomobile-work-124400139/lib/armeabi-v7a/libfyne_demo.so: fork/exec /home/###/Android/Sdk/ndk/23.0.7123448/toolchains/llvm/prebuilt/linux-x86_64/bin/arm-linux-androideabi-nm: no such file or directory
```

It seems they removed the arch specific `nm`, the `llvm-nm` is now used instead.

andydotxyz:
Fixes #2498

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/mobile/env.go
+++ b/cmd/fyne/internal/mobile/env.go
@@ -359,6 +359,8 @@ func (tc *ndkToolchain) Path(ndkRoot, toolName string) string {
 	switch toolName {
 	case "clang", "clang++":
 		pref = tc.ClangPrefix()
+	case "nm":
+		pref = "llvm"
 	default:
 		pref = tc.toolPrefix
 	}


andydotxyz:
You can verify this by running `fyne package -os android` on a computer with NDK of version 23

changkun:
What if someone is still using NDK with nm but not llvm?

andydotxyz:
> What if someone is still using NDK with nm but not llvm?

Is that possible for the supported toolchains?
I could not find out how long ago llvm was introduced but it seemed to be present for those I had access to (the last 3 major releases)

andydotxyz:
> What if someone is still using NDK with nm but not llvm?

Actually, my previous question does not even need answered - the `pref` variable that this change sets to "llvm" is used in the path 
```go
filepath.Join(ndkRoot, "toolchains", "llvm", "prebuilt", archNDK(), "bin", pref+"-"+toolName)
```

I think we can assume that the toolchain "llvm" will be available if the directory llvm is in the toolchains folder.
There may be an additional issue where we don't support really old NDK - but this change should not be a factor.

andydotxyz:
Review period lapsed, merging

PR #2504, opened 2021-09-26
internal/async: close unbounded channels gracefully

changkun:
### Description:

The previously introduced unbounded channel can be improved in the following aspects:

1. avoid misuse when closing the channel: if an unbounded channel was closed, one must drain out all elements, otherwise, there will be a leaking goroutine, and memories.
2. the internal buffer may hold cached elements until very late, which leads to a (pseudo) memory leak.

This PR improves the async's unbounded channel implementation so that when a channel is closed, it internally will process all elements and trying to deliver to a listener (if possible). Then terminates the created goroutine.

With this change, the internal/driver/mobile/app also uses the async package so that all unbounded channels are managed in one place.

Furthermore, as a specialization, struct{} type only uses a counter internally which actually improves the memory usage.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


internal/async/gen.go
```diff
@@ -111,15 +107,20 @@ func NewUnbounded{{.Name}}Chan() *Unbounded{{.Name}}Chan {
...
-				return
+			select {
+			case e, ok := <-ch.in:
+				if !ok {
+					close(ch.out)
+					return
+				}
+				q = append(q, e)
+			case <-ch.close:
+				goto closed
```
andydotxyz, line 118: Can this be done without `goto`? This makes code much harder to read, even `break label` is difficult. 
In this instance it is code to execute before return, so instead it could be defined as a helper function and called before a return I think?

andydotxyz (changes requested):

internal/async/gen.go
changkun: Done.

andydotxyz (approved):
Thanks, nice fix.
We normally would not include a panic in the library, as it should never panic, but this is internal usage so I guess its OK?

changkun:
The panic happens if and only if there is a clear misuse. If that happens, one should observe that during their development.

PR #2505, opened 2021-09-26
internal/driver/mobile: allow up to 120 fps on iOS

changkun:
### Description:

iOS devices now can have up to 120 fps display refresh rate, to let the drawing be more responsive, let the draw loop return sooner (previously can only return after 100ms  when there are too much work, meaning only 10fps was possible).

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/mobile/app/darwin_ios.go
+++ b/internal/driver/mobile/app/darwin_ios.go
@@ -213,6 +213,11 @@ func lifecycleVisible() { theApp.sendLifecycle(lifecycle.StageVisible) }
 //export lifecycleFocused
 func lifecycleFocused() { theApp.sendLifecycle(lifecycle.StageFocused) }
 
+// Maximum allowed framerate. Starting from iPhone 13 Pro, the display refresh
+// rate can up to 120 fps. To offer smoother experience, allow the draw loop
+// to return sooner so that the application can be more responsive.
+const defaultFramerate = time.Second / 120
+
 //export drawloop
 func drawloop() {
 	runtime.LockOSThread()
@@ -225,44 +230,8 @@ func drawloop() {
 		case <-theApp.publish:
 			theApp.publishResult <- PublishResult{}
 			return
-		case <-time.After(100 * time.Millisecond): // incase the method blocked!!
+		case <-time.After(defaultFramerate):
 			return
-		}
-	}
-}
-
-//export startloop
-func startloop(ctx C.GLintptr) {
-	go theApp.loop(ctx)
-}
-
-// loop is the primary drawing loop.
-//
-// After UIKit has captured the initial OS thread for processing UIKit
-// events in runApp, it starts loop on another goroutine. It is locked
-// to an OS thread for its OpenGL context.
-func (a *app) loop(ctx C.GLintptr) {
-	runtime.LockOSThread()
-	C.makeCurrentContext(ctx)
-
-	workAvailable := a.worker.WorkAvailable()
-
-	for {
-		select {
-		case <-workAvailable:
-			a.worker.DoWork()
-		case <-theApp.publish:
-		loop1:
-			for {
-				select {
-				case <-workAvailable:
-					a.worker.DoWork()
-				default:
-					break loop1
-				}
-			}
-			C.swapBuffers(ctx)
-			theApp.publishResult <- PublishResult{}
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


internal/driver/mobile/app/darwin_ios.go
```diff
@@ -225,48 +230,12 @@ func drawloop() {
 		case <-theApp.publish:
 			theApp.publishResult <- PublishResult{}
 			return
-		case <-time.After(100 * time.Millisecond): // incase the method blocked!!
+		case <-time.After(defaultFramerate):
```
andydotxyz, line 233: Are you sure that is how the delay works? I thought that how fast the screen refreshed was down to when the `theApp.publish` channel is sent? Decreasing the time in this segment breaks the select more often, but it does not do anything extra when that happens...?

I think more relevant is the `Run()` in the main mobile driver which has the following code:
```
draw := time.NewTicker(time.Second / 60)
```

andydotxyz (changes requested):

internal/driver/mobile/app/darwin_ios.go
changkun: > I thought that how fast the screen refreshed was down to when the theApp.publish channel is sent?

No. drawloop is executed on the system thread called by `drawInRect` (see drawin_ios.m), which executes all GL draw calls in DoWork. The `publish` channel is sent for calling glFlush, which will still be processed in this function (thread).

If you open the temporary generated Xcode file, running the application, debug panel shows you an Xcode measured fps number, which is literally measuring how fast the `drawInRect` returns.

changkun:
```
draw := time.NewTicker(time.Second / 60)
```

Only controls how regular the draw calls are emitted from the (pesudo) drawing thread. Note that these GL calls are actually send via a channel that actually executed in the `DoWork()`

andydotxyz: But don't we need to update both?
If the GL is updated to allow 120fps but we only send 60 times then we won't see the improvement will we?

changkun: > But don't we need to update both?

Not. The actual FPS only depends on how fast the drawInRect can return, irrelevant to how fast the run produces draw calls.

If each tick of draw in Run produces much more draw calls, the drawInRect can still return soon (which is vita for responsiveness); If each tick of draw in Run produces a little draw call, that's already what we want (drawInRect has less work to process and returns fast).

andydotxyz: If draw is called at 60FPS, how many difference frames will be drawn into your 120FPS output?
Surely it is only 60? animations will slice their timeline to 60 pieces and we won't see the benefit of 120 FPS output.

changkun: Done.

andydotxyz:
Doing this unconditionally will surely lead to degraded performance for anything before an iPhone 13 or an iPad Pro? I think we have to avoid drawing twice as many frames as we need on older devices

changkun:
> Doing this unconditionally will surely lead to degraded performance for anything before an iPhone 13 or an iPad Pro? I think we have to avoid drawing twice as many frames as we need on older devices

What can degrade the performance? If there is nothing to do (e.g. rendering), running the loop slightly faster still does nothing.

andydotxyz:
> What can degrade the performance? If there is nothing to do (e.g. rendering), running the loop slightly faster still does nothing.

I think I was meaning the opposite - if anything is animating (cursor, transition etc) then on a 60FPS device we will be drawing intermediate frames that will never be drawn.
This seems like a bad thing.

changkun:
> > What can degrade the performance? If there is nothing to do (e.g. rendering), running the loop slightly faster still does nothing.
> 
> 
> 
> I think I was meaning the opposite - if anything is animating (cursor, transition etc) then on a 60FPS device we will be drawing intermediate frames that will never be drawn.
> 
> This seems like a bad thing.

Not sure I could follow the reasoning, why intermediate frames  will never be drawn? Isn't the opposite side: drawn earlier, then nothing to draw.

andydotxyz:
If you tick the draw thread at 120fps and there is a transition animating it will attempt to make the intermediate render 120 times each second.
On a 60FPS device it will not be able to draw all of these frames, so half will be lost.

changkun:
Animation is a good argument. If we are intended to have a dynamic refresh rate, then this PR does not really fix it anymore. Close for now. Let's see if we can find a way to achieve a dynamic refresh rate.

PR #2526, opened 2021-10-05
Don't render carriage returns (\r). Fixes #2456

### LINKED ISSUE #2456: Carriage Return (\r) is rendered as space
### Describe the bug:
I've only tested the entry and label widgets so far, but it appears that when Fyne comes across `\r`, it is rendered as a space. Ideally, the application would NOT render it at all, while still potentially having the option to use `\r\n` for linebreaks, for compatibility with other Windows applications.

### To Reproduce:
In my case, I merely edited the textedit example, and inserted this line into the `Show()` function:

```Go
editor.entry.SetText("A test\nwith\rcarriage return,\r\nAnd if it understands?")
```

I also added a `\n\r` into the "Cursor Row" label (which is backwards from how Windows apps normally handle it, but still demonstrates how Fyne labels handle the carriage return)

### Screenshots:
image

 - **OS:** Windows
 - **Version:** 10
 - **Go version:** 1.17
 - **Fyne version:** 2.0.4

reyemxela:
### Description:

Changed both RichText (which most things use as a renderer) and TextGrid to ignore/remove \r characters from input.

Each fix seems to me like the most straight-forward approach, but if there are any tweaks I should make, please let me know.

Fixes #2456 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass. (some tests are failing in develop, but these changes don't seem to affect anything)

### DIFF
--- a/widget/richtext.go
+++ b/widget/richtext.go
@@ -359,6 +359,7 @@ func (t *RichText) updateRowBounds() {
 				continue
 			}
 			textSeg := seg.(*TextSegment)
+			textSeg.Text = strings.ReplaceAll(textSeg.Text, "\r", "") // remove "\r" characters that otherwise render as space
 			textStyle := textSeg.Style.TextStyle
 			textSize := textSeg.size()
 
--- a/widget/textgrid.go
+++ b/widget/textgrid.go
@@ -422,6 +422,9 @@ func (t *textGridRenderer) refreshGrid() {
 			if i >= t.cols { // would be an overflow - bad
 				continue
 			}
+			if r.Rune == '\r' { // remove "\r" characters that don't render correctly
+				continue
+			}
 			if t.text.ShowWhitespace && (r.Rune == ' ' || r.Rune == '\t') {
 				sym := textAreaSpaceSymbol
 				if r.Rune == '\t' {


widget/richtext.go
```diff
@@ -359,6 +359,7 @@ func (t *RichText) updateRowBounds() {
 				continue
 			}
 			textSeg := seg.(*TextSegment)
+			textSeg.Text = strings.ReplaceAll(textSeg.Text, "\r", "") // remove "\r" characters that otherwise render as space
```
andydotxyz, line 362: My understanding is that this comes as a result of loading files on (for example) windows right? So `SetText` may be a better place to do this (just a 1 time cost).

andydotxyz (changes requested):
Just a thought inline.
Plus this is a great thing to have a test for :)

widget/richtext.go
reyemxela: That's a good point. I personally haven't experienced this issue, but you're right, a \r would probably only happen initially. I'll change that.

reyemxela:
Moved checks to `SetText()` for labels, and `updateText()` for entries, as it's possible to paste carriage returns into an entry.

Also added basic tests for both cases.

As I was adding the tests, I realized that there are still ways this could get missed, such as setting `(label/entry).Text` directly. Honestly not sure how big of a deal that is.

andydotxyz:
> As I was adding the tests, I realized that there are still ways this could get missed, such as setting `(label/entry).Text` directly. Honestly not sure how big of a deal that is.

That is a very good point. Can we protect against that here just like SetText without too much more overhead? I think there is a check internally to see if the field was changed (as we have the caching internally)

reyemxela:
That's what I was trying to figure out, but didn't find it yet. I'll keep digging and mess with it.

andydotxyz:
> We need to make sure that this doesn't slow things down.

Yes we need to make sure it is not impacting performance. However the change is needed - loading a windows file will display blocks where the `\r` was.

reyemxela:
Definitely. I know it's a pretty rare edge case (it sounded like an easy fix when I saw the issue, so I thought it'd be a nice quick thing to tackle. lol).

As for the textgrid, there shouldn't be any noticeable impact. It's already looping through each character and handling spaces/tabs and other things, so it's just ignoring the \r there.

I was trying to simplify the Entry and Label fixes to only run when text is first set, but I'm running into other things there, including a potential bug in the label code. (if I can figure that out, I might make a separate PR for it)

A compromise could be to just leave the fix in the `SetText` functions and nowhere else, so it only runs once when actually setting the text. And if the user is trying to assign `(label/entry).Text` directly, they're responsible for cleaning up the text themselves.

andydotxyz:
> A compromise could be to just leave the fix in the `SetText` functions and nowhere else, so it only runs once when actually setting the text. And if the user is trying to assign `(label/entry).Text` directly, they're responsible for cleaning up the text themselves.

I'd rather not do this, in general the contract of `SetXxx(...)` is the same as `Xxx = ...; Refresh()`.

andydotxyz:
Is there an alternative approach that could take this forward? It would be good to get this in and keep the basic handling of fields vs Refresh() etc.

stuartmscott:
As this is a rendering issue, wouldn't it be better handled in the driver (maybe [here](https://github.com/fyne-io/fyne/blob/master/internal/painter/gl/gl_common.go#L88), and similar for software renderer)? `RenderedTextSize` would also need to ignore `\r`.

Consider someone making a notepad-style app, they open the app on Windows, load a file, type something, and then save it - only to find the app has stripped all the carriage returns from the file without warning and without an option to override this behaviour.

stuartmscott:
#2733

PR #2530, opened 2021-10-06
Fix issue #2525 (cursor position updates twice with a delay)

### LINKED ISSUE #2525: Mouse click in an entry updates cursor position twice, with a noticeable delay
### Describe the bug:
Suppose an entry widget has some text in it. User does a mouse click somewhere in the text and immediately starts typing "qwerty". When I do that, I usually get "ertyqw" because after I type one or two symbols (in this case, "qw") the cursor resets back to the position where I clicked, and the remaining symbols (in this case, "erty") are inserted before "qw".

This behavior is easily visible in a multiline entry: click at the end of some line and quickly press Enter to start a new line. Soon the cursor will return back to the end of the line you clicked. This does not require extreme speed, I can easily trigger this behavior even using the same hand to click the mouse first and press Enter afterwards.

### Bug location
In `widget/entry.go` both callbacks `Tapped()` and `MouseDown()` call `e.updateMousePointer`. For some reason `Tapped()` is invoked after a delay, allowing user to type some text in between these two callbacks. I'm not sure how input handling works in fyne, but maybe on a desktop machine it is enough to update the cursor only in MouseDown()? For example, by modifying Tapped() as follows:
```go
func (e *Entry) Tapped(ev *fyne.PointEvent) {
	if !e.Disabled() {
		e.requestFocus()
	}
	if fyne.CurrentDevice().IsMobile() && e.selecting {
		e.selecting = false
	}
	if fyne.CurrentDevice().IsMobile() { // Added this line
		e.updateMousePointer(ev.Position, false)
	} // Added this line
}
```
This fixes the problem on my machine.

### To Reproduce:
1. Open fyne_demo and navigate to "Widget/Entry" page.
2. Type something in the entry.
3. Click in the middle of the text you typed and start typing "qwerty".
4. You will probably see "ertyqw" or something like that instead of the intended result.

 - **OS:** MacOS
 - **Version:** 10.12 Sierra
 - **Fyne version:** 6e90820ea9ca4df836db5a99384c51073415571e

nullst:
### Description: 
Fixes #2525 following the strategy suggested there by @andydotxyz .

Lots of complications arise due to tests for widget.Entry. Essentially, the tests only use `Tapped()` callback to simulate user interaction, while in real applications it is always a sequence:
* either MouseDown() -> MouseUp() -> (double click delay) -> Tapped();
* or TouchDown() -> TouchUp() -> (double click delay) -> Tapped().

Fixing this problem in `widget/entry_internal_test.go` is easy: replace all calls to `.Tapped()` to a helper function that performs the sequence above (though, unlike real applications, it omits the double click delay).

Fixing this problem in `widget/entry_test.go` is harder: the test `TestEntry_Tapped` simulates an application window using `fyne/v2/test` package and sends the events by invoking `test.TapCanvas()`, which only sends `Tapped()`. There are a couple of options:
1. (seems wrong) Move `TestEntry_Tapped` to `entry_internal_test.go`, use widget callbacks directly instead of canvas interactions. I think internal tests are not expected to use `test.AssertRendersToMarkup` and do window-level simulations, but these are too valuable to get rid of them.
2. (seems wrong) Break the abstraction layer in `entry_test.go` and use widget callbacks there instead of TapCanvas(). Roughly the same issues as the option above. This conflicts with what widget tests do for all other widgets.
3. Add a function LeftClickCanvas() to the `test` package analogous to `TapCanvas()` that is already there.

I've implemented solution 3 in this pull request. I guess one issue is consistency: should all widget tests be ported to LeftClickCanvas() eventually instead of TapCanvas()?

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/test/test.go
+++ b/test/test.go
@@ -79,13 +79,6 @@ func AssertRendersToMarkup(t *testing.T, masterFilename string, c fyne.Canvas, m
 	return true
 }
 
-// ClickCanvas simulates a primary mouse button click at an absolute position on the canvas.
-func ClickCanvas(c fyne.Canvas, pos fyne.Position) {
-	if o, p := findMouseable(c, pos); o != nil {
-		clickPrimary(c, o.(desktop.Mouseable), &fyne.PointEvent{AbsolutePosition: pos, Position: p})
-	}
-}
-
 // Drag drags at an absolute position on the canvas.
 // deltaX/Y is the dragging distance: <0 for dragging up/left, >0 for dragging down/right.
 func Drag(c fyne.Canvas, pos fyne.Position, deltaX, deltaY float32) {
@@ -279,15 +272,6 @@ func findTappable(c fyne.Canvas, pos fyne.Position) (o fyne.CanvasObject, p fyne
 	return
 }
 
-func findMouseable(c fyne.Canvas, pos fyne.Position) (o fyne.CanvasObject, p fyne.Position) {
-	matches := func(object fyne.CanvasObject) bool {
-		_, ok := object.(desktop.Mouseable)
-		return ok
-	}
-	o, p, _ = driver.FindObjectAtPositionMatching(pos, matches, c.Overlays().Top(), c.Content())
-	return
-}
-
 func prepareTap(obj interface{}, pos fyne.Position) (*fyne.PointEvent, fyne.Canvas) {
 	d := fyne.CurrentApp().Driver()
 	ev := &fyne.PointEvent{Position: pos}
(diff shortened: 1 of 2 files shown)


test/test.go
```diff
@@ -274,6 +281,17 @@ func findTappable(c fyne.Canvas, pos fyne.Position) (o fyne.CanvasObject, p fyne
 	return
 }
 
+func findMouseable(c fyne.Canvas, pos fyne.Position) (o fyne.CanvasObject, p fyne.Position) {
+	matches := func(object fyne.CanvasObject) bool {
+		if _, ok := object.(desktop.Mouseable); ok {
+			return true
+		}
+		return false
```
Jacalz, line 289:
```suggestion
		_, ok := object.(desktop.Mouseable)
		return ok
```
You should be able to shorten that down to the following.

Jacalz (changes requested):
Thanks for the PR. Just a small thing I found. I wouldn't consider this a thorough review.

PR #2532, opened 2021-10-06
Remove FileIcon Selection

stuartmscott:
### Description:
Removes the deprecated FileIcon.SetSelected.

To be merged after the appropriate deprecation window has passed.

### Checklist:
- [ ] Tests included.
- [x] Tests all pass. Except TestWindow_HandleOutsideHoverableObject

#### Where applicable:


Jacalz:
Should we close this and then remove all deprecations when we approach 3.0?
This is going to be far behind upstream when it finally is time to merge it.

stuartmscott:
> far behind upstream

What other changes are planned for this file?

Jacalz:
It doesn't necessarily have to be that file that causes the issues. The actions are going to be out of sync (probably already are) and will either not run at all or give a false hope that things look fine.

andydotxyz:
It's true that the actions are out of sync, but merging develop or rebasing should fix that.
Long running branches need to be maintained in this way, but it doesn't mean there will be conflicts - just a little routine work to keep it aligned.

andydotxyz:
It was recently agreed that v3.0 is some distance off (i.e. 18 months or more), we have no plans to begin work on it yet.
Additionally we want to try and stay on top of PRs, so have added to the contributor guidelines that stale PRs (no activity in 6 months) will be closed.
Given that this change cannot be actioned this year I am closing it.

Thanks for your help on keeping our code clean.
I have linked from the 3.0 deprecation removal ticket so this can be re-opened and applied if it does not conflict.

Issue #1946, opened 2021-02-12
Focused text entry consumes 40%-80% of CPU in my app


ababo:
### Describe the bug:
I have several forms where text entries can take and lose focus. I noticed that when entry is focused (even without entering any text) the CPU consumption seriously increases.

### Screenshots

<img width="920" alt="Screenshot 2021-02-12 at 11 28 17" src="https://user-images.githubusercontent.com/503292/107757224-90790580-6d25-11eb-85a5-1d6a4c57f8a6.png">
<img width="904" alt="Screenshot 2021-02-12 at 11 28 48" src="https://user-images.githubusercontent.com/503292/107757235-92db5f80-6d25-11eb-9669-f12a778ecd79.png">

### To Reproduce:
Use any app that contains text entries.

### Reproducing conditions:
Platform: `OS X 11.1`
Go version: `go1.15.7 darwin/amd64`
Fyne version: `v2.0.0`

**Note:** This problem is not reproduced with Fyne v1.4.1.

Jacalz:
Can you please fill out that issue template when opening bug reports? I see that you have used it, but then removed all of the template. Please just use the template which contains the necessary information to diagnose the issues. Thanks for understanding 🙂

ababo:
@Jacalz updated

Jacalz:
Thanks 🙂

bebo-dot-dev:
I see this issue on Linux Mint x64 with fyne v2.0.3. CPU usage seems to be around 25% here and the application has to have focus / be topmost for CPU usage to spike.

andydotxyz:
Are you also using fyne_demo when you see the problem @bebo-dot-dev ?

bebo-dot-dev:
Hi @andydotxyz I wasn't using the fyne_demo when I saw this, I was playing with [my little proof of concept application](https://github.com/bebo-dot-dev/go-auth-fyne)

..but now that you mentioned it, yes I can see it with the fyne_demo application too and CPU usage does seem to be ~40% with fyne_demo

andydotxyz:
Looking into this further I think the issue is that the cursor pulses through every possible frame, so the entire UI is refreshing at 60 FPS. Because we're doing Go->C->OpenGL this is expensive even though we are not actually rendering any new textures.
There are probably 3 possible approaches:
* minimise the cursor pulse (i.e. add pause between a faster fade in and then fade out, so we refresh 1/3 the time)
[...]

Issue #436, opened 2019-09-04
Add Undo/Redo support for widget.Entry


andydotxyz:
Required for #434

andydotxyz:
As this ticket add to the entry popup

AndrewWPhillips:
I think this should not just be changes to the text but also changes to the selection, in which case it may even be required for read-only entry widgets.

andydotxyz:
Undoing selection sounds a little uncommon? If I make a change then copy 3 different bits of text then undo has to be executed 4 times to undo the actual change I intended...

xypwn:
@andydotxyz This can be closed now, right?

andydotxyz:
Ah yes thanks, I missed that

PR #2539, opened 2021-10-08
Add basic Undo/Redo functionality to Entry widget

nullst:
### Description:
This pull request introduces a basic Undo/Redo functionality in the Entry widget, fixing issue #436 . The realization works as is for me, but I have some questions, presented slightly below. I am also working on writing tests for the new functionality.

### Basic description

If `Entry.HistoryDisabled` is false, whenever a user performs an action like, e.g., "typed rune", "paste from clipboard", or "erase selection by pressing backspace", the full state of the entry widget is saved and appended to an action log. The state includes full text, cursor position, and scroll position. When user calls "Undo" on a history-enabled widget, the state is restored from the action log. Redo is implemented in a similar way.

To improve user experience, a sequence of "typed rune" actions performed in a quick succession is considered a single action, so Undo does not cancel just one letter at a time. Same for a sequence of deletions (e.g., pressing backspace several times).

Calling SetText() erases the history, with a new SetTextUndoable() function that adds a new event to the history log instead.

For simplicity, undo action resets the current selection to empty.

### Justification for the crude mechanism
Better history implementation would store only user actions and roll them back in a smart way when requested. Saving the whole widget state is slower and wasteful in terms of memory, but this is the approach used here. Why?

- **Resistant to modifications of Entry internals.** RichText had just been introduced in Fyne 2.1, and I expect further changes to Entry internals to follow soon. A smarter approach working with diffs between RichText trees may become obsolete as RichText progresses.
- **This is a first draft.** Internal optimizations may follow later. Designing public APIs is separate from the implementation details.
- **Performance of undo is not a bottleneck.** For short texts or short-lived entry widgets, the overhead is negligeable. For a long-lived entry with long text, well, this would not be the only performance issue (see issue #1946 etc).

#### Design questions

1. As mentioned, selection is just reset during the undo. Is this OK?
2. When some text is selected and user types the rune "k", the selected text is replaced by the letter "k", and currently this is registered as a single action. Should it be split into two? Erasing the selection, then typing k.
3. I created fyne.ShortcutUndo and fyne.ShortcutRedo globally and on desktop bound them to Ctrl+Z and Ctrl+Y respectively. Is this OK? I'm not sure about Ctrl+Y.
4. I added "Undo" and "Redo" menu items to the right-click menu of an entry with history enabled. But I'm not sure how common this is. Should we only show "Undo" by default? Maybe show no items and rely on user implicit knowledge of Ctrl+Z? Should there be a separator line between undo and the clipboard actions?
5. Should history be disabled for password entries?
6. What should I do with scroll position? I'm currently saving Entry.scroll.Offset and Entry.content.scroll.Offset and restoring them. This is wrong if entry has been resized, for example. In terms of user experience, ideally the scroll position should not be modified if the part of the text modified by the Undo() is already visible to the user, but this is impossible to implement nicely with a crude history mechanism.
7. This pull request probably adds more data races. For example, `restoreHistorySnapshot()` would ideally restore the widget state completely while holding propertyLock(), but the current implementation of `updateText()` is not particularly conductive to this (and `setFieldsAndRefresh` in general). Someone should take a look if there are severe problems with concurrency in this code, I'm not very experienced with this.

#### Implementation questions

- Since the undo stack behavior changes based on the delay between successive TypedRune() events, for the purpose of testing I used kind of a "dependency injection": Entry.timestamper is usually a reference to time.Now function, but may be changed in internal tests. Is this acceptable?
- I've added some basic tests for Undo/Redo, but I don't know if that's enough.

### Public API changes

Added to widget.Entry:

- CanUndo(), CanRedo() `func() bool`: return true if entry history is enabled and the corresponding action can be performed. Can be used by applications for disabling menu items or toolbar buttons.
- Undo(), Redo() `func()`: perform undo/redo.
- SetTextUndoable `func(text string)`: sets the text of the entry to the provided string and saves this as a separate action in the action log.

Modified behavior in widget.Entry:

- SetText() now erases the action log if history is enabled.

New shortcuts:

- fyne.ShortcutUndo, fyne.ShortcutRedo: new shortcuts with no associated data (just `struct{}`). On desktop they are bound to Ctrl+Z and Ctrl+Y.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -1274,6 +1274,12 @@ func (w *window) triggersShortcut(keyName fyne.KeyName, modifier desktop.Modifie
 		case fyne.KeyA:
 			// detect selectAll shortcut
 			shortcut = &fyne.ShortcutSelectAll{}
+		case fyne.KeyZ:
+			// detect undo shortcut
+			shortcut = &fyne.ShortcutUndo{}
+		case fyne.KeyY:
+			// detect redo shortcut
+			shortcut = &fyne.ShortcutRedo{}
 		}
 	}
 
--- a/shortcut.go
+++ b/shortcut.go
@@ -83,3 +83,19 @@ type ShortcutSelectAll struct{}
 func (se *ShortcutSelectAll) ShortcutName() string {
 	return "SelectAll"
 }
+
+// ShortcutUndo describes a shortcut undo action.
+type ShortcutUndo struct{}
+
+// ShortcutName returns the shortcut name
+func (se *ShortcutUndo) ShortcutName() string {
+	return "Undo"
+}
+
+// ShortcutRedo describes a shortcut redo action.
+type ShortcutRedo struct{}
+
+// ShortcutName returns the shortcut name
+func (se *ShortcutRedo) ShortcutName() string {
+	return "Redo"
+}
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -90,11 +90,11 @@ type Entry struct {
 	textSource   binding.String
 	textListener binding.DataListener
 
-	historyEnabled bool
+	HistoryEnabled bool
 	// timestamper is a function that returns the current moment. It is usually
 	// just time.Now(), but can be changed during widget tests.
 	timestamper   func() time.Time
-	historyOrigin entryHistoryState
+	historyOrigin *entryHistoryState
 	actionLog     []entryUserAction
 	redoOffset    int
 }
@@ -347,24 +347,6 @@ func (e *Entry) Hide() {
 	e.DisableableWidget.Hide()
 }
 
-func (e *Entry) HistoryEnable() {
-	e.propertyLock.Lock()
-	defer e.propertyLock.Unlock()
-
-	e.historyEnabled = true
-	e.timestamper = time.Now
-	e.registerInitialState()
-}
-
-func (e *Entry) HistoryDisable() {
-	e.propertyLock.Lock()
-	defer e.propertyLock.Unlock()
-
-	e.historyEnabled = false
-	e.timestamper = nil
-	e.registerInitialState()
-}
-
 // Keyboard implements the Keyboardable interface
 //
 // Implements: mobile.Keyboardable
@@ -498,7 +480,7 @@ func (e *Entry) SetText(text string) {
 	e.updateText(text)
 	e.updateCursor()
 
-	if e.historyEnabled {
+	if e.HistoryEnabled {
 		e.registerInitialState()
 	}
 }
@@ -509,7 +491,7 @@ func (e *Entry) SetTextUndoable(text string) {
 	e.updateText(text)
 	e.updateCursor()
 
-	if e.historyEnabled {
+	if e.HistoryEnabled {
 		e.registerAction(entryActionSetText)
 	}
 }
@@ -542,7 +524,7 @@ func (e *Entry) TappedSecondary(pe *fyne.PointEvent) {
 	super := e.super()
 
 	historyItems := make([]*fyne.MenuItem, 0)
-	if e.historyEnabled {
+	if e.HistoryEnabled {
 		historyItems = append(historyItems, fyne.NewMenuItem("Undo", func() {
 			super.(fyne.Shortcutable).TypedShortcut(&fyne.ShortcutUndo{})
 		}))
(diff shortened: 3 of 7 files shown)


andydotxyz:
This is amazing to have worked on thanks!

> If `Entry.HistoryEnabled` is true

Is it likely that developers would want to disable undo/redo? I'm not sure this is needed. If it is then certainly should not be disabled by default!

You're right that Entry will be changing as we move the ability to edit from a single style to rich styles.
The approach you take here is clean, I had not thought about doing it this way - but when you store only the `string` content it won't work if there is any rich text metadata available - which will happen before next release.

I wonder if you may be right about memory concerns. If someone loads a large text file then every time someone types a key it will add the whole amount again. So a 100k text file will be taking 10MB after just 100 keystrokes.

Just on API naming did you consider `CanUndo` instead of `IsUndoAvailable`? Last thought, `SetTextUndoable` I am not sure about - just like how history should be on by default I think we should maintain the history of `SetText` as app developers may do it on keystroke (there are some examples of this I think we have seen). Instead of that complex thought, how about `SetText` maintains history but a `ResetUndoHistory` or similar could be used if that was required?

nullst:
As for the future RichText compatibility, would it be reasonable to add a `Duplicate` method to the `RichTextSegment` interface that returns a deep copy of the segment? Then Undo mechanism could be implemented almost as simply as for plain text. (Making a deep copy by force using `reflect` package is possible, but not good).

This would have further benefits: in the rich text modification functions (`deleteFromTo` and `insertAt`) we can record which elements of the`RichText.Segments` slice were changed/added/deleted during those operations, and let the history log store only the (copies of) modified segments. This would be easier than computing substantial diffs between states, and maybe still valuable. Still extremely far from optimal since `RichText.Segments` is kind of a tree, not just a slice (see ParagraphSegment, ListSegment). 

(Maybe in the future Entry would be smart enough to represent very long texts as several different `TextSegment`s as an optimization, and potentially this could improve the performance generally, including Undo if realized as above.)

andydotxyz:
> * switching notes in [Fyne notes application](https://github.com/fynelabs/notes) calls SetText on the note-title and note-body

Good argument :) I agree now thanks.

> As for the future RichText compatibility, would it be reasonable to add a `DeepCopy` method to the `RichTextSegment` interface that returns a deep copy of the segment?

It would be worth looking at that yes. But you don't need to export the method as we're internal to `widget` package. But I think it would need to copy the whole tree.



> (Maybe in the future Entry would be smart enough to represent very long texts as several different `TextSegment`s as an optimization, and potentially this could improve the performance generally, including Undo if realized as above.)

Not sure if this is the right way to optimise or not, if we record the actual changes (and where the cursor was) we can avoid cloning trees and diffing, so don't need to update the data structure to work efficiently.

nullst:
> But you don't need to export the method as we're internal to widget package. But I think it would need to copy the whole tree.

Sure, that's a good point. Copying the tree is the "deep" part of the "deep copy" :)

> if we record the actual changes (and where the cursor was) we can avoid cloning trees and diffing

It wouldn't be easy to do this for a (future) fully RichText-based Entry without any "deepCopy" method on `RichTextSegment`s (assuming that RichText still provides `deleteFromTo` etc). For example, if a user deletes symbols from 100th to 400th of the text, and symbols 200-300 are a `ListSegment` with some items, we'll need to store a copy of that deleted list in the action log to be able to undo/redo this operation. I mean, either store a copy or compute some kind of a diff between trees, and store that.

Overall, I'm not sure an implementation of a smarter undo/redo mechanism is worth attempting before RichText has been stabilized and integrated into Entry.

nullst:
I think holding back until Entry embraces RichText fully is more sensible. With these modifications I now have an Undo/Redo functionality in an application I'm writing, which makes me content, and when RichText comes around, I'll try to port the modifications.

andydotxyz:
OK thanks. Moved it to draft so we can skim over it in the list for now.

andydotxyz:
We are tidying up [old PRs](https://github.com/fyne-io/fyne/wiki/Contributing#stale-pull-requests) and this one seems unlikely to merge soon.
However I know that `RichEntry` is aiming for the next major release - so are you able to help update this, then port it in @nullst or should it be closed and re-worked after the new widget changes?

nullst:
>  However I know that RichEntry is aiming for the next major release - so are you able to help update this, then port it in @nullst or should it be closed and re-worked after the new widget changes?

I'm not sure if I'll have time to help port this code, so maybe this pull request should be closed.

I'd just like to mention that since the undo mechanism implemented here is quite crude (as explained in the original description), essentially the **only** thing that's needed to adapt the code for `RichEntry` is a way to save and restore the content of RichEntry. For the usual `Entry` the code here just saves the text string (in the `historySnapshot` function) and sets it back (in the `restoreHistorySnapshot` function).

It seems that a proper way would be to add a `DeepClone()` function to the `RichTextSegment` interface and implement it for each segment type. Then possibly some complications could arise with respect to saving/restoring the cursor position and the scroll position of the widget, but that should be simple.

Jacalz:
This was replaced by https://github.com/fyne-io/fyne/pull/4251 a while back. Thanks for working on this @nullst.

PR #2540, opened 2021-10-09
Theme from JSON and user defined themes

### LINKED ISSUE #211: Proposal: Allow dynamic themes via JSON templates

andydotxyz:
### Description:
Allow importing theme from JSON data (useful for devs and/or sharing theme without the Go code).
And support a user specified theme file.

Fixes #211

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/settings.go
+++ b/app/settings.go
@@ -144,15 +144,16 @@ func (s *settings) loadSystemTheme() fyne.Theme {
 	data, err := fyne.LoadResourceFromPath(path)
 	if err != nil {
 		if !os.IsNotExist(err) {
-			fyne.LogError("Failed to parse user theme file: "+path, err)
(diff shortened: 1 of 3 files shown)


andydotxyz:
Thanks @Jacalz moved it to JSON. It's more readable than I expected, so am happy enough with this :)

Issue #2482, opened 2021-09-19
Entry.SetText may panic if called on a multiline entry with selected text


nullst:
### Describe the bug:

Suppose a multiline entry has a long text in it, with some part selected. If `SetText()` is called on the entry with a new text shorter than the original text, the selected region may be out of bounds of the new text. I don't know what the proper behavior should be, but as of now this results in a nil pointer dereference originating from the function `entryContentRendered.buildSelection` in widget/entry.go. An examination in debugger shows that in that function the object `r.content.entry` has: 

1. Updated (short) text.
2. Updated cursor position (at the very end of the new text since previous cursor position was even further).
3. Old values of selectRow and selectColumn (i.e., specifying a position that does not exist in the new text).

This leads to a wrong computation of `rowCount`, and eventually `lineSizeToColumn` is called with an out of bounds arguments.

Notes: 
1. At some moment before the segmentation fault the cursor position was modified to be consistent with the new text, but the selected position was not. Maybe this happens in `entryRenderer.Refresh()`, but I'm not sure.
2. I've originally discovered this when `.SetText()` was called while I, as a user, was in the process of changing the selection state by dragging the mouse. Probably any fix would cover this situation as well.
3. This happens regardless of whether the multiline entry has wrapping enabled.

### To Reproduce:
Steps to reproduce the behaviour:
1. Compile the example code below and run it.
2. Scroll down a bit and select some text.
3. Wait 5 seconds until the timer calls `entry.SetText()`.
4. See segmentation fault.

### Example code:
```go
package main

import (
	"time"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

var textFragment string = "ihsidf hasd fihasdf jas dfuias fduias fuibasd fibas dfoih kjhkhk uhuhuih"

var useWrapping bool = false

func main() {
	app := app.New()
	w := app.NewWindow("Entry with updating text")

	var longText string
	for i := 0; i < 60; i++ {
		longText += textFragment
		if !useWrapping {
			longText += "\n"
		}
	}

	entry := widget.NewMultiLineEntry()
	if useWrapping {
		entry.Wrapping = fyne.TextWrapWord
	}
	entry.SetText(longText)

	go func() {
		time.AfterFunc(5*time.Second, func() {
			entry.SetText(textFragment)
		})
	}()

	w.SetContent(entry)
	w.Resize(fyne.NewSize(float32(500), float32(300)))
[...]

PR #2547, opened 2021-10-12
Update selection indices during SetText to avoid crash in multiline entries

nullst:
### Description:
The Entry widget has an `updateCursor` function that is called after replacing the text in the entry: if the cursor position stored in `Entry.CursorRow` and `Entry.CursorColumn` is now out of bounds (e.g., if the new text is shorter than the original one), the position should be updated. Entry also has fields `selectRow` and `selectColumn` , but they are not updated when the text is replaced. I've modified SetText to also update those fields, thus fixing #2482.

There is a small catch. In the original `updateCursor` logic, if a cursor was at the position (column 5, row 10), and the updated text in the entry has only 2 rows, the new cursor position would be (column 5, row 2) if the second line has at least 5 characters. This in itself seems questionable to me (why not the last character of the second row? That's the closest we can get to "row 10" in the new text), but this logic becomes problematic if applied to select row/col as well. 

Namely, suppose the user selected characters from the 5th to 27th in row 10. Then a SetText is called with a new text having only 2 rows. Using the original logic for cursor movement we would get a new selected region, from the 5th to 27th characters in row 2! For me, as a user, this would be weird. I would expect nothing to be selected now. Thus I changed the position truncation logic for both cursor and selection to avoid this. However, some tests in `TestEntry_Submit` depended on the old behavior, so I modified them as well.

I added a new test: SetText on a multiline entry should move an out-of-bounds cursor position to the end of the last line.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/entry_test.go
+++ b/widget/entry_test.go
@@ -1417,23 +1417,27 @@ func TestEntry_Submit(t *testing.T) {
 	})
 	t.Run("NoCallback", func(t *testing.T) {
 		entry := &widget.Entry{}
+		resetEntry := func() {
+			entry.SetText("")
+		}
 		t.Run("SingleLine_Enter", func(t *testing.T) {
+			resetEntry()
 			entry.MultiLine = false
 			entry.SetText("a")
 			entry.TypedKey(&fyne.KeyEvent{Name: fyne.KeyEnter})
 			assert.Equal(t, "a", entry.Text)
 		})
 		t.Run("SingleLine_Return", func(t *testing.T) {
+			resetEntry()
 			entry.MultiLine = false
 			entry.SetText("b")
 			entry.TypedKey(&fyne.KeyEvent{Name: fyne.KeyReturn})
 			assert.Equal(t, "b", entry.Text)
 		})
 		t.Run("MultiLine_ShiftEnter", func(t *testing.T) {
+			resetEntry()
 			entry.MultiLine = true
 			entry.SetText("c")
-			entry.CursorRow = 0
-			entry.CursorColumn = 0
 			typeKeys(entry, keyShiftLeftDown, fyne.KeyReturn, keyShiftLeftUp)
 			assert.Equal(t, "\nc", entry.Text)
 			entry.SetText("d")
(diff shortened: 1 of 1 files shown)


nullst:
I've implemented the suggestions from the conversations above. Was it fine that I "resolved" them if I agree with them and addressed them in a commit? I don't know about best practices.

Another note: while this patch fixes #2482 in practice, it is not totally solid in theory. Namely, SetText first calls `updateText`, which performs `setFieldsAndRefresh`, which, if I understand correctly, at the end requests a visual refresh of the widget to be done on the main thread. After that, SetText updates cursor and selection positions. Now visual refresh takes a relatively long time, so it is *highly improbable* that the entry widget renderers would attempt to redraw the entry before the second half of SetText updates cursor/selection position. But in principle nothing forbids this: propertyLock is released at the end of updateText and acquired again at the start of updateCursorAndPosition, so technically updateCursorAndPosition could block on acquiring for a long time. And if a visual refresh happens during that time, #2482 may occur again. 

I wasn't able to figure out a neat way to avoid this potential problem. A dumb way would be to create a function `updateTextAssumingPropertyLock` that is exactly like `updateText`, but modifies the fields directly, assuming that the caller holds the propertyLock, and also create something like `updateCursorAndPositionAssumingPropertyLock`, perform the two of them after manually acquiring property lock, and then call refresh.

andydotxyz:
> Was it fine that I "resolved" them if I agree with them and addressed them in a commit? I don't know about best practices.

We  prefer the commentor to mark resolved, so they confirm the fix

nullst:
I think the fix is probably fine as is. The theoretical issue I sketched in https://github.com/fyne-io/fyne/pull/2547#issuecomment-942079908 will probably never happen. I'm not sure what happened with the windows platform tests -- they seem to fail without any details, can I just rerun the CI check for the pull request as is?

> Would it be sensible for the cursor truncation to be done in updateText?

Maybe, but `updateCursor` is also called from the Refresh loop separately, possibly to deal with changes in wrapping modes on the fly, or something like that. The current implementation seems OK for me now.

andydotxyz:
> I'm not sure what happened with the windows platform tests -- they seem to fail without any details, can I just rerun the CI check for the pull request as is?

Looks like a temporary infra fail, I have kicked it to try again

PR #2555, opened 2021-10-16
Fix refreshQueue not being consumed due to a mistake in FreeDirtyTextures()

### LINKED ISSUE #2548: Regression: too many Refresh() calls may now cause visual artifacts in the List widget
### Describe the bug:
Consider a List widget with 1000 `canvas.Text`s, with `updateFunc` callback set up to update the text and call Refresh() on each text element. Scrolling this list quickly (for example, using a touchpad with large scroll acceleration) sometimes results in visual artifacts that are not resolved on their own any time soon. By that I mean, waiting several seconds does not fix them, and sometimes even moving the mouse over the list (which highlights items on hover) does not refresh the text to the correct state even though the background is being updated. 

This is a **regression**: using the same example with fyne v2.0.4 (specifically, 2d4915c74c3c5ef497b22ffe9839e8f1ad4c01cf ) or ~with a recent commit c91d36c2c8fcd231f3104e4b967a309211753e65~ (UPD: happens on a recent commit as well if I enlarge the window vertically) does not lead to persistent visual artifacts. In those versions I can glimpse very similar distorted texts while scrolling very fast, but only momentarily, they are fixed the moment I stop scrolling.

Surprisingly, sometimes the visual glitch is fixed (as in: one moment you see a glitch, the other moment you see the correct text) without performing any invocations of `updateFunc` between those two moments. You can check this by uncommenting a logging line in the example code and watching the output in the terminal window. So perhaps this has something to do with visual cache.

I've attached the example code and a screenshot showing visual glitches on that example.

### To Reproduce:
1. Compile the example code on develop 0b8f09ed4847d6565895d794de8cc951912d886d (same for any commit starting from, including, a113cafeb52490f4a20d50611bdb3354ba43f5c3 )
2. Run it and scroll very fast.
3. You may see the visual glitches when you stop scrolling. Scrolling a bit further usually fixes the problem, but just waiting or hovering the cursor does not always work.

### Screenshots:

Screen Shot 2021-10-13 at 15 44 34

### Example code:
```go
package main

import (
	"fmt"
	"log"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/canvas"
	"fyne.io/fyne/v2/theme"
	"fyne.io/fyne/v2/widget"
)

func main() {
	app := app.New()
	w := app.NewWindow("List with many lines")

	recurringNames := []string{"aaaaaaaaa", "bb", "NNNNNNNNNNNNNNN"}

	list := widget.NewList(
		func() int {
			return 1000
		},
		func() fyne.CanvasObject {
			return canvas.NewText("", theme.ForegroundColor())
		},
		func(id widget.ListItemID, o fyne.CanvasObject) {
			o.(*canvas.Text).Text = fmt.Sprintf("%s %d", recurringNames[id%3], id)
			o.Refresh()
			// log.Printf("Called Refresh() on item ID %d\n", id)
		},
	)

	w.SetContent(list)
	w.Resize(fyne.NewSize(float32(500), float32(300)))
	w.ShowAndRun()
}
```

 - **OS:** MacOS
 - **Version:** 10.12 Sierra
 - **Fyne version:** 0b8f09ed4847d6565895d794de8cc951912d886d

nullst:
### Description:
This fixes #2548 as described in the long [comment](https://github.com/fyne-io/fyne/issues/2548#issuecomment-944947591) in the discussion. 

Introduces a new public method `EstimateLength()` to unbounded channels such as `refreshQueue` in `internal/driver/common/canvas.go`.

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.


nullst:
Note: I think the simpler strategy suggested by @changkun in the comments to the issue is also possible. Instead of creating an EstimateLength() method on unbounded channels, just keep the (atomic) length of the refresh queue in the canvas object. Computing it manually is easy since there is only one producer (Refresh()) and only one consumer (FreeDIrtyTextures()). Since Changkun implemented this idea, maybe his simpler version should be used? My solution is a bit overengineered, I guess.

andydotxyz:
It is very exciting that this fixes some very difficult to debug problems in renderers and cache/re-use.
This rocks. But I will take the steer from @changkun on what we should do.
This passes code review, but is it the right solution?
I would like to prep the v2.1.x release tomorrow so we have about 20 hours to agree on the impl :)

changkun:
@nullst Thanks for tagging me here again. Feel free to commit what's converged in the discussion of https://github.com/fyne-io/fyne/issues/2548 as I think you take the credit :)

PR #2564, opened 2021-10-19
Resolve multiple data races (for discussion)

changkun:
### Description:

The rendering thread and the user thread can race on the properties of widgets. This PR shows a very basic draft of fixing all potential data races in a basic example provided in #2548, and demonstrates potential new APIs or changes (not only from fyne side but also from user side) that might be needed to address the long observed data races for all APIs reported in #2509 and #1028. 

For instance, with this PR, users should update their code in this way:

```
package main

import (
	"fmt"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/canvas"
	"fyne.io/fyne/v2/theme"
	"fyne.io/fyne/v2/widget"
)

func main() {
	w := app.New().NewWindow("List with many lines")
	ts := []string{"aaaaaaaaa", "bb", "NNNNNNNNNNNNNNN"}

	w.SetContent(widget.NewList(
		func() int { return 1000 },
		func() fyne.CanvasObject {
			return canvas.NewText("", theme.ForegroundColor())
		},
		func(id widget.ListItemID, o fyne.CanvasObject) {
			// o.(*canvas.Text).Text = fmt.Sprintf("%s %d", ts[id%3], id) // Data race!
			o.(*canvas.Text).SetText(fmt.Sprintf("%s %d", ts[id%3], id))  // OK
			o.Refresh()
		},
	))
	w.Resize(fyne.NewSize(float32(500), float32(300)))
	w.ShowAndRun()
}

```

Updates #1028
Updates #2509 
Updates #2548

### DIFF
--- a/canvas/text.go
+++ b/canvas/text.go
@@ -25,12 +25,30 @@ type Text struct {
 // MinSize returns the minimum size of this text object based on its font size and content.
 // This is normally determined by the render implementation.
 func (t *Text) MinSize() fyne.Size {
-	return fyne.MeasureText(t.Text, t.TextSize, t.TextStyle)
+	t.propertyLock.Lock()
+	text := t.Text
+	t.propertyLock.Unlock()
+	return fyne.MeasureText(text, t.TextSize, t.TextStyle)
 }
 
 // SetMinSize has no effect as the smallest size this canvas object can be is based on its font size and content.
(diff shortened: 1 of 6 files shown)


changkun:
I think this PR served its purpose. Let's close it for now. I think will focus on getting the proposals done first (quite some busy months before).

PR #2566, opened 2021-10-19
Ensure atomicity of interface assignments in Label bindings

### LINKED ISSUE #2549: Scrolling list bound to data programmatically causes nil pointer dereference
### Describe the bug:

When scrolling a list bound to data using the keyboard programmatically, this results in a nil pointer dereference. This is after the list has been bound to an existing slice of string.

Sometimes this works, but more often than not I run into this issue, since keyboard scrolling using the arrow keys did not seem to be supported.

I have seen this behavior in the middle of list navigation in this manner, as well as at the end and near the beginning. I also see this when using the mouse wheel to scroll.

### To Reproduce:

declare a list and populate it with data (50-100 elements)

use a function like the one included below to map the keydowns to a scroll, where the keyup function will just disable the bool used to keep the scroll going, this function is called from a goroutine.

### Example code:
```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
	"time"
)

var list *widget.List
var ScrollStop bool
var slice []string
var count int

func Scroll(key fyne.KeyEvent, loc int) {
	time.Sleep(200 * time.Millisecond)
	for ScrollStop {
		time.Sleep(50 * time.Millisecond)
		switch key.Name {
		case fyne.KeyDown:
			loc += 1
			if loc > len(slice)-1 {
				ScrollStop = false
				break
			}
			list.Select(loc)
		case fyne.KeyUp:
			loc -= 1
			if loc < 0 {
				ScrollStop = false
				break
			}
			list.Select(loc)
		}
	}
}

func up(key *fyne.KeyEvent) {
	switch key.Name {
	case fyne.KeyDown:
		fallthrough
	case fyne.KeyUp:
		ScrollStop = false
	}
}

func down(key *fyne.KeyEvent) {
	switch key.Name {
	case fyne.KeyDown:
		list.Select(count + 1)
		ScrollStop = true
		go Scroll(*key, count+1)
	case fyne.KeyUp:
		list.Select(count - 1)
		ScrollStop = true
		go Scroll(*key, count-1)
	}
}

func selected(index int) {
	if index < 0 {
		index = 0
	} else if index > list.Length()-1 {
		index = list.Length() - 1
	}
	count = index
}

func main() {
	ScrollStop, count = false, 0

	slice = []string{}
	var data binding.ExternalStringList

	for i := 0; i < 100; i++ {
		slice = append(slice, "word")
	}

	a := app.NewWithID("com.vancise.listtest")
	w := a.NewWindow("List Test")

	data = binding.BindStringList(
		&slice,
	)

	list = widget.NewListWithData(data,
		func() fyne.CanvasObject {
			return widget.NewLabel("template")
		},
		func(i binding.DataItem, o fyne.CanvasObject) {
			o.(*widget.Label).Wrapping = fyne.TextTruncate
			o.(*widget.Label).Bind(i.(binding.String))
		})
	list.OnSelected = selected

	view := container.NewBorder(
		container.NewVBox(
			widget.NewLabel("List Test"),
			widget.NewSeparator()),
		nil,
		nil,
		nil,
		list)

	list.Select(0)

	if deskCanvas, ok := w.Canvas().(desktop.Canvas); ok {
		deskCanvas.SetOnKeyDown(down) //for monitoring navigation of the list in inquire mode
		deskCanvas.SetOnKeyUp(up)
	} else {
		panic("not desktop canvas")
	}

	w.Resize(fyne.NewSize(940, 660))
	w.SetContent(view)
	w.ShowAndRun()
}
```

### Error Report
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x415fce2]

goroutine 15 [running]:
fyne.io/fyne/v2/data/binding.(*boundExternalStringListItem).Get(0x0)
	<autogenerated>:1 +0x22
	/Users/nickvancise/go/pkg/mod/fyne.io/fyne/v2@v2.1.0/widget/label.go:148 +0x37
fyne.io/fyne/v2/data/binding.(*listener).DataChanged(0x1)
	/Users/nickvancise/go/pkg/mod/fyne.io/fyne/v2@v2.1.0/data/binding/binding.go:55 +0x1a
fyne.io/fyne/v2/data/binding.queueItem.func1.1()
	/Users/nickvancise/go/pkg/mod/fyne.io/fyne/v2@v2.1.0/data/binding/queue.go:19 +0x42
created by fyne.io/fyne/v2/data/binding.queueItem.func1
	/Users/nickvancise/go/pkg/mod/fyne.io/fyne/v2@v2.1.0/data/binding/queue.go:17 +0x45
Saving session...
...copying shared history...
...saving history...truncating history files...
...completed.

[Process completed]
```

 - OS: MacOS BigSur 11.6
 - go version go1.17 darwin/amd64
 - required fyne.io/fyne/v2 v2.1.0 in go mod file

Any help on debugging this problem is appreciated!

nullst:
### Description:

The Bind/Unbind API for widgets usually saves a copy of a `binding.DataItem` value provided by user, and then in Unbind() sets the saved value to `nil`. Since interface assignments are not atomic in go, these two assignments may clash during concurrent calls and produce an invalid value. Kind of a chimera: interface values in go are just pairs of pointers, so one pointer can be set before a concurrent assignment, but another pointer after.

I think this is not just a theoretical issue, but actually happens in practice and leads to crashes, see https://github.com/fyne-io/fyne/issues/2549#issuecomment-946675604 for a discussion of a case where I think a non-nil interface value with nil underlying pointer appears out of nowhere as this kind of chimera and causes a nil pointer dereference. I am not 100% sure that this is what happens, but on my machine modifying Label binding code to use atomic load/store for `Label.textSource` field makes it impossible to replicate the linked issue.

**OBSOLETE DESCRIPTION (you may skip this):** In this pull request I turned `Label.textSource` and `Label.textListener` into atomic values. The code is a bit ugly because `atomic.Value` explicitly panics if you try to store a nil value. So to simulate `label.textSource = nil` I wrapped not just binding.String, but a `struct{isSet bool, source binding.String}` in the atomic value, with a pair `{false, nil}` playing the role of nil.

**NEW DESCRIPTION**: In this pull request I abstracted the usage pattern needed for Bind/Unbind in Label code. I created `binding.BasicBinder` which is basically a pair (data item, callback) which exposes an interface for transparently changing each element of the pair independently. So you can switch the data item but have the same callback, which is exactly what Label (or any other widget except list) wants to do. I've added two concurrency-safe implementations for BasicBinder, one for go1.17 that uses atomic values (the CompareAndSwap method on atomic.Value only added in go1.17), and another for go <1.17 that uses mutexes, but seems to be noticeably slower (though I didn't do any benchmarks).

This fixes #2549 on my machine.

**Note:** all other widgets implementing Bind/Unbind are also theoretically susceptible to the same data corruption issue due to non-atomicity of interface assignments. So perhaps this should be fixed in all widgets. But if we assume that the only corruption that may happen is a mismatch of two pointers forming the underlying interface value (probably not guaranteed in a real world), I think only labels and entries can lead to crashes since strings are pointers.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/bind_helper.go
+++ b/widget/bind_helper.go
@@ -15,13 +15,6 @@ type basicBinder struct {
 	dataListenerPair     annotatedListener // access guarded by dataListenerPairLock
 }
 
-// SetCallback replaces the function to be called when the data changes.
-func (binder *basicBinder) SetCallback(f func(data binding.DataItem)) {
-	binder.callbackLock.Lock()
-	binder.callback = f
-	binder.callbackLock.Unlock()
-}
-
 // Bind replaces the data item whose changes are tracked by the callback function.
 func (binder *basicBinder) Bind(data binding.DataItem) {
 	listener := binding.NewDataListener(func() { // NB: listener captures `data` but always calls the up-to-date callback
@@ -42,6 +35,13 @@ func (binder *basicBinder) Bind(data binding.DataItem) {
 	binder.unbindLocked()
 	binder.dataListenerPair = listenerInfo
 	binder.dataListenerPairLock.Unlock()
+}
+
+// SetCallback replaces the function to be called when the data changes.
+func (binder *basicBinder) SetCallback(f func(data binding.DataItem)) {
+	binder.callbackLock.Lock()
+	binder.callback = f
+	binder.callbackLock.Unlock()
 }
 
 // Unbind requests the callback to be no longer called when the previously bound
--- a/widget/label.go
+++ b/widget/label.go
@@ -126,6 +126,18 @@ func (l *Label) Unbind() {
 	l.binder.Unbind()
 }
 
+func (l *Label) syncSegments() {
+	l.provider.Wrapping = l.Wrapping
+	l.provider.Segments = []RichTextSegment{&TextSegment{
+		Style: RichTextStyle{
+			Alignment: l.Alignment,
+			Inline:    true,
+			TextStyle: l.TextStyle,
+		},
+		Text: l.Text,
+	}}
+}
+
 func (l *Label) updateFromData(data binding.DataItem) {
 	if data == nil {
 		return
@@ -140,16 +152,4 @@ func (l *Label) updateFromData(data binding.DataItem) {
 		return
 	}
 	l.SetText(val)
-}
-
-func (l *Label) syncSegments() {
-	l.provider.Wrapping = l.Wrapping
-	l.provider.Segments = []RichTextSegment{&TextSegment{
-		Style: RichTextStyle{
-			Alignment: l.Alignment,
-			Inline:    true,
-			TextStyle: l.TextStyle,
-		},
-		Text: l.Text,
-	}}
 }


andydotxyz:
Thanks for working on this. Two quick questions:
1) Can it be done without exporting new APIs? We try not to add functions unless they really add developer functionality.
2) Do we need to maintain two versions for go 1.17 or before? Would it not be easier (and avoid duplication) to include the old version for now and a note that says once we require only Go 1.17 that it should be updated?

Issue #2562, opened 2021-10-19
Form validator


Cookie04DE:
### Is your feature request related to a problem? Please describe:

Sometimes validating each entry of a form on its own isn't enough.

Imagine the following scenario:
A form with two entries for numbers and the requirement that those numbers added up can't be above ten.
 
### Is it possible to construct a solution with the existing API?

You can work around the issue by adding the validator mentioned above to all relevant entries. This is especially difficult if those entries already have validators on their own. For example if the first number can't be two and the second one needs to be above four.

### Describe the solution you'd like to see:

A form wide validator would solve the problem. It would be of the type `func() error` and be evaluated every time one entry, checkbox etc. is changed.

If the function returns an error it should prevent the user from submitting the form (by disabling the submit button) and displaying the error message in red below the form similar to how an error on validating an entry is displayed.

andydotxyz:
Let us suppose instead that you have a validation that confirms that an entry is non-nil if a checkbox is checked.
On which object does this belong? which is checked first? When the checkbox changes you must fire the entry changed etc - this is not the same thing as multiple validations on a single item.
I'm not sure "a bunch of new functions" is really required - a `Form.Validator` field would surely suffice?

Issue #1268, opened 2020-08-27
META: List of validators to consider adding


Jacalz:
### Is your feature request related to a problem? Please describe:
Input validation is a very important part of application design in order to create secure applications. We should therefore provide more out of the box validators for developers to utilize inside their code.

### Is it possible to construct a solution with the existing API?
Yes. See #1267 for moving them out of the top level package.

### Describe the solution you'd like to see:
Please comment on this issue with example of validators that you think would make sense. We can update this list as we go.

#### Validators to consider:

- [ ] Email
- [ ] URL
- [ ] Phone numbers
- [ ] String length
- [ ] Custom password parameters

stuartmscott:
- [ ] List validator - for applying multiple validation criteria to an entry

andydotxyz:
Maybe ValidatorList sounds more like a group of validations than ListValidator which sounds like checking input is a list.

Alternatively (as it’s now in ‘data/validator’) how about ‘validator.RequireAll’?

fpabl0:
This would be a great addition. Until now, I have been using my own package with common validators to develop Fyne apps faster, but there would be really good if it comes by default :)

For `ValidatorList`, I use:
```go
validator.NewGroup(validators ...fyne.StringValidator) fyne.StringValidator
```

Then for common ones:

```go
validator.NotEmpty() // should be better Require ?
validator.Email()
validator.MinLength(min int)
validator.MaxLength(max int)
```

PR #2567, opened 2021-10-20
Add an AllString validator for chaining together string validators

Jacalz:
### Description:

This adds a new group validator for chaining together validators.
It is one of the validators mentioned in #1268 and might also (at least partially) help resolve the use case in #2562.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/validation/all.go
+++ b/data/validation/all.go
@@ -4,11 +4,11 @@ import (
 	"fyne.io/fyne/v2"
 )
 
-// NewAllString creates a validator that requires all of the passed string validators to pass.
+// NewAllStrings creates a validator that requires all of the passed string validators to pass.
 // In short, it combines multiple string validators into one.
 //
 // Since: 2.2
-func NewAllString(validators ...fyne.StringValidator) fyne.StringValidator {
+func NewAllStrings(validators ...fyne.StringValidator) fyne.StringValidator {
 	return func(text string) error {
 		for _, validator := range validators {
 			if err := validator(text); err != nil {
--- a/data/validation/all.go
+++ b/data/validation/all.go
@@ -4,11 +4,11 @@ import (
 	"fyne.io/fyne/v2"
 )
 
-// NewGroup creates a validator that groups together multiple validators into one.
-// All validators need to pass for the result to be valid.
+// NewAll creates a validator that requires all of the passed validators to pass.
+// In short, it combines multiple validators into one.
 //
 // Since: 2.2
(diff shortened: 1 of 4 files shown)


Jacalz:
I don't know if there is a use case for a "pass any" but if there is, we can certainly add one then. I agree that "All" is a better naming scheme and can work better with similar names in the future.

Jacalz:
> One last thought - `validation.NewAllString` vs `validation.AllStrings` - I know we normally used `New` in this package for new validators, but in this case we could probably argue it's not creating a new one? Just a thought.

It is kind of creating a new validator from the validators that you passed, so I think it kind of makes sense as it is?

andydotxyz:
Not sure that the singular form (no “s”) reads correctly does it?

Jacalz:
> Not sure that the singular form (no “s”) reads correctly does it?

Hmm, yes and no. I went with "String" because it takes string validators and returns just one string validator, which as it happens, also only takes one string as an argument. With "Strings" one could think that it is a validator that takes more than one string? I'm not sure what's best?

andydotxyz:
It is a tough one to know if we're getting it right. I think mostly "all string" doesn't read too well due to the order of words - how can you have "all" of a singular "string"... (thinking about group again "string group" flows, but "group string" doesn't - it would be "group of strings").

I am trying to explore this myself to see if anything becomes clear. My latest question is to compare with the hypothetical "Any" addition. So `validation.AnyString` does make sense, because it is asking for any one to check - but on that logic `validation.AllStrings` needs the "s" to communicate that many will be used to pass validation... I think.

Maybe someone with a fresh set of eyes could look over the various challenges / suggestions.

Jacalz:
I changed the name to `AllStrings`. Didn't seem like anyone else wanted to chime in on the matter.

stuartmscott:
> Didn't seem like anyone else wanted to chime in on the matter.

I did, but it was ignored - https://gophers.slack.com/archives/CK5U4BU87/p1635689185251500

Jacalz:
> > Didn't seem like anyone else wanted to chime in on the matter.
> 
> 
> 
> I did, but it was ignored - https://gophers.slack.com/archives/CK5U4BU87/p1635689185251500

Oh, sorry. I must have missed that at the time. However, I think we have already thought about that before. I had hoped that more people would jump in with concrete answers.

andydotxyz (approved):
I think the current naming works for what @stuartmscott was suggesting as well (i.e. `NewAllFloats`

PR #2580, opened 2021-10-23
geometry addition helpers

toaster:
### Description:

This PR introduces new `Add*` helpers to `fyne.Size` and `fyne.Position`.
With these helpers it is no longer necessary to create new  temporary vectors (`Size`, `Position`) just to add to an existing one. The benchmarks only cover the runtime of the methods themselves and do not consider the allocation time for the temporary vectors. The numbers would be even worse for the old approach then :).

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/geometry.go
+++ b/geometry.go
@@ -99,6 +99,21 @@ func (s Size) Add(v Vector2) Size {
 	return Size{s.Width + w, s.Height + h}
 }
 
+// AddHeight returns a new Size by adding height to the current one.
+func (s Size) AddHeight(height float32) Size {
+	return Size{s.Width, s.Height + height}
+}
+
+// AddWidth returns a new Size by adding width to the current one.
+func (s Size) AddWidth(width float32) Size {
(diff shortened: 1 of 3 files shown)


Jacalz:
> Seems sensible to me. Do we really need the benchmarks committed in though? I doesn't seem likely to me that we are going to want to compare the performance again in my opinion, especially if it is mostly about API usability and not primarily for performance.

@toaster I think you might have missed this comment. I understand why, because it was sent as a question and not a requesting changes. Would you want to do this as part of a separate PR?

toaster:
@Jacalz sorry for not answering the question. I didn’t miss it but simply forgot to write my thoughts on it:

I think the benchmarks document the performance gain and therefore should be part of the PR even if it is not the main reason for the convenience methods.
Plus, they are actually used to compare the different implementations in a regular test.

PR #2586, opened 2021-10-24
Add setters to dialogs

### LINKED ISSUE #1566: [FEATURE REQUEST] Change Title/Message in a Dialog
I have some long-running interactions that use `ProgressInfiniteDialog` and it would be great to simply be able to change the message shown as progress continued, instead of hiding the dialog and creating a completely new one every time a new stage of the process begins.

No

Consider indeterminate action with multiple steps, such as a program installation or download+extract.

```golang
total, progress, err := StartDownload("...")

// assume "total" was 0/-1/undefined for this example (e.g: webserver didn't tell us)
dlg := dialog.NewProgressInfinite("Downloading Update", "Starting download...", w)
select {
  case n, done := <-progress:
    if done { break }
    dlg.SetMessage(fmt.Sprintf("%d bytes retrieved...", n))
}

dlg.SetTitle("Extracting Update")
dlg.SetMessage("Starting file extraction...")
select {
  case f, done := <- events:
    if done { break }
    dlg.SetMessage(fmt.Sprintf("Extracting %s...", f)
}
```

andydotxyz:
Allowing dialogs to have content updated when shown.
First pass is SetTitle - easy because it applies to all.

Next has been requested `SetMessage` which applies to some.
Happy to get feedback about how we should expose this and if other setters should be added too

Fixes #1566

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/base.go
+++ b/dialog/base.go
@@ -22,11 +22,11 @@ const (
 type Dialog interface {
 	Show()
 	Hide()
-	SetDismissText(label string)
-	SetOnClosed(closed func())
+	SetDismissText(string)
+	SetOnClosed(func())
 	SetTitle(string)
 	Refresh()
-	Resize(size fyne.Size)
+	Resize(fyne.Size)
 
 	// Since: 2.1
 	MinSize() fyne.Size
@@ -152,6 +152,7 @@ func (d *dialog) SetOnClosed(closed func()) {
 	}
 }
 
+// SetTitle allows developers to set a new title for this dialog.
 func (d *dialog) SetTitle(title string) {
 	d.title = title
 	d.label.SetText(title)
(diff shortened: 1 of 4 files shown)


dialog/base.go
```diff
@@ -22,10 +22,11 @@ const (
...
 	Show()
 	Hide()
-	SetDismissText(label string)
-	SetOnClosed(closed func())
+	SetDismissText(string)
+	SetOnClosed(func())
+	SetTitle(string)
 	Refresh()
-	Resize(size fyne.Size)
+	Resize(fyne.Size)
```
Jacalz, line 29: I don't know to what extent people create truly custom dialogs using the dialog interface (if it is even possible), but isn't this a breaking change?

andydotxyz: Yes I guess that is strictly true. I suppose I need another approach.
The alternative is to expose new concrete types I suppose (better than many new `MutableDialog` interfaces).

andydotxyz:
I am closing this PR as @Jacalz has plans for a much better approach which will be underway soon.

PR #2587, opened 2021-10-25
Fix shortcuts not working in non-ASCII keyboard layouts

### LINKED ISSUE #1220: Copy-paste via keyboard don't work translated keyboard mappings on Windows.

nullst:
### Description:

People who use Russian language (or many other languages with primarily non-ASCII symbols) typically switch between several keyboard layouts throughout the day, for example between English layout and a Russian layout. The same physical key will correspond to both English symbol (e.g., "V") and Russian symbol ("М" for the most common Cyrillic layout; that's not the same symbol as English M, it just looks the same). Normally, people expect the physical key combination corresponding to "Ctrl+V" in English to also work in Russian (otherwise muscle memory goes haywire if you switch between the keyboard layouts often). This is how it usually works in Windows, Mac, and well-developed graphical toolkits on Linux. Note that this is not what people with AZERTY do, since they usually stick to a single keyboard layout and type both French and English with it.

This used not to work in Fyne, and this pull request fixes it: when a user presses an unknown-to-fyne key (like Russian "Ф") with a Control/Alt modifier, the `triggersShortcut` function will try to replace the unknown key with the equivalent ASCII symbol. I'm not sure if the method I use for translation actually works on all platforms, but at least it seems to work on my Mac OS Sierra.

Fixes #1220 .

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -1292,8 +1292,10 @@ func (w *window) triggersShortcut(localizedKeyName fyne.KeyName, key glfw.Key, m
 	// User pressing physical keys Ctrl+V while using a Russian (or any non-ASCII) keyboard layout
 	// is reported as a fyne.KeyUnknown key with Control modifier. We should still consider this
 	// as a "Paste" shortcut.
+	// See https://github.com/fyne-io/fyne/pull/2587 for discussion.
 	keyName := localizedKeyName
-	if (localizedKeyName == fyne.KeyUnknown) && (modifier == ctrlMod) {
+	resemblesShortcut := (modifier&(desktop.ControlModifier|desktop.SuperModifier) != 0)
+	if (localizedKeyName == fyne.KeyUnknown) && resemblesShortcut {
 		if asciiKey, ok := keyCodeMapASCII[key]; ok {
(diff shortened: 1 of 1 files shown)


toaster (approved):
Yeah, looks good!

I’ve also gained some knowledge (by starting my Windows machine again ;): `Alt+<Key>` is valid on Windows and I think it’s not that uncommon there. Therefore, it definitely would be a bad decision to forbid it. However, a dev using it has to be aware that it breaks things in one way ore another on Mac at least.

andydotxyz:
Thanks!

PR #2591, opened 2021-10-26
Use basicBinder helper in Check, Entry, ProgressBar, and Slider widgets

nullst:
### Description:

The current implementation of Bind/Unbind API in widgets such as Check, Entry, ProgressBar, and Slider is not thread-safe. As an example, all those widgets store the current data item and the registered listener in order to call `dataItem.RemoveListener(listener)` in Unbind(). If a concurrent call to `Bind(newDataItem)` is in progress, this may lead to RemoveListener being called on the `newDataItem`, but with the old listener value, which means that the previous listener will never be removed from the previous data item.

Previously Label widget had the same deficiencies, but they were fixed in #2566 by creating a helper type called "basicBinder" in the widget package which encapsulates basic bind logic. The same helper can be used in all other widgets. It prevents some concurrency issues like the one described above. Moreover, it simplifies the Bind/Unbind implementation. It is enough to define two callbacks:

* `thisWidget.updateFromData(data binding.DataItem)` that reads the current value from data and updates the widget to reflect it;
* `thisWidget.writeData(data binding.DataItem)` that writes the current state of the widget to the given data item.

As such, no direct manipulations with AddListener/RemoveListener are needed, which reduced the potential for concurrency bugs.

(Technical note: basicBinder is not an interface, the names are not relevant. Making basicBinder into an interface is possible, but that would expose methods like binder.SetCallback that should be private to the widget. I think storing a basicBinder as a field in the widget struct is a proper solution.)

As far as I know, the concurrency issues with those widgets aren't reported as bugs, but an issue like #2549 would occur as soon as someone creates a bound list with 1000 checkboxes instead of 1000 labels.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/bind_helper.go
+++ b/widget/bind_helper.go
@@ -37,6 +37,15 @@ func (binder *basicBinder) Bind(data binding.DataItem) {
 	binder.dataListenerPairLock.Unlock()
 }
 
+// CallWithData passes the currently bound data item as an argument to the
+// provided function.
+func (binder *basicBinder) CallWithData(f func(data binding.DataItem)) {
+	binder.dataListenerPairLock.RLock()
+	data := binder.dataListenerPair.data
+	binder.dataListenerPairLock.RUnlock()
+	f(data)
+}
+
 // SetCallback replaces the function to be called when the data changes.
 func (binder *basicBinder) SetCallback(f func(data binding.DataItem)) {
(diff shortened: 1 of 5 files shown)


andydotxyz (approved):
Nice way to use the new safer code, thanks :)

PR #2600, opened 2021-10-31
Adding Must methods to binding API

### LINKED ISSUE #2599: Binding "Must*" variants

Struct binding is a little clunky to use with all of the error handling involved. First the user must bind to the struct, and then they'll want to bind to fields within the struct. Consider this example.
https://play.golang.org/p/Afw8NqQv6lR

Another case shown in the playground link is when manipulating a bound value with `Get` and `Set`. In this case it's not immediately clear to the user why either method *should* return an error in the first place, but an error returned must be an error handled. This is troublesome enough that even fyne examples don't always handle this error, and the user should be able to opt out of handling it in favor of panicking.

An idiomatic way to accomplish this behavior could be borrowed from something like the regexp package in the standard library. It has "Must" variants of functions for situations where the user knows that if the regex fails to compile, something critical will not work, so the best option is to panic.
https://pkg.go.dev/regexp#MustCompile

The current solution is the problem. Errors must be handled - in some cases it's not clear why - and there should be an alternative.

Provide "Must" variants of binding interaction functions. See the go playground link above for comments showing suggestions with struct binding and working with field binding Get/Set.

drognisep:
### Description:

Adding Must variants to binding API

Fixes #2599 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.
  - Looks like there are some random benchmark failures locally

### DIFF
--- a/data/binding/binding.go
+++ b/data/binding/binding.go
@@ -99,7 +99,9 @@ func (b *base) trigger() {
 type Untyped interface {
 	DataItem
 	Get() (interface{}, error)
+	MustGet() interface{}
 	Set(interface{}) error
+	MustSet(interface{})
 }
 
 // NewUntyped returns a bindable interface{} value that is managed internally.
@@ -124,6 +126,14 @@ func (b *boundUntyped) Get() (interface{}, error) {
 	return b.val.Interface(), nil
 }
 
+func (b *boundUntyped) MustGet() interface{} {
+	val, err := b.Get()
+	if err != nil {
+		panic(err)
+	}
+	return val
+}
+
 func (b *boundUntyped) Set(val interface{}) error {
 	b.lock.Lock()
 	defer b.lock.Unlock()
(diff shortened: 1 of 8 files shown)


drognisep:
@andydotxyz I'm not sure what's failing with static analysis. What am I missing there?

drognisep:
@andydotxyz brought up some good points on this idea. This should be addressed with documentation, not panic options.
* As a rule things shouldn't panic in production. There are cases where an actual error would never be returned, and the error can be assigned to `_` in that case.
* Errors can convey legitimate conversion and validation problems and those should be handled appropriately and reported to the user rather than panicking.

All in all, the option to panic isn't more useful than ignoring a particular error completely (and can certainly be an anti-pattern instead), and it should be documented when errors will and will not be returned (and why) so the developer has better awareness.

andydotxyz:
> @andydotxyz I'm not sure what's failing with static analysis. What am I missing there?

It looks like we had a merge conflict that slipped in - we recently updated to needing the new format for build tags on develop and I think some open PRs didn't have that yet. I will go check develop is consistent so new PRs don't hit this again.

andydotxyz:
> I will go check develop is consistent so new PRs don't hit this again.

Yeah I was right, this is now fixed.

drognisep:
> > I will go check develop is consistent so new PRs don't hit this again.
> 
> Yeah I was right, this is now fixed.

Thank you! :)

PR #2609, opened 2021-11-03
widget: fix data race in bind test

changkun:
### Description:

This PR fixes the data races in the widget bind test.

```
go test -v -race -run=TestBasicBinder
```

Updates #1028
Updates #2509

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/bind_helper_test.go
+++ b/widget/bind_helper_test.go
@@ -1,43 +1,49 @@
 package widget
 
 import (
+	"sync/atomic"
 	"testing"
 
 	"fyne.io/fyne/v2/data/binding"
 	"github.com/stretchr/testify/assert"
 )
 
+type atomicInt struct{ val int64 }
+
+func (v *atomicInt) Set(val int) { atomic.StoreInt64(&v.val, int64(val)) }
+func (v *atomicInt) Get() int    { return int(atomic.LoadInt64(&v.val)) }
+
 func TestBasicBinder(t *testing.T) {
 	var binder basicBinder
-	var value int = -1
+	value := &atomicInt{-1}
 
 	item1 := binding.NewInt()
 
 	callback := func(data binding.DataItem) {
 		val, err := data.(binding.Int).Get()
 		assert.NoError(t, err)
-		value = val
+		value.Set(val)
 	}
 	binder.SetCallback(callback)
 
-	assert.Equal(t, -1, value)
+	assert.Equal(t, -1, value.Get())
 	item1.Set(100)
 	binder.Bind(item1)
 	waitForBinding()
-	assert.Equal(t, 100, value) // callback on the initial value
+	assert.Equal(t, 100, value.Get()) // callback on the initial value
 	item1.Set(150)
 	waitForBinding()
-	assert.Equal(t, 150, value) // callback on the changed value
+	assert.Equal(t, 150, value.Get()) // callback on the changed value
 
 	item2 := binding.NewInt()
 	item2.Set(200)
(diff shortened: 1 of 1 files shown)


widget/bind_helper_test.go
```diff
@@ -1,43 +1,49 @@
...
 	"testing"
 
 	"fyne.io/fyne/v2/data/binding"
 	"github.com/stretchr/testify/assert"
 )
 
+type atomicInt struct{ val int64 }
+
+func (v *atomicInt) Set(val int) { atomic.StoreInt64(&v.val, int64(val)) }
+func (v *atomicInt) Get() int    { return int(atomic.LoadInt64(&v.val)) }
```
andydotxyz, line 14: using `atomic.LoadInt64` will crash on 32bit as we saw.
Is it your intention to limit the unit tests to 64bit machines?

changkun: I think we won't have a chance to run tests on 32bit machine, would we?

Jacalz: Apart from on macOS, you should be able to compile and run 32bit binaries on a 64-but machine. We could try running them in our tests as well.

andydotxyz: I was also thinking that it means someone on a 32bit machine looking to compile would be restricted from getting test passing. It does not seem desirable. 
apart from that I suppose it means we can’t use a quick grep any more to make sure we are not using LoadInt64!

changkun: I feel horrible about thinking in a way to prevent people from using 64bit functions.
Furthermore, I could not even pass all tests on my 64bit machine today but now an argument about a single test may not pass on a 32bit machine is not desirable!

The sync/atomic doc only says:

- On 386, the 64-bit functions use instructions unavailable before the Pentium MMX.
- On non-Linux ARM, the 64-bit functions use instructions unavailable before the ARMv6k core.
- On ARM, 386, and 32-bit MIPS, it is the caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically. The first word in a variable or in an allocated struct, array, or slice can be relied upon to be 64-bit aligned.

https://pkg.go.dev/sync/atomic@go1.17.2#pkg-note-BUG

This means:

1. Only ARMs before  Pentium MMX and ARMv6k is not usable
2. If the developer properly handles the struct alignment, 64-bit atomics can still work properly on ARM, 386, and 32-bit MIPS.

BTW: The runtime garbage collector is using [atomic.Loadint64](https://cs.opensource.google/go/go/+/master:src/runtime/mgcmark.go;l=418?q=LoadInt64&ss=go%2Fgo&start=21), does that mean we can't even use Go at all??

Jacalz: They should work, you "just" need to align them correctly. I have seen people either using a [2]uint64 or putting the uint64 at the start of the struct fields.

andydotxyz: > BTW: The runtime garbage collector is using atomic.Loadint64, does that mean we can't even use Go at all??

I have no idea expertise in this area - I am just referencing the bug we worked on before that crashed apps on 32bit devices.
I would assume that the GC has some clever padding to avoid the issue, as clearly Go is working fine ok 32bit

andydotxyz:
Following the summary above about alignment it sounds like this won’t hit the same bug as before, so it could be approved (perhaps we should indeed be running 32bit tests in CI).
Now I’m wondering why it is closed…

PR #2613, opened 2021-11-03
widget: remove syncPool test

changkun:
### Description:

sync.Pool is encapsulated in the widget as syncPool for reusing CanvasObjects.
The implementation works but the test are flaky because syncPool.Put an object does not entirely guarantee the next Get will obtain a non-nil object.

Especially the test can fail under -race mode.

Remove it as the test itself is incorrect.

Updates #1028
Updates #2509

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


changkun:
I don't see there is a way to fix it. Removing them is the only way to fix it.

PR #2617, opened 2021-11-04
avoid layouts in Canvas.EnsureMinSize

### LINKED ISSUE #1655: Tree widget performance
### Describe the bug:
Tree widget triggers too many layouts, which affects performances. 
The user experience with a large tree is then not pleasant at all...

### To Reproduce:
Run the sample code below. 

For a size of 1000 entries, the `isBranch` is called over 1_000_000 times.
Investigating a bit, I found that it comes from `Tree.walkAll,` itself called in `treeContentRenderer.Layout`, which seems to be called around 1000 times.

As an aside, the List widget handles the same number of entries without any issues.

### Example code:

```go
package main

import (
	"fmt"

	"fyne.io/fyne"
	"fyne.io/fyne/app"
	"fyne.io/fyne/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Tree test")

	w.Resize(fyne.NewSize(600, 600))
	content := tree()

	w.SetContent(content)
	w.ShowAndRun()
}

func tree() fyne.CanvasObject {
	var count int
	childUIDs := func(uid widget.TreeNodeID) (c []widget.TreeNodeID) {
		fmt.Println("isBranch called ", count, "times")
		if uid == "" {
			out := make([]string, 1000)
			for i := range out {
				out[i] = fmt.Sprintf("node_%d", i)
			}
			return out
		} else if uid[0] == 'n' {
			return []string{"kid" + uid}
		} else {
			return nil
		}
	}
	createNode := func(branch bool) (o fyne.CanvasObject) {
		return widget.NewLabel("")
	}
	isBranch := func(uid widget.TreeNodeID) (ok bool) {
		count++
		return uid == "" || uid[0] != 'k'
	}
	updateNode := func(uid widget.TreeNodeID, branch bool, node fyne.CanvasObject) {
		node.(*widget.Label).SetText(uid)
	}

	return widget.NewTree(childUIDs, isBranch, createNode, updateNode)
}

```

 - OS: Linux
 - Version: Ubuntu 20.10
 - Go version:  1.15.3
 - Fyne version: 1.4.2

### Misc 
Fyne looks really promising, I'm looking forward to replace all my Go/Qt code !

benoitkugler:
### Description:
Try to avoid too many layouts in Canvas.EnsureMinSize by using a cache.

Fixes  #1655 : performance of the widget tree are then very decent 

I do not pretend this is a good solution, but the tests are passing. 

I don't really understand the logic of this function, but it seems strange to me that it is the parent which must be laid out instead of the current node CanvasObject (`objToLayout = node.parent.obj`).

 Wouldn't be possible to change this so that this behavior won't happen anymore ? (I mean, that the parent has to layout itself for each of its children)

Feel free to close if it seems to rash, or to guide me towards a better solution/understanding of the problem !

### DIFF
--- a/internal/driver/common/canvas.go
+++ b/internal/driver/common/canvas.go
@@ -65,6 +65,8 @@ func (c *Canvas) EnsureMinSize() bool {
 	}
 	var lastParent fyne.CanvasObject
 
+	seenObj := map[fyne.CanvasObject]struct{}{}
+
 	windowNeedsMinSizeUpdate := false
 	csize := c.impl.Size()
 	min := c.impl.MinSize()
@@ -93,8 +95,9 @@ func (c *Canvas) EnsureMinSize() bool {
 				}
 			}
 
-			if objToLayout != lastParent {
-				updateLayout(lastParent)
+			if _, hasSeen := seenObj[objToLayout]; !hasSeen {
+				updateLayout(objToLayout)
+				seenObj[objToLayout] = struct{}{}
 				lastParent = objToLayout
 			}
 		}
--- a/internal/driver/common/canvas_perf_test.go
+++ b/internal/driver/common/canvas_perf_test.go
@@ -1,0 +1,62 @@
+package common_test
+
+import (
+	"fmt"
+	"time"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/app"
+	"fyne.io/fyne/v2/widget"
+)
+
+const sizeTree = 1_000
+
+func Example() {
+	a := app.New()
+	w := a.NewWindow("Perf tree")
+
+	w.Resize(fyne.NewSize(600, 600))
+	content := tree()
+	w.SetContent(content)
+	time.AfterFunc(200*time.Millisecond, w.Close)
+	w.ShowAndRun()
+	fmt.Println("Number of isBranch calls:", count/100*100) // rounding
+
+	// with master version of Canvas.EnsureMinSize : 1028000
+	// with patched version of Canvas.EnsureMinSize : 29000
+
+	// Output: Number of isBranch calls: 1028000
+}
+
+var count int
+
+func tree() fyne.CanvasObject {
+	childUIDs := func(uid widget.TreeNodeID) (c []widget.TreeNodeID) {
+		if uid == "" {
+			out := make([]string, sizeTree)
+			for i := range out {
+				out[i] = fmt.Sprintf("node_%d", i)
+			}
+			return out
+		} else if uid[0] == 'n' {
+			return []string{"kid" + uid}
+		} else {
+			return nil
+		}
+	}
+	// Return a CanvasObject that can represent a Branch (if branch is true), or a Leaf (if branch is false)
+	createNode := func(branch bool) (o fyne.CanvasObject) {
+		return widget.NewLabel("")
+	}
+	// Return true if the given widget.TreeNodeID represents a Branch
+	isBranch := func(uid widget.TreeNodeID) (ok bool) {
+		count++
+		return uid == "" || uid[0] != 'k'
+	}
+	// Called to update the given CanvasObject to represent the data at the given widget.TreeNodeID
+	updateNode := func(uid widget.TreeNodeID, _ bool, node fyne.CanvasObject) {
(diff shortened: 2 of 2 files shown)


andydotxyz:
> Wouldn't be possible to change this so that this behavior won't happen anymore ? (I mean, that the parent has to layout itself for each of its children)

Potentially it could be optimised so the parent is only layed out once. This algorithm should be checking to see if any MinSize has changed, If it has then the parent must lay out again to accommodate.

Would it be possible to include a benchmark so we see how this is impacting things? I think your proposed change makes sense in the current code - though as you say there may be ways to improve the wider algorithm.

benoitkugler:
I'm not sure how to setup a proper benchmark, sorry. Is it possible to measure an "update" cycle ? 

As a first number, I measured the number of Tree.IsBranch() calls, in a program (example/main.go) which simply shows the tree and close after a small delay. 
For a 1-level tree with 1000 entries, the number of calls drops from about 1_000_000 to 29_000.

example/main.go
```diff
@@ -0,0 +1,60 @@
...
+		count++
+		return uid == "" || uid[0] != 'k'
+	}
+	// Called to update the given CanvasObject to represent the data at the given widget.TreeNodeID
+	updateNode := func(uid widget.TreeNodeID, _ bool, node fyne.CanvasObject) {
+		node.(*widget.Label).SetText(uid)
+	}
+
+	return widget.NewTree(childUIDs, isBranch, createNode, updateNode)
+}
```
Jacalz, line 60: It might be worth noting here that this wouldn't be allowed to be commited. We probably shouldn't fill the repo with a lot of random code examples. However, I do appreciate the idea :)

internal/driver/common/canvas.go
```diff
@@ -114,6 +118,64 @@ func (c *Canvas) EnsureMinSize() bool {
...
+		c.impl.Resize(csize.Max(min))
+	}
+
+	if lastParent != nil {
+		c.RLock()
+		updateLayout(lastParent)
+		c.RUnlock()
+	}
+	return windowNeedsMinSizeUpdate
+}
```
Jacalz, line 177: It's better to just add this in the original function instead of having both to benchmark in this way. I would suggest benchmarking before and after instead, using the git history to your help (i.e. benchmark on a different branch or by temporary going back to before the change).

Jacalz (changes requested):
Some idea for how to better make the PR fit for review.

Jacalz:
@benoitkugler Have you had any time to continue with this?

benoitkugler:
> @benoitkugler Have you had any time to continue with this?

Thanks for the feedback !
But no sorry I have not continued with this. I will eventually ! Should I close this PR in the meantime to avoid polluting the area ?

Jacalz:
> > @benoitkugler Have you had any time to continue with this?
> 
> Thanks for the feedback ! But no sorry I have not continued with this. I will eventually ! Should I close this PR in the meantime to avoid polluting the area ?

Great to hear! No need to close it. I was just sorting though some old PRs trying to see which were actively worked on. Feel free to ask in case you have any questions :)

benoitkugler:
So I've tried to clean things a bit. 
I'm still unsure to how properly benchmark the difference. Since the test suite does not use the desktop driver, I can't use it. 
Thus I've left the example, as an Example() function. 
However, the numbers in Output are likely to be platform dependent..

Jacalz (changes requested):
Sorry for taking some time to reach this. I'm not sure if we really should have the example in there. Wouldn't a regular `func BenchmarkX` be better perhaps? 

I think this makes a lot of sense but I don't know any better way to measure it. Maybe @changkun has a better idea on how to proceed with this?

andydotxyz:
Should we try to get this rebased onto a recent `develop`?
Looking at it again I wonder if it will break items that require a second pass for layout, such as multi-line wrapped text?

benoitkugler:
Hum honestly I'm really not sure this PR is the right way to go, more like a test. I have really no clue about the consequences on the framework.. 
So I would really don't mind if you close it.

benoitkugler:
Closing since widget tree performance has been improved.

PR #2620, opened 2021-11-05
Fix partial child URI returned by storage.List() for content: scheme …

### LINKED ISSUE #2619: storage.List doesn't return complete URI on Android for "content:" scheme
When trying to list the child of an URI (i.e. the content of the directory) which has the `content:` scheme on Android, the child URI are truncated.

In my case the file picker returns a directory the URI :
> content://com.android.externalstorage.documents/tree/primary:Android/media/Cards

And when listing the file within this directory with `storage.List()`, the children are missing half of their URI like
> primary:Android/media/Cards/test.md

Instead of
> content://com.android.externalstorage.documents/tree/primary:Android/media/Cards/test.md

 - **OS:** Android
 -  **Version:** 10
 - **Go version:**1.16.9
 - **Fyne version:** master, develop

lugu:
Fix partial child URI returned by storage.List() for content: scheme on Android

The current implementation only returns the DOCUMENT_ID instead of the full URI. Build the child URI using buildDocumentUriUsingTree.

Fixes #2619

### DIFF
--- a/internal/driver/mobile/android.c
+++ b/internal/driver/mobile/android.c
@@ -390,17 +390,18 @@ char* listContentURI(uintptr_t jni_env, uintptr_t ctx, char* uriCstr) {
 	jclass cursorClass = (*env)->GetObjectClass(env, cursor);
 	jmethodID next = find_method(env, cursorClass, "moveToNext", "()Z");
 	jmethodID get = find_method(env, cursorClass, "getString", "(I)Ljava/lang/String;");
+	jmethodID getChildURI = find_static_method(env, contractClass, "buildDocumentUriUsingTree", "(Landroid/net/Uri;Ljava/lang/String;)Landroid/net/Uri;");
 
 	char *ret = NULL;
 	int len = 0;
 	while (((jboolean)(*env)->CallBooleanMethod(env, cursor, next)) == JNI_TRUE) {
-		jstring name = (jstring)(*env)->CallObjectMethod(env, cursor, get, 0);
-		jobject childUri = (jobject)(*env)->CallStaticObjectMethod(env, contractClass, getChild, uri, name);
+		jstring childDocId = (jstring)(*env)->CallObjectMethod(env, cursor, get, 0);
+		jobject childUri = (jobject)(*env)->CallStaticObjectMethod(env, contractClass, getChildURI, uri, childDocId);
 		jclass uriClass = (*env)->GetObjectClass(env, childUri);
 		jmethodID toString = (*env)->GetMethodID(env, uriClass, "toString", "()Ljava/lang/String;");
-		jobject s = (*env)->CallObjectMethod(env, childUri, toString);
+		jstring s = (jstring)(*env)->CallObjectMethod(env, childUri, toString);
 
-		char *uid = getString(jni_env, ctx, name);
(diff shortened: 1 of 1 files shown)


internal/driver/mobile/android.c
```diff
@@ -390,17 +390,18 @@ char* listContentURI(uintptr_t jni_env, uintptr_t ctx, char* uriCstr) {
...
-		jobject childUri = (jobject)(*env)->CallStaticObjectMethod(env, contractClass, getChild, uri, name);
+		jstring childDocId = (jstring)(*env)->CallObjectMethod(env, cursor, get, 0);
+		jobject childUri = (jobject)(*env)->CallStaticObjectMethod(env, contractClass, getChildURI, uri, childDocId);
 		jclass uriClass = (*env)->GetObjectClass(env, childUri);
 		jmethodID toString = (*env)->GetMethodID(env, uriClass, "toString", "()Ljava/lang/String;");
-		jobject s = (*env)->CallObjectMethod(env, childUri, toString);
+		jstring s = (jstring)(*env)->CallObjectMethod(env, childUri, toString);
 
-		char *uid = getString(jni_env, ctx, name);
+		char *uid = getString(jni_env, ctx, s);
```
andydotxyz, line 404: Look at it more carefully I wonder if it was a typo - with just this 1 line change does it not return the expected data?
(I like the move to jstring above, but I don't know if the additional lookup method is needed)

andydotxyz (commented):
Thanks for this.
Looking at the code and our previously unused `s` I wonder if the additional lookup methods are required?

internal/driver/mobile/android.c
lugu: Absolutely, changing name into s in line 404 partially solve the original problem:

Here lies another bug: the returned URIs are "children" URIs (because built using [buildChildDocumentsUriUsingTree](https://developer.android.com/reference/android/provider/DocumentsContract#buildChildDocumentsUriUsingTree(android.net.Uri,%20java.lang.String)) instead of [buildDocumentUriUsingTree](https://developer.android.com/reference/android/provider/DocumentsContract#buildDocumentUriUsingTree(android.net.Uri,%20java.lang.String))). They are valid and accessible, but lost the original file name and extension.

An example would make this more concrete:
If the parent directory URI is:
> content://com.android.externalstorage.documents/tree/primary:Android/media/test

When fixing only the typo, the child URI becomes:
> content://com.android.externalstorage.documents/tree/primary%3AAndroid%2Fmedia%2Ftest/document/primary%3AAndroid%2Fmedia%2Ftest%2Ftest.md/children

Notice the suffix "/children".
With this URI, calling URI.Name() returns "children" and URI.Extension() does not return the original extension (".md"). If the directory contains several files, they are all named "children"... This is not convenient.

When using buildDocumentUriUsingTree, the child URI is:
> content://com.android.externalstorage.documents/tree/primary%3AAndroid%2Fmedia%2Ftest/document/primary%3AAndroid%2Fmedia%2Ftest%2Ftest.md

Now each child has a unique name and the expected extension.

PS: I wish those URIs were simpler and their path not encoded twice.

andydotxyz: Good catch, thanks for the clarification

andydotxyz (approved):

PR #2642, opened 2021-11-13
Installation capability is deprecated from go get. Need to use go install instead.

aysabzevar:
### Description:
"Starting in Go 1.17, installing executables with go get is deprecated. go install may be used instead."
Source: https://golang.org/doc/go-get-install-deprecation

The -i flag for go install is also deprecated and go install is enough.

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,13 +29,13 @@ If you're not sure if that's all installed or you don't know how then check out 
 
 Using the standard go tools you can install Fyne's core library using:
 
-    $ go get fyne.io/fyne/v2
+    $ go install fyne.io/fyne/v2
 
 # Widget demo
 
 To run a showcase of the features of Fyne execute the following:
 
-    $ go get fyne.io/fyne/v2/cmd/fyne_demo/
+    $ go install fyne.io/fyne/v2/cmd/fyne_demo/
     $ fyne_demo
 
 And you should see something like this (after you click a few buttons):
@@ -120,7 +120,7 @@ Using `go install` will copy the executable into your go `bin` dir.
 To install the application with icons etc into your operating system's standard
 application location you can use the fyne utility and the "install" subcommand.
 
-    $ go get fyne.io/fyne/v2/cmd/fyne
+    $ go install fyne.io/fyne/v2/cmd/fyne
     $ fyne install
 
 # Packaging for mobile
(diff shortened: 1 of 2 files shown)


andydotxyz:
It has not been removed, it has been deprecated. It still works.
The reason this difference matters is because we support Go 1.14, which does not support this new command in the same way.
If you think the documentation should be updated we will have to include both.

andydotxyz (changes requested):
This won’t work for all our supported versions (back to 1.14)

aysabzevar:
I am sorry @andydotxyz you are right. I didn't test it carefully.
What do you suggest?

andydotxyz:
So far we have held off making this change as listing both seemed messy.
However many people are getting confused so probably you are right we should update.
But I think we need to note that this is Go 1.16+ and keep the older docs in a section for older Go (immediately below the new instruction?)

README.md
```diff
@@ -29,13 +29,13 @@ If you're not sure if that's all installed or you don't know how then check out
 
 Using the standard go tools you can install Fyne's core library using:
 
-    $ go get fyne.io/fyne/v2
+    $ go install fyne.io/fyne/v2
```
Jacalz, line 32: For the record, I think this an incorrect change. You still use `go get` to fetch and install dependencies for your modules.

Jacalz (changes requested):
I agree with @andydotxyz on this. Also found something that seems to be wrong with the current change.

README.md
```diff
@@ -29,13 +29,13 @@ If you're not sure if that's all installed or you don't know how then check out
...
 
 # Widget demo
 
 To run a showcase of the features of Fyne execute the following:
 
-    $ go get fyne.io/fyne/v2/cmd/fyne_demo/
+    $ go install fyne.io/fyne/v2/cmd/fyne_demo/
```
Jacalz, line 38: It is probably also worth noting that this would fail on Go 1.16 and higher because it requires to add `@latest` at the end when using that command.

Jacalz:
I would suggest updating to something similar to https://github.com/fyne-io/developer.fyne.io/pull/57 if you would like to continue with this change.

Jacalz:
@aysabzevar Are you interested in continuing this?

PR #2646, opened 2021-11-15
Don't call OnScrolled if offset did not change

### LINKED ISSUE #1868: OnScrolled triggered when offset hasn't changed
`container.Scroll.OnScrolled` is called multiple times with the same offset, instead it should only be called when the offset has changed.

andydotxyz:
Fixes #1868

### Checklist:

- [x] Tests included.
- [x] Tests all pass. <- this is super strange... can't figure out why just yet

### DIFF
--- a/container/doctabs.go
+++ b/container/doctabs.go
@@ -69,10 +69,6 @@ func (t *DocTabs) CreateRenderer() fyne.WidgetRenderer {
 	r.box = NewHBox(r.create, r.action)
 	var lastX, lastY float32
 	r.scroller.OnScrolled = func(offset fyne.Position) {
-		// FIXME OnScrolled can be called when the offset hasn't changed (#1868)
-		if offset.X == lastX && offset.Y == lastY {
-			return
-		}
 		lastX = offset.X
 		lastY = offset.Y
 		r.updateIndicator(false)
--- a/container/doctabs.go
+++ b/container/doctabs.go
@@ -67,10 +67,7 @@ func (t *DocTabs) CreateRenderer() fyne.WidgetRenderer {
 	r.action = r.buildAllTabsButton()
 	r.create = r.buildCreateTabsButton()
 	r.box = NewHBox(r.create, r.action)
-	var lastX, lastY float32
 	r.scroller.OnScrolled = func(offset fyne.Position) {
-		lastX = offset.X
-		lastY = offset.Y
 		r.updateIndicator(false)
 	}
 	r.updateAllTabs()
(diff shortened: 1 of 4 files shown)


andydotxyz:
I meant this to be a draft PR sorry. I cannot figure out yet why the visual tests fail just because we are not triggering on duplicate OnScroll calls

andydotxyz:
> Have you managed to find the source for the error?

I'm getting there. Lots of widgets just worked off this false assumption.
I have found and resolved 2 issues. At least one more remains.

andydotxyz:
Finally got to the bottom of it :)
It was 2 items in general

*) Calling `Scroller.Refresh()` no longer calls `OnScrolled`, so we may have to refresh the content directly if required
*) The setting of `Scroller.Offset` and then calling `Refresh()` does not call `OnScrolled` which makes sense. It previously did.

Jacalz (commented):
Hmm, I think there might be something wrong here. When adding a few more tabs to the DocTabs in fyne_demo, the scroller jumps back to the start after a while of dragging it to the right.

I would go about replicating it like this:
- Go to the DocTabs section in fyne_demo.
- Press the "+" button to add a few tabs and thus get a scroller.
- Grab hold of the scroller with the mouse and drag it to the right.
- About 2/3 on the way to the right, the scroller jumps back to the start again.

andydotxyz:
> Hmm, I think there might be something wrong here. When adding a few more tabs to the DocTabs in fyne_demo, the scroller jumps back to the start after a while of dragging it to the right.

This is a good catch thanks @Jacalz, however the issue is on develop too, so not part of the changes here.

Jacalz:
> > Hmm, I think there might be something wrong here. When adding a few more tabs to the DocTabs in fyne_demo, the scroller jumps back to the start after a while of dragging it to the right.
> 
> 
> 
> This is a good catch thanks @Jacalz, however the issue is on develop too, so not part of the changes here.

Are you sure? I tested on develop and couldn't quite replicate it. I'll have to try again, I guess.

andydotxyz:
> Are you sure? I tested on develop and couldn't quite replicate it. I'll have to try again, I guess.

Quite sure - it seems to glitch when it's about to show the selected tab when scrolling.

Jacalz (dismissed):
You are entirely correct. Sorry for the fuss.

andydotxyz:
Not a problem thanks @Jacalz.
I decided that this was a behaviour change so did not belong on a bugfix release (given the changes we had to make internally).
If you would be OK to re-approve with this change I would be appreciative :)

Jacalz (approved):
Looks good twice :)

PR #2661, opened 2021-11-23
data/binding: use atomics

### LINKED ISSUE #2359: Use atomics for internal data items

It could be a good idea to try and use atomic instructions for data bindings where we have internal control of the variables, i.e. not binding to external ones. This should allow us to get rid of a lot of slow mutexes and decrease lock contention; resulting in better performance.

It should theoretically be possible. The standard library is currently lacking atomic functions for a lot of types, so we might want to either use https://pkg.go.dev/go.uber.org/atomic or wait until more are implemented in the standard library. I think the latter is the best solution as there is talk about expanding the standard library, see https://github.com/golang/go/discussions/47141#discussioncomment-994417.

Replacing mutexes with atomics where it makes sense and can be done in a clean way.

changkun:
### Description:
Use `sync/atomic.*` and `sync.Map` to optimize the following two cases:

Existing benchmarks:

```
name             old time/op  new time/op  delta
BoolToString-8   1.12µs ± 2%  1.10µs ± 2%    ~     (p=0.095 n=5+5)
FloatToString-8  1.42µs ± 0%  1.37µs ± 0%  -3.08%  (p=0.016 n=4+5)
IntToString-8    1.13µs ± 0%  1.12µs ± 2%    ~     (p=0.159 n=4+5)
StringToBool-8   1.46µs ± 2%  1.39µs ± 0%  -5.04%  (p=0.008 n=5+5)
StringToFloat-8  1.60µs ± 1%  1.51µs ± 0%  -5.62%  (p=0.000 n=5+4)
StringToInt-8    1.47µs ± 2%  1.41µs ± 0%  -4.17%  (p=0.008 n=5+5)
```

The actual performance gain might be even higher, but concerning new benchmarks would be another story.

Fixes #2359

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/binding.go
+++ b/data/binding/binding.go
@@ -56,41 +56,27 @@ func (l *listener) DataChanged() {
 }
 
 type base struct {
-	listeners []DataListener
-	lock      sync.RWMutex
+	listeners sync.Map // map[DataListener]bool
+
+	lock sync.RWMutex
 }
 
 // AddListener allows a data listener to be informed of changes to this item.
 func (b *base) AddListener(l DataListener) {
-	b.lock.Lock()
-	defer b.lock.Unlock()
-
-	b.listeners = append(b.listeners, l)
+	b.listeners.Store(l, true)
 	queueItem(l.DataChanged)
 }
 
 // RemoveListener should be called if the listener is no longer interested in being informed of data change events.
 func (b *base) RemoveListener(l DataListener) {
-	b.lock.Lock()
-	defer b.lock.Unlock()
-
-	for i, listen := range b.listeners {
-		if listen != l {
-			continue
-		}
-
-		if i == len(b.listeners)-1 {
-			b.listeners = b.listeners[:len(b.listeners)-1]
-		} else {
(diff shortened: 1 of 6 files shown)


Bluebugs:
I think that writing meaningful benchmark for locking mechanism is quite hard. In this case, a RWLock would take more memory and might trigger in some case an overhead that the Atomic would not. It also feel safer in term of code as you have to use the Atomic API, you can't forget it when accessing the value. So even without the benchmark, I think this is a meaningful improvement.

Jacalz (approved):
Nice work. Not only is it faster and potentially safer because you have to use the locking, it also neatly cleans up the code. Thanks for this :)

changkun:
@andydotxyz Haven't heard anything from your side, what might be your insights and how could we better proceed?

andydotxyz:
Looks good @changkun, not sure what question I was to be answering sorry.

andydotxyz (approved):
Neat :)

PR #2662, opened 2021-11-24
Use new logo for app icons and demo image

andydotxyz:
Just asset updates (and we don't need light/dark image in demo as the icon works with both)


Jacalz (approved):
Nice 👍

PR #2667, opened 2021-11-29
internal: use atomic, sync.Map, and friends for internal structures

changkun:
### Description:

Optimize the following two cases:

Updates #2359

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (changes requested):
Thanks for this :)

Looks like this has an import problem causing the compiler to complain.

Jacalz (approved):
Looks good to me. Thanks a lot. All of this really is both easier to read and avoids future issues with forgetting to acquire locks when accessing fields. The performance improvement is certainly very good as well :)

andydotxyz (approved):

changkun:
Thanks a lot for the reviewing and I'll merge it now.

PR #2668, opened 2021-11-29
fyne: use atomic, sync.Map, and friends for internal structures

changkun:
### Description:

Optimize the following two cases:

Updates #2359

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app.go
+++ b/app.go
@@ -62,20 +62,20 @@ type App interface {
 	Lifecycle() Lifecycle
 }
 
-// app is a App typed variable. Note that we cannot use atomic.Value as
-// this app may be replaced by different types.
-var app atomic.Value // internalApp
+// app contains an App variable, but due to atomic.Value restrictions on
+// interfaces we need to use an indirect type, i.e. appContainer.
+var app atomic.Value // appContainer
 
-// internalApp is a dummy container that holds an App instance. This
+// appContainer is a dummy container that holds an App instance. This
 // struct exists to guarantee that atomic.Value can store objects with
 // same type.
-type internalApp struct {
+type appContainer struct {
 	current App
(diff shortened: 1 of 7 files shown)


Jacalz (dismissed):
Looks good to me. Nice work.

app.go
```diff
@@ -62,26 +62,30 @@ type App interface {
 	Lifecycle() Lifecycle
 }
 
-var app App
-var appLock sync.RWMutex
+// app is a App typed variable. Note that we cannot use atomic.Value as
```
andydotxyz, line 65: I think I understand this comment but it is a little confusing.
Maybe "app contains an App variable, but due to atomic.Value restrictions on interfaces we need to use an indirect type `internalApp`"

andydotxyz (commented):
Just a couple of thoughts

app.go
changkun: Should be resolved in the latest commit.

app.go
```diff
@@ -62,26 +62,30 @@ type App interface {
...
+// this app may be replaced by different types.
+var app atomic.Value // internalApp
+
+// internalApp is a dummy container that holds an App instance. This
+// struct exists to guarantee that atomic.Value can store objects with
+// same type.
+type internalApp struct {
```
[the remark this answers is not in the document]
changkun: Should be resolved in the latest commit.

Jacalz (approved):

Jacalz:
Sorry. I just saw that I somehow accidentally approved after @andydotxyz requested changes.

andydotxyz (approved):
Nice, thanks

PR #2673, opened 2021-11-30
Look up config dir using XDG_CONFIG_HOME

Jacalz:
### Description:

This switches the config dir lookup from hardcoded `$HOME + /.config` to reading the $XDG_CONFIG_HOME environment variable using [os.UserConfigDir()](https://pkg.go.dev/os#UserConfigDir). This has one very big improvement. It means that Fyne apps running in flatpak don't need to pass though access to the cache directory to use the preference and document storage API. For everyone else, it's basically no change unless a custom `$XDG_CONFIG_HOME` is set in some situation (our old way is still the fallback).

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -46,9 +46,7 @@ func (a *fyneApp) SendNotification(n *fyne.Notification) {
 }
 
 func rootConfigDir() string {
-	homeDir, _ := os.UserHomeDir()
-
-	desktopConfig := filepath.Join(homeDir, ".config")
+	desktopConfig, _ := os.UserConfigDir()
 	return filepath.Join(desktopConfig, "fyne")
 }
 


Jacalz:
I figure that this could be in the release notes as something like "Fixed Preferences and Documents Storage APIs not working out of the box in flatpak".

app/app_xdg.go
```diff
@@ -46,9 +46,7 @@ func (a *fyneApp) SendNotification(n *fyne.Notification) {
 }
 
 func rootConfigDir() string {
-	homeDir, _ := os.UserHomeDir()
-
-	desktopConfig := filepath.Join(homeDir, ".config")
+	desktopConfig, _ := os.UserConfigDir()
```
changkun, line 49: Could we safely ignore the error here? After reading the doc, https://pkg.go.dev/os#UserConfigDir, this API may return an error if $HOME is not set. To me, this seems to be likely?

Jacalz: The same error would have been send out before as well. It is not a change in behaviour with this PR.

changkun: Indeed that's why I was wondering why the error was never reported. If there is an error, then the config path would be `$(pwd)/.config` because `homeDir` would be `""`?

Jacalz: It should probably be checked but it might be outside of the scope of this PR. Not having $HOME set is also, to the best of my knowledge, very rare. The same error would have been sent out by os.UserHomeDir() as well.

andydotxyz (approved):

PR #2676, opened 2021-12-05
Update deps for initial NetBSD support

Jacalz:
### Description:

This enables us to successfully build and run on NetBSD.

Screenshot from 2021-12-10 21-02-36

With this added, we now support all three major BSD versions (FreeBSD, OpenBSD and NetBSD).
The only missing BSD project is DragonflyBSD (`dragonfly` is the build tag for it). It would be cool to add that for completeness as it and illumos/solaris support are everything that is missing from supporting pretty much all unix-like operating systems that Go supports.

Installing on NetBSD can be done using the following command (as root):
```
pkgin install go pkgconf
```

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz (commented):
I think this missed "go mod tidy" on the latest version?

Jacalz:
I don't think so? I know that it was the case once up on a time, but the latest commit should have fixed that.

Jacalz:
Running `go mod vendor` removed one item from the modules.txt file. Was that perhaps what you meant?

andydotxyz:
Apologies, it seems that my comment missed your commit "Tidy the modules" which did indeed resolve it already.

andydotxyz (approved):

Jacalz:
No worries. I find it weird that Go 1.16 doesn't always complain about thing not having been tidied.

PR #2677, opened 2021-12-06
Update and add missing package comments

Jacalz:
### Description:

This updates some of the package comments and adds some new ones. After this change, we have package comments for every public library package. 

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/test/testapp.go
+++ b/test/testapp.go
@@ -1,4 +1,4 @@
-// Package test provides utility drivers for running UI tests without and without rendering.
+// Package test provides utility drivers for running UI tests without rendering to a screen.
 package test // import "fyne.io/fyne/v2/test"
 
 import (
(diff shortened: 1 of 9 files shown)


Jacalz:
Feel free to come with suggestions if you have ideas for better package comments. Every improvement counts.

andydotxyz (approved):

test/testapp.go
```diff
@@ -1,4 +1,4 @@
-// Package test provides utility drivers for running UI tests without rendering
+// Package test provides utility drivers for running UI tests without and without rendering.
```
stuartmscott, line 1: > // Package test provides utility drivers for running UI tests without and without rendering.

Duplicate "without"

```suggestion
// Package test provides utility drivers for running UI tests without rendering to a screen.
```

stuartmscott (changes requested):

test/testapp.go
Jacalz: A bit late of a comment, but thanks. My intent was to say "with or without" rendering, but your suggestion was even better :)

stuartmscott: "A bit late of a comment"?

Jacalz: I meant that I committed your change early this morning but sent the comment many hours later.

stuartmscott (approved):

PR #2678, opened 2021-12-07
Fix custom font derivative lookups

### LINKED ISSUE #2512: The text on button can't be show correctly when use imported font
### Describe the bug:

The set on button can't be show correctly when use imported font
The text in label can be shown correctly but the text on button can't.
### To Reproduce:

### Screenshots:

image

### Example code:

func init() {
	fontPaths := findfont.List()
	for _, path := range fontPaths {
		fmt.Println(path)
		//楷体:simkai.ttf
		//黑体:simhei.ttf
		//PingFang-Regular.ttf
		if strings.Contains(path, "PingFang-Regular.ttf") {
			os.Setenv("FYNE_FONT", path)
			break
		}
	}
	fmt.Println("=============")
}

 - **OS:** Windows10
 - **Version:**  Fyne 2.1.0

andydotxyz:
If we load the first custom font use it as fallback instead
This means that other languages can provide just regular font

Fixes #2512

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/theme/theme.go
+++ b/theme/theme.go
@@ -516,9 +516,15 @@ func (t *builtinTheme) initFonts() {
 	font := os.Getenv("FYNE_FONT")
 	if font != "" {
 		t.regular = loadCustomFont(font, "Regular", regular)
-		t.bold = loadCustomFont(font, "Bold", bold)
-		t.italic = loadCustomFont(font, "Italic", italic)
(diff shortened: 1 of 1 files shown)


Jacalz (approved):
Seems sensible. Should we back-port to 2.1.x?

andydotxyz:
Yes, thanks - cherry-picked

PR #2679, opened 2021-12-07
cmd/fyne: more consistent errors, or how I stopped using pkg/errors

Jacalz:
### Description:

The "pkg/errors" package was only used in a few places and it was quite
inconsistent where it was and was not used. This switches to using the
standard library error wrapping instead.

Cheers to you if you cought the reference of the commit/PR naming :)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


changkun (approved):

andydotxyz (approved):
Cool, and nice title.

PR #2682, opened 2021-12-09
Add `Bytes` for a `[]byte` binding

andydotxyz:
Just adding `[]byte` to the types generated

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/comparator_helper.go
+++ b/data/binding/comparator_helper.go
@@ -1,8 +1,6 @@
 package binding
 
 import (
-	"bytes"
-
 	"fyne.io/fyne/v2"
 )
 
@@ -14,8 +12,4 @@ func compareURI(v1, v2 fyne.URI) bool {
 		return false
 	}
 	return v1.String() == v2.String()
-}
-
-func compareBytes(v1, v2 []byte) bool {
-	return bytes.Equal(v1, v2)
 }
--- a/data/binding/comparator_helper.go
+++ b/data/binding/comparator_helper.go
@@ -1,8 +1,6 @@
 package binding
 
-import (
-	"fyne.io/fyne/v2"
-)
+import "fyne.io/fyne/v2"
 
 func compareURI(v1, v2 fyne.URI) bool {
 	if v1 == nil && v1 == v2 {
--- a/data/binding/comparator_helper.go
+++ b/data/binding/comparator_helper.go
@@ -17,5 +17,5 @@ func compareURI(v1, v2 fyne.URI) bool {
 }
 
 func compareBytes(v1, v2 []byte) bool {
-	return bytes.Compare(v1, v2) == 0
+	return bytes.Equal(v1, v2)
 }
--- a/data/binding/comparator_helper.go
+++ b/data/binding/comparator_helper.go
@@ -1,6 +1,10 @@
 package binding
 
-import "fyne.io/fyne/v2"
+import (
+	"bytes"
+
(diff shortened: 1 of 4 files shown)


Jacalz (commented):
Should this use atomics like the other bindings, to be more consistent?

data/binding/comparator_helper.go
```diff
@@ -11,3 +15,7 @@ func compareURI(v1, v2 fyne.URI) bool {
 	}
 	return v1.String() == v2.String()
 }
+
+func compareBytes(v1, v2 []byte) bool {
+	return bytes.Equal(v1, v2)
+}
```
Jacalz, line 21: I saw why this was added and then fixed thanks to Staticcheck. However, do we really need this wrapper function after that?

Jacalz (changes requested):

data/binding/comparator_helper.go
andydotxyz: Good catch, thanks

andydotxyz:
> Should this use atomics like the other bindings, to be more consistent?

Not sure I follow here. This change just adds a new type to the list we generate. Where should I have added atomics?

stuartmscott (commented):
The original suggestion was `binding.Byte` for `byte` and `binding.ByteList` for `byte[]`, what is the advantage of treating `byte[]` as a single item?

andydotxyz:
> The original suggestion was `binding.Byte` for `byte` and `binding.ByteList` for `byte[]`, what is the advantage of treating `byte[]` as a single item?

Basically a slice of bytes is typically a single item. Like a string or an image.
The `List` bindings are more synonymous with list widgets I would think - displaying many items to a user.
I agree that semantically a byte slice is a list, but as we are focused on how things work for a GUI space it seemed reasonable.
I think it's similar to how we have `String` instead of `RuneList`.

Jacalz:
> > Should this use atomics like the other bindings, to be more consistent?
> 
> 
> 
> Not sure I follow here. This change just adds a new type to the list we generate. Where should I have added atomics?

I thought we switched over a lot of the other binding types to use atomics instead of mutexes?

andydotxyz:
> I thought we switched over a lot of the other binding types to use atomics instead of mutexes?

The types are all created from the same generator. Not sure what code I committed that you're saying are incorrect...

Jacalz:
> > I thought we switched over a lot of the other binding types to use atomics instead of mutexes?
> 
> The types are all created from the same generator. Not sure what code I committed that you're saying are incorrect...

Ah, sorry. I was wrong. I thought the atom is were added to more binding types but I see now that it wasn’t the case. Ignore my concern.

data/binding/comparator_helper.go
```diff
@@ -1,6 +1,8 @@
 package binding
 
-import "fyne.io/fyne/v2"
+import (
+	"fyne.io/fyne/v2"
+)
```
Jacalz, line 5: This seems like an unnecessary change. I think it is a leftover from the wrapper that was removed.

Jacalz (changes requested):
Just one minor nitpick. Seems sensible to me otherwise. Nice work. Seems very useful.

data/binding/comparator_helper.go
andydotxyz: Oh yes sorry about that.

Jacalz (approved):
Looks good. Thanks :)

PR #2696, opened 2021-12-16
Add (URI based) icon and theme reference to theme JSON

### LINKED ISSUE #2577: Add font and icon support to JSON theme
The main work is all done in #211, but we need to find a way to add resources for icon/font.
Supporting path lookup (URI probably works best, hooking in to repos so we get http etc).
Also need a way to reference in-memory resources I think, not sure how yet.

andydotxyz:
Possibility of res:// URI type in the future, but at this time developers can use the theme API directly to use embedded resources.

Fixes #2577

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


theme/json.go
```diff
@@ -122,10 +140,26 @@ func (t *jsonTheme) Color(name fyne.ThemeColorName, variant fyne.ThemeVariant) c
...
 }
 
 func (t *jsonTheme) Icon(name fyne.ThemeIconName) fyne.Resource {
+	if val, ok := t.data.Icons[string(name)]; ok {
+		r, err := val.resource()
+		if err != nil {
+			fyne.LogError("Failed to parse URI", err)
+		} else {
+			return r
+		}
```
toaster, line 161: This is exactly the same as in `Font(…)`. Maybe it should be moved into `stringUri.resource()` which then would simply return `nil` if the resource is not present for whatever reason?

toaster (changes requested):
For discussion:
I would prefer a solution where the parsing of the JSON returns any possible error.
Thus, an icon or font URL should trigger the resource loading.
This will increase initial loading times, of course.

Regarding the implementation: I’d left two code clean-up tasks.
- [ ] inline `uriString.uri()`
- [ ] think about consolidation for URI parse error handling

andydotxyz:
> I would prefer a solution where the parsing of the JSON returns any possible error.
> Thus, an icon or font URL should trigger the resource loading.

Hmm, this would be counter to how we handle images (which are not loaded until the rendering needs it).
It is also possible that these are dynamic resources, so the error state may not be consistent over time.

toaster (approved):

PR #2701, opened 2021-12-21
Feature/sprintfbinding

Bluebugs:
### Description:
This introduce a new data binding NewSprintf that as its name specify will pretty print a series of data bindings and maintain it up to date. It does also work with Set, within the limitation of fmt.Scanf limitation.

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

#### Where applicable:


andydotxyz (commented):
You'll save a lot of space removing the braces from case blocks :)

andydotxyz:
Can anyone else see a link to kick these actions? It's not appearing for me today...

Jacalz:
I wonder, would this be a better fit for Fyne-x?

data/binding/printf.go
```diff
@@ -0,0 +1,254 @@
...
+	source []interface{}
+	err    error
+}
+
+// NewSprintf returns a String binding that format its content using the
+// format string and the provide additional parameter that must be other
+// data bindings.
+//
+// Since: 2.2
+func NewSprintf(format string, b ...interface{}) (String, error) {
```
stuartmscott, line 20: If the additional parameter must be a data binding, why not use `DataItem` here?
```suggestion
func NewSprintf(format string, b ...DataItem) (String, error) {
```

data/binding/printf.go
```diff
@@ -0,0 +1,254 @@
...
+	ret := &sprintfString{String: NewString(), format: format, source: make([]interface{}, 0)}
```
stuartmscott, line 21: Slices don't need to be initialized before calling `append`
```suggestion
	ret := &sprintfString{String: NewString(), format: format}
```

Bluebugs:
> I wonder, would this be a better fit for Fyne-x? Seems to me like it is kind of very specific use case.

I think the opposite. Assembling string for display is a general purpose task. Printf is after all a very common function for a reason. It also doesn't bring any additional dependencies and it is fairly small and simple.

data/binding/printf.go
Bluebugs: I used interface{} to copy the prototype of sprintf, but I agree a DataItem should work. Will update.

data/binding/printf.go
```diff
@@ -0,0 +1,250 @@
...
+
+func (s *sprintfString) Get() (string, error) {
+	if s.err != nil {
+		return "", s.err
+	}
+	return s.String.Get()
+}
+
+func (s *sprintfString) Set(str string) error {
+	data := make([]interface{}, 0)
```
Jacalz, line 131: Any reason to not just do `data := []interface{}` (applies to more lines)?

data/binding/printf.go
```diff
@@ -0,0 +1,250 @@
...
+			{
+				b, err := x.Get()
+				if err != nil {
+					s.err = err
+					return
+				}
+
+				data = append(data, b)
+				break
+			}
```
Jacalz, line 48:
```suggestion
		b, err := x.Get()
		if err != nil {
			s.err = err
			return
		}

		data = append(data, b)
		break
```
I'm not against it per say, but is there any specific reason that you are launching a new scope here?

data/binding/printf.go
```diff
@@ -0,0 +1,250 @@
...
+		case Float:
+			{
+				f, err := x.Get()
+				if err != nil {
+					s.err = err
+					return
+				}
+
+				data = append(data, f)
+				break
```
Jacalz, line 69: Go adds break statements to switch cases by default (applies to more places).

data/binding/printf.go
Bluebugs: Not knowing about this syntax :-) Will correct.

Bluebugs: It actually doesn't compile if I use the syntax you are suggesting.

data/binding/printf.go
Bluebugs: It allow me to use variable that are only valid inside that block and I do not have to start inventing a lot of other variable name that make things, I find, more confusing.

data/binding/printf.go
Bluebugs: Thanks, I forgot!

data/binding/printf.go
Jacalz: That's strange. Might be something specific to empty interfaces then?
It certainly does work for other types https://www.geeksforgeeks.org/slices-in-golang/.

stuartmscott: Try `var data []interface{}`

Jacalz: You are right. It should work with `var data []interface{}` though.

data/binding/printf.go
andydotxyz: Case is a new scope without the braces. And you don’t need break either, that is the default in Go (use `fallthrough` to do the C default)

data/binding/printf.go
```diff
@@ -0,0 +1,214 @@
...
+		case Float:
+			data = append(data, new(float64))
+		case Int:
+			data = append(data, new(int))
+		case Rune:
+			data = append(data, new(rune))
+		case String:
+			data = append(data, new(string))
+		case URI:
+			return fmt.Errorf("impossible to convert '%s' to fyne.URI type", str)
```
andydotxyz, line 142: Maybe we could do `storage.ParseURI`?

andydotxyz: I think the limitation is I’m the current implementation not an actual limitation of URI string conversions. Just parse a string here and parse it to URI in the next code block?

data/binding/printf_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	waitForItems()
+
+	sGenerated, err := sp.Get()
+
+	assert.Nil(t, err)
+	assert.NotNil(t, sGenerated)
+	assert.Equal(t, expected, sGenerated)
+
+	err = sp.Set("Bool false , Float 7.000000 , Int 42 , Rune 67 , String nospacestring")
```
andydotxyz, line 88: I think we should check strings with spaces as well, user data will likely have some :)

andydotxyz (changes requested):
I think we can support more types with a small change to how the parse code works

data/binding/printf_test.go
Bluebugs: It is actually a limitation of scanf. I didn't feel like trying to give it a different behavior as that could be more confusing.

data/binding/printf.go
Bluebugs: My idea for URI is logically to use them as a string, but as scanf requirement apply, I wasn't certain that all 'space' characters are properly escaped when generated and I wanted to keep this simple for introduction of the feature.

data/binding/printf.go
Bluebugs: Oh, I didn't know. Thanks, corrected now.

data/binding/printf_test.go
andydotxyz: I get that Scanf has some limitations, but this is a binding for SPrintf in a 2 way mode - will people be expecting the Scanf limitations to apply?
If this is indeed to be expected I think we need to be clear that it will not work if strings have spaces - possibly capturing it as an error in our binding - I am assuming that otherwise it will crash or behave unexpectecly?

data/binding/printf.go
andydotxyz: Good point, but as an encoded URI it should not contain any spaces

data/binding/printf_test.go
Bluebugs: It won't crash, it just will parse things in unexpected way or refuse to parse string. Go scanf rely on spaces to split the string given and extract each component. This means that it is not just string containing spaces, but even the way the format is specified that require to take spaces into account.

We could change the name or/and improve documentation to make it clear that this two way binding will have scanf limitation when used with for setting data.

data/binding/printf.go
Bluebugs: Ok, if we are sure that there is no space in fyne URI, then I should be able to process them as string. Will look at it.

data/binding/printf.go
```diff
@@ -0,0 +1,186 @@
+package binding
```
stuartmscott, line 1: nit: maybe this file should be `sprintf.go`

Jacalz (approved):
Looks sensible to me.

data/binding/sprintf.go
```diff
@@ -178,7 +180,16 @@ func (s *sprintfString) Set(str string) error {
...
 			}
 		case URI:
-			return fmt.Errorf("impossible to convert '%s' to fyne.URI type", str)
+			v := data[i].(*string)
+
+			if v == nil {
+				return fmt.Errorf("URI can not be nil in '%s'", str)
+			}
+
+			err := x.Set(storage.NewURI(*v))
```
andydotxyz, line 190: isn't `NewURI` deprecated?

andydotxyz (approved):
Nice one thanks

PR #2702, opened 2021-12-21
Don't run change listeners if setting same value

Jacalz:
### Description:

This effectively makes sure that we don't write to the disk, or update/run any change listeners, when
trying to set the exact same value again. Does this make sense?
I noticed that I was writing just after reading the values in wormhole-gui, but does is happen in normal circumstances?

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
I was trying to find non-timer based way to wait for the change but couldn’t see one built in.
Probably good to go for now.

PR #2703, opened 2021-12-22
Update icns to v2. Finally lets us drop pkg/errors entirely.

Jacalz:
### Description:

This not only includes my PR to stop using the `pkg/errors` package, but it also includes many fixes and improvements.
Some of which were made by Andrew last year. So i short, it cleans up our vendor folder by ~500 lines and should bring many improvements when packaging for macOS (and iOS too, I think?).

I would appreciate if someone with a development environment on macOS could test this and verify that packaging generates the icons as expected. Probably both for iOS and macOS.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.


andydotxyz:
I should have access to my iOS dev box tonight

andydotxyz (approved):
Yeah looks good, thanks for this :)

andydotxyz:
I will merge as something has gone wrong with the actions triggers

PR #2707, opened 2021-12-26
Adding custom icon support

xines:
### Description:

Adds the functionality for developers to add svg icon files to a custom folder within theme icons folder that only require gen.go like orginal to be used within projects.

- Custom icons are automatically loaded with no need to write functions upon adding more to the collection, just run gen.go and it will do the magic.

- Custom icons could be seen as a non-fyne standard, but still enable developers to have a quick method of adding icons to their project needs.

- Current way of implementing fyne icons require a lot of code to be written in **icons.go**, while it potentially could be much less if using custom icons method.

### Usage:

Any svg file placed in **theme/icons/custom** folder will be able to be called like so:

```
theme.CustomIcon("icon_name") // replace "icon_name" with svg file name to use

widget.NewButtonWithIcon("", theme.CustomIcon("thumb_up"), func() {
	log.Println("tapped thumb up")
})
```

Tell me what you think about this approach and give it a go using some free to use material icons from:
- https://fonts.google.com/icons

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.


Jacalz (commented):
Sorry, but I don't understand what this is trying to achieve. That gen.go file is only for bundling the local files in Fyne and is not an external API. Also, loading icons at runtime goes against our idea that binaries to should be static and portable.

A developer using Fyne shouldn't have to edit Fyne, the dependency, to add custom icons. Further more, this does not even seem possible when using Go modules.

xines:
> Sorry, but I don't understand what this is trying to achieve. That gen.go file is only for bundling the local files in Fyne and is not an external API. Also, loading icons at runtime goes against our idea that binaries to should be static and portable.
> 
> A developer using Fyne shouldn't have to edit Fyne, the dependency, to add custom icons. Further more, this does not even seem possible when using Go modules.

Hello,
This does not run gen.go or bundle anything upon runtime/loading of any sort? What it reads upon init is the 'custom' svg names used to call a pre run gen.go containing a map function that holds all the static resources, meaning all bundles are static and portable.

Jacalz:
> This does not run gen.go or bundle anything upon runtime/loading of any sort? What it reads upon init is the 'custom' svg names used to call a pre run gen.go containing a map function that holds all the static resources, meaning all bundles are static and portable.

But what about the point of running that gen command? If I am developing my own application using Fyne, it doesn't make sense (at least not in my opinion) to edit the code for my Fyne dependency to add custom icons and then run a custom generator script.

From looking at the pushed code, the map containing the custom icons doesn't seem to ever be appended to. The branch seems to also incorrect be based on master and thus contains unrelated changes outside of this PR.

andydotxyz:
Custom icons for an application should not be added to fyne.
Apps can use "fyne bundle" to include items in their UI. See https://developer.fyne.io/extend/bundle for more info.

xines:
> But what about the point of running that gen command? If I am developing my own application using Fyne, it doesn't make sense (at least not in my opinion) to edit the code for my Fyne dependency to add custom icons and then run a custom generator script.

Sorry i don't see how this is any different from typing in the bundle command for every given icon u want to add...

> From looking at the pushed code, the map containing the custom icons doesn't seem to ever be appended to. The branch seems to also incorrect be based on master and thus contains unrelated changes outside of this PR.

Well it works on my OS: windows 10.
- Yes mistake my bad accidentally forked the master instead of develop branch, sorry.



> Custom icons for an application should not be added to fyne. Apps can use "fyne bundle" to include items in their UI. See https://developer.fyne.io/extend/bundle for more info.

I strongly agree to disagree, i think Jacalz got this bad influenced due to his lack of understanding at first, but i'll close this and only report bugs from now on. 👍

andydotxyz:
> I strongly agree to disagree, i think Jacalz got this bad influenced due to his lack of understanding at first, but i'll close this and only report bugs from now on.

We welcome pull requests - but adding app specific icons belongs in app code not in Fyne upstream. I don’t know if you are using vendoring or go modules, but you can easily add icons to your app without modifying the Fyne project, as is documented in the page I linked.

xines:
> We welcome pull requests - but adding app specific icons belongs in app code not in Fyne upstream. I don’t know if you are using vendoring or go modules, but you can easily add icons to your app without modifying the Fyne project, as is documented in the page I linked.

I see, I'm using go modules.

andydotxyz:
> Sorry i don't see how this is any different from typing in the bundle command for every given icon u want to add...

Maybe you missed that “fyne bundle” also accepts a directory as a parameter?
Or that “go generate” can be instructed to run “fyne bundle”?

Jacalz:
It is also worth noting that the bundle command creates a file in your local project, not in the upstream Fyne code.

PR #2713, opened 2021-12-30
[WiP] Add more documentation to ThemedResource

### LINKED ISSUE #2712: ThemedResource documentation answer the 101

I wanted to theme SVGs I am supplying to Fyne for rendering a scalable grid of numbers (hacking at sudokgo still). I Made a post in the slack, implemented something (buggy and broken 😂 ) and @andydotxyz came back to me suggesting `ThemedResource`.

It was a fantastic shout, as I'd just made a method called exactly that, which returned a `StaticResource` from a `StaticResource` using `bytes.ReplaceAll` (no XML parsing, just wrote value hacks).

Navigating to the docs; it was very API driven, without many hints. So I was forced to read the very well written source code; using a waypoint Andrew provided. 

> Note that the builtin ThemedResource is public and used to create all the theme matching icons in Fyne/FyneDesk and other apps

Yes; I'm just going to add to documentation for n00bs like myself.

(ThemedResource)[https://developer.fyne.io/api/v2.1/theme/themedresource.html) to mirror more fleshed out examples, like [Canvas](https://developer.fyne.io/api/v2.1/canvas/) or [App](https://developer.fyne.io/api/v2.1/app/)

Lewiscowles1986:
### Description:

Fixes #2712 

* Adjusts documentation based on author experience.
  * mentions icon dimensions designed for in multiple places.
  * attempts to make is clear what the intent of each New{T} and struct {T} is for.
  * clarifies interface required.

Because of code organisation, it looks like package level docs are not available (this is a sub-package).
Unsure how docs are generated to look like they do, but it looks like more edits to markdown will be required. (maintenance burden?)

### Checklist:
- [ ] Updated markdown docs

### DIFF
--- a/theme/icons.go
+++ b/theme/icons.go
@@ -569,13 +569,16 @@ func (t *builtinTheme) Icon(n fyne.ThemeIconName) fyne.Resource {
 	return icons[n]
 }
 
-// ThemedResource is a resource wrapper that will return a version of the resource with the main color changed
-// for the currently selected theme.
+// ThemedResource is a resource wrapper used to represent a version of a resource.
+// It changes colors based on the active theme.
+// This is currently achieved using XML attributes.
+// This is designed for use with Icons of 24px x 24px.
 type ThemedResource struct {
 	source fyne.Resource
 }
 
-// NewThemedResource creates a resource that adapts to the current theme setting.
+// NewThemedResource creates a resource that adapts to the current theme primary color.
+// Receives any struct implementing the Resource interface of Content() and Name()
 func NewThemedResource(src fyne.Resource) *ThemedResource {
 	return &ThemedResource{
 		source: src,
@@ -597,13 +600,18 @@ func (res *ThemedResource) Error() *ErrorThemedResource {
 	return NewErrorThemedResource(res)
 }
 
-// InvertedThemedResource is a resource wrapper that will return a version of the resource with the main color changed
-// for use over highlighted elements.
+// InvertedThemedResource is a resource wrapper used to represent a version of a resource.
+// It changes colors based on the active theme.
+// It is for use with active states.
+// This is currently achieved using XML attributes.
+// This is designed for use with Icons of 24px x 24px.
 type InvertedThemedResource struct {
 	source fyne.Resource
 }
 
-// NewInvertedThemedResource creates a resource that adapts to the current theme for use over highlighted elements.
+// NewInvertedThemedResource creates a resource that adapts to the current theme.
+// It adapts for use over highlighted elements.
+// Receives any struct implementing the Resource interface of Content() and Name()
 func NewInvertedThemedResource(orig fyne.Resource) *InvertedThemedResource {
 	res := &InvertedThemedResource{source: orig}
 	return res
@@ -625,13 +633,18 @@ func (res *InvertedThemedResource) Original() fyne.Resource {
 	return res.source
 }
 
-// ErrorThemedResource is a resource wrapper that will return a version of the resource with the main color changed
-// to indicate an error.
+// ErrorThemedResource is a resource wrapper used to represent a version of a resource.
+// It changes colors based on the active theme.
+// It is designed to communicate error states, and indicate error.
+// This is currently achieved using XML attributes.
+// This is designed for use with Icons of 24px x 24px.
 type ErrorThemedResource struct {
 	source fyne.Resource
 }
 
-// NewErrorThemedResource creates a resource that adapts to the error color for the current theme.
+// NewErrorThemedResource creates a resource that adapts to the current theme.
+// It adapts for use over error-state elements.
+// Receives any struct implementing the Resource interface of Content() and Name()
 func NewErrorThemedResource(orig fyne.Resource) *ErrorThemedResource {
 	res := &ErrorThemedResource{source: orig}
 	return res
@@ -652,13 +665,16 @@ func (res *ErrorThemedResource) Original() fyne.Resource {
 	return res.source
 }
 
-// PrimaryThemedResource is a resource wrapper that will return a version of the resource with the main color changed
-// to the theme primary color.
+// PrimaryThemedResource is a resource wrapper used to represent a version of a resource.
+// It changes colors based on the active theme.
+// This is currently achieved using XML attributes.
+// This is designed for use with Icons of 24px x 24px.
 type PrimaryThemedResource struct {
 	source fyne.Resource
 }
 
-// NewPrimaryThemedResource creates a resource that adapts to the primary color for the current theme.
+// NewThemedResource creates a resource that adapts to the current theme primary color.
+// Receives any struct implementing the Resource interface of Content() and Name()
 func NewPrimaryThemedResource(orig fyne.Resource) *PrimaryThemedResource {
 	res := &PrimaryThemedResource{source: orig}
 	return res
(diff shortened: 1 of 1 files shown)


andydotxyz:
The API generator is `_gen/genapi.sh` in the `developer.fyne.io` repository.
Pacakage level docs are included - there is a "package xxx" at the top of each package output into the website.

theme/icons.go
```diff
@@ -569,13 +569,16 @@ func (t *builtinTheme) Icon(n fyne.ThemeIconName) fyne.Resource {
 	return icons[n]
 }
 
-// ThemedResource is a resource wrapper that will return a version of the resource with the main color changed
-// for the currently selected theme.
+// ThemedResource is a resource wrapper used to represent a version of a resource.
```
andydotxyz, line 572: We should not remove the main information from the first sentence - this is all that will be shown in some IDEs

theme/icons.go
```diff
@@ -569,13 +569,16 @@ func (t *builtinTheme) Icon(n fyne.ThemeIconName) fyne.Resource {
...
+// It changes colors based on the active theme.
+// This is currently achieved using XML attributes.
+// This is designed for use with Icons of 24px x 24px.
```
andydotxyz, line 575: There is no limitation here - we use 24x24 SVG files, but that is not relevant here - also the "px" is misleading as the coordinate system is not pixels in SVG

theme/icons.go
```diff
@@ -569,13 +569,16 @@ func (t *builtinTheme) Icon(n fyne.ThemeIconName) fyne.Resource {
...
 type ThemedResource struct {
 	source fyne.Resource
 }
 
-// NewThemedResource creates a resource that adapts to the current theme setting.
+// NewThemedResource creates a resource that adapts to the current theme primary color.
```
andydotxyz, line 580: Is this true? I think that it adapts to the theme foreground colour...

theme/icons.go
```diff
@@ -569,13 +569,16 @@ func (t *builtinTheme) Icon(n fyne.ThemeIconName) fyne.Resource {
...
+// Receives any struct implementing the Resource interface of Content() and Name()
```
andydotxyz, line 581: Is this needed? the parameter is `Resource` so it is clear that the parameter must implement `Resource`.

andydotxyz (changes requested):
A couple of thoughts on the first few lines - I think it applies to most of the entries, but did not want to overwhelm :)

theme/icons.go
Lewiscowles1986: Are you sure? The following (two popular releases of the SVG spec) and a range of other sources are referring to pixels.

* https://dev.w3.org/SVG/profiles/1.1F2/publish/coords.html#InitialCoordinateSystem
* https://www.w3.org/TR/SVG/coords.html#InitialCoordinateSystem
* https://www.sarasoueidan.com/blog/svg-coordinate-systems/#:~:text=You%20specify%20the%20size%20of%20the%20viewport%20using%20the%20width%20and%20height%20attributes%20on%20the%20outermost%20%3Csvg%3E%20element.
* https://www.digitalocean.com/community/tutorials/svg-svg-viewbox

theme/icons.go
Lewiscowles1986: Hmm, the words `main color` was the part I found misleading; like `return a version of the resource`.

I Think I picked the wrong word represent, rather than `used to alter the presentation of a resource`. but we're line length limited here right?

theme/icons.go
Lewiscowles1986: I was trying for this (from GoLand) in VSCode. I clearly slipped and hit my head 😄 
image

theme/icons.go
Lewiscowles1986: Maybe it was perfect. IDK...

theme/icons.go
```diff
@@ -652,13 +665,16 @@ func (res *ErrorThemedResource) Original() fyne.Resource {
...
+// PrimaryThemedResource is a resource wrapper used to represent a version of a resource.
+// It changes colors based on the active theme.
+// This is currently achieved using XML attributes.
+// This is designed for use with Icons of 24px x 24px.
 type PrimaryThemedResource struct {
 	source fyne.Resource
 }
 
-// NewPrimaryThemedResource creates a resource that adapts to the primary color for the current theme.
+// NewThemedResource creates a resource that adapts to the current theme primary color.
```
Jacalz, line 676: This line is incorrect. It is missing `Primary`.
```suggestion
// NewPrimaryThemedResource creates a resource that adapts to the current theme primary color.
```

theme/icons.go
Jacalz: I agree with @andydotxyz. Seems a bit unnecessary.

Jacalz (changes requested):

theme/icons.go
andydotxyz: Pretty sure, though it is subtle. From the spec you linked it does seem like I am wrong on first read:
> and one unit in the initial coordinate system equals one "pixel" (i.e., a px unit as defined in CSS2 ([CSS2], section 4.3.2)

However when you reflect a little more you see that a CSS pixel is not necessarily the same as a screen output pixel (which is what we mean in the very few instances that we mention pixel in Fyne). I.e. a raster pixel will be 1:1 with hardware, but CSS pixel will not:

i.e.: 
> Specifically, text, SVG and much of CSS will "just work" because of the automatic pixel scaling feature of the web

https://www.html5rocks.com/en/mobile/high-dpi/

theme/icons.go
andydotxyz: Line length limited is not the same as sentence length limited.
You could put some words on line 2 then a "." part way along - IIRC.

theme/icons.go
Lewiscowles1986: Thanks @Jacalz I made this change locally yesterday, but the generated docs are not what I was looking for. I mentioned in the Slack channel; I may close this out, or look for an alternative approach.

theme/icons.go
Lewiscowles1986: Well it will scale so no vector should ever be 1:1 device to screen pixel. Happy enough to remove `px` if it's confusing. For now I'm not sure here is the correct place to document. RN I may sink this PR; but leaving it to sit for a bit; with all these changes made in a local.

andydotxyz: > Well it will scale so no vector should ever be 1:1 device to screen pixel. Happy enough to remove `px` if it's confusing.

As we are in a vector toolkit the "pixel" word has specific meaning - i.e. its things that don't scale. A pixel here means an output pixel - and so the definition is not compatible with the "user scalable pixel" of SVG.

> For now I'm not sure here is the correct place to document. RN I may sink this PR; but leaving it to sit for a bit; with all these changes made in a local.

Tough to say - added docs in all places is good - some people come through API docs, others through the tutorials site so either can be valuable.

Jacalz: > For now I'm not sure here is the correct place to document. RN I may sink this PR; but leaving it to sit for a bit; with all these changes made in a local.
> Tough to say - added docs in all places is good - some people come through API docs, others through the tutorials site so either can be valuable.

I would just like to add that API documentation is best added inside the repository. Some people, me included, prefer to use https://pkg.go.dev and if we generate other/extra API documentation somewhere else, it won't be shown there.

PR #2715, opened 2022-01-02
Feature: menu shortcuts

### LINKED ISSUE #682: Complete Native Menu Implementation

toaster:
### Description:
This PR introduces basic shortcut support for menus.
The shortcuts still have to be registered explicitly.
Also, the shortcut is _not_ deactivated if the menu item is deactivated.

Fixes #682

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
This looks amazing thank you :).
I was about to approve as the code looks clean and clear, however I just wondered:
- the menu shortcut text is as clear as the menu text (and disabled appropriately) but should it be more subtle? (I think I have seen some places do that).
Would having the shortcut hints be in the disabled colour too light? do we need a "caption" colour that would make sense or similar?

toaster:
>     * the menu shortcut text is as clear as the menu text (and disabled appropriately) but should it be more subtle? (I think I have seen some places do that).
>       Would having the shortcut hints be in the disabled colour too light? do we need a "caption" colour that would make sense or similar?

Hm, I think this is a matter of personal taste, isn’t it? I can’t remember to ever have seen such an approach but I’m not against it. It’s probably worth to think over and create some mock-ups at least. Then we could decide if we want to do it.
I’ll open an issue for it, so we don’t forget about it and have a place for the discussion and the mock-ups.

andydotxyz:
> Hm, I think this is a matter of personal taste, isn’t it? I can’t remember to ever have seen such an approach but I’m not against it. It’s probably worth to think over and create some mock-ups at least. Then we could decide if we want to do it.

Strange, it must just be the GTK theme I have loaded :) thanks for opening a new ticket to discuss

andydotxyz (approved):
Excellent work, thanks!

Issue #1646, opened 2020-12-11
Support to deliver over the web :)


andydotxyz:
WASM work etc.
Maybe 2.0?

AlbinoGeek:
Related: #1598

Bluebugs:
Steps:
* Do handle window resize in a fully asynchronous way
* Bring back experimental code that use the fake glfw binding (There could be improved way to call WebGL, but as a first step, it should be the easiest to get going)
* Tie things up ideally with cloud infrastructure to deploy the app, enjoy :-)

andydotxyz:
Step 1 is the main challenge (considering that @Bluebugs basically solved 2 already).
And I don't think it can be solved universally for all GLFW drivers in Fyne.
If we move desktop windows to fully async we see more tearing / jittering / blank spaces on resize.
Is it possible that a) we need sync resize but async response from the canvas? or b) differrent resize code for web vs desktop (as mobile is already a different code path)?

andydotxyz:
Updated with the remaining list of what could be the complete feature set, unless we miss anything

andydotxyz:
Moving to F for the final feature mop-up

andydotxyz:
Just file dialogs left to do, but that won't get in to F - let's try and high priority it for then.

PR #2724, opened 2022-01-05
Feature/wasm return

Bluebugs:
### Description:
This add a simple working WebGL backend. It does work on the devices I have tried it on so far, but it is expected that their is limitation at this point. I consider it useful enough to be shared and enable other to contribute to its improvement.

### Checklist:

- [ ] Tests included.
- [X] Tests all pass.

#### Where applicable:

- [X] Updated the vendor folder (using `go mod vendor`).


andydotxyz:
I think this needs to pull in latest develop to get the updated build matrix so that it triggers the correct rules.

andydotxyz:
To match the verbs API already in I would propose "fyne serve" instead of "fyne server" and the `Device.IsBrowser` instead of `Device.IsWeb` (this one is less clear).

cmd/fyne_demo/main.go
```diff
@@ -181,9 +188,15 @@ func makeNav(setTutorial func(tutorial tutorials.Tutorial), loadPrevious bool) f
 				return
 			}
 			obj.(*widget.Label).SetText(t.Title)
+			if !supportedTutorial(t) {
```
andydotxyz, line 191: These nodes are re-used on scroll, so the "supported" case must reset the italic back to regular

dialog/file_goxjs.go
```diff
@@ -0,0 +1,34 @@
...
+// +build wasm js
+
+package dialog
+
+import (
+	"fyne.io/fyne/v2"
+)
+
+func (f *fileDialog) loadPlaces() []fyne.CanvasObject {
+	return nil
```
andydotxyz, line 11: Every one of these functions is a dummy placeholder - I think we need TODO items and maybe even a linked issue if this PR is to land without file support in the browser.

driver/desktop/key.go
```diff
@@ -48,6 +48,18 @@ const (
 	SuperModifier = fyne.KeyModifierSuper
 )
 
+// Action represents the change of state of a key or mouse button event
+type Action int
+
+const (
+	// Release Keyboard button was released
+	Release Action = 0
```
andydotxyz, line 56: Why was it necessary to add new public APIs? (and why desktop specific?)

andydotxyz: It looks like these are only used in the internal desktop driver code. If required, that is where they should be...

internal/painter/gl/gl_common.go
```diff
@@ -15,6 +15,12 @@ import (
 	"fyne.io/fyne/v2/internal/painter"
 )
 
+// Texture represents an uploaded GL texture
+type Texture cache.TextureType
+
+// NoTexture is the zero value for a Texture
+var NoTexture = Texture(cache.NoTexture)
```
andydotxyz, line 22: Not sure why this needed to be aligned with the `cache.NoTexture` - it was just a zero value before

internal/cache/texture_desktop.go
```diff
@@ -1,12 +1,13 @@
...
+// +build !android,!ios,!mobile,!js,!wasm,!web
 
 package cache
 
 // TextureType represents an uploaded GL texture
 type TextureType = uint32
 
-var noTexture = TextureType(0)
+// NoTexture used when there is no valid texture
+var NoTexture = TextureType(0)
```
andydotxyz, line 10: Does this need exposed? Noted below that it does not seem necessary to equate the cache and gl ZeroTextures

internal/painter/gl/gl_common.go
```diff
@@ -27,16 +33,6 @@ func logGLError(err uint32) {
 	}
 }
 
-func (p *glPainter) getTexture(object fyne.CanvasObject, creator func(canvasObject fyne.CanvasObject) Texture) Texture {
```
andydotxyz, line 30: I'm not sure I follow why this was moved to platform specific code - it seems the same as before when there was just this 1 instance...

andydotxyz (changes requested):
I gave this a once over and have a few questions. There seem to be many new APIs exposed (some publicly) that don't seem to be necessary, commented on them inline.
Also there are `js`, `wasm` and `web` tags - do we need all 3? Sometimes only two are used, othertimes all 3 - but I can't follow what the difference is or when they might be used.

andydotxyz:
Very exciting though, this is one cool feature!

andydotxyz:
Just for posterity marking this as progressing #1646

Bluebugs:
> Also there are `js`, `wasm` and `web` tags - do we need all 3? Sometimes only two are used, othertimes all 3 - but I can't follow what the difference is or when they might be used.

Indeed, all this tags are confusing and not easy to figure out. js is the tag used when you compile with gopherjs. wasm when you target web assembly with go. Finally web is a tag that allow to compile most of the web backend, but using glfw desktop library to allow native testing of the backend with as much of the code in as possible.

internal/cache/texture_desktop.go
Bluebugs: From memory there are test that rely on NoTexture which make it more meaningful and easier to read.

internal/painter/gl/gl_common.go
Bluebugs: Each version depends on a different version of OpenGL to compile. I will look at it more and see if there could be more in common code.

internal/cache/texture_desktop.go
Bluebugs: I think NoTexture is best being exposed as it is not always 0 (case for wasm). This allow code to not have to care about the compiler target.

internal/painter/gl/gl_common.go
Bluebugs: I think I managed to get that sorted out somehow.

internal/painter/gl/gl_common.go
Bluebugs: With GL Wasm it is a JS value and can not be compared to zero easily anymore.

dialog/file_goxjs.go
Bluebugs: Will work on that this afternoon. Pushing an updated PR, but more work for this is needed.

internal/painter/gl/gl_common.go
andydotxyz: The addition of `Valid` seems to remove the need for this?

internal/cache/texture_desktop.go
andydotxyz: The new `Valid` exposed API seems to do the same thing? I don't think both have to be new public API do they?

internal/painter/gl/gl_common.go
Bluebugs: We still need the cache.NoTexture, I will check if NoTexture in the gl package need to be exposed or not, but it certainly need to exist as cache.NoTexture need to be used to initialize the default value properly as with wasm it get initialized to undefined or something like that otherwise.

internal/cache/texture_desktop.go
Bluebugs: cache.NoTexture need to be a public API as with wasm it is not initialized to zero by default, but to undefined.

internal/painter/gl/gl_common.go
Bluebugs: I checked I properly did switch the gl NoTexture to gl.noTexture and stopped exporting it. So I think we should be good with what is exported and what is not.

internal/cache/texture_desktop.go
andydotxyz: Ugh, good point!

andydotxyz:
I have found one issue that I think is not captured. When I resize the browser window (on a HiDPI screen) the content no longer fills the window...

calc-resize

andydotxyz (commented):
It looks like duplication in the go.sum - does it need a "go mod tidy" run?

andydotxyz (approved):
Epic :)

andydotxyz:
I am happy with this thanks. I think @Jacalz mentioned doing a review on this as well, given it's really large then probably worth waiting a little longer to see if more folk want to feed in.

internal/driver/glfw/key.go
```diff
@@ -0,0 +1,13 @@
+package glfw
+
+// Action represents the change of state of a key or mouse button event
+type Action int
+
+const (
+	// Release Keyboard button was released
+	Release Action = 0
```
andydotxyz, line 8: Note, I know this is internal now so it's not in the public API - but I think it does not need to be exported?

cmd/fyne/commands/command.go
```diff
@@ -49,3 +49,10 @@ func NewPackager() Command {
 func NewReleaser() Command {
 	return &commands.Releaser{}
 }
+
+// NewServer returns a command that can serve wasm application over http.
+//
+// Deprecated: A better version will be exposed in the future.
+func NewServer() Command {
+	return &commands.Server{}
+}
```
Jacalz, line 58: Why introduce a new API that is already deprecated? We created the internal commands package to avoid this. From what I can see, this is not used anywhere so it should be able to safely be removed.

cmd/fyne/internal/commands/package-wasm.go
```diff
@@ -0,0 +1,107 @@
...
+		"\t\t\t\tvar main = document.querySelector('#main');\n"+
+		"\t\t\t\tmain.innerHTML = \"WebAssembly and WebGL are not supported in your browser\";"+
+		"\t\t\t}\n"+
+		"\n"+
+		"\t\t</script>\n"+
+		"\t</head>\n"+
+		"\t<body>\n"+
+		"\t\t<main id=\"wasm\"></main>\n"+
+		"\t</body>\n"+
+		"</html>\n")
```
Jacalz, line 66: It might be worth bundling these. Maintaining long strings is going to be painfull in the future. Consider using `cmd/fyne/internal/templates/` instead.

cmd/fyne/internal/commands/serve.go
```diff
@@ -0,0 +1,129 @@
...
+	flag.StringVar(&s.icon, "icon", "Icon.png", "The name of the application icon file")
+}
+
+// PrintHelp prints the help for the install command.
+//
+// Deprecated: Access to the individual cli commands are being removed.
+func (s *Server) PrintHelp(indent string) {
+	fmt.Println(indent, "The serve command setup a local http server that is necessary to test WebAssembly package.")
+	fmt.Println(indent, "Command usage: fyne serve [parameters]")
+}
```
Jacalz, line 69: I don't think you really need to add these methods. These deprecation notices are for the older commands that were moved out to the internal commands package, hince this should be removed when the new API is removed from there.

cmd/fyne/internal/commands/serve.go
```diff
@@ -0,0 +1,129 @@
...
+// Run runs the install command.
+//
+// Deprecated: A better version will be exposed in the future.
+func (s *Server) Run(_ []string) {
+	err := s.serve()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err.Error())
+		os.Exit(1)
+	}
+}
```
Jacalz, line 110: I think the above command applies to this one as well.

cmd/fyne_demo/main.go
```diff
@@ -149,7 +150,7 @@ func makeMenu(a fyne.App, w fyne.Window) *fyne.MainMenu {
 
 	// a quit item will be appended to our first (File) menu
 	file := fyne.NewMenu("File", newItem, checkedItem, disabledItem)
-	if !fyne.CurrentDevice().IsMobile() {
+	if !fyne.CurrentDevice().IsMobile() && !fyne.CurrentDevice().IsBrowser() {
```
Jacalz, line 153:
```suggestion
	device := fyne.CurrentDevice()
	if !device.IsMobile() && !device.IsBrowser() {
```
Just a minor suggestion that makes it a bit easier to read and probably also avoids fetching the driver twice.

cmd/fyne_demo/main.go
```diff
@@ -159,6 +160,13 @@ func makeMenu(a fyne.App, w fyne.Window) *fyne.MainMenu {
 	)
 }
 
+func supportedTutorial(t tutorials.Tutorial) bool {
+	if !t.SupportWeb && fyne.CurrentDevice().IsBrowser() {
+		return false
+	}
+	return true
```
Jacalz, line 167: Hmm, it's weird that staticcheck didn't catch this. You should be able to remove the if-statement and just return the negated boolean value of the result you get from `!t.SupportWeb && fyne.CurrentDevice().IsBrowser()`.

It could also be better to call this `unsupportedTutorial` given that it is used with !supported everywhere.

device.go
```diff
@@ -28,6 +28,7 @@ func IsHorizontal(orient DeviceOrientation) bool {
 type Device interface {
 	Orientation() DeviceOrientation
 	IsMobile() bool
+	IsBrowser() bool
```
Jacalz, line 31: We should be careful with this in the future given that this technically is a breaking change, but we can probably ignore that in this case.

internal/driver/glfw/device.go
```diff
@@ -16,22 +16,10 @@ func (*glDevice) Orientation() fyne.DeviceOrientation {
...
-		return 1.0 // macOS scaling is done at the texture level
-	}
-	if runtime.GOOS == "windows" {
-		xScale, _ := w.(*window).viewport.GetContentScale()
-		return xScale
-	}
-
-	return scaleAuto
+func (*glDevice) IsBrowser() bool {
+	return (runtime.GOARCH == "js" || runtime.GOOS == "js")
```
Jacalz, line 24: Just a minor nitpick, shall we remove the parenthesis?

internal/driver/glfw/device_goxjs.go
```diff
@@ -0,0 +1,37 @@
...
+var OperaID = regexp.MustCompile(`Opera Mini`)
+var WindowsMobileID = regexp.MustCompile(`IEMobile`)
+
+var navigator = js.Global().Get("navigator")
+var userAgent = navigator.Get("userAgent").String()
+var mobileCheck = AndroidId.MatchString(userAgent) ||
+	BlackBerryID.MatchString(userAgent) ||
+	iOSID.MatchString(userAgent) ||
+	OperaID.MatchString(userAgent) ||
+	WindowsMobileID.MatchString(userAgent)
```
Jacalz, line 25: Would it not be better to use `string.Contains` instead?

internal/driver/glfw/device_goxjs.go
```diff
@@ -0,0 +1,37 @@
...
+
+func (*glDevice) IsMobile() bool {
+	if mobileCheck {
+		return true
+	}
+	return false
```
Jacalz, line 31:
```suggestion
	return mobileCheck
```

test/device.go
```diff
@@ -23,3 +27,7 @@ func (d *device) SystemScale() float32 {
 func (d *device) SystemScaleForWindow(fyne.Window) float32 {
 	return 1
 }
+
+func (*device) IsBrowser() bool {
+	return (runtime.GOARCH == "js" || runtime.GOOS == "js")
```
Jacalz, line 32:
```suggestion
	return runtime.GOARCH == "js" || runtime.GOOS == "js"
```
Similar nitpick as above. We should try to avoid unnecessary clutter where we can :)

Jacalz (changes requested):
I did not have time today to test out any applications but this is my batch of code review comments. Thanks fro this amazing work :)

cmd/fyne/commands/command.go
Bluebugs: I kind of wanted to keep things consistent with all the other command. If this is not necessary, I can remove all of that easily.

internal/driver/glfw/device_goxjs.go
Bluebugs: I prefer MatchString here mostly for consistency and being able to group things a bit logically (I know it could be one giant match or a lot of independent Contains, kind of went in between).

cmd/fyne/commands/command.go
andydotxyz: Yeah well caught @Jacalz - we will be removing these around 3.0 which may be after 2.2.
They are there because it's technically public API that apps could have utilised, but adding new ones in the same place should not be needed/

internal/driver/glfw/device_goxjs.go
Jacalz: I mean, it's not just about the MatchString. I'm kind of worried about the unnecessary overhead of compiling a lot of regular expressions unnecessarily. Is it perhaps not as bad as I think?

cmd/fyne/commands/command.go
Jacalz: Yeah. It's better to just remove this. As Andy said, it's only there with the other commands because it has to be like that.

internal/driver/glfw/device_goxjs.go
Bluebugs: I don't think it could be a concern in term of speed or memory. This is called once at startup and the "compilation" of a regex is just the generation of a matrix automate. Shouldn't be much of a concern.

Jacalz: Yeah that's true, but what point is there to use regex when you don't need to? Wouldn't it be just simpler and clearer to use the strings functions? It just seems to me like it is a lot of unnecessary code.

Any option? @andydotxyz

Jacalz: Another thought. How about having all the device support checks in one regexp? It avoids unnecessary code clutter and would probably also avoid looping though the same content many times.

internal/driver/glfw/device_goxjs.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+var AndroidId = regexp.MustCompile(`Android`)
+var BlackBerryID = regexp.MustCompile(`BlackBerry`)
+var iOSID = regexp.MustCompile(`iPhone|iPad|iPod`)
+var OperaID = regexp.MustCompile(`Opera Mini`)
+var WindowsMobileID = regexp.MustCompile(`IEMobile`)
+
+var navigator = js.Global().Get("navigator")
+var userAgent = navigator.Get("userAgent").String()
```
Jacalz, line 20: I suspect that the return type of this is a byte slice. It's probably better to not convert it to a string and just use the `.Match()` method instead.

Jacalz (changes requested):

Jacalz (changes requested):
It would be great if we could get some sort of support for fallback icons or being able to read the FyneApp.toml files for getting the icons. If I don't copy an icon to the current folder, I get this error when trying to serve fyne_demo:
```
Missing application icon at "/home/jacob/Downloads/fyne-feature-wasm-return/cmd/fyne_demo/Icon.png"
```

We need some better documentation for this I think. What needs to be installed, how to build and so on. This is the error that I get when trying to serve fyne_demo:
```
[jacob@workstation fyne_demo]$ ~/go/bin/fyne serve --sourceDir .
stat /usr/lib/golang/misc/wasm/wasm_exec.js: no such file or directory
```
Building the binary using the js GOOS and wasm GOARCH worked without issues though.

Jacalz:
I can confirm that the missing wasm file was due to using Go 1.16. I can't quite install Go 1.17 in any simple way here on my Fedora machine, but I managed to at least get the serving to work by manually changing the path of that file to point to a downloaded Go 1.17 copy. The icon showed correctly but I didn't get any visual output (I didn't quite expect to either for that matter).

andydotxyz:
> I get this error when trying to serve fyne_demo:
> 
> ```
> Missing application icon at "/home/jacob/Downloads/fyne-feature-wasm-return/cmd/fyne_demo/Icon.png"
> ```

It takes the same "-icon …” parameter as the other packaging formats.
It does look like FyneApp.toml support may have been missed, but we could add that in a follow up item if required

andydotxyz:
> This is the error that I get when trying to serve fyne_demo:
> 
> ```
> [jacob@workstation fyne_demo]$ ~/go/bin/fyne serve --sourceDir .
> stat /usr/lib/golang/misc/wasm/wasm_exec.js: no such file or directory
> ```

Does it work better if you omit the source dir parameter?

Jacalz:
> > This is the error that I get when trying to serve fyne_demo:
> > ```
> > [jacob@workstation fyne_demo]$ ~/go/bin/fyne serve --sourceDir .
> > stat /usr/lib/golang/misc/wasm/wasm_exec.js: no such file or directory
> > ```
> 
> Does it work better if you omit the source dir parameter?

Nope. I think the file simply is missing in Go 1.16.

internal/driver/glfw/device_goxjs.go
Bluebugs: I think it is actually a JS Value like all the API that cross to the JS world.

Bluebugs:
Due to the nature of wasm not being officially a stable target for the go compiler, it does change/break things between release enough to break things. I plan to only support the latest version and anything below will be served by gopherjs. I do plan to add the logic for this in the subsequent PR related to enable gopherjs.

Bluebugs:
> It does look like FyneApp.toml support may have been missed, but we could add that in a follow up item if required

I have created #2750 to track this feature request.

internal/driver/glfw/device_goxjs.go
Bluebugs: Technically it would indeed avoid looping on the user agent string 4 times, in the entire run of the application, but I ruled that out from a maintenance and clarity point of view as this would mean mixing all the OS together with no quick glance check if a family does correctly cover a specific agent.

I do believe this piece of code is not performance critical, and that the main point of it is to be maintainable. Please review it under this angle.

Bluebugs:
> > > This is the error that I get when trying to serve fyne_demo:
> > > ```
> > > [jacob@workstation fyne_demo]$ ~/go/bin/fyne serve --sourceDir .
> > > stat /usr/lib/golang/misc/wasm/wasm_exec.js: no such file or directory
> > > ```
> > 
> > 
> > Does it work better if you omit the source dir parameter?
> 
> Nope. I think the file simply is missing in Go 1.16.

I have created #2751 to cleanly report and disable web assembly output when the compiler is to old.

internal/driver/glfw/device_goxjs.go
Jacalz: Ah, yeah. You're probably right.

internal/driver/glfw/device_goxjs.go
Jacalz: Alright then.

Jacalz (approved):
As I can't currently get my hands in Go 1.17, I'll just say that the code looks good to me now. Absolutely amazing work Cedric! :)

PR #2727, opened 2022-01-06
Refactoring: composite font face no longer uses face specifc fallback glyphs

toaster:
### Description:

A couple of golang.org/x/image/font.Face methods state to return `!ok` if a face does not contain a glyph.
The actual implementation is not really intuitive because it returns a placeholder glyph for many faces.

I changed the behaviour to not return the placeholder glyph anymore.
~~Instead a follow-up PR will introduce the rendering of the Unicode replacement character (U+FFFD).~~

~~Note: until the follow-up PR gets merged, this changes the rendering of missing glyphs from the hollow rectangle to nothing.~~

**Update:** I added the commits of the follow-up PR which add the rendering of the Unicode replacement character.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
> Note: until the follow-up PR gets merged, this changes the rendering of missing glyphs from the hollow rectangle to nothing.

Curious about this line - is there a reason not to combine the refactor and the fix of this into 1 PR, avoiding the regression described?

toaster:
> Curious about this line - is there a reason not to combine the refactor and the fix of this into 1 PR, avoiding the regression described?

Hm … not really. I’ll add the commits.

andydotxyz (approved):
Yeah makes sense, thanks.

Issue #2698, opened 2021-12-19
Some incorrect cursor movement in a multiline entry with wrapping


nullst:
### Describe the bug:
This is kind of a follow-up to #1737 : more edge cases for cursor positioning for entries with wrapping enabled. 

Suppose you have a multiline entry with word wrapping enabled. Some user actions move the cursor into an unexpected position. I've observed two problems, probably with the same underlying issue, thus I'm reporting them jointly:

#### Erasing the first symbol
Put the cursor after the first symbol on some line in a multiline entry with wrapping. Any line except the very first one would work. Pressing backspace erases the symbol, but then moves the cursor to the start of the **next** line instead of staying in the same line.

The additional cursor movement happens due to the `rowColFromTextPos` function in `entry.go`, specifically in this part:
https://github.com/fyne-io/fyne/blob/6c35798dfe076cd244b66f3f44736e0d68f733c8/widget/entry.go#L910-L912

This part of logic was created to fix #1737, but I guess it is executed in some undesirable cases too.

#### Space key in a word-wrapped entry

In a word-wrapping entry, sometimes pressing the space character creates a new (visual) line of text. In some cases at this moment cursor also moves one more row down than expected: instead of being moved to the new visual line, it's being put one line after that. 

It's easiest to reproduce by putting the cursor at the beginning of the last word on some line, and then pressing space enough times to move that word to the next visual line. You would see that the cursor is no longer at the beginning of that word but rather in the next line after it. It's also relatively easy to reproduce by typing a space in the middle of a long word in a way that the second part of the word no longer fits visually after the space character. Then the cursor would also jump two lines down instead of only one.

### To Reproduce:
1. Compile the example code below and run it.
2. To see the issue with space character, just press space. The long word would split into two and the cursor would jump two lines down.
3. To see the issue with erasing one character, press backspace several time until you erase the first symbol. You will see the cursor move one line down. Alternatively, just put the cursor so that pressing backspace would erase the first symbol of some line for the same result.

### Example code:
```go
package main

import (
	"time"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	app := app.New()
	w := app.NewWindow("Multiline entry demo")

	entry := widget.NewMultiLineEntry()
	entry.Wrapping = fyne.TextWrapWord
	entry.SetText("abc\ns\ndefsfdsfsdffd\nf hh\n")

	w.SetContent(entry)
[...]

PR #2723, opened 2022-01-05
 [widget.NewMultiLineEntry() Bugs report]

jorzion:
Fix the problem that the mouse cannot be used to adjust the cursor position in NewMultiLineEntry.

 

### Description:

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [x] Lint and formatter run with no errors.
- [ ] Tests all pass.

#### Where applicable:

- [x] Public APIs match existing style.
- [ ] Any breaking changes have a deprecation path or have been discussed.
- [ ] Updated the vendor folder (using `go mod vendor`).
- [ ] Check for binary size increases when importing new modules.

PR #2726, opened 2022-01-06
Fix cursor issue in Multiline Entry

jorzion:
Fix the problem that the mouse cannot be used to adjust the cursor position in NewMultiLineEntry. 

 

### Description:

Fixes #?

### Checklist:

- [ ] Tests included.
- [x] Lint and formatter run with no errors.
- [x] Tests all pass.

#### Where applicable:

- [ ] Public APIs match existing style.
- [ ] Any breaking changes have a deprecation path or have been discussed.
- [ ] Updated the vendor folder (using `go mod vendor`).
- [ ] Check for binary size increases when importing new modules.

PR #2729, opened 2022-01-07
Fix the logic error of rowColFromTextPos in Backspace

jorzion:
### Description:

https://github.com/fyne-io/fyne/pull/2723#issuecomment-1006180620

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -402,11 +402,11 @@ func (e *Entry) MouseDown(m *desktop.MouseEvent) {
 	}
 	e.propertyLock.Unlock()
 
+	e.updateMousePointer(m.Position, m.Button == desktop.MouseButtonSecondary)
+
 	if !e.Disabled() {
 		e.requestFocus()
 	}
-
-	e.updateMousePointer(m.Position, m.Button == desktop.MouseButtonSecondary)
 }
 
 // MouseUp called on mouse release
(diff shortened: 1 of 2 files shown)


jorzion:
If you have a better solution, please close this merge request.

Jacalz:
Does this in some way replace #2726?

jorzion:
The PR should have been submitted twice, but the unexpected result became the result of submitting two PRs in one submission action.

> Does this in some way replace #2726?

jorzion:
As long as the problem can be solved, the others are not important. :)

nullst:
Does this fix #2698 ? Seems to be related. Sorry, can't check it myself right now.

jorzion:
> Does this fix #2698 ? Seems to be related. Sorry, can't check it myself right now.

Yes, It seems to be the same problem.

andydotxyz:
> The PR should have been submitted twice, but the unexpected result became the result of submitting two PRs in one submission action.

I'm not sure what that means - should the older PR be replaced?
This includes commits from the older PR but also additional ones, I think we are not sure which should be reviewed for which fix.

Jacalz:
Any update on the above comment @jorzion? I'm sorry to have to bump this but I think we should try to squeeze this in for the next bug fix release.

andydotxyz:
It looks like between #2726 and the newer #2698 this PR is redundant. Sorry for picking up the latter, this PR seemed to have stagnated.

PR #2731, opened 2022-01-11
cmd/fyne: update error to report stderr content

lucor:
### Description:
This PR updates the packageUNIX() method to return the stdErr content for the tar command in case of error.
The additional info provided could be helpful to debug the error.

Fixes fyne-io/fyne-cross#85

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/package-unix.go
+++ b/cmd/fyne/internal/commands/package-unix.go
@@ -68,7 +68,7 @@ func (p *Packager) packageUNIX() error {
 			return fmt.Errorf("failed to write Makefile string: %w", err)
 		}
 
-		buf := bytes.Buffer{}
+		var buf bytes.Buffer
 		tarCmd := execabs.Command("tar", "-Jcf", p.name+".tar.xz", "-C", tempDir, "usr", "Makefile")
 		tarCmd.Stderr = &buf
 		if err = tarCmd.Run(); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/fyne/internal/commands/package-unix.go
```diff
@@ -67,9 +68,11 @@ func (p *Packager) packageUNIX() error {
 			return fmt.Errorf("failed to write Makefile string: %w", err)
 		}
 
+		buf := bytes.Buffer{}
```
stuartmscott, line 71: nit: I don't think it makes a difference, but I think looks neater
```suggestion
		var buf bytes.Buffer
```

cmd/fyne/internal/commands/package-unix.go
```diff
@@ -67,9 +68,11 @@ func (p *Packager) packageUNIX() error {
...
 		tarCmd := execabs.Command("tar", "-Jcf", p.name+".tar.xz", "-C", tempDir, "usr", "Makefile")
+		tarCmd.Stderr = &buf
 		if err = tarCmd.Run(); err != nil {
-			return fmt.Errorf("failed to create archive with tar: %w", err)
+			return fmt.Errorf("failed to create archive with tar: %s - %w", buf.Bytes(), err)
```
stuartmscott, line 75: nit: Again, might not make a difference
```suggestion
			return fmt.Errorf("failed to create archive with tar: %s - %w", buf.String(), err)
```

stuartmscott (approved):

PR #2733, opened 2022-01-13
Ignore Carriage Return when Measuring or Drawing Text

stuartmscott:
Carriage Return character is ignored in both the measure and draw loops.

Fixes: #2456

### Checklist:

- [x] Tests included.
- [x] Tests all pass. (except unrelated test failures: https://gophers.slack.com/archives/CK5U4BU87/p1642091780028600)


andydotxyz (approved):

andydotxyz (commented):
Apologies, I approved this as a good approach, but then I thought maybe there is need for an Entry test as well - when \r\n is the newline combination does it behave correctly with these elements simply not displayed?

andydotxyz:
Thanks @stuartmscott , I see multiline Entry does not limit to 1 line.
I guess 2 options here:
* Remove the multi-check test that shows it is wrong and add that to a new issue.
* Fix the issue on this ticket too.

I think the first is perfectly OK - just avoiding committing a test with wrong output as golden image.

andydotxyz (approved):
Thanks @stuartmscott - before closing can you link to a bug for the single line supporting multiple lines so we don't forget?

stuartmscott:
Moved to https://github.com/fyne-io/fyne/pull/2758

PR #2746, opened 2022-01-21
Unsafe pointer is a no go with gopherjs.

Bluebugs:
Thanksfully, there is only one system thread when running with gopherjs.
This means that all atomic call basically become direct access when
converted. This change does that for unsafe pointer as gopherjs doesn't
provide that atomic.

### Description:
gopherjs does not support unsafe pointer. This PR is just about removing
the only place where it shows up in Fyne tree.

### Checklist:

- [ ] Tests included.
- [X] Tests all pass.


changkun (approved):
Changes LGTM. Not entirely sure about gopherjs and wasm, but it seems that wasm has thread support and may be introduced in go. Is it wise to leave a TODO in the js version to remind the future? https://github.com/golang/go/issues/28631

Bluebugs:
Indeed, that why this is just targeting gopherjs with the build tag "js" (the build tag for web assembly is currently "wasm").

Jacalz (approved):
Seems sensible. Will go ahead and merge this time around.

PR #2748, opened 2022-01-23
Feature/non mac super symbol

toaster:
### Description:

This PR introduces a different rune representing the Super key on non-Apple systems, the “BLACK DIAMOND MINUS WHITE X” (`0x2756`, ❖) as fake Windows logo.

It also includes some missing documentation comments and replacement of deprecated constants.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


key.go
```diff
@@ -176,15 +176,40 @@ const (
...
+	//
+	// Since: 2.2
 	KeyModifierSuper
 )
+
+const (
+	// RuneModifierAlt is the rune that represents the Alt key.
+	//
+	// Since: 2.2
+	RuneModifierAlt = '⌥'
```
andydotxyz, line 206: Do these need to be public? Isn’t it a rendering detail rather than part of the user API?

andydotxyz (commented):
Looks great, but not sure I understand adding APIs for the runes we use in the menu…

key.go
toaster: Yes, you’re right.

andydotxyz (approved):
Nice thanks

PR #2752, opened 2022-01-25
Disable check for pinned thread with gopherjs.

Bluebugs:
### Description:
gopherjs only does have one system thread, so checking for thread and
pinning to a system thread doesn't work.

### Checklist:
- [X] Tests all pass.

### DIFF
--- a/internal/driver/glfw/driver.go
+++ b/internal/driver/glfw/driver.go
@@ -78,13 +78,6 @@ func (d *gLDriver) Quit() {
 	close(d.done)
 }
 
-func (d *gLDriver) Run() {
-	if goroutineID() != mainGoroutineID {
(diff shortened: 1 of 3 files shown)


Jacalz:
I think we did it this way to be consistent. If someone can run it from a goroutine in gopherjs, they will think that they can do so for the others as well.

Bluebugs:
The problem is that you don't have the notion of any thread at all in gopherjs, so the test never match and always panics. gopherjs is the only target I know of that do not have support for system thread.

andydotxyz:
I think that @Jacalz and I don't understand how gopherjs works in this case.
Will code like `go func() { a.Run() }()` compile and run or not?
Initially it looked like we should fix `goroutineID` instead, but my understanding is limited so more info could help :).

PR #2753, opened 2022-01-26
Feature/enforce go version

### LINKED ISSUE #2751: Require latest go version for compiling web assembly
The goal would be to disable with a clear error message web assembly as a target when the go compiler is to old.

No new API should be needed.

Checking the go version against a hard coded one and exiting fyne command if that version is to old. Currently only 1.17 is to be supported.

Bluebugs:
### Description:
We will only support specific go version for wasm. At the moment 1.17 is the minimal requirement.

Fixes #2751

### Checklist:

- [X] Tests all pass.

#### Where applicable:

- [X] Updated the vendor folder (using `go mod vendor`).

### DIFF
--- a/cmd/fyne/internal/commands/build.go
+++ b/cmd/fyne/internal/commands/build.go
@@ -6,6 +6,7 @@ import (
 	"runtime"
 	"strings"
 
+	version "github.com/mcuadros/go-version"
 	"golang.org/x/sys/execabs"
 )
 
@@ -16,6 +17,8 @@ type builder struct {
 }
 
 func (b *builder) build() error {
+	var versionConstraint *version.ConstraintGroup
+
 	goos := b.os
 	if goos == "" {
 		goos = targetOS()
@@ -55,11 +58,31 @@ func (b *builder) build() error {
 	if goos != "ios" && goos != "android" && goos != "wasm" {
 		env = append(env, "GOOS="+goos)
 	} else if goos == "wasm" {
+		versionConstraint = version.NewConstrainGroupFromString(">=1.17")
 		env = append(env, "GOARCH=wasm")
 		env = append(env, "GOOS=js")
 	}
 
 	cmd.Env = env
+
+	if versionConstraint != nil {
+		goVersion, err := execabs.Command("go", "version").Output()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "%s\n", string(goVersion))
+			return err
+		}
+
+		split := strings.Split(string(goVersion), " ")
+		// We are expecting something like: `go version goX.Y OS`
+		if len(split) != 4 || split[0] != "go" || split[1] != "version" || len(split[2]) < 5 {
+			return fmt.Errorf("invalid output for `go version`: `%s`", string(goVersion))
+		}
+
+		normalized := version.Normalize(split[2][2 : len(split[2])-2])
+		if !versionConstraint.Match(normalized) {
+			return fmt.Errorf("expected go version %v got `%v`", versionConstraint.GetConstraints(), normalized)
+		}
+	}
 
 	out, err := cmd.CombinedOutput()
 	if err != nil {
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 	github.com/jackmordaunt/icns/v2 v2.1.3
 	github.com/josephspurrier/goversioninfo v0.0.0-20200309025242-14b0ab84c6ca
 	github.com/lucor/goinfo v0.0.0-20210802170112-c078a2b0f08b
+	github.com/mcuadros/go-version v0.0.0-20190830083331-035f6764e8d2
 	github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e // indirect
 	github.com/srwiley/oksvg v0.0.0-20200311192757-870daf9aa564
 	github.com/srwiley/rasterx v0.0.0-20200120212402-85cb7272f5e9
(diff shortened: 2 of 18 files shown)


cmd/fyne/internal/commands/build.go
```diff
@@ -55,12 +58,32 @@ func (b *builder) build() error {
...
 		env = append(env, "GOOS="+goos)
 	} else if goos == "wasm" {
+		versionConstraint = version.NewConstrainGroupFromString(">=1.17")
 		env = append(env, "GOARCH=wasm")
 		env = append(env, "GOOS=js")
 	}
 
 	cmd.Env = env
 
+	if versionConstraint != nil {
```
andydotxyz, line 68: Could this be factored out into a method, and a couple of validation tests added so we can check the parse/version comparison?

cmd/fyne/internal/commands/build.go
```diff
@@ -55,12 +58,32 @@ func (b *builder) build() error {
 	if goos != "ios" && goos != "android" && goos != "wasm" {
 		env = append(env, "GOOS="+goos)
 	} else if goos == "wasm" {
+		versionConstraint = version.NewConstrainGroupFromString(">=1.17")
```
andydotxyz, line 61: As it's only used in this instance would it not be easier to just run the check here?

andydotxyz (changes requested):
I think you removed the mandatory "includes tests" checkbox item which is not good, especially as it should have tests...

cmd/fyne/internal/commands/build.go
Bluebugs: I am expecting to set the test to <=1.17 for gopherjs when I add it here. That's why I kept it as a general feature.

cmd/fyne/internal/commands/build.go
Bluebugs: I am trying to figure out how to write a test for this, but as it depends first on the output of `go version`, I am not sure how to proceed. Any suggestion?

andydotxyz: I would guess capturing certain known (or erroneous) outputs of `go version` and wire them into the test.
So the function would verify based on string not on a call out to a command.

cmd/fyne/internal/commands/build.go
andydotxyz: Fair enough

andydotxyz (changes requested):
I don't think this is testing the new addition of "we should error if built in js mode with Go < 1.17" it is testing how we run external tools to get the version info, that seems a different area

cmd/fyne/internal/commands/build.go
andydotxyz: Hmm, what I meant is that we can feed in captured versions of the "go version" tool output and test using that. It seems reasonable to assume that to go tool will work, if present.

cmd/fyne/internal/commands/build.go
```diff
@@ -6,16 +6,58 @@ import (
...
+	return runner.RunOutput(args...)
+}
+
 type builder struct {
 	os, srcdir, target string
 	release            bool
 	tags               []string
 }
 
+func checkVersion(runner runner, versionConstraint *version.ConstraintGroup) error {
```
lucor, line 33: The version could be also obtained using `runtime.Version()`. It could be good enough to check only against the minor version for now and avoid the introduction of a new dependency.
```suggestion
func checkVersion(minorConstraint int) (bool, error) {
	version := runtime.Version()
	minor, err := strconv.Atoi(version[4:6])
	if err != nil {
		return false, fmt.Errorf("invalid version: %s", version)
	}
	if minor < minorConstraint {
		return false, fmt.Errorf("expected Go minor version %d. Got %d from version %s", minor, minorConstraint, version)
	}
	return true, nil
}
```

Bluebugs: No, this won't work as there is no guaranty that the go installed is the same as the one used to compile the fyne command and so there is no link between runtime and compile time.

lucor: Oh, right we are going to invoke the go binary directly... I've missed that sorry.
Btw, if there are no downside the check about the minor version to avoid the addition of the dependency could be still valid.

andydotxyz: This is a good point. Realistically all we are doing is checking for "17" at the beginning of the version string...

Bluebugs: The version parsing is sadly not so easy as people could be using a git release of go for example. I do not see any drawback into relying on someone else work here instead of reinventing the wheel. The dependency only impact fyne command. It is MIT, small, well tested, used and more future proof that what we could come up with.

andydotxyz (changes requested):
A few thoughts inline.
I am a little concerned that there is about 300 lines of test code but 20 lines of new code...
Is this multiple new test areas in a single PR?

cmd/fyne/internal/commands/build.go
lucor: I see, but if I understood correctly, we need to check only if the minor version of the go binary used to compile in wasm is greater than `17` (this should be true until Go 2.0 is released). Looking at the code we are already extracting the version from the `go version` command output in a custom way so comparing directly the minor version should be good enough.

Anyway if we want to use the module, probably we would consider some alternatives like https://github.com/hashicorp/go-version since the one used seems archived.

cmd/fyne/internal/commands/runner.go
```diff
@@ -0,0 +1,46 @@
...
+	EnvSet(env []string)
+}
+
+type command struct {
+	exe string
+	dir *string
+	env []string
+}
+
+func (c *command) DirSet(dir string) {
```
andydotxyz, line 19: It should be Verb first (SetXxx)

cmd/fyne/internal/commands/build.go
Bluebugs: I did go with hashicorp go-version as it was MPL and not sure that would be ok for the project. If it is, I have no problem changing to it as they are pretty much equivalent.

andydotxyz (approved):

cmd/fyne/internal/commands/build.go
andydotxyz: I don't think MPL is OK for us, copying my Slack message here:

Following discussion on tickets I reviewed MPL v2 and I don't think we can use it... The main problematic clause is 3.2.1: "If You distribute Covered Software in Executable Form then: [1] such Covered Software must also be made available in Source Code Form..." - we cannot put that requirement on our users.

PR #2757, opened 2022-01-28
Feature/detect failure case

### LINKED ISSUE #2756: Improve initial web page for the web driver
Currently their is no user feedback during the initialization of the web driver. It would be good to provide user feedback during the progress of the initialization.

There should be no need for any new API.

The index.html provided by fyne could:
- detect that the browser does not have javascript support, display warning if not
- detect that the browser support WebGL, display warning if not
- detect that the browser support webassembly, fallback to gopherjs backend
- let the user know that the webassembly is being downloaded

Bluebugs:
### Description:
This is an initial patch that improve detection of failure case when starting a web application. There should be discussion and follow up PR to actually improve the visual of this.

Fixes #2756 

### Checklist:

- [X] Tests all pass.


cmd/fyne_demo/FyneApp.toml
```diff
@@ -2,4 +2,4 @@
   Icon = "../../theme/icons/fyne.png"
   Name = "Fyne Demo"
   ID = "io.fyne.demo"
-  Build = 6
```
andydotxyz, line 5: As this is not releasing a new version of these apps I don't think the build should be bumped...

andydotxyz (commented):
Seems like a good change - but I don't think the metadata update belongs

cmd/fyne_demo/FyneApp.toml
Bluebugs: Oh indeed.

andydotxyz (approved):
Should result in better debug info!

PR #2760, opened 2022-02-02
Generate an artifact with coverage report in each PR.

Bluebugs:
### Description:
This is just an experiment with github action to see how useful a go coverage report attached to each build is.

### DIFF
--- a/.github/workflows/platform_tests.yml
+++ b/.github/workflows/platform_tests.yml
@@ -38,12 +38,20 @@ jobs:
         go test -tags ci -covermode=atomic -coverprofile=coverage.out ./...
         coverage=`go tool cover -func coverage.out | grep total | tr -s '\t' | cut -f 3 | grep -o '[^%]*'`
         if (( $(echo "$coverage < 63" | bc) )); then echo "Test coverage lowered"; exit 1; fi
+        go tool cover -html=coverage.out -o coverage.html
       if: ${{ runner.os == 'Linux' }}
 
     - name: Update PR Coverage
       uses: shogo82148/actions-goveralls@v1
       with:
         path-to-profile: coverage.out
+      if: ${{ runner.os == 'Linux' && matrix.go-version == '1.17' && github.event_name == 'push' }}
+    
+    - name: Upload coverage report
+      uses: actions/upload-artifact@v2
+      with:
+          name: html-report
+          path: coverage.html
       if: ${{ runner.os == 'Linux' && matrix.go-version == '1.17' && github.event_name == 'push' }}
 
     - name: Build WebAssembly binary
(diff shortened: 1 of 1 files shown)


.github/workflows/platform_tests.yml
```diff
@@ -38,13 +38,21 @@ jobs:
...
     - name: Update PR Coverage
       uses: shogo82148/actions-goveralls@v1
       with:
         path-to-profile: coverage.out
       if: ${{ runner.os == 'Linux' && matrix.go-version == '1.17' && github.event_name == 'push' }}
+    
+    - name: Upload coverage report
+      uses: actions/upload-artifact@v2
+      with:
+          name: html-report
```
andydotxyz, line 53: Great idea, but this generic name probably should be improved so the artefact has clearer purpose.

andydotxyz (changes requested):

andydotxyz (approved):
:) thanks

Bluebugs:
I dropped the idea of posting a comment with the coverage report as github action do not allow PR that come from a forked repository to actually do get any write access on anything. A possible workaround would be to set a cron type of github action along what is shown here: https://github.com/nyurik/auto_pr_comments_from_forks , but this is way bigger than I really wanted to get done here.

Jacalz:
I can't seem to figure out where this coverage report is supposed to be found. Where should I look?

Jacalz:
Another thought. Would it be better to just run coverall on PRs instead of generating html files?

Bluebugs:
> I can't seem to figure out where this coverage report is supposed to be found. Where should I look?

I know took me time to figure out. You need to go in the platform_test summary section and then scroll all the way down. You will see an artifact section that now contain the result of this build. It will be available for 90 days.

Bluebugs:
> Another thought. Would it be better to just run coverall on PRs instead of generating html files?

I don't think it has the same purpose and the report is only related to the exact build that passed with the exact set of commit that match it find it easier to check this way (arguably would be nice if it wasn't in a zip file, but github wouldn't allow it).

On another though, it might be good to have coverall for the develop branch independently of master. What do you think?

andydotxyz:
> Another thought. Would it be better to just run coverall on PRs instead of generating html files?

I don't think so - coveralls is our public record of our code coverage. Having that change based on incoming PRs seems undesirable.

stuartmscott:
My PR in another project was initially blocked by coveralls because the coverage went down, and would update the comment after each new commit with the updated percentage. Seems to work quick well - https://github.com/bmaupin/go-epub/pull/56#issuecomment-998885366

Bluebugs:
@stuartmscott Thanks for pointing it out. It seems possible to setup coverall to callback once you push the coverage there. Might look into it next week, but this would work around the access right problem as we just need to give access to the bot.

Bluebugs:
https://github.com/fyne-io/fyne/pull/2785 is looking like a better solution.

PR #2762, opened 2022-02-03
Feature/coverage code terminal function

Bluebugs:
### Description:
This PR introduce proper test coverage of fyne build command by enabling mocking of the Command() call and checking that the right call are made along with the right handling of errors.

I am currently not targeting develop, but will as soon as the go version check is landed.

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne/internal/commands/mock_runner_test.go
+++ b/cmd/fyne/internal/commands/mock_runner_test.go
@@ -1,0 +1,96 @@
+package commands
+
+import (
+	"fmt"
+	"os"
+	"sort"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type mockRunOutputValue struct {
+	// Expected value
+	dir   interface{}
+	env   []string
+	osEnv bool
+	args  []string
+	// Returned value
+	ret []byte
+	err error
+
+	// To check if all expected call did happen
+	dirSet bool
+	envSet bool
+}
+
+type testCommandCall struct {
+	calls []mockRunOutputValue
+	index int
+	t     *testing.T
+}
+
+func (t *testCommandCall) RunOutput(args ...string) ([]byte, error) {
+	// Check that we have no more than the expected number of call
+	assert.Less(t.t, t.index, len(t.calls))
+	// Check that we have the expected number of parameters for this call
+	assert.Equal(t.t, len(t.calls[t.index].args), len(args))
+	// Check that each argument match our expectation
+	if len(t.calls[t.index].args) == len(args) {
+		for index, value := range args {
+			assert.Equal(t.t, t.calls[t.index].args[index], value)
+		}
+	} else {
+		fmt.Printf("%v\n", args)
+	}
+
+	ret, err := t.calls[t.index].ret, t.calls[t.index].err
+	t.index++
+
+	return ret, err
+}
+
+func (t *testCommandCall) DirSet(dir string) {
+	// Check that we have no more than the expected number of call
+	assert.Less(t.t, t.index, len(t.calls))
+
+	assert.Equal(t.t, t.calls[t.index].dir.(string), dir)
+	t.calls[t.index].dirSet = true
+}
+
+func (t *testCommandCall) EnvSet(env []string) {
+	// Check that we have no more than the expected number of call
+	assert.Less(t.t, t.index, len(t.calls))
+
+	// Prepare array for comparison
+	expectedEnv := t.calls[t.index].env
+	if t.calls[t.index].osEnv {
+		expectedEnv = append(expectedEnv, os.Environ()...)
+	}
+	sort.Strings(expectedEnv)
+	sort.Strings(env)
+
+	// First check length of expected and passed environment are equal
+	assert.Equal(t.t, len(expectedEnv), len(env))
+	if len(expectedEnv) == len(env) {
+		// Check each independent environement variable match our expectation
+		for index, value := range env {
+			assert.Equal(t.t, expectedEnv[index], value)
+		}
+	}
+	t.calls[t.index].envSet = true
+}
+
+func (t *testCommandCall) VerifyExpectation() {
+	// Expected as many call as we got
+	assert.Equal(t.t, len(t.calls), t.index)
+	// Check if every call really matched our expectaction
+	for _, value := range t.calls {
+		if value.dir != nil {
+			assert.Equal(t.t, true, value.dirSet)
+		}
+		if len(value.env) > 0 {
+			assert.Equal(t.t, true, value.envSet)
+		}
+	}
+}
--- a/cmd/fyne/internal/commands/mock_runner_test.go
+++ b/cmd/fyne/internal/commands/mock_runner_test.go
@@ -30,7 +30,7 @@ type testCommandCall struct {
 	t     *testing.T
 }
 
-func (t *testCommandCall) RunOutput(args ...string) ([]byte, error) {
+func (t *testCommandCall) runOutput(args ...string) ([]byte, error) {
 	// Check that we have no more than the expected number of call
 	assert.Less(t.t, t.index, len(t.calls))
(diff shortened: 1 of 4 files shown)


andydotxyz:
Seems like a great idea. Let's get it based on top of the go version check once on develop then review.

cmd/fyne/internal/commands/mock_runner_test.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type testCommandCall struct {
+	calls []mockRunOutputValue
+	index int
+	t     *testing.T
+}
+
+func (t *testCommandCall) runOutput(args ...string) ([]byte, error) {
+	// Check that we have no more than the expected number of call
```
andydotxyz, line 34: Here the comment does not match the code (no more than vs less)

cmd/fyne/internal/commands/mock_runner_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"fmt"
+	"os"
+	"sort"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type mockRunOutputValue struct {
+	// Expected value
```
andydotxyz, line 13: If variable names need to be commented then their name is not clear enough :)

andydotxyz (changes requested):
Nice idea, helpful in testing this rather low-on-coverage package.

However from a style point of view we prefer clear names over comments.
Rule of thumb: If the code needs a comment then it should be clearer.

cmd/fyne/internal/commands/mock_runner_test.go
andydotxyz: There were three instances of this, only one is fixed.

cmd/fyne/internal/commands/mock_runner_test.go
```diff
@@ -0,0 +1,101 @@
...
+	t     *testing.T
+}
+
+func (t *testCommandCall) runOutput(args ...string) ([]byte, error) {
+	// Check that we have less than the expected number of call
+	assert.Less(t.t, t.index, len(t.calls))
+	// Check that we have the expected number of parameters for this call
+	assert.Equal(t.t, len(t.calls[t.index].args), len(args))
+	// Check that each argument match our expectation
+	if len(t.calls[t.index].args) == len(args) {
```
andydotxyz, line 46: Isn't this already asserted above?

andydotxyz: By wrapping this in an if we won't test any arguments and potential failures will be missed

cmd/fyne/internal/commands/mock_runner_test.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+type testCommandCall struct {
+	calls []mockRunOutputValue
+	index int
+	t     *testing.T
+}
+
+func (t *testCommandCall) runOutput(args ...string) ([]byte, error) {
+	// Check that we have less than the expected number of call
```
andydotxyz, line 41: There are a lot of checks for t.calls size and t.index being in-sync.
It got me wondering... The mock has a different ordinance value to the real version (i.e. n calls instead of 1)
would it not be simpler if the `testCommandCall` was, as the name implies, singular?

andydotxyz (commented):
Trying to find ways to simplify this. The comments make me feel it's not totally clear, so maybe it can be clearer?

cmd/fyne/internal/commands/mock_runner_test.go
Bluebugs: I think it will be impossible/limited to use if it was only testing one call as that means we would for example not be able to test build() which might call go once or twice.

cmd/fyne/internal/commands/mock_runner_test.go
andydotxyz: > The assert, as far as I could see, don't seem to stop the code and it will lead to a crash if they do not match instead of just reporting an error.

6 lines above there is an assert for the size of `t.calls` against `t.index` which is then used blindly, this would produce the same crash if the assert fails.
If you want the test to stop at a test failure use `require` package instead of `assert` and the safety code can be avoided.

cmd/fyne/internal/commands/mock_runner_test.go
andydotxyz: I think I understand it. 
Should we not be checking then that t.index is exactly 1 more than the length of t.calls?

cmd/fyne/internal/commands/mock_runner_test.go
Bluebugs: Good suggestion. I will update with require.

cmd/fyne/internal/commands/mock_runner_test.go
Bluebugs: runOutput is called for each command that is to be executed. t.index is the number of call that did happen so far, starting from 0. t.calls is the array that contain all the expected call. So t.index must be within t.call for runOutput to be a valid call which mean t.index < len(t.calls).

cmd/fyne/internal/commands/mock_runner_test.go
```diff
@@ -39,14 +39,12 @@ type testCommandCall struct {
...
 	// Check that we have the expected number of parameters for this call
-	assert.Equal(t.t, len(t.calls[t.index].args), len(args))
+	require.Equal(t.t, len(t.calls[t.index].args), len(args))
 	// Check that each argument match our expectation
-	if len(t.calls[t.index].args) == len(args) {
-		for index, value := range args {
-			assert.Equal(t.t, t.calls[t.index].args[index], value)
-		}
+	for index, value := range args {
+		require.Equal(t.t, t.calls[t.index].args[index], value)
```
andydotxyz, line 47: Good use of `require`, but in this case should it not still be `assert` so we continue to test the other items? Best to check as much as you can without crashing.

cmd/fyne/internal/commands/mock_runner_test.go
```diff
@@ -76,26 +74,24 @@ func (t *testCommandCall) setEnv(env []string) {
...
 }
 
 func (t *testCommandCall) verifyExpectation() {
 	// Expected as many call as we got
-	assert.Equal(t.t, len(t.calls), t.index)
+	require.Equal(t.t, len(t.calls), t.index)
 	// Check if every call really matched our expectaction
 	for _, value := range t.calls {
 		if value.dir != nil {
-			assert.Equal(t.t, true, value.dirSet)
```
andydotxyz: As above, don't fail on first mismatch dir

cmd/fyne/internal/commands/mock_runner_test.go
```diff
@@ -76,26 +74,24 @@ func (t *testCommandCall) setEnv(env []string) {
...
-	assert.Equal(t.t, len(expectedEnv), len(env))
-	if len(expectedEnv) == len(env) {
-		// Check each independent environement variable match our expectation
-		for index, value := range env {
-			assert.Equal(t.t, expectedEnv[index], value)
-		}
+	require.Equal(t.t, len(expectedEnv), len(env))
+	// Check each independent environement variable match our expectation
+	for index, value := range env {
+		require.Equal(t.t, expectedEnv[index], value)
```
andydotxyz, line 80: Check all of env using `assert`

andydotxyz: If one is missing this is true - but what if one is incorrect... knowing if the others are also incorrect might be useful

Bluebugs: If one is incorrect the problem will happen very quickly as that will mess the sorting and the entire rest of the array might not match anymore (even if they do have the same number).

andydotxyz (approved):

PR #2764, opened 2022-02-05
Fix os/exec that has gotten back in

Jacalz:
### Description:

It seems like the `os/exec` package has managed to get back into the code base. This just fixes up a few cases where it was being used. For anyone not sure why this is the case, see https://go.dev/blog/path-security.

We probably should do some sort of blacklist of the package but I don't know how to do so in a good way so we'll just have to do it like this for now. I would appreciate if people could take a little extra care to not introduce `os/exec` again next time :)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/mobile/env.go
+++ b/cmd/fyne/internal/mobile/env.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"os/exec"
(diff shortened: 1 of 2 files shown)


andydotxyz (approved):
If we could block it that would be good. I also don’t know how…

PR #2765, opened 2022-02-05
Add selection validation

Jacalz:
### Description:

This is an initial version of selection validation for widgets that only requires items to be selected. I have added the support to CheckGroup and RadioGroup for now. I couldn't think of any good way to show the validaton to the user in this case, without using the HintText in `widget.Form`.

image

I intend to add more tests once people are happy with my idea.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
Sorry I forgot to comment here, will try to recall my comments from slack.
1) Should a "requires at least 1" be a Required validation instead? Or is this different / overlap? I'd like to start looking at form validations that check for required field.
2) Is the naming right? If the selection validation is just checking the number then I wonder if it really is a selectionCount (or similar) validation instead. Assuming other types could exist in the future..?

Jacalz:
A little late reply here, sorry about that. I’ll try to answer your questions below.
1. Hmm, good question. For required, I think we would need one for each type of validator. We could then have a `RequiredSelection()` that wants one item selected and also one `RequiredText()` that checks for non empty text. Does that seem sensible?
2. Naming it `SelectionCount` seems perfect to me. I’ll go ahead and update the PR with that.

andydotxyz:
> We could then have a RequiredSelection() that wants one item selected and also one RequiredText() that checks for non empty text. Does that seem sensible?

This feels overkill. We can just define that `Required` means that it must be "non-zero"` i.e. checked, not "", or len()>0? Not sure that extra, specific, APIs add to this.

Jacalz:
> This feels overkill. We can just define that `Required` means that it must be "non-zero"` i.e. checked, not "", or len()>0? Not sure that extra, specific, APIs add to this.

But those are two different validators. We can't just directly check no selection and no text input without different validators. How would you like that we do it instead?

andydotxyz:
> But those are two different validators. We can't just directly check no selection and no text input without different validators. How would you like that we do it instead?

I'm not sure that I agree that it cannot be done... but it may not be easy.
Perhaps it would be better / easier if `Required()` could be interface based instead of using the validator mechanism? Just thinking out loud.

Jacalz:
> I'm not sure that I agree that it cannot be done... but it may not be easy. Perhaps it would be better / easier if `Required()` could be interface based instead of using the validator mechanism? Just thinking out loud.

Hmm, I don’t know. I’m starting to feel like it probably is outside the scope of this PR.

andydotxyz:
> Hmm, I don’t know. I’m starting to feel like it probably is outside the scope of this PR.

Agreed.

I think that maybe the Required stuff should be separated so we can address it with the "*" indicator that forms might expect. So the required validator probably needs additional thought even if it is the approach, as we cannot currently tell if it represents a required, if that makes sense?

Jacalz:
That makes a lot of sense. The idea to make the form as required also seems like a splendid idea (I hadn’t thought about that). I’ll remove the required part of this PR.

Jacalz:
I have reverted the commits with the required stuff. I think this should be ready for a re-review now.

validation.go
```diff
@@ -15,3 +15,9 @@ type Validatable interface {
 //
 // Since: 1.4
 type StringValidator func(string) error
+
+// SelectionValidator is a function for requiring a specified
+// amount of selections to be made.
+//
+// Since: 2.2
+type SelectionValidator func(int) error
```
andydotxyz, line 23: This seems to be using the old naming I think.

data/validation/selectioncount.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"fmt"
+
+	"fyne.io/fyne/v2"
+)
+
+// NewSelectionCount returns a fyne.SelectionValidator that requires
+// that "required" amount or more items are selected.
+func NewSelectionCount(required int) fyne.SelectionValidator {
```
andydotxyz, line 11: I just realised this assumes that validation should only check a minimum. I would think that a maximum is more common, should this take 2 parameters instead?

andydotxyz (commented):
I think there's a few little issues left over

data/validation/selectioncount.go
Jacalz: Seems like a good idea. I'll get it updated.

validation.go
Jacalz: Good catch. I'll get that fixed.

Jacalz:
I think this should be ready for a re-review now.

validation.go
```diff
@@ -15,3 +15,8 @@ type Validatable interface {
 //
 // Since: 1.4
 type StringValidator func(string) error
+
+// CountValidator is a function sinature for validating counts (usually selections).
+//
+// Since: 2.2
+type CountValidator func(int) error
```
andydotxyz, line 22: I wonder, and sorry for back and forth, any reason this shouldn't be `IntValidator` to match `StringValidator` above? There is nothing about this type that requires it to refer to counts...

widget/check_group.go
```diff
@@ -18,10 +20,15 @@ type CheckGroup struct {
 	Options    []string
 	Selected   []string
 
+	Validator           fyne.CountValidator
```
andydotxyz, line 23: I can't help but shake the idea that this count/int validation isn't the primary type for `Check` (and possibly radio as well)... They are `string` or `[]string` data sources...
Will this possibly clash in the future if we want to validate the data more directly?
Should it be `SelectionValidator` instead?

validation.go
Jacalz: Good point. I hadn't thought about that at all.

widget/check_group.go
Jacalz: I guess you mean `StringValidator`? Hmm, that’s hard. I don’t think verifying strings makes sense for a widget where the user doesn’t input their own fields.

Sure, the developer might have set up some way to popular the items in the Select widget or Radio, but for the most part, it is just a case of users selecting from a predefined list. I personally think the the number of selections make more sense.

andydotxyz: Yes good points. But by calling this `Validator` we are committing to it being the main way of validating - but I am not sure that this makes sense, which is why I was suggesting the `SelectionValidation` name instead. When I look at other usage of `Validator` it is more comparing to the data itself rather than quantity of data selected...

Jacalz: Ah, I see now. I don't think I quite understood what you meant before. Given that the new validator was called `SelectionValidator` before, it caused some confusion.

andydotxyz: > Ah, I see now. I don't think I quite understood what you meant before. Given that the new validator was called `SelectionValidator` before, it caused some confusion.

Yeah sorry about that, I could have been clearer.

Jacalz:
Sorry for the force push. I had to get the PR rebased on develop to make the new CI actions run.

Jacalz:
As it currently stands, the Form is always displaying the error for these new validators and not respecting the widget focus like it is with the Entry. We need a better way of doing that. The way we are checking widget types and looking at internal state is not a good solution.

andydotxyz:
> As it currently stands, the Form is always displaying the error for these new validators and not respecting the widget focus like it is with the Entry.

I think after previous call we agreed that focus state should not be important for form validation - what matters is not bugging the user until they have had a chance to interact with the widget. I.e. new errors are always shown?

andydotxyz:
I have lost track of where we are on this PR if I'm honest. @Jacalz are there changes pending?

Jacalz:
I believe that I didn't quite know how to proceed and just let it decay :/

Jacalz:
I am also, to this day, not entirely sure how to proceed with this honestly. It always seems easier to build an API around how it is to be used but we never really got any input on usage

Jacalz:
I have rebased this on top of the latest work on `develop` now. This does include the change by Lucor to use `reflect` for looking up if the widget contains a validator or not. With some more eyes on this, I think we actually might get this through the door in a not too distant future :)

Jacalz:
This never really turned into something good. I'll close this.

PR #2773, opened 2022-02-12
Add a GitHub action to check license of dependencies

### LINKED ISSUE #2767: Check license of dependencies as part of Fyne PR/release process
To make sure that we do not inadvertently depend on a module with the wrong license it would be nice to integrate a tool in our github action that does check all the license for compatibility purpose (License compliance)

No new API should be necessary for this as this will just be part of our github workflow.

@lucor pointed out tools like: https://github.com/google/go-licenses for this purpose.

lucor:
### Description:

This PR adds a GitHub action to check license of dependencies using the [lucor.dev/lian](https://lucor.dev/lian) tool.

It will check the licenses of dependencies listed into the go.mod against a set of allowed types.

Fixes #2767

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/.github/workflows/license_check.yml
+++ b/.github/workflows/license_check.yml
@@ -1,0 +1,18 @@
+name: Check license of dependencies
+on: [push, pull_request]
+
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: WillAbides/setup-go-faster@v1.7.0
+      with:      
+        go-version: '1.18beta2'
+
+    - name: install lian
+      run: go install lucor.dev/lian@latest
+
+    - name: Check license of dependencies against go.mod
+      run: lian -d --allowed="Apache-2.0, BSD-2-Clause, BSD-3-Clause, MIT"
+
--- a/.github/workflows/license_check.yml
+++ b/.github/workflows/license_check.yml
@@ -8,7 +8,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: WillAbides/setup-go-faster@v1.7.0
       with:      
-        go-version: '1.18beta2'
+        go-version: '1.18.x'
 
     - name: install lian
       run: go install lucor.dev/lian@latest


.github/workflows/license_check.yml
```diff
@@ -0,0 +1,18 @@
...
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: WillAbides/setup-go-faster@v1.7.0
+      with:      
+        go-version: '1.18beta2'
+
+    - name: install lian
+      run: go install lucor.dev/lian@latest
```
Jacalz, line 14: Development versions in production? I’m not so sure about that. Can’t we stick to a stable release?

lucor: the tool requires go 1.18 so it is needed if we want to install via `go install`. 
We could install using the binary but in this way we need to use a specific version. (the tool works but it is into an early stage so an update could be required soon :) )
If the latest is ok, we could also integrate into the static check workflow instead of have a dedicated one.

andydotxyz: > the tool requires go 1.18 so it is needed if we want to install via `go install`.

We should not depend upon Go 1.18 before it is released IMHO.

Jacalz: I agree with @andydotxyz  about that. What is the reason for the 1.18 requirement after all? Not everyone will have Go 1.18 right as it is released, so it makes it a little harder to run locally for some. Is there any chance that the requirement could be lowered? If not, I think we sadly need to wait a few weeks for the stable release.

lucor: The tool uses [buildinfo.ReadFile](https://pkg.go.dev/debug/buildinfo@go1.18rc1#ReadFile) that is available starting from 1.18, hence the dependency at compile time.
We'd install from binary and update later to use `go install` but since go 1.18, as @Jacalz said, should be available in a few weeks I'd say it is ok to wait and update the PR once 1.18 is released. WDYT?

Jacalz: Ah, I see. Sounds great to me 👍

Jacalz: This should be unblocked now that Go 1.18 finally has been released :)

lucor: Done in [881be11](https://github.com/fyne-io/fyne/pull/2773/commits/881be112a5e16f505b6c614eb8d73cc8f0a08b3d)

.github/workflows/license_check.yml
```diff
@@ -0,0 +1,18 @@
+name: Check license of dependencies
+on: [push, pull_request]
```
Jacalz, line 2: Just a thought here. Do we need the workflow to run on push? It seems to me like any new module additions or updates would have to go though PRs either way.

Jacalz (approved):
Looks good to me. Just one question about if we actually need it to be run in every push.

.github/workflows/license_check.yml
Bluebugs: I feel it is a nice safety measure just in case, "belt and suspender" kind of approach.

Bluebugs (approved):

Jacalz:
I realised now that there is a minor problem with this new check. It is called “checks” just as the static analysis is. We might want to rename the static analysis workflow to “static_analysis” and perhaps give this one a better name as well. Another option would also be to move this check into the static analysis when we get that updated to Go 1.18.

PR #2794, opened 2022-02-17
If FixedSize is set without a preferred size we should fall back to minimum

### LINKED ISSUE #2784: Window not shown when SetFixedSize is used without Resize
### Describe the bug:

Window not shown when SetFixedSize is used without Resize.
Adding a call to Resize shows the window.

Seems to be a regression since 2.0.x

Discussion on https://gophers.slack.com/archives/CB4QUBXGQ/p1644939596121299

### Example code:

```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Hello")

		hello,
			hello.SetText("Welcome :)")
		}),
	))
	w.SetFixedSize(true)

	w.ShowAndRun()
}
```

 - **OS:** Mac os 
 - **Version:** 12.2.1 on an M1 
 - **Go version:** 1.17.6
 - **Fyne version:** 2.1.2

andydotxyz:
Avoid crashes by being larger than 1x1

Fixes #2784

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -219,6 +219,10 @@ func (w *window) FixedSize() bool {
 func (w *window) SetFixedSize(fixed bool) {
 	w.fixedSize = fixed
 
+	if fixed && (w.requestedWidth == 0 || w.requestedHeight == 0) {
+		bigEnough := w.canvas.canvasSize(w.canvas.Content().MinSize())
+		w.Resize(bigEnough)
+	}
 	if w.view() != nil {
 		w.runOnMainWhenCreated(w.fitContent)
 	}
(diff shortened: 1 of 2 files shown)


internal/driver/glfw/window.go
```diff
@@ -501,6 +505,11 @@ func (w *window) SetContent(content fyne.CanvasObject) {
 	}
 
 	w.canvas.SetContent(content)
+	if !visible && w.fixedSize && (w.requestedWidth == 0 || w.requestedHeight == 0) {
+		bigEnough := w.canvas.canvasSize(content.MinSize())
+		w.Resize(bigEnough)
+	}
```
Bluebugs, line 511: This look like a bit of duplicated code to me. Would it be possible to put most of the if tests along with the code inside the if in its own function?

Bluebugs (changes requested):

internal/driver/glfw/window.go
andydotxyz: Yeah could do. Initially it seemed like longer code - but the duplication is undesirable, will fix

Bluebugs (approved):

PR #2795, opened 2022-02-18
Fix issue with wrapping moving cursor too far after backspace

### LINKED ISSUE #2698: Some incorrect cursor movement in a multiline entry with wrapping

andydotxyz:
Fixes #2698

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -907,7 +907,8 @@ func (e *Entry) rowColFromTextPos(pos int) (row int, col int) {
 				row++
 			}
 			col = pos - b.begin
-			if canWrap && b.begin == pos && col == 0 && pos != 0 && row < (totalRows-1) {
+			// if this gap is at `pos` and is a line wrap, increment (safe to access boundary i-1)
+			if canWrap && b.begin == pos && pos != 0 && provider.rowBoundary(i-1).end == b.begin && row < (totalRows-1) {
 				row++
 			}
 		} else {
--- a/widget/entry_test.go
+++ b/widget/entry_test.go
@@ -512,12 +512,26 @@ func TestEntry_OnKeyDown_Backspace(t *testing.T) {
 	assert.Equal(t, 0, entry.CursorRow)
 	assert.Equal(t, 2, entry.CursorColumn)
 
-	key := &fyne.KeyEvent{Name: fyne.KeyBackspace}
-	entry.TypedKey(key)
+	backspace := &fyne.KeyEvent{Name: fyne.KeyBackspace}
+	entry.TypedKey(backspace)
 
 	assert.Equal(t, "H", entry.Text)
 	assert.Equal(t, 0, entry.CursorRow)
 	assert.Equal(t, 1, entry.CursorColumn)
+
+	entry = widget.NewMultiLineEntry()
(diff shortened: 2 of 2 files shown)


Jacalz (approved):
Nice work!

PR #2796, opened 2022-02-18
Resolve issue when last item of a composite line is non-text

### LINKED ISSUE #2695: Text and links in markdown are rendered in each other
### Describe the bug:
Text and links in markdown are rendered in each other

### Screenshots:
image

### Example code:
```go
container.NewTabItem("Example", container.New(
			layout.NewGridLayout(2),
			widget.NewRichTextFromMarkdown("Text [a link](https://github.com/fyne-io/fyne) more text [another link](https://github.com/fyne-io/fyne)"),
			)),
```

### Device:
 - **OS:** Windows
 - **Version:** 11
 - **Go version:** 1.17.3
 - **Fyne version:** 2.1.2

andydotxyz:
Fixes #2695

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/richtext.go
+++ b/widget/richtext.go
@@ -455,15 +455,15 @@ func (r *textRenderer) Layout(size fyne.Size) {
 			i++
 			_, isText := obj.(*canvas.Text)
 			if !isText && !inline {
+				if len(rowItems) != 0 {
+					width, _ := r.layoutRow(rowItems, rowAlign, left, yPos, lineWidth)
+					left += width
+				}
 				height := obj.MinSize().Height
 
 				obj.Move(fyne.NewPos(left, yPos))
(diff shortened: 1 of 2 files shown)


Jacalz (commented):
Should this really show the hovered line all the way to the end? Otherwise it's looking good.

image

andydotxyz:
That is a good question @Jacalz, but not one with a simple answer.
I'd agree that the visual outcome is not desirable - but if the element was something other than a link (image etc) should it take up the space? Maybe it is a special case for Hyperlink, or maybe the hyperlink should only underline under text instead of full width? I think it's a separate issue as if you put a link into a window content it will do the same outside of RichText.

Jacalz:
Okay, I see. I personally think it should only highlight under the actual text. Seems more sensible to me.

Issue #2775, opened 2022-02-13
Alternating grid.Add() and grid.Remove() lets program crash


midu-SA:
### Describe the bug:
I create 100 buttons (stored in an array), which I alternatingly add into and remove from a grid, using grid.Add() and grid.Remove(). After a while the program crashes.

### To Reproduce:
Steps to reproduce the behaviour:

Run the code provided below, where 
- a grid is created
- 100 buttons are created, which are stored in an array
- an add-button is created to add these 100 buttons to the grid using grid.Add() method
- a remove-button is created to remove these 100 buttons from the grid using grid.Remove() method

Now use the add-button and the remove-button alternatingly a few times. After a while the program crashes. The crash does not happen always at the same step. But it always crashes when grid.Remove() is executed. It seems to be a memory problem which is not deterministic, e.g. the crash might happen when button no. 17 is removed, or no. 35 or no. 72, or ...

The output in a terminal is e.g.
```
...
i= 32
i= 31
i= 30
i= 29
i= 28
i= 27
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x62679e]

goroutine 10 [running, locked to thread]:
fyne.io/fyne/v2/layout.(*gridLayout).MinSize(0xc0005c98d0, {0xc0000c4000, 0x1d, 0x6255a6})
        /home/opc/go/pkg/mod/fyne.io/fyne/v2@v2.1.2/layout/gridlayout.go:131 +0x2fe
fyne.io/fyne/v2.(*Container).MinSize(0xc0005c9849)
        /home/opc/go/pkg/mod/fyne.io/fyne/v2@v2.1.2/container.go:71 +0x72
fyne.io/fyne/v2/layout.(*boxLayout).MinSize(0xc0005c9849, {0xc0001712c0, 0x3, 0x0})
        /home/opc/go/pkg/mod/fyne.io/fyne/v2@v2.1.2/layout/boxlayout.go:140 +0x15d
fyne.io/fyne/v2.(*Container).MinSize(0x463c25)
        /home/opc/go/pkg/mod/fyne.io/fyne/v2@v2.1.2/container.go:71 +0x72
fyne.io/fyne/v2/internal/driver/glfw.(*glCanvas).MinSize(0xc0001219e0)
[...]

PR #2798, opened 2022-02-20
Container safety

Jacalz:
### Description:

This just adds a little more safety for containers to make sure that we don't add nil objects.
Not intended as a fix for #2775, but should improve the situation.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Agreed, thanks

PR #2802, opened 2022-02-23
Feature/web gopherjs build package serve

Bluebugs:
### Description:
Introduce gopherjs port and an automated mode to switch from wasm to gopherjs when the user is using Chrome. Default is still set to wasm as it doesn't require the installation of any additional tool.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


Bluebugs:
> Coverage decreased (-0.1%) to 61.461% when pulling **[112af47](https://github.com/fyne-io/fyne/commit/112af474076973896f5c2ceac9be08cc369254f1) on feature/web-gopherjs-build_package_serve** into **[87f66cf](https://github.com/fyne-io/fyne/commit/87f66cf150ca973b89a5c002fa38e005f3a5d142) on develop**.

I am working on a follow up that add more tests for the package operation, but that require the ability to mock util and should really be part of a follow up PR.

cmd/fyne/internal/commands/build.go
```diff
@@ -89,10 +99,10 @@ func (b *builder) build() error {
 		tags = append(tags, "release")
 	}
 	if len(tags) > 0 {
-		args = append(args, "-tags", strings.Join(tags, ","))
+		args = append(args, "--tags", strings.Join(tags, ","))
```
Jacalz, line 102: I think the actual Go syntax for flags is to prefer one dash over two. I think we might want to keep it as it was.

Jacalz (changes requested):
Just one minor note. Otherwise, it  looks good to me.

cmd/fyne/internal/commands/build.go
andydotxyz: Agreed, I think there are 7 instances, many are new - they should all be "-tags" iirc

cmd/fyne/internal/commands/build.go
```diff
@@ -46,37 +50,43 @@ func checkGoVersion(runner runner, versionConstraint *version.ConstraintGroup) e
...
 	if goos == "darwin" {
 		env = append(env, "CGO_CFLAGS=-mmacosx-version-min=10.11", "CGO_LDFLAGS=-mmacosx-version-min=10.11")
 	}
 
-	if goos == "windows" {
-		if b.release {
-			args = append(args, "-ldflags", "-s -w -H=windowsgui")
-		} else {
-			args = append(args, "-ldflags", "-H=windowsgui")
+	if !isWeb(goos) {
```
andydotxyz, line 80: There is an identical check 8 lines above, could it be simplified by combining?
Most likely by moving the `if goos == "darwin" ` into the `if !isWeb` :)

cmd/fyne/internal/commands/build.go
```diff
@@ -110,7 +120,7 @@ func (b *builder) build() error {
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "%s\n", string(out))
 	}
-	return err
+	return []string{b.target}, err
```
andydotxyz, line 123: Why is `b.build` returning `b.target`? This state is available to whoever called the build command, and as it is a method that executes change it should probably not return like a getter.

cmd/fyne/internal/commands/build_test.go
```diff
@@ -132,7 +134,7 @@ func Test_BuildWasmReleaseVersion(t *testing.T) {
 		},
 		{
 			expectedValue: expectedValue{
-				args:  []string{"build", "-ldflags", "-s -w", "-tags", "release"},
+				args:  []string{"build", "--tags", "release"},
```
andydotxyz, line 137: As noted above many "--tags" in this PR should remain as "-tags"

cmd/fyne/internal/commands/package-wasm.go
```diff
@@ -12,8 +10,11 @@ import (
 )
 
 type indexData struct {
-	Name       string
-	IsReleased bool
+	WasmFile     string
```
andydotxyz, line 13: This file name is "package-wasm.go" should it be part of the new "package-web.go" instead?

cmd/fyne/internal/util/file.go
```diff
@@ -27,6 +27,22 @@ func CopyExeFile(src, tgt string) error {
 	return copyFileMode(src, tgt, 0755)
 }
 
+// WriteFile write the content of data into a regular file, target.
+func WriteFile(target string, data []byte) error {
```
andydotxyz, line 31: I would recommend just using `ioutil.WriteFile` instead of adding a new helper

andydotxyz (changes requested):
Nice work, just a few things inline.

Broader question - do we want to offer the choice of "wasm", "gopherjs" and "web" in the `fyne serve`? Will any users want to control that? (especially given that we are providing alternatives to work around issues with some...)

I'm worried that we will get bug reports for example of "I'm using wasm and it crashes in Chrome"

cmd/fyne/internal/commands/package-wasm.go
Bluebugs: I kind of kept them separated as I was hoping that the maintenance of all this path would be temporary and we would get to wasm only shortly. This seems very unlikely now and I can definitively merge them into one file.

cmd/fyne/internal/util/file.go
Bluebugs: Oh, I didn't found that one. I only saw io.WriteFile which wasn't available in 1.14. Will switch to that.

andydotxyz: Yeah, the older `ioutil` package is being replaced by new `io` features and it will go away in the future - but not before Go 2.0 :)

cmd/fyne/internal/commands/package-web.go
```diff
@@ -9,60 +10,101 @@ import (
...
-	appDir := util.EnsureSubDir(p.dir, "web")
+type webData struct {
+	WasmFile     string
+	GopherJSFile string
+	IsReleased   bool
+	HasWasm      bool
+	HasGopherJS  bool
+}
 
+func packageWebInternal(w webData, appDir string, exeWasmSrc string, exeJSSrc string, icon string, release bool) error {
```
andydotxyz, line 21: The methods are now out-of-order, functions should appear after types and their methods.
We haven't found a way to automate this check sorry!

andydotxyz (approved):

Jacalz (approved):

PR #2807, opened 2022-02-25
Add support for OpenURL to the web drivers (GopherJS and Wasm).

### LINKED ISSUE #2736: Add OpenURL support to the web driver
The goal of this issue is to add the capability to open a new browser tab to a specific URL as part of porting the OpenURL API to the web driver.

There is no need to add any new API as this is about supporting existing API.

It should be possible to have the web driver trigger a JS call that will open a new window/tab that target the given URL.

Bluebugs:
### Description:
This make OpenURL work for both web drivers. This doesn't had any new dependencies, but make indirect one, direct.

Fixes #2736

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/app/app_goxjs.go
+++ b/app/app_goxjs.go
@@ -6,20 +6,12 @@
 package app
 
 import (
-	"errors"
-	"net/url"
-
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/theme"
 )
 
 func defaultVariant() fyne.ThemeVariant {
 	return theme.VariantDark
-}
-
-func (app *fyneApp) OpenURL(url *url.URL) error {
-	// TODO #2736
-	return errors.New("OpenURL is not supported yet with GopherJS backend.")
 }
 
 func (app *fyneApp) SendNotification(_ *fyne.Notification) {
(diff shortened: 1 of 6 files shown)


andydotxyz:
Looks good, but I wanted to check something. There are 3 web related tags "js" "wasm" and "web".
This PR does not handle the "web" target as far as I can see - is this a mistake?

Bluebugs:
> Looks good, but I wanted to check something. There are 3 web related tags "js" "wasm" and "web". This PR does not handle the "web" target as far as I can see - is this a mistake?

I didn't think it did make sense in the "web" target case, as the goal of the web target case at the moment is to run natively to test the web backend without going through a browser, but still using glfw-js and gl-js. I could move the web tag to use the native platform one, but I think it will add complexity to the maintenance of those platform with not useful benefit at this point.

andydotxyz:
> I didn't think it did make sense in the "web" target case, as the goal of the web target case at the moment is to run natively to test the web backend without going through a browser.

I agree, but when compiled with `-tags test_web_driver` which file will provide the `OpenURL` function? (in the past the goxjs file would handle all 3 targets)

Bluebugs:
> I agree, but when compiled with `-tags test_web_driver` which file will provide the `OpenURL` function? (in the past the goxjs file would handle all 3 targets)

Oh, I see. I think I forgot adding a file.

andydotxyz (approved):
Nice :)

PR #2808, opened 2022-02-25
Feature/mock util cmd

Bluebugs:
### Description:
This patch series introduce the ability to actually test the behavior of generating Package and use it to validate that packaging for web, gopherjs and wasm work and produce the same things over time.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/get.go
+++ b/cmd/fyne/internal/commands/get.go
@@ -8,7 +8,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"fyne.io/fyne/v2/cmd/fyne/internal/util"
 	"github.com/urfave/cli/v2"
 	"golang.org/x/sys/execabs"
 )
--- a/cmd/fyne/internal/commands/install.go
+++ b/cmd/fyne/internal/commands/install.go
@@ -10,7 +10,6 @@ import (
 
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/cmd/fyne/internal/mobile"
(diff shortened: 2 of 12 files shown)


cmd/fyne/internal/commands/get.go
```diff
@@ -8,7 +8,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"fyne.io/fyne/v2/cmd/fyne/internal/util"
```
andydotxyz, line 11: This (plus others later in) cannot be right?
Files don't compile if the import isn't used - so either this didn't compile before (actions should have caught it) or it doesn't compile now...

andydotxyz: I'm not sure how it's possible that the actions were passing before and after this change - can you explore?

andydotxyz (changes requested):

cmd/fyne/internal/commands/get.go
Bluebugs: They are right. That's how mock work, by intercepting the access to io operation and allowing them to get verified on how they are called and what they return. The package util file will do the default redirection, when no mock are in place while the package util mock test will do that redirection when running the tests.

Bluebugs: And yes, it does compile :-)

andydotxyz: My apologies, I completely missed that is how it works. Very slick :)

andydotxyz (approved):

PR #2812, opened 2022-02-28
Feature/systray

### LINKED ISSUE #283: Add system tray support
sometimes I need the app to minimize, and stay in the system tray, will `fyne` support system tray?

andydotxyz:
### Description:
Add support for system tray on macOS, Windows and Linux/BSD.
This adds a 6% increase in binary size, but I cannot find a way to do this with less.

Fixes #283

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
Can you clarify what you mean by whether “the icon approach” being right @Bluebugs ? I’m not sure what aspect is being questioned.

Bluebugs:
My understanding is that the current PR does use the application icon as the one for the systray icon. This might be limiting, especially because icon are usually "animated" or show some kind of state information by changing there content, which I don't think is necessarily expected for the application icon. Basically I am not sure that they have the same purpose.

andydotxyz:
> My understanding is that the current PR does use the application icon as the one for the systray icon. This might be limiting, especially because icon are usually "animated" or show some kind of state information by changing there content, which I don't think is necessarily expected for the application icon. Basically I am not sure that they have the same purpose.

Thanks. The plan is that this will be the fallback but developers will be able to specify an override, I hope this makes sense (I do still need to add that API).
Plan is that by default the icons will be the same (no user surprise), but that they can be animated etc by setting different ones.

Bluebugs:
> The plan is that this will be the fallback but developers will be able to specify an override, I hope this makes sense (I do still need to add that API). Plan is that by default the icons will be the same (no user surprise), but that they can be animated etc by setting different ones.

That does make sense indeed.

andydotxyz:
Icon replace is in and this seems feature complete.
I cannot figure how to make tests for it, but open to suggestions.

Bluebugs (dismissed):

stuartmscott:
> I cannot figure how to make tests for it, but open to suggestions.

Seems tricky as it is such an OS-dependent API. I'm not even sure you can programmatically trigger system tray elements on most OSes.

It might be good to add a couple of usage tests tho - to ensure it doesn't break if someone calls the APIs, or calls them in the wrong order, or updates the menu & icon every frame, and that the icon adjusts to theme-changes etc.

andydotxyz:
> It might be good to add a couple of usage tests tho - to ensure it doesn't break if someone calls the APIs, or calls them in the wrong order, or updates the menu & icon every frame, and that the icon adjusts to theme-changes etc.

But if we call any of these APIs then icons will appear and disappear from system tray, something we generally try to avoid in tests. Such testing should be in the `fyne-io/systray` package not here I would think.

andydotxyz:
> Running fyne test already flashes up windows and changes the menus in the top bar

Only on macOS IIRC. And it's a bit of a regression, we used to manage without visually interfering with the screen.

andydotxyz:
Sorry @Bluebugs I did an update and it invalidated your review :(

stuartmscott:
> I would happily add some Os specific testing to the "systray" project

Sounds good!

The only other thing to check is that the Icon updates on theme change

andydotxyz:
> The only other thing to check is that the Icon updates on theme change

Wow now that is an interesting thought. I don't think any of our app/window icon code refreshes on theme change.
Most icons like that are not part of the theme but I guess it is entirely possible that they could be and/or would want to match the current settings.

stuartmscott (approved):
Remaining items could land in a follow up PR. This feature is in high demand so let's land it and get it in the hands of devs for testing.

andydotxyz:
> Remaining items could land in a follow up PR.

Added #2842 to follow up on the icon theme change

andydotxyz:
Jacob said continue without him, so I did

Issue #2592, opened 2021-10-26
Windows: App Icon in Notification


luciferchase:
### Is your feature request related to a problem? Please describe:

There is no App Icon in the notification sent by Fyne App (At least in windows, haven't checked with other platforms). 

![fyne-notification](https://user-images.githubusercontent.com/48217707/138925843-f3ecc8ec-6075-4fdd-b03d-a25ccfbd6a06.png)

### Is it possible to construct a solution with the existing API?

No.

### Describe the solution you'd like to see:

Include app icon both in the toast and in the notification centre as well like this?

![notification-with-icon](https://user-images.githubusercontent.com/48217707/138926165-247d3100-a09d-4319-8c62-d5b2b55eb34a.png)

andydotxyz:
As best as I can tell this metadata comes from the packaged app metadata.
This metadata is only added when preparing for release ("fyne release") with the appropriate signing etc.
Can you please try packaging a release and install the result to see if it is fixed?

luciferchase:
I would love to do so, but unfortunately, I don't know how to do certification, signing etc. Sorry!

montovaneli:
Regarding this issue, I analyzed the app_windows.go file and saw that the toast template used is ToastText02, which does not allow images.

Moreover, in PowerShell, the Windows Toast notification requires a file path for the icon.

To work around this, I changed the PowerShell script template to:

```
const notificationTemplate = `$title = "%s"
$content = "%s"
$iconPath = "file:///%s" # The icon path must be a file URI

[Windows.UI.Notifications.ToastNotificationManager, Windows.UI.Notifications, ContentType = WindowsRuntime] > $null
$template = [Windows.UI.Notifications.ToastNotificationManager]::GetTemplateContent([Windows.UI.Notifications.ToastTemplateType]::ToastImageAndText02)
$toastXml = [xml] $template.GetXml()

$toastXml.GetElementsByTagName("text")[0].AppendChild($toastXml.CreateTextNode($title)) > $null
$toastXml.GetElementsByTagName("text")[1].AppendChild($toastXml.CreateTextNode($content)) > $null

$toastXml.GetElementsByTagName("image")[0].SetAttribute("src", $iconPath) # Set Icon

$xml = New-Object Windows.Data.Xml.Dom.XmlDocument
$xml.LoadXml($toastXml.OuterXml)
$toast = [Windows.UI.Notifications.ToastNotification]::new($xml)
[...]

Issue #936, opened 2020-05-05
Android notification uses fallback icon instead of app icon


andydotxyz:
We should show app icon not the android...

hkparker:
After working on #5667 I think the next step here is to allow the users to set a "notification icon" during `fyne package` that is used in all the notifications, and when one isn't provided fall back to the default icon like that PR does.  I think this is worth doing before trying to automatically turn the default icon into something that is acceptable to android.

If that sounds good I can start adding that to the existing PR, otherwise I'll close it out.

andydotxyz:
What it should fall back to is the app icon not the android fallback IMHO.

hkparker:
Ah yes that's what I meant, that's what the current PR does.

Most people I think would see an issue and create their own notification icon, but later we could also try to create a better notification icon automatically when one isn't provided.

PR #2813, opened 2022-02-28
Feature/metadataapi

### LINKED ISSUE #1940: sendnotification does not show app name on windows and never show icon you set
hey
for notification whatever i do name of app don't change like changing app name or set it on compile and ....
but it show app name appID still on windows

andydotxyz:
Add support for reading metadata through API.
More work to follow, including tests. Marked draft for discussion.

Fixes #1940 (indirectly)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


cmd/fyne/internal/commands/build.go
```diff
@@ -60,6 +66,24 @@ func (b *builder) build() error {
...
+		vars = append(vars, fmt.Sprintf("-X 'fyne.io/fyne/v2/internal/app.MetaID=%s'", b.id))
+	}
+	if b.name != "" {
+		vars = append(vars, fmt.Sprintf("-X 'fyne.io/fyne/v2/internal/app.MetaName=%s'", b.name))
+	}
+	if b.version != "" {
+		vars = append(vars, fmt.Sprintf("-X 'fyne.io/fyne/v2/internal/aapp.MetaVersion=%s'", b.version))
+	}
+	if b.buildNum != 0 {
+		vars = append(vars, fmt.Sprintf("-X 'fyne.io/fyne/v2/internal/app.MetaBuild=%d'", b.buildNum))
```
Jacalz, line 81: This is far from performance sensitive code, but isn't Sprintf rather unnecessary here?

cmd/fyne/internal/commands/build.go
```diff
@@ -60,6 +66,24 @@ func (b *builder) build() error {
...
+func (b *builder) generateMetaLDFlags() string {
+	var vars []string
+	if b.id != "" {
+		vars = append(vars, fmt.Sprintf("-X 'fyne.io/fyne/v2/internal/app.MetaID=%s'", b.id))
+	}
+	if b.name != "" {
+		vars = append(vars, fmt.Sprintf("-X 'fyne.io/fyne/v2/internal/app.MetaName=%s'", b.name))
+	}
+	if b.version != "" {
+		vars = append(vars, fmt.Sprintf("-X 'fyne.io/fyne/v2/internal/aapp.MetaVersion=%s'", b.version))
```
lucor, line 78:
```suggestion
		vars = append(vars, fmt.Sprintf("-X 'fyne.io/fyne/v2/internal/app.MetaVersion=%s'", b.version))
```

lucor (changes requested):
Just a small typo, apart that LGTM

app/meta.go
```diff
@@ -0,0 +1,30 @@
...
+	"fyne.io/fyne/v2"
+	intapp "fyne.io/fyne/v2/internal/app"
+)
+
+var (
+	meta fyne.AppMetadata
+)
+
+func init() {
+	build, err := strconv.Atoi(intapp.MetaBuild)
```
Jacalz, line 15: Is it not possible to inject an integer directly instead, or does it only support strings?

Jacalz (commented):
This looks very promising to me. I know that this is a draft but I just had two questions inline.

cmd/fyne/internal/commands/build.go
```diff
@@ -60,6 +66,24 @@ func (b *builder) build() error {
 	return err
 }
 
+func (b *builder) generateMetaLDFlags() string {
+	var vars []string
```
Bluebugs, line 70: Wouldn't it be possible to instead just have an array here?

andydotxyz: I’m sorry I don’t really understand. Do you mean it should be fixed length?

app/meta.go
andydotxyz: Yeah -X only supports strings

cmd/fyne/internal/commands/build.go
andydotxyz: Would you prefer strconv and a “+”?

cmd/fyne/internal/commands/build.go
andydotxyz: Thanks!

cmd/fyne/internal/commands/build.go
Jacalz: You can do however you like with that one, I guess. The ones above (which I now released was the intended source of this comment) through, are just strings so they should be as easy as a a simple concatenation.

andydotxyz: Ah I see, good point

cmd/fyne/internal/commands/build.go
Bluebugs: Sorry, I think I miss targeted the wrong line. I was talking about that series of if below which repeat the same code pretty much all over the place. Would be nicer if it was just an array being processed. Would be easier and cleaner also to extend over time.

app.go
```diff
@@ -88,6 +93,22 @@ func CurrentApp() App {
...
+	// ID is the unique ID of this application, used by many distribution platforms.
+	ID string
+	// Name is the human friendly name of this app.
+	Name string
+	// Version represents the version of this application, normally following semantic versioning.
+	Version string
+	// Build is the build number of this app, some times appended to the version number.
+	Build int
+	// Icon contains, if present, a resource of the icon that was bundled at build time.
+	Icon Resource
```
Jacalz, line 109: Doesn't this mean that each application now bundles the application icon twice?

Also, are we really sure that we want to include the icon this way? Seems like a very inefficient way to store it as a string and then read from Base64 each time we start the application.

cmd/fyne/internal/commands/build.go
andydotxyz: The alternative seems to be this, is it better?
```go
	buildIDString := ""
	if b.buildNum > 0 {
		buildIDString = strconv.Itoa(b.buildNum)
	}
	inserts := [][2]string{
		{"MetaID", b.id},
		{"MetaName", b.name},
		{"MetaVersion", b.version},
		{"MetaID", buildIDString},
	}
	var vars []string
	for _, item := range inserts {
		if item[1] != "" {
			vars = append(vars, "-X 'fyne.io/fyne/v2/internal/app."+item[0]+"="+item[1]+"'")
		}
	}
```

Bluebugs: Seems better and easier to extend.

andydotxyz:
I cannot figure out these test failures. They are passing correctly on my fork but failing here.
The report is specifying line numbers that don't exist (400 something on package_test.go that has just 180).
Anyone got an idea what is happening?

cmd/fyne/internal/commands/build.go
andydotxyz: Cool, done

app.go
andydotxyz: Where is the other place the icon is stored? Developers may have packaged the icon before this was possible, but we can move the app to use this icon by default I would think, so not packaged manually.

I cannot find another way to package it - the format put in must be something that can be serialised to the commandline args (a restriction of this process).

Bluebugs:
> I cannot figure out these test failures. They are passing correctly on my fork but failing here (and pass on my desktop too). The report is specifying line numbers that don't exist (400 something on package_test.go that has just 180). Also in my checkout / fork the test "Test_PackageWasm" does not appear, yet that is failing. Anyone got an idea what is happening?

This is very weird, but I have the impression that it is actually merging with develop and running the tests from develop. If you do a manual merge locally, you should be able to see them too.

andydotxyz:
> > I cannot figure out these test failures. They are passing correctly on my fork but failing here (and pass on my desktop too). The report is specifying line numbers that don't exist (400 something on package_test.go that has just 180). Also in my checkout / fork the test "Test_PackageWasm" does not appear, yet that is failing. Anyone got an idea what is happening?
> 
> This is very weird, but I have the impression that it is actually merging with develop and running the tests from develop. If you do a manual merge locally, you should be able to see them too.

Hmm, I thought I had done so, but maybe a day or two back was not enough ;)

andydotxyz:
> This is very weird, but I have the impression that it is actually merging with develop and running the tests from develop. If you do a manual merge locally, you should be able to see them too.

Amazing, this was bang on thanks. Fixed :)

Bluebugs (approved):

Jacalz:
I'll try to have a look at this later today or tomorrow.

app.go
Jacalz: I would vote for leaving the icon out for now and tackling that at a later date. Almost everyone has the icon packaged outside already, but that's only part of the problem that I see. If I remove that packaged icon and only rely on this metadata, I won't get any icon when I just build using `go build`.

andydotxyz: We could do it at another time I suppose yes.
I think we need to do either this or start shipping with icons in some other way - it's required to solve issues like #2592 and #936.

Jacalz: Can we not just use the icon that is passed to `app.SetIcon()` (and perhaps pass that into the metadata)? I think we need a better solution than passing it as a whole string with `fyne package`.

andydotxyz: Yes, users can bundle the icon manually and call SetIcon(). The metadata is static and will be the fallback so SetIcon will override it.
But why is the manual process favourable to an automated one?
An additional consideration - the MD5 is around half the number of bytes of the embedded resource, so not sure that it's a bad thing.

Jacalz: It's not necessarily that a manual process is better than an automated one. I'm simply saying that I think we need a better solution than including the icon as a string (and thus getting no icon when running `go build`) and parsing it on each application start. I don't know what you mean with MD5, that's a hashing algorithm (that doesn't change length).

Is it possible to do the Base64 decoding before adding it in, instead of after? If we do so instead, we at least avoid the slow reading and avoid essentially storing the icon twice in memory. But even if that works, I still think we need a better solution given the `go build` issue mentioned above.

andydotxyz: Sorry, by MD5 I meant Base64. It is not possible to add the raw bytes because this is LDFLAGS injecting a raw string into the source.

andydotxyz: Just for clarity, where in your app do you use the app icon that is required without the app being packaged?

Bluebugs:
This might be a silly idea, but could address @Jacalz concern. What do you think of having a "fyne metadata" command line that read the metadata and generate a go file instead. This could be used then by go generate. This would make go build actually work and provide metadata too.

Jacalz:
Can you elaborate a bit more in how you intend it to be used and how it is different from `fyne bundle`?

app.go
Jacalz: > Just for clarity, where in your app do you use the app icon that is required without the app being packaged?

I do use the icon inside the about page to display the application icon there. However, the big problem for me is that it means that, without packaging, the binaries will show no icon at all (or more like the "missing icon" icon) in the list of open applications. That will be very apparent on any posix operating systems (maybe on Windows as well, I can't remember).

This just seems like a step in the wrong direction to me. I'd suggest avoiding the icon inclusion here for now and try to make the Windows notifications just use the `app.SetIcon` icon instead.

Bluebugs:
> However metadata can change (version numbers etc) and forgetting to run this would mean the version info is out of date. Also exposing the `App.Metadata()` API will not be possible / reliable if the packaging of this data is optional...

I agree there is a synchronization problem. fyne package could maybe run go generate to make sure it is in sync, but for someone using go build, they would have to do it manually and thus the risk of not being in sync. Neither are ideal solution and both have drawback. I am personally fine with both as long as there drawback are documented.

andydotxyz:
> > However metadata can change (version numbers etc) and forgetting to run this would mean the version info is out of date. Also exposing the `App.Metadata()` API will not be possible / reliable if the packaging of this data is optional...
> 
> I agree there is a synchronization problem. fyne package could maybe run go generate to make sure it is in sync, but for someone using go build, they would have to do it manually and thus the risk of not being in sync. Neither are ideal solution and both have drawback. I am personally fine with both as long as there drawback are documented.

Ah cunning, combine the best of both. Inject latest into generated source code so go build still works. That sounds genius and may be possible. It might require another API because the app would have to inject the data instead of using the compiler. Will have to think about this more to see if it can be done more cleanly.

app.go
Jacalz: The step back is partly what you mention, but for me, the big issue is what I described above about the icon not being present in the operating system application list when you run it (unless you bundle a second icon or build using `fyne package`). As the `SetIcon()` call is used to give the icon to the operating system (so that it can show a correct icon among the open applications), that functionality would go missing without using the workarounds mentioned above.

andydotxyz: > As the `SetIcon()` call is used to give the icon to the operating system

This is highly platform specific so it's tough to talk in generalisations - however on most OS the app icon is provided by extra files in the package phase. The app icon is a fallback which is used primarily for app windows that don't have a specific icon. This is now picking up the packaged metedata as well. So from now on Linux window managers that use the window icon instead of the app icon will get this icon without going through the package phase - or requiring users to bundle data manually.

andydotxyz:
> This might be a silly idea, but could address @Jacalz concern. What do you think of having a "fyne metadata" command line that read the metadata and generate a go file instead. This could be used then by go generate. This would make go build actually work and provide metadata too.

I looked into this further and unfortunately I don't think it's practical. If the developer forgets to commit the file, and they use the `go build` then the compile will fail. With the ldflags approach in this PR it will not fail as the values fall back to default / "in development" info.

Jacalz:
I’m genuinely sorry but I would still vote for keeping the icon part out for now and implementing it at a later date. That way we can get more time to think about a better solution that doesn’t have the current drawbacks.

andydotxyz:
Info for the 3 different builds, "basic" is just `go build`, "bundled" adds the icon using `fyne bundle` and then "meta" uses this PR. I measure the time from start to main executing and from main executing to window entering foreground. This is with a 512x512 icon, 84kB.

| type | size | start time | show time |
| --- | --- | --- | --- |
| basic | - | 61ms | 238ms |
| bundle | +82kB | 56ms | 237ms |
| meta | +115kB | 59ms | 238ms |

Although the source code is much smaller (no bundle file, but even if there was the MD5 is half the size of the bundled byte array source) the compiled image data is around 2% smaller bundled (compared with image size) vs around 30% larger than the source image due to the MD5 storage format.
However you can see that startup time is pretty much the same.

I could look at different string encodings for the image data I guess.

andydotxyz:
Just a quick reminder of the current scanario:

* v2.1.4 and "go build": User does no icon bundling and no custom icon is set on a window or in the app bundle
* v2.1.4 with bundled icon and "go build": The users code will set the window icon but most OS will not show this for app icon.
* v2.1.4 with fyne package: OS will show app icon but based on OS the window icon may still be blank
* v2.1.4 with fyne package and bundled icon: Apps and windows will all have icon set.

As you can see above it basically requires user adding code into the app and bundling the icon manually to get it everywhere.

* This PR and "go build": icon is empty as before
* This PR and bundled icon with "go build": as above - as before, partial
* This PR and "fyne package": Icon will appear in all places without extra code
* This PR and "fyne package" with bundled icon: as before, all OK
* New possibility with this PR and "fyne build" the icon is included by default through FyneApp.toml or -icon param

The potential downside is that users bundling their own icon manually may have it stored twice when packaging for the benefit of controlling it's manual inclusion when not packaging and using "go build" only.

I can see that this is a tradeoff, but I think the lowered inconvenience for the primary target (packaging) is worth it for the possible downside during development (avoiding fyne tools and finding the icon not inserted unless code added).

Jacalz:
I think you are right. There is no better way around this. I don't know if it's worth it, but I guess we could consider using https://github.com/cristalhq/base64 if it makes any difference?

Another though that I had was if it makes sense to have an option to turn off the metadata? Something like a `nometadata` build tag that ignores loading it in and parsing it.

andydotxyz:
I tried some other encodings but they just didn't work (either invalid for a ldflags command line parameter, or much larger than comparison).
In the process I think I found that my previous measurements were not correct. So this shows updated size comparison:

Comparing this time to metadata included - but no icon metadata:
* base64 + 98k
* hex - 163k

So base64 is actually less than half the increase than I though it was compared to the efficient bundling of bytes directly (which was +82k). It makes me feel a little better.

lucor (approved):

Jacalz (approved):
Looks great. This is a really nice improvement. It looks like this branch isn't updated with the `fyne build`command but I guess that this will work it as well.

andydotxyz:
It didn't but I pushed an additional minor change over the top which added metadata to the serve command as well

PR #2816, opened 2022-03-01
Refactor, cleanup and use fyne bundle to store shaders.

### LINKED ISSUE #2804: Forgotten difference between WebGL mobile and WebGL desktop with GL ES
### Describe the bug:
Shader need to be emitted as GL ES when using WebGL on a Mobile platform.

### To Reproduce:
1. Try running fyne_demo or any fyne app inside firefox on an Android phone
2. Nothing is displayed as the shader can not compile

 - **OS:** Android/Firefox/Web backend
 - **Version:** All
 - **Go version:** 1.17
 - **Fyne version:** develop

### Fix
I somehow forgot in my series of patch introducing the wasm backend to switch shader to OpenGL ES when detecting that we are running on a mobile platform.

Bluebugs:
### Description:
This allow to use the same shader for both GL and GL ES case solving
issues for WebGL running on mobile. Additional benefit of this refactor is that now your IDE can actually do code completion/code highlight on the .glsl file.

Fixes #2804

I did manual test on the Web, Android native and Linux to check that it didn't break anything.

### Checklist:


andydotxyz:
The OpenGL (core) files were shader version 110 not 100 like the others.
I cannot remember why but I was pretty sure it was important.
GLES and OpenGL do not have the same shader version numbers as far as my research found...

Bluebugs:
> GLES and OpenGL do not have the same shader version numbers as far as my research found...

Indeed you are right. I think I am lucky that my desktop driver does support both GLSL 100 ES and GLSL 110 without complaining. I correct things to be supporting those two version of the language which are very identical if not for the precision information in our case.

andydotxyz (changes requested):
Nice work.
It looks like this was manually bundled - if that is the case can you please add the appropriate bundle commands to a "// go:generate" header? 
Then the "go generate" command will do it with no need to issue the specifics.

Bluebugs:
> Nice work. It looks like this was manually bundled - if that is the case can you please add the appropriate bundle commands to a "// go:generate" header? Then the "go generate" command will do it with no need to issue the specifics.

Oh, I was expecting the first call to fyne bundle to add the proper header to the file. I didn't check and missed that. Is there a reason for those header to not be added?

andydotxyz:
> Oh, I was expecting the first call to fyne bundle to add the proper header to the file. I didn't check and missed that. Is there a reason for those header to not be added?

Outputting the generator command to the file generated is circular problem of how does it first get created.
Also it is entirely possible that a file generated will not be able to determine the command which was used to generate it - it can be any command, so pretty much impossible to know for sure.

toaster:
I’d propose to use the file extensions mentioned [in this stackoverflow question](https://stackoverflow.com/questions/6432838/what-is-the-correct-file-extension-for-glsl-shaders).

These are also recognized by the GLSL GoLang-plugin I’m using :).

This would reduce the file names to `<name>.vert` and `<name>.frag`.

toaster:
While I understand the urge to DRY the shader code, I still would vote against the split of the “header” and the actual shader code.
I’d prefer complete shader files that can be compiled without further steps.

Bluebugs:
> While I understand the urge to DRY the shader code, I still would vote against the split of the “header” and the actual shader code. I’d prefer complete shader files that can be compiled without further steps.

Would you mind explaining further why you prefer so? I personally prefer to reduce the amount of code that need maintenance and is duplicated.

toaster:
> > While I understand the urge to DRY the shader code, I still would vote against the split of the “header” and the actual shader code. I’d prefer complete shader files that can be compiled without further steps.
> 
> Would you mind explaining further why you prefer so? I personally prefer to reduce the amount of code that need maintenance and is duplicated.

With full files, an IDE plug-in is able to check for errors.
With the header cut off, it is not.

Bluebugs:
> With full files, an IDE plug-in is able to check for errors. With the header cut off, it is not.

Hum, and in that case using Go template won't solve the problem as the IDE will not be able to figure out the template part either. This is a bit annoying. I guess two copies is better than 4. Until we have a better idea.

toaster (approved):

andydotxyz (approved):
Nice, much easier to work with

Issue #2534, opened 2021-10-07
[Windows] Calling external binaries causes flickering terminal


amo13:
### Describe the bug:

My application needs to periodically call an external binary (adb.exe android debug bridge) and evalutate its output. That is working fine so far. If I compile the application using `go build` then still, everything is working fine.
If I package the application using `fyne package -os windows -release` then unfortunately, for every call to the binary, a terminal window pops up and immediately disappears. (It makes no difference if I add `-executable myapp.exe` to the packaging command.)

### Device (please complete the following information):
 - **OS:** Windows
 - **Version:** 7
 - **Go version:** 1.17.1
 - **Fyne version:** 2.1.0

amo13:
Just for reference, in case anyone ends up here with the same problem:
[Here](https://github.com/amo13/Anarchy-Droid/issues/2#issuecomment-1008423373) is an elegant way to solve this by allocating a console and then hiding it using the windows system API.

PR #2821, opened 2022-03-05
Add a fyne build command.

Bluebugs:
### Description:
The build stage done by fyne has grown a lot and definitively simplify the life of the developer even if manually calling go build does still work. This just expose the build() so that it is also accessible, not just hidden behind the fyne package command.

This is also going to be useful for avoiding duplicated code by tools that could use fyne build to gerenate gopherjs and wasm output for vscode extension for example.

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne/internal/commands/build.go
+++ b/cmd/fyne/internal/commands/build.go
@@ -13,6 +13,7 @@ import (
 // Builder generate the executables.
 type Builder struct {
 	os, srcdir, target string
+	goPackage          string
 	release            bool
 	tags               []string
 	tagsToParse        string
@@ -64,10 +65,7 @@ func Build() *cli.Command {
 				if argCount != 1 {
 					return fmt.Errorf("incorrect amount of path provided")
 				}
-				if b.srcdir != "" {
-					return fmt.Errorf("the directory to package is already specified")
-				}
-				b.srcdir = ctx.Args().First()
+				b.goPackage = ctx.Args().First()
 			}
 
 			return b.Build()
@@ -179,6 +177,10 @@ func (b *Builder) build() error {
 			args = append(args, "-tags")
 		}
 		args = append(args, strings.Join(tags, ","))
+	}
+
+	if b.goPackage != "" {
+		args = append(args, b.goPackage)
 	}
 
 	if goos != "ios" && goos != "android" && !isWeb(goos) {
--- a/cmd/fyne/internal/commands/build.go
+++ b/cmd/fyne/internal/commands/build.go
@@ -28,7 +28,7 @@ func Build() *cli.Command {
 	return &cli.Command{
 		Name:        "build",
 		Usage:       "Build an application.",
-		Description: "You may specify the --executable to build, but --target will define the OS it is build for indiscriminate of the extention of the executable.",
+		Description: "You may specify the -o to build, but --target will define the OS it is build for indiscriminate of the extention of the output file.",
 		Flags: []cli.Flag{
 			&cli.StringFlag{
 				Name:        "target",
@@ -54,7 +54,7 @@ func Build() *cli.Command {
 			},
 			&cli.StringFlag{
 				Name:        "o",
-				Usage:       "The path to the executable to build, default is the current dir main binary",
+				Usage:       "Specify a name for the output file, default is based on the current directory.",
 				Destination: &b.target,
 			},
 		},
--- a/cmd/fyne/internal/commands/build.go
+++ b/cmd/fyne/internal/commands/build.go
@@ -7,14 +7,70 @@ import (
 	"strings"
 
 	version "github.com/mcuadros/go-version"
+	"github.com/urfave/cli/v2"
 )
 
-type builder struct {
+// Builder generate the executables.
+type Builder struct {
 	os, srcdir, target string
 	release            bool
 	tags               []string
+	tagsToParse        string
 
 	runner runner
+}
+
+// Build returns the cli command for building fyne applications
+func Build() *cli.Command {
+	b := &Builder{}
+
+	return &cli.Command{
+		Name:        "build",
+		Usage:       "Build an application.",
+		Description: "You may specify the --executable to build, but --target will define the OS it is build for indiscriminate of the extention of the executable.",
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:        "target",
+				Aliases:     []string{"os"},
+				Usage:       "The mobile platform to target (android, android/arm, android/arm64, android/amd64, android/386, ios, iossimulator).",
+				Destination: &b.os,
+			},
+			&cli.StringFlag{
+				Name:        "sourceDir",
+				Aliases:     []string{"src"},
+				Usage:       "The directory to package, if executable is not set.",
+				Destination: &b.srcdir,
+			},
+			&cli.StringFlag{
+				Name:        "tags",
+				Usage:       "A comma-separated list of build tags.",
+				Destination: &b.tagsToParse,
+			},
+			&cli.BoolFlag{
+				Name:        "release",
+				Usage:       "Enable installation in release mode (disable debug etc).",
+				Destination: &b.release,
+			},
+			&cli.StringFlag{
+				Name:        "executable",
+				Aliases:     []string{"exe"},
+				Usage:       "The path to the executable, default is the current dir main binary",
+				Destination: &b.target,
+			},
+		},
+		Action: func(_ *cli.Context) error {
+			return b.Build()
+		},
+	}
+}
+
+// Build parse the tags and start building
+func (b *Builder) Build() error {
+	if b.tagsToParse != "" {
+		b.tags = strings.Split(b.tagsToParse, ",")
+	}
+
+	return b.build()
 }
 
 func checkVersion(output string, versionConstraint *version.ConstraintGroup) error {
@@ -50,7 +106,7 @@ func checkGoVersion(runner runner, versionConstraint *version.ConstraintGroup) e
 	return checkVersion(string(goVersion), versionConstraint)
 }
 
-func (b *builder) build() error {
+func (b *Builder) build() error {
 	var versionConstraint *version.ConstraintGroup
 
 	goos := b.os
--- a/cmd/fyne/internal/commands/build.go
+++ b/cmd/fyne/internal/commands/build.go
@@ -28,7 +28,7 @@ func Build() *cli.Command {
 	return &cli.Command{
 		Name:        "build",
 		Usage:       "Build an application.",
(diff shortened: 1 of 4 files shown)


Jacalz:
Sorry but I'm not quite sure if I see the usefulness of this. I think you should either use the package command or a regular Go build command.

Bluebugs:
> Sorry but I'm not quite sure if I see the usefulness of this. I think you should either use the package command or a regular Go build command.

fyne build does remember for you all the parameters to pass to a go build to get a build done for a specific target. Sure, I can write a script or a command that copy the same behavior, but why not just expose it and make everyone life easier. The packaging step is mostly useful when preparing the deployment of the application, not much when doing testing and you want to be sure to have exactly the same binary as the one you get with fyne package.

Jacalz:
Absolutely. I just mean, is there really that much data that we are adding into the binaries? What I don’t see, it how much different it would be from a regular build command. Most of the packaged metadata is stored in external files as far as I know (but I might be wrong).

andydotxyz:
It feels almost like a circular discussion, in #2813 there is a push against having more fyne-build-added-benefits, and then here there is a question of whether fyne build is really adding much value.
Can't we add the improvements that simplify development and then see that this command is more valuable and could be approved?
If nothing else it's handy to know that the right "gui optimisation" flags are set - so we can stop having to answer #2534 etc

cmd/fyne/internal/commands/build.go
```diff
@@ -7,16 +7,72 @@ import (
...
+				Usage:       "A comma-separated list of build tags.",
+				Destination: &b.tagsToParse,
+			},
+			&cli.BoolFlag{
+				Name:        "release",
+				Usage:       "Enable installation in release mode (disable debug etc).",
+				Destination: &b.release,
+			},
+			&cli.StringFlag{
+				Name:        "executable",
```
andydotxyz, line 55: This seems not needed - as the exe is the result of build, so why would we pass in an alternative binary?

andydotxyz (changes requested):

Jacalz:
> It feels almost like a circular discussion, in #2813 there is a push against having more fyne-build-added-benefits, and then here there is a question of whether fyne build is really adding much value.

I'm truly sorry if I'm sounding negative or stubborn. All I'm trying to say, more or less, is that I'm sceptical about detracting from what is common ground when working with Go (such as `go build` binaries working out-of-the-box). I'm a bit worried that this sooner or later will lead to `go build` being phased out as more and more thing are moved into needing this build command.

However, you are very correct that the "gui optimisation" flags for Windows are a very good use case for this (one that I had not thought of). What I thought of, was the Linux and macOS binaires where almost all metadata is in separate files (the ones done by `fyne package`). If we can make this build command easy to use and similar to how a regular `go build` command can be used, I think this will be a good thing to add.

andydotxyz:
> I'm a bit worried that this sooner or later will lead to go build being phased out

A valid concern I guess, but the intro that @Bluebugs provided shows a different intent.
I think that continuing to support `go build` is a matter of project requirements than a single PR that makes alternative shorthands available.

>  the Linux and macOS binaires where almost all metadata is in separate files (the ones done by fyne package).

This becomes less true if we land #2813.

> If we can make this build command easy to use and similar to how a regular go build command can be used, I think this will be a good thing to add.

Is there anything in this PR that you feel is not easy to use, or should be updated to better match our aimed for simplicity of use?

Jacalz:
> Is there anything in this PR that you feel is not easy to use, or should be updated to better match our aimed for simplicity of use?

Not directly, no. I haven't had much time to test it out locally yet. I just mean that we should make it feel "similar" to the regular `go build` command, with a good user experience :)

Bluebugs:
Sorry, I can see the confusion if seen along #2813. My intent here is absolutely orthogonal and the goal is just to simplify developers life by providing a tool that will remember flags and environment variable for them when doing development/testing, so before packaging stage. This doesn't deprecate the use of go build or anything of the sort.

Also if third party tool want to get the binary without the packaging for new use that are out of scope for fyne itself, having that code accessible seems better than having them duplicate and maintain a parallel amount of code. I am currently thinking of providing a fyne-vscode command line that would do a different packaging and I don't think it should be part of the general purpose fyne command.

cmd/fyne/internal/commands/build.go
Bluebugs: Maybe it should match _go build_ command line and be _-o_ instead?

Jacalz: That sounds like a splendid idea.

cmd/fyne/internal/commands/build.go
```diff
@@ -7,16 +7,72 @@ import (
...
+				Aliases:     []string{"os"},
+				Usage:       "The mobile platform to target (android, android/arm, android/arm64, android/amd64, android/386, ios, iossimulator).",
+				Destination: &b.os,
+			},
+			&cli.StringFlag{
+				Name:        "sourceDir",
+				Aliases:     []string{"src"},
+				Usage:       "The directory to package, if executable is not set.",
+				Destination: &b.srcdir,
+			},
```
Jacalz, line 43: How about dropping this and instead using the command argument as input? This would make the command more like `go build ./cmd/fyne_demo` and we can avoid the need to remember the flag.

cmd/fyne/internal/commands/build.go
Jacalz: I wonder, is it better to keep consistency with other `fyne` commands, or stay close to the regular `go build`? I think the latter would make it easier to use for new users at the expense of us old chaps ;)

Jacalz (commented):
I'll have a closer look later in the week. Just a few thoughts in-line.

cmd/fyne/internal/commands/build.go
Jacalz: Good point. We can do so.

lucor:
Sorry for the late review. I'm +1 for having a dedicated build command.
First use case where it could be helpful is the support for the web target in fyne-cross (fyne-io/fyne-cross/issues/92)

cmd/fyne/internal/commands/build.go
andydotxyz: > How about dropping this and instead using the command argument as input? This would make the command more like `go build ./cmd/fyne_demo` and we can avoid the need to remember the flag.

Source dir and target package are not the same thing though... I think the go parameters take a package as parameter not directory.
Perhaps we need to move all our commands over time? "fyne get" is already there

cmd/fyne/internal/commands/build.go
andydotxyz: This looks confusing to me - the executable is an input parameter for all the other commands. file names are set according to the `name` parameter.
If we are to change this it should be for all commands, not just this one.

cmd/fyne/internal/commands/build.go
```diff
@@ -7,16 +7,90 @@ import (
...
+				Destination: &b.tagsToParse,
+			},
+			&cli.BoolFlag{
+				Name:        "release",
+				Usage:       "Enable installation in release mode (disable debug etc).",
+				Destination: &b.release,
+			},
+			&cli.StringFlag{
+				Name:        "executable",
+				Aliases:     []string{"exe", "o"},
```
andydotxyz, line 57: This just looks wrong - `exe` is input and `o` is output - how can they be aliases?

andydotxyz (changes requested):

cmd/fyne/internal/commands/build.go
Bluebugs: I have updated the documentation on the command to be more precise, but I think it depends on context. In this context, why would you expect that the build command take an executable as input when its goal is to build the executable? I don't see it as a conflict with other command, but actually keep consistency as it means you specify the executable in the same way.

I am fine dropping exe and executable and just use o, just found weird to not mirror other command.

andydotxyz: You’re right it does not make sense. Which is why I suggested removing it in the first place. If we want to add “-o” that would apply to all the commands not just this one. It is more important to be consistent with this tool than with the go build command when adding functionality. If we want to be more like go compiler then we need to apply that to all commands in this tool.

cmd/fyne/internal/commands/build.go
```diff
@@ -53,8 +53,7 @@ func Build() *cli.Command {
 				Destination: &b.release,
 			},
 			&cli.StringFlag{
-				Name:        "executable",
-				Aliases:     []string{"exe", "o"},
+				Name:        "o",
```
andydotxyz, line 56: I think these docs were copied from the old "executable" input parameter flag, so it's a little strange.
The -o which should go to all commands later, needs docs that will make sense. Perhaps:
"Specify a name for the output file, default is based on the current directory".

andydotxyz (changes requested):

cmd/fyne/internal/commands/build.go
```diff
@@ -7,16 +7,89 @@ import (
...
 }
 
+// Build returns the cli command for building fyne applications
+func Build() *cli.Command {
+	b := &Builder{}
+
+	return &cli.Command{
+		Name:        "build",
+		Usage:       "Build an application.",
+		Description: "You may specify the --executable to build, but --target will define the OS it is build for indiscriminate of the extention of the executable.",
```
Jacalz, line 31: This documentation should likely be tidied up, in a similar fashion to https://github.com/fyne-io/fyne/pull/2821/files#r825293352.

Jacalz (changes requested):

cmd/fyne/internal/commands/build.go
```diff
@@ -28,7 +28,7 @@ func Build() *cli.Command {
 	return &cli.Command{
 		Name:        "build",
 		Usage:       "Build an application.",
-		Description: "You may specify the --executable to build, but --target will define the OS it is build for indiscriminate of the extention of the executable.",
+		Description: "You may specify the -o to build, but --target will define the OS it is build for indiscriminate of the extention of the output file.",
```
andydotxyz, line 31: This sentence does not make sense. It looks like it was copied from package and manipulated to change the variable names. But it's content is still based on specifying an input executable.
How about something like the following?

`You can specify --target to define the OS to build for. The executable file will default to an appropriate name but can be overridden using -o`

andydotxyz (changes requested):
The build doc sentence doesn't scan. Provided suggested improvement.

andydotxyz (approved):
Thanks

lucor (approved):

Jacalz (approved):
Works perfectly.

Issue #2800, opened 2022-02-21
Window improperly sized if contains image with ImageFillOriginal and setFixedSize is set


kheyse-oqton:
### Describe the bug:

Regression where window does not take  image size into account if setfixedsize is true.

The regression is blocking us from upgrading to v2.1. Same code worked on v2.0

### To Reproduce:
Run the app

### Screenshots:

v2.0
<img width="459" alt="Screenshot 2022-02-18 at 11 12 36" src="https://user-images.githubusercontent.com/35029144/154996493-74796208-0e1f-486a-9208-f84d4090f1d5.png">
v2.1
<img width="236" alt="Screenshot 2022-02-18 at 11 12 08" src="https://user-images.githubusercontent.com/35029144/154996483-d4e38eab-64ce-4f8d-a524-e232bfbb914e.png">

### Example code:

```go
	header := canvas.NewImageFromResource(resourceHeaderSvg)
	header.FillMode = canvas.ImageFillOriginal

	w.SetContent(container.NewVBox(
		header,
	))

	w.SetFixedSize(true)

	w.ShowAndRun()
```

### Device (please complete the following information):
 - **OS:** Mac os 
 - **Version:** 12.2.1 on an M1 
 - **Go version:** 1.17.6
 - **Fyne version:** 2.1.2

andydotxyz:
This relates to the MinSize of `FillOriginal` is not known until the image is loaded after initial render request.

kheyse-oqton:
> This relates to the MinSize of `FillOriginal` is not known until the image is loaded after initial render request.

is there a workaround? because this blocks us from upgrading from 2.0

andydotxyz:
In general it is a bad idea to prop windows open using ImageFillOriginal - the results will be very different depending on device (i.e. with a 3* pixel density on iPhone or Retina/HiDPI on PC the window will be 1/3 or 1/2 of the expected size.
You should, wherever possible, use Fyne coordinates to describe size. Therefore in this instance `Window.Resize` would be a more platform-safe approach.

andydotxyz:
> I tried adding my image without ImageFillOriginal
> but I always get a squashed image that is nearly invisible

Yes, as an image which is pixel based has no innate size in a vector world the default is basically 1x1.
The right thing to do is to figure out how much space you want to allocate to your image and then fill it (the default mode).
If you want to fill remaining space in your window then a border layout may be all you need, with the image filling the central space.

andydotxyz:
[...]

PR #2829, opened 2022-03-08
Move the fixed size ensuring size to the end of window setup

### LINKED ISSUE #2819: window.Resize() does not work if SetFixedSize(true) is set after
### Describe the bug:
In Fyne 2.1.3 the order of  window.Resize() and SetFixedSize(true) matters. This is not the case in Fyne 2.1.2

### To Reproduce: 

This code only works on Fyne 2.1.2. On Fyne 2.1.3, it does not resize the window w which appears minimized :
```go
w.Resize(fyne.NewSize(finalWidth, finalHeight)) 
w.SetFixedSize(true)
```

Whereas this code works perfectly 
```go
w.SetFixedSize(true)
w.Resize(fyne.NewSize(finalWidth, finalHeight)) 
```

### Device :
 - **OS:**  Linux Manjaro 
 - **Version:**  up do date 2022-03-04
 - **Go version:**  go1.17.6 linux/amd64
 - **Fyne version:** 2.1.3

andydotxyz:
Fixes order of operation issues such as #2819

Fixes #2819

### Checklist:

- [ ] Tests included. <- subtlety in render size for some Linux WM
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -218,7 +218,6 @@ func (w *window) FixedSize() bool {
 
 func (w *window) SetFixedSize(fixed bool) {
 	w.fixedSize = fixed
-	w.ensureFixedHasSize()
 	if w.view() != nil {
 		w.runOnMainWhenCreated(w.fitContent)
 	}
@@ -501,9 +500,6 @@ func (w *window) SetContent(content fyne.CanvasObject) {
 	}
 
 	w.canvas.SetContent(content)
-	if !visible {
-		w.ensureFixedHasSize()
-	}
 
 	// show new canvas element
 	if content != nil {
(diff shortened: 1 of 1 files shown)


andydotxyz:
Tagged @Bluebugs on this as I think you worked on this recently

Bluebugs:
This sadly doesn't fix @kheyse-oqton problem in #2800. It is still a progress as I get a 0x0 window (and crash) with master and the provided example. I think it is an issue where the proper size is only really known after the windows has been built.

andydotxyz:
I removed #2800 from the fix list earlier today, this remains just a fix for #2819 as the description states above.
Can it be reviewed on the merits of that regression fix please?

Bluebugs (approved):
I am a bit uneasy that it doesn't fix #2800 as this patch change the behavior in that issue, but I can see that it does improve the situation. So let's move forward with this.

PR #2845, opened 2022-03-14
 fix some typos

cuishuang:


data/binding/convert_test.go
```diff
@@ -44,7 +44,7 @@ func TestBoolToStringWithFormat(t *testing.T) {
 	assert.Nil(t, err)
 	v, err = s.Get()
 	assert.Nil(t, err)
-	assert.Equal(t, "truely", v)
```
andydotxyz, line 47: Sadly this line (and the one below) is not going to work, as the test is for "true"+"ly".
Feel free to make a slightly larger change that does not mark it as a typo, but correcting the spelling is what has caused the build failures.

andydotxyz (changes requested):
Thanks for looking at this. Unfortunately one of the changes breaks tests, as noted inline.

data/binding/convert_test.go
cuishuang: Thanks.

andydotxyz (approved):
Thanks.

If you start working on larger changes please work on `develop` or the latest `release` branch as we don't normally work on `master`.

PR #2846, opened 2022-03-14
Update FixedSize window size if min is bigger

### LINKED ISSUE #2800: Window improperly sized if contains image with ImageFillOriginal and setFixedSize is set

andydotxyz:
Fixes #2800

### Checklist:

- [ ] Tests included. <- It's another one of those "measure visible window" issues we haven't the test infrastructure for yet.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -314,7 +314,12 @@ func (w *window) fitContent() {
 		w.viewLock.Unlock()
 	}
 	if w.fixedSize {
-		w.shouldWidth, w.shouldHeight = w.requestedWidth, w.requestedHeight
+		if w.shouldWidth > w.requestedWidth {
+			w.requestedWidth = w.shouldWidth
+		}
+		if w.shouldHeight > w.requestedHeight {
+			w.requestedHeight = w.shouldHeight
+		}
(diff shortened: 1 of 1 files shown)


andydotxyz:
Regarding tests I think we need to start having some visible tests that are not run all times (behind "full" tag or some such).
If we did that we could remove the visible tests from default run but add more like that in this full test mode?

Jacalz (approved):
Strange. I must have been in the wrong folder or something. Works now.
image

Bluebugs (approved):

PR #2848, opened 2022-03-14
Add RemoveAll to Container

stuartmscott:
### Description:
Adds an intuitive API for removing all objects from a container.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/container.go
+++ b/container.go
@@ -158,6 +158,12 @@ func (c *Container) Remove(rem CanvasObject) {
 	}
 }
 
+// RemoveAll updates the contents of this container to no longer include any objects.
+// Since: 2.2
+func (c *Container) RemoveAll() {
+	c.Objects = nil
+}
+
 // Visible returns true if the container is currently visible, false otherwise.
 func (c *Container) Visible() bool {
 	return !c.Hidden
--- a/container_test.go
+++ b/container_test.go
@@ -102,6 +102,16 @@ func TestContainer_Remove(t *testing.T) {
 	assert.Equal(t, float32(0), box2.Position().Y)
 }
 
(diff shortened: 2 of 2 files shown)


Jacalz (changes requested):
Great stuff. Just one minor note inline.

Jacalz (dismissed):
Thanks. Approved. However, shall we wait until the v2.1.4 changes have landed on develop? Then you could make sure to get the documentation updated to recommend this new method instead.

container.go
```diff
@@ -158,6 +158,13 @@ func (c *Container) Remove(rem CanvasObject) {
 	}
 }
 
+// RemoveAll updates the contents of this container to no longer include any objects.
+//
+// Since: 2.2
+func (c *Container) RemoveAll() {
+	c.Objects = nil
```
andydotxyz, line 165: I think it should Refresh() as well, method calls normally make sure we’re updating after the change.

andydotxyz (changes requested):
Great change, just missing the refresh call I think.
Jacob is right, we could hold back a couple of days until 2.1.4 comes in - but either way works

andydotxyz (dismissed):
As discussed it seems container functions don't refresh.
So matching that is right, and we can review them all together if we need to.
(plus 2.1.4 changes now on develop so you should be clear to update the new docs too)

Jacalz:
Should we get the the 2.1.4 changes pulled in, the docs updated, and then we can merge it? Should be ready to get that going now :)

andydotxyz (approved):

andydotxyz:
I completed the doc and resolved merge conflict so we could get this in

PR #2853, opened 2022-03-18
Provide a smoother menu refresh experience

### LINKED ISSUE #2852: Add a more elegant way of refreshing menus
This is a clarification of #2851. Please see that ticket for background.

The only way to refresh menus is by re-calling `window.SetMainMenu()`. Idiomatic Fyne would have us call `Refresh()`, but these are dumb structs.  Having to carry around objects which often have no other use is code smell, and re-adding the menu to the window as a way of refreshing changes is heavy-handed.

I suggest adding a `window.Refresh()` that refreshes decorations like menus or other possible future structures (decorations?) attached to it. This would still leak the window object when it might not otherwise be needed, but it's one less unnecessary object to carry around.

andydotxyz:
A work in progress, looking for discussion as it develops.
I think the following is required:

- [x] Refresh a Window MainMenu
- [x] Refresh a SystemTray Menu
- [x] Refresh any Menu
- [ ] (perhaps Refresh any MenuItem?)

Fixes #2852 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


menu.go
```diff
@@ -12,6 +18,29 @@ func NewMenu(label string, items ...*MenuItem) *Menu {
...
+		}
+	}
+
+	if d, ok := CurrentApp().Driver().(systemTrayDriver); ok {
+		if m == d.SystemTrayMenu() {
+			d.SetSystemTrayMenu(m)
+		}
+	}
+}
+
```
toaster, line 43: I’d expect the `Set*` methods to do nothing if the item to set is already set, i.e. I find the refresh effect unexpected. There should be a `RefreshMenu` method instead.

Thinking about it, IMO, it would be better if the menu would not know about the potential places where it might be used.
Instead, the sys-tray, the main menu widget and the menu widget should register a refresh callback if their fyne.(Main)Menu is being set. They should also deregister this if the menu is changed, of course.
Thus, the changed (main) menu would simply call all registered callbacks.

andydotxyz: I agree that would be a cleaner implementation - but I cannot figure out how to do that without additional functionality in our top level APIs.
Because `fyne.Menu` is a raw struct I am not sure how to support the chance callback hooks without adding complexity to the API - and also exposing the possibility that users could disconnect the callback...

If we do find a way to do it more neatly I would be happy to do so, but also I would hope it could sit behind the same clean public API that mimics how CanvasObject works.

Any thoughts on a way to proceed?

toaster: Okay, I thought a little bit about it and didn’t came up with an idea better than an internal registry to decouple the callback handling from the menu. But this is clearly to much complexity. The current approach is not really a blocker. It can be easily refactored if we have a better idea in the future.

However, the usage of `Set*` for triggering a refresh should be improved like I said in the first paragraph :).

andydotxyz: Thanks. I'm not totally clear about the `Set` comment - what would your suggested `RefreshMenu` do different to `Menu.Refresh()`? I'll try and catch you on Slack so we can disccus

toaster (approved):

PR #2855, opened 2022-03-19
RFC: Various changes to CI linting

Jacalz:
### Description:

I want to open this for various suggested changes regarding the linting we do as part of the CI checks. My suggestion is to replace `goimports` with [gofumpt](https://github.com/mvdan/gofumpt) and to stop using golint. As for the former, I think the stricter formatting is doing a great job (see diff for more info) and for the latter, it [has been deprecated](https://github.com/golang/lint/blob/master/README.md) for a while now. Most of what golint did (if not even everything) is handled by `go vet` and `staticcheck` already.

These changes should also make the static analysis quite a bit faster.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
This is basically working. It's just `gocyclo` that's complaining about the vendor folder. I have opened an issue with them about it.
I really want to be able to not have to remove the vendor folder for each run.

.github/workflows/static_analysis.yml
```diff
@@ -19,26 +19,20 @@ jobs:
...
+      run: test -z $(gofumpt -d -e . | tee /dev/stderr)
 
     - name: Gocyclo
       run: gocyclo -over 30 .
 
-    - name: Golint
-      run: golint -set_exit_status $(go list -tags ci ./...)
-
     - name: Staticcheck
-      run: staticcheck -go 1.14 ./...
```
Jacalz, line 44: FYI, this was removed as it picks it up from the module by default.

andydotxyz (commented):
In general this looks like a good change (except the noted CI failure).

However as it's changing the development style I wonder if it should be discussed with the group? Documentation and example git hooks would all need to be updated to avoid lots of PRs failing on open...

Jacalz:
> However as it's changing the development style I wonder if it should be discussed with the group? Documentation and example git hooks would all need to be updated to avoid lots of PRs failing on open...

Indeed. That is why I both opened this as RFC and added it to the agenda for the Friday meeting next week. I’m very well aware that this might be disruptive to peoples workflows. I want this to get all the attention that it needs :)

container/split.go
```diff
@@ -185,9 +185,12 @@ func (r *splitContainerRenderer) computeSplitLengths(total, lMin, tMin float32)
...
 var _ fyne.CanvasObject = (*divider)(nil)
-var _ fyne.Draggable = (*divider)(nil)
-var _ desktop.Cursorable = (*divider)(nil)
-var _ desktop.Hoverable = (*divider)(nil)
+
+var (
+	_ fyne.Draggable     = (*divider)(nil)
+	_ desktop.Cursorable = (*divider)(nil)
+	_ desktop.Hoverable  = (*divider)(nil)
+)
```
Jacalz, line 193: This is strange. It seems to me like these should have been in one block.

andydotxyz:
> Indeed. To try an alleviate the problems as much as possible, I have both opened this as RFC and added it to the agenda for the Friday meeting next week. I’m very well aware that this might be disruptive to peoples workflows. I want this to get all the attention that it needs :)

Ah I did not notice that good call!

container/split.go
andydotxyz: agreed. bug?

Jacalz: Yup. I’ve opened an issue upstream about it https://github.com/mvdan/gofumpt/issues/212.

Jacalz:
Looks like tests are passing now :)

container/split.go
Jacalz: I’ve done this change manually for now. It doesn’t complain without the fix to the automatic formatting so we are probably fine for now.

Jacalz:
Doesn't make sense at the moment. Will just drop golint in a different PR.

PR #2859, opened 2022-03-21
Optimization: minimal gradient texture

toaster:
### Description:

This PR reduces the size of textures used for linear gradients to be 1px high or wide.
Thus, it only contains pixels with different colour, i.e. the size is minimal.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/painter/gl/gl_common.go
+++ b/internal/painter/gl/gl_common.go
@@ -120,8 +120,16 @@ func (p *glPainter) newGlRasterTexture(obj fyne.CanvasObject) Texture {
 func (p *glPainter) newGlLinearGradientTexture(obj fyne.CanvasObject) Texture {
 	gradient := obj.(*canvas.LinearGradient)
 
-	width := p.textureScale(gradient.Size().Width)
-	height := p.textureScale(gradient.Size().Height)
+	w := gradient.Size().Width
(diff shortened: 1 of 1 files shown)


internal/painter/gl/gl_common.go
```diff
@@ -120,8 +120,16 @@ func (p *glPainter) newGlRasterTexture(obj fyne.CanvasObject) Texture {
 func (p *glPainter) newGlLinearGradientTexture(obj fyne.CanvasObject) Texture {
 	gradient := obj.(*canvas.LinearGradient)
 
-	width := p.textureScale(gradient.Size().Width)
-	height := p.textureScale(gradient.Size().Height)
+	w := gradient.Size().Width
+	h := gradient.Size().Height
+	switch gradient.Angle {
+	case 90, 70:
```
Jacalz, line 126: Why 70 degrees? It might just be that I’m missing but I want to know 🙂

toaster: Ha! very good catch! Obviously, my stubby fingers removed the `2`.
No test finds this because no one uses this particular angle.

andydotxyz (approved):
👍

PR #2869, opened 2022-03-26
Ci cleanup

Jacalz:
### Description:

Golint has been deprecated and frozen. Go vet and staticcheck should have us covered.
After some fixes upstream in gocyclo, we can also stop removing the vendor folder now.
The workflow has also been renamed to not be called "checks".

Added in the stricter security permissions and renamed the license workflow to not be called "checks".

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/.github/workflows/license_check.yml
+++ b/.github/workflows/license_check.yml
@@ -1,16 +1,20 @@
 name: Check license of dependencies
 on: [push, pull_request]
+permissions:
+  contents: read
 
 jobs:
-  checks:
+  licenses:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+      with:
+        persist-credentials: false
     - uses: WillAbides/setup-go-faster@v1.7.0
       with:      
         go-version: '1.18.x'
 
-    - name: install lian
+    - name: Install lian
       run: go install lucor.dev/lian@latest
 
     - name: Check license of dependencies against go.mod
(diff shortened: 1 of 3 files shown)


andydotxyz:
> After some fixes upstream in gocyclo, we can also stop removing the vendor folder now.

Not sure about this... As is demonstrated in the failure currently not all libraries that we vendor will share our sense of how strictly things should be enforced... any changes will be reversed next time someone "mod vendors" so I think it must still be ignored.

In the future we can strip out vendor anyhow (once go modules are standard for all versions we support).

Jacalz:
> any changes will be reversed next time someone "mod vendors" so I think it must still be ignored.

Sorry, I'm not sure I follow here. I haven't modified anything inside the vendor folder. The linting tools now all ignore the vendor folder (gocyclo was the only one that wasn't doing it before). Keeping the vendor folder in our case now is just the same as when we deleted it before. None of our tools are checking it either way.

andydotxyz:
> Sorry, I'm not sure I follow here. I haven't modified anything inside the vendor folder.

I did not mean to suggest you had modified the content, but that the "static_analysis" test failure was reporting incorrectly formatted code. This is where the problem is because we cannot modify the vendor folder to fit. This used to be ignored so I assumed the issue related to your description including "we can also stop removing the vendor folder now."...

Jacalz:
> 

Ah, I see. I had missed that fact that goimports failed. I see what you mean now. Because I didn't use it in the last PR, I missed that it looked at the vendor folder.

Jacalz:
This should be looking better now :)

andydotxyz (approved):

PR #2870, opened 2022-03-26
Form validators

### LINKED ISSUE #2562: Form validator

Cookie04DE:
### Description:
Adds the ability to add a form wide validator which get's run every time a widget inside the form get's changed. It has the following signature: `func(*widget.Form) error`, and if the returned error is not nil the error message get's displayed in red next to the submit button and the form get's disabled.

I am still working on the tests, since even the tests on the unchanged `develop` branch on the latest commit don't pass on macOS Monterey 12.3 arm64 go version go1.18 darwin/arm64: [output](https://pastebin.com/EV44qKbN).

Fixes #2562

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/form.go
+++ b/widget/form.go
@@ -147,7 +147,11 @@ func (f *Form) SetValidationError(err error) {
 	if (err == nil && f.validationError != nil) || (f.validationError == nil && err != nil) ||
 		err.Error() != f.validationError.Error() {
 		f.validationError = err
-		f.validationErrorText.Text = err.Error()
+		var errorText string
+		if err != nil {
+			errorText = err.Error()
+		}
+		f.validationErrorText.Text = errorText
 		if f.onValidationChanged != nil {
 			f.onValidationChanged(err)
 		}
--- a/widget/form.go
+++ b/widget/form.go
@@ -135,8 +135,10 @@ func (f *Form) SetOnValidationChanged(callback func(error)) {
 // SetValidationError manually updates the validation status until the next change in a child widget
 func (f *Form) SetValidationError(err error) {
 	if f.Validator == nil {
-		f.checkValidation(nil) // make sure the form get's enabled again if no widget inside it is invalid
 		return
+	}
+	if err == nil {
+		f.checkValidation(nil) // make sure the form get's enabled again if no widget inside it is invalid
 	}
 	if err == nil && f.validationError == nil {
 		return
--- a/widget/form.go
+++ b/widget/form.go
@@ -379,6 +379,7 @@ func (f *Form) CreateRenderer() fyne.WidgetRenderer {
 	f.updateButtons()
 	f.updateLabels()
 	f.checkValidation(nil) // will trigger a validation check for correct intial validation status
+	f.Validate()           // will trigger a validation check with f.Validator
 	return renderer
 }
 
--- a/widget/form.go
+++ b/widget/form.go
@@ -39,6 +39,7 @@ func NewFormItem(text string, widget fyne.CanvasObject) *FormItem {
 // If you change OnSubmit/OnCancel after the form is created and rendered, you need to call
 // Refresh() to update the form with the correct buttons.
 // Setting OnSubmit/OnCancel to nil will remove the buttons.
+// TODO
 type Form struct {
 	BaseWidget
 
@@ -47,13 +48,17 @@ type Form struct {
 	OnCancel   func()
 	SubmitText string
 	CancelText string
+	Validator  func(*Form) error
 
-	itemGrid     *fyne.Container
-	buttonBox    *fyne.Container
-	cancelButton *Button
-	submitButton *Button
+	itemGrid            *fyne.Container
+	buttonBox           *fyne.Container
+	cancelButton        *Button
+	submitButton        *Button
+	validationErrorText *canvas.Text
+	onValidationChanged func(error)
 
-	disabled bool
+	disabled        bool
+	validationError error
 }
 
 // Append adds a new row to the form, using the text as a label next to the specified Widget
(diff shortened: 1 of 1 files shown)


widget/form.go
```diff
@@ -241,6 +295,39 @@ func (f *Form) setUpValidation(widget fyne.CanvasObject, i int) {
...
+	if se, ok := widget.(*SelectEntry); ok {
+		se.OnChanged = func(string) {
+			f.Validate()
+		}
+	}
+	if s, ok := widget.(*Slider); ok {
+		s.OnChanged = func(float64) {
+			f.Validate()
+		}
+	}
```
Jacalz, line 330: Sorry, but I don't think this is the right way to tackle the problem. Not only are you overwriting `OnChanged` which may have been set by the user, but you are also hooking translations into (and hard-coding for) widgets that don't currently support validation.

Jacalz (changes requested):
Thanks a lot for opening this PR. I just had some thoughts on the implementation.
Would it not be better to hook into any validatable child widgets (using `SetOnValidationChanged()`) and update the validation of the form when they change? 
If widgets haven't changed their validation, there shouldn't be any reason for the validation
of the whole form to change, or so I think. Shouldn't just the `Submit` button be disabled, and not the whole form?

Cookie04DE:
While just hooking into validation state changes works for some use cases it doesn't for all.
Consider the following example:
A checkbox enabling advanced mode and an entry for you age.
If you enable advanced mode you have to enter an age of at least 18, else the age has to be below 18.
For the checkbox both states (checked and not checked) are valid and for the entry all whole numbers are valid.
If we only hook into the validation change the following could happen:
The user checks advanced mode and enters an age of 19. The form's `Validator` get's fired and returns `nil`, the form is submittable.
The user changes the value to 10 which is still valid for the entry and therefor the validation change doesn't fire. The form is still submittable despite being invalid according to the `Validator`.
As for the disabling of the entire form: I agree disabling the submit button only makes more sense. I will change it once I get around to it.

widget/form.go
Cookie04DE: Perhaps we could resolve the problem of overwriting `OnChanged`, by calling the original func that was there after we called `f.Validate()`. The user could still overwrite our function though, not sure how much of a problem that would be.
As I laid out in my other answer hooking into the validation change event isn't enough, and as far as I can tell there is no uniform way to hook into the `OnChanged` event of a widget since they all have a different signature for their `OnChanged` functions. Entries use `func(string)` for example while check boxes use `func(bool)`.
Hardcoding widgets and their `OnChanged` functions isn't elegant but I don't know how to avoid that.
I also wrote why checkboxes need to be involved at all (even if they don't support validation) in my second answer.

Jacalz: Hard-coding widgets is going to be viable in the long run. Any develop has the ability to create custom widgets and we can't keep track of those as well. This might require some more though to get right. I understand that it is a tricky problem to solve.

Jacalz:
As for the checkboxes, I do have some work on that matter in https://github.com/fyne-io/fyne/pull/2765. It adds support for selection validation.

> If you enable advanced mode you have to enter an age of at least 18, else the age has to be below 18.
For the checkbox both states (checked and not checked) are valid and for the entry all whole numbers are valid.

I am still in the belief that something like that would be best solved by https://github.com/fyne-io/fyne/pull/2567 or similar. If two validated widgets and their valid status depends on each other, I'd say that it is best to link them together in that way. With selection validation added in, perhaps it might make sense to have an `AllSelection()` and also some other function that allows chaining together different types of validators (if that's even possible)?

widget/form.go
Cookie04DE: Instead of updating on every widget change, we could leave that up to the developer. They would need to set the `OnChanged` function of very widget the `Validator` would check to include a call to `f.Validate()`. This would also be beneficial to the performance since the `Validator` would only be run if a widget, the validator actually cares about, changes.

Cookie04DE:
I was under the impression that the `Validator` field of an `Entry` for example should only make sure that the entries text is valid and not take any other data source into consideration.
So in a sense the `Form`'s `Validator` is responsible for the "bigger picture", checking if the data as a whole makes sense.
If two data sources don't fit together (in my example you entered an age of 10 and checked the advanced mode checkbox) where should this error be displayed? It could be displayed next to the entry but what if there are two entries? Should the error be displayed next to both or just one?
Both entries might contain data is valid on it's own but not in the "bigger picture", if you get what I mean.
Since the error is caused by multiple sources I think this sort of error should be displayed in a more "neutral" place, such as next to the submission button.

widget/form.go
andydotxyz: > Instead of updating on every widget change, we could leave that up to the developer. 

It sounds quite unintuitive which isn't really our approach.
Setting on a validator for Form should just work, and update at the right time.
If we can detect `Validatable` instead it would be much cleaner and extensible.

widget/form.go
```diff
@@ -47,13 +50,17 @@ type Form struct {
 	OnCancel   func()
 	SubmitText string
 	CancelText string
+	Validator  func(*Form) error
```
andydotxyz, line 53: Is there a way we can avoid this being form specific?
Something like the following would be really interesting:

`Form.Validator = validation.NewEqualText(email1, email2)`

A crass example but as you can see the `Form` is not passed in because it isn't really what is being validated...

PR #2871, opened 2022-03-27
Form Validation Status

chabad360:
### Description:

Allows interacting with the validation status of a form

### Checklist:

- [x] Tests included.
- [x] Tests all pass (only tested widgets, as other tests were already failing on this branch).

#### Where applicable:

- [ ] Public APIs have version numbers.

### DIFF
--- a/widget/form.go
+++ b/widget/form.go
@@ -273,8 +273,7 @@ func (f *Form) setValidationError(err error) {
 		return
 	}
 
-	if (err == nil && f.validationError != nil) || (f.validationError == nil && err != nil) ||
-		!errors.Is(err, f.validationError) {
+	if !errors.Is(err, f.validationError) {
 		if err == nil {
 			for _, item := range f.Items {
 				if item.invalid {
--- a/widget/form.go
+++ b/widget/form.go
@@ -274,7 +274,7 @@ func (f *Form) setValidationError(err error) {
 	}
 
 	if (err == nil && f.validationError != nil) || (f.validationError == nil && err != nil) ||
-		err.Error() != f.validationError.Error() {
+		!errors.Is(err, f.validationError) {
 		if err == nil {
 			for _, item := range f.Items {
 				if item.invalid {
--- a/widget/form.go
+++ b/widget/form.go
@@ -124,6 +124,37 @@ func (f *Form) Disabled() bool {
 	return f.disabled
 }
 
+// SetOnValidationChanged is intended for parent widgets or containers to hook into the validation.
+func (f *Form) SetOnValidationChanged(callback func(error)) {
+	if callback != nil {
+		f.onValidationChanged = callback
+	}
+}
+
+// SetValidationError manually updates the validation status until the next input change.
+func (f *Form) SetValidationError(err error) {
+	if err == nil && f.validationError == nil {
+		return
+	}
+
+	if (err == nil && f.validationError != nil) || (f.validationError == nil && err != nil) ||
+		err.Error() != f.validationError.Error() {
+		if err == nil {
+			for _, item := range f.Items {
+				if item.invalid {
+					err = item.validationError
+					break
+				}
+			}
+		}
+		f.validationError = err
+
+		if f.onValidationChanged != nil {
+			f.onValidationChanged(err)
+		}
+	}
+}
+
 // Validate validates the entire form.
 func (f *Form) Validate() error {
 	for _, item := range f.Items {
@@ -134,13 +165,6 @@ func (f *Form) Validate() error {
 		}
 	}
 	return nil
-}
-
-// SetOnValidationChanged is intended for parent widgets or containers to hook into the validation.
-func (f *Form) SetOnValidationChanged(callback func(error)) {
-	if callback != nil {
-		f.onValidationChanged = callback
-	}
 }
 
 func (f *Form) createInput(item *FormItem) fyne.CanvasObject {
@@ -242,29 +266,6 @@ func (f *Form) ensureRenderItems() {
 	f.itemGrid.Objects = append(f.itemGrid.Objects, objects...)
 }
 
-func (f *Form) setValidationError(err error) {
-	if err == nil && f.validationError == nil {
-		return
-	}
-
-	if (err == nil && f.validationError != nil) || (f.validationError == nil && err != nil) ||
-		err.Error() != f.validationError.Error() {
-		if err == nil {
-			for _, item := range f.Items {
-				if item.invalid {
-					err = item.validationError
-					break
-				}
-			}
-		}
-		f.validationError = err
-
-		if f.onValidationChanged != nil {
-			f.onValidationChanged(err)
-		}
-	}
-}
-
 func (f *Form) setUpValidation(widget fyne.CanvasObject, i int) {
 	updateValidation := func(err error) {
 		if err == errFormItemInitialState {
@@ -272,7 +273,7 @@ func (f *Form) setUpValidation(widget fyne.CanvasObject, i int) {
 		}
 		f.Items[i].validationError = err
 		f.Items[i].invalid = err != nil
-		f.setValidationError(err)
+		f.SetValidationError(err)
 		f.checkValidation(err)
 		f.updateHelperText(f.Items[i])
 	}
--- a/widget/form.go
+++ b/widget/form.go
@@ -124,6 +124,7 @@ func (f *Form) Disabled() bool {
 	return f.disabled
 }
 
+// Validate returns the validation status of the form.
 func (f *Form) Validate() error {
 	for i, item := range f.Items {
 		if w, ok := item.Widget.(fyne.Validatable); ok {
@@ -133,18 +134,25 @@ func (f *Form) Validate() error {
 				}
 				f.Items[i].validationError = err
 				f.Items[i].invalid = err != nil
-				f.SetValidationError(err)
+				f.setValidationError(err)
 				f.checkValidation(err)
 				f.updateHelperText(f.Items[i])
 				return err
 			}
 		}
 	}
-	f.SetValidationError(nil)
+	f.setValidationError(nil)
 	return nil
 }
 
-func (f *Form) SetValidationError(err error) {
+// SetOnValidationChanged is intended for parent widgets or containers to hook into the validation.
+func (f *Form) SetOnValidationChanged(callback func(error)) {
+	if callback != nil {
+		f.onValidationChanged = callback
+	}
+}
+
+func (f *Form) setValidationError(err error) {
 	if err == nil && f.validationError == nil {
 		return
 	}
@@ -158,12 +166,6 @@ func (f *Form) SetValidationError(err error) {
 		}
 
 		f.Refresh()
-	}
-}
-
-func (f *Form) SetOnValidationChanged(callback func(error)) {
-	if callback != nil {
-		f.onValidationChanged = callback
 	}
 }
 
@@ -273,7 +275,7 @@ func (f *Form) setUpValidation(widget fyne.CanvasObject, i int) {
 		}
 		f.Items[i].validationError = err
 		f.Items[i].invalid = err != nil
-		f.SetValidationError(err)
+		f.setValidationError(err)
 		f.checkValidation(err)
 		f.updateHelperText(f.Items[i])
 	}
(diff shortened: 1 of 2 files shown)


Jacalz:
It seems to me like this overlaps with https://github.com/fyne-io/fyne/pull/2870 quite a bit, but perhaps with different end results.

chabad360:
It does, sort of. This effectively just passes whether the submit button is disabled as a validation, while #2870 attempts to give the form its own validation abilities.
So there is definitely some overlap, but I don't think this will have a significant effect (if any) on the implementation of #2870.

Jacalz:
> So there is definitely some overlap, but I don't think this will have a significant effect (if any) on the implementation of #2870.

Sorry, but I unfortunately don't think that's true. As they are implementing the same methods, it would cause some massive merge conflicts if one is merged before the other.

chabad360:
Ah, yes, I suppose I was speaking a bit more theoretical. What I meant was that these changes should be almost entirely compatible (in concept) with those of #2870. I agree it would cause a merge conflict, but I don't think it would be very hard to resolve, as the conflicting functions are almost identical (other than a few lines). I'll doublecheck later today just to be sure.

chabad360:
Mmm, now that I've checked, the main complication would be in `SetValidationError()`, and even then it wouldn't be that bad, as #2870 would still need to implement at least part of my logic or there would be some issues (particularly around enabling the enable button despite having invalid entries). Point is, these two PRs are pretty compatible (at least from my perspective), @Cookie04DE can you give your opinion?

Cookie04DE:
Yes, I believe so too. More specifically I think `Validate` should check for any errors in the child widgets (your logic) and if there are none then check for the error returned by the form's validator (my logic). Though I think my `SetValidationError` would replace your `setValidationError`, since mine is exported.

chabad360:
There is a bit of logic from my `setValidationError` that you would need to use but other than that, it should be good. 

Also, I should probably just make it public.

andydotxyz:
> Also, I should probably just make it public.

We try never to expose new public API unless there is a clear use-case. So "just make it public" isn't something normally said ;).

chabad360:
Fair enough, I figured that since the only other place that has it, has it public, this should probably be public too.

andydotxyz:
> Fair enough, I figured that since the only other place that has it, has it public, this should probably be public too.

Yes I see that too, and I realise this is not straight forward. Maybe it was the right thing to do.
However not all validation has `SetValidationError` (public or private) so in the bigger picture of all validatable widgets, what is the right thing to do?

Apart from this point it looks like a great change and I am glad that @Cookie04DE is happy to build on it in the larger PR. What do you think @Jacalz ?

Jacalz:
> Apart from this point it looks like a great change and I am glad that @Cookie04DE is happy to build on it in the larger PR. What do you think @Jacalz ?

Sounds good to me. I'm glad that everyone seems to be happy with landing this first. It might make the other PR easier to review as well given that a subset of it already will have landed by then.

chabad360:
In some further testing, I discovered a bug where the error value didn't update properly because it was missing an initial `errFormItemInitialState` (so there was no change when the validation changed). This should be fixed with [`5ad11c7`](https://github.com/fyne-io/fyne/pull/2871/commits/5ad11c7062314a9335ee9e9a7f0f67acb23a63f3). 

But now it makes me wonder if there might be a better way of initializing this, perhaps adding an `Initialize()` call or similar to the interface so that it's clear you need to set it up before use, or can the widget perhaps do it itself on `SetOnValidationChange()` that way it's always certain to be cleared when the callback is set, etc. Either way, there needs to be a better way communicate this requirement, because if a custom widget isn't based on `Entry` it will have this problem (provided the changed state it enters is a `nil` error).

andydotxyz:
> Either way, there needs to be a better way communicate this requirement, because if a custom widget isn't based on `Entry` it will have this problem (provided the changed state it enters is a `nil` error).

I'm not really sure what this has to do with custom widgets or Entry either - it is an internal detail of how form items are validated isn't it?

chabad360:
It exists because forms don't use the error of an item upon initialization (because we want to be nice to the end users on launch). To make sure that when an `Entry` that is invalid changes it's status we do something about it (i.e. enable submit), we set the error of the `Entry` to `errFormItemInitialState` so when `onValidationChange()` gets called later, it's always updated.

The issue with this behavior is that it only happens for an `Entry`, anything else (like a form) needs to become invalid and then valid again (because `SetValidationError()` isn't in the interface, only in the `Entry`).

Now that I'm less tired, I think the best solution to the problem is just to make `SetValidationError()` part of the `Validator` interface. That way we can do [this](https://github.com/fyne-io/fyne/blob/master/widget/form.go#L237) on all form items instead of just `Entry`s.

andydotxyz:
Ah I see your point. What do you think @Jacalz ?

Jacalz:
Sorry for the late reply. I think it makes sense, but we need to be careful given that changing interfaces is a braking change.

chabad360:
So here's the question, should I add that change to this PR? Or should we make a separate one as an RFC to try and better alert anyone that may be caught by surprise from this change?

andydotxyz:
Our contract is not to break things until 3.0 - so we can't just add the new method to the interface - it will break all custom validators.
Is there no way that we can do this in another way that avoids putting this on developers? It seems like we are pushing something onto their code that should not be required but is an issue with how we have implemented validation...

Looking for out of the box ideas as to how we can get this right I guess :)

andydotxyz:
If you want to land this PR can it be done without the new method exposed? If this is to support embedding form in other form we can probably gloss over that for now...

chabad360:
> Our contract is not to break things until 3.0 - so we can't just add the new method to the interface - it will break all custom validators.

I'll admit I'm curious to see how many custom validators don't already have a `SetValidationError()`.

> Is there no way that we can do this in another way that avoids putting this on developers? It seems like we are pushing something onto their code that should not be required but is an issue with how we have implemented validation...

I'm not really sure there is, because any other container that holds validators would need to be able to do the same thing. So wither we have some not-simple logic to deal with this issue, or we just pre-set the validation error (which is nice and simple, but breaks an interface).

> If you want to land this PR can it be done without the new method exposed? If this is to support embedding form in other form we can probably gloss over that for now...

Yes it can, I've dealt with it in [`5ad11c7`](https://github.com/fyne-io/fyne/pull/2871/commits/5ad11c7062314a9335ee9e9a7f0f67acb23a63f3), but it's kinda dumb. Although, this isn't just an issue for that use case. If someone puts anything between the Form and the Validator, it will cause a problem.

chabad360:
Is there anything else I need to for this to get approved?

andydotxyz:
> Is there anything else I need to for this to get approved?

Yes, I don't think my previous comment about the new external function was really addressed - as far as I can see it is not needed in this implementation, so should be left private for further discussion about how we manage the complications of the initial validation state.

> This isn't just an issue for that use case. If someone puts anything between the Form and the Validator, it will cause a problem.

The same is true for any combination of elements, as validation has to be passed up to the top level - anything that is not `Validatable` in the chain will block it.

andydotxyz:
In summary of the above the issue of initial state for validation of a form should be separate and I think can be solved in a more elegant way (for example if the form was itself tracking if it had tested validation before assuming pass), so the additional external method may not be required when we solve it better.

chabad360:
If you want I can make it private. I made it public simply because the other implementations I came across had it public as well.

chabad360:
Perhaps you could leave an actual review on the code, that it's more clear what you want changed.

chabad360:
Alrighty, lmk if you need any other changes.

andydotxyz (approved):
Thanks, this looks good to me now. @Jacalz did you want a quick feedback as well?

Jacalz (changes requested):
I have a few notes on the code. Sorry for the late review.

widget/form.go
```diff
@@ -243,6 +268,29 @@ func (f *Form) setUpValidation(widget fyne.CanvasObject, i int) {
 	}
 }
 
+func (f *Form) setValidationError(err error) {
+	if err == nil && f.validationError == nil {
+		return
+	}
+
+	if (err == nil && f.validationError != nil) || (f.validationError == nil && err != nil) ||
+		err.Error() != f.validationError.Error() {
```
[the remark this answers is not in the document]
chabad360: To be honest, I copied this straight out of the logic for [entry](https://github.com/fyne-io/fyne/blob/master/widget/entry_validation.go#L40).

widget/form_test.go
```diff
@@ -355,3 +355,68 @@ func TestForm_HintsRendered(t *testing.T) {
...
+	}
+
+	validationError := false
+
+	form.SetOnValidationChanged(func(err error) {
+		if err != nil {
+			validationError = true
+		} else {
+			validationError = false
+		}
```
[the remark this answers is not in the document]
chabad360: Lol.

widget/form.go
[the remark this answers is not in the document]
Jacalz: I think that code was written before we supported Go 1.14 as minimum (Go 1.13 added the API in question and we supported Go 1.12 at the time).

widget/form.go
```diff
@@ -243,6 +268,29 @@ func (f *Form) setUpValidation(widget fyne.CanvasObject, i int) {
 	}
 }
 
+func (f *Form) setValidationError(err error) {
+	if err == nil && f.validationError == nil {
+		return
+	}
+
+	if (err == nil && f.validationError != nil) || (f.validationError == nil && err != nil) ||
+		!errors.Is(err, f.validationError) {
```
Jacalz, line 277: Sorry, I wasn't thinking straight when I sent the old comment. The correct way is to do it like this:
```suggestion
	if !errors.Is(err, f.validationError) {
```

Jacalz (changes requested):
I was slightly incorrect in my last review. This is the right way to do it.

widget/form.go
```diff
@@ -119,6 +124,25 @@ func (f *Form) Disabled() bool {
 	return f.disabled
 }
 
+// SetOnValidationChanged is intended for parent widgets or containers to hook into the validation.
```
Jacalz, line 127: Sorry for not realising this, but should we not have the `// The function might be overwritten by a parent that cares about child validation (e.g. widget.Form).` comment that the entry also has?

widget/form.go
```diff
@@ -119,6 +124,25 @@ func (f *Form) Disabled() bool {
...
+func (f *Form) SetOnValidationChanged(callback func(error)) {
+	if callback != nil {
+		f.onValidationChanged = callback
+	}
+}
```
Jacalz, line 132: In https://github.com/fyne-io/fyne/pull/2967/commits/23a0d89a27f50352396bc1005af0fa766e2a7595 I removed this from the entry. Don't you also think that it perhaps makes sense to allow removing the callback? This a comment not a request. I want to know your opinion :)

Jacalz (changes requested):
Sorry. I found two things but I swear that this is approved when you have answered :)

widget/form.go
chabad360: It's a good question, and I don't see why not. But it also leads to a bigger question, should it be able to have multiple callbacks (perhaps this question (of multiple) should move to #2964)?

widget/form.go
chabad360: I feel like a case where it would be overwritten (by say a parent form) is a bit rare, but I don't think there's a good reason why not.

widget/form.go
Jacalz: I think multiple ones would have to occur at a later date. We can start small at first, without introducing breaking changes into the API.

widget/form.go
Jacalz: I think it makes sense to at least indicate that it might be overwritten. The function is the same one as in entry and that one has the same comment.

widget/form.go
chabad360: fair enough, I'll add that and the corresponding test later tonight.

Jacalz (approved):
Looks good to me. Thanks. Sorry for all the back and forth. I have been very busy lately.
Btw, please don't force push when you don't have to. It has a habit of destroying the inline comments here on GitHub and generally makes reviewing harder.

chabad360:
> Btw, please don't force push when you don't have to. It has a habit of destroying the inline comments here on GitHub and generally makes reviewing harder.

Fair enough, I did it most recently because some commits from develop were accidentally merged into my branch and so I needed to drop them.

PR #2878, opened 2022-03-30
Update jackmordaunt/icns/v2 to v2.2.0

d4x1:
### Description:

[jackmordaunt/icns](https://github.com/JackMordaunt/icns) is a library that can convert *.png file to *.icns file, and it releases a new version with [tag v2.2.0](https://github.com/JackMordaunt/icns/tree/v2.2.0).  This latest release fixes [some bugs](https://github.com/JackMordaunt/icns/pull/10) and  can make generated icons' edges on macOS smooth.

Please consider merging it to Fyne's develop branch and hope it will help Fyne to be better.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/go-ole/go-ole v1.2.6
 	github.com/godbus/dbus/v5 v5.1.0
 	github.com/goki/freetype v0.0.0-20181231101311-fa8a33aabaff
-	github.com/jackmordaunt/icns/v2 v2.1.3
+	github.com/jackmordaunt/icns/v2 v2.2.0
 	github.com/josephspurrier/goversioninfo v0.0.0-20200309025242-14b0ab84c6ca
 	github.com/lucor/goinfo v0.0.0-20210802170112-c078a2b0f08b
 	github.com/mcuadros/go-version v0.0.0-20190830083331-035f6764e8d2
--- a/go.sum
+++ b/go.sum
@@ -194,8 +194,6 @@ github.com/hashicorp/serf v0.8.2/go.mod h1:6hOLApaqBFA1NXqRQAsxw9QxuDEvNxSQRwA/J
 github.com/ianlancetaylor/demangle v0.0.0-20181102032728-5e5cf60278f6/go.mod h1:aSSvb/t6k1mPoxDqO4vJh6VOCGPwU4O0C2/Eqndh1Sc=
 github.com/ianlancetaylor/demangle v0.0.0-20200824232613-28f6c0f3b639/go.mod h1:aSSvb/t6k1mPoxDqO4vJh6VOCGPwU4O0C2/Eqndh1Sc=
 github.com/inconshreveable/mousetrap v1.0.0/go.mod h1:PxqpIevigyE2G7u3NXJIT2ANytuPF1OarO4DADm73n8=
-github.com/jackmordaunt/icns/v2 v2.1.3 h1:LBBT9k6Rvnbx+peHFNVQU9klGs0jGol/pTd0EJf0+l8=
-github.com/jackmordaunt/icns/v2 v2.1.3/go.mod h1:6aYIB9eSzyfHHMKqDf17Xrs1zetQPReAkiUSHzdw4cI=
 github.com/jackmordaunt/icns/v2 v2.2.0 h1:Q+V+6c09VNTfRK1cQpBHC4qBkoLiHkxGXOKAQQSLGSw=
 github.com/jackmordaunt/icns/v2 v2.2.0/go.mod h1:6aYIB9eSzyfHHMKqDf17Xrs1zetQPReAkiUSHzdw4cI=
 github.com/josephspurrier/goversioninfo v0.0.0-20200309025242-14b0ab84c6ca h1:ozPUX9TKQZVek4lZWYRsQo7uS8vJ+q4OOHvRhHiCLfU=
--- a/go.sum
+++ b/go.sum
@@ -196,6 +196,8 @@ github.com/ianlancetaylor/demangle v0.0.0-20200824232613-28f6c0f3b639/go.mod h1:
 github.com/inconshreveable/mousetrap v1.0.0/go.mod h1:PxqpIevigyE2G7u3NXJIT2ANytuPF1OarO4DADm73n8=
 github.com/jackmordaunt/icns/v2 v2.1.3 h1:LBBT9k6Rvnbx+peHFNVQU9klGs0jGol/pTd0EJf0+l8=
 github.com/jackmordaunt/icns/v2 v2.1.3/go.mod h1:6aYIB9eSzyfHHMKqDf17Xrs1zetQPReAkiUSHzdw4cI=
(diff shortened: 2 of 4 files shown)


Jacalz (changes requested):
Thanks a lot for the PR. Are you certain that have actually updated the vendor folder with`go mod vendor`? There are way fewer changes than what I would expect.

Also, I just want to give a friendly tip on how to improve your PR. It is always a good idea to write an informative title for the PR and then write a description about why you want to merge the change. In other words, instead of writing "Update go.mod" you could write something like "Update icns/v2 to 2.2.0" and the write a description about the change and perhaps include a link to https://github.com/JackMordaunt/icns/pull/10. I hope this helps for this and future PRs :)

d4x1:
> Thanks a lot for the PR. Are you certain that have actually updated the vendor folder with`go mod vendor`? There are way fewer changes than what I would expect.
> 
> Also, I just want to give a friendly tip on how to improve your PR. It is always a good idea to write an informative title for the PR and then write a description about why you want to merge the change. In other words, instead of writing "Update go.mod" you could write something like "Update icns/v2 to 2.2.0" and the write a description about the change and perhaps include a link to [JackMordaunt/icns#10](https://github.com/JackMordaunt/icns/pull/10). I hope this helps for this and future PRs :)

hi @Jacalz 
Thanks for your advice, and it is helpful.  I update this PR's title and description.  And I have run the command `go mod vendor` in my branch, it is in line with expectations because only one file was updated.

Jacalz:
> And I have run the command `go mod vendor` in my branch, it is in line with expectations because only one file was updated.

Ah, I see now. Sorry about the fuss. GitHub mobile didn't display that diff for some reason.

Jacalz (approved):
Looks good to me. Thanks for opening the PR. Also, perhaps perhaps the most important, congratulations for landing your first contribution to Fyne :tada:

andydotxyz:
Really sorry to say this, but when running fyne package my app icons now look like this on macOS:
Screenshot 2022-04-01 at 13 55 58

It seems to only impact some image sizes but is a serious regression.
Can you look into this please @d4x1, or I may have to revert the merge to get back correct icon packaging.

andydotxyz:
The strangest thing is that if you examine the file all the image sizes are good, but when bundled into the app it doesn't always work well.

andydotxyz:
The difference between old, working, file and the new seems to be that icons are doubled up on the new file - 2 different DPI for each resolution. Which also results in the file being twice the size...

d4x1:
@andydotxyz 
Could you show me how reproduce this bug?

I check this potential bug as follows:
```
go install fyne.io/fyne/v2/cmd/fyne@develop   
fyne version
# output: fyne cli version: v2.1.4-0.20220401172244-2d1c546ccca1 
cd fyne_demo
fyne package
```
And I opened `Fyne Demo.app` and it works fine.
图片

As for the size, the icns v2.2.0 version will produce 10 icon images for incs file, while lower version will produce only 6 images. It's normal that icon file's size is larger then before. That's the reason I update lib icns.  
The icon looks great in Finder. like this:
图片


but when I move it into LaunchPad, it looks like this:
图片

andydotxyz:
Using "macOS Big Sur" simply do what you have above. However it is only visible in finder "list view" not the "icon view"

andydotxyz:
I also cannot replicate on Monterey

andydotxyz:
If I copy the item built on Big Sur to Monterey the icon is blank and appears corrupted. When I launch it on Monterey this issue appears to fix itself....

d4x1:
> 

I have on mac running Big Sur so i cannot directly debug and try to fix this problem. 

I have an idea,  we only need to check whether this 'broken icon'  was really broken.  You can start a new project with lib icns v2.2.0 and convert a simple a PNG to icns file. then  compare the different icns files generated on Big Sur and Montery. 

I think it will be helpful for us to find what's going wrong.

Here is my code:
```go
func main() {
	pngf, err := os.Open("./icon.png")
	if err != nil {
		log.Fatalf("opening source image: %v", err)
	}
	defer pngf.Close()
	srcImg, _, err := image.Decode(pngf)
	if err != nil {
		log.Fatalf("decoding source image: %v", err)
	}
	dest, err := os.Create("./icon.icns")
	if err != nil {
		log.Fatalf("opening destination file: %v", err)
	}
	defer dest.Close()
	if err := icns.NewEncoder(dest).Encode(srcImg); err != nil {
		log.Fatalf("encoding icns: %v", err)
	}
	fmt.Println("end")
}
```

Also, I will try to find the difference between  Big Sur and Montery in dealing with icns files,  it will be not easy( I am not familiar with macOS development).

andydotxyz:
The icons in the icns are not corrupted, it’s the usage of them that seems to break. As I noted above the difference i can see is how many icons there are of each size in the file, on examination each individual icon is ok, but the app icon rendering is not.

andydotxyz:
> hi @andydotxyz, As you can see, I haven't changed any logic on rendering icons in OS, so this PR maybe is not the root cause of this problem in theory.

Sorry I am not talking about Fyne's rendering of the app icons, this is in macOS interpretation of the icon file.
This PR is certainly the root cause of the issue because reversing the change removes the problem on Montery.

andydotxyz:
I have looked at the changelog of the icns project and I think the change is this one:

```
var osTypes = []OsType{
	{ID: "ic10", Size: uint(1024)},
	{ID: "ic14", Size: uint(512)},
	{ID: "ic09", Size: uint(512)},
	{ID: "ic13", Size: uint(256)},
	{ID: "ic08", Size: uint(256)},
	{ID: "ic07", Size: uint(128)}, 
	{ID: "ic12", Size: uint(64)},
+ 	{ID: "icp6", Size: uint(64)},
	{ID: "ic11", Size: uint(32)},
+	{ID: "icp5", Size: uint(32)},
+	{ID: "icp4", Size: uint(16)},
}```

What you can see is that there are new duplicate sizes. In a prior commit I added 2x support for certain sizes so what looks like duplicate is not. However this new change provides two different ways to look  up the 64x64 and 32x32 icons which could be problematic?

Just a thought, maybe we should report this upstream?

andydotxyz:
Update: I removed the new image formats and that was not the cause of the problem

d4x1:
Your testing provides an important clue,  and after searching about this problem, I found some other projects have encountered such problems. 
https://bugreports.qt.io/browse/QTBUG-74497
https://github.com/godotengine/godot/issues/23968

As [wiki](https://en.wikipedia.org/wiki/Apple_Icon_Image_format) says:
图片

It's certainly a macOS's bug(or feature). 

I think the best way to solve this problem is removing icp4、icp5、icp6 icons from lib icns.  And I test it, it works. The icons on Montery are still smooth which is my goal to submit this PR.

How are about your advice?

andydotxyz:
> I think the best way to solve this problem is removing icp4、icp5、icp6 icons from lib icns. And I test it, it works. The icons on Montery are still smooth which is my goal to submit this PR.

This sounds like the right thing to do. If you could open a PR upstream to do the change we could then use the version that results to resolve the issue for our packager.

andydotxyz:
Resolved with the latest PR, thanks @d4x1

PR #2880, opened 2022-03-30
custom FavoriteOrder

qmsu:
### Description:

custom FavoriteOrder

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -34,6 +34,11 @@ type favoriteItem struct {
 	loc     fyne.URI
 }
 
+type FavoriteOrder struct {
+	ShowName string
+	Name     string
+}
+
 type fileDialog struct {
 	file             *FileDialog
 	fileName         textWidget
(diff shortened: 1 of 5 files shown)


andydotxyz:
The title of this (and the documentation) does not seem to be right, this is not setting custom order, it is setting complete custom icons and indeed entries.
If that is desired then it should be updated and adding `FavoriteOrder` makes less sense.

In other discussions we have talked about having custom items appear under the standard icons... So if you want to add new items maybe that is a way to do it.
In short it seems like this PR is a mix of two different things, so looking for clarity before doing a code review.

qmsu:
i want to change the complete custom icons and indeed entries, and rename it.but now has not method to change。Are you considering adding features?

andydotxyz:
> i want to change the complete custom icons and indeed entries, and rename it.but now has not method to change。Are you considering adding features?

It depends on the reason for the change. Correcting names for local equivalents is planned in our localisation/internationalisation work. Adding custom elements should be separate from OS/standard favourites.
So to answer the question can you give more information on the use-case?
If it is complete new entries let's get the naming fixed so it does not imply just order change.

andydotxyz:
You don't need to close a PR if work is required on it... much easier to work through the changes here so developers can review the changes and offer assistance.

PR #2885, opened 2022-03-31
Feature: menu icons

### LINKED ISSUE #1752: MenuItem support Icon

toaster:
### Description:

This PR adds support for icons to (non-native) menus.
It also fixes a bug where the expand icon of a deactivated menu item with submenu was not greyed out.

Fixes #1752

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
This is excellent thanks @toaster. Is the plan to do the macOS icons in a separate PR?

toaster:
@andydotxyz, here’s the MacOS native support.

andydotxyz:
This is so cool!
One question - is it intentional that the `ColorizeSVG` is a new public method?
If so it's missing the "Since:" tag. If not then it could be internal for now...

toaster:
Ah, yes, I wanted to move it to `internal/painter` first but that gave me a cyclic dependency.
I could add a new internal package `svg` for this. I would also suggest to move the SVG to image conversion there.
WDYT?

andydotxyz:
> Ah, yes, I wanted to move it to `internal/painter` first but that gave me a cyclic dependency. I could add a new internal package `svg` for this. I would also suggest to move the SVG to image conversion there. WDYT?

This makes sense. Out of interest what was the cycle?

toaster:
> This makes sense. Out of interest what was the cycle?

When moving the method from `theme` to `internal/painter` the `theme` package of course gets dependent on `internal/painter` which already depends on `theme`.

andydotxyz:
Thanks for tidying this up. I am not quite sure of `PaintImageWithFillModeIgnored` - isn't this exactly the same as passing `PaintImage` where the image as any fill except `ImageFillOriginal`? It looks like an API that does not add functionality...

toaster:
> Thanks for tidying this up. I am not quite sure of `PaintImageWithFillModeIgnored` - isn't this exactly the same as passing `PaintImage` where the image as any fill except `ImageFillOriginal`? It looks like an API that does not add functionality...

Yes, but:
The places calling this method don’t provide a canvas (because they probably don’t even have one). If you do this to the other `PaintImage` method and the image happens to have `ImageFillOriginal`, the app will crash with an NPE.

I think an API should not be designed around the provided feature(s) but should fulfil purposes as exact as possible.
For OS icon creation (render a Fyne image into a plain Go image) we don’t want to care about things like fill mode or providing a canvas. Whereas these are important for rendering a widget onto the screen.
It might be that both tasks share a lot of common code but that’s an implementation detail.

One could argue that a single `PaintImage` method could ignore the fill mode when no canvas is given but I think this would be non-obvious behaviour which is hard to justify.

andydotxyz:
> If you do this to the other PaintImage method and the image happens to have ImageFillOriginal, the app will crash with an NPE.

Can we fix the nil crash rather than working around it?

If you don't care about the fill mode then don't set one and the `PaintImage` existing API will work, that is why we have sensible zero values for most fields. (the special case is FillOriginal which has to be set intentionally).

andydotxyz (approved):
Excellent, thank you so much!

Issue #2814, opened 2022-03-01
MultiLineEntry improperly displayed when having long text inside Scroll container


wsdexy:
### Describe the bug:
MultiLineEntry improperly displayed when having long text inside Scroll container, scroll down and you will see it.

### To Reproduce:
Run the Example code.

### Screenshots:
![bug_screen1](https://user-images.githubusercontent.com/17593203/156194507-56c39c6b-df6b-48a4-b2a4-e029d9e11ced.PNG)
![bug_screen2](https://user-images.githubusercontent.com/17593203/156194517-50595ce7-e50a-4306-a988-17df57a31a09.PNG)

### Example code:
```go
appInstance := app.NewWithID("appInstance")
win := appInstance.NewWindow("window")

longText := ""
for i := 0; i < 1024; i++ {
    longText += fmt.Sprintf("%v ", i)
}

entry := widget.NewMultiLineEntry()
entry.Wrapping = fyne.TextWrapWord
//entry.Disable()
entry.SetText(longText)

formItemSlice := []*widget.FormItem{
    &widget.FormItem{
        Text:     "entry",
        Widget:   entry,
        HintText: "",
    },
}
for i := 0; i < 32; i++ {
    formItemSlice = append(formItemSlice, &widget.FormItem{
        Text:     "Label",
        Widget:   widget.NewLabel("Label"),
        HintText: "Label",
    })
}

form := widget.NewForm(
    formItemSlice...,
)

win.SetContent(container.NewVScroll(form))
win.Resize(fyne.NewSize(800, 600))
win.ShowAndRun()
```

### Device (please complete the following information):
 - **OS:** win10
 - **Version:** 21H1
 - **Go version:** 1.17
 - **Fyne version:** v2.0

andydotxyz:
Can you please try updating to Fyne v2.1.3? The text, Entry and Scrolling have been updated since last January.

wsdexy:
![image](https://user-images.githubusercontent.com/17593203/156715863-eb887982-ee47-4fcb-aae4-a0cf1ccd2b63.png)
@andydotxyz 
Thanks for replay, updated to Fyne v2.1.3, still improperly displayed.

ranjith-bhide:
Hi All, 
I was able to find the root cause of this issue. 
I find this happening when we have a scroll container inside a scroll container. 
[...]

PR #2893, opened 2022-04-03
Width and Height params made non-negative for Opengl's Scissor Call

ranjith-bhide:
Fixes (#2814)

 

### Description:

Fixes (#2814)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/internal/clip.go
+++ b/internal/clip.go
@@ -77,5 +77,11 @@ func (i *ClipItem) Intersect(p fyne.Position, s fyne.Size) *ClipItem {
 	if p.Y+s.Height > i.pos.Y+i.size.Height {
 		ret.size.Height = (i.pos.Y + i.size.Height) - ret.pos.Y
 	}
+
+	if ret.size.Width < 0 || ret.size.Height < 0 {
+		ret.size = fyne.NewSize(0, 0)
+		return ret
+	}
+
 	return ret
 }
--- a/internal/driver/glfw/canvas.go
+++ b/internal/driver/glfw/canvas.go
@@ -275,6 +275,14 @@ func (c *glCanvas) paint(size fyne.Size) {
 			inner := clips.Push(pos, obj.Size())
 			c.Painter().StartClipping(inner.Rect())
 		}
+
+		if visible_area := clips.Top(); visible_area != nil {
+			_, paint_area := visible_area.Intersect(pos, size).Rect()
+			if paint_area.Width <= 0 || paint_area.Height <= 0 {
+				return
+			}
+		}
+
 		c.Painter().Paint(obj, pos, size)
 	}
 	afterPaint := func(node *common.RenderCacheNode) {


andydotxyz:
Thanks for working on this. Are you sure it’s the correct fix though? A negative scissor is certainly wrong, but a 0 scissor would result in nothing shown, so should we skip rendering it?

ranjith-bhide:
> so should we skip rendering it?

Yes, That is what is expectated.
and this would be one way of getting OpenGL Scissors to do it for us, following the generic design.

andydotxyz:
Looks to me like you found it :) yes indeed the clip stack is something to consider.
I guess you could push/pop the stack but just skip the paint call when rect is 0? Possibly could even skip the scissor open and close as they both know if the rectangle is 0 right?

andydotxyz:
> as in the afterPaint function the pop happens unconditionally for all scrollable objects, hence the pop may happen at an incorrect point.

I'm not sure about this - the afterPaint has the same opportunity to check the size of the object in the (cache) node?
Checking for zero size in both methods would assume the size does not change during walk but that seems reasonable.
Or have I missed something?

ranjith-bhide:
> I'm not sure about this - the afterPaint has the same opportunity to check the size of the object in the (cache) node?
Checking for zero size in both methods would assume the size does not change during walk but that seems reasonable.
Or have I missed something?

Sorry I was probably not clear when I mentioned the root cause.
It is not that the Object Itself is having the Width/Height as negative value. Its the `inner` rectangle which is the intersection of the parent Scroll object and the child Scroll Object that gets a negative width/height.
Screenshot from 2022-04-04 23-34-41

And as we are storing this `inner` variable only in the clips stack and nowhere else. if we do not push this to stack I dont think we will be able determine when this particular Object's `afterPaint` is getting called.

I hope this make sense @andydotxyz.

andydotxyz:
Ah yes thank you, I should have noticed sorry. I guess this is about the simplest fix for a complex problem then :)

ranjith-bhide:
> Ah yes thank you, I should have noticed sorry. I guess this is about the simplest fix for a complex problem then :)

I may be suggesting almost the same thing again, but the simplest fix for this problem I see is this:
```diff
diff --git a/internal/clip.go b/internal/clip.go
index d8983ce8..da8fbcd5 100644
--- a/internal/clip.go
+++ b/internal/clip.go
@@ -58,24 +58,28 @@ func (i *ClipItem) Rect() (fyne.Position, fyne.Size) {
 // Intersect returns a new clip item that is the intersection of the requested parameters and this clip.
 func (i *ClipItem) Intersect(p fyne.Position, s fyne.Size) *ClipItem {
        ret := &ClipItem{p, s}
        if i == nil {
                return ret
        }
 
        if ret.pos.X < i.pos.X {
                ret.pos.X = i.pos.X
                ret.size.Width -= i.pos.X - p.X
        }
        if ret.pos.Y < i.pos.Y {
                ret.pos.Y = i.pos.Y
                ret.size.Height -= i.pos.Y - p.Y
        }
 
        if p.X+s.Width > i.pos.X+i.size.Width {
                ret.size.Width = (i.pos.X + i.size.Width) - ret.pos.X
        }
        if p.Y+s.Height > i.pos.Y+i.size.Height {
                ret.size.Height = (i.pos.Y + i.size.Height) - ret.pos.Y
        }
+
+       /* Ensure Intersection Area Height and Width are always non-negative */
+       ret.size = ret.size.Max(fyne.NewSize(0, 0))
+
        return ret
 }
```

Here Im trying to ensure that the Returned rectangle of Intersection is always maintained to be of non-negative width and height. Because saying that the intersection of 2 rects have a negative length is absurd.
And thus this makes a bit more logical sense than simply seeming like a work-around (as in my original commit)

@andydotxyz Let me know if this makes sense, Ill update the PR.

andydotxyz:
Using this simple fix will not stop us from drawing the items that should not be visible. It is simpler but it does not save lost performance, which is a shame.

Instead of having a comment to explain code we prefer code that explains itself, so I would consider this alternative:
```go
       if ret.size.Width < 0 || ret.size.Height < 0 {
              return fyne.NewSize(0, 0)
       }
```

This code is clear in its intent - check for invalid sizes and return 0

ranjith-bhide:
> Using this simple fix will not stop us from drawing the items that should not be visible. It is simpler but it does not save lost performance, which is a shame.

Hi @andydotxyz. I spent a bit more time, and I see one way to actually achieve that performance optimisation. :) 
Let me know what you think.
```diff
index d8983ce..15e2afc 100644
--- a/internal/clip.go
+++ b/internal/clip.go
@@ -58,24 +58,30 @@ func (i *ClipItem) Rect() (fyne.Position, fyne.Size) {
 // Intersect returns a new clip item that is the intersection of the requested parameters and this clip.
 func (i *ClipItem) Intersect(p fyne.Position, s fyne.Size) *ClipItem {
 	ret := &ClipItem{p, s}
 	if i == nil {
 		return ret
 	}
 
 	if ret.pos.X < i.pos.X {
 		ret.pos.X = i.pos.X
 		ret.size.Width -= i.pos.X - p.X
 	}
 	if ret.pos.Y < i.pos.Y {
 		ret.pos.Y = i.pos.Y
 		ret.size.Height -= i.pos.Y - p.Y
 	}
 
 	if p.X+s.Width > i.pos.X+i.size.Width {
 		ret.size.Width = (i.pos.X + i.size.Width) - ret.pos.X
 	}
 	if p.Y+s.Height > i.pos.Y+i.size.Height {
 		ret.size.Height = (i.pos.Y + i.size.Height) - ret.pos.Y
 	}
+
+	if ret.size.Width < 0 || ret.size.Height < 0 {
+		ret.size = fyne.NewSize(0, 0)
+		return ret
+	}
+
 	return ret
 }
diff --git a/internal/driver/glfw/canvas.go b/internal/driver/glfw/canvas.go
index a632f9c..4b12c73 100644
--- a/internal/driver/glfw/canvas.go
+++ b/internal/driver/glfw/canvas.go
@@ -265,30 +265,38 @@ func (c *glCanvas) overlayChanged() {
 func (c *glCanvas) paint(size fyne.Size) {
 	clips := &internal.ClipStack{}
 	if c.Content() == nil {
 		return
 	}
 	c.SetDirty(false)
 	c.Painter().Clear()
 
 	paint := func(node *common.RenderCacheNode, pos fyne.Position) {
 		obj := node.Obj()
 		if _, ok := obj.(fyne.Scrollable); ok {
 			inner := clips.Push(pos, obj.Size())
 			c.Painter().StartClipping(inner.Rect())
 		}
+
+		if visible_area := clips.Top(); visible_area != nil {
+			_, paint_area := visible_area.Intersect(pos, size).Rect()
+			if paint_area.Width <= 0 || paint_area.Height <= 0 {
+				return
+			}
+		}
+
 		c.Painter().Paint(obj, pos, size)
 	}
 	afterPaint := func(node *common.RenderCacheNode) {
 		if _, ok := node.Obj().(fyne.Scrollable); ok {
 			clips.Pop()
 			if top := clips.Top(); top != nil {
 				c.Painter().StartClipping(top.Rect())
 			} else {
 				c.Painter().StopClipping()
 
 			}
 		}
 	}
 
 	c.WalkTrees(paint, afterPaint)
 }

```

andydotxyz:
Yeah looks great @ranjith-bhide

ranjith-bhide:
Hi @andydotxyz, I've updated the PR with changes okayed by you, Please take it forward for approval and merge.

andydotxyz (approved):
Excellent, thanks

PR #2896, opened 2022-04-07
Support mobile simulator on FreeBSD

Jacalz:
### Description:

This adds support for running the mobile simulator on FreeBSD.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/mobile/app/app.go
+++ b/internal/driver/mobile/app/app.go
@@ -2,8 +2,8 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-//go:build linux || darwin || windows
-// +build linux darwin windows
+//go:build freebsd || linux || darwin || windows
+// +build freebsd linux darwin windows
 
 package app
 
--- a/internal/driver/mobile/app/x11.c
+++ b/internal/driver/mobile/app/x11.c
@@ -2,8 +2,8 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-//go:build linux && !android
-// +build linux,!android
+//go:build linux && !android || freebsd
+// +build linux,!android freebsd
 
 #include "_cgo_export.h"
 #include <EGL/egl.h>
--- a/internal/driver/mobile/app/x11.c
+++ b/internal/driver/mobile/app/x11.c
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-//go:build linux && !android || freebsd
+//go:build (linux && !android) || freebsd
 // +build linux,!android freebsd
 
 #include "_cgo_export.h"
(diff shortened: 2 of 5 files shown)


Jacalz:
Sorry for the late answer. I gave a way my FreeBSD laptop to someone that needed it far more than I did.
I have FreeBSD up and running again on an older laptop now. This is the error that I get:
```
# fyne.io/fyne/v2/internal/driver/mobile/gl
In file included from ../../internal/driver/mobile/gl/work114.go:33:
./work.h:16:10: fatal error: 'GLES3/gl3.h' file not found
#include <GLES3/gl3.h>
```

Jacalz:
Yup. It seems like it tries to look for /usr/include/GLES3/gl3.h instead of /usr/local/include/GLES3/gl3.h (which is the correct location on the disk). I should have added this earlier, sorry. It was in a comment in Slack and I forgot to add it here.

Jacalz:
I have rebased this on the `develop` branch now. We agreed on the call that @andydotxyz will try to have a quick look if there is any simple way to get this to work. If not, we can merge this as it is given that it doesn't break anything but just gets us closer to working mobile simulator on FreeBSD (there are already some basic stuff in place for it).

andydotxyz:
I got it working locally :) it seems that `clang` is used by default and it breaks.
If you do `CC=gcc go run -tags mobile .` it works, with the following changes:

work.go needs to use `/usr/local/` instead of `/usr/local/X11R6` in the CFLAGS and LDFLAGS.
It seems you may also need to patch the godbus file  transport_unixcred_freebsd.go so that line 44 is `salign := int(C.sizeofPtr)` as well.

andydotxyz:
I'm happy for you to if you have time - am trying to coordinate a few other things this week.

Jacalz:
No bother. I'll get that fixed :)

Jacalz:
I pushed the changes for `work.go` now. Will spin up a FreeBSD VM when I have some time over but it might have to be after my exams.

andydotxyz:
Thanks for moving it on :) good luck with the exams!

Jacalz:
I had some time this evening actually and it is mostly working, I think. I opened the PR for godbus at https://github.com/godbus/dbus/pull/359 and I managed to get the `clang` compiler to compile with https://github.com/fyne-io/fyne/pull/2896/commits/8160c450c4a20d31bcdd599f05c4529a211c7074.

However, I am getting crashes about failing to link the OpenGL program. The strange thing there though is that I seem to be getting the exact same error on `develop` locally on Fedora. The latest stable release (and the upcoming bugfix release) is working.

Jacalz:
It is indeed working correctly when applying on `master` :partying_face:
Screenshot from 2023-03-14 21-35-15

I opened #3731 to keep track of the bug but I think this should be fine to merge before that is fixed :)

Jacalz (commented):
We should of course wait for the godbus PR to be merged first and then pull in here before merging :)

Jacalz:
I think this should be fine to merge now actually. After an updated FreeBSD install and after rebasing the branch on `develop`, I am no longer seeing the compile error in godbus. One of the developers at godbus thought the PR looked strange and I retested with a clean Go cache and with a fresh git clone just in case but the error is not there any more. I suspect something was bugged with the Go compiler version that FreeBSD had when testing a few weeks ago.

andydotxyz (approved):

PR #2901, opened 2022-04-08
Lifecycle hook

### LINKED ISSUE #311: How do I set the size and position of the window?
I want to set the size and position of the window. 
Either it can't be found or it doesn't work.

mrjrieke:
### Description:

Fixes #311 - How do I set the size and position of the window?

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
Separate to the discussion about whether this should be exposed (see discussion on #311) this seems to attach a resized event to application lifecycle but this refers to a window. Additionally it seems to assume that a device has a certain size /position but surely that too is window specific?
Not also that on mobile this seems to confuse safe area with window location… but safe area defines what can be draw inside a canvas.

mrjrieke:
This doesn't yet work for Mac or windows, so I will withdraw it until those two have been implemented.

PR #2906, opened 2022-04-09
Add initial notification icons support on Linux and BSD

Jacalz:
### Description:

This is a WIP PR for adding the support for showing the application icon in notifications on Linux and BSD. The code needs to be tidied up and the caching needs to be improved so that we only write it once and so on. 
image

It would also be great if we could reuse the application icon if it has been installed into the `/usr/share/icons` path but that might be a bit harder to pull off. Does anyone know an easy way to get the XDG icon corresponding to a specific name?

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -12,12 +12,15 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"sync"
 
 	"github.com/godbus/dbus/v5"
 
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/theme"
 )
+
+var once sync.Once
 
 func defaultVariant() fyne.ThemeVariant {
 	return theme.VariantDark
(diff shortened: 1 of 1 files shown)


app/app_xdg.go
```diff
@@ -48,6 +48,37 @@ func (a *fyneApp) SendNotification(n *fyne.Notification) {
 	}
 }
 
+func (a *fyneApp) cachedIconPath() string {
+	metadata := a.Metadata()
```
andydotxyz, line 52: Just use a.Icon(), it first checks settings and then metadata.

app/app_xdg.go
```diff
@@ -48,6 +48,37 @@ func (a *fyneApp) SendNotification(n *fyne.Notification) {
...
+	if metadata.Icon == nil || len(metadata.Icon.Content()) == 0 {
+		return ""
+	}
+
+	dirPath := filepath.Join(rootCacheDir(), a.uniqueID)
```
andydotxyz, line 57: This should be `a.UniqueID()` which does things when there is none set...

andydotxyz (changes requested):
Nice direction, just a few comments on the uniqueness/icon lookup

Jacalz:
Sorry for the force push I had to rebase on develop to get the latest CI actions.

andydotxyz (approved):
Pretty cool, thanks :)

PR #2907, opened 2022-04-09
Changed boundExternalUntypedListItem::setIfChanged to use DeepEqual...

kontza:
…  to be able to cope with structs that have embedded arrays in them.

### Description:
With this change one can use an array of structs containing embedded arrays in various operations, e.g. Append.

This came up in my app where I have an array of items containing not only primitive fields, but also an array of strings. When I appended an item to that array, my Fine app promptly crashed with `panic: runtime error: comparing uncomparable type`.

Fixes #(issue)

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/bindlists.go
+++ b/data/binding/bindlists.go
@@ -1,9 +1,12 @@
 // auto-generated
 // **** THIS FILE IS AUTO-GENERATED, PLEASE DO NOT EDIT IT **** //
-
 package binding
 
-import "fyne.io/fyne/v2"
+import (
+	"reflect"
+
+	"fyne.io/fyne/v2"
+)
 
 // BoolList supports binding a list of bool values.
 //
@@ -1327,7 +1330,7 @@ type boundExternalUntypedListItem struct {
 }
 
 func (b *boundExternalUntypedListItem) setIfChanged(val interface{}) error {
-	if val == b.old {
+	if reflect.DeepEqual(val, b.old) {
 		return nil
 	}
 	(*b.val)[b.index] = val
--- a/app.go
+++ b/app.go
@@ -1,4 +1,4 @@
-sackage fyne
+package fyne
 
 import (
 	"net/url"
--- a/data/binding/bindlists_test.go
+++ b/data/binding/bindlists_test.go
@@ -6,6 +6,29 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestDeepItems(t *testing.T) {
+	type DeepItem struct {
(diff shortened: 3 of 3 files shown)


data/binding/bindlists.go
```diff
@@ -1,9 +1,12 @@
 // auto-generated
 // **** THIS FILE IS AUTO-GENERATED, PLEASE DO NOT EDIT IT **** //
```
Jacalz, line 2: I think you missed this line. You need to to update the file that generates this file.

Jacalz (changes requested):
I noticed an issue with how an auto-generated file was created. Apart from that, I guess it might make sense.

Out of curiosity, how much slower is this for things that already was comparable?

data/binding/bindlists.go
kontza: I did notice it, but at that moment hadn't figured out how to generate those. Now I know, and notice, that my current skills are not good enough with this gen-stuff... Now every External list has `DeepEqual`, which is silly, say for boolean arrays. Is there a way I could control the generation so that for ExternalUntyped it would use DeepEquals?

And another thing I noticed: my PR is against master instead of develop as instructed. Should I delete this PR, and start over.

Jacalz (changes requested):
PRs should target against the `develop` branch instead of `master`. We then pick over the changes from the former into the latter once a new release is going out. Please rebase your changes on top of the `develop` branch and change the PR target. This is extra important now because there are various changes to GitHub workflows, which will (due to some of them being required) stop this from being merged in its current form.

Jacalz:
Don't get me wrong here. We are very thankful for your PR. Just a few changes needed to make it land 🙂

data/binding/bindlists.go
Jacalz: I think I wrote a another comment on the PR while you were sending yours here. It should be possible to fix up this PR but you can also create a new if you feel like that is easier.

Hmm, I'll have to look at that. I don't know how easy that will be.

kontza:
I'll start over with this from develop branch and using the generation sources.

data/binding/bindlists.go
kontza: Replying here to keep the context: after checking out how the generation works, I found out, that it was pretty easy: the `bindValues` struct in `gen.go` contains a field, `Comparator`, and if that field is empty equality is done with `==`, and if it is not empty, the given string is used as a function.

PR #2908, opened 2022-04-10
Add SetComparator to ExternalUntypedList

kontza:
### Description:
With this change ExternalUntypedList gets a new function: `SetComparator`. This way one can provide a custom comparator to be used in the package internal function `setIfChanged`. Such a comparator function could be `reflect.DeepEqual`.

This came up in my app where I have an array of items containing not only primitive fields, but also an array of strings. When I appended an item to that array, my Fine app promptly crashed with panic: runtime error: comparing uncomparable type.

p.s. This PR is a re-run of my incorrect one from the previous day. This time the change is in the string template that is used in generating the code, and the change is in the correct branch.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/bindlists.go
+++ b/data/binding/bindlists.go
@@ -222,7 +222,7 @@ type boundExternalBoolListItem struct {
 }
 
 func (b *boundExternalBoolListItem) setIfChanged(val bool) error {
-	if reflect.DeepEqual(val, b.old) {
+	if val == b.old {
 		return nil
 	}
 	(*b.val)[b.index] = val
@@ -666,7 +666,7 @@ type boundExternalFloatListItem struct {
 }
 
 func (b *boundExternalFloatListItem) setIfChanged(val float64) error {
-	if reflect.DeepEqual(val, b.old) {
+	if val == b.old {
 		return nil
 	}
 	(*b.val)[b.index] = val
@@ -888,7 +888,7 @@ type boundExternalIntListItem struct {
 }
 
 func (b *boundExternalIntListItem) setIfChanged(val int) error {
-	if reflect.DeepEqual(val, b.old) {
+	if val == b.old {
 		return nil
 	}
 	(*b.val)[b.index] = val
@@ -1110,7 +1110,7 @@ type boundExternalRuneListItem struct {
 }
 
 func (b *boundExternalRuneListItem) setIfChanged(val rune) error {
-	if reflect.DeepEqual(val, b.old) {
+	if val == b.old {
 		return nil
 	}
 	(*b.val)[b.index] = val
@@ -1332,7 +1332,7 @@ type boundExternalStringListItem struct {
 }
 
 func (b *boundExternalStringListItem) setIfChanged(val string) error {
-	if reflect.DeepEqual(val, b.old) {
+	if val == b.old {
 		return nil
 	}
 	(*b.val)[b.index] = val
--- a/data/binding/bindlists.go
+++ b/data/binding/bindlists.go
@@ -1,10 +1,10 @@
 // auto-generated
 // **** THIS FILE IS AUTO-GENERATED, PLEASE DO NOT EDIT IT **** //
-
 package binding
 
 import (
 	"bytes"
+	"reflect"
 
 	"fyne.io/fyne/v2"
 )
@@ -1553,7 +1553,7 @@ type boundExternalUntypedListItem struct {
 }
 
 func (b *boundExternalUntypedListItem) setIfChanged(val interface{}) error {
-	if val == b.old {
+	if reflect.DeepEqual(val, b.old) {
 		return nil
 	}
 	(*b.val)[b.index] = val
--- a/data/binding/bindlists.go
+++ b/data/binding/bindlists.go
@@ -1,5 +1,6 @@
 // auto-generated
 // **** THIS FILE IS AUTO-GENERATED, PLEASE DO NOT EDIT IT **** //
+
 package binding
 
 import (
@@ -221,7 +222,7 @@ type boundExternalBoolListItem struct {
 }
 
 func (b *boundExternalBoolListItem) setIfChanged(val bool) error {
-	if val == b.old {
+	if reflect.DeepEqual(val, b.old) {
 		return nil
 	}
 	(*b.val)[b.index] = val
@@ -665,7 +666,7 @@ type boundExternalFloatListItem struct {
 }
 
 func (b *boundExternalFloatListItem) setIfChanged(val float64) error {
-	if val == b.old {
+	if reflect.DeepEqual(val, b.old) {
 		return nil
 	}
 	(*b.val)[b.index] = val
@@ -887,7 +888,7 @@ type boundExternalIntListItem struct {
 }
 
 func (b *boundExternalIntListItem) setIfChanged(val int) error {
-	if val == b.old {
+	if reflect.DeepEqual(val, b.old) {
 		return nil
 	}
 	(*b.val)[b.index] = val
@@ -1109,7 +1110,7 @@ type boundExternalRuneListItem struct {
 }
 
 func (b *boundExternalRuneListItem) setIfChanged(val rune) error {
-	if val == b.old {
+	if reflect.DeepEqual(val, b.old) {
 		return nil
 	}
 	(*b.val)[b.index] = val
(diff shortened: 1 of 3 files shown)


Jacalz (approved):
This seems sensible to me, but let's hear what the others have to say before merging.

Bluebugs:
I am a bit worried with this change that make the test much more costly. As anyone done any benchmark to check this impact?

Jacalz:
> I am a bit worried with this change that make the test much more costly. As anyone done any benchmark to check this impact?

Good catch. I asked the question on the old PR but I forgot about asking it again here.

kontza:
I haven't done any benchmarking on that. I know that using reflection might be costly, but my use case seems to require a deep comparison. My ExternalUntypedList's entries are in the lines as the following:
```
type listItem struct {
  name string
  subItems []string
}
```
Without `DeepEqual`, `list.Append` dies on panic due to "comparing uncomparables."

Jacalz:
I think we might need to benchmark this to be sure and I also found some related information on the topic.

Related things that I found:
- https://www.sobyte.net/post/2021-12/struct-compare/
- https://github.com/docker/swarmkit/issues/198

Would it perhaps make more sense to allow people to implement custom comparator functions?

Bluebugs:
If a custom comparison functions could be set, with a default using a deep equal, I don't think we would be worried of performance.

Jacalz:
I didn't necessarily mean that you'd need to use the deep equal for that. If you could define a custom comparator for the strict above, it could just, j side the function, use regular compares for the two fields in the struct and return a bool for the result.

Jacalz:
What I am trying to say is that we could do something similar the "Comparable" interface in Java. If the data type that we get is comparable, we use it's own compare function and otherwise we just fall back to the deep equal.

Bluebugs:
I like that idea @Jacalz

kontza:
The second option sounds nice to me, too. How would one check in Go if something is comparable? AFAIK, there is no `equals()` as there is in Java.

Jacalz:
> The second option sounds nice to me, too. How would one check in Go if something is comparable? AFAIK, there is no `equals()` as there is in Java.

You just type cast to check if the variable implements the specified interface. Something like this:
```go
cmp, ok := object.(binding.Comparable)
if !ok {
    // Not comparable, and so on…
}
```

kontza:
What if we changed the generated code to look like this?
```
func BindUntypedList(v *[]interface{}) ExternalUntypedList {
	return BindUntypedListWithCustomComparator(v,nil)
}

type UntypedListItemComparator func(a interface{}, b interface{} ) bool

func BindUntypedListWithCustomComparator(v *[]interface{}, comparator UntypedListItemComparator) ExternalUntypedList {
	...
	b := &boundUntypedList{val: v, updateExternal: true, comparator: comparator}
```

This way existing code would bind untyped lists as they have earlier done. New code, wanting to to use custom comparator would then call
```
BindUntypedListWithCustomComparator(v, reflect.DeepEqual)
```

I came up with this as I couldn't figure out where to put the Comparable interface. It would make sense to have it somewhere in `bindlists.go`, but where it's used is at least two levels deep inside "private" types/methods.

Now, looking at the proposed `...WithCustomComparator` above...
Maybe `ExternalUntypedList.SetComparator(reflect.DeepEqual)` would be better after all. Then the given comparator function could be passed further down to the "private" parts of the package. Compatibility with old code would be preserved, new code could use the new `SetComparator` function, if needed.

kontza:
`ExternalUntypedList` has now `SetItemComparator`. `gen.go` template adds this only to those whose `Name == Untyped`.

kontza:
@Jacalz Hi, are you familiar with this static_analysis that now does not pass?

Jacalz:
> @Jacalz Hi, are you familiar with this static_analysis that now does not pass?

Yes. I think it is complaining about all new lines that have been added and removed in various places. You’ll see what I mean if you check the diff here on GitHub.

data/binding/bindlists.go
```diff
@@ -26,6 +26,7 @@ type BoolList interface {
 // ExternalBoolList supports binding a list of bool values from an external variable.
 //
 // Since: 2.0
+
```
Jacalz, line 29: This new line shouldn't be here.

data/binding/bindlists.go
```diff
@@ -219,7 +220,6 @@ type boundExternalBoolListItem struct {
 
 	old bool
 }
-
```
Jacalz, line 222: This line should be added back in.

data/binding/bindlists.go
```diff
@@ -1358,10 +1358,13 @@ type UntypedList interface {
 // ExternalUntypedList supports binding a list of interface{} values from an external variable.
 //
 // Since: 2.1
+type UntypedItemComparator func(a interface{}, b interface{}) bool
+
 type ExternalUntypedList interface {
```
Jacalz, line 1361: This is missing its comment. It seems to be the one that is above the `UntypedItemComparator`. That one should of course also have its own comment.

Jacalz (changes requested):
I have added a few example comment to indicate where formatting needs to be fixed. I was a bit short on time when I wrote my last message, so I hope that this clarifies things a bit :)

data/binding/bindlists.go
kontza: And all these should be changed in the `gen.go`, right?

Jacalz: Indeed. It was just easier to comment on the .go files as those had syntax highlighting.

kontza: Another question: as `bindlists` is a generated file, how can I control the output to add/remove newlines? I've tried adding several newlines (for testing purposes), and they all seem to be slurped by `go generate` step so that there are now extra newlines.

Jacalz: That sounds very strange. You should be able to just add and remove lines as you please. I can see the sources of the errors when I look at the generation strings.

andydotxyz: There should not be any strangeness - change the templates and re-generate. The newlines should be correct unless the template has been altered to introduce the line changes.

andydotxyz:
> With this change ExternalUntypedList gets a new function: SetComparator. This way one can provide a custom comparator to be used in the package internal function setIfChanged. Such a comparator function could be reflect.DeepEqual.

If this is for internal usage only then it shouldn't be exposed as a new public API should it?

data/binding/gen.go
```diff
@@ -388,10 +388,17 @@ type {{ .Name }}List interface {
 // External{{ .Name }}List supports binding a list of {{ .Type }} values from an external variable.
 //
 // Since: {{ .Since }}
+{{- if eq .Name "Untyped" }}
+type UntypedItemComparator func(a interface{}, b interface{}) bool
```
andydotxyz, line 392: This type definition could (should?) be outside of the generated files, as it is not created based on each type being generated

data/binding/gen.go
```diff
@@ -582,7 +605,28 @@ type boundExternal{{ .Name }}ListItem struct {
 	old {{ .Type }}
 }
 
+{{- if eq .Name "Untyped" }}
+func (b *boundExternal{{ .Name }}ListItem) setIfChanged(val {{ .Type }}, comparator UntypedItemComparator) error {
```
andydotxyz, line 609: Is this needed? Isn't the comparator internal state anyway? Instead of passing can't we just access the struct field?

data/binding/bindlists.go
andydotxyz: Ah I see - you have copied usage of "{{-" which instructs the templates to remove proceeding whitespace. When the item before is a blank line you might need "{{" instead.

andydotxyz:
@kontza we are preparing the next release, are you able to resolve the formatting issues so we can include this?

kontza:
Hi,

Sorry, this got pushed to background when the filesystem on my Mac crashed. After having rebuilt the system back to working order I almost totally forgot about this.

What's your planned release date? In a few days or further? I could return to this next week.

andydotxyz:
Thanks for the contributions, however we are now cleaning up [older PRs](https://github.com/fyne-io/fyne/wiki/Contributing#stale-pull-requests) to help us focus on new contributions.
Please re-open and update if you want to continue helping with this change.

PR #2913, opened 2022-04-11
Add user installation in HOME directory

### LINKED ISSUE #2909: Changing the Makefile template to respect freedesktop and local installation
Looking on the template file:
https://github.com/fyne-io/fyne/blob/master/cmd/fyne/internal/templates/data/Makefile

There are 2 majors problems if we want to install on Linux inside the "HOME" directory using the freedesktop standard.

Using; `make install PREFIX=~/.local`

1.  the icons is not placed in the freedesktop standard `~/.local/share/icons` directory
2. the `Exec` directive needs to be set with the absolute path to the binary

At this time, installing locally results on a desktop file which is not found in menu or activity menu, and the icon is not set.

At this time no, I need to fix the Makefile

I can propose 3 solutions as PR:

1. The first one is to respect all condition of freedesktop by using `sed` to replace the `Exec` directive with the absolute path to the binary, and place the icon file in `share/icons` instead of `share/pixmaps`, whatever if we install globally or locally
2. The second solution is to add "user-install" and "user-uninstall" directive (that can be mixed with the third solution below) which decide to install on `~/.local` and place files and make the `sed` commands
3. The second solution is to detect if the PREFIX is set to `~/.local`and make the replacement.
For example, I do this for my "nyrd" application:

```Makefile
# If PREFIX isn't provided, we check for $(DESTDIR)/usr/local and use that if it exists.
# Otherwice we fall back to using /usr.

LOCAL != test -d $(DESTDIR)/usr/local && echo -n "/local" || echo -n ""
LOCAL ?= $(shell test -d $(DESTDIR)/usr/local && echo "/local" || echo "")
PREFIX ?= /usr$(LOCAL)
HOME_INSTALL ?= $(shell [ "$(shell realpath $(PREFIX))" == "$(HOME)/.local" ] && echo "true" || echo "false")

default:
	# For simple users:
	# Run "make install PREFIX=~/.local" to install the application on your home directory.
	# Run "make uninstall PREFIX=~/.local" to uninstall the application on your home directory.
	#
	# Or globally:
	# Run "sudo make install" to install the application.
	# Run "sudo make uninstall" to uninstall the application.

install:
	install -Dm00644 usr/local/share/applications/nyrd.desktop $(DESTDIR)$(PREFIX)/share/applications/nyrd.desktop
	install -Dm00755 usr/local/bin/nyrd $(DESTDIR)$(PREFIX)/bin/nyrd
ifeq ($(HOME_INSTALL),true)
	install -Dm00644 usr/local/share/pixmaps/nyrd.png $(DESTDIR)$(PREFIX)/share/icons/nyrd.png
	sed -i 's,Exec=,Exec=$(HOME)/.local/bin/,' $(DESTDIR)$(PREFIX)/share/applications/nyrd.desktop
else
	install -Dm00644 usr/local/share/pixmaps/nyrd.png $(DESTDIR)$(PREFIX)/share/pixmaps/nyrd.png
endif 

uninstall:
	-rm $(DESTDIR)$(PREFIX)/share/applications/nyrd.desktop
	-rm $(DESTDIR)$(PREFIX)/bin/nyrd
	-rm $(DESTDIR)$(PREFIX)/share/pixmaps/nyrd.png
ifeq ($(HOME_INSTALL),true)
	-rm $(DESTDIR)$(PREFIX)/share/icons/nyrd.png
endif

# we can add this:
user-install:
	$(MAKE) install PREFIX=~/.local

user-uninstall:
	$(MAKE) uninstall PREFIX=~/.local
```

When the solution is accepted, I can of course make a pull request :)

metal3d:
### Description:

Append `user-install` and `user-uninstall` to make it possible to
install the application in the HOME directory respecting the freedesktop
recommendations.

Fixes #2909

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

There is no test for this it seems, but others tests are broken

### DIFF
--- a/cmd/fyne/internal/templates/data/Makefile
+++ b/cmd/fyne/internal/templates/data/Makefile
@@ -6,6 +6,11 @@ LOCAL ?= $(shell test -d $(DESTDIR)/usr/local && echo "/local" || echo "")
 PREFIX ?= /usr$(LOCAL)
 
 default:
+	# Install for simple user
+	# Run "make user-install" to install in ~/.local/
+	# Run "make user-uninstall" to uninstall from ~/.local/
+	#
+	# System install
 	# Run "sudo make install" to install the application.
 	# Run "sudo make uninstall" to uninstall the application.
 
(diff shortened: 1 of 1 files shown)


Jacalz:
Thanks a lot for this. PRs should target the `develop` branch, so you’ll probably need to rebase it and change the target

Jacalz:
Btw, the problem mentioned in the comment above is the reason why the CI won't run.

metal3d:
OK, I will remove this one and PR on `develop`, sorry

PR #2914, opened 2022-04-12
Implement paragraph padding for RichText widget

### LINKED ISSUE #2912: Increase paragraph padding in RichText widget

Hey, I'm trying to make use of the RichText widget to render markdown, but the padding between paragraph blocks is quite a bit smaller than most other renderers I've used.

I tried increasing my theme's padding value, this achieves the desired affect on my text, but then my other widgets look comical.

A formatting struct that could be passed to the RichText widget to configure things like line padding, word spacing, paragraph padding, and indentation size.

Or, formatting style brought inline with some popular markdown renderer (e.g. GitHub's):

image

_Left Fyne RichText widget; right GitHub gist_

alexrudd:
### Description:

~~_Opening as a draft as I'm having some issues getting the tests to run locally_~~

Introduces a new size name called `SizeNameParagraphPadding` and uses it in the rich text renderer to increase the padding between paragraphs.

Fixes #2912

image

The padding around most elements looks much better imo. The only problem is that the padding does not seem to be applied after lists, so they end up being cuddled too closely to the following element. This can be seen in the above screenshot.

Any idea why this might be?

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/theme/theme.go
+++ b/theme/theme.go
@@ -133,6 +133,12 @@ const (
 	// Since: 2.0
 	SizeNamePadding fyne.ThemeSizeName = "padding"
 
+	// SizeNameParagraphPadding is the name of theme lookup for paragraph
+	// padding size.
+	//
+	// Since: 2.2
+	SizeNameParagraphPadding fyne.ThemeSizeName = "paragraphPadding"
+
 	// SizeNameScrollBar is the name of theme lookup for the scrollbar size.
 	//
 	// Since: 2.0
@@ -319,6 +325,13 @@ func LightTheme() fyne.Theme {
 // Padding is the standard gap between elements and the border around interface elements.
 func Padding() float32 {
 	return current().Size(SizeNamePadding)
+}
+
+// ParagraphPadding is the standard gap between paragraphs with rich text.
+//
+// Since: 2.2
+func ParagraphPadding() float32 {
+	return current().Size(SizeNameParagraphPadding)
 }
 
 // PlaceHolderColor returns the theme's standard text color.
@@ -586,6 +599,8 @@ func (t *builtinTheme) Size(s fyne.ThemeSizeName) float32 {
 		return 20
 	case SizeNamePadding:
 		return 4
+	case SizeNameParagraphPadding:
+		return 10
 	case SizeNameScrollBar:
 		return 16
 	case SizeNameScrollBarSmall:
--- a/cmd/fyne_demo/tutorials/theme.go
+++ b/cmd/fyne_demo/tutorials/theme.go
@@ -46,6 +46,8 @@ func (customTheme) Size(s fyne.ThemeSizeName) float32 {
 	switch s {
 	case theme.SizeNamePadding:
 		return 8
+	case theme.SizeNameParagraphPadding:
+		return 20
 	case theme.SizeNameInlineIcon:
 		return 20
 	case theme.SizeNameScrollBar:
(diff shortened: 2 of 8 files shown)


theme/theme.go
```diff
@@ -321,6 +327,11 @@ func Padding() float32 {
 	return current().Size(SizeNamePadding)
 }
 
+// ParagraphPadding is the standard gap between paragraphs with rich text.
+func ParagraphPadding() float32 {
+	return current().Size(SizeNameParagraphPadding)
+}
+
```
alexrudd, line 334: This helper probably isn't needed as Paragraph Padding is only used in two places.

andydotxyz: These are also a public helper for people writing their own widgets... so maybe it is useful...
It should, however have a Since line.

theme/theme.go
```diff
@@ -133,6 +133,12 @@ const (
 	// Since: 2.0
 	SizeNamePadding fyne.ThemeSizeName = "padding"
 
+	// SizeNameParagraphPadding is the name of theme lookup for paragraph
+	// padding size.
+	//
+	// Since: 2.0
```
andydotxyz, line 139: This should be 2.2 not 2.0 :)

andydotxyz (changes requested):
This looks like a pretty good approach, a few comments inline.
It might be worth adding a test case somewhere so we verify the behaviour in unit tests.

alexrudd:
I've been using GitHub's markdown renderer as a reference, and this is as close as I've gotten:

image

I'm struggling to get the renderer to treat the List as a paragraph segment while the individual List Items as inline segments.

I'm also thinking that it might be good to have a separate padding value for lines of text rather than using the general `SizeNamePadding`. This would allow users to control how rich text renders without affecting any other widgets.

Jacalz:
Please don’t force-push your changes. It often destroys the review comments and makes it hard for us to see what has changed between commits. We’ll squash on merge if there are too many and unrelated commits. It’s better to just push normally with more commits as you go.

andydotxyz:
> I'm also thinking that it might be good to have a separate padding value for lines of text rather than using the general `SizeNamePadding`. This would allow users to control how rich text renders without affecting any other widgets.

I see your point but I'm not sure - line spacing is normally expressed as a multiple rather than a value isn't it? So what we are currently padding is "1x" and a future value could make it more (1.5) or less (0.8).
However it is possible that line spacing in general should be separate from padding - but i think that is a much bigger change than just in the rich-text widget... though I could be wrong.

andydotxyz:
> I'm struggling to get the renderer to treat the List as a paragraph segment while the individual List Items as inline segments.

Yes, this is a challenge. I think that relates to items being either inline or not, and list items are not inline (as they each need a new line) so we may have to look at inline / not-inline / paragraph as three possibilities instead. Or, and possibly easier, the paragraph wrap is only added around the text paragraph style?

andydotxyz:
@alexrudd did you manage to solve the list/paragraph issue? It would be good to get this merged.

alexrudd:
Hey, sorry for the slow responses. I had a few attempts but could never get this working satisfactorily. I think possibly a more wholistic approach to this problem might be needed. Happy for this PR to be closed.

andydotxyz:
This should be revisited on latest `develop` as we have changed the definition of padding somewhat and introduced line spacing as well.
Perhaps we should close this one and open a new, unless you think it can be adapted @alexrudd ?

alexrudd:
Think this should be closed :)

PR #2915, opened 2022-04-12
Add user installation in HOME directory

### LINKED ISSUE #2909: Changing the Makefile template to respect freedesktop and local installation

metal3d:
### Description:
Following the freedesktop recommendation and rules, installing in HOME on Linux needs to:

- place the icon in `~/.local/share/icons`
- set `Exec` to the absolute path to the binary

The current Makefile doesn't do the complete modification even using `PREFIX=~/.local`.

So, this changes add `user-install` and `user-uninstall` directives to make the installation working for end-user without the need of `sudo`

Fixes #2909 

### Checklist:

There are no Go file impacted

- [ ] Tests included.
- [ ] Tests all pass.


cmd/fyne/internal/templates/data/Makefile
```diff
@@ -6,6 +6,11 @@ LOCAL ?= $(shell test -d $(DESTDIR)/usr/local && echo "/local" || echo "")
 PREFIX ?= /usr$(LOCAL)
 
 default:
+	# Install for simple user
```
Jacalz, line 9: This seems wage and also like it could almost sound a bit rude. What is a "simple user"?
How about something like this instead (feel free to suggest something else if you like):
```suggestion
	# User install
```

Jacalz (changes requested):
Thanks for this. I had a comment inline about naming but there is also the issue that this isn't quite working. You need to run `go generate` inside `cmd/fyne/internal/templates` to make the changes in the template take effect.

I patched this up to work but I get this error when I try to `user-install` fyne_demo (btw, you need to change the FyneApp.toml name to Fyne_Demo because the Makefile doesn't seem to handle spaces in the name, but that's an issue for another time).
```
install -Dm00644 usr/local/share/applications/Fyne_Demo.desktop /home/jacob/.local/share/applications/Fyne_Demo.desktop
install -Dm00755 usr/local/bin/fyne_demo /home/jacob/.local/bin/fyne_demo
install -Dm00644 usr/local/share/pixmaps/Fyne_Demo.png /home/jacob/.local/share/icons/Fyne_Demo.png
sed -i -e "s/Exec=.*/Exec=/home/jacob/.local/bin/fyne_demo/g" /home/jacob/.local/share/applications/Fyne_Demo.desktop
sed: -e expression #1, char 17: unknown option to `s'
make: *** [Makefile:30: user-install] Error 1
```

Jacalz:
@metal3d Have you had any chance to look at continuing this? It would be great if we could get it landed for v2.2.0 :)

metal3d:
I will take a look today, sorry for the delay

Jacalz:
No worries. Just wanted to check that you hadn’t forgot it :)

metal3d:
> Thanks for this. I had a comment inline about naming but there is also the issue that this isn't quite working. You need to run `go generate` inside `cmd/fyne/internal/templates` to make the changes in the template take effect.

oops sorry

> Btw, you need to change the name in `FyneApp.toml` to Fyne_Demo because the Makefile doesn't seem to handle spaces in the name, but that's an issue for another time. The correct fix for that is to switch to using the app-id for the icons and desktop file, but that was an issue before this PR as well. I will send in a separate PR to fix that.

I propose to simply quote the names, I applied a simple `.Var | printf "%q"` that works here on my computer.

cmd/fyne/internal/templates/data/Makefile
```diff
@@ -5,15 +5,35 @@ LOCAL != test -d $(DESTDIR)/usr/local && echo -n "/local" || echo -n ""
...
+Icon := {{ .Icon | printf "%q" }}
+
 default:
+	# User install
+	# Run "make user-install" to install in ~/.local/
+	# Run "make user-uninstall" to uninstall from ~/.local/
+	#
+	# System install
 	# Run "sudo make install" to install the application.
 	# Run "sudo make uninstall" to uninstall the application.
```
Jacalz, line 19: Should these perhaps be updated to say say something about it not being in the user folder? I can't think of any better wording myself. Any ideas?

Jacalz (changes requested):
Thanks for the update. I had a little thing that I though of about the wording of one of the old comment. There is also an issue with the `sed` call when I try to install on my Fedora 35 laptop:
```
install -Dm00644 usr/local/share/applications/"Fyne Demo".desktop /home/jacob/.local/share/applications/"Fyne Demo".desktop
install -Dm00755 usr/local/bin/"fyne_demo" /home/jacob/.local/bin/"fyne_demo"
install -Dm00644 usr/local/share/pixmaps/"Fyne Demo.png" /home/jacob/.local/share/icons/"Fyne Demo.png"
sed -i -e "s/Exec=.*/Exec=/home/jacob/.local/bin/"fyne_demo"/g" /home/jacob/.local/share/applications/"Fyne Demo".desktop
sed: -e expression #1, char 17: unknown option to `s'
make: *** [Makefile:34: user-install] Error 1
```

cmd/fyne/internal/templates/data/Makefile
metal3d: I changed this to use coma instead of slashes on https://github.com/fyne-io/fyne/pull/2915/commits/4e8eff78d743aed32ff4d970bcbe0395bea735ad - have I missed something ?

metal3d: Fixed again in https://github.com/fyne-io/fyne/pull/2915/commits/8824eefb90e687a204494c8420e2f2d11a6a2609

metal3d: For the wording, maybe:

```
# Run "sudo make install" to install the application for all users.
# Run "sudo make uninstall" to uninstall the application for all users.
```

That's a bit what proposes Windows for some softwares, you can install for the current user only or for all the users.

The problem here is that if a user installed "locally", and then it uses `sudo` to remove the package, it will believe that that will remove the package...

Jacalz: Great. Thanks :)

Jacalz (approved):
This works great. Thanks a lot for the contribution :+1:

Issue #2899, opened 2022-04-08
Fyne CLI cannot detect %ANDROID_NDK_HOME% from fyne package


phathuynh5519:
### Describe the bug:

Folks, currently the fyne CLI cannot detect the environmental variable %ANDROID_NDK_HOME%, and fyne package command uses the default ndk-bundle from %ANDROID_HOME%/%ANDROID_SDK_HOME% location
```
D:\Demos\examples>refreshenv
Refreshing environment variables from registry for cmd.exe. Please wait...Finished..

D:\Demos\examples>fyne package -os android/arm64 -appID com.helloworld.fyne -icon Icon.png
no compiler for arm64 was found in the NDK (tried D:\Android\Sdk\ndk-bundle\toolchains\llvm\prebuilt\windows-x86_64\bin\aarch64-linux-android21-clang). Make sure your NDK version is >= r19c. Use `sdkmanager --update` to update it

D:\Demos\examples>echo %ANDROID_NDK_HOME%
D:\Android\Sdk\ndk\21.4.7075529

D:\Demos\examples>fyne package -os android/arm64 -appID com.helloworld.fyne -icon Icon.png
no compiler for amd64 was found in the NDK (tried D:\Android\Sdk\ndk-bundle\toolchains\llvm\prebuilt\windows-x86_64\bin\x86_64-linux-android21-clang). Make sure your NDK version is >= r19c. Use `sdkmanager --update` to update it
```

### To Reproduce:
Steps to reproduce the behaviour:
```
D:\Demos\examples>refreshenv
Refreshing environment variables from registry for cmd.exe. Please wait...Finished..

D:\Demos\examples>fyne package -os android/arm64 -appID com.helloworld.fyne -icon Icon.png
no compiler for arm64 was found in the NDK (tried D:\Android\Sdk\ndk-bundle\toolchains\llvm\prebuilt\windows-x86_64\bin\aarch64-linux-android21-clang). Make sure your NDK version is >= r19c. Use `sdkmanager --update` to update it

D:\Demos\examples>echo %ANDROID_NDK_HOME%
D:\Android\Sdk\ndk\21.4.7075529

D:\Demos\examples>fyne package -os android/arm64 -appID com.helloworld.fyne -icon Icon.png
[...]

PR #2919, opened 2022-04-13
Add flags to package cmd

lkebin:
### Description:
Add more flags to package command

- add `androidAPI` flag to set the Android API version
- add `iosVersion` flag to set the minimal version of the iOS SDK

PR #2922, opened 2022-04-14
Expose more gomobile flags to package command

lkebin:
Replace #2919

### DIFF
--- a/cmd/fyne/internal/commands/package-mobile.go
+++ b/cmd/fyne/internal/commands/package-mobile.go
@@ -16,11 +16,31 @@ import (
 )
 
 func (p *Packager) packageAndroid(arch string) error {
-	return mobile.RunNewBuild(arch, p.appID, p.icon, p.name, p.appVersion, p.appBuild, p.release, "", "")
+	return mobile.RunNewBuild(
+		mobile.WithTarget(arch),
+		mobile.WithBundleID(p.appID),
+		mobile.WithIcon(p.icon),
+		mobile.WithName(p.name),
+		mobile.WithVersion(p.appVersion),
+		mobile.WithBuild(p.appBuild),
+		mobile.WithRelease(p.release),
+		mobile.WithAndroidAPI(p.androidAPI),
+	)
 }
 
 func (p *Packager) packageIOS(target string) error {
-	err := mobile.RunNewBuild(target, p.appID, p.icon, p.name, p.appVersion, p.appBuild, p.release, p.certificate, p.profile)
+	err := mobile.RunNewBuild(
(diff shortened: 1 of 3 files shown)


lkebin:
Expose gomobile flags could resolve #2899

andydotxyz:
It it your intention that the minimum does not change - if so that probably helps with compatibility. However:

> However, if the API level of the platform is higher than the version declared by your app's targetSdkVersion, the system may enable compatibility behaviors to ensure that your app continues to work the way you expect.

(from https://developer.android.com/guide/topics/manifest/uses-sdk-element)
If the developer raises the target SDK higher than Fyne expects, then compatibility layers will be removed and features will stop working. If your aim is to fix the compile error then I don't think that this is the correct fix, as it opens up high risk of breakage at runtime.

Which compiler versions and device combinations have you tested against?

lkebin:
> It it your intention that the minimum does not change - if so that probably helps with compatibility. However:
> 
> > However, if the API level of the platform is higher than the version declared by your app's targetSdkVersion, the system may enable compatibility behaviors to ensure that your app continues to work the way you expect.
> 
> (from https://developer.android.com/guide/topics/manifest/uses-sdk-element) If the developer raises the target SDK higher than Fyne expects, then compatibility layers will be removed and features will stop working. If your aim is to fix the compile error then I don't think that this is the correct fix, as it opens up high risk of breakage at runtime (because the Android code is written by Fyne, against version 15, not by the developer using the target they have installed).
> 
> Which compiler versions and device combinations have you tested against?

Understood, I just want to fix the compiler error, not really want to setup targetSdkVersion of the app.

Let's close the PR at the moment.

lkebin:
> Ah, so you probably want the `compileSdkVersion` I guess? I wonder if we can fix the error in this way without exposing more flags. So target remains low and our code does not break - but we can discover what SDK is available and use it at compile time?

Actually, I want to fix this error:

```
no compiler for arm was found in the NDK (tried /Users/kb/Workspace/sdk/android/ndk-bundle/toolchains/llvm/prebuilt/darwin-x86_64/bin/armv7a-linux-androideabi16-clang). Make sure your NDK version is >= r19c. Use `sdkmanager --update` to update it
```

The latest NDK version doesn't have `armv7a-linux-androideabi16-clang` anymore.

andydotxyz:
Ok, that’s a great idea. If we have to update the minSDK then we must ensure all our code works at that minimum and update accordingly. If it is possible to use just a new compile version without impacting min we can keep compatibility and just find the lowest installed compiler to build. 
These possible solutions both avoid allowing developers to break Fyne behaviour accidentally.

PR #2927, opened 2022-04-16
Handle shortcut without modifier

scastiel:
### Description:

Fixes #(issue)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -865,7 +865,7 @@ func (w *window) triggersShortcut(localizedKeyName fyne.KeyName, key fyne.KeyNam
 		}
 	}
 
-	if shortcut == nil && modifier != 0 && !isKeyModifier(keyName) && modifier != fyne.KeyModifierShift {
+	if shortcut == nil && !isKeyModifier(keyName) && modifier != fyne.KeyModifierShift {
 		shortcut = &desktop.CustomShortcut{
 			KeyName:  keyName,
 			Modifier: modifier,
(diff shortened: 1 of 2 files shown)


andydotxyz:
Is ‘F’ really a shortcut? How does this differ to ‘OnTypedRune’?

scastiel:
@andydotxyz 

> Is ‘F’ really a shortcut? How does this differ to ‘OnTypedRune’?

Thank you! I suspected there was a better way to intercept key events 😉! (`canvas.SetOnTypedKey` is actually what I have been looking for.)

pascallohrer:
> Is ‘F’ really a shortcut? How does this differ to ‘OnTypedRune’?

While 'F' might not be your typical shortcut, I'd argue 'F2' is. It differs from 'OnTypedRune' (or in this case, 'OnTypedKey') in that registering a shortcut on the canvas always executes the registered action, while canvas.OnTypedKey is not executed while a Focusable is focused. Also, OnTypedKey is one singular function to which adding "shortcuts" is not possible; in order to achieve a similar API to 'canvas.AddShortcut', I must build my own "shortcut" handler with its own data holder and AddShortcut function, then set that as canvas.OnTypedKey.

Unless I'm doing something wrong, using single-key shortcuts with F-keys like F2 requires setting up such a custom key handler and then extending every single focusable widget I'm using so it calls this custom key handler before falling back to the base widget's TypedKey.

Am I doing something wrong?

andydotxyz:
In the current event setup it is not possible to have shortcuts overlap with key events - there is no "was there a matching shortcut" in the API we have.

So for now it is not possible to support "F1" as a shortcut, only as a key. You will indeed need to look for the key and run your shortcut function.

At some future breaking release we could consider requiring shortcut handlers to return whether or not it was executed/consumed.

PR #2942, opened 2022-04-28
Implement NewStringWithFormat method

### LINKED ISSUE #2890: Add binding.NewStringWithFormat
I.e. in my game, I have a word that the user needs to do something with, so I would want to show them something like
"Generated word: $word", instead of just $word

Shouldn't be very hard to implement since we have the same thing for int and floats

binding.NewStringWithFormat, working similar to binding.NewIntWithFormat

chrisbirster:
### Description:
Adds a new binding.NewStringWithFormat which is a wrapper over existing binding.NewSprintf. 

Fixes #2890 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/data/binding/sprintf.go
+++ b/data/binding/sprintf.go
@@ -32,6 +32,13 @@ func NewSprintf(format string, b ...DataItem) (String, error) {
 	return ret, nil
 }
 
+// NewStringWithFormat is a wrapper over NewSprintf a String binding that format
+// its content using the format string and the provide additional parameter that
+// must be other data bindings.
+func NewStringWithFormat(format string, b ...DataItem) (String, error) {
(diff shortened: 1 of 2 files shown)


andydotxyz (changes requested):
This isn't really consistent with the other `WithFormat` variants.
I had thought the intent was to return a string that simply had a format applied. i.e. no extra parameters but it contained a single %s to format something around the string content...

andydotxyz:
To be completely in-keeping I guess it would be `StringToStringWithFormat(source, "My string %s")` but the `NewStringWithFormat("My string %s")` could be a shorthand...

andydotxyz:
Continued in #3027 to try and land for v2.2.0

PR #2943, opened 2022-04-28
Bugs/gopherjs fix

Bluebugs:
### Description:
gopherjs support got broken when adding metadata support. This is not adding metadata support for gopherjs as that seems to introduce complexity at this point, but instead just get the build/package/serve passing again. Follow up issue will be created to address this issue.

### Checklist:
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/.github/workflows/platform_tests.yml
+++ b/.github/workflows/platform_tests.yml
@@ -54,9 +54,9 @@ jobs:
       run: go build
       if: ${{ matrix.go-version == '1.17' }}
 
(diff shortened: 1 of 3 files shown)


Bluebugs:
The issue to track adding metadata support to gopherjs: #2944

andydotxyz (approved):
Thanks for this.
We probably need a plan of how / what to do for JS builds that might want the metadata API.

PR #2952, opened 2022-05-01
Optimize the goroutineID function

Jacalz:
### Description:

It is not a lot faster but it does allocate about 60% less memory.

```
BenchmarkGoroutineIDnew-8   	  385348	      2847 ns/op	      66 B/op	       2 allocs/op
BenchmarkGoroutineIDold-8   	  400776	      2960 ns/op	     176 B/op	       3 allocs/op
```

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/driver_desktop.go
+++ b/internal/driver/glfw/driver_desktop.go
@@ -4,9 +4,9 @@
 package glfw
 
 import (
+	"bytes"
 	"runtime"
 	"strconv"
-	"strings"
 
 	"fyne.io/systray"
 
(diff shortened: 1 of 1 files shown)


MatejMagat305:
I m sorry for this notice:
if performance (speed) is important, try this ( in one project I was solving similar problem):
```
func goroutineIDMy() int {
	b, id := make([]byte, 64), 0
	b = b[:runtime.Stack(b, false)]
	b = b[bytes.IndexByte(b, ' ')+1:]
	b = b[:bytes.IndexByte(b, ' ')]
	for _, ch := range b {
		ch -= '0'
		id = id*10 + int(ch)
	}
	return n
}
```
it is a litle bit faster (average 100 ns/op - it is not much but maybe .......)
```
BenchmarkGoroutine-32       	  505766	      2376 ns/op
BenchmarkGoroutineMy-32       528980	      2144 ns/op
```

changkun:
Still a few ns faster: https://github.com/changkun/goid/blob/master/goid.go compare to whats posted here :)

MatejMagat305:
> Still a few ns faster: https://github.com/changkun/goid/blob/master/goid.go compare to whats posted here :)

I give up :) .........

Jacalz:
Thanks @MatejMagat305 and @changkun for your interesting ideas. I actually hadn’t thought about that parsing being possible but it makes a lot of sense now that I see it.

@changkun Would you like to open your own PR for that? I don’t feel like I can take credit for that wonderfully solution :)

changkun:
Sure :) See #2953.

Jacalz:
> Sure :) See #2953.

Wonderful. Thanks a lot :)

PR #2953, opened 2022-05-01
internal/driver/glfw: correct handle goroutineID

changkun:
### Description:

The current goroutine ID is handled with two issues:

1. main goroutine is assumed to be 1, but it may not be true;
2. goroutine ID is typed as int rather than uin64; this may cause overflow.

Let's handle them gracefully.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/driver.go
+++ b/internal/driver/glfw/driver.go
@@ -21,7 +21,11 @@ import (
 	"fyne.io/fyne/v2/storage/repository"
 )
 
-const mainGoroutineID = 1
+// mainGoroutineID stores the main goroutine ID.
+// This ID must be initialized in main.init because
+// a main goroutine may not equal to 1 due to the
+// influence of a garbage collector.
+var mainGoroutineID uint64
 
 var (
 	curWindow *window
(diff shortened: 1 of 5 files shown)


Jacalz:
Wonderful. I’ll have a look at this tomorrow and also see about benchmarking it because I’m curious :)

changkun:
Benchmark is not much interesting, though. The implementation is bounded by the `runtime.Stack` performance. Optimizing conversions are just a very few ns faster. I think it would be better for us to fix the current misuse and potential danger in the implementation:

```
name                      old time/op  new time/op  delta
GoroutineID/sequential-8  1.88µs ± 0%  1.78µs ± 0%  -5.38%  (p=0.000 n=4+5)
GoroutineID/concurrent-8  2.66µs ± 2%  2.53µs ± 2%  -5.07%  (p=0.008 n=5+5)
```

benchmark: https://go.dev/play/p/ZuN6iZMj2RT

Bluebugs (approved):

Jacalz:
> Optimizing conversions are just a very few ns faster. I think it would be better for us to fix the current misuse and potential danger in the implementation:

Absolutely. I just happen to be curious :)

andydotxyz (approved):
Clever to use `init` like this, thanks

PR #2956, opened 2022-05-02
android onLowMemory

MatejMagat305:
### Description:
This is only 2 lines of code - this is trivial change, clean memory at Android OS request ...

### DIFF
--- a/internal/driver/mobile/app/android.go
+++ b/internal/driver/mobile/app/android.go
@@ -57,6 +57,7 @@ import (
 	"log"
 	"mime"
 	"os"
+	"runtime/debug"
 	"strings"
 	"time"
 	"unsafe"
@@ -268,6 +269,7 @@ func onConfigurationChanged(activity *C.ANativeActivity) {
 
 //export onLowMemory
 func onLowMemory(activity *C.ANativeActivity) {
+	debug.FreeOSMemory()
 }
 
 var (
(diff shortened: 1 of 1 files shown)


changkun (approved):
Looks like a reasonable change to me. (why it was not there before? the API was introduced in go1.1…)

internal/driver/mobile/app/android.go
```diff
@@ -268,6 +269,7 @@ func onConfigurationChanged(activity *C.ANativeActivity) {
 
 //export onLowMemory
 func onLowMemory(activity *C.ANativeActivity) {
+	debug.FreeOSMemory()
```
changkun, line 272: I just recall that we could also call runtime.GC() before manually free OS memories so this is more effective:

```suggestion
	runtime.GC()
	debug.FreeOSMemory()
```

internal/driver/mobile/app/android.go
```diff
@@ -57,6 +57,7 @@ import (
 	"log"
 	"mime"
 	"os"
+	"runtime/debug"
```
changkun, line 60:
```suggestion
	"runtime"
	"runtime/debug"
```

changkun (changes requested):

MatejMagat305:
@changkun ok you aare bigger expert than me - I was thinking the freeOSMemory call gc exciplit ...

changkun (approved):
Thanks! There are two types of collectors in the runtime, one is the normal GC, and the other is called scavenger. I think debug.FreeOSMemory is only effective to the scavenger, which returns runtime managed memory to OS.

PR #2967, opened 2022-05-08
Various minor validation improvements

Jacalz:
### Description:

This rips out some of the changes in https://github.com/fyne-io/fyne/pull/2765 that I feel doesn't belong with the other changes in the PR. The changes are as follows (separated into one commit each):
- Avoid recreating the regexp error each time we get an invalid state. This should decrease the allocations and be faster over time when you write in an entry.
- We now use the `errors.Is` function that was added in Go 1.13. We can thus remove the hacky error checking and use something a bit more robust. This technically also allows wrapped errors to be correctly handled (if anyone even uses those for validation).
- Allow the onValidationChanged function to be removed in the entry. If we can remove the validator, we should be able to remove the function that runs when validation changes.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/validation/regexp.go
+++ b/data/validation/regexp.go
@@ -19,9 +19,10 @@ func NewRegexp(regexpstr, reason string) fyne.StringValidator {
 		return nil
 	}
 
+	err = errors.New(reason)
 	return func(text string) error {
 		if expression != nil && !expression.MatchString(text) {
-			return errors.New(reason)
+			return err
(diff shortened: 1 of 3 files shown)


andydotxyz (approved):
Thanks, sorry for the delay

Jacalz:
No worries. Thanks for the review :)

PR #2970, opened 2022-05-09
use release for android

MatejMagat305:
Hi, again me..............
### Description:
I sugess trivial changes to use "release" in android (it do not use it until now) - thrink binary, I hope (believe) it is relevant changes ......... 

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne/internal/mobile/build_androidapp.go
+++ b/cmd/fyne/internal/mobile/build_androidapp.go
@@ -76,19 +76,16 @@ func goAndroidBuild(pkg *packages.Package, bundleID string, androidArchs []strin
 		if err := mkdir(filepath.Dir(libAbsPath)); err != nil {
 			return nil, err
 		}
-		arguments := []string{
-			"-buildmode=c-shared",
-			"-o", libAbsPath,
-		}
-		if release && buildLdflags == "" { 
+		if release && buildLdflags == "" {
 			// if user want to release version the debug and DWARF is not nessery (useless), 
 			// but if user define ldflags, it will not duplicate (rewrite) 
-			arguments = append(arguments, "-ldflags", "-w -s")
+			buildLdflags = "-w -s"
 		}
 		err = goBuild(
 			pkg.PkgPath,
 			androidEnv[arch],
-			arguments...,
+			"-buildmode=c-shared",
+			"-o", libAbsPath,
 		)
 		if err != nil {
 			return nil, err
--- a/cmd/fyne/internal/mobile/build_androidapp.go
+++ b/cmd/fyne/internal/mobile/build_androidapp.go
@@ -76,11 +76,17 @@ func goAndroidBuild(pkg *packages.Package, bundleID string, androidArchs []strin
 		if err := mkdir(filepath.Dir(libAbsPath)); err != nil {
 			return nil, err
 		}
+		arguments := []string{
+			"-buildmode=c-shared",
+			"-o", libAbsPath,
+		}
+		if release && buildLdflags == "" {
+			arguments = append(arguments, "-ldflags", "-w -s")
+		}
 		err = goBuild(
 			pkg.PkgPath,
 			androidEnv[arch],
-			"-buildmode=c-shared",
-			"-o", libAbsPath,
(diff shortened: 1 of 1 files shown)


cmd/fyne/internal/mobile/build_androidapp.go
```diff
@@ -76,11 +76,17 @@ func goAndroidBuild(pkg *packages.Package, bundleID string, androidArchs []strin
 		if err := mkdir(filepath.Dir(libAbsPath)); err != nil {
 			return nil, err
 		}
+		arguments := []string{
+			"-buildmode=c-shared",
+			"-o", libAbsPath,
+		}
+		if release && buildLdflags == "" {
+			arguments = append(arguments, "-ldflags", "-w -s")
```
changkun, line 84: Could you also add a comment above this line and link back to a reference on the release build for android? Basically, explaining why `-ldflags -w -s` in this case?

MatejMagat305: is explaining: <br>
"the other OS (for example windows https://github.com/fyne-io/fyne/blob/7229e889d49c81a83b0b7e09400837f67f6ddad5/cmd/fyne/internal/commands/build.go#L32) has it, we do not need (on release version ) elf and debuh informations, ang we can reduce binary size of apk to 9.5 MB with one arch and aprox. 37.8 MB with all android version"
enough?

cmd/fyne/internal/mobile/build_androidapp.go
```diff
@@ -76,10 +76,12 @@ func goAndroidBuild(pkg *packages.Package, bundleID string, androidArchs []strin
 		if err := mkdir(filepath.Dir(libAbsPath)); err != nil {
 			return nil, err
 		}
+		/*
```
andydotxyz, line 79: This isn't how we format comments - just a line starting "//" uses much less space.
Is it possible to use something that is clearer English, perhaps:
"If building release and no ldflags are set then remove the useless debug and DWARF build options"

cmd/fyne/internal/mobile/build_androidapp.go
```diff
@@ -76,6 +76,10 @@ func goAndroidBuild(pkg *packages.Package, bundleID string, androidArchs []strin
 		if err := mkdir(filepath.Dir(libAbsPath)); err != nil {
 			return nil, err
 		}
+		// If building release and no ldflags are set then remove the useless debug and DWARF build options
+		if release && buildLdflags == "" {
+			buildLdflags = "-s -w"
```
changkun, line 81: I am getting more confused by the current change now. How does this change affect the `goBuild` at line 83? It seems this change is not affecting anything.

Moreover, I am still skeptical about whether we should heuristically assume that building a release does not need debug and DWARF info or not. From the gomobile's original command design, ldflags could be customized as bypass to the underlying go command, whereas fyne in this case, wraps the entire gomobile command and does not have this ability.

Instead of preassuming a `-s -w` for release, we might better to figure out a way to allow users to customize this themselves, such as to allow "-s -w" in `fyne build` (similarly, `-x` etc.) What do you think? @andydotxyz

MatejMagat305: well, there are few reason: <br>
1. in ordinary OS(windows, linux, mac) it look like
ordinary OS
2. I do not want to be conflict with 
IOS build
3. it has effect on 
LDFLAGS

changkun: Ah. I see. Thanks for the clarification!

andydotxyz: I think that customising ldflags is not something we need to be exporting until there is a compelling reason to.
Anything that ties us to implementation details through public API makes me nervous.
In general releases are intended to ship to stores so we keep them lean.
We may need to consider the option of generating DSYM or equivalent to suppose debugging a stripped binary in a future version.

andydotxyz (approved):
I think this looks good to me now - @changkun are you happy too?

PR #2976, opened 2022-05-14
Use Windows MessageBox API for error when GLFW can't create a Window.

### LINKED ISSUE #2689: Better error reporting / handling when OpenGL is not available
When a Fyne app crashes because it could not create a window this is normally due to a driver issues.
The OS or other dependency is not working.
We should report this better, so developers know what to do or where to report it.
Or to update their drivers etc?

The panics can get pasted here, but we have to interpret the data and often the developer could have solved it themselves.

Catch the panic and log the error with suggested fix/next steps and cleanly exit

Bluebugs:
### Description:
This use Windows native API to create a MessageBox that display an error for all problem related to the GL driver. Specifically useful when the OpenGL driver is not available.

Fixes #2689

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/driver.go
+++ b/internal/driver/glfw/driver.go
@@ -141,7 +141,7 @@ func (d *gLDriver) windowList() []fyne.Window {
 }
 
 func (d *gLDriver) initFailed(msg string, err error) {
-	fyne.LogError(msg, err)
+	logError(msg, err)
 
 	run.Lock()
 	if !run.flag {
--- a/internal/driver/glfw/driver_notwindows.go
+++ b/internal/driver/glfw/driver_notwindows.go
@@ -1,0 +1,10 @@
+//go:build !windows
+// +build !windows
+
+package glfw
+
+import "fyne.io/fyne/v2"
+
+func logError(msg string, err error) {
+	fyne.LogError(msg, err)
+}
--- a/internal/driver/glfw/driver_windows.go
+++ b/internal/driver/glfw/driver_windows.go
@@ -1,0 +1,50 @@
+package glfw
+
+import (
+	"fmt"
+	"runtime"
+	"syscall"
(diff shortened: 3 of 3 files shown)


toaster:
This PR let me think about error reporting in general and I’ve the following questions:

1. Shouldn’t GL errors always be fatal because a Fyne app without graphics output is fairly useless?
2. Shouldn’t we present errors on all platforms consistently (this does not only apply to GL errors), i.e. with some kind of native message box if the user is unlikely to see the log?

andydotxyz (approved):
Looks good to me thanks

Bluebugs:
>     1. Shouldn’t GL errors always be fatal because a Fyne app without graphics output is fairly useless?

In the code path that trigger this MessageBox, it is indeed fatal and the application will quit after the MessageBox is acknowledge by the user (driver.go:initFailed#149 and driver.go:initFailed#152). I would think if there is other path that should abort, they should call initFailed too.
 
>     2. Shouldn’t we present errors on all platforms consistently (this does not only apply to GL errors), i.e. with some kind of native message box if the user is unlikely to see the log?

That would be nice. It might be possible on macOS, but it won't on Linux/BSD which doesn't have a native message box API (Arguably on those OS, people are more used to the command line, so maybe less of an issue). Maybe opening another issue to look into how to do things on macOS would make sense?

toaster:
> That would be nice. It might be possible on macOS, but it won't on Linux/BSD which doesn't have a native message box API (Arguably on those OS, people are more used to the command line, so maybe less of an issue). Maybe opening another issue to look into how to do things on macOS would make sense?

I think, we could even try to make this work on Linux. At least we might try `notify-send` or `xmessage` or alike in addition to the command line error.

I’ll open an issue.

PR #2984, opened 2022-05-17
Adding .aab support for Android upload to google play.

### LINKED ISSUE #2663: Ability to generate Android Application Bundle - .aab

Since August 2021 the Play Store requires new applications to be delivered as .aab. However the fyne only compiles to .apk.

It is not possible to output .aab with fyne right now.

Fyne should be able to compile to .aab format.

andydotxyz:
Not perfect as it requires the Google `bundletool` installed (not part of Android SDK). But it seems better than nothing and we can improve over time.
Fixes #2663

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/install.go
+++ b/cmd/fyne/internal/commands/install.go
@@ -172,7 +172,7 @@ func (i *Installer) install() error {
 }
 
 func (i *Installer) installAndroid() error {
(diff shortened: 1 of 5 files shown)


andydotxyz:
I still need to work out how to use the non-Android-SDK `bundletool`...

PR #2988, opened 2022-05-18
Implement a nice splash screen for the web driver.

### LINKED ISSUE #2756: Improve initial web page for the web driver

Bluebugs:
### Description:
This nicely improve the splash screen when loading a Fyne app on the web. It use the application logo, its name and version to build this splash screen.

Fixes #2756

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
Looks like this broke the web package tests...

cmd/fyne/internal/commands/package.go
```diff
@@ -338,8 +338,16 @@ func (p *Packager) validate() error {
 		_, _ = fmt.Fprint(os.Stderr, "Parameter -executable is ignored for mobile builds.\n")
 	}
 
-	if p.name == "" || p.os == "wasm" || p.os == "gopherjs" || p.os == "web" {
-		p.name = exeName
+	if p.os == "wasm" || p.os == "gopherjs" || p.os == "web" {
+		if p.name == "" {
```
andydotxyz, line 342: I think this logic has broken the case where there is no name set, but the OS is not one of the web targets...
It should default to the exeName as it used to.

andydotxyz (changes requested):
The addition looks great, but I think a logic re-order has broken a condition for non-web code.

cmd/fyne/internal/commands/package.go
```diff
@@ -376,10 +376,6 @@ func calculateExeName(sourceDir, os string) string {
 
 	if os == "windows" {
 		exeName = exeName + ".exe"
-	} else if os == "wasm" {
```
andydotxyz, line 379: is it OK that this code is removed?

Bluebugs: Yes, I have moved the logic to add the extension in package-web.go where it is actually making more sense as it match the case where we are generating two executable and need to call build appropriately.

andydotxyz (approved):
Cool, thanks

PR #2990, opened 2022-05-19
Make 'web' the default output for `fyne serve` with added messaging t…

Bluebugs:
### Description:
This switch `fyne serve` to use the `web` output, instead of just wasm. As this require gopherjs to be installed, this also add an explicit warning for the user to install gopherjs if not available.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/build.go
+++ b/cmd/fyne/internal/commands/build.go
@@ -202,6 +202,12 @@ func (b *Builder) build() error {
 		versionConstraint = version.NewConstrainGroupFromString(">=1.17")
 		env = append(env, "GOARCH=wasm")
 		env = append(env, "GOOS=js")
+	} else if goos == "gopherjs" {
+		_, err := b.runner.runOutput("version")
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Can not execute `gopherjs version`. Please do `go install github.com/gopherjs/gopherjs@latest`.\n")
+			return err
+		}
 	}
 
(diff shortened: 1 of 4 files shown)


cmd/fyne/internal/commands/build.go
```diff
@@ -202,6 +202,12 @@ func (b *Builder) build() error {
 		versionConstraint = version.NewConstrainGroupFromString(">=1.17")
 		env = append(env, "GOARCH=wasm")
 		env = append(env, "GOOS=js")
+	} else if goos == "gopherjs" {
+		_, err := b.runner.runOutput("version")
```
andydotxyz, line 206: is version needed? I would have thought that `exec.LookPath` would be faster

andydotxyz (changes requested):
Lots of mocking for a version lookup.
If possible using LookPath would avoid this all

cmd/fyne/internal/commands/build.go
Bluebugs: It would be faster, but doesn't guarantee a working gopherjs. I did think that executing version would give more guarantee. If we are fine with a exec.LookPath, I can implement that.

Bluebugs:
> Lots of mocking for a version lookup. If possible using LookPath would avoid this all

If we use LookPath, I have to start mocking it too to make sure that the environment test are run in are not affected by unexpected change.

andydotxyz (approved):
I guess calling version won't be problematically slow, a shame we can't avoid all the boilerplate

PR #3008, opened 2022-05-26
Support reading .ico icon files if passed to packaging

### LINKED ISSUE #2412: Add support for ico file when compiling for windows
It is currently not possible to use a .ico file as an icon for the app using the `fyne.Window.SetIcon` function, though this might be a bit harder to do on OS's that are not Windows.
**BUT** when compiling for windows you also can't use a .ico file, it tries to make a .ico file from a format it can read (png)

img

You can use a png as the app icon and as the desktop icon

At least when compiling for Windows and you supply a .ico file it should use it instead of trying to make a new one

### Note:
This was discussed [here](https://discord.com/channels/118456055842734083/837686172494725160/881224551067942962) on the [Discord Gophers](https://discord.gg/0f1SbxBZjYq9jLBk) discord

andydotxyz:
Re-encode non-png packaging images

Fixes #2412 

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


cmd/fyne/internal/commands/package.go
```diff
@@ -412,6 +421,35 @@ func mergeMetadata(p *appData, data *metadata.FyneApp) {
...
+		return "", fmt.Errorf("failed to open source image: %w", err)
+	}
+	defer img.Close()
+	srcImg, _, err := image.Decode(img)
+	if err != nil {
+		return "", fmt.Errorf("failed to decode source image: %w", err)
+	}
+
+	tmpPath := filepath.Join(os.TempDir(), "fyne-ico-tmp.png")
+	out, err := os.Create(tmpPath)
```
Bluebugs, line 440: Would it be possible to remove the file once it is not necessary anymore?

andydotxyz: This is a good idea.
I don't think we have a hook for the processed completing at this time.
Can you think of a clean way to do that?

Bluebugs:
Just a bit of concern with this new dependencies regarding its license: https://github.com/biessek/golang-ico/issues/3

cmd/fyne/internal/commands/package.go
Bluebugs: doPackage use `defer p.removeBuild(files)`. Maybe make doPackage and validate return a `[]string` of files that need to be cleaned and set a defer in the caller to those function?

Jacalz, line 440: How about using [ioutil.TempFile()](https://pkg.go.dev/io/ioutil#TempFile) instead?
```suggestion
	out, err := ioutil.TempFile("", "fyne-ico-tmp.png")
```

Jacalz: We could create a folder in the temp directory using https://pkg.go.dev/io/ioutil#ReadDir and save everything there when we create temporary files. We then simply make sure that we just remove the whole folder before exiting the command (we could probably even hook up the interup signal https://www.tutorialspoint.com/how-to-handle-signals-in-golang so that it removes it before exiting using ctrl+c).

cmd/fyne/internal/commands/package.go
```diff
@@ -412,6 +421,35 @@ func mergeMetadata(p *appData, data *metadata.FyneApp) {
...
+	if err != nil {
+		return "", fmt.Errorf("failed to open image output file: %w", err)
+	}
+
+	err = png.Encode(out, srcImg)
+	if err != nil {
+		return "", fmt.Errorf("failed to encode icon: %w", err)
```
Jacalz, line 447: The variable `out` isn't closed if we get an error here.

Jacalz (changes requested):
A few possible improvements.

cmd/fyne/internal/commands/package.go
andydotxyz: I used that originally, but it's not in Go 1.14

Jacalz: Are you sure about that? It seems to exist in Go 1.14 if you look at this: https://pkg.go.dev/io/ioutil@go1.14#TempFile. The `os.CreateTemp` stuff was added in Go 1.16 but this does the same thing.

cmd/fyne/internal/commands/package.go
```diff
@@ -302,7 +306,14 @@ func (p *Packager) removeBuild(files []string) {
...
 }
 
-func (p *Packager) validate() error {
+func (p *Packager) validate() (err error) {
+	p.tempDir, err = ioutil.TempDir("", "fyne-package-*")
+	defer func() {
+		if err != nil {
+			_ = os.RemoveAll(p.tempDir)
+		}
+	}()
```
Jacalz, line 315: I am not sure that this does what is intended. It will not remove the temporary directory if it gets an error anywhere (not just if the `ioutil.TempDir("", "fyne-package-*")` call failed). It might be better to do this instead:
```suggestion
	p.tempDir, err = ioutil.TempDir("", "fyne-package-*")
	if err != nil {
		return err
	}
	defer os.RemoveAll(p.tempDir)
```

Jacalz (changes requested):
This looks really good. Just one comment on a potentially incorrect defer. I also think it might be good to get this updated once https://github.com/fyne-io/image/pull/1 lands.

cmd/fyne/internal/commands/package.go
andydotxyz: Sadly I cannot do that as it should only delete if there was an error.
My understanding is that because `err` is a return var it will be evaluated during the defer as the final value of `err` and so therefore if a later err is set it will catch that too...

Jacalz: Ah, I see. That is how it works as I have understood it as well. I just thought you meant to do something else.
If it only removes the tempdir on error, does this mean that the tempdir fills up every time you do a successful package?

andydotxyz: The directory is also deleted after a successful package.
I would have preferred a closer "create then delete" pairing of code, but we need it in validate and also package, which are called directly by other modes as well (release, install)

andydotxyz:
> FYI, there are also conflicts with the base branch that need to be resolved.

Thanks, fixed

Jacalz (approved):
Works like a charm. Thanks :+1:

PR #3012, opened 2022-05-29
fix clipping of large scrollable content

toaster:
### Description:

PR #2893 included an optimization to not paint an object if the intersection of the visible area and the object’s current position and size was empty. This computation does not consider that the object might have been scrolled. In effect, the background of large pop-ups (like the icon list in the demo app) vanished when scrolling down too much.

The optimization is removed for now. If it will ever be re-introduced, it has to consider scrolling and there have to be tests, of course.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/canvas.go
+++ b/internal/driver/glfw/canvas.go
@@ -276,14 +276,6 @@ func (c *glCanvas) paint(size fyne.Size) {
 			inner := clips.Push(pos, obj.Size())
 			c.Painter().StartClipping(inner.Rect())
 		}
-
-		if visibleArea := clips.Top(); visibleArea != nil {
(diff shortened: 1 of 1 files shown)


andydotxyz:
This code was to resolve a bug. Can it be improved instead of reverted? In the scroller if above it should be possible to capture the offset to fix the intersect calculations?

toaster:
This code did not resolve the bug. The other part of the PR did.
This code only tried to optimise painting but broke the scrolling of large contents.
Yes, the optimisation could be implemented in a working way, I guess. However, this is not the focus of this PR.

andydotxyz (approved):
Thanks for the explanation, let’s get this merged so we’re ready for next release

PR #3026, opened 2022-06-02
Allow hyperlink to override default behaviour

### LINKED ISSUE #2979: Feature request: custom hyperlink handling via click event
This is my first time opening an issue here (I don't usually use github). I read the docs, but wasn't sure about feature requests. My apologies if this is out of scope or incorrectly formatted (I'm happy to update if needed).

## Feature request description

This is not a bug, but is a gap that I think is unhandled. At present there is a hyperlink widget. It opens the given link in the default handler. I have verified that this works for http, gopher, gemini, finger etc. The correct handler based on my system configuration opens the link.

What I would like to be able to do it essentially catch the click and handle what happens within application code rather than passing control to an outside application. This can be done with a button, which is great. However, it doesn't solve for something like creating an internet client or having rich documentation of multiple pages within the application itself. I would like to be able to have a hypertext document (whether html or otherwise) render (not in scope of this, the rich text from markdown function works fine for this purpose) and be able to handle the click in my own code rather than port out to another application. I would think this could be solved by adding access to the hyperlink click event and letting a custom function be added to the widget that if nil will open in the default handler and if present the default handler will not be used and the given function will be called (with access to the underlying widget's link text and link url).

Adding this functionality in the hyperlink widget as well as the hyperlink segment of rich text objects could be very useful for building applications that can have linked content within the application or for the development of custom protocols for navigating around an application.

It is likely that the callback would need to be triggered by enter as well.

### Example code:

```
type Hyperlink struct {
	BaseWidget
	Text      string
	URL       *url.URL
	Alignment fyne.TextAlign
	Wrapping  fyne.TextWrap
	TextStyle fyne.TextStyle 
        Callback func(text, url string) // <--- add this, if nil default behavior will be used
}
```

andydotxyz:
Fixes #2979

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/hyperlink.go
+++ b/widget/hyperlink.go
@@ -23,6 +23,11 @@ type Hyperlink struct {
 	Wrapping  fyne.TextWrap  // The wrapping of the Text
 	TextStyle fyne.TextStyle // The style of the hyperlink text
 
+	// OnTapped overrides the default `fyne.OpenURL` call when the link is tapped
+	//
+	// Since: 2.2
+	OnTapped func() `json:"-"`
+
 	focused, hovered bool
 	provider         *RichText
 }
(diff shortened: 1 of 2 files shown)


Jacalz (approved):
Looks good to me.

PR #3027, opened 2022-06-02
Completing Chrisbirster new string with format

### LINKED ISSUE #2890: Add binding.NewStringWithFormat

andydotxyz:
Wrap up the good work started in #2942, but match our naming and return style.

Fixes #2890

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/data/binding/sprintf.go
+++ b/data/binding/sprintf.go
@@ -210,3 +210,16 @@ func (s *sprintfString) Set(str string) error {
 
 	return nil
 }
+
+// StringToStringWithFormat creates a binding that converts a string to another string using the specified format.
+// Changes to the returned String will be pushed to the passed in String and setting a new string value will parse and
+// set the underlying String if it matches the format and the parse was successful.
+//
+// Since: 2.2
+func StringToStringWithFormat(str String, format string) String {
+	if format == "%s" { // Same as not using custom formatting.
+		return str
+	}
+
+	return NewSprintf(format, str)
+}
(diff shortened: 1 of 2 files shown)


data/binding/sprintf.go
```diff
@@ -203,3 +203,16 @@ func (s *sprintfString) Set(str string) error {
...
+
+// StringToStringWithFormat creates a binding that converts a string to another string using the specified format.
+// Changes to the returned String will be pushed to the passed in String and setting a new string value will parse and
+// set the underlying String if it matches the format and the parse was successful.
+//
+// Since: 2.2
+func StringToStringWithFormat(str String, format string) String {
+	if format == "%s" { // Same as not using custom formatting.
+		return str
+	}
```
Jacalz, line 215: Might it be worth to also test this part in the tests that you added? It should probably even out the coverage.

Jacalz (approved):
This looks good to me but I added a comment that you might want to look at before merging.

data/binding/sprintf.go
andydotxyz: Good point, thanks!

Jacalz (approved):

Issue #3032, opened 2022-06-03
Add option to hide or show file extension in filedialog


Jacalz:
### Is your feature request related to a problem? Please describe:

It was mentioned in https://github.com/Jacalz/rymdport/issues/39 that it would be good to see the full file extensions for items within the file dialog. The icons have the extension text in them but it might be hard to read from the smaller icons in the list view.

### Is it possible to construct a solution with the existing API?

It should be. We just need to add a checkbox to the settings and then control how the item is displayed.
The most amount of code is likely to end up in the latter part.

### Describe the solution you'd like to see:

Something similar to the option to show hidden files.

PR #3033, opened 2022-06-03
dialog: show file extension in list view

Jacalz:
### Description:

Due to icons being a lot smaller for the list view, we should show the file extension to make it more clear to the user what file they are seeing. This PR makes that change and manages to clean up and optimize the code a bit in the process.

Fixes https://github.com/Jacalz/rymdport/issues/39 but should be superseded by #3032 for a future release.

image
image

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/fileitem.go
+++ b/dialog/fileitem.go
@@ -63,15 +63,6 @@ func (i *fileDialogItem) CreateRenderer() fyne.WidgetRenderer {
 	}
 }
 
-func fileName(path fyne.URI) (name string) {
-	pathstr := path.String()[len(path.Scheme())+3:]
-	name = filepath.Base(pathstr)
-	ext := filepath.Ext(name[1:])
-	name = name[:len(name)-len(ext)]
-
-	return
-}
-
 func (i *fileDialogItem) isDirectory() bool {
 	return i.dir
 }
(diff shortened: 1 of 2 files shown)


Jacalz:
The option in VSCode to only run a specific test is very useful but does apparently not help when you break other tests :see_no_evil:

dialog/fileitem.go
```diff
@@ -80,13 +71,13 @@ func (f *fileDialog) newFileItem(location fyne.URI, dir bool) *fileDialogItem {
...
+		name:     location.Name(),
 		dir:      dir,
 	}
 
-	if dir {
-		item.name = location.Name()
-	} else {
-		item.name = fileName(location)
+	if f.view == gridView {
+		ext := filepath.Ext(item.name[1:])
```
andydotxyz, line 79: I think this might crash for a file named "H"?

andydotxyz (changes requested):
Looks great, but I see a potential panic

dialog/fileitem.go
Jacalz: It seems like it actually doesn't. We were doing the same thing in the code that I removed.
image

andydotxyz: Oops! thanks

andydotxyz (approved):

PR #3035, opened 2022-06-03
fix: top Table items were invisible after window resize

### LINKED ISSUE #3034: top Table elements become invisible after window maximization
### Describe the bug:
The top elements (that were hidden because of scrolling) do not appear when you maximize the window (see screenshot). But when you click at those hidden elements they appear.

Similar to #3013 

### To Reproduce:
1. Create a window with a table with number of rows high enough that there is scrolling but low enough that there is no scrolling when the window is maximized
2. Scroll to the bottom of the table
3. Click on a random element
4. Maximize window

### Screenshots:

2022-06-03-232150_406x374_scrot

### Example code:
```go
package main

import (
	"time"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("table")
	t := createTable()
	w.SetContent(t)
	w.Resize(fyne.NewSize(640, 480))
	go func() {
		time.Sleep(time.Second)
		t.ScrollToBottom()
		time.Sleep(time.Second)
		t.Select(widget.TableCellID{Row: 15, Col: 1})
		time.Sleep(time.Second)
		w.SetFullScreen(true)
	}()
	w.ShowAndRun()
}

func createTable() *widget.Table {
	rowData := []string{"c1", "c2", "c3", "c4", "c5", "c6", "c7", "c8", "c9", "c10"}
	data := make([][]string, 0)
	const numOfRows = 20
	for i:=0; i<numOfRows; i++ {
		data = append(data, rowData)
	}
	t := widget.NewTable(
		func() (int, int) {
			return len(data), len(data[0])
		},
		func() fyne.CanvasObject {
			return widget.NewLabel("asdfghjkl")
		},
		func(cellID widget.TableCellID, obj fyne.CanvasObject) {
			obj.(*widget.Label).SetText(data[cellID.Row][cellID.Col])
		})
	return t
}

```

 - **OS:** Linux
 - **Version:** 5.17.9
 - **Go version:** 1.18.2
 - **Fyne version:** 3c1f987

cmitsakis:
### Description:
Fixes #3034 

This appears to be working.

Unfortunately I couldn't write tests for this one because I don't know how to check for visible elements, but I decided to send the PR now hoping it will be merged before the next release.

The issue was happening only when the size of the table increased as a result of the window size increasing. So in that case only I call ScrollToTop().

Losing scroll position doesn't seem to be a problem. The top row appears anyway, by calling ScrollToTop() I make it visible. But there might be cases where it causes a small inconvenience.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/table.go
+++ b/widget/table.go
@@ -512,7 +512,13 @@ func (c *tableCells) Resize(s fyne.Size) {
 	if s == c.size {
 		return
 	}
+	cSizeBeforeResize := c.size
 	c.BaseWidget.Resize(s)
+	if c.size.Height > cSizeBeforeResize.Height {
+		// table height has increased.
+		// call ScrollToTop() otherwise empty rows will appear at the top of the table.
+		c.t.ScrollToTop()
+	}
 	c.Refresh() // trigger a redraw
 }
 
--- a/widget/table_test.go
+++ b/widget/table_test.go
@@ -585,42 +585,30 @@ func TestTable_ScrollToBottomThenResize(t *testing.T) {
 		height  float32 = 10
 	)
 
-	const labelText = "a"
-
-	data := make([][]string, 0)
-	for i := 0; i < maxRows; i++ {
-		row := []string{}
-		for j := 0; j < maxCols; j++ {
-			row = append(row, labelText)
-		}
-		data = append(data, row)
-	}
-
-	createdLabels := make([]*Label, 0)
+	createdCells := make([]*canvas.Rectangle, 0)
 
 	table := NewTable(
 		func() (int, int) { return maxRows, maxCols },
 		func() fyne.CanvasObject {
-			l := NewLabel("")
-			createdLabels = append(createdLabels, l)
-			return l
+			templ := canvas.NewRectangle(color.Gray16{})
+			templ.SetMinSize(fyne.Size{Width: width, Height: height})
+			createdCells = append(createdCells, templ)
+			return templ
 		},
-		func(cellID TableCellID, o fyne.CanvasObject) {
-			o.(*Label).SetText(data[cellID.Row][cellID.Col])
-		})
+		func(cellID TableCellID, o fyne.CanvasObject) {})
 
 	w := test.NewWindow(table)
 	defer w.Close()
 
-	// choose a size small enough that there is vertical scroll
-	w.Resize(fyne.NewSize(100, 50))
+	// choose a size small enough that less than half of the rows are visible
+	w.Resize(fyne.NewSize(200, float32(maxRows/2-2)*(table.cells.cellSize.Height+theme.SeparatorThicknessSize())+2*theme.Padding()))
 	table.ScrollToBottom()
-	table.Select(TableCellID{Row: maxRows - 1, Col: 1})
-	// increase size big enough that there is no scroll
-	w.Resize(fyne.NewSize(1000, 1000))
+	// increase size enough so that there is no scroll
+	w.Resize(fyne.NewSize(200, float32(maxRows)*(table.cells.cellSize.Height+theme.SeparatorThicknessSize())+2*theme.Padding()))
 	var visibleCount int
-	for _, l := range createdLabels {
-		if l.Text == labelText {
+	var zeroSize = fyne.Size{Width: 0, Height: 0}
+	for _, c := range createdCells {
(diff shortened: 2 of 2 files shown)


Jacalz (changes requested):
Thanks for the PR. It would be good to add a test to make sure that the issue doesn’t appear again.

cmitsakis:
In Table it's not possible to access the cells because they are stored in TableCellsRenderer. That's why I couldn't create a test (although I created a test for my fix in List where I could access the items).

But I found a workaround that works. I store every label created by CreateCell() in a slice and check the text of those labels. It seems to work although I don't know if it's a good idea and I don't really understand it.

widget/table.go
```diff
@@ -512,7 +512,13 @@ func (c *tableCells) Resize(s fyne.Size) {
 	if s == c.size {
 		return
 	}
+	cSizeBeforeResize := c.size
 	c.BaseWidget.Resize(s)
+	if c.size.Height > cSizeBeforeResize.Height {
+		// table height has increased.
+		// call ScrollToTop() otherwise empty rows will appear at the top of the table.
+		c.t.ScrollToTop()
```
andydotxyz, line 520: I see that this fixes some instances of the issue, but if there are 1000 items and you are looking at 900 and the view gets larger then scrolling to the top won't be the right thing to do

widget/table_test.go
```diff
@@ -572,3 +572,58 @@ func (t *separatorThicknessZeroTheme) Size(n fyne.ThemeSizeName) float32 {
...
+		})
+
+	w := test.NewWindow(table)
+	defer w.Close()
+
+	// choose a size small enough that there is vertical scroll
+	w.Resize(fyne.NewSize(100, 50))
+	table.ScrollToBottom()
+	table.Select(TableCellID{Row: maxRows - 1, Col: 1})
+	// increase size big enough that there is no scroll
```
andydotxyz, line 619: Can we trigger it with a smaller increase, so it handles the case where there should still be hidden items, but that some visible ones are not showing up?

andydotxyz (changes requested):
Thanks for looking at this, but as you say there probably needs to be a more accurate fix for this issue

cmitsakis:
I improved the test. I found out how to calculate the exact size that triggers this bug.  Although when I run the test with -race it doesn't work as expected. very strange.

Also when I added this line
```go
log.Printf("c.size.Height (%v) > cSizeBeforeResize.Height (%v)\n", c.size.Height, cSizeBeforeResize.Height)
```
in Resize() before I call `c.t.ScrollToTop()`, I noticed that before I increase the size of the window c.size.Height = 109 and after I increase it c.size.Height = 110

So I'm wondering what would happen if I fixed the size to be 110 instead of 109. Not sure where this value is calculated.

I will look into it again tomorrow to see if I can improve my fix.

widget/table_test.go
cmitsakis: I believe it's not possible to trigger this bug if there are hidden items.

widget/table.go
cmitsakis: I noticed that the if condition becomes true only when the window is resized to a size big enough that there are no hidden items. So `ScrollToTop()` doesn't cause any problems.
But there should be a better fix. In my comment below I describe what I noticed about the value of `c.size.Height`.

andydotxyz: Your explanation makes sense - but it does not match what the code is doing now, so it will still scroll up some times when it is not desired.
I agree there is likely a better fix, have you any ideas?

Jacalz:
Please don’t force-push your changes. It messes with the commit history and make everything harder to review. We’ll squish when merging if we think it’s needed.

widget/table.go
cmitsakis: I agree there is likely a better fix, have you any ideas?
here is an idea I tried at first: https://github.com/fyne-io/fyne/commit/cfaeb96b6476d2d254809e19858f4e62cb736b94
I deleted it by force pushing because I noticed it doesn't work when I make the window smaller. but if I can fix it further it might be a better solution.

andydotxyz:
This fix was covered in a separate PR at some point, closing this change.

PR #3041, opened 2022-06-06
Update testify to bring in fix for CVE-2022-28948

Jacalz:
### Description:

This makes sure that we bring in a security fix for https://github.com/advisories/GHSA-hp87-p4gw-j4gq.
I was getting security warnings for this with Rymdport.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Looks good.
The release prep is on release branch already so you can merge this without impacting.
If it's high importance we can bring it into the branch if we have to roll another RC for any reason.

Jacalz:
For future reference: it was agreed that this isn't a blocker given that it isn't used by any core Fyne dependencies. As far as I can see, users depending on Fyne should not be impacted by the vulnerability.

PR #3047, opened 2022-06-07
internal/async: prevent unbounded channel may not be close properly

changkun:
### Description:

If there are more than 16 unprocessed objects, the unbounded channel may
not be closed gracefully because the existing close branch can lead to a
spin loop.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


changkun:
Hmmm. I still don't completely understand the behavior of a select statement when there is a default branch involved. The language spec seems to lack documentation on this type of use. I'll close this first until I fully understand what might be the best fix.

The initial issue is less problematic than the existing sync.UnboundedChann use in fyne only happens when the application terminates, and there is always a receiver that will never meet the spin loop.

Sorry for the noise.

PR #3049, opened 2022-06-08
fix: Table offset should not be negative

### LINKED ISSUE #2857: Negative TableCellID Row
### Describe the bug:

I get a negative TableCellID Row when table is too wide (?).

The bug seems to be due to this line
https://github.com/fyne-io/fyne/blob/7229e889d49c81a83b0b7e09400837f67f6ddad5/widget/table.go#L640
`r.cells.t.offset.Y` is negative for some reason.

Adding the following code fixes the issue:
```go
if r.cells.t.offset.Y < 0 {
	r.cells.t.offset.Y = 0
}
```

```
panic: runtime error: index out of range [-3]

goroutine 50 [running]:
main.main.func3({0xc000108270?, 0xc042a01800?}, {0x4470d78?, 0xc000108270?})
        /Users/xxx/Documents/project/cmd/test1/main.go:24 +0xa5
fyne.io/fyne/v2/widget.(*tableCellsRenderer).Refresh(0xc0000b4240)
        /Users/xxx/Documents/project/vendor/fyne.io/fyne/v2/widget/table.go:673 +0x691
fyne.io/fyne/v2/widget.(*BaseWidget).Refresh(0xc0000ba000?)
        /Users/xxx/Documents/project/vendor/fyne.io/fyne/v2/widget/widget.go:138 +0x2e
fyne.io/fyne/v2/widget.(*Table).finishScroll(0xc0003c0820)
        /Users/xxx/Documents/project/vendor/fyne.io/fyne/v2/widget/table.go:289 +0x4e
fyne.io/fyne/v2/widget.(*Table).ScrollToBottom(0xc0003c0820)
        /Users/xxx/Documents/project/vendor/fyne.io/fyne/v2/widget/table.go:215 +0x9f
main.main.func4()
        /Users/xxx/Documents/project/cmd/test1/main.go:33 +0x149
created by main.main
        /Users/xxx/Documents/project/cmd/test1/main.go:30 +0x319

Process finished with the exit code 2
```

### To Reproduce:
1. Run the below script

### Example code:
```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/layout"
	"fyne.io/fyne/v2/widget"
	"strconv"
	"time"
)

func main() {
	a := app.New()
	w := a.NewWindow("Some title")

	tableData := make([][]string, 0)
	table := widget.NewTable(func() (int, int) { return len(tableData), 3 },
		func() fyne.CanvasObject {
			item := widget.NewLabel("template")
			return item
		},
		func(i widget.TableCellID, o fyne.CanvasObject) {
			o.(*widget.Label).SetText(tableData[i.Row][i.Col])
		})
	table.SetColumnWidth(0, 40)
	table.SetColumnWidth(1, 80)
	table.SetColumnWidth(2, 80)

	go func() {
		time.Sleep(3 * time.Second)
		tableData = append(tableData, []string{strconv.Itoa(len(tableData) + 1), "test1", "test2"})
		table.ScrollToBottom()
	}()

	c := container.New(layout.NewMaxLayout(), table)

	w.SetContent(c)
	w.Resize(fyne.NewSize(300, 200))
	w.ShowAndRun()
}
```

 - **OS:** MacOS
 - **Version:** 12.1 Monterey
 - **Go version:** 1.18
 - **Fyne version:** 2.1.4

### LINKED ISSUE #2918: problem with table.ScrollToBottom
When the table is long and only part of it is displayed, then ScrollToBottom works fine.
When the table is short and is displayed in full, then ScrollToBottom fails.

The test bellow works with "go run . 12" and fails with "go run . 4"

```go
package main

import (
	"os"
	"strconv"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
    "fyne.io/fyne/v2/widget"
)

var tableTest *widget.Table
var theApp fyne.App
var theWindow fyne.Window

// -----
func onSomeCellSelected(id widget.TableCellID) {
	tableTest.ScrollToBottom()
}


// ----- main
func main() {
	arg := os.Args[1]
	len, _ := strconv.Atoi(arg)

	theApp = app.New()
	theWindow = theApp.NewWindow("test")
	theWindow.Resize(fyne.NewSize(200, 300))

	data := []string{"00", "01", "02", "03", "04", "05", "06", "07",
		"08", "09", "10", "11", "012", "013", "14", "15"}

	tableTest = widget.NewTable(
		func() (int, int) { return len, 1 },
		func() fyne.CanvasObject {
			return widget.NewLabel("****")
		},
		func(i widget.TableCellID, obj fyne.CanvasObject) {
			// i is index with i.Col and i.Row
			obj.(*widget.Label).SetText(data[i.Row])
		},
	)
	tableTest.OnSelected = onSomeCellSelected

	theWindow.SetContent(container.NewBorder(
		nil, nil, nil, nil, tableTest))
	theWindow.ShowAndRun()
}
```

cmitsakis:
### Description:

Fixes #2857 
edit: it seems it also fixes #2918 

Whenever t.scroll.Offset or t.offset is assigned a value, I added checks to make sure it is not negative.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz (commented):
Thanks for the PR. This seems to me like it is more of a workaround than a fix. I feels more like something is wrong in the code above where the offset is calculated. I might be wrong but that’s just my thoughts.

andydotxyz:
Can you help us get this landed @cmitsakis?

cmitsakis:
I will look into it this week

cmitsakis:
This bug happens because sometimes t.offset.Y becomes negative.
This happens after this line:
https://github.com/fyne-io/fyne/blob/56dd0f3e0f3233e5f72087cd44d04eeac24ee6dd/widget/table.go#L402
This is the code of this line:
```Go
y := cellY + cellHeight - t.content.Size().Height
```
You understand this code better than me, but if I understood correctly in the function ScrollToBottom() you find the last row, calculate the position (cellY + cellHeight), add subtract the height of the widget because you want to scroll not exactly at the end but a bit before so the last rows are visible. But if there are not enough rows that calculation results in a negative value. So I just make it 0 if it's negative. I believe that's correct.

The same applies to ScrollToTrailing().

In ScrollTo() I just made a change. I only check for negative value when you subtract the height of the widget because I believe it can't happen in other cases. I'm not 100% sure about this.

andydotxyz:
> > EDIT: just thought that maybe in that case we could just not scroll at all. I can test it.
> 
> just tested it
> 
> ```go
> y := cellY + cellHeight - t.content.Size().Height
> if y < 0 {
>         return
> }
> ```
> 
> works fine too. I think it's the cleanest solution.
> 
> if y<0, I noticed that offset is 0 anyway so there is no point in updating it's value.

Yes I think you are right here - exiting early is clean and also it is normally optimal too ;)

cmitsakis:
so I changed ScrollToBottom() and ScrollToTrailing() to return early.

I also just noticed that in ScrollTo() this bug cannot happen because the if statement guarantees the calculated offset is always positive:
https://github.com/fyne-io/fyne/blob/56dd0f3e0f3233e5f72087cd44d04eeac24ee6dd/widget/table.go#L381
so I removed my fix from ScrollTo()

cmitsakis:
do not accept this yet. I only tested it in a table with a couple of rows.

Just tested it in a table with more rows and after calling `ScrollToBottom()` I get this:
2023-05-24-200310_296x191_scrot

cmitsakis:
> do not accept this yet. I only tested it in a table with a couple of rows.
> 
> Just tested it in a table with more rows and after calling `ScrollToBottom()` I get this: 
and when I click I get the correct behavior which is this: 
it reminds me of this bug #3034

This is a different issue not caused by my fix. I opened #3918

Jacalz:
Thanks for your work on this. Does that mean that you want us to review this now? If so, we might be able to squeeze it in for v2.3.5 if everything goes according to plan

cmitsakis:
> Does that mean that you want us to review this now?

yes. I believe this is the best fix I can offer.

Jacalz (approved):
This looks good to me and I can verify that the linked issue is indeed fixed. Well done and many thanks for working on this @cmitsakis.

I am sorry that this took such a long time to get landed. It was a slip from our part and we will do our best to make sure that it does not happen again. You are of course always welcome to contribute improvements wherever you see fit :)

Jacalz:
@andydotxyz Would you mind merging and cherry-picking this over to `releases/v2.3.x` if you are happy with the changes?

andydotxyz (approved):

andydotxyz:
> @andydotxyz Would you mind merging and cherry-picking this over to `releases/v2.3.x` if you are happy with the changes?

Done and done :)

PR #3051, opened 2022-06-09
Fix #3050, update onTapped hooker for tabButton

### LINKED ISSUE #3050: Tab buttons stop working after removing a tab
### Describe the bug:

When removing a tab from AppTabs the tabs that are to the right of the removed tab cannot be selected with mouse clicks.
If there are tabs 1, 2, 3, 4 and tab 2 is removed then the tab buttons for 3 and 4 will not respond to mouse clicks.  If tabs 5 and 6 are then added, the buttons for tabs 5 and 6 will still work like normal.

### To Reproduce:
Running the below code, go to File -> New Tab two or three times.
Then select a tab that is not the last tab and click the "Close Tab" button, after this the tab buttons of tabs to the right of the removed tab will not work.
If another tab is added then removed then the last tab will be selected.

### Example code:

```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow()
	t := buildTabsCont()
	w.SetMainMenu(buildMenu(t))

	w.SetContent(t)
	w.ShowAndRun()
}

func buildMenu(t *container.AppTabs) *fyne.MainMenu {
	nt := fyne.NewMenuItem("New Tab", func() { newTab(t) })
	file := fyne.NewMenu("File", nt)
	m := fyne.NewMainMenu(file)
	return m
}

func buildTabsCont() *container.AppTabs {
	at := container.NewAppTabs(
		container.NewTabItem("Welcome", widget.NewLabel("Welcome info here")),
	)
	at.SetTabLocation(container.TabLocationTop)
	return at
}

func newTab(t *container.AppTabs) {
	b := widget.NewButton("Close Tab", func() { t.RemoveIndex(t.SelectedIndex()) })
	t.Append(container.NewTabItem("Tab Name", b))
}
```

### Device:
 - **OS:** Linux/Pop_OS
 - **Version:** 22.04
 - **Go version:** 1.18.1
 - **Fyne version:** v2.1

d4x1:
### Description:

Fixes #3050

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/container/apptabs.go
+++ b/container/apptabs.go
@@ -341,9 +341,8 @@ func (r *appTabsRenderer) buildTabButtons(count int) *fyne.Container {
 		item := r.appTabs.Items[i]
 		button, ok := r.buttonCache[item]
 		if !ok {
-			index := i // capture
 			button = &tabButton{
-				onTapped: func() { r.appTabs.SelectIndex(index) },
+				onTapped: func() { r.appTabs.Select(item) },
 			}
 			r.buttonCache[item] = button
 		}


container/apptabs.go
```diff
@@ -341,9 +341,8 @@ func (r *appTabsRenderer) buildTabButtons(count int) *fyne.Container {
 		item := r.appTabs.Items[i]
 		button, ok := r.buttonCache[item]
 		if !ok {
-			index := i // capture
 			button = &tabButton{
-				onTapped: func() { r.appTabs.SelectIndex(index) },
+				onTapped: func() { r.appTabs.Select(item) },
```
Jacalz, line 345: Hmm. I thought a bit more about this. The fix works but the change also results in that we loop through all tab items to see if the tab exists when we tap a it. That isn’t very elegant or performant.

PR #3065, opened 2022-06-12
Feature/cloud

andydotxyz:
Reviving my implementation of cloud hooks.
This is an implementation of proposal https://github.com/fyne-io/proposals/blob/main/cloud.md.

Draft for now, as I test with latest develop and changes in preference API, but should be good to go.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/pref_helper.go
+++ b/data/binding/pref_helper.go
@@ -80,25 +80,25 @@ func (m *preferencesMap) preferencesChanged(p fyne.Preferences) {
 	}
 }
 
-func (m *preferencesMap) migratePreferences(p1, p2 fyne.Preferences) {
-	old, loaded := m.prefs.Load(p1)
+func (m *preferencesMap) migratePreferences(src, dst fyne.Preferences) {
+	old, loaded := m.prefs.Load(src)
 	if !loaded {
 		return
 	}
 
-	m.prefs.Store(p2, old)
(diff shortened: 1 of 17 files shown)


data/binding/pref_helper.go
```diff
@@ -70,3 +79,26 @@ func (m *preferencesMap) preferencesChanged(p fyne.Preferences) {
 		item.checkForChange()
 	}
 }
+
+func (m *preferencesMap) migratePreferences(p1, p2 fyne.Preferences) {
```
Bluebugs, line 83: This look like it is moving data from fyne.Preferences to another. Maybe using src/dst instead of p1/p2 would make it more readable.

Bluebugs (approved):

andydotxyz:
I don't think you should have let me away with that @Bluebugs - there are no tests yet!

Bluebugs:
> I don't think you should have let me away with that @Bluebugs - there are no tests yet!
> 

Indeed and the coverage score even went down. Should have pointed that out.

Bluebugs (changes requested):
Indeed need tests and get the coverage at least on par if not up!

andydotxyz:
I think those tests cover most. Not certain if we get an updated report from Coveralls...

PR #3075, opened 2022-06-16
Features/detect fyne version

Bluebugs:
### Description:
Detect the fyne version used in go.mod to see if it does have the minimal requirement for enabling metadata generation.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/build.go
+++ b/cmd/fyne/internal/commands/build.go
@@ -196,7 +196,7 @@ func (b *Builder) injectMetadataIfPossible(runner runner, createMetadataInitFile
 	fyneGoModVersionNormalized := version.Normalize(fyneGoModVersion)
 	fyneGoModVersionConstraint := version.NewConstrainGroupFromString(">=2.2")
 	if fyneGoModVersion != "master" && !fyneGoModVersionConstraint.Match(fyneGoModVersionNormalized) {
-		return nil, fmt.Errorf("fyne command line version is more recent than the version used in go.mod")
+		return nil, nil
 	}
 
 	return createMetadataInitFile()
--- a/cmd/fyne/internal/commands/build.go
+++ b/cmd/fyne/internal/commands/build.go
@@ -189,16 +189,17 @@ func (b *Builder) createMetadataInitFile() (func(), error) {
 
 func (b *Builder) injectMetadataIfPossible(runner runner, createMetadataInitFile func() (func(), error)) (func(), error) {
 	fyneGoModVersion, err := getFyneGoModVersion(runner)
-	if err == nil {
-		fyneGoModVersionNormalized := version.Normalize(fyneGoModVersion)
-		fyneGoModVersionConstraint := version.NewConstrainGroupFromString(">=2.2")
-		if fyneGoModVersion == "master" || fyneGoModVersionConstraint.Match(fyneGoModVersionNormalized) {
-			return createMetadataInitFile()
-		} else {
-			return nil, fmt.Errorf("fyne command line version is more recent than the version used in go.mod")
-		}
+	if err != nil {
+		return nil, err
 	}
-	return nil, err
+
+	fyneGoModVersionNormalized := version.Normalize(fyneGoModVersion)
+	fyneGoModVersionConstraint := version.NewConstrainGroupFromString(">=2.2")
+	if fyneGoModVersion != "master" && !fyneGoModVersionConstraint.Match(fyneGoModVersionNormalized) {
+		return nil, fmt.Errorf("fyne command line version is more recent than the version used in go.mod")
+	}
+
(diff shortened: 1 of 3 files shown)


Bluebugs:
I have done this PR against develop, but ultimately I think it should go in the next v.2.2.2

cmd/fyne/internal/commands/build.go
```diff
@@ -159,19 +174,64 @@ func (b *Builder) build() error {
...
 			writeResource(b.icon, "fyneMetadataIcon", metadataInitFile)
 		}
 	}
-	metadataInitFile.Close()
+
+	return func() { os.Remove(metadataInitFilePath) }, err
+}
+
+func (b *Builder) injectMetadataIfPossible(runner runner, createMetadataInitFile func() (func(), error)) func() {
+	if fyneGoModVersion, err := getFyneGoModVersion(runner); err == nil {
```
andydotxyz, line 192: This `err` is never communicated

cmd/fyne/internal/commands/build.go
```diff
@@ -159,19 +174,64 @@ func (b *Builder) build() error {
 	metadataInitFilePath := filepath.Join(b.srcdir, "fyne_metadata_init.go")
 	metadataInitFile, err := os.Create(metadataInitFilePath)
 	if err != nil {
-		fyne.LogError("Failed to make metadata init file, omitting metadata", err)
+		return func() {}, err
```
andydotxyz, line 177: Why the 3 occurrences of `func() {}` instead of `nil`?

andydotxyz (changes requested):
Clever approach, I did not know that `go` could export module in JSON

cmd/fyne/internal/commands/build.go
```diff
@@ -128,29 +130,42 @@ func checkGoVersion(runner runner, versionConstraint *version.ConstraintGroup) e
...
+		Path    string
+		Version string
+	}
+}
 
-	goos := b.os
-	if goos == "" {
-		goos = targetOS()
+func getFyneGoModVersion(runner runner) (string, error) {
+	dependenciesOutput, err := runner.runOutput("mod", "edit", "-json")
```
lucor, line 144: Another possible approach could be to use the https://pkg.go.dev/golang.org/x/mod@v0.5.1/modfile#Parse

Bluebugs: Yes, I started with that, but couldn't figure out an API that find the go.mod file properly. The call to `go mod edit -json` solved that problem.

cmd/fyne/internal/commands/build.go
```diff
@@ -188,21 +187,18 @@ func (b *Builder) createMetadataInitFile() (func(), error) {
 	return func() { os.Remove(metadataInitFilePath) }, err
 }
 
-func (b *Builder) injectMetadataIfPossible(runner runner, createMetadataInitFile func() (func(), error)) func() {
-	if fyneGoModVersion, err := getFyneGoModVersion(runner); err == nil {
+func (b *Builder) injectMetadataIfPossible(runner runner, createMetadataInitFile func() (func(), error)) (func(), error) {
+	fyneGoModVersion, err := getFyneGoModVersion(runner)
+	if err == nil {
```
andydotxyz, line 192: Go (and Fyne) prefers exit early so happy path is left aligned - so exit with nil, err in this if err != nil, the rest gets less indented.

andydotxyz (changes requested):
Thanks, just 1 more tidy - don't put the body of a method in `if err == nil` but exit with the error early.

andydotxyz:
> I have done this PR against develop, but ultimately I think it should go in the next v.2.2.2

For sure we can pick it to the release branch after merge

andydotxyz (changes requested):
Sorry, just realised that with all errors getting printed the "older dep" isn't really an error...

cmd/fyne/internal/commands/build.go
```diff
@@ -189,16 +189,17 @@ func (b *Builder) createMetadataInitFile() (func(), error) {
...
-		}
+	if err != nil {
+		return nil, err
 	}
-	return nil, err
+
+	fyneGoModVersionNormalized := version.Normalize(fyneGoModVersion)
+	fyneGoModVersionConstraint := version.NewConstrainGroupFromString(">=2.2")
+	if fyneGoModVersion != "master" && !fyneGoModVersionConstraint.Match(fyneGoModVersionNormalized) {
+		return nil, fmt.Errorf("fyne command line version is more recent than the version used in go.mod")
```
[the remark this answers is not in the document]
andydotxyz: But the metadata API does not exist back then...

andydotxyz (approved):
Nice thanks

PR #3094, opened 2022-06-22
only quit applications when no windows remain if theres no systrayMen…

### LINKED ISSUE #3092: Do not close app when last window is closed but systrayMenu exists
I'm looking for a reliable way to have a systray app that has no visible window.
When there is no hidden window, the app closes after the last visible window is closed.

The systrayMenu is already offering a "Quit" entry, so we should not count windows.

d1ss0nanz:
### Description:
only quit application when no windows remain if there is no systrayMenu set.

Fixes #3092 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/window.go
+++ b/window.go
@@ -76,8 +76,9 @@ type Window interface {
 	// This will not destroy the window or cause the app to exit.
 	Hide()
 	// Close the window.
+	// If it is he "master" window the app will Quit.
 	// If it is the only open window and no menu is set via [desktop.App]
-	// SetSystemTrayMenu the app will Quit.
+	// SetSystemTrayMenu the app will also Quit.
 	Close()
 
 	// ShowAndRun is a shortcut to show the window and then run the application.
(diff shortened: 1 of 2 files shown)


andydotxyz (changes requested):
This probably needs to be documented in the Close() function for Window interface.

window.go
```diff
@@ -76,7 +76,8 @@ type Window interface {
 	// This will not destroy the window or cause the app to exit.
 	Hide()
 	// Close the window.
-	// If it is the only open window, or the "master" window the app will Quit.
+	// If it is the only open window and no menu is set via [desktop.App]
+	// SetSystemTrayMenu, or if it is the "master" window the app will Quit.
```
Bluebugs, line 80: I think even if it is the "master" window, as long as there is a system tray menu set, it won't quit. I feel the sentence here is a bit confusing to me.

andydotxyz: Are you sure @Bluebugs? I think closing a window marked as master will still exit the app.

andydotxyz (approved):
Great thanks

PR #3096, opened 2022-06-22
support controling visiblity of system tray menu items

### LINKED ISSUE #3095: Support to control visibility of system tray MenuItems
I would like to be able to control the visibility of a systrayMenu `MenuItem`.

### Is it possible to construct a solution with the existing API?C
Currently we have to add/remove items from the menu and then refresh it.

Add a `Visible` attribute to `MenuItem`, evalute it in `refreshSystray` and call `Hide()` or `Show()` on the underlying `systray.MenuItem` accordingly.

d1ss0nanz:
### Description:
Add a `Visible` attribute to `MenuItem`, evalute it in `refreshSystray` and call `Hide()` or `Show()` on the underlying `systray.MenuItem` accordingly.

Fixes #3095

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/menu.go
+++ b/menu.go
@@ -57,7 +57,7 @@ type MenuItem struct {
 	Icon Resource
 	// Since: 2.2
 	Shortcut Shortcut
-	// Since: 2.2.2
+	// Since: 2.3
 	Hidden bool
 }
 
--- a/menu.go
+++ b/menu.go
@@ -57,11 +57,13 @@ type MenuItem struct {
 	Icon Resource
 	// Since: 2.2
 	Shortcut Shortcut
+	// Since: 2.2.2
+	Visible bool
 }
 
 // NewMenuItem creates a new menu item from the passed label and action parameters.
 func NewMenuItem(label string, action func()) *MenuItem {
-	return &MenuItem{Label: label, Action: action}
+	return &MenuItem{Label: label, Action: action, Visible: true}
 }
 
 // NewMenuItemSeparator creates a menu item that is to be used as a separator.
--- a/menu.go
+++ b/menu.go
@@ -58,12 +58,12 @@ type MenuItem struct {
 	// Since: 2.2
 	Shortcut Shortcut
 	// Since: 2.2.2
-	Visible bool
+	Hidden bool
 }
 
 // NewMenuItem creates a new menu item from the passed label and action parameters.
 func NewMenuItem(label string, action func()) *MenuItem {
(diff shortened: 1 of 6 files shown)


d1ss0nanz:
Good points.

Will work on it.

Bluebugs (commented):
I just wonder if it would be possible to have some test for this. Otherwise it looks pretty good.

menu.go
```diff
@@ -57,6 +57,8 @@ type MenuItem struct {
 	Icon Resource
 	// Since: 2.2
 	Shortcut Shortcut
+	// Since: 2.2.2
```
andydotxyz, line 60: We don't allow API changes in patch releases so this will need to be for 2.3.
This is also why we only use 2 digits for API level :)

andydotxyz (changes requested):
Looks good, but the API version is a bit too keen!

andydotxyz:
> I just wonder if it would be possible to have some test for this. Otherwise it looks pretty good.

This might be tricky as I don't think we have test infrastructure in place for system tray Fyne integration.
That said we could literally add a system tray icon and run the code - though it is starting to get intrusive for development. Perhaps it can be pushed back and added with other "full-test" flag that has been discussed before?
(Allowing all unit tests other than this tag to run without showing anything on screen...)

menu.go
d1ss0nanz: Done.

d1ss0nanz:
> I just wonder if it would be possible to have some test for this. Otherwise it looks pretty good.
Sure. Where should I put the tests?
There must be a bunch of related tests, where I just can extend the coverage, right?

andydotxyz:
I think your last commit was wrong @d1ss0nanz - it added a blank file to a `cmd/systray_demo` package. Let's not have more executables. The system tray is demoed in fyne_demo like everything else.

As for unit testing I expect that alongside the menu_test in drivers would be sufficient. We can't really test the systray aspects as that rendering is inside the `fyne.io/systray` package.

Jacalz:
Hi @d1ss0nanz. Do you plan on continuing this PR? It is currently out of date from develop so you will need to rebase on develop to get to pass tests. I believe the last message from Andrew should contain solutions to the remaining work before we can get this merged :)

andydotxyz:
It's been almost a year since last commit and no response in 2 months - closing.

PR #3101, opened 2022-06-26
Add new colours to the theme. Plus more flexible ThemedResource

### LINKED ISSUE #1861: `Danger`, `Warning`, `Success` alternatives for `Primary` colours
## The Ask

Perhaps widgets like the `ProgressBar` and `Button` could support other colours as defined within the theme?

We currently have the following possibilities:

- `Primary` being what we use now for HighImportance buttons and progressbar fill
- `Background` being what we use now for other buttons and progressbar bg

But frankly, I would like us to have these as well:

- `Error` for (frankly, everything, see below.)
- `Danger` for both buttons, card border and progressbar fill
- `Warning` for both buttons, card border and progressbar fill
- `Success` for both buttons and progressbar fill

These are important both for animation state targets when these can be animated to show the result of the action, as well as showing dangerous action and/or actions that must be taken with care/caution.

### `Error`

Straight out of Material Design Specification:

[image](https://material.io/design/color/the-color-system.html#color-theme-creation)
[They define a theme colour named `Error`. This image is also a link to the spec. page.]

For all other colours, also consider the section in the above link called:

**Additional colors for data visualization**

### `Danger`

This image comes straight from GitHub's Repository Settings page:

image
[Although to match with current `fyne` design, only the filled danger (top button) would be implemented.]

There are hundreds of examples of destructive actions, and they should be **very visually dissimilar** to standard non-destructive actions, from a user experience perspective.

### `Warning`

Can be used:

- as an Animation target
- anywhere the user should be especially cautious, or warned

Examples include:

- Indicating work in progress that is sensitive and therefore should not be aborted or interrupted (yes, you really can communicate all this with just a colour.)
- Card border around potentially-dangerous-but-not-destructive-actions (see `Danger`)

### `Success`

Can be used:

- as an Animation target
- as an icon tint/colour to show success, or any other indication of success.

Examples include:

- Background colour in a success flash notice (not in material design, because toast messages are sorta deprecated.)
- CI test status icon colour
- Background colour of an already faded-out (pressed and disabled) fades to green to show completion
- `+` buttons in CRUD interfaces, distinguished from pencil icons / up/down arrows, delete icons, etc
- visual separation inside any sort of dense toolbar (see previous example)

andydotxyz:
Added the new colour codes as suggested in the issue.
Skipped "Danger" as it is basically the same as Error. We could also have Pass/Fail but they too are just synonyms for Success and Error.

Fixes #1861

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/test/markup_renderer.go
+++ b/test/markup_renderer.go
@@ -137,7 +137,7 @@ func (r *markupRenderer) setResourceAttr(attrs map[string]*string, name string, 
 	}
 
 	var variant string
-	switch rsc.(type) {
+	switch t := rsc.(type) {
 	case *theme.DisabledResource:
 		variant = "disabled"
 	case *theme.ErrorThemedResource:
@@ -147,7 +147,10 @@ func (r *markupRenderer) setResourceAttr(attrs map[string]*string, name string, 
 	case *theme.PrimaryThemedResource:
 		variant = "primary"
 	case *theme.ThemedResource:
-		variant = "default"
+		variant = string(t.ColorName)
+		if variant == "" {
+			variant = "default"
+		}
 	default:
 		r.setStringAttr(attrs, name, rsc.Name())
 		return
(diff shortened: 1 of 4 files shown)


andydotxyz:
Colour preview:

colors

andydotxyz:
Auto-accepting following no feedback

PR #3108, opened 2022-06-29
implement fyne.Draggable interface and expose Mouse event hooks

MrWaggel:
### Description:
Added the fyne.Draggable interface to widget.Tree, also exposing the Mouse events.
My motivation is that users can expand the functionality of the Tree.

Example;

example_tree

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/tree.go
+++ b/widget/tree.go
@@ -22,14 +22,20 @@ type Tree struct {
 	BaseWidget
 	Root TreeNodeID
 
-	ChildUIDs      func(uid TreeNodeID) (c []TreeNodeID)                     `json:"-"` // Return a sorted slice of Children TreeNodeIDs for the given Node TreeNodeID
-	CreateNode     func(branch bool) (o fyne.CanvasObject)                   `json:"-"` // Return a CanvasObject that can represent a Branch (if branch is true), or a Leaf (if branch is false)
-	IsBranch       func(uid TreeNodeID) (ok bool)                            `json:"-"` // Return true if the given TreeNodeID represents a Branch
-	OnBranchClosed func(uid TreeNodeID)                                      `json:"-"` // Called when a Branch is closed
-	OnBranchOpened func(uid TreeNodeID)                                      `json:"-"` // Called when a Branch is opened
-	OnSelected     func(uid TreeNodeID)                                      `json:"-"` // Called when the Node with the given TreeNodeID is selected.
-	OnUnselected   func(uid TreeNodeID)                                      `json:"-"` // Called when the Node with the given TreeNodeID is unselected.
-	UpdateNode     func(uid TreeNodeID, branch bool, node fyne.CanvasObject) `json:"-"` // Called to update the given CanvasObject to represent the data at the given TreeNodeID
+	ChildUIDs         func(uid TreeNodeID) (c []TreeNodeID)                            `json:"-"` // Return a sorted slice of Children TreeNodeIDs for the given Node TreeNodeID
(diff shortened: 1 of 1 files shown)


MrWaggel:
Also any reason behind checking if anonymous function are nil
https://github.com/fyne-io/fyne/blob/043f68d10554b3f95a21452904b87694971b6cc5/widget/tree.go#L285

Insead of 
`if t.OnUnselected != nil {
    t.OnUnselected(uid)
}
`

Jacalz:
> Also any reason behind checking if anonymous function are nil?

We’d be crashing the whole application if we didn’t ;)

Jacalz (commented):
I don’t think this is the right solution. It might even already be doable. You just need to extend the tree widget and overwrite the functions. See https://developer.fyne.io/extend/extending-widgets for more information.

MrWaggel:
> > Also any reason behind checking if anonymous function are nil?
> 
> We’d be crashing the whole application if we didn’t. Nil pointers are scary. Calling a nil pointer function is a bit (simplified, thought not exactly) like trying to jump to a memory address that doesn't exist. The code will not be able to jump back to the call-site and the program thus crashes.

No I mean the way it is checked for nil.
https://github.com/fyne-io/fyne/blob/043f68d10554b3f95a21452904b87694971b6cc5/widget/tree.go#L100-L102

Instead of
```
if  t.OnBranchClosed != nil {
     t.OnBranchClosed(uid)
}
```

MrWaggel:
I know you can extend the widgets, but the Drag and mouse events are called back from the the treeNode structure which isn't exposed, and which are created by the Tree widget.

Jacalz:
> No I mean the way it is checked for nil.
> 
> https://github.com/fyne-io/fyne/blob/043f68d10554b3f95a21452904b87694971b6cc5/widget/tree.go#L100-L102
> 
> Instead of
> 
> ```go
> if  t.OnBranchClosed != nil {
>      t.OnBranchClosed(uid)
> }
> ```

I honestly have no idea. Less to write maybe? It doesn’t have any other benefit at all as far as I can see.

andydotxyz:
> > No I mean the way it is checked for nil.
> > https://github.com/fyne-io/fyne/blob/043f68d10554b3f95a21452904b87694971b6cc5/widget/tree.go#L100-L102
> > 
> > Instead of
> > ```go
> > if  t.OnBranchClosed != nil {
> >      t.OnBranchClosed(uid)
> > }
> > ```
> 
> I honestly have no idea. Less to write maybe? It doesn’t have any other benefit at all as far as I can see. IMHO, it’s actually less readable than your example.

Isn't the theory that it avoids the value becoming nil before the execution of the function? (race condition?)

Jacalz:
> Isn't the theory that it avoids the value becoming nil before the execution of the function? (race condition?)

I haven’t checked it but I don’t think it would help with that. As it is just a function pointer, it is probably just creating a new pointer to the same value and not helping with race conditions.

Bluebugs:
> > Isn't the theory that it avoids the value becoming nil before the execution of the function? (race condition?)
> 
> I haven’t checked it but I don’t think it would help with that. As it is just a function pointer, it is probably just creating a new pointer to the same value and not helping with race conditions.
> 
> EDIT: I might not have thought that through. It's not the function that changes value but the location that it points to. Yes, you might be right about that :)

It is definitively preventing a race condition. The power of copy! :-)

Issue #1090, opened 2020-06-09
draw rounded edges in Rectangle


MJacred:
### Is your feature request related to a problem? Please describe:
For better readability and a more diverse collection of shapes. Different Components in Material Design have them, like [Material buttons](https://material.io/components/buttons), tags/[chips](https://material.io/components/chips), [cards](https://material.io/components/cards), ...

### Is it possible to construct a solution with the existing API?
desired outcome not possible at this time

### Describe the solution you'd like to see:
define in `canvas/Rectangle.go Rectangle struct` additional members: cornerX1, cornerX2, cornerY1, cornerY2,  
then define defaults in `theme/theme.go` per Widget: Makes sense for Button, Entry, Select, PopUp, TabContainer Button, ProgressBar, Spinner (wip). And other widgets which will probably come later: [Frame](https://developer.gnome.org/gtk3/stable/GtkFrame.html), [Switch](https://developer.gnome.org/gtk3/stable/GtkSwitch.html).  
Maybe let the user overwrite them in code?

see border radius in CSS

andydotxyz:
I’m a little torn here on whether the individual corner configuration is needed? Adding “CornerRadius” to Rectangle could be a nice addition. However if it was to be 4 new fields then it might hint towards a RoundedRectangle type that offers more flexibility? 

Just some thoughts really - no conclusion other than it sounds like a good idea in principle.

MJacred:
Hm.. one `CornerRadius` probably suffices. I proposed the 4 for maximum configuration possibility.

I'd have no problem with an extra `RoundedRectangle`, which embeds `Rectangle`. So that both work harmoniously together, we'd probably need a `Rect` interface which both rectangle structs implement?

fpabl0:
@andydotxyz Isn't it possible to just add a new field to `canvas.Rectangle` called `BorderRadius`? Then the canvas.Rectangle could be any kind of rectangle rounded or not. BorderRadius could be private and defined from Theme API. I am not sure how this can be implemented but looking at `internal/painter/draw.go` the method `DrawRectangle` could use `rasterx.AddRoundRect` instead of `rasterx.AddRect` to enable border radius. This feature would be very helpful to make the current widgets more beautiful :)

andydotxyz:
To add new canvas primitives (or features to them) we need to add in the software and OpenGL renderers.
[...]

PR #3121, opened 2022-07-04
Add rounded edges to Rectangle

renlite:
### Description:

#1090
Implementation of a (Round)Rectangle as primitive Open GLObject. 

- The property "fyne.RectangleRadius" extends canvas.Rectangle struct.
- For the left and right side a different radius can be defined.
- For the left and right side different segments can be defined. Default is 8 segments for one rounded edge of a rectange. The higher the value the more triangles are calculated and rendered.
- A rectangle with stroke is rendered in one communication/stream to the GPU. A shape with opaque colors and stroke can be rendered with composition.
- If radius is 0 a normal rectangle will be drawn.
- Only required slices are rendered, so I hope the performance will be ok.

Limitation: Shapes with stroke and only one segment don't work properly. As a work around for shapes with opaque colors you could use composition. As this is not the main use for this canvas object I don't plan to implement it.

### Checklist:

- [x] Tests included.
- [x] Formatter run with no errors.
- [x] Lint run with no errors.
- [x] Tests all pass.

#### Where applicable:

- [X] GL ES and software-fallback not implemented
- [X] Please check: rendering (method: drawRectangle https://github.com/renlite/fyne/blob/c24f9a641cf0bf39f2e0c0da1694aed85132c239/internal/painter/gl/draw.go#L110) is called twice
- [X] Old method: drawRectangle was commented out
- [X] See last discussion #1090
- [X] Example included. (https://github.com/renlite/fyne/blob/develop/example/flexrect.go)

### DIFF
--- a/internal/painter/gl/texture.go
+++ b/internal/painter/gl/texture.go
@@ -142,26 +142,6 @@ func (p *painter) newGlRasterTexture(obj fyne.CanvasObject) Texture {
 	return p.imgToTexture(rast.Generator(int(width), int(height)), rast.ScaleMode)
 }
 
-/*
-func (p *painter) newGlRectTexture(obj fyne.CanvasObject) Texture {
-	rect := obj.(*canvas.Rectangle)
-	if rect.StrokeColor != nil && rect.StrokeWidth > 0 {
-		return p.newGlStrokedRectTexture(rect)
-	}
-	if rect.FillColor == nil {
-		return noTexture
-	}
-	return p.imgToTexture(image.NewUniform(rect.FillColor), canvas.ImageScaleSmooth)
-}
-
-func (p *painter) newGlStrokedRectTexture(obj fyne.CanvasObject) Texture {
-	rect := obj.(*canvas.Rectangle)
-	raw := paint.DrawRectangle(rect, paint.VectorPad(rect), p.textureScale)
-
-	return p.imgToTexture(raw, canvas.ImageScaleSmooth)
-}
-*/
-
 func (p *painter) newGlTextTexture(obj fyne.CanvasObject) Texture {
 	text := obj.(*canvas.Text)
 	color := text.Color


Jacalz:
Thanks for opening the PR. Would you mind changing the title name to be more specific about what the PR is doing (instead of which issue it fixes)? :)

andydotxyz:
> [x] Tests included. (https://github.com/renlite/fyne/blob/develop/example/flexrect.go)

When we say "tests included" we mean unit tests, not example code.
Please don't remove the other checks from the PR template - the "where applicable" can be removed, but the rest is required.

renlite:
> Amazing thanks!
> 
> Can you please put example code into the `fyne_demo` cmd rather than creating a new package for this PR?

Sorry, I don't know how to do that. Could you give me a link where to put the example code.

renlite:
> 

Which lint tool should be used if any?

Jacalz:
> > 
> 
> 
> 
> Which lint tool should be used if any?

The ones used in the static analysis action should always be used.

Bluebugs:
This is pretty cool. Looking at the code, I realized that maybe it will lead to a faster frame rate when using the web backend. Sadly, something is broken there with this patch. I think it is a bug in the gl-js layer and not in this patch. I am going to look more into it, but please do not land this patch until confirmed that it work for the web backend properly.

Bluebugs:
I figured out that there is no GL ES variant of the shader which lead all other backend to fail. I kind of half fixed it locally, but then I see errors in the web console:
```
WebGL warning: drawArraysInstanced: Vertex fetch requires 10, but attribs only supply 6.
WebGL warning: drawArraysInstanced: Vertex attrib array 2 is enabled but has no buffer bound.
```

renlite:
> I figured out that there is no GL ES variant of the shader which lead all other backend to fail. I kind of half fixed it locally, but then I see errors in the web console:
> 
> ```
> WebGL warning: drawArraysInstanced: Vertex fetch requires 10, but attribs only supply 6.
> WebGL warning: drawArraysInstanced: Vertex attrib array 2 is enabled but has no buffer bound.
> ```

I havn't looked at GL ES so far. Shaders for GL ES are not defined and built at the moment nor implemented in eg. gl_es.go->Init(). My next steps are to split the method flexRectCoords() in draw.go to fullfill the lint checks. I hope the GL 110 and GL ES are compatible. I think in that case only the calling stack for GL ES must be changed.

Bluebugs (approved):
This is working well for me and looks good. Really great contribution!

andydotxyz (changes requested):
I ran fyne_demo on a macOS computer and there was lots of flickering rectangles.
Various UI elements (buttons, tab bar underline) does not show by default but appears when you hover.
Not sure how it happened, but certainly this PR introduced that.

internal/painter/gl/draw.go
```diff
@@ -97,13 +97,65 @@ func (p *painter) drawRaster(img *canvas.Raster, pos fyne.Position, frame fyne.S
...
+	if aS == 0 {
+		p.ctx.Uniform4f(strokeColorUniform, 0, 0, 0, 0)
+	} else {
+		alphaS := float32(aS)
+		colF := []float32{float32(rS) / alphaS, float32(gS) / alphaS, float32(bS) / alphaS, alphaS / 0xffff}
+		p.ctx.Uniform4f(strokeColorUniform, colF[0], colF[1], colF[2], colF[3])
+	}
+	p.logError()
+
+	println("drawRectangle: ", len(points), triangleXYPoints)
```
andydotxyz, line 149: Please remove this debug

Bluebugs (changes requested):
Seems there are issue left with tests.

andydotxyz:
Probably worth noting that mobile doesn't compile... see the error in the mobile test runs:
```
Error: internal/painter/gl/gl_gomobile.go:194:13: c.glContext.DisableVertexAttribArray undefined (type gl.Context has no field or method DisableVertexAttribArray)
```

You can check this locally by using `-tags mobile`.

renlite:
FIle: ```gl_gomobile.go```
The context of GL Mobile has an EnableVertexAttributeArray function but no DisableVertexAttributeArray function. The gl_core.gl does not need to disable the vertex. Is there a wrapper function missing for mobileContext? It could be that the mobile GL doesn't need the DisableVertexAttributeArray to work as expected.
```go
func (c *mobileContext) EnableVertexAttribArray(attribute Attribute) {
	c.glContext.EnableVertexAttribArray(gl.Attrib(attribute))
}

func (c *mobileContext) DisableVertexAttribArray(attribute Attribute) {
}
```
I tested fyne_demo:
- [x] Windows:
- [x] Web (wasm) (event performance)
- [ ] Linux ?
- [ ] macOS
- [ ] mobile Android ?
- [ ] mobile iOS ?

I don't know why the demo has flickering on macOS which works on Windows and Web wasm. What GL version does macOS support: Open GL or GL ES?  macOS is main using Metal API.
Could you please test it on Linux Desktop and Android.

andydotxyz:
MacOS uses the same OpenGL code as Windows and Linux, no Metal.

ErrorNoInternet:
Seems like I only see flickering when I use my dedicated GPU (Arch Linux, KDE Plasma, Nvidia GT 720M)
Using integrated graphics is fine though

https://user-images.githubusercontent.com/61157623/187200765-fafa90a1-d73f-47f5-a6ca-b7a1b65fea92.mp4

renlite:
> Seems like I only see flickering when I use my dedicated GPU (Arch Linux, KDE Plasma, Nvidia GT 720M) Using integrated graphics is fine though
> 
>  fyne_demo.mp4

Thanks for the tests and the mp4. I see in the terminal a warning: C7050 'color' might be used before being initialized. This warning happens only with the Nvidia graphic card. I have an integrated Intel HD 5500 graphic card on Windows and it works. I will have a look at the fragment shader but at the moment I have no idea whrere to start, to find the issue. I don't think the warning is the reason.
- Could it be an Open GL driver problem?
- Or, a FrameBufferObject Problem?

andydotxyz:
It works well on my Dell XPS running Debian/FyshOS :)

Bluebugs:
I wonder if it isn't a float problem. The shader doesn't initialize color with a default value and rely instead on comparing switch_var to a float. Maybe those float are not always the value we expect due to float approximate value? For example if the float where 16 bits instead of 32? I don't know, just throwing an idea here. I wonder if there is an online simulator for shader somewhere.

renlite:
> I wonder if it isn't a float problem. The shader doesn't initialize color with a default value and rely instead on comparing switch_var to a float. Maybe those float are not always the value we expect due to float approximate value? For example if the float where 16 bits instead of 32? I don't know, just throwing an idea here. I wonder if there is an online simulator for shader somewhere.

Thank you for the direction every idea is welcome. It seems some rectangles are loosing the color. I will try to init the color variable a fix red color, to see if the missing rectangles will be painted in red.

renlite:
> It works well on my Dell XPS running Debian/FyshOS :) WIndows is OK too. This is an Intel Iris X graphics card

Thanks for the information, this is positive.

andydotxyz:
It is worth saying thanks for all the hard work. When this works it works really well. Just a few things to figure out and it will rock!

andydotxyz:
Android (emulator) now compiles and runs well, thanks

Jacalz (changes requested):
Thanks for this. There still seems to be static analysis checks failing and also lowered coverage causing the tests to fail.

internal/painter/gl/painter.go
```diff
@@ -11,10 +11,12 @@ import (
...
-	"line_es":   {shaderLineesVert.StaticContent, shaderLineesFrag.StaticContent},
-	"simple":    {shaderSimpleVert.StaticContent, shaderSimpleFrag.StaticContent},
-	"simple_es": {shaderSimpleesVert.StaticContent, shaderSimpleesFrag.StaticContent},
+	"line":         {shaderLineVert.StaticContent, shaderLineFrag.StaticContent},
+	"line_es":      {shaderLineesVert.StaticContent, shaderLineesFrag.StaticContent},
+	"simple":       {shaderSimpleVert.StaticContent, shaderSimpleFrag.StaticContent},
+	"simple_es":    {shaderSimpleesVert.StaticContent, shaderSimpleesFrag.StaticContent},
+	"rectangle":    {shaderRectangleVert.StaticContent, shaderRectangleFrag.StaticContent},
+	"rectangle_es": {shaderRectangleesVert.StaticContent, shaderRectangleesFrag.StaticContent},
 }
```
Jacalz, line 20: You might also want to change this to a switch statement in line with https://github.com/fyne-io/fyne/pull/3244. I won't force you but it seems better than to have me wait for your PR, or not wait and create merge conflicts here :)

renlite:
> Thanks for this. There still seems to be static analysis checks failing and also lowered coverage causing the tests to fail.

I think the static analysis checks failing belongs to ```texture.go``` - two unused func.. It was merged with fork sync.

Jacalz:
> > Thanks for this. There still seems to be static analysis checks failing and also lowered coverage causing the tests to fail.
> 
> I think the static analysis checks failing belongs to `texture.go` - two unused func.. It was merged with fork sync.

Exactly. If they are unused and no longer severe a purpose, then they should of course be removed.

internal/painter/gl/painter.go
renlite: Please have a look at the coding.

renlite:
Could somebody please test the new fragment shader and show the result on MacOS or on NVIDIA? Is there still flickering or are some fragments painted in red?
Is there still the warning: C7050 'color' might be used before being initialized?

andydotxyz:
It works great on my iMac, but not the macBook Pro. Video attached:


https://user-images.githubusercontent.com/294436/188606431-b6d17433-cd81-4ef6-899e-358093deca1b.mov

ErrorNoInternet:
Still flickering for me (NVIDIA)...
The color warning is gone though

renlite:
I activated DisableVertexAttribArray() in gl_core.go. On Windows it works as previously.
@andydotxyz  and @ErrorNoInternet could you please try it again. Thank you.

renlite:
> > > Thanks for this. There still seems to be static analysis checks failing and also lowered coverage causing the tests to fail.
> > 
> > 
> > I think the static analysis checks failing belongs to `texture.go` - two unused func.. It was merged with fork sync.
> 
> Exactly. If they are unused and no longer serve a purpose, then they should of course be removed.

I did not change texture.go, the code failing static analysis checks is not from me.

Jacalz:
> > > > Thanks for this. There still seems to be static analysis checks failing and also lowered coverage causing the tests to fail.
> 
> > > 
> 
> > > 
> 
> > > I think the static analysis checks failing belongs to `texture.go` - two unused func.. It was merged with fork sync.
> 
> > 
> 
> > Exactly. If they are unused and no longer serve a purpose, then they should of course be removed.
> 
> 
> 
> I did not change texture.go, the code failing static analysis checks is not from me.

It is caused by your changes. The functions in question are unused because you you removed the places that they were used. If the functions are unused and no longer serve a purpose, you should remove them.

Jacalz:
The unused functions should be removed and not commented out. We have a principle to not keep commented out code.

ErrorNoInternet:
> I activated DisableVertexAttribArray() in gl_core.go. On Windows it works as previously.
> @andydotxyz  and @ErrorNoInternet could you please try it again. Thank you.

Still flickering

Bluebugs:
Just a random though here, comparison with float is not a precise thing. It might be a difference in precision that make those test not pass (maybe). See https://stackoverflow.com/questions/54520743/how-to-compare-float-in-glsl-scripts .

Bluebugs:
I may have a proposition that might be a lot of change to this PR, but might be actually resulting in simpler and easier to understand code. After reading https://www.warp.dev/blog/how-to-draw-styled-rectangles-using-the-gpu-and-metal and https://stackoverflow.com/questions/59197671/glsl-rounded-rectangle-with-variable-border. I started playing with shader toy and got the following shader this evening: https://www.shadertoy.com/view/NldBDH . The main change is that the geometry we need to  push is simply the quad that represent the rectangle and the fragment shader compute the distance from the center to the edge of the rectangle to take appropriate decision. Let me know what you think of this approach? But I believe this has more chance to work on all GPU.

Bluebugs:
I actually ended up adding linear gradient support too.

renlite:
@Bluebugs 
I forked the develop branche and implemented my last RoundRect coding to start some tests, but now I get following cgo gl error. I used Go1.17 and Go1.19, but the same error (Win10). It is a runtime error. The windows appears shortly before gl throws following exception:
Do you have an idea?

```
Exception 0xc0000005 0x0 0xc 0x7ffe74c44fa0
PC=0x7ffe74c44fa0
signal arrived during external code execution

runtime.cgocall(0x7ff685188720, 0xc000831740)
        c:/go/src/runtime/cgocall.go:158 +0x4a fp=0xc000831718 sp=0xc0008316e0 pc=0x7ff684c8486a
github.com/go-gl/gl/v3.2-core/gl._Cfunc_glowDrawArrays(0x7ffea5b59370, 0x4, 0x0, 0x6)
        _cgo_gotypes.go:6468 +0x52 fp=0xc000831740 sp=0xc000831718 pc=0x7ff68504f032
github.com/go-gl/gl/v3.2-core/gl.DrawArrays(...)
        C:/_project/renlite/fyne/vendor/github.com/go-gl/gl/v3.2-core/gl/package.go:9566
fyne.io/fyne/v2/internal/painter/gl.(*coreContext).DrawArrays(0x7ff685a7d170?, 0x4?, 0x0?, 0x6?)
        C:/_project/renlite/fyne/internal/painter/gl/gl_core.go:156 +0x30 fp=0xc000831768 sp=0xc000831740 pc=0x7ff6850706b0
fyne.io/fyne/v2/internal/painter/gl.(*painter).drawRectangle(0xc0009fe000, 0xc0004180e0, {0x8318d8?, 0xc0?}, {0x831870?, 0xc0?})
        C:/_project/renlite/fyne/internal/painter/gl/draw.go:140 +0x476 fp=0xc000831800 sp=0xc000831768 pc=0x7ff6850687b6
fyne.io/fyne/v2/internal/painter/gl.(*painter).drawObject(0xc00014a540?, {0x7ff6856fd6a0?, 0xc0004180e0?}, {0x84c8c19d?, 0x7ff6?}, {0x855472a0?, 0x7ff6?})
        C:/_project/renlite/fyne/internal/painter/gl/draw.go:86 +0x16d fp=0xc000831850 sp=0xc000831800 pc=0x7ff68506814d
fyne.io/fyne/v2/internal/painter/gl.(*painter).Paint(0x7ff684c8e45f?, {0x7ff6856fd6a0, 0xc0004180e0}, {0x84c8c19d?, 0x7ff6?}, {0x831908?, 0xc0?})
        C:/_project/renlite/fyne/internal/painter/gl/painter.go:88 +0x74 fp=0xc0008318a0 sp=0xc000831850 pc=0x7ff685071854
fyne.io/fyne/v2/internal/driver/glfw.(*glCanvas).paint.func1(0x7ff684ef7df1?, {0x84ef847c?, 0x7ff6?})
        C:/_project/renlite/fyne/internal/driver/glfw/canvas.go:279 +0x102 fp=0xc000831908 sp=0xc0008318a0 pc=0x7ff68515a562
fyne.io/fyne/v2/internal/driver/common.(*Canvas).walkTree.func1({0x7ff6856fd6a0?, 0xc0004180e0}, {0x4180e0?, 0xc0?}, {0xc0000?, 0xc0?}, {0x1757d81?, 0xe?})
        C:/_project/renlite/fyne/internal/driver/common/canvas.go:438 +0x277 fp=0xc000831960 sp=0xc000831908 pc=0x7ff685075757
fyne.io/fyne/v2/internal/driver.walkObjectTree({0x7ff6856fd6a0, 0xc0004180e0}, 0x0, {0x7ff6856fdac0, 0xc0000c0000}, {0x0?, 0x418da000?}, {0x42af6800?, 0x42be5000?}, {0x40c00000?, ...}, ...)
        C:/_project/renlite/fyne/internal/driver/util.go:169 +0x29b fp=0xc000831a28 sp=0xc000831960 pc=0x7ff684eef09b
fyne.io/fyne/v2/internal/driver.walkObjectTree.func1(...)
        C:/_project/renlite/fyne/internal/driver/util.go:176
fyne.io/fyne/v2/internal/driver.walkObjectTree({0x7ff6856fdac0, 0xc0000c0000}, 0x0, {0x7ff6856fd400, 0xc0000ac050}, {0x0?, 0x0?}, {0x0?, 0x0?}, {0x0?, ...}, ...)
        C:/_project/renlite/fyne/internal/driver/util.go:190 +0x41d fp=0xc000831af0 sp=0xc000831a28 pc=0x7ff684eef21d
fyne.io/fyne/v2/internal/driver.walkObjectTree.func1(...)
        C:/_project/renlite/fyne/internal/driver/util.go:176
fyne.io/fyne/v2/internal/driver.walkObjectTree({0x7ff6856fd400, 0xc0000ac050}, 0x0, {0x0, 0x0}, {0x84cc3731?, 0x7ff6?}, {0x0?, 0x0?}, {0x831c38?, ...}, ...)
        C:/_project/renlite/fyne/internal/driver/util.go:190 +0x41d fp=0xc000831bb8 sp=0xc000831af0 pc=0x7ff684eef21d
fyne.io/fyne/v2/internal/driver.WalkVisibleObjectTree({0x7ff6856fd400?, 0xc0000ac050?}, 0xc0000841a0?, 0x200000003?)
        C:/_project/renlite/fyne/internal/driver/util.go:134 +0x4c fp=0xc000831c20 sp=0xc000831bb8 pc=0x7ff684eeedac
fyne.io/fyne/v2/internal/driver/common.(*Canvas).walkTree(0xc000000000?, 0xc000089040, 0xc000831d80, 0xc000831d50)
        C:/_project/renlite/fyne/internal/driver/common/canvas.go:459 +0x165 fp=0xc000831cd8 sp=0xc000831c20 pc=0x7ff685075345
fyne.io/fyne/v2/internal/driver/common.(*Canvas).WalkTrees(0xc0000ba000, 0xc000831d50?, 0x7ff685159ca6?)
        C:/_project/renlite/fyne/internal/driver/common/canvas.go:371 +0x3c fp=0xc000831d30 sp=0xc000831cd8 pc=0x7ff685074e9c
fyne.io/fyne/v2/internal/driver/glfw.(*glCanvas).paint(0xc0000ba000, {0x0?, 0x0?})
        C:/_project/renlite/fyne/internal/driver/glfw/canvas.go:291 +0xfe fp=0xc000831db0 sp=0xc000831d30 pc=0x7ff68515a33e
fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).repaintWindow.func1()
        C:/_project/renlite/fyne/internal/driver/glfw/loop.go:203 +0x8f fp=0xc000831e00 sp=0xc000831db0 pc=0x7ff68515d7af
fyne.io/fyne/v2/internal/driver/glfw.(*window).RunWithContext(0xc0004189a0?, 0xc0000b3e30)
        C:/_project/renlite/fyne/internal/driver/glfw/window.go:903 +0x4f fp=0xc000831e20 sp=0xc000831e00 pc=0x7ff685165c4f
fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).repaintWindow(0xc0000b3e98?, 0xc00005c520?)
        C:/_project/renlite/fyne/internal/driver/glfw/loop.go:194 +0x4a fp=0xc000831e58 sp=0xc000831e20 pc=0x7ff68515d6ea
fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).drawSingleFrame(0xc0000b3f90?)
        C:/_project/renlite/fyne/internal/driver/glfw/loop.go:102 +0x1cb fp=0xc000831ef8 sp=0xc000831e58 pc=0x7ff68515cecb
fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).startDrawThread.func1()
        C:/_project/renlite/fyne/internal/driver/glfw/loop.go:250 +0x1a6 fp=0xc000831fe0 sp=0xc000831ef8 pc=0x7ff68515dc26
runtime.goexit()
        c:/go/src/runtime/asm_amd64.s:1594 +0x1 fp=0xc000831fe8 sp=0xc000831fe0 pc=0x7ff684ce6841
created by fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).startDrawThread
        C:/_project/renlite/fyne/internal/driver/glfw/loop.go:226 +0xea

goroutine 1 [select, locked to thread]:
runtime.gopark(0xc000835ea0?, 0x3?, 0xa0?, 0x5c?, 0xc000835e32?)
        c:/go/src/runtime/proc.go:363 +0xd6 fp=0xc000835c58 sp=0xc000835c38 pc=0x7ff684cbbc56
runtime.selectgo(0xc000835ea0, 0xc000835e2c, 0x7ff6851322ee?, 0x0, 0x7ff6856fac68?, 0x1)
        c:/go/src/runtime/select.go:328 +0x7dc fp=0xc000835d98 sp=0xc000835c58 pc=0x7ff684cc9ffc
fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).runGL(0xc0004189a0)
        C:/_project/renlite/fyne/internal/driver/glfw/loop.go:121 +0x19f fp=0xc000835ee0 sp=0xc000835d98 pc=0x7ff68515d15f
fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).Run(0xc0004189a0)
        C:/_project/renlite/fyne/internal/driver/glfw/driver.go:168 +0x75 fp=0xc000835f00 sp=0xc000835ee0 pc=0x7ff68515b5d5
fyne.io/fyne/v2/internal/driver/glfw.(*window).ShowAndRun(0xc0000b8000)
        C:/_project/renlite/fyne/internal/driver/glfw/window.go:230 +0x2c fp=0xc000835f18 sp=0xc000835f00 pc=0x7ff685160dac
main.main()
        C:/_project/renlite/fyne/cmd/hello/main.go:22 +0x1cd fp=0xc000835f80 sp=0xc000835f18 pc=0x7ff6851705ad
runtime.main()
        c:/go/src/runtime/proc.go:250 +0x1fe fp=0xc000835fe0 sp=0xc000835f80 pc=0x7ff684cbb8be
runtime.goexit()
        c:/go/src/runtime/asm_amd64.s:1594 +0x1 fp=0xc000835fe8 sp=0xc000835fe0 pc=0x7ff684ce6841

goroutine 2 [force gc (idle)]:
...
```

renlite:
Thank you for your reply. I have looked at this, but did not find a difference. My code is unchanged since it worked. I will check it again.
It seems the c (cgo) interop to GL is not that perfect and there could be a timelag in the communication between GC based goroutines and c code.
I was thingking about the reason why GPUs with more power cause flickering and GPUs with lower performance work. Maybe Go communication is not fast enough to send the complete vertex stream to the GPU and so the GPU paints only the fragments it receives. Or there is a lock or GC problem between the goroutines and the GPU.
I would like to disable the round corners and antializing for the primitive rectangle, so only a simple rectangle having only 6 points (2 triangles) will be send and rendered. Theoretically this should work on all graphic cards as this shape would be less complex than a antialized line. If this works than I would activate step by step more rectangle features.

renlite:
It works now. I overlooked initialization of the glcontext.

https://github.com/renlite/fyne

My understanding of fyne rendering process so far:

1) fyne uses one Canvas to render all visible Objects (fyne: CanvasObejects) on screen.
2) The ``` func runGL(...)``` loop at any point calls ```glDriver.repaintWindow(...)``` and later ```canvas.Paint(canvas.Size)``` is invoked. Canvas.Size is the size of the rendering Window. Then ```Canvas.WalkTree(...)``` is called and in the ```driver.WalkVisibleObjectTree(tree.root.obj, bc, ac)``` happens the rendering of the CanvasObjects like Line, Rectangle, Texture.
3) I did not analyze further but what I noticed that when a repaint of a window is necessary, all the Shapes (CanvasObjects) of a Window are new calculated and are sent to the GPU. It does not matter whether the CanvasObject changed or not, the calculation of the Points of all CanvasObject is done again and again. I have implemented a ```println()``` of the calculated Points for the Line and the Rectangle in the file: ```/internal/painter/gl/draw.go```.  Now you can see, how many Points are re-calculated and how often. Line is a const of 6 Points and Rectangle depends on the use. A rectangle without stroke and without radius needs the same 6 Points as a Line, but with less calculation because an antialiazing is not necessary. This means if you use the new Rectangle in the simplest form Rectangle should be faster than a Line.
4) As I mentioned for every event the entire window is new calculated/rendered. This is especially in the case of animation not optimal. In fyne the animation causes a repaint of the window for every step. I found out that in the simplest example ```/cmd/hello/main.go``` after a click on the button the window and the position and points of the rectangles are recalculated about 13 times even though the rectangles coords of the button did not change.
image
- Could you please check the example on the GPUs with flickering and inform me of the result?
5) This behavior that all CanvasObjects of a window are recalculated/rerendered although only one widget (the animated Button) has changed is to see in the example ```/cmd/shapes/main.go```. It seems the Window is marked as dirty and not the single CanvasObject. At the moment rendering seems not to offer the state of a CanvasObject and the possibility to cache and reuse the calculated Points.
This example uses different rectangles with Points between 6 and 714. But as long as the Button is not clicked and animation started a repainting or rerendering should not happen.
image
- Could you please check the example on the GPUs with flickering and inform me of the result?
6) As I reported erlier for the first (initial) draw of the Window all CanvasObjects are caculated twice. I don't understand why? Maybe it has to do with the functions ```func beforeChildren(...)``` and ```func afterChildren(...)``` in the ```func WalkTree(...)```.  It could be that the rendering process needs to walk twice over the CanvasObjectTree. In this case a caching of the Points would help too.
7) The example ```/cmd/fyne_demo/main.go``` is special because the calculation of the CanvasObjects repeats constantly after once the tab "Animation" is selected. When the animation is stopped or the tab is leaved the repainting is continuing. Maybe the hidden animations in the tabs "Canvas" and "Animation" causes the permanent recalculation. I suppose this and the missing caching could be the reason for flickering on some GPUs. I will prepare the same example but wihout animation, to see what happens.
image
- Could you please check the recalculation of this example? If you start with the "Welcome" tab and stay on this tab, no recalculation and no flickering should be. Try to change the theme "Dark" or "Light".  Could you please inform what happens?

Solution:(?)
- To avoid the time consuming recalculation of the coords the CanvasObject or the primitive GLObjects (Line and Rectangle) could have a property of type array with the cached coordinates. If the CanvasObject is dirty than coords_cache will be deleted from the rendering process (WalkTree) and the ```func drawRectangle(...)``` will check if cache exists and reuse the data or calculate new coords.
I don't know if the GPU offers a cache with an ID for repainting, so the buffer on to the GPU could be reused too?
- Rethink the animation process/rendering process? Animation could get an different widget type with its own canvas on the screen. I think this would lead to more redesign of the core rendering process.
- To offer a setting for the Button with or without Ripple Effect. Maybe changing the background color on click is for the user as good as an animation on the button and maybe it offers a better userbility.

andydotxyz:
This doesn't sound right. A CanvasObject is not re-rendered until Refresh is called. However if you call Refresh at the root it will (except inside scroll?) invalidate all the children as well.
A move animation for example should not invalidate and rendering, but instead will reposition cached textures.

A window being marked dirty is not causing a complete re-render, that is just marking that we should paint another frame (instead of freezing the content).
If your example calls Content().Refresh() regularly then I could imagine seeing the behaviour you describe.

It is possible that vertex calculations are re-done for any dirty frame because we don't track when an object has moved (for an additional "position refresh" consideration) though there is an issue related to this whereby move should produce a smaller level refresh (and therefore avoid the re-Calc happening if this is not registered).

Much of what looks like re-rendering in your debugging should actually be hitting texture caches and not actually calling the draw methods. It could be more efficient but it certainly should not be triggering complete redraws for every frame.

Bluebugs:
I think you are both talking about something different. First, with OpenGL on all OS except Linux (for both X and Wayland), there is no partial rendering. You have to render the full screen for every frame.

On Linux, you have an extension, buffer age and swap buffer with damage, that allow you to only update part of the window that changed and notify the compositor of the change that was made. This is pretty good for 2D application where there is only one blinking cursor and nothing else changing. This kind of improvement is mostly about energy saving and less about performance as you would still need to figure out how to get to 60 fps for a full window update.

As you don't have this infrastructure on any other OS, this optimization is not implemented in Fyne at the moment. So the problematic is we need to always do a full screen redraw, whatever changed in the window. What slow us down in maintaining 60 fps are in order, computing things on the CPU that will require large upload to the GPU (image, glyph), then large redraw on the GPU (things like scrolling) and finally vertex upload (even if computed, they are small amount of computation and are not consuming that much bandwidth). In the case of Fyne, the upload of image and glyph are what a Refresh can trigger (there are cache in place to limit those). In case of a scroll, as long as the object being moved are not getting Refresh, they should end up in a buffer and this would avoid some cost for scrolling. Each of the vertex are reuploaded for everything that changed on screen.

The rendering to an intermediate buffer is an optimization for the case where you do redraw the same thing a lot on screen, as you can just reuse it and avoid geometry and stuff to be pushed again. But there is a cost in doing so, you need twice the bandwidth as you would if thing were rendered once. So for fast scrolling things or animation at 60fps, having intermediate buffer can actually become a penalty and slow things down. It is a tradeoff, a way to simplify the way rendering is done and give good enough performance for the general purpose case which is what Fyne is currently optimized for and what @andydotxyz is describing above.

Now, I think that OpenGL is also not the best API to do the kind of optimization @renlite seems to think of. Uploading all the vertex information once and only modifying the one that need to be, seems a pain to do in GL. I would rather push that kind of optimization to when we focus on doing a brand new painter with something like Vulkan or WebGPU. We could in that case design the painter to actually do push all vertex once and then send change command executed on the GPU side a bit like game engine do. Having the ability to do delta like that would mean supporting easily the Linux extension mentioned above. Along with an estimator of the complexity of a layout and the amount of animation going on in it, it would also be possible to have a much more improved policy for when to do buffering.

All of that describe a next generation painter, and before going there, it would be good to have all the shaders and primitive accelerated by the GPU. I think it would be very useful to have tracing in place that inform us of the time it takes to render a frame and the latency between a change hitting Refresh and the time it took to get to screen. Maybe even write a benchmark application that people can run and give us a report of hardware performance/OS/architecture in use. I would advocate for the work in the this paragraph first, before looking at trying to work on next gen painter as I am sure they would reveal low hanging fruits.

renlite:
Please start https://github.com/renlite/fyne/blob/develop/cmd/shapes/main.go and watch the output in the terminal.

In the https://github.com/renlite/fyne/blob/develop/internal/painter/gl/draw.go there are two functions for primitive GLObject drawing or calculation of coords for later upload of vertex to OpenGL.

```go
func (p *painter) drawLine(line *canvas.Line, pos fyne.Position, frame fyne.Size) {
	if line.StrokeColor == color.Transparent || line.StrokeColor == nil || line.StrokeWidth == 0 {
		return
	}
	points, halfWidth, feather := p.lineCoords(pos, line.Position1, line.Position2, line.StrokeWidth, 0.5, frame)

	// (...)
	println("Line_6 ")
	p.ctx.DrawArrays(triangles, 0, 6)
}


func (p *painter) drawRectangle(rect *canvas.Rectangle, pos fyne.Position, frame fyne.Size) {
	points := p.flexRectCoords(pos, rect, 0.5, frame)
	p.ctx.UseProgram(p.rectangleProgram)

	// (...)
	print("Rect_", triangleXYPoints, " ")
        p.ctx.DrawArrays(triangles, 0, triangleXYPoints)
}
```` 
These functions calculates the coordinates of the CanvasObjects and make the upload of the vertex to the GPU. The intensive calculation is done in the lines ```points, halfWidth, feather := p.lineCoords(pos, line.Position1, line.Position2, line.StrokeWidth, 0.5, frame)  points := p.flexRectCoords(pos, rect, 0.5, frame)```.  You will see in the output that these functions are called for CanvasObject where the coordinates did not change. Eg. after a Button click the animation of the click effect causes a new call of these functions and a new calculation of all CanvasObjects on the screen, not only of the rectangles that belong to the button. This means the heavy weight calculation of the vertex coords is IMO unnecessary called.
Would the in_param (CanvasObject) ```canvas.Line``` and ```canvas.Rectangele``` own a property of eg. ```CoordsCache``` than it would be possible to decide in these functions if a new calculation of the coords have to be done or the unchanged cache could be used. The upload of the CanvasObject would be unchanged. So the FW would have to manage the cache in the WalkTree. Did the coords change (CanvasObject is dirty) the cache would be deleted.

I did not look at the TextureObjects.

I have prepared a lighter version of the fyne_demo: https://github.com/renlite/fyne/blob/develop/cmd/light/main.go

It would be interesting if the flickering still apears after starting the examples "hello", "shapes" and "light".

Bluebugs:
I don't believe that computing vertex is something we can classify as heavy. That's why I put it at the end of the list in my comment above. I do believe that actually the main impact is in the upload process that actually require the gpu to wait for it before it can work. That's why game try to keep all the geometry information on the gpu, small update are pushed and it limits the gpu having to wait on the upload. Latency and memory bandwidth are actually the problem. Computation is overwhelming starving on those.

If we are to be talking about optimization and not just adding features, I would highly recommend to build a benchmark application, add tracing to fyne and see were time is spent. My past experience on performance optimization is not to go there with assumption on what is taking time, but build a benchmark and measure things.

renlite:
GPU waiting for the upload process (T) = Calculation coords (T1) + Sending vertex to GPU (T2)
You are right to messure the time for T1 and T2 is an important next step. To start the prepared examples on the hardware with flickering would give us a first information whether the assumption is right. I don't have the hardware with flickering to check it.
I will implement the time output in the obove mentioned functions.
Additionally: If you click into a InputField in the "light" example the animation of the curser is forcing a lot of repainting.

ErrorNoInternet:
https://user-images.githubusercontent.com/61157623/207014811-bcff0bbe-5c9c-47cb-852d-1d0fdeb470a3.mp4

Also, the light version of fyne_demo wouldn't close properly
(sorry for the bad quality, had to compress it down to 9 MB in order to upload it)

renlite:
Thank you for the tests. It helpsto better understand the issue. There is a suprise that the line with more calculation than simple rect has no flickering. It could be the length of the vertex. (4 : 7). I expected that "hello" and "light" would work. Currently only "hello" works.
I will make some changes and implement the time component.

ErrorNoInternet:
> Currently only "hello" works.

Doesn't work 100% though, when I hover my mouse over the button, the background color becomes "invisible" and when I resize the window a ton, I can see the same artifacts (only half of the button's background is rendered)

renlite:
I have implemented time measuring in https://github.com/renlite/fyne/blob/develop/internal/painter/gl/draw.go and uploaded new examples. Would you please run the examples and publish the results.
https://github.com/renlite/fyne/tree/develop/cmd/rect_6
https://github.com/renlite/fyne/tree/develop/cmd/rect_30
https://github.com/renlite/fyne/tree/develop/cmd/rect_100%2B

ErrorNoInternet:
Hovered over the button, clicked 3 times, then closed the window:
rect_6: [intel-rect_6.txt](https://github.com/fyne-io/fyne/files/10216169/intel-rect_6.txt), [nvidia-rect_6.txt](https://github.com/fyne-io/fyne/files/10216171/nvidia-rect_6.txt)
rect_30: [intel-rect_30.txt](https://github.com/fyne-io/fyne/files/10216172/intel-rect_30.txt), [nvidia-rect_30.txt](https://github.com/fyne-io/fyne/files/10216173/nvidia-rect_30.txt)
rect_100+: [intel-rect_100+.txt](https://github.com/fyne-io/fyne/files/10216174/intel-rect_100%2B.txt), [nvidia-rect_100+.txt](https://github.com/fyne-io/fyne/files/10216175/nvidia-rect_100%2B.txt)

Note: `prime-run` sends rendered frames from the GPU's GDDR5 memory to the system's DDR3 memory and renders them (if I don't use NVIDIA Prime and let the dGPU take over Xorg, the flickering still exists)

Here are the results without using NVIDIA Prime (letting the NVIDIA card take over Xorg):
rect_6: [full-nvidia-rect_6.txt](https://github.com/fyne-io/fyne/files/10216284/full-nvidia-rect_6.txt)
rect_30: [full-nvidia-rect_30.txt](https://github.com/fyne-io/fyne/files/10216295/full-nvidia-rect_30.txt)
rect_100+: [full-nvidia-rect_100+.txt](https://github.com/fyne-io/fyne/files/10216287/full-nvidia-rect_100%2B.txt)

Arch Linux, i7-4710HQ, 8GB x2 DDR3, GTX 860M (GDDR5)

andydotxyz:
I would be very surprised if it is a CPU time issue - complex apps have been built with Fyne and none have exhibited problems like flickering textures. There may be a timing issue elsewhere but my personal feeling is that the shader may either have a bug or may be being passed unexpected data.

Out of curiosity did you try the simplified shader that @Bluebugs posted before? It managed to avoid drawing the radius in segments, perhaps that helps us to find out if the flicker is in the shader or in the coord calculations?

Either way we should probably move discussion off a closed PR so we can focus on going forward.

Bluebugs:
I agree with @andydotxyz, I would also be surprised if the issue for the flickering is CPU related. My thinking for this is that, rectangle are the only one in the UI to have this trouble every other element are correct. Also the GPU will draw all the command up to the swap buffer, it will then wait for the rendering queue to finish emptying and only then swap the buffer. So if we are to slow to actually push a frame, the outcome is frame rate drop, not half a frame being drawn. There is some other issue at play for the flickering.

renlite:
@ErrorNoInternet  Thank you for the tests. Does it mean that on Nvidia the simplest example with only 6 point rectangles does not work either. I will look at the shaders if I can find an optimisation to reduce the vertex length from 7 to 4.
@andydotxyz The example from @Bluebugs is an other technologie to paint the shapes. There a some open questions:
- Does it work with OpenGL 100?
- What ist the vertexShader? Only a base rectangle
- I think you loose the option to define a separate radius for all edges.
- We have a working code but unfortunately not on some hardware.

I would like to make some experiments with WebGPU und WGSL in the browser and would like to try the way with fragmentShader.

Bluebugs:
It is a very simple shader, there won't be a problem with OpenGL100 and GL ES 110. The shape is just a simple rectangle, no complexity there. Indeed the way it is currently written, it is the same radius for all corner. The benefit of that shader is that you also get linear gradient for the stroke and fill.

I kind of think your technique is closer to my other proposal to have a bunch of very simple independent triangular shape and build complex shape from them. I do think both work, but having a dedicated shader just for rounded rectangle would pay off considering the amount of time rounded rectangle shape are used in modern UI.

renlite:
There is a first version of GPU rendered stroke rectangle. This prototype doesn't have round edges. The normalization of vertex shader is calculated on the GPU too.
https://github.com/renlite/fyne
Would you please try it.

Bluebugs:
Just tested, it does work (as you say no round edges). This is on the github.com/renlite/fyne@develop branch for whoever want to test next.

renlite:
@Bluebugs Thank you for the test.
New version is available on github.com/renlite/fyne@develop branch. Now with round rectangle GPU rendered. Example .\cmd\rect_100+\main.go uses round rects.
image

Bluebugs:
Nice, it does work for me with the rounded rectangle. Maybe something to pay attention to when doing the PR, but the commit log looks weird.

ErrorNoInternet:
> There is a first version of GPU rendered stroke rectangle. This prototype doesn't have round edges. The normalization of vertex shader is calculated on the GPU too. https://github.com/renlite/fyne Would you please try it.

Looks like this for me, regardless of what card I'm using (notice the buttons)
image
image
image
(Those `Xlib` errors seem to appear when I press the PrintScreen button to take screenshots)

renlite:
@ErrorNoInternet Thanks for the response! I assume flickering is gone.
This is an unexpected issue. It seems the coords are not OK. There is a new example .\cmd\rect_test\main.go. I changed the outside color to see the space in use and implemented a shell output to see the coords (not normalized) for the rect submitted to the shaders.
image
image

ErrorNoInternet:
@renlite 

Yup, flickering is gone.
However, I'm seeing different results with different cards again. 

Intel:
image

NVIDIA:
image

renlite:
I thought it has something to do with scale and went through the code. I didn't find the reason in the fragment shader.
The first result (@ErrorNoInternet) shows correct rectangle space from vertex shader, but the second one has less vertical space for the first two rectangles. All rectangles have not the correct centerPosition. Scale seems not to produce the error, the results between scale 1 and scale 1.2 are almost identical.
At the moment I don't know why the issue happens. To see the frame size I implemented a println and in the fragment shader, if the size is not x800.0 and  y600 a red outer color will appear in the 'rect_test' example.

renlite:
It seems the real size of the screen/window in the fragment shader is not the submitted frame size but a scaled form. Maybe I have to consider the scale in the fragment shader.

Bluebugs:
I haven't looked at your code, but you might need to get the exact pixel size and not the windows pseudo pixels size which is scaled. I have found that a bit confusing as it is hard to see when it is scaled and when it is not. In your case, as things are sent to GL, you want to have the real pixels size.

Bluebugs:
I can confirm it is a scale issue, you can try with different FYNE_SCALE value different from 1 and you will see the issue.

andydotxyz:
> I have found that a bit confusing as it is hard to see when it is scaled and when it is not. 

Scaled values (our main coordinate system) are `float32`. Pixel coordinates (limited to the render/driver code) are `int` values.

Potential confusion arises later because OpenGL is working with floats from 0.0 and 1.0. However that is specific to the GL painter and not something that should impact other code.

renlite:
> Potential confusion arises later because OpenGL is working with floats from 0.0 and 1.0. However that is specific to the GL painter and not something that should impact other code.

Vertex coordinates are working different from fragment coordinates.  Fragment coords can be normalized to 0.0 -> 1.0 if the real size of the screen is known, but this is not necessary. From this view some var names in the fragment shader are not correct, and should be changed later.

renlite:
The code has been updated: pixScale is submitted to OpenGL and used in fragment shader. It has no impact on my hardware with scale 1.

Bluebugs:
This does work for me and look pretty good. It is definitively reducing massively the amount of texture uploaded to the GPU.

I think that with the same shader, you should be able to provide the implementation for circle to be fully running on the GPU too without texture upload.

Just a quick review of the shader.

   Division is slow and costly and 2.0/frame_size.x along with 2.0/frame_size.y are not changing for the entire run of the shader. I guess you can just precalculate them and just do vec4(vert.x*scale.x - 1.0, 1.0 - vert.y*scale.y, 0, 1) where scale.x == 2.0/frame_size.x and scale.y == 2.0/frame_size.y.
   
   A shader get its entire code, including all the if branch, executed for every run of the shader. The `if ( radius == 0.0 ){` in the shader will actually lead to a more costly run and the code for rounded rectangle should work just fine with radius == 0.0.

For people that are interested in reviewing the PR and might now know so much about the details of shaders, the way to think a bit about shader, they are executing things for multiple pixels, geometry, ... in parallel. That parallel execution is done using mainly a Single Instruction Multiple Data design. To make it work with `loop` and `if` construct, each instruction use a mask to disable/enable which pixels/geometry/... the instruction is manipulating. That's why `if` do not provide an optimization or shortcut as they do on the CPU, because all the branch are executed, but with a different mask.

Overall this feel very close to a PR and very early in our release cycle, would be amazing to get!

renlite:
Does it work for @ErrorNoInternet and @andydotxyz too?

Jacalz:
It seems to work for me on Intel UHD Graphics 620 but the grey rectangles go red when I resize the window.

renlite:
> It seems to work for me on Intel UHD Graphics 620 but the grey rectangles go red when I resize the window (in the example application).

Thanks, the grey to red change is ok, it is the only possible type of debugging on the GPU. This will be deleted later.

ErrorNoInternet:
It looks fine on my first monitor thats 1650x1080, however, when I drag it to my 1920x1080 monitor it looks like this:
image
The window also becomes larger, so I guess it just scaled up

renlite:
Thank you for your feedback. That seems to be also ok. The last rect without round edges is not correct but this is because I did not scale the size in shader for this rect type.  There are open questions to discuss (as mentioned by @Bluebugs) whether to intergrate the simple rect or to write a separate shader. The simple rect (with stroke) is massively used so a simpler shader would save power.
> The window also becomes larger, so I guess it just scaled up

I would say this behavior is ok too but I can't anwer this question.

Bluebugs:
> The simple rect (with stroke) is massively used so a simpler shader would save power.

This is absolutely true.

> > The window also becomes larger, so I guess it just scaled up
> 
> I would say this behavior is ok too but I can't anwer this question.

Yes, when moving window between screen with different scale, the window min size will be recalculated and most of the time it will be different due to a difference in pixels density that lead to a bigger window.

renlite:
@Bluebugs @andydotxyz 
I will clean up the code and prepare it for a PR. I think it is clearer to make a new PR with a link to this one. Do you agree?

@Bluebugs 
>Division is slow and costly  ...

Yes, this is a good input to reduce calculations in the shaders for variables that are constant through all threads. I know that the fragment shader creates for all pixels of the rendering area a thread to paint the color, but is this true for the vertex shaders too ? I don't know if the vertex shader uses only the number (threads) of used points - in our case 6 = two triangles - or the whole number of pixels.  If vertex shaders creates threads for all pixels too then it would be better to make the normalization in Go and not in OpenGL.
```gl_Position = vec4(2.0*vert.x/frame_size.x - 1.0, 1.0 - 2.0*vert.y/frame_size.y, 0, 1);```
 ```attribute vec2 normal;``` are not used at the moment and the submit would happen in one buffer.

> I think that with the same shader, you should be able to provide the implementation for circle to be fully running on the GPU too without texture upload.

Yes, this was my plan to do it with a new PR. Maybe with an own fragment shader. I think there is a faster version for circle only.

@Bluebugs @andydotxyz  @Jacalz  @ErrorNoInternet 
To get optimized rendering I would create a separate ```CanvasObject``` for the ```Rectangle``` and ```RoundRectangle``` both inheriting from ```baseObject```.  
This idea has following reasons:
1) Better performance for Rectangle (massively used)
2) The fragment shader is prepared to support different strokes (top, bottom, left, right). ```RoundRectangle``` can have only one stroke width.
3) Because it is not necessary, the ```Rectangle``` is not antialiazed. I did not try it but I think it should work to create a ```Rectangle``` with the object ```RoundRectangle (Radius 0.0)```. This would be much more expensive (for many objects) but would offer smooth steps if desired by the programmer.

@andydotxyz could you please test the example /cmd/rect_test/main.go (branche develop) whether flickering is gone on your hardware?

Bluebugs:
> @Bluebugs @andydotxyz I will clean up the code and prepare it for a PR. I think it is clearer to make a new PR with a link to this one. Do you agree?

Absolutely, will be easier to do a proper review from a clean start.
 
> @Bluebugs
> 
> > Division is slow and costly  ...
> 
> Yes, this is a good input to reduce calculations in the shaders for variables that are constant through all threads. I know that the fragment shader creates for all pixels of the rendering area a thread to paint the color, but is this true for the vertex shaders too ? I don't know if the vertex shader uses only the number (threads) of used points - in our case 6 = two triangles - or the whole number of pixels. If vertex shaders creates threads for all pixels too then it would be better to make the normalization in Go and not in OpenGL. `gl_Position = vec4(2.0*vert.x/frame_size.x - 1.0, 1.0 - 2.0*vert.y/frame_size.y, 0, 1);` `attribute vec2 normal;` are not used at the moment and the submit would happen in one buffer.

For the vertex shader, it should be per vertex. So having a bit of optimization there is still a good thing, but no need to do head scratching one :-)

> > I think that with the same shader, you should be able to provide the implementation for circle to be fully running on the GPU too without texture upload.
> 
> Yes, this was my plan to do it with a new PR. Maybe with an own fragment shader. I think there is a faster version for circle only.

I agree, there is a faster one for circle and it could enable adding more option later to it (like partial arc and stuff). Still, this is already a pretty good step forward, so I would be perfectly ok with just reusing the rounded rectangle shader for now.

> @Bluebugs @andydotxyz @Jacalz @ErrorNoInternet To get optimized rendering I would create a separate `CanvasObject` for the `Rectangle` and `RoundRectangle` both inheriting from `baseObject`. This idea has following reasons:
> 
>     1. Better performance for Rectangle (massively used)
> 
>     2. The fragment shader is prepared to support different strokes (top, bottom, left, right). `RoundRectangle` can have only one stroke width.

I think having multiple stroke width for each side of the rectangle to be maybe a bit unnecessary at this stage. All the modern design I have seen are using the same stroke around the rectangle. It might be a marginal use case that would be better left to a generic vector drawing primitive.

>     3. Because it is not necessary, the `Rectangle` is not antialiazed. I did not try it but I think it should work to create a `Rectangle` with the object `RoundRectangle (Radius 0.0)`. This would be much more expensive (for many objects) but would offer smooth steps if desired by the programmer.

I don't think you need a new primitive, you just need to check in the painter if the radius is != 0 and switch the shader accordingly. I prefer to have less primitive exposed, but with more feature and put all the optimization logic in the painter as much as we can.

andydotxyz:
When run on macOS I see this:

Screenshot 2022-12-31 at 14 17 35

> @Bluebugs @andydotxyz
I will clean up the code and prepare it for a PR. I think it is clearer to make a new PR with a link to this one. Do you agree?

Yes please, a nice clean start with the new code.

It may be worth noting that if `Radius == Height/2` then the ends of the rectangle should be a semi-circle, but I don't think it is, so when adjusting the scaling calculations maybe check this too?

> To get optimized rendering I would create a separate CanvasObject for the Rectangle and RoundRectangle both inheriting from baseObject.

I don't think the API should be designed around implementation details. We can internally load two different shaders based on whether `Radius == 0` and/or `StrokeWidth` and `StrokeColor` settings as well. This way the user of the API does not have to worry and can turn radius on and off for the shape.

renlite:
image
Some notes to the actual state:
- Code redesign with a lot of optimizations in vertex and fragment shaders.
- Implementation of scaling for all relevant params.
- Original ```cmd/fyne_demo/main.go``` should now work as expected.
- Split between Rectangle and RoundRectangle depending on Radius. 
- In the example ```cmd/rect_test/maint.go``` is a try to paint a circle with ```round_rectangle.frag``` shader. I think it looks quite good.
- At the bottom of the example  ```cmd/rect_test/main.go``` are two rectangles. The left one is rendered by ```rectangle.frag``` and the right one by ```round_rectangle.frag```. The right rectangle is antialized and has a radius of 0.01.

Still missing:
- To define radius by pixel and in percent. At the moment only pixel. I would say this needs a struct of Radius with both values. If both values are defined then use pixel or throw an exception.
- The circle shape is still a texture. Replace texture rendering with ```round_rectangle.frag``` shader if the result looks good.

Please let me know whether all works as expected.

ErrorNoInternet:
@renlite Works perfectly fine on both my monitors!

andydotxyz:
Oh, for the screenshots addition it looks glitch-free on my HiDPI Linux laptop too:

renlite:
Thank you for the tests. I will include the OpenGL ES path and the radius in percentage and then make a PR.

Issue #2083, opened 2021-03-12
iPhone 12 mini cannot tap on widgets in right-most 12dp elements


andydotxyz:
Seems to relate to some issue with scaling coordinates on the X-axis.
Not clear why rendering scale correctly but the taps don't.

### To Reproduce:
Steps to reproduce the behaviour:
1. build an iOS app with a narrow widget on right-hand side (i.e. slider)
2. Try to interact
3. Notice that that around 12dp does not respond. tapping to the left of the widget seems to get signals.

### Device (please complete the following information):
 - **OS:** iOS
 - **Version:** 1.14
 - **Go version:** 1.15
 - **Fyne version:** 2.0.1

PR #3123, opened 2022-07-04
Use UIScreen bounds instead of nativeBounds for DisplayMetrics

### LINKED ISSUE #3122: Iphone incorrect click coordinates in zoomed screen view
### Describe the bug:

When using a zoomed display zoom view on IOS, click positions registered in fyne apps are incorrect relative to actual user clicks. Simply, when I click in the center of the screen, fyne registers the click a bit to the top left.
### To Reproduce:
1. Set zoomed view in Settings -> Display & Brightness -> View -> Zoomed
2. Get fyne calc from app store.
3. Click on the top section of any button.
4. See that click is not registered on button clicked. For example, clicking on top section of button with number `5` in fyne calc, you will get button with number `8` clicked.

### Screenshots:

N/A

### Example code:

I believe the bug is in internal driver code, so all fyne IOS apps are susceptible to the bug.

 - **OS:** IOS
 - **Version:** 15.5
 - **Go version:** go1.17.1 darwin/amd64
 - **Fyne version:** v2.2.1 and develop

cedy:
### Description:

Fixes #3122

### Checklist:

- [] Tests included. <- I don't think tests applicable here.
- [x] Tests all pass. <- Same tests that fail on master/develop for me, fail after my fix.

### DIFF
--- a/internal/driver/mobile/app/darwin_ios.go
+++ b/internal/driver/mobile/app/darwin_ios.go
@@ -84,8 +84,8 @@ var DisplayMetrics struct {
 
 //export setDisplayMetrics
 func setDisplayMetrics(width, height int, scale int) {
-	DisplayMetrics.WidthPx = width
-	DisplayMetrics.HeightPx = height
+	DisplayMetrics.WidthPx = width * scale
+	DisplayMetrics.HeightPx = height * scale
 }
 
 //export setScreen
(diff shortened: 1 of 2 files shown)


andydotxyz (commented):
Thanks so much for this change. Can you confirm that the apps still work in landscape as well? :)
I wonder if this fixes #2083 as well

cedy:
Thank you for the awesome open sourced golang framework! I'm glad that I'm able to contribute. 
I've just retested it on fyne calculator app as well, works well in both orientations and view modes.

andydotxyz (approved):

andydotxyz:
This is now resolved on `develop` and ready for `v2.2.3` as well, thanks @cedy

andydotxyz:
Sorry I had to revert the application of this change. On further testing for the release candidate I see that it breaks popup content sizing :(

cedy:
Sorry to hear that, here is popup widget that I’m getting with patch applied. On zoomed and standard view. What is the right behavior?
44F60B94-95AB-41DA-B6E5-E219B55A2A4D
E740D505-0038-47D2-BAB4-289941DA83D7

andydotxyz:
Ah, a different kind of popup - it was `ShowPopUpMenuAtPosition` that was broken in my tests.
Will try to get a screenshot.

andydotxyz:
In fyne_demo you can see that it pops up the menu at the wrong position on the `Buttons` tab.
The instance I saw in my own application was here where the menu does not fill the bounds.
Not sure why they appear as a different break, I think the root cause is in `Driver.AbsolutePositionForObject` returning incorrect values with this patch applied. But this is a guess.
IMG_1849
IMG_1848

PR #3136, opened 2022-07-12
Fix #3113 Use lock in Container.Remove

### LINKED ISSUE #3113: Container.Remove() from multiple goroutines causes panic
### Describe the bug:
Calling Container.Remove() from multiple goroutines at the exact same time causes a panic
Container.Add() seems to be working fine (unless you're trying to add 10k+ items on a 2 fps display then it crashes)
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x7bd2b9]

goroutine 78 [running]:
fyne.io/fyne/v2/layout.(*boxLayout).Layout(0xc0003b19d8, {0xc0004e8000, 0x7d, 0xc000167250?}, {0x48732e?, 0x0?})
        /home/ryan/go/pkg/mod/fyne.io/fyne/v2@v2.2.2/layout/boxlayout.go:63 +0x399
fyne.io/fyne/v2.(*Container).layout(...)
        /home/ryan/go/pkg/mod/fyne.io/fyne/v2@v2.2.2/container.go:186
fyne.io/fyne/v2.(*Container).Remove(0xc00037fe40, {0xb0fe98?, 0xc0003f9e00})
        /home/ryan/go/pkg/mod/fyne.io/fyne/v2@v2.2.2/container.go:139 +0x1fb
main.removeLater({0xb0fe98, 0xc0003f9e00})
        /home/ryan/Downloads/FyneContainerCrash/main.go:43 +0x54
created by main.main
        /home/ryan/Downloads/FyneContainerCrash/main.go:29 +0xf0
exit status 2
```

### To Reproduce:
Example code here: https://go.dev/play/p/pSAsuB7XU-5

1. Create a container
2. Add a ton of objects to that container
3. Launch a goroutine for each object
4. Have each goroutine remove an object from the container at the same time
5. Panic

 - **OS:** Linux (amd64)
 - **Version:** 5.18.7-arch1-1
 - **Go version:** 1.18.3
 - **Fyne version:** 2.2.1

cedy:
### Description:

Use lock around objects slice to prevent race condition.

Fixes #3113

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container.go
+++ b/container.go
@@ -1,5 +1,9 @@
 package fyne
 
+import (
+	"sync"
+)
+
 // Declare conformity to CanvasObject
 var _ CanvasObject = (*Container)(nil)
 
@@ -10,8 +14,9 @@ type Container struct {
 	position Position // The current position of the Container
 	Hidden   bool     // Is this Container hidden
 
-	Layout  Layout         // The Layout algorithm for arranging child CanvasObjects
-	Objects []CanvasObject // The set of CanvasObjects this container holds
+	Layout      Layout // The Layout algorithm for arranging child CanvasObjects
+	objectsLock *sync.Mutex
+	Objects     []CanvasObject // The set of CanvasObjects this container holds
 }
 
 // NewContainer returns a new Container instance holding the specified CanvasObjects.
@@ -27,7 +32,8 @@ func NewContainer(objects ...CanvasObject) *Container {
 // Deprecated: Use container.NewWithoutLayout() instead
 func NewContainerWithoutLayout(objects ...CanvasObject) *Container {
 	ret := &Container{
-		Objects: objects,
+		Objects:     objects,
+		objectsLock: &sync.Mutex{},
 	}
 
 	ret.size = ret.MinSize()
@@ -40,8 +46,9 @@ func NewContainerWithoutLayout(objects ...CanvasObject) *Container {
 // Deprecated: Use container.New() instead
 func NewContainerWithLayout(layout Layout, objects ...CanvasObject) *Container {
 	ret := &Container{
-		Objects: objects,
-		Layout:  layout,
+		Objects:     objects,
+		objectsLock: &sync.Mutex{},
+		Layout:      layout,
 	}
 
 	ret.size = layout.MinSize(objects)
@@ -126,6 +133,8 @@ func (c *Container) Remove(rem CanvasObject) {
 		return
 	}
 
+	c.objectsLock.Lock()
+	defer c.objectsLock.Unlock()
 	for i, o := range c.Objects {
 		if o != rem {
 			continue
--- a/container.go
+++ b/container.go
@@ -60,6 +60,8 @@ func (c *Container) Add(add CanvasObject) {
 		return
 	}
 
+	c.objectsLock.Lock()
+	defer c.objectsLock.Unlock()
 	c.Objects = append(c.Objects, add)
 	c.layout()
 }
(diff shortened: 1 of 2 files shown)


container.go
```diff
@@ -10,8 +14,9 @@ type Container struct {
 	position Position // The current position of the Container
 	Hidden   bool     // Is this Container hidden
 
-	Layout  Layout         // The Layout algorithm for arranging child CanvasObjects
-	Objects []CanvasObject // The set of CanvasObjects this container holds
+	Layout      Layout // The Layout algorithm for arranging child CanvasObjects
+	objectsLock *sync.Mutex
```
Jacalz, line 18: Isn’t it possible to not use a pointer here and thus avoid the need to pass a new value each time the struct is created? The current way would crash anyone using containers without the constructors.

container.go
```diff
@@ -1,5 +1,9 @@
 package fyne
 
+import (
+	"sync"
+)
```
Jacalz, line 5: Let’s get this on one row please.

Jacalz (changes requested):
Thanks for the fix. I have a few suggestions.

container.go
Jacalz:
```suggestion
	objectsLock sync.Mutex
```

cedy: Definitely, no need for a pointer. Thanks

container.go
cedy: 👍

andydotxyz:
Would it be worth adding to `Add` at the same time? It will suffer a version of this crash as well I think...

ErrorNoInternet:
> Would it be worth adding to `Add` at the same time? It will suffer a version of this crash as well I think...

Just did some testing, and it seems like the more widgets you try to add, the higher the chance of crashing. Sometimes it crashes when I'm adding more than 512 widgets (I would say a 30% chance of crashing), but when it doesn't crash, not all of the 512 widgets are added. I can only see about 10-20 widgets in the container.

So yes, adding a mutex to `Add` would be nice.

cedy:
> Would it be worth adding to `Add` at the same time? It will suffer a version of this crash as well I think...

Indeed, added there as well

container.go
```diff
@@ -10,8 +12,9 @@ type Container struct {
 	position Position // The current position of the Container
 	Hidden   bool     // Is this Container hidden
 
-	Layout  Layout         // The Layout algorithm for arranging child CanvasObjects
-	Objects []CanvasObject // The set of CanvasObjects this container holds
+	Layout      Layout // The Layout algorithm for arranging child CanvasObjects
+	objectsLock sync.Mutex
```
Jacalz, line 16: Should we just call it just `lock`?

Jacalz (changes requested):
Looking good. Thanks. Just one small wording change that I think would be beneficial.

container.go
cedy: 👍

Jacalz (dismissed):
Thanks. This looks sensible to me. I just want a second opinion before we merge as I'm only a novice when it comes to mutexes.

Jacalz:
Ah, looks like you have some formatting that needs to be fixed as well.

cedy:
> Ah, looks like you have some formatting that needs to be fixed as well.

yeah, annoying when go fmt should run on file save but it doesn't, fixed now :)

Jacalz:
It looks like tests are failing as well.

cedy:
> It looks like tests are failing as well.

Seem to pass now after rebase

Bluebugs (approved):
This does improve things significantly. A good step forward.

Jacalz (approved):
Looks like this might have been waiting for my review. Sorry for the wait.

PR #3137, opened 2022-07-13
Fix incorrect size and position for radio focus indicator

Jacalz:
### Description:

This fixes an issue where the incorrect amount of padding was added to the focus indicator in `radioItem`.

#### Before:
image

#### After:
image

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
> It would be nice to split commit next time with cosmetic change (like *0.5 -> /2 or reordering code) that are orthogonal to the fix to a different commit to make review a bit easier.

Sure thing. Sorry about that.

PR #3155, opened 2022-07-21
Fix installing instructions for fyne CLI

kataras:
As titled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Using `go install` will copy the executable into your go `bin` dir.
 To install the application with icons etc into your operating system's standard
 application location you can use the fyne utility and the "install" subcommand.
 
-    $ go get fyne.io/fyne/v2/cmd/fyne
+    $ go install fyne.io/fyne/v2/cmd/fyne@latest
     $ fyne install
 
 # Packaging for mobile


Jacalz:
> /cc @Jacalz

Please avoid doing that unless absolutely necessary. We sometimes request reviews when it makes sense because a particular developer has more knowledge of a specific piece of code. Anyone in the Fyne org has access to the dashboard that shows everything that happens and will respond as soon as we have time.

README.md
```diff
@@ -121,7 +121,7 @@ Using `go install` will copy the executable into your go `bin` dir.
 To install the application with icons etc into your operating system's standard
 application location you can use the fyne utility and the "install" subcommand.
 
-    $ go get fyne.io/fyne/v2/cmd/fyne
+    $ go install fyne.io/fyne/v2/cmd/fyne@latest
```
Jacalz, line 124: This doesn't work for Go versions < 1.16 and as we support Go 1.14, we can't just show a command that will only work for some. You need to do something similar to https://developer.fyne.io/started/#installing instead.

Jacalz (changes requested):

README.md
kataras: Look on top of this command, it says:

> "To install the application with icons etc into your operating system's standard
application location you can use the fyne utility and the "install" subcommand."


So it already describes the `install` subcommand of go>1.16 versions. Do you believe new developers who want to install fyne will use versions below 1.16? I don't think so, but if you do, let's keep it as it is.

Jacalz: No, it does not describe the new install command. It is the same install command just that what you have changed it to doesn’t work for older Go releases.

Yes, I believe that some users will use an older Go version because they are what’s available by default for some older LTS Linux releases.

andydotxyz: Notice the difference between "fyne install" and "go install" - that may be causing some confusion.

Jacalz:
Friendly ping. Are you interested in continuing this? @kataras

rexposadas:
In the README, we can simply make a distinction on what installation command to use depending on the go version.

Jacalz:
I'll close this as stale. Feel free to reopen if you want to continue working on it.

PR #3158, opened 2022-07-23
Fix build error when using 'fyne build' and no icon

andydotxyz:
Turns out that icon can be missing because "fyne build" does not require metadata.
This change fixes the resulting compile error for that case

### Checklist:

- [ ] Tests included. <- "fyne build" any fyne app with no icon file of FyneApp.toml
- [x] Tests all pass.


Bluebugs (approved):

fuskiid:
I've tested this and I can confirm it has resolved the error in my personal environment I had previously reported over discord. It is now correctly generating the Fyne resource on build. Thanks :)

PR #3162, opened 2022-07-26
Correctly transition cloud providers in the test app

andydotxyz:
This was missed in the test app so it did not behave consistently on provider switching out.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/test/testapp.go
+++ b/test/testapp.go
@@ -132,7 +132,7 @@ func (a *testApp) transitionCloud(p fyne.CloudProvider) {
 	}
 
 	// after transition ensure settings listener is fired
-	a.settings.apply() //just a test - "cedric"
+	a.settings.apply()
 }
 
 // NewApp returns a new dummy app used for testing.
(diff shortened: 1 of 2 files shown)


test/testapp.go
```diff
@@ -106,6 +111,30 @@ func (a *testApp) lastAppliedTheme() fyne.Theme {
...
+		a.prefs = internal.NewInMemoryPreferences()
+	}
+
+	for _, l := range listeners {
+		a.prefs.AddChangeListener(l)
+		l() // assume that preferences have changed because we replaced the provider
+	}
+
+	// after transition ensure settings listener is fired
+	a.settings.apply() //just a test - "cedric"
```
Bluebugs, line 135: I am not sure I need to be in the comment for posterity here :-)

andydotxyz: oops!

andydotxyz, line 135:
```suggestion
	a.settings.apply()
```

Bluebugs (approved):

Issue #3042, opened 2022-06-06
Add new theme updates


andydotxyz:
Previews and possible API considerations to be added here, this is a placeholder issue.

andydotxyz:
The full design needs to be exported, but here are some details to show what is being worked on:
<img width="620" alt="Screenshot 2022-07-11 at 13 01 48" src="https://user-images.githubusercontent.com/294436/178260186-714a1d09-f142-44e5-a271-16904af9617f.png">
<img width="621" alt="Screenshot 2022-07-11 at 13 02 03" src="https://user-images.githubusercontent.com/294436/178260192-6056778a-e321-42fc-9837-bf2914084fe2.png">
<img width="339" alt="Screenshot 2022-07-11 at 13 02 28" src="https://user-images.githubusercontent.com/294436/178260200-50409442-70e3-4674-9dea-30c26a248c09.png">
<img width="336" alt="Screenshot 2022-07-11 at 13 02 37" src="https://user-images.githubusercontent.com/294436/178260206-1c7782d2-e5b2-4bfc-a555-f6125d8c9861.png">
<img width="622" alt="Screenshot 2022-07-11 at 13 03 02" src="https://user-images.githubusercontent.com/294436/178260209-ffc77680-58fc-43e3-9b21-727871e6e58c.png">
<img width="622" alt="Screenshot 2022-07-11 at 13 03 18" src="https://user-images.githubusercontent.com/294436/178260212-f2089042-a5cd-4cb0-8588-5c782391ad9e.png">

ErrorNoInternet:
Will there be rounded corners? :eyes:

c1ngular:
well done , looks refreshing .

ddkwork:
good,  i want change the button text size when mouse in and out,but i found
the text style not export this field.

andydotxyz:
>  i want change the button text size when mouse in and out,but i found
the text style not export this field.

That is not supported by the standard button. 
To do this you would need to add your own button that altered size based on desktop events.

Jacalz:
Just adding it here that we discussed this during the last call. It seems like the text selection colour for dark theme is incorrect (too dark) and the section colour used in tree is still the old one for light theme. We need to check this through before closing as entirely done.

Jacalz:
An update on things that remain to be done (as far as I know):
- [ ] Update the selection colour for the light theme with new colour.
- [ ] Update the text selection colour in the dark theme to have better contrast.
[...]

Issue #2817, opened 2022-03-02
  Unexpected vertical scroll bar in entry 


BabySid:
### Describe the bug:
As the title describes, when the grid layout is used, the vertical scroll bar appears in the entry

### To Reproduce:
Steps to reproduce the behaviour:
- Run the code below

### Screenshots:
![图片](https://user-images.githubusercontent.com/6658943/156390825-b01704f5-e69f-4a59-b64c-034fb1ad8401.png)

### Example code:
```go
func main() {
	myApp := app.New()
	myWindow := myApp.NewWindow("Entry Widget")

	input := widget.NewEntry()
	input.SetPlaceHolder("Enter text...")

	content := container.NewGridWithColumns(2, input, widget.NewButton("btn", nil))

	myWindow.SetContent(content)
	myWindow.ShowAndRun()
}
```

### Device (please complete the following information):
 - **OS:** Windows10
 - **Version:** 19044
 - **Go version:** go1.17.5
 - **Fyne version:** 2.1.3

andydotxyz:
I cannot replicate this, can you provide any more info about actions, display scale etc.
The screenshot is not the default window size, did you have to resize it to see the issue?

BabySid:
> I cannot replicate this, can you provide any more info about actions, display scale etc. The screenshot is not the default window size, did you have to resize it to see the issue?

Um, yes. Because the default window size is too small, I lengthened the window manually.

However, I have rewritten the following code and still have the same problem.

![图片](https://user-images.githubusercontent.com/6658943/156484305-8c5001e5-1241-4543-aca0-968d6fe7123e.png)

```go
func main() {
	myApp := app.New()
	myWindow := myApp.NewWindow("Entry Widget")

	input := widget.NewEntry()
	input.SetPlaceHolder("Enter text...")

	top := container.NewGridWithColumns(2, input, widget.NewButton("btn", nil))

	myWindow.SetContent(container.NewBorder(top, nil, nil, nil, widget.NewLabel("hello world")))

	myWindow.Resize(fyne.NewSize(600, 400))
	myWindow.ShowAndRun()
}
```

In addition, if I change layout to HBox, although the input box is too small to meet my expectations, at least the vertical scroll bar does not exist. Very strange.
![图片](https://user-images.githubusercontent.com/6658943/156485279-2b176279-a4b0-40c1-9bdf-f3f1afbf4245.png)

BabySid:
[...]

PR #3163, opened 2022-07-29
Starting work on new theme. Update colours and sizes initially

andydotxyz:
Set button to not have a shadow and reduce cursor size to match input border
This is broad strokes stuff that does not require new APIs.

It should be noted that this design moves from transparent background on buttons - and we have removed their shadows.

Progresses #3042 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.
Screenshot 2022-07-29 at 14 25 02
Screenshot 2022-07-29 at 14 24 59
Screenshot 2022-07-29 at 14 24 59


andydotxyz:
> This looks pretty good. Just one thing is weird in the data binding. The text entry with the number are a weird shadow artefacts. It is the only one. Any idea why?

I think that may be this issue: https://github.com/fyne-io/fyne/issues/2817 which I hope to look at shortly - basically a rounding error that needs addressed (not in theme code)

Bluebugs (approved):
Looks great!

Issue #1904, opened 2021-02-05
Cannot disable container.TabItem


pohanlSUM:
### Is your feature request related to a problem? Please describe:
We cannot disable container.TabItem using attribute inside container.TabItem

### Is it possible to construct a solution with the existing API?
if tabs.CurrentTabIndex() != 0 && someCondition {
	fmt.Println("Disable!")
	tabs.SelectTab(tabs.Items[0])
}

### Describe the solution you'd like to see:
There is an attribute in container.TabItem named "Disable", default value is false.
When Disable is set to true, user cannot select this tab

pohanlSUM:
widget.treeNode has same problem

andydotxyz:
Good request for TabItem. However TreeNode is not the same, that is just an ID and it’s up to the developer to manage the presentation of the tree nodes.

Aakash-Jakhmola:
@andydotxyz This needs to be closed.

andydotxyz:
On develop ready for v2.3.0

PR #3172, opened 2022-08-01
Allow TabItem to be Disabled

gavinwade12:
### Description:

Addresses #1904 by adding methods to AppTabs and DocTabs for disabling TabItems.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.
~~I was having the issue with the dialog tests failing, so I installed Go 1.17 (currently on 1.18.5) and ran the tests with it. However, I'm still having issues with (seemingly) unrelated tests failing that are also failing in other PRs. Here's an [example](https://github.com/fyne-io/fyne/runs/7603055303?check_suite_focus=true); although, it's strange the 1.14-ubuntu failed and 1.17-ubuntu passed here when I am running the tests with 1.17 on Linux Mint.~~
I have an open PR for the randomly-failing test (#3175).

#### Where applicable:


container/apptabs.go
```diff
@@ -89,6 +88,34 @@ func (t *AppTabs) CurrentTabIndex() int {
 	return t.current
 }
 
+// Enable enables the specified TabItem.
+//
+// Since: 2.2
```
andydotxyz, line 93: 2.2 is already released, this would need to be 2.3

andydotxyz: Also we have a `Disablable` interface with `Enable` which enables the whole widget, so this name overloading may be confusing. I would recommend `EnableItem` to indicate that it's function is different (same with `Disable` etc).

andydotxyz (changes requested):
Thanks for the contribution.

A few thoughts inline.
Plus you'll need to address the static analysis failures

container/apptabs.go
gavinwade12: Thanks for the review! I believe I've made your suggested updates, and the static analysis is now passing. There is now just the `TestForm_Disabled` test that is sometimes failing, but I believe it's unrelated to this change. The same test fails when I run it prior to my changes e.g. `git checkout 2504bea3c6a5c97a1d5b45d7c3a39128de6fba5a && go test ./...`.

andydotxyz: There does seem to be an instability in that form test. All others are now fixed but that one remains a mystery

gavinwade12: I believe I may have solved the issue with the form test: https://github.com/fyne-io/fyne/pull/3175.

container/tabs.go
```diff
@@ -29,22 +29,22 @@ func (ti *TabItem) enable() {
 	}
 }
 
+func (ti *TabItem) disable() {
```
andydotxyz, line 32: private functions (not exported) should be below public (exported) ones.
Apologies that this is not automatically enforced yet, we are working on it.
More info about order of code in a file at: https://github.com/fyne-io/fyne/wiki/Contributing#code-style

container/apptabs.go
```diff
@@ -89,6 +88,34 @@ func (t *AppTabs) CurrentTabIndex() int {
 	return t.current
 }
 
+// EnableItem enables the specified TabItem.
```
andydotxyz, line 91: We try to keep stuff alphabetical, please put enable after disable.
https://github.com/fyne-io/fyne/wiki/Contributing#code-style

andydotxyz (changes requested):
Looking good, but can you change the order of functions please - noted the contributor doc inline.
We really need to make that an automated check...

container/tabs.go
gavinwade12: Sorry about that! I admit I should have read through that a bit more thoroughly the first time. I believe things are correct now.

container/apptabs.go
gavinwade12: This is fixed now. My apologies for the delay!

gavinwade12:
Hey Andy, my apologies for the delay in cleaning this up. I think it's ready to go now!

andydotxyz (approved):
Great, thanks!

PR #2262, opened 2021-05-24
Simple attempt to fix form tests 🤞

fpabl0:
### Description:

Use `test.ApplyTheme` instead of `app.Settings().SetTheme` to avoid data races in form tests.

### Checklist:

~~- [ ] Tests included.~~
- [x] Lint and formatter run with no errors.
- [x] Tests all pass.

PR #3175, opened 2022-08-03
Fix Random TestForm_Disabled Failure

gavinwade12:
### Description:
This fixes the randomly-failing `TestForm_Disabled` test. I narrowed this down to a race with `cache.Renderer`. `test.NewApp()` spawns a goroutine that calls `app.ApplySettings` when settings are changed. This was racing with `app.NewWindow("")` to create the form's renderer, so `CreateRenderer` was occasionally being called on the form twice. This led to separate pointers for the `itemGrid` where the cached renderer didn't know about any of the form's objects.

### Checklist:
- [x] Tests all pass.

### DIFF
--- a/internal/cache/widget.go
+++ b/internal/cache/widget.go
@@ -32,11 +32,7 @@ func Renderer(wid fyne.Widget) fyne.WidgetRenderer {
 	if !ok {
 		rinfo = &rendererInfo{renderer: wid.CreateRenderer()}
 		renderersLock.Lock()
-		if erinfo, ok := renderers[wid]; ok {
-			rinfo = erinfo
-		} else {
-			renderers[wid] = rinfo
-		}
+		renderers[wid] = rinfo
 		renderersLock.Unlock()
 	}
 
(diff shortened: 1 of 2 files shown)


gavinwade12:
Actually, I'm not sure this is the correct fix. It can be made fail by removing the second form item from the `NewForm` call and moving it to an `AppendItem` after `test.NewWindow`. It's quite late here, so I'll have to revisit this tomorrow with a clearer mind.

andydotxyz:
Thanks for looking in to this. Post what you find in the morning and we can review or discuss :)

fpabl0:
I think the solution for this, is the same as before when we had flaky tests on Form widget too (see #2262). Basically:
`app.Settings().SetTheme(theme.LightTheme())` should be replaced with `test.ApplyTheme(t, theme.LightTheme())`.

That should solve this, hopefully :)

gavinwade12:
Thank you, @fpabl0! This seems to do the trick. My previous change was also causing some unexpected results. My apologies for the delay in looking at this.

fpabl0 (approved):
Looks good to me! Thanks :)

PR #3178, opened 2022-08-03
feat: create new folder from file/folder dialog

### LINKED ISSUE #3174: Create new folder directly from FolderDialog
### Checklist

I have tried searching through all issues but am unable to find any that addresses this. The closest I found is #2981 but it is not exactly the same.

Currently, the FolderDialog seems to lack the feature of creating a new folder to be selected. For example, if the directory `/Users/me/` exists, but I want to select the folder `/Users/me/folder` which does not already exist, I would have to create the folder separately first before selecting it from the FolderDialog.

I believe it is possible to add this functionality via FileDialog. What I have in mind is something along of the lines shown below.

```
type fileDialog struct {
        ...
        newFolder        *widget.Button
        ...
}
```

```
f.newFolder = widget.NewButton(label, func() {
        // Get curr dir via f.dir
        // Create folder via os.Mkdir
        // Refresh dialog
})
```

As proposed above.

Beebeeoii:
### Description:

This PR introduces the ability to create new folder directly from the file/folder dialog. This is motivated by the hassle of having to create a folder separately before being able to select it via the folder dialog.

Changes were made to `makeUI()` to include a create folder button in the options container which would open up a entry form dialog for user to input the new folder name before executing the `mkdir` syscall. Directory is then refreshed to enable user to select the newly created folder. 

Fixes #3174 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -246,7 +246,7 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 			}
 
 			newFolderPath := filepath.Join(f.dir.Path(), newFolderEntry.Text)
-			createFolderErr := os.MkdirAll(newFolderPath, 0700)
+			createFolderErr := os.MkdirAll(newFolderPath, 0750)
 			if createFolderErr != nil {
 				fyne.LogError(
 					fmt.Sprintf("Failed to create folder with path %s", newFolderPath),
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -246,7 +246,7 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 			}
 
 			newFolderPath := filepath.Join(f.dir.Path(), newFolderEntry.Text)
-			createFolderErr := os.MkdirAll(newFolderPath, 0000)
+			createFolderErr := os.MkdirAll(newFolderPath, 0700)
 			if createFolderErr != nil {
 				fyne.LogError(
 					fmt.Sprintf("Failed to create folder with path %s", newFolderPath),
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -254,9 +254,9 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 	})
 
 	optionsbuttons := container.NewHBox(
+		newFolderButton,
 		toggleViewButton,
 		optionsButton,
-		newFolderButton,
 	)
 
 	header := container.NewBorder(nil, nil, nil, optionsbuttons,
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -233,9 +233,30 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 		}
 	})
 
+	newFolderButton := widget.NewButtonWithIcon("", theme.FolderNewIcon(), func() {
+		newFolderEntry := widget.NewEntry()
+		ShowForm("New Folder", "Create Folder", "Cancel", []*widget.FormItem{
+			{
+				Text:   "Name",
+				Widget: newFolderEntry,
+			},
+		}, func(s bool) {
+			if !s || newFolderEntry.Text == "" {
+				return
+			}
+
+			createFolderErr := os.MkdirAll(filepath.Join(f.dir.Path(), newFolderEntry.Text), 0700)
+			if createFolderErr != nil {
+				ShowInformation("Error", "Folder cannot be created", f.file.parent)
+			}
+			f.refreshDir(f.dir)
+		}, f.file.parent)
+	})
+
 	optionsbuttons := container.NewHBox(
 		toggleViewButton,
 		optionsButton,
+		newFolderButton,
 	)
 
 	header := container.NewBorder(nil, nil, nil, optionsbuttons,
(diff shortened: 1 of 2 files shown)


Jacalz (changes requested):
Thanks for opening this PR. Please note that PRs should target the `develop` branch as that is where all feature development happens.

Also, would you mind sharing an image of what your change looks like? While we always can compile the code and test ourselves, it’s always useful to have an image :)

Beebeeoii:
@Jacalz my bad ! thank you for the reminder... will make the appropriate changes.

Beebeeoii:
Here are some screenshots of how it looks like:

Folder Dialog:
image

Inputting new folder name:
image

Workflow:
Peek 2022-08-04 08-45

andydotxyz:
Thanks so much for this.
I have a question for folk following this - should the add folder be further left? (leaving settings as the right most item).
So the left or middle of the 3?... personally I wonder if being 1st might make more sense, but curious to see what others think.

Beebeeoii:
@andydotxyz yup i think makes more sense for settings to be the right most. Maybe we can try putting create folder to be the left most?

Jacalz:
I agree. It would make more sense if it was on the left :)

Beebeeoii:
just updated this MR to reflect the change. The folder dialog looks something like this now:

image

dialog/file.go
```diff
@@ -233,7 +233,28 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
...
+				Text:   "Name",
+				Widget: newFolderEntry,
+			},
+		}, func(s bool) {
+			if !s || newFolderEntry.Text == "" {
+				return
+			}
+
+			createFolderErr := os.MkdirAll(filepath.Join(f.dir.Path(), newFolderEntry.Text), 0700)
+			if createFolderErr != nil {
```
Jacalz, line 249: It would be good if we could log the error using `fyne.LogError`.

Jacalz (changes requested):
Thanks. I have an improvement and one question inline.

dialog/file.go
```diff
@@ -233,7 +233,28 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
...
+				ShowInformation("Error", "Folder cannot be created", f.file.parent)
```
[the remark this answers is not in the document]
Beebeeoii: Ah yes. `ShowError()` is indeed better for this context. However, I would propose that we have a short error message like "Folder cannot be created" as users might not be able to understand the technical error message generated by `createFolderErr`

dialog/file.go
Beebeeoii: thanks ! missed that one out.

Beebeeoii:
> Also, you need to update the tests, because they are failing.

Hey @Jacalz thanks for reviewing my PR. I have just got to updating the tests and also included one test case for folder creation dialog. TIA !

dialog/file.go
```diff
@@ -233,7 +233,33 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
...
+				Text:   "Name",
+				Widget: newFolderEntry,
+			},
+		}, func(s bool) {
+			if !s || newFolderEntry.Text == "" {
+				return
+			}
+
+			newFolderPath := filepath.Join(f.dir.Path(), newFolderEntry.Text)
+			createFolderErr := os.MkdirAll(newFolderPath, 0000)
```
Jacalz, line 249: The permission values do not seem correct to me. I think you likely want to use 0700 instead.

Jacalz (changes requested):
Thanks. This is very close now :)

dialog/file.go
andydotxyz: Sorry to be a pain, but shouldn't it be 0733? These folders are not restricted to user home folders so they could be writing files that other users should see...

andydotxyz:
Looks great, but inline note about group/other read&list permission for new folders.

dialog/file.go
Beebeeoii: thanks for pointing that out. actually i would think that it should be 0755? we would like the newly created folders to be seen by others (via `ls`) and that others can `cd` into those directories. these corresponds to `r` and `x` respectively. what we might not want is for others to modify the directory's contents (e.g. renaming of subfiles), which corresponds to `w`. `r-x` corresponds to 0755.

honestly, it might be better to have permissions set to 0777 since like you pointed out, they might be creating folders for others to write to as well, and fyne shouldn't restrict on that?

not sure if i grasped the concepts correctly but these are my thoughts. thanks a bunch !

andydotxyz: You're quite right - I meant 755. However you could be on to something with 777. If the system umask applies to this it would make complete sense. What do you think @Jacalz?

Jacalz: Hmm. I had a look at what `gosec` is recommending and I think they don’t want anything higher than 0750. Will have a closer look at that later today.

Jacalz: I ran the numbers through a Unix permission calculator and I think 0750 makes sense. That means read/write/execute for user and read/execute for group with none for other.

Beebeeoii: agree with that reasoning

andydotxyz (approved):
Looks great, are you around @Jacalz to approve the change you requested?

Jacalz:
I will have a look later today 🙂

Jacalz (approved):
Looks good to me. Thanks for the PR :)

PR #3179, opened 2022-08-03
Cleanup of glfw driver window files

Jacalz:
### Description:

This PR cleans up some of the maps that are used in the `internal/driver/glfw/window_*`.go files.
- De-duplicate a some code between desktop and goxjs.
- Remove trivial `keyCodeMapASCII` in favour of a function with simple rune value manipulation.
- Replace  `keyCodeMap` (it was basically just checking mapping "a" to "a" to check if it existed) with a switch statement instead. Benchmark shows about a 3x improvement.
- Use a function for `keyNameMap` instead. Allows for some clever ASCII value mappings in intervals.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window_common.go
+++ b/internal/driver/glfw/window_common.go
@@ -6,6 +6,26 @@ import (
 	"github.com/go-gl/glfw/v3.3/glfw"
 )
 
+// Cursor mode values.
+const (
+	CursorNormal   int = glfw.CursorNormal
+	CursorHidden   int = glfw.CursorHidden
+	CursorDisabled int = glfw.CursorDisabled
+)
+
+var cursorMap map[desktop.StandardCursor]*glfw.Cursor
+
+// Input modes.
+const (
+	CursorMode             glfw.InputMode = glfw.CursorMode
+	StickyKeysMode         glfw.InputMode = glfw.StickyKeysMode
+	StickyMouseButtonsMode glfw.InputMode = glfw.StickyMouseButtonsMode
+	LockKeyMods            glfw.InputMode = glfw.LockKeyMods
+	RawMouseMotion         glfw.InputMode = glfw.RawMouseMotion
+)
+
+const defaultTitle = "Fyne Application"
+
 var keyCodeMap = map[glfw.Key]fyne.KeyName{
 	// non-printable
 	glfw.KeyEscape:    fyne.KeyEscape,
--- a/internal/driver/glfw/window_common.go
+++ b/internal/driver/glfw/window_common.go
@@ -1,10 +1,5 @@
 package glfw
 
-import (
-	"fyne.io/fyne/v2"
-	"github.com/go-gl/glfw/v3.3/glfw"
-)
-
 const defaultTitle = "Fyne Application"
 
 func unknownKey(key string) bool {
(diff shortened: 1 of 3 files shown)


Jacalz:
This should likely be working now. The fact that `glfw-js` is importing a package named `glfw` made me think that I could just move the code out to a common file but that would mean importing a different `glfw` package.

Bluebugs:
> This should likely be working now. The fact that `glfw-js` is importing a package named `glfw` made me think that I could just move the code out to a common file but that would mean importing a different `glfw` package. Confusing.

The things is that `glfw-js` is wrapping `glfw`, so theoretically we could just lean completely rely on glfw-js for everything. I haven't done that as it was missing some bits and I was afraid of breaking things, but it is still a possible path forward.

internal/driver/glfw/window_common.go
```diff
@@ -0,0 +1,36 @@
...
+	'`':  {},
+
+	';': {},
+	'+': {},
+	'=': {},
+
+	'[':  {},
+	'\\': {},
+	']':  {},
+}
```
Jacalz, line 36: This feels ugly to me. It might be worth it to just move the checks into the function but I am honestly not perfectly happy with that approach either.

There is some overlap in the ascii table that avoids having to check each character individually though.

Jacalz: I removed it entirely. There are quite a few overlaps and it made sense to handle them in code instead of using a map.

internal/driver/glfw/window_goxjs.go
```diff
@@ -388,8 +311,7 @@ func keyToName(code glfw.Key, scancode int) fyne.KeyName {
 	}
 
 	//	keyName := glfw.GetKeyName(code, scancode)
-	//	ret, ok = keyNameMap[keyName]
-	//	if !ok {
+	// if unknownKey(keyName) {
```
changkun, line 314: Are these commented code remain necessary to stay here?

changkun (commented):
Very interesting approach to do a key mapping by inference.
I wondered how frequent these functions get called.

Key mapping is always a pain because there are always uncovered special cases, and building a configuration table was the most commonly used approach.

Furthermore, what would happen if we had a non-ASCII keyboard? Are we generally constrained by glfw keymapping?

internal/driver/glfw/window_goxjs.go
Jacalz: I think they are supposed to be fixed sometime in the future when the web driver gets more features. Am I right, @Bluebugs?

Bluebugs: Yes, I still need to investigate that code path.

Jacalz:
Hmm. I just tried to improve what we already had but still make it give the same results. I must admit that I don't know the code well enough to tell how non-ASCII keyboards and glfw keymappings fit into it all.

Jacalz:
From what I can tell from looking at https://github.com/golang/go/commit/844b410922b2ea7d22b651c536178e79696749f6 and the linked conversation within, it seems like the best way to do this kind of mapping (as https://github.com/fyne-io/fyne/blob/638ae24e4bebfc0b6dbaabf8db5807eeabfafc61/internal/driver/glfw/window_desktop.go#L392) actually is using a switch statement. It can be optimized by the compiler but it can not prove that the map is never mutated, The switch statement should also be a lot faster in Go 1.19 due to it using jump tables for ints (glfw keys in this case) and strings.

Bluebugs (approved):

PR #3185, opened 2022-08-05
Call ExtendBaseWidget in form.Refresh method, fixes #3153

### LINKED ISSUE #3153: Segfault when adding items directly in form struct
### Checklist

### Describe the bug

I get a SEGFAULT when calling `.Refresh()` on a form where all items were all added in the struct creation.

I followed [this example](https://github.com/fyne-io/fyne/blob/master/cmd/fyne_demo/tutorials/widget.go#L306), but added everything as `Items` when creating the struct -- so no `form.Append()`.

This issue may be the same as [#1967](https://github.com/fyne-io/fyne/issues/1967), but I'm not sure.

The problem seems to be when [ensureRenderItems](https://github.com/fyne-io/fyne/blob/master/widget/form.go#L221) assumes `itemGrid` is not null. This is not the case when adding items directly to the form struct.

If adding items directly is not recommended, maybe the tutorial could be updated.
But a guard on this code would probably be safer.

### How to reproduce

Call `start(false)` on the example code below and it works fyne.
Call `start(true)` on and it segfaults.

### Screenshots

_No response_

### Example code

```go
import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func start(segfault bool) {
	a := app.New()
	w := a.NewWindow("Manny")

	name := widget.NewEntry()
	name.SetPlaceHolder("John Smith")
	item := &widget.FormItem{Text: "Name", Widget: name}

	form := &widget.Form{}
	if segfault {
		form.Items = []*widget.FormItem{item}
	} else {
		form.AppendItem(item)
	}
	form.Refresh()

	w.SetContent(form)
	w.ShowAndRun()
}
```

### Fyne version

v2.2.3

### Go compiler version

go1.18.4

### Operating system

macOS

### Operating system version

11.6.5

### Additional Information

_No response_

fpabl0:
### Description:

Fixes #3153 

Calling `ExtendBaseWidget` in `form.Refresh` is needed to ensure `f.super()` returns a non-nil widget (when initializing Form through struct fields without calling any method that actually call `ExtendBaseWidget`), otherwise `cache.Renderer` will not be able to create any renderer.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/form.go
+++ b/widget/form.go
@@ -89,6 +89,7 @@ func (f *Form) MinSize() fyne.Size {
 
 // Refresh updates the widget state when requested.
 func (f *Form) Refresh() {
+	f.ExtendBaseWidget(f)
 	cache.Renderer(f.super()) // we are about to make changes to renderer created content... not great!
 	f.ensureRenderItems()
 	f.updateButtons()
--- a/widget/form_test.go
+++ b/widget/form_test.go
@@ -416,3 +416,15 @@ func TestForm_SetOnValidationChanged(t *testing.T) {
 	assert.False(t, validationError)
 
 }
+
+func TestForm_RefreshFromStructInit(t *testing.T) {
+	form := &Form{
+		Items: []*FormItem{
+			{Text: "Entry", Widget: NewEntry()},
+		},
+	}
+
+	assert.NotPanics(t, func() {
+		form.Refresh()
+	})
+}


andydotxyz (approved):
great catch thanks

PR #3190, opened 2022-08-06
Remove prompt so that the copy button copies a valid command

rexposadas:
### Description:

I would like to the copy buttons to copy a valid command. This would mean removing the prompt from what is copied.


andydotxyz (approved):
Thanks

PR #3202, opened 2022-08-12
Minor refactor and cleanup of dialog code

Jacalz:
### Description:

This is a small refactor and cleanup of the dialog code. It just renames a method (it was doing a lot more than just setting up buttons) and removes some unnecessary logic (creating an image with nil icon and an image with no icon is the same thing).

### Checklist:

- [ ] Tests included.
  - Already covered by tests.
- [x] Tests all pass.


andydotxyz (approved):
:+1:

PR #3215, opened 2022-08-21
fix(bug): #3212

### LINKED ISSUE #3212: NewCheckWithData().OnChanged not synchronized with its binding data
### Checklist

### Describe the bug

When the widget "NewCheckWithData" is changend and its function "OnChanged" is called the corresponding binding data ist not set corret respectively the function "Get()" does not return the correct value.

Maybe it is a timing problem but calling "waitForBinding()" does not help at this stage.

### How to reproduce

Run the code provided and the result is:
Value start (checkbox): true
Value start (binding):  true
Value change to (param): false
Value change to (checkbox): false
Value change to (binding):  **true** 

The binding value must be "false" instead of "true" after unchecking the checkbox.

### Screenshots

_No response_

### Example code

package main

import (
	"fmt"
	"time"

	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

// Is nowhere really described but probably necessary
// See https://github.com/fyne-io/fyne/blob/master/widget/check_test.go
func waitForBinding() {
	time.Sleep(time.Millisecond * 100) // data resolves on background thread
}

func main() {
	a := app.New()

	bindValue := binding.NewBool()
	err := bindValue.Set(true)
	if err != nil {
		fmt.Println("Error")
	}

	checkBoxBind := widget.NewCheckWithData("Binding checkbox", bindValue)
	waitForBinding() // must be run otherwise the checkBoxBind.Checked has the wrong value

	checkValue, _ := bindValue.Get()
	fmt.Printf("Value start (checkbox): %t\n", checkBoxBind.Checked)
	fmt.Printf("Value start (binding):  %t\n", checkValue)

	checkBoxBind.OnChanged = func(checked bool) {
		fmt.Printf("Value change to (param): %t\n", checked)
		fmt.Printf("Value change to (checkbox): %t\n", checkBoxBind.Checked)
		checkValue2, _ := bindValue.Get()
		fmt.Printf("Value change to (binding):  %t\n", checkValue2)
	}

	w.SetContent(checkBoxBind)
	w.ShowAndRun()
}

### Fyne version

2.2.3

### Go compiler version

1.18.3

### Operating system

Windows

### Operating system version

Windows 10 Enterprise 20H2 (19042.1826)

### Additional Information

_No response_

QIN2DIM:
### Description:

Fixes #3212 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/check.go
+++ b/widget/check.go
@@ -126,7 +126,7 @@ func (c *Check) Bind(data binding.Bool) {
 	}
 }
 
-// SetChecked sets the the checked state and refreshes widget
+// SetChecked sets the checked state and refreshes widget
 func (c *Check) SetChecked(checked bool) {
 	if checked == c.Checked {
 		return
@@ -135,6 +135,7 @@ func (c *Check) SetChecked(checked bool) {
 	c.Checked = checked
 
 	if c.OnChanged != nil {
+		c.binder.CallWithData(c.writeData)
 		c.OnChanged(c.Checked)
 	}
 


andydotxyz (commented):
This does not make sense. The `OnChanged` after `Bind` is called will call `CallWithData` automatically.

andydotxyz:
As discussed this is not the right fix. We should avoid overloading `OnChanged` rather than second guessing whether it has been replaced. The proposed change will call the setter twice in normal usage.

PR #3218, opened 2022-08-23
improve the color dialog

zivoy:
### Description:

Fixed a crash when calling SetColor without calling Refresh first.
I also fixed the advanced mode color selection being added to the history when clicking the color palette.

some other things changed:
- made history max length seven so it will be the same length as the color palette and won't leave a single color on the next line
- changed the middle gray in the grayscale colour picker to be closer to halfway (.5 * 255)
- made the advanced mode accordion keep its state after a refresh
- added refreshing after writing to history so that the history will show when the same dialog is opened next time
- set the color on advanced mode when selecting from color palette
- updated fyne demo so that it doesn't recreate the advanced mode colour selector every time the button is clicked

~~Fixes~~
fixed it myself instead of opening an issue

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


zivoy:
I added a proper colour preview and a background to the colour wheel
image

andydotxyz:
Thanks for this, unfortunately with other work going on there are conflicts on this files.
Can we request please that you split the fix and the new features into separate changes? It is easier to review them separately.
Given the theme changes moving I would recommend starting 2 new, smaller, PRs on latest develop instead of continuing this one, sorry.

zivoy:
split the pr into the 2 parts as requested, see #3233 and #3234

PR #3231, opened 2022-08-29
New theme removes separator in collections for padding

andydotxyz:
This will make some people happy :)
Also adapts accordion padding, which loops in someones suggestion that the content touching next header didn't look great.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
Oh no, it merged in my change to have a light coloured scrollbar on dark theme as well.
Not what I intended, but they are all theme fixes so I think it's not worth unpicking it all.

Jacalz (commented):
This is looking very good but I do have a suggestion for a potential improvement.
The image below is how it currently looks:
image

This image is instead using the button colour as the separator colour. I rather like how this looks. I suspect that you could get by with the colour and then making the separators thinner. Thoughts?
image

Jacalz (approved):
I realise now that the latter makes it look more like how it was before. Is that a good think or a bad thing? Either way, I will approve this just in case :)

andydotxyz:
Ah the colour I used was in the designs we have for separating, so figured I would use it here to replace the old (disabled?) colour.

PR #3233, opened 2022-09-01
New features for color dialog

zivoy:
I am splitting #3218 into multiple parts, as requested by @andydotxyz.
This pull request contains the new additions, which are:
- a background for semi-transparent colours in the colour picker wheel
- a colour picker component that shows the last selected colour


dialog/color.go
```diff
@@ -166,16 +167,30 @@ func newColorButtonBox(colors []color.Color, icon fyne.Resource, callback func(c
 	return container.NewGridWithColumns(8, objects...)
 }
 
-func newCheckeredBackground() *canvas.Raster {
-	return canvas.NewRasterWithPixels(func(x, y, _, _ int) color.Color {
-		const boxSize = 8
+const (
```
andydotxyz, line 170: Can you move the const to the top of the file? We try to keep things in order as: https://github.com/fyne-io/fyne/wiki/Contributing

dialog/color.go
```diff
@@ -166,16 +167,30 @@ func newColorButtonBox(colors []color.Color, icon fyne.Resource, callback func(c
...
+	boxSize       = 8
+	numberOfRings = 12
+)
 
+func newCheckeredBackground(radial ...bool) *canvas.Raster {
```
andydotxyz, line 175: This looks confusing, I appreciate that it's to keep the old call consistent but as it's internal it can be changed and I think it would be cleaner as just `bool`.

dialog/color.go
```diff
@@ -166,16 +167,30 @@ func newColorButtonBox(colors []color.Color, icon fyne.Resource, callback func(c
...
+	rect := func(x, y, _, _ int) color.Color {
```
andydotxyz, line 176: it seems wasteful to create `rect` then `f` then re-assign it later if `radial`... If you put an `if radial` at the top then you could simply return this radial version, avoiding both branches running.

andydotxyz (changes requested):
Very nice additions, just a few notes inline about code style.

dialog/color.go
zivoy: yes, but both branches need to run for radial anyway and this way the colours for the boxes are only in one place.
what do you think about moving the part that makes the function for `canvas.NewRasterWithPixels` to be another function and having it make a loop of recursion?
something like
```go
func newCheckeredBackground(radial bool) *canvas.Raster {
	return canvas.NewRasterWithPixels(checkeredBackgroundRasterFunction(radial))
}

func checkeredBackgroundRasterFunction(radial bool) func(x, y, w, h int) color.Color {
	if !radial {
		return func(x, y, _, _ int) color.Color {
			if (x/checkeredBoxSize)%2 == (y/checkeredBoxSize)%2 {
				return color.Gray{Y: 58}
			}

			return color.Gray{Y: 84}
		}
	}

	rect := checkeredBackgroundRasterFunction(false)
	return func(x, y, w, h int) color.Color {
		r, t := cmplx.Polar(complex(float64(x)-float64(w)/2, float64(y)-float64(h)/2))
		x = int((t + math.Pi) / (2 * math.Pi) * checkeredNumberOfRings * checkeredBoxSize)
		y = int(r)
		return rect(x, y, 0, 0)
	}
}
```

andydotxyz: That is very clever, but I think a little confusing to read.
You were probably right to declare the function as `rect` initially, but then it is easier to read if you `if !radial {return new raster etc...}`,  then the rest of the method sets up the radial and you can return the `f` func as an anonymous pass to the raster init.

Though as you point out it is more complex than I realised, so the initial implementation isn't so far from the simplest as I thought, sorry. But in general returning early is a good thing :).

andydotxyz:
Close enough ;) will need to resolve the merge conflict on those test images before we can test and approve, sorry.
You can just move the failed image over the master (which I am sure you know by now).

andydotxyz (approved):

PR #3234, opened 2022-09-01
bug fixes for color dialog

zivoy:
The second part after splitting #3218.
this pull request contains the bug fixes, which are:
- fixing a crash when calling `SetColor` after setting `Advanced` to true without calling `Refresh`
- when selecting a colour from the colour palette, the current advanced selection no longer gets added to the history as well
- changed the history length to 7 so it lines up with the colour palette
- changed the middle gray to be closer to halfway up the grayscale scale
- made the advanced mode accordion keep its state after a refresh
- refresh when writing to history so that when the same dialog is reshown, the history will be up to date 
- update the selected colour in advanced mode after selecting from the colour palette
- made the fyne demo not recreate the advanced mode dialog every time it's opened to showcase the fixes
- fixed the checkered background on the colour palette/history
- shrank the colour selection buttons to 30x30, so they line up with the checkered background


dialog/color_button.go
```diff
@@ -95,19 +94,21 @@ type colorButtonRenderer struct {
 func (r *colorButtonRenderer) Layout(size fyne.Size) {
 	r.rectangle.Move(fyne.NewPos(0, 0))
 	r.rectangle.Resize(size)
+	r.background.Resize(size)
 }
 
 func (r *colorButtonRenderer) MinSize() fyne.Size {
-	return r.rectangle.MinSize().Max(fyne.NewSize(32, 32))
+	return r.rectangle.MinSize().Max(fyne.NewSize(30, 30))
```
andydotxyz, line 101: I think this is unrelated, the checked pattern is not part of this PR

andydotxyz (changes requested):
Looking good, easier to test these fixes - all great catches thanks.
Not sure about resizing the items though, wasn't that related to the hatching pattern in the other PR?

dialog/color_button.go
zivoy: The other pr is for the circular hatched pattern. This is to make the hatch pattern on the color buttons fit the box
As I mentioned in the discord, the code to add the pattern was already there, it just wasn't showing since it wasn't being resized to fit the area in the layout function (line 97)

andydotxyz: > The other pr is for the circular hatched pattern. This is to make the hatch pattern on the color buttons fit the box As I mentioned in the discord, the code to add the pattern was already there, it just wasn't showing since it wasn't being resized to fit the area in the layout function (line 97)

Aha I understand, sorry I missed that before.
Would it not be a smaller change to set the squares of the hatching to 8pt instead of 10? Then the UI layout remains the same but the bug fix focuses on the presentation, which is what you are fixing...

zivoy: That is also an option, though if this is done the advanced picker needs to be scaled so the preview won't be cut off. But I can do that and then just leave it like that and have it be fixed in the other pr where I can control the size of the preview directly

andydotxyz: Not quite sure I follow, sorry - are you saying there is an element of the advanced is aligned to the 10pt multiple?

zivoy: Nevermind, I thought it fit perfectly on this, I will change it to be 8
image

andydotxyz (approved):
excellent thanks

PR #3244, opened 2022-09-05
Replace theme maps with switch statements

Jacalz:
### Description:

This refactors and cleans up some of the theme code to use use functions with switch statements instead of static maps.
It has been said before but static maps can not be optimised by the compiler because it can not prove that they never change.
This decreases binary size by about 0,06% and is much more efficient.

### Checklist:

- [ ] Tests included.
- Should already be covered. It is also hard to write useful tests for functions that only switch on inputs.
- [x] Tests all pass.

### DIFF
--- a/theme/theme.go
+++ b/theme/theme.go
@@ -639,7 +639,7 @@ func currentVariant() fyne.ThemeVariant {
 	return fyne.CurrentApp().Settings().ThemeVariant()
 }
 
-func primaryColorNamed(name string) color.Color {
+func primaryColorNamed(name string) color.NRGBA {
 	switch name {
 	case ColorRed:
 		return color.NRGBA{R: 0xf4, G: 0x43, B: 0x36, A: 0xff}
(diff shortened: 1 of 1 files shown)


andydotxyz (changes requested):
Looks like a good change, but the functions are out of order

Jacalz:
Good point. I have fixed it now

Jacalz:
I have updated the PR to also include work for the icons code. However, I am not entirely sure why the tests are failing. Nothing seems incorrect to me.

andydotxyz:
The tests look like they are failing because the test render to xml icon name lookup is broken

Jacalz:
It feels strange that it would happen so suddenly. Will update the branch against develop and see if that changes anything (probably not).

andydotxyz:
Feels vaguely logical from an outside perspective as the icon lookup code was changed in the theme package here - maybe test package needs to be updated to match?

andydotxyz:
A really I figured out what it is - you moved from static resources (only ever one) to dynamic where it creates a new resource every time asked. So the test lookup is no longer getting a true when it compares a resource to the canonical call (as they are now different and indeed duplicate).

This actually seems like unnecessary object creation, perhaps this form of switch is not desirable.

Jacalz:
I think you are right. The creation should be relatively cost-less because the function calls are inlined, but you are right that it isn't ideal. I will revert the commits about that tomorrow.

Jacalz:
I have reverted it now. Should be good to go for a re-review :)

andydotxyz (approved):

Issue #3024, opened 2022-06-01
Caret position lost when resizing a MultilineEntry


matwachich:
### To Reproduce:
1. Run the example
2. Position the caret somewhere in the text
3. Resize the window
4. Try to type something and see the changed caret position

### Example code:
```go
package main

import (
	lorem "github.com/drhodes/golorem"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Tester")

	e := widget.NewMultiLineEntry()
	e.Wrapping = fyne.TextWrapWord
	e.SetText(lorem.Paragraph(10, 15))

	w.SetContent(e)
	w.ShowAndRun()
}
```

### Device (please complete the following information):
 - **OS:** Windows 10
 - **Version:** 10.0.18363.2212
 - **Go version:** 1.18.2
 - **Fyne version:** 2.1.4

andydotxyz:
Resolved on `develop` branch ready for release

PR #3259, opened 2022-09-13
Move to the shared go-text project for our text shaping

### LINKED ISSUE #1623: Text shaping for Unicode strings

Take this code:
```
// main.go:
package main

import (
	"fyne.io/fyne/app"
	"fyne.io/fyne/widget"
)

func main() {
	myApp := app.New()
	myWindow := myApp.NewWindow("Window")
	text := "سلام"
	myWindow.SetContent(widget.NewLabel(text))

	myWindow.ShowAndRun()
}
```
Use a Persian font like [Nazli](https://candyfonts.com/font/nazli.htm) and run it:
`$ FYNE_FONT=Nazli.ttf go run main.go`
You'll get this:
s
Which is just the default glyphs of the input Unicode characters shown left to right and is simply wrong. What it would look like if it is shown right to left after substituting correct glyphs for characters depending on surrounding characters and changing the spacing of each glyph correctly:
s2

I don't think the API needs to change. The way text is rendered in the internal parts is the issue.

Using go bindings of a text rendering library like [Pango](https://pkg.go.dev/github.com/jmoiron/gotk3/pango)

andydotxyz:
A few things still to do, but this is using the go-text so we can base better text handling and i18n.

- [x] Update some broken tests
- [ ] Create a go-text renderer and use it - not per-rune but per-shaping.Out - moved to #3338
- [x] Demonstrate the value add to justify the increase in file size...

```
-rwxr-xr-x   1 andy  staff    15M 13 Sep 19:32 hello-develop
-rwxr-xr-x   1 andy  staff    18M 13 Sep 19:31 hello-gotext
```

This is an increase of 18%. We should see if anything can be done to reduce the incrase

Fixes #1623 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
There is a broken test that I have part disabled, it was not passing and/or is this issue: #3024

Bluebugs:
I just tested on my laptop and the screen which has a different scale factor does not behave correctly. The object around the text are properly sized, but the font size rest stuck as if scale was still one.

andydotxyz:
Thanks @Bluebugs that should be fixed now. As is the GopherJS issue.
There are further changes to be included from go-text project so I will make sure that the replace directive is removed before we release!

Bluebugs (approved):

PR #3262, opened 2022-09-15
Optimize GoString() method of StaticResource

Jacalz:
### Description:

I noticed that the `serialise.go` file used `buffer.WriteString(fmt.Sprint(v))` instead of `fmt.Fprint(&buffer, v)` and also used `bytes.Buffer` instead of `strings.Builder`. This improves speed by about 20%.

```
BenchmarkToGoBefore-8        	110071	     10744 ns/op	    3425 B/op	     100 allocs/op
BenchmarkToGoAfter-8          	132193	      8545 ns/op	    1456 B/op	       6 allocs/op
```

I am not entirely sure how much this is used. Should we consider deprecating it?

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz (approved):
Cool thanks

PR #3263, opened 2022-09-15
Fix RichText with two mulitline in a row

### LINKED ISSUE #3117: Text line not displayed in RichText
### Describe the bug:
In a RichText, when adding an Inline TextSegment{} with a .Text finishing with a line break (\r\n or \n) ; and just after, a TextSegment{} that have a multiline .Text (containing \r\n or \n), then the first line of that second TextSegment{} is not displayed.

It can get displayed after resetting the Segments and calling Refresh on the RichText

### To Reproduce:
Run the code

### Screenshots:
image

### Example code:
```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("TEST")

	rt := widget.NewRichText(
		&widget.TextSegment{
			Text: "A Title\r\n",
			Style: widget.RichTextStyle{
				Inline:    true,
				TextStyle: fyne.TextStyle{Bold: true},
			},
		},
		&widget.TextSegment{
			Text: "Here is some text,\r\nwith a line break",
		},
	)

	btn := widget.NewButton("Refresh", func() {
		rt.Segments = []widget.RichTextSegment{
			&widget.TextSegment{
				Text: "A Title\n",
				Style: widget.RichTextStyle{
					Inline:    true,
					TextStyle: fyne.TextStyle{Bold: true},
				},
			},
			&widget.TextSegment{
				Text: "Here is some text,\nwith a line break",
			},
		}
		rt.Refresh()
	})

	w.SetContent(container.NewBorder(nil, btn, nil, nil, rt))
	w.ShowAndRun()
}

```

 - **OS:** Windows 10
 - **Go version:** 1.18.2
 - **Fyne version:** 2.2.2

### Workaround
I noticed that you can avoid this bug by removing the trailing \r\n from the title, and placing it the begening of the multiline text
image

NicolasPerrenoud:
### Description:
Rich text with multiple segments with multiple lines in a row are not displayed correctly. Because wrong line was selected in the cache.

Fixes #3117

### Checklist:
- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Great catch, thanks!

PR #3266, opened 2022-09-16
add time aware RuneEvent

dmowcomber:
### Description:
Add RuneEvent type and SetOnTtpedRuneEvent func to have a time aware typed rune logic. 
this is useful when looking for a Rune that was typed after the user was prompted for a key press.
Runes typed before of after a prompt can be ignored by comparing the RuneEvent.When() time with the time of a user prompt. 

WIP. New tests need to be added and existing tests need to be fixed…

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/canvas.go
+++ b/canvas.go
@@ -39,6 +39,7 @@ type Canvas interface {
 
 	OnTypedRune() func(rune)
 	SetOnTypedRune(func(rune))
+	SetOnTypedRuneEvent(func(*RuneEvent))
 	OnTypedKey() func(*KeyEvent)
 	SetOnTypedKey(func(*KeyEvent))
 	AddShortcut(shortcut Shortcut, handler func(shortcut Shortcut))
--- a/event.go
+++ b/event.go
@@ -1,5 +1,7 @@
 package fyne
 
+import "time"
+
 // HardwareKey contains information associated with physical key events
 // Most applications should use KeyName for cross-platform compatibility.
 type HardwareKey struct {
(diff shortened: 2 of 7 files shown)


andydotxyz:
I'm not sure I completely grasp the use-case here. Is the order of event guarantee not sufficient? Or the Validator system? 
Adding a new type and event hook for this purpose where it benefits only runes seems strange - why is it needed and/or why don't keys, taps etc. need it too?

dmowcomber:
I'm attempting to make an application similar to a speak and spell but with a single letter to help kids lean their letters.
I'm not familiar with the Validator System you mentioned.
Good point if this is to be added for Runes it should probably be added to all key press types.
Although I realize now that a time based event can be added to the caller side with something like this:
```
	go func() {
		w.Canvas().SetOnTypedRune(func(char rune) {
			runeEvents <- &runeEvent{
				char:      char,
				eventTime: time.Now(),
			}
		})
	}()
```
I had TTS speech calls being made in my `SetOnTypedRune` func previously which were slowing the handling of those key presses previously. Separating the logic helps.
So maybe this logic isn't needed in the Fyne code base.

andydotxyz:
I think what you describe makes sense. I general putting slow/expensive code in callbacks is a bad idea and spawning goroutines or passing down a channel is a clean solution.
Best of luck with your app!

dmowcomber:
Thanks for the feedback!

PR #3268, opened 2022-09-18
Improve and optimize the bundling code

Jacalz:
### Description:

This cleans up, uses more performant APIs and adds a test for bundling files. The end result is less code and 30% faster bundling per resource (should be a larger difference for large bundled contents). Benchmark results are as follows:
```
BenchmarkWriteResourceOld-8   	 36675	     28070 ns/op	    1232 B/op	      12 allocs/op
BenchmarkWriteResourceNew-8      62576	     19818 ns/op	     960 B/op	       9 allocs/op

```

It also cleans up some of the metadata code. This now uses the .GoString() method of *StaticResource (that recently was sped up by 30%) to add the icon contents directly inline instead of to a global variable. It also has the benefit of not needing a []byte("") conversion as gopls never touches the temporary file (the normal bundling uses strings because gopls leaks memory otherwise).
The second change is avoiding casting []byte() to a string in the metadata loading by using a better choice of APIs.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


cmd/fyne/internal/commands/build.go
```diff
@@ -280,18 +280,21 @@ func createMetadataInitFile(srcdir string, app *appData) (func(), error) {
...
-		if app.icon != "" {
-			writeResource(app.icon, iconResName, metadataInitFile)
-		} else {
-			v := fmt.Sprintf("var %s = (fyne.Resource)(nil)\n", iconResName)
-			_, err = metadataInitFile.Write([]byte(v))
-			if err != nil {
-				fyne.LogError("Error writing icon placeholder", err)
-			}
+	app.ResGoString = "nil"
+	if app.icon != "" {
```
andydotxyz, line 284: It feels like this is the wrong way round - should an icon specified on the command line not override the one in the metadata?

andydotxyz (changes requested):

cmd/fyne/internal/commands/build.go
Jacalz: I am not sure I follow. If the `app.icon` fields points to a file, we load the file and embed the icon data. This seems to me like it is the same behaviour as before or am I missing something?

andydotxyz (approved):
Sorry I missed that the metadata read and file output for icon were different passes

Jacalz:
No worries. Thanks for the approval 🙂

PR #3273, opened 2022-09-19
Add lock protection for map read/write operations

### LINKED ISSUE #3269: fatal error: concurrent map read and map write
### Checklist

### Describe the bug

On some occasion when I run my app I experience the following issue: 
```
fatal error: concurrent map writes
...
goroutine 42 [running]:
fyne.io/fyne/v2/internal/painter.CachedFontFace({0x0, 0x0, 0x0, 0x0, 0x0}, 0xc000fcd1b8)
    /Users/vk/Work/Languages/Go/gopath/pkg/mod/fyne.io/fyne/v2@v2.2.3/internal/painter/font.go:70 +0x4a5
```
or
```
fatal error: concurrent map read and map write
...
fyne.io/fyne/v2/internal/painter.CachedFontFace({0x0, 0x0, 0x0, 0x0,0x0}, 0xc0005eb2e0)
   /Users/vk/Work/Languages/Go/gopath/pkg/mod/[2]fyne.io/fyne/v2@v2.2.3/internal/painter/font.go:64 +0x314
```

Upon code inspection, I found that indeed I see that the following lines:
https://github.com/fyne-io/fyne/blob/master/internal/painter/font.go#L59
https://github.com/fyne-io/fyne/blob/master/internal/painter/font.go#L64
https://github.com/fyne-io/fyne/blob/master/internal/painter/font.go#L70
will lead to this issue since the update in map is not protected.

### How to reproduce

There is no clear way to reproduce the problem, but upon code inspections it is clearly see where the bug is, see
https://github.com/fyne-io/fyne/blob/master/internal/painter/font.go#L59
https://github.com/fyne-io/fyne/blob/master/internal/painter/font.go#L64
https://github.com/fyne-io/fyne/blob/master/internal/painter/font.go#L70
which will lead to unprotected map updates.

### Screenshots

_No response_

### Example code

There is no clear way to reproduce the issue.

### Fyne version

v2.2.3

### Go compiler version

1.19

### Operating system

macOS

### Operating system version

11.6.8

### Additional Information

The code itself reveals the issue as used map is not protected on read/write operations and will lead to aforementioned issue in concurrent environment.

vkuznet:
### Description:

Add lock protection for map read/write operations. I made it backward compatible with older go release. Starting with Go 1.19 these changes may be replaced with `sync.Map` (concurrent map).

Fixes #3269

### Checklist:

- [ ] Tests included. **I'm not aware of reliable test procedure for concurrent map read/write operations.**
- [x] Tests all pass.

### DIFF
--- a/internal/painter/font.go
+++ b/internal/painter/font.go
@@ -61,18 +61,17 @@ func CachedFontFace(style fyne.TextStyle, opts *truetype.Options) font.Face {
 	}
 
 	comp := val.(*fontCacheItem)
-	lock := sync.RWMutex{}
-	lock.RLock()
+	comp.RLock()
 	face := comp.faces[*opts]
-	lock.RUnlock()
+	comp.RUnlock()
 	if face == nil {
 		f1 := truetype.NewFace(comp.font, opts)
 		f2 := truetype.NewFace(comp.fallback, opts)
 		face = newFontWithFallback(f1, f2, comp.font, comp.fallback)
 
-		lock.Lock()
+		comp.Lock()
 		comp.faces[*opts] = face
-		lock.Unlock()
+		comp.Unlock()
 	}
 
 	return face
@@ -297,6 +296,7 @@ type ttfFont interface {
 }
 
 type fontCacheItem struct {
+	sync.RWMutex
 	font, fallback *truetype.Font
 	faces          map[truetype.Options]font.Face
 }
--- a/internal/painter/font.go
+++ b/internal/painter/font.go
@@ -61,13 +61,18 @@ func CachedFontFace(style fyne.TextStyle, opts *truetype.Options) font.Face {
 	}
 
 	comp := val.(*fontCacheItem)
+	lock := sync.RWMutex{}
+	lock.RLock()
 	face := comp.faces[*opts]
+	lock.RUnlock()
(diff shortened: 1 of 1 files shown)


Jacalz (changes requested):
Thanks for the PR. However, this does unfortunatelly not seem correct. You are creating a new lock object each time so it is not going to guard anything as far as I can see.

The right approach is either to put the mutex in fontCacheItem or switch to `sync.Map` for the map implementation.

vkuznet:
@Jacalz you're right, I made necessary changes and now lock is part of `fontCacheItem` object.

Jacalz:
Thanks. There is also the issue that the PR is targeting the master branch. All PRs should target the develop branch per the PR template.

vkuznet:
@Jacalz , I already see the changes in develop branch, so I think you or someone else has already applied them. Do you need anything from me? Should we close the issue?

andydotxyz:
If the changes are already on develop then the PR should be closed.
Sorry about the duplication, but we do most of our dev work on that branch.
If we'd been preparing a minor release next we could have brought this in to a release branch, but we are entering the final stages of preparing v2.3.0 so develop is the next branch to release.

PR #3282, opened 2022-09-25
[driver/glfw] fix TestWindow_CaptureTypedShortcut

toaster:
### Description:

This fixes GLFW driver’s `TestWindow_CaptureTypedShortcut` which is a non-CI test.

The key simulation was wrong. Also the key-name used for custom shortcuts uses lower case letters.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass (CI only!).

### DIFF
--- a/internal/driver/glfw/window_test.go
+++ b/internal/driver/glfw/window_test.go
@@ -1494,16 +1494,16 @@ func TestWindow_CaptureTypedShortcut(t *testing.T) {
 	w.Canvas().Focus(content)
 
 	w.keyPressed(nil, glfw.KeyLeftControl, 0, glfw.Press, glfw.ModControl)
-	w.keyPressed(nil, glfw.KeyLeftShift, 0, glfw.Press, glfw.ModControl)
-	w.keyPressed(nil, glfw.KeyF, 0, glfw.Press, glfw.ModControl)
-	w.keyPressed(nil, glfw.KeyLeftShift, 0, glfw.Press, glfw.ModControl)
-	w.keyPressed(nil, glfw.KeyLeftControl, 0, glfw.Release, glfw.ModControl)
-	w.keyPressed(nil, glfw.KeyF, 0, glfw.Release, glfw.ModControl)
+	w.keyPressed(nil, glfw.KeyLeftShift, 0, glfw.Press, glfw.ModControl|glfw.ModShift)
+	w.keyPressed(nil, glfw.KeyF, 0, glfw.Press, glfw.ModControl|glfw.ModShift)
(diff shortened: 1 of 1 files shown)


internal/driver/glfw/window_test.go
```diff
@@ -1494,16 +1494,16 @@ func TestWindow_CaptureTypedShortcut(t *testing.T) {
...
+	w.keyPressed(nil, glfw.KeyLeftShift, 0, glfw.Press, glfw.ModControl|glfw.ModShift)
+	w.keyPressed(nil, glfw.KeyF, 0, glfw.Press, glfw.ModControl|glfw.ModShift)
+	w.keyPressed(nil, glfw.KeyF, 0, glfw.Release, glfw.ModControl|glfw.ModShift)
+	w.keyPressed(nil, glfw.KeyLeftShift, 0, glfw.Release, glfw.ModControl)
+	w.keyPressed(nil, glfw.KeyLeftControl, 0, glfw.Release, 0)
 
 	w.WaitForEvents()
 
 	assert.Equal(t, 1, len(content.capturedShortcuts))
-	assert.Equal(t, "CustomDesktop:Control+F", content.capturedShortcuts[0].ShortcutName())
```
andydotxyz, line 1506: I worry that this is indicative of a break we have made. @Jacalz did the key code changes recently moved the shortcut encoding to use lower case instead of upper?

Jacalz: I don't think it should have broken it, I know I did take care in trying to preserve the behaviour of capital letters. However, I guess there is a possibility that it did. I'll look into it

Jacalz: The only thing I can think of is that https://github.com/fyne-io/fyne/blob/3bccfbad8779941137d3682bb37bd7639c302669/internal/driver/glfw/window_desktop.go#L532-L539 somehow messed with it but I don't quite see how.

andydotxyz: I can see that the old "keyNameMap" made sure that "a" was called KeyNameA, which is "A". The new cast breaks this as far as I can see https://github.com/fyne-io/fyne/commit/1abcf044c4982d99dce9237cb830cdf55fd4a10f#diff-3e0900ca2848440b1f1a2a6b75716de12701abede08a0e8c53f3f0184f02446dL504

Jacalz: You are right. Something is wonky there. It might be best to just revert that PR and start over fresh.

Jacalz: @toaster You might want to revert https://github.com/fyne-io/fyne/commit/1abcf044c4982d99dce9237cb830cdf55fd4a10f. I'll get a better solution together some other time.

Jacalz:
This should now be solved. Will close this :)

PR #3299, opened 2022-09-30
Remove call to Refresh from the rendering logic.

### LINKED ISSUE #2991: Image should not call Refresh from inside the Render thread
Calling Refresh from within the Render thread expose race condition between the Render thread and the user of Fyne API that are out of the user control. It also means that it require multiple iteration in the render thread to have the proper layout when loading an image.

This is an open question here, but the logical things to do would be to have an image open the image, read the image dimension, apply minsize, call Refresh on the canvas, render thread would pick up the open file and finalize the loading of image.

For doing that flow, we might either need to introduce a new field in the image structure or maybe introduce a shadow structure internal to the canvas API.

Until it is tried, I do not have a really good idea of what the solution to this is.

Bluebugs:
### Description:
This is a first attempt at removing the use of Refresh from within the render goroutine and allow first frame with image to be correct with the right min size properly set. It does introduce a few new API in the image object that need discussion along with some refactoring in the internal directory.

One known issue is that some rectangle that should be translucent are opaque somehow, but I do not understand how the image code is affecting it yet.

Fixes #2991 

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


canvas/image.go
```diff
@@ -106,9 +203,7 @@ func NewImageFromFile(file string) *Image {
 // Since: 2.0
 func NewImageFromURI(uri fyne.URI) *Image {
 	if uri.Scheme() == "file" && len(uri.String()) > 7 {
-		return &Image{
-			File: uri.String()[7:],
-		}
+		return NewImageFromFile(uri.String()[7:])
```
Jacalz, line 206: I know that this was wrong before, but you can use `uri.Path()` instead of slicing the string. Also, why does this call the function instead of how it was before given that the function doesn't do anything else?

canvas/image.go
```diff
@@ -136,31 +231,120 @@ func NewImageFromReader(read io.Reader, name string) *Image {
...
+		if err != nil {
+			return fyne.NewSize(0, 0), err
+		}
+		width, height = config.Width, config.Height
+		img.aspect = float32(width) / float32(height)
+	}
+
+	c := fyne.CurrentApp().Driver().CanvasForObject(img)
+	if c == nil {
+		return fyne.NewSize(0, 0), fmt.Errorf("object is not attached to a canvas yet")
```
Jacalz, line 345:
```suggestion
		return fyne.NewSize(0, 0), errors.New("object is not attached to a canvas yet")
```

Jacalz (changes requested):
Nice work. Just had a small look at this and added some comments in-line

canvas/image.go
Bluebugs: Good idea about uri.Path(). As for the why, I think it is better practice to reuse code rather than duplicate even for simple case.

Jacalz: I see. I thought there was an inconsistency because the others were not calling the functions but they are using other parameters. Totally agree with you. Sorry for the fuzz.

andydotxyz:
Thanks for working on this. Looks like the PR template has not been filled out though

andydotxyz (changes requested):
There are 3 new public APIs on Image here, are they necessary for this to work? If not then unexport, if they are then add the missing "Sinece" docs

andydotxyz:
Can you comment on the move of scale and svg helpers to new packages? This has led to a lot of duplication in the usage of the methods (scale.ScaleInt etc)

Bluebugs:
> Can you comment on the move of scale and svg helpers to new packages? This has led to a lot of duplication in the usage of the methods (scale.ScaleInt etc).
> 
> If refactoring like that is to be done it should be treated like white space changes and done in a separate commit so we can see it apart from code changes and the reason would be documented in the history.

Indeed they required refactoring as they needed to be used from the canvas.Image code instead from the internal code. This lead to circular dependencies for scale if it stayed at the root of the package. The change for svg is necessary to match the split between minsize calculation and rendering which weren't differentiated before. I will look into splitting those commit.

andydotxyz:
Perhaps if we find a way that does not require new APIs on the Image type the refactoring is not required either?

andydotxyz:
Would it not be possible to maintain this data outside of the object itself that is controlled by a internal space that avoids the new dependency?

Bluebugs:
It might be possible to have a shadow object that live in internal somewhere and canvas.Image proxy things to it. Not to sure how this would look like. Will see if that can work.

Bluebugs:
The test that is failing is a weird one. It does pass locally if I test all the tests in that directory, but fail if I specifically run just that test. The difference seems to be one pixel in the size of the label (while the image size stay stable). It is possible that it was stable before because there was multiple call to Refresh from the image, but that wouldn't really explain why it pass when it is part of multiple tests being run.

The difference is just in the size of the button 86 x 36 vs 87 x 36. The image is otherwise visually identical.

canvas/image.go
```diff
@@ -79,22 +108,97 @@ func (i *Image) Resize(s fyne.Size) {
 	}
 
 	i.baseObject.Resize(s)
+	i.updated = false
+	i.Refresh()
+}
 
-	Refresh(i)
+// Aspect will return the original content aspect after it was last Refresh
+func (i *Image) Aspect() float32 {
```
Bluebugs, line 116: This is necessary for the painter to be able to properly scale the Image and can not be guessed from it as the SVG resize might lead to a different aspect.

andydotxyz: Why would resizing change the aspect of the image source?

andydotxyz (commented):
This seems like multiple fixes in one PR, if the fix is to remove a race condition it would be best not to change behaviour like Refresh at the same time.

canvas/image.go
```diff
@@ -136,31 +204,150 @@ func NewImageFromReader(read io.Reader, name string) *Image {
...
+	return ""
+}
+
+func (i *Image) updateReader() error {
+	i.reader = nil
+
+	var fd io.Reader
+	if i.Resource != nil {
+		fd = bytes.NewReader(i.Resource.Content())
+		i.isSVG = svg.IsResourceSVG(i.Resource)
```
andydotxyz, line 246: the `isSVG` is not reset in an `else` case

canvas/image.go
```diff
@@ -136,31 +204,150 @@ func NewImageFromReader(read io.Reader, name string) *Image {
...
+		i.isSVG = svg.IsFileSVG(i.File)
+	}
+	i.reader = fd
+	return nil
+}
+
+func (i *Image) calculateMinSize() error {
+	var size fyne.Size
+
+	if i.Resource != nil || i.File != "" {
```
andydotxyz, line 263: shouldn't this guard on `i.reader` instead of this or condition?

canvas/image.go
```diff
@@ -79,22 +93,78 @@ func (i *Image) Resize(s fyne.Size) {
 	}
 
 	i.baseObject.Resize(s)
+	i.Refresh()
+}
 
-	Refresh(i)
+// Aspect will return the original content aspect after it was last Refresh
```
andydotxyz, line 99: "refreshed".

Missing "Since: 2.4"

canvas/image.go
```diff
@@ -79,22 +93,78 @@ func (i *Image) Resize(s fyne.Size) {
...
+func (i *Image) Aspect() float32 {
+	if i.aspect == 0 {
+		i.Refresh()
+	}
+	return i.aspect
+}
+
+func (i *Image) MinSize() fyne.Size {
```
andydotxyz, line 107: missing docs

andydotxyz (changes requested):
Getting really close thanks.

However the 2 new packages with a single file, where the names are stuttering, (svg, scale) are not in-keeping. If they really have to be moved out of `internal` then may I suggest `internal/image` instead?

Bluebugs:
I think the svg module make sense in its own as it is here. `IsFileSVG` and `IsResourceSVG` could be renamed without the SVG in their name as it does repeat when used, but I can't come up with a better name if you have a suggestion.

andydotxyz:
> I think the svg module make sense in its own as it is here.

Can you explain why `svg` and `scale` make sense in their own package? Like I said this is not normally how Go is done (larger packages, only created when there is a clear need), and it is not how we had it before.

Bluebugs:
> > I think the svg module make sense in its own as it is here.
> 
> Can you explain why `svg` and `scale` make sense in their own package? Like I said this is not normally how Go is done (larger packages, only created when there is a clear need), and it is not how we had it before.

The `svg` module was a pre existing module that does all the manipulation to load svg file for Fyne internally. `scale` only contain translation operation to convert from canvas coordinate to screen coordinate. I do not see any relation between the two.

I was unaware of the rule to try to have large package. In this case, I don't see where it would make sense to have scale moved to. `driver` is the only one which seems maybe close as it contain functions that operate on the canvas.

Still, I think this is forcing things and as large package are a preference when there isn't a clear need. In this case, I think having a standalone package for scale make more sense and give an opportunity for a more readable code base.

andydotxyz (approved):

Jacalz (approved):
Just approving to remove my request for changes. Don't really feel that I have the technical understanding to review this code more than what I already have done

PR #3301, opened 2022-10-01
Crash parsing invalid file uri

### LINKED ISSUE #3275: Crash parsing invalid file URI
### Checklist

### Describe the bug

We used the Fuzz engine to modify some Test(TestFromJSON) data, and then the following crash appeared. We hope to get the help of the developer to confirm whether it is a real bug.

### How to reproduce

### Screenshots

crash log:
```
panic: runtime error: slice bounds out of range [5:4] [recovered]
        panic: runtime error: slice bounds out of range [5:4]

goroutine 35 [running]:
testing.tRunner.func1.2({0x62e6c0, 0xc0001d43d8})
testing.tRunner.func1()
panic({0x62e6c0, 0xc0001d43d8})
fyne.io/fyne/v2/storage/repository.ParseURI({0xc0001b9770, 0x4})
        /home/zjx/workspace/gowork/src/purelib/fyne/storage/repository/parse.go:63 +0x737
fyne.io/fyne/v2/storage.ParseURI(...)
        /home/zjx/workspace/gowork/src/purelib/fyne/storage/uri.go:35
fyne.io/fyne/v2/theme.uriString.resource({0xc0001b9770, 0xc00019a128})
        /home/zjx/workspace/gowork/src/purelib/fyne/theme/json.go:82 +0x1f
fyne.io/fyne/v2/theme.(*jsonTheme).Font(0xc0001b0528, {0x0, 0x0, 0x1, 0x0, 0x0})
        /home/zjx/workspace/gowork/src/purelib/fyne/theme/json.go:146 +0xbb
fyne.io/fyne/v2/theme.FuzzTestFromJSON({0xc00036f600, 0x22, 0x200})
        /home/zjx/workspace/gowork/src/purelib/fyne/theme/json_test.go_TestFromJSON_test.go:74 +0x143
```

### Example code

```go
        var1:="0000"
	var2:=`{"Fonts":{"monospace":"file"}}0000{"Fonts":{"monospace":"file"}}0000{"Fonts":{"monospace":"file"}}0000{"Fonts":{"monospace":"file"}}0000{"Fonts":{"monospace":"file"}}0000{"Fonts":{"monospace":"file"}}0000{"Fonts":{"monospace":"file"}}0000{"F`
	repository.Register(var1, intRepo.NewFileRepository())
	th, err := FromJSON(var2)
	assert.Nil(t, err)
	th.Font(fyne.TextStyle{Monospace: true}).Name()
```

### Fyne version

638ae24e4bebfc0b6dbaabf8db5807eeabfafc61

### Go compiler version

1.17.8

### Operating system

Linux

### Operating system version

Linux r920 4.15.0-159-generic #167-Ubuntu

### Additional Information

_No response_

WetDesertRock:
### Description:
storage/repository/parse.go assumed that there would always be a colon terminating the scheme. As seen in #3275 this led to an unexpected panic.

Note that I'm not as familiar with the usage of URI's in fyne, and this does make the code less lenient and throw errors more often (in theory). For example, "path/to/file" now expected to throw an error. After some research I think this is acceptable behavior as I don't see any way for those 'URI's to work. Let me know if this isn't a valid assumption and I can resubmit with a different fix in place (targeting just the reported issue)

Fixes #3275

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/storage/repository/parse.go
+++ b/storage/repository/parse.go
@@ -44,20 +44,12 @@ func NewFileURI(path string) fyne.URI {
 // Since: 2.0
 func ParseURI(s string) (fyne.URI, error) {
 	// Extract the scheme.
-	scheme := ""
-	validScheme := false
-	for i := 0; i < len(s); i++ {
-		if s[i] == ':' {
-			validScheme = true
-			break
-		}
-		scheme += string(s[i])
-	}
-	scheme = strings.ToLower(scheme)
-
-	if !validScheme {
+	colonIndex := strings.Index(s, ":")
+	if colonIndex == -1 {
 		return nil, errors.New("invalid URI, scheme must be present")
 	}
+
+	scheme := strings.ToLower(s[:colonIndex])
 
 	if scheme == "file" {
 		// Does this really deserve to be special? In principle, the
--- a/storage/repository/parse.go
+++ b/storage/repository/parse.go
@@ -44,8 +44,8 @@ func NewFileURI(path string) fyne.URI {
 // Since: 2.0
 func ParseURI(s string) (fyne.URI, error) {
 	// Extract the scheme.
-	colonIndex := strings.Index(s, ":")
-	if colonIndex == -1 {
+	colonIndex := strings.IndexByte(s, ':')
+	if colonIndex <= 0 {
 		return nil, errors.New("invalid URI, scheme must be present")
 	}
 
--- a/storage/repository/parse.go
+++ b/storage/repository/parse.go
@@ -1,6 +1,7 @@
 package repository
 
 import (
+	"errors"
 	"path/filepath"
 	"runtime"
 	"strings"
@@ -44,13 +45,19 @@ func NewFileURI(path string) fyne.URI {
 func ParseURI(s string) (fyne.URI, error) {
 	// Extract the scheme.
 	scheme := ""
+	validScheme := false
 	for i := 0; i < len(s); i++ {
 		if s[i] == ':' {
+			validScheme = true
 			break
 		}
 		scheme += string(s[i])
 	}
 	scheme = strings.ToLower(scheme)
+
+	if !validScheme {
+		return nil, errors.New("invalid URI, scheme must be present")
+	}
 
 	if scheme == "file" {
 		// Does this really deserve to be special? In principle, the
--- a/storage/repository/parse_test.go
+++ b/storage/repository/parse_test.go
@@ -24,3 +24,17 @@ func TestParseURI(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, "file://C:/tmp/foo.txt", uri.String())
 }
+
+func TestParseInvalidURI(t *testing.T) {
+	uri, err := ParseURI("/tmp/foo.txt")
+	assert.NotNil(t, err)
+	assert.Nil(t, uri)
+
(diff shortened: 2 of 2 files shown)


storage/repository/parse.go
```diff
@@ -44,14 +45,20 @@ func NewFileURI(path string) fyne.URI {
...
 			break
 		}
 		scheme += string(s[i])
 	}
 	scheme = strings.ToLower(scheme)
 
+	if !validScheme {
+		return nil, errors.New("invalid URI, scheme must be present")
+	}
+
```
Jacalz, line 61: The previous code here actually looks rather inefficient and not that good. I am thinking that it would be better to just replace the whole loop with something like this:
```go
colonIndex := strings.Index(s, "://")
if colonIndex == -1 {
     return nil, errors.New("invalid URI, scheme must be present")
}

scheme := s[:colonIndex]
```
This should be cleaner and a lot better than performing string concatenation all the time.

Jacalz (changes requested):
Thanks for the PR. While the fix seems to work, I think the previous solution to getting the scheme was slow and unnecessary complex. It would be better if we just removed it in favour of a better solution. I have proposed one in the comments.

storage/repository/parse.go
WetDesertRock: Good call. It appears that this code also needs to support URNs (storage/uri_test.go:26) so I've implemented your suggestion however just limiting it to the colon.

storage/repository/parse.go
```diff
@@ -43,14 +44,12 @@ func NewFileURI(path string) fyne.URI {
 // Since: 2.0
 func ParseURI(s string) (fyne.URI, error) {
 	// Extract the scheme.
-	scheme := ""
-	for i := 0; i < len(s); i++ {
-		if s[i] == ':' {
-			break
-		}
-		scheme += string(s[i])
+	colonIndex := strings.Index(s, ":")
```
Jacalz, line 47: This is the incorrect API to use for single characters. It makes more sense to use `strings.IndexByte` in the case.

storage/repository/parse.go
```diff
@@ -43,14 +44,12 @@ func NewFileURI(path string) fyne.URI {
...
+	if colonIndex == -1 {
```
Jacalz, line 48: Should we check `== 0` here as well? Your test indicates that this already produces an error but does so further down in the chain. Thus, the question is if we should check it here when we already have the index?

This is merely a question and not a code suggestion.

Jacalz (changes requested):
Thanks. Good call about the URI not necessarily having `://`.

storage/repository/parse.go
WetDesertRock: This would indicate that the scheme is blank. I can't imagine a good use case for this, but I wouldn't mind deferring to someone who has more experience with how these URI's are used. I don't think the RFC allows blank schemes at any rate. I'll throw that change in.

andydotxyz: It's not allowed but given the number of bugs open for crashes with invalid URIs we should probably be more defensive.

Jacalz (approved):
This looks good to me now. Thanks your for all the work that has gone into it and congratulations on your first contribution to Fyne 🙂

PR #3304, opened 2022-10-03
Refactor Android manifest into internal/templates

Jacalz:
### Description:

This is just a simple cleanup that moves the template to the common templates package and also moves the data struct to where it is being used. The main benefit here is basically only that it is easier to find the manifest without having to search as we have all templates in the common place.

I don't have any Android devices to test this on, so I'd prefer that someone can verify that this regresses nothing. Normally I would think it is fine but the template seems to accidentally have used `html/template` before and I don't want to risk anything.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/mobile/build_androidapp.go
+++ b/cmd/fyne/internal/mobile/build_androidapp.go
@@ -20,10 +20,20 @@ import (
 	"strings"
 
 	"fyne.io/fyne/v2/cmd/fyne/internal/mobile/binres"
+	"fyne.io/fyne/v2/cmd/fyne/internal/templates"
 	"fyne.io/fyne/v2/cmd/fyne/internal/util"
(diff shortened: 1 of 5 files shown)


andydotxyz:
Just FYI the android emulator from Android Studio etc runs the same APK as a real device so you can verify locally in a way you can't with iOS simulator.

Jacalz:
> Just FYI the android emulator from Android Studio etc runs the same APK as a real device so you can verify locally in a way you can't with iOS simulator.

Trust me, I have tried that but never had any luck with it. It is always terribly slow and seems to have trouble launching the applications for me.

andydotxyz (approved):
Working great, sorry it took so long to test

Jacalz:
No worries. It is kind of my fault for not having any working Android setup

PR #3305, opened 2022-10-03
Move shaderSource lookup to switch statement

Jacalz:
### Description:

This replaces the map with a small function using a switch case. The cost is 33 units in this case and the compiler can thus inline it for even more improved performance.

This seems like a case where tests are hard to do. We can't quite hard-code each check. Will have to do without for now. Should get coverage if we add tests to the function that uses it.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/painter/gl/painter.go
+++ b/internal/painter/gl/painter.go
@@ -10,8 +10,8 @@ import (
 	"fyne.io/fyne/v2/theme"
 )
 
-func shaderSourceNamed(fileName string) ([]byte, []byte) {
-	switch fileName {
+func shaderSourceNamed(name string) ([]byte, []byte) {
+	switch name {
 	case "line":
 		return shaderLineVert.StaticContent, shaderLineFrag.StaticContent
 	case "line_es":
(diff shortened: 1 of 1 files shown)


internal/painter/gl/painter.go
```diff
@@ -128,7 +136,7 @@ func (p *painter) createProgram(shaderFilename string) Program {
 	// Why a switch over a filename?
 	// Because this allows for a minimal change, once we reach Go 1.16 and use go:embed instead of
 	// fyne bundle.
-	sources := shaderSources[shaderFilename]
+	sources := shaderSourceNamed(shaderFilename)
 	vertexSrc, fragmentSrc := sources[0], sources[1]
```
Bluebugs, line 140: If the only use case is to split this in vertex and fragment shader, why not return two []byte{} instead of one [2][]byte{}.

Jacalz: Good call. I'll get that sorted tomorrow.

Bluebugs (approved):

PR #3312, opened 2022-10-05
fix: crash in storage.List

### LINKED ISSUE #3287: crash in storage.List()
### Checklist

### Describe the bug

We used the Fuzz engine to modify some Test(TestInMemoryRepositoryListing) data, and then the following crash appeared. Although Test does not handle the error returned by (storage.ParseURI, storage.CanList(foo)), such a crash occurs when corrupt data is passed into the storage.List()

### How to reproduce

### Screenshots

CrashLog
```
panic: runtime error: index out of range [-1] [recovered]
        panic: runtime error: index out of range [-1]

goroutine 19 [running]:
testing.tRunner.func1.2({0x6f8480, 0xc0001643d8})
testing.tRunner.func1()
panic({0x6f8480, 0xc0001643d8})
fyne.io/fyne/v2/internal/repository.(*InMemoryRepository).List(0xc000140180, {0x789360, 0xc00012c360})
        /home/zjx/workspace/gowork/src/purelib/crashConfirm/fyne/internal/repository/memory.go:281 +0x330
fyne.io/fyne/v2/storage.List({0x789360, 0xc00012c360})
        /home/zjx/workspace/gowork/src/purelib/crashConfirm/fyne/storage/uri.go:481 +0x58
fyne.io/fyne/v2/internal/repository.TestInMemoryRepositoryListing(0x0)
        /home/zjx/workspace/gowork/src/purelib/crashConfirm/fyne/internal/repository/memory_test.go:334 +0x309
```

### Example code

```go
//part of the Test Code
	m := NewInMemoryRepository("000")
	repository.Register("dht", m)
	m.Data["/foo"] = []byte{1, 2, 3}
	m.Data["/foo/bar"] = []byte{1, 2, 3}
	m.Data["/foo/baz/"] = []byte{1, 2, 3}
	m.Data["/foo/baz/quux"] = []byte{1, 2, 3}

	foo, _ := storage.ParseURI("dht:?00000")

	canList, err := storage.CanList(foo)
	listing, err := storage.List(foo)
```

### Fyne version

commit 638ae24e4bebfc0b6dbaabf8db5807eeabfafc61 (HEAD -> master, origin/master, origin/HEAD)

### Go compiler version

1.17.8

### Operating system

Linux

### Operating system version

OS: Linux r920 4.15.0-159-generic #167-Ubuntu

### Additional Information

_No response_

b5710546232:
### Description:

Fixes #3287
I added checking for length of prefix that caused application crash.

### DIFF
--- a/internal/repository/memory_test.go
+++ b/internal/repository/memory_test.go
@@ -33,6 +33,26 @@ func TestInMemoryRepositoryRegistration(t *testing.T) {
 	assert.Equal(t, m2, repo)
 }
 
+func TestInMemoryRepositoryParsingWithEmptyList(t *testing.T) {
+	m := NewInMemoryRepository("000")
+	repository.Register("dht", m)
+	m.Data["/foo"] = []byte{1, 2, 3}
+	m.Data["/foo/bar"] = []byte{1, 2, 3}
+	m.Data["/foo/baz/"] = []byte{1, 2, 3}
+	m.Data["/foo/baz/quux"] = []byte{1, 2, 3}
+
+	foo, err := storage.ParseURI("dht:?00000")
+	assert.Nil(t, err)
+
+	canList, err := storage.CanList(foo)
+	assert.NotNil(t, err)
+	assert.Equal(t, canList, false)
+
+	listing, err := storage.List(foo)
+	assert.Nil(t, err)
+	assert.Equal(t, len(listing), 0)
+}
+
 func TestInMemoryRepositoryParsing(t *testing.T) {
 	// set up our repository - it's OK if we already registered it
 	m := NewInMemoryRepository("mem")
--- a/internal/repository/memory.go
+++ b/internal/repository/memory.go
@@ -278,7 +278,8 @@ func (m *InMemoryRepository) List(u fyne.URI) ([]fyne.URI, error) {
 	// solves the edge case where you have say '/foo/bar' and
 	// '/foo/barbaz'.
 	prefix := u.Path()
-	if prefix[len(prefix)-1] != '/' {
+
+	if len(prefix) > 0 && prefix[len(prefix)-1] != '/' {
 		prefix = prefix + "/"
 	}
 


internal/repository/memory_test.go
```diff
@@ -33,6 +33,26 @@ func TestInMemoryRepositoryRegistration(t *testing.T) {
 	assert.Equal(t, m2, repo)
 }
 
+func TestInMemoryRepositoryParsingWithEmptyList(t *testing.T) {
+	m := NewInMemoryRepository("000")
+	repository.Register("dht", m)
+	m.Data["/foo"] = []byte{1, 2, 3}
+	m.Data["/foo/bar"] = []byte{1, 2, 3}
```
andydotxyz, line 40: This data does not seem to be needed

internal/repository/memory_test.go
```diff
@@ -33,6 +33,26 @@ func TestInMemoryRepositoryRegistration(t *testing.T) {
...
+	m.Data["/foo/baz/"] = []byte{1, 2, 3}
+	m.Data["/foo/baz/quux"] = []byte{1, 2, 3}
+
+	foo, err := storage.ParseURI("dht:?00000")
+	assert.Nil(t, err)
+
+	canList, err := storage.CanList(foo)
+	assert.NotNil(t, err)
```
andydotxyz, line 48: Is this right? Asking if it can list returning an error...
Then the next call to list on something that cannot is checking for no error. Seems backwards somehow.

andydotxyz (changes requested):
Asked a few questions but mostly the data setup seems like it isn't needed

b5710546232:
You means in the test case?

internal/repository/memory_test.go
b5710546232: I agree

internal/repository/memory_test.go
b5710546232: Or i do not need to call this one?
Because root case is  in `List` of InMemoryRepository
prefix from path is return as empty string then when it called`prefix[len(prefix)-1]` It will be out of bound.

b5710546232: Or can i delete the test case ?

andydotxyz:
The comment was inline showing that 4 lines of the test case were unneeded, looked like a copy paste issue.
Please don't remove the test case, we should have coverage of all fixes

b5710546232:
Okay i will bring it back.

internal/repository/memory_test.go
andydotxyz: I think I understand your case here. The data is just rather hard to understand as you have a custom scheme and a custom structure of data - perhaps something more traditional that performs the same test?

PR #3315, opened 2022-10-07
Change link of repojacking vulnerable link

yakirk:
Hello from Hacktoberfest :)
The link to purell is vulnerable to repojacking (due to a typo in the org name, it redirects to an unexisting project), you should change the link to the current name of the project. if you won't change the link, an attacker can open the linked repository and attacks users that trust your links

### DIFF
--- a/vendor/github.com/fredbi/uri/README.md
+++ b/vendor/github.com/fredbi/uri/README.md
@@ -59,7 +59,7 @@ which provides a workable but loose implementation of the RFC.
 
 This package concentrates on RFC 3986 strictness for URI validation. 
 At the moment, there is no attempt to normalize or auto-escape strings. 
-For url normalization, see github.com/PuertokitoBio/purell.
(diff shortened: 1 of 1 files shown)


Jacalz (changes requested):
Thanks for the PR, I appreciate it. However, I have a few notes about why this is not the correct fix.

1. The claim that we are vulnerable to repo jacking but the link is not in our README but in a vendor we project. The likelihood that this affects anyone is pretty small.
2. You are updating a vendor see project in the vendor folder. The change should be submitted upstream to https://github.com/fredbi/uri instead.
3. Our PR template specifies that all PRs should target the develop branch and not the master branch.

The right way forward is to open the PR upstream instead and once it has been merged there, you can open a PR against `develop` here. Thanks for understanding.

Jacalz:
Closing per above comment. Feel free to open a new PR when the notes above have been solved.

PR #3317, opened 2022-10-09
Allow nil Action for systray menu items

WetDesertRock:
### Description:
Throughout fyne the MenuItem Action can be set to nil if it shouldn't execute an action. This behavior isn't supported when setting the menu using SetSystemTrayMenu. This PR fixes that.

Note that no tests are included as it doesn't appear that the systray stuff is covered by tests (and adding tests is a bit out of scope of my intended fix). If I just missed where the tests are located, let me know and I can add this into those.

For example (borrowed and adapted from fyne_demo):
```go
func makeTray(a fyne.App) {
		h := fyne.NewMenuItem("Hello", nil)
		h.Icon = theme.HomeIcon()
		menu := fyne.NewMenu("Hello World", h)
		desk.SetSystemTrayMenu(menu)
	}
}
```

### Checklist:
- [o] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/driver_desktop.go
+++ b/internal/driver/glfw/driver_desktop.go
@@ -110,7 +110,9 @@ func (d *gLDriver) refreshSystray(m *fyne.Menu) {
 		fn := i.Action
 		go func() {
 			for range item.ClickedCh {
-				fn()
+				if fn != nil {
+					fn()
+				}
 			}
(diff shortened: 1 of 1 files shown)


andydotxyz (approved):
Thanks so much for this contribution :)

PR #3318, opened 2022-10-09
expose storage.ErrNotExists for non existing documents

### LINKED ISSUE #3083: Expose a storage.ErrNotExists for non existing documents

My code cannot check an error returned by, such as app.Storage().Save() with errors.Is()

Such as `return os.ErrNotExist` for `document with name * does not exist`

Please use standard errors from Go packages instead of using errors.New()

rangzen:
### Description:

Fixes #3083

Let me know all the changes you need because I am pretty sure this one will not be right in the first place.

### Checklist:

- [x] Tests included.
- [ ] Tests all pass. (I don't pass all the tests because of the Android things.)

### DIFF
--- a/storage/errors.go
+++ b/storage/errors.go
@@ -7,4 +7,9 @@ var (
 	//
 	// Since: 2.3
 	ErrNotExists = errors.New("document does not exist")
+
+	// ErrAlreadyExists may be thrown by docs. E.g., save a document twice.
+	//
+	// Since: 2.3
+	ErrAlreadyExists = errors.New("document already exists")
 )
--- a/storage/errors.go
+++ b/storage/errors.go
@@ -1,0 +1,10 @@
+package storage
+
+import "errors"
+
+var (
+	// ErrNotExists may be thrown by docs. E.g., save an unknown document.
+	//
+	// Since: 2.2.2
+	ErrNotExists = errors.New("document does not exist")
+)
--- a/storage/errors.go
+++ b/storage/errors.go
@@ -3,13 +3,13 @@ package storage
 import "errors"
 
 var (
-	// ErrNotExists may be thrown by docs. E.g., save an unknown document.
-	//
-	// Since: 2.3
-	ErrNotExists = errors.New("document does not exist")
-
 	// ErrAlreadyExists may be thrown by docs. E.g., save a document twice.
 	//
 	// Since: 2.3
 	ErrAlreadyExists = errors.New("document already exists")
+
+	// ErrNotExists may be thrown by docs. E.g., save an unknown document.
+	//
+	// Since: 2.3
+	ErrNotExists = errors.New("document does not exist")
 )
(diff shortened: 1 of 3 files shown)


storage/errors.go
```diff
@@ -0,0 +1,10 @@
+package storage
+
+import "errors"
+
+var (
+	// ErrNotExists may be thrown by docs. E.g., save an unknown document.
+	//
+	// Since: 2.2.2
```
andydotxyz, line 8: This should be Since 2.3 - we don't add new API in patch releases (plus v2.3.0 is the next release). As you can see the api docs are grouped with just 2 digits to match this design.

andydotxyz (changes requested):
This looks like a great approach. I wonder if there are any other places in Storage or Docs that should be using the new error?

Jacalz (dismissed):
Looks good to me. Thanks for the PR

storage/errors.go
```diff
@@ -0,0 +1,15 @@
...
+var (
+	// ErrNotExists may be thrown by docs. E.g., save an unknown document.
+	//
+	// Since: 2.3
+	ErrNotExists = errors.New("document does not exist")
+
+	// ErrAlreadyExists may be thrown by docs. E.g., save a document twice.
+	//
+	// Since: 2.3
+	ErrAlreadyExists = errors.New("document already exists")
```
andydotxyz, line 14: Can you please order these alphabetically? It's part of our code style.

andydotxyz (changes requested):
Looks great, just a const ordering request

storage/errors.go
rangzen: Done. Can you point me to a document, I don't find anything about that. If this document doesn't exist, I will try to compile these rules in the Contributing in another MR.

rangzen: https://github.com/fyne-io/fyne/pull/3325

storage/errors.go
```diff
@@ -0,0 +1,15 @@
...
+
+	// ErrNotExists may be thrown by docs. E.g., save an unknown document.
+	//
+	// Since: 2.3
+	ErrNotExists = errors.New("document does not exist")
+)
```
Jacalz, line 15: I realised that there already is an error in one of the other files. It would be better to have all of them in the same place

rangzen: If you are talking about the one in repository, I saw it but it is another package so I though that these errors make more sense with the storage. prefix than the repository. prefix.

Jacalz: Ah, good point. Makes sense

andydotxyz (approved):

Jacalz (approved):
Looks good. Nice work!

Jacalz:
Congratulations on your first contribution to Fyne and thanks for your work :)

rangzen:
Thank you @Jacalz ! It is an interesting project. Happy to have discovered it! I will probably try some experiment on the mobile part.

PR #3323, opened 2022-10-09
add CheckGroup.Remove

### LINKED ISSUE #3124: Func for deleting item in checkboxes
Hello, I'm want to make an app for my home usage. I've added few select boxes and amout of items in checkboxes depends on selected choose. 
I find on API that you can add, but can't delete
Is there any possibility you can add this staff? Or maybe you can help me come up with different solution?
image
image

rangzen:
### Description:

Fixes #3124

### Checklist:

- [x] Tests included.
- [ ] Tests all pass. (I don't pass all the tests because of the Android things.)

#### Where applicable:

### DIFF
--- a/widget/check_group.go
+++ b/widget/check_group.go
@@ -50,18 +50,13 @@ func (r *CheckGroup) Append(option string) {
 //
 // Since: 2.3
 func (r *CheckGroup) Remove(s string) bool {
-	found := -1
 	for i, o := range r.Options {
 		if strings.EqualFold(s, o) {
-			found = i
-			continue
+			r.Options = append(r.Options[:i], r.Options[i+1:]...)
+			return true
 		}
 	}
-	if found == -1 {
-		return false
-	}
-	r.Options = append(r.Options[:found], r.Options[found+1:]...)
-	return true
+	return false
 }
 
 // CreateRenderer is a private method to Fyne which links this widget to its renderer
--- a/widget/check_group.go
+++ b/widget/check_group.go
@@ -47,6 +47,7 @@ func (r *CheckGroup) Append(option string) {
 }
 
 // Remove removes the first occurrence of the specified option found from a CheckGroup widget.
+// Return true if an option was removed.
 //
 // Since: 2.3
 func (r *CheckGroup) Remove(option string) bool {
--- a/widget/check_group.go
+++ b/widget/check_group.go
@@ -4,6 +4,7 @@ import (
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/canvas"
 	"fyne.io/fyne/v2/internal/widget"
+	"strings"
 )
 
 // CheckGroup widget has a list of text labels and checkbox icons next to each.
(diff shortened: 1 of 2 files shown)


widget/check_group.go
```diff
@@ -44,6 +45,24 @@ func (r *CheckGroup) Append(option string) {
...
 
+// Remove removes the first specified option found from a CheckGroup widget.
+//
+// Since: 2.3
+func (r *CheckGroup) Remove(s string) bool {
+	found := -1
+	for i, o := range r.Options {
+		if strings.EqualFold(s, o) {
+			found = i
+			continue
```
Jacalz, line 56: This seems like it should be a `break`. However, please see my other comment about this not being necessary.

widget/check_group.go
```diff
@@ -44,6 +45,24 @@ func (r *CheckGroup) Append(option string) {
...
+		}
+	}
+	if found == -1 {
+		return false
+	}
+	r.Options = append(r.Options[:found], r.Options[found+1:]...)
+	return true
+}
```
Jacalz, line 64: Would it not be better to do something like this?
```go
	for i, o := range r.Options {
		if strings.EqualFold(s, o) {
			r.Options = append(r.Options[:i], r.Options[i+1:]...)
			return true
		}
	}
	return false
}
```

Jacalz (changes requested):
Thanks for the PR. I have a suggestion on how the code can be cleaned up a bit.

widget/check_group.go
rangzen: Oh my... Of course, it is a break...

widget/check_group.go
rangzen: Possible, but even in my version I am afraid of the moving states with no lock between append and remove. It is even worse to change states in a loop but that is your call :) I'll made the changes.

widget/check_group.go
```diff
@@ -44,6 +46,19 @@ func (r *CheckGroup) Append(option string) {
 	r.Refresh()
 }
 
+// Remove removes the first specified option found from a CheckGroup widget.
+//
+// Since: 2.3
+func (r *CheckGroup) Remove(s string) bool {
```
Jacalz, line 52: I would prefer to call the parameter `option` yo be consistent with `.Append`.

widget/check_group.go
```diff
@@ -44,6 +46,19 @@ func (r *CheckGroup) Append(option string) {
 	r.Refresh()
 }
 
+// Remove removes the first specified option found from a CheckGroup widget.
```
Jacalz, line 49: I think this wording is a bit unclear. It sounds almost as if it just removes the first item in the slice.

Jacalz (changes requested):
Thanks. Just a few small notes on naming and wording.

widget/check_group.go
rangzen: Done. Good catch.

widget/check_group.go
rangzen: Is this one better? [76ab4fb](https://github.com/fyne-io/fyne/pull/3323/commits/76ab4fb557ace3f4d21502c0246baa296217a8fd)

rangzen: We can also add remove on index, but it means adding an error. And since Options is public I am not sure of the value added.

andydotxyz (changes requested):
Thanks for this, a couple of thoughts:
* you need to refresh after a change like this
* what happens if the item is currently selected?
* the documentation does not mention the meaning of the return book
* we try to keep methods in alphabetical order so this new one is in the wrong place.

rangzen:
> Thanks for this, a couple of thoughts:
> 
>     * you need to refresh after a change like this
> 
>     * what happens if the item is currently selected?
> 
>     * the documentation does not mention the meaning of the return bool
> 
>     * we try to keep methods in alphabetical order so this new one is in the wrong place.

All done. I will test the selected case later.

rangzen:
>     * what happens if the item is currently selected?

Test the case, no error in the GUI, but internally, CheckGroup.Selected state is wrong. Fixed in the last commit.

andydotxyz (approved):
Great thanks. Sorry that I missed the previous fix was committed already

widget/check_group.go
andydotxyz: Let us know if this is fixed @Jacalz so it can land

Jacalz (approved):

PR #3325, opened 2022-10-10
add Coding style

rangzen:
### Description:

I got some points to fix in the review, but it appears that it is not written somewhere, so it's not always obvious.


andydotxyz (changes requested):
Probably better to link to the full guideline https://github.com/fyne-io/fyne/wiki/Contributing#code-style

rangzen:
Oh nice, I didn't find that one, I will fix it with the link :+1:

andydotxyz (approved):
Great thanks

PR #3330, opened 2022-10-12
With Chrome 106 it is now possible to run fyne wasm output.

Bluebugs:
### Description:
Chrome 106 did solve its bug when running fyne application targeting webassembly. So let's switch to serving wasm output to chrome 106 and older.

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


Jacalz:
Can you add a TODO, or open an issue, to remind us to remove this once Chrome 106 is out minimum version?

Bluebugs:
> Can you add a TODO, or open an issue, to remind us to remove this once Chrome 106 is out minimum version?

I actually do not know how we would decide that. If you have suggestion, I can add that to an issue.

andydotxyz:
Looks fine. Would be good if you can fill the PR template though.

Bluebugs:
> Would be good if you can fill the PR template though.

I really never know what to do when this is outside of the go code base. There is no test I can think of for it, and none of the other question seems relevant. What is the expected action on those checkbox if they are not relevant to this case?

andydotxyz:
If the items under "where applicable" are not applicable we remove them instead of leaving them unchecked (otherwise GitHub reports them as incomplete tasks).
I agree there are no tests that make sense at this time

andydotxyz (approved):

PR #3331, opened 2022-10-13
Fix #3267 android entry on submitted is not working

### LINKED ISSUE #3267: Android - Entry - OnSubmitted is not working
### Checklist

### Describe the bug

OnSubbmited event is not firing when pressing `Enter` key on Android keyboard

Keyboard image:
image

### How to reproduce

1. Run code on Android
2. Press Enter
3. Nothing happens (no logs in logcat for example)

### Screenshots

_No response_

### Example code

```go
func main() {
	a := app.NewWithID("example")

	w := a.NewWindow("example")
	cntr := widget.NewEntry()
	cntr.OnSubmitted = func(s string) {
		fmt.Println("SUBMITTED: ", s)
	}

	w.SetContent(cntr)

	w.Resize(fyne.NewSize(640, 480))
	w.ShowAndRun()
}
```

### Fyne version

2.2.3 and also develop

### Go compiler version

go1.18 linux/amd64

### Operating system

Android

### Operating system version

Lineage OS

### Additional Information

_No response_

apiotrowski312:
### Description:

With `IME_ACTION_DONE` we were not sending `\n` to `keyboardTyped()` function. If I get everything correctly, this is the only way to fire `OnSubmitted` function.

This fix works only with single-line entry.

I tested changes on _Microsoft Swift_ keyboard, on a phone with _Lineage OS_. I do not own any other phone so I was not able to test it on other devices.

Fixes #3267

### Checklist:

I did not touch Go code.

- [ ] Tests included.
- [ ] Tests all pass.


apiotrowski312:
As I stated, this fix works only with single line entry, multi line does not support submit at all. In a code fyne assume `OnSubmitted` should be fired when we press `Shift + Enter`. Should it work the same on mobile?

andydotxyz:
> As I stated, this fix works only with single line entry, multi line does not support submit at all. Fyne assume `OnSubmitted` should be called when we press `Shift + Enter`. Should it work the same on mobile?

Good point, yes the mobile keyboard behaviour should be consistent I would think

apiotrowski312:
@andydotxyz I wanted to do it in this PR. But I am afraid I will not be able too. 
I spent some time trying to figure out how to detect shift key being pressed and I faced a bug where my shift key is not sending any events at all :shrug:

andydotxyz:
> @andydotxyz I wanted to do it in this PR. But I am afraid I will not be able too. I spent some time trying to figure out how to detect shift key being pressed and I faced a bug where my shift key is not sending any events at all 🤷

That's OK. If you can open another issue and reference it then we can accept this fix I think.

apiotrowski312:
@andydotxyz https://github.com/fyne-io/fyne/issues/3339

andydotxyz (approved):
Thanks, I appreciate this and we can come back to the multi line case as you noted.

PR #3332, opened 2022-10-14
Add ability to specify custom metadata.

Bluebugs:
### Description:
Custom metadata can be defined and picked from the Development or Release section of FyneApp.toml or specified on the command. In the later case it will override any custom metadata provided in the FyneApp.toml. This is useful if you want to provide for example a remote self update url that depends on if the build is a release or a development release. Or if you want to provide some kind of API key during the build and do not want to store it in your repository.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


app.go
```diff
@@ -79,6 +79,14 @@ type App interface {
...
 	// Since: 2.3
 	SetCloudProvider(CloudProvider) // configure cloud for this app
+
+	// CustomMetadata allow you to get metadata specified at compile time
+	// It will contain the Release section if the binary was compiled in Release mode, the Development section otherwise
+	// The value can also be injected on the command line during compilation if you do not want to store them in your
+	// source repository
+	//
+	// Since: 2.3
+	CustomMetadata(name string) (custom string, exist bool)
```
andydotxyz, line 89: Would this not be better attached to the `AppMetadata` struct?

app/meta.go
```diff
@@ -20,3 +24,21 @@ func SetMetadata(m fyne.AppMetadata) {
...
+func (a *fyneApp) CustomMetadata(name string) (custom string, exist bool) {
+	custom, exist = customMetadata[name]
+	return
+}
+
+// SetCustomMetadata allow fyne to set the custom metadata map once at the start of your application
+// You shouldn't use it if you do not know what you are doing
+//
+// Since: 2.3
+func SetCustomMetadata(custom map[string]string) bool {
```
andydotxyz, line 37: Do we need a setter here? the other custom data isn't set in this way...
I'm not sure why this blocks a second set, and the need for customMetadataSet.

andydotxyz (changes requested):
Thanks for the addition, but I think it can be made more simple by adding it to `AppMetadata` instead of the `App` interface.

app.go
Bluebugs: AppMetadata is a structure not an interface. I think that hiding the data structure with an interface allow for more potential use case in the future with less chance of breaking API.

app/meta.go
Bluebugs: I guess that it comes down to how we want to expose this data. If we are sure that a map is a good solution in all use case and exposing directly the data type is ok, then that would make sense to not have a custom metadata setter. I am not very confident in exposing the map directly and feel like an interface give more freedom.

andydotxyz: I think there are a few steps between a global setter in the `app` package and directly exposing the `map` in `AppMetadata` are there not?

app.go
Bluebugs: I am not to sure how this would look like. AppMetadata is implemented in the fyne module while the metadata is currently provided by the app module. Attaching along the AppMetadata means that the implementation will be in fyne for custom metadata. It also means that the setter for it should be in fyne module or that we make the map public.

app/meta.go
Bluebugs: The setter does not enforce the form in which it is stored in the metadata struct. It also doesn't define how it is accessed. Only the way to set it.

andydotxyz: If it must be set as a `map[string]string` that does rather constrain things, so keeping the flexibility seems to have limited benefit.
Moreover if the flexible approach requires accessing new metadata in a different place in our API I don't see that as being desirable. We should not require a new metadata access point to support custom fields.

app.go
andydotxyz: Given that this is where the app metadata is stored it should at least be explored.
I don't see a strong reason not to follow the pattern that all other metadata has taken so far.

cmd/fyne/internal/commands/build.go
```diff
@@ -331,6 +332,12 @@ func injectMetadataIfPossible(runner runner, srcdir string, app *appData,
 		return nil, nil
 	}
 
+	fyneGoModVersionConstraintOlderThan203 := version.NewConstrainGroupFromString(">=2.3")
+	if fyneGoModVersionConstraintOlderThan203.Match(fyneGoModVersionNormalized) {
+		app.VersionOlderThan203 = true
+	}
+	log.Println("fyne version:", fyneGoModVersionNormalized)
```
andydotxyz: I don't think this log is needed

cmd/fyne/internal/commands/build.go
```diff
@@ -331,6 +332,12 @@ func injectMetadataIfPossible(runner runner, srcdir string, app *appData,
 		return nil, nil
 	}
 
+	fyneGoModVersionConstraintOlderThan203 := version.NewConstrainGroupFromString(">=2.3")
+	if fyneGoModVersionConstraintOlderThan203.Match(fyneGoModVersionNormalized) {
+		app.VersionOlderThan203 = true
```
andydotxyz, line 336: I guess you mean `VersionAtLeast23` not `VersionOlderThan203`?

cmd/fyne/internal/commands/build.go
```diff
@@ -65,6 +67,11 @@ func Build() *cli.Command {
 				Usage:       "Specify a name for the output file, default is based on the current directory.",
 				Destination: &b.target,
 			},
+			&cli.GenericFlag{
```
andydotxyz, line 70: Am I right that you can have only 1 custom field? I'm not familiar with `GenericFlag` but it seems like setting a second will overwrite the first...

andydotxyz (changes requested):
A few fixes suggested. Also I wonder how we would have multiple metadata parameters?

cmd/fyne/internal/commands/build.go
Bluebugs: Oh, I was thinking I did remove it, but seems not. Good catch!

cmd/fyne/internal/commands/build.go
Bluebugs: Actually, if you look into flags.go, keyValueFlag is actually storing multiple key/value pairs in a map and the cli parser will call Set as often as needed. So it will store as many custom metadata as you want :-)

cmd/fyne/internal/commands/build.go
andydotxyz: Given that it is just an identifier and not a int why not `VersionAtLeast2_3` which would allow 2_11 if required

Bluebugs: Indeed, switched to VersionAtLeast2_3.

andydotxyz (approved):
Nice thanks

alfonmga:
I find confusing the design decision about how to specify custom metadata.
I miss some way to add metadata for both modes (`release` and `development`) in the `FyneApp.toml` file.

Sorry for the bump, but I have been trying to understand how the metadata system works for a while and wondering why when I ran `fyne package` it was removing the `[Custom]` section I specified in the `FyneApp.toml` file...

Cheers.

alfonmga:
> I find confusing the design decision about how to specify custom metadata. I miss some way to add metadata for both modes (`release` and `development`) in the `FyneApp.toml` file.
> 
> Sorry for the bump, but I have been trying to understand how the metadata system works for a while and wondering why when I ran `fyne package` it was removing the `[Custom]` section I specified in the `FyneApp.toml` file...
> 
> Cheers.

To make a constructive criticism, I propose this improvement:

```go
fyneApp.Metadata().Build // fyne package --metadata
fyneApp.Metadata().Custom // Common metadata specified in the  `FyneApp.toml` `[Custom]` section.
fyneApp.Metadata().Release // Release metadata specified in the  `FyneApp.toml` `[Release]` section.
fyneApp.Metadata().Development // Development metadata specified in the  `FyneApp.toml` `[Development]` section.
```
It may involve breaking changes.

Bluebugs:
> I find confusing the design decision about how to specify custom metadata.
> I miss some way to add metadata for both modes (`release` and `development`) in the `FyneApp.toml` file.
> 
> Sorry for the bump, but I have been trying to understand how the metadata system works for a while and wondering why when I ran `fyne package` it was removing the `[Custom]` section I specified in the `FyneApp.toml` file...

There is two section supported with this pr, development and release which are included in the binary depending on how you build it. The reasoning for only including development in a development build and release information in a release is to reduce the potential for leak of api keys and things like that to a production binary. This can be overridden on the command line when building if necessary.

I guess in some case you end up with Metadata duplicated between both release and development. It would make sense I guess to add another section for the value common to both mode. Would that cover your problem?

Jacalz:
I just want to point out that it is better to open a discussion when it comes to questions. This is a PR that was merged four months ago and is not the right place for implementation discussions.

alfonmga:
> There is two section supported with this pr, development and release which are included in the binary depending on how you build it. The reasoning for only including development in a development build and release information in a release is to reduce the potential for leak of api keys and things like that to a production binary. This can be overridden on the command line when building if necessary.

It's well thought out. Thanks for the PR.

> I guess in some case you end up with Metadata duplicated between both release and development. It would make sense I guess to add another section for the value common to both mode. Would that cover your problem?

Yes, it would cover my problem. 
I would love to get rid of duplicated code.

PR #3346, opened 2022-10-20
Make difference between package -release and release on mobile.

Bluebugs:
### Description:
It is expected that package -release generate a binary stripped from as much information as possible (On mobile it means stripping debug information, but not stripping of symbol). While it is expected that release generate a signed package for consumption by stores.

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/mobile/build.go
+++ b/cmd/fyne/internal/mobile/build.go
@@ -273,10 +273,14 @@ var (
 )
 
 // RunNewBuild executes a new mobile build for the specified configuration
-func RunNewBuild(target, appID, icon, name, version string, build int, release bool, cert, profile string) error {
+func RunNewBuild(target, appID, icon, name, version string, build int, release, distribution bool, cert, profile string) error {
 	buildTarget = target
 	buildBundleID = appID
-	buildRelease = release
+	buildRelease = distribution
+	if release {
+		buildLdflags = "-w"
+		buildTrimpath = true
+	}
 
 	cmd := cmdBuild
 	cmd.Flag = flag.FlagSet{}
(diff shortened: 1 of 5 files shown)


andydotxyz:
Looks good to me, can Jacob or one other confirm our discussion from #fyne-contributors makes sense and is solved by this?

cmd/fyne/internal/mobile/build.go
```diff
@@ -273,10 +273,14 @@ var (
...
 
 // RunNewBuild executes a new mobile build for the specified configuration
-func RunNewBuild(target, appID, icon, name, version string, build int, release bool, cert, profile string) error {
+func RunNewBuild(target, appID, icon, name, version string, build int, release, distribution bool, cert, profile string) error {
 	buildTarget = target
 	buildBundleID = appID
-	buildRelease = release
+	buildRelease = distribution
+	if release {
+		buildLdflags = "-w"
```
Jacalz, line 281: Should this not have `-s` as well? The release binaries usually use `-ldflags="-s -w"` but I am not sure why this only has `-w`.

Jacalz (commented):
The code makes sense but I am not entirely sure what the discussion in #fyne-contributors was about. Left a small comment inline.

cmd/fyne/internal/mobile/build.go
Bluebugs: > Should this not have `-s` as well? The release binaries usually use `-ldflags="-s -w"` but I am not sure why this only has `-w`.

On android the application is a .so linked dynamically with the jvm. -s remove the symbol which will break the dynamic linking.  -w remove the dwarf symbol, also known as debugging symbol. This are safe to remove as only user by debugger.

Jacalz: Ah, I see. Sounds sensible

andydotxyz (approved):

lucor (approved):

PR #3348, opened 2022-10-21
Split container hidden sizing

### LINKED ISSUE #3232: Split Container does not respect item's Visible status
### Checklist

### Describe the bug

When an item in a Split container is hidden, the split should should act is if there is nothing there for that item.  Neither the `Layout` nor the `MinSize` function examine the `Visible` status of the items.

### How to reproduce

1. Run the attached code, which is a Split container with buttons on the left and a Label on the right
2. Drag the slider around to verify it works properly.  Note that if you drag the slider right to the MinSize of the Label, the pointer moves, but the slider does not.  While still holding the click, drag the slider back to the left and you can see that the pointer and the slider handle become out of sync.
3. Click the Hide button 
4. Drag the slider around.  Note that it behaves exactly the same, but with the label hidden.  I would expect that with the label hidden, I could drag the slider all the way to the right edge of the window

### Screenshots

CleanShot 2022-08-29 at 16 23 39

### Example code

```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	myApp := app.New()
	myWindow := myApp.NewWindow("Form Layout")

	label := widget.NewLabel("I am a longish label that just sits here")

	buttons := container.NewVBox(
		widget.NewButton("Show", func() {label.Show()}),
		widget.NewButton("Hide", func() {label.Hide()}),
	)

	c := container.NewHSplit(buttons, label)

	myWindow.SetContent(c)
	myWindow.Resize(fyne.NewSize(500, 300))
	myWindow.ShowAndRun()
}
```

### Fyne version

2.2.3

### Go compiler version

go version go1.19 darwin/amd64

### Operating system

macOS

### Operating system version

Monterey 12.5.1

### Additional Information

_No response_

itsjustdel:
### Description:

Addresses Split Container not respecting item's visible status if item is not in a container.  When calculating the size of each split, my fix checks if the item is visible and returns a size of 0 if not.

Fixes #3232

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container/split.go
+++ b/container/split.go
@@ -95,7 +95,7 @@ func (r *splitContainerRenderer) Layout(size fyne.Size) {
 	var dividerSize, leadingSize, trailingSize fyne.Size
 
 	if r.split.Horizontal {
-		lw, tw := r.computeSplitLengths(size.Width, r.split.Leading.MinSize().Width, r.split.Trailing.MinSize().Width)
+		lw, tw := r.computeSplitLengths(size.Width, r.minLeadingWidth(), r.minTrailingWidth())
 		leadingPos.X = 0
 		leadingSize.Width = lw
 		leadingSize.Height = size.Height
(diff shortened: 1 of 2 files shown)


Bluebugs (approved):
It does make sense actually, and there is tests \o/

Jacalz (approved):
Nice work. Looks good to me :)

andydotxyz (approved):
Thanks

PR #3349, opened 2022-10-21
Fix issue where implicit selection deletion can break text state

### LINKED ISSUE #3340: Data bound entry text replacing selection is ignored
### Checklist

### Describe the bug

a data bound entry will not correctly replace selected text with new input

### How to reproduce

* Go to a `NewEntryWithData`
* Add some text
* Select a few characters
* Type another
* See that the selected text is removed, but the new text does not get added

### Screenshots

_No response_

### Example code

```go
e := NewEntryWithData(data)
test.Type(e, "My content")
// select "con"
test.Type(e, "d")
assert.Equal(t, "My dtent", e.Text)
```

### Fyne version

2.2+

### Go compiler version

1.18

### Operating system

macOS

### Operating system version

Big Sur

### Additional Information

_No response_

andydotxyz:
Fixes #3340

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -728,7 +728,10 @@ func (e *Entry) TypedRune(r rune) {
 
 	// if we've typed a character and we're selecting then replace the selection with the character
 	if selecting {
+		cb := e.OnChanged
+		e.OnChanged = nil // don't propagate this change to binding etc
 		e.eraseSelection()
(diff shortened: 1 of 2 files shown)


Jacalz (approved):
Makes sense. Nice work :+1:

PR #3352, opened 2022-10-24
Support verbose output (issue #3069)

### LINKED ISSUE #3069: Add debug option and verbose output for all commands executed by fyne while building a release
I found that I can't get new release on macOS (darwin) while executing the following command:
```
fyne release -os darwin -appID com.example.app -appVersion 1.0 -appBuild 1 -profile development -category utilities --cert W123X123XYZ
2022/06/14 13:15:15 Fyne error:  Codesign failed
2022/06/14 13:15:15   Cause: exit status 1
2022/06/14 13:15:15   At: /Users/vk/Work/Languages/Go/gopath/pkg/mod/fyne.io/fyne/v2@v2.1.4/cmd/fyne/internal/commands/release.go:308
unable to codesign application bundle
```
Upon fyne codebase inspection I found that it runs few commands behind the scene and it would be extremely helpful to get verbose output of this commands, especially if there is an error. This can be achieved by adding debug option to release which will provide  the following:
- exact command fyne release execute
- both stdout and stderr of that comnand
- its exit status
This will allow developers to reproduce every step of the release build.

No

```
fyne release -debug ...

# and it will provide something like:
Command executed: codesign ...
Stdout of command output: bla-bla
Stderr of command output: error-error
Status of command: 1
```

vkuznet:
### Description:
End-user should be able to see which internal commands `fyne release` executes during build and other actions. I added `-verbose` flag to enable such output.

Fixes fyne-io/tools#55


cmd/fyne/internal/commands/release.go
```diff
@@ -12,8 +12,8 @@ import (
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/cmd/fyne/internal/mobile"
 	"fyne.io/fyne/v2/cmd/fyne/internal/templates"
+	cli "github.com/urfave/cli/v2"
```
Jacalz, line 15: This change does not seen correct. The name of the package is already `cli`.

cmd/fyne/internal/commands/release.go
```diff
@@ -104,6 +104,11 @@ func Release() *cli.Command {
 				Usage:       "Windows: the developer identity for your Microsoft store account",
 				Destination: &r.developer,
 			},
+			&cli.BoolFlag{
+				Name:        "verbose",
+				Usage:       "verbose output of used commands",
+				Destination: &r.verbose,
+			},
```
Jacalz, line 111: You might want to alias this to "-v".

cmd/fyne/internal/commands/release.go
```diff
@@ -176,6 +183,9 @@ func (r *Releaser) Run(params []string) {
 		return
 	}
 
+	if r.verbose {
+		mobile.Verbose = true
+	}
```
Jacalz, line 188: Should it not be enough to do `mobile.Verbose = r.verbose`?

cmd/fyne/internal/commands/release.go
```diff
@@ -194,6 +204,9 @@ func (r *Releaser) releaseAction(_ *cli.Context) error {
 		return err
 	}
 
+	if r.verbose {
+		mobile.Verbose = true
+	}
```
Jacalz, line 209: Same as comment above.

Jacalz (changes requested):
Thanks for opening the PR. I added a few suggestions inline

cmd/fyne/internal/commands/release.go
vkuznet: I think it is correct since all other imports you have in code use `v2`, see
- https://github.com/fyne-io/fyne/blob/develop/cmd/fyne/main.go#L10
- https://github.com/fyne-io/fyne/blob/develop/cmd/fyne/internal/commands/install.go#L14
- https://github.com/fyne-io/fyne/blob/develop/cmd/fyne/internal/commands/build.go#L13
etc.

I can remove `cli` alias though and leave only `"github.com/urfave/cli/v2"`

Jacalz: I only meant that the alias was incorrect. Sorry for any confusion

cmd/fyne/internal/commands/release.go
vkuznet: done

cmd/fyne/internal/commands/release.go
vkuznet: done

vkuznet:
Now, with this change I see the following output of fyne:
```
fyne release -help
..
   --verbose, -v                      verbose output of used commands (default: false)
```
and when I invoke fyne I see this:
```
fyne release -os iossimulator -verbose -appID io.github.vkuznet -appVersion 1.0 -appBuild 1 -profile developer -cert W2R33GWXXXX
echo "// !$*UTF8*$!
...
" > $WORK/main.xcodeproj/project.pbxproj
echo "<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
....
" > $WORK/main/Images.xcassets/AppIcon.appiconset/Contents.json
xcrun lipo -o $WORK/main/main -create
xcrun xcodebuild -configuration Release -project $WORK/main.xcodeproj -allowProvisioningUpdates DEVELOPMENT_TEAM=67N29KRXXX
```
So, indeed it prints all commands it uses during a build.

cmd/fyne/internal/commands/release.go
vkuznet: Moreover, I see the following build error if I do not use `v2`:
```
go build
package fyne.io/fyne/v2/cmd/fyne
	imports fyne.io/fyne/v2/cmd/fyne/internal/commands
	imports github.com/urfave/cli: no Go files in /Users/vk/tmp/fyne/vendor/github.com/urfave/cli
```
So, based on your go.mod file it is required to have v2, see https://github.com/fyne-io/fyne/blob/develop/go.mod#L28

Therefore, I only removed alias `cli` and left `"github.com/urfave/cli/v2"`

andydotxyz:
Thanks so much for doing this, however I have a question / thought.
Verbose is more logging, to help with information - but is a "debug" flag not normally used to print out actual commands to the console? I guess the question is, aren't the commands used to run it and the output of those commands different levels of inspection?

vkuznet:
@andydotxyz , I'm not sure I understand what you are asking for. Do you want to rename `-verbose` to `-debug`? So far there is no `-debug` option in `fyne release` set. To me there are equivalent, but it depends on definition and scope of the flag.

vkuznet:
But I need to add that I disagree how `fyne release` works since it embed all commands which require code changes. For instance, if tomorrow Apple will change `xcrun`, or `xcodebuild` or anything (in this regard Android too), then fyne requires to adopt those changes. It would be more flexible and portable if you will not embed any commands but rather provide a shell script to build on different platforms, e.g.
- on iOS you provide `fyne_release_ios.sh` script which will include Apple specific set of commands to build a release,
- on Android you provide `fyne_release_android.sh` which will perform a different set of commands, etc.

Such scripts will not only educate users about how build should be done, but will allow much easier maintenance and less code writing. There is nothing wrong to supplement any codebase with aux scripts to perform certain actions. The shell scripts are portable too and within them we can invoke `go`, `xcodebuild`, `echo`, or any other commands you use so far from Go code itself.

andydotxyz:
I'm not suggesting a rename of the parameters, outputting the command results with "-verbose" makes sense. I was just wondering (out loud) if the commands which generate the output would be classified as debug instead of verbose?

For example gradle splits these to different logging levels: https://docs.gradle.org/current/userguide/logging.html
One warning in their docs is that outputting commands can leak security information, which you won't need when looking for command output...

andydotxyz:
> There is nothing wrong to supplement any codebase with aux scripts to perform certain actions.

Indeed, but it is important to remember that the `fyne` command, like most Go, aims for single binary distribution. To use shell scripts external to the app would require either an additional installer, or for the binary to unpack scripts at runtime just to run them. And then you have versioning/change detection to worry about - when we ship an update to `xcrun` calls or `Play Store` requirements we cannot blindly overwrite files that we have encouraged users to edit...

As long as we can maintain a reasonable release cycle (which for the most part we do) nobody should be left out in the cold when a change happens. The `fyne` command, much like all of Fyne, should "just work" and we believe it does.

vkuznet:
I do not have strong opinion about verbose/debug argument and will rely on your decision. And, I understand single binary distribution argument too, but as a developer I need to know what is happened in a build since it allows me to debug issues. In my case, I was battling with output error to identify certificate problem. For that I was need to see all output commands fyne is doing. About security concern, it is valid point too, but it boils down to what print command is doing, does it prints only command itself of its output/errors stdout/stderr streams. But this I consider as separate issue from ability to see the commands.

Anyway, please let me know if you need anything else from this RP?

andydotxyz:
> Anyway, please let me know if you need anything else from this RP?

Let's get @Jacalz opinion on the parameters.
But firstly we need you to address the build failure - complexity has risen too high.

vkuznet:
@andydotxyz , I doubt because changes I made, but rather that existing code was close to certain threshold. For instance from `gocyclo -over 30 .` output I see the following: https://coveralls.io/builds/53535562/source?filename=cmd%2Ffyne%2Finternal%2Fcommands%2Frelease.go#L107 As such I only added one block for command line option and it cause to overflow your code coverage. The changes in this PR is negligible but due to the fact that code was already close to complexity threshold the coverage drop was noticeable.

Bluebugs:
It is actually the || verbose that make gocyclo fail and the static analysis pass to fail, not the coverage going down. An easy fix would actually to use buildX in the mobile package instead of introducing a new variable there.

andydotxyz:
> The changes in this PR is negligible but due to the fact that code was already close to complexity threshold the coverage drop was noticeable.

This is probably true. We have thresholds set so that we avoid complexity going too high. The changes here cause it to go over, so the code should be simplified (either splitting into simpler methods, or do as @Bluebugs suggests). All code has the same constraint and over time we are hoping to reduce this complexity number further!

vkuznet:
I hear different opinions here. Please tell me exactly how you want to proceed with this. I see the following options:
- re-factor code to pass coverage and other tests, but nobody told me what is your vision on that.
- remove `verbose` in favor of `buildX` but it will require to move `buildX` from mobile.go to release.go, again I need clear message here how you want to proceed.
- I can change `buildX || verbose` if statements, again I need a clear message what is your proposal here

Please come to consensus and outline clear how you envision to proceed, then I can change code accordingly.

Bluebugs:
> Please come to consensus and outline clear how you envision to proceed, then I can change code accordingly.

The consensus is to have a solution that doesn't increase code complexity (which gocyclo is a proxy). Whatever you feel best in your opinion to pass that requirement will be acceptable here. So any of the solution you are pointing are fine solution. It is up to your choice to do what you prefer.

Bluebugs:
I can go a bit further and highlight draw back of each solution:
- refactoring increase likeliness of bug and require more work
- removing verbose in build.go in favor of using buildX, require likely to expose buildX and rename it to BuildX then directly use it in release.go. Likely multiple place to change due to the rename, but give a chance to be sure no unintended effect come from using buildX.
- changing the if statement is likely more tricky as you need to figure out how to appease gocyclo in doing that change.

I would personally go with the second option, but I will be ok with the other two.

vkuznet:
I opt-ed to the second option as it seems less controversial, and in fact it allows now to see full build options and commands like:
```
GOOS=ios GOARCH=arm64 CC=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang CXX=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ CGO_CFLAGS=-isysroot/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator15.2.sdk -mios-simulator-version-min=7.0 -arch arm64 CGO_CXXFLAGS=-isysroot/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator15.2.sdk -mios-simulator-version-min=7.0 -arch arm64 CGO_LDFLAGS=-isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator15.2.sdk -mios-simulator-version-min=7.0 -arch arm64 CGO_ENABLED=1 go build -tags ios release -x -ldflags -w -trimpath -ldflags=-w -o=$WORK/arm64 github.com/vkuznet/ecmxcrun lipo -o $WORK/main/main -create $WORK/arm64
mkdir -p $WORK/main/assets
...
```
which I consider a plus. But as was pointed out it required to switch from local `buildX` to global `BuildX` variable. I still left `-verbose` as a flag though in release command, and it is identical to `-x` flag in build command. I do not know if it is acceptable from your point of view, but I think `-x` is less intuitive in release command and may confuse people with such choice. Please let me know your opinion.

Bluebugs (approved):

andydotxyz (changes requested):
As I tried to explain earlier I don't think public global variables set to configure the build is the right approach. Surely you can reuse "buildX" like @Bluebugs suggested without having to leak this internal detail from the package?

cmd/fyne/internal/commands/release.go
```diff
@@ -12,8 +12,8 @@ import (
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/cmd/fyne/internal/mobile"
 	"fyne.io/fyne/v2/cmd/fyne/internal/templates"
-
 	"github.com/urfave/cli/v2"
+
 	"golang.org/x/sys/execabs"
```
Jacalz, line 17: Can you please restore this to how it was?

Jacalz (changes requested):
I agree with Andy. Just have a small request above that

vkuznet:
@andydotxyz , I doubt I understand what you're asking. Please re-iterate as I don't see that you expressed clearly what you had in mind. In other words, if I'll keep `buildX` I can't use it in `release.go` because it is not public, i.e. I can't set `mobile.buildX` in release.go as it is not visible. The only way to set it in release.go was to make it public which is what I did. The whole problem is that in release.go we need to have a flag and based on this flag we need to change `buildX` in mobile package in order to force its printouts.

andydotxyz:
Sorry I've not been more clear @vkuznet - you're right it cannot be set from a different package if it is not public, but this is desirable. As we are enabling "verbose" what I would expect is to pass that into the mobile build and then internally it can use that to set `buildX` - this way you can re-use the existing variables without exposing the internal workings.
I hope this makes sense. Essentially going back to @Bluebugs first comment you can avoid the duplication of `verbose` and `buildX` by simply setting it from the variable passed in (as you were doing before) to avoid adding the new verbose flag in the mobile package.

vkuznet:
@andydotxyz , in order to do what you ask for we either need to add new public function to build package, e.g
```
// package mobile
func SetFlags(verbose bool) {
    buildX = verbose
}
```
which will set appropriate build flags (so far buildX but it can be extended further), or may be more flexible and extendable approach and convert build flags of mobile package into public struct, e.g. 
```
// package mobile
type BuildFlags struct {
   buildA bool
   buildX bool
...
}
```
which then can be used elsewhere (the build package will initialize it in `init()`, while release package will initialize it using its own set of input options).

In first case, we can call `SetFlags` from release codebase, while if we choose the latter we can pass into mobile any supported flags from the release codebase, e.g. in release user can specify `ldflags`, etc. My original idea with this PR was to minimize the changes, but from user and design point of view it would be better to give full control of all build flags to release codebase. Once again, please provide feedback that I'll not waste time to implement something which will require further debate.

andydotxyz:
We pass build parameters into `mobile` via `RunNewBuild`. Is it not possible to continue with that pattern to avoid public global variables?

vkuznet:
@andydotxyz No, it is not possible since there is no calls to this function from release codebase. Once again, my target was to do the following:
```
fyne release -verbose ...
```
and it should provide me full output of commands used to make a release on specific target. Again, please give me directions how to proceed since I do not aware of internal design and how different packages should communicate with each other. Since I was doing release for ios it triggered mobile build, and for that I used verbose flag which required changes into mobile package. But I did not explored how release will trigger build on different targets, e.g. android or linux, etc.

andydotxyz:
Isn't the path:

```
release.go: Run() -> r.Packager.Run()
package.go: Run() -> p.doPackage()
package.go: doPackage -> p.packageIOS()
package-mobile.go -> mobile.RunNewBuild()
```

So, if `verbose` was stored in `Packager` (and then it could nicely help with being verbose for non-release package too) the variable will be available at the moment it should be passed to mobile package.

vkuznet:
@andydotxyz , thanks for the pointers, now it is clear (as I'm new to the code I was not aware this flow). So, I put verbose into Packager, reverted buildX and hopefully now code will satisfies your vision. Please have a look after the build checks.

vkuznet:
can someone trigger the tests to see how new commits fits into them.

Jacalz:
Sure. I've approved them now

vkuznet:
All checks are passed but merging is still blocked. Do you need anything from me to proceed forward with this?

andydotxyz:
I think there is an unresolved comment from Jacob if you read up.
Also as this is only logging xxx run calls it seems like there are some other exec like the android build packages - are there more things that should be printed? (Checking package as well now that it's not release specific)

cmd/fyne/internal/commands/release.go
vkuznet: resolved now

vkuznet:
I resolved extra line comment (sorry it is done automatically by vim-go). And, I turned on `buildV` option which is what build_android.go uses to print verbose output. Please review.

Jacalz (dismissed):
The code looks clean and good to me. Thanks for working on this. Let's make sure that someone else, with better knowledge of the mobile code, also reviews this before merging :)

andydotxyz:
Can you please rebase on the latest develop @vkuznet? It should fix the remaining tests

vkuznet:
I did rebase, please fire up tests.

vkuznet:
all tests passed, do you need anything else from me?

vkuznet:
@andydotxyz is anything else blocking this PR to be merged?

andydotxyz (commented):
I think this is an open question following the merging of -x and -v to one parameter

andydotxyz:
Apologies - i left that inline comment 3 days ago but it seems not to have submitted. Fixed now

andydotxyz:
@Jacalz @Bluebugs @vkuznet any thoughts on turning on verbose and debug in one flag (possible leak of password to logger)

vkuznet:
I do not have real preferences. But I think it depends if you want to support backward compatibility and compatibility with go-mobile, if so two flags should be there, otherwise you may merge into one.

Jacalz:
I'd vote for two flags. Seems a bit scary to potentially print passwords and secrets to the standard output

andydotxyz:
I'm not too concerned about go-mobile compatibility, but I do think they split it for a reason.
Perhaps we can define `verbose` as providing more info on what the code is doing and `debug` as literally printing the executables called?

andydotxyz:
It seems like there was agreement that verbose (command output) and debug (logging the commands to be run) should be separate flags - are you interested to complete this PR @vkuznet ?

vkuznet:
I have limited time this time of the year, but I resolved existing conflict and committed the changes. Feel free to trigger new build to check tests. And, I may come back to this issue in couple of weeks once I back home.

Jacalz:
The cmd/fyne tool has moved into the https://github.com/fyne-io/tools project now. Feel free to continue the work there. I will close this.

PR #3356, opened 2022-10-26
force hide horizontal and vertical scrollbars on wasm

fuskiid:
### Description:
Scrollbar is shown for fyne apps running in the browser using wasm, adding `scroll="no" style="overflow: hidden"` to body will forcefully turns off these unwanted scrollbars.

Fixes #2827

PR #3357, opened 2022-10-27
fixed accidental removal of (getChromeVersion() < 106) from bundled r…

fuskiid:
### Description:
Fixes regression of #3356 which removed chrome version check from bundled resource.


andydotxyz (approved):
Thanks for fixing

PR #3359, opened 2022-10-28
Some nice new test helpers for rendering images

andydotxyz:
A few quality of life updates to image based testing.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


test/test.go
```diff
@@ -29,6 +30,47 @@ func AssertCanvasTappableAt(t *testing.T, c fyne.Canvas, pos fyne.Position) bool
...
+//
+// Since 2.3
+func AssertObjectRendersToImage(t *testing.T, masterFilename string, o fyne.CanvasObject, msgAndArgs ...interface{}) bool {
+	c := NewCanvasWithPainter(software.NewPainter())
+	c.SetPadded(false)
+	size := o.MinSize().Max(o.Size())
+	c.SetContent(o)
+	c.Resize(size) // ensure we are large enough for current size
+
+	img := c.Capture()
```
toaster, line 48: nit: I’d inline this or maybe even call `AssertRendersToImage` but that’s more a personal preference.

andydotxyz: Good point, I missed that due to the order I coded it in I think, thanks

test/test_test.go
```diff
@@ -33,6 +33,56 @@ func TestAssertCanvasTappableAt(t *testing.T) {
...
+	c := test.NewCanvas() // no painter, would be cycle - images will be blank
+	c.SetContent(canvas.NewCircle(color.Black))
+	c.Resize(fyne.NewSize(10, 10))
+
+	t.Run("non-existing master", func(t *testing.T) {
+		tt := &testing.T{}
+		assert.False(t, test.AssertRendersToImage(tt, "non_existing_master.png", c), "non existing master is not equal to rendered image")
+		assert.True(t, tt.Failed(), "test failed")
+		_, err := os.Stat("testdata/failed/non_existing_master.xml")
+		exists := os.IsNotExist(err)
```
toaster, line 60: From the documentation of `IsNotExist`:
> New code should use errors.Is(err, fs.ErrNotExist).

Do we need to stick to `os.IsNotExist` to support older Go versions? AFAICS, `errors.Is` was added in Go 1.13.

toaster: Oh, and the outcome of `os.IsNotExist(err)` is if the file does _not_ exist. Therefore this test seems to be a false positive.

andydotxyz: We don't need to - I just thought it read more cleanly. Wasn't the `fs.ErrNotExist` they recommend using not more recent than 1.14 though?

toaster: But the test runs say it’s okay, what am I missing?

toaster (changes requested):
Please check why the tests about the failure files don’t fail.

test/test_test.go
toaster: > Wasn't the fs.ErrNotExist they recommend using not more recent than 1.14 though?

Probably not, `os.IsNotExist` checks for `fs.ErrNotExist`.

andydotxyz: It will in a modern package, but was the `io/fs` not new in 1.16?

andydotxyz:
Apologies for the bad tests @toaster - I must have been half asleep when I coded them up. Fixed now!

toaster:
Cool!

andydotxyz:
> Cool!

Thanks so much, and apologies for the embarrassing test mistakes!

PR #3364, opened 2022-10-30
First outline of what app skeleton could be

andydotxyz:
### Description:
Start work on what will be a super helpful container for app layout.

looking for suggestions of other great stuff to include…

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container/skeleton.go
+++ b/container/skeleton.go
@@ -1,0 +1,38 @@
+package container
+
+import (
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/layout"
+	"fyne.io/fyne/v2/widget"
+)
+
+// AppSkeleton is a container that assists in laying out a complete application.
+//
+// since: 2.3
+type AppSkeleton struct {
+	widget.BaseWidget
+	content *fyne.Container
+}
+
+// NewAppSkeleton creates a widget that lays out a standard application container.
+// The content passed in will fill the space left after any accessories are packed in.
+//
+// Since: 2.3
+func NewAppSkeleton(content fyne.CanvasObject) *AppSkeleton {
+	a := &AppSkeleton{content: NewBorder(nil, nil, nil, nil, content)}
+	a.ExtendBaseWidget(a)
+	return a
+}
+
+// CreateRenderer is an internal method for returning the information for rendering this skeleton.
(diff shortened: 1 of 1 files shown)


Bluebugs:
I have been thinking about this and I wonder if the intent shouldn't be in what kind of application it is rather than about the content. And ideally something you define in the constructor.

For example, if I plan to do a classic desktop application, I would expect to pass the content, a toolbar, a status bar and a menu. If I plan to do a mobile application, I would guess a menu, a toolbar and a series of tabs that contain each a list. If I plan to do a desktop agent, I would expect a systray menu, a status bar and a series of tab. Their is also the modern desktop application, something a bit like we do in fyne_demo, except no top menu but just a side list with icon to navigate between panel (close to the mobile application, but instead of list it tends to be more frequently a grid or just panel). That last one might be actually the one that could be truly cross platform in usage as there could be a migration path from it to the classic mobile layout.

Another more specialized one could be the dashboard application which is how things like home assistant lovelace and many other application look like. It is basically a menu (with icons on the left) and a series of cards on the right. Using something like fyne-x responsive layout to automatically deal with those, it could be quite powerful for cross platform dashboard kind of application. The number of cards/entry being limited, it can be treated differently from the mobile application long list problem. This one could also be truly cross platform, I think.

andydotxyz:
> For example, if I plan to do a classic desktop application, I would expect to pass the content, a toolbar, a status bar and a menu. If I plan to do a mobile application, I would guess a menu, a toolbar and a series of tabs that contain each a list

I appreciate the thoughts, but the concept of splitting APIs based on desktop vs mobile etc goes agains the design principles of a platform-agnostic GUI toolkit in my opinion.

I agree that it should be a clean API based on the type of application but I think "list and detail" is more the type of app than "desktop with toolbar"...

Bluebugs:
> I appreciate the thoughts, but the concept of splitting APIs based on desktop vs mobile etc goes agains the design principles of a platform-agnostic GUI toolkit in my opinion.

This is a good point. Still their are family of application that can't have the same features when in a pure mobile, tablet or TV mode. But then, when you get to add a real keyboard to an iPad or use something like Samsung Dex, those features should be there. Really sorry, I do not have an idea here on how this could be expressed by developers.

Issue #3386, opened 2022-11-07
RagioGroup miscalulates label widths in horizontal mode


grkuntzmd:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

When long labels are used in the RagioGroup in horizontal mode, they overlap.

### How to reproduce

`go run  .`

### Screenshots

![image](https://user-images.githubusercontent.com/703718/200319477-ff4b62da-fe28-4a5c-b522-d0ccdab83950.png)

### Example code

```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Radio Group Bug")

	rg := widget.NewRadioGroup([]string{"Easy", "Medium", "Hard", "Expert"}, func(s string) {})
	rg.Horizontal = true

	w.SetContent(rg)
	w.ShowAndRun()
}
```

### Fyne version

2.2.3

### Go compiler version

1.19.3

### Operating system

Windows

### Operating system version

Windows 11 Pro Version	22H2

vinser:
It's a pity but the bug still is being on. Why it's marked as Unverified?

In horizontal mode Layout method calculates average width size for radioGroup items/ 
```go
// Layout the components of the radio widget
func (r *radioGroupRenderer) Layout(_ fyne.Size) {
	count := 1
	if r.items != nil && len(r.items) > 0 {
		count = len(r.items)
	}
	var itemHeight, itemWidth float32
	minSize := r.radio.MinSize()
	if r.radio.Horizontal {
		itemHeight = minSize.Height
		itemWidth = minSize.Width / float32(count) // HERE IS THE BUG
	} else {
		itemHeight = minSize.Height / float32(count)
		itemWidth = minSize.Width
	}

	itemSize := fyne.NewSize(itemWidth, itemHeight)
	x, y := float32(0), float32(0)
	for _, item := range r.items {
		item.Resize(itemSize)
		item.Move(fyne.NewPos(x, y))
		if r.radio.Horizontal {
			x += itemWidth
		} else {
			y += itemHeight
		}
	}
}

```
The bug makes horizontal items to overlap if they have different width and it looks poor :(
 
[...]

PR #3389, opened 2022-11-09
Fix issue #3386 - Fix issue #3386 - RagioGroup miscalculates label widths in horizontal mode

grkuntzmd:
The horizontal radio group miscalculates the width of labels when there is a mix of short and long labels. It uses the total label length, with padding added, divided by the number of items, instead of the more correct maximum min width.


andydotxyz:
The code looks good - but the tests that are failing need to be copied from "testdata/failed" folder into "testdata" then add them to git, it will show you the how the change impacts some test cases. Then we can make sure it just impacts the expected output.

Jacalz:
Closing in favor of #3763 that landed recently. Thanks for working on this.

PR #3394, opened 2022-11-11
Take two at optimizing key translation

Jacalz:
### Description:

This is a rework of the improvements in #3179. Not only are the shortcut issues fixed but certain other aspects have
also been improved further. For example, I noted that a few glfw keys (and strings) just map directly to ascii so
we can often just reuse the values directly without even having to switch on them.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window_desktop.go
+++ b/internal/driver/glfw/window_desktop.go
@@ -526,11 +526,9 @@ func keyCodeToKeyName(code string) fyne.KeyName {
 	}
 
 	char := code[0]
-
-	relativeLetter := char - 'a'
-	if relativeLetter <= 'z' {
+	if char >= 'a' && char <= 'z' {
 		// Our alphabetical keys are all upper case characters.
-		return fyne.KeyName('A' + relativeLetter)
+		return fyne.KeyName('A' + char - 'a')
 	}
 
 	switch char {
--- a/internal/driver/glfw/window_desktop.go
+++ b/internal/driver/glfw/window_desktop.go
@@ -531,7 +531,36 @@ func keyCodeToKeyName(code string) fyne.KeyName {
 		return fyne.KeyName('A' + char - 'a')
 	}
 
-	return fyne.KeyName(char)
+	switch char {
+	case '[':
+		return fyne.KeyLeftBracket
+	case '\\':
+		return fyne.KeyBackslash
+	case ']':
+		return fyne.KeyRightBracket
+	case '\'':
+		return fyne.KeyApostrophe
+	case ',':
+		return fyne.KeyComma
+	case '-':
+		return fyne.KeyMinus
+	case '.':
+		return fyne.KeyPeriod
+	case '/':
+		return fyne.KeySlash
+	case '*':
+		return fyne.KeyAsterisk
+	case '`':
+		return fyne.KeyBackTick
+	case ';':
+		return fyne.KeySemicolon
+	case '+':
+		return fyne.KeyPlus
+	case '=':
+		return fyne.KeyEqual
+	}
+
+	return fyne.KeyUnknown
 }
 
 func keyToName(code glfw.Key, scancode int) fyne.KeyName {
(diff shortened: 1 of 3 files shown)


andydotxyz (changes requested):
Looks like 
```
	"[":  fyne.KeyLeftBracket,
	"\\": fyne.KeyBackslash,
	"]":  fyne.KeyRightBracket,

	"'": fyne.KeyApostrophe,
	",": fyne.KeyComma,
	"-": fyne.KeyMinus,
	".": fyne.KeyPeriod,
	"/": fyne.KeySlash,
	"*": fyne.KeyAsterisk,
	"`": fyne.KeyBackTick,

	";": fyne.KeySemicolon,
	"+": fyne.KeyPlus,
	"=": fyne.KeyEqual,
```
have been removed? That doesn't seem right.

Also how come the gocyclo ignore was added?

Jacalz:
> have been removed? That doesn't seem right.

The idea is that they are all length one and can be converted automatically using `fyne.KeyName(char)` but I realise now that, while the code works for those values, it might produce invalid `fyne.KeyName` variants for other one length strings. Will get that fixed up.

>  Also how come the gocyclo ignore was added?

That was also present in the old PR. For some reason it likes huge maps but not switch statements.

andydotxyz:
Any idea why the drop in coverage is high?
Combined with adding complexity ignore lines makes me feel a little nervous, especially so close to release.
Is it worth revisiting after v2.3.0 so we can resolve those issues and test it thoroughly?

Jacalz:
I would prefer to have it in this release as it does improve memory usage quite a bit. I think I remember seeing almost a 1MB decrease. Will add some extra tests to it. Perhaps the functions aren't as well tested as I thought they were

Jacalz:
I added tests for all the key translation stuff now. Everything passes and works as expected using the old input maps as reference for tests to run :)

Jacalz:
We unfortunately won't get any coverage until https://github.com/fyne-io/fyne/pull/3448 has been merged. Sorry for that but to be fair, I wasn't the one that added an explicit check for Go 1.17 :see_no_evil:

andydotxyz:
GopherJS does not work with 1.19 iirc, so the check is there for reasons. Maybe we should not have moved everything off 1.17 so fast, as part of the web code still requires it.

Jacalz:
Yeah, I agree. Will get that sorted

Jacalz:
Sorry for the force-push. Had to get it to run the Go 1.17 workflows again now that we reverted back

Jacalz:
Coverage has now increased by 0.2% :)

internal/driver/glfw/window_desktop.go
```diff
@@ -400,167 +400,181 @@ func convertMouseButton(btn glfw.MouseButton, mods glfw.ModifierKey) (desktop.Mo
...
-	glfw.KeyKP5: fyne.Key5,
-	glfw.Key6:   fyne.Key6,
-	glfw.KeyKP6: fyne.Key6,
-	glfw.Key7:   fyne.Key7,
-	glfw.KeyKP7: fyne.Key7,
-	glfw.Key8:   fyne.Key8,
-	glfw.KeyKP8: fyne.Key8,
-	glfw.Key9:   fyne.Key9,
-	glfw.KeyKP9: fyne.Key9,
+	case glfw.KeyKP0:
```
andydotxyz, line 498: The comment above makes me think it's important we check the codes for things like Azerty, so should this not be matching before, per:
`case glfw.key0, glfw.KeyKP0:` 
instead?

andydotxyz (changes requested):

internal/driver/glfw/window_desktop.go
Jacalz: I moved them all into the switch. Should be good to go :)

andydotxyz (approved):
Looks good thanks.
Did an attempt at testing with AZERTY too and it seems happy.

PR #3398, opened 2022-11-12
Fix widgets using theme.BackgroundColor() in dialogs

Jacalz:
### Description:

The dialogs use a different color for the background now. It looks better to have transparency rather than the wrong background.
image

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Thanks

PR #3403, opened 2022-11-13
RichText image

### LINKED ISSUE #2366: Add image support to rich text
New RichText needs more features :)

You could split a RichText into two and place an image between them - but that won't support inline images

Block and inline images as part of the RichText Segment APIs

andydotxyz:
Insert an image and set the rendering to be @2x native pixels as discussed in our call.
Best compromise between consistency across platforms and a reasonable clarity on high-end devices.

Fixes #2366

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (approved):
Looks good to me. Nice work :+1:

PR #3408, opened 2022-11-15
add Table.SetRowHeight and List.SetItemHeight to match Table.SetColumnWidth

andydotxyz:
Much better support for variable height rows

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/list.go
```diff
@@ -210,6 +223,43 @@ func (l *List) UnselectAll() {
 	}
 }
 
+func (l *List) visibleItemHeights(itemHeight float32, length int) (visible map[int]float32, offY float32, minRow, maxRow int) {
```
Bluebugs, line 226: This doesn't seems ideal in term of performance as it will force a reallocation, recalculate and freeing every time the list is updated. It would be better if the map was allocated as part of the widget structure and just updated if the position of the visible item change.

widget/list.go
```diff
@@ -409,12 +459,26 @@ func (l *listLayout) Layout([]fyne.CanvasObject, fyne.Size) {
...
+	if f := l.list.Length; f == nil {
+		return fyne.NewSize(0, 0)
+	} else {
+		items = f()
 	}
-	return fyne.NewSize(0, 0)
+
+	height := float32(0)
+	templateHeight := l.list.itemMin.Height
+	for item := 0; item < items; item++ {
```
Bluebugs, line 471: This is, I fear, going against the current efficiency of a list and might trigger performance regression. It would be safer to iterate over the itemHeights content and then add the templateHeight * unknown size item. This would make performance for current user of the list not change.

widget/list.go
```diff
@@ -95,6 +95,19 @@ func (l *List) MinSize() fyne.Size {
...
 }
 
+// SetItemHeight supports changing the height of the specified list item. Items normally take the height of the template
+// returned from the CreateItem callback. The height parameter uses the same units as a fyne.Size type and refers
+// to the internal content height not including the divider size.
+//
+// Since: 2.3
+func (l *List) SetItemHeight(id ListItemID, height float32) {
+	if l.itemHeights == nil {
+		l.itemHeights = make(map[ListItemID]float32)
```
Bluebugs, line 105: I am a bit worried of this bit of code as it means every access to itemHeights could get access to a nil and they should all be protected. Not to sure how to address this, maybe having a function for all API entry point to check if this map is allocated and allocate it if not.

Bluebugs (changes requested):

widget/list.go
andydotxyz: I just copied the code that has been in table for the last major release.
Do we need to review that as well?

widget/list.go
andydotxyz: It is safe to index, we always check `ok` to know the value is real.

The code:
```go
func main() {
	var m map[int]string

	v, ok := m[4]
	fmt.Println("Hello", v, ok)
}
```

Returns "Hello  false"

widget/list.go
andydotxyz: We are just iterating an index, I think it would require benchmarks to show at what number of overrides this is slower than all the map accesses and calculating the gaps and doing the maths at the end.
As above it is the same code we have been using in Table for some time. Does it need fixed there as well, or can we return to it later as we optimise?

widget/list.go
Bluebugs: Oh, I didn't know that. Good!

widget/list.go
Bluebugs: There is a change, as far as I can see, compared to the previous code, it is now iterating over the table which it wasn't before. If I understand it correctly, it is starting from zero and might break early, so as your list of content get longer and you are displaying the end of the list, this start to be a more costly operation which will impact current users even if they do not use the custom height. Minimizing the impact of switching from a O(1) algorithm to an O(n) seems like a good idea here. My understanding is that this won't be called as often as MinSize, so I will be ok if we delay optimizing this part, but it would be good to take note that rendering the list will now be a function of the number of items which wasn't the case before.

andydotxyz:
In the prior column code the `Table` widget iterated the entire length in exactly this way, just search for `columnWidths` in `table.go`

Bluebugs:
> In the prior column code the `Table` widget iterated the entire length in exactly this way, just search for `columnWidths` in `table.go`

I am indeed not to worried with the `Table` users as you point out they were already going through a O(n) case before this PR, but this isn't the case with the `List` users.

andydotxyz:
>  For people who do not use the custom height feature, it will instantly slow them down.

I think this is the main point. For people with "some" custom heights it gets complicated, and for people with "a lot" of custom heights the map lookup could even be slower than the list iteration.

How about I add an escape clause at the top that checks if there are no custom heights, then just returns the old calculation and skips the loop?

Bluebugs:
> > For people who do not use the custom height feature, it will instantly slow them down.
> 
> I think this is the main point. For people with "some" custom heights it gets complicated, and for people with "a lot" of custom heights the map lookup could even be slower than the list iteration.

I don't think that the change I propose will have any impact on the custom height case. Currently your code for MinSize does:
```
for item := 0; item < items; item++ {
    itemHeight, ok := l.list.itemHeights[item]
```
This means it is checking in the map for each item even when they do not exist. It does iterate, compute the hash and does a lookup for each item.

What I do think it should be:
```
height := 0
total := 0
for idx, itemHeight := range l.list.itemHeights {
   if idx < items {
       total++
       height += itemHeight
   }
}
height += (items - total) * templateHeight
```

The worst case scenario now will be for only people that add more items height to the map than there is items in the list. Otherwise this is still O(1) when there is no item height and O(n) when there is custom items height. The cost of those iteration, in comparison with the previous loop, it does not compute the hash as it doesn't need to do a lookup, but just need to iterate. When the len(map) == len(list) as a worst case scenario, it would still be more efficient than the proposed code does here.

> How about I add an escape clause at the top that checks if there are no custom heights, then just returns the old calculation and skips the loop?

I don't think there is a benefit in keeping this code as is and would prefer that we do not have it improved, but if you really want to keep it in this form, I guess the early exit would do the job to prevent the worst case scenario for existing users.

andydotxyz:
> I don't think there is a benefit in keeping this code as is and would prefer that we do not have it improved.

I don't really understand this sentence sorry. Was the "not" a typo?

Bluebugs:
> Was the "not" a typo?

Indeed, sorry.

andydotxyz:
> The worst case scenario now will be for only people that add more items height to the map than there is items in the list.

I think this is a slight mis-characterisation. The `visibleItemHeights` will exit when it reaches the bottom of the visible scroll.
When iterating the map there is no guarantee of order, so we are forced to iterate all the way to the end, so the current code that is `O(n)` could at worst case be the same as a heavily populated map. But when the list is not fully scrolled the `n` could be much shorter than a full map range lookup.

Bluebugs:
> > The worst case scenario now will be for only people that add more items height to the map than there is items in the list.
> 
> I think this is a slight mis-characterisation. The `visibleItemHeights` will exit when it reaches the bottom of the visible scroll. When iterating the map there is no guarantee of order, so we are forced to iterate all the way to the end, so the current code that is `O(n)` could at worst case be the same as a heavily populated map. But when the list is not fully scrolled the `n` could be much shorter than a full map range lookup.

My main concern and the above code was for MinSize. I understand that the change for visibleItemHeights are more complex than that (and why I suggested a cache earlier) and I will be fine if that one does just the early exit at this point.

andydotxyz:
Ah good point, `MinSize` is more trouble. Adding the escape clause for List and Table in MinSize...

Bluebugs:
> Ah good point, `MinSize` is more trouble. Adding the escape clause for List and Table in MinSize...

I think I didn't convey what I had in mind correctly. I think `MinSize` should be optimized with the code I showed above and `visibleItemHeights` should get an early exit code. The reason I think `MinSize` should be optimized instead of an early exit, is that I do prefer one code path for everyone and if we can made that code path the same complexity as before, it is a better solution. As for `MinSize` it seems doable, I think that should be the way forward. For `visibleItemHeights` it doesn't seems to be doable and so I am fine with an early exit for the moment.

andydotxyz:
I'm worried about premature optimisation and adding complexity at a late stage.

This should have landed over a week ago so wondering if we should just forget about this desirable addition and revisit for a future release.

andydotxyz:
I then added custom heights to every 10th row, as:
```go
	for i := 0; i < 10_000; i+= 10 {
		list.SetItemHeight(i, 12)
	}
```

And got the following results:
```
1 took 1.616637ms
2 took 807.449µs
3 took 1.147593ms
4 took 1.286658ms
5 took 1.494822ms
avg 1.270631ms
```

My conclusion here is that there does not seem to be anything to worry about.

andydotxyz:
The first response is obviously doing a lot of initial minsize and creating the cached items.
The subsequent ones will be doing a little layout calculations but omit creating all the child widgets, hence why they are faster.

I could iterate over the iterating to get more data, but these seemed indicative on my PC.

Bluebugs (approved):

Issue #2657, opened 2021-11-22
Update Linux/XDG app settings to delegate theme to new fd.o color-scheme


andydotxyz:
Updating the `app/app_xdg.go` function `defaultVariant` to look up the value now defined "globally" https://github.com/flatpak/xdg-desktop-portal/blob/d7a304a00697d7d608821253cd013f3b97ac0fb6/data/org.freedesktop.impl.portal.Settings.xml#L33-L45

Jacalz:
We might possibly also be able to read the gsettings value for `org.gnome.desktop.interface gtk-theme` and see if it contains "dark". This is, as an example, how you would set dark theme on Gnome using the terminal:
```
gsettings set org.gnome.desktop.interface gtk-theme 'Adwaita-dark'
```

Jacalz:
I have dug a bit deeper and this appears to be a proper freedesktop specification now  that should be supported in Gnome 42 and ElementaryOS 6.1 (probably KDE Plasma in the future as well). This link has pretty much all the details: https://blogs.gnome.org/alexm/2021/10/04/dark-style-preference/

andydotxyz:
That blog post is helpful, but I still cannot find any documentation that is not in a gnome blog post or wiki ;)

Jacalz:
True. That part is a bit weird but I think the above link should have most of what is needed to implement it. I did found out about this from the elementary OS 6.1 release notes though: https://blog.elementary.io/elementary-os-6-1-available-now/.

andydotxyz:
Don't they get it for free from using the latest Gnome libs? i.e. it may still be proposed at the Gnome side rather than ratified standard...

Jacalz:
Maybe. Seems like they really want it to be a standard though. Maybe the standard just hasn't been published yet? Seems like they have good intentions for making it a standard either way. We are probably closer now than we ever have been with this.

metal3d:
I made an implementation here: https://github.com/fyne-io/fyne-x/pull/48

Jacalz:
@metal3d Would you care to bring the functionality specific to this issue into the main code base here? We already follow light/dark themes on Windows/macOS and it would be great if we do so on Linux as well :)

metal3d:
Yes, I can propose the theme here.

Of course, as you see, I found ways to follow dark/light theme on Linux, but I think there is a few of decision to take on what I did.
But we can discuss on the PR.

Let me a few minute to make the PR 😄

metal3d:
[...]

PR #3414, opened 2022-11-19
Detect theme variant for Gnome and KDE

metal3d:
### Description:

This detects theme variant for Gnome and KDE.

- For Gnome, read values from `gsettings` to get the color-scheme (can be prefer-dark, prefer-light or "default"
- For KDE, read the kdeglobals file and calculate the brightness if the theme is not the standard one - this method can also be applied for older Gnome version reading GTK themes, but is this useful?
- Theme change detection: - Gnome: connect to dbus and wait for theme change - KDE, no idea at this time, but switching from KDE to Gnome has changed the Gnome theme... So maybe KDE is using dbus too.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -53,14 +53,14 @@ func findThemeVariant() fyne.ThemeVariant {
 	case "gnome", "xfce", "unity", "gnome-shell", "gnome-classic", "mate", "gnome-mate":
 		return findGnomeThemeVariant()
 	case "kde", "kde-plasma", "plasma":
-		return findKdeThemeVariant()
+		return findKDEThemeVariant()
 	default:
 		return theme.VariantDark
 	}
 }
 
 // find the current KDE theme variant. At this time, no solution.
-func findKdeThemeVariant() fyne.ThemeVariant {
+func findKDEThemeVariant() fyne.ThemeVariant {
 	homedir, err := os.UserHomeDir()
 	if err != nil || homedir == "" {
 		// there is a problem, fallback to dark theme
@@ -237,6 +237,10 @@ func watchTheme() {
 	}
 }
 
+func themeChanged() {
+	fyne.CurrentApp().Settings().(*settings).setupTheme()
+}
+
 func watchGnomeTheme() {
 	// connect to dbus to detect color-schem theme changes
 	conn, err := dbus.SessionBus()
@@ -256,20 +260,10 @@ func watchGnomeTheme() {
 	defer conn.Close()
 	dbusChan := make(chan *dbus.Signal, 10)
 	conn.Signal(dbusChan)
-	currentTheme := fyne.CurrentApp().Settings().Theme()
 	for sig := range dbusChan {
 		for _, v := range sig.Body {
-			switch v {
-			case "color-scheme":
-				if currentTheme == theme.DefaultTheme() {
-					variant := findGnomeThemeVariant()
-					switch variant {
-					case theme.VariantLight:
-						fyne.CurrentApp().Settings().SetTheme(theme.LightTheme())
-					case theme.VariantDark:
-						fyne.CurrentApp().Settings().SetTheme(theme.DarkTheme())
-					}
-				}
+			if v == "color-scheme" {
+				themeChanged()
 			}
 		}
 	}
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -9,9 +9,14 @@
 package app
 
 import (
+	"io/ioutil"
+	"log"
 	"net/url"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/godbus/dbus/v5"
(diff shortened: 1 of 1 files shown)


Jacalz:
Would it, by any chance, be possible to stick to only the BDus values mentioned in https://github.com/fyne-io/fyne/issues/2657?

It seems less elegant and more prone to error if we should parse files specific to KDE or Gnome. The issue in question mentions a new standard and I think it would be wise to instead only rely on that. This way we reduce the risk of weird issues if the structures of the files we are parsing change in the future

Bluebugs:
> Would it, by any chance, be possible to stick to only the BDus values mentioned in https://github.com/fyne-io/fyne/issues/2657?

I would think that this should be the first method tried. As a fallback and due to the large number of old os supported by fyne, we need to still try the files. 

> It seems less elegant and more prone to error if we should parse files specific to KDE or Gnome. The issue in question mentions a new standard and I think it would be wise to instead only rely on that. This way we reduce the risk of weird issues if the structures of the files we are parsing change in the future

The file structure shouldn't massively change as this could also a problem for the application that exist today and use this file.

metal3d:
I can avoid to use `gsettings` for Gnome and read the value from DBus. It's a complex code, but it gives the `unit8` value of the variant.
KDE doesn't use the DBus property for now.

There are 2 possibilities:
- I only change the call to `gsettings` and we accept, at this time, to read the config file of KDE for it
- I remove everything to only use the DBus value, but this will not work on KDE for a certain time

As @Bluebugs mentioned, KDE didn't change the structure of the file, and will not for a while.

Using only DBus drastically reduces the code size. But KDE will not be supported. And AFAIK, it could be for a long time... KDE is not famous to quickly follow the standards.

BTW: KDE is messy, and it changes the Gnome/GTK theme each time I use it – it does this to ensure that GTK apps will match the Breeze theme (or whatever the theme you choose on KDE preferences). That means that each time I switch my session, and I'm back to Gnome, it changed the icon theme, the color theme, add window buttons, and many changes... It's exhausting, annoying, clearly not a behavior I appreciate.
I personally don't use it, I installed KDE to make tests, that's a regret now 😄 

So, my opinion now: I personally prefer to use DBus with standard freedesktop values, because it's standard... Now, what are the Fyne rules? Do we accept "default dark theme" for KDE and wait them to respect the standards?

I don't have the power to decide 😉

app/app_xdg.go
```diff
@@ -32,6 +37,107 @@ func (a *fyneApp) OpenURL(url *url.URL) error {
...
+		return findGnomeThemeVariant()
+	case "kde", "kde-plasma", "plasma":
+		return findKdeThemeVariant()
+	default:
+		return theme.VariantDark
+	}
+}
+
+// find the current KDE theme variant. At this time, no solution.
+func findKdeThemeVariant() fyne.ThemeVariant {
```
andydotxyz, line 63: I think the name should be `findKDEThemeVariant` with casing?

app/app_xdg.go
```diff
@@ -122,5 +228,49 @@ func rootCacheDir() string {
...
+	currentTheme := fyne.CurrentApp().Settings().Theme()
+	for sig := range dbusChan {
+		for _, v := range sig.Body {
+			switch v {
+			case "color-scheme":
+				if currentTheme == theme.DefaultTheme() {
+					variant := findGnomeThemeVariant()
+					switch variant {
+					case theme.VariantLight:
+						fyne.CurrentApp().Settings().SetTheme(theme.LightTheme())
```
andydotxyz, line 268: This (and the line 2 down) is wrong - it will override developer set themes.
Try to copy the darwin code (`app_desktop_darwin.go`) and use the call to `setupTheme()` instead.

andydotxyz (changes requested):
This is excellent, thanks so much.
The `SetTheme` needs to be changed though, as it will override developer settings.

andydotxyz:
This is so cool, thanks :)

> So, my opinion now: I personally prefer to use DBus with standard freedesktop values, because it's standard... Now, what are the Fyne rules? Do we accept "default dark theme" for KDE and wait them to respect the standards?

This would be my preference - but as you say it will not work for KDE for some time... so maybe the KDE code needs to remain.
For Gnome I think the DBus should be OK - this is a new setting if I understand correctly so older non-DBus versions won't have it anyway will they?

app/app_xdg.go
```diff
@@ -32,6 +37,107 @@ func (a *fyneApp) OpenURL(url *url.URL) error {
...
+			}
+		}
+	}
+
+	// If we reach this point, it means that the LookAndFeelPackage key is not present in the kdeglobals file
+	// we can try to calculate the theme variant from the current KDE theme using the WM activeBackground key
+	for _, line := range lines {
+		if strings.HasPrefix(line, "activeBackground=") {
+			// read the value, it's a color in the form of r,g,b
+			col := strings.Split(line, "=")[1]
```
andydotxyz, line 102: As per your fyne-x bigger PR it would be good if this colour parsing was in a separate function.
We have the approach that a comment inline often shows that the code should have been in another function with appropriate name. Code is more resilient than docs ;)

metal3d:
> This would be my preference - but as you say it will not work for KDE for some time... so maybe the KDE code needs to remain.

So, for now, I will leave the code like this (with changed you asked)

>For Gnome I think the DBus should be OK - this is a new setting if I understand correctly so older non-DBus versions won't have it anyway will they?

Actually, I made this detection very recently. Months ago, I prepared themes for Gnome variants and KDE/Plasma:
- Gnome, I was using `gjs` to get colors (and icons), and I made brightness computation
- KDE, mainly the same as in this PR, reading file and calculate the brightness.

I decided to not reimplement the Gnome theme reading with `gjs` because of the complexity (it calls javascripts, following the current theme for detected version of target GTK version... see https://github.com/fyne-io/fyne-x/pull/48 which is now a draft)

I really don't know if I should take some parts of what I did in fyne-x, or if I should use DBus value only for Gnome and remove the KDE specific case later.

In any case, I've got solutions. The hard part is to decide.

metal3d:
To complete the answer, I can do (for Gnome):

*Solution 1:*
Only use DBUS

*Solution 2:*
Try DBUs, fallback to `gjs` call and calculate the brightness

*Solution 3:*
Only make `gjs` call (I don't like)

app/app_xdg.go
metal3d: fixed

app/app_xdg.go
metal3d: I made variation, the darwin code calls C implementaition.

app/app_xdg.go
metal3d: The newest code is pushed, is it OK for you?

Jacalz:
I am quite certain that I have read somewhere that KDE Plasma actually support the dark style preference. A quick Google search did also seem to indicate that it is the case: https://www.reddit.com/r/kde/comments/rwizt0/freedesktop_dark_style_preference_for_kde_apps/

app/app_xdg.go
andydotxyz: Ah yes indeed, good implementation for pure Go version

metal3d:
> I am quite certain that I have read somewhere that KDE Plasma actually support the dark style preference. A quick Google search did also seem to indicate that it is the case: https://www.reddit.com/r/kde/comments/rwizt0/freedesktop_dark_style_preference_for_kde_apps/

I made many tests to check – when I select the light/dark theme in KDE / Plasma on Fedora 36, it doesn't change anything on the DBus property. It changes the kdeglobals file, it breaks Gnome theme by forcing GTK and icon theme, but the DBus property is still set to the same value.

Maybe it changed in Fedora 37 (I didn't upgrade yet), but at this time, it's not the case.

So, I wonder what to do.

If I only use this DBus variable, newest KDE will be OK and Gnome will be OK. But, previous KDE version will not respond.

Because it's only to detect the variant, that is not a breaking feature, and because the color-scheme variable in DBus is the "correct" way to decide... Should we not, finally, leave the previous behavior for desktop manager that are not yet ready to use it? And so, leave the previous behavior (dark theme by default) for them?

I mean, the decision could be:

- DBus color scheme found:
   - OK, set the color variant to the value found
   - watch the variant change in DBus to react on changes
- else
   - set color variant to dark
   - no watch

This will simplify the source code too.

Jacalz:
I am personally leaning more to what you are describing above. I’d be interested in what the others think of it

metal3d:
Actually, if we change the behavior by what I described, the functions should be renamed to "XXXFreedesktop" - that will not be specific to Gnome or KDE 😄

Bluebugs:
Sounds like a good plan to me too.

metal3d:
I will change source code in a few minutes/hours, if it's not OK for @andydotxyz I can revert.

andydotxyz:
> I will change source code in a few minutes/hours, if it's not OK for @andydotxyz I can revert.

I am happy to go with the consensus here - we can always iterate as we get user feedback.

Jacalz (changes requested):
Thanks. I annoyed a few minor things that can be tidied up but it is getting very close

app/app_xdg.go
```diff
@@ -32,6 +33,43 @@ func (a *fyneApp) OpenURL(url *url.URL) error {
...
+	if err = call.Store(&value); err != nil {
+		log.Println("failed to read theme variant from D-Bus", err)
+		return theme.VariantDark
+	}
+
+	switch value {
+	case 0:
+		return theme.VariantLight
+	case 1:
+		return theme.VariantDark
```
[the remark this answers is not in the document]
metal3d: Right, I wondered if that was interesting for the reading. I can change.

app/app_xdg.go
```diff
@@ -122,5 +159,39 @@ func rootCacheDir() string {
...
+	}
+	defer conn.Close()
+
+	dbusChan := make(chan *dbus.Signal, 10)
+	conn.Signal(dbusChan)
+
+	for sig := range dbusChan {
+		for _, v := range sig.Body {
+			if v == "color-scheme" {
+				themeChanged()
```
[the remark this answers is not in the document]
metal3d: Ho, `break` sorry, changed!

Jacalz (changes requested):
I just tested this on Fedora 36 with Gnome 42 and it does not seem to work. Nothing happens when I change between light and dark mode. The app just always use the dark mode no matter what preference I set in the settings

app/app_xdg.go
```diff
@@ -32,6 +33,41 @@ func (a *fyneApp) OpenURL(url *url.URL) error {
...
+		"color-scheme",
+	)
+	if call.Err != nil {
+		log.Println("failed to read theme variant from D-Bus", call.Err)
+		return theme.VariantDark
+	}
+
+	var value uint8
+	if err = call.Store(&value); err != nil {
+		log.Println("failed to read theme variant from D-Bus", err)
```
Jacalz, line 58:
```suggestion
		fyne.LogError("Failed to read theme variant from D-Bus", err)
```

app/app_xdg.go
```diff
@@ -122,5 +157,40 @@ func rootCacheDir() string {
...
+		log.Println("Unable to connect to session D-Bus", err)
+		return
+	}
+
+	if err := conn.AddMatchSignal(
+		dbus.WithMatchObjectPath("/org/freedesktop/portal/desktop"),
+		dbus.WithMatchInterface("org.freedesktop.portal.Settings"),
+		dbus.WithMatchMember("SettingChanged"),
+	); err != nil {
+		log.Println(err)
```
Jacalz, line 180: Let's use `fyne.LogError` here as well.

app/app_xdg.go
```diff
@@ -122,5 +157,40 @@ func rootCacheDir() string {
...
+
+func themeChanged() {
+	fyne.CurrentApp().Settings().(*settings).setupTheme()
+}
+
+// connect to dbus to detect color-schem theme changes in portal settings.
+func watchFreedekstopThemeChange() {
+	conn, err := dbus.SessionBus()
+	if err != nil {
+		log.Println("Unable to connect to session D-Bus", err)
```
Jacalz, line 171: Let's use `fyne.LogError` here as well.

app/app_xdg.go
```diff
@@ -122,5 +157,40 @@ func rootCacheDir() string {
...
+		dbus.WithMatchObjectPath("/org/freedesktop/portal/desktop"),
+		dbus.WithMatchInterface("org.freedesktop.portal.Settings"),
+		dbus.WithMatchMember("SettingChanged"),
+	); err != nil {
+		log.Println(err)
+		return
+	}
+	defer conn.Close()
+
+	dbusChan := make(chan *dbus.Signal, 10)
```
Jacalz, line 185: Is there any specific reason that this has a length 10 buffer? Do we need as much?

Jacalz (changes requested):
I now retested after having upgraded to Fedora 37 and Gnome 43. It works flawlessly. I don't know why I did not see it working before. Nice work :slightly_smiling_face: 
Just found a few last notes. Should be very close to landing :+1:

app/app_xdg.go
metal3d: Changed, it seems to work actually with chan size to 0

Jacalz (approved):
Works great. Thanks for working on this. It will make Fyne applications work so much better on Linux :)

Jacalz:
@andydotxyz Did we decide that it was fine to merge this even though we are in a feature freeze? I have a vague recollection of talking about that but I can't remember.

andydotxyz:
Yes I think it's good - "bug fixes" still allowed and this is technically one. Though an epic one 😀 .
Once we get the blockers closes it will be code freeze and no more new fixes unless blocking.

Jacalz:
Great. You still have required changes. Feel free to review and merge if you feel that it is ready :)

andydotxyz (approved):
Great thanks

PR #3415, opened 2022-11-19
Completing the new theme

### LINKED ISSUE #3042: Add new theme updates

andydotxyz:
I updated the contrast as best I could whilst sticking to the design and the way selection colour works.
If we want to revisit it further I will have to go re-visit the designs with our designer.

Fixes #3042 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz:
Thanks for this. The new text selection colour for the dark looks a lot better but the light theme still seems quite similar to what it was before?
image

andydotxyz:
> Thanks for this. The new text selection colour for the dark looks a lot better but the light theme still seems quite similar to what it was before? image

I don't know if it's the monitor but those colours look quite different to me?

Jacalz:
Indeed. That is because the the left one is the mock-up image for the new theme (taken form the issue mentioned above)

andydotxyz:
Oh. I misunderstood your comparison.

If I make the colour closer to the light theme look then it generates a small difference on the dark theme that was originally noted for change. It is a part transparent overlay using the same colour for both variants.

From the designs we have two different opacities for selection on light and dark.
I can move to multiple colours for different variants but if there is nothing wrong with the contrast I am tempted to use the colour we have.

Thoughts?

Jacalz (approved):
Makes sense. It's better than before :)

PR #3424, opened 2022-11-24
Update README.md with proper go install invocations

paboum:
### Description:

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,8 +178,8 @@ It is recommended that you install the following additional apps:
 
 | app | go install | description |
 | --- | ------ | ----------- |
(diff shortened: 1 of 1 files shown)


andydotxyz (commented):
Unfortunately "@latest" only works for go since 1.16, but we support back to 1.14 so we tend not to specify version unless providing an alternative syntax for older installs...

andydotxyz:
I didn't mean this needed to be closed - there was more to this than just changing to "@latest" wasn't there?

PR #3431, opened 2022-11-26
Fix crazy peak in memory when moving selection

### LINKED ISSUE #3426: Memory usage surge when selecting/appending MultilineEntry text
### Checklist

### Describe the bug

My fyne App is just a node.exe launcher. It launcher node.exe and capture its output and write to a text entry.
Memory usage surge when selecting/appending MultilineEntry text as the following gif shows.
Memory usage keeps growing when the output of the node script is captured to the text entry. After about an hour of running, the memory usage peak at an astonishing 12G.

### How to reproduce

1. Create a multiline entry with a lot text content
2. Keep Selecting text in the entry and monitor the memory usage of the app

### Screenshots

20221124151652_rec_

image

### Example code

```go
	text := widget.NewMultiLineEntry()
	text.SetText("Press button 'Run' to Start")
	text.SetMinRowsVisible(16)

       var c *cmd.Cmd = cmd.NewCmdOptions(cmdOptions,  'node', newArgs...)
```

```go
		go func() {
			complete := c.Status().Complete
			for !complete && c.Stdout != nil || c.Stderr != nil {
				complete = c.Status().Complete
				if !complete {
					if !runButton.Disabled() {
						runButton.Disable()
					}
					if stopButton.Disabled() {
						stopButton.Enable()
					}
				}
				select {
				case line, open := <-c.Stdout:
					// OK
					if !open {
						c.Stdout = nil
						continue
					}
					log.Println(line)
					text.SetText(text.Text + "\n" + line)
				case line, open := <-c.Stderr:
					// Error
					if !open {
						c.Stderr = nil
						continue
					}
					fmt.Fprintln(os.Stderr, line)
					text.SetText(text.Text + "\n" + line)
				}
			}
			stopButton.Disable()
			runButton.Enable()
		}()
		c.Start()
		a.Lifecycle().SetOnStopped(func() {
			log.Println("Stopped")
			c.Stop()
		})
```

### Fyne version

2.2.3

### Go compiler version

1.17

### Operating system

macOS

### Operating system version

12.5.1

### Additional Information

_No response_

andydotxyz:
There is an underlying issue where text refresh memory is not freed

but here we are able to avoid refresh which is always good.
Fixes #3426

### Checklist:

- [ ] Tests included. <- an optimisation
- [x] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -1158,17 +1158,23 @@ func (e *Entry) truncatePosition(row, col int) (int, int) {
 
 func (e *Entry) updateMousePointer(p fyne.Position, rightClick bool) {
 	row, col := e.getRowCol(p)
-	e.setFieldsAndRefresh(func() {
-		if !rightClick || rightClick && !e.selecting {
-			e.CursorRow = row
-			e.CursorColumn = col
-		}
+	e.propertyLock.Lock()
 
-		if !e.selecting {
-			e.selectRow = row
-			e.selectColumn = col
-		}
-	})
+	if !rightClick || rightClick && !e.selecting {
+		e.CursorRow = row
+		e.CursorColumn = col
+	}
+
+	if !e.selecting {
+		e.selectRow = row
+		e.selectColumn = col
+	}
+	e.propertyLock.Unlock()
+
+	r := cache.Renderer(e.content)
+	if r != nil {
+		r.(*entryContentRenderer).moveCursor()
+	}
 }
 
 // updateText updates the internal text to the given value
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -1160,7 +1160,7 @@ func (e *Entry) updateMousePointer(p fyne.Position, rightClick bool) {
 	row, col := e.getRowCol(p)
 	e.propertyLock.Lock()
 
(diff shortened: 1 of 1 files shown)


widget/entry.go
```diff
@@ -1158,17 +1158,23 @@ func (e *Entry) truncatePosition(row, col int) (int, int) {
...
-			e.CursorColumn = col
-		}
+	e.propertyLock.Lock()
 
-		if !e.selecting {
-			e.selectRow = row
-			e.selectColumn = col
-		}
-	})
+	if !rightClick || rightClick && !e.selecting {
```
Jacalz, line 1163: You should be able to simplify the Boolean logic slightly:
```suggestion
	if !rightClick || !e.selecting {
```
As you can see, the truth tables are the same:
image
image

Jacalz (changes requested):
Nice work. Just noted a small simplification to the Boolean logic :)

widget/entry.go
andydotxyz: Good point. Missed that as it was just an indent change. fixing

Jacalz (approved):
Works well. I do not see any drastic increase when moving the cursor :+1:

PR #3440, opened 2022-11-29
Update workflows

Jacalz:
### Description:

This brings in the latest fixes and improvements to our workflows.
- We now use Go 1.19.x instead of 1.17.x for running all tests (except those on Go 1.14.x that test the old support). The formatting with Go 1.19 is a lot stricter due to it now supporting markdown (mostly applies to the very large comments so should generally not be an issue). I added in all the changes that `goimports` was complaining about.
- It also applies a suggestion that I got about getting rid of the "too menu arguments" error when `goimports` fails (tested it and it works well).
- I also noticed that `staticcheck was a few versions behind. 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/repository/http.go
+++ b/internal/repository/http.go
@@ -70,7 +70,7 @@ func constructURI(u fyne.URI) string {
 // Exists checks whether the the resource at u returns a
 // non "404 NOT FOUND" response header.
 //
-// # The method is a part of the implementation for repository.Repository.Exists
+// Implements: repository.Repository
 //
 // Since: 2.1
 func (r *HTTPRepository) Exists(u fyne.URI) (bool, error) {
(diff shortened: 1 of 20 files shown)


Bluebugs (approved):

internal/repository/http.go
```diff
@@ -89,7 +89,7 @@ func (r *HTTPRepository) Exists(u fyne.URI) (bool, error) {
 // Reader provides a interface for reading the body of the response received
 // from the request to u.
 //
-// The method is a part of the implementation for repository.Repository.Exists
```
andydotxyz, line 92: This doesn't seem right - these items are not headings.
Occurs in a couple of places - just noting it here

andydotxyz (changes requested):
Nice update.
A few comments of medium size moved 1 liner text to headings, can they just be sentences instead?

internal/repository/http.go
Jacalz: Good point. I sorted out the issue and also fixed a comment that accidentally said "Sine:" instead of "Since:" (we don't want it to periodically be in releases, hehe).

andydotxyz (approved):
Nice thanks

Jacalz:
I think you need to modify allowed workflows. I can't merge this :/

andydotxyz:
Sorted :) If we can remove the Go version number from the workflow title this will stop being an additional step.

PR #3448, opened 2022-12-01
.github/workflows: Make sure to run all Go 1.19 tests

Jacalz:
This makes sure to enable all tests to run using Go 1.19. Apparently they are checking for only Go 1.17 which is a not so elegant solution. This is a quick-fix to make sure that everything runs as expected.

 

### Description:

This makes sure to enable all tests to run using Go 1.19. Apparently they are checking for only Go 1.17 which is a not so elegant solution. This is a quick-fix to make sure that everything runs as expected.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/.github/workflows/platform_tests.yml
+++ b/.github/workflows/platform_tests.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.14, 1.19]
+        go-version: [1.14, 1.17]
         os: [ubuntu-latest, windows-latest, macos-latest]
 
     steps:
@@ -44,7 +44,7 @@ jobs:
       uses: shogo82148/actions-goveralls@v1
       with:
         path-to-profile: coverage.out
-      if: ${{ runner.os == 'Linux' && matrix.go-version == '1.19' }}
+      if: ${{ runner.os == 'Linux' && matrix.go-version == '1.17' }}
 
     - name: Build WebAssembly binary
       env:
@@ -52,11 +52,11 @@ jobs:
         GOARCH: wasm
       working-directory: cmd/fyne_demo
       run: go build
-      if: ${{ matrix.go-version == '1.19' }}
+      if: ${{ matrix.go-version == '1.17' }}
 
     - name: Build GopherJS and Wasm full website
       run: |
         go install github.com/gopherjs/gopherjs@latest
         go install ./cmd/fyne
         cd cmd/fyne_demo && fyne package --target=web
-      if: ${{ matrix.go-version == '1.19' && runner.os == 'Linux' }}
+      if: ${{ matrix.go-version == '1.17' && runner.os == 'Linux' }}
--- a/.github/workflows/platform_tests.yml
+++ b/.github/workflows/platform_tests.yml
@@ -44,7 +44,7 @@ jobs:
       uses: shogo82148/actions-goveralls@v1
       with:
         path-to-profile: coverage.out
-      if: ${{ runner.os == 'Linux' && matrix.go-version == '1.17' }}
+      if: ${{ runner.os == 'Linux' && matrix.go-version == '1.19' }}
 
     - name: Build WebAssembly binary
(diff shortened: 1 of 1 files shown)


andydotxyz (approved):

andydotxyz (changes requested):
I think GooherJS won't work with 1.19

andydotxyz:
Looks like the correct fix is actually yo put 1.17 back for the web build/test

fuskiid:
> I think GooherJS won't work with 1.19

GopherJs is still pending [1.19 suppport](https://github.com/gopherjs/gopherjs/issues/1140) 

which itself is on hold until [1.18 Generic support]( https://github.com/gopherjs/gopherjs/issues/1013)

Jacalz:
Yeah. I'll get this sorted later today

andydotxyz (approved):

.github/workflows/platform_tests.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.14, 1.19]
+        go-version: [1.14, 1.17]
```
Bluebugs, line 12: Can't we have 1.14, 1.17 and 1.19? As the gopherjs test is only enable for 1.17, that would work, no?

andydotxyz: That is adding a lot of time to the compiling.
Really I think we should use 19 for most, and add another test run for the web build which can use 1.17

.github/workflows/platform_tests.yml
```diff
@@ -44,19 +44,19 @@ jobs:
...
+      if: ${{ runner.os == 'Linux' && matrix.go-version == '1.17' }}
 
     - name: Build WebAssembly binary
       env:
         GOOS: js
         GOARCH: wasm
       working-directory: cmd/fyne_demo
       run: go build
-      if: ${{ matrix.go-version == '1.19' }}
+      if: ${{ matrix.go-version == '1.17' }}
```
Bluebugs, line 55: I would really prefer if wasm was checked against 1.19 as the idea is that wasm come out of latest go and that work, while gopherjs require an older go as backup solution for problematic browser.

.github/workflows/platform_tests.yml
Bluebugs: That make sense indeed.

Jacalz: We had more than two Go versions earlier but it gets very slow in the long run. I prefer to keep it as it is

.github/workflows/platform_tests.yml
Jacalz: As stated in my previous comment, I prefer to only have two Go versions at the time. We will slow down our CI times a lot otherwise. Lets just keep this PR simple and fix the issue to make it behave as before

.github/workflows/platform_tests.yml
Jacalz: Yeah, I agree with @andydotxyz. That seems like a good idea. Will add that to my list

Jacalz: You are also free to tackle that @Bluebugs if you are interested :)

nevkontakte:
A drive-by comment on behalf of the GopherJS project. Go 1.19 support is on hold merely because I (personally) don't have time to work on both Go 1.19 support and generics at the same time, and I feel like working on generics is more important. However, if folks are interested in making Go 1.19 support appearing sooner, this would be a welcome contribution :) I even [wrote a guide](https://docs.google.com/document/d/1QD_jl9G1jwL2ye_ZeFMtqIEEYZ9cNIHbV3n3R4pXdYA/edit) to make this easier for the newer contributors.

Cheers!

andydotxyz:
Great pointer @nevkontakte and thanks so much for the hard work, we appreciate it.
Hopefully once we have our current release done someone can spare a little time to come and help.

PR #3449, opened 2022-12-02
Fix rendering of compound glyphs

### LINKED ISSUE #598: Displaying unicode or different language like Bengali  doesn't work.
I'm trying to make an application using Fyne, where I need to display Bengali. But didn't find a way to display it. Is there a way to display Unicode/different language?

andydotxyz:
Move from iterating on rune to glyphID.
This is not a full i18n solution as we still have to handle the font being to latin-alphabet-base, but it resolves the issues when a font is provided...

Fixes #598

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
Screenshot 2022-12-02 at 14 25 35

The above test could be added to the code, but it would mean bundling another font into the code so I did not do it as part of this PR. Though as we do more in i18n we may need/want to anyway

Bluebugs (approved):

Bluebugs (approved):

PR #3450, opened 2022-12-02
Make sure that Resize call with a canvas Scale() == 1 express its pre…

### LINKED ISSUE #3402: Initial text rendering is blurry on develop
### Checklist

### Describe the bug

I noticed when testing `cmd/hello ` for #3401 that the initial text that is rendered on start-up seems to be blurry but resizing or updating the window in some other way, such as pressing the button, makes it redraw without being blurry.

### How to reproduce

1. Compile and run two versions of `cmd/hello` at the same time.
2. Resize, or press the button, in one of the windows.
3. Notice that the text initially is blurry but then gets sharper (and slightly brighter) after the window is updated.

### Initial draw:
Screenshot from 2022-11-13 19-30-52

### Window resized:
image

### Button pressed:
image

### Example code

Just use `cmd/hello` for tests.

### Fyne version

4c67d93aa35be8646ba9c4354acb0492d5115f4d

### Go compiler version

1.18.7

### Operating system

Linux

### Operating system version

Fedora 36

### Additional Information

_No response_

Bluebugs:
### Description:
The issue for the blurry first frame is actually not related to text, but to an imprecision in the canvas size that lead to a blurry output more visible by text.

Fixes #3402 

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/internal/driver/glfw/canvas.go
+++ b/internal/driver/glfw/canvas.go
@@ -105,6 +105,9 @@ func (c *glCanvas) PixelCoordinateForPosition(pos fyne.Position) (int, int) {
 }
 
 func (c *glCanvas) Resize(size fyne.Size) {
+	// This might not be the ideal solution, but it effectively avoid the first frame to be blurry due to the
+	// rounding of the size to the loower integer when scale == 1. It does not affect the other cases as far as we tested.
+	// This can easily be seen with fyne/cmd/hello and a scale == 1 as the text will happear blurry without the following line.
 	nearestSize := fyne.NewSize(float32(math.Ceil(float64(size.Width))), float32(math.Ceil(float64(size.Height))))
 
 	c.Lock()
--- a/internal/driver/glfw/canvas.go
+++ b/internal/driver/glfw/canvas.go
@@ -105,8 +105,9 @@ func (c *glCanvas) PixelCoordinateForPosition(pos fyne.Position) (int, int) {
 }
 
 func (c *glCanvas) Resize(size fyne.Size) {
+	pixelsSize := fyne.NewSize(float32(math.Ceil(float64(size.Width))), float32(math.Ceil(float64(size.Height))))
 	c.Lock()
-	c.size = size
+	c.size = pixelsSize
 	c.Unlock()
 
 	for _, overlay := range c.Overlays().List() {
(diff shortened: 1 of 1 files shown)


internal/driver/glfw/canvas.go
```diff
@@ -105,8 +105,9 @@ func (c *glCanvas) PixelCoordinateForPosition(pos fyne.Position) (int, int) {
 }
 
 func (c *glCanvas) Resize(size fyne.Size) {
+	pixelsSize := fyne.NewSize(float32(math.Ceil(float64(size.Width))), float32(math.Ceil(float64(size.Height))))
```
andydotxyz, line 108: I think this is more likely `nearestPixelSize` or something shorter... as it's not really a pixel size.

However this maths does not result in that either - it is currently getting a pixel at the coursest level and multiplying by scale, in effect. So on a HighDPI with fractional content it may be too wide/high.
Consider checking out `roundToPixel` in `internal/painter/gl/draw.go` as this will return the closest pixel. Though in this instance we probably want `Ceil` not `Round` as the canvas must always be at least the minimum size, not rounded down.

andydotxyz (changes requested):
Glad you have found a fix, not sure it's quite right though - the precision will be lost on higher DPI or FYNE_SCALE values.

internal/driver/glfw/canvas.go
Bluebugs: Wouldn't it be logical that all rounding to pixels logic use the same logic (Ceil or Round) throughout the code?

andydotxyz: It depends on the intent of the code.
To find the closest pixel you want to round to closest. To get the size that is required to fit the content you must round up as the canvas/window should never be smaller than required

Bluebugs: I went with nearestSize. I tried to follow all the callers and callees path to see if there was a better location, but the canvas size really need to be adjusted as it is accessed in a few place from the painter directly not being passed around. Also their is multiple callers that have different way of coming at the number and again the call to Resize is where everyone gather, so at this point, I think that's the best place for it.

andydotxyz: I think this isn't going to be the right value though. For lowDPI (1dp=1px) it will work cool, but what about if we're on super-high DPI? Like iOS:

25dp at 4* DPI is 100px, then if the value was 24.6 then this would return 25 - the 100px equivalent, but it should be 24.5, which is 98px. Meaning that the canvas will be 2px bigger than the window.

andydotxyz: I guess in the other direction where the content is too small may be worse as we'll get a 2px stripe at the bottom of the window.

andydotxyz: It sounds like you're confusing canvas and window size - this fix is for canvas size only, so I'm not sure I follow the logic.

Windows move in pixel size increments, so surely we want the canvas to match that?

andydotxyz (approved):

typenoob:
This doesn't look like the best approach though, as you commented.
And this causes an accuracy problem #5625, when setting a custom display scaling factor.

PR #3460, opened 2022-12-05
Make sure entry border rounds down to nearest pixel on trailing edge.

### LINKED ISSUE #3379: Entry text is missaligned on develop (to be v2.3.0)
### Checklist

### Describe the bug

The old theme had a line at the bottom of the entry and I think the new theme still tries to take the line offset into consideration. Basically the text seems to be off from center (slightly above).

I first noticed this when having two versions of Rymdport next to each other:
image

Can also be seen in fyne_demo by noticing that the text is slightly higher than centered compared to the I-beam.
[Screencast from 05-11-22 23:06:08.webm](https://user-images.githubusercontent.com/25466657/200143104-410c4b99-ae29-42c9-b12a-327708b2fc61.webm)

### How to reproduce

1. Open fyne_demo on develop and one on v2.2.x.
2. Notice that text is offset incorrectly on develop.

### Screenshots

_No response_

### Example code

See fyne_demo code.

### Fyne version

3ac369895b751a2a35667322252e35679cc78539

### Go compiler version

1.18.7

### Operating system

Linux

### Operating system version

Fedora 36

### Additional Information

_No response_

andydotxyz:
Fixes  #3379

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (approved):
This seems a bit like a scary hack and I would have preferred a more elegant solution but it fixes the issue alright

andydotxyz:
> This seems a bit like a scary hack and I would have preferred a more elegant solution but it fixes the issue alright

I agree with this sentiment, but unless we move the rectangle rendering to draw inside instead of on the edge then it's going to be needed :(

PR #3467, opened 2022-12-11
Fix blocking theme issues

### LINKED ISSUE #3462: New 2.3.0 theme suggestions
### Checklist

There are some little things that I think should be improved in the new theme:

Check mark should be white
image

Menu items and buttons should have a little more right-padding to by symmetrical
image
image

_No response_

image

andydotxyz:
Menu padding/align and contrast for checks

Fixes #3462

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/progressbar.go
+++ b/widget/progressbar.go
@@ -65,7 +65,7 @@ func (p *progressRenderer) Layout(size fyne.Size) {
 func (p *progressRenderer) applyTheme() {
 	p.background.FillColor = progressBackgroundColor()
 	p.bar.FillColor = theme.PrimaryColor()
-	p.label.Color = theme.ForegroundColor()
+	p.label.Color = theme.BackgroundColor()
 	p.label.TextSize = theme.TextSize()
 }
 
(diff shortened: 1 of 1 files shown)


Jacalz (changes requested):
A few tests are failing but it seems good otherwise. I also noticed that the progress bar text is a little hard to read in the light theme. I think it might have the wrong color:
image

andydotxyz:
> A few tests are failing but it seems good otherwise. I also noticed that the progress bar text is a little hard to read in the light theme. I think it might have the wrong color: image

Thanks, updated to match the fix of Check :)

Jacalz:
Hmm. The progress bar seems a bit problematic. It is better in some places now and worse in others:
image
image
image
image

Jacalz:
I think the solution would be to always keep the progress bar text white. Seems to give the best contrast in my opinion.

andydotxyz:
> I think the solution would be to always keep the progress bar text white. Seems to give the best contrast in my opinion.

Unfortunately that approach codes something theme specific into the progress bar, so could break with other themes, or would require new APIs in the theme to return a colour that works as you describe.
Given we can't do either of those at this stage I think the options are to go as-is or to draw a background behind the text.
What I would like to do is render the text on background as foreground colour, and that over the primary section to be background colour. But at this time we don't have the ability to crop the text rendering in that way I don't think.

Jacalz (approved):
Makes sense. This is better than before at least :)

PR #3472, opened 2022-12-13
Add static flag to statically link all cgo packages with fyne

### LINKED ISSUE #3459: cmd/fyne package can't find FyneApp.toml when -src option has given
### Checklist

### Describe the bug

I am trying to build project with fyne cmd but I get an error:
`Missing application icon at "cmd\myapp\Icon.png"`

My project structure is like this:
```
root
--cmd
----myapp
------assets
--------icon.png
------main.go
--pkg
--FyneApp.toml
--Makefile
```

FyneApp.toml
```
Website = "https://www.example.com"

[Details]
  Icon = "cmd/myapp/assets/icon.png"
  Name = "myapp"
  ID = "com.myapp.app"
  Version = "0.1.0"
  Build = 50
```

I am running this command in the root but it doesn't work and gives the error:
`fyne package -release -os windows -src cmd/myapp`

I tried moving icon to cmd/myapp/Icon.png but it also doesn't work

Moving FyneApp.toml to cmd/myapp and running command `fyne package -release -os windows` in cmd/myapp works so I think the problem happens only with -src option

What I want is fyne should look FyneApp.toml in cwd and load it from there so I can put FyneApp.toml to root and run fyne package in root

### How to reproduce

1. Create a project with same structure like above
2. Run `fyne package -release -os windows -src cmd/myapp` in root

### Screenshots

_No response_

### Example code

_No response_

### Fyne version

2.2.3

### Go compiler version

1.19.1

### Operating system

Windows

### Operating system version

11

### Additional Information

_No response_

hismailbulut:
### Description:

New static flag will be passed to go ldflags to only statically link cgo packages
This is a very important flag because dynamically linked cgo libraries will need dll's to be distributed with binary

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.


Bluebugs:
I think we need to make this an option as otherwise building against dll will be impossible. It would require adding a flag to build, package and release and propagate the information along.

hismailbulut:
You are totally right. I did it for myself and I created pr to generate ideas. It absolutely needs more testing and most possibly fixing.

cmd/fyne/internal/commands/package.go
```diff
@@ -360,11 +360,10 @@ func (p *Packager) validate() (err error) {
 		return errors.New("parameter -sourceDir is currently not supported for mobile builds. " +
 			"Change directory to the main package and try again")
 	}
-	os.Chdir(p.srcDir)
 
 	p.appData.CustomMetadata = p.customMetadata.m
 
-	data, err := metadata.LoadStandard(p.srcDir)
+	data, err := metadata.LoadStandard(baseDir)
```
andydotxyz, line 366: I disagree with this change. The .toml file should be read from the source dir over the current directory.
If we go with basedir instead then builds can change based on current directory and the app metadata is detached from the app it describes.

andydotxyz (changes requested):

andydotxyz:
I don't think static should be the default, as @Bluebugs says. I also added another thought inline which seems counter-productive too.

cmd/fyne/internal/commands/package.go
hismailbulut: In my opinion it is better to run fyne commands where FyneApp.toml is. Most of the build systems needs configuration or script in root folder. If you are concerning about multiple main programs in one project here is my suggestion. fyne will first look at the srcdir for FyneApp.toml as you said but if it couldn't find it there, it will look for it in the cwd where fyne command runs. I still respect your decision.

hismailbulut: Also I only have access to windows so I can't make required changes for linux and especially macos. It is better to close this pr and make yourself when you have time.

hismailbulut:
@Bluebugs @andydotxyz I didn't want to ping but maybe we can make it until the new version

andydotxyz:
I'm nervous about changing the current directory handling inside this PR about static linking. I have tried to tidy up this and inevitably it ends up breaking some image lookups or other aspects that are still relative. Can it be split out so this focuses just on the build setting.

Additionally I think you missed applying this to "fyne release" command?

hismailbulut:
It may be but this pr should fix the issue. I am not sure which parts are broken because of removing chdir, because codebase is pretty big and complicated to lookup for me. I can add static flag to release but I am unable to test it because certificates etc.

hismailbulut:
If you want I can create separate issue (feature request) for static linking and you can work on them separately by ignoring this pr

andydotxyz:
I don't want to dismiss the hard work you have done here - but it would be easier to test and verify if it's focused on just the build flags at this time - that makes it about only 12 combinations (macOS, Linux, windows, BSD x build, package, release). If you add in chdir and passing in path to build instead the combinations for testing seriously explode.

Jacalz:
I agree with Andy. It is easier to review if the the PR just fixes one thing at the time.

Jacalz (commented):
I just tested this and I get some strange results. I am not sure how much the static linking actually does. Test results follow below:

## Binary sizes
I find the binary sizes here to be strange. I would have expected static linking, if it actually is working, to produce larger binaries than dynamically linking. Somehow it just seems to enable release ldflags for some reason (given the similar size).

**Dynamic:** 25.1 MB
**Dynamic, release:** 18.6MB
**Static:** 18.6MB
**Dynamic, release:** 18.6MB

## Test of linked symbols
Running `ldd` on both static and dynamic binaries results in basically the same dependencies being reported as linked (apart from the hexadecimal numbers changing). I suspect that the static linking isn't doing what it it supposed to do.
```
linux-vdso.so.1 (0x00007ffe7c5d6000)
	libGL.so.1 => /lib64/libGL.so.1 (0x00007f1c1aa82000)
	libX11.so.6 => /lib64/libX11.so.6 (0x00007f1c1a93b000)
	libXrandr.so.2 => /lib64/libXrandr.so.2 (0x00007f1c1a92e000)
	libXxf86vm.so.1 => /lib64/libXxf86vm.so.1 (0x00007f1c1a927000)
	libXi.so.6 => /lib64/libXi.so.6 (0x00007f1c1a913000)
	libXcursor.so.1 => /lib64/libXcursor.so.1 (0x00007f1c1a904000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f1c1a824000)
	libXinerama.so.1 => /lib64/libXinerama.so.1 (0x00007f1c1a81f000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f1c1a642000)
	libGLX.so.0 => /lib64/libGLX.so.0 (0x00007f1c1a610000)
	libXext.so.6 => /lib64/libXext.so.6 (0x00007f1c1a5fb000)
	libGLdispatch.so.0 => /lib64/libGLdispatch.so.0 (0x00007f1c1a542000)
	libxcb.so.1 => /lib64/libxcb.so.1 (0x00007f1c1a515000)
	libXrender.so.1 => /lib64/libXrender.so.1 (0x00007f1c1a508000)
	libXfixes.so.3 => /lib64/libXfixes.so.3 (0x00007f1c1a4ff000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f1c1ab1b000)
	libXau.so.6 => /lib64/libXau.so.6 (0x00007f1c1a4f9000)
```

These were my test results on Linux. Should I expect something different or am I just doing something wrong? What is the use case of static linking tat you are looking for?

cmd/fyne/internal/commands/build.go
```diff
@@ -277,7 +291,10 @@ func (b *Builder) build() error {
 		return err
 	}
 
-	b.runner.setDir(b.srcdir)
```
andydotxyz, line 280: I agree that abspath is a good update - but can it still be passed to `setDir` so we don't change the semantics of path lookups inside the runner?

andydotxyz (changes requested):
Thanks so much, this is way more concise

Jacalz (changes requested):
See my comment above. I am not entirely sure that static linking works at this stage.

hismailbulut:
> Running ldd on both static and dynamic binaries results in basically the same dependencies being reported as linked (apart from the hexadecimal numbers changing). I suspect that the static linking isn't doing what it it supposed to do.

Are you using clang or gcc? I think it will not work with clang because I realized llvm's linker lld's corresponding flag for `-static` is `--static`.

Jacalz:
> Are you using clang or gcc? I think it will not work with clang because I realized llvm's linker lld's corresponding flag for -static is --static.

I am using `gcc`. Clang is not installed at the moment.

hismailbulut:
I also don't quite understand actually. These are my results:
```
C:\ismailbulut\Projeler\nftsiren>ldd cmd\nftsiren\nftsiren_release.exe
        ntdll.dll => /c/WINDOWS/SYSTEM32/ntdll.dll (0x7ffb84f30000)
        KERNEL32.DLL => /c/WINDOWS/System32/KERNEL32.DLL (0x7ffb835d0000)
        KERNELBASE.dll => /c/WINDOWS/System32/KERNELBASE.dll (0x7ffb82520000)
        apphelp.dll => /c/WINDOWS/SYSTEM32/apphelp.dll (0x7ffb7f740000)
        GDI32.dll => /c/WINDOWS/System32/GDI32.dll (0x7ffb83090000)
        win32u.dll => /c/WINDOWS/System32/win32u.dll (0x7ffb828c0000)
        gdi32full.dll => /c/WINDOWS/System32/gdi32full.dll (0x7ffb82c00000)
        msvcp_win.dll => /c/WINDOWS/System32/msvcp_win.dll (0x7ffb82350000)
        ucrtbase.dll => /c/WINDOWS/System32/ucrtbase.dll (0x7ffb82ae0000)
        USER32.dll => /c/WINDOWS/System32/USER32.dll (0x7ffb848c0000)
        msvcrt.dll => /c/WINDOWS/System32/msvcrt.dll (0x7ffb83420000)
        ole32.dll => /c/WINDOWS/System32/ole32.dll (0x7ffb844b0000)
        combase.dll => /c/WINDOWS/System32/combase.dll (0x7ffb84120000)
        RPCRT4.dll => /c/WINDOWS/System32/RPCRT4.dll (0x7ffb83780000)
        SHELL32.dll => /c/WINDOWS/System32/SHELL32.dll (0x7ffb838a0000)
        OPENGL32.dll => /c/WINDOWS/SYSTEM32/OPENGL32.dll (0x7ffb68470000)
        ADVAPI32.dll => /c/WINDOWS/System32/ADVAPI32.dll (0x7ffb82ea0000)
        sechost.dll => /c/WINDOWS/System32/sechost.dll (0x7ffb84650000)
        GLU32.dll => /c/WINDOWS/SYSTEM32/GLU32.dll (0x7ffb6f940000)
        libwebpdemux-2.dll => /mingw64/bin/libwebpdemux-2.dll (0x7ffb7db60000)
        libwebp-7.dll => /mingw64/bin/libwebp-7.dll (0x7ffaf97c0000)
        libwebpmux-3.dll => /mingw64/bin/libwebpmux-3.dll (0x7ffb70010000)
        dxcore.dll => /c/WINDOWS/SYSTEM32/dxcore.dll (0x7ffb7fb60000)

C:\ismailbulut\Projeler\nftsiren>ldd cmd\nftsiren\nftsiren_release_static.exe
        ntdll.dll => /c/WINDOWS/SYSTEM32/ntdll.dll (0x7ffb84f30000)
        KERNEL32.DLL => /c/WINDOWS/System32/KERNEL32.DLL (0x7ffb835d0000)
        KERNELBASE.dll => /c/WINDOWS/System32/KERNELBASE.dll (0x7ffb82520000)
        apphelp.dll => /c/WINDOWS/SYSTEM32/apphelp.dll (0x7ffb7f740000)
        GDI32.dll => /c/WINDOWS/System32/GDI32.dll (0x7ffb83090000)
        win32u.dll => /c/WINDOWS/System32/win32u.dll (0x7ffb828c0000)
        gdi32full.dll => /c/WINDOWS/System32/gdi32full.dll (0x7ffb82c00000)
        msvcp_win.dll => /c/WINDOWS/System32/msvcp_win.dll (0x7ffb82350000)
        ucrtbase.dll => /c/WINDOWS/System32/ucrtbase.dll (0x7ffb82ae0000)
        USER32.dll => /c/WINDOWS/System32/USER32.dll (0x7ffb848c0000)
        msvcrt.dll => /c/WINDOWS/System32/msvcrt.dll (0x7ffb83420000)
        ole32.dll => /c/WINDOWS/System32/ole32.dll (0x7ffb844b0000)
        combase.dll => /c/WINDOWS/System32/combase.dll (0x7ffb84120000)
        RPCRT4.dll => /c/WINDOWS/System32/RPCRT4.dll (0x7ffb83780000)
        SHELL32.dll => /c/WINDOWS/System32/SHELL32.dll (0x7ffb838a0000)
        OPENGL32.dll => /c/WINDOWS/SYSTEM32/OPENGL32.dll (0x7ffb68470000)
        ADVAPI32.dll => /c/WINDOWS/System32/ADVAPI32.dll (0x7ffb82ea0000)
        sechost.dll => /c/WINDOWS/System32/sechost.dll (0x7ffb84650000)
        GLU32.dll => /c/WINDOWS/SYSTEM32/GLU32.dll (0x7ffb6f940000)
        dxcore.dll => /c/WINDOWS/SYSTEM32/dxcore.dll (0x7ffb7fb60000)
```

As you can see it only affects libwebp, but it works because all other dll's can be found in every windows system by default.

Jacalz:
I tried with `CGO_LDFLAGS=-static` (found that somewhere online) and did get some compiler errors about not finding libraries. Does it work for you?

hismailbulut:
> I tried with `CGO_LDFLAGS=-static` (found that somewhere online) and did get some compiler errors about not finding libraries. Does it work for you?

Try this:
```
CGO_LDFLAGS="-extldflags='static'"
```

Also it can not statically link header only libraries. I think this is because why it links others dynamic

hismailbulut:
I meant not header only but your system only has it's headers and .so but not .o

cmd/fyne/internal/commands/build.go
hismailbulut: done

Jacalz:
> > I tried with `CGO_LDFLAGS=-static` (found that somewhere online) and did get some compiler errors about not finding libraries. Does it work for you?
> 
> Try this:
> 
> ```
> CGO_LDFLAGS="-extldflags='-static'"
> ```
> 
> Also it can not statically link header only libraries. I think this is because why it links others dynamic

Does not seem to work. Getting a warning instead:
```
# fyne.io/fyne/v2/cmd/fyne_demo
/usr/bin/ld: warning: cannot find entry symbol xtldflags='-static'; defaulting to 0000000000426d00
```

hismailbulut:
> /usr/bin/ld: warning: cannot find entry symbol xtldflags='-static'; defaulting to 0000000000426d00

Are you sure you are writing it correctly? It should start with -ext with 'e'

Jacalz:
> > /usr/bin/ld: warning: cannot find entry symbol xtldflags='-static'; defaulting to 0000000000426d00
> 
> Are you sure you are writing it correctly? It should start with -ext with 'e'

Yes, I copied what you wrote. I suspect that one "-" results in it thinking -e is a flag.
I tried adding one more "-" and I get this instead:
```
gcc: error: unrecognized command-line option ‘--extldflags='-static'’
```

hismailbulut:
Sorry my bad. It is already CGO_LDFLAGS and what you did at first should work. I found a thread about it
https://groups.google.com/g/golang-nuts/c/7YH5CseB4pc
So it only works with in build command but not in CGO_LDFLAGS

Jacalz:
I also tried `go build -ldflags="-extldflags='--static'"` (basically using two -- instead of one -) and that means that I get the same errors as when using `CGO_LDFLAGS="-static" (same for one or two minus there)`:
```
/usr/lib/golang/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/ld: cannot find -lGL: No such file or directory
/usr/bin/ld: cannot find -lX11: No such file or directory
/usr/bin/ld: cannot find -lXrandr: No such file or directory
/usr/bin/ld: cannot find -lXxf86vm: No such file or directory
/usr/bin/ld: cannot find -lXi: No such file or directory
/usr/bin/ld: cannot find -lXcursor: No such file or directory
/usr/bin/ld: cannot find -lm: No such file or directory
/usr/bin/ld: cannot find -lXinerama: No such file or directory
/usr/bin/ld: cannot find -lGL: No such file or directory
/usr/bin/ld: cannot find -lc: No such file or directory
collect2: error: ld returned 1 exit status
```

hismailbulut:
> go build -ldflags="-extldflags='--static'"

Try this one with one '-'

`go build -ldflags -extldflags="-static"`

Jacalz:
@hismailbulut That results in the same errors posted above.

hismailbulut:
I am not sure why it doesn't work. But you tried this pr and it worked and built succesfully with -static option, isn't it?

Jacalz:
> I am not sure why it doesn't work. But you tried this pr and it worked and built succesfully with -static option, isn't it?

Yes and no, depending on how you look at it. Nothing is statically linked but there are no build errors.

hismailbulut:
> Yes and no, depending on how you look at it. Nothing is statically linked but there are no build errors.

Have you tried installing static builds of some of these libraries? I think it will statically link if you have it installed.

Jacalz:
> > Yes and no, depending on how you look at it. Nothing is statically linked but there are no build errors.
> 
> Have you tried installing static builds of some of these libraries? I think it will statically link if you have it installed.

No, I have not. Didn't know that was a thing that I had to do. I think we need better documentation about this flag.

Bluebugs:
I was going to point out that on many Linux distribution, static library are in a separate package and won't be present which is the error you are seeing @Jacalz . So when you get the error it is the right command, but your system doesn't have the static library installed.

Jacalz (changes requested):
@Bluebugs In that case, I think we have the wrong command because I don't see any errors when building with the current implementation. I also don't know why the binary is stripped when building with `-static`. That seems confusing.

My notes are as follows then:
- We might have the wrong command, or something is off with quoting arguments?
- Building with `-static` strips binaries for some reason?
- We need to document the flag better. Should that be here or somewhere else?

Bluebugs:
I don't know how the go compiler work, but if you compile statically, there is no symbol to export as they are all linked and resolved. Maybe that's what you mean.

Jacalz:
I just see that it seems to do the exact same as `-ldflags="-s -w"` (or the size is at least the same).

hismailbulut:
The -static flag will be passed directly to external linker. It has no relations with either go compiler or the go linker. In that case we all using gcc's 'ld'. And the flag may change depending of which linker the c compiler are using currently. I created this pr because I needed the -static flag and fyne doesn't package already compiled binaries. So I couldn't compile package myself and let fyne to package it with metadata. And maybe in near future someone may need some other linker or compiler flag to build his app and will create another pr for just doing what he wants. It may better to just externalize these ldflags and gcflags and let users control how their app will be built like the '--tags' flag cmd/fyne currently has. But when we have --ldflags then we have to deal with conflicts. We somehow need to force users to not use -s -w and -H=windowsgui to prevent any conflicts with -release flag and windows builds. But I think these are not really a problem we could just process them and remove/add the flags we want or let users do what they want and override the default flags we added for ldflags. 

> I just see that it seems to do the exact same as -ldflags="-s -w" (or the size is at least the same).

I tried the four possibilities and these are the results:

**fyne package**: 35716kb
**fyne package -static**: 36571kb
**fyne package -release**: 16644kb
**fyne package -release -static**: 17413kb

As you can see they are all correct in my tests in windows and -static doesn't strip the binary

Bluebugs:
Indeed @hismailbulut, I agree with your statement and it is a correct assessment of what this patch does. static compilation is something the linker will control and the exact result will depends on which linker/os you are using.

Bluebugs (approved):

Jacalz:
Absolutely. I'm just trying to sort things out and understand it. Sorry for all the fuzz 😅

Jacalz:
Will look at this further tomorrow one last time.

andydotxyz (changes requested):
The tests will need to pass before we can approve this PR...

Jacalz:
> As you can see they are all correct in my tests in windows and -static doesn't strip the binary

For what it's worth, this is what I am seeing on Linux:
```
-rwxr-xr-x. 1 jacob jacob 25106552  9 jan 10.51 fyne_demo
-rwxr-xr-x. 1 jacob jacob 18648168  9 jan 10.50 fyne-demo-release
-rwxr-xr-x. 1 jacob jacob 18648168  9 jan 10.49 fyne-demo-static
-rwxr-xr-x. 1 jacob jacob 18648168  9 jan 10.50 fyne-demo-static-release
```
I don't know if it matters much but it is what I am seeing.

hismailbulut:
> I don't know if it matters much but it is what I am seeing.

You were right, I fixed it thanks

hismailbulut:
> The tests will need to pass before we can approve this PR...

I didn't understand why tests are failing. Do I need to create myTest folder or something?

Bluebugs:
> > The tests will need to pass before we can approve this PR...
> 
> I didn't understand why tests are failing. Do I need to create myTest folder or something?

The test are checking that we call the go build command with the expected information. Every time we change what we pass to the test, it will break. You can also add a new test to make sure that when we specify a static build, it pass the correct flags.

Jacalz:
Hi @hismailbulut. Do you plan on fixing the tests? I believe it is the only remaining item before we can merge this.

hismailbulut:
I stopped working on this because I realized this option causes every library linked statically including `libc` etc. and I think it should not be used for fyne applications. People can add this option themselves by using `CGO_LDFLAGS` environment variable. You can close this pr.

PR #3476, opened 2022-12-15
Minor dialog refactor/cleanup

Jacalz:
### Description:

This is just a very small change that removes an old inefficient method that basically
took one slice and then created a new one. Removing it is faster and, in my opinion,
leads to more clear code. It also just moves  some code into another file.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):

andydotxyz:
Probably best to hold up merging this - we're in RC mode so only blocking issues are being added to the pending release (currently on develop branch)

Jacalz:
Indeed. I am aware. Had no plan to merge this yet. If I hadn't opened the PR, I would just have forgotten it :)

PR #3478, opened 2022-12-16
Use an hbox to reduce label resize in Fyne_Demo

Bluebugs:
### Description:
This further reduce the amount of Refresh triggered on Label in fyne_demo during resize. This significantly reduce the amount of call to Harfbuzz. It is a simple improvement that doesn't introduce much risk at this point in time.

### Checklist:
- [ ] Tests all pass.

### DIFF
--- a/cmd/fyne_demo/main.go
+++ b/cmd/fyne_demo/main.go
@@ -204,7 +204,7 @@ func makeNav(setTutorial func(tutorial tutorials.Tutorial), loadPrevious bool) f
 			return ok && len(children) > 0
 		},
 		CreateNode: func(branch bool) fyne.CanvasObject {
-			return widget.NewLabel("Collection Widgets")
+			return container.NewHBox(widget.NewLabel("Collection Widgets"))
 		},
 		UpdateNode: func(uid string, branch bool, obj fyne.CanvasObject) {
 			t, ok := tutorials.Tutorials[uid]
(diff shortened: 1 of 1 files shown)


Jacalz (commented):
I am sorry, but I am actually against this. It is only a workaround and not a general solution for that fixes the issue for more programs. It complicates the code and makes it harder to follow. If anything, it just hides it and makes it harder to reproduce in fyne_demo

Bluebugs:
I think it comes down to how we expect Fyne to behave. The base optimization path is to not call Refresh. This is what this patch does. It is current best practice. The downside of this approach is that it rely on the developer doing the right things even when it is not the simplest solution.

If we think that to keep the API usage simple, call to Refresh should not be costly, then there is a lot of optimization that have to be pushed all around in Fyne. This seems to be a change in how we approach Fyne optimization. It might be overall the right thing to do and make sense, but it won't happen fast.

In the mean time and for the time being, we are still advising developers to not trigger resize unnecessarily in there design. fyne_demo is not a benchmark for performance, it shows how to write an application using Fyne. I do think we should have a benchmark application that does show all the area where we want to improve performance (like making Refresh on Text or Image faster), but that is not the job of fyne_demo.

Jacalz:
But we can't just wrap every widget in a HBox container to save on Refresh calls. We should fix the underlying issue and make sure widgets only refresh only when necessary. We should not suggest developers to make their code more complex and unintuitive just to get improved performance.

I am sorry but I just don't see the benefit of this. I understand that the big fix is going to be complicated but leaving the issue unresolved and proposing sketchy workarounds is not going to look good for us

andydotxyz:
I think this is superseded by #3481, which essentially makes all trivial text labels minsize and positions them in a simplified layout algorithm. So that will fix all 1-line labels, buttons and entries.

Issue #3471, opened 2022-12-13
Memory spikes and lag when resizing window


Jacalz:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

When using v2.3.0 on development, I am seeing very laggy and slow window resize. It is almost as if whole sizes of the widow are skipped. I am sometimes seeing large spikes in memory usage increases right at the same time as the window resize slows down.

A similar issue occurs when running under Wayland but there is less skipping of frames and more just general slowness.

### How to reproduce

1. Open up fyne_demo.
2. Resize the window up and down. Very large changes in size seems to how the most effect.
3. Notice that the window sometimes lags and stutters when trying to resize.

### Screenshots

[laggy-resize-x.webm](https://user-images.githubusercontent.com/25466657/207436227-fa4c6e13-d4e6-41fd-b25a-5d0581547933.webm)

### Example code

It is eneough to use fyne_demo for this

### Fyne version

c0cde7ab1c251bba361c130eec24b68fe89e7d34

### Go compiler version

1.19.3

### Operating system

Linux

### Operating system version

Fedora 37

Jacalz:
Some interesting notes. Using v2.2.4, it kind of lags a bit sometimes but generally feels a lot better than it does now. Building using `-tags wayland` did initially show slight issues but having trouble replicating there now

andydotxyz:
Should be addressed by #3481

Jacalz:
Were there further optimizations to do in the future? If so, should we leave this open after merging that PR?

andydotxyz:
> Were there further optimizations to do in the future? If so, should we leave this open after merging that PR?

There are always more optimisations to be done. If my PR resolves the title "Memory spikes and lag when resizing window" then I think we should close this ticket accordingly.

Jacalz:
Sure. I'll make sure to review the PR this evening. If we close this then, we should probably make sure to open a new issue to track further improvements :)

andydotxyz:
The PR was merged a while back, is there more required work in this space for the next release?

Jacalz:
Looks great on master and develop. Closing as done :)

PR #3481, opened 2022-12-18
Add some text layout optimisations

andydotxyz:
### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (approved):
This is a noticeably big improvement for the performance. Resizing is a lot faster most of the time but it is still far from smooth and I do see occasional long slowdowns sometimes (but a lot less often).

I am happy with merging this and no longer holding up v2.3.0 but I don't quite think it entirely solves #3471. There are still some lag and spikes in memory usage as far as I can see.

PR #3484, opened 2022-12-20
Remove old fredbi/uri workaround

Jacalz:
### Description:

Now that https://github.com/fredbi/uri/issues/3 has been fixed, we can remove the old workaround.
I also removed the haveAuthority bool` field as it always seems to be set to true.

All the tests pass as the code is a lot cleaner. However, I do not have a through understanding of of the uri standards are supposed to work so it would be good to get someone with a broader understanding to have a look at it. Opening this for inclusion into v2.4.0 at the earliest.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/storage/repository/parse.go
+++ b/storage/repository/parse.go
@@ -28,9 +28,8 @@ func NewFileURI(path string) fyne.URI {
 	}
 
 	return &uri{
-		scheme:        "file",
-		haveAuthority: true,
-		path:          path,
+		scheme: "file",
+		path:   path,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


charlesdaniels (approved):
I think this should be OK.

I don't remember the nature of the problem I was running into with `net/url` that made me do `haveAuthority`... presumably if the tests are passing this is fixed though. Maybe double check and make sure the test coverage for `.String()` is good? It's pretty short so getting it to 100% should be easy if it isn't there already.

Glad we can finally get rid of that kludge!

PR #3497, opened 2022-12-24
Add ctrl-left/right to move along a word

andydotxyz:
### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -928,8 +928,18 @@ func (e *Entry) registerShortcut() {
 			}
 		})
 	}
+	selectMoveWord := func(se fyne.Shortcut) {
+		if !e.selecting {
+			e.selectColumn = e.CursorColumn
+			e.selectRow = e.CursorRow
+			e.selecting = true
+		}
+		moveWord(se)
+	}
 	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyLeft, Modifier: fyne.KeyModifierShortcutDefault}, moveWord)
+	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyLeft, Modifier: fyne.KeyModifierShortcutDefault | fyne.KeyModifierShift}, selectMoveWord)
 	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyRight, Modifier: fyne.KeyModifierShortcutDefault}, moveWord)
+	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyRight, Modifier: fyne.KeyModifierShortcutDefault | fyne.KeyModifierShift}, selectMoveWord)
 }
 
 func (e *Entry) requestFocus() {
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -936,9 +936,13 @@ func (e *Entry) registerShortcut() {
 		}
 		moveWord(se)
 	}
-	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyLeft, Modifier: fyne.KeyModifierShortcutDefault}, moveWord)
+	unselectMoveWord := func(se fyne.Shortcut) {
+		e.selecting = false
+		moveWord(se)
+	}
+	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyLeft, Modifier: fyne.KeyModifierShortcutDefault}, unselectMoveWord)
 	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyLeft, Modifier: fyne.KeyModifierShortcutDefault | fyne.KeyModifierShift}, selectMoveWord)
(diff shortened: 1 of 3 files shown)


widget/entry.go
```diff
@@ -898,6 +898,38 @@ func (e *Entry) registerShortcut() {
...
+		row := e.textProvider().row(e.CursorRow)
+		start, end := getTextWhitespaceRegion(row, e.CursorColumn, true)
+		if start == -1 || end == -1 {
+			return
+		}
+
+		e.setFieldsAndRefresh(func() {
+			if s.(*desktop.CustomShortcut).KeyName == fyne.KeyLeft {
+				if e.CursorColumn == 0 {
+					if e.CursorRow > 0 {
```
Jacalz, line 912: I suppose these could use a single if-statement.

widget/entry.go
```diff
@@ -898,6 +898,38 @@ func (e *Entry) registerShortcut() {
...
+						e.CursorRow--
+						e.CursorColumn = len(e.textProvider().row(e.CursorRow))
+					}
+				} else {
+					e.CursorColumn = start
+				}
+			} else {
+				if e.CursorColumn == len(e.textProvider().row(e.CursorRow)) {
+					if e.CursorRow < e.textProvider().rows()-1 {
```
Jacalz, line 921: Same as comment above.

Jacalz (changes requested):
Well done. Just some minor comments on removing redundant if-statements to lessen the indentation a bit.

On a different note, should we add selection at the same time? Selecting words with shift+ctrl+left/right is very convenient :)

andydotxyz:
> On a different note, should we add selection at the same time? Selecting words with shift+ctrl+left/right is very convenient :)

Hmm good point. It will expand existing selections as you would expect, but does not have code to initiate a selection

widget/entry.go
andydotxyz: I don't think so - the outer if has an else, the inner is a condition of the first `if`.

widget/entry.go
andydotxyz: As before - the else applies to the outer `if`, not to both.

widget/entry.go
Jacalz: But it has the same effect, has it not? If we have `e.CursorColumn == 0 && e.CursorRow > 0` and the first part of the and operation fails, we will fall into the else statement.

widget/entry.go
Jacalz: It should be equivalent to just do `e.CursorColumn == len(e.textProvider().row(e.CursorRow)) && e.CursorRow < e.textProvider().rows()-1`.

andydotxyz:
Updated to fully support selection as well, thanks @Jacalz.
But the points about simplifying `if` don't seem to apply as far as I can tell.

widget/entry.go
Jacalz: I don't see why this would be wrong. Avoids the triple if statements and instead keeps it down to two.
```go
if e.CursorColumn == 0 && e.CursorRow > 0 {

} else {
	
}
```

widget/entry.go
Jacalz: I don't see why this would be wrong. Avoids the triple if statements and instead keeps it down to two.
```go
if e.CursorColumn == len(e.textProvider().row(e.CursorRow)) && e.CursorRow < e.textProvider().rows()-1 {

} else {

}
```

Jacalz (commented):
I do not quite a agree with that. It should be possible to remove the most inner if-statement and keep the indentation to two instead of three steps in. Make the code a bit more readable in my opinion.

Jacalz (changes requested):
I noticed that the selection isn't quite behaving as expected in one regard (absolutely perfect otherwise). Having something selected and then just pressing `ctrl` + `left/right` should dismiss the selection and only move the cursor. However, it currently seems to still apply the selection (even through `shift` isn't pressed).

widget/entry.go
andydotxyz: What about the case where `e.CursorColumn = 0` and `e.CursorRow = 0`?
In the original code we do nothing, in your revised code it calls to else, which it shouldn't (as far as I can see)...

andydotxyz:
> I noticed that the selection isn't quite behaving as expected in one regard (absolutely perfect otherwise). Having something selected and then just pressing `ctrl` + `left/right` should dismiss the selection and only move the cursor. However, it currently seems to still apply the selection (even through `shift` isn't pressed).

Good point, thanks.

widget/entry.go
andydotxyz: As above it will change the action for one of the cases. Are you sure that is a desirable change?

Jacalz: You are right. Sorry for the fuss

Jacalz (changes requested):
Looking really good. Can we just add a test case for the selection dismiss?

andydotxyz:
Thanks @Jacalz, added

Jacalz (approved):
Well done. This improves text ergonomics a lot :)

PR #3516, opened 2022-12-29
Don't log a common error - lots of environments don't have this exported

### LINKED ISSUE #3515: org.freedesktop.portal.Desktop not exist
### Checklist

### Describe the bug

it is not fatal bug, but I have tried 2.2 vs 2.3 version on termux-proot-arch (except litle bit bigger binary 13M vs 16M) I have found interasting log:    `org.freedesktop.portal.Desktop not exist...`

### How to reproduce

1. run andronix-proot on termux
2. compile current version
3. see log

### Screenshots

Screenshot_20221229_201438_VNC Viewer

### Example code

```
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Hello")

		hello,
			hello.SetText("Welcome :)")
		}),
	))

	w.ShowAndRun()
}
```

### Fyne version

2.3

### Go compiler version

1.19

### Operating system

Linux

### Operating system version

arch (android 13 on sm-p619)

### Additional Information

version show neofetch ....

andydotxyz:
Fixes #3515

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -48,7 +48,7 @@ func findFreedestktopColorScheme() fyne.ThemeVariant {
 		"color-scheme",
 	)
 	if call.Err != nil {
-		fyne.LogError("failed to read theme variant from D-Bus", call.Err)
+		// many desktops don't have this exported yet
 		return theme.VariantDark
 	}
 


Jacalz (commented):
Is it not better to keep the error? In my opinion, this is not much different than a systray enabled application giving an error when there is no systray support.

Would it perhaps be better to instead also print something about the desktop potentially not supporting dark theme preference?

andydotxyz:
It's not really the same. This is a spec set up by one of many platforms that others don't respect. It means that a large percentage of apps launched on Linux will show this error.

I was responding to an issue which was opened because of just such a setup. 

It's not like systray - in that instance functionality will be limited. In this case it is likely of zero impact, because a desktop without this property may be setting dark mode in some other way.

Jacalz:
But there could be genuine errors that we don't see on platforms with working support?

andydotxyz:
This is one of many locations that can error in the method. This one is looking up the service which provides this data. If it is not present (but the dbus is working, as no earlier error) I'm not sure what else it could mean...

If looking for comparisons it's a bit like running on an Apple Intel and presenting an error because an M1 specific extension cannot be found. Something that a lot of users will see and is quite reasonable to ignore.

Jacalz (approved):
Alright then.

Issue #3498, opened 2022-12-25
TextWrapWord will cause crash in RichText


aynakeya:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

If a label in a Scroll container have its text starts with a space.  and its Wrapping equals `TextWrapWord`.

eg
```
l :=widget.NewLabel(" AAAA")
l.Wrapping = fyne.TextWrapWord
```

Then program will crash at `RichText.go` with index out of range -1

![image](https://user-images.githubusercontent.com/32156054/209456184-86811807-90cb-4c92-8805-b5bb7f2c74a0.png)

### How to reproduce

Run code in Example code will cause program to crash

### Screenshots

![image](https://user-images.githubusercontent.com/32156054/209456140-cb32de6d-4a23-4323-b662-c4beede930ee.png)

### Example code

```
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Hello World")

	texts := make([]fyne.CanvasObject, 32)
	for i := 0; i < len(texts); i++ {
		l := widget.NewLabelWithStyle(
			" AAAA",
			fyne.TextAlignCenter, fyne.TextStyle{})
		l.Wrapping = fyne.TextWrapWord
		texts[i] = l
	}
	vbox := container.NewVBox(texts...)
	scroll := container.NewScroll(vbox)
	w.SetContent(scroll)
	w.Resize(fyne.NewSize(360, 540))
	w.ShowAndRun()
}
```

### Fyne version

v2.2.4

### Go compiler version

1.19

### Operating system

Windows

### Operating system version

Windows 10

aynakeya:
Maybe its the same problem as #3291

andydotxyz:
Please can you test against v2.3.0 which was released yesterday. Lots of text code was upgraded.

aynakeya:
> Please can you test against v2.3.0 which was released yesterday. Lots of text code was upgraded.

Problem still exists in v2.3.0
![image](https://user-images.githubusercontent.com/32156054/209735816-eeaa9ab8-42b7-4f35-af91-6154c4828d24.png)

[...]

PR #3523, opened 2022-12-30
Fix crashes for RichText

### LINKED ISSUE #3291: Crash in widget.(*RichText).updateRowBounds.func1
### Checklist

### Describe the bug

We used the Fuzz engine to modify some Test(TestEntry_MultilineWrapping_DeleteWithBackspace) data, and then the following crash appeared. Originally I thought it was similar to the issue #3288 & #3289, but this requires two other corrupt data to trigger, and finally crash in different location.

### How to reproduce

### Screenshots

CrashLog
```
panic: runtime error: index out of range [-1] [recovered]
        panic: runtime error: index out of range [-1]

goroutine 100 [running]:
testing.tRunner.func1.2({0x87e8c0, 0xc000024288})
testing.tRunner.func1()
panic({0x87e8c0, 0xc000024288})
fyne.io/fyne/v2/widget.(*RichText).updateRowBounds.func1({0xc00083e6e0, 0x1, 0xc0005adb78})
        /home/zjx/workspace/gowork/src/purelib/crashConfirm/fyne/widget/richtext.go:381 +0xbc5
fyne.io/fyne/v2/widget.(*RichText).updateRowBounds(0xc0001646e0)
        /home/zjx/workspace/gowork/src/purelib/crashConfirm/fyne/widget/richtext.go:411 +0x106
fyne.io/fyne/v2/widget.(*RichText).Refresh(0xc0001646e0)
        /home/zjx/workspace/gowork/src/purelib/crashConfirm/fyne/widget/richtext.go:89 +0x1e
fyne.io/fyne/v2/widget.(*RichText).insertAt(0xc0001646e0, 0x5, {0xc0005add78, 0x1})
        /home/zjx/workspace/gowork/src/purelib/crashConfirm/fyne/widget/richtext.go:226 +0x285
fyne.io/fyne/v2/widget.(*Entry).TypedRune(0xc00000c1e0, 0x30)
        /home/zjx/workspace/gowork/src/purelib/crashConfirm/fyne/widget/entry.go:738 +0x11b
fyne.io/fyne/v2/test.typeChars({0xc0005ade20, 0x6, 0xc0005ade68}, 0xc0005adea0)
        /home/zjx/workspace/gowork/src/purelib/crashConfirm/fyne/test/test.go:311 +0x39
fyne.io/fyne/v2/test.Type({0x976960, 0xc00000c1e0}, {0x8ab7bd, 0x6})
        /home/zjx/workspace/gowork/src/purelib/crashConfirm/fyne/test/test.go:232 +0xbc
fyne.io/fyne/v2/widget_test.TestEntry_MultilineWrapping_DeleteWithBackspace(0x0)
        /home/zjx/workspace/gowork/src/purelib/crashConfirm/fyne/widget/entry_test.go:428 +0xc5
```

### Example code

```go
//part of the Test Code
	entry := widget.NewMultiLineEntry()
	entry.Resize(fyne.NewSize(64, -3.0268908e+12))
	test.Type(entry, "\x20\x30\x30\x30\xD4")
	test.Type(entry, "\x20\x30\x30\x30\xD4\x20")
```

### Fyne version

commit 638ae24e4bebfc0b6dbaabf8db5807eeabfafc61 (HEAD -> master, origin/master, origin/HEAD)

### Go compiler version

1.17.8

### Operating system

Linux

### Operating system version

OS: Linux r920 4.15.0-159-generic #167-Ubuntu

### Additional Information

_No response_

alexballas:
### Description:
Handle the case where the beginning of the label or entry text is a space.

Fixes #3291 #3498

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/richtext_test.go
```diff
@@ -867,6 +867,10 @@ func TestText_findSpaceIndex(t *testing.T) {
 			text: "ww wwww www wwwww",
 			want: 11,
 		},
+		"space beginning": {
+			text: " ww",
+			want: 2,
```
andydotxyz, line 872: It doesn't seem right - the space should be found at position 0, should it not?

andydotxyz (changes requested):
Thanks for working on this. I'm glad the crash is gone, but the additional test seems incorrect.

widget/richtext_test.go
alexballas: Hello @andydotxyz. To my understanding, findSpaceIndex is basically a way to check where words are being separated. A space at the beginning of a word is not really separating anything so it would not make sense to treat it as a separator. I agree that as it is now, it's not super clear. I could try to move the change to the `lineBounds` function and check for the zero value there and leave findSpaceIndex to do its things

alexballas: The same could be said about the `fallback` return in findSpaceIndex. Returning the fallback value does not really make it clear if it found a space at lets say return value X or it found none.

alexballas: Actually the above statement I stroke through is correct
If you have "text", you get an index 3 (fallback)
If you have "text " you get again index 3, not 4

andydotxyz: I was just going by the comments of the method, about returning the last space on a string. Does it not fix the bug if, for leading space, you return 0?
If not then this seems like a good fix - perhaps the method doc just needs a little updating.

alexballas: Unfortunately returning 0 due to leading spaces is what seems to trigger the crash. I'll update the comment tomorrow and let you know, for further approvals

alexballas: > I was just going by the comments of the method

apologies, by method, you mean the function `findSpaceIndex` right?

alexballas:
@andydotxyz messing with findSpaceIndex does feel a bit confusing at the end of the day, would a change like the following make more sense?
```go
spaceIndex := findSpaceIndex(sub, fallback)
if spaceIndex == 0 {
	spaceIndex = fallback
}

high = low + spaceIndex
```

andydotxyz:
That sounds reasonable yes. Though it's still not entirely clear why "fallback" is the correct position to choose. I guess the real problem is that start and end become the same?

If i double tap to select a string with a space at the beginning what happens?

alexballas:
>  If i double tap to select a string with a space at the beginning what happens?
Only the visible text is selected, not the spaces.

Regarding your question about using fallback when 0 is returned. By using fallback we tell the app to "consider" the rest of the text line as the first instance of the word and not prematurely push the first word to the next line before we start splitting it.

Example:
If we return fallback when 0:
Screenshot from 2022-12-31 11-29-12

If we return 1 when 0 (we split prematurely):
Screenshot from 2022-12-31 11-31-55

Any other value would just split the word inconsistently

alexballas:
@andydotxyz I run some check on some other text editors, and the screenshot where if there is a space at the beginning it pushes the whole word below, seems to be the expected behaviour.

andydotxyz:
> @andydotxyz I run some checks on some other text editors, and the screenshot where if there is a space at the beginning it pushes the whole word below, seems to be the expected behaviour.

Agreed. Upon further testing I see simple editors move the whole word that must wrap to a new line - the whitespace is above as it is superfluous to the word wrapping. The number of the spaces does not matter, it is a whitespace block and it should be consistent IMHO.

alexballas:
ty! It should fine now, could you please also confirm when the time permits?

andydotxyz (changes requested):
This has broken other things, such as this 1-liner in fyne_demo
wrapchange

alexballas:
Hello @andydotxyz and happy new YEAR!
This issue does not look related to this fix. It's already broken on develop. We should raise a new ticket for it.

alexballas:
@andydotxyz specifically this change seems to be causing the fyne_demo issue https://github.com/alexballas/fyne/blob/d46e4f78b05aae5489a197c799f7b476dfe2dc14/widget/richtext.go#L48

andydotxyz:
Can you merge develop in to your PR and see how it works out? The glitch found above has been fixed upstream.

alexballas:
> Can you merge develop in to your PR and see how it works out? The glitch found above has been fixed upstream.

Merged Develop into the fix branch. I don't see the glitch you mentioned.

andydotxyz (approved):
Thanks

PR #3525, opened 2022-12-31
Add the ability to disable ToolbarActions #2306

### LINKED ISSUE #2306: Add the ability to disable ToolbarActions
ToolbarActions and MenuItems needs to be disableable for the similar reasons and coherence with DisableableWidgets. Last corresponds to solved #234 .

pravin772:
### Description:
Issue addressed Add the ability to disable ToolbarActions #2306
The change is inspired from same properties as Button for ToolbarAction.
Comments from issue were helpful and followed during code change.

Fixes #2306 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


pravin772:
Hi,
Seen this issue open and labeled "good first issue". So, tried to implement this feature. Happy to contribute more.
Please let me know if any change necessary. Happy to comply with required changes. Thanks

widget/toolbar.go
```diff
@@ -37,7 +38,17 @@ func (t *ToolbarAction) SetIcon(icon fyne.Resource) {
...
+		Icon:        icon,
+		OnActivated: onActivated,
+	}
+	return toolbarAction
+}
+
+// SetDisabled updated the disabled property of ToolbarAction
+func (t *ToolbarAction) SetDisabled(disabled bool) {
+	t.disabled = disabled
+	t.ToolbarObject().Refresh()
```
andydotxyz, line 51: This doesn't seem right - it creates a new object visual and then refreshes it instead of refreshing the current UI.

andydotxyz (changes requested):
Unfortunately embedding DisablableWidget is not the right thing to do, as these are not widgets.

Matching the functions it provides does make sense though.

widget/toolbar.go
pravin772: Thanks for review. Does t.Refresh() makes sense? That will re-render same element.

pravin772:
> Unfortunately embedding DisablableWidget is not the right thing to do, as these are not widgets.
> 
> Matching the functions it provides does make sense though.

Do you reckon should we use separate disabled property in ToolbarAction struct?

andydotxyz:
Yes I think so - replicating DisabledWidget without bringing in BaseWidget along with it

pravin772:
Added disabled property with all equivalent methods as DisabledWidget. Could you suggest how can I refresh that element.
Cannot use t.Refresh() as it is no more BaseWidget. Moreover, do I need to update styles specific for disabled ToolbarAction?

andydotxyz (changes requested):
This looks clean, but it does not seem to refresh the widget when `disabled` changes.

pravin772:
Hi, I was not able to use t.Refresh() as it is not BaseWidget anymore. However, I found that t.ToolbarObject().Refresh() can be used as t.ToolbarObject() gets a button to render this ToolbarAction. Let me know if that works otherwise please could you suggest any solution. Thanks

andydotxyz:
> Hi, I was not able to use t.Refresh() as it is not BaseWidget anymore. However, I found that t.ToolbarObject().Refresh() can be used as t.ToolbarObject() gets a button to render this ToolbarAction. Let me know if that works otherwise please could you suggest any solution. Thanks

Indeed, you'll have to find the right thing to refresh. Unfortunately `ToolbarObject()` creates a new instance, so you are refreshing something not on screen.
If you test this in a real app you should notice that it does not update if you toggle it after drawn for the first time

PR #3529, opened 2023-01-01
Implement keyboard navigation in file dialog

Jacalz:
### Description:

This greatly improves the usability of the file dialog
by making sure that it can be navigated with a keyboard.
- Hover previous and next items using tab and shift+tab.
- Select hovered item by pressing space.

I also noticed that the background was refreshed twice, so I removed that. Should avoid a redraw, I think :)

For #1515

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/fileitem.go
+++ b/dialog/fileitem.go
@@ -30,15 +30,16 @@ type fileDialogItem struct {
 	dir      bool
 
 	hovered bool
+	focused bool
 }
 
 func (i *fileDialogItem) FocusGained() {
-	i.hovered = true
+	i.focused = true
 	i.Refresh()
 }
 
 func (i *fileDialogItem) FocusLost() {
-	i.hovered = false
+	i.focused = false
 	i.Refresh()
 }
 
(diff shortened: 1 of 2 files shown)


andydotxyz:
Nice improvement - but should we not consider the file panel as "1 widget" so arrow keys move around it like the list/table/tree improvements?
A file listing could be really long so the tab/shift tab would be very slow to get to buttons

Jacalz:
Good point. Will add that support.

Jacalz:
I had a lot of troubles with the focusing using left/right and up/down. It is not that I do not know how to do it, but it basically just want working as expected in some cases and resulted in extremely ugly code. Can we not merge this for now and then improve by adding the other support later? This is a big improvement already in my opinion.

andydotxyz:
I guess we could merge it - but are we OK that the key navigation this supports will change (alongside List/Tree/Table I guess?).

Looking for others thoughts here I guess.

Jacalz:
Sorry for the late answer. I believe you are right that we need a better solution here. Will close this for now to focus one some other stuff (lots of project work at the uni taking up time).

Jacalz:
Will move the performance improvements into another PR in the meantime :)

PR #3530, opened 2023-01-01
Implement kayboard navigation in widget.Tree

Jacalz:
### Description:

This implements the Focusable interface for Tree Nodes. Added keyboard navigation:
- Navigate between tree nodes using `tab` and `shift + tab`.
- Select a tree node by pressing space.
- Open a branch by pressing `right` and close using `left`. 

For https://github.com/fyne-io/fyne/issues/1515

Opening as draft for now until I have implemented tests.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/tree.go
+++ b/widget/tree.go
@@ -712,6 +712,12 @@ func (n *treeNode) TypedRune(_ rune) {
 func (n *treeNode) TypedKey(key *fyne.KeyEvent) {
 	if key.Name == fyne.KeySpace {
 		n.tree.Select(n.uid)
+	} else if n.isBranch {
+		if key.Name == fyne.KeyRight {
+			n.tree.OpenBranch(n.uid)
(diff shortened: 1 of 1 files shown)


andydotxyz:
It seems like up/down would pair better with left/right than tab/shift+tab do...

Jacalz:
But the focusing controls that landed in the List widget a while ago just uses Tab?

andydotxyz:
> But the focusing controls that landed in the List widget a while ago just uses Tab?

Ah. Oops. On playing with it I don't think that was right.
Unfortunately it needs work on List as it will only tab through visible widgets - it really should handle moving up and down the list of items full, not just those rendered.

Jacalz:
I'll close this and work on some other items that I prioritize higher. My backlog is getting rather big and with my limited spare time,  I prefer to work on improvements that I will be using in the near future. Sorry.

PR #3531, opened 2023-01-02
Implement focus and keyboard controls for slider

Jacalz:
### Description:

This implements focus and hover for the slider widget to make it follow the material design guidelines better.
I have also added support for moving the slider using the arrow keys.

For #1515 

Opening as draft for now until I have fixed some things and added tests.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/slider.go
+++ b/widget/slider.go
@@ -2,7 +2,6 @@ package widget
 
 import (
 	"fmt"
-	"image/color"
 	"math"
 
 	"fyne.io/fyne/v2"
@@ -256,11 +255,10 @@ func (s *Slider) CreateRenderer() fyne.WidgetRenderer {
 	track := canvas.NewRectangle(theme.InputBackgroundColor())
 	active := canvas.NewRectangle(theme.ForegroundColor())
 	thumb := &canvas.Circle{FillColor: theme.ForegroundColor()}
-	focusIndicator := &canvas.Circle{FillColor: color.Transparent}
 
-	objects := []fyne.CanvasObject{track, active, thumb, focusIndicator}
+	objects := []fyne.CanvasObject{track, active, thumb}
 
-	slide := &sliderRenderer{widget.NewBaseRenderer(objects), track, active, thumb, focusIndicator, s}
+	slide := &sliderRenderer{widget.NewBaseRenderer(objects), track, active, thumb, s}
 	slide.Refresh() // prepare for first draw
 	return slide
 }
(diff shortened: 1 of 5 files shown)


andydotxyz:
This is really nice.
Not sure about the gap in the middle of the focused indicator - is that according to material, or could we make it solid and/or a wider gap in the middle?

Jacalz:
Hmm. I don't know why there is a gap in the middle. I am just using the circle stroke. Shouldn't that just be outside of the fill?

andydotxyz:
Stroke is half inside half outside the fill. To make it fully outside set the stroke to half the size and increase the circle diameter by stroke (and move it up/left by stroke/2)

Jacalz:
Hmm. I don't think that will work as the stroke colour is partly transparent. Will revert back to using two circles instead.

Jacalz:
I have one strange issue that I don't know why it is happening. The focus circle seems to be one pixel too far to the right on my screen. Any ideas why? Rounding error?

andydotxyz:
> I have one strange issue that I don't know why it is happening. The focus circle seems to be one pixel too far to the right on my screen. Any ideas why? Rounding error?

Looking at the markup test files they don't have the same centre point

andydotxyz:
I see an off-by-one but only because it's not possible to align centrally with current numbers I think?

slidezoom

widget/slider.go
```diff
@@ -321,6 +387,11 @@ func (s *sliderRenderer) Layout(size fyne.Size) {
 
 	s.thumb.Move(thumbPos)
 	s.thumb.Resize(fyne.NewSize(diameter, diameter))
+
+	focusIndicatorSize := fyne.NewSize(theme.IconInlineSize()+theme.InnerPadding(), theme.IconInlineSize()+theme.InnerPadding())
```
andydotxyz, line 391: Maybe if this was coded relative to the calculated `diameter` (i.e. `diameter+theme.InnerPadding()`) the issue may be resolved? Just wondering as it could be a rounding error somewhere and that might work around it...

Jacalz: I tried that initially but had some issues with getting the size to match the radio buttons. Will have a shot at it again though :)

Jacalz: Tried a lot of things to work around the possible rounding error but to no avail (even tried using `theme.InnerPadding() + 1.75` which the `delta` variable becomes if simplified). I have no idea how to take this further with this issue unfortunately. I would appreciate some help if you have the time :sweat_smile:

Jacalz: I believe that the code should be mostly fine in all other areas though. I have just added the remaining comments, moved some methods around to be in alphabetical order and fixed tap focus to work correctly.

Jacalz: This looks to be solved now thanks to your slider wobble fix :)

Jacalz:
I will rebase this once https://github.com/fyne-io/fyne/pull/3651 has landed :)

andydotxyz (approved):
Looks good now thanks.
I wonder though - should we have "since" headers on methods that are just added for getting signals from the driver? Not sure so accepting in case it's not a thing to add.

Jacalz:
I had to rebase onto develop to get the workflows to pass. I also added the "Since 2.4" comments. I think it makes sense to be consistent even though they are mostly meant to be used by the driver :)

andydotxyz (approved):

PR #3539, opened 2023-01-04
Add OpenGL (ES)/GPU rendered rectangle and rectangle with rounded edges

renlite:
### Description:
This PR replaces the PR https://github.com/fyne-io/fyne/pull/3121. The difference to this PR is the technique how the primitive OpenGL objects are rendered. In the PR https://github.com/fyne-io/fyne/pull/3121 the objects were mainly calculated on the CPU whereas in this PR the objects are primarily calculated on the GPU. 
The new OpenGL rendering should bring better performance and the possibility to paint antialized rectangles with rounded edges. With this technology it's possible to paint a circle. In a second step the texture based circle could be switched too.
Thanks to: @Bluebugs for support and @andydotxyz @ErrorNoInternet @Jacalz for tests.

Fixes issue https://github.com/fyne-io/fyne/issues/1090

### Checklist:

- [x] Tests included.
- [X] Tests all pass.


cmd/rect_test/main.go
```diff
@@ -0,0 +1,79 @@
+package main
```
Bluebugs, line 1: The "Tests" checkbox is actually for a test that would pass via `go test`, not a binary to manually check the result. Actually what is in this file should likely be included in fyne_demo/canvas section.

internal/painter/gl/draw.go
```diff
@@ -83,7 +82,11 @@ func (p *painter) drawObject(o fyne.CanvasObject, pos fyne.Position, frame fyne.
 	case *canvas.Raster:
 		p.drawRaster(obj, pos, frame)
 	case *canvas.Rectangle:
-		p.drawRectangle(obj, pos, frame)
+		if o.(*canvas.Rectangle).Radius.Pixel == 0 && o.(*canvas.Rectangle).Radius.Percent == 0 {
```
Bluebugs, line 85: :+1:

internal/painter/gl/draw.go
```diff
@@ -97,12 +100,147 @@ func (p *painter) drawRaster(img *canvas.Raster, pos fyne.Position, frame fyne.S
...
+	frameWidthScaled := roundToPixel(frame.Width*p.pixScale, 1.0)
+	frameHeightScaled := roundToPixel(frame.Height*p.pixScale, 1.0)
+	p.ctx.Uniform4f(frameSizeUniform, frameWidthScaled, frameHeightScaled, 0.0, 0.0)
+
+	rectCoordsUniform := p.ctx.GetUniformLocation(p.roundRectangleProgram, "rect_coords")
+	x1Scaled := roundToPixel(points[0]*p.pixScale, 1.0)
+	x2Scaled := roundToPixel(points[4]*p.pixScale, 1.0)
+	y1Scaled := roundToPixel(points[1]*p.pixScale, 1.0)
+	y2Scaled := roundToPixel(points[9]*p.pixScale, 1.0)
+	p.ctx.Uniform4f(rectCoordsUniform, x1Scaled, x2Scaled, y1Scaled, y2Scaled)
```
Bluebugs, line 194: There is a bit of duplicated code between the initialization for both shader, it would be great to refactor the common code as much as possible.

internal/painter/gl/draw.go
```diff
@@ -231,6 +369,66 @@ func (p *painter) lineCoords(pos, pos1, pos2 fyne.Position, lineWidth, feather f
 	}, halfWidth, featherWidth
 }
 
+/*
+func (p *painter) flexLineCoordsNew(pos, pos1, pos2 fyne.Position, lineWidth, feather float32, frame fyne.Size) ([]float32, float32, float32) {
```
Bluebugs, line 373: We prefer to not have dead code. Would you mind removing this until it is of use?

Bluebugs (changes requested):
Looking pretty good. Just a small more cleanup in my opinion. I will try to dedicate some time for testing on mobile and web later this week.

internal/painter/gl/draw.go
renlite: 3 new func, code reuse - done

cmd/rect_test/main.go
Jacalz: Indeed. This would have to be removed as we shouldn't add more applications into the cmd/ directory.

renlite: Yes, this is still an open point. I haven't had time yet.
Next steps will be:
- Test implementation
- Extend fyne_demo
- Delete cmd/rect_test

In the meantime you can use the rect_test for web and mobile testing.

geometry.go
```diff
@@ -140,3 +140,9 @@ func (s Size) Subtract(v Vector2) Size {
 func (s Size) SubtractWidthHeight(width, height float32) Size {
 	return Size{s.Width - width, s.Height - height}
 }
+
+// RectangleRadius describes a radius in pixel or in percentage
+type Radius struct {
+	Pixel   float32
```
andydotxyz, line 146: I don't think this is a good idea - the toolkit API is all vector and therefore pixel parameters should not be used.

Can we just use `float32` instead of `Radius` - the `Width` and `Height` are the same so that will be what developers are expecting. We also don't have percentage based calculations elsewhere so I don't think it's needed.

andydotxyz (changes requested):
I like how this is shaping up - and agree with the other comments.
However I don't agree with the need for a `Radius` type - and I certainly don't think we should use pixel definition in the middle of a vector toolkit.

geometry.go
renlite: from @andydotxyz in https://github.com/fyne-io/fyne/pull/3121.
> It may be worth noting that if Radius == Height/2 then the ends of the rectangle should be a semi-circle, but I don't think it is, so when adjusting the scaling calculations maybe check this too?

This statement was for me the wish to have a percentage alternative for radius. 

How to solve this? The developer uses a container that changes the size depending on the windows' size.  The included rectangle changes the size dynamically too.
- The rectangle should look always the same, eg. semi-circle

In this case it would even be necessary to define the StrokeWidth as percentage to get an identical looking rectangle.
If you don't like the struct it could be named:
```go
Radius float32
RadiusRelation  float32 //RadiusDynamic RadiusPercent
StrokeWidth float32
StrokeWidthRelation float32 // StrokeWidthDynamic StrokeWidthPercent
```
It could be possible to remove the percentage from the rectangle and to define a separate shape eg ```FlexRectagle```, ```DynamicRectagle```, where the Radius and SrokeWidth are always handled in percentage. This could be implemented later after this PR. I would prefer the solution with an own Shape-Type. It would be faster and would have less complexity.

Please let me know how to change the code.

andydotxyz: I think you are putting to much smart into dumb canvas objects.
Something resizing the rectangle can set it's radius to be different if it is trying to maintain a specific shape.
Just like if someone wants a square - they have to ensure that the width and height match even if the container space is rectangular not square.

I don't think it's worth a more complicated API at this level to handle corner cases like that automatically. The canvasobjects are our primitive draw type - widgets using them are where the layout and render details are calculated.

renlite: OK, I will remove the percentage and the Radius will be ```Radius float```.

cmd/rect_test/main.go
renlite: Open points done.

geometry.go
renlite: Refactor Radius done.

internal/painter/gl/draw.go
```diff
@@ -97,12 +96,111 @@ func (p *painter) drawRaster(img *canvas.Raster, pos fyne.Position, frame fyne.S
...
 }
 
-func (p *painter) drawRectangle(rect *canvas.Rectangle, pos fyne.Position, frame fyne.Size) {
-	if (rect.FillColor == color.Transparent || rect.FillColor == nil) && (rect.StrokeColor == color.Transparent || rect.StrokeColor == nil || rect.StrokeWidth == 0) {
-		return
+func (p *painter) drawRectangle(
+	rect *canvas.Rectangle,
+	pos fyne.Position,
+	frame fyne.Size,
+) {
```
Jacalz, line 103: I don't think we really do this with the input parameters. Would prefer if you left them on one line like before.

internal/painter/gl/draw.go
```diff
@@ -97,12 +96,111 @@ func (p *painter) drawRaster(img *canvas.Raster, pos fyne.Position, frame fyne.S
...
+	p.ctx.DrawArrays(triangles, 0, 6)
+	p.logError()
+	p.freeBuffer(vbo)
+}
+
+func (p *painter) drawRoundRectangle(
+	rect *canvas.Rectangle,
+	pos fyne.Position,
+	frame fyne.Size,
+) {
```
Jacalz, line 152: Same as the comment above.

internal/painter/gl/draw.go
```diff
@@ -278,6 +376,40 @@ func rectInnerCoords(size fyne.Size, pos fyne.Position, fill canvas.ImageFill, a
 	return size, pos
 }
 
+func (p *painter) vecRectCoords(
+	pos fyne.Position,
+	rect *canvas.Rectangle,
+	frame fyne.Size,
+) []float32 {
```
Jacalz, line 383: Same as the comment above.

internal/painter/gl/draw.go
```diff
@@ -297,3 +429,30 @@ func roundToPixelCoords(size fyne.Size, pos fyne.Position, pixScale float32) (fy
...
+
+// Returns FragmentColor(red,green,blue,alpha) from fyne.Color
+func getFragmentColor(col color.Color) (float32, float32, float32, float32) {
+	r, g, b, a := col.RGBA()
+	if a == 0 {
+		return 0, 0, 0, 0
+	} else {
+		alpha := float32(a)
+		return float32(r) / alpha, float32(g) / alpha, float32(b) / alpha, alpha / 0xffff
+	}
```
Jacalz, line 441: You can remove the unnecessary else-statement here. Avoids some indentation and makes code more readable.
```suggestion
	}

	alpha := float32(a)
	return float32(r) / alpha, float32(g) / alpha, float32(b) / alpha, alpha / 0xffff
```

internal/painter/gl/shaders.go
```diff
@@ -45,3 +45,37 @@ var shaderSimpleesVert = &fyne.StaticResource{
...
+}
+
+// rectangle_es.vert is used for rectangle_es.frag and round_rectangle_es.frag
+var shaderRectangleesVert = &fyne.StaticResource{
+	StaticName: "rectangle_es.vert",
+	StaticContent: []byte(
+		"#version 100\r\n\r\n#ifdef GL_ES\r\n# ifdef GL_FRAGMENT_PRECISION_HIGH\r\nprecision highp float;\r\n# else\r\nprecision mediump float;\r\n#endif\r\nprecision mediump int;\r\nprecision lowp sampler2D;\r\n#endif\r\n\r\nattribute vec2 vert;\r\nattribute vec2 normal;\r\n\r\nvoid main() {\r\n    vert;  //Workaround, because WebGL optimizes away attributes unused\r\n    gl_Position = vec4(normal, 0, 1);\r\n}\r\n"),
+}
```
Jacalz, line 81: I think you need to re-generate this using `go generate` in this directory.
The comments indicate to me that the files has been manually edited afterwards and some variable names do not match compared to what I get when generating the file locally. As it is a generated file, we need to be able to regenerate it without making manual modifications afterwards.

cmd/fyne_demo/tutorials/canvas.go
```diff
@@ -45,5 +45,15 @@ func canvasScreen(_ fyne.Window) fyne.CanvasObject {
...
+		&canvas.Rectangle{
+			FillColor:   color.NRGBA{R: 255, G: 200, B: 0, A: 180},
+			StrokeColor: color.NRGBA{R: 255, G: 120, B: 0, A: 255},
+			StrokeWidth: 4.0,
+			Radius:      20},
+		&canvas.Rectangle{
+			FillColor:   color.NRGBA{R: 255, G: 200, B: 0, A: 180},
+			StrokeColor: color.NRGBA{R: 255, G: 120, B: 0, A: 255},
+			StrokeWidth: 2.0,
+			Radius:      45},
```
Jacalz, line 57: Would it not make more sense to place these after the rectangle? Also, the second one just looks like another circle. Do we need to showcase both (the other ones are only one for each type of canvas object)?

Jacalz (changes requested):
Well done. I had a look through the code and just found a few small things that I would like to have fixed up :)

internal/painter/gl/draw.go
renlite: You are right.

internal/painter/gl/draw.go
renlite: I changed the formating of importing parameters to one line, but IMO the readability of the func is much better if you use more lines.  
Example one line:
```Go
func (p *painter) lineCoords(pos, pos1, pos2 fyne.Position, lineWidth, feather float32, frame fyne.Size) ([]float32, float32, float32) {
	// Shift line coordinates so that they match the target position.
	xPosDiff := pos.X - fyne.Min(pos1.X, pos2.X)
```
The line seems not to have an end and the func signuture is confusing.

internal/painter/gl/shaders.go
renlite: I did't know that and used ```fyne bundle``` for the single files.

andydotxyz:
I think we will also have to add support to rendering the radius in software renderer. Should we address that here too, or add it as a follow-on item?

Bluebugs:
> I think we will also have to add support to rendering the radius in software renderer. Should we address that here too, or add it as a follow-on item?

I think this PR is already big enough and complex enough in its form. I would think it would be better to have an issue created with pointer to where this change should be done and have a follow up issue address it. Just to keep things as simple as it can be.

internal/painter/gl/shaders.go
renlite: After ```go generate``` of ```shaders.go``` there are a lot "\r".

renlite:
The requested changes by @Jacalz have been completed.

internal/painter/gl/shaders.go
Jacalz: Hmm. You are right. That's strange and not entirely desirable. Are you on Windows by any chance?

andydotxyz: There are some weird encoding issues on windows consoles. Using the "fyne bundle" with "-o" instead of terminal redirect ">>" works around this.

andydotxyz (changes requested):
Looking really close thanks!

andydotxyz:
Latest version is working really well, thanks so much for working so hard on this!

Jacalz:
> We have tried to avoid manual bundle commands, preferring go:generate headers - have we missed some?

I cannot seem to answer the comment above but we do have go:generate headers. I ran `go generate` in that folder and it was also what I suggested @renlite  to do.

andydotxyz:
> > We have tried to avoid manual bundle commands, preferring go:generate headers - have we missed some?
> 
> 
> 
> I cannot seem to answer the comment above but we do have go:generate headers. I ran `go generate` in that folder and it was also what I suggested @renlite  to do.

Sorry my comment was posted out of order - I was meaning to reply to @renlite, also suggesting that the generate headers were preferable to manual commands.

canvas/rectangle.go
```diff
@@ -16,6 +16,7 @@ type Rectangle struct {
 	FillColor   color.Color // The rectangle fill color
 	StrokeColor color.Color // The rectangle stroke color
 	StrokeWidth float32     // The stroke width of the rectangle
+	Radius      float32     // The radius of the rectangle corners
```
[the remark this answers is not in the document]
renlite: Done.

renlite:
Is there still something to do from my side?

canvas/rectangle_test.go
```diff
@@ -23,3 +23,25 @@ func TestRectangle_FillColor(t *testing.T) {
 
 	assert.Equal(t, c, rect.FillColor)
 }
+
+func TestRectangle_Radius(t *testing.T) {
```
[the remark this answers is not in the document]
renlite: Sorry, I don't know what you mean with "visual" or "markup" test? I have created a visual example in /cmd/rect_test/main.go which I deleted after functional tests.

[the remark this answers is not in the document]
Jacalz: The problem with visual tests using only a main package is that they do not contribute to any test coverage reports and also can not be automatically run by our test suite. To instead test, on a reliable basis, that things are behaving correctly, you can use the [fyne.io/fyne/v2/test](https://pkg.go.dev/fyne.io/fyne/v2/test) package.

internal/painter/gl/shaders.go
```diff
@@ -8,40 +8,70 @@ import "fyne.io/fyne/v2"
 var shaderLineFrag = &fyne.StaticResource{
 	StaticName: "line.frag",
 	StaticContent: []byte(
-		"#version 110\n\nuniform vec4 color;\nuniform float lineWidth;\nuniform float feather;\n\nvarying vec2 delta;\n\nvoid main() {\n    float alpha = color.a;\n    float distance = length(delta);\n\n    if (feather == 0.0 || distance <= lineWidth - feather) {\n        gl_FragColor = color;\n    } else {\n        gl_FragColor = vec4(color.r, color.g, color.b, mix(color.a, 0.0, (distance - (lineWidth - feather)) / feather));\n    }\n}\n"),
+		"#version 110\r\n\r\nuniform vec4 color;\r\nuniform float lineWidth;\r\nuniform float feather;\r\n\r\nvarying vec2 delta;\r\n\r\nvoid main() {\r\n    float alpha = color.a;\r\n    float distance = length(delta);\r\n\r\n    if (feather == 0.0 || distance <= lineWidth - feather) {\r\n        gl_FragColor = color;\r\n    } else {\r\n        gl_FragColor = vec4(color.r, color.g, color.b, mix(color.a, 0.0, (distance - (lineWidth - feather)) / feather));\r\n    }\r\n}\r\n"),
```
Jacalz, line 11: I think you want to remove all the \r. I don't think they are desirable. Windows is probably causing the weird issues.

Jacalz (changes requested):
You are very close. Just a few notes.

canvas/rectangle_test.go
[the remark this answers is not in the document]
renlite: I thing the test ```func TestRectangle_FillColor(t *testing.T)``` is almost the same as the Raidus test.
I realy don't understand how to test the primitive draw type with the https://pkg.go.dev/fyne.io/fyne/v2/test. What would be the sense of such a test? It is not a widget to check the Text or to simulate an Input with Type(), or so. Do you have an example I can look at? What is a MarkupTest?
Or, could you create a test coding, I can include?

canvas/rectangle.go
```diff
@@ -16,6 +16,8 @@ type Rectangle struct {
 	FillColor   color.Color // The rectangle fill color
 	StrokeColor color.Color // The rectangle stroke color
 	StrokeWidth float32     // The stroke width of the rectangle
+	// Since: 2.4
+	Radius float32 // The radius of the rectangle corners
```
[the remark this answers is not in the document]
renlite: Done.

internal/painter/gl/shaders.go
renlite: Removed \r from shaders.go.

internal/painter/gl/draw.go
```diff
@@ -98,11 +97,102 @@ func (p *painter) drawRaster(img *canvas.Raster, pos fyne.Position, frame fyne.S
 }
 
 func (p *painter) drawRectangle(rect *canvas.Rectangle, pos fyne.Position, frame fyne.Size) {
-	if (rect.FillColor == color.Transparent || rect.FillColor == nil) && (rect.StrokeColor == color.Transparent || rect.StrokeColor == nil || rect.StrokeWidth == 0) {
-		return
```
Jacalz, line 102: Is there any specific reason as to why this was removed?

canvas/rectangle_test.go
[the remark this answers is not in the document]
Jacalz: You are right that that particular test isn't very helpful. The sense of the test is that it tries to render the object and then outputs either a markup of renderer contents or an image, depending on which function you use. What you want is basically to test that some input to a function gives the expected output, more or less. There is an example of using image tests in the text object: https://github.com/fyne-io/fyne/blob/8c2509518b2df442a6b748d9b07754739592e6d7/canvas/text_test.go#L192

internal/painter/gl/draw.go
renlite: I think it was the old func that was deleted and written from scratch for GL. The rectangle is used very often and this if-statement is called repeatedly during rendering and makes the execution slower. I would say the developer sees on the screen if the rectangle is visible and would change the code. Is there an advantage of this if-statement?

Jacalz: I think the intent is to have transparent rectangles (and similar) not be rendered at all and thus cost basically nothing. Let's check with the others before adding it back though.

canvas/rectangle_test.go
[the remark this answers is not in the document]
Bluebugs: Doesn't that rely on the software renderer actually and so doesn't cover this pr?

internal/painter/gl/draw.go
Bluebugs: An if statement like that is not going to cost anything, but indeed save unnecessary transfer to the gpu. It actually improve things.

renlite: I put the if statement a call stack higher and fixed double type assertion. Later a func could manage this color and stroke checks because the circle shape would need it too.
```Go
	case *canvas.Rectangle:
		if (obj.FillColor == color.Transparent || obj.FillColor == nil) && (obj.StrokeColor == color.Transparent || obj.StrokeColor == nil || obj.StrokeWidth == 0) {
			return
		}
		if obj.Radius == 0 {
			p.drawRectangle(obj, pos, frame)
		} else {
			p.drawRoundRectangle(obj, pos, frame)
		}
```

canvas/rectangle_test.go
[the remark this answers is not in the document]
renlite: Is there anything left for me to do here?

Jacalz (dismissed):
I don't know what to do about the test but LGTM apart from that :)

Bluebugs:
> I don't know what to do about the test but LGTM apart from that :)

I think the test should be part of the software painter change as a follow up issue as I think they will rely on it. If that's ok for you, let's create a follow up issue and merge this in.

Bluebugs (approved):

Bluebugs:
I have created #3547 for this followup task. If that's good for you @Jacalz, we just need @andydotxyz review before we can merge this.

Jacalz:
Sounds good to me. The only thing is that I think the point from @andydotxyz about the tests not testing anything stands though. If we can't test it without the software render, I personally think we can just remove the tests here but let's hear what Andy has to say first :)

andydotxyz:
> Sounds good to me. The only thing is that I think the point from @andydotxyz, about the tests not testing anything, still stands though. If we can't test it without the software render support, I personally think we can just remove the tests here but let's hear what Andy has to say first :)

Indeed there is a follow up task for the software renderer - but the markup renderer should be updated to understand this new field anyway. Not sure if we can use it in the test depending on circular imports

Jacalz:
What I meant was, should we keep the tests that were added in this PR? They only test that a field is set?

andydotxyz:
> Indeed there is a follow up task for the software renderer - but the markup renderer should be updated to understand this new field anyway. Not sure if we can use it in the test depending on circular imports

Apologies, I meant to provide a little more detail but was afk:

```diff
diff --git a/test/markup_renderer.go b/test/markup_renderer.go
index a47b706f..28e5901e 100644
--- a/test/markup_renderer.go
+++ b/test/markup_renderer.go
@@ -332,6 +332,7 @@ func (r *markupRenderer) writeRectangle(rct *canvas.Rectangle, attrs map[string]
        r.setColorAttr(attrs, "fillColor", rct.FillColor)
        r.setColorAttr(attrs, "strokeColor", rct.StrokeColor)
        r.setFloatAttr(attrs, "strokeWidth", float64(rct.StrokeWidth))
+       r.setFloatAttr(attrs, "translucency", rct.Radius)
        r.writeTag("rectangle", true, attrs)
 }
 
```

I think that the tests there should be replaced by something that actually tests - such as:

```go
	test.AssertObjectRendersToMarkup(t, "rounded_rect.xml", rect)
```

andydotxyz:
For example the test:

```go
func TestRectangle_Radius(t *testing.T) {
	rect := &canvas.Rectangle{
		FillColor:   color.NRGBA{R: 255, G: 200, B: 0, A: 180},
		StrokeColor: color.NRGBA{R: 255, G: 120, B: 0, A: 255},
		StrokeWidth: 2.0,
		Radius:      25}

	rect.Resize(fyne.NewSize(50, 50))
	test.AssertObjectRendersToMarkup(t, "rounded_rect.xml", rect)
}
```

renders


```xml
<canvas size="50x50">
	<content>
		<rectangle fillColor="rgba(255,200,0,180)" radius="25" size="50x50" strokeColor="rgba(255,120,0,255)" strokeWidth="2"/>
	</content>
</canvas>
```

andydotxyz:
Just to clarify, I too am very happy with what this PR is and it's working well - if we get texts fixed up I'll be happy to approve too

andydotxyz:
I added an example implementation of the software renderer to the issue @Bluebugs created in case anyone was curious / looking to add it.

renlite:
> For example the test:
> 
> ```go
> func TestRectangle_Radius(t *testing.T) {
> 	rect := &canvas.Rectangle{
> 		FillColor:   color.NRGBA{R: 255, G: 200, B: 0, A: 180},
> 		StrokeColor: color.NRGBA{R: 255, G: 120, B: 0, A: 255},
> 		StrokeWidth: 2.0,
> 		Radius:      25}
> 
> 	rect.Resize(fyne.NewSize(50, 50))
> 	test.AssertObjectRendersToMarkup(t, "rounded_rect.xml", rect)
> }
> ```
I will replace the new test funcs (I created) in [canvas/rectangle_test.go](https://github.com/fyne-io/fyne/pull/3539/files/48df130369665050982ebb739be3526e6e86ff23#diff-c6c57c0a99e493eb1ca09368417c3226cc2591f36169de149554ba399b57e000) with this test func.

renlite:
Go test fails but in testadata/failed/rounded_rect.xml is available? Radius is missing in XML.
```Shell
PS C:\_project\renlite\fyne> go test .\canvas\rectangle_test.go
--- FAIL: TestRectangle_Radius (0.00s)
    test.go:115: Master not found at C:\_project\renlite\fyne\canvas\testdata\rounded_rect.xml. Markup written to C:\_project\renlite\fyne\canvas\testdata\failed\rounded_rect.xml might be used as master.
FAIL
FAIL    command-line-arguments  0.426s
FAIL
```

```XML
<canvas size="50x50">
	<content>
		<rectangle fillColor="rgba(255,200,0,180)" size="50x50" strokeColor="rgba(255,120,0,255)" strokeWidth="2"/>
	</content>
</canvas>
```

andydotxyz:
> Go test fails but in testadata/failed/rounded_rect.xml is available? Radius is missing in XML.

You need to copy the file from "testdata/failed" to "testdata" and commit it.

But for the radius to be present you need to include my adjustment to the markup writer:
https://github.com/fyne-io/fyne/pull/3539#issuecomment-1374928955

renlite:
@andydotxyz Thank you for the infos. Test is working.

Bluebugs (approved):

andydotxyz (approved):
Amazing thanks!

Jacalz:
Well done @renlite! 🙂

renlite:
Thanks for the support by @Bluebugs and all the tests and good feedback by @andydotxyz @Jacalz @ErrorNoInternet.

PR #3540, opened 2023-01-05
Allow for doctabs.CreateTabs to take its time before returning a value.

Bluebugs:
### Description:
This is a likely use case of showing up a dialog.NewForm and wait for the result. The problem is that, at least on linux, the input thread is mixed with the rendering thread. So when a button is tapped and the interaction is blocked, it does block the main loop.

Another potential fix for this would be to make the Tapped callback happen in a goroutine too. I didn't go with that change as it did feel a bit bigger in scope, but it might make sense to actually spawn a goroutine for each Tapped callback as it might make this kind of bugs impossible.

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/container/doctabs.go
+++ b/container/doctabs.go
@@ -301,10 +301,12 @@ func (r *docTabsRenderer) buildAllTabsButton() (all *widget.Button) {
 func (r *docTabsRenderer) buildCreateTabsButton() *widget.Button {
 	create := widget.NewButton("", func() {
 		if f := r.docTabs.CreateTab; f != nil {
-			if tab := f(); tab != nil {
-				r.docTabs.Append(tab)
-				r.docTabs.SelectIndex(len(r.docTabs.Items) - 1)
-			}
+			go func() {
+				if tab := f(); tab != nil {
+					r.docTabs.Append(tab)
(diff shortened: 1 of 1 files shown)


Bluebugs:
Potential side effect of running event in a go routine as a general solution is the comment regarding QueueEvent that seems to exist to make event happen in order, but I don't see why it would not and a goroutine there will means an event would be processed in parallel to others instead of serially. Maybe a difference between OS.

andydotxyz:
> This is a likely use case of showing up a dialog.NewForm and wait for the result.

I don't understand how a change to `DocTabs` fixes an issue in `NewForm`?

> So when a button is tapped and the interaction is blocked, it does block the main loop.

This doesn't seem right - button taps don't run on the main thread, we have an event queue.

> Another potential fix for this would be to make the Tapped callback happen in a goroutine too. I didn't go with that change as it did feel a bit bigger in scope, but it might make sense to actually spawn a goroutine for each Tapped callback as it might make this kind of bugs impossible.

We cannot do that as events must be processed in the order that the user performed them. We previously used coroutines and removed it for that reason.

PR #3546, opened 2023-01-08
Fix #3459 by changing srcdir to it's absolute path

### LINKED ISSUE #3459: cmd/fyne package can't find FyneApp.toml when -src option has given

hismailbulut:
### Description:

Using relative paths with os.Chdir brokes joined file paths, to prevent this we can just use absolute filepaths without changing anything other

Fixes #3459 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne/internal/commands/build.go
+++ b/cmd/fyne/internal/commands/build.go
@@ -90,6 +90,13 @@ func Build() *cli.Command {
 // Build parse the tags and start building
 func (b *Builder) Build() error {
 	if b.srcdir != "" {
+		// Use absolute srcdir
+		if !filepath.IsAbs(b.srcdir) {
+			absSrcDir, err := filepath.Abs(b.srcdir)
+			if err == nil {
+				b.srcdir = absSrcDir
+			}
+		}
 		dirStat, err := os.Stat(b.srcdir)
 		if err != nil {
 			return err
@@ -297,6 +304,10 @@ func createMetadataInitFile(srcdir string, app *appData) (func(), error) {
 
 	app.ResGoString = "nil"
 	if app.icon != "" {
+		// Icon path should relative to FyneApp.toml and it is in the srcdir
+		if !filepath.IsAbs(app.icon) {
+			app.icon = filepath.Join(srcdir, app.icon)
+		}
 		res, err := fyne.LoadResourceFromPath(app.icon)
 		if err != nil {
 			fyne.LogError("Unable to load medadata icon file "+app.icon, err)
--- a/cmd/fyne/internal/commands/build.go
+++ b/cmd/fyne/internal/commands/build.go
@@ -287,7 +287,7 @@ func createMetadataInitFile(srcdir string, app *appData) (func(), error) {
 	data, err := metadata.LoadStandard(srcdir)
 	if err == nil {
 		// When icon path specified in metadata file, we should make it relative to metadata file
-		data.Details.Icon = util.MakePathRelative(srcdir, data.Details.Icon)
+		data.Details.Icon = util.MakePathRelativeTo(srcdir, data.Details.Icon)
 
 		app.mergeMetadata(data)
 	}
--- a/cmd/fyne/internal/commands/build.go
+++ b/cmd/fyne/internal/commands/build.go
@@ -90,13 +90,7 @@ func Build() *cli.Command {
 // Build parse the tags and start building
 func (b *Builder) Build() error {
 	if b.srcdir != "" {
-		// Use absolute srcdir
-		if !filepath.IsAbs(b.srcdir) {
-			absSrcDir, err := filepath.Abs(b.srcdir)
-			if err == nil {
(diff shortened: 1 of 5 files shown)


cmd/fyne/internal/commands/build.go
```diff
@@ -90,6 +90,13 @@ func Build() *cli.Command {
 // Build parse the tags and start building
 func (b *Builder) Build() error {
 	if b.srcdir != "" {
+		// Use absolute srcdir
```
andydotxyz, line 93: It would be better to check and correct `srcdir` earlier so that it is consistent (validate etc can also use this).
Build will not always be called so later stages will not have this correction run.
It may also avoid the duplication in the PR.

andydotxyz (changes requested):
I'd rather not mutate the configuration during the build phase, this seems prone to complications.

cmd/fyne/internal/commands/build.go
hismailbulut: Builder.Build function is the first thing called when user starts fyne build and only called by this command. I don't see anywhere earlier than this. Other commands just calls underlying build function. Thats why I have to add it to Package.validate because it seems release and install commands also calling it and we don't need to add another correction for them.

Bluebugs: Yes, we do have `fyne build` command that try to mimick the `go build` command.

andydotxyz: > Builder.Build function is the first thing called when user starts fyne build and only called by this command. I don't see anywhere earlier than this. Other commands just calls underlying build function

What I meant is that this may not be called. If the exe specified already exists then it won't try to build.

cmd/fyne/internal/util/file.go
```diff
@@ -39,6 +39,31 @@ func EnsureSubDir(parent, name string) string {
 	return path
 }
 
+// EnsureAbsPath returns the absolute path of the given file if it is not already absolute
+func EnsureAbsPath(path string) string {
+	if !filepath.IsAbs(path) {
```
andydotxyz, line 44: We prefer to exit early when we can instead of having method contents indented in long ifs.
i.e.
```go
if file path.IsAbs(path) {
    return path
}
```

cmd/fyne/internal/util/file.go
```diff
@@ -39,6 +39,31 @@ func EnsureSubDir(parent, name string) string {
...
+			return abs
+		}
+	}
+	return path
+}
+
+// MakePathRelative joins root with path if path is not absolute and exists in root
+// both root and path must non-empty, otherwise path will be returned
+func MakePathRelative(root, path string) string {
+	if root != "" && path != "" && !filepath.IsAbs(path) {
```
andydotxyz, line 58: As above, the inverse of this check is a chance to return early, then the rest of the method at the top-level is the purpose of the code

cmd/fyne/internal/util/file.go
```diff
@@ -39,6 +39,31 @@ func EnsureSubDir(parent, name string) string {
...
+		} else {
+			return abs
+		}
+	}
+	return path
+}
+
+// MakePathRelative joins root with path if path is not absolute and exists in root
+// both root and path must non-empty, otherwise path will be returned
+func MakePathRelative(root, path string) string {
```
andydotxyz, line 57: This name seems misleading - the path returned is absolute.
Perhaps you meant `MakePathRelativeTo`?

andydotxyz (changes requested):
Thanks, this reads cleanly - just some comments about style

andydotxyz (approved):
Thanks

PR #3561, opened 2023-01-12
Update Linux/XDG default theme selection

lucor:
### Description:
This commit update the Linux/XDG default theme selection as per https://github.com/flatpak/xdg-desktop-portal/blob/1.16.0/data/org.freedesktop.impl.portal.Settings.xml#L32-L46

Relates to: #2657 #3414
### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -58,8 +58,12 @@ func findFreedestktopColorScheme() fyne.ThemeVariant {
 		return theme.VariantDark
 	}
 
+	// See: https://github.com/flatpak/xdg-desktop-portal/blob/1.16.0/data/org.freedesktop.impl.portal.Settings.xml#L32-L46
+	// 0: No preference
+	// 1: Prefer dark appearance
+	// 2: Prefer light appearance
 	switch value {
-	case 0:
+	case 2:
 		return theme.VariantLight
 	default:
 		return theme.VariantDark
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -62,12 +62,15 @@ func findFreedestktopColorScheme() fyne.ThemeVariant {
 	// 0: No preference
 	// 1: Prefer dark appearance
 	// 2: Prefer light appearance
-	if value == 1 {
+	switch value {
+	case 2:
+		return theme.VariantLight
+	case 1:
 		return theme.VariantDark
+	default:
+		// Default to light theme to support Gnome's default see https://github.com/fyne-io/fyne/pull/3561
+		return theme.VariantLight
 	}
-
-	// Default to light theme for 0, 2 and unknown values to support Gnome's default see https://github.com/fyne-io/fyne/pull/3561
-	return theme.VariantLight
 }
 
 func (a *fyneApp) SendNotification(n *fyne.Notification) {
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -62,15 +62,12 @@ func findFreedestktopColorScheme() fyne.ThemeVariant {
 	// 0: No preference
 	// 1: Prefer dark appearance
 	// 2: Prefer light appearance
-	switch value {
-	case 2:
-		return theme.VariantLight
-	case 1:
+	if value == 1 {
 		return theme.VariantDark
-	default:
-		// Default to light theme to support Gnome's default see https://github.com/fyne-io/fyne/pull/3561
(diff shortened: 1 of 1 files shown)


Jacalz (changes requested):
It seems to me that the previous behavior was doing the right thing (or Gnome is not following the protocol correctly) because this change makes the switch between dark/light not work anymore for me.

I would suggest that we use light theme for both 0 and 2 instead.

andydotxyz:
But previously dark was our default if nothing was specified outside of this lookup...

Jacalz:
> But previously dark was our default if nothing was specified outside of this lookup...

Yes, but what's the point of having this theme switching support if it isn't possible to switch between light and dark? With this change, the application is always dark no matter what you do.

We have already passed the stage where we know that the desktop supports the protocol so those that do not will still default to dark.

app/app_xdg.go
```diff
@@ -58,8 +58,12 @@ func findFreedestktopColorScheme() fyne.ThemeVariant {
...
+	// See: https://github.com/flatpak/xdg-desktop-portal/blob/1.16.0/data/org.freedesktop.impl.portal.Settings.xml#L32-L46
+	// 0: No preference
+	// 1: Prefer dark appearance
+	// 2: Prefer light appearance
 	switch value {
-	case 0:
+	case 2:
 		return theme.VariantLight
 	default:
 		return theme.VariantDark
```
Jacalz, line 69: I'd suggest that we remove the switch and just do it like this:
```go
if value == 1 {
    return theme.VariantDark
}

return theme.VariantLight
```

This should make sure that we handle 2 for light and also support switching on Gnome.

andydotxyz: Yes I guess this is probably the best we can do - but I feel it would be good to know if we are working around a GNOME bug (that will be resolved) or if the implement this standard only partially?

Jacalz: I honestly don't know but I wouldn't want to break support for one of the most used/popular desktops either way.

Jacalz: Since this has been the behaviours since Gnome 42, I would think it is their desired behaviour, but I honestly don't know. Wouldn't want to break support on one of the most popular desktops though.

lucor: Indeed, it depends what theme we want show for the default value `0`.
It looks like gnome `42` will use the light one where no preference as set (i.e. fresh install), so the suggestion above makes sense.
Tested on Ubuntu 22.04 (Gnome 42.5) and works as expected in both cases.
It seems the only thing we need to decide is which theme apply as default when the portal preference is supported and if the default theme in case of error is the same or always black.

Jacalz: I am using Gnome 43 and it behaves just like 42. I'd say that we default to dark in the case of an error as it would be most like how we do on other platforms.

lucor: Make sense, updated in 0dffef90e81abc620eb1eb7ba8b49fa7b709204b to support Gnome's default

andydotxyz:
Thanks for getting a good solution to this one

andydotxyz (dismissed):

app/app_xdg.go
```diff
@@ -58,13 +58,19 @@ func findFreedestktopColorScheme() fyne.ThemeVariant {
...
 	}
 
+	// See: https://github.com/flatpak/xdg-desktop-portal/blob/1.16.0/data/org.freedesktop.impl.portal.Settings.xml#L32-L46
+	// 0: No preference
+	// 1: Prefer dark appearance
+	// 2: Prefer light appearance
 	switch value {
```
Jacalz, line 65: Would it not be better to just keep this to one case? Seems a bit unnecessary with three.
```go
if value == 1 {
    return theme.VariantDark
}

// Default to light theme for 0 and 2 to support Gnome's default see https://github.com/fyne-io/fyne/pull/3561
return theme.VariantLight
```

andydotxyz: To me it seems good to keep them separate due to the conflicting documentation linked in the method

lucor: I left the switch to have the code document directly the values that we handle here for `value`
Updated in dc029901192007b3df31bcc09a4670af2806ea69

Jacalz (approved):

andydotxyz (approved):
Thanks

PR #3563, opened 2023-01-12
Bugs/mimicking go build

Bluebugs:
### Description:
`fyne build` try to mimick `go build`. It was missing a proper handling of the case when we need to generate the metadata in a subdirectory. This cover this case.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.


Bluebugs:
Note: we do need this patch in a release before we can make a release of fyne-cross as this break using fyne-cross when building a subdirectory.

cmd/fyne/internal/commands/build.go
```diff
@@ -282,6 +287,25 @@ func (b *Builder) build() error {
 	return err
 }
 
+func (b *Builder) computeSrcDir(fyneGoModRunner runner) (string, error) {
+	if b.goPackage == "" {
+		return b.srcdir, nil
+	}
+
+	srcdir := b.srcdir
+	if strings.HasPrefix(b.goPackage, "./") {
```
andydotxyz, line 296: This doesn't seem right? `./` is relative to the current directory rather than a declared source directory isn't it?

cmd/fyne/internal/commands/build.go
```diff
@@ -282,6 +287,25 @@ func (b *Builder) build() error {
...
+		srcdir = filepath.Join(srcdir, b.goPackage)
+	} else {
+		// get the output of go env GOROOT
+		goroot, err := fyneGoModRunner.runOutput("env", "GOROOT")
```
andydotxyz, line 300: Why GOROOT? Isn't code stored in GOPATH?

cmd/fyne/internal/commands/build.go
andydotxyz: Also what about the "." case? That is simply cwd.

cmd/fyne/internal/commands/build.go
```diff
@@ -197,14 +197,19 @@ func (b *Builder) build() error {
 		}
 	}
 
+	srcdir, err := b.computeSrcDir(fyneGoModRunner)
```
andydotxyz, line 200: Is there a reason this is not fixed earlier in the build validation? Not sure about the possible impact of using a difference srcdir for build compared to other phases...

andydotxyz (commented):
Lots of new logic here - not sure I follow it all.

If we want to match how go compiler handles it can we use their APIs instead of trying to work it out?

cmd/fyne/internal/commands/build.go
Bluebugs: go build take a module path. It can be relative, in which case it start in the current location or it will look up in GOROOT for it.

Bluebugs: > If we want to match how go compiler handles it can we use their APIs instead of trying to work it out?

I don't see any API for the purpose of finding where go build is going to look where the source it is asked to build are. Did you have one in mind?

cmd/fyne/internal/commands/build.go
Bluebugs: go build something is looking up in GOROOT for me.

cmd/fyne/internal/commands/build.go
Bluebugs: The reason is that once we call go build, it should build the binary where we expect too. The problem fixed here is to locate where the Icon and the metadata should be as that depend on where go build is going to look for the source code.

cmd/fyne/internal/commands/build.go
andydotxyz: GOROOT is where the runtime and standard library are stored, not where user code goes

andydotxyz (approved):

PR #3566, opened 2023-01-13
Fix rounded rectangle with no stroke

### LINKED ISSUE #3565: Rounded rectangles blank if no stroke
### Checklist

### Describe the bug

Rounded rectangles only work if stroke colour is set and > 0 `StrokeWidth`.

### How to reproduce

In fyne_demo `canvas` tutorial set `StrokeWidth` to 0 on the demo and see the rectangle disappear.

### Screenshots

_No response_

### Example code

```go
rect.Radius = 5
rect.StrokeWidth = 0
```

### Fyne version

fe77710a1d

### Go compiler version

1.18

### Operating system

macOS

### Operating system version

all

### Additional Information

_No response_

andydotxyz:
This fixes the bug and also a hidden graphical glitch that was hiding packed next to the stroke colour, it blended through white or black to transparent so created outlines:

Screenshot 2023-01-13 at 15 36 48

Its looks like blending to transparent is not working, instead it is picking up the `r, g, b` even though alpha is `0.0`.

Fixes #3565

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


internal/painter/gl/shaders/round_rectangle.frag
```diff
@@ -21,29 +19,30 @@ float calc_distance(vec2 p, vec2 b, float r)
...
     {
-        if (distance < 0.0)
-        {
-            to_color = fill_color;   
-        } 
-        
-        distance = abs(distance) - stroke_width_half;
+        from_color = fill_color;
+        to_color = fill_color;
+        to_color[3] = 0.0; // blend the fill colour to alpha
```
Bluebugs, line 34: It seems that in all case we need to set to_color[3] to zero, in which case it would be good to move it outside and after the if.

Bluebugs (changes requested):
Can you update the _es variant of the shader too?

internal/painter/gl/shaders/round_rectangle.frag
andydotxyz: Not in all cases, no - only 2 out of 3. When the fill is being drawn we want the original colour alpha

Bluebugs: The code for the shader look like this after this patch:
```
    to_color[3] = 0.0; // blend the stroke colour to alpha

    if (stroke_width_half == 0.0)
    {
        from_color = fill_color;
        to_color = fill_color;
        to_color[3] = 0.0; // blend the fill colour to alpha
    }
```

If you move to_color[3] = 0.0, I think it does exactly the same things as now.
```
    if (stroke_width_half == 0.0)
    {
        from_color = fill_color;
        to_color = fill_color;
    }
    to_color[3] = 0.0; // blend the stroke or fill colour to alpha

```

andydotxyz: Ah those to, before the third case. I see yes

Bluebugs (approved):

PR #3568, opened 2023-01-13
Make sure that all the build put their result in p.dir.

Bluebugs:
### Description:
All build when building a subdir using -src should output there content in the current directory. Windows and Linux target where not following that requirement and are corrected with this patch.

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.


andydotxyz (approved):
Nice catches

Issue #2147, opened 2021-04-07
Dynamic Menus


Serverfrog:
### Is your feature request related to a problem? Please describe:
We have the Possibility to use Bindings to change dynamically the, except the Menu and MenuItem's.
These have only have a initial Value for Label's and after that no Possibility to change without recreating the Structure of the Application (Window or restarting the application)

### Is it possible to construct a solution with the existing API?
I'm not sure how in the back end it is possible to rerender Item's but due to the fact that the widget.* Components have the possibility it should not be impossible to do.

### Describe the solution you'd like to see:
This could be in two ways.

- The "WidgetWay" with `fyne.NewMenuItem()` and `fyne.NewMenuItemWithData`
- a Method where someone can Change the Label and its rerender then handled there `SetTitle()`

andydotxyz:
I think a lower impact change would be to use the current structures but to support re-reading the changes.
Perhaps `fyne.Menu.Refresh()` would do this, but we have to reflect in our internal menu as well as the macOS native ones.

PR #3596, opened 2023-01-22
Allow removal of the Quit entry in the Taskbar Menu.

EssGeeEich:
### Description:
If developers wish to use the Taskbar Menu, as it stands today, the code forces you to have a "Quit" entry in said menu.

The string is also hardcoded.
This simple limitation makes it extremely frustrating (or downright impossible!) to localize a software.

While the correct approach would be to fully remove the "Quit" entry without looking back, such a thing would break the tutorial code and its YouTube video.

This is why I opted to add a public member to the Menu struct - which lets developers control whether the "Quit" entry should exist or not.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -162,4 +162,5 @@ func newAppWithDriver(d fyne.Driver, id string) fyne.App {
 type systrayDriver interface {
 	SetSystemTrayMenu(*fyne.Menu)
 	SetSystemTrayIcon(resource fyne.Resource)
+	SetSystemTrayQuitText(title string, tooltip string)
 }
--- a/app/app_desktop_darwin.go
+++ b/app/app_desktop_darwin.go
@@ -37,6 +37,13 @@ func (a *fyneApp) SetSystemTrayIcon(icon fyne.Resource) {
 	a.Driver().(systrayDriver).SetSystemTrayIcon(icon)
 }
 
+// SetSystemTrayQuitText sets the title and tooltip strings for the Quit option that's available by default on a SystemTray Menu.
+func (a *fyneApp) SetSystemTrayQuitText(title string, tooltip string) {
+	if desk, ok := a.Driver().(systrayDriver); ok {
+		desk.SetSystemTrayQuitText(title, tooltip)
+	}
+}
+
 func defaultVariant() fyne.ThemeVariant {
 	if C.isDarkMode() {
 		return theme.VariantDark
--- a/app/app_windows.go
+++ b/app/app_windows.go
@@ -95,6 +95,11 @@ func (a *fyneApp) SetSystemTrayIcon(icon fyne.Resource) {
 	a.Driver().(systrayDriver).SetSystemTrayIcon(icon)
 }
 
+// SetSystemTrayQuitText sets the title and tooltip strings for the Quit option that's available by default on a SystemTray Menu.
+func (a *fyneApp) SetSystemTrayQuitText(title string, tooltip string) {
+	a.Driver().(systrayDriver).SetSystemTrayQuitText(title, tooltip)
+}
+
 func escapeNotificationString(in string) string {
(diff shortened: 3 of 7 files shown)


andydotxyz:
This doesn't feel like the right fix. You have applied a new API to the `Menu` type which is used in many places (main menu, pop-up menu etc) but used in only one place.
I would recommend considering how we have approached this in the main-menu code where there is a similar Quit item.

I don't know if your aim is translation or removal for other reasons. But for translation we should do it properly with full localisation in a future release. If the aim is to just hide the ability to quit then be aware that we do not support this feature, as the OS defined lifecycle of application software should be respected.

EssGeeEich:
Hello @andydotxyz,
If I understand correctly, there is no contribution that I can do at the moment to change how the things are right now.

Still, after reading your reasoning I came to the conclusion that the less conflicting option would be to allow the developer to simply change the Quit option's text / tooltip.

A revert plus a proper commit have been pushed as of now.
This is out of personal curiosity - it is not fully clear to me if this issue will eventually be tackled sooner or later, or if my pull request will be considered from the team.

If you are sure that you want to tackle this issue differently, then I suppose you should close this pull request without a merge, but at the moment I need a way to localize my program and I believe that this is the closest thing I can write without derailing my original project and I don't mind sharing my solution with everyone else.

After the last two commits, you'd be able to change the Quit text as follows:

```
a := app.New()

	menu := fyne.NewMenu("GoTalk")
	desk.SetSystemTrayMenu(menu)
	desk.SetSystemTrayQuitText("Esci", "Chiude il programma")
}
```

You can also change the Quit Text before creating the menu:

```
a := app.New()

	menu := fyne.NewMenu("GoTalk")
	desk.SetSystemTrayQuitText("Esci", "Chiude il programma")
	desk.SetSystemTrayMenu(menu)
}
```

andydotxyz:
I appreciate that there is no way to do this without a patch at this time - but my hesitance is that anything added in the API to make it work now will have to remain for years - so it is important that we find a solution that is not working around the short term needs.
The current API is indeed more focused on the need, but once localisation is in place it will be unnecessary.
Note also that if this is to be fixed we should probably find a solution that works for the main menu as well because it suffers the same problem.

An open question in the past has been is there a way to mark a menu item as having a certain characteristic (like "Quit") without just comparing the text (what we do at the moment, which breaks as you see)?
Perhaps if we do that then apps could provide a quit menu item that has different text, and that would be folded in to the localisation solution in the future without APIs that need cleaned up (as a "quit" menu item carries meaning beyond the text content).
This also relates to an issue requesting a special edit menu area (#2147) so if we get it right there may be a wider benefit.

I hope this makes sense, sorry it is more complex than you would like for what seems like a simple fix.

The alternative is that we get moving with translations sooner rather than later and get them in to 2.4 - which I think we would be open to, but is a lot more work.

EssGeeEich:
I suppose that for the time being I will close this PR.
I might open an issue in relation to the translation side - maybe every now and then I could contribute something.

andydotxyz:
Apologies that this didn't deliver what you were looking for. We could leave it open as a request for translating the default menu items? We hope to be working in the next few months.

EssGeeEich:
No need to apologize!
I don't mind helping in my free time.
If translation is something that needs some work on, I might as well (try to) contribute something!

andydotxyz:
Sweet thanks

leessmin:
Seeing your discussions, I would like to offer my opinion. I think it's better to have an API that can set the text, because sometimes I don't need quit but exit. That's my opinion

andydotxyz:
With translations in 2.5 that should be possible. It's not the same as removing the Quit option, which the OP asked for.

leessmin:
Thank you Andy for your answer, I hope 2.5 will come soon, and thank you fyne developers for your hard work

PR #3606, opened 2023-01-25
We should only correct icon path if it is specified in metadata file

hismailbulut:
### Description:

This will fix a mistake previosly made by me in #3546 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz (approved):
Thanks

PR #3613, opened 2023-01-29
dialog: Remove unnecessary refresh call from fileitem

Jacalz:
### Description:

This PR  removes an unnecessary refresh call that the fileitem was doing each time it was refreshed. The background will be refreshed automatically by the `Show()/Hide()` calls and need not be refreshed once more.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/fileitem.go
+++ b/dialog/fileitem.go
@@ -48,8 +48,7 @@ func (i *fileDialogItem) Tapped(_ *fyne.PointEvent) {
 }
 
 func (i *fileDialogItem) CreateRenderer() fyne.WidgetRenderer {
-	background := canvas.NewRectangle(theme.PrimaryColor())
-	background.Hide()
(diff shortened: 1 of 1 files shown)


andydotxyz (commented):
Are you certain about this?
`Show` will not refresh if it is already visible.
This means if you change the fill colour and just call `Show` instead of refresh it won't change...

Now we have a faster rectangle it may be desirable to use `color.Transparent` instead of `Hide` which would allow a removal of the double refresh in a different way.

Jacalz:
Good point. This PR worked fine with my testing but your suggestion seems better. Will implement that :)

Jacalz:
You were right. That was a much nicer and better option :D

andydotxyz (approved):

PR #3632, opened 2023-02-06
Clear render cache when software rendering

elliot-aeroqual:
### Description:
Each time testcanvas.Refresh is called we tell the render cache that it can be cleaned.
This gives the render cache a chance to be cleared when using the test canvas for software rendering.
This should have no impact on most users.

There isn't an open issue for this but it is essentially the same as [2069](https://github.com/fyne-io/fyne/issues/2069)

### Checklist:

- [x] Tests all pass

### DIFF
--- a/test/testcanvas.go
+++ b/test/testcanvas.go
@@ -9,6 +9,7 @@ import (
 	"fyne.io/fyne/v2/driver/desktop"
 	"fyne.io/fyne/v2/internal"
 	"fyne.io/fyne/v2/internal/app"
+	"fyne.io/fyne/v2/internal/cache"
 	"fyne.io/fyne/v2/theme"
 )
 
@@ -88,6 +89,7 @@ func NewTransparentCanvasWithPainter(painter SoftwarePainter) WindowlessCanvas {
 }
 
 func (c *testCanvas) Capture() image.Image {
+	cache.Clean(true)
 	bounds := image.Rect(0, 0, internal.ScaleInt(c, c.Size().Width), internal.ScaleInt(c, c.Size().Height))
 	img := image.NewNRGBA(bounds)
 	if !c.transparent {


andydotxyz:
Thanks for the contribution - but this is surely too fast? The refresh call here is not called once per paint but once per object - so this could be called hundreds of times per draw frame.
Calling that fast seems like a bad idea - what problem are you improving?

elliot-aeroqual:
> Thanks for the contribution - but this is surely too fast? The refresh call here is not called once per paint but once per object - so this could be called hundreds of times per draw frame. Calling that fast seems like a bad idea - what problem are you improving?

Yes it is called more than required but the cache.Clean returns early almost all of the time at least.
The other place I considered putting it was in testcanvas.Capture().

You might not remember but I discussed this with you briefly on Discord. The problem is that if you are using software rendering then the cache is never cleared. This becomes a problem if you use the List widget at least because the cache holds references to widgets (list items) that would otherwise be GC'd.

I considered making an exported function for clearing the cache instead but that also doesn't feel right. 
Anyway I'm happy to have any other suggestions because I'm also not that fond of this change.

elliot-aeroqual:
> > Thanks for the contribution - but this is surely too fast? The refresh call here is not called once per paint but once per object - so this could be called hundreds of times per draw frame. Calling that fast seems like a bad idea - what problem are you improving?
> 
> Yes it is called more than required but the cache.Clean returns early almost all of the time at least. The other place I considered putting it was in testcanvas.Capture().
> 
> You might not remember but I discussed this with you briefly on Discord. The problem is that if you are using software rendering then the cache is never cleared. This becomes a problem if you use the List widget at least because the cache holds references to widgets (list items) that would otherwise be GC'd.
> 
> I considered making an exported function for clearing the cache instead but that also doesn't feel right. Anyway I'm happy to have any other suggestions because I'm also not that fond of this change.

I was thinking about this last night and I can't remember why I thought putting it in testcanvas.Capture() was a worse option so I'm thinking about moving it there now if that would be better @andydotxyz .

andydotxyz:
Thanks for this.
I'm still not sure - if you clear the cache before drawing every time then the cache is surely useless?
What problem is this PR solving?

elliot-aeroqual:
> Thanks for this. I'm still not sure - if you clear the cache before drawing every time then the cache is surely useless? What problem is this PR solving?

The issue is [2069](https://github.com/fyne-io/fyne/issues/2069) which was addressed in the mobile and glfw drivers [here](https://github.com/fyne-io/fyne/pull/2101). This is to try to address that same issue but for the software driver.

My understanding of the cache clearing is it is only removing items that haven't been used for > 1 minute so clearing it before drawing makes no difference if you are redrawing the same widgets continuously. But if you change the content on the canvas then after 1 minute the renderers for content that is no longer being drawn will be cleared.

andydotxyz:
Ah apologies you are quite right - I had mistaken it for a cache reset!

andydotxyz (approved):
Thanks

elliot-aeroqual:
> Thanks
👍
Thanks for Fyne

PR #3633, opened 2023-02-06
Use new font render from go-text

### LINKED ISSUE #3401: Drastic increase in memory usage on develop
### Checklist

### Describe the bug

I noted during some testing that the memory usage seems to have roughly doubled between v2.2.4 and develop (to be v2.3.0). I initially launched `fyne_demo` and it had a 1.6x increase but switching the view resulted in it becoming a 2x increase:

Screenshot from 2022-11-12 23-06-26

I observed a very similar 2x increase in Rymdport but at initial start without pressing anything.

### How to reproduce

1. Open `fyne_demo` on master branch and on develop branch
2. Open a system monitor and observe the memory usage

### Screenshots

See screenshots inline above ^

### Example code

It does not really matter what code is being used. The demo reproduces the same problem.

### Fyne version

4c67d93aa35be8646ba9c4354acb0492d5115f4d

### Go compiler version

1.18.7

### Operating system

Linux

### Operating system version

Fedora 36

### Additional Information

_No response_

andydotxyz:
Opening work of moving to go-text - for discussion at this stage.
Some code tidying to be done and I don't think the vertical align is quite right.
Test failures will be addressed when it gets close as its mostly golden images.

Fixes #3401 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

This represents a big improvement in memory usage, though a small slow-down.
A further PR will be addressing the level at which we cache text draw which should easily make up for the draw speed regression. Or improvements to go-text/render may as well!

golang.org/x/text: 13M app, uses 102M -> 164M RAM

2023/01/26 11:28:53 Initial: 8.640313ms
2023/01/26 11:28:53 Refresh average 7.850099ms

go-text/render: 13M app, uses 94M -> 111M RAM

2023/01/26 18:21:22 Initial: 10.654775ms
2023/01/26 18:21:22 Refresh average 10.242469ms


internal/painter/font_test.go
```diff
@@ -25,11 +26,12 @@ func TestCachedFontFace(t *testing.T) {
...
-			for _, r := range tt.wantGlyphs {
-				_, ok := got.GlyphAdvance(r)
-				assert.True(t, ok, "symbol font should include: %c", r)
-			}
+			_ = painter.CachedFontFace(tt.style, 14, 1)
+			// TODO fix this and remove rule / glyph ambiguity
+			//for _, r := range tt.wantGlyphs {
+			//	_, ok := got.Font.GlyphAdvance(r)
+			//	assert.True(t, ok, "symbol Font should include: %c", r)
+			//}
```
Jacalz, line 34: Is this intended to remain?

Jacalz (changes requested):
This is looking really well. The results for Rymdport as follows:

Before: 58MB
After: 46MB

It is also worth noting that memory usage also seems to have stabilized a lot since v2.3.1. With resizing the window a lot for a while, I am seeing this:

Before: 90MB
After: 53MB

It is climbing a lot less. I don't know if it was a one time thing or just my system being weird (not entirely scientific tests either) but this seems like a huge win :)

internal/painter/font_test.go
andydotxyz: A good question. I can't quite figure out if there is an equivalent test with the new setup...

andydotxyz:
Apologies for the force-push. The merge with develop appears to be creating some very strange test failures which I can't quite figure out

Jacalz (approved):
Well done. This is already a huge improvement for memory usage and I can't wait to see more improvements further down the road :)

PR #3637, opened 2023-02-07
expose ErrResourceNotFound

### LINKED ISSUE #3353: app.Storage.Open() to return a known error when file not exist
### Checklist

similar to #3083 , I suggest Open() to use an exported error for users to compare.
Android. fyne 2.2.3
Thank you.

Currently Open() returns "resource not found at URI", which is hard to compare for users.

[internal/driver/mobile/file_android.go](https://github.com/fyne-io/fyne/blob/a83492533a2c9f56dc3e7e1d100ca9b4216edfb6/internal/driver/mobile/file_android.go)
`return nil, errors.New("resource not found at URI")`
to return some error we can match to

LilJohny:
### Description:

- Added ErrResourceNotFound
- Used it in file_android.go in all place relevant
- Added tests to check new changes

Fixes #3353 

### Checklist:

- [X] Tests included.
- [ ] Tests all pass.


andydotxyz:
Can you please address the static check issues @LilJohny ?

LilJohny:
@andydotxyz working on it

LilJohny:
@andydotxyz Should be fixed now

andydotxyz (changes requested):
Thanks for this, a few issues I think:

* the file file_android_test.go should not be in the `test` package but in the `internal/driver/mobile` surely?
* The public error is not actually public (buried in the android driver instead of `storage` package.
* It only applies in the case of android, should we not add it to the generic file storage system as well? (internal/repository/file.go)

Lastly I am not sure if this should be distinct from `storage.ErrNotExists` but open to discussion on that one...

LilJohny:
@andydotxyz Thanks for the feedback! Sorry about the mistakes, I'll fix the file placement and make the error public. Thanks again for pointing this out!

LilJohny:
But on using the storage error I can't agree – they have error message document not found, which is not exactly what need here as I see it.

LilJohny:
And regarding internal/repository/file.go – it doesn't contain errors.New(...) , in case of errors it returns nils, or forwards the result of os. methods. If you are sure, I can update it to use this new error too.

LilJohny:
And sorry for child errors, I am relatively new to go, now I figured how to run the pipelines in my fork, so such problems won't be repeated.

andydotxyz:
> And regarding internal/repository/file.go – it doesn't contain errors.New(...) , in case of errors it returns nils, or forwards the result of os. methods. If you are sure, I can update it to use this new error too.

It is a little more work. The key thing here is consistency - the point of returning a known error is that it will always be used for certain types of error. But as it exists now it only represents an android resource not found. But the method is clearly not android specific as the abstraction works for all OS.

LilJohny:
Okay, I am happy to work on this too

andydotxyz:
> Okay, I am happy to work on this too

Sweet, thanks so much

LilJohny:
@andydotxyz Please take a look. I implemented two errors – platform independent and android specific. Also android specific error is platform independent, but not other way around.

andydotxyz:
Thanks, I'm not sure why we need an android specific error?
Earlier I was intending to suggest the error can be used to represent that same error case for other OSes, rather than being Android specific. That way we keep the abstraction and users don't have to worry about platform errors or checks.

LilJohny:
My point here is that Android resource is different abstraction compared to file on Unix systems, and it has some characteristic limitations. But if you persist we can go with single error for platform-specific and platform-independent code.

andydotxyz:
> Android resource is different abstraction compared to file on Unix

I don't understand. The point of our abstraction is so that developers don't have to write platform specific code - it won't be great if we had different abstractions for each platform (it's not really an abstraction at that point).

LilJohny:
Okay, I`ll use same object.

Jacalz:

PR #3638, opened 2023-02-07
Correct all version used while building application to ensure Windows…

Bluebugs:
### Description:
Compiling for Windows with a short AppVersion, like 1.0, lead to the version not being consistent from Windows perspective and prevent the starting of the application on up to date Windows.

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/package.go
+++ b/cmd/fyne/internal/commands/package.go
@@ -259,7 +259,13 @@ func (p *Packager) buildPackage(runner runner) ([]string, error) {
 }
 
 func (p *Packager) combinedVersion() string {
-	return fmt.Sprintf("%s.%d", p.AppVersion, p.AppBuild)
+	versions := strings.Split(p.AppVersion, ".")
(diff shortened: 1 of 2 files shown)


Bluebugs:
@Jacalz This is why you do not get a working binary on Windows at the moment when fyne-cross use fyne to build. We would want this patch to be in the minor release, I think.

andydotxyz:
Are you sure this is the cause @Bluebugs ? The metadata that is being used does have 3 version digits and a build...
https://github.com/Jacalz/rymdport/blob/main/FyneApp.toml

Jacalz:
Indeed. It seems like that shouldn't quite impact my use case?

Bluebugs:
@Jacalz indeed, their was another bug in geoffrey that was preventing this also to work. It was a small one in fyne-cross, this one here and one more in geoffrey.

andydotxyz (approved):

PR #3649, opened 2023-02-10
add termux available

MatejMagat305:
well I would like to create video to fyne similar to [my video](https://youtu.be/7aUh39w_-2Q), would you agree?
On the video I show how to compile ndk apk in vlang (similar to golang) I think fyne it can done too
### Description:
add path on termux for fyne cmd (I asume that from apk cmd fyne will never call or will have similar structure like termux)

Fixes #(issue)
none
### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne/internal/mobile/env.go
+++ b/cmd/fyne/internal/mobile/env.go
@@ -342,6 +342,9 @@ func archNDK() string {
 			arch = "x86_64"
 			break
 		}
+		if runtime.GOOS == "android" {
+		   return "linux-aarch64"
+		}
 		fallthrough
 	default:
 		panic("unsupported GOARCH: " + runtime.GOARCH)


Bluebugs:
That's interesting. Would you mind correcting the coding style?

Do you use a tablet with termux to develop with it? I am guessing there is more to support as you would want to install a fyne app from fyne cli and I would be surprised if that worked out of the box, but never the less, this is an interesting step forward.

MatejMagat305:
sorry, I forgot you support 16+ - that method work on 28+ :(

Jacalz:
You are not using any methods? What is it that only works on 1.18 (I assume it is the version you mean).

MatejMagat305:
I mean version of android api ..., I was asume that it will be work when I enable path ...

MatejMagat305:
I'm realy sorry it is seems like that go doesn't know open clang on termux ...

Issue #3650, opened 2023-02-10
Add Tappable to widget.Slider 


alexballas:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [X] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

I was not able to click on slider to quickly move it. I had to drag it instead.

### Describe the solution you'd like to see.

A solution where widget.Slider implements the fyne.Tappable interface and properly handles the scroll positioning when tapping on it.

alexballas:
Since this is now merged, I'm closing it

PR #3651, opened 2023-02-10
Add tappable to widget.Slider

alexballas:
### Description:
Adding tappable to widget.Slider

Fixes [#3650](https://github.com/fyne-io/fyne/issues/3650)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/slider.go
+++ b/widget/slider.go
@@ -81,31 +81,23 @@ func (s *Slider) DragEnd() {
 // Dragged function.
 func (s *Slider) Dragged(e *fyne.DragEvent) {
 	ratio := s.getRatio(&(e.PointEvent))
-
 	lastValue := s.Value
 
 	s.updateValue(ratio)
-
-	if s.almostEqual(lastValue, s.Value) {
-		return
-	}
-
-	s.Refresh()
-
-	if s.OnChanged != nil {
-		s.OnChanged(s.Value)
-	}
+	s.positionChanged(lastValue, s.Value)
 }
 
 // Tapped function.
 func (s *Slider) Tapped(e *fyne.PointEvent) {
 	ratio := s.getRatio(e)
-
 	lastValue := s.Value
 
 	s.updateValue(ratio)
+	s.positionChanged(lastValue, s.Value)
+}
 
-	if s.almostEqual(lastValue, s.Value) {
+func (s *Slider) positionChanged(lastValue, currentValue float64) {
+	if s.almostEqual(lastValue, currentValue) {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


Jacalz:
You might wish to close this in favour https://github.com/fyne-io/fyne/pull/3531. It implements the full focus controls.

Jacalz:
Hmm. Sorry never mind. We are just doing similar but not the same things. It might be better to try and work together in this :)

alexballas:
Hi @Jacalz, would you like me to push the same changes to your branch and close this one?

alexballas:
Unless this gets approved now and you use that as the base for your branch

Jacalz:
I might be able to cherry-pick you commit into my branch. That way we will avoid a few headaches with merge conflicts.

Jacalz:
It also might just be simpler to merge this as that PR seems to be a bit stuck on a (possible) rounding error. Will try to review this when I have some time over.

andydotxyz (approved):
Looks good to me, thanks!
Best wait for @Jacalz as well given the note above.

Jacalz (changes requested):
This is great, thanks. The only thing I would like to fix is that there now is a lot of code duplication between `Dragged()` and `Tapped()` (they are basically the same). It seems to me like you should be able to just have `Dragged()` call `Tapped()` with it's point event.

andydotxyz:
> This is great, thanks. The only thing I would like to fix is that there now is a lot of code duplication between `Dragged()` and `Tapped()` (they are basically the same). It seems to me like you should be able to just have `Dragged()` call `Tapped()` with it's point event.

A good assertion, though I would prefer they both call a common method like `positionChanged` or similar, so it can't accidentally break if tapped and dragged change (for example if the request for `OnChangeEnded` were called from the `Tapped` method...

andydotxyz:
It is probably worth noting that `SetValue` is almost identical as well (with extra `Refresh`) - so possibly it could be even tidier with that common method being used by all three?

Jacalz:
Indeed. Sounds like a good plan to me 👍

alexballas:
Do you believe the code duplication in this case justifies the creation of a new method?
We're effectively talking about those lines
```go
	if s.almostEqual(lastValue, s.Value) {
		return
	}

	s.Refresh()

	if s.OnChanged != nil {
		s.OnChanged(s.Value)
	}
}
```
 Keeping things simple and leaving each method as they are sounds cleaner and aligns more to the Go way.
If we used a new method, we could also try to transfer the Value setting into that new method but we'd then need to account for two cases. One where just the ratio is passed and one where a direct value is pass (for the case of the SetValue method). It all sounds like unnecessary complexity.

alexballas:
I did apply the changes however, let me know

Jacalz (approved):
Looks good to me. Well done :+1: 
I just want the opinion of @andydotxyz before merging :)

andydotxyz (approved):
Thanks so much. Always nice to get a new feature whist reducing the lines of code in the widget ;)

PR #3663, opened 2023-02-14
Features/grey font

Bluebugs:
### Description:
This should reduce the amount of memory allocated and uploaded to the GPU. Hopefully making shaving some of the peak memory consumption off. This is just a small step in improving font rendering in Fyne.

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.


Bluebugs:
@Jacalz hopefully this reduce the performance problem you noticed earlier when resizing windows.

internal/painter/gl/draw.go
```diff
@@ -216,9 +217,44 @@ func (p *painter) drawText(text *canvas.Text, pos fyne.Position, frame fyne.Size
...
+		color = theme.ForegroundColor()
+	}
+
 	// text size is sensitive to position on screen
 	size, _ = roundToPixelCoords(size, text.Position(), p.pixScale)
-	p.drawTextureWithDetails(text, p.newGlTextTexture, pos, size, frame, canvas.ImageFillStretch, 1.0, 0)
+	p.drawGrayTexture(text, p.newGlTextTexture, pos, size, frame, color, 0)
+}
+
+func (p *painter) drawGrayTexture(o fyne.CanvasObject, creator func(canvasObject fyne.CanvasObject) Texture,
```
andydotxyz, line 230: Passing in `Color` to something called `Gray` seems a little misleading. Should it be more like `drawUniformTexture` or `DrawSingleChannelTexture`?

andydotxyz (changes requested):
Just a comment inline about naming.
But it got me wondering, should "gray" in the file names maybe be "uniform" or "singlchannel" or something that indicates that it is about masking on a single colour rather than being grayscale?

Jacalz (approved):
This seems to make things slightly better, I think. There was not a huge difference on my workstation. Will have a look at this on my laptop tomorrow :)

andydotxyz (approved):

PR #3677, opened 2023-02-24
Refactor GitHub workflow runners

Jacalz:
### Description:

This brings a major rework of the workflows that we run for tests and static analysis. Below is a list of all the improvements that have been made. However, in short, the biggest improvement is moving web tests out to their own workflow so we finally can run tests on Go > 1.17 :)

List of changes:
- Update to run all tests (except for Go 1.14 tests and new Web workflow) on Go 1.20.
- Use `actions/checkout@v3` instead of `actions/checkout@v2`.
- Update `WillAbides/setup-go-faster` to `v1.8.0`.
- Moved coverage generation to the Static Analysis workflow. As these were run only on Linux, it did not make sense to check it on every iteration. It also saves us some time as the old check ran tests twice (for both Go versions) but only pushed one of them to coveralls.
- All web tests are moved out to a separate Web Tests workflow that can keep a Go version that gopherjs supports.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/.github/workflows/license_check.yml
+++ b/.github/workflows/license_check.yml
@@ -7,12 +7,12 @@ jobs:
   licenses:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
       with:
         persist-credentials: false
-    - uses: WillAbides/setup-go-faster@v1.7.0
+    - uses: WillAbides/setup-go-faster@v1.8.0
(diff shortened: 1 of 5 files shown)


Bluebugs (approved):
Nice! Thanks for the update!

andydotxyz (approved):
Cool thanks, updating branch rules to match

PR #3683, opened 2023-02-26
Various fixes / improvements to image scaling code

andydotxyz:
I went to do some work and found some areas the image scaling could be improved. Fixes:

* Lots of errors logged for a normal image constructor
* Raw images with Original fill did not have scaled sizes set
* Compiler warning about inconsistent receiver naming
* Update methods to not function on side-effects

The code that logged 3 errors was:

```go
func main() {
	app := app.New()
	window := app.NewWindow("Hello world!")

	window.SetContent(canvas.NewImageFromResource(theme.LogoutIcon()))
	window.ShowAndRun()
}
```

Clearly not something that actually contains errors.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (dismissed):
Looks good to me. However, I am still getting these errors in fyne-demo, which I thought would be fixed by this PR:
```
2023/02/26 20:40:26 Fyne error:  Failed to load image
2023/02/26 20:40:26   Cause: no matching image source
2023/02/26 20:40:26   At: /home/jacob/Downloads/fyne-fix-imagescale/canvas/image.go:129
```

andydotxyz:
> Cause: no matching image source

It doesn't fix that error - that one isn't related to when something refreshes.
I'm not sure why that error is appearing actually...

andydotxyz:
Ah, is this on the data binding panel? I think it looks like an image with nil content is triggering this output... Should I strip out the error - or instead put in nil check for the various fields that, if all nil/zero, would skip the minsize/aspect code path?

Bluebugs:
Oh, I didn't consider that an all nil image refresh would be OK. I guess adding no check make sense.

canvas/image.go
```diff
@@ -261,97 +263,97 @@ func (i *Image) updateReader() error {
...
+
+		if i.FillMode == ImageFillOriginal {
+			return i.setMinSizeForPixels(width, height)
+		}
 	} else if i.Image != nil {
 		original := i.Image.Bounds().Size()
-		size = fyne.NewSize(float32(original.X), float32(original.Y))
-		i.aspect = size.Width / size.Height
+		i.aspect = float32(original.X) / float32(original.Y)
+		if i.FillMode == ImageFillOriginal {
```
Bluebugs, line 280: This looks like something that could be refactored between the two if path.

Bluebugs (changes requested):

canvas/image.go
andydotxyz: Short of moving the if into the method and calling it `setMinSizeForPixelsIfOriginalFill` I'm not sure how?
I did consider it but this layout seemed to be clearer.

Bluebugs: Looking at the code path both `if i.FillMode == ImageFillOriginal` seems to be doing the same things and are the last thing done in the `if`. There is also no other code path that could go past this three branch if. This automatically trigger a question for me, why is the code looking similar and duplicated, if there is no reason for it, it is misleading and duplicated. Otherwise if this code is not duplicated, then it is confusing.

andydotxyz: I'm not certain I follow - you mean that the code should be refactored so the `if` only occurs once? 
To do so we have to increase the scope of the width/height variables and move the code that uses the pixel information outside of the context where it is created...

```go
func (i *Image) updateAspectAndMinSize() error {
+       var pixWidth, pixHeight int
+
        if i.reader != nil {
                width, height, aspect, err := i.imageDetailsFromReader(i.reader)
                if err != nil {
                        return err
                }
                i.aspect = aspect
-
-               if i.FillMode == ImageFillOriginal {
-                       return i.setMinSizeForPixels(width, height)
-               }
+               pixWidth, pixHeight = width, height
        } else if i.Image != nil {
                original := i.Image.Bounds().Size()
                i.aspect = float32(original.X) / float32(original.Y)
-               if i.FillMode == ImageFillOriginal {
-                       return i.setMinSizeForPixels(original.X, original.Y)
-               }
+               pixWidth, pixHeight = original.X, original.Y
        } else {
                return errors.New("no matching image source")
        }
 
+       if i.FillMode == ImageFillOriginal {
+               return i.setMinSizeForPixels(pixWidth, pixHeight)
+       }
        return nil
 }
```

I personally don't see that this is preferable, but if it is required to land this tidy-up then I will add it.

Bluebugs: I would prefer something like that, indeed.

canvas/image.go
```diff
@@ -261,88 +263,102 @@ func (i *Image) updateReader() error {
...
+	}
+	driver := app.Driver()
+	if driver == nil {
+		return errors.New("no current driver")
+	}
+	c := driver.CanvasForObject(i)
+	if c == nil {
+		return nil // this will happen a lot during init
+	}
+	dpSize, err := scale.ToFyneSize(i, width, height)
```
Bluebugs, line 333: scale.ToFyneSize does all the access check above (fyne.CurrentApp().Driver().CanvasForObject(i) isn't necessary anymore)

Bluebugs (changes requested):

Bluebugs (approved):

PR #3687, opened 2023-02-27
Add option to set button importance for confirm dialogs

Jacalz:
### Description:

This adds support for changing the importance of the confirm button in the confirm and custom confirm dialogs. This is especially useful when using the new danger and warning importance values that were added in 2.3.

Unblocks https://github.com/Jacalz/rymdport/issues/48

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz:
I am very confused as to why that test is failing. Doesn't seem to be related to my change?

dialog/custom.go
```diff
@@ -23,12 +23,35 @@ func NewCustom(title, dismiss string, content fyne.CanvasObject, parent fyne.Win
...
+	NewCustom(title, dismiss, content, parent).Show()
+}
+
+// CustomConfirm is a custom dialog with dismiss and confirm buttons.
+//
+// Since 2.4
+type CustomConfirm struct {
+	*dialog
+
+	confirm *widget.Button
```
Bluebugs, line 39: Wouldn't it be possible to inherit from the ConfirmDialog here instead of duplicating things?

Bluebugs:
> I am very confused as to why that test is failing. Doesn't seem to be related to my change?

I wonder if it doesn't change the number of object generated by the button and so the count of object change.

dialog/confirm.go
```diff
@@ -20,6 +20,13 @@ func (d *ConfirmDialog) SetConfirmText(label string) {
 	d.win.Refresh()
 }
 
+// SetConfirmImportance sets the importance level of the confirm button.
+//
+// Since 2.4
+func (d *ConfirmDialog) SetConfirmImportance(importance widget.ButtonImportance) {
```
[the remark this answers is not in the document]
Jacalz: All other dialogs literally do it this way. I think it feels more inconsistent to it the other way, especially since you can't actually create dialogs by just creating the structs (all of the data is set up in the `NewXyz` functions).

dialog/custom.go
Jacalz: Should be resolved now

Bluebugs (approved):

Bluebugs:
I am ok with the change, just need to figure out why the tests get borked.

Bluebugs (dismissed):

Jacalz:
I figured it out. The test code is looking at the amount of images in the `testdata` directory. I updated the counters and also added some warnings, which hopefully avoids some confusion when the next person runs into the same issue :)

dialog/confirm.go
```diff
@@ -34,9 +42,9 @@ func NewConfirm(title, message string, callback func(bool), parent fyne.Window)
 			d.hideWithResponse(true)
 		},
 	}
-	d.create(container.NewGridWithColumns(2, d.dismiss, confirm))
+	d.create(container.NewHBox(layout.NewSpacer(), d.dismiss, confirm, layout.NewSpacer()))
```
andydotxyz, line 45: Is this change what we want? It was set up so the buttons are an even size before. I think that is how all the dialogs are, so isn't this inconsistent now?

If you want to change so they are not any more I don't think we need spacers as it is centred anyway.

andydotxyz (changes requested):
I think the change I highlighted inline makes the buttons different size/layout to the other dialogs.

dialog/confirm.go
Jacalz: I made the change to be more consistent with the other dialogs. From what I could tell, the majority of dialogs (at least the custom ones and form) use a HBox with spacers.

However, I do agree that it probably makes more sense to use grids for even sizing. I can roll-back this change and open a follow-up PR to move the others to use a grid instead if you'd like?

andydotxyz: I would think this makes sense, yes

Jacalz: Cool. I've just reverted the commit that introduced the change. Should be fine now

andydotxyz (approved):
Thanks :)

Issue #3684, opened 2023-02-26
Hangs on repeated calls to SetSelected in table.


dhowlett99:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

Using the latest version fyne.io/fyne/v2 v2.3.1 on Mac Catalina 10.15.7, go version go1.20 darwin/amd64

I'm building a table which has multiple widget types  (wrapped in a container) in each field of the table. 
Hide and Show are called to build a table with multiple widget types see screenshot.

I use SetSelected to set the correct option from an array of options in the select widgets of the table.

If you replace the SetSelected calls with just Selected the problem goes away, but the UI shows 'Set Selected' instead of the correct option.

### How to reproduce

On a MAC Build the example code and simply scroll up and down and within a few minutes the program will hang and never return.

### Screenshots

![image](https://user-images.githubusercontent.com/78086928/221433208-f866f368-34e4-4e49-834e-3d0b1e069fc7.png)

### Example code

```
package main

import (
	"fmt"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

const id int = 0
const tpe int = 1
const group int = 2
const number int = 3
const name int = 4
const label int = 5
const desc int = 6
const addr int = 7
const del int = 8
const add int = 9
const channel int = 10

func main() {
	myApp := app.New()
	myWindow := myApp.NewWindow("Table Widget")

	groupOptions := []string{"1", "2", "3", "4"}
	numberOptions := []string{"1", "2", "3", "4", "5", "6", "7", "8"}
	typeOptions := []string{"rgb", "scanner", "switch", "projector"}

	var data = [][]string{}

	// scan the fixtures structure for the selected fixture.
	for fixture := 0; fixture < 32; fixture++ {
		newFixture := []string{}
		newFixture = append(newFixture, fmt.Sprintf("%d", fixture))
		newFixture = append(newFixture, "rgb")
		newFixture = append(newFixture, fmt.Sprintf("%d", 1))
		newFixture = append(newFixture, fmt.Sprintf("%d", 1))
		newFixture = append(newFixture, fmt.Sprintf("name#%d", fixture))
[...]

PR #3693, opened 2023-03-02
fix mac hang

dhowlett99:
### Description:

Fixes #(issue
 Hangs on repeated calls to SetSelected in table. #3684 

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/table.go
+++ b/widget/table.go
@@ -723,6 +723,8 @@ func (r *tableCellsRenderer) MinSize() fyne.Size {
 
 func (r *tableCellsRenderer) Refresh() {
 	r.cells.propertyLock.Lock()
+	defer r.cells.propertyLock.Unlock()
+
 	oldSize := r.cells.cellSize
 	r.cells.cellSize = r.cells.t.templateSize()
 	if oldSize != r.cells.cellSize { // theme changed probably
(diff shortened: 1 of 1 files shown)


andydotxyz (changes requested):
Unfortunately a `defer` here is not OK - at the end we call `UpdateCell` (set to the local variable `updateCell`, which must not be within a lock.
User code should not be called when we are locked as they can call into any of our public API, which would be a deadlock

dhowlett99:
Andy,  If I move the r.cells.propertyLock.Unlock() to just above the block just above the updateCell code it hangs. So something in the updateCell is causing a deadlock on Mac.  

If I follow the code down I end up at `func (w *BaseWidget) ExtendBaseWidget` which immediatley had another property lock. `w.propertyLock.Lock()`  I just don't have enough knowledge on Fyne to decipher whats happening here.

All I can tell you is if you call teh updateCell code without the lock you end up in dead lock on the Mac.

Let me know if there's any thing I can do to help. Perhaps a conf call ?

andydotxyz:
> there is a stack trace in the original issue But I don't know why you refer to the updateCell Method as user code. Its a method on the table struct table.go line 30. from what i can see you should have the r.cells.propertyLock in place before messing with its internal cells.

The func() variable `updateCell` that I was referring to is set to the content of `r.cells.t.UpdateCell`. Therefore it must not be invoked with the table lock in place, as it is an exit point for the internal code.

Jacalz:

PR #3695, opened 2023-03-02
Bugs/checkbox no refresh

### LINKED ISSUE #3694: Checkboxes don't properly refresh on click. No tickbox. - Develop Branch -
### Checklist

### Describe the bug

Branch: Develop

Noticed when I click a checkbox, it doesn't display the tick icon but the icon just becomes darker. 
This issue was introduced with the following commit [https://github.com/fyne-io/fyne/commit/a769ce8a5617e2ffd0d74b280031e987b7325c24](https://github.com/fyne-io/fyne/commit/a769ce8a5617e2ffd0d74b280031e987b7325c24 )

It seems a refresh related issue. Please refer to the following video.
https://user-images.githubusercontent.com/11015965/222529570-6f0cc0ae-eea8-40c4-ae92-cd40d6490496.mp4

The issue can be easily replicated with any checkbox.

### How to reproduce

Please run the example code

### Screenshots

_No response_

### Example code

```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	myApp := app.New()
	myWindow := myApp.NewWindow("Test")
	medialoop := widget.NewCheck("One", func(b bool) {})
	medialoop.OnChanged = func(b bool) {
	}

	myWindow.SetContent(medialoop)
	myWindow.ShowAndRun()
}
```

### Fyne version

develop 

### Go compiler version

1.20

Ubuntu 18.04

### Additional Information

_No response_

Bluebugs:
### Description:
Checkbox code is modifying the icon without calling Refresh. It was working before as any code that triggered a Refresh on the screen would have walked the icon in the render thread and triggered the refresh. This is still something I consider an improper use of the API and a bug in the check box.

Fixes #3694 

### Checklist:
- [x] Tests included.
- [ ] Tests all pass.


andydotxyz:
Thanks for looking at this. However the method you fixed is called fr within `Refresh` which does request a refresh after a few methods update properties... so this fix doesn't feel quite right - it's adding another refresh in so the issue may lie deeper?

Bluebugs:
I couldn't find any call to refresh on the icon anywhere else, but I may have missed it. Where did you see that call?

andydotxyz:
Isn't it the last line in Refresh? I think it's the old type, maybe changing that is the minimal fix?

Bluebugs:
You mean the call on canvas.Refresh?

Bluebugs:
I think our expectation of what Refresh does might actually differ. The documentation on CanvasObject says: `// Refresh must be called if this object should be redrawn because its inner state changed.`. This means to me, that calling Refresh on the parent widget should not be sufficient to trigger a rendering update of the children object. The inner state of a container is the child object position and size, not the content of those child object as those are their inner state, not the parent object inner state.

I would not expect that calling Refresh on the parent would implicitly trigger the redraw of all children in a cascade update. I think it was because we have the painter directly access object properties during WalkTree that we pickup some change even when there wasn't any.

It make me a bit worried that now we have the shader pick font or rectangle color directly from the canvas object, you could manage to write code that change text or rectangle color without calling Refresh.

andydotxyz:
> This means to me, that calling Refresh on the parent widget should not be sufficient to trigger a rendering update of the children object.

This is not how it is designed, `canvas.Refresh` will invalidate the texture of all children as well, it always has done. Without this refreshing `Window.Content` would not redraw the whole screen.

Many widgets choose not to invalidate their entire content, but if they ask the canvas to invalidate themself then all child items will be redrawn.

Bluebugs:
I am guessing there is no limit to the amount of invalidation currently being done? This explain the resize of a window problem a lot better to me. Even image that do not change in size will get invalidated, rerendered and reuploaded for every resize step of the window.

If this is the understanding of what Refresh on a container should do, we can implement that recursive behavior in the application thread instead of the rendering thread and that would mimick the past behavior.

I would argue that this is a bad design and not clear from the API documentation that this should be expected. I think, we should better require explicit call to Refresh instead of implicit and clarify accordingly the API documentation. The application and the widget should be able to control what get refreshed, so that unecessary operation are not duplicated.

andydotxyz:
> This explain the resize of a window problem a lot better to me. Even image that do not change in size will get invalidated, 

This isn't quite true. When a window is resized it calls `Resize` not `Refresh` - then each child is resized according to layout - and each child that is changing size has `Resize` called. Items which are resized can then decide if they should call `Refresh` or not. Plenty of items (like Scroll) do not pass this information to children.

>  The application and the widget should be able to control what get refreshed, so that unecessary operation are not duplicated.

I don't see why a cascading refresh results in the lack of control you are describing.
It is not common for someone to request the whole UI is refreshed. But if they do, then that is what has been requested and we execute.
Every widget is absolutely in control of what it asks to be refreshed when a state changes.

I am sure the docs could be improved, but as we have already discussed pushing off the `canvas.Refresh` as a public method we need to be careful it all aligns.

andydotxyz:
> we can implement that recursive behavior in the application thread instead

I guess that is true, but I am not sure if that is desirable. Currently no textures are removed until we start the paint cycle, which seems correct - as only the painter should know when they are no longer needed.
See the `Canvas.FreeDirtyTextures` in `internal/common/canvas`.

Bluebugs:
> > This explain the resize of a window problem a lot better to me. Even image that do not change in size will get invalidated,
> 
> This isn't quite true. When a window is resized it calls `Resize` not `Refresh` - then each child is resized according to layout - and each child that is changing size has `Resize` called. Items which are resized can then decide if they should call `Refresh` or not. Plenty of items (like Scroll) do not pass this information to children.

This doesn't make logical sense to me. It only take one parent anywhere to trigger a Refresh call on itself to invalidate all of its Children. If the scroll is inside a split, all the children of the scroll are also a child of split and will get invalidated too. I don't see how you wouldn't have that completely cascading effect. And checking fyne_demo on main, you can see that resizing widgets/input leads to all image being regenerated.

> > The application and the widget should be able to control what get refreshed, so that unecessary operation are not duplicated.
> 
> I don't see why a cascading refresh results in the lack of control you are describing. It is not common for someone to request the whole UI is refreshed. But if they do, then that is what has been requested and we execute. Every widget is absolutely in control of what it asks to be refreshed when a state changes.

I do not understand in this case where is the barrier of the Refresh. If any parent widget when calling Refresh invalidate all children, children do not have a way to prevent it from propagating.

Bluebugs:
> > we can implement that recursive behavior in the application thread instead
> 
> I guess that is true, but I am not sure if that is desirable. Currently no textures are removed until we start the paint cycle, which seems correct - as only the painter should know when they are no longer needed. See the `Canvas.FreeDirtyTextures` in `internal/common/canvas`.

I think that would actually be better. FreeDirtyTexture seems to recursively walk every object that is being Refresh. This means that you would have multiple walk of all the children. What we could do instead, is explicitly walk the children in the application and mark them once for dirty texture. Then Canvas.FreeDirtyTextures would just walk a single time all the invalidated children (if that is what we want to keep as a behavior).

andydotxyz:
> This doesn't make logical sense to me. It only take one parent anywhere to trigger a Refresh call on itself to invalidate all of its Children. If the scroll is inside a split, all the children of the scroll are also a child of split and will get invalidated too.

But `Resize` should not imply `Refresh`. When a widget decides to refresh itself then it should follow that it means all content is to be refreshed.
It does seem that `Split` container has a bug where it refreshes the child widgets after resizing them, I will open a PR for that...

Bluebugs:
Indeed, there seems to be some misuse of Refresh that lead me to misunderstand its purpose. This means that this is not the right way to solve this bug. Will look at it again and look at the container that are calling Refresh to much. Might be good to improve Refresh documentation too.

andydotxyz:
I can see now why the test fails - the `test.Canvas` does not walk children in `Refresh` like the main drivers do.

However that would not answer why this does not work on an app - so there may still be more digging to do!

andydotxyz:
(You could work around this by forcing images in the software painter to always refresh, like the following, but it doesn't seem like the right fix:)

```diff
+++ b/internal/painter/software/painter.go
@@ -36,6 +36,7 @@ func (*Painter) Paint(c fyne.Canvas) image.Image {
                )
                switch o := obj.(type) {
                case *canvas.Image:
+                       o.Refresh()
                        drawImage(c, o, pos, base, clip)
```

Bluebugs:
Yeah, that's what we want to avoid indeed :-)

PR #3698, opened 2023-03-03
Don't refresh items when we don't know they have changed

andydotxyz:
Resize (and move) is all a container should do to child items.

### Checklist:

- [ ] Tests included. <- if the existing tests still pass that is a pass :)
- [x] Tests all pass.


Bluebugs:
Why is there also a change to the hello example?

andydotxyz:
> Why is there also a change to the hello example?

I don't know! I removed all other changes from my stage I thought. Will fix...

andydotxyz:
Apologies. It seems renaming a file had staged a remove/add which I missed. Fixed now.

Bluebugs (approved):

PR #3702, opened 2023-03-04
deps: Update golang.org/x/text to v0.6.0 for security fix

Jacalz:
### Description:

Already fixed on develop but this seems like something we actually want to backport.
Did not update to latest version as it requires Go 1.17 or later (makes a compelling case for Go 1.17 as next base version).

Fixes https://github.com/fyne-io/fyne/security/dependabot/4

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


Jacalz:
Hmm. The update of goldmark seems to make test fail. Let's just not backport this now then.

andydotxyz:
The gold mark upgrade making a failing test was ok - I updated the test on develop. It just changed empty heading from being "#" text to "" heading, which was ok for me :)

Jacalz:
Thanks. I'll try to incorporate that then :)

andydotxyz (approved):
Thanks

andydotxyz:
After this I wonder if we should prepare v2.3.2 including the recent macOS M2 fixes as well?

Jacalz:
That seems like a sensible plan 👍

PR #3705, opened 2023-03-05
Deprecate theme.FyneLogo() function

### LINKED ISSUE #3296: Deprecate theme.FyneLogo() from theme package
### Checklist

We should try to clean up the theme package for icons that we don't need. This Fyne logo was used back in the day but right now it is basically only used in `fyne_demo`, one single place in the driver as well as in a bunch of tests.

We should remove it, yes.

Deprecate the icon and move the tests to using something else. We can bundle the icon in `fyne_demo` and make sure that the internal driver don't set the logo when no other is provided. It might be better for applications to not have any when the developer does not select one. This makes it more clear that they should add their own icon.

Jacalz:
### Description:

This deprecates the `theme.FyneLogo()` function as discussed on a call a few months back. Developers should use their own application icons instead. Anyone that needs a Fyne logo can bundle it manually.

Fixes #3296

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


internal/driver/glfw/driver_desktop.go
```diff
@@ -40,7 +40,7 @@ func (d *gLDriver) SetSystemTrayMenu(m *fyne.Menu) {
 			} else if fyne.CurrentApp().Icon() != nil {
 				d.SetSystemTrayIcon(fyne.CurrentApp().Icon())
 			} else {
-				d.SetSystemTrayIcon(theme.FyneLogo())
+				d.SetSystemTrayIcon(theme.WarningIcon())
```
Jacalz, line 43: I suppose it is worth discussing what to do about systray. Should we use a warning icon, a question icon or just not set one at all?

andydotxyz: We have to set something surely? Without it there will be no display in the tray.

Jacalz: I guess it depends on how the system tray handles no icons. At least for app icons, a "no icon" icon is usually displayed if you don't provide one. Perhaps the same thing happens with system tray?

andydotxyz: Sadly it doesn't. At least not on some Os. Those just display a blank space that can be tapped. And i think I've seen some make it quite narrow as well (like it's "" instead of a missing icon).

andydotxyz:
I think it needs the `FyneApp.toml` references updated too

Jacalz:
> But yes we need to figure what the default for tray/app icon is and keep it consistent. I'm not sure that no-icon works in the tray context?

Absolutely. You are right. It is currently changed to use the warning icon but the error icon could work as well. Another idea of mine was to use the question icon. What would you prefer?

vcscsvcscs (dismissed):
LGTM

andydotxyz:
> > But yes we need to figure what the default for tray/app icon is and keep it consistent. I'm not sure that no-icon works in the tray context?
> 
> Absolutely. You are right. It is currently changed to use the warning icon but the error icon could work as well. Another idea of mine was to use the question icon. What would you prefer?

Maybe instead of re-using we should actually have an "icon missing" icon? then app developers can replace it using the theme without changing one of those dialog icons...

Jacalz:
That sounds like a good plan. I'll go ahead with getting that sorted in a few days

internal/driver/glfw/driver_desktop.go
Jacalz: I have switched to use the `BrokenImageIcon` from fynedesk.

internal/driver/glfw/testdata/windows_no_hover_outside_object.xml
```diff
@@ -1,47 +1,33 @@
-<canvas size="200x300">
```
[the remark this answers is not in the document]
Jacalz: Good point. I made sure to restore it

Jacalz:
I had to force-push a rebase on develop. Something seemed strange with the test where it expected `foreground_` at the beginning of the name yet none of the other files had it.

andydotxyz:
> I had to force-push a rebase on develop. Something seemed strange with the test where it expected `foreground_` at the beginning of the name yet none of the other files had it.

The foreground_ prefix should be on every default themed resource name now, it was a fix to a cache overlap bug earlier this month

Jacalz:
Yeah, I remember it being added but the test were strange. None of the other icon tests included the change so I didn’t either. However, my code still failed because it was missing `foreground`z. Strange. Either way, this should be fine now :)

andydotxyz (approved):
Nice

PR #3706, opened 2023-03-06
Remove unecessary call to Refresh on Linux.

Bluebugs:
### Description:
This remove excessive call to Refresh when resizing a window on Linux.

I have not noticed any issue after removing the call to Refresh and it doesn't make logical sense. It also has a direct performance impact and Refresh should not be called due to just a Layout change which is what a Resize is.

I think @Jacalz has tested it with Wayland and has also not noticed any adverse effect.

### Checklist:
- [ ] Tests all pass.


Jacalz (approved):
Indeed. I retested and am not seeing any difference except for improved performance. Nicely found :+1:

PR #3710, opened 2023-03-08
Deprecate MaxLayout in favour of StackLayout

Jacalz:
### Description:

This was discussed quite a long time ago by Andrew. Let's implement this to avoid confusion about what MaxLayout does.

### Checklist:

- [x] Tests included (kinda renamed).
- [x] Tests all pass.

#### Where applicable:


andydotxyz (approved):
Thanks. And timely - it seems the name was causing more confusion recently

PR #3712, opened 2023-03-09
Don't leak file handles for image

andydotxyz:
This also allows us not to hold files open for faster refresh. Trade off is a bit more memory used

Sadly this means we are now storing the compressed *and* decoded versions of bitmap images loaded from a file.
I think this fix is important. But maybe we need to follow up with a more memory efficient overall solution?

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/canvas/image.go
+++ b/canvas/image.go
@@ -245,21 +245,22 @@ func (i *Image) name() string {
 func (i *Image) updateReader() error {
 	i.reader = nil
 
-	var fd io.Reader
 	i.isSVG = false
 	if i.Resource != nil {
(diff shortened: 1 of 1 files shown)


Bluebugs (approved):
This make sense and I can think the simpler way to solve it might be with the shadow object as they will get destroyed by the internal painter/canvas code once an object is not used.

andydotxyz:
> We should open an issue to document that we should improve this.

Thanks, opened https://github.com/fyne-io/fyne/issues/3713

PR #3719, opened 2023-03-10
Cleanup/remove unecessary

### LINKED ISSUE #3713: Images should use less memory
### Checklist

### Describe the bug

The image handling code on `develop` uses a lot of memory. Current count is:

* cache contents of the file (to close the reader)
* The tee reader populates a buffer which may duplicate the above to support reading image metadata
* The image data is then decided into pixels which are also cached
* The image is then rendered to a texture cache duplicating the pixel info

we should be able to remove 3/5 of those caches. At least 2 should be cleared without need for any api changes.

### How to reproduce

Use `NewImageFromFile` (or resource, or reader etc)

### Screenshots

_No response_

### Example code

`NewImageFromReader(“myfile.png”)`

### Fyne version

develop

### Go compiler version

All

Any

### Additional Information

_No response_

Bluebugs:
### Description:
Quick tidying around image.

For #3713 

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/painter/gl/texture.go
+++ b/internal/painter/gl/texture.go
@@ -113,15 +113,11 @@ func (p *painter) newGlCircleTexture(obj fyne.CanvasObject) Texture {
 func (p *painter) newGlImageTexture(obj fyne.CanvasObject) Texture {
 	img := obj.(*canvas.Image)
 
-	width := p.textureScale(img.Size().Width)
-	height := p.textureScale(img.Size().Height)
-
-	tex := paint.PaintImage(img, p.canvas, int(width), int(height))
-	if tex == nil {
+	if img.Image == nil {
 		return noTexture
 	}
 
-	return p.imgToTexture(tex, img.ScaleMode)
+	return p.imgToTexture(img.Image, img.ScaleMode)
 }
 
 func (p *painter) newGlLinearGradientTexture(obj fyne.CanvasObject) Texture {
--- a/internal/painter/gl/texture.go
+++ b/internal/painter/gl/texture.go
@@ -113,11 +113,15 @@ func (p *painter) newGlCircleTexture(obj fyne.CanvasObject) Texture {
 func (p *painter) newGlImageTexture(obj fyne.CanvasObject) Texture {
 	img := obj.(*canvas.Image)
 
-	if img.Image == nil {
+	width := p.textureScale(img.Size().Width)
+	height := p.textureScale(img.Size().Height)
+
+	tex := paint.PaintImage(img, p.canvas, int(width), int(height))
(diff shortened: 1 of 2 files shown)


andydotxyz:
It looks like a useful change, but it does not fix the memory issues that it marks as done.
More duplication exists in the `canvas.Image` load/decode code.

Bluebugs:
It is indeed a partial fix. Is there a different keyword in that case?

andydotxyz:
> It is indeed a partial fix. Is there a different keyword in that case?

You just need to reference the #number and it will link. "fixes" specifically activates the issue close workflow.

Bluebugs:
Thanks, it is corrected now.

Jacalz:
I updated it to say "For #nr". I think that is how it is usually done :)

internal/painter/gl/texture.go
```diff
@@ -113,15 +113,11 @@ func (p *painter) newGlCircleTexture(obj fyne.CanvasObject) Texture {
 func (p *painter) newGlImageTexture(obj fyne.CanvasObject) Texture {
 	img := obj.(*canvas.Image)
 
-	width := p.textureScale(img.Size().Width)
-	height := p.textureScale(img.Size().Height)
-
-	tex := paint.PaintImage(img, p.canvas, int(width), int(height))
```
andydotxyz, line 119: If this really isn't needed here then can you apply the same fix to the `software/draw.go` as well?
Then I don't think this `PaintImage` method is needed any more either...

andydotxyz (changes requested):

internal/painter/gl/texture.go
Bluebugs: The software backend does actually use PaintImage to do the scaling. You can see that the size being passed is actually different from the size of the image in the software case and not in the case of GL.

andydotxyz: Isn't it the same here? PaintImage is being given the current size of the object, not the size of the image pixels

Bluebugs: It isn't exactly the same, but I think you are right. I will drop that part of the change for now and come back to it later.

andydotxyz:
I'm looking at the change and it looks like the `Refresh` method sets the object field through `updateReader` then sets it to `nil` inside the same method.
I could be wrong but would it not be tidier to remove that as an object field and just handle it locally?

Bluebugs:
Seems doable, will look into it.

Bluebugs:
@andydotxyz that worked out.

andydotxyz (approved):
Thanks

PR #3722, opened 2023-03-12
XML Encode the windows XML data

### LINKED ISSUE #3195: When packaging an ampersand in "Name" causes an error
### Checklist

### Describe the bug

When packaging a Fyne app I cam across an issue where I had the application name in the `FyneApp.toml` containing an ampersand:

```toml
[Details]
  Icon = "logo.png"
  Name = "App With & In Name"
  ID = "com.example.app-with-ampersand-in-name"
  Version = "1.1.0"
  Build = 1
```

This resulted in an error `The application has failed to start because its side-by-side configuration is incorrect. Please see the application event log or use the command-line sxstrace.exe tool for 
more detail` similar to #1521

After reviewing the Windows Event Log there were entries showing `Error in manifest or policy file "C:\path\to\app.exe" on line 2. Invalid Xml syntax`

After replacing the `&` with `and` the issue went away.

I would assume the name is not being properly escaped in the generated manifest as an ampersand must be escaped as `&amp;` in XML.

### How to reproduce

1. Create `FyneApp.toml` with `Name` containing and ampersand character
2. Package using `fyne package -os windows`
3. Attempt to run app

### Screenshots

_No response_

### Example code

No code example, issue is with app manifest generation of `fyne` command line tool. 

### Fyne version

2.2.3

### Go compiler version

1.19

### Operating system

Windows

### Operating system version

Windows 11

### Additional Information

_No response_

andydotxyz:
Fixes #3195

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/encode.go
+++ b/cmd/fyne/internal/commands/encode.go
@@ -1,8 +1,9 @@
 package commands
 
-import "strings"
+import (
+	"html"
+)
 
 func encodeXMLString(in string) string {
-	amped := strings.ReplaceAll(in, "&", "&amp;")
-	return strings.ReplaceAll(strings.ReplaceAll(amped, "<", "&lt;"), ">", "&gt;")
+	return html.EscapeString(in)
 }
--- a/cmd/fyne/internal/commands/encode.go
+++ b/cmd/fyne/internal/commands/encode.go
@@ -1,0 +1,7 @@
+package commands
+
+import "strings"
+
+func encodeXMLString(in string) string {
+	return strings.ReplaceAll(in, "&", "&amp;")
+}
--- a/cmd/fyne/internal/commands/encode.go
+++ b/cmd/fyne/internal/commands/encode.go
@@ -3,5 +3,6 @@ package commands
 import "strings"
 
 func encodeXMLString(in string) string {
-	return strings.ReplaceAll(in, "&", "&amp;")
+	amped := strings.ReplaceAll(in, "&", "&amp;")
+	return strings.ReplaceAll(strings.ReplaceAll(amped, "<", "&lt;"), ">", "&gt;")
 }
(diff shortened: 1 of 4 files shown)


cmd/fyne/internal/commands/encode.go
```diff
@@ -0,0 +1,8 @@
+package commands
+
+import "strings"
+
+func encodeXMLString(in string) string {
+	amped := strings.ReplaceAll(in, "&", "&amp;")
+	return strings.ReplaceAll(strings.ReplaceAll(amped, "<", "&lt;"), ">", "&gt;")
```
Jacalz, line 7: I wonder if there perhaps isn't a better alternative already in the standard library?
[This link](https://stackoverflow.com/questions/22906722/how-to-encode-special-characters-in-xml) seems to indicate that there are quite a few special characters.

Some googling on "html escape string in Go" seems to bring this up: https://stackoverflow.com/questions/38065765/how-to-escape-the-html-using-golang.

Jacalz: Using [html.EscapeString](https://pkg.go.dev/html#EscapeString) seems to be the right way to go.

cmd/fyne/internal/commands/encode.go
```diff
@@ -0,0 +1,9 @@
+package commands
+
+import (
+	"html"
+)
+
+func encodeXMLString(in string) string {
+	return html.EscapeString(in)
```
Jacalz, line 8: I don't think we need the wrapper function nor the tests if we are to use the built in function.

Jacalz (changes requested):

cmd/fyne/internal/commands/encode.go
andydotxyz: Given that HTML and XML are not actually the same I felt the clarity was helpful.

Jacalz: Good point

Jacalz (approved):

PR #3732, opened 2023-03-15
Fix linking mobile simulator on Apple M1/2

andydotxyz:
Mobile simulator would not link on any Go newer than 1.15 due to broken build flags.
We can remove a lot of this complexity once 1.14 is not supported any more

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/mobile/gl/work.go
+++ b/internal/driver/mobile/gl/work.go
@@ -2,9 +2,9 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-//go:build (darwin || linux || openbsd || freebsd) && (go1.15 || go1.16 || go1.17 || go1.18 || go1.19 || go1.20 || go1.21)
+//go:build (darwin || linux || openbsd || freebsd) && go1.15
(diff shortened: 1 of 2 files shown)


Jacalz (commented):
Are you sure that this changes anything?As far as I know, any Go release after version X will enable the goX build tag. The previous build tags were already working as far as I know.

On top of that, isn’t it just go11X and not with a dot in between?

andydotxyz:
According to the documentation it does have a dot in the middle. Which page are you seeing that doesn't?

> As far as I know, any Go release after version X will enable the goX build tag. The previous build tags were already working as far as I know.

Good catch! thanks, it was just the "." typo then...

Jacalz:
Hmm. Interesting. The old builds tags were indeed not working because I see the Go 1.14 code still running.

Jacalz:
Looks like we figured out the same solution. It was just a missing dot :)

Jacalz (approved):
Well done. A nice and simple fix :)
Go vet really should flag this...

PR #3744, opened 2023-03-17
Bugs/resize table

Bluebugs:
### Description:
Table doesn't properly resize as their is two scroller object, one in `Table` and one in `tableRenderer` that must be kept in sync and aren't.

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/widget/table.go
+++ b/widget/table.go
@@ -410,7 +410,7 @@ type tableRenderer struct {
 }
 
 func (t *tableRenderer) Layout(s fyne.Size) {
-	t.scroll.Resize(s)
+	t.t.scroll.Resize(s)
 	t.moveIndicators()
 }
 


andydotxyz:
I'm not sure if it's worth noting or not but I cannot replicate the issue on macOS

andydotxyz:
I'm curious if the following also fixes it? The Linux and "non-linux: drivers are out of sync with the following change:

```diff
 func (w *window) platformResize(canvasSize fyne.Size) {
        w.canvas.Resize(canvasSize)
+       d.repaintWindow(w)
 }
```

Bluebugs:
Calling repaintWindow trigger a crash calling `d, ok := fyne.CurrentApp().Driver().(*gLDriver)`.

andydotxyz:
Ah oops threads.
What about:
```diff
--- a/internal/driver/glfw/window_linux.go
+++ b/internal/driver/glfw/window_linux.go
@@ -3,5 +3,8 @@ package glfw
 import "fyne.io/fyne/v2"
 
 func (w *window) platformResize(canvasSize fyne.Size) {
-       w.canvas.Resize(canvasSize)
+       runOnDraw(w, func() {
+               w.canvas.Resize(canvasSize)
+               d.repaintWindow(w)
+       })
 }
```

Bluebugs:
This is working too, but it is definitively feeling slower than with just the correction on the table. It might be because forcing a repaint will introduce an additional frame to render when we are already pushing. Is there any other things that d.repaintWindow do that Resize doesn't?

andydotxyz:
> This is working too, but it is definitively feeling slower than with just the correction on the table. It might be because forcing a repaint will introduce an additional frame to render when we are already pushing. Is there any other things that d.repaintWindow do that Resize doesn't?

I think it's more of a threading issue... `Resize` will result in a request to repaint, if required. Whereas the `repaintWindow` will draw for sure.

andydotxyz:
I am trying to replicate on my Linux setup and cannot.
The fyne_demo tables all appear to resize as expected.
How can we replicate the problem to validate the fix?

andydotxyz (changes requested):
This change doesn't make sense - the variables reference the same scroller.
This is the same pattern used in List and Tree which work there.

Something else must be going on, but I cannot replicate on any OS with fyne_demo so don't know how to help.

Bluebugs:
I have tried to make a simpler example, but I have so far failed. This is weird. No idea why there is no propagation between those two in my case and not in general.

Bluebugs:
I finally figured out how to reproduce it. It is related to how I am actually extending the table widget somehow:
```
package main

import (
	"fmt"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

type table struct {
	widget.Table

	data bool
}

func main() {
	app := app.New()
	w := app.NewWindow("Hello")

	t := &table{Table: *widget.NewTable(func() (int, int) {
		return 100, 100
	}, func() fyne.CanvasObject {
		return widget.NewLabel("Long text to force space allocation")
	}, func(i widget.TableCellID, o fyne.CanvasObject) {
		o.(*widget.Label).SetText(fmt.Sprintf("%d,%d", i.Row, i.Col))
	}),
		data: true,
	}
	t.ExtendBaseWidget(t)

	w.SetContent(container.NewBorder(
		widget.NewLabel("header"), widget.NewLabel("footer"),
		nil, nil,
		t))
	w.Resize(fyne.NewSize(800, 600))
	w.ShowAndRun()
}
```

PR #3751, opened 2023-03-20
Add helper to automatically inject pprof in a fyne application.

Bluebugs:
### Description:
The goal of this PR is to provide better infrastructure in fyne tooling to get better bug report regarding memory usage mostly, but performance in general. It use the same trick as fyne metadata to inject pprof and run a web server to access it allowing for easy gathering of data.

It is obvious that this can be done with fyne help, but considering how little report we get contain enough information for acting, this seems like a necessary feature to improve Fyne and its ecosystem.

### Checklist:
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
I'm keen to hear what others think, but this feels a little like scope creep for our packaging tool?
That said we do have plans to provide a bunch of debugging tools and visual inspectors etc - should we consider this inside the Fyne tool or a separate debug tool?

Alternatively would keeping this in a new "fyne debug" command help to avoid parameter explosion for all commands?

Jacalz:
I think this sounds like a good idea. Having it part of the `fyne build` seems like a sensible and convenient way to help both us and users profile applications.

I suppose that a `fyne debug` command could be useful in the future with more debugging tools (maybe even adding some integration of `delve` if it makes sense). However, I am not sure how it would work to use that when building?

Jacalz:
Thinking about it though, it might not make sense to have `-pprof` for packaging.

Bluebugs:
> Thinking about it though, it might not make sense to have `-pprof` for packaging.

Agreed, that's why I only added it to the build command.

Bluebugs:
In my opinion, I would expect that debug tool might have performance impact (like running something inside `delv` for example or injecting additional visual information). This is opposite of multiple use case for `pprof` which is trying to track memory and cpu usage without impacting performance. I would think that conflating `pprof` and a `fyne debug` command might lead to higher expectation of what kind of tool we can use for debugging.

Jacalz:
Yeah, maybe. I don't have anything against adding it to the build commanded. It seems like a sensible and very helpful addition to me.

andydotxyz:
Then that works for me too

cmd/fyne/internal/commands/build.go
```diff
@@ -67,6 +69,17 @@ func Build() *cli.Command {
...
 			},
+			&cli.BoolFlag{
+				Name:        "pprof",
+				Usage:       "Enable pprof profiling.",
+				Destination: &b.pprof,
+			},
+			&cli.IntFlag{
+				Name:        "pprof-port",
+				Usage:       "Specify the port to use for pprof profiling.",
+				Value:       6060,
```
andydotxyz, line 80: Is 6060 a normal default?
We use 8080 in "fyne serve", shouldn't we be in the same range for all our port defaults?

Bluebugs: Yes, it is. That's the port that all net pprof example use: https://pkg.go.dev/net/http/pprof

andydotxyz (approved):

PR #3757, opened 2023-03-23
Issue #3492: Adjust scrolling speed for windows and linux

### LINKED ISSUE #3492: Enhance scroll wheel/touchpad scroll on desktop
### Checklist

The scrolling using mouse wheel (or laptop touchpad) is too slow, comparing to other applications with the same mouse (obviously).

_No response_

Increase scrolling speed/pages per scroll or any other solution you see is good.

dweymouth:
### Description:

Fixes #3492 

Adjusting scrolling speed for Windows and Linux to more closely match the behavior of other applications. I have not tested any scrolling adjustments on any platform other than Win/Mac/Linux.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


dweymouth:
The base of this PR looks like it picked up @Jacalz's security.md commit. This is the tweaks I made to scrolling speed for my app (which does not target BSD and I did not test on any BSD system). I would be inclined to think that BSD should probably share the same scroll speed as windows and linux, and that Mac is the odd one out due to its builtin momentum scrolling.

Jacalz:
Did you accidentally base it on the `master` branch and not `develop`?

Jacalz:
Also, please make sure to fill out the description file with a short summary and a correct "Fixes #issue-nr" line.

We also want the checklist to be filled out. If the "when applicable" checks are not applicable, you can remove those.

dweymouth:
Yes, I did base against master, sorry about that. I'll rebase against develop this evening. 

> I'm also kind of thinking about if it would be cleaner to just move const fields into two different files (protected by build tags). I'm personally not a huge fan of runtime.GOOS checks but it might just be a personal preference of mine.

I actually agree with this. This PR is just the tweak that I used for my app, Andy asked me to open it so you can test it on different systems. Once it's determined which systems (ie bsd) will need a tweak to the scroll speed then it should be moved to new files protected by build tags IMO.

Bluebugs:
Oh, that look fun. I am betting that the acceleration is actually different in the web case depending on the platform it is running on. I am guessing we will have to look at the user agent again. Absolutely great discovery :-D

Jacalz (commented):
I can confirm that this makes the scrolling feel a lot more consistent on both Linux and FreeBSD (tested on Gnome and Mate desktops respectively). I would perhaps prefer it to be ever so slightly slower but it is hard to say. It is definitely an improvement.

matwachich (commented):
Seems more like a quick dirty fix than a definitive one.

Jacalz:
> Seems more like a quick dirty fix than a definitive one.

I don't see it being any worse than what we already have, especially with the items discussed in https://github.com/fyne-io/fyne/pull/3757#issuecomment-1481519879 being implemented. If anything, it improves consistency on multiple platforms. Do you have a suggestion for a better way to handle this?

andydotxyz:
@Jacalz I think this got left behind - what are your thoughts?

Jacalz:
I am pretty happy with it but would prefer to have the scrolling speeds determined by build tags instead of methods with runtime lookups from operating systems. With that change done, I'd be more than happy to approve and merge :)

dweymouth:
Was it determined if other platforms should have the scrolling speed adjusted as well? I only tested on Win/Mac/Linux

Jacalz:
I tested on FreeBSD with the conclusion that your change was an improvement. I suspect that the same is true for NetBSD and OpenBSD as well but I don't know for sure.

dweymouth:
Cool, so basically all the desktop platforms except Mac. I'll update this sometime this week to use build tags and cover all the *BSD with the scrolling adjustments.

dweymouth:
Opened this as a fresh pull #3921 fixing the baseline branch and using conditional compilation instead of `runtime.GOOS` as discussed

PR #3763, opened 2023-03-26
Use max of radio labels to lay out horizontal

### LINKED ISSUE #3386: RagioGroup miscalulates label widths in horizontal mode

andydotxyz:
Fixes #3386

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/radio_group.go
+++ b/widget/radio_group.go
@@ -173,13 +173,15 @@ func (r *radioGroupRenderer) MinSize() fyne.Size {
 	height := float32(0)
 	for _, item := range r.items {
 		itemMin := item.MinSize()
-		if r.radio.Horizontal {
-			height = fyne.Max(height, itemMin.Height)
-			width += itemMin.Width
-		} else {
-			width = fyne.Max(width, itemMin.Width)
-			height += itemMin.Height
-		}
+
+		width = fyne.Max(width, itemMin.Width)
+		height = fyne.Max(height, itemMin.Height)
+	}
+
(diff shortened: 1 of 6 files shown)


andydotxyz:
> Does this not result in a very big minimum size if one of the labels is much bigger than the others? It might just be me but I can't quite see why the old code has the issue.

There are two valid layouts, min size and equal (grid) - long text in just one will make grid take a lot more than min size, but it is more uniform.
The issue here is that it was previously an average of all - so long ones overlapped short by some amount. (min size was as you imagined, but the layout then split it evenly)

andydotxyz:
@Jacalz was your comment to be a "request change" or are you happy with this PR?

PR #3764, opened 2023-03-26
Don't block main thread with our lifecycle callbacks

### LINKED ISSUE #3724: Calling SetContent() in SetOnEnteredForeground() hangs application
### Checklist

### Describe the bug

This bug can be reproduced by modifying the "Hello, World!" example with a lifecycle event (as shown below).

When calling SetContent() inside a Lifecycle().SetOnEnteredForeground(), the application will hang with no error and you will be forced to terminate the application.

When running with `go run -tags mobile` the application will run as expected with no hang.

### How to reproduce

Run example code provided on desktop

### Screenshots

_No response_

### Example code

Example of hanging code:
```go 
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Hello")

		hello,
			hello.SetText("Welcome :)")
		}),
	))

	a.Lifecycle().SetOnEnteredForeground(func() {
			hello,
			widget.NewButton("Goodbye!", func() {
				hello.SetText("Bye :(")
			}),
		))
	})

	w.ShowAndRun()
}
```

### Fyne version

v2.3.1

### Go compiler version

go1.20.2

Arch Linux (6.2.5-zen1-1-zen)

### Additional Information

_No response_

andydotxyz:
How they are called depends on the context, ideally an event queue.

Fixes #3724

### Checklist:

- [ ] Tests included. <- not sure how, race is in runtime driver callbacks
- [x] Tests all pass.

### DIFF
--- a/internal/app/lifecycle.go
+++ b/internal/app/lifecycle.go
@@ -40,34 +40,38 @@ func (l *Lifecycle) SetOnStopped(f func()) {
 	l.onStopped.Store(f)
 }
 
-// TriggerEnteredForeground will call the focus gained hook, if one is registered.
-func (l *Lifecycle) TriggerEnteredForeground() {
+// OnEnteredForeground returns the focus gained hook, if one is registered.
+func (l *Lifecycle) OnEnteredForeground() func() {
 	f := l.onForeground.Load()
-	if ff, ok := f.(func()); ok && ff != nil {
-		ff()
+	if ff, ok := f.(func()); ok {
+		return ff
 	}
+	return nil
 }
 
-// TriggerExitedForeground will call the focus lost hook, if one is registered.
-func (l *Lifecycle) TriggerExitedForeground() {
+// OnExitedForeground returns the focus lost hook, if one is registered.
+func (l *Lifecycle) OnExitedForeground() func() {
 	f := l.onBackground.Load()
-	if ff, ok := f.(func()); ok && ff != nil {
-		ff()
+	if ff, ok := f.(func()); ok {
+		return ff
 	}
+	return nil
 }
 
-// TriggerStarted will call the started hook, if one is registered.
-func (l *Lifecycle) TriggerStarted() {
+// OnStarted returns the started hook, if one is registered.
+func (l *Lifecycle) OnStarted() func() {
 	f := l.onStarted.Load()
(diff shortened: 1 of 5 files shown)


andydotxyz:
I think then we have to add an event handling loop for the overall application...
Adding a simpler wait will just put us in a blocking situation again.

Or I could make it just impact the foreground/background window code for now and not deal with the start/stopped which has this complication?

Bluebugs:
I am not to sure what the intent of this change is, but couldn't the user of those API just call `go func...` to spin their code in an asynchronous fashion if that is what they need?

andydotxyz:
> I am not to sure what the intent of this change is, but couldn't the user of those API just call `go func...` to spin their code in an asynchronous fashion if that is what they need?

Yes developers could work around the bug - but we should not deadlock.
The change is to resolve the bug linked in the description. As I said I could reduce it to just the foreground/background to use the event queues that exist and avoid the intermediate `go` usage for the app where there is not a threaded event queue.

andydotxyz:
Just reviving this PR. The crash in fyne_demo is resolved.
I don't know if this is good to go or not. As discussed above calling public APIs from other public callbacks should not deadlock. Is this introducing unexpected behaviour?

andydotxyz:
(the test failure has been resolved on the theme widgets PR)

internal/app/lifecycle.go
```diff
@@ -48,45 +48,55 @@ func (l *Lifecycle) SetOnStopped(f func()) {
...
 
-// TriggerStopped will call the stopped hook, if one is registered,
-// and an internal stopped hook after that.
-func (l *Lifecycle) TriggerStopped() {
+// OnStopped returns the stopped hook, if one is registered.
+func (l *Lifecycle) OnStopped() func() {
 	f := l.onStopped.Load()
-	if f != nil && *f != nil {
-		(*f)()
+	f2 := l.onStoppedHookExecuted
```
Jacalz, line 84: Can we change `f` and `f2` to be a bit more descriptive here?

Jacalz (changes requested):
Thanks. Looks mostly good, just a suggested naming change inline :)

Jacalz (approved):

PR #3772, opened 2023-03-29
Add CurrentKeyModifiers API to desktop driver

dweymouth:
### Description:

Adds a new API `CurrentKeyModifiers() fyne.KeyModifier` to the desktop driver to return the set of key modifiers that are currently held down.

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz (commented):
This overall looks good now.
There is a conflict unfortunately, sorry if it was caused by a delay in reviewing.
When updating to fix the merge conflict it might be worth moving the if/elseif/else to a switch, after all Go is happy to support multiple values per-case :).

andydotxyz (approved):
Thanks so much, this looks good to me. Anyone else around to check it out?

Jacalz (approved):
Looks good. A nice and clean implementation. Well done :)

PR #3777, opened 2023-04-01
Minor performance improvement for dialogs

Jacalz:
### Description:

This implements two minor performance improvements for the dialog. Firstly, it removes an unnecessary refresh within the `create` function because the call to `d.win.Show()` already does a refresh before showing. It also makes sure that we are not creating a new modal popup when setting the button row.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/base.go
+++ b/dialog/base.go
@@ -119,6 +119,11 @@ func (d *dialog) create(buttons fyne.CanvasObject) {
 	d.win = widget.NewModalPopUp(content, d.parent.Canvas())
 }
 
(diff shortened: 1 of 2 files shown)


Jacalz:
I think the base dialog mostly should be up to the standards of the other Fyne codebase as of having this landed. Some of the larger dialogs could do with refactors but I feel that the code has come a long way compared to v2.3.x.

andydotxyz (approved):
Good catch

PR #3778, opened 2023-04-01
Improve the VBox and HBox documentation

Jacalz:
### Description:

This adds more information for `VBox` and `HBox` to clarify that they render at `MinSize`.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Good idea, do we need to update the developer site layout page as well?

Jacalz:
> Good idea, do we need to update the developer site layout page as well?

Thanks. I don't think we necessarily need to but it probably wouldn't hurt. The documentation at https://developer.fyne.io/explore/layouts seems better than what we had here earlier.

PR #3780, opened 2023-04-01
AUTHORS: Update the email for Jacalz

Jacalz:
### Description:

Thanks to Tutanota, I now have sponsored premium so I can have email aliases and avoid exposing my private email to the internet. Hence, I should probably add my development alias instead of no email at all :)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Welcome :)

Issue #3543, opened 2023-01-06
Look at slow duduplication in Radio- and CheckGroup


Jacalz:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

Both RadioGroup and CheckGroup remove duplicated items each time they are refreshed. This is slow for large number of items and unnecessary to do multiple times.

The duplication was added because multiple items were showing as selected if the had the same string. We should try to  remove duplication and just make sure that it is not showing multiple items as selected.

### How to reproduce

Radio: https://github.com/fyne-io/fyne/blob/8c2509518b2df442a6b748d9b07754739592e6d7/widget/radio_group.go#L116
Check: https://github.com/fyne-io/fyne/blob/8c2509518b2df442a6b748d9b07754739592e6d7/widget/check_group.go#L149

### Example code

See links above ^

### Fyne version

v2.3.0

### Go compiler version

1.19.3

### Operating system

Linux

### Operating system version

Fedora 36

Jacalz:
The other option is of course to do the de-duplication less often but it would be good avoid if it all together if we can.

andydotxyz:
I liked the idea of removing the (slow) code and ensuring that only an item tapped is selected to better fix the original bug.

alx3dev:
What about adding code to check if item exist before append?

function to be added:
```go
func NotIncluded(options []string, value string) bool {
	result := true
	for x := 0; x < len(options); x++ {
		if options[x] == value {
			result = false
		}
	}
	return result
}
```  
then in append:
```go
// Append adds a new option to the end of a RadioGroup widget.
func (r *RadioGroup) Append(option string) {
	if NotIncluded(r.Options, option) {
		r.Options = append(r.Options, option)
		r.Refresh()
	}
}

```

Bluebugs:
Options is public, so it would bypass the Append check :-( A solution is to cache the last value of Options when it was checked and only do the costly operation when there is a change.

andydotxyz:
I think on the call we found a way to avoid needing the de-duplication code at all, by fixing the initial issue that multiple items were selected if they were duplicates.

Jacalz:
Indeed. I will try to remove the function entirely and just make sure to not display multiple items, with the same name, as selected when one is clicked.

Jacalz:
[...]

PR #3782, opened 2023-04-02
Remove unnecessary refreshes from check- and radiogroup

Jacalz:
### Description:

The individual item will be refreshed by itself when tapped but the parent was then forcing a refresh of the entire item slice directly after. This does not need to happen as the necessary state already is updated without updating all other options as well.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/check_group.go
+++ b/widget/check_group.go
@@ -142,7 +142,6 @@ func (r *CheckGroup) itemTapped(item *Check) {
 	if r.OnChanged != nil {
 		r.OnChanged(r.Selected)
 	}
-	r.Refresh()
 }
 
 func (r *CheckGroup) update() {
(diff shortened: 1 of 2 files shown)


Jacalz:
We might have to still do this if we remove the deduplication of items in https://github.com/fyne-io/fyne/issues/3543. However, it can probably be done in a better way with only updating items with the exact same option name. Until then, this saves us both memory and improves the performance.

Solving that issue is far from trivial and I need to find a good solution. The duplicate state that needs to be managed within the RadioGroup and CheckGroup is quite troublesome.

Jacalz:
Gosh. There is even more to it. The call to `SetSelected()` means that we are calling refresh for all items twice on each press. Will need to look into this even more.

Jacalz:
Never mind. There are two different `SetSelected` being called. This is fine.

Jacalz:
The logic here seems fine for Check but I realised that it is fundamentally wrong for Radio but still seems to work either way. Will try to sort it out in a different way.

PR #3785, opened 2023-04-04
Trigger image.Refresh outside of the rendering thread.

Bluebugs:
### Description:
Canvas.Image.Refresh() needs to be called outside of the rendering thread. This modification by walking the object tree to call Image.Refresh() before notifying the render thread make sure that all children image are properly refresh before being updated.

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/common/canvas.go
+++ b/internal/driver/common/canvas.go
@@ -5,6 +5,7 @@ import (
 	"sync/atomic"
 
 	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/canvas"
 	"fyne.io/fyne/v2/internal"
 	"fyne.io/fyne/v2/internal/app"
 	"fyne.io/fyne/v2/internal/async"
@@ -288,6 +289,22 @@ func (c *Canvas) Painter() gl.Painter {
 
 // Refresh refreshes a canvas object.
 func (c *Canvas) Refresh(obj fyne.CanvasObject) {
+	walkNeeded := false
+	switch obj.(type) {
+	case *fyne.Container:
+		walkNeeded = true
+	case fyne.Widget:
+		walkNeeded = true
+	}
+
+	if walkNeeded {
+		driver.WalkCompleteObjectTree(obj, func(co fyne.CanvasObject, p1, p2 fyne.Position, s fyne.Size) bool {
+			if i, ok := co.(*canvas.Image); ok {
+				i.Refresh()
+			}
+			return false
+		}, nil)
+	}
 	c.refreshQueue.In(obj)
 	c.SetDirty()
 }
--- a/internal/driver/common/canvas.go
+++ b/internal/driver/common/canvas.go
@@ -38,6 +38,8 @@ type Canvas struct {
 
 	painter gl.Painter
 
+	serializeRefresh sync.Mutex
+
 	// Any object that requestes to enter to the refresh queue should
 	// not be omitted as it is always a rendering task's decision
 	// for skipping frames or drawing calls.
@@ -298,13 +300,16 @@ func (c *Canvas) Refresh(obj fyne.CanvasObject) {
 	}
 
 	if walkNeeded {
+		c.serializeRefresh.Lock()
 		driver.WalkCompleteObjectTree(obj, func(co fyne.CanvasObject, p1, p2 fyne.Position, s fyne.Size) bool {
 			if i, ok := co.(*canvas.Image); ok {
 				i.Refresh()
 			}
 			return false
 		}, nil)
+		c.serializeRefresh.Unlock()
 	}
+
 	c.refreshQueue.In(obj)
 	c.SetDirty()
 }
--- a/internal/driver/common/canvas.go
+++ b/internal/driver/common/canvas.go
@@ -199,12 +199,25 @@ func (c *Canvas) FocusPrevious() {
 func (c *Canvas) FreeDirtyTextures() (freed uint64) {
 	freeObject := func(object fyne.CanvasObject) {
 		freeWalked := func(obj fyne.CanvasObject, _ fyne.Position, _ fyne.Position, _ fyne.Size) bool {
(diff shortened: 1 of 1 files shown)


andydotxyz (commented):
This does indeed fix the issue.
I wonder if there is a performance impact though - `Refresh` previously just added to a queue but now the call blocks until the walk is done.
Often that may not be a problem - but if the UI contains large images won't it now hang the UI until the image loads are all complete?

Bluebugs:
Loading large image would have happened previously inside the render thread, blocking it until the image is loaded. With this code, the load will happen in the thread that is sending update to the render thread. So it won't block the render thread and it should in general make frame rate more stable. The downside is that we are not doing a synchronized update anymore. All the image child object are updated first, then the other object are invalidated. So you could have a frame where image display first, then other component get updated. That being theory, as Fyne is mostly having eventual consistency as a synchronization pattern, I think that's ok.

andydotxyz:
It seems like we are moving further from content that can draw synchronously...
Is that a problem? Is moving this blocking call into the user's app code the right thing to do?
Anyone else got thoughts? @Jacalz or @changkun ?

Bluebugs:
What do you mean by draw synchronously here?

andydotxyz:
It was in reply to your sentence:

> The downside is that we are not doing a synchronized update anymore.

changkun (commented):
Does every `(*Canvas).Refresh` get serialized? Can two `(*Canvas).Refresh` be called concurrently?

Bluebugs:
> Does every `(*Canvas).Refresh` get serialized? Can two `(*Canvas).Refresh` be called concurrently?

Indeed, you are right. I should have this inside a lock I think.

Bluebugs:
> It was in reply to your sentence:
> 
> > The downside is that we are not doing a synchronized update anymore.

Ah, that is not so much about synchronization of drawing as Fyne does eventual consistency when it comes to drawing, but this does not apply on the destruction of texture which was a synchronized operation (the construction of those texture is still not synchronized and wasn't).

With this change, the destruction could happen twice before the widget Refresh is fully taken into account. Once when doing the call to image.Refresh() which also load the image, then a frame get rendered (so invalidate and reupload the texture), and finally the widget Refresh() call finish which require a final frame to be rendered.

Saying that I just realized that this code is actually uploading the image texture to the GPU twice. Will correct it.

Anyway, prior to widget.Refresh() happening, there is the possibility that resizing some of the child object (like text) trigger also the same synchronization problem as the Resize on Text would trigger a Refresh prior to reaching the Refresh of the Widget, so you could end up with the Text size and Refresh happening on a different rendered frame than the one triggered by the Widget call to Refresh (It also would do a double texture upload/generation, but that can't be avoided at this point).

internal/driver/common/canvas.go
```diff
@@ -288,6 +304,25 @@ func (c *Canvas) Painter() gl.Painter {
...
+
+	if walkNeeded {
+		c.serializeRefresh.Lock()
+		driver.WalkCompleteObjectTree(obj, func(co fyne.CanvasObject, p1, p2 fyne.Position, s fyne.Size) bool {
+			if i, ok := co.(*canvas.Image); ok {
+				i.Refresh()
+			}
+			return false
+		}, nil)
+		c.serializeRefresh.Unlock()
```
changkun, line 323: If we are serializing the refresh call, do we need a benchmark to indicate performance changes here?

Bluebugs: What would you measure in the benchmark? Past behavior was loading image in the render thread which would have generated frame drop there, but this is now moved to the user thread. So less framedrop, but impact on user thread.

Also we do not have the ability to know when a change hit the screen, so measuring the time between a Refresh to when we get the change to screen is difficult. Do you have any suggestion maybe?

changkun: I am a bit disconnected from the problem we are trying to solve here. Is there a code snippet to reproduce the "framedrop" issue? It sounds to me that this is a decision tradeoff regarding higher performance but not fully correct or lower performance but fully correct. I think, if the "framedrop" only happens during loading, then the optimization can be done a bit better and guarantee the "lock" behavior only happens during loading. To benchmark, it would consider two cases: 1) refresh without loading; 2) refresh during loading.

Bluebugs: For a "framedrop" to happen, it would have to be in the render thread, but the idea is to move all image.Refresh call outside of the render thread. As the image.Refresh is now outside of the render thread, there is no loading during the rendering of the frame.

So with the image.Refresh always happening outside of the render thread, it means the loading is not happening in the same place as with past version of Fyne. This logically move the time spend in loading image from the render thread to the thread calling Refresh. As the loading is done before notifying the render thread, there is no framedrop as the image loading is not blocking the rendering thread. Still the total time spend on the task remain the same as long as there is no excess call to Refresh.

With that in mind, I do not see what to benchmark as we do not have the ability to know when a change outside of the render thread has been rendered. If we measure only the Refresh time, it won't be comparable as we moved where the computation is happening.

andydotxyz: Discussed on the call this change does make sense and it is hard to show in benchmarking.
We agreed that it is part of a wider improvement to how `Canvas.Refresh` is used as well so we can lead to better performance overall.

andydotxyz (approved):

Jacalz:
Will review later today

internal/driver/common/canvas.go
changkun: I looked up the notes [here](https://github.com/fyne-io/meetings/blob/main/2023/2023-04-14.md#image-render-on-event-thread-3785) and agreed that it would be helpful to document the practices for calling `Refresh`

changkun (approved):
I would still aim for a minimal example to reproduce the problem and benchmark it despite also understanding the challenge to reproduce it.

PR #3789, opened 2023-04-05
There is no signal on the web.

Bluebugs:
### Description:
This allow for compiling fyne application with tinygo dev. The resulting appliation is not functional yet.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/driver.go
+++ b/internal/driver/glfw/driver.go
@@ -6,10 +6,8 @@ import (
 	"bytes"
 	"image"
 	"os"
-	"os/signal"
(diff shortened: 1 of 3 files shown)


andydotxyz:
The title says this relates to web, but the impl. is only one of the web targets.
I wondered if it should be in the (poorly named) driver_mobile.go, renamed as driver_web.go (as that has all the web flags set)

Bluebugs:
Indeed, that make more sense. Done.

andydotxyz (approved):

PR #3794, opened 2023-04-07
Remove uses of multiple deprecated functions/methods

Jacalz:
### Description:

There are mostly just various uses of theme.LightTheme() and theme.DarkTheme() in tests remaining but I don't think we have a good path forward for those.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Nice tidy

PR #3801, opened 2023-04-09
Add OnChangeEnded callback API to Slider

### LINKED ISSUE #3652: API request: add OnChangeEnded API to the Slider widget
### Checklist

In some apps it is necessary to differentiate ValueChanged events to a slider that are occurring during a drag action, vs a ValueChanged that occurs at the end of a user interaction, in order to trigger backend logic only in the 2nd case.

Yes, by extending Slider, overriding `DragEnd` (and later overriding `Tapped` as well once tap-to-set support lands in the slider widget), to keep track of when the user interaction ends yourself

A new callback API to the slider widget `OnChangeEnded func()` which allows application code to be notified when a user interaction that is changing the slider value has completed. Maybe the signature should be `OnChangeEnded func(float)` where the last set value is passed as well

dweymouth:
### Description:

Adds an OnChangeEnded callback to Slider that is invoked when a drag ends and the value has changed,
or when a Tapped event changes the slider value.

Fixes #3652 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


widget/slider.go
```diff
@@ -183,6 +199,7 @@ func (s *Slider) TypedKey(key *fyne.KeyEvent) {
 			s.SetValue(s.Value + s.Step)
 		}
 	}
+	s.checkInvokeOnChangeEnded()
```
dweymouth, line 202: This should be moved to the SetValue function, since that function calls OnChanged, and should call OnChangeEnded as well if the value is updated. Will push a fix.

widget/slider.go
```diff
@@ -121,11 +127,21 @@ func (s *Slider) positionChanged(lastValue, currentValue float64) {
...
 	s.Refresh()
 
+	s.pendingChange = true
 	if s.OnChanged != nil {
 		s.OnChanged(s.Value)
 	}
 }
 
+func (s *Slider) checkInvokeOnChangeEnded() {
+	if s.pendingChange {
```
andydotxyz, line 137: We prefer early exit so main body is not indented. (I.e. invert logic and return instead)

widget/slider.go
```diff
@@ -121,11 +127,21 @@ func (s *Slider) positionChanged(lastValue, currentValue float64) {
 
 	s.Refresh()
 
+	s.pendingChange = true
 	if s.OnChanged != nil {
 		s.OnChanged(s.Value)
 	}
 }
 
+func (s *Slider) checkInvokeOnChangeEnded() {
```
andydotxyz, line 136: Maybe something simpler like `fireChangeEnded`?

andydotxyz (changes requested):
Thanks. Just a few things on style. More info is on the wiki contributing pages

widget/slider.go
dweymouth: What about `checkFireChangeEnded`? I like 'check' or similar being part of the function name to emphasize that if there is no pending change, it will not invoke the callback. Thoughts?

widget/slider.go
Jacalz: FYI @dweymouth: We prefer to have the reviewer mark their own comments as resolved :)

widget/slider.go
andydotxyz: I know what you mean but the word "check" is ambiguous to convey that.
It also checks that there is a OnChangeEnded - so which does it refer to.
Does it also require a version that will call the fire even if nothing has changed?

dweymouth: Good point, I am fine with your first suggestion. Will update

dweymouth: updated

andydotxyz (approved):
Great thanks

Issue #58, opened 2019-01-01
Text rendering is bad, not-aliased 


dz0ny:
Example on Ubuntu Bionic + Gnome on go 1.11.4

![image](https://user-images.githubusercontent.com/239513/50572410-79366280-0dc0-11e9-8be3-3c8dde2c0b42.png)

Any ideas on workaround?

andydotxyz:
Hi, I have seen something similar on macOS - I think the text render code we import varies between platform.

If you run with FYNE_SCALE=2 does it look better (the UI will be bigger though)

dz0ny:
Well kinda, but you can still see same font issues.

![image](https://user-images.githubusercontent.com/239513/50586524-0db0cb80-0e7a-11e9-9ff1-8b897193d8c1.png)

andydotxyz:
I have removed a workaround that was added for macOS - can you update and let me know if this makes things better? Thanks

dz0ny:
![image](https://user-images.githubusercontent.com/239513/50592718-9852f400-0e95-11e9-9f6d-14cc412f9a1b.png)

Seems bit better to me, but scaling is broken now:

<pre><font color="#FF361E"><b>➜</b></font><font color="#FFFFFF"><b> </b></font><b>env</b> <font color="#2AA7E7">FYNE_SCALE=2FYNE_SCALE=2</font> <font color="#2AA7E7">go</font> <font color="#2AA7E7">run</font> <font color="#2AA7E7">main.go</font>
# github.com/go-gl/glfw/v3.2/glfw
In file included from ../.gobu/global/pkg/mod/github.com/go-gl/glfw@v0.0.0-20181213070059-819e8ce5125f/v3.2/glfw/c_glfw_linbsd.go:24:
../.gobu/global/pkg/mod/github.com/go-gl/glfw@v0.0.0-20181213070059-819e8ce5125f/v3.2/glfw/glfw/src/linux_joystick.c: In function ‘_glfwInitJoysticksLinux’:
../.gobu/global/pkg/mod/github.com/go-gl/glfw@v0.0.0-20181213070059-819e8ce5125f/v3.2/glfw/glfw/src/linux_joystick.c:224:42: warning: ‘%s’ directive output may be truncated writing up to 255 bytes into a region of size 9 [-Wformat-truncation=]
             snprintf(path, sizeof(path), &quot;%s/%s&quot;, dirname, entry-&gt;d_name);
                                          ^~~~~~~
[...]

Issue #53, opened 2018-12-26
Dependencies are not installed using "go get"


andydotxyz:
As reported by @officialgupta using "go get install github.com/fyne-io/fyne" does not install the GL and svg dependencies...

Let's solve this so that it's a single step install for people who have Go and CGo working.

andydotxyz:
The best workaround at the moment is:

```bash
go get -u github.com/fyne-io/fyne
cd $GOPATH/src/github.com/fyne-io/fyne
go get -u ./...
```

Is there a way we can stop the additional step needed to recursively get?

andydotxyz:
I set up dependency vendoring - this should remove the problems with downloading dependencies as they are packaged in the project

PR #3806, opened 2023-04-10
Pick newer go-text with lower memory allocation.

Bluebugs:
### Description:
After https://github.com/go-text/typesetting/issues/53 landed, this significantly improve the amount of memory allocated during the drawing of a string. Picking up latest go-text lead to significantly less memory spike and allocation in my tests. Would be good to see if it solve any of #3471 or at least help.

#### Where applicable:


dweymouth:
Very much looking forward to this as it might finally unblock allowing users to set their own font in my app and allow theme variant switching without memory issues. Do you need to call the new `SetFontCacheSize` API that was introduced in [go-text/typesetting#54](https://github.com/go-text/typesetting/pull/54)?

andydotxyz:
Might be worth holding off a few more hours for a more complete reduction in initial memory requirement go-text/typesetting#58

Bluebugs:
#58 seems like it will address the problem with large font, but there might be still some work on that one. With #53, I hope that the pick usage seen by @Jacalz is gone as I don't think he was using a special font for his tests. It would still be interesting to know if we are making general progress there or if we should look more into our allocation pattern.

andydotxyz:
Good point. As it's about memory usage is there a before/after benchmark or measurement to verify?

Bluebugs:
It does help in my own local tests, but I was hoping @Jacalz could re run the test he was doing with his app that was showing the issue in #3471.

Jacalz:
I see about a 2 to 3 megabyte reduction. Nothing huge but a nice improvement :)

Bluebugs:
Dang, I was hoping for more :-( Back at looking at our allocation profile I guess. Maybe you can use fyne build --pprof and share the resulting allocs file.

Jacalz:
Sounds like a good plan. I just did some quick tests today. Will do some more thorough testing when I have a chance :)

Jacalz:
I created some memory profiles using [rymdport](https://github.com/Jacalz/rymdport) on the [next](https://github.com/Jacalz/rymdport/tree/next) branch. The profiles were made by just clicking through the tabs a couple times. Memory usage does climb to around 60-70MB from around 55MB (according to the system monitor when collecting the profiles).

I believe that the one called `before.pprof` (on current develop branch) has slightly lower memory usage over all but it might just have been a fluke as it was not what I got when testing yesterday. The profile `after.pprof` is with this path applied. I hope that the profiles are helpful to figure out where memory is allocated/leaking.

[profiles.zip](https://github.com/fyne-io/fyne/files/11225675/profiles.zip)

Bluebugs:
Was that the result of http://localhost:6060/debug/pprof/heap or http://localhost:6060/debug/pprof/allocs? I have the impression of the former. In that case, it seems our permanent used memory cost clearly come from font being fully loaded in memory and the patch pointed by Andrew should help. I would still like to see the result of allocs as that might point to why the GC keep memory high and doesn't give it back to the OS.

Jacalz:
I believe it was the former, like you say, but I didn't use the built in pprof but instead the `pkg/pprof` package because I couldn't figure out how to collect the file. Will give allocs a go tomorrow.

Jacalz:
We might want to make that more clear with the flag (I never got a chance to review the original PR before merging). I'll give it a shot later today if I can manage :)

Jacalz:
Here's a memory allocation profile of Rymdport when using this PR: [allocs-with-pr.zip](https://github.com/fyne-io/fyne/files/11240243/allocs-with-pr.zip)

Jacalz (approved):
It works fine. No reason to not approve :)

andydotxyz (approved):

PR #3822, opened 2023-04-13
Bugs/handle larger data set list

### LINKED ISSUE #3816: Performance issue in widget.List
### Checklist

### Describe the bug

Working with a really simple program using huge `widget.List`, I can see that there is two problems.

At the top of the list everything is working well, but as we go closer to the bottom two leaks appears:
1. It's start to be more and more slow (lag)
2. An offset appear when we select an item, that make the list goes up with no reason (the closer to the end, the bigger the offset is)

Video to see the problem: https://www.loom.com/share/0c2d5393fad9451cad7e8b0bb5d26676

### How to reproduce

We just have to create a huge widget.List (actually I did it with 300000 items).

You can use this repo to reproduce (I added a TextField when you can put a number to directly go to the specified line).
https://github.com/LaGregance/fyne-listview-issue
Then just run `go run .`

### Screenshots

Video: https://www.loom.com/share/0c2d5393fad9451cad7e8b0bb5d26676

### Example code

```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
	"strconv"
)

func main() {
	a := app.New()
	win := a.NewWindow("Fyne ListView Issue")
	win.SetMaster()
	win.Resize(fyne.NewSize(1024, 700))

	listSize := 300000
	listWidget := widget.NewList(
		func() int {
			return listSize
		},
		func() fyne.CanvasObject {
			return widget.NewLabel("Template Object")
		},
		func(id widget.ListItemID, item fyne.CanvasObject) {
			item.(*widget.Label).SetText(strconv.Itoa(id) + " - aaaaaaa")
		},
	)

        // For convenience only, field to go to a specified line
	gotoField := widget.NewEntry()
	gotoField.OnSubmitted = func(idStr string) {
		id, err := strconv.Atoi(idStr)
		if id > 0 && err == nil {
			listWidget.Select(id)
		}
		gotoField.SetText("")
	}
        // ---------------------------------------------------

	win.SetContent(container.NewBorder(gotoField, nil, nil, nil, listWidget))
	win.ShowAndRun()
}
```

### Fyne version

2.3.3

### Go compiler version

1.20.1

MacOS Ventura 13.2.1 (M1)

### Additional Information

_No response_

Bluebugs:
### Description:
Improve performance of list to be O(1) when manipulating large list with no custom item height.

Fixes #3816

### Checklist:
- [x] Tests all pass.

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -251,8 +251,8 @@ func (l *List) UnselectAll() {
 	}
 }
 
-func (l *List) visibleItemHeights(itemHeight float32, length int) (visible []float32, offY float32, minRow, maxRow int) {
-	maxRow = length
+func (l *List) visibleItemHeights(itemHeight float32, length int) (visible []float32, offY float32, minRow int) {
+	maxRow := length
 	rowOffset := float32(0)
 	isVisible := false
 	visible = []float32{}
@@ -600,19 +600,20 @@ func (l *listLayout) updateList(refresh bool) {
 	}
 
 	wasVisible := l.visible
-	visible := make(map[ListItemID]*listItem)
-	cells := []fyne.CanvasObject{}
 
 	l.list.propertyLock.Lock()
-	visibleRowHeights, offY, minRow, maxRow := l.list.visibleItemHeights(l.list.itemMin.Height, length)
+	visibleRowHeights, offY, minRow := l.list.visibleItemHeights(l.list.itemMin.Height, length)
 	l.list.propertyLock.Unlock()
 	if len(visibleRowHeights) == 0 && length > 0 { // we can't show anything until we have some dimensions
 		l.renderLock.Unlock() // user code should not be locked
 		return
 	}
 
+	visible := make(map[ListItemID]*listItem, len(visibleRowHeights))
+	cells := make([]fyne.CanvasObject, len(visibleRowHeights))
+
 	y := offY
-	for index := 0; index < maxRow-minRow; index++ {
+	for index := range visibleRowHeights {
 		row := index + minRow
 		itemHeight := visibleRowHeights[index]
 		size := fyne.NewSize(width, itemHeight)
@@ -631,7 +632,7 @@ func (l *listLayout) updateList(refresh bool) {
 
 		y += itemHeight + separatorThickness
 		visible[row] = c
-		cells = append(cells, c)
+		cells[index] = c
 	}
 
 	l.visible = visible
--- a/widget/list.go
+++ b/widget/list.go
@@ -260,13 +260,15 @@ func (l *List) visibleItemHeights(itemHeight float32, length int) (visible []flo
 		return
 	}
 
+	padding := theme.Padding()
+
 	for i := 0; i < length; i++ {
 		height := itemHeight
 		if h, ok := l.itemHeights[i]; ok {
 			height = h
 		}
 
-		if rowOffset <= l.offsetY-height-theme.Padding() {
+		if rowOffset <= l.offsetY-height-padding {
 			// before scroll
 		} else if rowOffset <= l.offsetY {
 			minRow = i
@@ -279,7 +281,7 @@ func (l *List) visibleItemHeights(itemHeight float32, length int) (visible []flo
 			break
 		}
 
-		rowOffset += height + theme.Padding()
+		rowOffset += height + padding
 		if isVisible {
 			visible = append(visible, height)
 		}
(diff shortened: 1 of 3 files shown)


widget/list.go
```diff
@@ -250,23 +251,39 @@ func (l *List) UnselectAll() {
...
 	rowOffset := float32(0)
 	isVisible := false
-	visible = make(map[int]float32)
+	visible = []float32{}
 
 	if l.scroller.Size().Height <= 0 {
 		return
 	}
 
+	// theme.Padding is a slow call, so we cache it
```
andydotxyz, line 264: Doesn't this just repeat what the code shows?

widget/list.go
```diff
@@ -250,23 +251,39 @@ func (l *List) UnselectAll() {
...
+	padding := theme.Padding()
+
+	if len(l.itemHeights) == 0 {
+		offY = l.offsetY - itemHeight - padding
```
andydotxyz: This doesn't seem right - it should be the offset of the first visible item which is aligned on row boundaries.
This calculation looks like it is always 1 cell off-screen... what about when at 0, or when an item is half visible?

andydotxyz: More likely the fast way to calculate would be offsetY modulus padded height.

widget/list.go
```diff
@@ -570,20 +600,22 @@ func (l *listLayout) updateList(refresh bool) {
...
 		return
 	}
 
+	visible := make(map[ListItemID]*listItem, len(visibleRowHeights))
+	cells := make([]fyne.CanvasObject, len(visibleRowHeights))
+
 	y := offY
-	for row := minRow; row < maxRow; row++ {
-		itemHeight := visibleRowHeights[row]
+	for index := range visibleRowHeights {
```
andydotxyz, line 616: This change seems to be adding calculations for no benefit - the previous loop should have been fine.

widget/list.go
```diff
@@ -250,23 +251,52 @@ func (l *List) UnselectAll() {
...
+
+		offY = float32(math.Floor(float64(l.offsetY/paddedItemHeight))) * paddedItemHeight
+		minRow = int(math.Floor(float64(offY / paddedItemHeight)))
+		maxRow = int(math.Ceil(float64((offY + l.scroller.Size().Height) / paddedItemHeight)))
+
+		if minRow < 0 {
+			minRow = 0
+			offY = 0
+		}
+		if minRow > length {
```
andydotxyz, line 278: This doesn't seem right somehow. Why would the minimum row be the index 1 after the last item?

widget/list.go
```diff
@@ -250,23 +251,52 @@ func (l *List) UnselectAll() {
...
+			minRow = length
+		}
+
+		if maxRow > length {
```
andydotxyz, line 282: I think this is off by one. A row index should never reach the length surely?

andydotxyz (commented):
I haven't had a chance to run the code yet but there seem to be some small mathematical issues.

widget/list.go
Bluebugs: I am afraid that with time the knowledge of why this is done will be lost.

widget/list.go
Bluebugs: As the commit state, using range reduce the array bound check. It is generally a better practice to use range when possible.

widget/list.go
Bluebugs: The default value for maxRow is set to length at the top of the function.

widget/list.go
Bluebugs: Indeed.

widget/list.go
andydotxyz: I guess it was a `min >= I > max` type if thing then, sorry

widget/list.go
```diff
@@ -250,38 +251,64 @@ func (l *List) UnselectAll() {
...
 			// before scroll
 		} else if rowOffset <= l.offsetY {
 			minRow = i
 			offY = rowOffset
 			isVisible = true
 		}
-		if rowOffset < l.offsetY+l.scroller.Size().Height {
-			maxRow = i + 1
-		} else {
+		if !(rowOffset < l.offsetY+l.scroller.Size().Height) {
```
andydotxyz, line 305: Shouldn't it be `>=` rather than `!(<)`?

andydotxyz (approved):

PR #3827, opened 2023-04-15
This fix a crash when no data is provided and just header is expected.

Bluebugs:
### Description:
The visual output is maybe not correct and might require to be discussed, but it seems we do not draw any header if we have no data. I think the expectation would be to draw the top header.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
There is a good chance this goes away when the headers non-sticky code is removed. Should it be revisited after that tidy up?

Bluebugs:
Maybe, just right now if you end up with an empty table it crash always making it hard to use table with dynamic content. Can wait a few definitively.

andydotxyz (approved):

PR #3831, opened 2023-04-17
Fix GL rectangle painting

toaster:
### Description:
The rounded rectangle feature introduced an unwanted side effect where rectangles without stroke color got this explicitly set to transparent when being painted.
This failed the GL tests which are not yet run on CI. There are still other failures there. This PR only fixes the several XML diffs of the `TestWindow_HandleOutsideHoverableObject` test:
```
-							<rectangle fillColor="inputBackground" pos="1,1" size="186x31"/>
+							<rectangle fillColor="inputBackground" pos="1,1" size="186x31" strokeColor="rgba(0,0,0,0)"/>
```

### Checklist:

- [ ] Tests included.
- [x] Tests all pass. (CI only)


Bluebugs:
Can I help by testing additional implementation?

There is a known bug with WebGL on Firefox where Firefox decide to drop from the shader variable that are unused and so we fail to upload that data on the go side. This is not related to your patch and was there before. I need to look at it, but will wait for this to land first.

andydotxyz (approved):
Nice improvement thanks

Verified on the Apple M2 as well

PR #3832, opened 2023-04-17
Correct path and correctly escape path when running windows install.

### LINKED ISSUE #3809: fine install command fails on Windows 10
### Checklist

### Describe the bug

On Windows 10 "fyne install" makes dir in Program File and shortcut in Start Menu but don't move built exe from src folder to Program Files

### How to reproduce

1. Run "fyne install” in source dir 
2. It makes folder in Program files and  shortcut in Start Menu
3. Executable remain in source dir but not been moved in correspondent folder of  Program files

### Screenshots

_No response_

### Example code

```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Hello")

		hello,
			hello.SetText("Welcome :)")
		}),
	))

	w.ShowAndRun()
}
```

### Fyne version

fyne version v2.3.1

### Go compiler version

go1.20.2 windows/amd64

Windows 10

### Additional Information

_No response_

Bluebugs:
### Description:
Make sure we actually copy the newly created executable from the right location to where we want it to be installed.

Fixes #3809

### Checklist:
- [ ] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/package-windows.go
+++ b/cmd/fyne/internal/commands/package-windows.go
@@ -108,19 +108,23 @@ func (p *Packager) packageWindows(tags []string) error {
 		return fmt.Errorf("failed to rebuild after adding metadata: %w", err)
 	}
 
-	appPath := p.exe
 	appName := filepath.Base(p.exe)
 	if filepath.Base(p.exe) != p.Name {
 		appName = p.Name
 		if filepath.Ext(p.Name) != ".exe" {
 			appName = appName + ".exe"
 		}
-		appPath = filepath.Join(p.dir, appName)
 		os.Rename(filepath.Base(p.exe), appName)
 	}
 
 	if p.install {
-		err := runAsAdminWindows("copy", "\"\""+appPath+"\"\"", "\"\""+filepath.Join(p.dir, appName)+"\"\"")
+		wd, err := os.Getwd()
+		if err != nil {
+			return fmt.Errorf("failed to locate current working directory")
+		}
+		appPath := filepath.Join(wd, appName)
+
+		err = runAsAdminWindows("copy", appPath, filepath.Join(p.dir, appName))
 		if err != nil {
 			return fmt.Errorf("failed to run as administrator: %w", err)
 		}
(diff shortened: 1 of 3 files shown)


cmd/fyne/internal/commands/package-windows.go
```diff
@@ -131,8 +135,15 @@ func (p *Packager) packageWindows(tags []string) error {
 func runAsAdminWindows(args ...string) error {
 	cmd := "\"/c\""
 
-	for _, arg := range args {
-		cmd += ",\"" + arg + "\""
+	for idx, arg := range args {
+		if idx == 0 {
```
andydotxyz, line 139: This can't be right? Only adding a comma for 0 index?
I think this would benefit from a unit test showing the correct format.

andydotxyz (changes requested):
It doesn't seem like the right encoding but it's hard to tell without a little test confirming how it comes together.

cmd/fyne/internal/commands/package-windows.go
Bluebugs: Actually cmd.exe does parse the string given to it as the command to execute. So it means the entire arguments list has to be in the same argument passed to it. It is closed at the end of the loop, if there was more than one. I will see to add a unit test.

andydotxyz: As requested above can you please add a unit test?
The mix of single and double quotes is hard to parse here. With args "1" and "2" it looks like it will produce: 
* (before) `"/c","1","2"`
* (after) `"/c" ,'1 "2"'`

Without seeing the expected output I can't figure if this is desired...

andydotxyz (changes requested):
As in other message, please add a unit test to show the expected output to confirm the formatting.

andydotxyz (approved):
Thanks. The examples demonstrate why the seemingly peculiar mix if brackets and raw parameters works.

I do have a concern that without escaping the first parameter it may fall down on some things (like running a command/binary that is not builtin) but it probably doesn't apply here.

PR #3837, opened 2023-04-18
Fix the miss-align drag when passing through less than min-size

andydotxyz:
This impacted table (row and col drag) and split containers.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Bluebugs (approved):
Looks good.

PR #3838, opened 2023-04-18
Bugs/recursive deadlock

Bluebugs:
### Description:
Small lock fixes, I came across while debugging a deadlock. This address all dead lock I could see in my application. It would be nice to figure a more systematic way to detect them.

### Checklist:
- [x] Tests all pass.

### DIFF
--- a/internal/driver/common/canvas.go
+++ b/internal/driver/common/canvas.go
@@ -68,21 +68,25 @@ func (c *Canvas) EnsureMinSize() bool {
 	}
 	var lastParent fyne.CanvasObject
 
-	c.RLock()
-	defer c.RUnlock()
-
 	windowNeedsMinSizeUpdate := false
 	csize := c.impl.Size()
 	min := c.impl.MinSize()
+
+	c.RLock()
+	defer c.RUnlock()
 
 	ensureMinSize := func(node *RenderCacheNode) {
 		obj := node.obj
 		cache.SetCanvasForObject(obj, c.impl)
 
+		c.RUnlock()
 		if !obj.Visible() {
+			c.RLock()
 			return
 		}
 		minSize := obj.MinSize()
+		c.RLock()
+
 		minSizeChanged := node.minSize != minSize
 		if minSizeChanged {
 			objToLayout := obj
@@ -91,7 +95,9 @@ func (c *Canvas) EnsureMinSize() bool {
 				objToLayout = node.parent.obj
 			} else {
 				windowNeedsMinSizeUpdate = true
+				c.RUnlock()
 				size := obj.Size()
+				c.RLock()
 				expectedSize := minSize.Max(size)
 				if expectedSize != size && size != csize {
 					objToLayout = nil
--- a/internal/driver/common/canvas.go
+++ b/internal/driver/common/canvas.go
@@ -68,6 +68,9 @@ func (c *Canvas) EnsureMinSize() bool {
 	}
 	var lastParent fyne.CanvasObject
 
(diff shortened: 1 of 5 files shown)


andydotxyz (commented):
This doesn't seem quite right - why would the container hold a lock whilst calling the public method of the layout, but not whilst accessing the private field of the container that the lock is on?

andydotxyz:
It's in the `updateLayout` function.

Bluebugs:
The lock is on the Canvas not on the container (Layout() can end up calling the Scale() function on the canvas and lead to a dead lock). I am not sure I understand your question.

widget/table.go
```diff
@@ -121,6 +121,7 @@ func NewTableWithHeaders(length func() (int, int), create func() fyne.CanvasObje
 	t := NewTable(length, create, update)
 	t.ShowHeaderRow = true
 	t.ShowHeaderColumn = true
+	t.ExtendBaseWidget(t)
```
andydotxyz, line 124: This change breaks the ability to extend table doesn't it?
We use the extend-later pattern in all other widgets.

andydotxyz (changes requested):
I don't think we should change the semantics of extending table to address lock issues

andydotxyz:
Apologies I misread the unlock/lock as lock/unlock

widget/table.go
Bluebugs: In `NewTable`, we are calling it first. It should be in one place only or it lead to a wrong order in lock usage. If the right place is in `CreateRenderer` then I can remove it from `NewTable` and that also solve the problem. As NewTableWithHeaders was new, I assumed that it was the one to break past behavior here.

andydotxyz: I think the pattern we use is that it is CreateRenderer and also in MinSize (because sometimes you need it extended before rendering).

internal/driver/common/canvas.go
```diff
@@ -560,13 +567,19 @@ type renderCacheTree struct {
 	root *RenderCacheNode
 }
 
-func updateLayout(objToLayout fyne.CanvasObject) {
+func (c *Canvas) updateLayout(objToLayout fyne.CanvasObject) {
 	switch cont := objToLayout.(type) {
 	case *fyne.Container:
 		if cont.Layout != nil {
-			cont.Layout.Layout(cont.Objects, cont.Size())
+			layout := cont.Layout
```
andydotxyz, line 574: Why do you need to lock to check `cont.Layout` but not `cont.Objects`?

andydotxyz (commented):
A couple of potential issues, let me know what you think

widget/entry.go
```diff
@@ -1747,9 +1747,7 @@ func (r *entryContentRenderer) moveCursor() {
 	r.buildSelection()
 	r.content.entry.propertyLock.RLock()
 	provider := r.content.entry.textProvider()
-	provider.propertyLock.RLock()
```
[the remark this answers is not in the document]
Bluebugs: propertyLock is under the ownership of provider. The call provider.lineSizeToColumn does use that same lock via rowBoundary to access the property of the object.

internal/driver/common/canvas.go
Bluebugs: Indeed, good catch.

widget/entry.go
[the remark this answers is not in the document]
andydotxyz: So it does, thanks

andydotxyz (approved):
Looks good thanks

PR #3840, opened 2023-04-20
Fix usage of images in native menus and system tray menu items.

andydotxyz:
Those images are not on a canvas so will always fail the  check

### Checklist:

- [ ] Tests included. <- native only :(
- [x] Tests all pass.


internal/painter/image.go
```diff
@@ -16,6 +16,9 @@ import (
 // to fit its original size. If it doesn’t, PaintImage does not paint the image but adjusts its min size.
 // The image will then be painted on the next frame because of the min size change.
 func PaintImage(img *canvas.Image, c fyne.Canvas, width, height int) image.Image {
+	if img.Size().IsZero() && width != 0 && height != 0 { // if the image has not been realised for some reason, set it up
+		img.Resize(fyne.NewSize(float32(width), float32(height)))
```
Bluebugs, line 20: I have slight concern here. When using the software painter, this code path could be triggered which would then happen in the render thread. Is it possible to either make drawImage in the software painter call paintImage directly or pass an additional information that says "this is not in the render thread"/something along that line.

andydotxyz: I'm not sure about that - it's getting to complex for a simple problem.
The alternative I wondered about is should the image constructor call the refresh code immediately? That feels worse to me because then your whole UI load will hang until all the images it references are decoded...

andydotxyz: What we are trying to work with here is that image contains more than just the metadata required to paint it now - it contains the actual paint cache.
With an "unrefreshed" image it can never be painted because even if it is refreshed the size is zero so the cached pixels are empty.

I'm not sure of the right way forward tbh. The code above is essentially checking for "invalid" cases caused when the image is drawn with no size but an allocated area. I suppose I could just put it behind a `c == nil` guard - not sure if that makes it better or not.

Bluebugs: Oh, indeed, behind a `c == nil` sounds like a good solution here as it address my concern.

Bluebugs (approved):

PR #3843, opened 2023-04-21
improve LogError

toaster:
### Description:

This PR adds another line of context to `LogError` output. This is at least useful for _all_ messages generated by Fyne itself because it does not help to know the location of the logging in Fyne but one needs to know the location of the caller of the logging Fyne method (maybe even further but that has to be proven).

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/canvas_test.go
+++ b/internal/driver/glfw/canvas_test.go
@@ -200,7 +200,7 @@ func TestGlCanvas_Focus(t *testing.T) {
 
 	foreign := &focusable{id: "o2e1"}
 	c.Focus(foreign)
-	assert.False(t, foreign.focused, "does not focus foreign object")
+	assert.False(t, foreign.focused, "does not focus foreign object but logs a Fyne error")
 	assert.True(t, o2e.focused)
 }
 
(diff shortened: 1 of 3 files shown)


toaster:
Mea culpa … didn’t know that we actually have a test for the error logger :).

Jacalz:
I wonder if there perhaps is a better way to generate a stack trace (instead of using `runtime.Callers()` in a loop)? I did some research and found [runtime.Stack()](https://pkg.go.dev/runtime#Stack), [debug.Stack()](https://pkg.go.dev/runtime/debug#Stack) and [debug.PrintStack()](https://cs.opensource.google/go/go/+/go1.20.3:src/runtime/debug/stack.go;l=15).

andydotxyz:
What this originally aimed to do was print the most likely like that caused a problem.
With the improvements it gets more helpful, but we should not put in a full stack in my opinion as it would get too long.
I'm not sure if the `byte[]` output of the suggested methods is really something we can work with?

Jacalz:
I suppose you’re right. I’m not entirely sure about this change. Like you say, the initial implementation is nice because it prints the most likely error location. I was quite happy with the current behaviour of the logging. 

What this PR is trying to do seems to me like what you usually would use a debugger for. Or am I perhaps too old school in saying that (I’ve done way too much C++ recently)?

Bluebugs:
I think log are for user report error and debugger are for developers. In this case, having a full stack report can sometime be useful, but not always. I would propose to either have the stack display when doing not doing a release build or when an environment variable is set.

Jacalz:
That sounds like a good plan. I'm definitely in favour of release builds (using that build tag) keeping the logging as is and debug builds providing more information.

andydotxyz:
We also have a handy "debug" tag that is perhaps where a full stack report would live?

PR #3844, opened 2023-04-22
Remove unecessary and excessive lock.

Bluebugs:
### Description:
I added this lock previously with the assumption that walking the tree was lock less. This was an incorrect assumption as lock are being taken accordingly when walking the trees. This lock can also be taken by the user thread, which can lead to a Refresh of a subobject while the parent is being refreshed. This can lead to dead lock.

### Checklist:
- [x] Tests all pass.

### DIFF
--- a/internal/driver/common/canvas.go
+++ b/internal/driver/common/canvas.go
@@ -38,8 +38,6 @@ type Canvas struct {
 
 	painter gl.Painter
 
-	serializeRefresh sync.Mutex
-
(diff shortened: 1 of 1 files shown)


changkun (approved):
Removing the lock is likely to introduce another source of a data race, which will be harder to resolve. To me, it is necessary: 1) introduce at least a test that can enable `-race`; 2) try to use condition variable

PR #3854, opened 2023-05-01
Don't hold lock when calling user code that could trigger a deadlock.

Bluebugs:
### Description:
While testing fyne_demo in advanced panel and resizing the window would always lead to a deadlock. This might be what @Jacalz was experiencing earlier.

### Checklist:
- [x] Tests all pass.


andydotxyz (approved):
Thanks. Fixes an issue I noticed on fyne_demo tab changes too

PR #3861, opened 2023-05-03
Features/faster table

### LINKED ISSUE #3860: Scrolling in Table Widget is slow with a lot of rows
### Checklist

### Describe the bug

If i add several hundreds of thousands of rows into a table, scrolling becomes very slow, and uses lots of cpu resources.

### How to reproduce

[fynetable.zip](https://github.com/fyne-io/fyne/files/11390178/fynetable.zip)

build, run, push "Test" button, and try to scroll (to the bottom)

### Screenshots

_No response_

### Example code

```go
```

### Fyne version

2.3.3 but also develop branch from 2023-05-02

### Go compiler version

1.20.3

Windows 11

### Additional Information

_No response_

Bluebugs:
### Description:
Remove O(n) cost from the code path when there is no custom width/height cell defined.

It would be nice to move away from using map and use two array instead, but that was a lot more change than I was wanting to do here. Also as it is a lot easier to change a column or row size with table, we might have to make that code path faster too at some point, but for now this does speed things for the most common case.

Fixes #3860 

### Checklist:
- [x] Tests all pass.

### DIFF
--- a/widget/table.go
+++ b/widget/table.go
@@ -732,6 +732,9 @@ func (t *Table) visibleRowHeights(rowHeight float32, rows int) (visible map[int]
 		return
 	}
 
+	// theme.Padding is a slow call, so we cache it
+	padding := theme.Padding()
+
 	stick := t.StickyRowCount
 	for i := 0; i < rows; i++ {
 		height := rowHeight
@@ -739,7 +742,7 @@ func (t *Table) visibleRowHeights(rowHeight float32, rows int) (visible map[int]
 			height = h
 		}
 
-		if rowOffset <= t.offset.Y-height-theme.Padding() {
+		if rowOffset <= t.offset.Y-height-padding {
 			// before visible content
 		} else if rowOffset <= headHeight || rowOffset <= t.offset.Y {
 			minRow = i
@@ -752,7 +755,7 @@ func (t *Table) visibleRowHeights(rowHeight float32, rows int) (visible map[int]
 			break
 		}
 
-		rowOffset += height + theme.Padding()
+		rowOffset += height + padding
 		if isVisible || i < stick {
 			visible[i] = height
 		}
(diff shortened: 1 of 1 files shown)


Jacalz (approved):
This does look fine on my end but it  would be good to get an opinion from @andydotxyz before merging :)

widget/table.go
```diff
@@ -487,7 +487,8 @@ func (t *Table) columnAt(pos fyne.Position) int {
 		pos.X += t.content.Offset.X
 		offX += t.stuckXOff
 	}
-	for x := offX; i < end; x += visibleColWidths[i-1] + theme.Padding() {
+	padding := theme.Padding()
```
andydotxyz, line 490: I admit I had assumed the compiler would perform this optimisation!

andydotxyz (commented):
I'm unsure about this... There are new functions to calculate the sticky row/col sizes - but the overall visible calculation methods still do the same thing.
So there may not be duplicate code exactly but it does seem to be duplicating effort and/or could be simplified so we don't do the same thing in pairs of methods.

widget/table.go
Bluebugs: It actually can not and never will be able to. As it is behind a lock and shared across multiple goroutine, it can not know when it is acceptable to reuse the previous result of the call and has to assume the value changed between two calls. 

In general, if the compiler can not prove that the value it is accessing can not be changed by anyone else, it will redo a fetch.

Bluebugs:
> I'm unsure about this... There are new functions to calculate the sticky row/col sizes - but the overall visible calculation methods still do the same thing.
> So there may not be duplicate code exactly but it does seem to be duplicating effort and/or could be simplified so we don't do the same thing in pairs of methods.

Returning a map and passing it around kind of make that difficult. I could not manage to easily remove it as the header is potentially at an index far away from the rest and the user of the function are passing data all over. It might be doable with further refactoring to calculate the information about the header in one function, but not trivially and it will significantly increase the complexity of the pr. Maybe a follow up work.

andydotxyz (approved):

PR #3863, opened 2023-05-04
Animation smoothness improvement

Hypnotriod:
### Description:
Moved animation calculations from the separate routine to the drawing routine, to be sync with the draw call. Eliminate the race conditions and improve the animation smoothness. (FPS wasn't changed)

### Checklist:
- [x] Tests included. `TestGLDriver_StartAnimation`, `TestGLDriver_StopAnimation`, `TestGLDriver_StopAnimationImmediatelyAndInsideTick` are **modified**!
- [ ] Tests all pass. The only `4106/4114` tests are `passed`, as well as `4106/4114` tests are `passed` in the `master` branch

### Warning
Since I'm not fully familiar with the project structure, so I can't cover all the cases where this improvement can cause issues. I'm not insisting on this commit to be merged, but may be used as an solution example.

### DIFF
--- a/internal/animation/animation_test.go
+++ b/internal/animation/animation_test.go
@@ -14,6 +14,8 @@ import (
 )
 
 func TestGLDriver_StartAnimation(t *testing.T) {
+	// ticker simulates animations ticker goroutine
+	ticker := time.NewTicker(time.Second / 60).C
 	done := make(chan float32)
 	run := &Runner{}
 	a := &fyne.Animation{
@@ -24,6 +26,8 @@ func TestGLDriver_StartAnimation(t *testing.T) {
 
 	run.Start(a)
 	select {
+	case <-ticker:
+		run.TickAnimations()
 	case d := <-done:
 		assert.Greater(t, d, float32(0))
 	case <-time.After(100 * time.Millisecond):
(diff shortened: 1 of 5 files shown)


Hypnotriod:
As far as I understood it was done intentionally, so user can stop/launch animations from the animation `Tick` callback without deadlock. The only thing `Stop` does, is modify the `stopped` flag with `atomic.StoreUint32` call. So in the worst case animation will tick one extra time, I guess...

andydotxyz (commented):
Does this only apply to
The GL Driver? With this new public api to tick I worry that it's broken or impacted the mobile driver.

Hypnotriod:
> Does this only apply to The GL Driver? With this new public api to tick I worry that it's broken or impacted the mobile driver.

Sorry, but yes, I only did this for the GL driver only.
Does the mobile package have its own loop? Where should I look?

Hypnotriod:
I guess I figured it out.

andydotxyz:
@Hypnotriod I just noticed this is on the wrong branch - please re-play it on develop where we do our development work (see https://github.com/fyne-io/fyne/wiki/Contributing).

PR #3864, opened 2023-05-04
Only call updateLayout once on the parent.

Bluebugs:
### Description:
updateLayout was called for each children that has a min size change instead of one time once all the children have been updated. This lead to some large performance impact on Tree specifically, but all nested container with large number of object would have been impacted.

### Checklist:
- [x] Tests all pass.

### DIFF
--- a/internal/driver/common/canvas.go
+++ b/internal/driver/common/canvas.go
@@ -64,8 +64,6 @@ func (c *Canvas) EnsureMinSize() bool {
 	if c.impl.Content() == nil {
 		return false
 	}
-	var lastParent fyne.CanvasObject
-
 	windowNeedsMinSizeUpdate := false
 	csize := c.impl.Size()
 	min := c.impl.MinSize()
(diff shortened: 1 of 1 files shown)


internal/driver/common/canvas.go
```diff
@@ -515,6 +509,8 @@ type RenderCacheNode struct {
 	// it should free all associated resources when released
 	// i.e. it should not simply be a texture reference integer
 	painterData interface{}
+
+	needUpdateLayout bool
```
Jacalz, line 513: Does this need to be within the struct for recursion? It looks to me like it only is needed within the function above and could be a local variable instead, or am I missing something?

Jacalz (commented):
Nicely done. Just one comment in-line on the struct field perhaps not being necessary?

Jacalz:
Also, please remember to check the checkmarks in the PR template.

andydotxyz:
> Also, please remember to check the checkmarks in the PR template.

And don't remove the "tests included" its presence is not optional. If there are no tests, leave it there unchecked so we are fully aware that there are no new tests for this PR.

internal/driver/common/canvas.go
Bluebugs: Indeed, good idea.

Jacalz (approved):

PR #3866, opened 2023-05-04
Close manifest after writing it so it can be removed later in code

roffe:
Fixes a bug where the fyne package command would fail on Windows with error:

failed to remove manifest: remove C:\Users\roffe\go\src\github.com\roffe\t7logger\t7logger.exe.manifest: The process cannot access the file because it is being used by another process.

### DIFF
--- a/cmd/fyne/internal/commands/package-windows.go
+++ b/cmd/fyne/internal/commands/package-windows.go
@@ -63,10 +63,10 @@ func (p *Packager) packageWindows(tags []string) error {
 			CombinedVersion: p.combinedVersion(),
 		}
 		err := templates.ManifestWindows.Execute(manifestFile, tplData)
-		manifestFile.Close()
 		if err != nil {
 			return fmt.Errorf("failed to write manifest template: %w", err)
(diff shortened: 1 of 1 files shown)


cmd/fyne/internal/commands/package-windows.go
```diff
@@ -63,6 +63,7 @@ func (p *Packager) packageWindows(tags []string) error {
 			CombinedVersion: p.combinedVersion(),
 		}
 		err := templates.ManifestWindows.Execute(manifestFile, tplData)
+		manifestFile.Close()
```
Jacalz, line 66: Can you please do the close after the error? It is better to handle the error where it is created.

Jacalz (changes requested):
Thanks for the PR. I left a comment in-line about better placement for the close.

Jacalz (approved):
Thanks for the fix and congratulations on landing your first PR to Fyne :)

 I wish that the static analysis would catch unclosed readers and writers. I suspect there might be more places where we are missing close calls.

PR #3871, opened 2023-05-06
Update fsnotify to v1.6.0

Jacalz:
### Description:

Update fsnotify to [v1.6.0](https://github.com/fsnotify/fsnotify/releases/tag/v1.6.0) for improved cpu and memory usage as well as some fancy new syntax that is a lot more readable.

I tested this locally by using fyne_settings, fyne_demo while also removing the settings file.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Tested on macOS as well, working well

PR #3872, opened 2023-05-06
Update the golang.org/x/* deps (except for mobile) 

Jacalz:
### Description:

Now that Go 1.17 is the lowest supported version, we can finally update the `golang.org/x/*`  packages for various improvements and cleanups. The only one not updated was `golang.org/x/mobile` because I don't have any mobile device to test on and don't feel comfortable doing that update.

I have tested `cmd/fyne` and `cmd/fyne_demo` with these changes. Every update is in a separate commit and everything seems to be working as expected.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
We probably need to figure out here the x/mobile is coming from as we have our internal fork so if both are compiled in we are wasting space.

Jacalz:
We only seem to be using the `golang.org/x/mobile/exp/f32` package within some goxjs file. Nothing else exists in the vendor folder. I went ahead and updating it as it was mostly just cleanups and formatting changes (https://github.com/golang/mobile/commits/master/exp/f32).

Jacalz:
It is also worth noting that I don't think we have that package locally within the repository.

andydotxyz (approved):

PR #3875, opened 2023-05-07
Add desktop systray quit menu item only if it's not already present

bzdtchk:
### Description:
Add desktop systray quit menu item only if it's not already present

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/internal/driver/glfw/driver_desktop.go
+++ b/internal/driver/glfw/driver_desktop.go
@@ -125,12 +125,7 @@ func (d *gLDriver) refreshSystray(m *fyne.Menu) {
 	systray.ResetMenu()
 	d.refreshSystrayMenu(m, nil)
 
-	systray.AddSeparator()
-	quit := systray.AddMenuItem("Quit", "Quit application")
-	go func() {
-		<-quit.ClickedCh
-		d.Quit()
-	}()
+	addMissingQuitForMenu(m, d)
 }
 
 func (d *gLDriver) refreshSystrayMenu(m *fyne.Menu, parent *systray.MenuItem) {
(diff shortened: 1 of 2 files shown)


internal/driver/glfw/driver_desktop.go
```diff
@@ -183,3 +178,25 @@ func catchTerm(d *gLDriver) {
 		break
 	}
 }
+
+func addMissingQuitForMenu(menu *fyne.Menu, d *gLDriver) {
```
andydotxyz, line 182: This looks like a duplicate of `addMissingQuitForMainMenu` but with the first parameter different.

Would it not make sense to call one from the other? The `Menu` is found in `menus.Items[0]` so it should be a simple call.

andydotxyz (changes requested):
Thanks this seems like a good idea, however let's not duplicate the method to use it in system tray. The quit item is, after all, appended to a `Menu` in both cases.

internal/driver/glfw/driver_desktop.go
andydotxyz: I don't know how, but I haven't managed to explain. The following diff seemed to work for me:

```diff
index f46945175..9d7624973 100644
--- a/internal/driver/glfw/menu.go
+++ b/internal/driver/glfw/menu.go
@@ -15,30 +15,6 @@ func buildMenuOverlay(menus *fyne.MainMenu, w *window) fyne.CanvasObject {
 }
 
 func addMissingQuitForMainMenu(menus *fyne.MainMenu, w *window) *fyne.MainMenu {
-       var lastItem *fyne.MenuItem
-       if len(menus.Items[0].Items) > 0 {
-               lastItem = menus.Items[0].Items[len(menus.Items[0].Items)-1]
-               if lastItem.Label == "Quit" {
-                       lastItem.IsQuit = true
-               }
-       }
-       if lastItem == nil || !lastItem.IsQuit { // make sure the first menu always has a quit option
-               quitItem := fyne.NewMenuItem("Quit", nil)
-               quitItem.IsQuit = true
-               menus.Items[0].Items = append(menus.Items[0].Items, fyne.NewMenuItemSeparator(), quitItem)
-       }
-       for _, item := range menus.Items[0].Items {
-               if item.IsQuit && item.Action == nil {
-                       item.Action = func() {
-                               for _, win := range w.driver.AllWindows() {
-                                       if glWin, ok := win.(*window); ok {
-                                               glWin.closed(glWin.view())
-                                       } else {
-                                               win.Close() // for test windows
-                                       }
-                               }
-                       }
-               }
-       }
+       addMissingQuitForMenu(menus.Items[0], w.driver)
        return menus
 }
```

andydotxyz: However make sure that the full quit action is taken - the version you added is not as thorough as the one in the original main menu quit code...

andydotxyz (changes requested):
I'm not sure why another review was requested - the duplicate code I tried to help with is still present.

Jacalz:

Issue #3870, opened 2023-05-05
Panic when using autogenerated quit in menu if some windows have not been shown


imran-iq:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

Hi :wave: . First off amazing library, I am having a lot of fun building a UI with it.

I am running into a bug regarding the menu (I think). If I run the code provided and quit the app from the main menu without showing the second window at least once. The program will panic/sigsegv on exit with the following:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x99e754]

goroutine 12 [running]:
github.com/go-gl/glfw/v3.3/glfw.(*Window).SetShouldClose.func1(0x10000000045f8d6?)
        /home/iiqbal/go/pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20221017161538-93cebf72946b/window.go:380 +0x14
github.com/go-gl/glfw/v3.3/glfw.(*Window).SetShouldClose(0x9eaa20?, 0x10?)
        /home/iiqbal/go/pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20221017161538-93cebf72946b/window.go:380 +0x25
fyne.io/fyne/v2/internal/driver/glfw.(*window).closed(0xc00004a200?, 0x0?)
        /home/iiqbal/go/pkg/mod/fyne.io/fyne/v2@v2.3.4/internal/driver/glfw/window_desktop.go:320 +0x25
fyne.io/fyne/v2/internal/driver/glfw.addMissingQuit.func1()
        /home/iiqbal/go/pkg/mod/fyne.io/fyne/v2@v2.3.4/internal/driver/glfw/menu.go:35 +0x87
fyne.io/fyne/v2/widget.(*menuItem).trigger(0xc000538240)
        /home/iiqbal/go/pkg/mod/fyne.io/fyne/v2@v2.3.4/widget/menu_item.go:205 +0x3c
fyne.io/fyne/v2/widget.(*menuItem).Tapped(0xc00008bfb0?, 0xc0004cd801?)
        /home/iiqbal/go/pkg/mod/fyne.io/fyne/v2@v2.3.4/widget/menu_item.go:151 +0x58
[...]

PR #3878, opened 2023-05-08
Guard for nil when closing windows

imran-iq:
### Description:

As described in: https://github.com/fyne-io/fyne/issues/3870#issuecomment-1536719761
When the autogenerated quit is called from the main menu, if any windows were created but not shown, the program with panic. This is because the viewport would be nil.

This can be fixed with a nil check in `closed`

Fixes  https://github.com/fyne-io/fyne/issues/3870

### Checklist:

- [x] Tests included.
- [ ] Tests all pass.

The test that I added fails without my added changes, verified with `go test fyne.io/fyne/v2/internal/driver/glfw -run ClosedBeforeShow` however on my local machine I have some tests that fail irregardless of my changes (padding/resizing/hoverable) so I am not checking the last box

### DIFF
--- a/internal/driver/glfw/window_desktop.go
+++ b/internal/driver/glfw/window_desktop.go
@@ -318,7 +318,9 @@ func (w *window) refresh(_ *glfw.Window) {
 }
 
 func (w *window) closed(viewport *glfw.Window) {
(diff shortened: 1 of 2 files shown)


andydotxyz (approved):
Excellent thanks

andydotxyz:
Many thanks and congratulations on your first Fyne patch :)

PR #3888, opened 2023-05-14
Support disabling of animation - either as a user or developer

### LINKED ISSUE #1813: Add option to disable UI animations
This should not turn off all animation usage, but should be made available to UI so that animations that think they are optional can opt out of displaying and just show the last frame instead.

For example:
* Tab indicator
* Text cursor
* Buttton taps

Non-optional:
* progress bar

andydotxyz:
A user configuration is available - or a new `no_animations` build tag...

Fixes #1813

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -1560,6 +1560,9 @@ type entryContentRenderer struct {
 }
 
 func (r *entryContentRenderer) Destroy() {
+	if !fyne.CurrentApp().Settings().ShowAnimations() {
+		return
+	}
 	r.content.entry.cursorAnim.stop()
 }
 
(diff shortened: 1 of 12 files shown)


andydotxyz:
Just a ping on this @Jacalz or @Bluebugs - otherwise it will default approve tomorrow.

widget/entry.go
```diff
@@ -1560,6 +1560,9 @@ type entryContentRenderer struct {
 }
 
 func (r *entryContentRenderer) Destroy() {
+	if !fyne.CurrentApp().Settings().ShowAnimations() {
+		return
```
Jacalz, line 1564: Can this not leak if the user disables animations after one has started?

andydotxyz: Thanks for this. I had thought that, but loop forever is on so they won't end. I have removed the guard here so it will always try to stop.

Jacalz (approved):

PR #3889, opened 2023-05-15
Add boolean operator Not, And and Or data bindings.

Bluebugs:
### Description:
This add a logical Not, And and Or boolean data binding operator.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/data/binding/bool.go
+++ b/data/binding/bool.go
@@ -1,0 +1,118 @@
+package binding
+
+type not struct {
+	Bool
+}
+
+var _ Bool = (*not)(nil)
+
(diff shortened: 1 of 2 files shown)


andydotxyz (changes requested):
This is a nice addition, thanks.
However I think the "New" prefix trips up the reading a little. In other conversion constructors we have avoided it (like BoolToString) so I wonder if `binding.Not(binding.NewBool(...))` is more in-keeping.

andydotxyz (approved):

PR #3892, opened 2023-05-16
fyne box layout do care about user defined size

lazywhite:
### Description:

Feat: if user defined size for canvasObject, box layout will not just ignore them

### Checklist:

- [ x ] Tests included.
- [ x ] Lint and formatter run with no errors.
- [ x ] Tests all pass.

#### Where applicable:

- [ x ] Public APIs match existing style and have Since: line.
- [ x ] Any breaking changes have a deprecation path or have been discussed.
- [ x ] Updated the vendor folder (using `go mod vendor`).
- [ x ] Check for binary size increases when importing new modules.

### DIFF
--- a/layout/boxlayout.go
+++ b/layout/boxlayout.go
@@ -73,9 +73,9 @@ func (g *boxLayout) Layout(objects []fyne.CanvasObject, size fyne.Size) {
 			continue
 		}
 		if g.horizontal {
-			total += child.MinSize().Width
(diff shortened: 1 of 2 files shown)


andydotxyz (changes requested):
On reflection I don't think this approach can ever work... the container will no longer ever be able to shrink back down to the true MinSize for its content widgets.

You've basically implemented a second layer MinSize, which is not what Size is for

lazywhite:
You are right

PR #3894, opened 2023-05-17
Mention "go mod tidy" in README

thebigbone:
### Description:

Running ```go mod tidy``` will fix the missing dependencies and create go.sum file.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,10 @@ Using the standard go tools you can install Fyne's core library using:
 
     go get fyne.io/fyne/v2
 
+Run the following if there are any issues with the installation of dependencies:
+
+    go mod tidy
+
 # Widget demo
 
 To run a showcase of the features of Fyne execute the following:
@@ -177,10 +181,10 @@ However, if looking to support Fyne in a bigger way on your operating system the
 
 It is recommended that you install the following additional apps:
 
-| app | go install | description |
-| --- | ------ | ----------- |
-| fyne_settings | `fyne.io/fyne/v2/cmd/fyne_settings` | A GUI for managing your global Fyne settings like theme and scaling |
-| apps | `github.com/fyne-io/apps` | A graphical installer for the Fyne apps listed at https://apps.fyne.io |
+| app           | go install                          | description                                                            |
+| ------------- | ----------------------------------- | ---------------------------------------------------------------------- |
+| fyne_settings | `fyne.io/fyne/v2/cmd/fyne_settings` | A GUI for managing your global Fyne settings like theme and scaling    |
+| apps          | `github.com/fyne-io/apps`           | A graphical installer for the Fyne apps listed at https://apps.fyne.io |
 
 These are optional applications but can help to create a more complete desktop experience.
 
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Using the standard go tools you can install Fyne's core library using:
 
     go get fyne.io/fyne/v2
 
-Run the following if there are any issues with the installation of dependencies:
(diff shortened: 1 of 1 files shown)


andydotxyz:
If you "go get ..." and then immediately "go mod tidy" it will remove the fyne import from the go.mod file won't it?
The analysis will see that it's not yet used in the code and assume it wasn't intended.

thebigbone:
> If you "go get ..." and then immediately "go mod tidy" it will remove the fyne import from the go.mod file won't it? The analysis will see that it's not yet used in the code and assume it wasn't intended.

No. It keeps the fyne in go.mod. It just fixes any dependencies issue, in case fyne isn't installed correctly (which was the issue I faced)

thebigbone:
> If code to use fyne hasn’t been written they, yes, running “go mod tidy” will indeed do what @andydotxyz wrote.
> 
> However, you also don’t technically have to even run “go get” nowadays as writing the code and then running “go mod tidy” will download the dependency. I’m honestly not sure how we want to proceed.

I understand. Should I close this PR then?

Jacalz:
Not necessarily. I'm kind of leaning towards `go mod tidy` being better than `go get` as works in all cases. What do you think @andydotxyz?

andydotxyz:
That is only the right way to go if there is already fyne code written - but at that point in the instructions there won't be...

Jacalz:
Yeah sure. That's true. I'm not really saying that we should switch it to a `go mod tidy` there. It's just that from my point of view, the `go get` doesn't really do anything for the developer as it will be downloaded when running `go mod tidy` either way (and puts the go.mod file in a kind of limbo until that happens).

Most IDEs handles this rather well. The Go compiler has made some strange decisions in regards to module handling and I thing it is hard to write documentation for it in a good way. I wish that `go get` just did a tidy automatically (and that it shouldn't be on us to teach users how Go modules work) :/

andydotxyz:
Agreed on all those points yes.
I think we considered removing that line before but I'm not sure.
Alternatively we could have this new item in there as "in case of import or compile errors"?

thebigbone:
> Alternatively we could have this new item in there as "in case of import or compile errors"?

This would be a better choice.

Jacalz:
I agree as well. Sounds like a good plan 👍

thebigbone:
Any update on this?

Jacalz:
I thought we had decided that the changes mentioned in https://github.com/fyne-io/fyne/pull/3894#issuecomment-1552865260 had to me implemented but there have been no commits since then.

thebigbone:
I thought the first commit was supposed to be the new item. 
> Run the following if there are any issues with the installation of dependencies:
Should I just remove the above line and add the following line instead?
> In case of import or compile errors, run the following

Jacalz:
I'm personally thinking about something like this: "After having imported a new module and before compiling the code for the first time, you need to run the following command: `go mod tidy`". Don't take it word for word as there is likely a better way to say it. Basically, my idea is that we need to say that changes to the imported modules require `go mod tidy` to be run, but also it shouldn't be done before writing any code that uses the module (as that removes the dependency as mentioned above).

thebigbone:
Will the new commit work?

Jacalz:
Thanks. I'm happy with that. How about you @andydotxyz?

andydotxyz (approved):

andydotxyz:
> Thanks. I'm happy with that. How about you @andydotxyz?

Yeah thanks

Jacalz (approved):

Issue #3896, opened 2023-05-19
theme.*Resource wrapper types call svg.Colorize() on raster resources


davidhaymond:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

When using raster images as resources such as button icons, they are assumed to be SVG, causing errors like the following to be logged:

```
2023/05/19 11:10:44 Fyne error:  could not load SVG, falling back to static content:
2023/05/19 11:10:44   Cause: XML syntax error on line 3: invalid UTF-8
2023/05/19 11:10:44   At: /home/david/go/pkg/mod/fyne.io/fyne/v2@v2.3.4/internal/svg/svg.go:27
```

This happens because the `Content()` methods of the `theme.*Resource` wrapper types in `theme/icons.go` pass the content of the underlying resource to `svg.Colorize()` without checking to see if the underlying resource is SVG. This causes `svg.Colorize()` to log XML parsing errors. The following types have this problem:

- `theme.ThemedResource'
- `theme.InvertedThemedResource'
- `theme.ErrorThemedResource'
- `theme.PrimaryThemedResource'
- `theme.DisabledResource'

### How to reproduce

1. Create a button with a raster icon.
2. Cause one of the themed resource wrapper types to be used (e.g. disable the button to cause the icon to be wrapped in a `theme.DisabledResource`).

See the example code below.

### Example code

```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Hello World")

	icon, _ := fyne.LoadResourceFromURLString("https://mirrors.creativecommons.org/presskit/icons/cc.png")
	btn := widget.NewButtonWithIcon("", icon, func() {})
	btn.Disable()
	w.SetContent(btn)
	w.ShowAndRun()
}
```

### Fyne version

2.3.4

### Go compiler version

1.20.4

### Operating system and version

EndeavorOS

andydotxyz:
This is an interesting point.
I wonder if perhaps logging an error is correct, but that it should be more appropriate.
Otherwise won't we get more bugs opened to say the raster icon is not disabling when a button is disabled? :)

davidhaymond:
[...]

PR #3897, opened 2023-05-19
Add documentation comment to button Icon property

dweymouth:
### Description:
Add a documentation comment to the button's Icon property to specify that SVG resources will update their appearance with the theme and disabled state, while raster resources will not.

Related to #3896 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/button.go
+++ b/widget/button.go
@@ -66,6 +66,9 @@ var _ fyne.Focusable = (*Button)(nil)
 type Button struct {
 	DisableableWidget
 	Text string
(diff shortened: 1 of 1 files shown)


andydotxyz (changes requested):
Thanks for this, but perhaps "bitmap" could be used instead of "raster" to avoid the collision with a canvas primitive of the same name?

dweymouth:
I think this comment doesn't make sense to have on Button, it might make sense to put it on ThemedResource instead

PR #3915, opened 2023-05-24
Some better wording for the changelog

Jacalz:
### Description:

This improves some wording for the changelog of v2.3.5.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
This isn't how we usually word it. I missed the "### Fixed" header which would have made more sense but normally the issues are describing the problem not the fix I think.

Jacalz:
If we are to describe a log of changes, isn’t it better to describe what we have done/fixed instead of what the problem was? It seems clearer to me and probably does so as well for users/developers with less knowledge of the internals of Fyne

Jacalz:
But maybe my comment above only applies for when there is no ###fixed label? I think you might be right but I still believe that there are some improvements to be made

Jacalz:
Neve mind. Ignore my two previous messages. I see what you mean and I believe to have addressed your concerns :)

andydotxyz (approved):
Thanks

PR #3921, opened 2023-05-27
Adjust scrolling speed for non-Mac desktop OSes 

### LINKED ISSUE #3492: Enhance scroll wheel/touchpad scroll on desktop

dweymouth:
### Description:
Fresh PR based off of pull #3757, fixing incorrect base branch (now based off of develop), and using conditional compilatino rather than `runtime.GOOS` checks.

Fixes #3492 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


dweymouth:
No idea why the goimports check is failing - maybe someone can take a look?

internal/driver/glfw/window.go
```diff
@@ -17,9 +17,6 @@ import (
 )
 
 const (
-	scrollAccelerateRate   = float64(5)
-	scrollAccelerateCutoff = float64(5)
-	scrollSpeed            = float32(10)
 	doubleClickDelay       = 300 // ms (maximum interval between clicks for double click detection)
 	dragMoveThreshold      = 2   // how far can we move before it is a drag
 	windowIconSize         = 256
```
Jacalz, line 22: This is the reason that `goimports` is failing. With the long names removed from the block, it wants the equal signs to be brought more to the left.

dweymouth: Yeah, makes sense. Usually VSCode does `gofmt` on save but I think it timed out since it hadn't finished loading the full repo state when I saved the file. Fixed now

internal/driver/glfw/scroll_speed_default.go
Jacalz, line 1: Does it not make more sense to call this `_darwin` instead? It seems a bit odd to me that the default file is the one that is only for one operating system.

Jacalz (changes requested):
Thanks for working on this. I just have one minor request for a better name of one of the files.

internal/driver/glfw/scroll_speed_default.go
dweymouth: Is the GLFW driver only used for desktop OSes? I wasn't sure if some other device/platform may be using the GLfW driver

Jacalz: Yup. The mobile operating systems use the `mobile` driver which is formed from the `gomobile` project. It is only enabled when building with `-tags mobile`.

dweymouth: I guess the build guards could be simplified then to `darwin` and `!darwin` and the files renamed to `scroll_speed_darwin.go` and `scroll_speed_default.go`. What do you think?

Jacalz: Sounds good to me

dweymouth: Done!

Jacalz (approved):
Thanks. This looks good to me. I'll stand by waiting for a second approval before we merge.

PR #3932, opened 2023-05-31
Lists of types in preferences API

andydotxyz:
### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/preferences.go
+++ b/app/preferences.go
@@ -109,6 +109,10 @@ func (p *preferences) loadFromFile(path string) (err error) {
 
 	p.InMemoryPreferences.WriteValues(func(values map[string]interface{}) {
 		err = decode.Decode(&values)
+		if err != nil {
+			return
+		}
+		convertStringLists(values)
 	})
 
 	p.prefLock.Lock()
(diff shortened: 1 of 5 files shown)


app/preferences.go
```diff
@@ -109,6 +109,10 @@ func (p *preferences) loadFromFile(path string) (err error) {
 
 	p.InMemoryPreferences.WriteValues(func(values map[string]interface{}) {
 		err = decode.Decode(&values)
+		if err != nil {
+			return
```
Jacalz, line 113: Should we not do a `fyne.LogError()` on the value?

Jacalz (commented):
Looks good. Seems like a nice addition. Just one comment inline.

app/preferences.go
andydotxyz: I don't think so, no - this is exiting the write before applying conversion with an error set to the return value on the parent function call. So the caller will get the error and should log it if required.

Jacalz: Alright. Sounds good.

Jacalz (approved):

PR #3934, opened 2023-06-01
Bug report template now mentions not using .zip files

Jacalz:
### Description:

I have noticed that some people send in their code examples as`.zip` files. This can be a security issue for us because downloading and extracting arbitrary archives can be dangerous. This adds it as one of the things not to do. 

(This targets `master` as it changes user facing data on GitHub)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Nice thanks

PR #3937, opened 2023-06-01
Improved name consistency

vikkio88:
### Description:
This was done to avoid error in Icon List on the docs.
on [HERE](https://developer.fyne.io/explore/icons) those icons are presented as `ViewZoomIn` and `ViewZoomOut`, but in the code they are `ZoomIn` and `ZoomOut` the doc is generated by looking at the file names so I decided to make the naming more consistent.

this is what it looks now
image

and this is where they are generated from https://github.com/fyne-io/developer.fyne.io/blob/master/_gen/genicons.sh#L14

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


andydotxyz (changes requested):
You cannot change a public API like this. I think the docs need fixed rather than changing exposed APIs

vikkio88:
I only changed the names, the API already exposes the ZoomIn and ZoomOut
names?

andydotxyz:
It looks like the API has changed:

image

vikkio88:
this is too much changes for little gain, so will close it

PR #3948, opened 2023-06-05
Support for responding to back button on Android

### LINKED ISSUE #2910: [Mobile] Add back button handling
On Android, to my knowledge the back button isn't "caught" as a keyboard event, and *always* quit (minimize) the app.
This prevent us for building "proper" apps where Back can be used to quit a submenu / go to the previous screen.

I'm not sure if just adding a KeyEvent (how ?) would be enough or if it is rooted deeper.
I don't think the other 2 buttons (recent-app & home) need handling because their use is identical whatever the app is running on screen and the OS probably lock them. There is also historically the "menu" button but it's not really used anymore I don't think it matters.
If someone have an idea I could take a shot at it.

I'm not sure if handling it as a key Event is the best option or goes well with the guidelines ? 
But I think it is absolutely essential we can at least block the behavior of "back button > exit the app", even better handle it. I've made an app for someone, their biggest annoyance is pressing the back button to exit a window/form (as would be normal in any app) and have the whole app minimized.

andydotxyz:
Still todo:

- [x] Discuss if keyboard code and/or shortcut is the right way to deliver
- [x] Make sure that apps by default will handle it correctly, unless app uses the event...
 
Fixes #2910 

### Checklist:

- [ ] Tests included. <- we can't test this native lifecycle yet
- [x] Tests all pass.

Current usage:

```go
	w.Canvas().SetOnTypedKey(func(ev *fyne.KeyEvent) {
		if ev.Name == mobile.KeyBack {
			log.Println("Intercepted and issued back")
			a.Driver().(mobile.Driver).GoBack()
		}
	})
```


andydotxyz:
Marking as work in progress as I think the two remaining TODO items do need discussion to get it right.

MatejMagat305:
you are bigger expert than me, but some time ago I meet (in my oppinion simpler) method for this:
```
int32_t app_handle_event(struct android_app* app, AInputEvent* event) {
    if (AKeyEvent_getKeyCode(event) == AKEYCODE_BACK) {
        // actions on back key
        return 1; // <-- prevent default handler
    };
    // ...
    return 0;
}
```
but as I now see it is incompatible struct type ...
I ask [GPT](https://chat.openai.com/share/be18d760-d9ca-4d1b-a444-46dbf3e16479) and:
```
#include <android_native_app_glue.h>

void handle_event(struct android_app* app, AInputEvent* event) {
    ANativeActivity* activity = app->activity;
    
    if (AKeyEvent_getKeyCode(event) == AKEYCODE_BACK) {
        // Actions on back key
        ANativeActivity_finish(activity);
    }
    // Handle other events
}

void ANativeActivity_onCreate(ANativeActivity* activity, void* savedState, size_t savedStateSize) {
    struct android_app* app = activity->instance;

    // Set the event handler
    app->onInputEvent = handle_event;

    // Your other initialization code

    // Main loop
    while (1) {
        int events;
        struct android_poll_source* source;

        while (ALooper_pollAll(-1, nullptr, &events, (void**)&source) >= 0) {
            if (source != nullptr) {
                source->process(app, source);
            }

            if (app->destroyRequested != 0) {
                // The activity is being destroyed, cleanup and exit the loop
                // Your cleanup code
                return;
            }
        }

        // Your main loop code
    }
}
```
do you think that would be working?

andydotxyz:
@MatejMagat305 which part are you trying to simplify?

One of the reasons it is not simple is because the back button is a specific event with it's own callback handler in the app on the Java side...

MatejMagat305:
sorry, my mistake

internal/driver/mobile/app/android.c
```diff
@@ -85,6 +86,7 @@ void ANativeActivity_onCreate(ANativeActivity *activity, void* savedState, size_
 		hide_keyboard_method = find_static_method(env, current_class, "hideKeyboard", "()V");
 		show_file_open_method = find_static_method(env, current_class, "showFileOpen", "(Ljava/lang/String;)V");
 		show_file_save_method = find_static_method(env, current_class, "showFileSave", "(Ljava/lang/String;Ljava/lang/String;)V");
+		finish_method = find_method(env, current_class, "finishActivity", "()V");
```
MatejMagat305, line 89: this should call `finish` instead of `finishActivity`, because it is crashing

andydotxyz: I don't think so... are you sure that you have built the new fyne command to package the app when testing this code?

MatejMagat305: oh, sorry, my tablet was restarted and now it going, really sorry

MatejMagat305:
it can be done like that toohttps://github.com/andydotxyz/fyne/pull/2

andydotxyz:
> it can be done like that toohttps://github.com/[andydotxyz/pull/2](https://github.com/andydotxyz/fyne/pull/2)

Are you sure about this? If you don't add:
```java
    @Override
    public void onBackPressed() {
```
then the default back behaviour will occur.

And without the ability to call the `GoBack` behaviour then someone intercepting the key cannot decide to go back from an activity vs issuing some other behaviour...

MatejMagat305:
> > it can be done like that toohttps://github.com/[andydotxyz/pull/2](https://github.com/andydotxyz/fyne/pull/2)
> 
> Are you sure about this? If you don't add:
> 
> ```java
>     @Override
>     public void onBackPressed() {
> ```
> 
> then the default back behaviour will occur.
> 
> And without the ability to call the `GoBack` behaviour then someone intercepting the key cannot decide to go back from an activity vs issuing some other behaviour...

I was try it with fyne/main and my changes and it catch it: 
back

MatejMagat305:
we can also insert `if callback == nill {return 0}`

andydotxyz:
Are you certain you're using the released `fyne` command that matches the base library version for this change, and not one that I was working on before? This code is inserted into the tool, not added in the library in the compiler.

The issue with going backwards isn't just if there is a callback or not, its so that the callback can decide whether or not to permit the back action... If the developer wants to display a confirmation before going back then how can you do it without providing the `GoBack` call?

MatejMagat305:
> 

well take or do not ...
return 0 is default behaviour, return 1 is not allow ...., if you want that behaviour you can done that callback will return boolean reprezentation of allow default ..., or call "finish" activity ...

andydotxyz:
> or call "finish" activity ...  

How will the Go code be able to call finish activity without the bridge code exposing that API?

MatejMagat305:
> > or call "finish" activity ...
> 
> How will the Go code be able to call finish activity without the bridge code exposing that API?

change 89 row in android.c to 		
`finish_method = find_method(env, current_class, "finish", "()V");` and it will close activity (without any change in C finish)...., I released  one principial problem when I try this:

you call `setCurrentContext` in `on_create` it is clewer, but when I was close activity and resume it context is probably new and you have not actual context, well you can therefore refuse my sugestion, but it would be good fix context to new ...

andydotxyz:
> make your origin like this:
> 
> ```go
>     w.Canvas().SetOnTypedKey(func(ev *fyne.KeyEvent) {
> 		if ev.Name == mobile.KeyBack {
> 			log.Println("Intercepted and issued back")
> 			a.Driver().Quit()
> 		}
> 	})
> ```

Unfortunately this is an over-simplification. Going "back" is not the same as quitting an app.

Also because our keyboard handlers don't return whether or not something was intercepted the underlying driver does not know if you are going to handle the back button or not.
I appreciate you working on this, but you haven't convinced me there is a simpler way than what I have proposed.

MatejMagat305:
> > make your origin like this:
> > ```go
> >     w.Canvas().SetOnTypedKey(func(ev *fyne.KeyEvent) {
> > 		if ev.Name == mobile.KeyBack {
> > 			log.Println("Intercepted and issued back")
> > 			a.Driver().Quit()
> > 		}
> > 	})
> > ```
> 
> Unfortunately this is an over-simplification. Going "back" is not the same as quitting an app.
> 
> Also because our keyboard handlers don't return whether or not something was intercepted the underlying driver does not know if you are going to handle the back button or not. I appreciate you working on this, but you haven't convinced me there is a simpler way than what I have proposed.

well, it is not my library ..., but please look at context at start and resume ..., that is hide and it would cause problems in futures ...

andydotxyz:
> but please look at context at start and resume ..., that is hide and it would cause problems in futures ...

I'm sorry I don't understand this. The lifecycle code already handles the context on resumed apps?

MatejMagat305:
> > but please look at context at start and resume ..., that is hide and it would cause problems in futures ...
> 
> I'm sorry I don't understand this. The lifecycle code already handles the context on resumed apps?

well I release that, if I close app (put on background) and open it again, the pointers to context and vm (clazz and jvm) are not actual ... (not work if you want call for example finish() on activity) the solution would be use function to set actual pointers: https://github.com/andydotxyz/fyne/pull/2/files#diff-5d878dec0c0c4e09096aef345b55e1b73598999dae4a4a73f851ac1c47b7cd33R123, again your solution is good and probably this error has not efekt to them, but for example bluetooth which I prepare it is important to have actual pointers ...

if you want to try it (what I mean), you can: 
1. copy code with Quit() (above)
2. copy https://github.com/MatejMagat305/fyne/tree/feature/backbutton to vendor
3. remove command `C.set_global()` from functions `onResume` and `onStart`: https://github.com/andydotxyz/fyne/pull/2/files#diff-5d878dec0c0c4e09096aef345b55e1b73598999dae4a4a73f851ac1c47b7cd33R123
4. compile and install to android
5. open and try exit - it will work
6. open again (from bg)
7. try to trap exit - it will probably not working because the pointers to clazz and vm are different

andydotxyz:
I still don't understand. I am using `RunNative` to call functions into Android. The VM and env parameters are passed in then so why would they be stale?

MatejMagat305:
> I still don't understand. I am using `RunNative` to call functions into Android. The VM and env parameters are passed in then so why would they be stale?

please try it (my code), my pure search show me that the VM and CTX working, but in some situation when I need request permision or do with actual pointers ...

or look the code in onCreate(...) save VM and CTX into global variables, after app moving into background and re-start app will get in onResume(...) new VM and CTX which are different from  global variables which are not make actual, this would be hiden error in some case ...

andydotxyz:
> or look the code in onCreate(...) save VM and CTX into global variables, after app moving into background (close app and reopen from bg) and re-start app will get in onResume(...) new VM and CTX which are different from global variables which are not make actual, this would be hiden error in some case ...

Yes you cannot (and should not) save these to global variables and assume they work forever.

andydotxyz:
I don't think we are on the same page. You can see that in the Fyne code `internal/driver/mobile/app/android.c` makes a call to `setCorrentContext` every time the activity is re-created.

So the context available in the package variables is always up to date...

andydotxyz:
> are not actualiting at onResume(...) something is wrong when I try call finish()

My apologies I finally saw this happen - it seemed to not always be a problem.

Hopefully https://github.com/fyne-io/fyne/pull/3948/commits/8b545a8ada78f8d41fb2d16dc3cc5446cbb8ed1f resolves that aspect of this PR?

andydotxyz:
As promised @Jacalz here is a video of the work :) 

https://github.com/fyne-io/fyne/assets/294436/6bea515a-14ab-40a6-b8be-37aeac8d52ac

andydotxyz:
Just for record the app in the video above is just this:

```go
package main

import (
	"fmt"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/driver/mobile"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Hello")

	count := 0
	hello := widget.NewLabel("Trapped back 0 times")
	w.Canvas().SetOnTypedKey(func(ev *fyne.KeyEvent) {
		if ev.Name == mobile.KeyBack {
			count++
			hello.SetText(fmt.Sprintf("Trapped back %d times", count))
		}
	})
	back := widget.NewButton("Go Back", nil)
	if drv, ok := a.Driver().(mobile.Driver); ok {
		back.OnTapped = func() {
			drv.(mobile.Driver).GoBack()
		}
	} else {
		back.Disable()
	}
		hello,
		back,
	))

	w.ShowAndRun()
}
```

Jacalz (approved):
Thanks. Well done. This looks great :)

MatejMagat305:
> > are not actualiting at onResume(...) something is wrong when I try call finish()
> 
> My apologies I finally saw this happen - it seemed to not always be a problem.
> 
> Hopefully [8b545a8](https://github.com/fyne-io/fyne/commit/8b545a8ada78f8d41fb2d16dc3cc5446cbb8ed1f) resolves that aspect of this PR?

nice, thank ...

maybe call origin function OnResume(...) in go is useless but thank for this fix

PR #3950, opened 2023-06-05
Clarify issue template to say “full stack trace”

Jacalz:
### Description:

Reporters should always include the full stack trace. Make it more obvious that it is the case (and so we can point to the template in case people don’t).

This targets the master branch so that any changes to the template can be seen here on GitHub.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Nice

Jacalz:
Thanks. I noticed that it may or may not have been helpful on one of the issues recently. It seems like the issue template really needs to be bullet proof without chance of misinterpretation 😅

PR #3952, opened 2023-06-06
fix: support for theme switching

herry-go:
Description: support for theme switching

Bug: replace the funtion Settings().SetTheme(theme.LightTheme()) or Settings().SetTheme(theme.DarkTheme())

Log: theme switch

 

### Description:

Fixes #(issue)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -14,7 +14,7 @@ body:
 - type: textarea
   attributes:
     label: Describe the bug
-    description: A clear and concise description about the bug. Please include the stack trace in the case of a crash.
+    description: A clear and concise description about the bug. Please include the full stack trace in the case of a crash.
   validations:
     required: true
 - type: textarea


andydotxyz:
This isn't needed and replicates deprecated functionality (`Settings.SetTheme(theme.LightTheme)`).
This overrides user preferences and is a bad UX. to specify a theme for your app use a custom theme.

herry-go:
> 这不是必须的，而且会复制已弃用的功能 ( `Settings.SetTheme(theme.LightTheme)`)。 这个会覆用户首选项目并且是糕点的用户体验，这就是我们停止支持它的原因。
> 
> 要为您的应用指定主题，请使用自定义主题。在您的主题中，您可以根据需要指定亮色或暗色。

This submission is to support dynamic theme switching, which is different from the previous feature because I found during actual testing that ('Settings. SetTheme (theme. LightTheme)') will call the funtion  -> initFonts(), which will cause font garbled. However, this commit will only switch themes

andydotxyz:
Dynamic theme switching is already supported (see fyne_demo). If there is a bug with garbled fonts then that is what should be fixed - no new APIs required.

herry-go:
> Dynamic theme switching is already supported (see fyne_demo). If there is a bug with garbled fonts then that is what should be fixed - no new APIs required.

Ok, thank you for your reply very much. I'm trying to fix the font garbled issue after theme switching

Issue #142, opened 2019-03-04
Add drag and drop support


andydotxyz:
Add the events for these gestures.

Explore further what's required to integrate with system DnD

boussou:
it would be useful to allow it  or not programmatically.

andydotxyz:
Can you expand on that thought? Do you mean that widgets should have an API for controlling if DnD is enabled? Do you mean for creating custom widgets or for buitlin ones?

boussou:
Sure. I would like to be able to decide if/when to allow clipboard & drag and drop.
I mean it could be important for security reasons for example. 
I think having that kind of setting available "framework wide" & set at startup would be sufficient.

rausm:
Would this (http://anton.maurovic.com/posts/win32-api-approach-to-windows-drag-and-drop/) help with the Windows side of things ?

andydotxyz:
Possibly relates to URI registry? How do we handle unknown or custom defined types etc

user1121114685:
it`s work now?

andydotxyz:
> it`s work now?

This has not been completed.
The work has been moved out to the release after the one we are working on now.

user1121114685:
> > it`s work now?
> 
> This has not been completed.
> The work has been moved out to the release after the one we are working on now.

Very much looking forward to

PresleyHank:
> 
> 
> > it`s work now?
> 
> This has not been completed.
> The work has been moved out to the release after the one we are working on now.

Look forward to drag and drop

arlin-github:
Look forward to drag and drop

arlin-github:
When will it be finished？

andydotxyz:
This does not have a completion date. It is temtatively placed in the release /after/ the one we are working on. This would suggest late 2021 - but nothing is committed yet.

dtrehas:
Please find an idea how java's  SWT (close to C) toolkit implemented Drag and Drop in Mac, GTK, Windows API:
 
https://github.com/eclipse/eclipse.platform.swt/tree/master/bundles/org.eclipse.swt/Eclipse%20SWT%20Drag%20and%20Drop

dtrehas:
> Please find an idea how java's SWT (close to C) toolkit implemented Drag and Drop in Mac, GTK, Windows API:
> 
> https://github.com/eclipse/eclipse.platform.swt/tree/master/bundles/org.eclipse.swt/Eclipse%20SWT%20Drag%20and%20Drop

[...]

PR #3954, opened 2023-06-07
Add desktop support for window wide drag and drop

Jacalz:
### Description:

This adds initial support for a kind of window-wide drag and drop support that we might be able to get into v2.4.0. We can probably expand drag and drop in the future fore use cases like having it in specific areas and within `CanvasObjects` but this is similar to how you can add shortcuts to the whole window or to individual widgets. The other parts of development will take time and I hope that this can be a good stop-gap in the meantime.

For #142

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/internal/driver/glfw/window_desktop.go
+++ b/internal/driver/glfw/window_desktop.go
@@ -100,6 +100,7 @@ type window struct {
 
 	onClosed           func()
 	onCloseIntercepted func()
+	onDropped          func(fyne.Position, []fyne.URI)
 
 	menuTogglePending       fyne.KeyName
 	menuDeactivationPending fyne.KeyName
@@ -143,17 +144,19 @@ func (w *window) CenterOnScreen() {
 }
 
 func (w *window) SetOnDropped(dropped func(pos fyne.Position, items []fyne.URI)) {
-	w.viewport.SetDropCallback(func(win *glfw.Window, names []string) {
-		if dropped == nil {
-			return
-		}
+	w.runOnMainWhenCreated(func() {
+		w.viewport.SetDropCallback(func(win *glfw.Window, names []string) {
+			if dropped == nil {
+				return
+			}
 
-		uris := make([]fyne.URI, len(names))
-		for i, name := range names {
-			uris[i] = storage.NewFileURI(name)
-		}
+			uris := make([]fyne.URI, len(names))
+			for i, name := range names {
+				uris[i] = storage.NewFileURI(name)
+			}
 
-		dropped(w.mousePos, uris)
+			dropped(w.mousePos, uris)
+		})
 	})
 }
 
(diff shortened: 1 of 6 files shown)


internal/driver/glfw/window_desktop.go
```diff
@@ -141,6 +142,23 @@ func (w *window) CenterOnScreen() {
 	}
 }
 
+func (w *window) SetOnDropped(dropped func(pos fyne.Position, items []fyne.URI)) {
+	runOnDraw(w, func() {
```
Jacalz, line 146: It seems like the passed function that I want to run on the draw thread never actually runs. The code within works fine if I put it within one of the other functions that already are passed to `runOnDraw()` in other places within the codebase. Any help on why this queue never gets us anywhere would be helpful.

window.go
```diff
@@ -70,6 +70,11 @@ type Window interface {
 	// Since: 1.4
 	SetCloseIntercept(func())
 
+	// SetOnDropped allows setting a window-wide callback to intercept dropped items.
```
andydotxyz, line 73: Probably "receive" not "intercept"

internal/driver/glfw/window_desktop.go
andydotxyz: Probably you shouldn't manually get GLFW contexts and issue calls. Normally the GLFW calls get issued on the main not the draw thread - draw is OpenGL context, main is the GLFW context. `w.viewport.SetDropCallback` should be the right way to call it...

andydotxyz (changes requested):
Just some thoughts inline.

However I have a bigger question... Is dropping to the window with no widget information useful? We don't have a way to get the widget at a passed position, so perhaps at a minimum we should send the widget it was over instead of just the raw position? (making it a relative position with object could be useful)...

Jacalz:
> However I have a bigger question... Is dropping to the window with no widget information useful? We don't have a way to get the widget at a passed position, so perhaps at a minimum we should send the widget it was over instead of just the raw position? (making it a relative position with object could be useful)...

I think it is useful as it is now. There are cases where it doesn't quite matter where someone drops a file (like in Rymdport where I basically just want to switch to the `Send` tab unless the user isn't there already) or when the whole window is like a big drop target (one example would be like some kind of game or such).

Sending the widget it was dropped over just seems like it would be a duplicated effort when we later on add support for a widget to register a drop callback, doesn't it? I personally think that this fits in quite nicely in its current state with potential for an interface that makes the widget receive dropped items.

andydotxyz (approved):
Sorry this one slipped through my review list.
Looks great now.

andydotxyz (approved):
Helpful doc is helpful :)

PR #3958, opened 2023-06-07
Specify custom desktop file fields in FyneApp.toml

Jacalz:
### Description:

This PR adds support for specifying custom metadata fields for Linux and BSD. It is currently hooked up to provide support for adding `GenericName`, `Categories`, `Comment`, and `Keywords` to the desktop files. These fields are the ones that I had to specify manually in [io.github.jacalz.rymdport.desktop](https://github.com/Jacalz/rymdport/blob/main/internal/assets/unix/io.github.jacalz.rymdport.desktop) for Flatpak packaging to be accepted.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


cmd/fyne_demo/FyneApp.toml
```diff
@@ -3,10 +3,16 @@
...
-  Build = 10
+  Build = 11
 
 [Development]
   HelperText = "This binary was build with debug symbol"
 
 [Release]
   HelperText = "This binary was build without debug symbol"
+
+[LinuxAndBSD]
```
Bluebugs, line 14: Why not `Freedesktop` here as that the specification they apply to.

Bluebugs (commented):
That's a great idea!

cmd/fyne_demo/FyneApp.toml
Jacalz: Andrew and I discussed it on a call and decided that we could more settings there in the future, unrelated to desktop files, it's the current naming. All of our users/developers may or may not know what Freedesktop is.

andydotxyz, line 14: If we make it "Linux,BSD" or "Linux&BSD" it will be machine parsable if we decide to match go's approach to this logic :).

andydotxyz (changes requested):
Apologies these pending comments did not get sent :(

cmd/fyne_demo/FyneApp.toml
Jacalz: I tried that but it does not seem to be supported. I tried `+`, `,`, and `&`  before opening this PR but neither seem to be supported.

cmd/fyne_demo/FyneApp.toml
```diff
@@ -3,10 +3,16 @@
...
+  GenericName = "Fyne Demo"
+  Categories = "Development;"
```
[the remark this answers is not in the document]
Jacalz: I don't know if it technically is required but the official documentation uses it.
https://freedesktop.org/wiki/Howto_desktop_files/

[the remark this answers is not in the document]
Jacalz: A TOML list sounds like a great idea :)

cmd/fyne_demo/FyneApp.toml
```diff
@@ -13,6 +13,6 @@
 
 [LinuxAndBSD]
   GenericName = "Fyne Demo"
-  Categories = "Development;"
+  Categories = ["Development"]
   Comment = "A demo of Fyne and its capabilities."
-  Keywords = "demo;fyne;"
+  Keywords = ["demo", "fyne"]
```
Bluebugs, line 18: Nice

Bluebugs (approved):

cmd/fyne_demo/FyneApp.toml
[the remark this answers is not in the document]
andydotxyz: > I don't know if it technically is required but the official documentation uses it.
> 
> https://freedesktop.org/wiki/Howto_desktop_files/

By using FDo formatting you still require folk to understand the spec - which we should avoid if possible just like how naming avoids it.

cmd/fyne_demo/FyneApp.toml
```diff
@@ -3,10 +3,16 @@
...
 [Development]
-  HelperText = "This binary was built with debug symbol"
+  HelperText = "This binary was built with debug symbols"
 
 [Release]
-  HelperText = "This binary was built without debug symbol"
+  HelperText = "This binary was built without debug symbols"
+
+[LinuxAndBSD]
+  GenericName = "Fyne Demo"
```
andydotxyz, line 15: What is generic name, and how often should it differ from the Name specified in the details section above?

cmd/fyne_demo/FyneApp.toml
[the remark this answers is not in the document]
Jacalz: Indeed but I've switched to TOML lists now so it should no longer be an issue.

cmd/fyne_demo/FyneApp.toml
Jacalz: For Rymdport, I have "File Transfer" as GenericName. Just like Firefox has "Web Browser". I hope that helps.

andydotxyz: Ah yea thanks. I guess maybe it should be "demo app" or "developer tools" so it's clearer why they are different in this example?

Jacalz: That's a good point. I'll get that fixed

Jacalz: Perhaps "Toolkit Demo" even?

andydotxyz: Sounds good

andydotxyz (approved):
Great addition thanks

Jacalz:
Thanks. This is going to make the release binaries in Geoffrey so much nicer for Rymdport's users :)

p4k1tz:
The documentation for custom metadata fields needs to be added to https://docs.fyne.io/started/metadata.html

Jacalz:
> The documentation for custom metadata fields needs to be added to https://docs.fyne.io/started/metadata.html

Please open an issue for that instead of commenting on a PR that was merged a year ago.

PR #3960, opened 2023-06-08
A few cleanups now that Go1.17 is the minimum version

Jacalz:
### Description:

This PR applies two changes that we can do with Go 1.17 as the base.
- Removes the `work_go114.go` file that isn't needed anymore.
- Removes the `vendor` folder now that modules are fully supported.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
Thanks for doing this. Should probably update the PR template line about vendor too?

PR #3965, opened 2023-06-11
widget: Add the GridWrap collection widget from @dweymouth

Jacalz:
### Description:

Initial implementation was by @dweymouth over at https://github.com/fyne-io/fyne-x/pull/56. This PR includes that work plus selection support and some of the improvements that I have done (see https://github.com/fyne-io/fyne-x/pull/61 and https://github.com/fyne-io/fyne-x/compare/master...Jacalz:gridwrap-selection?expand=1).

This was done as a PR directly to Fyne and not fyne-x as the `.super()` call from `widget.BaseWidget` wasn't public and not using it made the selection work bad. All of the changes in this PR compared to the widget in fyne-x can be found here: https://github.com/fyne-io/fyne-x/compare/master...Jacalz:gridwrap-selection?expand=1

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/gridwrap.go
+++ b/widget/gridwrap.go
@@ -73,8 +73,6 @@ func NewGridWrapWithData(data binding.DataList, createItem func() fyne.CanvasObj
 }
 
 // CreateRenderer is a private method to Fyne which links this widget to its renderer.
-//
-// Since: 2.4
 func (l *GridWrap) CreateRenderer() fyne.WidgetRenderer {
 	l.ExtendBaseWidget(l)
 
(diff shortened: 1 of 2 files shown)


andydotxyz:
I don't think we need the "Since: 2.4" line when it is a function attached to a type that has "Since: 2.4" do we?
I think in the past we've just put it on constructors and types until the functions are newly added to an older type.

Jacalz:
It should be fixed now. Feel free to review when you have the time for it :)

andydotxyz (approved):
Great to have this upstream, thanks

PR #3970, opened 2023-06-13
Fix/preflistoverwrite

andydotxyz:
Fix a crash where a preference list overwrite was not not tested.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


internal/preferences.go
```diff
@@ -270,9 +271,26 @@ func (p *InMemoryPreferences) remove(key string) {
...
-		p.lock.Unlock()
-		return
+	if reflect.TypeOf(value).Kind() == reflect.Slice {
+		s := reflect.ValueOf(value)
+		old := reflect.ValueOf(p.values[key])
+		if p.values[key] != nil && s.Len() == old.Len() {
+			changed := false
+			for i := 0; i < s.Len(); i++ {
+				if s.Index(i).Interface() != old.Index(i).Interface() {
+					changed = true
```
Jacalz, line 281: Would it not be wise to break out of the loop once we find that it has changed?
```suggestion
					changed = true
					break
```

Jacalz (changes requested):
One comment inline about not needing to go through the whole slice if we notice changes.

andydotxyz:
> Thanks. Nice catch. I left one comment inline about not needing to go through the whole slice if we notice changes.

Excellent thanks - applied

Jacalz (approved):

PR #3974, opened 2023-06-14
Fix issue where macOS retina windows would show blurry icons

andydotxyz:
They were generating at 1x scale instead of 2x (due to draw before canvas attached)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


internal/driver/common/canvas.go
```diff
@@ -98,7 +98,11 @@ func (c *Canvas) EnsureMinSize() bool {
 
 	ensureMinSize := func(node *RenderCacheNode, pos fyne.Position) {
 		obj := node.obj
-		cache.SetCanvasForObject(obj, c.impl)
+		cache.SetCanvasForObject(obj, c.impl, func() {
```
Bluebugs, line 101: I am concerned that this is running in the rendering thread. If the problem is just happening on SetContent, wouldn't calling Refresh there make sense?

andydotxyz: It happens long after SetContent.
The problem is that the canvas pixscale is not realised until the window draws - it seems to be around the 3rd time an image is refreshed in the setup process due to configuring and resize etc.

Bluebugs: Hum, I am wondering if this is a Mac behavior as the sequence for configuring/resizing is OS dependent and I haven't noticed a blurriness even on my screen where scale != 1. I would feel better about this change if it was a Refresh call applied on the entire canvas actually.

andydotxyz: Yes, this is currently only visible on macOS - that is currently the only OS that sets a pixel scale.
Just to be clear this is not canvas scale.
On a macOS retina screen with scale 1.5, for example:
* Widget size = 20
* Screen coordinate size = 30 // that's size * scale
* Actual pixel size = 60 // that's the screen coordinates * retina scaling (2)

I'm not entirely sure why refreshing the whole canvas is better than objects when they are added to the canvas, but I am pretty sure that would miss the case where objects move canvas, or are added after the initial setup as well. In this case we tell it to update as soon as we know what the pixel scale is.

andydotxyz:
Turns out this actually impacts HiDPI screens on other OS as well, Linux does show it on my laptop (where the screen is very highDPI and the system/user scale combo comes out at 2.8 (reported in fyne_demo)

Jacalz (commented):
I don't personally notice a difference on my 27" 1440p display as it perhaps isn't HIDPI enough. I don't think I have enough knowledge of this part of the code base to know if this is a good solution or not. If it's fine, I would prefer to leave the reviewing to @Bluebugs or @changkun.

andydotxyz:
I think you'd need a (close to) 4k screen or a retina macOS laptop/display. Once you get one of them it's obvious ;)

internal/driver/common/canvas.go
andydotxyz: Hmm, what should we do then?
The current state is not ok as we've lost crispness on retina/uhdpi screens so some action must be taken.

andydotxyz:
I spoke to @Bluebugs on a call earlier and we agreed that this is frustrating because it's corrected on the wrong thread but that we could not find a way to do it better.
As there is a regression in develop can we get this approved and if inspiration strikes us I can make a better solution.

Bluebugs (approved):

Jacalz (approved):
Sounds good then. Sounds like we want an issue opened for tracking that.
On another note, I can confirm that this fixes the image refresh issues.

PR #3976, opened 2023-06-14
layout: Avoid many calls to theme.Padding() and child.MinSize()

Jacalz:
### Description:

Some of the optimizations that @Bluebugs did for the collection widgets removed many calls to the relatively slow theme lookups. This PR applies this to layouts as well. On top of that, it also makes sure that we do not call `MinSize()` on child objects many times because we don't necessarily know how fast they will be (and function calls are generally slow etc.).

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Bluebugs (approved):
Nice!

PR #3983, opened 2023-06-17
Fix bugs with appTabs render

### LINKED ISSUE #3980: tabItem did not removed on screen after remove all tabItems from AppTabs
### Checklist

### Describe the bug

Remove all tabs from AppTabs, the last tab item did not remove on screen.

### How to reproduce

1、create new container.AppTabs without container.TabItems,  show it in window.
2、the view shown on window is an empty page with a seperator line on top. this line is between tabs and content.
3、create a new container.TabItems and add in AppTabs. it is ok while render on window.
4、remove this tabItem. The tab content is removed on window,  but the tab did not removed, the tab title still remain on window.

### Screenshots

before
add
bug

### Example code

```go
templabel := widget.NewLabel("Contents for this " +title + " Tab")
c := widget.NewButton("Close", func() {
	clearTab(title)     //Close Tap Item
})

tItem := container.NewTabItem(title,
	container.NewVBox(
		templabel,c),
)
tabContainer.Append(tItem)         //Add Tap Item
```

This is clearTab func done:
```go
tabContainer.RemoveIndex(idx)
tabContainer.Refresh()
keyWindow.Content().Refresh()
```

### Fyne version

v2.3.3

### Go compiler version

1.17.7

macOS Catalina 10.15.7

### Additional Information

I found the problem is this code : 
[https://github.com/fyne-io/fyne/blob/1834d704a305f4729cce65213f5b5c3098e1d993/container/apptabs.go#L278-L301](url)

which line is :
```
for i := len(r.appTabs.Items); i > 0; i-- {
```

I change to :
```
for i := len(r.appTabs.Items); i >= 0; i-- {
```

and the problem is fixed.

spxvszero:
### Description:

Fixes #3980

I also fix the appTabs indicator and seperator line does not show properly.

you can compare with this demo code : 
```go
func main() {
	a := app.New()
	w := a.NewWindow("Hello")

	closeTab1Btn := widget.NewButton("Close Tab1", nil)
	tab := container.NewAppTabs(
		container.NewTabItem("Tab 1", container.NewVBox(widget.NewLabel("Content of tab 1"),
			closeTab1Btn)),
	)
	closeTab1Btn.OnTapped = func() {
		tab.RemoveIndex(0)
	}

	left := container.NewVBox(widget.NewButton("Add", func() {
		btn := widget.NewButton("Close", nil)

		item := container.NewTabItem(fmt.Sprint(rand.Int()),btn)

		btn.OnTapped = func() {
			tab.Remove(item)
		}
		tab.Append(item)
	}))

	split := container.NewHSplit(left, tab)
	split.Offset = 0.2

	w.SetContent(split)
	w.Resize(fyne.NewSize(640,480))
	w.ShowAndRun()
}
```

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container/apptabs.go
+++ b/container/apptabs.go
@@ -279,7 +279,7 @@ func (r *appTabsRenderer) Layout(size fyne.Size) {
 	// Try render as many tabs as will fit, others will appear in the overflow
 	if len(r.appTabs.Items) == 0 {
 		r.updateTabs(0)
-	}else {
+	} else {
 		for i := len(r.appTabs.Items); i > 0; i-- {
 			r.updateTabs(i)
 			barMin := r.bar.MinSize()
@@ -396,7 +396,7 @@ func (r *appTabsRenderer) updateIndicator(animate bool) {
 		r.divider.Hide()
 		r.indicator.Hide()
 		return
-	}else {
+	} else {
 		r.divider.Show()
 		r.indicator.Show()
 	}
--- a/container/apptabs.go
+++ b/container/apptabs.go
@@ -277,18 +277,22 @@ type appTabsRenderer struct {
 
 func (r *appTabsRenderer) Layout(size fyne.Size) {
 	// Try render as many tabs as will fit, others will appear in the overflow
-	for i := len(r.appTabs.Items); i >= 0; i-- {
-		r.updateTabs(i)
-		barMin := r.bar.MinSize()
-		if r.appTabs.location == TabLocationLeading || r.appTabs.location == TabLocationTrailing {
-			if barMin.Height <= size.Height {
-				// Tab bar is short enough to fit
-				break
-			}
-		} else {
-			if barMin.Width <= size.Width {
-				// Tab bar is thin enough to fit
-				break
+	if len(r.appTabs.Items) == 0 {
+		r.updateTabs(0)
+	}else {
+		for i := len(r.appTabs.Items); i > 0; i-- {
+			r.updateTabs(i)
+			barMin := r.bar.MinSize()
+			if r.appTabs.location == TabLocationLeading || r.appTabs.location == TabLocationTrailing {
+				if barMin.Height <= size.Height {
+					// Tab bar is short enough to fit
+					break
+				}
+			} else {
+				if barMin.Width <= size.Width {
+					// Tab bar is thin enough to fit
+					break
+				}
 			}
 		}
 	}
--- a/container/apptabs.go
+++ b/container/apptabs.go
@@ -393,12 +393,8 @@ func (r *appTabsRenderer) buildTabButtons(count int) *fyne.Container {
 
 func (r *appTabsRenderer) updateIndicator(animate bool) {
 	if r.appTabs.current < 0 {
-		r.divider.Hide()
 		r.indicator.Hide()
 		return
-	} else {
-		r.divider.Show()
-		r.indicator.Show()
 	}
 
 	var selectedPos fyne.Position
--- a/container/apptabs.go
+++ b/container/apptabs.go
@@ -277,7 +277,7 @@ type appTabsRenderer struct {
 
 func (r *appTabsRenderer) Layout(size fyne.Size) {
 	// Try render as many tabs as will fit, others will appear in the overflow
-	for i := len(r.appTabs.Items); i > 0; i-- {
+	for i := len(r.appTabs.Items); i >= 0; i-- {
 		r.updateTabs(i)
 		barMin := r.bar.MinSize()
 		if r.appTabs.location == TabLocationLeading || r.appTabs.location == TabLocationTrailing {
@@ -389,8 +389,12 @@ func (r *appTabsRenderer) buildTabButtons(count int) *fyne.Container {
 
 func (r *appTabsRenderer) updateIndicator(animate bool) {
 	if r.appTabs.current < 0 {
+		r.divider.Hide()
 		r.indicator.Hide()
 		return
+	}else {
+		r.divider.Show()
+		r.indicator.Show()
 	}
 
 	var selectedPos fyne.Position
--- a/container/apptabs_internal_test.go
+++ b/container/apptabs_internal_test.go
@@ -13,6 +13,9 @@ import (
 func TestAppTabs_tabButtonRenderer_SetText(t *testing.T) {
 	item := &TabItem{Text: "Test", Content: widget.NewLabel("Content")}
 	tabs := NewAppTabs(item)
+	// bug report #3980 : buttons will be rendered while already empty tabs. after fixed it, there has another logic that
+	// buttons will not be rendered while there has not enough space, so we add a resize logic here to make it render the button.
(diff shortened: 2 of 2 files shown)


andydotxyz:
Thanks for this @spxvszero but you'll need to fix the failures in CI before it will get reviewed

container/apptabs_internal_test.go
```diff
@@ -13,6 +13,9 @@ import (
 func TestAppTabs_tabButtonRenderer_SetText(t *testing.T) {
 	item := &TabItem{Text: "Test", Content: widget.NewLabel("Content")}
 	tabs := NewAppTabs(item)
+	// bug report #3980 : buttons will be rendered while already empty tabs. after fixed it, there has another logic that
```
andydotxyz, line 16: I don't think we should leave bug references in code once they are resolved

spxvszero: well. i just want to explain why i need to add code in other test code, i wlll remove the bug reference

Jacalz:
I'm sorry that this has taken some time to review. It is indeed a good idea to request a re-review (top right on the little circling arrows) once you feel happy enough with the code to do so. Commits can go unnoticed depending on notification settings here on GitHub. It makes it a bit easier for us to know that we should have a look at it again :)

container/apptabs.go
```diff
@@ -277,18 +277,22 @@ type appTabsRenderer struct {
...
+					break
+				}
+			} else {
+				if barMin.Width <= size.Width {
+					// Tab bar is thin enough to fit
+					break
+				}
 			}
 		}
 	}
```
Jacalz, line 298: We prefer to have the main path not indented and instead keep any special cases in their own if-statements with a return statement instead. Something like this:
```go
	if len(r.appTabs.Items) == 0 {
		r.updateTabs(0)
		return
	}
	
        for i := len(r.appTabs.Items); i > 0; i-- {
	        r.updateTabs(i)
	        barMin := r.bar.MinSize()
	        if r.appTabs.location == TabLocationLeading || r.appTabs.location == TabLocationTrailing {
		        if barMin.Height <= size.Height {
			        // Tab bar is short enough to fit
			        break
		        }
	        } else {
		        if barMin.Width <= size.Width {
			        // Tab bar is thin enough to fit
			        break
		        }
	        }
        }
```

Jacalz (changes requested):
Thanks for this. I just have a minor comment about keeping the code in line with our preferred code style. Apart from that, it looks good to me

container/apptabs.go
spxvszero: @Jacalz Thanks for reviewing. It looks clear in this style. But we cannot return in this position, we need excute the rest code after `updateTabs`, so i wrapped for-loop in if-else statement even it looks not pretty.
https://github.com/fyne-io/fyne/blob/79968626fe10c3cd49b978bc2f0d635bc8e18d31/container/apptabs.go#L278-L305

Jacalz: Ah, I see now. Sorry about that. It seems like GitHub didn't display that part at all in the diff and didn't make it very clear that the function was longer than that.

spxvszero: it doesn't matter :)

Jacalz: Cool. I'll review this later today and approve if everything's working fine :)

container/apptabs.go
```diff
@@ -389,8 +393,12 @@ func (r *appTabsRenderer) buildTabButtons(count int) *fyne.Container {
 
 func (r *appTabsRenderer) updateIndicator(animate bool) {
 	if r.appTabs.current < 0 {
+		r.divider.Hide()
```
andydotxyz, line 396: Are you sure the divider should be hidden?
Just because there are no tabs does not mean that this widget becomes invisible. Won't the MinSize still leave room for it?

container/apptabs_internal_test.go
```diff
@@ -47,3 +47,28 @@ func Test_tabButtonRenderer_DeleteAdd(t *testing.T) {
 	tabs.SelectTab(item2)
 	assert.Equal(t, pos, indicator.Position())
 }
+
+func Test_tabButtonRenderer_EmptyDeleteAdd(t *testing.T) {
+	item1 := &TabItem{Text: "Test", Content: widget.NewLabel("Content")}
+	tabs := NewAppTabs()
+
+	//cause tab button was build with enough space, without enough space, btns will not created.
```
andydotxyz, line 55: This doesn't read very clearly.
```suggestion
	// ensure enough space for buttons to be created.
```

container/apptabs.go
spxvszero: @andydotxyz  I am pretty sure this divider should be hidden, show you the picture, it looks better after hidden the divider.
divide


And the tabs minSize calculation does not contains divider, so i think it is ok for doing this.
https://github.com/fyne-io/fyne/blob/79968626fe10c3cd49b978bc2f0d635bc8e18d31/container/tabs.go#L367-L399

andydotxyz: No other widgets exhibit the behaviour of appearing invisible if they have no content. I wonder what the use-case is for the bar disappearing if items are removed?

ismyhc: I agree it looks much better with the bar hidden, but my question is why would you ever have no tabs in this widget? In the case of not having tabs, it should probably be up to the user to hide the apptabs widget I think.

Jacalz: Yeah, no tabs in a tab container seems a bit strange. Most applications that I know of close the whole window if you close the last tab. That, or just hiding the tab widget itself if necessary, seems like something the developers can do already

ismyhc: @spxvszero Okay yeah, got it. I think in that case you'd either hide or not show the apptabs when there are none.

ismyhc: Any chance you were able to finish this up @spxvszero ? Sorry to bug, but would really like to see this get merged in as the current bug on develop looks terrible. 

I guess I could just use your fork instead. lol

Jacalz: It could be wise to separate out the bug fix and the behaviour change. I think it should make this land more quickly

spxvszero: @andydotxyz @ismyhc @Jacalz Sorry for replying so late. I think we cannot figure out all the use case and stand for their perspectives. But we can make it looks more natural and comfortable, so i did these change. I agree with @Jacalz , i should separate out the bug fix and the behavior change, i discard the behavior change.Thanks for reviewing.

ismyhc (approved):
This looks good. As suggested leaving the  behavior changes for another request.

andydotxyz (approved):
Great thanks

Jacalz:
FYI: I'll review and merge this later today.

Jacalz (approved):
This looks great. Thanks for the contribution and congratulations on landing your first PR to Fyne. I also want to add that I especially appreciated the test code that you added to the description. It made it nice and easy to review the bug fix part of this :)

PR #3986, opened 2023-06-18
Add helpers for square size and offset position

Jacalz:
### Description:

I have noticed that it is quite common to create sizes and positions that are "squared", with the same values in both places. This PR introduces helpers for simplifying code to make it less to write and easier to read as there is less information to parse on one line. Another benefit is that it allows there to be less theme lookups without having to cache the value in a variable first (less to write, simpler).

On top of introducing that API, this PR also adapts some existing code to use the new API.   

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/geometry.go
+++ b/geometry.go
@@ -42,8 +42,8 @@ func NewPos(x float32, y float32) Position {
 	return Position{x, y}
 }
 
-// NewSquarePos returns a newly allocated Position with the same x and y position.
-func NewSquarePos(length float32) Position {
+// NewSquareOffsetPos returns a newly allocated Position with the same x and y position.
+func NewSquareOffsetPos(length float32) Position {
 	return Position{length, length}
 }
 
--- a/geometry.go
+++ b/geometry.go
@@ -42,6 +42,11 @@ func NewPos(x float32, y float32) Position {
 	return Position{x, y}
 }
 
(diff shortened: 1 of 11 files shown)


andydotxyz (commented):
This is a nice helper, and SquareSize is clear. But is there a better name for SquarePos?

Jacalz:
My initial idea was `NewEqualPos` but went with square for consistent. Would you prefer that one instead?

andydotxyz:
I'm not really a fan, something about Equal implies a test or comparison I think.
"diagonal" may be relevant but I can't see a way to use it. 
In matrices such a behaviour may be called unit or identity - can that help?

Jacalz:
Yeah, me neither. It's hard to come up with a good name for it. I'll have to think some more about it.

Jacalz:
I discussed this with two colleagues at work. We concluded that `fyne.NewSquareDisplacedPos()` was the most correct we could think of. What do you think?

andydotxyz:
Interesting... what about Offset instead of Displaced as I think we use that term elsewhere.

andydotxyz (changes requested):
Sorry I missed this before but it needs the "Since".
Good otherwise

andydotxyz (approved):

PR #3989, opened 2023-06-19
run all tests on CI

toaster:
### Description:

This PR changes the platform tests to run all tests (at least on Linux).
To get the GLFW tests stable on Linux is not that easy because there is a kind of race/concurrency issue which I did not solve. However, we have a workaround which works in 99% of the cases (I still see failures in “patched” tests every now and then). There might be tests that I missed because they fail less often. These have to be adjusted when a failure occurs.

### Checklist:

- [ ] Tests included.
- 🤞 Tests all pass.


andydotxyz (approved):
Nice to have this all working thanks!

geometry_benchmark_test.go
```diff
@@ -35,32 +35,32 @@ func BenchmarkSize_Subtract(b *testing.B) {
 // because this slows down the speed by factor 10.
 func TestPosition_Add_Speed(t *testing.T) {
```
Jacalz, line 36: Would it not make more sense to make these into actual benchmarks (which are turned off by default) instead of adding a build tag for disabling them?

Jacalz (commented):
I couldn't quite stay away from a minor code-only (no testing) review during my time off :)

Jacalz (changes requested):
Wrong review type in that last comment. Sorry about that

geometry_benchmark_test.go
toaster: I’m undecided on this one. I think it is out-of-scope for this PR but probably would not veto a new PR changing this.

Jacalz: Alright. I don't necessarily agree that it is out of scope but I will approve and open a PR for that.

PR #3990, opened 2023-06-20
Improve performance on obtaining current theme

adamantike:
### Description:

The `current()` util is used by many functions in the `theme` package, and can be potentially called many times during normal app navigation.

Benchmark results for this refactor:
```shell
# Before
BenchmarkTheme_current-16       123340401                9.857 ns/op
# After
BenchmarkTheme_current-16       263862091                5.018 ns/op
```

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/theme/theme.go
+++ b/theme/theme.go
@@ -687,11 +687,16 @@ func (t *builtinTheme) Size(s fyne.ThemeSizeName) float32 {
 }
 
 func current() fyne.Theme {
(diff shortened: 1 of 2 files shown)


andydotxyz (commented):
This is an excellent catch, thanks so much!

andydotxyz (approved):

Issue #3879, opened 2023-05-09
Text not visible in fyne app


islippers:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

All text in any content tab or button is not visible. Not sure if there is a setting somewhere that needs to be set. 
I have also ran the fyne demo app with the same results. See screen shot. 

### How to reproduce

Run go run fyne.io/fyne/v2/cmd/fyne_demo@latest on linux mint.

### Screenshots

![image](https://github.com/fyne-io/fyne/assets/30820690/f9b93b03-6a30-4cd9-a696-097663875657)

### Example code

go run fyne.io/fyne/v2/cmd/fyne_demo@latest

### Fyne version

latest 

### Go compiler version

1.18

### Operating system and version

Linux Mint

andydotxyz:
One other person has reported this and it seems to relate to the graphics driver. Can you compare what hardware you are using and what OpenGL / graphics driver versions?

heejit:
I am using Windows7  vm via Virtual-Box in ubuntu 20.04 I have used 

fyne version develop : fyne.io/fyne/v2 v2.3.6-0.20230614212946-594e3eb17bc8

![Screenshot_2023-06-15_13-46-25](https://github.com/fyne-io/fyne/assets/11409425/ebe7f024-7223-4ecb-b36f-f74903a3f853)

Vm configuration:
![windows7-vm-Screenshot_system_config](https://github.com/fyne-io/fyne/assets/11409425/e6f63218-a333-49e3-ab69-3d85473e1763)

Jacalz:
Hi @heejit. Can you please test on `develop` now? I think it might have been resolved.

heejit:
I have tested with :  fyne.io/fyne/v2 v2.3.6-0.20230812072749-e63d4e8030ef 

it is working fine thanks.

PR #3991, opened 2023-06-20
Adding emoji support, in this iteration using an embedded font

### LINKED ISSUE #3790: Text doesn't render in Windows VM
### Checklist

### Describe the bug

Text doesn't render on Windows (11) VirtualBox VM running in an Ubuntu host. Where the text should be is just an empty space. Hardware acceleration is enabled for the VM. This is a regression since earlier Fyne versions on the same VM with the same settings rendered text just fine. This problem is not present in a non-VM install of Windows.

### How to reproduce

Create a Windows 11 VM inside VirtualBox on Ubuntu (no idea if this is present with other combos of virtualization software, host OS, etc). Run the latest fyne demo.

### Screenshots

[screenshot](https://user-images.githubusercontent.com/5303666/230256584-3f8a0df7-23b1-4588-8bd8-14c62c60373f.png)

### Example code

fyne demo @ latest

### Fyne version

2.3.3

### Go compiler version

1.19.4 windows/amd64

Windows 11 (VM guest)

### Additional Information

_No response_

andydotxyz:
Emoji support using a flat SVG file bundled in apps.
Additionally to get colour font I stepped out the single-channel-text work which had associated bugs anyhow... should discuss.

Source font files go from 6MB to 18MB (maybe we should see if go:embed helps?
Debug app size on this machine went up 29%, stripped release binary was around 27%.

This increase in size is something to discuss. On later iterations we will add font lookup support so apps can be packaged without these fonts, but for now I don't know if it's worth making emoji specifically as an optional compile element?

Fixes #3790, #3879

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
I cannot replicate the test failure on hyperlink focus - with 1.17 or 1.20 - anyone got ideas?

andydotxyz:
> I cannot replicate the test failure on hyperlink focus - with 1.17 or 1.20 - anyone got ideas?

Never mind, it was just temporary... no idea how that happened but it's gone

Jacalz:
As discussed on Slack, it would be nice to either have this `opt-in` or `opt-out`. The large binary size increase impacting a lot of applications, even those that have no need for emoji support, does not feel great. 

> Source font files go from 6MB to 18MB (maybe we should see if go:embed helps?

My testing that I did a while back seemed to indicate that it is ever so slightly bigger to use `go:embed` (we're talking kilobytes at most probably) because of the import of the embed package, I assume. Our own bundling is actually better there.

andydotxyz:
It's now opt-out with the `no_emoji` build tag, matching the new tags that Tilo has added for other disabling aspects

Jacalz (commented):
Are all the GL reverts supposed to be there? 

The code looks good otherwise and works great here. I'm "only" seeing a 22% size increase for stripped binaries actually. It looks better than I expected but the build tag is still a very nice addition 😀

andydotxyz:
Without any GL revert the emoji will be in greyscale because the text was moved to single channel.
I could have left some in and/or turned it off with a flag, but given the code that would be off is causing two (linked) bugs it seemed sensible to revert the optimisation and call them fixed at the same time.

Jacalz:
Ah, I see. I couldn't find any reference to why it was reverted so I thought it was a mistake

Jacalz (approved):

andydotxyz:
Sorry there was a merge conflict crept into go.mod, updated.
Can you re-approve @Jacalz ?

PR #3996, opened 2023-06-23
dialog: Add .Submit() to form dialog

Jacalz:
### Description:

This allows a form dialog to be hidden and submitted if all validators are passing. The new API cna be very useful when implementing an entry that should close the dialog when submitted. Just calling hide will run the OnClosed() with false in that case so the user will not be able to submit, even if the validation is passing.

Unblocks https://github.com/Jacalz/rymdport/issues/55

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/dialog/entry.go
+++ b/dialog/entry.go
@@ -9,7 +9,7 @@ import (
 //
 // Deprecated: Use dialog.NewFormDialog() or dialog.ShowFormDialog() with a widget.Entry inside instead.
 type EntryDialog struct {
-	*formDialog
+	*FormDialog
 
 	entry *widget.Entry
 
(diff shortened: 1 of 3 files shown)


andydotxyz:
Wouldn't this just be `FormDialog.Submit()`? I can't think of a use-case where a dialog stays open after the primary action has been performed.

Jacalz:
Good point. I'll rename the method.

andydotxyz:
Cool.

I do wonder if the failed validation behaviour is right though. If a form is submitted with validation failures it doesn't do anything right? So would it be OK if calling submit on a form dialog also didn't submit (and didn't hide). There will after all be errors on the screen.

Jacalz:
Sorry for the misunderstanding @andydotxyz. I believe that this should address your concern now :)

andydotxyz (approved):

aohanhongzhi:
I want to submit information in dialog.Form，but the backend server will check it . and when the backend server reponse error , I hope the dialog.Form will not dismiss and tip a message . so how should I to solve it ?

Jacalz:
@aohanhongzhi This is a pull request that was merged a month ago. It is not the right place to ask questions regarding API usage. If you have a question regarding Fyne, you should open a discussion question (using the GitHub discussion feature) or join the #fyne channel on slack.

PR #4006, opened 2023-06-26
Remove flaky tests that made sure benchmarks were fast enough

Jacalz:
### Description:

I have had these tests fail multiple times locally due to differences in hardware, compiler versions etc. Remove the tests that verify that benchmarks are fast enough and remove the no_bench tag now that there only are benchmarks (which are only enabled when running tests with -bench)

### Checklist:

- [ ] Tests included. Tests removed?
- [x] Tests all pass.

### DIFF
--- a/geometry.go
+++ b/geometry.go
@@ -52,6 +52,7 @@ func NewSquareOffsetPos(length float32) Position {
 // Add returns a new Position that is the result of offsetting the current
 // position by p2 X and Y.
 func (p Position) Add(v Vector2) Position {
+	// NOTE: Do not simplify to `return p.AddXY(v.Components())`, it prevents inlining.
 	x, y := v.Components()
 	return Position{p.X + x, p.Y + y}
 }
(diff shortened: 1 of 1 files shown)


andydotxyz (commented):
This seems like a reasonable change, to simplify things is good. But was there a reason (like in the comments) that they had to be here?

Jacalz:
I think it was there just to prove that the new methods were faster than the old ones. I never personally thought that the benchmark result test verification provided a whole lot given that the methods themselves are very unlikely to get slower than the old methods.

andydotxyz:
Maybe the warning that appears next to the removed tests should somehow be translated into the code warning against accidental slowdowns by simplifying the code?

Jacalz:
Good point. I had gotten the impression that the tests were testing the other way around. A simple NOTE comment should be fine. I also verified that Go 1.20 does mark the functions as not inline-able if the simplification is made (and a 10x slowdown). This should be fine for a re-review.

Jacalz:
I will open a bug report for the Go compiler. That slowdown does look like a genuine bug in the inliner to me.

andydotxyz (approved):

PR #4007, opened 2023-06-26
Add new truncation mode, ellipsis

andydotxyz:
And adjust text tests to be real measuring now as we can't fake it!

This is the beginning of a migration to go-text for line wrap and truncation, but this new mode was easier than migrating the existing ones...

It does not support multi-line truncation (adding ... on 2nd or 3rd line)
In fact a point of discussion is that the current truncation tests /do/ support multi-line but they do it by truncating every line which seems peculiar. Revisiting in a follow-on item

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/richtext.go
+++ b/widget/richtext.go
@@ -10,9 +10,13 @@ import (
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/canvas"
 	"fyne.io/fyne/v2/internal/cache"
+	paint "fyne.io/fyne/v2/internal/painter"
 	"fyne.io/fyne/v2/internal/widget"
 	"fyne.io/fyne/v2/layout"
 	"fyne.io/fyne/v2/theme"
+	"github.com/go-text/typesetting/di"
+	"github.com/go-text/typesetting/shaping"
+	"golang.org/x/image/math/fixed"
 )
 
 const (
@@ -634,6 +638,10 @@ func (r *textRenderer) Refresh() {
 			} else if i == len(bound.segments)-1 && len(bound.segments) > 1 {
 				txt.Text = string(runes[:bound.end])
 			}
+			if bound.ellipsis && i == len(bound.segments)-1 {
+				txt.Text = txt.Text + "…"
+			}
+
 			if concealed(seg) {
 				txt.Text = strings.Repeat(passwordChar, len(runes))
 			}
@@ -812,6 +820,10 @@ func findSpaceIndex(text []rune, fallback int) int {
 	return curIndex
 }
 
+func float32ToFixed266(f float32) fixed.Int26_6 {
+	return fixed.Int26_6(float64(f) * (1 << 6))
+}
+
 // lineBounds accepts a slice of Segments, a wrapping mode, a maximum line width and a function to measure line width.
 // lineBounds returns a slice containing the boundary metadata of each line with the given wrapping applied.
 func lineBounds(seg *TextSegment, wrap fyne.TextWrap, firstWidth, maxWidth float32, measurer func([]rune) float32) []rowBoundary {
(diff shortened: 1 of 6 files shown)


widget/richtext.go
```diff
@@ -934,15 +951,52 @@ func splitLines(seg *TextSegment) []rowBoundary {
...
+			lines = append(lines, rowBoundary{[]RichTextSegment{seg}, len(lines), low, high, false})
 			low = i + 1
 		}
 	}
-	return append(lines, rowBoundary{[]RichTextSegment{seg}, len(lines), low, length})
+	return append(lines, rowBoundary{[]RichTextSegment{seg}, len(lines), low, length, false})
+}
+
+func truncateLimit(s string, text *canvas.Text, limit int, ellipsis []rune) (int, bool) {
+	s = strings.ReplaceAll(s, "\r", "")
```
Jacalz, line 962: I am a bit worried about this for long texts. We are going to iterate over all of the text and allocate a new string that does not contain the carriage return characters. Is there are way to avoid this?

andydotxyz: Honestly I didn't even think as this code was copied out of an earlier shaping but if work. 
I don't think we can avoid iterating, but we could move to 2 pass where it iterates to check and if required the. Calls replace... but I suspect the call is already doing this.

I suspect we need bigger refactors here to avoid multi-passes - maybe there is a way to replace the returns in one pass earlier in the shaping.

Jacalz: I think it is good to open an issue for it then. If it is copied from earlier code then we need not stall this PR on it.

andydotxyz: Thanks. I looked at this again with a fresh mind and resolved it.
The source was improved to avoid the string replace - and the truncation shaping just ignores it, as the precision is not so important there (we will just truncate a little earlier if many \r are present...

andydotxyz: > The source was improved to avoid the string replace

OK, that wasn't quite possible as it broke kerning. Instead I put it in a `Contains` check to try and avoid the memory allocation.

I looked into the `ReplaceAll` which delegates to `WriteString` which means it cannot simply return same pointer when string doesn't contain the search string unfortunately. So this seemed like a good balance.

Jacalz: I think you are iterating twice if you check contains before replacing. Looking at the source, it does that check internally already. If there are carriage returns though, it will have to allocate a new string with all items replaced.

Jacalz: I can open issue for tracking it and similar issues in the codebase, later today

andydotxyz: I agree this absolutely is iterating twice. It does that to avoid allocating memory. From my reading of the internals it will create a new buffer the same size and copy into it - even if nothing changes.
I thought you wanted to avoid the allocation. As far as I can see it is impossible (using stdlib) to iterate only once and also to avoid any memory allocations for the happy path.

andydotxyz: p.s. it seems right that we are optimising for the case of no \r, which this now does.

Jacalz: Looking at the `strings.Replace()` function, which `ReplaceAll` uses, it looks to me like it counts the number of occurrences and avoids allocating if there aren't any.
```
	// Compute number of replacements.
	if m := Count(s, old); m == 0 {
		return s // avoid allocation
	}
```
It thus seems to me like we don't need the call to contains?

andydotxyz: Oh you're right, I misread the code.
Backed this out and it doesn't seem there is more to do.

Jacalz: Having a look at the [walkString](https://github.com/fyne-io/fyne/blob/1834d704a305f4729cce65213f5b5c3098e1d993/internal/painter/font.go#L260) function on the `master` branch, it seems like we already where ignoring the carriage return while looping over the characters. However, the [walkString](https://github.com/fyne-io/fyne/blob/03384a7a3aa88f328724e399301ad38b321dcd84/internal/painter/font.go#L159) function in `develop` does that replace call instead (which will basically allocate a whole new string if the string contains any carriage returns). That seems a bit like a regression to me. I really do think we should get that fixed. It is outside of the scope of this PR do you think?

Jacalz: I think I answered that question myself earlier in the thread. I'll open an issue :)

Jacalz: I see now that you removed the second replace call and the one that was there already is the only one. Nice.

Jacalz:
I'll try to test this tomorrow

Jacalz (approved):
Looks good nice work. Works rather well in the file dialog (I patched it up to sue this as a test) :) 
image

PR #4011, opened 2023-06-29
Add support for input in the Exec= field of desktop files

Jacalz:
### Description:

This allow the developer to pass any specific flags or arguments that are necessary but also allows passing [exec codes](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s07.html) to tell the desktop that the file supports things like opening folders.

Example of Rymdport using `%F` to support one or more files being opened:
[Screencast from 2023-06-29 20-54-29.webm](https://github.com/fyne-io/fyne/assets/25466657/2f6ae98f-7e78-4d85-a8d1-d89050896c5d)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/package-unix.go
+++ b/cmd/fyne/internal/commands/package-unix.go
@@ -19,6 +19,7 @@ type unixData struct {
 	Categories       string
 	Comment          string
 	Keywords         string
+	ExecInput        string
 }
 
 func (p *Packager) packageUNIX() error {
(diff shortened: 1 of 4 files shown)


andydotxyz:
Isn't it possible to do much more than input specification with this? The name sounds quite specific.
Maybe `ExecParams` or similar would be more applicable?

Jacalz:
Yeah, I hinted a bit about doing more than inputs in the PR description but I had thought about it as input to the `Exec` field. Using `ExecParams` seems like a much better name :)

Jacalz:
I can't quite decide if it makes sense with an array here as well. What do you think?

andydotxyz (approved):
Thanks. I don't think an array would make sense here as it just complicates space handling for what is a basic string output.

Jacalz:
Thanks. That was basically what made me think that a regular string would be better in some ways :)

PR #4012, opened 2023-07-02
Reduce updateItem calls

### LINKED ISSUE #3826: Add APIs for refreshing individial items in collections
### Checklist

This was discussed on the call yesterday. There are a few cases where you know which item index needs a refresh but the only way to refresh it is to refresh the whole collection using `.Refresh()`. We should add an API to refresh individual items in collections (cells in Table, rows in List and so on). As a bonus, the collection widget can be smart and avoid a refresh if the item is not currently visible.

It should be possible

Add an API to refresh individual items in collections (cells in Table, rows in List and so on).

SMQuazi:
### Description:
Currently, placing a print in the `updateItem` callback for a list shows that the function is being called for all items that are visible for every change made. This makes expensive tasks repeat whenever there is any sort of change to the list, and locks up the screen for a long time. This change aims to only call the function for a given item only once, when the item becomes visible. 

The unit test shows a slight difference, however placing a print statement in the updateItem function seems to show a drastically more apparent change, as show below. 

### also, Fixes
#3826  (partially - for lists only)

#### Print out of list IDs - on loading of a list
currently
```
7.9.0.6.3.4.5.8.10.1.2.5.8.9.3.4.6.7.10.0.1.2.1.3.5.8.7.9.10.0.2.4.6.0.1.2.4.10.3.5.6.7.8.9.3.4.7.8.10.0.2.6.9.1.5.0.3.5.7.10.1.2.4.6.8.9.7.8.9.0.3.4.5.6.10.1.2.6.7.8.2.3.4.9.10.0.1.5.6.7.8.9.0.2.5.10.1.3.4.8.9.10.1.2.4.5.7.0.3.6.1.2.3.4.5.6.7.9.10.0.8.0.5.8.10.1.2.3.4.6.7.9.10.0.1.2.4.5.6.7.3.8.9.1.3.5.10.9.0.2.4.6.7.8.3.5.7.8.9.0.2.4.6.10.1.3.4.8.10.0.2.5.6.7.9.1.8.10.2.7.3.4.5.6.9.0.1.1.2.3.10.9.0.4.5.6.7.8.1.2.5.8.0.3.4.6.7.9.10.8.9.4.5.7.3.6.10.0.1.2.0.1.2.4.8.3.5.6.7.9.10.3.7.0.2.5.6.8.9.10.1.4.1.3.7.9.10.0.2.4.5.6.8.5.6.8.10.0.2.4.9.1.3.7.10.0.2.8.9.6.7.1.3.4.5.1.3.5.8.10.0.2.4.6.7.9.
```

with the change
```
7.8.10.9.0.1.2.3.4.5.6.2.3.4.10.0.1.5.6.7.8.9.
```

####  Print out of list IDs - scrolling down 2 items
current
```
1.4.6.7.8.9.0.2.3.5.10.10.0.1.5.7.8.2.3.4.6.9.3.5.6.7.8.10.0.1.2.4.9.8.9.10.0.4.5.6.7.1.2.3.4.5.7.8.9.11.3.2.6.10.1.2.3.6.8.9.11.1.5.7.10.4.11.2.6.7.8.9.1.3.4.5.10.8.4.6.7.5.9.10.11.1.2.3.5.7.9.1.2.6.8.10.11.3.4.10.11.1.3.5.7.8.2.4.6.9.6.10.1.2.3.4.5.7.8.9.11.7.11.3.4.5.6.8.9.1.2.10.6.7.8.9.1.2.3.4.10.5.11.7.9.10.11.4.5.6.8.1.2.3.10.11.3.5.4.6.7.8.9.1.2.2.4.6.8.9.10.1.3.5.7.11.3.5.7.11.1.2.8.9.10.4.6.2.3.5.6.7.9.11.1.4.8.10.8.11.3.4.6.7.10.1.2.5.9.1.2.6.7.8.10.11.3.4.5.9.11.1.3.4.7.8.9.10.2.5.6.11.4.6.8.10.7.9.1.2.3.5.5.6.7.8.9.10.1.3.11.2.4.6.7.9.11.3.5.4.8.10.1.2.2.5.8.10.1.3.4.6.7.9.11.3.5.6.8.11.1.4.7.9.10.2.9.10.11.1.2.3.5.8.4.6.7.6.8.10.1.2.3.9.11.4.5.7.3.4.6.9.10.11.1.2.5.7.8.9.11.1.3.6.7.8.2.4.5.10.4.5.11.1.3.6.7.8.9.10.2.3.4.5.10.12.2.7.8.9.11.6.2.5.7.10.11.12.3.4.6.8.9.2.7.8.9.10.3.4.5.6.11.12.7.8.9.10.11.2.3.4.5.6.12.7.9.11.10.12.2.3.4.5.6.8.4.5.8.11.12.10.2.3.6.7.9.8.11.2.3.4.5.12.6.7.9.10.2.3.4.9.10.5.6.7.8.11.12.3.4.7.10.11.12.2.6.8.9.5.6.10.9.11.2.3.4.5.7.8.12.12.5.8.10.11.7.9.2.3.4.6.7.9.4.5.6.8.10.11.2.3.12.3.5.6.8.10.11.12.2.4.7.9.5.6.8.9.11.12.3.4.7.10.2.4.5.6.8.2.3.7.9.10.11.12.2.4.5.12.11.3.6.7.8.9.10.10.11.5.7.8.6.9.12.2.3.4.2.3.4.5.10.12.6.7.8.9.11.5.6.7.8.9.2.3.4.10.11.12.7.8.9.12.2.4.5.6.3.10.11.4.8.10.11.12.3.5.6.7.9.2.3.5.6.7.8.9.10.2.11.12.4.5.6.9.10.2.3.8.11.12.4.7.
``` 

with changes
```
11.12.
```

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -93,19 +93,23 @@ func (l *List) MinSize() fyne.Size {
 	return l.BaseWidget.MinSize()
 }
 
+// RefreshItem refreshes a single item, specified by the item ID passed in.
+//
+// Since 2.4
 func (l *List) RefreshItem(id ListItemID) {
 	if l.scroller != nil {
-		l.BaseWidget.Refresh()
-		lo := l.scroller.Content.(*fyne.Container).Layout.(*listLayout)
-		visible := lo.visible
-		canvas := fyne.CurrentApp().Driver().CanvasForObject(lo.list)
-		var focused fyne.Focusable
-		if canvas != nil {
-			focused = canvas.Focused()
-		}
-		if item, ok := visible[id]; ok {
-			lo.setupListItem(item, id, focused == item)
-		}
+		return
+	}
+	l.BaseWidget.Refresh()
+	lo := l.scroller.Content.(*fyne.Container).Layout.(*listLayout)
+	visible := lo.visible
+	canvas := fyne.CurrentApp().Driver().CanvasForObject(lo.list)
+	var focused fyne.Focusable
+	if canvas != nil {
+		focused = canvas.Focused()
+	}
+	if item, ok := visible[id]; ok {
+		lo.setupListItem(item, id, focused == item)
 	}
 }
 
@@ -671,9 +675,9 @@ func (l *listLayout) updateList(newOnly bool) {
 	l.renderLock.Unlock() // user code should not be locked
 
 	if newOnly {
-		for nId, nItem := range visible {
-			if _, ok := wasVisible[nId]; !ok {
-				l.setupListItem(nItem, nId, focused == nItem)
+		for row, obj := range visible {
+			if _, ok := wasVisible[row]; !ok {
+				l.setupListItem(obj, row, focused == obj)
 			}
 		}
 	} else {
--- a/widget/list.go
+++ b/widget/list.go
@@ -9,7 +9,6 @@ import (
 	"fyne.io/fyne/v2/canvas"
 	"fyne.io/fyne/v2/data/binding"
 	"fyne.io/fyne/v2/driver/desktop"
-	"fyne.io/fyne/v2/internal/cache"
 	"fyne.io/fyne/v2/internal/widget"
 	"fyne.io/fyne/v2/theme"
 )
@@ -96,14 +95,7 @@ func (l *List) MinSize() fyne.Size {
 
 func (l *List) Refresh() {
 	if l.scroller != nil {
-		impl := l.super()
-		if impl == nil {
-			return
-		}
-
-		render := cache.Renderer(impl)
-		render.Refresh()
-
+		l.BaseWidget.Refresh()
 		lo := l.scroller.Content.(*fyne.Container).Layout.(*listLayout)
 		visible := lo.visible
 		canvas := fyne.CurrentApp().Driver().CanvasForObject(lo.list)
--- a/widget/list.go
+++ b/widget/list.go
@@ -93,19 +93,23 @@ func (l *List) MinSize() fyne.Size {
 	return l.BaseWidget.MinSize()
 }
 
+// RefreshItem refreshes a single item, specified by the item ID passed in.
+//
+// Since 2.4
 func (l *List) RefreshItem(id ListItemID) {
-	if l.scroller != nil {
-		l.BaseWidget.Refresh()
-		lo := l.scroller.Content.(*fyne.Container).Layout.(*listLayout)
-		visible := lo.visible
-		canvas := fyne.CurrentApp().Driver().CanvasForObject(lo.list)
-		var focused fyne.Focusable
-		if canvas != nil {
-			focused = canvas.Focused()
-		}
-		if item, ok := visible[id]; ok {
-			lo.setupListItem(item, id, focused == item)
-		}
+	if l.scroller == nil {
+		return
+	}
+	l.BaseWidget.Refresh()
+	lo := l.scroller.Content.(*fyne.Container).Layout.(*listLayout)
+	visible := lo.visible
+	canvas := fyne.CurrentApp().Driver().CanvasForObject(lo.list)
+	var focused fyne.Focusable
+	if canvas != nil {
+		focused = canvas.Focused()
+	}
+	if item, ok := visible[id]; ok {
+		lo.setupListItem(item, id, focused == item)
 	}
 }
 
@@ -671,9 +675,9 @@ func (l *listLayout) updateList(newOnly bool) {
 	l.renderLock.Unlock() // user code should not be locked
 
 	if newOnly {
-		for nId, nItem := range visible {
-			if _, ok := wasVisible[nId]; !ok {
-				l.setupListItem(nItem, nId, focused == nItem)
+		for row, obj := range visible {
+			if _, ok := wasVisible[row]; !ok {
+				l.setupListItem(obj, row, focused == obj)
 			}
 		}
 	} else {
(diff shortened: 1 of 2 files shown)


andydotxyz:
Thanks for looking at this. However every item must also update if Refresh is call. I think that was missed or not tested in the unit test you added.

SMQuazi:
Changes made and tests passed (locally atleast. waiting on PR checks). It's not where I want it to be performance wise. Still getting less calls (description update) but would ideally like it only to call updateItem 1x when it becomes visible.

SMQuazi:
Looks like it's working as desired and all green.

widget/list.go
```diff
@@ -93,6 +93,36 @@ func (l *List) MinSize() fyne.Size {
...
+
+func (l *List) Refresh() {
+	lo, visible, focused := l.refreshBase()
+	for id, item := range visible {
+		item.updated = false
+		lo.setupListItem(item, id, focused == item)
+	}
+}
+
+func (l *List) RefreshItem(id ListItemID) {
```
SMQuazi, line 119: I'm not sure what access this refresh should have.

Jacalz: There is an issue (https://github.com/fyne-io/fyne/issues/3826k) open to track adding support for refreshing individual items. If you want to do this for List, then I'm fine with that

SMQuazi: Nice! I'll leave this as is then.

widget/list.go
```diff
@@ -93,6 +93,36 @@ func (l *List) MinSize() fyne.Size {
...
+		var focused fyne.Focusable
+		if canvas != nil {
+			focused = canvas.Focused()
+		}
+		return lo, visible, focused
+	}
+	return nil, nil, nil
+}
+
+func (l *List) Refresh() {
```
andydotxyz, line 111: By adding this function you are cutting out the existing Refresh work in the renderer (having overridden the BaseWidget.Refresh()).
Just looking at it this may lead to duplications, or could be redundant code in renderer now.

SMQuazi: I'm not sure I follow. The BaseWidget refresh is being called in refreshBase. I haven't looked much into it yet, but it looks like it handles the renderer refresh.

SMQuazi:
Reworked to adhere to the current flow with minimal changes.

widget/list.go
```diff
@@ -93,6 +93,22 @@ func (l *List) MinSize() fyne.Size {
...
+		visible := lo.visible
+		canvas := fyne.CurrentApp().Driver().CanvasForObject(lo.list)
+		var focused fyne.Focusable
+		if canvas != nil {
+			focused = canvas.Focused()
+		}
+		if item, ok := visible[id]; ok {
+			lo.setupListItem(item, id, focused == item)
+		}
+	}
```
Jacalz, line 109: We prefer to keep the main code path not indented as it is a bit easier to read, i.e. you just put the if-statement on top:
```go
if l.scroller == nil {
    return
}
```

widget/list.go
```diff
@@ -93,6 +93,22 @@ func (l *List) MinSize() fyne.Size {
 	return l.BaseWidget.MinSize()
 }
 
+func (l *List) RefreshItem(id ListItemID) {
```
Jacalz, line 96: Please add a comment for new plunking APIs. Look at the function below this one to see how it is supposed to be done (substituting the version with 2.4 in this case).

widget/list.go
```diff
@@ -654,8 +670,16 @@ func (l *listLayout) updateList(refresh bool) {
 	l.list.scroller.Content.(*fyne.Container).Objects = objects
 	l.renderLock.Unlock() // user code should not be locked
 
-	for row, obj := range visible {
-		l.setupListItem(obj, row, focused == obj)
+	if newOnly {
+		for nId, nItem := range visible {
```
Jacalz, line 674: Can you please keep the variables consistent with the loop below? I think I personally prefer row and obj for naming.

Jacalz (changes requested):
Thanks. That does look better. I didn't have time to test the code locally but I added some comments inline.

widget/list.go
```diff
@@ -93,6 +93,26 @@ func (l *List) MinSize() fyne.Size {
 	return l.BaseWidget.MinSize()
 }
 
+// RefreshItem refreshes a single item, specified by the item ID passed in.
+//
+// Since 2.4
```
Jacalz, line 98: The colon needs to be included for it to be parsed correctly on our developer website:
```suggestion
// Since: 2.4
```

Jacalz (changes requested):

andydotxyz (approved):
Looks like a great improvement, thanks for helping with this.

I guess we need to match `RefreshItem` with table and tree equivalents too, but that could be a follow up PR.

SMQuazi:
> Looks like a great improvement, thanks for helping with this.
> 
> I guess we need to match `RefreshItem` with table and tree equivalents too, but that could be a follow up PR.

Thanks for guiding me through my first PR here. Hopefully more to come.

Jacalz (approved):
Thanks for the PR. This looks good to me as well. Congratulations on landing your first contribution to Fyne. We appreciate your work. Don't hesitate to reach out  if you have any questions or want to discuss anything. A lot of discussions around Fyne take place  in the #fyne and #fyne-contributors channels on Gophers Slack if you wish to join :)

PR #4013, opened 2023-07-02
Change StringWithFallback return to value

nh3000-org:
### Description:
StringWithFallback was returning the string validator rather than the string causing a type mismatch.

### DIFF
--- a/internal/preferences.go
+++ b/internal/preferences.go
@@ -1,6 +1,7 @@
 package internal
 
 import (
+	"fmt"
 	"sync"
 
 	"fyne.io/fyne/v2"
@@ -183,7 +184,7 @@ func (p *InMemoryPreferences) StringWithFallback(key, fallback string) string {
 		return fallback
 	}
 
-	return value.(string)
+	return fmt.Sprint(value)
 }
 
 // SetString saves a string value for the given key


internal/preferences.go
```diff
@@ -183,7 +184,7 @@ func (p *InMemoryPreferences) StringWithFallback(key, fallback string) string {
 		return fallback
 	}
 
-	return value.(string)
+	return fmt.Sprint(value)
```
Jacalz, line 187: I don't quite think this is the right approach. It definitely solves the issue but it is a bit brute-force as it will be quite slow and basically cause a copy if there is a string that is passed to it (which we probably expect to be there most of the time). I would rather prefer if we check that the value actually is a string and log an error, or something like that, if it isn't.

Jacalz (changes requested):
Thanks for the PR. I appreciate the contribution :)

I do have a few notes about this change:

- All pull requests should target the `develop` branch and not `master`.  All of the development happens there and we then move the changes into master once it is time for a new release. This means that master always tracks the latest stable. It isn't what people are used to on GitHub but it, in my opinion, a much better system.

- Please don't remove the check boxes in they PR templates. They are meant as more or less mandatory tasks for what we want a PR to have completed before being opened.
  - In short, I think it would be good to add tests for this. 

- If the binding code you were using (I assume) incorrectly set the string validator instead of the string, then that issue should be fixed there. Would you mind sharing the code that led to the issue?

I understand that there are a lot of things to think about here as we have quite an elaborate set of rules and guidelines, but don't worry. As a first-time contributor I also understand that it isn't easy, so feel free to reach out if you have any questions or need any help. We'll get you through this :)

andydotxyz:
There is no "string validator" in this code - the type returned is the Go builtin `string`. By printing the content to a string you are using more memory and hiding type conversion errors.

internal/preferences.go
nh3000-org: Hi,
Problem is value.(string) returns a boolean and not the string.  I do not think this is the expected behavior.  Also since it is an interface, there may not be much else that can be done.  I must admit I am new to go, if there is a preferred way to handle this, I will gladly comply.  As far as error handling, that is a good idea.  Ill add tests once we are in agreement about the patch.

andydotxyz: Erm I think there is a misunderstanding here.
`mything.(string)` will return a string (or panic if mything was not a string or castable). It certainly does not return a bool.

andydotxyz: P.s. `string` is not an interface, it is the builtin type representing a sequence of `rune`s.

nh3000-org:
Android phone was not properly removing preferences on un-install.  Manually cleared the data in app settings fixed the problem.  Thanks to ALL for the help.

PR #4017, opened 2023-07-03
Minor cleanup and performance improvements to tabs

Jacalz:
### Description:

This PR does the following:
- Removes a case of double refresh when showing Apptabs or DocTabs.
- Only check for mismatched tab items when hints are actually enabled.
- Caches some object sizes and uses `fyne.NewSquareSize()` for improved performance.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Neat :)

PR #4027, opened 2023-07-04
Improvements to application startup time and memory usage

Jacalz:
### Description:

This PR cleans up some of the app startup code to avoid an issue where preferences were loaded multiple times on startup. There are some other cleanups included as well for more performance and memory use improvements.

#### Multiple loads

We were storing a known struct as an interface and then directly doing an interface check to verify that it contained the method `load()`. However, we don't need to do an interface when we already know the type so we can remove it and know that we will have loaded the preferences. Now that we know the preferences are loaded already, we can remove the exact same interface check that was done before setting up document storage.

This improves the startup speed by only loading and parsing the preferences file once instead of twice during startup. The lack of interface casting improves things a bit further.

#### Test-only field

The app struct contained a field that was basically only used in order to intercept the command execution. This was only ever used for a single test and that test also wasn't very good due to not running in ci and so on. Removing the field uses slightly less memory but it also made the test not work. Another test was added to keep the percentage stable while the old, and bad test, was removed.  

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -14,8 +14,6 @@ import (
 	"fyne.io/fyne/v2/internal/app"
 	intRepo "fyne.io/fyne/v2/internal/repository"
 	"fyne.io/fyne/v2/storage/repository"
-
-	"golang.org/x/sys/execabs"
 )
 
 // Declare conformity with App interface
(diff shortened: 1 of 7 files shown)


andydotxyz (approved):
Nice thanks

PR #4039, opened 2023-07-08
Don't panic when loading preferences with wrong type

nh3000-org:
### Description:
Fix panics in case of type change or file corruption

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/preferences.go
+++ b/internal/preferences.go
@@ -41,11 +41,11 @@ func (p *InMemoryPreferences) BoolListWithFallback(key string, fallback []bool) 
 	if !ok {
 		return fallback
 	}
-	_, ok1 := value.([]bool)
-	if !ok1 {
+	valb, ok := value.([]bool)
+	if !ok {
 		return fallback
 	}
-	return value.([]bool)
+	return valb
 }
 
 // BoolWithFallback looks up a boolean value and returns the given fallback if not found
@@ -54,11 +54,11 @@ func (p *InMemoryPreferences) BoolWithFallback(key string, fallback bool) bool {
 	if !ok {
 		return fallback
 	}
-	_, ok1 := value.(bool)
-	if !ok1 {
+	valb, ok := value.(bool)
+	if !ok {
 		return fallback
 	}
-	return value.(bool)
+	return valb
 }
 
 // ChangeListeners returns the list of listeners registered for this set of preferences.
@@ -80,14 +80,14 @@ func (p *InMemoryPreferences) FloatListWithFallback(key string, fallback []float
 	if !ok {
 		return fallback
 	}
-	_, ok1 := value.([]float64)
-	if ok1 {
-		return value.([]float64)
+	valf, ok := value.([]float64)
+	if ok {
+		return valf
 	}
-	_, ok2 := value.([]int)
-	if ok2 {
-		flts := make([]float64, len(value.([]int)))
-		for i, f := range value.([]int) {
+	vali, ok := value.([]int)
+	if ok {
+		flts := make([]float64, len(vali))
+		for i, f := range vali {
 			flts[i] = float64(f)
 		}
 		return flts
@@ -98,16 +98,15 @@ func (p *InMemoryPreferences) FloatListWithFallback(key string, fallback []float
 // FloatWithFallback looks up a float64 value and returns the given fallback if not found
 func (p *InMemoryPreferences) FloatWithFallback(key string, fallback float64) float64 {
 	value, ok := p.get(key)
-
 	if !ok {
 		return fallback
 	}
-	_, ok1 := value.(float64)
-	if ok1 {
-		return value.(float64)
+	valf, ok := value.(float64)
+	if ok {
+		return valf
 	}
-	vali, ok2 := value.(int)
-	if ok2 {
+	vali, ok := value.(int)
+	if ok {
 		return float64(vali)
 	}
 	return fallback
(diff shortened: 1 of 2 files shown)


internal/preferences.go
```diff
@@ -41,7 +41,10 @@ func (p *InMemoryPreferences) BoolListWithFallback(key string, fallback []bool)
 	if !ok {
 		return fallback
 	}
-
+	_, ok1 := value.([]bool)
```
Jacalz, line 44: There is no need to call this `ok1`. Just change it like this:
```go
	_, ok = value.([]bool)
```

internal/preferences.go
```diff
@@ -41,7 +41,10 @@ func (p *InMemoryPreferences) BoolListWithFallback(key string, fallback []bool)
...
+	if !ok1 {
+		return fallback
+	}
 	return value.([]bool)
```
Jacalz, line 48: You are doing a second cast here which is unnecessary work. Just save the first parameter of the `_, ok` call above (replacing `_` with `value`).

internal/preferences.go
```diff
@@ -78,22 +80,37 @@ func (p *InMemoryPreferences) FloatListWithFallback(key string, fallback []float
...
+		}
+		return flts
+	}
+	return fallback
 }
 
 // FloatWithFallback looks up a float64 value and returns the given fallback if not found
 func (p *InMemoryPreferences) FloatWithFallback(key string, fallback float64) float64 {
 	value, ok := p.get(key)
+
```
Jacalz, line 101: Please remove this empty line.

internal/preferences.go
```diff
@@ -129,16 +148,16 @@ func (p *InMemoryPreferences) IntWithFallback(key string, fallback int) int {
 	if !ok {
 		return fallback
 	}
-
-	// integers can be de-serialised as floats, so support both
```
Jacalz, line 133: Please keep this comment. It was providing useful information.

Jacalz (changes requested):
Thanks for the PR. I left a few comments on improving the variable naming. Most of the comments do apply in more places than where I added them (a lot of similar code in various functions).

One question for the other reviewers: Should we really keep the casting between float and int? At least for slices, this does seem like a very slow operation.

Jacalz:
I got got to add, would you mind giving the PR a more description name?

andydotxyz:
> Thanks for the PR. I left a few comments on improving the variable naming. Most of the comments do apply in more places than where I added them (a lot of similar code in various functions).
> 
> One question for the other reviewers: Should we really keep the casting between float and int? At least for slices, this does seem like a very slow operation.

The float/int casting is required because Go loads JSON ints as floats...

Jacalz:
> The float/int casting is required because Go loads JSON ints as floats...

Sure, but if all numbers are stored are slices of floats then surely all of our `IntList()` methods that we have would not have worked before this PR?

andydotxyz:
Ah yes maybe some of it could be avoided. As long as it doesn't introduce type issues between int/float in load/save happy to remove what we can.

internal/preferences.go
nh3000-org: done

andydotxyz (commented):
This looks good to me, I think we need to resolve the casting discussion with @Jacalz

andydotxyz:
> Sure, but if all number lists are stored as slices of floats then surely all of our `IntList()` methods that we have would not have worked before this PR?

It is possible we missed a test case in the new number list code?...

Jacalz:
I did a quick test on develop and int lists do work fine on `develop` without any casts from float in the slices we return. Do we really want to do the costly translation then? I'd be with it just returning the fallback if I try to load a FloatList() when I have set an IntList() on that key. It seems better to just make the developer understand that they used the wrong ket rather than allocating a new slice and copying over all the values.

FYI: The code we have to casting over slices if incorrect for float in this PR. If we decide on keeping the casting, we never want to cast float to int as any possible decimal precision is rounded down.

nh3000-org:
> FYI: The code we have to casting over slices if incorrect for float in this PR. If we decide on keeping the casting, we never want to cast float to int as any possible decimal precision is rounded down.

Agree with rounding, but is fallback better than return 1 rather than 1.0

I can change it. Let me know....

andydotxyz:
I think the rounding is a red-herring. Ints can be deserialised as floats, but they will have ".0" after the decimal place because they were ints in the first place... so if int cast is here to cope with JSON storage then there is no rounding to worry about?

nh3000-org:
I have tested on linux, android and tests run as expected.

internal/preferences_test.go
```diff
@@ -198,6 +229,7 @@ func TestPrefs_SetSameSliceValue(t *testing.T) {
 
 	// We should not fire change when it hasn't changed.
 	for i := 0; i < 2; i++ {
+
```
Jacalz, line 232: Let's not add new lines outside of the code related to the PR. It is more consistent with our other code to not have a new line here.

internal/preferences.go
```diff
@@ -129,16 +148,17 @@ func (p *InMemoryPreferences) IntWithFallback(key string, fallback int) int {
...
-	val, ok := value.(float64)
+	// integers can be de-serialised as floats, so support both
+	valf, ok := value.(float64)
 	if !ok {
-		return 0
+		return fallback
 	}
-	return int(val)
+	return int(valf)
+
```
Jacalz, line 161: Can you please remove the newline?

Jacalz (changes requested):
Thanks. This is looking good. Sorry for the confusion with the casting. What you have now makes sense.

I've left two comments about removing some newlines that I think you might have accidentally added in. Apart from that, I think we need a test for the []int to []float slice cast because that wasn't there before. This should be good to go after that :)

internal/preferences.go
nh3000-org: Done....

internal/preferences_test.go
nh3000-org: Done..

nh3000-org (commented):
Tested on linux and android.

andydotxyz (approved):
Excellent to have more safety thanks :)

Jacalz (approved):
Thanks

PR #4043, opened 2023-07-09
Force writing preferences to file when application stops (fixes #3170)

### LINKED ISSUE #3170: Preferences don't all save when written in `CloseIntercept`
### Checklist

### Describe the bug

When using `fyne.App.Run()` to start an application, only first written preference is saved in Preferences store.

### How to reproduce

1. Launch example code.
2. Take a look at `${XDG_CONFIG_HOME}/fyne/some.test/preferences.json`
3. Be surprised that only width is saved.

### Screenshots

_No response_

### Example code

```go
package main

import (
        "fyne.io/fyne/v2"
        "fyne.io/fyne/v2/app"
)

var (
        a fyne.App
        wnd fyne.Window
)

func main() {
        a = app.NewWithID("some.test")
        wnd = a.NewWindow("Test")

        wnd.Resize(fyne.Size{
                Width: float32(a.Preferences().IntWithFallback("mainwindow/width", 800)),
                Height: float32(a.Preferences().IntWithFallback("mainwindow/height", 600)),
        })

        wnd.SetCloseIntercept(closeIntercept)

        wnd.Show()

        a.Run()
}

func closeIntercept() {
        windowSizes := wnd.Canvas().Size()

        a.Preferences().SetInt("mainwindow/width", int(windowSizes.Width))
        a.Preferences().SetInt("mainwindow/height", int(windowSizes.Height))

        wnd.Close()
}
```

### Fyne version

2.2.3 and today's develop at https://github.com/fyne-io/fyne/commit/b71c0f2e5ab70eefa34e6a0b9a1643ffbaa68915

### Go compiler version

1.18.4

### Operating system

macOS

### Operating system version

macOS 12.5

### Additional Information

This problem does not appear when using `fyne.Window.ShowAndRun`.

nullst:
### Description:

Fyne's preferences mechanism implements a debouncing strategy for modifications, it saves the preferences changes to the filesystem at most every 100ms. As explained in issue #3170, the implementation has a bug: if an application modifies at least two preferences (say "A" and "B") during the `OnStopped` lifecycle trigger, we get this:

1. Preference "A" modified, the preferences change listener saves the result to file.
2. Preference "B" modified, the preferences change listener **does not write** the result to file, but instead schedules the save 100ms later.
3. Application exits.
4. After about 100ms, the updated value of the preference "B" would be written to file, but the application is not running anymore. Thus only the value of the preference "A" is updated.

This pull request forces the application to write preferences to file on exit. It does this by adding a new internal lifecycle event, `AfterStoppedTrigger`, which is evaluated after the user-provided `Stopped` lifecycle callback. This modification does not change any public-facing interfaces, only internal implementations.

Fixes #3170.

### Example code

For completeness, here's an example code that demonstrates that the issue #3170 applies to the lifecycle triggers as well. It starts working as intended after this patch.

```go
package main

import (
	"fmt"
	"math/rand"
	"time"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/layout"
	"fyne.io/fyne/v2/widget"
)

func main() {
	application := app.NewWithID("preferences-bug")
	window := application.NewWindow("Preferences bug")

	randomSource := rand.NewSource(time.Now().UnixNano())
	rng := rand.New(randomSource)
	value := rng.Intn(1000)

	label1 := widget.NewLabel(fmt.Sprintf("1st value: %d", application.Preferences().IntWithFallback("one", 0)))
	label2 := widget.NewLabel(fmt.Sprintf("2nd value: %d", application.Preferences().IntWithFallback("two", 0)))
	button := widget.NewButton(fmt.Sprintf("Quit and write %d to both preferences", value), func() {
		application.Quit()
	})
	box := container.New(layout.NewVBoxLayout(), label1, label2, button)
	window.Resize(fyne.NewSize(float32(200), float32(100)))
	window.SetContent(box)

	application.Lifecycle().SetOnStopped(func() {
		application.Preferences().SetInt("one", value)
		application.Preferences().SetInt("two", value)
	})

	window.ShowAndRun()
}
```

### Checklist:

- [ ] Tests included. <- I don't know how to test lifecycle triggers; besides, fyne's tests only set up in-memory preferences, so I don't know how to test that the file is saved correctly.
- [x] Tests all pass.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -148,6 +148,11 @@ func newAppWithDriver(d fyne.Driver, id string) fyne.App {
 	fyne.SetCurrentApp(newApp)
 
 	newApp.prefs = newApp.newDefaultPreferences()
+	newApp.lifecycle.(*app.Lifecycle).SetAfterStopped(func() {
+		if prefs, ok := newApp.prefs.(*preferences); ok {
+			prefs.forceImmediateSave()
+		}
+	})
 	newApp.settings = loadSettings()
 	store := &store{a: newApp}
 	store.Docs = makeStoreDocs(id, newApp.prefs, store)
--- a/app/app.go
+++ b/app/app.go
@@ -148,7 +148,7 @@ func newAppWithDriver(d fyne.Driver, id string) fyne.App {
 	fyne.SetCurrentApp(newApp)
 
 	newApp.prefs = newApp.newDefaultPreferences()
-	newApp.lifecycle.(*app.Lifecycle).SetAfterStopped(func() {
+	newApp.lifecycle.(*app.Lifecycle).SetOnStoppedHookExecuted(func() {
 		if prefs, ok := newApp.prefs.(*preferences); ok {
 			prefs.forceImmediateSave()
 		}
--- a/app/preferences.go
+++ b/app/preferences.go
@@ -62,7 +62,6 @@ func (p *preferences) saveToFile(path string) error {
 	p.savedRecently = true
 	p.prefLock.Unlock()
 	defer p.resetSavedRecently()
-
 	err := os.MkdirAll(filepath.Dir(path), 0700)
 	if err != nil { // this is not an exists error according to docs
 		return err
--- a/app/preferences.go
+++ b/app/preferences.go
@@ -19,11 +19,24 @@ type preferences struct {
 	savedRecently       bool
 	changedDuringSaving bool
 
-	app *fyneApp
+	app                 *fyneApp
+	needsSaveBeforeExit bool
 }
 
 // Declare conformity with Preferences interface
 var _ fyne.Preferences = (*preferences)(nil)
+
+// forceImmediateSave writes preferences to file immediately, ignoring the debouncing
+// logic in the change listener. Does nothing if preferences are not backed with a file.
+func (p *preferences) forceImmediateSave() {
+	if !p.needsSaveBeforeExit {
+		return
+	}
+	err := p.save()
+	if err != nil {
+		fyne.LogError("Failed on force saving preferences", err)
+	}
+}
 
 func (p *preferences) resetSavedRecently() {
 	go func() {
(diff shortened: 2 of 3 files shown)


andydotxyz (changes requested):
Thanks for looking at this. Unfortunately using the public lifecycle API isn't a workable solution because somebody's app code would override it and break your fix.

nullst:
>  Unfortunately using the public lifecycle API isn't a workable solution because somebody's app code would override it and break your fix.

Yep, that's why this patch does not use or modify public lifecycle API.

The public interface Lifecycle in `fyne/lifecycle.go` is not changed, I only added an exported function to the (private) structure `Lifecycle` in `fyne/internal/app/lifecycle.go`.

andydotxyz:
Ah clever, I missed that sorry.
I wonder if a single callback in that instance is sufficient? Because normally the usage of it is separated from the other lifecycle usage (when a developer uses this api it is all in main.go, so easy to read).

So if we do the same thing for another item of code we will break this one. Would a slice of them be suitable? (Also can it use naming that matches, such as `onDestroy` or something... we use "on" prefix to indicate it is a callback)

nullst:
> So if we do the same thing for another item of code we will break this one.

True, I just followed the approach taken for other lifecycle triggers. Since this is not a public API, it wouldn't be hard to switch to a slice of functions if at some point that would be necessary. I think for most internal purposes there are more direct solutions for executing code at the application exit, here (as discussed in the comments of #3170) the point of the lifecycle approach is to avoid introducing a public `Sync()` method for preferences.

> Also can it use naming that matches, such as onDestroy or something... we use "on" prefix to indicate it is a callback

How about `onStoppedHookExecuted`? More descriptive, no need to explain the priorities between `onStopped` and `onDestroy`, sounds like an internal API name.

andydotxyz (commented):
This looks great.
I think that it could avoid the `atomic.Value` as the setter is always called before quit and it doesn't change through the lifecycle.

nullst:
>  I think that it could avoid the atomic.Value as the setter is always called before quit and it doesn't change through the lifecycle.

It could, but I used the atomic approach to keep the lifecycle code consistent, thus easier to understand. I can switch to just storing `func ()` if you prefer, though.

nullst:
I've switched from the atomic.Value approach to storing just a func().

andydotxyz (approved):
Thanks for this

PR #4044, opened 2023-07-09
added: SetText method on Check widget

mbaklor:
### Description:

Added the option to SetText on check widget.
Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


widget/check.go
```diff
@@ -284,6 +284,12 @@ func (c *Check) TypedRune(r rune) {
 // TypedKey receives key input events when the Check is focused.
 func (c *Check) TypedKey(key *fyne.KeyEvent) {}
 
+// SetText sets the text of the Check
```
andydotxyz, line 287: Thanks for this, you would add since in the comment before the function definition, using:
```go
//
// Since: v2.4
```

widget/check_test.go
```diff
@@ -96,3 +96,12 @@ func TestNewCheckWithData(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, false, v)
 }
+
+func TestCheck_SetText(t *testing.T) {
+	check := &widget.Check{Text: "Test"}
+	check.SetText("Other Text")
```
andydotxyz, line 102: I'm not sure what this line or the Refresh below do?

andydotxyz (changes requested):
Thanks so much for this, just a couple of comments inline.

widget/check_test.go
mbaklor: I honestly don't know, this specific test I copied from the Label widget, I think there it has to do with render tests but I'm not sure. Deleting and committing again

mbaklor:
changes were implemented as requested

PR #4047, opened 2023-07-10
Added Importance field on Label to color the text

blaize:
### Description:

I'm making a cross platform GUI app that manage acounting data and I wanted to color texts for negative values (red) and positive values (green). Also some warning messages if the client havn't paid should be printed in orange.
I can use canvas.Text but then I lost the TextStyle{} and Wrapping feature.
Also, using label and text in the same row of a List are not exactly aligned.

So I made this little patch based on the logic of ButtonImportance to add an Importance to a label. The Importance color the message on the label.

I'm not sure about names, and if it need to add an importance that print in Primary color and Secondary color.
But I think a lot of use case needs success, warning and error messages in the app. 
Coloring text make the message more clear and the interface more intuitive.

Fixes #(issue)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
I hadn't thought about this, could be cool.
I wonder if instead of adding text importance as a new API maybe we should either re-use the button ones or move to a less widget-specific naming that can be used elsewhere?

Or just export the ColorName directly on Label - not sure about that either...
Maybe others have thoughts on the naming - always good to get a few thoughts.

blaize:
> I wonder if instead of adding text importance as a new API maybe we should either re-use the button ones or move to a less widget-specific naming that can be used elsewhere?

That what I was thinking at first, but I didn't want to use ButtonImportance for something called Label, and I also didn't want to move too much code for my first attempt so I made this. But yes, ideally an "importance" type that work for multiple widget would be the best.

Jacalz:
I think we should deprecate `widget.ButtonImportance` and create a `widget.Importance` (or similar) and use that instead. We might even make the types entirely compatible by doing something like `type Importance = ButtonImportance`.

blaize:
I commit a patch to use widget.Importance instead of widget.ButtonImportance.
Also updated label and other widget that used ButtonImportance.

widget/button.go
```diff
@@ -21,7 +21,8 @@ type ButtonIconPlacement int
 // ButtonImportance represents how prominent the button should appear
 //
 // Since: 1.4
-type ButtonImportance int
+// Deprecated: Use widget.Importance instead
```
Jacalz, line 24: I believe that you need to have the deprecation as a new paragraph.
```suggestion
//
// Deprecated: Use widget.Importance instead
```

widget/button.go
```diff
@@ -21,7 +21,8 @@ type ButtonIconPlacement int
...
+type ButtonImportance Importance
```
Jacalz, line 25: Your change is currently backwards-incompatible as any code looking like this won't compile:
```go
var importance widget.ButtonImportance = widget.MediumImportance
```
You can fix it by using an equal sign like mentioned in https://github.com/fyne-io/fyne/pull/4047#issuecomment-1631252683:
```suggestion
type ButtonImportance = Importance
```

widget/importance.go
```diff
@@ -0,0 +1,24 @@
+package widget
+
+type Importance int
```
Jacalz, line 3: Please add a comment and a `Since: 2.4` line to this type.

widget/importance.go
```diff
@@ -0,0 +1,24 @@
...
+	//
+	// Since 2.3
+	DangerImportance
+	// WarningImportance applies a warning theme to the widget.
+	//
+	// Since 2.3
+	WarningImportance
+
+	// SuccessImportance applies a success theme to the widget.
+	SuccessImportance
```
Jacalz, line 23: Hmm. I'm not sure about this as an importance. Success does not seem like an importance level to me. Can we leave that out for now and focus on adding importance support to label?

Jacalz (changes requested):
Thanks. I added a few notes inline. Also, would you mind adding some tests? You can probably have a look at the existing importance tests for Button.

widget/importance.go
blaize: I will need the possibility to print green (success theme) message. Don't know how it can be called, but it's usefull to print text like "successfully done something" or "up-to-date something". In my case, it will be to print positive value like +200. Negative are printed in red (DangerImportance)

blaize: Maybe HighImportance on label print in Success color instead of Primary color ?

blaize:
Oh ok sorry I didn't know. I will not touch next time ;)

Thanks.

blaize:
I added initialization on Medium importance on the label. It's not necessary because Medium = 0 but I think it's more readable.
Maybe I can do the same for button if you want, or you can discard the last commit.

Also, I copied the test and I notified this strange call in many test :
```
test.NewApp()
defer test.NewApp()
```

I don't understand why this is called second time at the end of the test. Can you explain me ?

widget/button_test.go
```diff
@@ -312,3 +312,21 @@ func TestButton_ChangeTheme(t *testing.T) {
...
+	var imp widget.ButtonImportance = widget.HighImportance
+
+	btn := widget.NewButton("test", func() {})
+	btn.Importance = imp
+	btn.Refresh()
+
+	w := test.NewWindow(btn)
+	defer w.Close()
+
+	test.AssertImageMatches(t, "button/compat_importance.png", w.Canvas().Capture())
```
Jacalz, line 331: I don't think you need any image for this if you just want to make sure that it compiles (please also start comments with a space and then a capital letter):
```suggestion
	// Test backward compatibility of widget.Importance
	var imp widget.ButtonImportance = widget.HighImportance

	btn := widget.NewButton("test", func() {})
	btn.Importance = imp
```
Remember to delete the corresponding image as well.

widget/label.go
```diff
@@ -36,9 +38,10 @@ func NewLabelWithData(data binding.String) *Label {
 // NewLabelWithStyle creates a new label widget with the set text content
 func NewLabelWithStyle(text string, alignment fyne.TextAlign, style fyne.TextStyle) *Label {
 	l := &Label{
-		Text:      text,
-		Alignment: alignment,
-		TextStyle: style,
+		Text:       text,
+		Alignment:  alignment,
+		TextStyle:  style,
+		Importance: MediumImportance,
```
Jacalz, line 44: This should not be needed. `MediumImportance` is equal to zero and thus is the default.

Jacalz:
FYI: Please fill out all of the checkboxes in the PR description.

Jacalz:
The code looks good to me. I'll do a proper review and test locally when I have some time over.

widget/importance.go
andydotxyz: We should not be inconsistent - high importance does not mean success.

I think that adding success here makes sense - the naming may not be optimal, but we have been using it on button since last release so why not go with it?

Jacalz: Have we been using it on Button since last release? We didn't have it as an importance before as far as I can see

Jacalz: I agree that success in itself is good to add but yeah, not sure about naming there. I can let it pass if you're happy with it

Jacalz: Success probably is the better one in that case. It is what we call the colour

andydotxyz: > Have we been using it on Button since last release? We didn't have it as an importance before as far as I can see

My apologies there was not a success button before. I kindof assumed there was!
I'm not too bothered one way or another but if the purpose is to expose the meaning here in general I don't think there's a reason to leave off success.

Jacalz: Agreed. Let's go for that then. Sorry for the back and forth @blaize

blaize: Ok no problem. So I will add SuccessImportance. Should I have to also patch the Button widget to have it react to SuccessImportance ?

Jacalz: Sorry for the late reply. I see you added it. Thanks :)

Jacalz (approved):
This looks good to me. Thanks. Well done

andydotxyz (approved):
Great thanks for adding this

PR #4052, opened 2023-07-11
Fix postion of line during deletion

andydotxyz:
### Description:

Fixes #(issue)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container/doctabs.go
+++ b/container/doctabs.go
@@ -244,7 +244,12 @@ func (r *docTabsRenderer) Layout(size fyne.Size) {
 	r.updateCreateTab()
 	r.updateTabs()
 	r.layout(r.docTabs, size)
+
+	// lay out buttons before updating indicator, which is relative to their position
+	buttons := r.scroller.Content.(*fyne.Container)
+	buttons.Layout.Layout(buttons.Objects, buttons.Size())
 	r.updateIndicator(r.docTabs.transitioning())
+
 	if r.docTabs.transitioning() {
 		r.docTabs.setTransitioning(false)
 	}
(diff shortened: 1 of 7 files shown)


container/doctabs.go
```diff
@@ -244,7 +244,12 @@ func (r *docTabsRenderer) Layout(size fyne.Size) {
 	r.updateCreateTab()
 	r.updateTabs()
 	r.layout(r.docTabs, size)
+
+	// lay out buttons before updating indicator, which is relative to their position
+	buttons := r.scroller.Content.(*fyne.Container)
+	buttons.Layout.Layout(buttons.Objects, buttons.Size())
```
Jacalz, line 250: Is it not possible to just do `r.scroller.Content.Resize(r.scroller.Content.Size())` or something along those lines?
Looking at the container code, it seems to me like it does the same thing without casting.

Jacalz: Maybe something like this would work? Untested:
```suggestion
	buttons := r.scroller.Content
	buttons.Resize(buttons.Size())
```

andydotxyz:
Apologies I opened the PR on a computer where the tests don't work and forgot to come back to it as I took a few days off. It's the M1/2 test issue that is hiding me back, need to get to the bottom of it!

container/doctabs.go
andydotxyz: I don't think so - when you resize something to the same size it was before the layout will not be called.

Jacalz: Alright

Jacalz (approved):

Issue #4061, opened 2023-07-14
Find a more elegant way to avoid binding data "echos"


andydotxyz:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

Input widgets (such as Entry) call binding setters when they change, and these are then echoed back to the widget with a delay. This can result in a jitter or lost input letters so a "ignore delay" is being used to avoid the feedback.

It feels like there could be a better way, such as when setting list which listener should not be called back. This may not be trivial, or may require a breakage, but should be considered for a future version.

### How to reproduce

Things like bug #3887 show the problem, though they are being fixed on a case-by-case for now.

### Example code

Use a two-way binding, like `String` and set regularly - you will see old state being restored.

### Fyne version

2.3.5

### Go compiler version

1.20.4

### Operating system and version

macOS

dan-frohlich:
This seems like a common publish / subscribe problem. That is the subscribers should be able to Id and ignore messages they produced.

If the bound data callbacks and registry had a param containing a field uniquely identifying the UI object, that bound data item could potentially not send update events to the object which sourced the change. Alternately, the id could be passed to all subscribers and each subscriber could choose how to handle calls regarding changes sourced from itself.

andydotxyz:
I feel like Entry now (on develop) has a good model for this (we know the change comes from binding so we can avoid pushing it back out).

Issue #3774, opened 2023-03-30
binding.String===>Quick refresh *b.val will appear with new data reset by a call to OnChange


MrWen:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

```go
// updateText updates the internal text to the given value
func (e *Entry) updateText(text string) {
	var callback func(string)
	e.setFieldsAndRefresh(func() {
		changed := e.Text != text
		e.Text = text
		e.syncSegments()
		e.text.updateRowBounds()

		if e.Text != "" {
			e.dirty = true
		}

		if changed {
			callback = e.OnChanged
		}
	})

	e.Validate()

	if callback != nil {
		callback(text)
	}
}

func (e *Entry) writeData(data binding.DataItem) {
	if data == nil {
		return
	}
	textTarget, ok := data.(binding.String)
	if !ok {
		return
	}
	curValue, err := textTarget.Get()
	if err == nil && curValue == e.Text {
		e.conversionError = nil
		return
	}
	e.conversionError = textTarget.Set(e.Text)
}
```

### How to reproduce

Because refreshing Entry and Set are asynchronous operations。

Quick call String.Set。 Old data overwrites new data when writeData is executed。

### Example code

```go
binding := binding.NewString()
entry := widget.NewMultiLineEntry()
entry.Bind(binding)

go func() {
var num int
 for {
str :=fmt.Sprintf("%d", num)
binding.Set(str)
}
}
```

### Fyne version

2.3.2

### Go compiler version

1.20

### Operating system and version

Window11

andydotxyz:
Fixed on develop and release/v2.3.x branch

PR #4062, opened 2023-07-14
Fix/3887

### LINKED ISSUE #3887: Entry with data binding misses keystrokes or panics on fast input
### Checklist

### Describe the bug

I'm using a Barcode-Scanner to input data into an "widget.NewEntryWithData". If the Data is long, and has about a 12ms delay between keystrokes, the input either gets cut or fyne panics.

The length of the string that is needed to hang the application seems to depend on the performance of the computer.
On slower machines I have seen this bug with string length of 10.

A normal Entry (without data binding) works as expected.

### How to reproduce

1. run example code
2. focus input
3. use Barcode scanner (or simulate it with something like xdotool)
4. Panic

```
panic: runtime error: slice bounds out of range [6:5]

goroutine 15 [running]:
fyne.io/fyne/v2/widget.(*RichText).insertAt(0xc0001942c0, 0x6, {0xc000535f20, 0x1})
        /home/alex/go/pkg/mod/fyne.io/fyne/v2@v2.3.4/widget/richtext.go:236 +0x390
fyne.io/fyne/v2/widget.(*Entry).TypedRune(0xc0001ba1c0, 0x72)
        /home/alex/go/pkg/mod/fyne.io/fyne/v2@v2.3.4/widget/entry.go:747 +0x1a5
fyne.io/fyne/v2/internal/driver/glfw.(*window).processCharInput.func1()
        /home/alex/go/pkg/mod/fyne.io/fyne/v2@v2.3.4/internal/driver/glfw/window.go:786 +0x25
fyne.io/fyne/v2/internal/driver/common.(*Window).RunEventQueue(0x0?)
        /home/alex/go/pkg/mod/fyne.io/fyne/v2@v2.3.4/internal/driver/common/window.go:35 +0x3e
created by fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).createWindow.func1
        /home/alex/go/pkg/mod/fyne.io/fyne/v2@v2.3.4/internal/driver/glfw/window.go:946 +0x136
exit status 2
```

### Screenshots

_No response_

### Example code

```go
package main

import (
        "fyne.io/fyne/v2/app"
        "fyne.io/fyne/v2/container"
        "fyne.io/fyne/v2/widget"
)

func main() {
        myApp := app.New()
        window := myApp.NewWindow("Test")
        sb := binding.NewString()
        c := container.NewVBox(
                widget.NewEntryWithData(sb),
                widget.NewLabelWithData(sb),
        )
        window.SetContent(
                c,
        )
        window.ShowAndRun()
}
```

### Fyne version

2.3.4 and 2.3.5

### Go compiler version

1.20.3 and 1.20.5

Linux (6.1), Windows 10 and Windows 11

### Additional Information

On Linux (with X11) you can simulate fast input with:
```sh
go run . & sleep 5; xdotool type "thisisareallylongstringthatmostlikelykillsyourappandorwontappearinitsentirety0123456789"
```

andydotxyz:
### Description:
Fix race conditions and the data setting echo causing panic and/or loss of input when fast-typing on bound entry.

This points to a potential bigger issue of binding set echoes, which I have entered as a new issue #4061.

Fixes #3887 and #3774

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


widget/entry.go
```diff
@@ -1178,7 +1169,7 @@ func (e *Entry) updateCursorAndSelection() {
 }
 
 func (e *Entry) updateFromData(data binding.DataItem) {
-	if data == nil {
+	if data == nil || e.lastChange.After(time.Now().Add(-bindIgnoreDelay)) {
```
Jacalz, line 1172:
```suggestion
	if data == nil || e.lastChange.After(time.Now().Sub(bindIgnoreDelay)) {
```

Jacalz (changes requested):
This is looking very good. I just have one minor comment inline

widget/entry.go
```diff
@@ -17,8 +18,9 @@ import (
 )
 
 const (
-	multiLineRows = 3
-	wordSeparator = "`~!@#$%^&*()-=+[{]}\\|;:'\",.<>/?"
+	bindIgnoreDelay = time.Millisecond * 100 // ignore incoming DataItem fire after we have called Set
```
dweymouth, line 21: Would this end up ignoring data updates from user code that happen more frequently than 100 ms?

andydotxyz: It's not ignoring changes faster than 100ms - it is ignoring external updates for a widget being edited by the user, for 10th sec.
I'm not sure if that is what you meant, but given a typical user typing speed this seemed reasonable.

It is possible that this will fail to register correctly if you modify the entry and then 50ms later a remote source changes the data, for these minute corner cases, and for code cleanliness, I opened the other ticket where we should try and find a solution that is not time-based but will be much further reaching.

Jacalz (approved):

PR #4070, opened 2023-07-19
Don't write out 'LinuxAndBSD' section on every package

andydotxyz:
Make it optional instead :)

### Checklist:

- [ ] Tests included. <- I think we have omitted the serialising of TOML from tests?...
- [x] Tests all pass.


Jacalz (approved):
Nice catch. I 100% agree with you about it not being added to every package.

PR #4075, opened 2023-07-19
Fix make user-install breaking exec param

Jacalz:
### Description:

The `sed` command in the Makefile was a bit too inexact and removed everything to the right of the `Exec=` key. This broke parameters added using `ExecParams` in `FyneApp.toml`. Make the command match the specific part to replace more exact to fix the parameters being stripped away.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Good catch

Issue #2462, opened 2021-09-12
navigating in entry quickly with ctrl key


zivoy:
### Is it possible to construct a solution with the existing API?
You can make a custom entry with key binds for ctrl + left and right arrows and move the curser using it, but this should be a base feature

### Describe the solution you'd like to see:

When in an entry, you should be able to use the ctrl + left/right arrow keys to quickly naviage by jumping between spaces (or other characters) instead of moving one character at a time.
you should also be able to hold ctrl + shift + an arrow key to select a word

andydotxyz:
Thanks @dweymouth for resolving this (on develop) for v2.4.0 :)

PR #4084, opened 2023-07-22
Support CTRL + Backspace for deleting word in Entry

Jacalz:
### Description:

This adds support for pressing `ctrl + backspace` for deleting a whole word.
Opening as draft until I have added support for `ctrl + delete` for deleting a whole word to the left as well.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -941,10 +941,27 @@ func (e *Entry) registerShortcut() {
 		e.selecting = false
 		moveWord(se)
 	}
+	removeWord := func(se fyne.Shortcut) {
+		row := e.textProvider().row(e.CursorRow)
+		start, end := getTextWhitespaceRegion(row, e.CursorColumn, true)
+		if start == -1 || end == -1 {
+			return
+		}
+
+		e.propertyLock.Lock()
+		pos := e.cursorTextPos()
+		e.textProvider().deleteFromTo(start, pos)
+		e.CursorRow, e.CursorColumn = e.rowColFromTextPos(pos - 1)
+		e.propertyLock.Unlock()
+
+		e.updateTextAndRefresh(e.textProvider().String())
+	}
+
 	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyLeft, Modifier: fyne.KeyModifierShortcutDefault}, unselectMoveWord)
 	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyLeft, Modifier: fyne.KeyModifierShortcutDefault | fyne.KeyModifierShift}, selectMoveWord)
 	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyRight, Modifier: fyne.KeyModifierShortcutDefault}, unselectMoveWord)
 	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyRight, Modifier: fyne.KeyModifierShortcutDefault | fyne.KeyModifierShift}, selectMoveWord)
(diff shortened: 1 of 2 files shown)


widget/entry.go
```diff
@@ -941,10 +941,27 @@ func (e *Entry) registerShortcut() {
...
+		e.propertyLock.Unlock()
+
+		e.updateTextAndRefresh(e.textProvider().String())
+	}
+
 	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyLeft, Modifier: fyne.KeyModifierShortcutDefault}, unselectMoveWord)
 	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyLeft, Modifier: fyne.KeyModifierShortcutDefault | fyne.KeyModifierShift}, selectMoveWord)
 	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyRight, Modifier: fyne.KeyModifierShortcutDefault}, unselectMoveWord)
 	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyRight, Modifier: fyne.KeyModifierShortcutDefault | fyne.KeyModifierShift}, selectMoveWord)
+	e.shortcut.AddShortcut(&desktop.CustomShortcut{KeyName: fyne.KeyBackspace, Modifier: fyne.KeyModifierShortcutDefault}, removeWord)
```
dweymouth, line 964: On Mac the shortcut for this is Option (alt) + backspace. I think we should try to follow this convention, either using different files with build guards to define the shortcuts, or runtime.GOOS checks

Jacalz: Hmm, interesting. I updated it to use the alt shortcut on macOS instead.

Jacalz: Interesting find: It seems like Go is smart enough to optimize away runtime.GOOS checks. It is a trivial optimization but nevertheless, I am impressed :)
https://go.godbolt.org/z/b8GYaxz6Y

andydotxyz (approved):
This is slick!
I know it's still draft but you added the other shortcut now so I wondered if that was just oversight.

Jacalz:
Thanks but it is still draft because there seems to be a few bugs that I need to sort out :/

dweymouth:
I wonder if we could address https://github.com/fyne-io/fyne/issues/2462 in this PR (or a follow up) as well :)

Jacalz:
Don't we already support that?

dweymouth:
No we don't, not yet. On Mac it's the same with Alt instead of Ctrl/Cmd being the modifier.

Jacalz:
I suppose that issue should be closed and there should be a specific one for macOS using Alt key instead in that case. The idea of the original issue seems to be fixed (ctrl is supported). It doesn't necessarily seem like something that has to be part of this PR. You're more than welcome to follow up with a PR yourself. I only have very infrequent access to my macOS computer so can't verify the change ;)

dweymouth:
> I suppose that issue should be closed and there should be a specific one for macOS using Alt key instead in that case. The idea of the original issue seems to be fixed (ctrl is supported). It doesn't necessarily seem like something that has to be part of this PR. You're more than welcome to follow up with a PR yourself. I only have very infrequent access to my macOS computer so can't verify the change ;)

Oh, my bad - the Ctrl+left/right was added on develop and I didn't realize it since it wasn't present in my app building against 2.3.x. All that would be needed to match MacOS conventions would be to use your `deleteModifier` value for all the word-navigation shortcuts instead of ShortcutDefault modifier (and rename it to `editModifier` or something more general?)

Jacalz:
What I am trying to say is that this is quite broken at the moment and it will be a while until I get it fixed up.  You might want to open a PE if you want it on v2.4:0.

Jacalz:
I haven't had time to fix up the issues here. Closing for now and opening https://github.com/fyne-io/fyne/issues/4524 in case anyone wants to pick it up.

PR #4090, opened 2023-07-23
Added AppendLine method to widget.Entry

mbaklor:
### Description:

Added an `AppendLine` method to `entry`, to make appending to multiline entry easier.
Wasn't sure what to do if the entry isn't multiline so I set it to append with a space instead of a newline. Come to think of it maybe I should make the method just `Append`?

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -474,26 +474,17 @@ func (e *Entry) SetText(text string) {
 	e.updateCursorAndSelection()
 }
 
-// AppendLine appends the text to the end of the entry adding a new line if the entry isn't empty,
-// or a space if multiline isn't enabled
+// Appends the text to the end of the entry
 //
 // Since: 2.4
-func (e *Entry) AppendLine(text string) {
-	e.propertyLock.RLock()
-	ml := e.MultiLine
-	st := e.Text
-	e.propertyLock.RUnlock()
+func (e *Entry) Append(text string) {
+	e.propertyLock.Lock()
+	provider := e.textProvider()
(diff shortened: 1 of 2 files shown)


Jacalz:
Thanks for the contribution but I’m not so sure about this change. Appending text can be done very easily by the developer and I don’t think we should hard code it as lines. If you want to add a new line, you can just add a new line character. Adding with a space in between if it isn’t multiline also seems to be kind of unexpected. If you remove both of those parts of the methods it seems like something that can be done in one line by any developer wanting to append text.

andydotxyz:
This is a tough one. I broadly agree with @Jacalz except that maybe this can be done more efficiently. The text provider has smarter ways to append text I think, so it could be called through to avoid the string manipulation route.

Regarding naming `Append` works if there are no more characters inserted (as this is not a "line based" editor). So maybe Append is one method and AppendLine a useful helper that only works on multi line?

widget/entry.go
```diff
@@ -474,6 +474,19 @@ func (e *Entry) SetText(text string) {
 	e.updateCursorAndSelection()
 }
 
+// Appends the text to the end of the entry
```
andydotxyz, line 477: Sorry, but the comment should begin "Append" for the linter to be happy - i.e. "Append will add the given text to the end of the Entry"...

andydotxyz (approved):

andydotxyz:
Thank so much for this!

PR #4091, opened 2023-07-24
cleaned up typedKey methods on entry

mbaklor:
### Description:

while skimming the code I found a bunch of methods in `widget.Entry` that take in a bool parameter of `multiLine` but don't use it, so I removed it. 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz (approved):
Thanks so much for helping keep our code tidy :)

Jacalz:
I’m surprised that our CI didn’t catch this. Thanks for the PR :)

PR #4094, opened 2023-07-25
Add tree binding types and use it in Tree widget

### LINKED ISSUE #3995: ability to create Tree widgets with data bindings to arbitrary types?
### Checklist

i wish there was the option to use data bindings to custom types with the Tree widget, similarly to the [List widget](https://developer.fyne.io/binding/list).

yes, but it's NOT pretty. essentially, what i have to do is:

1. create a helper function that takes a slice of custom structs as an argument and an ORM instance as arguments
2. loop over the slice, checking if each element has a parent and adding it to the "root" if true (via the `CreateNode` method)
3. convert each slice element's primary key to a string and append it to an array (also via the `CreateNode` method)
4. create an update function that queries a database for the item identified by the `TreeNodeID` argument to update the label accordingly
5. return the tree

the addition of a function called something like `widget.NewTreeWithData`, whose arguments are functions that take a `binding.DataItem` as opposed to a plain string/`widget.TreeNodeID`.

andydotxyz:
Fixes #3995 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/bindtrees.go
+++ b/data/binding/bindtrees.go
@@ -1,0 +1,1739 @@
+// auto-generated
+// **** THIS FILE IS AUTO-GENERATED, PLEASE DO NOT EDIT IT **** //
+
+package binding
+
+import (
+	"bytes"
+
+	"fyne.io/fyne/v2"
+)
+
+// BoolTree supports binding a tree of bool values.
+//
+// Since: 2.0
+type BoolTree interface {
+	DataTree
+
+	Append(parent, id string, value bool) error
+	Get() (map[string][]string, map[string]bool, error)
+	GetValue(id string) (bool, error)
+	Prepend(parent, id string, value bool) error
+	Set(ids map[string][]string, values map[string]bool) error
+	SetValue(id string, value bool) error
+}
+
+type ExternalBoolTree interface {
+	BoolTree
+
+	Reload() error
+}
+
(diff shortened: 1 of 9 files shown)


Jacalz (approved):
Nice. Well done. We should probably add the tree to the data binding tab in `fyne_demo` but I will not hold up this PR on that :)

andydotxyz:
Thanks. I wasn't sure about fyne_demo as the binding tab is getting too complicated. Think maybe we need to split it down more into individual widgets and show normal vs data bind for each of them instead.

Future work indeed.

williambrode:
I just noticed this code in `gen.go`:

```go
	for _, b := range binds {
		if b.Since == "" {
			b.Since = "2.0"
		}

		writeFile(listFile, list, b)
		if b.Name == "Untyped" {
			continue // interface{} is special, we have it in binding.go instead
		}

		writeFile(treeFile, tree, b)
		if b.Name == "Untyped" {
			continue // interface{} is special, we have it in binding.go instead
		}
```

Checking `b.Name` twice is at least not necessary - but was the intention to leave out the Untyped Tree, or to include it?  I was just hoping to try to use an Untyped Tree binding.  The comment about it being in binding.go doesn't seem to apply to UntypedTree.

andydotxyz:
Good spot. The UntypedTree should have been included.

PR #4100, opened 2023-07-26
Complete keyboard handling for all collection widgets

andydotxyz:
This was a 2-parter really:

1) Fix list widget so that the focus could move through all elements
2) Add the same handling for Table and Tree as well

Now all of them focus as 1 item and use arrow keys to move the focus around internally.
This is a change for anyone who had extended Lists, but mostly they had to work around this so the change is a positive one.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ More detailed release notes can be found on the [releases page](https://github.c
 
 * Go 1.17 or later is now required.
 * Icons for macOS bundles are now padded and rounded, disable with "-use-raw-icon"
+* Focus handling for List/Tree/Table are now at the parent widget not child elements
 
 
 ## 2.3.5 - 6 June 2023
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -536,7 +536,20 @@ func (w *window) processMouseClicked(button desktop.MouseButton, action action, 
 	}
 
 	if wid, ok := co.(fyne.Focusable); !ok || wid != w.canvas.Focused() {
-		w.canvas.Unfocus()
+		ignore := false
+		_, _, _ = w.findObjectAtPositionMatching(w.canvas, mousePos, func(object fyne.CanvasObject) bool {
+			switch object.(type) {
+			case fyne.Focusable:
+				ignore = true
+				return true
+			}
+
+			return false
+		})
+
+		if !ignore { // if a parent item under the mouse has focus then ignore this tap unfocus
+			w.canvas.Unfocus()
+		}
 	}
 
 	w.mouseLock.Lock()
--- a/widget/gridwrap.go
+++ b/widget/gridwrap.go
@@ -13,8 +13,9 @@ import (
 	"fyne.io/fyne/v2/theme"
 )
 
-// Declare conformity with Widget interface.
+// Declare conformity with interfaces.
 var _ fyne.Widget = (*GridWrap)(nil)
+var _ fyne.Focusable = (*GridWrap)(nil)
 
 // GridWrapItemID is the ID of an individual item in the GridWrap widget.
 //
@@ -35,6 +36,8 @@ type GridWrap struct {
 	OnSelected   func(id GridWrapItemID)                         `json:"-"`
 	OnUnselected func(id GridWrapItemID)                         `json:"-"`
 
+	currentFocus  ListItemID
+	focused       bool
 	scroller      *widget.Scroll
 	selected      []GridWrapItemID
 	itemMin       fyne.Size
@@ -85,6 +88,23 @@ func (l *GridWrap) CreateRenderer() fyne.WidgetRenderer {
 	layout.Resize(layout.MinSize())
 
 	return newGridWrapRenderer([]fyne.CanvasObject{l.scroller}, l, l.scroller, layout)
+}
+
+// FocusGained is called after this GridWrap has gained focus.
+//
+// Implements: fyne.Focusable
+func (l *GridWrap) FocusGained() {
+	l.focused = true
+	l.scrollTo(l.currentFocus)
+	l.Refresh() // TODO l.RefreshItem(l.currentFocus)
+}
+
+// FocusLost is called after this GridWrap has lost focus.
+//
+// Implements: fyne.Focusable
+func (l *GridWrap) FocusLost() {
+	l.focused = false
+	l.Refresh() // TODO l.RefreshItem(l.currentFocus)
 }
 
 // MinSize returns the size that this widget should not shrink below.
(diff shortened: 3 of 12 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ More detailed release notes can be found on the [releases page](https://github.c
 
 * Go 1.17 or later is now required.
 * Icons for macOS bundles are now padded and rounded, disable with "-use-raw-icon"
+* Focus handling for List/Tree/Table are now at the parent widget not child elements
```
dweymouth, line 12: This behavior should be added to the new GridWrap widget as well

internal/driver/glfw/window.go
```diff
@@ -536,7 +536,20 @@ func (w *window) processMouseClicked(button desktop.MouseButton, action action,
 	}
 
 	if wid, ok := co.(fyne.Focusable); !ok || wid != w.canvas.Focused() {
-		w.canvas.Unfocus()
+		ignore := false
```
dweymouth, line 539: Reading the code here I can't really understand what effect this change is having and why it was necessary, but could it possibly be impacting those with custom lists that handle selection/focus themselves?

widget/list.go
```diff
@@ -239,6 +254,39 @@ func (l *List) ScrollToTop() {
...
 
+// TypedKey is called if a key event happens while this List is focused.
+//
+// Implements: fyne.Focusable
+func (l *List) TypedKey(event *fyne.KeyEvent) {
+	switch event.Name {
+	case fyne.KeySpace:
+		l.Select(l.currentFocus)
+	case fyne.KeyDown:
+		if f := l.Length; f != nil && l.currentFocus >= f()-1 {
```
dweymouth, line 265: Can we add public APIs to get the ListItemID of the first and last visible rows? (ie a public API similar to `.currentFocus` but just getting viewport info) - that way if someone needs to handle the focus themselves by overriding these Focusable methods they have all the public APIs needed to scroll the list if necessary

and same goes for Table, Tree, and GridWrap as well

widget/list.go
```diff
@@ -239,6 +254,39 @@ func (l *List) ScrollToTop() {
...
+			return
+		}
+		l.RefreshItem(l.currentFocus)
+		l.currentFocus++
+		l.scrollTo(l.currentFocus)
```
dweymouth, line 270: Can the `scrollTo` API be made public as well (same rationale as above comment)

widget/list.go
```diff
@@ -86,6 +89,23 @@ func (l *List) CreateRenderer() fyne.WidgetRenderer {
 	return newListRenderer(objects, l, l.scroller, layout)
 }
 
+// FocusGained is called after this List has gained focus.
+//
+// Implements: fyne.Focusable
+func (l *List) FocusGained() {
```
dweymouth, line 95: More of a general comment than specific to this PR, but is it possible with the focus system for a user to override a Focusable widget and effectively make it non-focusable by calling `FocusNext()` in the `FocusGained` implementation? Say if there were a use case where it really was easier to handle focus in the row widgets rather than the list itself

widget/list.go
dweymouth: Oh, I misread the implementation first time around and thought it was calling scrollTo only if the next-to-focus item was outside of the viewport. I guess these APIs aren't needed then but it still might be nice to have

widget/list.go
andydotxyz: If you know what "next" is you can ask the canvas to focus it in your "FocusGained", yes.

CHANGELOG.md
Jacalz: I agree. Likely just an oversight as it is such a new widget :)

andydotxyz: That's in now. Just needs some more tests.
Sorry I missed it - I was going by fyne_demo which it was not in, so I added it there too.

internal/driver/glfw/window.go
andydotxyz: Previously if you tapped an item that was not Focusable inside/over a focused item it would remove the focus (assuming they were different widgets).
What this does it changes it so that whilst "over" a focused widget it will only change focus if the child is Focusable.

andydotxyz: So for those with Focusable child elements who worked around this previous list focus handling it should still work.

widget/list.go
andydotxyz: I'm not sure I quite follow - this isn't tabbing through items inside a list any more, the list is focused and up/down move the focused item. Tabbing moves focus to the next item in chain - either the next sibling widget or any Focusable items inside the visible list widget items.
Does that help?

widget/list.go
andydotxyz: That would be consistent with the other collections

dweymouth: Nevermind, there's already the public ScrollTo, which just does bounds checking and then calls scrollTo

andydotxyz: Erm... as far as I can see List already has a public ScrollTo

internal/driver/glfw/window.go
dweymouth: Sounds like a good change 👍

widget/tree.go
```diff
@@ -717,6 +815,11 @@ func (n *treeNode) MouseOut() {
 
 func (n *treeNode) Tapped(*fyne.PointEvent) {
 	n.tree.Select(n.uid)
+	canvas := fyne.CurrentApp().Driver().CanvasForObject(n.tree)
+	if canvas != nil {
+		canvas.Focus(n.tree)
+	}
+	n.tree.currentFocus = n.uid
```
dweymouth, line 822: Need Refresh or RefreshItem calls here as well

widget/gridwrap.go
```diff
@@ -459,7 +511,13 @@ func (l *gridWrapLayout) setupGridItem(li *gridWrapItem, id GridWrapItemID, focu
 		f(id, li.child)
 	}
 	li.onTapped = func() {
+		canvas := fyne.CurrentApp().Driver().CanvasForObject(l.list)
+		if canvas != nil {
+			canvas.Focus(l.list)
+		}
+
 		l.list.Select(id)
+		l.list.currentFocus = id
```
dweymouth, line 520: Need Refresh() call here to update the focus indicator background rectangles. If you tap a new cell the focus indicator doesn't disappear from the previously-focused cell until you mouse over it.

widget/gridwrap.go
```diff
@@ -179,6 +199,70 @@ func (l *GridWrap) ScrollToOffset(offset float32) {
...
+		if f := l.Length; f != nil {
+			count = f()
+		}
+		// TODO l.RefreshItem(l.currentFocus)
+		l.currentFocus += l.getColCount()
+		if l.currentFocus >= count-1 {
+			l.currentFocus = count - 1
+		}
+		l.scrollTo(l.currentFocus)
+		l.Refresh() // TODO l.RefreshItem(l.currentFocus)
```
dweymouth, line 220: Not a blocker for this PR, but I think we should get the RefreshItem API in for GridWrap before 2.4 if possible

dweymouth (changes requested):

widget/list.go
dweymouth: the focus API also has `.FocusNext()` so should be good in any case

widget/gridwrap.go
andydotxyz: Ah yes sorry - it is a separate PR that Tilo was looking at - hence the TODO here.
If we get to next week and it's still missing then I will go ahead and complete that aspect separately.

dweymouth (approved):

Jacalz (approved):
Wonderful stuff. Makes the widgets super easy to navigate. Well done :)

PR #4103, opened 2023-07-28
Complete rounding by adding selection

### LINKED ISSUE #3981: New default theme updates (Rounded corners) missing on some of the built in widgets.
### Checklist

Im currently working in develop for the new default theme which has a lot of new rounded items like buttons, entries, etc. This is amazing step forward in an out of the box good style for fyne. Ive found a handful of other builtins that IMO should also adopt round corners to complete this look.

- Card border needs rounded corners
- List selection needs rounded corners
- Tree selection needs rounded corners
- Table selection needs rounded corners
- Dropdown menu container when opened needs rounded corners
- All dialog containers / splash needs rounded corners

It's also worth noting that list & tree selection should probably not have separators (or at least be able to be turned off). Hiding this looks sooo much better imo. I understand this can be done via theme and making separator transparent, but that affects using separators in other areas "I think ".

Also these may be more of taste and need more thought to be changed but could potentially be solved with rounded corners

- Scroll indicators look out of place. Consider rounded corners
- Split indicator looks out of place. Not exactly sure design direction here
- Tabs/Doc Tabs really look bad. Consider using current button style. Maybe even like gtk grouped buttons.

Attaching some images describing some of the issues above.

Screenshot 2023-06-17 at 8 09 11 AM
Screenshot 2023-06-17 at 8 10 58 AM
Screenshot 2023-06-17 at 8 11 14 AM

_No response_

I think a lot of this was probably just missed in the default theme update. Some of these need more thought probably.

Once I understand how these work I could possibly help on this :)

Ive created this at @andydotxyz request instead of just putting slack one by one as I find them :)

andydotxyz:
This is all we can do without rounded shadows I think...

Fixes #3981

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (commented):
I think we might want to work a little bit more on the spacing for the rounded selection. 
For example, the spacing in the Table widget looks a little bit off:
image
It also kind of seems to not be centered in list and tree. Larger gap at the top?
image

I think it would be nice to do something like Adwaita where the selection has a slight bit of padding on the outside (though I suspect we want less if that works with our design):
image

Jacalz (changes requested):
I think you also might want to update the dialog fileitem to use the new selection radius.

andydotxyz:
I'm not sure I follow which was missed, pretty sure I tested file dialog which was moved to GridWrap?

Jacalz:
> I'm not sure I follow which was missed, pretty sure I tested file dialog which was moved to GridWrap?

File dialog was never moved to GridWrap

andydotxyz:
I am quite confused about what is going on then.
Here is what I see:

andydotxyz:
I see on `dialog/file.go` line 481:

```go
	f.files = container.NewGridWrap
```

dweymouth:
@Jacalz is right - that's the GridWrap container, not the GridWrap widget

andydotxyz:
> @Jacalz is right - that's the GridWrap container, not the GridWrap widget

Aha absolutely. Thanks.
But is there more work to be done here, it looks correct to me...

Jacalz:
What I meant is that I couldn't see any mention in this PR about updating the `fileItem` to use the new radius. It still uses the input radius and I suspect it should use this section radius?

andydotxyz:
Aha I understand. I'd forgotten That some selection was using input radius. I will check for others.

andydotxyz:
That should fix it. I have a WiP move to GridWrap but the selection handling is proving a little complex so I'd like to open it as a different PR

Jacalz:
No problem. I wasn't suggesting it as part of this one :)

Did you have a look at my comment regarding the alignment?

andydotxyz:
> Did you have a look at my comment regarding the alignment?

I missed them.

Regarding spacing / padding around the rectangles - this has not changed since last release.
On a lowDPI screen as screenshotted the spacers are 1px wide in a 4px space - this will cause what you observed - it has been the same for a couple of releases now.
I don't think we should add more space around curved rectangles than we did around sharp ones.

I don't know exactly what you mean by "a slight bit of padding on the outside", but I am pretty sure that accommodating something like that would potentially mess with the otherwise standard alignment and/or be complicated additions to the theme API.

Jacalz:
Alright. I'll have a second look at this this eve thing if I can

andydotxyz:
Yes. We have an underline on tabs. If we round the rectangles those lines will look peculiar won't they?

If you can think of a nice layout I'll look into it but I couldn't find a way that made sense.

Jacalz:
I assumed that was the case, thanks. Just wanted to be sure. The only thing I can think of is having only the top two corners rounded but that isn't currently possible. I'll approve this :)

Jacalz (approved):

Issue #4086, opened 2023-07-22
Occasional deadlock on application start


Jacalz:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

I have noticed sometimes that `fyne_demo` sometimes freezes with a black screen on startup. It seems like we have some kind of deadlock issue affecting it. I haven't found a good reproducer but I'm opening this so we can try to get it fixed before issuing the release.

### How to reproduce

1. Open `fyne_demo` various times
2. You will occasionally (but not very often) see a black screen

### Example code

Have only seen it with `fyne_demo` up to this point but that might just be because I've spent the most time there and less on my personal projects...

### Fyne version

4a0f23166f221c9c5f9ecc657e0638aeabfee2d6

### Go compiler version

Go 1.20.6

### Operating system and version

Fedora Linux 38

vivekprm:
I was checking the race conditions for fyne_demo. Found few DATA_RACE warnings. Can I work on it?

`
WARNING: DATA RACE
Read at 0x00c00019daa0 by main goroutine:
  fyne.io/fyne/v2/internal/widget.(*BaseRenderer).Objects()
      /Users/vmishra/go/src/github.com/vivekprm/opensource/fyne/internal/widget/base_renderer.go:26 +0x37
  fyne.io/fyne/v2/widget.(*textRenderer).Objects()
      <autogenerated>:1 +0x1d
  fyne.io/fyne/v2/internal/driver.walkObjectTree()
      /Users/vmishra/go/src/github.com/vivekprm/opensource/fyne/internal/driver/util.go:160 +0x201
  fyne.io/fyne/v2/internal/driver.walkObjectTree.func1()
      /Users/vmishra/go/src/github.com/vivekprm/opensource/fyne/internal/driver/util.go:176 +0x473
  fyne.io/fyne/v2/internal/driver.walkObjectTree()
`

`
WARNING: DATA RACE
Read at 0x00c0001f0ca0 by main goroutine:
  fyne.io/fyne/v2/internal/driver.walkObjectTree()
      /Users/vmishra/go/src/github.com/vivekprm/opensource/fyne/internal/driver/util.go:189 +0x404
  fyne.io/fyne/v2/internal/driver.walkObjectTree.func1()
[...]

PR #4105, opened 2023-07-29
Fix potential infinite loop (create renderer -> refresh -> create renderer) in menu item

andydotxyz:
This was uncommon but could happen.
CreateRenderer should never initiate a Refresh as it can always reach this infinite loop.

### Checklist:

- [ ] Tests included. <- not possible in test driver with its threading model
- [x] Tests all pass.


Jacalz (approved):
Looks good. I think this might solve #4086 but I'm not sure. Started `fyne_demo` around 30 times without it happening. I am not sure how often it occurred before so it is hard to say if it was just a coincidence or not.

PR #4112, opened 2023-07-31
Hyperlink: Invoke OnTapped callback when typing space bar while focused

dweymouth:
### Description:
The Hyperlink widget can be activated by typing the space bar while focused in addition to tapping it, but the key event handler did not invoke the user-supplied OnTapped callback if set.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Nice catch thanks

PR #4118, opened 2023-08-01
sugest to avoid to memory leak

MatejMagat305:
avoid memory leak resume and sugest to avoid to call GO functions for empty life cycle functions ...

### DIFF
--- a/internal/driver/mobile/app/android.c
+++ b/internal/driver/mobile/app/android.c
@@ -70,8 +70,6 @@ static int main_running = 0;
 void processOnResume(ANativeActivity *activity) {
 	JNIEnv* env = activity->env;
 	setCurrentContext(activity->vm, (*env)->NewGlobalRef(env, activity->clazz));
-
-    onResume(activity);
 }
 
 // Entry point from our subclassed NativeActivity.
--- a/internal/driver/mobile/app/android.c
+++ b/internal/driver/mobile/app/android.c
@@ -67,7 +67,7 @@ jint JNI_OnLoad(JavaVM* vm, void* reserved) {
 static int main_running = 0;
 
 // ensure we refresh context on resume in case something has changed...
-void processOnResume(ANativeActivity *activity) {
+void onResume(ANativeActivity *activity) {
 	JNIEnv* env = activity->env;
 	setCurrentContext(activity->vm, (*env)->NewGlobalRef(env, activity->clazz));
 }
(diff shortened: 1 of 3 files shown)


internal/driver/mobile/app/android.c
```diff
@@ -70,10 +70,12 @@ static int main_running = 0;
 void processOnResume(ANativeActivity *activity) {
 	JNIEnv* env = activity->env;
 	setCurrentContext(activity->vm, (*env)->NewGlobalRef(env, activity->clazz));
-
-    onResume(activity);
```
andydotxyz, line 74: This cannot be right can it?
It looks like you have moved 3 of the callbacks from Go to C but deleted this one?

andydotxyz (commented):
I'm not sure about the way that some callbacks are deleted and others moved.

Why not make the PR just about the memory leak fix?

MatejMagat305:
> I'm not sure about the way that some callbacks are deleted and others moved.
> 
> Why not make the PR just about the memory leak fix?

this lines: https://github.com/fyne-io/fyne/pull/4118/files#diff-86b6487000ebad8b8c46272d0aa615cdbec1dba15013d82bed007ffa593e5464R90 is memory leaks, others are only therefore nothing theese functions do and call GO from C is recomended to eliminate ...

Jacalz:
But you have, as stated in https://github.com/fyne-io/fyne/pull/4118#discussion_r1289639647,  removed one of the functions. Why?

andydotxyz:
I think I see why it looks inconsistent - because the `onResume` was written in C to do something but not in Go. But it was called `processOnResume`. If that was renamed to `onResume` in the C file then I think we would have a consistent code and this could be approved.
Is that OK @MatejMagat305 ?

MatejMagat305:
ok

Jacalz:
Why did you close it? Andy just suggested changes

MatejMagat305:
ach soory, I will changes tomorow

andydotxyz (approved):
Thanks for making these optimisations

Jacalz (approved):
Thanks. If there are more places where we are passing Go functions back into C, we probably want to remove those as well some time in the future. It is quite slow from what I have understood (but a lot faster in Go 1.21).

MatejMagat305:
> Thanks. If there are more places where we are passing Go functions back into C, we probably want to remove those as well some time in the future. It is quite slow from what I have understood (but a lot faster in Go 1.21).

well there are some functions like `onContentRectChanged` or `onNativeWindowCreated`, which are empty, but I did not know which one will you plan use ..., life time I dare move because I saw that life time are binding with running loop (are call from other way ...)

PR #4119, opened 2023-08-02
Fix memory leak in fontSize cache

### LINKED ISSUE #4108: Memory leak in font metrics cache
### Checklist

### Describe the bug

The `fontSize` cache [here](https://github.com/fyne-io/fyne/blob/develop/internal/cache/text.go#L10) is only cleaned out on settings change, and it stores arbitrary user-supplied strings in the map key. It should be made an expiring cache and have expired entries cleaned out during the periodic clean task.

### How to reproduce

Run fyne_demo and scroll through all rows of the list/table demos, or any other text-heavy Fyne app.

### Screenshots

_No response_

### Example code

fyne-demo

### Fyne version

develop

### Go compiler version

go1.19.4 darwin/amd64

Mac OS

### Additional Information

_No response_

dweymouth:
### Description:
The fontSizeCache which stores size information for a given string and font size+style was only cleaned out on settings change. Since it stores arbitrary strings from user code it would be a slow leak over time especially for text-heavy apps.

This PR makes the cache entries expire with the standard expiry time and adds the cleanup to the Clean task that is run on paint events. I am open to discussion if the cleaning should be done on a different schedule and/or the expiry should be adjusted.

Fixes #4108 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/cache/base.go
+++ b/internal/cache/base.go
@@ -45,6 +45,7 @@ func Clean(canvasRefreshed bool) {
 		return
 	}
 	destroyExpiredSvgs(now)
+	destroyExpiredFontMetrics(now)
 	if canvasRefreshed {
 		// Destroy renderers on canvas refresh to avoid flickering screen.
 		destroyExpiredRenderers(now)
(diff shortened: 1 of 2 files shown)


andydotxyz (approved):
Good catch thanks

PR #4121, opened 2023-08-02
Call OnChange with every updateText, fixes #4117

### LINKED ISSUE #4117: Backspace and Delete key Do not Fire OnChanged Event
### Checklist

### Describe the bug

Entry widget do not fire Onchanged event on Delete and Backspace Key.

### How to reproduce

Attached code

### Screenshots

_No response_

### Example code

package main

import (
	"fmt"

	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	fyneApp := app.NewWithID("com.xyz")
	mainWin := fyneApp.NewWindow("Bug Report")
	mainWin.SetMaster()
	mainWin.Show()

	w := widget.NewEntry()
	w.SetPlaceHolder("On Backspace and Delete Key OnChanged event do not fire")
	w.OnChanged = func(v string) {
		fmt.Println("On Changed called")
	}

	mainWin.SetContent(w)

	fyneApp.Run()
}

### Fyne version

fyne.io/fyne/v2 v2.3.6-0.20230801132204-0b324c4306a0

### Go compiler version

1.20

Linux

### Additional Information

_No response_

mbaklor:
### Description:

This adds a call to `entry.OnChanged()` every time `entry.updateText` is called.
While I was at it I also added a call to `e.Validate` where I didn't see one.

Fixes #4117 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -832,9 +832,8 @@ func (e *Entry) cutToClipboard(clipboard fyne.Clipboard) {
 	e.copyToClipboard(clipboard)
 	e.setFieldsAndRefresh(e.eraseSelection)
 	e.propertyLock.Lock()
-	content := e.textProvider().String()
 	if e.OnChanged != nil {
-		e.OnChanged(content)
+		e.OnChanged(e.Text)
 	}
 	e.propertyLock.Unlock()
 	e.Validate()
@@ -1055,9 +1054,8 @@ func (e *Entry) selectingKeyHandler(key *fyne.KeyEvent) bool {
 		// clears the selection -- return handled
 		e.setFieldsAndRefresh(e.eraseSelection)
 		e.propertyLock.Lock()
-		content := e.textProvider().String()
 		if e.OnChanged != nil {
-			e.OnChanged(content)
+			e.OnChanged(e.Text)
 		}
 		e.propertyLock.Unlock()
 		e.Validate()
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -485,9 +485,11 @@ func (e *Entry) Append(text string) {
 	changed := e.updateText(content)
 	e.propertyLock.Unlock()
 
-	e.Validate()
-	if changed && e.OnChanged != nil {
-		e.OnChanged(content)
+	if changed {
+		e.Validate()
+		if e.OnChanged != nil {
+			e.OnChanged(content)
+		}
 	}
 	e.Refresh()
 }
@@ -669,9 +671,11 @@ func (e *Entry) TypedKey(key *fyne.KeyEvent) {
 		e.selecting = false
 	}
 	e.propertyLock.Unlock()
-	e.Validate()
-	if changed && e.OnChanged != nil {
-		e.OnChanged(content)
+	if changed {
+		e.Validate()
+		if e.OnChanged != nil {
+			e.OnChanged(content)
+		}
 	}
 	e.Refresh()
 }
@@ -849,9 +853,11 @@ func (e *Entry) eraseSelection() {
 	content := provider.String()
 	changed := e.updateText(content)
 
-	e.Validate()
-	if changed && e.OnChanged != nil {
-		e.OnChanged(content)
+	if changed {
+		e.Validate()
+		if e.OnChanged != nil {
+			e.OnChanged(content)
+		}
 	}
 }
 
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -833,7 +833,9 @@ func (e *Entry) cutToClipboard(clipboard fyne.Clipboard) {
 	e.setFieldsAndRefresh(e.eraseSelection)
 	e.propertyLock.Lock()
 	content := e.textProvider().String()
-	e.OnChanged(content)
+	if e.OnChanged != nil {
+		e.OnChanged(content)
+	}
 	e.propertyLock.Unlock()
 	e.Validate()
 }
@@ -1054,7 +1056,9 @@ func (e *Entry) selectingKeyHandler(key *fyne.KeyEvent) bool {
 		e.setFieldsAndRefresh(e.eraseSelection)
 		e.propertyLock.Lock()
 		content := e.textProvider().String()
-		e.OnChanged(content)
+		if e.OnChanged != nil {
+			e.OnChanged(content)
+		}
 		e.propertyLock.Unlock()
 		e.Validate()
 		return true
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -831,6 +831,11 @@ func (e *Entry) cutToClipboard(clipboard fyne.Clipboard) {
 
 	e.copyToClipboard(clipboard)
 	e.setFieldsAndRefresh(e.eraseSelection)
+	e.propertyLock.Lock()
+	content := e.textProvider().String()
+	e.OnChanged(content)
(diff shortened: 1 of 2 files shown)


widget/entry.go
```diff
@@ -481,9 +481,13 @@ func (e *Entry) Append(text string) {
 	e.propertyLock.Lock()
 	provider := e.textProvider()
 	provider.insertAt(provider.len(), text)
-	e.updateText(provider.String())
+	changed := e.updateText(provider.String())
 	e.propertyLock.Unlock()
 
+	e.Validate()
+	if changed && e.OnChanged != nil {
+		e.OnChanged(provider.String())
```
Jacalz, line 489: You should avoid calling `provider.String()` multiple times. If you look at the code you will see that it allocates an entirely new string each time. You probably want to save the string to a variable.

widget/entry.go
```diff
@@ -658,11 +662,15 @@ func (e *Entry) TypedKey(key *fyne.KeyEvent) {
 	}
 
 	e.propertyLock.Lock()
-	e.updateText(provider.String())
+	changed := e.updateText(provider.String())
 	if e.CursorRow == e.selectRow && e.CursorColumn == e.selectColumn {
 		e.selecting = false
 	}
 	e.propertyLock.Unlock()
+	e.Validate()
```
Jacalz, line 670: This leads to unwanted side effects. Pressing tab inside a multiline entry calls validation twice but we also validate the text even in all cases where we know that nothing has changed (i.e. moving the cursor using tab keys but also pressing keys like home, page up, end etc.). I don't think doing validation that often will be good if users have slow validators.

widget/entry.go
```diff
@@ -836,7 +844,12 @@ func (e *Entry) eraseSelection() {
 	e.CursorRow, e.CursorColumn = e.rowColFromTextPos(posA)
 	e.selectRow, e.selectColumn = e.CursorRow, e.CursorColumn
 	e.selecting = false
-	e.updateText(provider.String())
+	changed := e.updateText(provider.String())
+
+	e.Validate()
```
Jacalz, line 849: This change also leads to verifying twice (potentially running `OnChanged` twice as well, I would assume) when selecting something and then pressing another key using `TypedKey` or `TypedRune` to replace the text.

Jacalz (changes requested):
Thanks for the PR. This  does indeed fix the issue but I noticed that introduces some unwanted side effects that I think we need to handle. Some of the functions that now validate and run `OnChanged` does unfortunately later on call them again. I left some comments inline about what we need to look out for.

Jacalz:
FYI @mbaklor. I think the `Fixes #` needs to be on a separate line for GitHub to understand that it should be closed. I updated the text for you :)

mbaklor:
Very fair, I actually threw the validate in all those cases as an afterthought because I think hitting backspace should validate as well. I'll take a look tomorrow and remove unnecessary ones and think how best to validate backspace 😁👍

widget/entry.go
mbaklor: When erasing selected text by typing the callback is ignored, so I felt confident using the OnChanged here, however looking again at where `eraseSelection` gets called, it looks like the only place that doesn't then call the callback itself is when pressing backspace or delete while in selecting mode? not sure how best to handle that.

widget/entry.go
mbaklor: np, fixed in next commit

mbaklor:
okay as you can see I'm very good at using the comments thing in the code lol, I just tested (manually, I should've probably written a test but I only thought of checking after the commit 😅) and looks like everything now calls validate and onchanged only once

andydotxyz:
I haven't tested myself, but at a glance I would be concerned that highlighting text and the entering a key may validate and fire handlers twice - once when clearing selection and another when entering the new text.
The "clearSelection" was created as a helper so be sure to check all the places it is called.

mbaklor:
Within `TypedRune` there's a little snippet to clear `OnChanged` before erasing selection, so I figured it was fine, I'll go over where `eraseSelection` is called again just to be sure

mbaklor:
okay so there were some repetitions of the callbacks from `eraseSelection`, so I removed the callbacks from there and into where it wasn't called when not in `eraseSelection`. Hope this was the correct choice, lmk if that's not the solution

Jacalz:
There are a lot of test failures (crashes). Would you mind having a look at those?

mbaklor:
oh my god I should not push commits at 1:00 AM (not that now isn't 1:00 AM but it is what it is)

yes I didn't check that `OnChanged` exists before calling it. fixed now. commit incoming 😅

andydotxyz (approved):
Looks great, thanks so much

widget/entry.go
```diff
@@ -817,6 +831,13 @@ func (e *Entry) cutToClipboard(clipboard fyne.Clipboard) {
 
 	e.copyToClipboard(clipboard)
 	e.setFieldsAndRefresh(e.eraseSelection)
+	e.propertyLock.Lock()
+	content := e.textProvider().String()
+	if e.OnChanged != nil {
+		e.OnChanged(content)
```
Jacalz, line 837: I realized that this still is creating a new copy of the string. Looking at `eraseSelection`, it calls `e.updateText()` which sets the `e.Text` field and as we are in a lock we can simply reused that instead of recreating the string once more removing the need to allocate a new string.

```suggestion
	if e.OnChanged != nil {
		e.OnChanged(e.Text)
```

widget/entry.go
```diff
@@ -1033,6 +1054,13 @@ func (e *Entry) selectingKeyHandler(key *fyne.KeyEvent) bool {
 	case fyne.KeyBackspace, fyne.KeyDelete:
 		// clears the selection -- return handled
 		e.setFieldsAndRefresh(e.eraseSelection)
+		e.propertyLock.Lock()
+		content := e.textProvider().String()
+		if e.OnChanged != nil {
+			e.OnChanged(content)
```
Jacalz, line 1060: Same comment as above applies here.

```suggestion
		if e.OnChanged != nil {
			e.OnChanged(e.Text)
```

Jacalz (changes requested):
Thanks. This is looking great. I just released that we still are allocating new strings in this code so I added some suggestions inline to remedy the issue.

Sorry for the long review process. This will be ready to land very soon, I promise :)

Jacalz:
@mbaklor Would you be able to get the last two changes done soon? This is one of the remaining release blockers and it would be good to have it resolved as soon as possible :)

mbaklor:
Yes absolutely, I was planning on getting it done tonight, I'll see if I can get to it earlier but I doubt

Jacalz:
Great. Thanks. There are other blockers that haven't been solved yet so a few hours earlier or later today isn't a problem

mbaklor:
> Sorry for the long review process. This will be ready to land very soon, I promise :)

No worries, I'm honestly glad for the long and thorough process, I'm still rather insecure in my golang and happy to have someone else make sure what I'm contributing won't break the library 😉

Jacalz (approved):
This looks great now and solves one of the important blocker bugs which is wonderful. Thanks for working on this and congratulations on landing your first change into Fyne :)

PR #4122, opened 2023-08-02
Move the FileOpen/Save dialog to new GridWrap widget

andydotxyz:
### Checklist:

- [x] Tests included.
- [x] Tests all pass.


dialog/file.go
```diff
@@ -445,44 +448,63 @@ func (f *fileDialog) setLocation(dir fyne.URI) error {
...
+				dir := f.data[id]
+				parent := id == 0 && len(dir.Path()) < len(f.dir.Path())
+				_, isDir := dir.(fyne.ListableURI)
+				o.(*fileDialogItem).setLocation(dir, isDir || parent, parent)
+			},
 		)
+		f.files.(*widget.GridWrap).OnSelected = func(id widget.GridWrapItemID) {
+			f.selectedID = id
+			f.setSelected(f.data[id], id)
+		}
```
Jacalz, line 494: You should be able to switch `widget.GridWrapItemID` to `int` and put the function next to count and template. It should make it possible to reuse the exact same function for both :)

If it is possible to use the same functions for the update functions, you should be able to do the same thing there.

dialog/file_test.go
```diff
@@ -341,9 +352,10 @@ func TestShowFileSave(t *testing.T) {
 		return
 	}
 
-	// This will only execute if we have a file in the home path.
-	// Until we have a way to set the directory of an open file dialog.
-	test.Tap(target)
+	abs, err := filepath.Abs("./testdata/")
```
Jacalz, line 355: From static analysis: `his value of err is never used`

dialog/fileitem.go
```diff
@@ -19,55 +17,48 @@ const (
...
+		item:    i,
+		icon:    icon,
+		text:    text,
+		objects: []fyne.CanvasObject{icon, text},
 	}
 }
 
 func (i *fileDialogItem) isDirectory() bool {
 	return i.dir
 }
```
Jacalz, line 42: Unused. Can be deleted.

Jacalz (changes requested):
Nicely done. Very impressive and such a nice change. Feels smoother and the new keyboard navigation works wonders. Just left a few comments inline. Should be easy to fix :)

dialog/file.go
```diff
@@ -326,48 +328,49 @@ func (f *fileDialog) loadFavorites() {
...
 
-	f.files.Objects = icons
 	f.files.Refresh()
 	f.filesScroll.Offset = fyne.NewPos(0, 0)
 	f.filesScroll.Refresh()
 }
 
 func (f *fileDialog) setLocation(dir fyne.URI) error {
+	if f.selectedID > -1 {
+		f.files.(interface{ Unselect(int) }).Unselect(f.selectedID)
```
Jacalz, line 372: You can probably move this interface into its own type and add `.Refresh()` you use that instead of `fyne.Widget` in the fileDialog struct and avoid casting the value here.

dialog/file.go
```diff
@@ -445,44 +448,63 @@ func (f *fileDialog) setLocation(dir fyne.URI) error {
...
+				dir := f.data[id]
+				parent := id == 0 && dir != f.dir
+				_, isDir := dir.(fyne.ListableURI)
+				o.(*fileDialogItem).setLocation(dir, isDir, parent)
+			},
+		)
+		f.files.(*widget.List).OnSelected = func(id widget.ListItemID) {
+			f.selectedID = id
+			f.setSelected(f.data[id], id)
+		}
```
Jacalz, line 507: If you use a `&widget.List{}` type here, you can avoid having to cast the variable to get the `OnSelected` field and just set al of the functions directly instead.

Jacalz (changes requested):
Sorry, missed two comments :sweat_smile:

dialog/file_test.go
andydotxyz: Thanks, I missed that as I was iterating the test

dialog/file.go
andydotxyz: Great idea thanks. Though I extended `fyne.Widget` so it could be passed to container constructor too.

Jacalz (approved):
Nice. Well done :)

PR #4128, opened 2023-08-03
Move to a truncate mode separate to wrapping

andydotxyz:
This avoids combinatorial explosion when we start to wrap AND truncate

Tasks:

- [x] Fix API to add multiple truncate modes to different wrap modes
- [x] Fix ellipsis mode for multi-line non-wrapped
- [x] Add support for truncation on wrap (use space available)
- [x] Add ellipsis support when truncating wrapped content

Possibly a future PR as this is not actually an API change (as this requires more discussion and is not actually an API change):
- [ ] Update Entry API for the "force no scroll" truncation mode (uses TextWrapOff at the moment, TextWrapTruncate is the scroll version)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/text.go
+++ b/text.go
@@ -13,6 +13,22 @@ const (
 	TextAlignTrailing
 )
 
+// TextTruncation controls how a `Label` or `Entry` will truncate its text.
+// The default value is `TextTruncateOff` which will not truncate.
+//
+// Since: 2.4
+type TextTruncation int
+
+const (
+	TextTruncateOff TextTruncation = iota
+	TextTruncateClip
+	// TextTruncateEllipsis is like regular truncation except that an ellipses (…) will be inserted
+	// wherever text has been shortened to fit.
+	//
+	// Since: 2.4
+	TextTruncateEllipsis
+)
+
 // TextWrap represents how text longer than the widget's width will be wrapped.
 type TextWrap int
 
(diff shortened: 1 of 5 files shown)


text.go
```diff
@@ -13,6 +13,25 @@ const (
...
+	// TextTruncateOff means no truncation will be applied, it is the default.
+	// This means that the minimum size of a text block will be the space required to display it fully.
+	TextTruncateOff TextTruncation = iota
+	// TextTruncateClip will truncate text when it reaches the end of the available space.
+	TextTruncateClip
+	// TextTruncateEllipsis is like regular truncation except that an ellipses (…) will be inserted
+	// wherever text has been shortened to fit.
+	//
+	// Since: 2.4
+	TextTruncateEllipsis
```
dweymouth, line 32: I remember there being an issue somewhere suggesting a middle ellipsis truncation mode eventually as well. Would it make more sense to name this one `TextTruncateTrailingEllipsis` then? (it is a bit of a mouthful) or is it ok to eventually have `TextTruncateEllipsis` and `TextTruncateMiddleEllipsis`?

andydotxyz: Good question. It depends on whether the leading/middle/trailing is to be separated from the ellipsis itself (some locales will want something other than an ellipsis).
I am guessing we will handle the rune in our localisation work so "ellipsis" is synonymous with the generic concept not the actual rune. In which case I think your suggestion is a good one.

That said, trailing vs leading depends on text direction - so I am pretty tempted to stick with current naming to mean "use ellipsis or similar, and truncate at the normal position for current direction" and we can add "middle ellipsis" in the future - or even something for the "other" end...

andydotxyz:
Marking as release blocker as an API added for v2.4.0 wasn't right - this feature addition resolves.

Jacalz (approved):
Looks good. Well done. I noticed some things that can be improved in the existing code in some places. Will open a PR for that once this has been merged.

Jacalz:
Also, was the truncation left out of Entry intentionally or not?

andydotxyz:
> Also, was the truncation left out of Entry intentionally or not?

The existing "workaround" usage of truncation variable to set scrolling mode still works - but I think we should find a better way. Not including as I cannot see the best way forward yet. The existing setup isn't great and when we fix it we should make it much better / clearer.

PR #4132, opened 2023-08-03
Fix keybinding for Entry shortcuts on Mac OS

### LINKED ISSUE #2462: navigating in entry quickly with ctrl key

dweymouth:
### Description:
On Mac, Alt is the modifier used for moving left and right by words in text entries, instead of Cmd. Also, Cmd+Left/Cmd+Right behave like home and end on Mac (since most Mac keyboards don't have home/end)

As a follow-up, Cmd+Shift+Left/Right should select from the cursor position to the beginning/end of the line, but we don't have those shortcuts in for the other platforms yet either so that can be done as a separate task

Fixes #2462

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -3,6 +3,7 @@ package widget
 import (
 	"image/color"
 	"math"
+	"runtime"
 	"strings"
 	"time"
 	"unicode"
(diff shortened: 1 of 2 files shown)


andydotxyz (approved):
Nice one thanks so much :)

PR #4135, opened 2023-08-04
Minor redesign of tabs to use rounded corners

### LINKED ISSUE #4130: Improve design of tabs to fit better with rounded corners
### Checklist

The current design of the tab bar is inherently hard to adapt to the new theme look with rounded corners. it is especially true for the selection. We should think about updating the design to fit in better with the new theme. This is how it looks in contrast to the rounded button:
image

Just making the selection rectangle rounded would not fit in with the tab bar below.

It depends on the ideas being used. If we want to solve the idea about using rounded top corners then no. Otherwise yes.

One idea that I initially had was to solve the issue by using rounded corners on the top part of the selection. I don't if it is the best solution but it is (apart from the fact that we can't individually control rounded corners yet) more or less the simplest as we won't need to get rid of the tab bar. See mockup below:
image

### Some other inspiration:
How tabs in Firefox look:
image

How tabs look in Adwaita:
image

Jacalz:
### Description:

This implements my suggestion about adding rounded hover selection and also rounds off the indicator. I am very happy with the look of this. 

#### Video demo in Rymdport:
[Screencast from 2023-08-04 22-22-34.webm](https://github.com/fyne-io/fyne/assets/25466657/43a82bac-db3c-48bd-9023-dc827ad6ba20)

#### Images:
image
Screenshot from 2023-08-04 22-40-02

Fixes #4130 

### Checklist:

- [ ] Tests included. <- Covered by existing tests
- [x] Tests all pass.


andydotxyz (approved):
Looks good thanks for figuring a nice solution

PR #4140, opened 2023-08-06
generic list binding

pseidemann:
### Description:

implement (experimental) generic list binding.

I implemented two new interfaces List and ExternalList which are identical to their untyped counterparts, except that all types which were `interface{}` are now replaced by a type parameter. the original interfaces are now just using the new implementation with a hardcoded type `any`. thus allowing using the old api as before and using the new generic api at the same time. the implementation itself virtually didn't change.

I noticed there are no tests for the existing untyped implementation? this change should keep them green if they were tests for it. let me know if I missed something.

this is meant as a demo/proof of concept how generics could be used to get rid of `interface{}` apis, which require type annotations everywhere. all other binding implementations could be updated in a similar fashion.

caveats:

- for generics we need to support at least go 1.18
- before go 1.20, the `any` type didn't satisfy `comparable`, which is annoying if we want to use the new implementation for the old api (used in this demo but could be changed to be 1.18 compatible)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/data/binding/binding.go
+++ b/data/binding/binding.go
@@ -89,6 +89,12 @@ type Untyped interface {
 	Set(interface{}) error
 }
 
+type Typed[E any] interface {
+	DataItem
+	Get() (E, error)
+	Set(E) error
+}
+
 // NewUntyped returns a bindable interface{} value that is managed internally.
 //
 // Since: 2.1
--- a/data/binding/bindlists.go
+++ b/data/binding/bindlists.go
@@ -1355,6 +1355,17 @@ type UntypedList interface {
 	SetValue(index int, value interface{}) error
 }
 
+type List[E any] interface {
+	DataList
+
+	Append(value E) error
+	Get() ([]E, error)
+	GetValue(index int) (E, error)
+	Prepend(value E) error
+	Set(list []E) error
+	SetValue(index int, value E) error
+}
+
 // ExternalUntypedList supports binding a list of interface{} values from an external variable.
 //
 // Since: 2.1
@@ -1364,11 +1375,21 @@ type ExternalUntypedList interface {
 	Reload() error
 }
 
+type ExternalList[E any] interface {
+	List[E]
+
+	Reload() error
+}
+
 // NewUntypedList returns a bindable list of interface{} values.
 //
 // Since: 2.1
 func NewUntypedList() UntypedList {
-	return &boundUntypedList{val: &[]interface{}{}}
+	return NewList[any]()
+}
+
+func NewList[E comparable]() List[E] {
+	return &boundList[E]{val: &[]E{}}
 }
 
 // BindUntypedList returns a bound list of interface{} values, based on the contents of the passed slice.
@@ -1376,27 +1397,31 @@ func NewUntypedList() UntypedList {
 //
 // Since: 2.1
 func BindUntypedList(v *[]interface{}) ExternalUntypedList {
+	return BindList(v)
+}
+
+func BindList[E comparable](v *[]E) ExternalList[E] {
 	if v == nil {
-		return NewUntypedList().(ExternalUntypedList)
+		return NewList[E]().(ExternalList[E])
 	}
 
-	b := &boundUntypedList{val: v, updateExternal: true}
+	b := &boundList[E]{val: v, updateExternal: true}
 
 	for i := range *v {
-		b.appendItem(bindUntypedListItem(v, i, b.updateExternal))
+		b.appendItem(bindListItem(v, i, b.updateExternal))
 	}
 
 	return b
 }
 
-type boundUntypedList struct {
+type boundList[E comparable] struct {
 	listBase
 
 	updateExternal bool
-	val            *[]interface{}
+	val            *[]E
 }
 
-func (l *boundUntypedList) Append(val interface{}) error {
+func (l *boundList[E]) Append(val E) error {
 	l.lock.Lock()
 	defer l.lock.Unlock()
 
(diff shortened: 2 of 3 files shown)


Jacalz:
Thanks for the PR but I’m afraid that we can’t merge this. We are using Go 1.17 as the base/minimum Go version for the upcoming release on `develop`. Generics were not added until Go 1.18.

pseidemann:
hi @Jacalz,
this is quite unfortunate. any chance the version can be bumped?

go 1.17 is now 2 years old and with the release of go 1.21 this august, it will be unsupported since three go releases.

only the last two "major" releases are supported:

> Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release. We fix critical problems, including [critical security problems](https://go.dev/security), in supported releases as needed by issuing minor revisions (for example, Go 1.6.1, Go 1.6.2, and so on).

from: https://go.dev/doc/devel/release

so 1.17 is unsupported since go 1.19 and the latest is 1.20 (and this month will be 1.21)

Jacalz:
We are aware of the release cycles and maintenance of old Go versions but the short answer is basically that we want Fyne to be able to run on as many platforms as possible.

Many LTS versions of Linux (Ubuntu, Debian, etc.) ship with an older version of Go (still gets security patches backported from newer versions though) and we want developers on those systems to be able to use Fyne. The Go team have also been dropping support for macOS versions with every other release (though 1.18 and 1.19 seem to have been quite calm there), basically turning old Apple computers into e-waste even though there are a lot of people still using them.

We usually try to follow the Go version in the latest stable Debian release. At the moment that is [1.19](https://packages.debian.org/bookworm/golang-go). Apart from Go 1.18 upping the minimum iOS version to 12, there doesn't seem to be any breakages. However, we decided to stick with 1.17 to allow all of the tolling and best practices  around generics to stabilise. The upcoming release will have to stick to Go 1.17 but we would have to have a meeting to discuss potentially changing it for the release after that (v2.5.0 or whatever it turns out to be). It certainly isn't impossible that we might raise the version but I can't guarantee anything.

We really should write up a common answer to that question because it has come up some many times over the year.

Jacalz:
I'm also not entirely sure about introducing a half-baked version of generics. A lot of useful parts of it feels a bit like it hasn't been fully feature complete up until 1.20 or the upcoming 1.21 in terms of the `cmp` package.

pseidemann:
I understand 👍 

so normally you would have been on 1.19 now, but because of the introduction of generics in 1.18, you wanted to wait a bit before opening pandora's box?

> We really should write up a common answer to that question because it has come up some many times over the year.

this sounds like that there is actually demand in bumping the version. I think it will be worth it, if the project takes full advantage of generics.

the bump is not urgent and it sounds like you might discuss it in the future anyway, so I might need to be just a bit more patient.

if/when generics will be possible, I'm happy to take another stab at integrating them 🙂

pseidemann:
in case anyone is interested, this is how one can _wrap_ the existing implementation to use generics.
it doesn't work for "external" though, since the pointer would be lost during the conversion from typed to untyped.
but instead of `Reload()` one could just always use `Set()`.

<details>
  <summary>list wrapping</summary>

  ```go
  package bindingx

  import "fyne.io/fyne/v2/data/binding"

  type List[E any] interface {
    binding.DataList

    Append(value E) error
    Get() ([]E, error)
    GetValue(index int) (E, error)
    Prepend(value E) error
    Set(list []E) error
    SetValue(index int, value E) error
  }

  type list[E any] struct {
    untyped binding.UntypedList
  }

  // constructors

  func NewList[E any]() List[E] {
    return &list[E]{
      untyped: binding.NewUntypedList(),
    }
  }

  // binding.DataItem implementation

  func (b *list[E]) AddListener(l binding.DataListener) {
    b.untyped.AddListener(l)
  }

  func (b *list[E]) RemoveListener(l binding.DataListener) {
    b.untyped.RemoveListener(l)
  }

  // binding.DataList implementation

  func (l *list[E]) GetItem(index int) (binding.DataItem, error) {
    return l.untyped.GetItem(index)
  }

  func (l *list[E]) Length() int {
    return l.untyped.Length()
  }

  // List implementation

  func (l *list[E]) Append(value E) error {
    return l.untyped.Append(value)
  }

  func (l *list[E]) Get() ([]E, error) {
    val, err := l.untyped.Get()
    return toTypedSlice[E](val), err
  }

  func (l *list[E]) GetValue(index int) (E, error) {
    val, err := l.untyped.GetValue(index)
    return val.(E), err
  }

  func (l *list[E]) Prepend(value E) error {
    return l.untyped.Prepend(value)
  }

  func (l *list[E]) Set(list []E) error {
    return l.untyped.Set(toAnySlice[E](list))
  }

  func (l *list[E]) SetValue(index int, value E) error {
    return l.untyped.SetValue(index, value)
  }

  // helpers

  func toTypedSlice[E any](s []any) []E {
    typed := make([]E, len(s))
    for i, v := range s {
      typed[i] = v.(E)
    }
    return typed
  }

  func toAnySlice[E any](s []E) []any {
    untyped := make([]any, len(s))
    for i, v := range s {
      untyped[i] = v
    }
    return untyped
  }
  ```
</details>

<details>
  <summary>value wrapping</summary>

  ```go
  package bindingx

  import "fyne.io/fyne/v2/data/binding"

  type Value[E any] interface {
    binding.DataItem

    Get() (E, error)
    Set(value E) error
  }

  type value[E any] struct {
    untyped binding.Untyped
  }

  func New[E any]() Value[E] {
    return &value[E]{
      untyped: binding.NewUntyped(),
    }
  }

  // binding.DataItem implementation

  func (b *value[E]) AddListener(l binding.DataListener) {
    b.untyped.AddListener(l)
  }

  func (b *value[E]) RemoveListener(l binding.DataListener) {
    b.untyped.RemoveListener(l)
  }

  // Value implementation

  func (b *value[E]) Get() (E, error) {
    val, err := b.untyped.Get()
    return val.(E), err
  }

  func (b *value[E]) Set(value E) error {
    return b.untyped.Set(value)
  }
  ```
</details>

Jacalz:
> so normally you would have been on 1.19 now, but because of the introduction of generics in 1.18, you wanted to wait a bit before opening pandora's box?

Yeah, basically.

> this sounds like that there is actually demand in bumping the version. I think it will be worth it, if the project takes full advantage of generics.

I wouldn't call it a huge demand really. It was mostly as case about updating from Go 1.14 as our previous minimum version. Apart from developers wanting new features, there isn't much demand from users as far as I see. Given the Go 1 stability guarantees, we can always compile with the latest versions :)

Jacalz:
> the bump is not urgent and it sounds like you might discuss it in the future anyway, so I might need to be just a bit more patient.

Indeed :)

PR #4148, opened 2023-08-09
Add range/column helper function

mgazza:
### Description:
Added some helper functions to aid in ranging over the text grid

### Checklist:

- [x] Tests included.
- [ ] Tests all pass.


andydotxyz (changes requested):
Thanks, this is great work - but will require API change to meet our style. Rather than a detailed review inline let's do this high level for the first pass.

Primarily the `Options` pattern is not something we use - so the type and functions utilising this should be updated to remove this (note also that the static functions it requires pollute the API of the entire widget package just for TextGrid usage).
Only expose APIs that are actually required for this work (i.e. InvertColor could be private).

Is `HighlightedTextGridStyle` truly required? it could be an internal detail I think.
Even if it is there seems to be no need to include the `Original` alongside `Highlighted`, as one could be inferred from other data I think?

Mostly I wonder - can the state of highlighting not be tracked in the `TextGrid` instead of the `TextGridCell(Style)`? More like how Entry handles selection, which would feel more consistent if this is indeed for selection usage...

p.s. also is this "Selection" or "Highlight" - I am not sure if we have used the term highlight anywhere else, should TextGrid be adding it? And if it is selection - perhaps the style should use the theme selection somehow instead of just inverting?

mgazza:
I was actually thinking that the HighlightedTextGridStyle could replace the CustomTextGridStyle with the Functional Options pattern. 
I agree that InvertColor should be private.
I could move more of the functionality from FyneTerm down to this in terms of mouse position tracking, drag selection, etc. 

This becomes a lot more obvious I think when we try to implement blink/flash. as the style will need to alternate on a clock and if the user currently has that highlighted the style will need to invert on each clock.

Please check out the changes made in https://github.com/codelaboratoryltd/terminal/tree/add-clipboard
I have added selection, block selection, right-click copy

In the meantime, I will take a look at how Entry manages selection.

andydotxyz:
> I was actually thinking that the HighlightedTextGridStyle could replace the CustomTextGridStyle with the Functional Options pattern.

We should not move to functional options for a single widget. That is an API design decision and would have to be reflected globally to make sense. We don't use the pattern and adopting it would be a much wider discussion.

> This becomes a lot more obvious I think when we try to implement blink/flash. as the style will need to alternate on a clock and if the user currently has that highlighted the style will need to invert on each clock.

Not sure I agree on this, the theme currently includes all the info needed to do this - see how Entry is implemented already

andydotxyz:
Fair enough. But if you think inverting is the only way to do it then exporting new API types to allow different sort of selection colouring doesn't seem necessary either?

There's always outlining as an option too ;)

mgazza:
I've heavily scaled back the changes, the title of this needs updating to align now

widget/textgrid_test.go
```diff
@@ -9,7 +9,6 @@ import (
 	"fyne.io/fyne/v2/canvas"
 	"fyne.io/fyne/v2/test"
 	"fyne.io/fyne/v2/theme"
-
```
andydotxyz, line 12: Import groups should be separated like this was - please return this blank line

widget/textgrid.go
```diff
@@ -310,6 +309,97 @@ func (t *TextGrid) refreshCell(row, col int) {
 	r.refreshCell(row, col)
 }
 
+// ForRangeFn calls back for each cell and row in the selection
+func (t *TextGrid) ForRangeFn(blockMode bool, startRow, startCol, endRow, endCol int, eachCell func(cell *TextGridCell), eachRow func(row *TextGridRow)) {
```
andydotxyz, line 313: regarding naming the `Fn` is not really needed, ForRange should be enough (as the function type is part of the signature). The documentation does not seem complete - `blockMode` is not explained.
Also the need for a row callback compared to cell is not clear - and the row callback is not called on first row - so it's more like a "eachNewRow" not just eachRow...?

You need to add the since like (blank comment line followed by a new comment // Since: 2.5" to mark this as a new API (same for the text getter added as well).

andydotxyz (changes requested):
Thanks for this, but we'll need clearer docs to support this new API.

andydotxyz:
Thinking more about this - is the range function needed as public API?
If it is to enable selection then the selection API is the public aspect and the text would be `SelectedText()` - so is this iteration required publicly?

mgazza:
implementing SelectedText() would require the tracking of the selected range and mode on the TextGrid. I can just put all this back to the term. That's how I had done it originally before trying to upstream here.

widget/textgrid.go
mgazza: Thank you, I will make these changes

andydotxyz:
> implementing SelectedText() would require the tracking of the selected range and mode on the TextGrid. I can just put all this back to the term. That's how I had done it originally before trying to upstream here.

If this is indeed related only to SelectText then add it wherever that code is to avoid extra APIs.
I had thought it was heading into the toolkit from what you'd shared. If it is to be only in the terminal then I don't see the benefit in putting it upstream here.

widget/textgrid.go
```diff
@@ -310,6 +309,135 @@ func (t *TextGrid) refreshCell(row, col int) {
...
+// ForRange iterates over a range of cells and rows within a TermGrid, optionally applying a function to each cell and row.
+//
+// Parameters:
+// - blockMode (bool): If true, the iteration is done in block mode, meaning it iterates through rows and applies the cell function for each cell in the specified column range.
+// - startRow (int): The starting row index for the iteration. Rows are 0-indexed.
+// - startCol (int): The starting column index for the iteration within the starting row. Columns are 0-indexed.
+// - endRow (int): The ending row index for the iteration.
+// - endCol (int): The ending column index for the iteration within the ending row.
+// - eachCell (func(cell *widget.TextGridCell)): A function that takes a pointer to a TextGridCell and is applied to each cell in the specified range. Pass `nil` if you don't want to apply a cell function.
+// - eachRow (func(row *widget.TextGridRow)): A function that takes a pointer to a TextGridRow and is applied to each row in the specified range. Pass `nil` if you don't want to apply a row function.
```
andydotxyz, line 321: This is misleading - it implies that it will be called for the first row, but it isn't

andydotxyz:
> If this is indeed related only to SelectText then add it wherever that code is to avoid extra APIs.
> I had thought it was heading into the toolkit from what you'd shared. If it is to be only in the terminal then I don't see the benefit in putting it upstream here.

I mention this just because it looks like it does not actually need any internal access to perform the functions - feel free to point out if I am wrong about that as it would mean a requirement to land in the main project.

mgazza:
OK happy to close this.

PR #4149, opened 2023-08-10
Update the padding of text under a form item

### LINKED ISSUE #4137: form hint text has confusing padding to next widget
### Checklist

### Describe the bug

the `HintText` and the validation message has more padding to the widget it belongs to, rather than to the next widget. I think this is visually confusing, since the hint is meant for the preceding widget.

a fix would be to reduce the top padding and/or increase the bottom padding of hints/validation messages.

### How to reproduce

open fyne demo

### Screenshots

Screenshot 2023-08-05 at 18 34 37

### Example code

fyne demo

### Fyne version

2.3.5

### Go compiler version

1.20.6

macOS 13.5

### Additional Information

_No response_

andydotxyz:
Fixes #4137

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/form.go
```diff
@@ -371,3 +370,21 @@ func NewForm(items ...*FormItem) *Form {
...
+
+	objs[1].Move(fyne.NewPos(theme.InnerPadding(), itemHeight+theme.InnerPadding()/2))
+	objs[1].Resize(fyne.NewSize(size.Width, objs[1].MinSize().Width))
+}
+
+func (f formItemLayout) MinSize(objs []fyne.CanvasObject) fyne.Size {
+	min0 := objs[0].MinSize()
+	min1 := objs[1].MinSize()
+
+	minWidth := fyne.Max(min0.Width, min1.Width)
```
Jacalz, line 388: Should this not be called "maxWidth"?

widget/form.go
```diff
@@ -371,3 +370,21 @@ func NewForm(items ...*FormItem) *Form {
...
+	return fyne.NewSize(minWidth, min0.Height+min1.Height).AddWidthHeight(0, theme.InnerPadding())
```
Jacalz, line 389: No need for the extra add here. You can move it into the size creation.
```suggestion
	return fyne.NewSize(minWidth, min0.Height+min1.Height+theme.InnerPadding())
```

Jacalz (changes requested):
I pushed a commit to fix the tests. Just one inline comment about a variable name and a minor cleanup while you're at it.

widget/form.go
andydotxyz: No, the max of to minimums is still a minimum

andydotxyz:
> I pushed a commit to fix the tests. Just one inline comment about a variable name and a minor cleanup while you're at it.

Thanks - between replying and accepting the simplification I think we're good?

Jacalz (approved):
Sorry, I thought I had approved

pseidemann:
thanks 👍

PR #4151, opened 2023-08-10
Update deps for v2.4.0 release

Jacalz:
### Description:

We have gotten a lot of PRs from Chinese security researchers trying to update packages for security issues that do not affect us and we have in general been behind on versions due to previously having Go 1.14 as minimum version. Update dependencies to be on top of it before releasing v2.4.0.

I have not updated typesetting, packages related to wasm support or urfave/cli (uses generics and Go 1.18). For the first two, there are probably I don't have enough knowledge to know if everything is ready to be updated.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


Jacalz:
Had to revert the change to the new fyne-io/image package. The license check is still complaining. I am honestly considering just moving gobmp into the fyne-io/image project as the upstream developer was a pain working with and had no interest in merging my PR. He commited is own solution without even trying to say what was wrong with my PR like it was some kind of issue and not a PR I had opened.

andydotxyz (approved):
Good idea

dweymouth:
Do you need to `go mod vendor` in this PR or do we do that as a follow-up?

Jacalz:
> Do you need to `go mod vendor` in this PR or do we do that as a follow-up?

Nope. The vendor folder has been removed on `develop`. GitHub takes PR templates from `master`, that's why it is still mentioned there.

PR #4162, opened 2023-08-15
Add note to stacklayout about single object

Jacalz:
### Description:

This adds a note to the documentation for stacklayout (and the corresponding container) that a single object has no effect on that object. 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container/layouts.go
+++ b/container/layouts.go
@@ -102,7 +102,8 @@ func NewPadded(objects ...fyne.CanvasObject) *fyne.Container {
 
 // NewStack returns a new container that stacks objects on top of each other.
 // Objects at the end of the container will be stacked on top of objects before.
-// A stack with a single object does not change the size or position of that object.
+// Having only a single object has no impact as CanvasObjects will
+// fill the available space even without a Stack.
 //
 // Since: 2.4
 func NewStack(objects ...fyne.CanvasObject) *fyne.Container {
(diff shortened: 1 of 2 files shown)


container/layouts.go
```diff
@@ -102,6 +102,7 @@ func NewPadded(objects ...fyne.CanvasObject) *fyne.Container {
 
 // NewStack returns a new container that stacks objects on top of each other.
 // Objects at the end of the container will be stacked on top of objects before.
+// A stack with a single object does not change the size or position of that object.
```
andydotxyz, line 105: This doesn't seem right - it will resize it after all. It's more accurate to say that it doesn't provide any advantage as that is the default behaviour?

andydotxyz (changes requested):
Comment doesn't seem quite right

container/layouts.go
Jacalz: Yeah, you're right. I meant that it wouldn't change the size. I didn't know how to word that in a good way. Will have to think a bit more about it.

dweymouth: Why not just "A stack with a single object is a no-op" similar to the message you added to the layout itself?

Jacalz: I wanted something more descriptive for the more commonly used container but maybe that was just a simple case of overthinking :)

andydotxyz: Maybe along the lines of "having only a single item has no impact as widgets will fill available space without Stack"?

Jacalz: Sounds perfect. Thanks :)

andydotxyz (approved):
Nice

PR #4167, opened 2023-08-18
Ensure window is rendered when re-used

### LINKED ISSUE #4163: Window doesn't render when called for the first time from system tray and the last window was closed
### Checklist

### Describe the bug

I have an app set up with a system tray and splash screen, where the main window is hidden on startup until you open it from the systray. What I found is that if I don't render the window before the splash screen closes the window won't render. On windows by resizing it renders, on ubuntu it doesn't even do that.

I did find that using `Hide` on the splash screen instead of `Close` avoids the issue, so it's something about the last visibly window calling `Close` I guess? Couldn't find anything I could understand in the sources.

### How to reproduce

build an app with a splash screen, no other window showing on startup, and a systray to open a window.

### Screenshots

I'm not on my windows machine now so I'll grab that later but here's the linux:
image

### Example code

```go
func main() {
	a := app.New()
	label := widget.NewLabel("Window with content!")
	win := a.NewWindow("Render Test")
	win.SetContent(label)
	showWin := fyne.NewMenuItem("Show win", func() {
		win.Show()
	})
	menu := fyne.NewMenu("tray", showWin)

		desk.SetSystemTrayMenu(menu)
	}
	if drv, ok := a.Driver().(desktop.Driver); ok {
		splash := drv.CreateSplashWindow()
		splash.SetContent(widget.NewLabel("Splash with content!"))
		splash.Show()
		go func() {
			time.Sleep(time.Second * 3)
			splash.Close()
		}()
	}
	win.SetCloseIntercept(func() {
		win.Hide()
	})
	a.Run()
}
```

### Fyne version

both latest (2.3.5) and develop (v2.3.6-0.20230815202934-6f2c727d711a)

### Go compiler version

1.20.5

windows 11, 10, kubuntu 5.27.4

### Additional Information

_No response_

andydotxyz:
Fixes #4163

### Checklist:

- [ ] Tests included. <- GL runtime :(
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -130,6 +130,10 @@ func (w *window) Show() {
 
 func (w *window) doShow() {
 	if w.view() != nil {
+		runOnDraw(w, func() {
+			w.driver.repaintWindow(w)
+		})
+
 		w.doShowAgain()
 		return
 	}


Jacalz (approved):

internal/driver/glfw/window.go
```diff
@@ -130,6 +130,10 @@ func (w *window) Show() {
 
 func (w *window) doShow() {
 	if w.view() != nil {
+		runOnDraw(w, func() {
+			w.driver.repaintWindow(w)
+		})
```
Jacalz, line 135: Actually, I had another look and wonder if it perhaps would be clearer to put this inside `doShowAgain()` instead?

Jacalz (approved):

PR #4170, opened 2023-08-20
Allow the compiler to optimize out platform specific dead code

Jacalz:
### Description:

Using constants for `drawOnMainThread` and `isWayland` allows the prove pass in the compiler to remove dead code. This means that any specific code for Wayland (mostly future work there) or M1/M2 doesn't get compiled in on other platforms. This improves binary sizes and improves performance by avoiding branching in some cases.

FYI: I verified on https://go.godbolt.org using a similar code structure that this indeed makes it now optimize out more dead code. 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/driver.go
+++ b/internal/driver/glfw/driver.go
@@ -27,15 +27,13 @@ import (
 // influence of a garbage collector.
 var mainGoroutineID uint64
 
-var (
-	curWindow *window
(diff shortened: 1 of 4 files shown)


andydotxyz (approved):
cool

PR #4171, opened 2023-08-21
Fix regression causing flickery windows when hiding

Jacalz:
### Description:

This reverts commit 5bac179573ed3b525dbd68ff1c991f0b3c18fa9b.

Reason for revert: This code runs when hiding and closing windows as well. Always doing a show results in flickering when trying to hide a second window by pressing a button inside that window. I don't know of a better solution so I just revert the PR for now to avoid the bug.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -801,10 +801,6 @@ func (w *window) processCharInput(char rune) {
 
 func (w *window) processFocused(focus bool) {
 	if focus {
-		// On Mac OS, hidden windows can be re-shown from the dock icon, which issues
-		// a GLFW window focus callback. We need to make sure the run loop is re-activated.
-		go w.doShow()
-
 		if curWindow == nil {
 			fyne.CurrentApp().Lifecycle().(*app.Lifecycle).TriggerEnteredForeground()
 		}


andydotxyz (changes requested):
It seems that the issue of flickering is Linux (or at least not Darwin) so perhaps this could be put inside a GOOS check instead of being reverted?

Jacalz:
Would you mind just testing this with the following code first? It should allow you to replicate it easily.
```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("First")

	w2 := a.NewWindow("Second")
	w2.SetContent(widget.NewButton("Hide", w2.Hide))
	w2.Show()

	w.Resize(fyne.NewSquareSize(50))
	w.ShowAndRun()
}
```

andydotxyz:
OMG yes that replicates on macOS too. With both windows visible.
But the first one shouldn't be...
Reverting fixes. That is the way - we need a different fix to this bug!

andydotxyz (approved):

Issue #2506, opened 2021-09-27
An idle application with no visible windows still uses noticeable CPU time


nullst:
### Describe the bug:
I am writing a desktop GUI application that would be usually running in the background, with minimized window that is opened from time to time. I've discovered that any fyne application, including `fyne_demo` or even a simple window with a single label in it, puts some consistent CPU load even if the application window is not visible and no work is being done. It's not a large amount of CPU usage, but over time it adds up to a battery drain, making even the most basic fyne application the 8th largest energy consumer on my laptop if left running for 20 hours (assuming that a large portion of those 20 hours are spent in the web browser). That's more demanding than, e.g., Telegram chat client even though that program is actually being used actively for some portion of those 20 hours, not just stalling in the background.

On my computer, the background load (of any fyne application that does no work in the background) is 3-4% CPU usage, as reported by "Activity Monitor" of my OS. This can be confirmed with standard golang CPU profiling: it reports 1.27s of CPU time during 30.03s of running in the background:

[![Profiling idle fyne application](https://i.ibb.co/rpC9qDT/default-settings.png)](https://svgshare.com/i/a_f.svg)

As one can see from this not very informative graphic, a small amount of work (~0.17s out of 1.27s CPU usage) is being done in `glfw.PollEvents` as invoked in `runGL`, but the majority of time is spent by golang's scheduling mechanism (`runtime.schedule` and its friends). I don't know what exactly is this scheduling mechanism working on. It is my impression that channel communication, for example, performing`select` over several channels, may be part of this.

### Ticker tuning helps, but not completely

Fyne's `runGL` from `internal/driver/glfw/loop.go` is an event loop which, in particular, polls GLFW events 60 times a second, triggered by the messages on a channel created by `time.Ticker`. This frequency is constant, regardless of whether the application window is visible or not. Even though profiling does not indicate that `runGL` is a significant CPU consumer, it is possible that some part of `runtime.schedule` CPU usage is the overhead for the `select` statement reading from several channels in `runGL` loop, since that `select` is performed 60 times per second. This is in fact the case.

As an experiment, I reduced the ticker frequency (by editing `loop.go`) to a rather extreme "1 event per second". This reduced background CPU load by a factor of two. Only 0.63s of CPU time were used during 30s of running in the background:

[![Ticker tuning](https://i.ibb.co/D7FK4g4/ticker-tuning.png)](https://svgshare.com/i/aZi.svg)

[...]

PR #4172, opened 2023-08-21
Clean up some driver code

Jacalz:
### Description:

This backports some changes that I did while trying to solve https://github.com/fyne-io/fyne/issues/2506. It improves the performance slightly be allowing NewGLDriver() to be inlined (now cost 79, with 80 as limit) and removes an unnecessary field from `runFlag`. Nothing major just makes the code a bit cleaner.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Cleaner, thanks

PR #4173, opened 2023-08-21
Only process events in run loop when they happen

Jacalz:
### Description:

This PR changes the run loop to use `glfw.WaitEvents()` instead and use `glfw.PostEmptyEvent()` when necessary to only ever process events when there are any to process. 
NOTE: This currently includes the code from #4172 as well. 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
It feels really late in the process to be changing the runloop for 2.3.0 - there are *so* many platform idiosyncrasies that we can be pretty sure this will trigger something.
Maybe to land after release for following cycle?

dweymouth:
I would like to help get this in for 2.4.2 if possible, but I feel like I'll need a deeper understanding of how the run loop is designed - is there any design documentation somewhere? I'll try to review based on what I know though

internal/driver/glfw/loop.go
```diff
@@ -112,74 +114,81 @@ func (d *gLDriver) runGL() {
...
-	for {
-		select {
-		case <-d.done:
-			eventTick.Stop()
-			d.drawDone <- nil // wait for draw thread to stop
+
+	// Post empty event to not get stuck in the loop on first iteration.
+	PostEmptyEvent()
+
+	for range eventTick.C {
```
dweymouth, line 123: Do we even need `eventTick` anymore? Can this just be a "forever" loop now that we are blocking with `d.waitForEvents()`?

internal/driver/glfw/loop.go
```diff
@@ -112,74 +114,81 @@ func (d *gLDriver) runGL() {
...
 				}
+			}
 
-				newWindows = append(newWindows, win)
+			newWindows = append(newWindows, win)
 
-				if drawOnMainThread {
-					d.drawSingleFrame()
-				}
+			if drawOnMainThread {
```
dweymouth, line 181: Now that we are waiting for GLFW events this won't run 60 times a second anymore so I think we need a separate way to handle drawing on the main thread.

I'm kind of thinking out loud here but maybe the whole main loop should be a big select statement from multiple channels? 1) available GLFW events. 2) a channel signalling when a refresh is needed (because user code called .Refresh(), 3) a 60 fps animation ticker that only runs if a) we are drawing on the main thread, and b) an animation is currently in progress. So if none of these channels have available input the whole main loop can be entirely asleep in the select until something needs to be done

Jacalz:
Sorry for the late reply @dweymouth. I'll try to have a look at this next week and see about answering your questions.

andydotxyz:
Just FYI a change like this has the potential to break some OS combinations so putting it in a point release makes me nervous. 
Of course they could be overcome by lots of testing in the field I guess, but typically such a change is landed months before a release and v2.4.2 is already overdue so maybe not too realistic?

Jacalz:
I agree. Landing in 2.4.x is probably not a good plan

internal/driver/glfw/loop.go
Jacalz: I don't think a big select statement would work? I have no idea how to get the blocking `d.waitForEvents()` to send on a channel instead? We'd have to be blocked on that function (which has to run on the main thread) which makes it so that we can't get other items that are sent on the channels. That it problematic.

internal/driver/glfw/loop.go
Jacalz: My idea for keeping it was if we got more inputs that 60 times per second but you might be right.

Jacalz:
I felt that it was easier to start over with #4399. Let's continue the discussion there.

PR #4175, opened 2023-08-21
Move away from struct creation using new() when we have access to fields

Jacalz:
### Description:

It seems like we (and many other projects but that's just speculation) are moving away from using new() when creating structs (using new(int) still is useful). This PR changes a few places where it seemed to make sense to not use new().

My opinion is that it is a lot clearer to create the struct directly, without using new(), and we also get direct field access which can decrease inlining costs and can sometimes allow the compiler to optimize things a bit more (but I am doing this for readability more than anything else in this case). 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -407,10 +407,9 @@ func (w *window) processMouseMoved(xpos float64, ypos float64) {
 	isMouseOverDragged := w.objIsDragged(mouseOver)
 	w.mouseLock.RUnlock()
 	if obj != nil && !isObjDragged {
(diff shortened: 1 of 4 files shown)


andydotxyz (approved):
This makes sense. Do you want to have it in the style guide as well?

Jacalz:
Sounds like a good idea. Apart from a few tests, a few new(int) (and similar) in data binding, and code we inherited from gomobile we don't really use new() anywhere.

Jacalz:
I added it to the style guide: https://github.com/fyne-io/fyne/wiki/Contributing

PR #4177, opened 2023-08-22
Cleanup GLFW init to avoid needing a sync.Once

Jacalz:
### Description:

We previously called `glfw.Init()` before doing anything that needed it to be called (i.e. when creating a window and when starting the run loop) which led to init running multiple times. The solution was to protect it using a `sync.Once`  but I realized that  we can avoid it by just doing init on driver creation instead. This cleans up the code and makes window creation slightly faster as an added benefit.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
I noticed that I accidentally pushed two unrelated commits here. Will move those out to a separate PR.

andydotxyz (approved):

PR #4184, opened 2023-08-23
Don't allocate a slice of windows 60 times per second

Jacalz:
### Description:

We were allocating and appending to a slice 60 times per second to modify the window list if windows were supposed to close. This updates the runloop to keep track of how many windows need to be closed and only allocate a slice when we need to.

With this, we should be doing less each iteration of the runloop (especially for applications with many windows) but a little more when removing windows which I think is a worthy tradeoff given closing is a lot less common. I also tuned the allocation of the closing windows slice to only allocate as much as it needs to. However, I am having trouble getting accurate readings in regards to how much effect it has. The numbers are just all over the place and I don't think my system monitor on Linux is accurate enough for CPU usage. I would appreciate if someone could see if they notice a decrease in CPU usage. @dweymouth perhaps?

Updates #2506

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -279,7 +279,7 @@ func (w *window) destroy(d *gLDriver) {
 	w.DestroyEventQueue()
 	cache.CleanCanvas(w.canvas)
 
-	if w.master || len(w.driver.windowList()) == 1 {
+	if w.master {
 		d.Quit()
 	} else if runtime.GOOS == "darwin" {
 		go d.focusPreviousWindow()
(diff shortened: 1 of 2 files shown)


internal/driver/glfw/window.go
```diff
@@ -279,7 +279,7 @@ func (w *window) destroy(d *gLDriver) {
 	w.DestroyEventQueue()
 	cache.CleanCanvas(w.canvas)
 
-	if w.master {
+	if w.master || len(w.driver.windowList()) == 1 {
```
andydotxyz, line 282: == 1 is only exit criteria on a desktop app with system tray turned on isn't it?

Jacalz: Hmm. I though it would be no different than the check that quit if there were 0 windows. Will have to look into that.

Jacalz: I moved it back to where it was

andydotxyz (approved):
yeah nice improvement thanks

PR #4186, opened 2023-08-23
Fix progressbar text having the wrong color sometimes

Jacalz:
### Description:

I noticed in Rymdport that the text color flickered from white to black after being rendered for the first time. There have also been reports of the text having a different color sometimes which I assume relates to the fact that we never refreshed the label.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Oops!

PR #4196, opened 2023-08-25
Lock and unlock less for BaseWidget.Move

Jacalz:
### Description:

Instead of first locking to set size and then locking again to get the implementation we can do it while we hold the first lock. This also matches the behaviour in all other methods.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/widget/base.go
+++ b/internal/widget/base.go
@@ -72,9 +72,10 @@ func (w *Base) Position() fyne.Position {
 func (w *Base) Move(pos fyne.Position) {
 	w.propertyLock.Lock()
 	w.position = pos
+	impl := w.impl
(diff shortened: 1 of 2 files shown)


Jacalz:
TODO: Benchmark the effect of this and apply the same to internal/widget if useful.

andydotxyz (dismissed):
thanks

Jacalz:
Benchmark results: This seems to bring down the time to call `Move()` (including creation of position) from 68ns to 55ns.

Jacalz:
Sorry, needs a new approval. I added the same change to the internal basewidget.

andydotxyz (approved):

PR #4197, opened 2023-08-25
Avoid unnecessary atomic app loads

Jacalz:
### Description:

I'm not a huge fan of this repaint function (that inline interface check looks strange) but we can at least change three atomic app lookups to one.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
TODO: FIx up https://github.com/fyne-io/fyne/blob/1834d704a305f4729cce65213f5b5c3098e1d993/internal/widget/base.go#L165 as well. Maybe move into a common place?

andydotxyz (approved):

andydotxyz:
> Maybe move into a common place?

Just read the notes about dependency cycle ;)

PR #4206, opened 2023-08-30
Fix issue where fullscreen input would cover app in landscape keyboard mode

andydotxyz:
### Checklist:

- [ ] Tests included. <- only happened on Samsung hardware
- [x] Tests all pass.


andydotxyz:
Landing automatically as no contrary comments

PR #4207, opened 2023-08-31
Close an open branch with the left-key on a tree

Roemer:
### Description:
This change allows to close a branch in a tree with the left-key but keeps the focus on it. With an additional left-key, it will (as before) go up to the next parent node.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/tree.go
+++ b/widget/tree.go
@@ -348,11 +348,17 @@ func (t *Tree) TypedKey(event *fyne.KeyEvent) {
 		t.ScrollTo(t.currentFocus)
 		t.RefreshItem(t.currentFocus)
 	case fyne.KeyLeft:
-		t.walk(t.Root, "", 0, func(id, p TreeNodeID, _ bool, _ int) {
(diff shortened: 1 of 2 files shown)


andydotxyz:
> I did not add tests as I think we cannot test keyboard input right now.

You should probably check out `TestTree_Focus` in `tree_internal_test.go` where the existing key handling is tested.

Roemer:
> > I did not add tests as I think we cannot test keyboard input right now.
> 
> You should probably check out `TestTree_Focus` in `tree_internal_test.go` where the existing key handling is tested.

Thanks, I just added a test for keyboard navigation.

Roemer:
Something failed that I did not touch. What to do?

andydotxyz:
There is a flakey test we need to fix, have asked it to re-run

Roemer:
All green now.

andydotxyz (approved):

PR #4214, opened 2023-09-03
expose entry attribe Scroll value and support one entrance to set it

sunvim:
### Description:

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -132,6 +132,29 @@ func NewPasswordEntry() *Entry {
 	return e
 }
 
+// ScrollDirection represents the directions in which a Scroll can scroll its child content.
+type ScrollDirection widget.ScrollDirection
+
+// Constants for valid values of ScrollDirection.
+const (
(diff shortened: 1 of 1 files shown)


andydotxyz:
These types are already exposed in the container package.

Jacalz:
Please also note that we, according to the PR template and contributor documentation, want PRs to target the `develop` branch and not `master`.

Jacalz:
If anything, the scrolling constants should be in the `fyne` package. I also don’t quite think we need a setter when we already have access to the field easily.

sunvim:
> These types are already exposed in the container package.

yeah , I see!

sunvim:
> If anything, the scrolling constants should be in the `fyne` package. I also don’t quite think we need a setter when we already have access to the field easily.

okkk ,got it

PR #4215, opened 2023-09-03
Stop using deprecated theme, improve docs

### LINKED ISSUE #4208: LightTheme and DarkTheme are being deprecated without any replacement
### Checklist

### Describe the bug

From the code:

// LightTheme defines the built-in light theme colors and sizes.
//
// Deprecated: This method ignores user preference and should not be used, it will be removed in v3.0.

However, I can't see any obvious replacement which would allow me to change the theme variant for the user.  I have a settings menu where they can select Light or Dark, just as the fyne demo has.  Fyne demo even still uses the deprecated function.  I believe the comment should be updated to suggest another way to set the theme variant, so users can keep the functionality once its removed. 

### How to reproduce

Use LightTheme(), once its removed, dunno how to change theme variant.

### Screenshots

_No response_

### Example code

From fyne demo

```go
	themes := container.NewGridWithColumns(2,
		widget.NewButton("Dark", func() {
			a.Settings().SetTheme(theme.DarkTheme())
		}),
		widget.NewButton("Light", func() {
			a.Settings().SetTheme(theme.LightTheme())
		}),
	)
```

### Fyne version

latest

### Go compiler version

1.19.1

Windows 10

### Additional Information

_No response_

andydotxyz:
Fixes #4208

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne_demo/theme.go
+++ b/cmd/fyne_demo/theme.go
@@ -1,0 +1,17 @@
+package main
+
+import (
+	"image/color"
+
+	"fyne.io/fyne/v2"
+)
+
+type forceVariantTheme struct {
+	fyne.Theme
+
+	variant fyne.ThemeVariant
+}
+
+func (f *forceVariantTheme) Color(name fyne.ThemeColorName, _ fyne.ThemeVariant) color.Color {
+	return f.Theme.Color(name, f.variant)
+}
--- a/cmd/fyne_demo/theme.go
+++ b/cmd/fyne_demo/theme.go
@@ -6,12 +6,12 @@ import (
 	"fyne.io/fyne/v2"
 )
 
-type forceVariantTheme struct {
+type forcedVariant struct {
 	fyne.Theme
 
 	variant fyne.ThemeVariant
 }
 
-func (f *forceVariantTheme) Color(name fyne.ThemeColorName, _ fyne.ThemeVariant) color.Color {
+func (f *forcedVariant) Color(name fyne.ThemeColorName, _ fyne.ThemeVariant) color.Color {
 	return f.Theme.Color(name, f.variant)
 }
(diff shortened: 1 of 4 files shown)


Jacalz:
The deprecated functions are used a lot throughout the codebase in tests (remove `staticcheck.conf` temporarily and run staticcheck in the project root). Should we add some kind of wrapper for tests as well?

andydotxyz:
I think maybe those are old tests from before the test theme was added - should they not use that instead of the older light/dark APIs?

andydotxyz:
The coverage report seems to be comparing this to master, and it's the previously landed commits that have coverage drop... rather strange!

Jacalz:
> The coverage report seems to be comparing this to master, and it's the previously landed commits that have coverage drop... rather strange!

Strange. Have you created the branch from master?

Jacalz:
> I think maybe those are old tests from before the test theme was added - should they not use that instead of the older light/dark APIs?

Maybe, I don't know? I think some of them are from after that theme was added. I have a vague recollection of some of them using both light and dark to make sure that the widgets look as expected in both themes but I might be wrong.

cmd/fyne_demo/theme.go
```diff
@@ -0,0 +1,17 @@
+package main
+
+import (
+	"image/color"
+
+	"fyne.io/fyne/v2"
+)
+
+type forceVariantTheme struct {
```
Jacalz, line 9: Would this not be better named as `forcedThemeVariant`?

Jacalz (commented):
Sorry for the late review :/

andydotxyz:
> > I think maybe those are old tests from before the test theme was added - should they not use that instead of the older light/dark APIs?
> 
> Maybe, I don't know? I think some of them are from after that theme was added. I have a vague recollection of some of them using both light and dark to make sure that the widgets look as expected in both themes but I might be wrong.

Many of the instances just need a theme, we cannot use 'theme.DefaultTheme' as it will be different based on user preferences. So a more advanced test handler that provides two theme variants programmatically will be required. If OK with you this could be a different PR.

Jacalz (approved):

PR #4219, opened 2023-09-05
Usage of deprecated commands would crash due to nil AppData

andydotxyz:
This is an issue also for usage of the not-deprecated `Getter`

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/release.go
+++ b/cmd/fyne/internal/commands/release.go
@@ -28,8 +28,7 @@ var macAppStoreCategories = []string{
 
 // Release returns the cli command for bundling release builds of fyne applications
 func Release() *cli.Command {
-	r := &Releaser{}
-	r.appData = &appData{}
+	r := NewReleaser()
 
 	return &cli.Command{
 		Name:  "release",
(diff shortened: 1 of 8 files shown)


cmd/fyne/internal/commands/release.go
```diff
@@ -149,6 +148,13 @@ type Releaser struct {
 	password     string
 }
 
+// NewReleaser returns a command that can handle the packaging a GUI apps for release from local Fyne source code.
+func NewReleaser() *Releaser {
+	r := &Releaser{}
+	r.appData = &appData{}
+	return r
```
Jacalz, line 155: Should we not do it like this to be consistent with the others?
```suggestion
	return &Releaser{appData: &appData{}}
```

Jacalz (changes requested):
Thanks. Just one minor change for consistency.

cmd/fyne/internal/commands/release.go
andydotxyz: No, that won't work - the appData is not a field of Releaser but inherited from a parent.

Jacalz: Ah, I see. Alright then. That is honestly one of the few things with Go that I strongly dislike

Jacalz (approved):

PR #4225, opened 2023-09-08
Update table.go for ease of dev usage/self documentation (retry)

Ju-B:
When using the widget, i always mix rows an columns in the lenght function.

This change allows IDE/LSPs to display a usefull signature for the function.


andydotxyz (approved):
Thanks, much clearer

PR #4232, opened 2023-09-10
Move to js GOOS flag from gopherjs to be consistent with tags

andydotxyz:
Let's not be beholden to gophers if we find a different path for js output...

If this is approved we can back-port to v2.4.1 before web is officially supported and we are locked in.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
Merged as per Slack discussion and no negatives after 1 week

PR #4240, opened 2023-09-12
Correctly align mobile menu content

andydotxyz:
Fix it so the mobile menu button aligns with the inside of the canvas, not the window edge

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz (approved):
Nice catch

PR #4244, opened 2023-09-13
Add Bold and Underline to TextGrid widget

mgazza:
### Description:

Enable the text style to be passed in as part of the TextGridStyle. this will allow font options to be passed down

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/text.go
+++ b/text.go
@@ -61,6 +61,9 @@ type TextStyle struct {
 	Symbol bool // Use the system symbol font.
 	// Since: 2.1
 	TabWidth int // Width of tabs in spaces
+	// Since: 2.5
+	// Currently only supported by the TextGrid widget.
+	Underline bool // Should text be underlined.
 }
 
 // MeasureText uses the current driver to calculate the size of text when rendered.
--- a/widget/textgrid.go
+++ b/widget/textgrid.go
@@ -43,12 +43,15 @@ type TextGridRow struct {
 
 // TextGridStyle defines a style that can be applied to a TextGrid cell.
 type TextGridStyle interface {
+	Style() fyne.TextStyle
 	TextColor() color.Color
 	BackgroundColor() color.Color
 }
 
 // CustomTextGridStyle is a utility type for those not wanting to define their own style types.
 type CustomTextGridStyle struct {
+	// Since: 2.5
+	TextStyle        fyne.TextStyle
 	FGColor, BGColor color.Color
 }
 
@@ -60,6 +63,11 @@ func (c *CustomTextGridStyle) TextColor() color.Color {
 // BackgroundColor is the color a cell should use for the background.
 func (c *CustomTextGridStyle) BackgroundColor() color.Color {
 	return c.BGColor
+}
+
+// Style is the text style a cell should use.
+func (c *CustomTextGridStyle) Style() fyne.TextStyle {
+	return c.TextStyle
 }
 
 // TextGrid is a monospaced grid of characters.
@@ -350,12 +358,15 @@ func (t *textGridRenderer) appendTextCell(str rune) {
 	text.TextStyle.Monospace = true
 
 	bg := canvas.NewRectangle(color.Transparent)
-	t.objects = append(t.objects, bg, text)
+
+	ul := canvas.NewLine(color.Transparent)
+
+	t.objects = append(t.objects, bg, text, ul)
 }
 
 func (t *textGridRenderer) refreshCell(row, col int) {
 	pos := row*t.cols + col
-	if pos*2+1 >= len(t.objects) {
+	if pos*3+1 >= len(t.objects) {
 		return
 	}
 
@@ -367,26 +378,59 @@ func (t *textGridRenderer) setCellRune(str rune, pos int, style, rowStyle TextGr
 	if str == 0 {
 		str = ' '
 	}
+	rect := t.objects[pos*3].(*canvas.Rectangle)
+	text := t.objects[pos*3+1].(*canvas.Text)
+	underline := t.objects[pos*3+2].(*canvas.Line)
 
 	th := t.text.Theme()
 	v := fyne.CurrentApp().Settings().ThemeVariant()
-
-	text := t.objects[pos*2+1].(*canvas.Text)
-	text.TextSize = th.Size(theme.SizeNameText)
 	fg := th.Color(theme.ColorNameForeground, v)
+	text.TextSize = th.Size(theme.SizeNameText)
+	textStyle := fyne.TextStyle{}
+	var underlineStrokeWidth float32 = 1
+	var underlineStrokeColor color.Color = color.Transparent
+
 	if style != nil && style.TextColor() != nil {
 		fg = style.TextColor()
 	} else if rowStyle != nil && rowStyle.TextColor() != nil {
 		fg = rowStyle.TextColor()
 	}
+
+	if style != nil {
+		if style.Style().Bold {
+			underlineStrokeWidth = 2
+			textStyle = fyne.TextStyle{
+				Bold: true,
+			}
+		}
+		if style.Style().Underline {
+			underlineStrokeColor = fg
+		}
+	} else if rowStyle != nil {
+		if rowStyle.Style().Bold {
+			underlineStrokeWidth = 2
+			textStyle = fyne.TextStyle{
+				Bold: true,
+			}
+		}
+		if rowStyle.Style().Underline {
+			underlineStrokeColor = fg
+		}
+	}
+
 	newStr := string(str)
-	if text.Text != newStr || text.Color != fg {
+	if text.Text != newStr || text.Color != fg || textStyle != text.TextStyle {
 		text.Text = newStr
 		text.Color = fg
+		text.TextStyle = textStyle
 		t.refresh(text)
 	}
 
-	rect := t.objects[pos*2].(*canvas.Rectangle)
+	if underlineStrokeWidth != underline.StrokeWidth || underlineStrokeColor != underline.StrokeColor {
+		underline.StrokeWidth, underline.StrokeColor = underlineStrokeWidth, underlineStrokeColor
+		t.refresh(underline)
+	}
+
 	bg := color.Color(color.Transparent)
 	if style != nil && style.BackgroundColor() != nil {
 		bg = style.BackgroundColor()
@@ -401,10 +445,10 @@ func (t *textGridRenderer) setCellRune(str rune, pos int, style, rowStyle TextGr
 
 func (t *textGridRenderer) addCellsIfRequired() {
 	cellCount := t.cols * t.rows
-	if len(t.objects) == cellCount*2 {
+	if len(t.objects) == cellCount*3 {
 		return
 	}
-	for i := len(t.objects); i < cellCount*2; i += 2 {
+	for i := len(t.objects); i < cellCount*3; i += 3 {
 		t.appendTextCell(' ')
 	}
 }
@@ -468,7 +512,7 @@ func (t *textGridRenderer) refreshGrid() {
 
 		line++
 	}
-	for ; x < len(t.objects)/2; x++ {
+	for ; x < len(t.objects)/3; x++ {
 		t.setCellRune(' ', x, TextGridStyleDefault, nil) // trailing cells and blank lines
 	}
 }
@@ -513,10 +557,17 @@ func (t *textGridRenderer) Layout(size fyne.Size) {
 	cellPos := fyne.NewPos(0, 0)
 	for y := 0; y < t.rows; y++ {
 		for x := 0; x < t.cols; x++ {
-			t.objects[i*2+1].Move(cellPos)
+			// rect
+			t.objects[i*3].Resize(t.cellSize)
+			t.objects[i*3].Move(cellPos)
 
-			t.objects[i*2].Resize(t.cellSize)
-			t.objects[i*2].Move(cellPos)
+			// text
+			t.objects[i*3+1].Move(cellPos)
+
+			// underline
+			t.objects[i*3+2].Move(cellPos.Add(fyne.Position{X: 0, Y: t.cellSize.Height}))
+			t.objects[i*3+2].Resize(fyne.Size{Width: t.cellSize.Width})
+
 			cellPos.X += t.cellSize.Width
 			i++
 		}
--- a/widget/textgrid_test.go
+++ b/widget/textgrid_test.go
@@ -27,7 +27,7 @@ func TestTextGrid_CreateRendererRows(t *testing.T) {
 	rend := test.WidgetRenderer(grid).(*textGridRenderer)
 	rend.Refresh()
 
-	assert.Equal(t, 12, len(rend.objects))
+	assert.Equal(t, 18, len(rend.objects))
 }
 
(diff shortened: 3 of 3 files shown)


widget/textgrid.go
```diff
@@ -47,9 +47,15 @@ type TextGridStyle interface {
 	BackgroundColor() color.Color
 }
 
+// TextGridTextStyle defines the text style that can be applied to a TextGrid cell.
+type TextGridTextStyle interface {
```
andydotxyz, line 51: Because the `TextGridTextStyle` must always be a `TextGridStyle` should it extend I to bake it into the API?

p.s. It's missing the `// Since: 2.5` comment

widget/textgrid.go
```diff
@@ -47,9 +47,15 @@ type TextGridStyle interface {
...
+	TextStyle() *fyne.TextStyle
+}
+
 // CustomTextGridStyle is a utility type for those not wanting to define their own style types.
 type CustomTextGridStyle struct {
 	FGColor, BGColor color.Color
+	Text             *fyne.TextStyle
```
andydotxyz, line 58: Wouldn't `Style` or `TextStyle` be more appropriate than `Text` for a style field?
Perhaps it should match naming elsewhere like on the `canvas.Text` or `widget.Label`?

p.s. It's missing the `// Since: 2.5` comment

widget/textgrid.go
```diff
@@ -62,6 +68,11 @@ func (c *CustomTextGridStyle) BackgroundColor() color.Color {
 	return c.BGColor
 }
 
+// TextStyle is the text style a cell should use for the text.
```
andydotxyz, line 71: This also needs a since line

andydotxyz (changes requested):
Looks good. A few comments inline about since notes and naming.

Would it be possible to add a test?

Also I think you might find that although the APIs are added they do not take effect on output...
You'll need to extend the code in `internal/painter/font.go` I think (line 35 onwards). It will likely be necessary to pass the style in to the actual theme `Font` call instead of using the helper methods which are limited to monospace vs all other styles...

widget/textgrid.go
mgazza: I've replaced this with two bool values

widget/textgrid.go
mgazza: added

mgazza:
quick demo
image

widget/textgrid.go
```diff
@@ -47,9 +47,19 @@ type TextGridStyle interface {
...
+	TextGridStyle
+	Bold() bool
+	Underlined() bool
+}
+
 // CustomTextGridStyle is a utility type for those not wanting to define their own style types.
 type CustomTextGridStyle struct {
-	FGColor, BGColor color.Color
+	FGColor, BGColor     color.Color
+	IsBold, IsUnderlined bool
```
andydotxyz, line 62: These fields will need a since line too, sorry

widget/textgrid.go
```diff
@@ -345,12 +369,15 @@ func (t *textGridRenderer) appendTextCell(str rune) {
 	text.TextStyle.Monospace = true
 
 	bg := canvas.NewRectangle(color.Transparent)
-	t.objects = append(t.objects, bg, text)
+
+	ul := canvas.NewLine(color.Transparent)
```
andydotxyz, line 373: Won't having 50% more objects always visible slow things down?
If adding them cannot be avoided at least avoid showing them... or maybe we need an optimisation in line drawing to ignore transparent lines? (I don't think that exists)...

widget/textgrid.go
```diff
@@ -47,9 +47,19 @@ type TextGridStyle interface {
 	BackgroundColor() color.Color
 }
 
+// TextGridTextStyle defines the text style that can be applied to a TextGrid cell.
+//
+// Since: 2.5
+type TextGridTextStyle interface {
+	TextGridStyle
+	Bold() bool
+	Underlined() bool
```
andydotxyz, line 56: Isn't it `Underline` not `Underlined` in most places?

andydotxyz (changes requested):

widget/textgrid.go
andydotxyz: Taking this path means that adding another value later will need another interface... Using TextStyle (which is a concrete struct) would make it possible to add more I think

andydotxyz:
> You'll need to extend the code in internal/painter/font.go I think (line 35 onwards). It will likely be necessary to pass the style in to the actual theme Font call instead of using the helper methods which are limited to monospace vs all other styles...

I'm curious about how you got the demo working with that font cache not patched - did you find that it wasn't a problem for some reason? How was the bold monospace font looked up in that demo?

widget/textgrid.go
mgazza: Text Style would need the underline property adding.

widget/textgrid.go
mgazza: It doesn't appear any slower to me. I do wonder if we should have some benchmarks around this, however.

widget/textgrid.go
mgazza: Underline only appears in the processing logic to refer to the actual line.

mgazza:
> > You'll need to extend the code in internal/painter/font.go I think (line 35 onwards). It will likely be necessary to pass the style in to the actual theme Font call instead of using the helper methods which are limited to monospace vs all other styles...
> 
> I'm curious about how you got the demo working with that font cache not patched - did you find that it wasn't a problem for some reason? How was the bold monospace font looked up in that demo?

overriding the theme was all that was needed
image

widget/textgrid.go
andydotxyz: Indeed. But do you think underline is only useful for TextGrid? Or would we perhaps consider having it broadly available under a simple style value?

andydotxyz:
Yup I appreciate that the API was designed to make this possible - but I'm pretty sure you'll get cache collisions inside the toolkit as it does not understand bold monospace.

Happy to be proved wrong on this of course, but it's a concern.

widget/textgrid.go
mgazza: No, it could/would make sense. Just making you aware of the implications. I'm finding it very dificult to determine the scope of changes to fyne and associated projects as in some cases you suggest less invasive and some other cases more invasive. The implications of adding this to TextStyle would mean it would remain unimplemented by anything that uses it and other options like italics which will not be implemented within the TextGrid/Renderer.

mgazza:
> Yup I appreciate that the API was designed to make this possible - but I'm pretty sure you'll get cache collisions inside the toolkit as it does not understand bold monospace.
> 
> Happy to be proved wrong on this of course, but it's a concern.

I haven't seen any, it sounds like an ideal unit test candidate, however. Is there a table test which I can add a test case to?

widget/textgrid.go
andydotxyz: even just running top should push the terminal and display CPU usage at the same time

widget/textgrid.go
andydotxyz: Sorry, I was meaning in most typographic APIs is it not referred to as underline?

I.e. it is not "bolded" or "italicised" either...

andydotxyz:
> I haven't seen any, it sounds like an ideal unit test candidate, however. Is there a table test which I can add a test case to?

The problem will be in `CachedFontFace` of `internal/painter/font.go` and there is a `font_internal_test.go` used to test caching and other non-public API.

mgazza:
When running the tests locally, I get quite a few failures. I have checked to ensure no more/other tests are failing due to this work. I can't see how `CachedFontFace` is affected, it appears to call the theme font in every case from what I can see. I may be missing something however.

andydotxyz:
> When running the tests locally, I get quite a few failures. I have checked to ensure no more/other tests are failing due to this work. I can't see how `CachedFontFace` is affected, it appears to call the theme font in every case from what I can see. I may be missing something however.

I can't figure out how! The OpenGL driver text rendering uses the cached face to draw. it is created in `internall/painter/gl/texture.go:155`, and the text style passed in uses only the Monospace field in the case of mono, ignoring the others. I cannot find any other routes to get to `DrawString` which is the entry point for rendering text...

andydotxyz:
I just realised that `TextGridTextStyle` does not handle Italics... so would that be another new interface in the future? It's looking more and more like a duplication of the `TextStyle` struct!

andydotxyz:
> When running the tests locally, I get quite a few failures.

There should not be - there is an intermittent issue due to window sizing and OS not always allowing it during the test IIRC. If the CI does not see failures then it's probably OK but worth opening a bug to report that it's not all working at your end.

andydotxyz:
> When running the tests locally, I get quite a few failures.

Latest develop should have all the test failures resolved.
Shall we try and get something based on this in to 2.5? Would be good to have the `TextStyle` used on more widgets.

mgazza:
Sorry for the delay, yeah it would be good to have this into 2.5

widget/textgrid.go
mgazza: These methods act as queries, determining whether a specific style is currently applied. "IsUnderline" wouldn't be appropriate because it suggests an action to check for underlining, rather than reflecting the current state of the text.
if Bold() accurately reflects the presence of bold formatting, then "Underlined()" follows the same logic. It's not "IsBolded" because the text either is bolded or not.
While "underline" might be more common in typography, "Underlined()" aligns with the existing naming convention and emphasizes the query nature of these methods.  This prioritizes readability and consistency within the codebase.

andydotxyz:
That might be tough, yesterday was feature freeze. I remain concerned that adding "TextGridTextStyle" in addition to "TextGridStyle" and "TextStyle" is confusing and the discussion about bold/bolded underline/underlined would be avoided by using the existing APIs

widget/textgrid.go
andydotxyz: I don't see how this is consistent, "Bold" and "Italic" used elsewhere refer to state. Their "is set" equivalent in getters would be "IsItalic" or "Italicised".
Following this "Underline" is the state version and underlined refers to an action that occurred.

As said in other places though, I'm still not sure why it isn't a better plan to use existing "TextStyle" API rather than a specific style API for this widget.

drunlade: This is a strange discussion but I'm going to weigh in anyway.

```
You would say:      "Is it bold?"
You would not say:  "Is it bolded?"

You would say:      "Is it italic?"
You would not say:  "Is it italicised?"

You would say:      "Is it underlined?"
You would not say:  "Is it underline?"
```

```go
if text.Bold() { log.Debug("Text is bold") }

if text.Italic() { log.Debug("Text is italic") }

if text.Underlined() { log.Debug("Text is underlined") }
```

Put another way:

"Italic" is a adjective/noun, it describes what something is, its a name or state of something.
"Underline" (in how it's used here) is a verb, it's an action, it is not a description or name and doesn't represent state.

If we used your version the code would be:

```go
if text.Underline() { log.Debug("Text is underline") }
```

Which reads and sounds (especially if you say it out loud) very wrong.

drunlade: > As said in other places though, I'm still not sure why it isn't a better plan to use existing "TextStyle" API rather than a specific style API for this widget.

This would make sense, having different versions of effectively the same thing isn't great, assuming there's no gotchas lurking under the code for why this wouldn't work.

mgazza: To weigh in a bit more, I think the confusion is coming from the fact this is a method on a struct and not an enum like more traditional API's.
Since this is a method it feels more subject to OO programming naming, like we have suggested. If this was an enum then I suspect this whole thing would go away.
some other GUI API's for reference. 
https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-createfonta?redirectedfrom=MSDN
https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/javax/swing/text/StyleConstants.CharacterConstants.html

mgazza: >>As said in other places though, I'm still not sure why it isn't a better plan to use existing "TextStyle" API rather than a specific style API for this widget.

>This would make sense, having different versions of effectively the same thing isn't great, assuming there's no gotchas lurking under the code for why this wouldn't work.

The only gotchas would be that I'm only implementing this for this single widget, all other widgets can be updated subsequently by the parties wanting that functionality.

drunlade: > > This would make sense, having different versions of effectively the same thing isn't great, assuming there's no gotchas lurking under the code for why this wouldn't work.
> 
> The only gotchas would be that I'm only implementing this for this single widget, all other widgets can be updated subsequently by the parties wanting that functionality.

I think it's reasonable to use the existing TextStyle then, add the new features, support them in the widget you need it for and anyone who wants to add support to the other widgets can if/when it's needed.

andydotxyz: Ah, difference between field and method yes this is causing confusion.

> The only gotchas would be that I'm only implementing this for this single widget, all other widgets can be updated subsequently by the parties wanting that functionality.

We should not add temporary APIs or items that are widget specific but will be broadened. Any public API will remain part of the toolkit indefinitely. Given there are feature requests open for underline it seems pretty reasonable to think it will be added in the meantime we could add some documentation about it not being supported in all locations until that is no longer true.

mgazza:
I hope the last commit aligns more with your thinking?

widget/textgrid.go
```diff
@@ -43,25 +43,15 @@ type TextGridRow struct {
...
-// Since: 2.5
-type TextGridTextStyle interface {
-	TextGridStyle
-	Bold() bool
-	Underlined() bool
-}
-
 // CustomTextGridStyle is a utility type for those not wanting to define their own style types.
 type CustomTextGridStyle struct {
+	TextStyle        fyne.TextStyle
```
andydotxyz, line 53: I think this will still need the since line

text.go
```diff
@@ -45,6 +45,8 @@ type TextStyle struct {
 	Symbol bool // Use the system symbol font.
 	// Since: 2.1
 	TabWidth int // Width of tabs in spaces
+	// Since 2.5
```
andydotxyz, line 48: Missing colon on this line (after "Since").
It is probably worth adding a line that this is currently only supported in TextGrid so we don't mislead until it is available everywhere.

andydotxyz:
> I hope the last commit aligns more with your thinking?

This seems much more consistent thanks.
There is a conflict stopping CI from running unfortunately. Either way I added some small thoughts inline.

mgazza:
> > I hope the last commit aligns more with your thinking?
> 
> This seems much more consistent thanks. There is a conflict stopping CI from running unfortunately. Either way I added some small thoughts inline.

Thanks, I'll add the comments and tidy up the branch so its much more atomic :D

andydotxyz (approved):
Thanks so much, this is great to have in.

andydotxyz:
I guess the apps would need a bold monospace font in the theme to see the difference on that right? Perhaps we should consider with the new system font lookup to find local defaults that are more flexible? A later PR works for that.

Also, I am curious - is italic being included for the TextGrid lots of extra work or would it follow the same model?

Thanks again for sticking through the process here.

mgazza:
The same model will work for italics. yes you have to return a different font for bold and italics in the font lookup

andydotxyz (approved):

PR #4245, opened 2023-09-13
Adding inner window and multi window containers

andydotxyz:
### Checklist:

- [x] Tests included.
- [x] Tests all pass.

innerwin


andydotxyz:
Feel free to test it out. Much testing and docs required I think, but looking for feedback on the API & widgets/layout etc

andydotxyz:
Same windows build issues as https://github.com/fyne-io/fyne/pull/4262 - unrelated to this code

c1ngular:
What is this for ? Window embedding ?

andydotxyz:
> What is this for ? Window embedding ?

for making some content of your app appear within a window inside the app. Like the MDI interfaces from Windows https://en.wikipedia.org/wiki/Multiple-document_interface

c1ngular:
@andydotxyz  i don't understand , so it is not an actual native window , but an UI widget/container looks like a window visually ?

andydotxyz:
It looks like a window. Nothing to do with native windows in border or content. See screenshot at the top or try out the fyne_demo in this branch.

c1ngular:
@andydotxyz  I see , thanks

container/innerwindow.go
```diff
@@ -0,0 +1,201 @@
...
+func (i *innerWindowRenderer) Layout(size fyne.Size) {
+	pad := theme.Padding()
+	pos := fyne.NewSquareOffsetPos(pad / 2)
+	size = size.Subtract(fyne.NewSquareSize(pad))
+	i.LayoutShadow(size, pos)
+
+	i.bg.Move(pos)
+	i.bg.Resize(size)
+
+	pad = theme.Padding()
```
Jacalz, line 109: Are you not setting the value to the same one here?

Jacalz (changes requested):
Nice work. This is very cool. I left a comment inline but also have some thoughts below. You don't necessarily have to address all of them but I think it might be wise to at least consider them.
- Resizing does not change the mouse cursor. Is that expected?
- I think the title would look better if it was centered in the window instead of being left-aligned.
- I feel like we need some sort of way to control the side of the buttons. It looks a bit out of place with inner windows having buttons to the left and my main window having them to the right. I also wander which of these should be the default? With Windows users being more than Mac-users I suspect that controls on the right side would be preferable?
-  The icon that is displayed in the top right is clickable but does nothing when it is clicked. Should it just be an icon and not a button?
- There seems to be less padding on the top of the window than on the other sides? See image below.

image
image

Sorry for the rambly sort of review. This really is great stuff but I think it can be made even better :)

andydotxyz:
> * Resizing does not change the mouse cursor. Is that expected?

GLFW does not have a standard cursor for this yet sadly :( - in the future this could be addressed

> I feel like we need some sort of way to control the side of the buttons. It looks a bit out of place with inner windows having buttons to the left and my main window having them to the right. I also wander which of these should be the default? With Windows users being more than Mac-users I suspect that controls on the right side would be preferable? Or do we change default depending on the OS?

Let's address this in a future release and attempt to switch them based on OS. Not sure how we juggle GTK vs KDE vs Enlightenment though!

> There seems to be less padding on the top of the window than on the other sides? See image below.

The padding on 3 sides is the size of padding, on the top it is the window bar - adding more padding didn't look good

I have addressed the other points. If you agree with the summary above we should be good. I can add a ticket to add and detect the button side so it's not forgotten.

Jacalz:
I think the centered label would have to be centered in the window, not centered between the buttons. Sorry. I understand that it makes the layout a bit more complicated but a relatively simple custom layout should do the job
image

andydotxyz:
> I think the centered label would have to be centered in the window, not centered between the buttons. Sorry. I understand that it makes the layout a bit more complicated but a relatively simple custom layout should do the job
> 
> image
> 

If you want the window to be propped open by title that's a simple custom layout - but this really needs truncation.
What should we truncate to if it's centered in the window?

Jacalz:
Why would having it cantered in the window break truncation? As long as the custom layout handles MinSize correctly, it seems like it should work. All my other windows on my computer has truncation for the title and test in the centre.

andydotxyz:
Because there is a different space required on the left and right it cannot simply be centered with truncate turned on.
It would have to center as long as there is space and then switch to some other mode when the space needed is less than available if assuming even padding but more than is actually required because the right side is so much smaller in terms of button space.

That seems quite complex and given that left align is standard on Windows 11 I'm tempted to go back to that for now rather than chase this truncation algorithm...

Jacalz (approved):
Agreed. Approved

Jacalz (approved):

Jacalz (approved):

PR #4246, opened 2023-09-13
Make Button rendering consistent

### LINKED ISSUE #4243: Inconsistent rendering of Button widget
### Checklist

### Describe the bug

The rendering of the button widget is different before the button is first clicked and after the button has been clicked. Specifically, the box used for the animation of the button has no color and a size of (0,0) prior to the first click and a size that is the same as the button and a color of transparent after the first click. This creates a problem for tests that use the test.AssertObjectRendersTo... and test.AssertRendersTo... functions.

### How to reproduce

Run the attached test code. The first run will fail because there is nothing in testdata, but after you move the two .xml files from failed to testdata it will succeed. Now compare beforeClick.xml to afterClick.xml. The rendering difference is clear.

### Screenshots

image

### Example code

```go
package widget

import (
	"testing"
	"time"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/test"
	"fyne.io/fyne/v2/widget"
)

func TestButtonBg(t *testing.T) {
	button := widget.NewButton("ClickMe", nil)
	time.Sleep(100 * time.Millisecond)
	test.AssertObjectRendersToMarkup(t, "beforeClick.xml", button)
	button.Tapped(&fyne.PointEvent{})
	time.Sleep(100 * time.Millisecond)
	test.AssertObjectRendersToMarkup(t, "afterClick.xml", button)
}

```

### Fyne version

2,4,0

### Go compiler version

1.21.1

Windows 10

### Additional Information

_No response_

pbrown12303:
### Description:
Changed the default rendering of Button so that the animation background rectangle starts off with the same rendering as it has after the button has been clicked

Unfortunately, this affected all of the xml testdata files that contain buttons (87 of them). All test result files have been updated and the full fyne test suite passed with the excaption of some errors in internal/driver/glfw which do not appear to have anything to do with this change.

Fixes #4243 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


pbrown12303:
This is strange. There is only one source file that got changed - all the rest are testdata .xml files. It's acting as if the test data did not get pushed before the tests were run.

andydotxyz:
Should "dialog-custom-custom-buttons.xml" be present?

Jacalz:
@pbrown12303 Would you mind giving the PR a more descriptive title?

widget/button.go
```diff
@@ -99,6 +99,7 @@ func (b *Button) CreateRenderer() fyne.WidgetRenderer {
 	b.background = canvas.NewRectangle(theme.ButtonColor())
 	b.background.CornerRadius = theme.InputRadiusSize()
 	tapBG := canvas.NewRectangle(color.Transparent)
+	tapBG.FillColor = &color.NRGBA{R: 0, G: 0, B: 0, A: 0}
```
Jacalz, line 102: This seems strange. The `canvas.NewRectangle()` constructor already sets `FillColor`. Is this line really needed?

internal/driver/glfw/__debug_bin1201983206.exe
Jacalz, line 1: Please delete this file. Always check what you are adding before committing.

Jacalz (changes requested):

internal/driver/glfw/__debug_bin1201983206.exe
pbrown12303: Done

widget/button.go
pbrown12303: Yes. The tap animation does not return the color to transparent (a nil color). It explicitly sets it to this value. The other alternative would be to alter the animation logic to return the color to nil at the end, but I think this would be a Pandora's box. I certainly do not know enough about animation to even begin to assess the feasibility of this kind of change.

pbrown12303:
@andydotxyz Added missing dialog-custom-custom-buttons.xml

widget/button.go
Jacalz: What I am trying to say is that you can replace `color.Transparent` on the line above instead of setting the field directly after.

However, if this is just a workaround for the animation doing strange things it seems like we ought to fix the animation instead? We are using `color.Transparent` all over the place and if we need this workaround everywhere then something seems very wrong.

Thoughts? @andydotxyz

pbrown12303: Yes, I see. I didn't realize that the value supplied to the NewRectangle() function was the background color - I thought it was the line (edge) color.

With respect to the use of color.Transparent, I suspect it is only an issue with animation. I'll let you and @andydotxyz discuss and draw conclusions - I'll go with whatever you decide.

andydotxyz: Colour.Transparent is not `nil`... 

https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/image/color/color.go%3Bl=345

not sure what nil is being referenced!

Jacalz: Indeed. That was my though as well

pbrown12303: Sorry, I was wrong about transparent being nil: it is a struct with only the A value specified. In any case, I found a better fix. I modified the Button animation so that if the fade (i.e. the A value of the color) reaches zero, the assigned value for the bg.FillColor is color.Transparent, which was the default initial value. Updated test results accordingly.

pbrown12303:
Looks like the expected test results did not get pushed before the tests were run.

andydotxyz:
> Looks like the expected test results did not get pushed before the tests were run.

GitHub doesn't usually get the order wrong. Are you sure all the fixes are pushed?

pbrown12303:
Everything is pushed. GitHub has no way of knowing the relationship between the _test.go files and the .xml test results. The only linkages are the quoted text strings that are the arguments to the test.Assert... functions.

Jacalz:
> GitHub has no way of knowing the relationship between the _test.go files and the .xml test results. The only linkages are the quoted text strings that are the arguments to the test.Assert... functions

That doesn't seem very likely. All other tests (except for a few spurious tests failing sometimes) are passing

andydotxyz:
> Everything is pushed. GitHub has no way of knowing the relationship between the _test.go files and the .xml test results. The only linkages are the quoted text strings that are the arguments to the test.Assert... functions. 

GitHub doesn't understand it true - but it does know when a commit has been pushed and will clearly wait for all the files to be present before it starts the CI run...

andydotxyz:
It looks to me that both failures came from the same test case. I have asked it to re-run just in case, but it seems to me that "go test -tags mobile ./..." is failing every time and for the same reason.

Jacalz:
I don't think the mobile driver supports running in Windows?

andydotxyz (approved):
Great thanks for the tidy here

andydotxyz:
Looks good to me - @Jacalz ?

Jacalz (approved):
Looks good. Thanks

PR #4250, opened 2023-09-14
Introduce new API so Form doesn't have to hardcode validation for Entry

### LINKED ISSUE #4147: Dialog form validation has passed, but submit button is not available.
### Checklist

### Describe the bug

In my code, using a customized Input for the Entry results in the submit button being disabled in the dialog form, even when validation passes.

### How to reproduce

When using a custom Entry it.

### Screenshots

QQ截图20230809174748

### Example code

```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/data/validation"
	"fyne.io/fyne/v2/dialog"
	"fyne.io/fyne/v2/widget"
)

type Input struct {
	widget.Entry
}

func (e *Input) KeyDown(key *fyne.KeyEvent) {
	//...
}

func NewInput() *Input {
	e := &Input{}
	e.ExtendBaseWidget(e)
	return e
}

func main() {
	a := app.New()
	w := a.NewWindow("Hello")

		hello,
			hello.SetText("Welcome :)")
		}),
	))

	username := widget.NewEntry()
	username.Validator = validation.NewRegexp(`\S`, "required")
	password := NewInput()
	password.Validator = validation.NewRegexp(`\S`, "required")

	items := []*widget.FormItem{
		widget.NewFormItem("Username", username),
		widget.NewFormItem("Password", password),
	}
	dialog.ShowForm("Login...", "Log In", "Cancel", items, func(b bool) {
		if !b {
			return
		}

	}, w)

	w.Resize(fyne.NewSize(500, 400))
	w.ShowAndRun()
}
```

### Fyne version

2.3.5

### Go compiler version

1.19

Windows11

### Additional Information

_No response_

Jacalz:
### Description:

This will likely fix #4147

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


Jacalz:
This part is still problematic: https://github.com/fyne-io/fyne/blob/be9cd4b67990836a8d9f518b0c3b5a59b1f0349e/widget/form.go#L312-L314

Jacalz:
The APIs here are very rough around the edges but I would like some opinions on it to try to shape it into something more fit for landing in the codebase. This should solve the problems we had with extended widgets.

validation.go
```diff
@@ -1,5 +1,28 @@
...
 
+// ExpandedValidator expands on the Validatable interface to allow the widget to expose more context to the parent.
+//
+// Since: 2.5
+type ExpandedValidator interface {
+	Validatable
+
+	// HasValidator makes it possible for a parent that cares about child validation (e.g. widget.Form)
+	// to know if the developer has attached a validator or not.
+	HasValidator() bool
```
Jacalz, line 11: Hmm. Might be able to move this into `ShouldDisplayValidation`?

Jacalz:
Hmm. This does still not seem to fix the issue mentioned above. However, it no longer relates to there being an extended entry any more so that's a win. The issue lies somewhere else now :/

andydotxyz (changes requested):
I think the logic in naming `ShouldDisplayValidation` is backwards...

Also this introduces a new type with 4 methods for what looks like a single new function - can you share why?
Lastly, can this not be done with an anonymous interface instead of defining new APIs? If the fix was to resolve extended Entries we don't need to expose any new type as it is a case of Entry implementing a known function only...

Jacalz:
> I think the logic in naming ShouldDisplayValidation is backwards...

Hehe yeah indeed. Nice catch 🙈

> Also this introduces a new type with 4 methods for what looks like a single new function - can you share why?

I'm entirely aware that this API is very rough around the edges. I just basically added a new method incrementally for each place where we were using internal API details from Entry. It definitely needs to be cleaned up, like https://github.com/fyne-io/fyne/pull/4250#discussion_r1387255501 and potentially more. However, I doubt that all four methods can be added as only one. I'm more than happy to receive suggestions, or work together with someone, to try to find a cleaner solution :)

I will try to clean up the API and slim it down now that it actually works like before minus the Entry specific code. However, it would be nice with some thoughts about better solutions to the internal API usage that gets removed here.

> Lastly, can this not be done with an anonymous interface instead of defining new APIs? If the fix was to resolve extended Entries we don't need to expose any new type as it is a case of Entry implementing a known function only...

It can yes, but it avoids parts of what I'm trying to achieve here. The idea was never to work on this just to fix the issue. I was just hoping to fix it at the same time by more generally making the Form not hardcode validation to Entry. Basically, anyone should be able to create a widget that has enough context to display useful validation messages in a Form.

This also doesn't even fix the issue linked above. It just happens to make the extended entry work the same as the not extended one 😅

Jacalz:
I have looked at this some more now and I don't quite know how to simplify it any more. I can't think of a way of reworking Form to not require the methods that I have added.

andydotxyz:
Sorry I let this sit for a while, I will take a closer look and see why each of these methods is needed and try to think of simplifications myself.

widget/entry_validation.go
```diff
@@ -47,6 +63,14 @@ func (e *Entry) SetValidationError(err error) {
 	}
 }
 
+// ShouldDisplayValidation returns true if the entry has not been interacted with
```
andydotxyz, line 66: The dirty check seems to be backwards here - if it is not dirty it should not display validation?

validation.go
```diff
@@ -1,5 +1,28 @@
...
+
+	// SetOnFocusChanged is intended for parent widgets or containers to hook into the change of focus.
+	// The function might be overwritten by a parent that cares about child validation (e.g. widget.Form).
+	SetOnFocusChanged(func(bool))
```
andydotxyz, line 15: This feels like exposing internal details - I can see why it was needed but I don't think it's the cleanest way.
Can't we wrap the widgets if they are already `Focusable` and intercept the `FocusGained` and `FocusLost`?

validation.go
```diff
@@ -1,5 +1,28 @@
...
+
+	// SetValidationError manually updates the validation status until the next input change.
+	SetValidationError(error)
```
andydotxyz, line 18: I'm not sure about this either - it means that validation moves from a action to a state.
I would look carefully at where it is used now - I think that the new methods added should remove the necessity to set an initial state, especially if we have interception of focus events.

andydotxyz (changes requested):
Thanks for this, good to support extendible types etc for validation.
However I think the new interface exposes, or requires, to much - see my comments inline. I think we can leave it at `HasValidator` and `ShouldDisplayValidation` - these are clearly extra metadata about validation so fits within the interface idea.

Jacalz:
There are just too many issues making this problematic to fix in a good way without introducing unwanted APIs. I'll likely never get to fixing this...

PR #4251, opened 2023-09-14
Add Entry undo/redo functionality

xypwn:
### Description:

Implements #436

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### Notes:
I wrote this independently of #2539.

My implementation is fairly extensible, although currently tied to Entry by name.
It works by keeping track of the changes made by the user, NOT by making snapshots. This makes it more efficient and flexible than the other implementation.

### Basic Architecture:
`interface entryUndoAction`: Represents a single user interaction. It contains the functions `Undo` and `Redo`, which modify the current Entry content to undo/redo that user interaction.

`interface entryMergeableUndoAction`: Extends entryUndoAction. Contains the `TryMerge` method, which allows for merging two actions into one. This is useful because, for example when typing a word, we want to undo that whole word at once, and not have to spam `Ctrl+Z` for every character.

`struct entryInsertAction`: Implements `entryMergeableUndoAction` and represents a string insertion or deletion. It only saves the position and string of the edit, i.e. no snapshots, as mentioned previously. It can merge with itself, but doesn't merge across word boundaries.

`struct entryUndoStack`: This is the main interface for dealing with the undo/redo system. It has the following methods:
- `Undo`/`Redo`: Executes the current undo/redo operation on the current string and returns the new string as well as the executed operation.
- `Add`/`MergeOrAdd`: Records the given action, allowing for  `Undo`/`Redo` to be called afterwards. `MergeOrAdd` also tries to merge the given action with the undo action already on top of the stack; if it can't merge, it just reverts to call `Add`.
- `Clear` clears all records in the undo stack. This is done after SetText() is called, for example, also meaning that SetText() modifications aren't recorded.

---

Feel free to drop in any suggestions :)


widget/entry.go
```diff
@@ -525,6 +532,12 @@ func (e *Entry) TappedSecondary(pe *fyne.PointEvent) {
 	clipboard := fyne.CurrentApp().Driver().AllWindows()[0].Clipboard()
 	super := e.super()
 
+	undoItem := fyne.NewMenuItem("Undo", func() {
```
andydotxyz, line 535: this can be inlined without the new `func()` - just pass `e.undo` as the action parameter.

widget/entry.go
```diff
@@ -546,7 +559,7 @@ func (e *Entry) TappedSecondary(pe *fyne.PointEvent) {
 	} else if e.Password {
 		menu = fyne.NewMenu("", pasteItem, selectAllItem)
 	} else {
-		menu = fyne.NewMenu("", cutItem, copyItem, pasteItem, selectAllItem)
+		menu = fyne.NewMenu("", undoItem, redoItem, cutItem, copyItem, pasteItem, selectAllItem)
```
andydotxyz, line 562: Maybe add a separator in there as well?

widget/entry.go
```diff
@@ -1941,3 +2027,179 @@ func getTextWhitespaceRegion(row []rune, col int, expand bool) (int, int) {
...
+	TryMerge(next entryMergeableUndoAction) bool
+}
+
+// Declare conformity with entryMergeableUndoAction interface
+var _ entryMergeableUndoAction = (*entryInsertAction)(nil)
+
+// entryInsertAction implements entryMergeableUndoAction.
+// It represents the insertion/deletion of a single string at a
+// position (e.g. "Hello" => "Hello, world", or "Hello" => "He").
+type entryInsertAction struct {
```
andydotxyz, line 2062: this naming seems a little off - there is a `bool` that changes whether it is insert or delete, but insert is in the name

andydotxyz (changes requested):
This is really excellent work, thanks so much.
A couple of items about naming inline but it is clear and well laid out, thanks.

andydotxyz:
One possible bug to be improved - if I "undo" and there are no items left it seems to put the cursor back to the top left, scrolling content if appropriate. I don't think that should happen?

xypwn:
Hi, glad you like it! I'm not home right now, but when I am I'll be able address these issues and make the necessary changes.

Regarding the naming, insert was essentially vim terminology. But I think I'll change it to something like "modifyAction" instead, since that makes a lot more sense.

Regarding the bug, that's because the new cursor position is default initialized to 0 and if it isn't an insertAction, it uses that default value instead of just returning. Should be an easy fix.

Also, I think it would make sense to expose Undo and Redo as public functions, since you might want that for text editors, for example.

xypwn:
Alright, I made all the changes you requested + I made Undo/Redo public functions

widget/entry.go
```diff
@@ -487,6 +488,44 @@ func (e *Entry) SetText(text string) {
 	e.propertyLock.Unlock()
 }
 
+// Undo un-does the last modifying user-action.
+func (e *Entry) Undo() {
```
andydotxyz, line 492: The new public methods must also include in the doc:
```
\\
\\ Since: 2.5
```

andydotxyz (changes requested):
This looks excellent, thanks for all the fixes.
Please annotate the new APIs with the version number as commented inline

widget/entry.go
xypwn: Done.

andydotxyz: This is not resolved, you missed the ":".

Also, why did you move the public methods back down to the private area? They should be ordered alphabetically with public appearing above private...

xypwn: Okay, I hope it's all correct now. I was a bit tired yesterday.

widget/entry.go
```diff
@@ -869,11 +941,19 @@ func (e *Entry) eraseSelection() {
 		return
 	}
 
+	erasedText := provider.String()[posA:posB]
```
Jacalz, line 944: You are building and allocating an entire new copy of the string contents from all richtext segments here. Can you not use the `entry.Text` field in this case?

widget/entry.go
```diff
@@ -615,17 +648,29 @@ func (e *Entry) TypedKey(key *fyne.KeyEvent) {
 
 		e.propertyLock.Lock()
 		pos := e.cursorTextPos()
+		deletedText := provider.String()[pos-1 : pos]
```
Jacalz, line 651: Same as the comment below.

widget/entry.go
```diff
@@ -525,6 +555,8 @@ func (e *Entry) TappedSecondary(pe *fyne.PointEvent) {
 	clipboard := fyne.CurrentApp().Driver().AllWindows()[0].Clipboard()
 	super := e.super()
 
+	undoItem := fyne.NewMenuItem("Undo", e.Undo)
+	redoItem := fyne.NewMenuItem("Redo", e.Redo)
```
Jacalz, line 559: How common is it to have undo and redo in the context menu? It seems to me like most applications don't expose it there.

Jacalz (changes requested):
Thanks for working on this. I have a few comments inline.

widget/entry.go
```diff
@@ -615,17 +648,29 @@ func (e *Entry) TypedKey(key *fyne.KeyEvent) {
...
+			Position: pos - 1,
+			Text:     deletedText,
+		})
 		e.propertyLock.Unlock()
 	case fyne.KeyDelete:
 		pos := e.cursorTextPos()
 		if provider.len() == 0 || pos == provider.len() {
 			return
 		}
+		deletedText := provider.String()[pos : pos+1]
```
Jacalz, line 665: Missed this one. Same comment about using text field.

Jacalz (changes requested):

widget/entry.go
xypwn: QT has undo/redo in the context menu in every text box by default. Also all popular web browsers. GTK apps are the only counter example I can think of. I don't think it's unreasonable to expose it in the context menu.

xypwn: A possible improvement though would be to only show undo/redo if there is actually anything to undo/redo

xypwn: > A possible improvement though would be to only show undo/redo if there is actually anything to undo/redo

I just implemented that, it was pretty simple

widget/entry.go
Jacalz: But the code you are referencing is called after the text has changed? What I am referring to is called when I believe them to be in sync

widget/entry.go
andydotxyz: Normally it would be enable/disable rather than add/remove - having menus be consistent in size and layout makes it easier to navigate them at speed.

widget/entry.go
xypwn: I just made the changes.

widget/entry.go
xypwn: I mean, it *is* less cluttered. Also that's how Chromium does it. And if you want to operate at speed, you'll want to use keyboard shortcuts anyway. But your project, your design decisions :)

xypwn:
What's it with the tests not passing? Is that my fault?

andydotxyz:
There seems to be a bug with the windows / go 1.17 that has impacted all builds in the last 3 days. Multiple projects have noted it - we may need to downgrade the patch version for that build until it's fixed.

shortcut.go
```diff
@@ -144,3 +144,51 @@ func (se *ShortcutSelectAll) Mod() KeyModifier {
 func (se *ShortcutSelectAll) ShortcutName() string {
 	return "SelectAll"
 }
+
+// ShortcutUndo describes a shortcut undo action.
+type ShortcutUndo struct{}
```
andydotxyz, line 149: These new types need the extra:

```go
//
// Since: 2.5
```

andydotxyz (changes requested):
Just 2 types missing the Since line sorry I missed that before.

Otherwise I think it's great - @Jacalz ?

shortcut.go
xypwn: Ah yes. Didn't think of that either.

I added it.

andydotxyz (approved):
Amazing thanks

Jacalz:
I'll try to review tomorrow or on Monday. Been a bit busy lately

Jacalz (approved):
Thanks @xypwn for a well written PR both in terms of code and the extensive description. Your work is much appreciated. Looks good to me.

Also, congratulations on landing your first PR for Fyne :partying_face:

xypwn:
Let's gooo! Thank you all!

andydotxyz:
So excited this is in - thanks a lot :) v2.5.0 looks great already and it's still 3-4 months away!

xypwn:
> @andydotxyz Can you please share the release plan for these changes? I cannot find them in the master branch.

It will be added in v2.5.0

andydotxyz:
> > @andydotxyz Can you please share the release plan for these changes? I cannot find them in the master branch.
> 
> It will be added in v2.5.0

Exactly right. As you can see we are working towards feature freeze now, but you can use it on @develop branch before then.
https://github.com/fyne-io/fyne/wiki/Releases

PR #4261, opened 2023-09-18
Disable toolbar

### LINKED ISSUE #2306: Add the ability to disable ToolbarActions

Rakshana2300:
### Description:
Issue addressed Add the ability to disable ToolbarActions. 

Fixes #2306 

### Checklist:

- [x] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/widget/toolbar.go
+++ b/widget/toolbar.go
@@ -24,7 +24,7 @@ type ToolbarAction struct {
 func (t *ToolbarAction) ToolbarObject() fyne.CanvasObject {
 	button := NewButtonWithIcon("", t.Icon, t.OnActivated)
 	button.Importance = LowImportance
-	if t.disable == true {
+	if t.disable {
 		button.Disable()
 	}
 	return button
--- a/widget/toolbar.go
+++ b/widget/toolbar.go
@@ -17,14 +17,14 @@ type ToolbarItem interface {
 type ToolbarAction struct {
 	Icon        fyne.Resource
 	OnActivated func() `json:"-"`
-	disable     bool
+	disabled    bool
 }
 
 // ToolbarObject gets a button to render this ToolbarAction
 func (t *ToolbarAction) ToolbarObject() fyne.CanvasObject {
 	button := NewButtonWithIcon("", t.Icon, t.OnActivated)
 	button.Importance = LowImportance
-	if t.disable {
+	if t.disabled {
 		button.Disable()
 	}
 	return button
@@ -40,12 +40,11 @@ func (t *ToolbarAction) SetIcon(icon fyne.Resource) {
 
 // NewToolbarAction returns a new push button style ToolbarItem
 func NewToolbarAction(icon fyne.Resource, onActivated func()) *ToolbarAction {
-	t := &ToolbarAction{
+	return &ToolbarAction{
 		icon,
 		onActivated,
 		false,
 	}
-	return t
 }
 
 // ToolbarSpacer is a blank, stretchable space for a toolbar.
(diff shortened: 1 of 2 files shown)


Rakshana2300:
Hi @andydotxyz and @Jacalz, I've came across this issue and took the opportunity to work on it. Please let me know if further changes are required.
Thanks!

widget/toolbar.go
```diff
@@ -17,13 +17,16 @@ type ToolbarItem interface {
 type ToolbarAction struct {
 	Icon        fyne.Resource
 	OnActivated func() `json:"-"`
+	disable     bool
```
Jacalz, line 20: Wording this as "disable" sounds more like you want someone to disable it. I think "disabled" is a better name as the field specifies if the component is disabled or not.

widget/toolbar.go
```diff
@@ -37,7 +40,12 @@ func (t *ToolbarAction) SetIcon(icon fyne.Resource) {
 
 // NewToolbarAction returns a new push button style ToolbarItem
 func NewToolbarAction(icon fyne.Resource, onActivated func()) *ToolbarAction {
-	return &ToolbarAction{icon, onActivated}
+	t := &ToolbarAction{
+		icon,
+		onActivated,
+		false,
+	}
+	return t
```
Jacalz, line 48: I'd rather keep it the way it was returning the struct directly. There is no need to store it in a variable first. It is also a good idea to switch to named values given that you don't have to specify all field (false is the default value for bool).

```suggestion
	return &ToolbarAction{
		Icon: icon,
		OnActivated: onActivated,
	}
```

widget/toolbar.go
```diff
@@ -112,6 +120,15 @@ func NewToolbar(items ...ToolbarItem) *Toolbar {
 	return t
 }
 
+func (t *Toolbar) Disable() {
+	for _, item := range t.Items {
+		_, ok := item.(*ToolbarAction)
+		if ok {
+			item.(*ToolbarAction).disable = true
+		}
```
Jacalz, line 128: You don't need to cast the item twice here. You can keep the reference and just use that (I used disabled for the wording):
```suggestion
		if action, ok := item.(*ToolbarAction); ok {
			action.disabled = true
		}
```

widget/toolbar.go
```diff
@@ -112,6 +120,15 @@ func NewToolbar(items ...ToolbarItem) *Toolbar {
 	return t
 }
 
+func (t *Toolbar) Disable() {
```
Jacalz, line 123: Please add a comment to this public method. It also needs a `Since: ` line at the bottom.

Jacalz (changes requested):
Thanks for the PR @Rakshana2300. I haven't tested it yet but just provided a minor code review to give you an idea of how the code can be improved.

widget/toolbar.go
Rakshana2300: Hi @Jacalz, thanks for the review. I'll modify the code to include the above mentioned changes

Jacalz: Thanks. FYI, you don't need to tag any of us unless it is unclear who you are directing your message too. Not doing so is actually preferred because we don't have to notified directly but can look at things when we have time. We have a project dashboard where we see al new PRs and we get notifications for discussions we are taking part in.

Jacalz: Sorry but you added the wrong version in the since line. 2.4 has already been released. The upcoming feature  release is 2.5.

andydotxyz:
Thanks so much for looking at this, but I think this isn't a fix for the linked issue - it disables the whole toolbar not just an action.

andydotxyz (commented):
Thanks for this. But I think this won't work if the items are visible already...

andydotxyz:
Was this closed because it became impossible after feedback or you ran out of time to work on it? Just curious so we know if it should be taken forward somehow.
Thanks for your help!

Rakshana2300:
Hey Andy, I got held up in some other and couldn't able to pick this up. I closed it so that maybe others can work on it.

andydotxyz:
OK thanks. Feel free to come back when you're available if nobody else has picked it up.

PR #4265, opened 2023-09-20
Fix Compilation with Android NDK r26

BieHDC:
### Description:

Fixes Compilation with [Android NDK r26](https://github.com/android/ndk/wiki).
Does not break compatibility with r25c.
This is in preperation to add this path of building Android Applications as an option.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/mobile/android.c
+++ b/internal/driver/mobile/android.c
@@ -41,7 +41,7 @@ static jmethodID find_static_method(JNIEnv *env, jclass clazz, const char *name,
 	return m;
 }
 
-char* getString(uintptr_t jni_env, uintptr_t ctx, jstring str) {
+const char* getString(uintptr_t jni_env, uintptr_t ctx, jstring str) {
 	JNIEnv *env = (JNIEnv*)jni_env;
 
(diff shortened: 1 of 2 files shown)


internal/driver/mobile/android.c
```diff
@@ -178,12 +178,12 @@ char* readStream(uintptr_t jni_env, uintptr_t ctx, void* stream, int len, int* t
 		return NULL;
 	}
 
-	char* bytes = malloc(sizeof(char)*count);
+	jbyte* bytes = (jbyte*)malloc(sizeof(char)*count);
```
andydotxyz, line 181: Can you guarantee jbyte and char are the same size, or should the sizeof parameter be updated too?

andydotxyz (commented):
Thanks for this - just a question inline

internal/driver/mobile/android.c
BieHDC: Oversight. Fixed.

andydotxyz (approved):
Thanks so much

andydotxyz:
Thanks so much, we appreciate your contribution.

PR #4277, opened 2023-09-23
More space for file names

### LINKED ISSUE #2165: NewFolderOpen: incomplete filenames

andydotxyz:
Also fix up spacing/padding for folder dialog in general

Fixes #2165

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


dialog/fileitem.go
```diff
@@ -76,52 +77,53 @@ func (f *fileDialog) newFileItem(location fyne.URI, dir, up bool) *fileDialogIte
...
 
-	padding = fyne.NewSize(theme.Padding(), theme.Padding()*4)
-	return fyne.NewSize(fileInlineIconSize+s.text.MinSize().Width, fileTextSize).Add(padding)
+	textMin := s.text.MinSize()
+	return fyne.NewSize(fileInlineIconSize+textMin.Width+theme.Padding(), textMin.Height)
 }
 
-func (s fileItemRenderer) Refresh() {
+func (s *fileItemRenderer) Refresh() {
+	s.fileTextSize = widget.NewLabel("M\nM").MinSize().Height // cache two-line label height
```
Jacalz, line 118: Just wandering if this can be improved in any way? Seems a bit strange to create a new temporary label just to get the minsize but I understand if there isn't a better way.

Jacalz (approved):
Nice stuff. A huge improvement. Approved but please consider my inline comment before merging.

dialog/fileitem.go
andydotxyz: I agree that it doesn't look neat - but I tried the alternative of keeping a widget around and checking the MinSize regularly and it was slower.

andydotxyz:
There was no file change in the "merge conflict" weird!

PR #4278, opened 2023-09-24
FileDialog view layout preference

JordanGoulder:
### Description:
Prior to this change, the FileDialog always started with grid view regardless of the previously selected view. This pull request adds a `fileDialogViewLayout` key to app preferences to save the last view layout selected. The next time a FileDialog is show it will start with the last view layout selected by the user.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -208,12 +208,12 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 		title = label + " Folder"
 	}
 
-	view := fyne.CurrentApp().Preferences().Int(viewLayoutKey)
-	if view != int(listView) {
-		view = int(gridView)
+	view := viewLayout(fyne.CurrentApp().Preferences().Int(viewLayoutKey))
+	if view != listView {
+		view = gridView
 	}
 
-	f.setView(viewLayout(view))
+	f.setView(view)
 
 	f.loadFavorites()
 
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -206,7 +206,13 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 		title = label + " Folder"
 	}
 
-	f.setView(gridView)
+	viewPreference := fyne.CurrentApp().Preferences().StringWithFallback("fileDialogViewLayout", "grid")
+	if viewPreference == "grid" {
+		f.setView(gridView)
+	} else {
+		f.setView(listView)
+	}
+
 	f.loadFavorites()
 
 	f.favoritesList = widget.NewList(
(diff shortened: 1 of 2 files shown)


dialog/file.go
```diff
@@ -206,7 +206,13 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 		title = label + " Folder"
 	}
 
-	f.setView(gridView)
+	viewPreference := fyne.CurrentApp().Preferences().StringWithFallback("fileDialogViewLayout", "grid")
+	if viewPreference == "grid" {
+		f.setView(gridView)
+	} else {
+		f.setView(listView)
+	}
```
Jacalz, line 214: I don't think we gain much from using a string for this. Sure, it makes the JSON a bit more human readable but I don't think it is very useful? It is an internal detail after all. If you use an int for the preference, you can avoid the if-statement when loading it into the setView. It will also default to 0 by default so no need to load with a default.
```suggestion
	viewPreference := fyne.CurrentApp().Preferences().Int("fileDialogViewLayout")
	f.setView(viewLayout(viewPreference))
```

dialog/file.go
```diff
@@ -499,10 +512,12 @@ func (f *fileDialog) setView(view viewLayout) {
 		grid := widget.NewGridWrap(count, template, update)
 		grid.OnSelected = choose
 		f.files = grid
+		fyne.CurrentApp().Preferences().SetString("fileDialogViewLayout", "grid")
 	} else {
 		list := widget.NewList(count, template, update)
 		list.OnSelected = choose
 		f.files = list
+		fyne.CurrentApp().Preferences().SetString("fileDialogViewLayout", "list")
 	}
```
Jacalz, line 521: With my suggestion to not store strings in the preferences, you can hoist it out of the if-statement.
```suggestion
	} else {
		list := widget.NewList(count, template, update)
		list.OnSelected = choose
		f.files = list
	}
	
	fyne.CurrentApp().Preferences().SetInt("fileDialogViewLayout", int(view))
```

Jacalz (changes requested):
Thanks for working on this. It is a great improvement. I just left some ideas inline about simplifying the code a bit :)

I also have a though about the preference lookup name "fileDialogViewLayout". It is not necessarily a problem but it is worth noting that the name theoretically can conflict with a name the developer already is using. I wonder if we should have some internal standard for how to handle it? Maybe prefixing with "fyne:" like how "go:build" tags and such work?

dialog/file.go
JordanGoulder: I was back and forth on this but I think int does make sense since it is internal. I have updated to make this change.

Jacalz (changes requested):
Thanks. Just one minor suggestion

Jacalz:
Please also have a look at simplifying `makeUI()` a bit. The cyclomatic complexity is a bit too high. You might want to separate something out to a function.

dialog/file.go
```diff
@@ -206,7 +208,13 @@ func (f *fileDialog) makeUI() fyne.CanvasObject {
 		title = label + " Folder"
 	}
 
-	f.setView(gridView)
+	view := fyne.CurrentApp().Preferences().Int(viewLayoutKey)
+	if view != int(listView) {
+		view = int(gridView)
+	}
+
+	f.setView(viewLayout(view))
```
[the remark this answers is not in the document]
JordanGoulder: Great suggestion! Done.

Jacalz (approved):
Thanks for the PR. It looks good to me :)

PR #4294, opened 2023-10-02
fyne release: on codesign error, report stderr/out 

glycerine:
and keep the inputs around for manual debugging by re-running the codesign
command that errored.

### DIFF
--- a/cmd/fyne/internal/commands/release.go
+++ b/cmd/fyne/internal/commands/release.go
@@ -282,7 +282,7 @@ func (r *Releaser) nameFromCertInfo(info string) string {
 	return cn[pos+3:]
 }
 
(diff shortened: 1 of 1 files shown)


glycerine:
I couldn't figure out the errors I was dealing with from codesign without the stderr/stdout, and then checking that I had resolved the situation manually was only possible by retaining the files that are input to the command, and knowing the command that would be attempted.

This simply provides everything needed for troubleshooting without having to hack the fyne code again.  

I provided this simply as a suggestion, based on the experience of what I needed in order to resolve the problems I had. 

There's no need to adopt it in its present form.  Do it however you like, or not at all.  Adapt it, if it helps. 

I'm busy with other stuff and so I'll close this to avoid getting future github update/notice spam.

PR #4309, opened 2023-10-10
Add canvas refresh to richtext renderer refresh

### LINKED ISSUE #4298: RichText Refresh doesn't work when replacing the whole Segments slice
### Checklist

### Describe the bug

Trying to replace the entire text of a rich text and calling refresh does not actually refresh until some external force (resizing the window or hiding and showing the window) force an update. This isn't the case when appending to the rich text segments or when replacing the text of a single segment, only when replacing the whole slice of segments

### How to reproduce

1. add a RichText widget
2. change the segments to a new slice with the new text segment
3. call Refresh

### Screenshots

_No response_

### Example code

```go
func main() {
	a := app.New()
	w := a.NewWindow("")

	rt := widget.NewRichTextWithText("ORIGINAL TEXT")
	go func() {
		time.Sleep(time.Second * 3)
		// appending works
		rt.Segments = append(rt.Segments, &widget.TextSegment{Text: "append text", Style: widget.RichTextStyleInline})
		rt.Refresh()

		time.Sleep(time.Second * 3)
		// so does replacing the content
		rt.Segments = slices.Delete(rt.Segments, 1, len(rt.Segments))
		rt.Segments[0].(*widget.TextSegment).Text = "remove all but first segment and update it"
		rt.Refresh()

		time.Sleep(time.Second * 3)
		// THIS doesn't work
		rt.Segments = []widget.RichTextSegment{&widget.TextSegment{Text: "replace the whole slice", Style: widget.RichTextStyleInline}}
		rt.Refresh()
	}()

	w.SetContent(rt)
	w.ShowAndRun()
}
```

### Fyne version

2.4.0

### Go compiler version

1.21.0

windows 10

### Additional Information

I have not yet tested this on ubuntu or windows 11 but I feel like this isn't an OS problem so I'm not bothering for now.

Also, when force refreshing with a mouse resize of the screen it looks like the RichText min size isn't updated either, so I suspect this is in fact the Layout function in the Refresh but I honestly couldn't parse what that function did and I'm out of time to check

mbaklor:
### Description:

Ensures refresh happens even when segment slice completely replaced. Looks like `TextRenderer.Refresh()` only calls `canvas.Refresh(...)` on the canvas texts in the text segments, and only when they already exist in the cache, so I added a `canvas.Refresh(r.obj.super())` to make sure it always happens, even when the text segment is completely replace by a new instance.
Tests not included because I couldn't replicate the issue in a test, looks like the `testDriver` and `glDriver` have a different impl of `CanvasForObject(obj)` where the `testDriver` never returns `nil` so it never doesn't refresh.

Fixes #4298 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/richtext.go
+++ b/widget/richtext.go
@@ -707,6 +707,7 @@ func (r *textRenderer) Refresh() {
 	r.obj.propertyLock.Unlock()
 
 	r.Layout(r.obj.Size())
+	canvas.Refresh(r.obj.super())
 }
 
 func (r *textRenderer) layoutRow(texts []fyne.CanvasObject, align fyne.TextAlign, xPos, yPos, lineWidth float32) (float32, float32) {


dweymouth:
I think I will defer approving this to someone else with more familiarity with the internal refreshing system. But one concern I'd have is that this are adding a new refresh call to all codepaths, including ones that didn't trigger the original issue. Have you done any performance regression testing on this to make sure this is not making RichText refreshes more expensive in all cases?

mbaklor:
I did not, but I asked if there was a way to check if the internal segments were changed and was told no, so I took a page from `form` and `AppTabs` and added the canvas refresh regardless. If this isn't the recommended way let me know I'll gladly edit this

andydotxyz (approved):
It's quite appropriate to ask the canvas to refresh our object in the widget Refresh call - that is what ensures our caches invalidate. It could be a slowdown - but when someone calls Refresh on the widget it would be assumed that the data has changed anyway.

PR #4313, opened 2023-10-13
Ability to have theme overrides for a widget/container

andydotxyz:
As discussed at contributor meeting, API updated to simpler and less exposed info.
Does not include all widgets, that will be a follow-on PR.
Additional optimisations could follow to reduce time spent in setting/walking the list of overridden widget themes.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
This isn't the whole codebase but it's the APIs and techniques required.
Done: Button, Label, Hyperlink.

Rather than drag on a large PR for much longer and get them all I'd like to land this one as soon as the API/code structure etc is approved and track remaining widgets in another issue with multiple PRs.

pbrown12303 (commented):
A thought occurred to me after today's discussion. If you go the route of having a special container (or widget) to which the alternate theme can be applied and you wait until the rendering to actually apply it to the children (as I think was suggested during the discussion), then if the application queries one of the children for a theme-based property between the time the parent's theme has changed and a rendering has occurred, the result will be incorrect.

andydotxyz:
That is a good point. It will require a path to make sure that it is known before CreateRenderer is called...

container/theme.go
```diff
@@ -0,0 +1,43 @@
+package container
+
+import (
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/internal/cache"
+	"fyne.io/fyne/v2/widget"
+)
+
+type ThemeOverride struct {
```
Jacalz, line 9: Should this type be exported? It is missing a comment and the zero value does not seem to work?

internal/cache/widget.go
```diff
@@ -57,6 +57,14 @@ func DestroyRenderer(wid fyne.Widget) {
 	if rinfo != nil {
 		rinfo.renderer.Destroy()
 	}
+	delete(overrides, wid)
+	for r, w := range forWidgets {
+		if w == wid {
+			delete(forWidgets, r)
+			return
```
Jacalz, line 64: Should this not be a `break`?

theme/theme.go
```diff
@@ -24,6 +25,10 @@ const (
 	variantNameUserPreference fyne.ThemeVariant = 2 // locally used in builtinTheme for backward compatibility
 )
 
+var (
+	defaultTheme fyne.Theme
+)
```
Jacalz, line 30:
```suggestion
var defaultTheme fyne.Theme
```

widget/widget.go
```diff
@@ -163,6 +169,15 @@ func (w *BaseWidget) super() fyne.Widget {
 	return impl
 }
 
+// theme returns a cached theme instance for this widget (or its extending widget).
+func (w *BaseWidget) theme() fyne.Theme {
+	if w.themeCache == nil {
+		w.themeCache = theme.CurrentForWidget(w.impl)
+	}
+
+	return w.themeCache
```
Jacalz, line 178: This should use the locks?

widget/widget.go
```diff
@@ -163,6 +169,15 @@ func (w *BaseWidget) super() fyne.Widget {
 	return impl
 }
 
+// theme returns a cached theme instance for this widget (or its extending widget).
+func (w *BaseWidget) theme() fyne.Theme {
```
Jacalz, line 173: I wonder if this would be better off exported for other developers to enjoy?

Jacalz (changes requested):
Thanks. This is looking better. Left some suggestions

widget/widget.go
andydotxyz: I did wonder about that. But if we do, people might thing that overriding it will use a different theme - so I wanted to avoid that confusion.

Jacalz: Hmm. I don't know. Seems like a very useful thing to have access to.

widget/widget.go
```diff
@@ -163,6 +169,21 @@ func (w *BaseWidget) super() fyne.Widget {
...
+	w.propertyLock.RLock()
+	cached := w.themeCache
+	w.propertyLock.RUnlock()
+	if w.themeCache == nil {
```
Jacalz, line 177:
```suggestion
	if cached == nil {
```

Jacalz (changes requested):

widget/widget.go
andydotxyz: That's fair - I added it as a public API and tried to be clear that it's a utility.

widget/widget.go
andydotxyz: Oh, missed that before sorry - it's in now

andydotxyz:
All merged and up to date now, thanks for the reviews :)

Jacalz:
Sorry for the late review. It seems like there is a conflict that needs to be resolved,

Jacalz (approved):
I'm happy with this now but given that it is such a large and widespread change, I'd prefer if someone more review before merging :+1:

andydotxyz:
Maybe @dweymouth, @Bluebugs or @toaster are around? I think they were all on the call when this new API was proposed by Tilo.

andydotxyz:
Sorry @Jacalz loads of merge conflicts.
I also removed a recently added container theme test as it was making a whole load of CI tests unstable. I cannot fathom what it is doing and have dug quite deep. Hopefully that's not a big loss!

internal/cache/theme.go
```diff
@@ -0,0 +1,64 @@
+package cache
+
+import "fyne.io/fyne/v2"
+
+var (
+	forWidgets = make(map[fyne.Resource]fyne.Widget)
```
dweymouth, line 6: Is there a way to do without this forWidget map, it seems a bit weird to me and like it's an inversion of encapsulation in a way. Since ThemedResource is a widget, is it not just taken care of by the overrides map?

Anyway, there's another bigger issue with ThemedResources with this (see other comment) and in reworking for that we may find a solution for this as well

theme/icons.go
```diff
@@ -633,7 +649,7 @@ func NewSuccessThemedResource(src fyne.Resource) *ThemedResource {
 }
 
 // NewThemedResource creates a resource that adapts to the current theme setting.
```
dweymouth, line 651: There is an issue with ThemedResource and SVG caching that this will introduce. SVGs are cached by resource name, and themed resource prepends the theme color *name* to the base resource name. This worked when there was only one color for any given color name, and the entire cache was cleaned out on theme variant changes. But now we can have multiple colors for a given theme color name, because we can have multiple active themes throughout the app.

Perhaps we need to introduce a lower-level item like ColoredResource (it doesn't need to be part of the public API, at least initially), that is like ThemedResource but it is parametrized by an actual color.Color value rather than a theme color name, and ThemedResource is a higher-level wrapper of ColoredResource.

dweymouth (changes requested):

theme/icons.go
andydotxyz: Good catch.
I think the solution is one that would solve both of your points. To have the resource cached to the scope of an override container.
I cannot see how right now, but that should fix the scope change with the minimal change outside this feature... I hope!

internal/cache/theme.go
andydotxyz: The point above is interesting, though based on "Since ThemedResource is a widget" which is false unfortunately. I think the point below did open up a way to remove this breaking of encapsulation.

andydotxyz:
OK there is a complex interconnected problem to solve here that are causing import loop headaches. There are 3 choices:

1) A version of the original `ThemedResource.ForWidget` or `WidgetThemedResource.Owner` where we are wrapping a themed resource and remembering which widget it is being used in. Slightly better than the first try.
2) The version most recently here where the cache package tracks the widget for each resource - adapted to work around the cache issue raised.
3) Move the `ThemedResource` to the top level namespace (with a deprecated alias for the old) that works around the import loop. Like we planned to do for the scroll directions consts.

I'm leaning towards 3 but I didn't want to implement a third iteration of this code if it's not something reviewers agree with.

dweymouth:
The SVG caching issue I mentioned is lower-level and I don't think any of the above would fix it (they're totally orthogonal). The issue is that in the SVG rasterization cache https://github.com/fyne-io/fyne/blob/master/internal/cache/svg.go we cache by name, and the ThemedResource prepends the theme color name to create a unique resource name for caching. But this only worked when a theme color name could only be associated with a single color at once per app. I think we need to change the resource naming for themed resources to have the color RGBA values form part of the resource name instead

andydotxyz:
Actually I have a fix locally that works around it with multiple caches so there is no name collision. That way widgets sharing the override share the same cache.

andydotxyz:
Edit, the ID update is easier - using the scope of different caches to simply prefix the IDs works quite well. Just working on how the import loop issue for ThemedResource may work for the `ColorName` to ensure the cache updates correctly if that value changes.

Jacalz:
You might be able to change `ColorName` to be an alias to `string` and just use strings for any internal calls.

andydotxyz:
The colour name type is already top level, it's being able to look up the new value that's a challenge. 

I've realised there are two solutions:

1) a top level interface "ThemedResource" that adds a lookup function the theme package can implement.

2) the theme package can offer the entry point to allowing theme override for resources and they could extract a pointer to the colour name field for the cache to use.

I think 1 wins - 2 feels a little hacky for a reason I can't quite put my finger on.

andydotxyz:
Is this looking good now @dweymouth, @Jacalz ?

internal/cache/theme.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+var overrides = make(map[fyne.Widget]*overrideScope)
+
+type overrideScope struct {
+	th      fyne.Theme
+	cacheID string
+}
+
+var nextID = 1
```
Jacalz, line 17: This seems bound to cause race conditions and potential collisions between ids. Use an atomic counter (use `.Add(1)` and the return value)?

We might need a lock around the map as well maybe?

andydotxyz: Should all be sorted now, sorry for the re-review

Jacalz (approved):

andydotxyz:
Thanks @Jacalz - very exciting :)

I think that @dweymouth has a bit of an outage but I'll wait a little longer in case they are back and have time to confirm the fix.

dweymouth:
> Thanks @Jacalz - very exciting :)
> 
> I think that @dweymouth has a bit of an outage but I'll wait a little longer in case they are back and have time to confirm the fix.

Yep, still have intermittent internet access. Hopefully should be fully back online by end of tomorrow, but no guarantees. I will try to take a look at this later in the coming week

widget/button.go
```diff
@@ -167,7 +166,7 @@ func (b *Button) MouseOut() {
 
 // SetIcon updates the icon on a label - pass nil to hide an icon
 func (b *Button) SetIcon(icon fyne.Resource) {
-	b.Icon = icon
+	b.Icon = cache.OverrideResourceTheme(icon, b)
```
dweymouth, line 169: Hmm, I feel like having a widget public `XYZ` field and the `SetXYZ` function result in `wid.XYZ` being something other than what was passed into `SetXYZ` is confusing behavior. I don't think we have a precedent for `wid.SetXYZ(f00)` being in any way different, behavior-wise, from `wid.XYZ = foo; wid.Refresh()`?

dweymouth (changes requested):

internal/cache/theme.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Content returns the underlying content of the resource adapted to the current text color.
+func (res *WidgetResource) Content() []byte {
+	name := res.ThemeColorName()
+	if name == "" {
+		name = "foreground"
+	}
+
+	th := themeForResource(res)
+	return svg.Colorize(res.ThemedResource.Content(), th.Color(name, fyne.CurrentApp().Settings().ThemeVariant()))
```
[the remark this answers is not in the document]
dweymouth: OK you can ignore this particular comment because I thought of a way to resolve this in the general case here: https://github.com/fyne-io/fyne/pull/4558

widget/button.go
andydotxyz: This is simply a shortcut required to adopt a resource when being set on a button. It replaces the prior "ForWidget" field on the resource.
If setting an icon resource manually and the developer wants it to reflect theme override it could be called manually as it's public api.

I'm not sure what the alternative is. If you have suggestions please say. As far as I can see we either link the resource to the widget, or we wrap the resource when the widget adopts it. I could certainly remove this line of code for a "clean setter" but then developers would have to manually adopt the resource for the widget. Is that better?

test/markup_renderer.go
```diff
@@ -130,6 +131,9 @@ func (r *markupRenderer) setResourceAttr(attrs map[string]*string, name string,
 	if rsc == nil {
 		return
 	}
+	//if strings.Contains(name, "_") {
```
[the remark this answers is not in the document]
andydotxyz: Thanks, done

widget/button.go
Jacalz: This also isn't something custom widgets will be able to do. That can be problematic.

andydotxyz: I'm not following what custom widgets can't do?

dweymouth: I guess the alternative could be calling `cache.OverrideResourceTheme` on the icon in the Refresh function? There could be other alternatives, I just think that we should stick to the invariant that `wid.SetFoo(bar)` results in `wid.Foo == bar` always (where there is both a publicly exported `Foo` property and the setter)

andydotxyz: Won't that infinitely wrap the resource?
Either way it will result in the field access not returning the same as Set was passed. Potentially worse the field will change over time...

dweymouth: I was referring to having a new private field `themedIcon` or similar in the button, that on Refresh, we set `btn.themedIcon = cache.OverrideResourceTheme(btn.Icon, btn)`. Then the button uses `themedIcon` internally instead of `Icon` directly

andydotxyz: Ah thank you. Let me explore that.

widget/button.go
```diff
@@ -369,37 +378,38 @@ func (r *buttonRenderer) applyTheme() {
...
-				r.icon.Resource = res.Original()
-				r.icon.Refresh()
+				icon = res.Original()
+			}
+		case *theme.ThemedResource, fyne.ThemedResource:
+			if r.button.Importance == HighImportance || r.button.Importance == DangerImportance || r.button.Importance == WarningImportance || r.button.Importance == SuccessImportance {
+				icon = theme.NewInvertedThemedResource(res)
 			}
 		}
+		r.icon.Resource = cache.OverrideResourceTheme(icon, r.button)
```
Jacalz, line 392: I'm still a bit worried about the fact ghat a custom widget wouldn't be able to set the resource this way given that the `cache` package is internal.

andydotxyz: I can expose it through a public method on `container.ThemeOverride` like `AdoptResource` or `OverrideResource`?

Jacalz: That sounds like a good plan 👍

andydotxyz: Hmm. actually it can't be there - that is handy for app developers not custom widget developers.
`BaseWidget` is a place it could be - but then every single widget would have that function on it.
The `theme` package remains an option but I think that goes against moving to the container type recently.

I am thinking that this maybe benefits from being part of the "could be done from the render stack" which came up on a call but is an additional layer of work/refactoring. That way any resource in an overridden container could be interpreted that way in a future PR, removing the need for any more public API?

dweymouth: Good point

Jacalz: That does sound like an even better solution although I understand that it will require quite a lot of work

Jacalz (dismissed):
Approved with the requirement that an issue is opened as a blocker against v2.5.0 to make sure that the last remaining improvements are resolved before then.

andydotxyz:
Thanks @Jacalz I added https://github.com/fyne-io/fyne/issues/4596

Jacalz (approved):

dweymouth (approved):

PR #4320, opened 2023-10-15
Hyperlink: resize underline and focus rectangles to text width; ignore tap outside of text area

### LINKED ISSUE #3528: Hyperlink enhancement: underline and tappable area shouldn't be wider than the text label
### Checklist

If a hyperlink widget is placed in a layout that makes it wider than needed to fully display the text label, the full area of the hyperlink widget is tappable and the underline extends past the end of the text, the full width of the widget.

Yes, but it's really hacky:
* If you want to support textwrap truncate, create a custom widget that extends (or wraps) a `widget.Hyperlink` and overrides minSize to return something sensible to keep it from shrinking to zero
* place the `widget.Hyperlink` or your custom hyperlink wrapper above in a HBox container with a spacer, and use this container where you'd use `widget.Hyperlink` otherwise. The layout will shrink the hyperlink to only show the text and take up the rest of the space to the right with the spacer

Ideally, the desktop pointer should only change when hovering over the text label, tapping/clicking the widget outside the width of the text label should do nothing, and the visual underline should never extend past the width of the text.

dweymouth:
### Description:
Constrain the focus rectangle, underline, and tappable area to the part of the hyperlink widget that is actually displaying text, if the widget itself is sized larger than the text display area.

Fixes #3528

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.


andydotxyz:
I think there are new test failures (window tests on the ubuntu CI) is it possible the hyperlink is used in them and this change is impacting the outcomes?

dweymouth:
It seems like this has uncovered a platform difference between Ubuntu and the other desktop OSes - I can see from adding debug logs that MouseIn is being called on the Hyperlink widget, which then sets the hovered flag to true so that the Cursor() function will return Pointer cursor on the next invocation - but it seems like it calls Cursor first, and only once, where on Mac/Windows, it either must call cursor twice, or calls it after MouseIn.

I think we can either remove Hyperlink from that test, or else correct for this platform difference in the driver

Jacalz (approved):
Nice. Well done :)

andydotxyz (approved):
This looks like a great fix, I agree with the change in order of cursor lookup.
But has it caused more test failures? Attempting to re-run the 3 failures

andydotxyz:
I cannot re-run them but it seems that the cursor tests on Ubuntu are all failing?

andydotxyz:
@dweymouth it looks like tests are failing on Ubuntu due to some Cursor issues.
Can you check it out and see if perhaps they were testing for the wrong thing with the previous behaviour?

andydotxyz (changes requested):
To resolve test failures on Linux

dweymouth:
From what I could tell the tests were failing because we actually have "incorrect" behavior - in that the widget is asked for its cursor type *before* the mouse move is processed, so if the mouse moves in one event onto the hyperlink's tappable area, the cursor doesn't actually change until the next mouse event.

I am on vacation right now so probably won't really look at this until next week, so if someone wanted to take it over, go for it! But the problem was that the Cursor function is called synchronously while mouse events were placed in an event queue before being sent to the widget. So the challenge is to find out a way to ensure Cursorable widgets always have their Cursor function called after each mouse movement, so if they need to update their cursor type in response to the mouse movement, they can do so.

Jacalz:
It would be nice to get this landed. Just a friendly reminder so to say :)

dweymouth:
I think we need @andydotxyz to make an executive decision here - 

* Do we decide we can ignore the "incorrect" behavior with little user impact, and just re-work the tests to pass and file a ticket to adjust the order of mouse and cursor processing later?
* If not, what is the best way to go about restructuring the main loop to ensure mouse movements are processed before cursor updates?

andydotxyz:
I think the first of the two may be ok assuming the test fixes are pretty simple.
If the change can have a comment referencing the new issue tracking it would be good. I assume we can still check the desirable behaviour by triggering mouse move a second time?

andydotxyz (approved):
Thanks

PR #4326, opened 2023-10-17
Feature/mobile key codes

BogdanTheGeek:
### Description:
Add support for android gamepad(or tv remote) keycodes. These are basically required for apps running on android TV.

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/driver/mobile/key.go
+++ b/driver/mobile/key.go
@@ -5,7 +5,8 @@ import (
 )
 
 const (
-	// KeyBack represents the back button which may be hardware or software
+	//
+	// Since: 2.5
 	KeyBack       fyne.KeyName = "Back"
 	KeyDPadLeft   fyne.KeyName = "DPadLeft"
 	KeyDPadRight  fyne.KeyName = "DPadRight"


BogdanTheGeek:
For now, I have only added and tested the dpad keys to get some feedback.
I have tested these on an android phone with an xbox controller and on a fire tv stick with the included remote.

princessmortix:
+1 for this feature.

andydotxyz (changes requested):
Thanks so much for this, a great addition.
We should consider adding support for DPad into the focus manager so the apps are traversable on a TV device. But this PR enables such work so let's land it :).

I had wondered if this needed to be "gamepad support" in the API but I think this probably just falls to documentation.

Talking of which I think the costs in driver/mobile/key.go need appropriate documentation and it should include:
```go
//
// Since: 2.5
```

to mark when they were added. I think I missed this for KeyBack last release!

andydotxyz:
Thanks for looking at this - a great approach to starting gamepad support. I think it needs a bit of docs as noted inline but neat solution.

BogdanTheGeek:
> Thanks for looking at this - a great approach to starting gamepad support. I think it needs a bit of docs as noted inline but neat solution.

Should I also add the remaining supported keycodes(Gamepad, play/pause, camera etc?) to this PR?
I am not sure how these would be supported on other platforms like Linux, Mac, iOS. I assume, the same keycode to key name lookup is implemented for all other platforms. If pointed in the right direction, I can add them for Mac and Linux, the only other ones I can test on.

andydotxyz:
There are a few assumptions thrown in there. What is a camera or play/pause button on a desktop?
As it's in mobile right now all we need to test is Android and iOS (though I'm not sure how you hook up the dpad to an iOS device - but we should try and figure it out!)

BogdanTheGeek:
> There are a few assumptions thrown in there. What is a camera or play/pause button on a desktop? As it's in mobile right now all we need to test is Android and iOS (though I'm not sure how you hook up the dpad to an iOS device - but we should try and figure it out!)

you can connect any bluetooth remote or gamepad to any desktop or mobile device and get the same key events.

BogdanTheGeek:
@andydotxyz anything else required to merge this?

andydotxyz:
> you can connect any bluetooth remote or gamepad to any desktop or mobile device and get the same key events.

Then maybe this isn't mobile specific after all? If it's expected that this should work on all platforms then maybe landing it in mobile space was not the right approach - apologies if it was my assumption

> @andydotxyz anything else required to merge this?

I guess it needs to have complete support for whatever platforms should be supported. If it is mobile specific then the iOS driver looks to be missing code. If however it's not mobile specific (as in above message) then maybe it needs to be revised for working-anyway codes?

PR #4327, opened 2023-10-17
Update golang.or/x/sys to bring in security fix in x/net

Jacalz:
### Description:

This fixes https://github.com/fyne-io/fyne/security/dependabot/7 and https://github.com/fyne-io/fyne/security/dependabot/6. Nothing that immediately affected our use of the package but is nice to have updated because GitHub flags all downstream projects using Fyne as well. We might want o backport this to the release channel as well.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Thanks

Jacalz:
Merged. You can backport if you'd like :)

PR #4331, opened 2023-10-18
Run tests on Go 1.21 instead of 1.20

Jacalz:
### Description:

Makes sure that we can still land PRs for the current release once #4330 lands in develop.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Thanks

PR #4337, opened 2023-10-21
Adding triple tap to select current row for widget.Entry

### LINKED ISSUE #4328: Triple clicking in an Entry widget should select all text
### Checklist

This is fairly standard behavior across OSes and UI toolkits. We should aim to support it.

Yes, by extending Entry to override Tapped and implementing the behavior yourself

Triple clicking in the entry should select all text in a single line entry, or all of the current line in a multi-line entry

dweymouth:
### Description:

Fixes #4328 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/common/driver.go
+++ b/internal/driver/common/driver.go
@@ -5,6 +5,14 @@ import (
 	"fyne.io/fyne/v2/internal/cache"
 )
 
+const (
+	// moved here from the platform driver packages to allow
+	// referencing in widget package without an import cycle
+
+	DesktopDoubleClickDelay = 300 // ms
+	MobileDoubleClickDelay  = 500 // ms
+)
+
 // CanvasForObject returns the canvas for the specified object.
 func CanvasForObject(obj fyne.CanvasObject) fyne.Canvas {
 	return cache.GetCanvasForObject(obj)
--- a/internal/driver/common/driver.go
+++ b/internal/driver/common/driver.go
@@ -5,14 +5,6 @@ import (
 	"fyne.io/fyne/v2/internal/cache"
 )
 
-const (
-	// moved here from the platform driver packages to allow
-	// referencing in widget package without an import cycle
-
-	DesktopDoubleClickDelay = 300 // ms
-	MobileDoubleClickDelay  = 500 // ms
-)
-
 // CanvasForObject returns the canvas for the specified object.
 func CanvasForObject(obj fyne.CanvasObject) fyne.Canvas {
 	return cache.GetCanvasForObject(obj)
--- a/internal/driver/consts/consts.go
+++ b/internal/driver/consts/consts.go
@@ -1,9 +1,0 @@
-package consts
-
-const (
-	// moved here from the platform driver packages to allow
-	// referencing in widget package without an import cycle
-
-	DesktopDoubleClickDelay = 300 // ms
-	MobileDoubleClickDelay  = 500 // ms
-)
--- a/internal/driver/consts/consts.go
+++ b/internal/driver/consts/consts.go
@@ -1,0 +1,9 @@
+package consts
+
+const (
+	// moved here from the platform driver packages to allow
+	// referencing in widget package without an import cycle
+
+	DesktopDoubleClickDelay = 300 // ms
+	MobileDoubleClickDelay  = 500 // ms
+)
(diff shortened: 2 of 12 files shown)


widget/entry.go
```diff
@@ -21,6 +21,9 @@ import (
 const (
 	bindIgnoreDelay = time.Millisecond * 100 // ignore incoming DataItem fire after we have called Set
 	multiLineRows   = 3
+
+	desktopDoubleClickDelay = 300 // ms - keep in sync with internal/driver/glfw/window.go
```
andydotxyz, line 25: Can't we access the driver constants from here to save the need to keep them in sync manually?

andydotxyz: As it's buried in the internal package it doesn't really make much difference :) I'd probably just export them where they are for now, I don't think there is a need for a new package - and we don't really add them unless needed or much cleaner.

dweymouth:
Looks like there's still an import cycle in test code. I'll take a look in a bit

internal/driver/consts/consts.go
Jacalz, line 1: I wonder if it would be better to just have these in the `internal/driver` package, similar to how we do with the `internal/` package? It would also avoid you having to (for clarity obviously) change the name to `driverconsts`.

Jacalz (dismissed):
This is wonderful. What a difference in usability! Thanks :D
Just left one comment on the placement of the consts :)

internal/driver/consts/consts.go
dweymouth: Finding a place to put these without causing import cycles is a challenge! I think I had it there originally but it caused an import cycle. There is still an unresolved import cycle so they will need to be moved somewhere else yet again but I haven't looked into where

Jacalz: Hmm. Yeah, that's problematic. I'm not a huge fan of package names that have to be changed for clarity when imported but I don't know a better way really :/

internal/driver/common/driver.go
```diff
@@ -5,6 +5,14 @@ import (
 	"fyne.io/fyne/v2/internal/cache"
 )
 
+const (
+	// moved here from the platform driver packages to allow
+	// referencing in widget package without an import cycle
+
+	DesktopDoubleClickDelay = 300 // ms
```
andydotxyz: This probably could be exposed just as DoubleClickDelay with build constraints setting it appropriately.
Just a thought, it would take the logic out of the Entry to switch based on platform

dweymouth: How about making it a driver public API? `driver.DoubleClickDelay() time.Duration` So if other 3rd party widgets needed to have triple click - or to emulate DoubleTapped with Tapped + time comparisons (as I need to do in Supersonic to have list rows instantly tappable while also being DoubleTappable), then they could be consistent with the builtin ones

dweymouth: Of course in the meantime until 2.5 we still need a solution for Entry if we want to get this in a 2.4.x point release. But if we have a long-term plan we could just duplicate the constant 🤷‍♂️

andydotxyz: It's in the 2.5 milestone not 2.4.x currently. We don't usually add features in point release and I'm not sure we could list this as a bug fix ;)

dweymouth: Since this is targeted for 2.5 anyway, I reworked with a new public API DoubleTapDelay on the driver interface. This also allows for 3rd party widgets to be able to match double click delay for whatever custom logic they may have. And is also future compatible for getting the value from the OS instead of having it be a compile-time constant, for OSes that provide an API.

Jacalz:
I wonder, should we expose an interface for making a widget triple tappable instead of exposing the double tap delay?

dweymouth:
We figured that Entry was probably the only widget that needed triple tap. Also personally I'd like the double tap delay to be exposed since I have my list rows in Supersonic implement their own double tap logic (so that tap can instantly trigger to select the row of a tracklist while double tap can play a track). Having a widget have both Tappable and DoubleTappable means, by design, it won't get the tap event until after the delay, which is not what I wanted in my case.

andydotxyz:
> Edit: either that - or a way to request for a specific widget that the Tapped callback should be invoked immediately even if it may become a DoubleTap if tapped again

This is tough to be convinced of - as `Tapped` is an abstract request for the tap event, as you know we have lower level events for when the widget wants to do smarter things. Double tapping, and indeed Triple, are not things that most widgets will be doing (I hope!) as there is no discoverability to these actions.

PR #4347, opened 2023-10-28
Add the ability to disable ToolbarActions #2306

maruu:
### Description:
Added ability to disable ToolbarActions.
I chose to embed the button directly, since it already handles the disable-state logic.
Also, since the button is returned as a pointer, no separate refreshing logic is required. This seems to have been an issue with previous PRs for this issue.

Please let me know if the "Since:" comment with version 2.5 is fine, or if you prefer another version.

I don't necessarily like that the state of Icon & OnActivated are now essentially duplicated (once in the ToolbarAction and once in the underlying Button itself), but I did not want to break backward compatibility by changing the public interface. Let me know what your thoughts are on this.

Fixes https://github.com/fyne-io/fyne/issues/2306

I have tested the code using the following addition to the fyne-demo (`cmd/fyne_demo/tutorials/widget.go`):
```
func makeToolbarTab(_ fyne.Window) fyne.CanvasObject {
	paste := widget.NewToolbarAction(theme.ContentPasteIcon(), func() { fmt.Println("Paste") })

	t := widget.NewToolbar(widget.NewToolbarAction(theme.MailComposeIcon(), func() { fmt.Println("New") }),
		widget.NewToolbarSeparator(),
		widget.NewToolbarSpacer(),
		widget.NewToolbarAction(theme.ConfirmIcon(), func() {
			if paste.Disabled() {
				paste.Enable()
			} else {
				paste.Disable()
			}
		}),
		widget.NewToolbarSpacer(),
		widget.NewToolbarAction(theme.ContentCutIcon(), func() { fmt.Println("Cut") }),
		widget.NewToolbarAction(theme.ContentCopyIcon(), func() { fmt.Println("Copy") }),
		paste,
	)

	return container.NewBorder(t, nil, nil, nil)
}
```

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/toolbar.go
+++ b/widget/toolbar.go
@@ -36,16 +36,22 @@ func (t *ToolbarAction) SetIcon(icon fyne.Resource) {
 }
 
 // Enable this ToolbarAction, updating any style or features appropriately.
+//
+// Since: 2.5
 func (t *ToolbarAction) Enable() {
 	t.button.Enable()
 }
 
 // Disable this ToolbarAction so that it cannot be interacted with, updating any style appropriately.
+//
+// Since: 2.5
 func (t *ToolbarAction) Disable() {
 	t.button.Disable()
 }
 
 // Disabled returns true if this ToolbarAction is currently disabled or false if it can currently be interacted with.
+//
+// Since: 2.5
 func (t *ToolbarAction) Disabled() bool {
 	return t.button.Disabled()
 }
--- a/widget/toolbar.go
+++ b/widget/toolbar.go
@@ -22,6 +22,7 @@ type ToolbarAction struct {
 
 // ToolbarObject gets a button to render this ToolbarAction
 func (t *ToolbarAction) ToolbarObject() fyne.CanvasObject {
+	t.button.Icon = t.Icon
 	t.button.OnTapped = t.OnActivated
 	return t.button
 }
--- a/widget/toolbar.go
+++ b/widget/toolbar.go
@@ -13,42 +13,17 @@ type ToolbarItem interface {
 	ToolbarObject() fyne.CanvasObject
 }
 
-type toolbarActionButton struct {
-	Button
-
-	t *ToolbarAction
-}
-
-func newToolbarActionButton(t *ToolbarAction) *toolbarActionButton {
-	b := &toolbarActionButton{t: t}
-	b.ExtendBaseWidget(b)
-	return b
-}
-
-func (b *toolbarActionButton) Refresh() {
-	b.Icon = b.t.Icon
-	b.OnTapped = b.t.OnActivated
-
-	if b.t.Disabled {
-		b.Button.Disable()
-	} else {
-		b.Button.Enable()
-	}
-}
-
 // ToolbarAction is push button style of ToolbarItem
 type ToolbarAction struct {
 	Icon        fyne.Resource
 	OnActivated func() `json:"-"`
-	// Since: 2.5
-	Disabled bool
+	button      *Button
 }
 
 // ToolbarObject gets a button to render this ToolbarAction
 func (t *ToolbarAction) ToolbarObject() fyne.CanvasObject {
-	b := newToolbarActionButton(t)
-	b.Refresh()
-	return b
+	t.button.OnTapped = t.OnActivated
+	return t.button
 }
 
 // SetIcon updates the icon on a ToolbarItem
@@ -56,11 +31,37 @@ func (t *ToolbarAction) ToolbarObject() fyne.CanvasObject {
 // Since: 2.2
 func (t *ToolbarAction) SetIcon(icon fyne.Resource) {
 	t.Icon = icon
+	t.button.SetIcon(t.Icon)
+	t.button.Refresh()
+}
+
+// Enable this ToolbarAction, updating any style or features appropriately.
+//
+// Since: 2.5
+func (t *ToolbarAction) Enable() {
+	t.button.Enable()
+}
+
+// Disable this ToolbarAction so that it cannot be interacted with, updating any style appropriately.
+//
+// Since: 2.5
+func (t *ToolbarAction) Disable() {
+	t.button.Disable()
+}
+
+// Disabled returns true if this ToolbarAction is currently disabled or false if it can currently be interacted with.
+//
+// Since: 2.5
+func (t *ToolbarAction) Disabled() bool {
+	return t.button.Disabled()
 }
 
 // NewToolbarAction returns a new push button style ToolbarItem
 func NewToolbarAction(icon fyne.Resource, onActivated func()) *ToolbarAction {
-	return &ToolbarAction{Icon: icon, OnActivated: onActivated}
+	button := NewButtonWithIcon("", icon, onActivated)
+	button.Importance = LowImportance
+
+	return &ToolbarAction{Icon:icon, OnActivated: onActivated, button: button}
 }
 
 // ToolbarSpacer is a blank, stretchable space for a toolbar.
--- a/widget/toolbar.go
+++ b/widget/toolbar.go
@@ -32,7 +32,6 @@ func (t *ToolbarAction) ToolbarObject() fyne.CanvasObject {
 func (t *ToolbarAction) SetIcon(icon fyne.Resource) {
 	t.Icon = icon
 	t.button.SetIcon(t.Icon)
-	t.button.Refresh()
 }
 
 // Enable this ToolbarAction, updating any style or features appropriately.
--- a/widget/toolbar.go
+++ b/widget/toolbar.go
@@ -17,14 +17,13 @@ type ToolbarItem interface {
 type ToolbarAction struct {
 	Icon        fyne.Resource
 	OnActivated func() `json:"-"`
+	button      *Button
 }
 
 // ToolbarObject gets a button to render this ToolbarAction
 func (t *ToolbarAction) ToolbarObject() fyne.CanvasObject {
-	button := NewButtonWithIcon("", t.Icon, t.OnActivated)
-	button.Importance = LowImportance
-
-	return button
+	t.button.OnTapped = t.OnActivated
+	return t.button
 }
 
 // SetIcon updates the icon on a ToolbarItem
@@ -32,12 +31,31 @@ func (t *ToolbarAction) ToolbarObject() fyne.CanvasObject {
 // Since: 2.2
 func (t *ToolbarAction) SetIcon(icon fyne.Resource) {
 	t.Icon = icon
-	t.ToolbarObject().Refresh()
+	t.button.SetIcon(t.Icon)
+	t.button.Refresh()
+}
+
+// Enable this ToolbarAction, updating any style or features appropriately.
+func (t *ToolbarAction) Enable() {
+	t.button.Enable()
+}
+
+// Disable this ToolbarAction so that it cannot be interacted with, updating any style appropriately.
+func (t *ToolbarAction) Disable() {
+	t.button.Disable()
+}
+
+// Disabled returns true if this ToolbarAction is currently disabled or false if it can currently be interacted with.
+func (t *ToolbarAction) Disabled() bool {
+	return t.button.Disabled()
 }
 
 // NewToolbarAction returns a new push button style ToolbarItem
 func NewToolbarAction(icon fyne.Resource, onActivated func()) *ToolbarAction {
-	return &ToolbarAction{icon, onActivated}
+	button := NewButtonWithIcon("", icon, onActivated)
+	button.Importance = LowImportance
+
+	return &ToolbarAction{Icon:icon, OnActivated: onActivated, button: button}
 }
 
 // ToolbarSpacer is a blank, stretchable space for a toolbar.
--- a/widget/toolbar.go
+++ b/widget/toolbar.go
@@ -19,6 +19,12 @@ type toolbarActionButton struct {
 	t *ToolbarAction
 }
 
+func newToolbarActionButton(t *ToolbarAction) *toolbarActionButton {
+	b := &toolbarActionButton{t: t}
+	b.ExtendBaseWidget(b)
+	return b
+}
+
 func (b *toolbarActionButton) Refresh() {
 	b.Icon = b.t.Icon
 	b.OnTapped = b.t.OnActivated
@@ -40,7 +46,7 @@ type ToolbarAction struct {
 
 // ToolbarObject gets a button to render this ToolbarAction
 func (t *ToolbarAction) ToolbarObject() fyne.CanvasObject {
-	b := &toolbarActionButton{t: t}
+	b := newToolbarActionButton(t)
 	b.Refresh()
 	return b
 }
--- a/widget/toolbar.go
+++ b/widget/toolbar.go
@@ -17,22 +17,16 @@ type ToolbarItem interface {
 type ToolbarAction struct {
 	Icon        fyne.Resource
 	OnActivated func() `json:"-"`
-	button      *Button
+	button      Button
 }
 
 // ToolbarObject gets a button to render this ToolbarAction
 func (t *ToolbarAction) ToolbarObject() fyne.CanvasObject {
-	if t.button == nil {
-		// lazy initialization of the button, when the ToolbarAction 
-		// has not been created via the constructor
-		t.button = createToolbarActionButton(t.Icon, t.OnActivated)
-	} else {
-		// synchronize properties
-		t.button.Icon = t.Icon
-		t.button.OnTapped = t.OnActivated
-	}
-
-	return t.button
+	// synchronize properties
+	t.button.Icon = t.Icon
+	t.button.OnTapped = t.OnActivated
+	
+	return &t.button
 }
 
 // SetIcon updates the icon on a ToolbarItem
@@ -64,16 +58,10 @@ func (t *ToolbarAction) Disabled() bool {
 	return t.button.Disabled()
 }
 
-func createToolbarActionButton(icon fyne.Resource, onActivated func()) *Button {
-	button := NewButtonWithIcon("", icon, onActivated)
-	button.Importance = LowImportance
-
-	return button
-}
 
 // NewToolbarAction returns a new push button style ToolbarItem
 func NewToolbarAction(icon fyne.Resource, onActivated func()) *ToolbarAction {
-	return &ToolbarAction{Icon: icon, OnActivated: onActivated, button: createToolbarActionButton(icon, onActivated)}
+	return &ToolbarAction{Icon: icon, OnActivated: onActivated}
 }
 
 // ToolbarSpacer is a blank, stretchable space for a toolbar.
--- a/widget/toolbar.go
+++ b/widget/toolbar.go
@@ -22,8 +22,16 @@ type ToolbarAction struct {
 
 // ToolbarObject gets a button to render this ToolbarAction
 func (t *ToolbarAction) ToolbarObject() fyne.CanvasObject {
-	t.button.Icon = t.Icon
-	t.button.OnTapped = t.OnActivated
+	if t.button == nil {
+		// lazy initialization of the button, when the ToolbarAction 
+		// has not been created via the constructor
+		t.button = createToolbarActionButton(t.Icon, t.OnActivated)
+	} else {
+		// synchronize properties
+		t.button.Icon = t.Icon
+		t.button.OnTapped = t.OnActivated
+	}
+
 	return t.button
 }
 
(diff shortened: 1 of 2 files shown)


widget/toolbar.go
```diff
@@ -17,27 +17,51 @@ type ToolbarItem interface {
...
+//
+// Since: 2.5
+func (t *ToolbarAction) Disabled() bool {
+	return t.button.Disabled()
 }
 
 // NewToolbarAction returns a new push button style ToolbarItem
 func NewToolbarAction(icon fyne.Resource, onActivated func()) *ToolbarAction {
-	return &ToolbarAction{icon, onActivated}
+	button := NewButtonWithIcon("", icon, onActivated)
```
dweymouth, line 61: The Fyne APIs are designed to support initialization using the constructor or using the struct directly (mostly for extending widgets), so you can't assume the button will have been initialized the first time that SetIcon/Enable/etc is called - so perhaps add a private helper function `ensureButton` that will construct the button if it is nil.

widget/toolbar.go
```diff
@@ -17,27 +17,51 @@ type ToolbarItem interface {
 type ToolbarAction struct {
 	Icon        fyne.Resource
 	OnActivated func() `json:"-"`
+	button      *Button
```
dweymouth, line 20: I'm not sure how I feel about embedding button directly here. What I might prefer is adding a `Disabled  bool` here, and creating a private `toolbarActionButton` type here that extends `widget.Button`, holds a reference to the parent `ToolbarAction`, and in its overridden `Refresh` function, syncs the state from the `ToolbarAction` to the button, and then call `BaseWidget.Refresh()`.

Then, in `ToolbarObject()`, create and return a new `toolbarActionButton` that has the reference to `t`.

maruu: Hi, thanks for reviewing the PR so quickly.

I think I got your point about the private toolbarActionButton, although I am not entirely sure why you prefer this option above using the button type directly as a member of the ToolbarAction type. Could you please elaborate more on this decision?

widget/toolbar.go
maruu: Understood.

Another option could also be to not save a pointer to the button, but the button itself, right? If the ToolbarAction is created via struct initializer directly, this way the button would automatically be in a valid state. Of course, if we want to keep the button embedded into the ToolbarAction type, then the state of the ToolbarAction would need to be synchronized in the ToolbarObject() call (since if the Icon is initialized via the struct initialization, currently it is not applied to the Button)

widget/toolbar.go
dweymouth: The Fyne APIs are designed to allow changing properties through setter functions (eg `Disable()`) or properties (`wid.Disabled = true`). With embedding the button directly the Disable property will not be a part of the API for ToolbarAction. Also we try to separate the state from the renderer - see the separation between `Widget` and `WidgetRenderer` in the codebase. The button is what's used to render the toolbar action, not really an intrinsic property of the action.

maruu: > The button is what's used to render the toolbar action, not really an intrinsic property of the action.
Got it, thank you.

I implemented the suggested `toolbarActionButton`. I intentionally used the `Disabled` property, since this is also what the `MenuItem` in `menu.go` uses, and I wanted to stay consistent here.
What I somehow dislike about the property approach is, that it requires an explicit refresh, which would not be the case if one would use the function-API (`Enable()`/`Disable()`) instead of the property (`Disabled`), because the refresh could be hidden in the function-API.

Let me know what you think.

widget/toolbar.go
```diff
@@ -13,31 +13,48 @@ type ToolbarItem interface {
...
+
+	t *ToolbarAction
+}
+
+func (b *toolbarActionButton) Refresh() {
+	b.Icon = b.t.Icon
+	b.OnTapped = b.t.OnActivated
+
+	if b.t.Disabled {
+		b.Button.Disable()
```
maruu, line 27: Note that this calls Refresh implicitly, so I avoided calling it explicitly again after.

maruu:
I also experience some weird rendering behavior in the fyne_demo after making my changes. After a refresh, the alignment of the toolbar items seems to be off. I am not sure if this a local bug for me, or if it is caused by my changes:

Before:
grafik

After refresh:
grafik

The code that I am using is

```
func makeToolbarTab(_ fyne.Window) fyne.CanvasObject {
	paste := widget.NewToolbarAction(theme.ContentPasteIcon(), func() { fmt.Println("Paste") })

	var t *widget.Toolbar

	t = widget.NewToolbar(widget.NewToolbarAction(theme.MailComposeIcon(), func() { fmt.Println("New") }),
		widget.NewToolbarSeparator(),
		widget.NewToolbarSpacer(),
		widget.NewToolbarAction(theme.ConfirmIcon(), func() {
			if paste.Disabled {
				paste.Disabled = false
			} else {
				paste.Disabled = true
			}
			t.Refresh()
		}),
		widget.NewToolbarSpacer(),
		widget.NewToolbarAction(theme.ContentCutIcon(), func() { fmt.Println("Cut") }),
		widget.NewToolbarAction(theme.ContentCopyIcon(), func() { fmt.Println("Copy") }),
		paste,
	)

	return container.NewBorder(t, nil, nil, nil)
}
```

Could you try it on your side and let me know if it happens as well on your machine?

andydotxyz:
Wasn't  implementing the disablable interface desirable though? Maybe it isn't possible as the toolbar action is not really a widget?

maruu:
> Wasn't implementing the disablable interface desirable though? Maybe it isn't possible as the toolbar action is not really a widget?

I intentionally used the Disabled property, since this is also what the `MenuItem` in `menu.go` uses, and I wanted to stay consistent here.

I am not sure how you handled such situations in the past. Let me know if I should change the implementation, or how you want to continue.

widget/toolbar.go
```diff
@@ -13,31 +13,48 @@ type ToolbarItem interface {
...
 }
 
 // ToolbarObject gets a button to render this ToolbarAction
 func (t *ToolbarAction) ToolbarObject() fyne.CanvasObject {
-	button := NewButtonWithIcon("", t.Icon, t.OnActivated)
-	button.Importance = LowImportance
-
-	return button
+	b := &toolbarActionButton{t: t}
+	b.Refresh()
```
dweymouth, line 44: Is the Refresh here necessary? I don't think so since this function should only be called to get a new Button to render the toolbar action, which is immediately placed on the canvas for display which will cause it to get refreshed automatically?

widget/toolbar.go
```diff
@@ -13,31 +13,48 @@ type ToolbarItem interface {
 	ToolbarObject() fyne.CanvasObject
 }
 
+type toolbarActionButton struct {
```
dweymouth, line 16: I think you may need to create a constructor for this and call `ExtendBaseWidget` in the constructor

widget/toolbar.go
maruu: The toolbarRenderer `Refresh()` code does not call the `Refresh()` function of the underlying objects, but only creates new ones using `ToolbarObject()`. Since the underlying `Button` of the `toolbarActionButton` is only updated with the `Refresh()` call, which is not happening by the `toolbarRenderer`, I chose to implement it here. Alternatively, I can also add a `Refresh()` call in the `toolbarRenderer`:

https://github.com/fyne-io/fyne/blob/ccc6a1c19a6d16988da3e468f08f25d8100cd4b4/widget/toolbar.go#L145

Let me know what you prefer.

maruu:
By the way, what is the reasoning behind not making the toolbarItems widgets?
According to my current understanding, we would not be having these issues if the toolbarItems were normal widgets, right?
For example, a `menuItem` also is a widget: 
https://github.com/fyne-io/fyne/blob/ccc6a1c19a6d16988da3e468f08f25d8100cd4b4/widget/menu_item.go#L42

widget/toolbar.go
maruu: Thanks, that fixed my rendering issue! Added a commit.

maruu:
Rebased branch to current develop state

andydotxyz:
> By the way, what is the reasoning behind not making the toolbarItems widgets?
> 
> According to my current understanding, we would not be having these issues if the toolbarItems were normal widgets, right?
> 
> For example, a `menuItem` also is a widget: 
> 
> https://github.com/fyne-io/fyne/blob/ccc6a1c19a6d16988da3e468f08f25d8100cd4b4/widget/menu_item.go#L42

It's about the public API. Internally menuItem is a widget, but the public api fyne.MenuItem is not. Using public APIs that are abstractions allows us to keep the widgets consistent.

A container will give the user control over its content through adding arbitrary widgets, but it is rare for a widget to do this. If you want a "toolbar that accepts anything" then wouldn't that just be a VBox with a background rectangle?

That said I don't know what the issue is here, the ToolbarObject returns a CanvasObject for its "implementation detail" and this can be a widget, disablable widget or anything else. So why would this be easier if the main object was a widget?

widget/toolbar.go
andydotxyz: Typically widgets are created with the correct state so that a Refresh is not required before they first draw.

> The toolbarRenderer Refresh() code does not call the Refresh() function of the underlying objects, but only creates new ones using ToolbarObject()

Given this information, and that as far as I can see the only way to ask this item to refresh is to call `ToolBar.Refresh()` once the item is `Disabled`, is the custom type and additional refresh paths needed?

widget/toolbar_test.go
```diff
@@ -90,3 +90,50 @@ type toolbarLabel struct {
...
+
+func TestToolbarAction_Disable(t *testing.T) {
+	testIcon := theme.InfoIcon()
+	toolbarAction := NewToolbarAction(testIcon, nil)
+	object := toolbarAction.ToolbarObject().(*toolbarActionButton)
+
+	assert.False(t, object.Disabled())
+
+	toolbarAction.Disabled = true
+	object.Refresh()
```
andydotxyz, line 102: I don't think end users of the API have access to this call.
How will they refresh?

andydotxyz (changes requested):
I can't quite see how the disabled item is refreshed without internal API access.
I try to give more detail inline - let me know if it is not clear.

widget/toolbar.go
```diff
@@ -13,31 +13,54 @@ type ToolbarItem interface {
...
+	b.Refresh()
+	return b
 }
 
 // SetIcon updates the icon on a ToolbarItem
 //
 // Since: 2.2
 func (t *ToolbarAction) SetIcon(icon fyne.Resource) {
 	t.Icon = icon
-	t.ToolbarObject().Refresh()
```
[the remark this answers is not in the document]
maruu: You might be right - I think the expected behavior for SetIcon in the fyne API is to call Refresh internally, and not wait for the user to call refresh on the toolbar, right?

maruu:
Any idea why the tests are always failing with:

```
D:\a\fyne\setup-go-faster\go\1.17.13\x64\pkg\tool\windows_amd64\link.exe: running gcc failed: exit status 1
C:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\Users\RUNNER~1\AppData\Local\Temp\go-link-3164744842\000008.o: in function `_cgo_preinit_init':
c:\go\src\runtime\cgo/gcc_libinit_windows.c:30: undefined reference to `__imp___iob_func'
C:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\Users\RUNNER~1\AppData\Local\Temp\go-link-3164744842\000008.o: in function `x_cgo_sys_thread_create':
c:\go\src\runtime\cgo/gcc_libinit_windows.c:60: undefined reference to `__imp___iob_func'
C:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\Users\RUNNER~1\AppData\Local\Temp\go-link-3164744842\000008.o: in function `x_cgo_notify_runtime_init_done':
c:\go\src\runtime\cgo/gcc_libinit_windows.c:101: undefined reference to `__imp___iob_func'
C:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\Users\RUNNER~1\AppData\Local\Temp\go-link-3164744842\000009.o: in function `x_cgo_thread_start':
c:\go\src\runtime\cgo/gcc_util.c:18: undefined reference to `__imp___iob_func'
C:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\Users\RUNNER~1\AppData\Local\Temp\go-link-3164744842\000010.o: in function `_cgo_sys_thread_start':
c:\go\src\runtime\cgo/gcc_windows_amd64.c:31: undefined reference to `__imp___iob_func'
collect2.exe: error: ld returned 1 exit status
```
? I don't see any connection to the code that I have changed.

Jacalz:
@maruu It is not your fault. It’s a bug in versions of Go older than 1.20. We need to look at how we should address that on our next contributor meeting for Fyne.

widget/toolbar.go
andydotxyz: This call to refresh would not be needed if the disabled state is set correctly when you create the button - that would match how the code works elsewhere.

widget/toolbar.go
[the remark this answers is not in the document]
andydotxyz: Refresh should be called in a setter - so the state is drawn.
However calling it on the result of `ToolbarObject` won't do much as that is a new instance.

This line should be fixed not removed - feel free to address it or to leave it for now (but don't remove as that is also wrong).

widget/toolbar_test.go
andydotxyz: I agree - can you use public APIs in this test so we can be sure it's working for end users and not just in unit tests?

andydotxyz (changes requested):
More thoughts inline - good to have the feature working though.

widget/toolbar.go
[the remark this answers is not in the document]
maruu: @andydotxyz You are right. At the moment, the only way to change anything in the `ToolbarAction` and have it visible in the `Toolbar` that the `ToolbarAction` has been assigned to is by calling the `Toolbar's` `Refresh` function. This will then re-render the `ToolbarAction` with the new state. This is a result of the change proposed here: https://github.com/fyne-io/fyne/pull/4347#discussion_r1375256904. Before that, I used to save a reference to the `Button` which is drawn on the screen in the `ToolbarAction`, which made it possible to update the state of the `Button` directly via the `ToolbarAction's` setters & properties.

With the current implementation I currently do not see any way to refresh the `ToolbarAction` if it has already been added to a `Toolbar`, other than calling the `Toolbar's` `Refresh` function. If we want to be able to update the `ToolbarAction` and immediately have the state reflected in the `Toolbars`, then I think it is necessary to save a reference to the rendered `Button` in the `ToolbarAction`, but this would go against the proposed change from https://github.com/fyne-io/fyne/pull/4347#discussion_r1375256904.

Let me know which approach you prefer so I can implement it.

widget/toolbar.go
maruu: I think this would only move the call to `Refresh` into the constructor of the `toolbarActionButton` as I also need to make sure that the `Icon` & the `OnTapped` handler are set correctly.

widget/toolbar.go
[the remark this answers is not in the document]
andydotxyz: That is a good point, perhaps @dweymouth can comment so I don't contradict their thoughts.

widget/toolbar.go
andydotxyz: What I am thinking is no refresh should be needed here. Setting the disabled state on creation would avoid it would it not?

maruu: Are you thinking about something like this?

```
func newToolbarActionButton(t *ToolbarAction) *toolbarActionButton {
	b := &toolbarActionButton{t: t}
	b.ExtendBaseWidget(b)

	b.Icon = t.Icon
	b.OnTapped = b.t.OnActivated
	b.disabled = b.t.Disabled

	return b
}
```

Note that I also need to set the `Icon` & `OnTapped` handler here, since we are not calling `Refresh` afterwards, and therefore these fields would not be set.

Having this code though results in a nasty duplication in my opinion, because the same fields are set when the state is synchronized with the `ToolbarAction`

```
func newToolbarActionButton(t *ToolbarAction) *toolbarActionButton {
	b := &toolbarActionButton{t: t}
	b.ExtendBaseWidget(b)

	b.Icon = t.Icon
	b.OnTapped = b.t.OnActivated
	b.disabled = b.t.Disabled

	return b
}

func (b *toolbarActionButton) Refresh() {
	b.Icon = b.t.Icon
	b.OnTapped = b.t.OnActivated

	if b.t.Disabled {
		b.Button.Disable()
	} else {
		b.Button.Enable()
	}
}
```

I would propose to wait for the results of our discussion in https://github.com/fyne-io/fyne/pull/4347#discussion_r1439335003, as this could change this affected code part.

widget/toolbar.go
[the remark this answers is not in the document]
dweymouth: I think embedding the button is fine in this case as we have a reason to do so. My original thought was that we'd want a publicly settable `Disabled bool` field, but I see that I was mistaken as all the Disableable widgets do not export a public property- you only interact through the getters and setters.

maruu:
Hi @andydotxyz @dweymouth, as discussed here: https://github.com/fyne-io/fyne/pull/4347#discussion_r1440613377 I changed the PR to now use the Button directly. The benefits from my point of view are that the state management is not duplicated, but can be reused from the Button itself.

I again tested the functionality with an extension of the fyne_demo app (see my initial PR description for the code if you want to give it a try). Let me know what you think.

maruu:
Rebased branch to current develop state

andydotxyz (changes requested):
Looking good, but this will crash if you just create a &ToolbarAction{} directly as the constructor function is optional.

maruu:
> Looking good, but this will crash if you just create a &ToolbarAction{} directly as the constructor function is optional.

You are totally correct. Since I cannot store a reference to the button directly (because it cannot be copied) I chose to use lazy initialization. I also added a test for this, hope that's okay. Let me know what you think.

widget/toolbar.go
```diff
@@ -22,8 +22,16 @@ type ToolbarAction struct {
...
-	t.button.OnTapped = t.OnActivated
+	if t.button == nil {
+		// lazy initialization of the button, when the ToolbarAction 
+		// has not been created via the constructor
+		t.button = createToolbarActionButton(t.Icon, t.OnActivated)
+	} else {
+		// synchronize properties
+		t.button.Icon = t.Icon
+		t.button.OnTapped = t.OnActivated
+	}
```
maruu, line 33: I was unsure about this, as it violates the happy-path-left pattern. But I also did not want to duplicate the `return t.button` line. Let me know what you think/prefer.

maruu: I think there is a case that would still generate a problem though. If I would keep it like it is now, and create the `ToolbarAction` without the constructor, and then would use any of the setters, there would still be a crash. So I would need to pull the lazy initialization in all of the setters as well, or synchronize the state once in the `ToolbarObject` function, after lazy initializing it there. But this would require a property for the Disabled state again in the `ToolbarAction` so I can also synchronize it after the lazy initialization in the`ToolbarObject` function....

Sigh... am I missing something obvious/simple here?

andydotxyz: The code already has this problem I realised - the constructor is optional and there is no guarantee that `ToolbarObject` is called before you call `Disable`.

It's possible that the easiest fix here is to embed `Button` instead of `*Button` so you know it is always initialised. Then it's just a case of ensuring the state of the embedded button is correct?

widget/toolbar.go
```diff
@@ -17,27 +17,63 @@ type ToolbarItem interface {
 type ToolbarAction struct {
 	Icon        fyne.Resource
 	OnActivated func() `json:"-"`
+	button      *Button
```
Jacalz, line 20: You can change this field to not be a pointer. Saves you having to do a nil check. It also avoid needing to construct it manually. Just set the fields you want. Just return a reference to the field in places where it wants it to be a pointer.

widget/toolbar.go
Jacalz: My comment above likely means that you can avoid the nil-check here.

widget/toolbar.go
```diff
@@ -17,27 +17,63 @@ type ToolbarItem interface {
...
+func (t *ToolbarAction) Disabled() bool {
+	return t.button.Disabled()
+}
+
+func createToolbarActionButton(icon fyne.Resource, onActivated func()) *Button {
+	button := NewButtonWithIcon("", icon, onActivated)
+	button.Importance = LowImportance
+
+	return button
 }
```
Jacalz, line 72: This constructor should no longer be necessary with my comment above.

Jacalz (changes requested):
Thanks for working on this. I have a few suggestions that will clean up the code a bit once resolved

widget/toolbar.go
maruu: Thanks, this simplifies the code a lot.

Jacalz:
Please resolve the issues reported by the static analysis workflow

andydotxyz (approved):
Well done and thanks, this is a great addition

Jacalz (approved):

Issue #4301, opened 2023-10-06
WASM Copy / Paste


rmasci:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

Tried running several fyne apps using wasm. The apps all run, you see the window in the browser, but copy / paste is not available nor can you open files, save files etc. 

### How to reproduce

1. Simple App, Code pasted in 'Example Code' section
2. Run fyne serve --port 8080 --target wasm
3. In browser, try to copy / paste. I can't select any text I type, I can't copy it to the clipboard and if I have text on the clipboard I can't paste it in. 
4. I've tried more complex applications, none of them seem to work. Tried in Windows / Mac, browsers: Edge, Firefox, Safari

### Example code

``` go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Test")
	w.SetContent(widget.NewMultiLineEntry())
	w.Resize(fyne.Size{Height: 640, Width: 480})
	w.ShowAndRun()
}
```

### Fyne version

fyne.io/fyne/v2 v2.4.0

### Go compiler version

go version go1.21.1

### Operating system and version

Windows 11, Mac Ventura M2

### Additional Information

Every app I try, including apps found on https://apps.fyne.io/ seem to work with copy / paste. I also noticed trying to open a file / save a file doesn't work either. All this works on a desktop app without a problem, just not in wasm.

Bluebugs:
Indeed, this is currently not implemented in the wasm and the js backend. The change actually have to be done in the glfw-js module at https://github.com/fyne-io/glfw-js/blob/master/browser.go#L925 and https://github.com/fyne-io/glfw-js/blob/master/browser_wasm.go#L941 . I currently do not have the time to work on this, but it should be a possible task for someone beginning to look at fyne code base and its dependencies.

sujalcodes3:
I would love to contribute to this issue, but I am a beginner. Can you please suggest me where to start?

Bluebugs:
[...]

PR #4348, opened 2023-10-29
Mac based browser shortcut fix

gcottom:
### Description:
previously when in the browser on a mac you would have to use the ctrl key to perform keyboard shortcuts which is not intuitive on a mac where cmd is typically used instead of ctrl. this change adds a check in window initialization that will check if the build is browser based (based on build constraints). At runtime there is a javascript check of the navigator.platform value. if this value contains mac, the modifier key handles default shortcut keys with cmd key instead of ctrl key. other operating systems will see no change. Note that this does not change custom defined shortcuts. If you define a shortcut as "ctrl+f" it will be interpreted as ctrl+f meaning that even on a mac you will have to use the ctrl key. This would be in line with the expectation that if you were to run a browser based terminal, that ctrl+c wouldn't copy text but would kill the process. 

Fixes #(#4301)
This is part 2 of a 2 part fix. Part 1 is in fyne/glfw-js [Pull Request 13](https://github.com/fyne-io/glfw-js/pull/13). Part 1 addresses the clipboard issue in the browser #4301 . Part 2 which is this PR, addresses keyboard shortcuts not working correctly in Mac browsers which was discovered during the testing of the fix created in part 1. 

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

@andydotxyz this is the solution I proposed in [glfw-js Pull Request 14](https://github.com/fyne-io/glfw-js/pull/14). I will now close that pull request.

Thank you.


internal/driver/glfw/window.go
```diff
@@ -832,7 +832,7 @@ func (w *window) processFocused(focus bool) {
 func (w *window) triggersShortcut(localizedKeyName fyne.KeyName, key fyne.KeyName, modifier fyne.KeyModifier) bool {
 	var shortcut fyne.Shortcut
 	ctrlMod := fyne.KeyModifierControl
-	if runtime.GOOS == "darwin" {
+	if runtime.GOOS == "darwin" || isMacOSBrowser() {
```
andydotxyz, line 835: This could be simplified - put the GOOS check in place if "false" in the shortcuts_other and call it "isMacOSRuntime instead.
It will also fix the issue where you are running the server on macOS but accessing from a windows browser.

andydotxyz (changes requested):
Thanks so much.
I made a suggestion theat keeps the word "browser" out of the general code but also fixes issue with macOS Server and windows client.

internal/driver/glfw/web_shortcuts_js.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/gopherjs/gopherjs/js"
+)
+
+// Checks if the browser is runnning on Mac OSX
+func isMacOSBrowser() bool {
+	if strings.Contains(strings.ToLower(js.Global.Get("window").Get("navigator").Get("platform").String()), "mac") {
+		return true
+	} else {
+		return false
+	}
```
Jacalz, line 18: Just return the Boolean value directly instead of checking if it is true and then returning a true value.

internal/driver/glfw/web_shortcuts_wasm.go
```diff
@@ -0,0 +1,19 @@
...
+	"syscall/js"
+)
+
+// Checks if the browser is runnning on Mac OSX
+func isMacOSBrowser() bool {
+	if strings.Contains(strings.ToLower(js.Global().Get("window").Get("navigator").Get("platform").String()), "mac") {
+		return true
+	} else {
+		return false
+	}
```
Jacalz, line 18: Same thing here.

Jacalz (changes requested):

internal/driver/glfw/window.go
gcottom: renamed to isMacOSRuntime. removed web/browser refererence. put runtime check in the shortcuts_other.

internal/driver/glfw/web_shortcuts_js.go
gcottom: good point, I've made the changes

internal/driver/glfw/web_shortcuts_wasm.go
gcottom: good point, I've made the changes

dweymouth:
There are other places we assign shortcuts dynamically based on runtime.GOOS. `widget/entry.go` being one (only one I know of without digging around). It would be great to get this improvement there as well!

andydotxyz:
True, but I think that's a bigger fix as this is internal to the GLFW driver code for the initial fix?

gcottom:
From what I can see the only place that shortcuts are set outside of what I've changed here is in widget/entry.go. But to add this the shortcut fix would have to live outside of the glfw driver. Any suggestions on how to move forward?

dweymouth:
I think we will just move forward with this for now as the fix in widget/entry.go would require a new API, which probably should be public, and would wait until 2.5

andydotxyz:
> I think we will just move forward with this for now as the fix in widget/entry.go would require a new API, which probably should be public, and would wait until 2.5

That works for me. Can you please rebase this on the `develop` branch @gcottom as that is where we do our development work, then this could be merged.

andydotxyz:
The other half of this is now landed, so you can update the version this depends on when rebasing and it should be fully testable.

gcottom:
@andydotxyz rebased and updated dependency and set to develop branch.

lorypelli:
I'm on windows and ALL shortcuts are broken for me

andydotxyz:
> I'm on windows and ALL shortcuts are broken for me

You are commenting on a pull request here. Do you mean that with this change it is broken, or it was broken for you before and after?

lorypelli:
> > I'm on windows and ALL shortcuts are broken for me
> 
> You are commenting on a pull request here. Do you mean that with this change it is broken, or it was broken for you before and after?

After the new update, context menu is working, but still all shortcuts are broken...

Jacalz:
This is certainly not the place to report a bug. This is a pull request for fixing shortcuts when running Fyne apps in a browser. If something isn’t working then please open a bug report instead…

andydotxyz:
> After the new update

You tested this PR and it fixed popup menus? That doesn't seem likely.

lorypelli:
> > After the new update
> 
> You tested this PR and it fixed popup menus? That doesn't seem likely.

Is this the correct PR for my issue, is this related to shortcuts like CTRL + C or CTRL + V or F12 not working?

andydotxyz:
This should fix all copy-paste shortcuts. When you appply the patch it contains

PR #4358, opened 2023-11-02
updated URI.Extension() documentation

### LINKED ISSUE #4357: Incorrect documentation example on URI.Extension() function
### Checklist

### Describe the bug

While I was working on some file validation, I came across this bug. The official golang `path/filepath` package returns extentions with the dot included (e.g `.csv`) but the `URI.Extention()` function is incorrectly documented as returning only the extension as follows;
```
// Extension should return the file extension of the resource
// referenced by the URI. For example, the Extension() of
// 'file://foo/bar.baz' is 'baz'. May return an empty string if the
// referenced resource has none.
// Extension() string
```
Although this is a very minor issue, I think it might lead some developers (like me) to lose some amount of time debugging before they realize that the problem is from the documentation and not from their `reader.URI.Extension() == 'csv'` if condition :smile: 

I have created a pull request that updates the documentation.

### How to reproduce

1. Call a `ShowFileOpen` function and choose a file with an extension
2. print the value of `reader.URI().Extension()`

### Screenshots

_No response_

### Example code

```go
dialog.ShowFileOpen(func(reader fyne.URIReadCloser, err error) { // suppose we selected a csv file
	if err != nil || reader == nil {
		return
        }
        fmt.Println(reader.URI().Extension()) // outputs '.csv'
}
```

### Fyne version

2.4.1

### Go compiler version

1.21.0

Ubuntu 22.04

### Additional Information

_No response_

roskee:
Updated URI.Extension() documentation to clarify that it returns the extension with the dot included.

 

### Description:

I have updated the documentation on URI.Extension() to clarify that the returned string includes the dot.

Fixes #4357 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Great catch thanks

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 This file lists the main changes with each version of the Fyne toolkit.
 More detailed release notes can be found on the [releases page](https://github.com/fyne-io/fyne/releases). 
 
-## 2.4.1 - 8 October 2023
```
dweymouth, line 6: I think you may have based your branch off of master instead of develop, or otherwise accidentally picked up this unrelated diff?

roskee: Yeah. It was based on master. Sorry! I will close this one and create a new pull request.

roskee:
incorrect base branch `master`

PR #4375, opened 2023-11-08
Avoid allocating a new slice on every call to drawSingleFrame

dweymouth:
### Description:

Preallocate a slice to reuse to avoid memory allocations in a tight loop

Side note - after testing this in Supersonic for awhile it seems like the GC is more inclined to give excess memory back to the OS now since we're not allocating so frequently :)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/loop.go
+++ b/internal/driver/glfw/loop.go
@@ -73,7 +73,7 @@ func runOnDraw(w *window, f func()) {
 // window canvases that are dirty on a single frame.
 // So its memory impact should be negligible and does not
 // need periodic shrinking.
-var refreshingCanvases = make([]fyne.Canvas, 0)
+var refreshingCanvases []fyne.Canvas
 
 func (d *gLDriver) drawSingleFrame() {
 	for _, win := range d.windowList() {
--- a/internal/driver/glfw/loop.go
+++ b/internal/driver/glfw/loop.go
@@ -67,8 +67,15 @@ func runOnDraw(w *window, f func()) {
 	<-done
 }
 
+// Preallocate to avoid allocations on every drawSingleFrame.
+// Note that the capacity of this slice can only grow,
+// but its length will never be longer than the total number of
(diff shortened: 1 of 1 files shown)


internal/driver/glfw/loop.go
```diff
@@ -67,8 +67,15 @@ func runOnDraw(w *window, f func()) {
 	<-done
 }
 
+// Preallocate to avoid allocations on every drawSingleFrame.
+// Note that the capacity of this slice can only grow,
+// but its length will never be longer than the total number of
+// window canvases that are dirty on a single frame.
+// So its memory impact should be negligible and does not
+// need periodic shrinking.
+var refreshingCanvases = make([]fyne.Canvas, 0)
```
Jacalz, line 76: You don't have make it into zero size. Doing so does an unnecessary call to the runtime per https://stackoverflow.com/a/29164565.  Just do this instead:
```suggestion
var refreshingCanvases []fyne.Canvas
```

Jacalz (changes requested):
Thanks. That's a good improvement. Just one minor note inline :)

internal/driver/glfw/loop.go
dweymouth: The alternative I guess is to use a best-guess size rather than 0 in the initializer, but at this point it really is unnecessary optimization so I'll just go with your suggestion :)

Jacalz: To be fair, I do not necessarily only say this because it technically should be faster but I do prefer to only make a slice when I need to set a non default capacity. Not using make for zero size does make me have to parse less code if that makes sense?

Jacalz (approved):

internal/driver/glfw/loop.go
Jacalz: I see now that it already was make'd to zero. Not you fault but I have never really understood why people do it. It is more code to write as well ;)

PR #4389, opened 2023-11-15
Add Remove to data binding lists and trees

### LINKED ISSUE #3100: Function "Remove" or "RemoveItem" on binding lists
When working with binded lists you can Append and Prepend items, but it's not possible to Remove them. 
This makes using binded lists more difficult.

You can "Get" the list, manipulate it outside and then "Set" the new list.

A "Remove" function which accept an index of the item that should be removed.

andydotxyz:
Adding the missing functions.
Removing a tree item removes all its children.

Fixes #3100

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/bindlists.go
+++ b/data/binding/bindlists.go
@@ -114,25 +114,27 @@ func (l *boundBoolList) Remove(val bool) error {
 	l.lock.Lock()
 	defer l.lock.Unlock()
 
-	if len(*l.val) == 0 {
+	v := *l.val
+	if len(v) == 0 {
 		return nil
 	}
-	if (*l.val)[0] == val {
-		*l.val = (*l.val)[1:]
-	} else if (*l.val)[len(*l.val)-1] == val {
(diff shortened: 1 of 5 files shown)


andydotxyz:
Naming to be checked here I guess. List might add `RemoveIndex` in the future, should there be an equivalent for Map and/or should `Remove` be `RemoveKey`? I was following the Golang `delete` as passing the key but that may not fit with our API design.

Jacalz:
I'll try to have a look tomorrow or on Thursday if possible.

andydotxyz:
Thanks, I think it should all be sorted now.

Jacalz (changes requested):
Looking good. Just some staticcheck fixes that need to be done now that it apparently understands what `v` is (bug?).

andydotxyz:
Oh good point I missed that. It was a bug that was hiding. Added a new test to confirm

PR #4396, opened 2023-11-17
Ensure the widget updates to show cursor position

### LINKED ISSUE #4181: Pasting text in Entry does not update cursor position display
### Checklist

### Describe the bug

When pasting some text in an Entry widget, the caret/cursor position is not update.
It is always shown at the begining of the pasted text.

It's a display bug only, since when, after the paste, we type some text, it is inserted at the right position (after the pasted text).

### How to reproduce

Paste text in an Entry widget.
See the cursor position (wrong).
Type some text, the cursor is updated to the right position.

### Screenshots

Typing some text
image

After paste
image

Re-type some text (cursor is updated to the right position)
image

### Example code

```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("")

	w.SetContent(widget.NewMultiLineEntry())
	w.ShowAndRun()
}
```

### Fyne version

v2.3.6-0.20230822222624-fce948601c72

### Go compiler version

1.21

Windows 10

### Additional Information

_No response_

andydotxyz:
Fixes #4181

### Checklist:

- [ ] Tests included. <- tests were passing already, this is a render glitch that the software render tests would not catch either
- [x] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -910,6 +910,7 @@ func (e *Entry) pasteFromClipboard(clipboard fyne.Clipboard) {
 
 	e.updateTextAndRefresh(provider.String())
 	e.CursorRow, e.CursorColumn = e.rowColFromTextPos(pos + len(runes))
+	e.Refresh() // placing the cursor (and refreshing) happens last
 }
 
 // placeholderProvider returns the placeholder text handler for this entry


widget/entry.go
```diff
@@ -910,6 +910,7 @@ func (e *Entry) pasteFromClipboard(clipboard fyne.Clipboard) {
 
 	e.updateTextAndRefresh(provider.String())
 	e.CursorRow, e.CursorColumn = e.rowColFromTextPos(pos + len(runes))
```
dweymouth, line 912: It looks like we're refreshing at least twice now per paste (3 times if e.selecting)? Is there a way to refactor this down to one refresh (or max 2)? Or does it not really matter since they're all but guaranteed to be on the same frame anyway?

andydotxyz: I looked at that and it would mean introducing 2 new methods, breaking down the content of the two above.
I figured that given this is on a paste operation it wasn't likely a big deal...

dweymouth: Sounds good, maybe let's create a ticket tagged with optimization for reducing Entry refreshes? probably not high priority at all but I think something that should eventually be looked into

dweymouth (approved):

widget/entry.go
andydotxyz: Done :)

PR #4398, opened 2023-11-18
Reduce allocations in list scroll updates

dweymouth:
### Description:

Previously, we were allocating at least 3 new slices and one new map on every call to updateList.

This PR re-uses the same allocated slice capacity to place the objects that will be drawn each update, and switches to using slices with reusable capacity instead of maps to keep track of what's visible. Since we only expect max 50 or so list rows to ever be visible - and also we can use binary search since we always insert things in ascending row order - this should be even faster than maps.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -123,7 +123,10 @@ func (l *List) RefreshItem(id ListItemID) {
 	}
 	l.BaseWidget.Refresh()
 	lo := l.scroller.Content.(*fyne.Container).Layout.(*listLayout)
-	if item, ok := lo.searchVisible(lo.visible, id); ok {
+	lo.renderLock.RLock() // ensures we are not changing visible info in render code during the search
+	item, ok := lo.searchVisible(lo.visible, id)
+	lo.renderLock.RUnlock()
+	if ok {
 		lo.setupListItem(item, id, l.focused && l.currentFocus == id)
 	}
 }
@@ -537,7 +540,7 @@ type listLayout struct {
 	visible           []itemAndID
 	wasVisible        []itemAndID
 	visibleRowHeights []float32
-	renderLock        sync.Mutex
+	renderLock        sync.RWMutex
 }
 
 func newListLayout(list *List) fyne.Layout {
--- a/widget/list.go
+++ b/widget/list.go
@@ -664,6 +664,7 @@ func (l *listLayout) updateList(newOnly bool) {
 		return
 	}
 
+	oldVisibleLen := len(l.visible)
 	l.visible = l.visible[:0]
 	oldChildrenLen := len(l.children)
 	l.children = l.children[:0]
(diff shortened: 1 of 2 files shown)


widget/list.go
```diff
@@ -724,3 +743,31 @@ func (l *listLayout) updateSeparators() {
...
+			return visible[mid].item, true
+		}
+		if visible[mid].id > id {
+			high = mid - 1
+		} else {
+			low = mid + 1
+		}
+	}
+	return nil, false
+}
```
Jacalz, line 764: I wonder if it would be cleaner to just use `sort.Search()` here? Might be slightly slower but does it matter?

Jacalz (commented):
Thanks. Nice work. I just have a comment (not necessarily requesting changes) inline :)

dweymouth:
Oh, I think we have a race condition - and it existed before, too. It seems like there is nothing preventing a user-code call to list.RefreshItem running concurrently with the layout logic that is computing what items are visible.

widget/list.go
dweymouth: Oh yeah, good call.

dweymouth: Done

Jacalz (dismissed):
Looks good to me. Just wondering if we perhaps should add a benchmark for the list widget? Makes it easier to verify performance improvements and so on.

dweymouth:
> > I actually think there might be another race condition here unless updateList always runs in the same goroutine.
> 
> The `updateList` is on the renderer, so the idea was that it was always running in the rendering goroutine. That may not be true but it was the expectation.

OK looking again it's definitely not true since it's invoked (through offsetUpdated) via public APIs like ScrollToItem that can be called from any goroutine. So we either need a way to post a re-render message to the rendering goroutine, or come up with new locking or to create ephemeral `visible`, `wasVisible` slices (obtained via a new sync.Pool to reuse allocated memory) that can be safely read in the non-locked part of `updateList`.

dweymouth:
Any ideas @andydotxyz on what is the right way forward? I think there are at least two options:
* make the whole `updateList` protected by the lock including when it invokes user callbacks. If we do this, we need to spawn a new goroutine every time `updateList` is called from any public API, to prevent possible deadlocking
* go back to making the second half of `updateList` properly re-entrant by using local slices of `visible`/`wasVisible` that aren't concurrently accessed anywhere. Use sync.Pool to reuse allocated slice memory so we don't allocate on every scroll

andydotxyz:
I don't have any specific thoughts sorry.
The general idea is that public methods will lock internal state if they have to - and private methods should be allowed to assume locks are in place. This is something that is emerging though and plenty of code does not follow it yet.

Is that the way forward or should we think more?

dweymouth:
I implemented bullet point 2 from my last comment - creating local slices of visible and wasVisible that can safely be read in the non-locked portion of updateList, and reusing allocated slice memory with a new sync.Pool added to the listLayout.

Jacalz:
Cool. Well done. Will have a look in a day or two 👍

andydotxyz (approved):
Thanks for working on this performance

PR #4401, opened 2023-11-18
WIP: idea/suggestion: Add ContainerWidget code to make using containers as widgets easier

sdassow:
### Description:

Lately I've had the need to wrap containers in widgets, and I came up with some rather simple wrapper code to put a container into a widget.
The more I'm copying this code around, the more I think it could be useful to others as well.

Basic usage looks like this:
```go
type MyWidget struct {
        ContainerWidget
        ...
}
// ... somewhere in the code...
x := &MyWidget{}
x.ExtendBaseWidget(x)
x.Container = container.New(...)
```

This also allows to add methods that can be used to manipulate the state of the widget.
While it looks very simple, it's been proven very useful so far, and since it seems like very common boilerplate code I thought it might be worth having in the widget code.

This isn't polished yet and certainly would need more/better documentation as well as tests I think.
For now it's meant to gather feedback and see if there is any interest in having this.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/container_widget.go
+++ b/widget/container_widget.go
@@ -1,0 +1,75 @@
+package widget
+
+import (
+	"fyne.io/fyne/v2"
+)
+
+// Container widget helps to use a container as a widget.
+type ContainerWidget struct {
+	BaseWidget
+	*fyne.Container
+}
+
+type ContainerWidgetRenderer struct {
+	*fyne.Container
+}
+
+func (r *ContainerWidgetRenderer) Layout(size fyne.Size) {
+	r.Container.Layout.Layout(r.Container.Objects, size)
+}
+func (r *ContainerWidgetRenderer) MinSize() fyne.Size {
+	return r.Container.MinSize()
+}
+func (r *ContainerWidgetRenderer) Refresh() {
+	r.Container.Refresh()
+}
+func (r *ContainerWidgetRenderer) Objects() []fyne.CanvasObject {
+	return r.Container.Objects
+}
+func (r *ContainerWidgetRenderer) Destroy() {}
+
+var _ fyne.CanvasObject = (*ContainerWidget)(nil)
+var _ fyne.WidgetRenderer = (*ContainerWidgetRenderer)(nil)
+
+// implement CanvasObject interface
+func (t *ContainerWidget) CreateRenderer() fyne.WidgetRenderer {
+	return &ContainerWidgetRenderer{
+		Container: t.Container,
+	}
+}
+
+func (t *ContainerWidget) Hide() {
+	t.Container.Hide()
+}
+
+func (t *ContainerWidget) MinSize() fyne.Size {
+	return t.Container.MinSize()
+}
+
+func (t *ContainerWidget) Move(pos fyne.Position) {
+	t.Container.Move(pos)
+}
+
+func (t *ContainerWidget) Position() fyne.Position {
+	return t.Container.Position()
+}
+
+func (t *ContainerWidget) Refresh() {
+	t.Container.Refresh()
+}
+
+func (t *ContainerWidget) Resize(size fyne.Size) {
+	t.Container.Resize(size)
+}
+
+func (t *ContainerWidget) Show() {
+	t.Container.Show()
+}
+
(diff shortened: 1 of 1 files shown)


Jacalz:
Thanks for looking into this. I believe that you might be duplicating something that already exists (kind of). Just writing this to perhaps save you some time :)

Are you looking for something like this? https://pkg.go.dev/fyne.io/x/fyne@v0.0.0-20231020065621-89b4a4aea27d/wrapper

We also have `widget.NewSimpleRenderer` which makes it quite easy to wrap objects.

sdassow:
> Thanks for looking into this. I believe that you might be duplicating something that already exists (kind of). Just writing this to perhaps save you some time :)
> 
> Are you looking for something like this? https://pkg.go.dev/fyne.io/x/fyne@v0.0.0-20231020065621-89b4a4aea27d/wrapper

Related but different. Thanks for pointing at the fyne-x repo though, I have to remember checking there before as well (the other wrappers there could be useful for something else).
This seemed so fundamental (at least for what I'm doing), I assumed it would be worth to be in fyne core. In the end I simply want to give back a bit and share valuable code, and the fyne-x repo might be a good place for it as well.

> We also have `widget.NewSimpleRenderer` which makes it quite easy to wrap objects.

Unfortunately it behaves a bit different with regards to `MinSize()` being forwarded to `fyne.CanvasObject` instead of `fyne.Layout` I believe, which makes it appear with additional space on top in my tests.

dweymouth:
> Unfortunately it behaves a bit different with regards to MinSize() being forwarded to fyne.CanvasObject instead of fyne.Layout I believe, which makes it appear with additional space on top in my tests.

I think you may be missing an ExtendBaseWidget somewhere then? This is the pattern I use which works well for me:

```go
type MyCustomWidget struct
    widget.BaseWidget
    // all your widget's state, including its sub-widgets
}

func NewMyCustomWidget() *MyCustomWidget {
    w := &MyCustomWidget{ /* init if needed*/ }
    w.ExtendBaseWidget(w)
    return w
}

func (w *MyCustomWidget) CreateRenderer() fyne.WidgetRenderer {
    c := container.NewWhatever(...) // do your layout here
    return widget.NewSimpleRenderer(c)
}
```

andydotxyz:
> > We also have `widget.NewSimpleRenderer` which makes it quite easy to wrap objects.
> 
> 
> 
> Unfortunately it behaves a bit different with regards to `MinSize()` being forwarded to `fyne.CanvasObject` instead of `fyne.Layout` I believe, which makes it appear with additional space on top in my tests.

Using `NewSimpleRenderer(NewContainer(...))` should function exactly as the new container widget you have proposed.

I don't think we should go with a "container widget" as they mean different things - containers lay things out with no interactivity and widgets are where behaviour is added (though I appreciate that the lines are blurred slightly). 

If a helper such as that you have committed is really useful then why limit it to Container? Couldn't it actually be any CanvasObject?

sdassow:
Ok, while changing my existing container widget code to use the `SimpleRenderer` still doesn't look as it should, I've switched another widget to use it and that is working perfectly fine. So something in my code is to blame, and I will go after it myself.

As for the suggestion/idea, I think this can be solved with better documentation.
I'm thinking of adding a paragraph to the page **Writing a Custom Widget** to show this pattern with example code. Expect a PR for the documentation soon.

Thanks for the friendly feedback. Closing this now.

sdassow:
FYI: successfully replaced all occurrences of my `ContainerWidget` code with the `SimpleRenderer` pattern in my current project, and all is well.

Jacalz:
Nice :)

andydotxyz:
> As for the suggestion/idea, I think this can be solved with better documentation.
> 
> I'm thinking of adding a paragraph to the page **Writing a Custom Widget** to show this pattern with example code. Expect a PR for the documentation soon.

That would be good. I think there is/was an open PR with this intent but it wasn't really meeting our expectations so perhaps you can make something clearer?

Jacalz:
See https://github.com/fyne-io/developer.fyne.io/pull/138 @andydotxyz

PR #4402, opened 2023-11-18
Remove constant debug field from canvas

Jacalz:
### Description:

This clears out a small amount of memory usage by moving a build-constant field to a global constant variable. Most importantly, it makes Go have a better time deadcode eliminating methods that we only want to compile in for debug (especially the reflect code that gets the name of the canvasobject).

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/debug_disabled.go
+++ b/internal/debug_disabled.go
@@ -1,0 +1,8 @@
+//go:build !debug
+// +build !debug
+
+package internal
+
+// BuildTypeIsDebug is true if built with -tags debug. This value exists as a constant
+// so the Go compiler can deadcode eliminate reflect from non-debug builds.
+const BuildTypeIsDebug = false
--- a/internal/debug_enabled.go
+++ b/internal/debug_enabled.go
@@ -1,0 +1,8 @@
+//go:build debug
+// +build debug
+
+package internal
+
+// BuildTypeIsDebug is true if built with -tags debug. This value exists as a constant
+// so the Go compiler can deadcode eliminate reflect from non-debug builds.
+const BuildTypeIsDebug = true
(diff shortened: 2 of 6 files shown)


internal/debug_disabled.go
```diff
@@ -0,0 +1,8 @@
+//go:build !debug
+// +build !debug
+
+package internal
+
+// BuildTypeIsDebug is true if built with -tags debug. This value exists as a constant
+// so the Go compiler can deadcode eliminate reflect from non-debug builds.
+const BuildTypeIsDebug = false
```
dweymouth, line 8: I think we already have a constant for this: https://github.com/fyne-io/fyne/blob/master/app/app_debug.go#L8C12-L8C12

Other than that, this is a nice improvement!

dweymouth (changes requested):

internal/debug_disabled.go
Jacalz: We can't use that field directly as it isn't exported. We can however do `fyne.CurrentApp().Settings().BuildType() == fyne.BuildDebug` but that's quite a slow lookup and I'm not sure to which degree it manages to deadcode eliminate it. I think Andrew cached the debug value like he did to avoid the app lookup + settings + buildtype lookup in the hot rendering path.

dweymouth: I wonder if we should move all the build type compile-time constants here then - seems a bit like an anti-pattern to have them defined twice in two different packages. Up to you if you want to address it in this PR now or just merge it and revisit later

Jacalz: I agree with you. I'll create a separate package for it, I think. A package like `internal/buildtype` could also have a common way to look up if we are running on Wayland or not for example.

Jacalz: I believe this should be fixed now. I'm very happy with the solution personally. Might be worth it to extend to mobile as well but need to see if it manages to deadcode that or not first

dweymouth (approved):

andydotxyz (approved):
Nice unification here thanks

PR #4403, opened 2023-11-18
Minor driver cleanup

Jacalz:
### Description:

It was a failure to reapply the init changes. Too many problems surface. I kept the other improvements from the PR and cleaned up some more things.

- The animation runner struct object didn't need to be a pointer. This allows us to not manually create the object, which in turn means that we get rid of some code and the constructor for the GLDriver is thus now inlineable.
- A macOS-specific function got updated to avoid multiple interface casts.
- The `glDevice` struct does not need to part of the driver. It is an empty struct, we can just return it each time someone asks for it.

~~This replaces my previous PR which had to be reverted due to an issue with the application not having started. This cleans up some of the init code to avoid calling init multiple times (when starting and when creating a new windows). It also includes an extra change to set up the device field more coherently.~~

~~Replaces #4177~~

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -24,7 +24,7 @@ type fyneApp struct {
 	uniqueID string
 
 	cloud     fyne.CloudProvider
-	lifecycle fyne.Lifecycle
+	lifecycle app.Lifecycle
 	settings  *settings
 	storage   fyne.Storage
 	prefs     fyne.Preferences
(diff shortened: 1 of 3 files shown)


andydotxyz:
Hmm, not sure.
This moves the OpenGL init from "when showing first window" to "when the app is loaded".
So it's much earlier in the lifecycle. Is this a good thing?
If an app loads and never shows a window we don't need OpenGL to be loaded before but now we do...

Jacalz:
Hmm. I think you're right but I'm really not a fan of the way it is called so many times unnecessarily now. Maybe there is a better way. I need to think about it...

Jacalz:
I haven't managed to come up with any good solutions. I rebased the PR and kept one of the other improvements from it plus expanded a bit on that idea. Now it's just a small cleanup.

andydotxyz (approved):

Issue #3377, opened 2022-11-05
Use Flatpak portals for file handling


Jacalz:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [X] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

Fyne applications, like https://beta.flathub.org/apps/details/io.github.jacalz.rymdport, running in Flatpak containers currently require the developers to allow a few XDG-directories to be enabled for the file and folder dialogs to work. Any other directories that the application needs access to will require those directories to be added as well.

It would be better to instead use the [Flatpak portals](https://libportal.org/) (over DBus presumably) to open OS file dialogs that then let the user select a file or folder that the application then will get access to.

### Is it possible to construct a solution with the existing API?

Yes. We need to make sure that the file and folder dialogs open the corresponding Flatpak portals for selecting or creating files/folders. The code for URI and storage code might also need to be adapted to handle this correctly.

### Describe the solution you'd like to see.

I don't know if this has been done using Go before but seems, per https://libportal.org/libportal.html, like this should be doable over Dbus. Worst case, we need to create an optional CGo wrapper around https://github.com/flatpak/libportal.

Bluebugs:
I think that the support for flatpak call for making a libportal in go. Dbus and go isn't that difficult and it would fit well in helping the go ecosystem on Linux.  I think we should have a fyne-io/flatpak-portal module for this.

PR #4406, opened 2023-11-18
Implement support for XDG Desktop Portal

### LINKED ISSUE #3378: Use Flatpak portals for notifications
### Checklist

Fyne applications, like https://beta.flathub.org/apps/details/io.github.jacalz.rymdport, currently have to enable `--talk-name=org.freedesktop.Notifications` to support notifications over Dbus. It would be better to instead use the Flatpak portal for this as it would avoid having to enable more than permissions and also would allow the Gnome/Plasma applications settings to turn on and off notifications for the application.

Yes. We should just send the notification in a different way that what we currently are doing.

I don't know if this has been done using Go before but seems, per https://libportal.org/libportal.html, like this should be doable over Dbus. Worst case, we need to create an optional CGo wrapper around https://github.com/flatpak/libportal.

Using `go build -tags flatpak` might be required to solve this.

Jacalz:
### Description:

This is a WIP PR to implement support for using XDG Desktop Portal (https://flatpak.github.io/xdg-desktop-portal/) for interacting with the host system (notifications, file choosers etc.). At this point only notifications are added.

I have decided to only use these when we compile with `-tags flatpak` for now. If we want to use native file dialogs outside of the sandbox then that's a different decision to be had.

Fixes #3378, #3377

### File and folder dialogs

Everything except for file filters and changing the dismiss button text has been set up and is working perfect. The first one seems like it might be problematic to get working but should be possible. The latter will not be possible to implement (and probably doesn't matter that much). 

Running inside Flatpak with no added file or notification permissions: 
[Screencast from 2023-11-19 21-50-07.webm](https://github.com/fyne-io/fyne/assets/25466657/f0a271d5-bb78-4817-9fe4-12f0dfcb955b)

#### Opening URLS

The `app.OpenURL()` method has been updated to use the `OpenURI` portal. Not 100% necessary because the `xdg-open` command we used before already works but going through the portal is the recommended approach as far as I know. It should allow the user to theoretically deny an app from opening links.

#### Notifications

From the documentation:
```
The portal only accepts serialized GThemedIcon and GBytesIcons. Both of these have the form (sv). For themed icons, the string is "themed", and the value is an array of strings containing the icon names. For bytes icons, the string is "bytes", and the value is a bytestring containing the icon data in png, jpeg or svg form. For historical reasons, it is also possible to send a simple string for themed icons with a single icon name. 
```

I decided to go with the last part where we can use the name of the application icon. It has the downside of resulting in applications having no icon unless they have been installed. It is perhaps not optimal but it is a lot better than no icon at all and I didn't want to mess around with some GTK icon serialization stuff.

The code is currently using the appID as the icon key even though we are building tarballs where the icon is named after the application name in  `fyne package`. Flatpak and other tools want the desktop files and icons to be named after the application id and not the application name so we were doing the wrong thing all along. I will update the code to use that instead.

No icon:
image

Using `io.github.jacalz.rymdport` as appID (with it installed) works:
image

#### TODO:
- [x] ~~Switch Linux/BSD tarballs using `fyne package` to use appid instead of app name for files.~~
- [x] Look at dialog.FileDialog and dialog.FolderDialog implementations with `FileChooser`.
- [x] Look into opening links using `OpenURI`.
- [x] Implement a way to get the parent window.
- [ ] Get parent window working on Wayland as well.
- [ ] Get file filters working.
- [x] Only use these portals when compiling with `-tags flatpak`.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
I think we need to decide on when to use the local file dialogs and when not to. They do work even when outside of Flatpak so it isn't necessarily as simple as just using them in Flatpak.

Jacalz:
Never mind. I've decided to only use these portals when compiling with a new `wayland` buildtag.

Jacalz:
I couldn't figure out how to get the Wayland window handle so I think I'll leave that until another time. The way file filters work also seem to be very different so I think it would be wise to tackle that later as well. This should be good to be reviewed now. It is working very nicely :)

andydotxyz (approved):
This is really cool thanks so much. 

Can you open a new issue (or PR? ;) ) to track the missing waylaid handle lookup and file filter etc?

Jacalz:
Sure. I'll open an issue for now but will likely try to implement it myself going forward. Especially now with Wayland takin shape

PR #4409, opened 2023-11-21
Fix *BSD systems using the wrong (and slow) window resize

Jacalz:
### Description:

I realized that we had decided to group Linux into its own implementation without considering that *BSD systems would want to use the same fast file path.

Draft: Currently untested on BSD. The logic is sound and I just didn't forget my branch so opened.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
I think this ought to go into a v2.4.x release when merged.

sdassow:
Oh, nice! Going to try this later on OpenBSD. Noticed it before, but didn't start debugging since I'm also hacking on the window manager I'm using and couldn't rule out that it wasn't my fault so far as I'm actually working on window resizing logic.

sdassow:
Tested and works as intended on OpenBSD.

Jacalz:
Thanks for testing 👍

andydotxyz (approved):

PR #4417, opened 2023-11-23
Glfw remove old workaround

Jacalz:
### Description:

See https://github.com/fyne-io/fyne/pull/1136. The clipboard no longer crashes when pasting empy content (also verified locally). The clipboard also no longer require access to the window struct. Clean up the code. As the clipboard no longer needs the viewport, we can also remove a possible crash when getting a clipboard from a window that has yet to be set up.

I also added a comment to a workaround that is still needed.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
See https://github.com/fyne-io/fyne/issues/4418 as well.

Issue #4370, opened 2023-11-05
Incorrect row header width in widget.Table


vinser:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

Neither  SetColumnWidth nor template in CreateHeader do not set row header width up. 
See the first fixed column on screenshot below

### How to reproduce

Run the code below

### Screenshots

![image](https://github.com/fyne-io/fyne/assets/14954512/1caa4879-bc61-4a5d-8c7f-1cd5fd5e6769)

### Example code

```go
import (
	"fmt"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	myApp := app.New()
	w := myApp.NewWindow("Table with Borders")

	t := widget.NewTableWithHeaders(
		func() (int, int) { return 5, 5 },
		func() fyne.CanvasObject {
			return widget.NewLabel("Cell 000, 000")
		},
		func(id widget.TableCellID, o fyne.CanvasObject) {
			label := o.(*widget.Label)
			label.SetText(fmt.Sprintf("Cell %d, %d", id.Row, id.Col))
		})

	t.SetColumnWidth(-1, 100)

	t.CreateHeader = func() fyne.CanvasObject {
		return widget.NewLabel("Row 000")
	}

	t.UpdateHeader = func(id widget.TableCellID, o fyne.CanvasObject) {
		l := o.(*widget.Label)
		if id.Col == -1 {
			l.SetText(fmt.Sprintf("Row %d", id.Row))
		} else {
			l.SetText(fmt.Sprintf("Column %d", id.Col))
		}
	}

	content := container.NewBorder(nil, nil, nil, nil, t)

	w.Resize(fyne.NewSize(400, 300))
	w.CenterOnScreen()
	w.SetContent(content)
	w.ShowAndRun()
}
```

### Fyne version

2.4.1

### Go compiler version

1.21.3

### Operating system and version

Ubuntu 22.04 arm64, Windows 10 amd64

Ju-B:
i encounter the same issue.

The code to handle header sizing do not use t.columnWidths nor t.rowHeights.
The size is always the one from the template's Minsize cached in t.headerSize.

[...]

PR #4421, opened 2023-11-24
Correct row header width in widget.Table

Ju-B:
### Description:
Allows to set width and height of headers using methods: SetColumnWidth and SetRowHeight with index -1

Fixes #([4370](https://github.com/fyne-io/fyne/issues/4370))

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window_notxdg.go
+++ b/internal/driver/glfw/window_notxdg.go
@@ -1,5 +1,5 @@
-//go:build !linux
-// +build !linux
+//go:build !linux && !freebsd && !openbsd && !netbsd
+// +build !linux,!freebsd,!openbsd,!netbsd
 
 package glfw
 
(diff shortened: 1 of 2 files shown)


andydotxyz:
Thanks for your contribution. I think there is some cross-branch issues here though, lots of unrelated commits have been pulled in.

Also does this not break the default header size? The new `headerSize` cache will default to 0,0 won't it?

Ju-B:
I'm not familiar at all with github and git in general, so I probably
messed up that part.
If you can advise how to edit this pull request, I'll do or I can close and
make a new one.

The default behaviour is not changed.
In createrenderer of Table and refresh of tablerenderer the cachesize is
and was set to minsize of header I added code to take into account if
height or width  has been specified.

 If no set is done, the default will be the header template's size as
before. This header behavior is coherent with the cells behavior.

An other implementation could avoid using the maps to store user specified
height or width but could rely on :
rowheaderwidth,columnheaderheight *int
Added to the Table type.
Benefit is that table.Minsize() will take the fast path if only the header
size is changed.
Do you have some thoughts?

Le ven. 24 nov. 2023, 23:17, Andy Williams ***@***.***> a
écrit :

PR #4425, opened 2023-11-27
Use internal/build for mobile lookup as well

Jacalz:
### Description:

Does not seem to help deadcode elimination anything but it seems nice to clean up some code and have all build type information in one common place? It does seem to introduce a few bugs in device tests so draft for now.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.


andydotxyz:
I think you'll need to do iOS or Android or mobile - the mobile tag is only used when we are simulating mobile not when it is inferred from the target OS.

PR #4433, opened 2023-12-01
Move hints check into internal/build

Jacalz:
### Description:

Just making sure that we are using the `internal/build` package everywhere for better cohesion.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (commented):
The check for "HasHints" could be removed as calling the function is no-op when disabled. Mostly it's there to help avoid calculations required to log a hint

andydotxyz (approved):

PR #4446, opened 2023-12-10
Reduce allocations in GridWrap on scroll updates

dweymouth:
### Description:

Similar to #4398 but for the GridWrap widget.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/gridwrap.go
+++ b/widget/gridwrap.go
@@ -3,6 +3,7 @@ package widget
 import (
 	"fmt"
 	"math"
+	"sort"
 	"sync"
 
 	"fyne.io/fyne/v2"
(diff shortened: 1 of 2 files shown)


Jacalz:
The tests don't seem to run. That seems strange

Jacalz:
Yeah maybe

dweymouth:
The tests ran after another commit that was just a typo fix

andydotxyz:
Thanks for looking into this - it certainly sounds desirable.
Are there any metrics / indicative benchmark we can check it against?

dweymouth:
I didn't run profiling myself but doing a memory profile with pprof and scrolling around a bunch in the GridWrap in Fyne Demo should show fewer overall allocations than the same thing without this change

andydotxyz (approved):
Thanks for this

PR #4462, opened 2023-12-16
Add Notification & watchTheme & stopWatching Support for web platform

### LINKED ISSUE #2735: Add notification support to the web driver
The web driver does not support the notification API. It would be nice to have a full local notification support for the web driver.

There is no need for a new API as it is just about supporting an existing API on a new platform.

The local notification API as described here should do the job:
https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API

(Do not confuse this with push notification)

Dorbmon:
### Description:

Add Notification API: Now we can send notification using browser's api.
Add watchTheme, stopWatching: Now we can listen the change of dark or light theme.

Fixes
#2735 
#1646 
### Checklist:

- ☐ Tests included.
- ☑ Lint and formatter run with no errors.
- ☑ Tests all pass.

### DIFF
--- a/app/app_goxjs.go
+++ b/app/app_goxjs.go
@@ -56,11 +56,17 @@ func (app *fyneApp) SendNotification(n *fyne.Notification) {
 func rootConfigDir() string {
 	return "/data/"
 }
+
+var themeChanged = js.FuncOf(func(this js.Value, args []js.Value) interface{} {
+	if len(args) > 0 && args[0].Type() == js.TypeObject {
+		fyne.CurrentApp().Settings().(*settings).setupTheme()
+	}
+	return nil
+})
+
 func watchTheme() {
-	js.Global().Call("matchMedia", "(prefers-color-scheme: dark)").Call("addEventListener", "change", js.FuncOf(func(this js.Value, args []js.Value) interface{} {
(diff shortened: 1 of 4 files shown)


andydotxyz:
On my testing the theme change monitoring already worked. Can you confirm the steps you used to show that it doesn't?

Dorbmon:
Yes. Current code of develop branch added the listener in the app/app_theme.js . I think it is not consistent with other implementation like Darwin.

Jacalz:
If that code already exists and you are adding another implementation, shouldn't the old code be removed?

Dorbmon:
Now it is removed.

app/app_goxjs.go
```diff
@@ -6,14 +6,67 @@
...
+	if len(args) > 0 && args[0].Type() == js.TypeObject {
+		fyne.CurrentApp().Settings().(*settings).setupTheme()
+	}
+	return nil
+})
+
+func watchTheme() {
+	js.Global().Call("matchMedia", "(prefers-color-scheme: dark)").Call("addEventListener", "change", themeChanged)
+}
+func stopWatching() {
```
andydotxyz, line 70: To match the `watchTheme` this method should have `Theme` in the name

Issue #4397, opened 2023-11-17
Entry refreshes too often


andydotxyz:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

In interactive steps (especially when cursor moves, like paste) the entry is refreshed many times.
We should try to reduce any refreshes that occur in quick succession in case they result in multiple frame draws.

### How to reproduce

paste lots?

### Example code

n/a

### Fyne version

v2.4.2

### Go compiler version

1.20.2

### Operating system and version

All

### Additional Information

As discussed in #4396

Dorbmon:
How about to add a flag to disable the refresh? We can set that flag before we call some functions, then disable the flag after the procedure we need. After that, we can do the refresh manually.
All we need to do is to add the code to check the refresh flag in the Refresh() and update the flag where we need.

Jacalz:
I think it is better to rework all the internal methods to not refresh and then have the exported methods refresh (or something along those lines). Some kind of refactoring basically. Having a flag in the struct to disable a specific function seems like a bad practice in my opinion.

PR #4465, opened 2023-12-17
Avoid unnecessary refresh in Entry widget.

Dorbmon:
### Description:

There are two ideas for reducing the unnecessary (Without doing deep modification to the internal packages). 

First is when Validate() is called, we check if the validate has refreshed the widget. If so, we won't refresh.

Second is that, we used to just call extra refresh to ensure the cursor to be at the correct position(#4396). We can avoid that extra refresh by replacing the "e.updateTextAndRefresh" with updateText and calling the callback function manually.

I still think introducing a flag to forbid the refresh is a better idea so that we can avoid refreshing without modifying too many codes and also, without writing too many no-refresh version helper functions. Although it is not that elegant :)
Fixes #(issue)
#4397 
### Checklist:

- [ ] Tests included.
- [X] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -1,6 +1,7 @@
 package widget
 
 import (
+	"errors"
 	"image/color"
 	"math"
 	"runtime"
@@ -893,11 +894,13 @@ func (e *Entry) TypedRune(r rune) {
 	})
 	e.propertyLock.Unlock()
 
-	e.Validate()
+	if err := e.Validate(); err == nil || errors.Is(err, e.validationError) {
+		// Under this situation, the validate function won't call refresh, so do it here
+		e.Refresh()
+	}
 	if cb != nil {
 		cb(content)
 	}
-	e.Refresh()
 }
 
 // TypedShortcut implements the Shortcutable interface
@@ -1018,6 +1021,11 @@ func (e *Entry) pasteFromClipboard(clipboard fyne.Clipboard) {
 		// format clipboard content to be compatible with single line entry
 		text = strings.Replace(text, "\n", " ", -1)
 	}
+	if text == "" {
+		// The content of the clipboard can be the unsupported type or just empty.
+		// Then we do nothing about this.
+		return
+	}
 	provider := e.textProvider()
 	runes := []rune(text)
 	pos := e.cursorTextPos()
@@ -1028,11 +1036,19 @@ func (e *Entry) pasteFromClipboard(clipboard fyne.Clipboard) {
 		Position: pos,
 		Text:     text,
 	})
+	content := provider.String()
+	e.updateText(content)
+	cb := e.OnChanged
 	e.propertyLock.Unlock()
 
-	e.updateTextAndRefresh(provider.String())
 	e.CursorRow, e.CursorColumn = e.rowColFromTextPos(pos + len(runes))
-	e.Refresh() // placing the cursor (and refreshing) happens last
+	if err := e.Validate(); err == nil || errors.Is(err, e.validationError) {
+		// the validate() won't call refresh under this condition, so we need to do it manually
+		e.Refresh() // placing the cursor (and refreshing) happens last
+	}
+	if cb != nil { // we have made sure that the paste content is not empty, so we always need to call the callback
+		cb(content)
+	}
 }
 
 // placeholderProvider returns the placeholder text handler for this entry
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -950,14 +950,15 @@ func (e *Entry) cutToClipboard(clipboard fyne.Clipboard) {
 	}
 
 	e.copyToClipboard(clipboard)
-	e.setFieldsAndRefresh(e.eraseSelection)
+	e.setFields(e.eraseSelection)
 	e.propertyLock.RLock()
 	cb := e.OnChanged
 	e.propertyLock.RUnlock()
 	if cb != nil {
 		cb(e.Text)
 	}
-	e.Validate()
+	e.validate()
+	e.Refresh()
 }
 
 // eraseSelection removes the current selected region and moves the cursor
@@ -1229,14 +1230,15 @@ func (e *Entry) selectingKeyHandler(key *fyne.KeyEvent) bool {
 	switch key.Name {
 	case fyne.KeyBackspace, fyne.KeyDelete:
 		// clears the selection -- return handled
-		e.setFieldsAndRefresh(e.eraseSelection)
+		e.setFields(e.eraseSelection)
 		e.propertyLock.RLock()
 		cb := e.OnChanged
 		e.propertyLock.RUnlock()
 		if cb != nil {
 			cb(e.Text)
 		}
-		e.Validate()
+		e.validate()
+		e.Refresh()
 		return true
 	case fyne.KeyReturn, fyne.KeyEnter:
 		if e.MultiLine {
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -1759,7 +1759,6 @@ func (r *entryRenderer) ensureValidationSetup() {
 
 		r.entry.validate()
 
-		r.Refresh()
 	}
 }
 
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -1475,7 +1475,7 @@ func (e *Entry) updateText(text string) bool {
 // This should not be called under a property lock
 func (e *Entry) updateTextAndRefresh(text string) {
 	var callback func(string)
-	e.setFields(func() {
+	e.setFieldsAndRefresh(func() {
 		changed := e.updateText(text)
 
 		if changed {
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -1,7 +1,6 @@
 package widget
 
 import (
-	"errors"
 	"image/color"
 	"math"
 	"runtime"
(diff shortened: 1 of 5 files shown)


widget/entry.go
```diff
@@ -893,11 +894,13 @@ func (e *Entry) TypedRune(r rune) {
 	})
 	e.propertyLock.Unlock()
 
-	e.Validate()
+	if err := e.Validate(); err == nil || errors.Is(err, e.validationError) {
+		// Under this situation, the validate function won't call refresh, so do it here
```
Jacalz, line 898: You are hard-coding an assumption about the method you are calling. This is bound to be problematic when someone changes the method's behaviour and forgets to update all occurrences of this assumption. It will basically introduce code-rot in the long run.

Jacalz (commented):
I won't request changes to block this just because I have a different opinion but I'll let someone else in the team decide. 

However, I am still under the impression that an unexported `e.validate()` that doesn't refresh would be a better option. The exported `Validate()` could then call the unexported version and then refresh.

Jacalz:
Also, fyi, I don’t believe this solves the issue in question entirely. There are quite a few places where we are refreshing twice.

widget/entry.go
Dorbmon: Yes. So introduce the validate() without refresh might be a better option.

Dorbmon:
Now I added the validate() which won't refresh and changed Validate() into a wrapper of validate() with refreshing.

Dorbmon:
> Also, fyi, I don’t believe this solves the issue in question entirely. There are quite a few places where we are refreshing twice+ outside of validation and pasting.

Actually I think the best way to do the most optimization is to add the refresh barrier flag(maybe not a flag, can also be a counter, when it is 0, means no barrier). For example. the user might do some change with refresh in the callback function, you won't know if they will do that. Only if we set a barrier flag stopping the refresh function from really doing refresh, can we avoid the unnecessary refresh. Also, it is convenient because we can just set the barrier flag at the top level of the code.

widget/entry_validation.go
```diff
@@ -27,13 +36,13 @@ func (e *Entry) SetOnValidationChanged(callback func(error)) {
 	e.onValidationChanged = callback
 }
 
-// SetValidationError manually updates the validation status until the next input change
-func (e *Entry) SetValidationError(err error) {
+// SetValidationError manually updates the validation status until the next input change. Returns if the validation error is really set.
+func (e *Entry) SetValidationError(err error) bool {
```
Jacalz, line 40: You are breaking the public facing API here. Please revert this change.

Jacalz (changes requested):
Looking better in my opinion. Though please note that you are currently breaking the stable API guarantee.

Jacalz:
You might want to have a look at all the uses of `setFieldsAndRefresh()`. Saw a few more cases where refresh happens twice.

Dorbmon:
> You might want to have a look at all the uses of `setFieldsAndRefresh()`. Saw a few more cases where refresh happens twice (`Undo() for example).

I believe I fix all this kind of duplicated refresh now.

widget/entry_validation.go
Dorbmon: OK

Jacalz:
You might also wish to have a look at `ensureValidationSetup()`. I think it is only called when we refresh later.

widget/entry.go
```diff
@@ -1218,14 +1230,15 @@ func (e *Entry) selectingKeyHandler(key *fyne.KeyEvent) bool {
...
+		e.setFields(e.eraseSelection)
 		e.propertyLock.RLock()
 		cb := e.OnChanged
 		e.propertyLock.RUnlock()
 		if cb != nil {
 			cb(e.Text)
 		}
-		e.Validate()
+		e.validate()
+		e.Refresh()
```
Jacalz, line 1241: You have move the refresh to before the callback in many other places. I suppose we want to be as consistent as possible. Not sure what is best but refreshing before potentially calling a long-running callback seems sensible?

Dorbmon:
> You might also wish to have a look at `ensureValidationSetup()`. I think it is only called when we refresh later.

How about directly remove the Refresh() from ensureValidationSetup() as ensureValidationSetup() is only used inside the package.

Jacalz:
> > You might also wish to have a look at `ensureValidationSetup()`. I think it is only called when we refresh later.
> 
> 
> 
> How about directly remove the Refresh() from ensureValidationSetup() as ensureValidationSetup() is only used inside the package.

That was my idea as well. Had a look through all of the Entry code looking at which unexported methods called Refresh() and then looked at where they were called.

widget/entry.go
Dorbmon: Yes. You are right.

Dorbmon: But the validate() won't call the refresh. If so, we need to also move the validate. Is this consistent with the previous behavior? I found in the original code, sometimes we call the Validate() before the callback function ,but sometimes after it.

Jacalz: I think it makes more sensor to validate before? The user might want to check the validation status maybe?

Jacalz: I honestly don't know. We should be consistent though

Jacalz:
Seems like there are a few test failures

Dorbmon:
I got some problems about the test. I ran the test "Entry_SetValidationError" manually on my local machine(macOS). And I got the error:
```
Test:           TestEntry_SetValidationError
                Messages:       Image did not match master. Actual image written to file:///xxxx/fyne/widget/testdata/failed/entry/validation_set_valid.png.
```
But I checked the acutal image, it looks same to the standard image(file:///xxxx/fyne/widget/testdata/entry/validation_set_valid.png.)

Jacalz:
Try to update the files (drag the files from the failed/ folder). There might be just some pixels in difference I think.

Dorbmon:
I don't understand. I can pass the "Entry_SetValidationError" test on my local machine...

andydotxyz:
> I don't understand. I can pass the "Entry_SetValidationError" test on my local machine...

Are you running on a M2 Mac by chance? They have somehow not the same image output to all other operating system.
So revert the image changes and see if the CI passes on this PR.
I still have to get to the bottom of why arm64 image output is different to all other arch...

Dorbmon:
> > I don't understand. I can pass the "Entry_SetValidationError" test on my local machine...
> 
> Are you running on a M2 Mac by chance? They have somehow not the same image output to all other operating system. So revert the image changes and see if the CI passes on this PR. I still have to get to the bottom of why arm64 image output is different to all other arch...

Yes. I run on M2 pro chip. I've reverted the change but it still has some errors....

andydotxyz:
> > > I don't understand. I can pass the "Entry_SetValidationError" test on my local machine...
> > 
> > 
> > Are you running on a M2 Mac by chance? They have somehow not the same image output to all other operating system. So revert the image changes and see if the CI passes on this PR. I still have to get to the bottom of why arm64 image output is different to all other arch...
> 
> Yes. I run on M2 pro chip. I've reverted the change but it still has some errors....

There is a test failure - the Entry border is no longer refreshing. When the state is invalid the border should be red... Comparing `form/hint_initial.png`

With your PR:
hint_initial

Before:
hint_initial

Dorbmon:
But on my local machine, it works seems well? 
image
But still has error on my local machine...

andydotxyz:
That isn't the same code. Run the unit test and check the output - you'll see that broken image I shared

Jacalz:
I get the same result as @andydotxyz when running your code locally. I think there is a bug somewhere resulting in the border around the Entry not updating.

Jacalz:
It would be great to try and get this landed soon. What I found strange though is that it seems to work fine when testing locally but, like noted above, the images for the Form widget seem to show that the border isn't updated correctly. Is there perhaps some Refresh missing somewhere?

Dorbmon:
I don't have any idea how to solve this... I don't why the behavior is different between running on the local machine and the GitHub Action.

Jacalz:
Andy, do you have any idea for what could be wrong? I've looked at the code but can't figure out what it can be.

andydotxyz:
Test differing from the widget in real life is normally because refresh happens too late or misses something. The main UI refreshes a lot more due to user actions etc.
Given that this PR was removing refreshes it seems likely that this is the issue.

Jacalz:
@Dorbmon Are you sure that you cannot replicate the test failures locally? The updated test files will be in `widget/testdata/failed` and its subfolders.

Dorbmon:
@Jacalz I ran again on my local machine. I got some error:
Test:           TestForm_Disable_Validation
                Messages:       Image did not match master. Actual image written to file:///xxx/fyne/widget/testdata/failed/form/disable_validation_disabled_valid.png.
The follow image is the actual image:
image


But it looks different from CI's error:
Test:       	TestForm_Disable_Validation
        	Messages:   	Image did not match master. Actual image written to file:///home/runner/work/fyne/fyne/widget/testdata/failed/form/disable_validation_enabled_invalid.png.

andydotxyz:
> [better name](https://github.com/fyne-io/fyne/pull/4465/commits/a1a11fcbe866abb15e6a17df180cfa5eb39dab94)

It looks like you have pushed a load of test debug code and/or commented out test cases.
With some of them disabled I can see why you might find that different lines of the test are failing - as it will stop at the first image that does not match.

Jacalz:
@Dorbmon Would you have anything against me picking this up and trying to get the double refreshes resolved? I think it would be nice to land this in v2.4.5 but I also think it might be easier to start fresh and constantly run the tests to avoid running into these issues again :)

Jacalz:
I opened #4680 for now. I can close it if you are not happy with me taking on the work.

Dorbmon:
> I opened #4680 for now. I can close it if you are not happy with me taking on the work.

Just feel free to take this job. I am so busy these days and have no idea how to solve this:)

Jacalz:
Alright. Thanks and thanks for working on this. You are always welcome back to working on PRs once you have more time :)

PR #4470, opened 2023-12-21
Various improvements to InnerWindow

andydotxyz:
Including:
* Indication that corners are draggable
* Padded by default
* Ability to switch content
* Better updating of state when refreshed.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


container/innerwindow.go
```diff
@@ -25,15 +24,15 @@ type InnerWindow struct {
...
+	content *fyne.Container
 }
 
 // NewInnerWindow creates a new window border around the given `content`, displaying the `title` along the top.
 // This will behave like a normal contain and will probably want to be added to a `MultipleWindows` parent.
 //
 // Since: 2.5
 func NewInnerWindow(title string, content fyne.CanvasObject) *InnerWindow {
-	w := &InnerWindow{title: title, content: content}
+	w := &InnerWindow{title: title, content: NewPadded(content)}
```
dweymouth, line 35: I think by the "Principle of Least Surprise", inner windows should be unpadded by default, just like top-level windows. What about introducing a secondary constructor `NewPaddedInnerWindow` that sets it up with a padded container, and having this constructor set up a stack (ie unpadded) container?

Jacalz: Top level windows are padded by default though, aren’t they?

dweymouth: Oh wait, you're right. Disregard :)

dweymouth (approved):

PR #4474, opened 2023-12-22
4472 focus list item on select

### LINKED ISSUE #4472: When clicking a list item for the first time - scroll jumps back to item 0.
### Checklist

### Describe the bug

If you create a list and then select a list item at the bottom for the user, then when they try to change to a new item in the list, it will jump back to the top of the list before selecting that item.  This makes for a confusing UI experience, since no scroll should be needed (the item is visible).

The cause seems to be that the list item is not "focused" when we do list.Select().  So when the list itself is clicked - it resets to the focused item (0), then it focuses the item the user clicked on.  We only set the focused item in the list when it is clicked - but shouldn't we also set it when we've selected that item for the user?

### How to reproduce

list := NewList(...)
list.Select(id)

User then clicks an item in the list and it will adjust the scroll when its not necessary.

### Screenshots

_No response_

### Example code

list := NewList(...)
list.Select(id)

### Fyne version

2.4.0

### Go compiler version

1.19

Windows 10

### Additional Information

_No response_

williambrode:
### Description:

Fixes #4472

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,1 +1,1 @@
-github: [fyne-io, andydotxyz, toaster, Jacalz, changkun]
+github: [fyne-io, andydotxyz, toaster, Jacalz, changkun, dweymouth]
--- a/widget/list.go
+++ b/widget/list.go
@@ -205,6 +205,9 @@ func (l *List) Select(id ListItemID) {
 	if id < 0 || id >= length {
 		return
 	}
+	if !fyne.CurrentDevice().IsMobile() {
+		l.currentFocus = id
+	}
 	old := l.selected
 	l.selected = []ListItemID{id}
 	defer func() {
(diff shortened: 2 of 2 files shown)


Jacalz:
The includes a commit from Drew. Have you based the branch on the `master` branch?

Jacalz:
Also, please always fill out the checklist in the template and try to add some tests.

andydotxyz:
I think the tests won't run because of the same (re)base issue showing the stray commit.
Can you please try cherry picking your commit onto latest develop then force pushing to this branch again @williambrode? Hopefully that would resolve everything.

Jacalz (changes requested):
This seems to be working now. Can you please add a test for it?

williambrode:
I ran the test on `develop` without my changes and confirmed it fails.

williambrode:
Mobile test failed - I can either change the test or change the code.  Does someone know why the code doesn't focus the list for mobile?  I was trying to change as little as possible but figured it might be worth asking.

```go
li.onTapped = func() {
		if !fyne.CurrentDevice().IsMobile() {
			canvas := fyne.CurrentApp().Driver().CanvasForObject(l.list)
			if canvas != nil {
				canvas.Focus(l.list)
			}

			l.list.currentFocus = id
		}

		l.list.Select(id)
```

PR #4477, opened 2023-12-22
Add multi-select for list

### LINKED ISSUE #2575: Add multiple selection
Split out from the parent #1395, we next want to add multiple selection.
(a future ticket will add range selection, however we decide to progress with that).

From the ticket:
---

This is a simple implementation with no ranges and no optimisations

Add boolean Table.MultiSelect bool, defaults to false. Then we just add func (t *Table) SetSelection(TableCellID...) that iterates through existing selection to cancel them then iterates through the new ones to select.

List would have List.MultiSelect bool and func (l *List) SetSelection(ListItemID...), that iterates cancelling excisting, then selecting those passed.

dweymouth:
### Description:

* Adds new enum widget.SelectionMode (values: SelectionSingle, SelectionMultiple, SelectionNone)
  - This only effects the tap/key handling, not the programmatic APIs
* Adds tap/key handling for multi-select, respecting SelectionMode setting
* Adds a few new APIs related to multi-select
  - `SelectOnly(ListItemID)`
  - `SelectAll()`
  - `SetSelection([]ListItemID)`
* The `list.Select` API now adds to the selection (which is what the documentation said), rather than replacing it

### TODO/ Concerns:
* tests
* This uses a slice of IDs to keep track of selection, and creates a new selection slice on any unselect operation. This makes thread-safety fairly easy, but comes with some performance tradeoffs:
  - Repeatedly unselecting items causes a lot of memory churn (UnselectAll is implemented efficiently)
  - **setupListItem** is now **O(n)** in number of selected items. This is a potential concern
  - Notifying onSelected/onUnselected callbacks is **O(n^2)** in operations that select multiple IDs at once (SelectAll, SetSelection)

Fixes #2575 (for list)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne_demo/tutorials/collection.go
+++ b/cmd/fyne_demo/tutorials/collection.go
@@ -96,7 +96,27 @@ func makeListTab(_ fyne.Window) fyne.CanvasObject {
 	list.SetItemHeight(5, 50)
 	list.SetItemHeight(6, 50)
 
-	return container.NewHSplit(list, container.NewCenter(hbox))
+	split := container.NewHSplit(list, container.NewCenter(hbox))
+	radio := widget.NewRadioGroup([]string{"Single", "Multiple", "None"}, func(sel string) {
+		switch sel {
+		case "Single":
+			list.SelectionMode = widget.SelectionSingle
+		case "Multiple":
+			list.SelectionMode = widget.SelectionMultiple
+		case "None":
+			list.SelectionMode = widget.SelectionNone
+		}
+	})
+	radio.Horizontal = true
+	radio.Selected = "Single"
+
+	return container.NewBorder(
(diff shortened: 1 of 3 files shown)


lidldev (approved):

gi-yt:
any update on this?

dweymouth:
No updates, and I have implemented multi-select in my own Fyne app by having the row widgets instead of the list handle it. And I've realized it's a lot better in my case to have the selection state as part of the data model instead of managed by the list because it enables things like sorting the data with the selection following the data instead of the row number.

So I have lost a lot of motivation in moving this forward... but it's open source so feel free to pick it up yourself, or copy it into your app's codebase, or take a look at my app (Supersonic) for how to manage multi-select in your app code instead of the list

Issue #1851, opened 2021-01-23
Complete support for Wayland clients


andydotxyz:
Currently in testing with `-tags wayland`, this needs a checklist of work to complete.

- [x] Scale detection does not work correctly
~- [ ] Window is not always filled when first shown (resize fixes)~ (this is not a Wayland specific issue - #4964)
~- [ ] Menu bar can be moved up under titlebar, perhaps when window minsize shrinks~ (cannot reproduce any more)
- [x] Find a solution for the window decorations... Request SSD or render something like FyneDesk?
- [x] Read scale information from the environment

Discussed on a recent call, wayland is not quite ready to be default on and auto-switch, but we will complete support in v2.5.x and test thoroughly with it defaulting on in v2.6.0 estimated.

~- [ ] Requires -tags wayland, update GLFW once that is removed and update our code accordingly~

brianredbeard:
As a note for anyone who comes along, you'll also need the headers for `xkbcommon` (e.g. `libxkbcommon-devel` in RPM based distros, `libxkbcommon-dev` in DEB based ones, etc).

brianredbeard:
Also, affected by go-gl/glfw#316:

```bash
[~/Projects/golang/src/fyne.io/fyne/v2/cmd/fyne_demo]$ go build -o fyne_demo-wayland -tags wayland . ;  echo $?
# github.com/go-gl/glfw/v3.3/glfw
In file included from ./glfw/src/internal.h:190,
                 from ./glfw/src/context.c:30,
                 from ../../../../../../pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20210311203641-62640a716d48/c_glfw.go:4:
./glfw/src/wl_platform.h:60:10: fatal error: wayland-xdg-decoration-client-protocol.h: No such file or directory
   60 | #include "wayland-xdg-decoration-client-protocol.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
2
```

andydotxyz:
Thanks for noticing this. Fix is now on `release/v2.0.x` branch ready for 2.0.3 release

brianredbeard:
Definitely getting some weird behavior.  Presently thinking through how to record a short video to demonstrate, but at a minimum the following errors are thrown on Fedora 33 using Wayland + GNOME3:

```
[...]

PR #4481, opened 2023-12-23
Update Glfw to v3.3.9 and improve Wayland support

Jacalz:
### Description:

Glfw 3.3.9 includes a bunch of fixed and improvements. Most notably, it now uses `libdecor` on Wayland meaning that the decorations work and are much more usable. I've also removed some workarounds (I personally asked for one of Andrew's several year old commits to be backported upstream) and made sure to silence the majority of warnings/errors that are printed when running on Wayland.  

There are still weird scaling and other issues but Wayland should now be quite usable (and now more than enough that you can actually use the app enough to fix the other issues) :)

For #1851

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Excellent thanks so much for this, looks great in weston

PR #4482, opened 2023-12-24
Make Slider a Disableable widget

### LINKED ISSUE #3551: Allow Disable and Enable for Slider widget
### Checklist

I often use the slider widget as input and it's not possible to disable and enable it out of the box.

The only work around is creating a custom slider widget and that is a huge inconvenience for such a simple feature. It requires one to extend not only the widget but also the renderer to provide the necessary disabled state styling. Things get messy fast.

It would be nice if the Slider widget like other input widget could implement the Disableable interface right out the box. I believe it is not such a huge fix, I could even offer to send in a PR for it.

dweymouth:
### Description:

Fixes #3551 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Nice addition thanks

PR #4493, opened 2023-12-29
Fix ScrollToBottom and ScrollToTop

### LINKED ISSUE #2917: func (s *Scroll) ScrollToBottom() not scrolling
### Describe the bug:
I used `ScrollToBottom` and found it didn't actually scroll to the bottom at all.  I was able to work around the issue by calling a couple of the lines that `ScrollToBottom` calls without the unnecessary stuff:
```go
scrollable.Offset.Y = scrollable.Content.MinSize().Height - scrollable.Size().Height
scrollable.Base.Refresh()
```
### To Reproduce:
Here is the code that creates a container with scrollable text.  Using scrollable.ScrollToBottom() didn't work for me (it loads scrolled to the top).
```go
func GetLogsTab(ctx context.Context) fyne.CanvasObject {
	scrollable := container.NewVScroll(widget.NewLabel(config.LogBuffer.String()))
	scrollable.ScrollToBottom()
	return container.NewMax(scrollable)
}
```

 - **OS:** Windows 10
 - **Version:** 10.0.19044
 - **Go version:** 1.16.5
 - **Fyne version:**  v2.1.4

lorypelli:
### Description:
I managed to fix scroll to bottom issue, I tested and it works...

Fixes #2917 

### Checklist:

- [ ] Tests included.
- [X] Tests all pass.

#### Where applicable:

### DIFF
--- a/internal/widget/scroller.go
+++ b/internal/widget/scroller.go
@@ -393,13 +393,16 @@ func (s *Scroll) CreateRenderer() fyne.WidgetRenderer {
 
 // ScrollToBottom will scroll content to container bottom - to show latest info which end user just added
 func (s *Scroll) ScrollToBottom() {
+	if s.Content.MinSize().Height > s.Size().Height {
+		s.Offset.Y = s.Content.MinSize().Height - s.Size().Height
+		s.Base.Refresh()
+	}
 	s.scrollBy(0, -1*(s.Content.MinSize().Height-s.Size().Height-s.Offset.Y))
-	s.Refresh()
 }
 
 // ScrollToTop will scroll content to container top
 func (s *Scroll) ScrollToTop() {
-	s.scrollBy(0, -s.Offset.Y)
+	s.scrollBy(0, s.Offset.Y)
 }
 
 // DragEnd will stop scrolling on mobile has stopped


andydotxyz:
I have realised that the tests are basically no-ops as the X/Y and assertions were confused.

Try the following test cases (maybe commit them in this PR if you agree with them):
```go
func TestScrollContainer_ScrollToTop(t *testing.T) {
	rect := canvas.NewRectangle(color.Black)
	rect.SetMinSize(fyne.NewSize(100, 500))
	scroll := NewScroll(rect)
	scroll.Resize(fyne.NewSize(100, 50))

	scroll.scrollBy(0, -100)
	assert.Equal(t, float32(100), scroll.Offset.Y)

	scroll.ScrollToTop()
	assert.Equal(t, float32(0), scroll.Offset.Y)
}

func TestScrollContainer_ScrollToBottom(t *testing.T) {
	rect := canvas.NewRectangle(color.Black)
	rect.SetMinSize(fyne.NewSize(100, 500))
	scroll := NewScroll(rect)
	scroll.Resize(fyne.NewSize(100, 50))
	scroll.ScrollToBottom()
	assert.Equal(t, float32(450), scroll.Offset.Y)
}
```

internal/widget/scroller.go
```diff
@@ -393,13 +393,16 @@ func (s *Scroll) CreateRenderer() fyne.WidgetRenderer {
...
+		s.Base.Refresh()
+	}
 	s.scrollBy(0, -1*(s.Content.MinSize().Height-s.Size().Height-s.Offset.Y))
-	s.Refresh()
 }
 
 // ScrollToTop will scroll content to container top
 func (s *Scroll) ScrollToTop() {
-	s.scrollBy(0, -s.Offset.Y)
+	s.scrollBy(0, s.Offset.Y)
```
andydotxyz, line 405: Good catch!

internal/widget/scroller.go
```diff
@@ -393,13 +393,16 @@ func (s *Scroll) CreateRenderer() fyne.WidgetRenderer {
 
 // ScrollToBottom will scroll content to container bottom - to show latest info which end user just added
 func (s *Scroll) ScrollToBottom() {
+	if s.Content.MinSize().Height > s.Size().Height {
+		s.Offset.Y = s.Content.MinSize().Height - s.Size().Height
+		s.Base.Refresh()
```
andydotxyz, line 398: This fix doesn't seem right - making a change then refreshing but then also making another change.

Really it should call `scrollBy` so the fix should be revising the parameters to that method.
And that `Refresh` looks like it's not needed any more.

andydotxyz (changes requested):
Great work but the `ScrollToBottom` does not seem correct - it's got two codepaths and can be simplified as noted inline.

internal/widget/scroller.go
Jacalz: You should also save the result of `MinSize()` and `Size()` instead of computing it multiple times in succession.

TsungPao:
修正好了嗎?

andydotxyz:
Google translate:

> Has it been corrected?

No. If the issue was resulted this would have been closed.

Issue #3786, opened 2023-04-04
CheckGroup and RadioGroup have multiple slices of items


Jacalz:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

When trying to remove some refresh calls from `RadioGroup` and `CheckGroup`, I ran into some strange issues where thing seemed to be refreshed even though I removed callas that should make it not look correct. This led me to look more closely into the code and I then noticed that both (their code is based on each other so the implementation is like 90% the same, hence the single issue) contain multiple slices of items.

The widget has one slice and the renderer has two slices. The slice of `fyne.CanvasObject` within the renderer has to remain to be returned as part of the `Objects()` method but it should not need a second copy within the renderer. The problem is that both copies are getting updated on refresh as well as refreshed.

This duplication of state seems seems unnecessary and removing it would have the benefit of not only making the code easier to understand but also use less memory. I tried to remove one of the slices but I was getting strange test failures. i am struggling to wrap my head around why it is this way.

#### Some examples for RadioGroup (the exact same principle applies for CheckGroup):

First there is the field within the `RadioGroup` item:
https://github.com/fyne-io/fyne/blob/018a2e14545aa1e854bea3b1d590be272ed62080/widget/radio_group.go#L21

Then there is also an entirely separate slice within the renderer:
https://github.com/fyne-io/fyne/blob/018a2e14545aa1e854bea3b1d590be272ed62080/widget/radio_group.go#L135

Both of these are updated and refreshed separately in both the main widget and within the renderer:
https://github.com/fyne-io/fyne/blob/018a2e14545aa1e854bea3b1d590be272ed62080/widget/radio_group.go#L115-L131
https://github.com/fyne-io/fyne/blob/018a2e14545aa1e854bea3b1d590be272ed62080/widget/radio_group.go#L193-L212

### How to reproduce

This is not a visual bug in that sense. The code works fine but is just perhaps more complicated than necessary.

### Example code

See code snippets above.

### Fyne version

v2.3.3

### Go compiler version

1.19.7

### Operating system and version

Fedora Linux 37

PR #4495, opened 2023-12-29
RadioGroup improvements

dweymouth:
### Description:

Fixes: #3786, #3543 (for RadioGroup)

Improvements to RadioGroup
* Now supports multiple items with the same label, the one tapped will be selected only
* Removed slice of `radioItem` from the widget type (only renderer deals with the radioItems)
* Except for the `*_Duplicate*` tests, no test logic is changed; only other test changes are how we access items now that internal impl is different

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/radio_group_extended_test.go
+++ b/widget/radio_group_extended_test.go
@@ -42,7 +42,6 @@ func TestRadioGroup_Extended_Unselected(t *testing.T) {
 		selected = sel
 	})
 	radio.Selected = selected
-	radio.Refresh() // sets up selectedIndex
 	extendedRadioGroupTestTapItem(t, radio, 0)
 
 	assert.Equal(t, "", selected)
--- a/widget/radio_group_extended_test.go
+++ b/widget/radio_group_extended_test.go
@@ -22,7 +22,7 @@ func newextendedRadioGroup(opts []string, f func(string)) *extendedRadioGroup {
 	ret.Options = opts
 	ret.OnChanged = f
 	ret.ExtendBaseWidget(ret)
-	ret.update() // Not needed for extending Radio but for the tests to be able to access items without creating a renderer first.
+	//ret.update() // Not needed for extending Radio but for the tests to be able to access items without creating a renderer first.
 
 	return ret
 }
@@ -32,7 +32,7 @@ func TestRadioGroup_Extended_Selected(t *testing.T) {
 	radio := newextendedRadioGroup([]string{"Hi"}, func(sel string) {
 		selected = sel
 	})
-	radio.items[0].Tapped(&fyne.PointEvent{Position: fyne.NewPos(theme.Padding(), theme.Padding())})
+	extendedRadioGroupTestTapItem(t, radio, 0)
 
 	assert.Equal(t, "Hi", selected)
 }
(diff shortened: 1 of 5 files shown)


Jacalz:
Nice. Well done :)

Jacalz:
A quick look at the failing tests seem to suggest that disabling the widget no longer works as expected?

dweymouth:
> A quick look at the failing tests seem to suggest that disabling the widget no longer works as expected?

Interesting, I tested the disabling manually and it seems to work fine

andydotxyz:
Ah, got it - the group is not passing the item state into the items when It creates them - only in the `Refresh()` call is it applied.

I was able to fix some tests as follows, the rest would be passed similarly I imagine:
```go
@@ -15,9 +15,11 @@ var _ desktop.Hoverable = (*radioItem)(nil)
 var _ fyne.Tappable = (*radioItem)(nil)
 var _ fyne.Focusable = (*radioItem)(nil)
 
-func newRadioItem(label string, onTap func(*radioItem)) *radioItem {
+func newRadioItem(label, sel string, onTap func(*radioItem)) *radioItem {
        i := &radioItem{Label: label, onTap: onTap}
        i.ExtendBaseWidget(i)
+
+       i.Selected = sel == label
        return i
 }
```

dweymouth:
Hmm, good find, not sure how to fix it though as 

> i.Selected = sel == label

is just reintroducing the same issue (all items with the same label would be selected).

Maybe the tests need to be updated since it works in the "real world" so-to-speak

andydotxyz:
It only works real world because it's refreshing after display for some reason.
Initial state should be correct.
Maybe some of the new code needs to be called on setup?

dweymouth:
Looks like all remaining test failures are flukes/unrelated

andydotxyz:
> Looks like all remaining test failures are flukes/unrelated

Agreed. One was fixed on the pending theme widget branch and the other is intermittent.

widget/radio_group_extended_test.go
```diff
@@ -43,7 +42,8 @@ func TestRadioGroup_Extended_Unselected(t *testing.T) {
 		selected = sel
 	})
 	radio.Selected = selected
-	radio.items[0].Tapped(&fyne.PointEvent{Position: fyne.NewPos(theme.Padding(), theme.Padding())})
+	radio.Refresh() // sets up selectedIndex
```
andydotxyz, line 45: This line should not be needed any more?

widget/radio_group_internal_test.go
```diff
@@ -57,15 +57,16 @@ func TestRadioGroup_Unselected(t *testing.T) {
 		selected = sel
 	})
 	radio.Selected = selected
-	radio.items[0].Tapped(&fyne.PointEvent{Position: fyne.NewPos(theme.Padding(), theme.Padding())})
+	radio.Refresh() // sets up selectedIndex
```
andydotxyz, line 60: Is this line needed? I don't think it should be

andydotxyz (changes requested):
This is looking great - but I think two Refreshes in the tests should not be needed now that the initial state is correct?

Jacalz (approved):
I'm happy when Andrew is happy. Thanks for fixing this. I'm glad to see that we allocating less garbage for the GC to deal with :)

andydotxyz (approved):
Perfect thanks. I'll merge this in over the failing tests then cherry pick the fix off my long standing PR to get this resolved too.

PR #4498, opened 2023-12-31
cmd/fyne: Initial migration to fyne.io/tools

Jacalz:
### Description:

This is for https://github.com/fyne-io/tools/pull/1. I don't know how we should handle the migration here. Should the command still run like it does now or should it error out and ask the user if it should automatically download the new tool?

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne/main.go
+++ b/cmd/fyne/main.go
@@ -13,7 +13,8 @@ import (
 )
 
 func main() {
-	fmt.Println("NOTE: This tool is deprecated. Install fyne.io/tools/cmd/fyne for the latest version.")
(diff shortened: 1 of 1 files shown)


Bluebugs:
I would think printing the exact line to give to `go install` would be best. No magic update at this point. Alternate solution would be to include a self update mechanism, but not sure what the reaction from the users to that would be.

Jacalz:
I have updated it to print the installation line.

andydotxyz (approved):
Looks good

PR #4509, opened 2024-01-03
Reduce lock contention in base structs

Jacalz:
### Description:

This improve performance and reduces lock contention by using atomics in base structs for widgets and canvas objects. Some methods like `Position()` and `Size()` are now inlineable and this also fixes various places where there was racy field access bypassing the locks.

The implementation here uses a clever solution, that I came up with, for storing `fyne.Size` (and min in the case of baseObject) and `fyne.Position`  in a single uint64. It uses some clever bit manipulation to store the two raw float32 bytes in a single uint64. This is to make sure that the position and size arguments in `Resize()` and `Move()` do not escape to the heap (like they do when using `atomic.Pointer[]`) and to avoid needing two atomic operations (better performance wise and no synchronisation issues between X and Y coordinates).

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


widget/button.go
```diff
@@ -358,7 +358,7 @@ func (r *buttonRenderer) applyTheme() {
 	r.button.applyButtonTheme()
 	r.label.Segments[0].(*TextSegment).Style.ColorName = theme.ColorNameForeground
 	switch {
-	case r.button.disabled:
+	case r.button.Disabled():
```
andydotxyz, line 361: If we are going with the idea that public methods are called unlocked but private should be within the lock then this isn't the right fix.
If that mantra isn't going to work going forward then we should make another plan for setting expecation of lock/thread safety as we go more into resolving these issues.

andydotxyz (commented):
Very clever approach, nice use of 1.18/1.19 features too.
But I'm not sure we should be attaining more locks inside private helper methods, are those not supposed to be under lock by default?

widget/button.go
andydotxyz: That would be more consistent, yes. If it seems ugly there could be an internal function doing that?

Jacalz: A little more to write but I wouldn't call it ugly :)

Jacalz:
Had to rebase to fix a merge conflict. Should be ready for a review now :)

canvas/base.go
```diff
@@ -11,14 +11,15 @@ import (
 	"sync"
 
 	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/internal/async"
```
dweymouth, line 14: Is `atomic` not a better name for this internal package? Yes it conflicts with the Go builtin `atomic` package, but in most places (including here) we don't need to import the stdlib atomic package, and the naming is much more clear that IMO it is worth it if we occasionally have to alias the import as `fyneAtomic` or similar

dweymouth:
I'm also curious if you've benchmarked this at all to see that lock contention was indeed a problem. (Still a worthwhile change regardless, but I'm curious of the magnitude of improvement we might expect from this.)

Jacalz:
I haven't gotten to benchmarking it. In theory, I can't quite think of any reason for it to be slower. Atomics should be as fast or faster than mutexes, some methods can be unlined now (not possible to inline functions with mutexes and/or defers) and some methods now do less atomic operations (I.e. those that use CAS). We should also avoid stalling the render thread when updating widget state. Any modification of locked widget data would make our render thread stall when trying to get the position, size and MinSize.

I don't really know how I should benchmark it?

canvas/base.go
Jacalz: I figured that we already have the package there so it is better to keep it there and also avoid the conflict?

dweymouth: OK yeah, somehow in my mind I was thinking that internal/async was a new package with this PR

dweymouth (dismissed):

Jacalz:
I put together a quick benchmark to compare the atomic `Size` and `Position` types against simple `sync.Mutex` and `sync.RWMutex` implementations. Just by doing a load and store, we are getting an almost 18x speedup compared to the `sync.RWMutex` and this is without any lock contention whatsoever:

```
goos: linux
goarch: amd64
pkg: fyne.io/fyne/v2/internal/async
BenchmarkAtomicLoadAndStore-8    	408523082	         2.481 ns/op	       0 B/op	       0 allocs/op
BenchmarkMutexLoadAndStore-8     	60645063	        19.00 ns/op	       0 B/op	       0 allocs/op
BenchmarkRWMutexLoadAndStore-8   	26073561	        42.25 ns/op	       0 B/op	       0 allocs/op
```  

Code:
```go
// MutexSize is an atomic version of fyne.Size.
// Loads and stores are guaranteed to happen using a single atomic operation.
type MutexSize struct {
	size fyne.Size
	lock sync.Mutex
}

// Load performs an atomic load on the fyne.Size value.
func (s *MutexSize) Load() fyne.Size {
	s.lock.Lock()
	size := s.size
	s.lock.Unlock()
	return size
}

// Store performs an atomic store on the fyne.Size value.
func (s *MutexSize) Store(size fyne.Size) {
	s.lock.Lock()
	s.size = size
	s.lock.Unlock()
}

// RWMutexSize is an atomic version of fyne.Size.
// Loads and stores are guaranteed to happen using a single atomic operation.
type RWMutexSize struct {
	size fyne.Size
	lock sync.RWMutex
}

// Load performs an atomic load on the fyne.Size value.
func (s *RWMutexSize) Load() fyne.Size {
	s.lock.Lock()
	size := s.size
	s.lock.Unlock()
	return size
}

// Store performs an atomic store on the fyne.Size value.
func (s *RWMutexSize) Store(size fyne.Size) {
	s.lock.Lock()
	s.size = size
	s.lock.Unlock()
}

var globalSize fyne.Size

func BenchmarkAtomicLoadAndStore(b *testing.B) {
	local := fyne.Size{}
	size := async.Size{}

	for i := 0; i < b.N; i++ {
		new := fyne.Size{Width: float32(i), Height: float32(i)}
		if new == size.Load() {
			continue
		}

		size.Store(new)
	}

	globalSize = local
}

func BenchmarkMutexLoadAndStore(b *testing.B) {
	local := fyne.Size{}
	size := async.MutexSize{}

	for i := 0; i < b.N; i++ {
		new := fyne.Size{Width: float32(i), Height: float32(i)}
		if new == size.Load() {
			continue
		}

		size.Store(new)
	}

	globalSize = local
}

func BenchmarkRWMutexLoadAndStore(b *testing.B) {
	local := fyne.Size{}
	size := async.RWMutexSize{}

	for i := 0; i < b.N; i++ {
		new := fyne.Size{Width: float32(i), Height: float32(i)}
		if new == size.Load() {
			continue
		}

		size.Store(new)
	}

	globalSize = local
}
```

internal/async/vector.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func uint64fromTwoFloat32(a, b float32) uint64 {
+	x := uint64(math.Float32bits(a))
+	y := uint64(math.Float32bits(b))
+	return (y << 32) | x
+}
+
+func twoFloat32FromUint64(combined uint64) (float32, float32) {
+	x := uint32(combined & 0x00000000FFFFFFFF)
```
Jacalz, line 49: This might be entirely unnecessary now that I think about it. The downcast ought to clip off the high bits.

internal/async/vector.go
```diff
@@ -0,0 +1,52 @@
...
+	y := uint32(combined & 0xFFFFFFFF00000000 >> 32)
```
Jacalz, line 50: The logical and-operation might be entirely unnecessary here as well. All of the lower bits ought to be shifted out during the right shift.

Jacalz (commented):
Just doing a minor self-review on the phone. Don't mind me :)

dweymouth:
I think it's worth adding a simple round trip unit test (store and load an atomic size and check that it's the same), just since bit manipulation can sometimes be hard to verify just in your head

Jacalz:
@dweymouth That's already been added a few hours ago :)

dweymouth:
> @dweymouth That's already been added a few hours ago :)
> 
> However, I plan to add some more tests of the bit manipulation logic once I've clean that code up to remove the unnecessary logical and operations.

Whoops, my bad :D I think the tests you've added are probably sufficient since they necessarily invoke both bit-shift functions, but up to you if you want to add tests that invoke them directly.

Jacalz:
I added the tests you requested. I also agree what you said about the other tests. Decided to not test it at a lower level because it is highly dependent on bit structure and it’s too much of an internal detail to hard-code into the tests in my opinion :)

dweymouth (approved):

PR #4515, opened 2024-01-05
Support custom icons in a FileIcon

andydotxyz:
By allowing URIs to have an optional `Icon()` method we can have custom items shown in the `FileIcon`

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/uri.go
+++ b/uri.go
@@ -91,3 +91,12 @@ type ListableURI interface {
 	// List returns a list of child URIs of this URI.
 	List() ([]URI, error)
 }
+
+// URIWithIcon describes a URI that should be rendered with a certain icon in file browsers.
+//
+// Since: 2.5
+type URIWithIcon interface {
+	URI
+
+	Icon() Resource
+}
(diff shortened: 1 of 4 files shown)


Jacalz (commented):
My initial reaction to this is that a URI itself shouldn't really have anything to do with what icon it displays. In the widgets case it is a matter of an internal mapping from MimeType to icon.

I think I'd prefer to instead update the FileIcon widget to allow some kind of custom mapping that translates specific paths (or maybe patterns or MimeTypes?) to specific logos. This would allow to more easily just define that `*.go` files should have a Go logo in `defyne` for example, without having to write custom URIs. What do you think?

Jacalz:
I suppose I mean something like a `UriMimeTypeFilter` or whatever we call it for the file dialog but for filtering in custom icons (if that makes sense) 😅

Jacalz (approved):
Never mind. It is a nice solution :)

PR #4522, opened 2024-01-07
Make staticcheck checks more agressive

Jacalz:
### Description:

Poorly chosen identifiers, consistent receiver names in struct methods, and exported APIs should have comments beginning with type name. These are the checks that now are running as well.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Thanks, glad to have a few issues found and fixed

PR #4525, opened 2024-01-07
Fix possible crash in atomic function pointers

Jacalz:
### Description:

Fixes a few cases where we could crash due to two sources of nil.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Thanks for catching this

PR #4529, opened 2024-01-07
Fix the theme of hidden tabs if it changes

### LINKED ISSUE #4344: Changing theme while application is running doesn't change some parameters on some widgets
### Checklist

### Describe the bug

When changing theme from dark to light (or vice versa) some widgets like buttons doesn't change it's theme. E.g buttons doesn't change background until mouseover, never changes font color.

### How to reproduce

1. Create a window with tabs and buttons placed in tabs. Open that app.
2. Open fyne_settings, change theme from dark to light (or vise versa).
3. Switch back to running application, change active tab. Some buttons doesn't change it's background and almost no buttons changed font color.

### Screenshots

image

### Example code

I'll add code later if neccessary. Just create a window, put tabs in it, but some buttons in tabs containers.

### Fyne version

2.3.5, 2.4.1

### Go compiler version

1.21.3

Arch Linux latest @ 20231026

### Additional Information

_No response_

andydotxyz:
Fixes #4344

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
I cannot replicate an test failures locally...

container/tabs_test.go
```diff
@@ -19,3 +23,26 @@ func TestTabButton_Icon_Change(t *testing.T) {
...
+
+	tabs := NewAppTabs(
+		NewTabItem("a", widget.NewLabel("a")),
+		NewTabItem("b", widget.NewLabel("b")))
+	w := test.NewWindow(tabs)
+	w.Resize(fyne.NewSquareSize(100))
+
+	initial := w.Canvas().Capture()
+
+	a.Settings().SetTheme(theme.DarkTheme())
```
dweymouth, line 40: This is really nitpicky since it's just test code, but I feel like we shouldn't be introducing *any* new code that calls deprecated functions (DarkTheme() and LightTheme()). Maybe we need a test theme package that has dark and light themes that aren't publicly exposed?

dweymouth (changes requested):
Glad to see this get fixed - once this lands I can remove a longstanding workaround from Supersonic

container/tabs_test.go
Jacalz: I agree. Then we can turn deprecation warnings back on in staticcheck!

cmd/fyne_demo/main.go
```diff
@@ -238,10 +239,10 @@ func makeNav(setTutorial func(tutorial tutorials.Tutorial), loadPrevious bool) f
 
 	themes := container.NewGridWithColumns(2,
 		widget.NewButton("Dark", func() {
-			a.Settings().SetTheme(&forcedVariant{Theme: theme.DefaultTheme(), variant: theme.VariantDark})
+			a.Settings().SetTheme(test.DarkTheme())
```
dweymouth, line 242: Hmm, not sure how I feel about importing internal/test in fyne demo. I think in this case the duplicated code of the `forcedVariant` theme is good, since it A) makes fyne_demo depend on nothing but the public APIs, so it is a "true" demo of what is possible for 3rd party apps, and B) does show a code example for a custom theme, which is a common user question now, esp. after the deprecation of DarkTheme and LightTheme

dweymouth (changes requested):

dweymouth (approved):

PR #4533, opened 2024-01-10
Reduce calls to theme lookup functions in Slider

dweymouth:
### Description:
A small optimization to reduce calls to theme lookup functions

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz (approved):
Nice work. I think there might still be more widgets that need this treatment. I saw in your Check PR that the Layout function of the Check widget is replicating many calls.

PR #4538, opened 2024-01-12
Support disabling the screen blank/lock

### LINKED ISSUE #4534: Allow apps to block screen locking/screensaver
### Checklist

I was thinking of building a mobile monitoring app, and wanted the screen to remain on while the app is running.

I looked into it and there are simple native APIs for it.

On iOS you can call
```objective-c
[[NSOperationQueue mainQueue] addOperationWithBlock:^ {
    [UIApplication sharedApplication].idleTimerDisabled = YES;
}];
```
(Thanks Andy for finding this, I only found the Swift API)

On Android it's split into two options, either keep screen on or keep CPU running, not entirely sure we need the latter and the former doesn't require permissions:
```java
public class MainActivity extends Activity {
  @Override
  protected void onCreate(Bundle savedInstanceState) {
    super.onCreate(savedInstanceState);
    setContentView(R.layout.activity_main);
    getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
  }
}
```
https://developer.android.com/develop/background-work/background-tasks/scheduling/wakelock
It looks like there's also a C version of this code but I could only find mention of it and not the API itself.

_No response_

I think probably just add to the mobile driver API a method that calls the native APIs, preferably something that can be set and reset, so you can keep screen on for specific long lasting tasks and let it lock otherwise

andydotxyz:
API proposed and implementation for iOS and Android.
Windows/macOS/Linux X11&Wayland could be another PR?

Fixes #4534

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/device.go
+++ b/device.go
@@ -31,6 +31,11 @@ type Device interface {
 	IsBrowser() bool
 	HasKeyboard() bool
 	SystemScaleForWindow(Window) float32
+
+	// SetDisableScreenBlanking allows an app to ask the device not to sleep/lock/blank displays
+	//
+	// Since: 2.5
+	SetDisableScreenBlanking(bool)
 }
 
 // CurrentDevice returns the device information for the current hardware (via the driver)
--- a/device.go
+++ b/device.go
@@ -31,11 +31,6 @@ type Device interface {
 	IsBrowser() bool
 	HasKeyboard() bool
 	SystemScaleForWindow(Window) float32
-
-	// SetDisableScreenBlanking allows an app to ask the device not to sleep/lock/blank displays
-	//
-	// Since: 2.5
-	SetDisableScreenBlanking(bool)
 }
 
 // CurrentDevice returns the device information for the current hardware (via the driver)
--- a/canvas/base.go
+++ b/canvas/base.go
@@ -11,14 +11,15 @@ import (
 	"sync"
 
 	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/internal/async"
 )
 
 type baseObject struct {
-	size     fyne.Size     // The current size of the canvas object
-	position fyne.Position // The current position of the object
-	Hidden   bool          // Is this object currently hidden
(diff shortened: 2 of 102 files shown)


dweymouth:
Should this really be a desktop API as well? Isn't asking the screen not to sleep pretty much a mobile-only thing?

andydotxyz:
I was going on what the linked issue described - the way that macOS, Windows and Linux allow apps to stop screensaver in the same way the mobile stops the screen switch off / lock.

dweymouth:
> I was going on what the linked issue described - the way that macOS, Windows and Linux allow apps to stop screensaver in the same way the mobile stops the screen switch off / lock.
> 
> > to reflect that some OSes/DEs may refuse the request or not support it at all
> 
> I've not found one that can block this request yet.

Ah, OK, hadn't seen the latest comment activity on the ticket. I'd say if we can at least verify that we can support it on Windows, Mac, and X11, then we can go with the `Set...` naming as proposed in this PR. The actual implementation could wait until a follow-up PR

device.go
```diff
@@ -31,6 +31,11 @@ type Device interface {
 	IsBrowser() bool
 	HasKeyboard() bool
 	SystemScaleForWindow(Window) float32
+
+	// SetDisableScreenBlanking allows an app to ask the device not to sleep/lock/blank displays
```
dweymouth, line 35: The `Device` interface comment will need updating since it no longer just "provides information" about the device

dweymouth (changes requested):

device.go
dweymouth: Alternatively, the API could be on the Driver type

andydotxyz: Good point - best not to change the intent of an interface ;)

dweymouth (approved):
I think this is good to go now, as long as we remember to add the desktop implementation before pulling it into a release

PR #4539, opened 2024-01-14
Clean up code for Radio and Check and fix TODO about icons

Jacalz:
### Description:

Fixes a TODO about exposing new icons for filled Radio and Check. Also cleans up some of the rendering code while at it.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/check.go
+++ b/widget/check.go
@@ -283,21 +283,25 @@ func (c *checkRenderer) MinSize() fyne.Size {
 
 // Layout the components of the check widget
 func (c *checkRenderer) Layout(size fyne.Size) {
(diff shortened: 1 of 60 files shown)


widget/check.go
```diff
@@ -337,9 +339,6 @@ func (c *checkRenderer) updateResource() {
 		bgRes.ColorName = theme.ColorNameBackground
 	}
 	if c.check.disabled.Load() {
-		if c.check.Checked {
-			res = theme.NewThemedResource(theme.CheckButtonCheckedIcon())
-		}
```
Jacalz, line 342: For the reviewer: This was removed because it is already handled by the if-statement above.

andydotxyz (approved):
Oh thanks for catching this

Issue #3607, opened 2023-01-26
Translation


EssGeeEich:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [X] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

At this point in time, there is no one way to handle translations in fyne.

### Is it possible to construct a solution with the existing API?

Currently, translation is partially feasible using external libraries.

### Describe the solution you'd like to see.

It would be nice to have a built-in translation (maybe even localization?) system.

andydotxyz:
Absolutely a great idea - something we plan to start work on soon.
I think this is a duplicate of #2928?

EssGeeEich:
I do not think that this is, strictly speaking, a duplicate of #2928.
Internationalization is a big project in its entirety - which should be more or less split in, I believe, three topics:
- Translation (obviously)
- Localization (eg. time/date formats vary between different countries, think date format differences between US/EU)
- Unicode/Font support (eg. handling and rendering the characters properly, which I suppose this is mostly good to go thanks to golang strings, but fyne seems to be lacking a system font support, as mentioned in the other topic)

While translation is part of internationalization, I think translation, localization and Unicode/rendering should be split apart and not be mixed up, work-wise, moreover #2928 was mostly about the font side.
Another message will follow with a proposal of what I was thinking.

andydotxyz:
Thanks - just FYI the unicode/rendering side should be ready, as of `v2.3.0`.

EssGeeEich:
API-wise, I propose the following user-facing APIs:

**fyne/v2/lang**
`lang.Language`
Literally just `type Language int`.
Basically an enumeration that will contain every language.

`lang.*`
Constants that represent a language (with a type of `lang.Language`).

`lang.LoadLanguage(lang.Language, Resource) error`
Loads a Resource as a localization language.
This might be a file format that might allow users to use websites like Crowdin.

`lang.SetCurrentLanguage(lang.Language) error`
Selects a language as the current one.

`lang.CurrentLanguage() lang.Language`
Retrieves the current language.
Should return english by default.

`lang.Tr(text string) string`
Translates a string into the currently selected language.
Falls back to english if available, returns the original string otherwise.

`lang.TrExt(text string, disambiguation string, count int) string`
Translates a string into the currently selected language.
[...]

PR #4545, opened 2024-01-16
Add support for translating applications

### LINKED ISSUE #2928: About internationalization related support
I'm trying to develop desktop app which should display chinese.but chinese is not supported by fyne's default font .

I know I can change font to display chinese by set theme , but I want reserve default font and display chinese.

I try to reserve more than one font,and can set the priority of different fonts, if the font with the highest priority cannot be displayed, try to find a font with a lower priority.

Its might be this:

`

t := getFontsByPath(path)
app.Settings().addfont(t)

`

or

`

t := getFonts.ByPath(path)
app.Settings().Setfonts(default,t)

`

I'm wondering if you need this feature.
sry for my poor english

andydotxyz:
Open for comment or testing at this time. Still to do:

- [x] Providing locale info ~with setter~ in Device metadata
- [x] Allow apps to inject their own translations
- [x] Translating plurals and other complex structures

possibly follow-on is:

- [ ] Create tool to parse code and generate template translation
- [ ] Update app if device language changes
- [ ] Add localisation for time, date, numbers and money

Fixes #2928, #3607

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


go.mod
```diff
@@ -18,10 +18,12 @@ require (
 	github.com/go-text/typesetting v0.1.0
 	github.com/godbus/dbus/v5 v5.1.0
 	github.com/jackmordaunt/icns/v2 v2.2.6
+	github.com/jeandeaual/go-locale v0.0.0-20220711133428-7de61946b173
```
Jacalz, line 21: Isn't it a bit worrying that the package hasn't been updated for two years and imports Fyne v1?

andydotxyz: It's the type of library where it's coded once and it just works so that aspect does not worry me.
I could offer them a v2 upgrade for Fyne and see where it goes?

Jacalz: I think that would be good. We don't want to compile two versions of Fyne into each application 😅

andydotxyz:
I decided that locale get/set didn't make sense. Limited it to a getter on Device for now

lang/lang.go
```diff
@@ -0,0 +1,185 @@
...
+		return
+	}
+
+	for _, f := range files {
+		data, err := fs.ReadFile(dir + "/" + f.Name())
+		if err != nil {
+			fyne.LogError("failed to read bundled translation", err)
+			continue
+		}
+		bundle.MustParseMessageFileBytes(data, f.Name())
```
Jacalz, line 171: This will panic if it can't parse, right? Do we want that?

lang/lang.go
```diff
@@ -0,0 +1,185 @@
...
+func addLanguage(data []byte, name string) error {
+	_, err := bundle.ParseMessageFileBytes(data, name)
+	return err
+}
+
+func init() {
+	bundle = i18n.NewBundle(language.English)
+	bundle.RegisterUnmarshalFunc("json", json.Unmarshal)
+
+	loadTranslationsFromFS(translations, "translations")
```
Jacalz, line 134: Should we really parse all translations for all languages on startup? This will not only slow down startup with many languages but also add extra memory pressure. Would it not be better to only load the requested language?

lang/lang.go
```diff
@@ -0,0 +1,185 @@
...
+		data, err := fs.ReadFile(dir + "/" + f.Name())
+		if err != nil {
+			fyne.LogError("failed to read bundled translation", err)
+			continue
+		}
+		bundle.MustParseMessageFileBytes(data, f.Name())
+
+		name := "en"
+		if !strings.Contains(f.Name(), "template") {
+			name = f.Name()[5 : len(f.Name())-5]
```
Jacalz, line 175: This might panic if the length is too short.

lang/lang.go
```diff
@@ -0,0 +1,185 @@
...
+		fyne.LogError("Translation failure", err)
+		return fallbackWithData(key, fallback, d0)
+	}
+	return ret
+}
+
+// AddTranslations allows an app to load a bundle of translations.
+// The language that this relates to will be inferred from the resource name, for example "fr.json".
+// The data should be in json format.
+func AddTranslations(r fyne.Resource) error {
```
Jacalz, line 115: Just adding a comment here that we discussed on the call about if resource is the right type here? With `embed` being more and more popular it might be a bit annoying to wrap it into a resource (and I and maybe more people mostly think of resources as images). Thoughts?

Jacalz (changes requested):
Looking nice. A few thoughts inline :)

lang/lang.go
andydotxyz: How will you know which language files are needed?
If the user reports a Brazilian locale we need to load the Portuguese language, for example (if nothing exists for Brazil). 
Similarly Canadian French locale is an English language country with a French language pack. The lookup code will do this correctly, but only if we have loaded the languages needed.

If performance is an issue we could split the parse into two, loading all the files first to do the lookup and then parsing the messages only from those needed. However we must also ensure that the fallback path is honoured, so that an "en_GB" and "en" are not assumed to be the same and the more specific could provide only the messages which differ.

lang/lang.go
andydotxyz: I think we agreed to add embed.FS support as an additional option - as we need the file name and file content to support the features, which is where the translations for locale method comes in.
I can add an FS helper though.

lang/lang.go
andydotxyz: As before this is used for internal and we have "base." prefix.
I can add safety though if this method may ne used for other purposes.

andydotxyz: Fixed anyhow, this was getting exposed to developers through the suggested FS public API

lang/lang.go
andydotxyz: Added

lang/lang.go
andydotxyz: Fixed anyhow, this was getting exposed to developers through the suggested FS public API

andydotxyz:
@Jacalz I think the fixes are all in place, can you give this the go-ahead?

Jacalz (approved):
Nice. We might want to make sure that this doesn't slow down application starts goign forward but it works fine now :)

PR #4547, opened 2024-01-17
Add a workflow that fyne developer can call to help with Fyne QA.

Bluebugs:
### Description:
Introduce a GitHub Action workflow that Fyne developer can call from their repository to ensure the coming release are of the highest quality.

Likely the README content can be improved to make sure people use this with the right intention. After all this is not a paid service, but something to help an open source project.

### Checklist:

#### Where applicable:

### DIFF
--- a/.github/workflows/fyne-qa-all.yml
+++ b/.github/workflows/fyne-qa-all.yml
@@ -1,0 +1,29 @@
+name: Run tests on all maintained fyne branch
+
+on:
+  workflow_call:
+    inputs:
+      tags:
+        description: 'Tags to pass when running tests'
+        required: false
+        default: ''
+        type: string
+
+jobs:
+  qa-develop:
+    uses: fyne-io/fyne/.github/workflows/fyne-qa.yml@develop
+    with:
+      branch: develop
+      tags: ${{ inputs.tags }}
+
+  qa-master:
+    uses: fyne-io/fyne/.github/workflows/fyne-qa.yml@develop
+    with:
+      branch: master
+      tags: ${{ inputs.tags }}
+        
+  qa-v24:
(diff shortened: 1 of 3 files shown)


Jacalz:
Should this really be in the main Fyne repository?

andydotxyz:
> Should this really be in the main Fyne repository?

Oh an interesting thought. Maybe the new tools repo is the most logical place for such build support?

Jacalz:
Yeah, I think so too. We might want more workflows. Like we talked about in Edinburgh, it might make sense to have a workflow that help with building Fyne and installing dependencies etc.

Bluebugs:
I am not sure, this workflow is related to the version of fyne as a dependencies. So if we ever need to adjust anything related to a change in dependencies in the develop branch, we will need to patch the workflow at the same time we land the change in develop. I think they are definitively something that must be in sync. It is why the direction I give in the README is to use @develop version, so as to pick the one that is always in sync with the dependencies of the develop branch.

Jacalz:
Our tools repo is going to have to follow the Fyne version more or less either way given that we have the `cmd/fyne` tool there as well. I don't think we should be adding lots of external workflows (there might be more ones that makes sense) to this repository.

Bluebugs:
Just thinking about this a bit more. The fyne qa workflow should rely on fyne cmd tool from its master branch only. If there is a need for a fyne cmd qa workflow, it should test with fyne master branch only. User won't keep them in sync (like today) and testing both scenario are important in this case. I can update this workflow accordingly and provide a matching pr for the fyne tool repository.

andydotxyz:
The fyne tool and dependency version should always be kept the same. Some apps (Android mostly) won't compile or run if they are not in sync.
This is something to be worked on, but for now it is a requirement.

PR #4551, opened 2024-01-21
Improve performance when creating progressbar renderers

Jacalz:
### Description:

Around a 30+% performance improvement generally. Caches some icon values and allocates the canvas objects inline into the outer struct instead of allocating them in multiple places. We cal likely apply similar changes in many more places and avoid allocating the struct in one place and then allocate each canvas object in a different place. Less for the GC to handle and we run less of a risk for nil pointer de-references.

Benchstat output over 10 runs:
```
goos: linux
goarch: amd64
pkg: fyne.io/fyne/v2/widget
                 │   old.txt   │               new.txt               │
                 │   sec/op    │   sec/op     vs base                │
Progressbar-8      491.7n ± 5%   287.4n ± 3%  -41.55% (p=0.000 n=10)
ProgressbarInf-8   491.8n ± 6%   392.5n ± 2%  -20.21% (p=0.000 n=10)
geomean            491.8n        335.8n       -31.71%

                 │  old.txt   │              new.txt               │
                 │    B/op    │    B/op     vs base                │
Progressbar-8      452.0 ± 0%   420.0 ± 0%   -7.08% (p=0.000 n=10)
ProgressbarInf-8   452.0 ± 0%   356.0 ± 0%  -21.24% (p=0.000 n=10)
geomean            452.0        386.7       -14.45%

                 │  old.txt   │              new.txt               │
                 │ allocs/op  │ allocs/op   vs base                │
Progressbar-8      7.000 ± 0%   4.000 ± 0%  -42.86% (p=0.000 n=10)
ProgressbarInf-8   7.000 ± 0%   5.000 ± 0%  -28.57% (p=0.000 n=10)
geomean            7.000        4.472       -36.11%
```

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/progressbar.go
+++ b/widget/progressbar.go
@@ -15,8 +15,8 @@ import (
 
 type progressRenderer struct {
 	widget.BaseRenderer
-	background, bar *canvas.Rectangle
-	label           *canvas.Text
+	background, bar canvas.Rectangle
+	label           canvas.Text
 	progress        *ProgressBar
 }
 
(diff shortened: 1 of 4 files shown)


dweymouth (approved):
Nice improvement!

PR #4552, opened 2024-01-21
Optimize radio item renderer to be created 50% faster

Jacalz:
### Description:

Instead of allocating one struct for the renderer and one struct for each canvas object that is part of the renderer, we can embed the types into the renderer and just allocate the larger struct once. This makes for less allocations but we are also theoretically helping the processor cache by keeping the items closer together in memory instead of in different locations. I also made sure to avoid setting a few fields twice. Doing this makes the code not only cleaner but also almost 50% faster (likely due to fewer allocations but using the fields later on might also be faster) for creating the renderer.

```
goos: linux
goarch: amd64
pkg: fyne.io/fyne/v2/widget
                      │   old.txt   │               new.txt               │
                      │   sec/op    │   sec/op     vs base                │
RadioCreateRenderer-8   664.0n ± 2%   351.4n ± 2%  -47.09% (p=0.000 n=10)

                      │  old.txt   │              new.txt              │
                      │    B/op    │    B/op     vs base               │
RadioCreateRenderer-8   736.0 ± 0%   704.0 ± 0%  -4.35% (p=0.000 n=10)

                      │  old.txt   │              new.txt               │
                      │ allocs/op  │ allocs/op   vs base                │
RadioCreateRenderer-8   9.000 ± 0%   4.000 ± 0%  -55.56% (p=0.000 n=10)
```

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/radio_item.go
+++ b/widget/radio_item.go
@@ -37,19 +37,8 @@ type radioItem struct {
 //
 // Implements: fyne.Widget
 func (i *radioItem) CreateRenderer() fyne.WidgetRenderer {
-	focusIndicator := canvas.NewCircle(color.Transparent)
-	icon := canvas.NewImageFromResource(theme.RadioButtonFillIcon())
-	over := canvas.NewImageFromResource(theme.NewThemedResource(theme.RadioButtonIcon()))
-	label := canvas.NewText(i.Label, theme.ForegroundColor())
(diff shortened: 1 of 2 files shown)


andydotxyz (approved):
Sweet improvement :)

PR #4558, opened 2024-01-24
Unwrap multiple layers of themed resources

dweymouth:
### Description:

This PR adds a new `unwrapResource` function to retrieve the "base" resource from a nested chain of themed resource types, and calls it in the Name and Content functions of all ThemedResource types to achieve two improvements:

* Avoid multiple calls to svg.Colorize, that allocate excess memory and waste CPU
* Avoid caching multiple versions of the same SVG rasterization with differing prefix names like "error_foreground_base-resource.svg" and "error_base-resource.svg".

Fixes #(issue)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Thanks

Issue #4360, opened 2023-11-03
Scroll bar should have rounded corners, at least in its expanded state


dweymouth:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

It looks like the scroll bars were missed when rounded rectangles were added to the builtin widgets

### How to reproduce

Run fyne demo

### Example code

Fyne demo

### Fyne version

2.4.1

### Go compiler version

go1.21.1

### Operating system and version

macOS Ventura M2

dweymouth:
On develop for 2.5

PR #4559, opened 2024-01-24
Scroll and scrollbar padding and rounding

ZolAnder85:
### Description:

I was trying to achieve a non-overlapping scrollbar and rounded, non-touching scrollbar-thumbs.
I ended up coding it so I can use it.
I think – since I have already done it – it is worth adding as it could only be achieved by rewriting the whole scroller source.
With the default theme values I added, results should look identical.

I don't know what version you would be adding this, so please add the since comments.
I hope this still counts as a "small contribution" :)

Implements #4360 too, if you tweak the default theme.

Here is an image what I was looking for:

Demo


Jacalz (changes requested):
Hi. Thanks for the contribution. Having rounded scrollbars is a really nice change. There are a few things I'd like to changed here so please see below:

- Per the PR template, we do all of the feature development on the `develop` branch. Would you mind targeting that instead of `master`?

- Instead of just making it possible to make the scrollbar rounded, can you make it rounded in the default theme? The idea with https://github.com/fyne-io/fyne/issues/4360 is of course to have that as the default after all.

- Would you mind separating rounded scrollbar and padding into two separate PRs? I see that as two different changes and the latter might require a bit more discussions.

> I don't know what version you would be adding this, so please add the since comments.

We only add bug fixes to `2.4.x` (and other path releases) so please add `Since: 2.5` lines.

Again, we really do appreciate the contribution. I hope my many requested changes put you off :)

Jacalz:
Forgot a `don't` in the last sentance (edited that in afterwards). I don't mean to put you off is what I meant :sweat_smile:

Andari85:
I did understand it either way, don't worry. I have limited amount of time to spend on this, so it might take some time.
I added these as one PR, because I felt that corner rounding without the padding does not look right. I also did not want to decide to change default looks. But I am glad if you like it. Anyway, I hope I won't have to make you wait for too long.

andydotxyz (changes requested):
This looks like a nice improvement, but it seems like it could be done without 5 new size types and associated package methods.
One thing to note is that we have moved to `theme.Size(...)` to avoid API explosion so that cuts the API surface in half.

It is not immediately obvious what the 3 different padding values are - a padding to avoid overlap makes sense but the others don't so much. Additionally I don't think there is much value in making the "limit" customisable - we should always avoid them being too small for tapping.

Andari85:
Sorry, but I have no time for this. If you want to add these changes, feel free to use the changes. I will keep the fork for now, but I will eventually remove it.

andydotxyz:
> Sorry, but I have no time for this. If you want to add these changes, feel free to use the changes. I will keep the fork for now, but I will eventually remove it.

Sorry to hear that. If you have time later and would like to come back to it please do, and consider the comments left earlier.
Best of luck.

Issue #4156, opened 2023-08-14
High cpu usage when drawing a table with 288 items


roffe:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

Displaying a simple table uses 6-14% CPU on a Intel i9-9900k.

Moving the mouse pointer around in the window spikes 1 core to almost 100% CPU

clicking and marking a cell is noticeably laggy and takes upwards half a second for the blue border to move

I can see the update function is constantly being called just by moving the mouse pointer in the window which might be the culprit causing it to call SetText on all 288 entries even though nothing changed

### How to reproduce

Run example program which has 288 data points that needs to be displayed in the table

### Example code

```go
package main

import (
	"strconv"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

var yValues = []int{6200, 5820, 5440, 5060, 4680, 4300, 3920, 3540, 3160, 2780, 2400, 2020, 1640, 1260, 880, 700}
var xValues = []int{25, 75, 125, 180, 240, 300, 360, 420, 480, 540, 600, 660, 720, 780, 840, 900, 1050, 1300}

var valueMap = [][]float64{
	{0.86, 0.90, 0.92, 0.94, 0.97, 1.00, 1.11, 1.13, 1.15, 1.16, 1.17, 1.19, 1.21, 1.24, 1.32, 1.32, 1.40, 1.45},
	{0.86, 0.90, 0.92, 0.94, 0.97, 1.00, 1.02, 1.11, 1.13, 1.13, 1.14, 1.15, 1.17, 1.18, 1.27, 1.30, 1.37, 1.42},
	{0.85, 0.89, 0.92, 0.95, 0.97, 1.01, 1.01, 1.01, 1.09, 1.10, 1.11, 1.11, 1.13, 1.14, 1.21, 1.25, 1.35, 1.40},
	{0.86, 0.90, 0.93, 0.95, 0.97, 0.99, 1.00, 1.01, 1.02, 1.07, 1.09, 1.10, 1.12, 1.13, 1.15, 1.22, 1.29, 1.34},
[...]

PR #4560, opened 2024-01-24
Reduce CPU consumption of Entry cursor by around 80%

### LINKED ISSUE #1946: Focused text entry consumes 40%-80% of CPU in my app

andydotxyz:
### Description:
Reduce CPU load by around 80% by fading over only 1/5th the time.
Also avoid re-calculating MinSize when the visible content has not changed.

Fixes #1946, #4156

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -13,6 +13,7 @@ import (
 	"fyne.io/fyne/v2/data/binding"
 	"fyne.io/fyne/v2/driver/desktop"
 	"fyne.io/fyne/v2/driver/mobile"
+	"fyne.io/fyne/v2/internal/async"
 	"fyne.io/fyne/v2/internal/cache"
 	"fyne.io/fyne/v2/internal/widget"
 	"fyne.io/fyne/v2/theme"
@@ -96,7 +97,7 @@ type Entry struct {
 	binder          basicBinder
 	conversionError error
 	lastChange      time.Time
-	minCache        *fyne.Size
+	minCache        async.Size
 	multiLineRows   int // override global default number of visible lines
 
 	// undoStack stores the data necessary for undo/redo functionality
@@ -400,11 +401,9 @@ func (e *Entry) KeyUp(key *fyne.KeyEvent) {
 //
 // Implements: fyne.Widget
 func (e *Entry) MinSize() fyne.Size {
-	e.propertyLock.RLock()
-	cached := e.minCache
-	e.propertyLock.RUnlock()
-	if cached != nil {
-		return *cached
+	cached := e.minCache.Load()
+	if !cached.IsZero() {
+		return cached
 	}
 
 	e.ExtendBaseWidget(e)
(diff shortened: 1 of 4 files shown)


dweymouth (changes requested):

widget/entry.go
```diff
@@ -474,6 +485,14 @@ func (e *Entry) Redo() {
 	e.Refresh()
 }
 
+func (e *Entry) Refresh() {
+	e.propertyLock.Lock()
+	e.minCache = nil
+	e.propertyLock.Unlock()
+
+	e.BaseWidget.Refresh()
+}
```
Jacalz, line 494: I'm not saying that it need to be done now or should be done, but maybe it would make sense to have some kind of `MinSize` cache built in to the basewidget? I suppose it is something we might want to do for all widgets more or less (unless the computation is very cheap).

widget/entry.go
```diff
@@ -96,6 +96,7 @@ type Entry struct {
 	binder          basicBinder
 	conversionError error
 	lastChange      time.Time
+	minCache        *fyne.Size
```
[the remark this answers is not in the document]
Jacalz: I suppose the default 0,0 value might be fine as a not-set value because Entry never gets that small, right?

widget/entry.go
andydotxyz: I did propose the BaseWidget inclusion some time back but there was pushback for a reason I cannot recall.
Let's get it in here and see how useful it is and it could be pulled up without impact later.

widget/entry.go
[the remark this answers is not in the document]
andydotxyz: Thanks, updated

dweymouth (approved):

widget/entry.go
Jacalz: It might be worth exploring again. Either as part of the base widget or as some kind of ready cache-system that developers can add to their widgets (maybe using `async.Size` internally?). We should at least look at caching MinSize in more widgets :)

dweymouth:
This should be pulled into v2.4.5 if possible as it's another huge optimization, but like the one for animation runner, it may need minor reworking to not use `async.Size` and just use a regular fyne.Size with locking instead

Jacalz:
It might be worth just waiting for v2.5.0 if the changes are as big as they are

dweymouth:
> It might be worth just waiting for v2.5.0 if the changes are as big as they are

I think it might just be undoing the last commit https://github.com/fyne-io/fyne/pull/4560/commits/42ab4dde88610c031a44a330e71e329bd1e936e7

Plus, if the new threading model is landed in 2.5, which is the current plan, async.Size may not even be needed anymore as widgets can assume their API calls would arrive from a single thread

dweymouth:
This is a *huge* optimization for any Fyne app where the user spends any non-trivial amount of time in text entry. I think we really should get it in 2.4.5

PR #4566, opened 2024-01-26
Dismiss the iOS soft keyboard if 'Done' button is tapped

andydotxyz:
### Checklist:

- [ ] Tests included. <- requires iOS hardware to run (simulator no help)
- [x] Tests all pass.


dweymouth (approved):
I am not setup to test on iOS but this looks simple enough and I trust that you've validated it :)

PR #4569, opened 2024-01-27
Don't allocate new slices on each animation tick

### LINKED ISSUE #4445: Animation runner allocates new slices every frame
### Checklist

### Describe the bug

When animations are running, the animation runner allocates a new slice of active animations 60 times a second [here](https://github.com/fyne-io/fyne/blob/master/internal/animation/runner.go#L73). This leads to a lot of work for the garbage collector and possibly increased system memory use of Fyne apps.

### How to reproduce

Run any app that has animations. This is not really a bug though but a ticket to track implementing memory optimizations in the animation runner.

### Screenshots

_No response_

### Example code

n/a

### Fyne version

2.4.2

### Go compiler version

n/a

n/a

### Additional Information

_No response_

dweymouth:
### Description:

Another try at fixing the excess allocations in the animation runner in a simpler way than #4451, with no public API change. This does not address the non-idempotency of starting an animation (one can still be started twice), but aims only to avoid the performance penalty of allocating slices every tick.

Fixes #4445 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/animation/animation_test.go
+++ b/internal/animation/animation_test.go
@@ -47,6 +47,8 @@ func TestGLDriver_StopAnimation(t *testing.T) {
 		t.Error("animation was not ticked")
 	}
 	run.Stop(a)
+	// Animations are really stopped asynchronously
+	time.Sleep(time.Second/60 + 100*time.Millisecond)
 	run.animationMutex.Lock()
 	assert.Zero(t, len(run.animations))
 	run.animationMutex.Unlock()
--- a/internal/animation/animation_test.go
+++ b/internal/animation/animation_test.go
@@ -47,9 +47,9 @@ func TestGLDriver_StopAnimation(t *testing.T) {
 		t.Error("animation was not ticked")
 	}
 	run.Stop(a)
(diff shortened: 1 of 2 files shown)


andydotxyz (approved):
Thanks

Jacalz:
Starting a goroutine allocates memory as well. Have you benchmarked that this improves performance?

dweymouth:
> Starting a goroutine allocates memory as well. Have you benchmarked that this improves performance?

No, though I was operating under the assumption that since animations tick 60 times a second, the number of ticks will vastly outnumber the number of Stop calls, so worse Stop performance was acceptable. The other PR I have in draft for 2.5 would overall be a better solution, though it still has some unresolved issues

Jacalz:
Oh yeah, you are absolutely right there. Benchmarking is very important though and I think we ought to do so for any non-trivial optimizations. If not doing so for making sure that we don't accidentally regress, then do it  to make me excited for the nice improvements ;)

You can leave it this time though since it's already merged and I agree that you are right with your statement :)

Issue #4572, opened 2024-01-27
tests fail with thousands of "Assertion failure in -[NSMenu _lockForMainMenuItemArray], NSMenu.m:1106"


josharian:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

Run `go test ./...`. Result is thousands (or more, I didn't wait for it to stop spewing) of lines like:

```
2024-01-27 07:58:05.095 glfw.test[33753:5608525] *** Assertion failure in -[NSMenu _lockForMainMenuItemArray], NSMenu.m:1106
```

### How to reproduce

See above.

### Example code

Again, see above.

### Fyne version

caf50f6407d7f8a5abc91064338ce9bc3b1820bb

### Go compiler version

1.21.6

### Operating system and version

macOS Sonoma

### Additional Information

I would be happy to try my hand at a fix, but I don't see any clear hints from grepping the source code about where to even start looking.

andydotxyz:
Do these actually cause a failure, or were they outputting some warnings only?

Jacalz:
I wonder if this is M1/M2 specific and potentially even a bug in GLFW upstream? This is not something I’ve seen locally on Linux with amd64

andydotxyz:
I don't think we have seen it on our Apple test devices either. Perhaps it was lost amongst the image issues.

josharian:
For my future reference, here's the tail of the failure:

```
[lots of these lines]
2024-01-27 15:56:26.282 glfw.test[63243:5971505] *** Assertion failure in -[NSMenu _lockForMainMenuItemArray], NSMenu.m:1106
panic: test timed out after 10m0s
running tests:
	TestDarwinMenu (9m56s)

goroutine 121 [running]:
testing.(*M).startAlarm.func1()
	/Users/josh/go/1.21/src/testing/testing.go:2259 +0x320
created by time.goFunc
	/Users/josh/go/1.21/src/time/sleep.go:176 +0x38

goroutine 1 [select, locked to thread]:
fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).runGL(0x140000fe180)
	/Users/josh/x/fyne/internal/driver/glfw/loop.go:120 +0x14c
fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).Run(0x140000fe180)
[...]

PR #4575, opened 2024-01-28
make image comparison tests less brittle

### LINKED ISSUE #4573: image rendering tests are fragile
### Checklist

### Describe the bug

Run `go test ./...`. I get multiple test failures around rendering images not matching perfectly. The diffs appear quite small.

Example:

```
       	           	--- Expected
        	           	+++ Actual
        	           	@@ -4251,3 +4251,3 @@
        	           	 00010990  ff fd fd ff ff ff ff ff  f2 c0 c0 ff c8 01 01 ff  |................|
        	           	- 000109a0  c8 00 00 ff d4 35 35 ff  ff ff ff ff ff ff ff ff  |.....55.........|
        	           	+ 000109a0  c8 00 00 ff d4 36 36 ff  ff ff ff ff ff ff ff ff  |.....66.........|
        	           	 000109b0  ff ff ff ff ff fd fd ff  ff ff ff ff cd 17 17 ff  |................|
        	Test:       	TestDocTabs_ApplyTheme
        	Messages:   	Image did not match master. Actual image written to file:///Users/josh/x/fyne/container/testdata/failed/doctabs/desktop/single_custom_theme.png.
FAIL
FAIL	fyne.io/fyne/v2/container	0.789s
```

### How to reproduce

See above.

### Screenshots

_No response_

### Example code

Again, see above.

### Fyne version

caf50f6407d7f8a5abc91064338ce9bc3b1820bb

### Go compiler version

1.21.6

macOS Sonoma

### Additional Information

How would you like to approach fixing this? Do you want to allow some tolerance in pixel values? Or have multiple allowed images it can match?

josharian:
- internal/test: allow images to be +/- 1 RGB value per pixel

### Description:

This provides robustness when the rendering engine
behaves (ever so) slightly differently.

Fixes #4573 (I think, I can't be entirely sure
because my tests still fail because of #4572)

While we're here, let gofumpt add an octal marker to 0o755.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/test/util.go
+++ b/internal/test/util.go
@@ -8,6 +8,7 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -40,6 +41,13 @@ func AssertImageMatches(t *testing.T, masterFilename string, img image.Image, ms
 	masterPix := pixelsForImage(t, raw) // let's just compare the pixels directly
 	capturePix := pixelsForImage(t, img)
 
+	// On darwin/arm64, there are slight differences in the rendering.
+	// Use a slower, more lenient comparison. If that fails,
+	// fall back to the strict comparison for a more detailed error message.
+	if runtime.GOOS == "darwin" && runtime.GOARCH == "arm64" && pixCloseEnough(masterPix, capturePix) {
+		return true
+	}
+
 	var msg string
 	if len(msgAndArgs) > 0 {
 		msg = fmt.Sprintf(msgAndArgs[0].(string)+"\n", msgAndArgs[1:]...)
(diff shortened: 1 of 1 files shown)


josharian:
(Test failures were real...didn't see them locally because they were masked by even more test failures. Seems like CI coverage on macOS is somehow insufficient?)

andydotxyz:
Is this still with good performance? last time I tried to do a pixel iteration route it was 20* slower than the assert library comparison

josharian:
> Is this still with good performance?

Does it matter in practice? I did a quick spot check with this on and off and none of the tests were obviously slower at the top level.

I can certainly make it faster...at the cost of time and effort and code complexity. I'd suggest that we start by making it work, and then make it faster if/when it turns out to matter.

andydotxyz:
> I can certainly make it faster...at the cost of time and effort and code complexity. I'd suggest that we start by making it work, and then make it faster if/when it turns out to matter.

That sounds reasonable. However I wonder - could it be put into an GOOS check (as this only impacts M1/M2 or arm64) meaning that anything not Darwin (or arm64?) still runs the original faster code path?

josharian:
Done. PTAL.

andydotxyz:
Nice one thanks

andydotxyz (approved):
Nice one thanks

PR #4577, opened 2024-01-28
widget: allow slight floating point variation

josharian:
This test was failing locally. Exact comparison of floating point values is fraught.

### Checklist:

- [x] Tests included.
- [ ] Tests all pass.


josharian:
Test failure appears to be unrelated.

andydotxyz (approved):
Sweet, I didn't know about `InEpsilon` that's handy

PR #4578, opened 2024-01-28
add clickable icon

Artem3213212:
### Description:
Add new widget ClickableIcon

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/clickable_icon.go
+++ b/widget/clickable_icon.go
@@ -1,0 +1,77 @@
+package widget
+
+import (
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/canvas"
+	"fyne.io/fyne/v2/theme"
+)
+
+type ClickableIcon struct {
+	BaseWidget
+
+	OnTapped func() `json:"-"`
+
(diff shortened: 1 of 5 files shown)


Jacalz:
Thanks for the contribution, we appreciate you taking the time to write this up. However, this is already something you can do easily be extending the regular icon widget or automate with [wrapper.MakeTappable](https://pkg.go.dev/fyne.io/x/fyne/wrapper#MakeTappable) and a `widget.Icon{}`. I don’t think we need another widget to do so.

Also FYI, please target PRs towards the `develop` as per the information in the contributor documentation and PR template.

andydotxyz:
This also has a huge amount of duplication with the `Icon` type - if you embed that in another type it is just the case of adding the tap handler and a constructor.

Just so you know you can add these handlers to any widget through extension. We do not want (or need) to add such handler versions of all widgets to make it possible.
This documentation even shows how to make what you have opened in this PR https://docs.fyne.io/extend/extending-widgets

Issue #4568, opened 2024-01-26
data races in fyne demo


josharian:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

I started seeing what looked like memory corruption in my (very simple) app. I checked with the race detector, and saw races. I didn't see any of my code in the data races, so I ran the demo app, and it reproduces with the demo app 100%. So I think there is an issue with fyne itself.

### How to reproduce

```
go run -race fyne.io/fyne/v2/cmd/fyne_demo@latest
```

Click on Canvas.

Result:

```
==================
WARNING: DATA RACE
Read at 0x00c00148ebf8 by main goroutine:
  fyne.io/fyne/v2/internal/painter/gl.(*painter).newGlLinearGradientTexture()
      /Users/josh/pkg/mod/fyne.io/fyne/v2@v2.4.3/internal/painter/gl/texture.go:113 +0x68
  fyne.io/fyne/v2/internal/painter/gl.(*painter).newGlLinearGradientTexture-fm()
      <autogenerated>:1 +0x44
  fyne.io/fyne/v2/internal/painter/gl.(*painter).getTexture()
      /Users/josh/pkg/mod/fyne.io/fyne/v2@v2.4.3/internal/painter/gl/texture.go:36 +0x50
  fyne.io/fyne/v2/internal/painter/gl.(*painter).drawTextureWithDetails()
      /Users/josh/pkg/mod/fyne.io/fyne/v2@v2.4.3/internal/painter/gl/draw.go:193 +0x70
  fyne.io/fyne/v2/internal/painter/gl.(*painter).drawGradient()
      /Users/josh/pkg/mod/fyne.io/fyne/v2@v2.4.3/internal/painter/gl/draw.go:35 +0x84
  fyne.io/fyne/v2/internal/painter/gl.(*painter).drawObject()
      /Users/josh/pkg/mod/fyne.io/fyne/v2@v2.4.3/internal/painter/gl/draw.go:85 +0x1c4
  fyne.io/fyne/v2/internal/painter/gl.(*painter).Paint()
      /Users/josh/pkg/mod/fyne.io/fyne/v2@v2.4.3/internal/painter/gl/painter.go:93 +0x6c
  fyne.io/fyne/v2/internal/driver/glfw.(*glCanvas).paint.func1()
[...]

PR #4582, opened 2024-01-29
Make sure that Accordion is using locking

Jacalz:
### Description:

There are more widgets to fix up but this gets Accordion up to speed as an initial test and to make sure that I am on the right track. Tests all pass and `fyne_demo` does not show any deadlocks so far.

For #4568 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/accordion.go
+++ b/widget/accordion.go
@@ -28,33 +28,45 @@ func NewAccordion(items ...*AccordionItem) *Accordion {
 
 // Append adds the given item to this Accordion.
 func (a *Accordion) Append(item *AccordionItem) {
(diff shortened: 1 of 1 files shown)


dweymouth (approved):

dweymouth:
Ah, I should have reviewed this more closely. Took another look and I see some issues. For one, between the two critical sections in RemoveIndex, the length could change again and the index that was previously checked to be in range could now be out of range. The whole thing up until Refresh needs to be a single locked section.

dweymouth:
Made PR #4589 to fix

PR #4583, opened 2024-01-30
Attempt to make Container fully thread-safe

dweymouth:
### Description:
This is an attempt to make Container fully thread-safe, as long as user code never directly accesses the `Objects` property after construction (and uses the provided methods instead). In order to do this the Objects slice must frequently be copied while holding a lock before proceeding into an operation that cannot be done under lock (ie calls user code) due to risk of deadlock. To avoid generating lots of garbage we pool the `[]CanvasObject` slices in a tiered pool system.

This PR is using copy-on-read, or more precisely, copy-before-reading-in-non-critical-section ;) for the Objects slice. It adds an O(n) factor to various container operations, but has bounded allocations due to slice pooling. An alternative would be to use copy-on-write (see #4600), where every time the Objects slice is changed a new slice is generated and it is never updated in place. This would have better performance when changes to container contents are infrequent, but has no real upper-bound on the amount of GC garbage that could be generated in the case where container contents are being frequently updated.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container.go
+++ b/container.go
@@ -92,6 +92,7 @@ func (c *Container) AddObject(o CanvasObject) {
 func (c *Container) Hide() {
 	c.lock.Lock()
 	if c.Hidden {
+		c.lock.Unlock()
 		return
 	}
 
--- a/container.go
+++ b/container.go
@@ -86,6 +86,9 @@ func (c *Container) Hide() {
 // MinSize calculates the minimum size of a Container.
 // This is delegated to the Layout, if specified, otherwise it will mimic MaxLayout.
 func (c *Container) MinSize() Size {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
 	if c.Layout != nil {
 		return c.Layout.MinSize(c.Objects)
 	}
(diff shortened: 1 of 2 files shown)


Jacalz:
I think holding the lock once calling refresh is causing deadlocks?

dweymouth:
> I think holding the lock once calling refresh is causing deadlocks?

Hmm, I was thinking some test is stuck waiting for the lock, since fyne_demo ran fine (fyne ;) with this change. In any case, we need to hold the lock throughout the whole duration of looping over Objects, though maybe it's not even worth it at this point since there's no way Objects can be made thread-safe since it's a public field

andydotxyz:
We have found that it is required to remove locks when calling out to public/user code

dweymouth:
Yeah I think I will close this PR as the design does not allow for Container to be fully thread-safe anyway

Jacalz:
It is probably worth it to make it as thread safe as possible in the other areas through. Does decrease the total number of races

Jacalz:
Having most other issues solved will likely also simplify things when we do the rest?

dweymouth:
The problem is that we need to lock during the Refresh call, since we're iterating over Objects. But we also can't, since the call to Refresh on the child objects potentially invokes user code. What we really need to I guess is deep copy the slice with the lock, then release the lock and iterate over that copy. This generates a lot of garbage though unless we pool the slices, and since there are thread safety errors like this all over Fyne, I wonder if it makes sense to introduce an internal, centralized pool of `[]CanvasObject` slices, and that should be done as part of a large effort all across the codebase?

dweymouth:
Also, in a sense, locking to protect access to Objects doesn't really solve anything, at least not perfectly, since user code can still change it at any time. But at least it would guard against Fyne code being responsible for a data race

Jacalz:
I mostly mean that we can add locking around the other fields. I'm entirely aware that the slices are problematic :)

dweymouth:
will not be needed with new threading model

PR #4589, opened 2024-02-01
fix accordion locking

dweymouth:
### Description:
There was a locking error in accordion where we would check the length of Items with a lock, then drop the lock, and later re-acquire it to do something that is dependent on length. The length could have changed during the interim so the lock needs to be held for the entire time.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz (approved):
Thanks. You are indeed right. Nice catch!

PR #4590, opened 2024-02-01
Make Activity start and stop idempotent

dweymouth:
### Description:
Make sure start and stop are idempotent so the activity can't be started twice.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/activity.go
+++ b/widget/activity.go
@@ -2,6 +2,7 @@ package widget
 
 import (
 	"image/color"
+	"sync/atomic"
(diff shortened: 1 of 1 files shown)


Jacalz (dismissed):
This seems to create a deadlock for the Activity tab?

Jacalz (approved):

Jacalz:
Nice and clean solution. I really like atomics and the CAS operation :)

PR #4592, opened 2024-02-01
Add locking to Hyperlink

dweymouth:
### Description:

Protects access to hyperlink.Text and hyperlink.Url which are not safe to access concurrently.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz (commented):

dweymouth:

Good catch. Added renderer locking.

andydotxyz (dismissed):
Thanks for this!

Jacalz (changes requested):
This seems to deadlock when going the welcome screen in `fyne_demo`.

dweymouth:
Ah, bad merge I guess. Will take a look.

Jacalz (approved):

PR #4593, opened 2024-02-01
Add missed locking to Check

dweymouth:
### Description:
Lock when setting properties.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz (commented):

dweymouth:

There was indeed a missed lock in one place in the renderer. Addressed

andydotxyz (approved):
Yup.

Over time hopefully we can move to the private methods as standard being able to assume a lock. It's not where the code is at right now but one of those things that I think should become automatic as we standardise on this.

PR #4594, opened 2024-02-01
Add locking to Button

dweymouth:
### Description:

Add locking to Button for thread safety

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz (commented):

dweymouth:

I think it does? The read lock is held for the duration of buttonRenderer.Refresh and for the second half of Layout when it accesses button properties, and also in CreateRenderer of the button.

Jacalz:
> I think it does? The read lock is held for the duration of buttonRenderer.Refresh and for the second half of Layout when it accesses button properties, and also in CreateRenderer of the button.

It is not held during MinSize from what I can see.

dweymouth:
> > I think it does? The read lock is held for the duration of buttonRenderer.Refresh and for the second half of Layout when it accesses button properties, and also in CreateRenderer of the button.
> 
> It is not held during MinSize from what I can see.

The renderer MinSize function actually doesn't read any properties from the button struct. We could add a lock to the renderer but I think Fyne actually calls renderer methods single-threaded, if I understand correctly.

PR #4598, opened 2024-02-01
GH-4595: Add SetView function to file dialog

### LINKED ISSUE #4595: Allow application to set default view (list/grid) of file dialog before showing it
### Checklist

I think the issue has already been described here: https://github.com/fyne-io/fyne/issues/2165
I essentially have the same pains, but instead of having to click on the list-view toggle button all of the time, I would like to show the dialogs with that view by default.

It might be able to get a reference to the button, as is done in the tests, but I would prefer a more clear API.
https://github.com/fyne-io/fyne/pull/2251/files#diff-3144205fac6ab76f332735320fee204ddb6963bcc76a40684806537d5a999534R441

When creating a dialog, like `dialog.NewFileOpen()` I would propose to have a method on the returned object like `SetDefaultView()` which lets me select whether to use the `Grid` or `List` view. I would expect that this is set before the `Show()` method is called.

With some help from you (i.e. drafting the solution together), I would be happy to help implement it.

maruu:
### Description:
This change adds a `SetView()` function to `FileDialog`s.
It can be used to set the default view of the `FileDialog`.
The view can still be changed with the toggle button.
If the user has already stored a preference for the view in its user settings, then this overrides the default view setting for the `FileDialog`.

Please note that this changes the public API of the FileDialog.
Also, I changed the indizes of the views, but there is an update path for users which already have preferences stored somewhere. The default is still the GridView, as it has been before.

Fixes #4595 

### Checklist:

- [x] Tests included.
- [x] Tests all pass. - the internal/driver/glfw drivers seem to fail, although I did not change anything in this code area. Is this currently expected on the develop branch?

#### Where applicable:


dialog/file_test.go
```diff
@@ -505,13 +505,86 @@ func TestView(t *testing.T) {
...
+	assert.Equal(t, theme.ListIcon(), toggleViewButton.Icon)
+}
+
 func TestViewPreferences(t *testing.T) {
 	win := test.NewWindow(widget.NewLabel("Content"))
 
 	prefs := fyne.CurrentApp().Preferences()
 
 	// set viewLayout to an invalid value to verify that this situation is handled properly
-	prefs.SetInt(viewLayoutKey, -1)
```
maruu, line 514: I am not sure when this would be the case.
I checked that `Preferences().Int()` returns 0 if the key is not found, so I think the current code should be safe and backward compatible.

dialog/file.go
```diff
@@ -17,11 +17,12 @@ import (
 	"fyne.io/fyne/v2/widget"
 )
 
-type viewLayout int
```
dweymouth, line 20: Public comments are needed for all the new exported types and constants, including the "Since: 2.5" line.

maruu: thanks! Done

maruu:
> It also got me thinking... should this really override user preference? In general we assume that the user knows their circumstance better than app developer...

I think there might be a misunderstanding. I tried to implement it in such a way, that if there is a preference already set, the preference overrules the view set via SetView() before showing the dialog.

dialog/file.go
```diff
@@ -774,6 +787,17 @@ func (f *FileDialog) SetFileName(fileName string) {
 	}
 }
 
+// SetView changes the default display view of the FileDialog
```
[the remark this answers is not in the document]
maruu: At the moment I only tested the SetView() function if it is called before the dialog is shown (i.e. which is why it states in the documentation that the function should be called before showing the dialog, and which is why I chose the wording "default display view" here). Let me know if you want anything changed.
Personally, I only see the usecase of setting the view before the dialog is shown. Afterwards, when the dialog is already active, the user can toggle it via the respective button anyway.

andydotxyz:
Ah then I misread what you said sorry.

But given you call the internal "setView" if the dialog exists then I'm pretty sure this will change the view any time you call the method.

maruu:
> But given you call the internal "setView" if the dialog exists then I'm pretty sure this will change the view any time you call the method.

I think that is correct, yes. But what would also be required is to handle the state of the toggleViewButton as well, as this is not done in the setView function at the moment. As I stated earlier, I specifically focused on the usecase that the view should be changed before the dialog is opened, so I can avoid this. Let me know what you think, and whether you wish to extend the solution so it also handles the state of the toggleViewButton, so the `SetView` function also properly works when called when the dialog is already open.

andydotxyz:
I guess it should either be extended to update the button status to be applicable at all calls.
Or reduced to only be used to set the default and never actually change the display.
The former seems more useful - we don't have any other setters that only work before display I don't think.

maruu:
> I guess it should either be extended to update the button status to be applicable at all calls.
> The former seems more useful - we don't have any other setters that only work before display I don't think.

I updated the PR and added tests to check if the dialog is updated properly, even if it is already opened.
Let me know what you think :-)

maruu:
Rebased to current develop branch state

maruu:
I am not sure why the "Mobile Tests / mobile_tests (1.21.x) (pull_request)" tests are failing. Is this already known/expected?

andydotxyz:
Please don't rebase and force push a PR that has begun review - it means we can't see change history.

dialog/file.go
```diff
@@ -555,15 +554,22 @@ func (f *fileDialog) setView(view ViewLayout) {
...
 		f.files = grid
+		toggleViewButtonIcon = theme.ListIcon()
 	} else {
 		list := widget.NewList(count, template, update)
 		list.OnSelected = choose
 		f.files = list
+		toggleViewButtonIcon = theme.GridIcon()
 	}
+
+	f.toggleViewButton.SetIcon(toggleViewButtonIcon)
```
andydotxyz, line 571: I think it's 2 lines less code if you just call `SetIcon` instead of adding the `toggleViewButtonIcon` variable in this context

andydotxyz:
For these tests to pass I think you need to merge in develop again - there was a bad state in the static check a few weeks back and I think that is when you rebased.

dialog/file.go
Jacalz: FYI: It is common practice for us to let the one opening the review comment mark them as done and not the one fixing them. It is fine for now but it is worth noting.

dialog/file.go
maruu: Thanks. I also found another occurrence that seemed too verbose

maruu:
> Please don't rebase and force push a PR that has begun review - it means we can't see change history.

Sure, thanks for the feedback. I am still learning the open-source contribution etiquette and feedback like this helps me get better. I will switch to merging the develop state from now on

dialog/file.go
maruu: > FYI: It is common practice for us to let the one opening the review comment mark them as done and not the one fixing them. It is fine for now but it is worth noting.

Sure, thanks for the feedback. I am still learning the open-source contribution etiquette and feedback like this helps me get better

Jacalz: No problem. Happy to guide you :)

maruu:
Hi @andydotxyz any idea when we can continue with the review process here? Thanks!

dialog/file_test.go
andydotxyz: As the comment above says, it is setting an invalid value which is a good test case.
Preferences could be set by an external force and so we should be safe and handle unexpected values.

andydotxyz (commented):
Sorry, I got caught in a serious backlog.

Not sure about the change to a test-case to remove one of the things it is testing - can that go back to how it was before?

dialog/file_test.go
maruu: Thanks, you are right. I added code to handle invalid values as well.

andydotxyz (approved):
Thanks so much, this looks good to me now.

maruu:
Thanks for finding this bug - it should be fixed now. Are you usually testing the default minsizes of popups? If the minsizes are important, I think it could make sense to write a test for it. Do you want me to write a test for it?

andydotxyz:
> I think it could make sense to write a test for it. Do you want me to write a test for it?

Agreed, but in this instance I don't think it's needed.
The reason being that I noticed another bug that the view you have chosen (last time) changes the minsize of the file dialog.
This bug should be resolved but it was there before you opened this PR.

The sizes are now back to what they were and we can look at the minsize in another change.

andydotxyz (approved):
Excellent, thanks so much for this

Issue #4599, opened 2024-02-01
Add more Objects manipulation functions to Container, and deprecate direct slice access


dweymouth:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [X] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

As an ongoing effort for thread-safety, user code should be discouraged from manipulating exported fields directly, especially slices. We should provide several more functions to manipulate the Objects slice in Container in a thread-safe way, such as:

* container.Replace(objA, objB)
* container.ReplaceAt(int, obj)
* container.RemoveAt(int)
* container.Swap(int, int)
* container.Insert(int, obj)
* container.SetObjects(objects) - for whole content replacement

etc...

Basically make it possible to do everything needed without directly manipulating the Objects slice, and then deprecate access to the Objects slice directly.

### Describe the solution you'd like to see.

described above.

dweymouth:
This won't be necessary with the new proposed threading model.

andydotxyz:
Won't some of these like SetObjects still be helpful?

dweymouth:
Sure, I guess so. I suppose it depends if we want to standardize that Setters are thread-safe (and internally `wid.SetFoo(fooVal)` is basically `fyne.RunOnMain(func() { wid.Foo = fooVal; wid.Refresh() })`

If we go this way, then yes `container.SetObjects` could just be runOnMain ( () => container.Objects = obj )

dweymouth:
Without the new threading model, `container.SetObjects` would be absolutely needed, as in there's no other safe way to accomplish it. With the new thread model, it would be a shorthand

andydotxyz:
I think where we got to on the call is that Setters are guaranteed thread safe, as was intended in current code. It may just be a shorthand but it's an intuitive one.

PR #4600, opened 2024-02-01
Make Container thread-safe with copy-on-write

dweymouth:
### Description:
An alternate approach from #4583 to make Container thread-safe. This time using copy-on-write, where any mutation to c.Objects *MUST* create a new slice, so that a shallow copy - ie just the slice header - of c.Objects can be taken under lock which is then safe to iterate through without a lock (since it's immutable).

Compared to #4583 this approach would have better performance when container contents are not frequently mutated, ie read operations on c.Objects are much more common than writes. But it has no upper bound on how much GC garbage it can generate if container contents are being frequently updated.

Choosing which implementation depends on what we expect the use patterns of Container to be. In any case, we need #4599 in combination with either of these approaches, and also to make sure the renderer locks the container lock when it reads the container contents as well.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/container.go
+++ b/container.go
@@ -48,7 +48,7 @@ func NewContainerWithLayout(layout Layout, objects ...CanvasObject) *Container {
 	}
 
 	ret.size = layout.MinSize(objects)
-	ret.layout()
+	ret.layout(objects, ret.size)
 	return ret
 }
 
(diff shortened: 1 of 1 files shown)


andydotxyz:
Which to choose is complex. I expect that a border container content changes rarely, but that a stack possibly changes a lot.
But we have to solve this with the right for majority cases as I don't think multiple different solutions hold up in the library.

> any mutation to c.Objects MUST create a new slice

What is the implication if they don't? As I expect a lot of folk won't do it or may forget and/or older code won't have that approach and the upgrade will likely be transparent.

dweymouth:
> What is the implication if they don't? As I expect a lot of folk won't do it or may forget and/or older code won't have that approach and the upgrade will likely be transparent.

The implication would be that, with this solution, the following is *not* thread safe in this approach but would be OK in the other approach:

```go
container.lock.Lock()
container.Objects[0] = newObj // or anything that changes the contents of container.Objects.
                                                   // must create and assign new objects slice instead
container.lock.Unlock()
```

In either case, accessing the Objects slice would have to be protected by a lock, which is why we need to deprecate the field for direct access by user code, at least after container construction, and provide enough objects manipulation functions on the container that are thread-safe.

andydotxyz:
That's cool, I was just checking I suppose that it was nothing more serious than being unsafe in that mode. Like possible panic etc. A solution that has corners that are still not safe may be OK in the meantime as it's better than we have and does not require any breaking change.

PR #4606, opened 2024-02-02
Correctly handle URIs like c: on Windows

### LINKED ISSUE #4605: URIs for Windows (c:) are handled incorrectly
### Checklist

### Describe the bug

Currently in fyne, Windows file URIs look like: `file://c:/path` which is actually incorrect as in that notation, the mandatory separator for the authority is missing. This would correctly be `file:///c:/path`.
See for example: https://en.wikipedia.org/wiki/File_URI_scheme

### How to reproduce

See example code

### Screenshots

_No response_

### Example code

```go
// Create a file uri in Windows with c:\ notation
uri := storage.NewFileURI(`c:\foo\bar`)
// This test currently fails
assert.Equal(t, "file:///c:/foo/bar", uri.String())
```

### Fyne version

0ace386

### Go compiler version

1.21.6

Windows 11

### Additional Information

This should be cleaned up so that
1. The URI format is now correct
2. This allows further improvements in the file storage, especially regarding support of UNC paths (will be a separate ticket).

Roemer:
### Description:
This change makes sure that file URIs on Windows with drive letters are handled correctly.
It kind of is a breaking change but old formats (`storage.NewURI("file://C:\\foo\\bar\\baz\\")`) are still accepted but converted to the new format.

Fixes #4605

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/internal/repository/file.go
+++ b/internal/repository/file.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
+	"runtime"
 	"strings"
 
 	"fyne.io/fyne/v2"
@@ -170,6 +172,11 @@ func (r *FileRepository) Parent(u fyne.URI) (fyne.URI, error) {
 
 	// trim the scheme
 	s = strings.TrimPrefix(s, fileSchemePrefix)
+
+	// Only for Windows: If the path is a drive root, no parent is possible
+	if runtime.GOOS == "windows" && regexp.MustCompile(`(?i)^/[a-z]:$`).MatchString(s) {
+		return nil, repository.ErrURIRoot
+	}
 
 	// Completely empty URI with just a scheme
 	if s == "" {
--- a/storage/repository/parse.go
+++ b/storage/repository/parse.go
@@ -3,6 +3,7 @@ package repository
 import (
 	"errors"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 
@@ -22,6 +23,10 @@ func NewFileURI(path string) fyne.URI {
 	// or NT style paths, with / or \, but when we reconstruct
 	// the URI, we want to have / only.
 	if runtime.GOOS == "windows" {
+		// Make sure that Windows paths (eg "c:\") also correctly start with a "/"
+		if regexp.MustCompile(`(?i)^[a-z]:`).MatchString(path) {
+			path = "/" + path
+		}
 		// seems that sometimes we end up with
 		// double-backslashes
 		path = filepath.ToSlash(path)
--- a/storage/uri_test.go
+++ b/storage/uri_test.go
@@ -114,7 +114,7 @@ func TestURI_Scheme(t *testing.T) {
 func TestURI_Name(t *testing.T) {
 	assert.Equal(t, "text.txt", storage.NewURI("file:///text.txt").Name())
 	assert.Equal(t, "library.a", storage.NewURI("file:///library.a").Name())
-	assert.Equal(t, "directory", storage.NewURI("file://C:/directory/").Name())
+	assert.Equal(t, "directory", storage.NewURI("file:///C:/directory/").Name())
 	assert.Equal(t, "image.JPEG", storage.NewFileURI("C:/image.JPEG").Name())
 }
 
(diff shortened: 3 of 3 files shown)


internal/repository/file.go
```diff
@@ -171,6 +173,11 @@ func (r *FileRepository) Parent(u fyne.URI) (fyne.URI, error) {
 	// trim the scheme
 	s = strings.TrimPrefix(s, fileSchemePrefix)
 
+	// Only for Windows: If the path is a drive root, no parent is possible
+	if runtime.GOOS == "windows" && regexp.MustCompile(`(?i)^/[a-z]:$`).MatchString(s) {
```
Jacalz, line 177: Compiling the regexp each time is going to be incredibly slow. It is better to cache this in some way. The initial thought would be to use a global variable but it seems unnecessary to do that for all operating systems.

Roemer: I suspect there are not thousands of these calls per second in 99.99% of all applications, but I can for example move it to a local "global" variable and compile it only once there. Or to a utility method and put the global there. The downside is: globals are not nice.
The alternative could be a utility struct like:
```go
var windowsDrivePattern = regexp.MustCompile(`(?i)^/[a-z]:$`)
type uriUtility struct {
}

func (f *uriUtility) MatchesDrivePattern(path string) bool {
	return windowsDrivePattern .MatchString(path)
}
```

dweymouth (changes requested):
Can we add a Windows case to `TestURI_Path` as well?

internal/repository/file.go
dweymouth: Personally I have no problem with a `var windowsDrivePattenRegex = regexp.MustCompile(...)` or similar name added to the top of this file, and then just matching directly against it here (ie no need for `uriUtility`). @Jacalz do you have a different opinion?

storage/uri_test.go
```diff
@@ -52,6 +52,11 @@ func TestURIPath(t *testing.T) {
 	u, err = storage.ParseURI(s)
 	assert.Nil(t, err)
 	assert.Equal(t, "example:animal:ferret:nose", u.Path())
+
+	s = "file:///C:/over/there"
+	u, err = storage.ParseURI(s)
+	assert.Nil(t, err)
+	assert.Equal(t, "/C:/over/there", u.Path())
```
dweymouth, line 59: Returning a path like this for Windows is wrong. The logic needs to be updated in the `Path` function to strip out the extra slash for file:// urls on Windows.

dweymouth (changes requested):

internal/repository/file.go
Roemer: I added the variables now.

storage/uri_test.go
Roemer: True, I added the logic accordingly. There are more and more `if runtime.GOOS == "windows"` and even more will come when UNC will be supported as well but I don't see a way around this.

dweymouth: Hmm, a question I guess (and maybe this is one for @andydotxyz) - do we want to use runtime.GOOS as the trigger or do we want to handle "windows-looking" file URLs the same (ie correctly) regardless of what OS we're running on?

Roemer: If this is discussed, the handling of UNC paths on Windows should also be discussed.
Simple example:
`file://server/path` in Windows almost always means that any file operation then should be done on `\\server\path`, which is the UNC path build out of this URI. All `os` and `io` methods on Windows can handle the UNC path correctly but would fail with `server/path`.

internal/repository/file.go
```diff
@@ -16,6 +18,9 @@ import (
 // for string processing
 const fileSchemePrefix string = "file://"
 
+// Regex pattern to match Windows drive paths
```
andydotxyz, line 21: it's a drive identifier isn't it - not a path?

storage/repository/parse.go
```diff
@@ -11,6 +12,9 @@ import (
 	"fyne.io/fyne/v2"
 )
 
+// Pattern to find Windows drive paths
+var windowsDrivePrefixhPattern = regexp.MustCompile(`(?i)^[a-z]:`)
```
andydotxyz, line 16: is `prefixh` a typo?

storage/repository/parse_test.go
```diff
@@ -20,9 +20,9 @@ func TestParseURI(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, "file:///tmp/foo.txt", uri.String())
 
-	uri, err = ParseURI("file://C:/tmp/foo.txt")
```
andydotxyz, line 23: You've stated that legacy URI will still work - so please leave the old test in so we can ensure it stays working.

storage/repository/uri.go
```diff
@@ -13,6 +15,9 @@ import (
 // Declare conformance with fyne.URI interface.
 var _ fyne.URI = &uri{}
 
+// Regex pattern to match Windows drive paths
```
andydotxyz, line 18: This probably needs a note about them being URI related or the slash in the middle may be unobvious or look like duplication of the other prefix pattern

andydotxyz (changes requested):
Thanks for this, a couple of small things inline

andydotxyz:
I think there are some missed test changes as well...

PR #4610, opened 2024-02-04
Cache some computations and values in GridWrap

dweymouth:
### Description:

Improve GridWrap performance by caching some theme calls and the column count calculation.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/gridwrap.go
+++ b/widget/gridwrap.go
@@ -44,6 +44,7 @@ type GridWrap struct {
 	itemMin       fyne.Size
 	offsetY       float32
 	offsetUpdated func(fyne.Position)
+	colCountCache int
 }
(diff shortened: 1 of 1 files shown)


andydotxyz (commented):
It's not immediately obvious from the PR, but will this work if you don't use the constructor, initial state setting can be problematic.

dweymouth:
> It's not immediately obvious from the PR, but will this work if you don't use the constructor, initial state setting can be problematic.

Good catch, I just updated the uninitialized value to 0, since we were actually treating 1 as the first valid column count.

andydotxyz (approved):
Cool, thanks

PR #4614, opened 2024-02-06
Add IME support

### LINKED ISSUE #618: The input method is located outside the app
The **input indicator box** is located outside the app. 
windows and mac.  
like this:  
avatar
avatar

kanryu:
### Description:
As everyone knows, fyne does not currently support IME. This is fundamentally caused by the fact that glfw does not support IME, and cannot be resolved by fyne's efforts alone. So I solved this problem by adding a hack to both go-gl/glfw and fyne. Please check it on your PC.

Fixes #618

## Supported environments

- supported
  - Windows
  - Mac
  - Linux(X11, Wayland)
- not supported
  - Android
  - iOS

## How to try this

```bash
git clone https://github.com/kanryu/fyne
git clone https://github.com/kanryu/glfw
cd fyne
go mod edit -replace github.com/go-gl/glfw/v3.3/glfw=../glfw/v3.4/glfw
go run cmd/fyne_demo/main.go
```

https://github.com/fyne-io/fyne/assets/759165/89c858dc-beaf-4b19-948f-77baee4aa2a6

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/canvasobject.go
+++ b/canvasobject.go
@@ -75,6 +75,14 @@ type Focusable interface {
 	TypedKey(*KeyEvent)
 }
 
+// Preeditable describes any CanvasObject that can respond to IME inputs.
+// Originally, it should be a method of Focusable, but only a few widgets need PreeditChanged,
+// and adding a method to Focusable will have a big impact on others, so defined as a new interface.
+type Preeditable interface {
+	// PreeditChanged is a hook called by window with IME is on and typed multi-byte characters inputed
+	PreeditChanged(preedit string)
+}
+
 // Scrollable describes any CanvasObject that can also be scrolled.
 // This is mostly used to implement the widget.ScrollContainer.
 type Scrollable interface {
--- a/canvasobject.go
+++ b/canvasobject.go
@@ -75,12 +75,16 @@ type Focusable interface {
 	TypedKey(*KeyEvent)
 }
 
+type CursorPositionChangedCallback func(pos Position)
+
 // Preeditable describes any CanvasObject that can respond to IME inputs.
 // Originally, it should be a method of Focusable, but only a few widgets need PreeditChanged,
 // and adding a method to Focusable will have a big impact on others, so defined as a new interface.
 type Preeditable interface {
 	// PreeditChanged is a hook called by window with IME is on and typed multi-byte characters inputed
 	PreeditChanged(preedit string)
+
+	ReceiveCursorPositionChangedCallback(callback CursorPositionChangedCallback)
 }
 
 // Scrollable describes any CanvasObject that can also be scrolled.
--- a/canvasobject.go
+++ b/canvasobject.go
@@ -75,7 +75,7 @@ type Focusable interface {
 	TypedKey(*KeyEvent)
 }
 
-type CursorPositionChangedCallback func(pos Position)
+type CursorPositionChangedCallback func(pos Position, size Size)
 
 // Preeditable describes any CanvasObject that can respond to IME inputs.
 // Originally, it should be a method of Focusable, but only a few widgets need PreeditChanged,
--- a/internal/driver/common/canvas.go
+++ b/internal/driver/common/canvas.go
@@ -191,6 +191,15 @@ func (c *Canvas) Focused() fyne.Focusable {
 	return mgr.Focused()
 }
 
+// Preeditable return the current preeditable object only when focused
+func (c *Canvas) Preeditable() fyne.Preeditable {
+	focused := c.Focused()
+	if focused == nil {
+		return nil
+	}
+	return focused.(fyne.Preeditable)
+}
+
 // FocusGained signals to the manager that its content got focus.
 // Valid only on Desktop.
 func (c *Canvas) FocusGained() {
--- a/internal/driver/common/canvas.go
+++ b/internal/driver/common/canvas.go
@@ -197,7 +197,10 @@ func (c *Canvas) Preeditable() fyne.Preeditable {
 	if focused == nil {
 		return nil
 	}
-	return focused.(fyne.Preeditable)
+	if preeditable, ok := focused.(fyne.Preeditable); ok {
+		return preeditable
+	}
+	return nil
 }
 
 // FocusGained signals to the manager that its content got focus.
--- a/internal/driver/glfw/loop_desktop.go
+++ b/internal/driver/glfw/loop_desktop.go
@@ -11,8 +11,11 @@ import (
 	"github.com/go-gl/glfw/v3.3/glfw"
 )
 
+const GLFW_X11_ONTHESPOT glfw.Hint = 0x00052002
+
 func (d *gLDriver) initGLFW() {
 	initOnce.Do(func() {
+		glfw.InitHint(GLFW_X11_ONTHESPOT, 1) // enable IME callbacks, hint must be set before Init()
 		err := glfw.Init()
 		if err != nil {
 			fyne.LogError("failed to initialise GLFW", err)
--- a/internal/driver/glfw/window_desktop.go
+++ b/internal/driver/glfw/window_desktop.go
@@ -685,10 +685,11 @@ func (w *window) imeStatus(_ *glfw.Window) {
 	fmt.Println("ime status ", counterIme)
 	counterIme++
 	if preeditable := w.canvas.Preeditable(); preeditable != nil {
+		// This callback function is called by the focused widget and tells the system the exact canvas coordinates
+		// of the text cursor on the window. The system will display an IME Candidate Window with those coordinates
 		preeditable.ReceiveCursorPositionChangedCallback(func(pos fyne.Position) {
-			monitor := w.getMonitorForWindow()
-			xscale, yscale := monitor.GetContentScale()
-			w.viewport.SetPreeditCursorRectangle(int(pos.X*xscale), int(pos.Y*yscale), 100, 100)
+			canvasScale := w.canvas.scale
+			w.viewport.SetPreeditCursorRectangle(int(pos.X*canvasScale), int(pos.Y*canvasScale), 100, 100)
 		})
 	}
 }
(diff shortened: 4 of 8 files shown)


canvasobject.go
```diff
@@ -75,6 +75,14 @@ type Focusable interface {
 	TypedKey(*KeyEvent)
 }
 
+// Preeditable describes any CanvasObject that can respond to IME inputs.
+// Originally, it should be a method of Focusable, but only a few widgets need PreeditChanged,
+// and adding a method to Focusable will have a big impact on others, so defined as a new interface.
```
andydotxyz, line 80: I agree with this sentiment - adding it to Focusable would be a breaking change.
However can it be worded less like a workaround. 
i.e. "Widgets that implement `Focusable` may also want to implement `PreeditChanged` to handle IME input?

internal/driver/common/canvas.go
```diff
@@ -191,6 +191,15 @@ func (c *Canvas) Focused() fyne.Focusable {
 	return mgr.Focused()
 }
 
+// Preeditable return the current preeditable object only when focused
```
andydotxyz, line 194: This will crash if the input is not Preeditable. It should not be needed as it will always be the same object returned by `Focused()`

internal/driver/glfw/window_desktop.go
```diff
@@ -677,6 +678,32 @@ func (w *window) charInput(viewport *glfw.Window, char rune) {
 	w.processCharInput(char)
 }
 
+var counterIme int
+
+// IME
```
andydotxyz, line 683: A comment should not be needed here. Where comments are added it usually implies naming is insufficient.

internal/driver/glfw/window_desktop.go
```diff
@@ -677,6 +678,32 @@ func (w *window) charInput(viewport *glfw.Window, char rune) {
...
+func (w *window) preedit(
+	_ *glfw.Window,
+	preeditCount int,
+	preeditString string,
+	blockCount int,
+	blockSizes string,
+	focusedBlock int,
+	caret int,
+) {
+	fmt.Println(
```
andydotxyz, line 697: Please don't leave in debug

internal/driver/glfw/window_desktop.go
```diff
@@ -677,6 +678,32 @@ func (w *window) charInput(viewport *glfw.Window, char rune) {
...
+		"preedit preeditCount", preeditCount,
+		"preeditString", preeditString,
+		"blockCount", blockCount,
+		"focusedBlock", focusedBlock,
+		"caret", caret,
+	)
+	w.processPreedit(preeditString)
```
andydotxyz, line 704: It looks like it would be easier to just inline processPreedit here as it is barely a single line and only called from this location

widget/entry.go
```diff
@@ -102,6 +105,7 @@ type Entry struct {
 
 // NewEntry creates a new single line entry widget.
 func NewEntry() *Entry {
+	//	e := &Entry{Wrapping: fyne.TextTruncate, PreeditText: "(日本語)"}
```
andydotxyz, line 108: Please don't leave dead code in comments

widget/entry.go
```diff
@@ -39,8 +40,9 @@ var _ fyne.Tabbable = (*Entry)(nil)
 // Entry widget allows simple text to be input when focused.
 type Entry struct {
 	DisableableWidget
-	shortcut fyne.ShortcutHandler
-	Text     string
+	shortcut    fyne.ShortcutHandler
+	Text        string
+	PreeditText string
```
andydotxyz, line 45: Why is this a public field? it feels like it should be an internal implementation if required at all?

widget/entry.go
```diff
@@ -1745,10 +1800,12 @@ func (r *entryContentRenderer) buildSelection() {
 	}
 
 	provider := r.content.entry.textProvider()
+	preedit := r.content.entry.preeditProvider()
 	// Convert column, row into x,y
 	getCoordinates := func(column int, row int) (float32, float32) {
 		sz := provider.lineSizeToColumn(column, row)
-		return sz.Width, sz.Height*float32(row) - theme.InputBorderSize() + theme.InnerPadding()
+		szpe := preedit.lineSizeToColumn(column, row)
+		return float32(math.Max(float64(sz.Width), float64(r.cursor.Position().X+szpe.Width))), sz.Height*float32(row) - theme.InputBorderSize() + theme.InnerPadding()
```
andydotxyz, line 1808: Are you sure this is correct? Isn't the pre-edit only on the current cursor line? This seems to add the "szpe" irrespective of the row being measured...

andydotxyz (changes requested):
Thanks for working on this really complex task.
I have some suggestions inline.

internal/driver/common/canvas.go
kanryu: @andydotxyz It may be counter-intuitive to you, but this code works fine. Unlike C casts, Go interface casts return nil if the interface is not satisfied.

canvasobject.go
kanryu: That comment certainly makes the relationship with Focusable clearer. People who see this code for the first time may not understand the deep relationship between Focusable and Preeditable.

internal/driver/glfw/window_desktop.go
kanryu: @andydotxyz  Unfortunately processPreedit and Preeditable.PreeditChanged are temporary interfaces. This is because the original Preedit response requires a more complex response. It is necessary to additionally display the conversion candidates for input content provided by the IME and the IME's own caret.

internal/driver/common/canvas.go
andydotxyz: It would only happen if you started to interact with an input that had not implemented that interface.
Something like the checkbox I would think for example.

internal/driver/glfw/window_desktop.go
andydotxyz: OK. Should we mark this as a draft PR then instead of one ready for review?

kanryu:
スクリーンショット 2024-02-14 11 36 16
Full input is now available on Mac.

widget/entry.go
kanryu: @andydotxyz  OK, Removed.

widget/entry.go
kanryu: @andydotxyz OK, privated

widget/entry.go
kanryu: @andydotxyz OK, removed

internal/driver/glfw/window_desktop.go
kanryu: @andydotxyz ok, removed.

qizhanchan:
image
I cloned the latest branch and run the demo, the entry's 'select all' focus effect is confusing.

Jacalz:
@kanryu If you are experiencing a bug on develop, report it as an issue instead of as a comment on a PR please. Thanks.

kanryu:
> I cloned the latest branch and run the demo, the entry's 'select all' focus effect is confusing.
@qizhanchan This issue is not caused by this PR's fix, but also occurs with the original Fyne Entry. So it's a Fyne problem.

kanryu:
> If you are experiencing a bug on develop, report it as an issue instead of as a comment on a PR please. Thanks.

@Jacalz Which problem in this PR are you pointing out? This PR involves a very large source code modification.

Jacalz:
Ah, sorry @kanryu. I meant to be tagging @qizhanchan in my message. I assumed that he was referring to a bug on the develop branch.

internal/driver/glfw/window_desktop.go
```diff
@@ -765,11 +807,15 @@ func (w *window) create() {
 		win.SetRefreshCallback(w.refresh)
 		win.SetContentScaleCallback(w.scaled)
 		win.SetCursorPosCallback(w.mouseMoved)
+		win.SetPreeditCallback(w.preedit)
```
andydotxyz, line 810: This won't compile for me. I guess the `go.mod` is missing some alterations?

internal/driver/glfw/loop_desktop.go
```diff
@@ -11,8 +11,11 @@ import (
 	"github.com/go-gl/glfw/v3.3/glfw"
 )
 
+const GLFW_X11_ONTHESPOT glfw.Hint = 0x00052002
```
andydotxyz, line 14: Won't this be declared in the GLFW project?
I guess it relates to my other comment about not being sure of how to compile or which GLFW version is needed

andydotxyz (commented):
Sorry it has taken so long to get back on this.

I can't figure out how to test this - is there info about the versions of GLFW etc missing form go.mod?

internal/driver/glfw/loop_desktop.go
kanryu: @Jacalz It's difficult to explain in a few words because the work spanned multiple projects, but achieving this PR required major modifications to go-gl/glfw, and they effectively refused to merge. Because of this, we are currently unable to proceed with this PR any further.
Other than that, I think this PR is of a quality worthy of being tested by many people.

It's up to the people at the Fyne project to decide whether or not this PR should be marked as draft. Not that I would complain.

Jacalz: @kanryu I am entirely aware of the situation. My personal opinion is that if a PR can't be tested as it is, i.e. isn't ready to be merged, it should be a draft. Those that are interested in testing it can do so but we maintainers don't have to temper that it isn't ready.

andydotxyz: > @andydotxyz Thanks for trying. As I wrote at the beginning of this PR, in order to check the operation of this PR, you need to clone both the modified fyne and glfw and change go.mod.

This can be done with a replace directive in the go.mod so that this PR can simply be checked out and tested without pre-requisites. That also makes it clear of the requirements to compile.

andydotxyz:
Converting to draft as this is not in a state where we would be reviewing for merge.

andydotxyz:
We care very much about adding IME support, but this PR cannot be incorporated as it depends on work in progress which is not accepted to the upstream projects.

As it has been open for a long time there are many conflicts meaning it will be hard to get up to date.

Once the dependant code has been landed upstream it would be excellent to re-visit this and get support added.

PR #4615, opened 2024-02-06
Hide image in dialog background when not needed, saved CPU

### LINKED ISSUE #4574: High CPU use when showing CustomDialogs
### Checklist

### Describe the bug

Showing a custom confirm dialog using `dialog.ShowCustomConfirm` uses high CPU while the dialog is displayed. After dismissing the dialog the CPU settles back down to normal.

### How to reproduce

Run the example code, with an Activity Monitor/Task Manager/etc window open, and click the button to show the custom confirm dialog

### Screenshots

_No response_

### Example code

```go
package main

import (
	"log"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/dialog"
	"fyne.io/fyne/v2/widget"
)

func main() {
	myApp := app.New()
	myWindow := myApp.NewWindow("Custom Confirm CPU bug")

	button := widget.NewButton("Show dialog", func() {
		dialog.ShowCustomConfirm("Custom confirm", "Confirm text", "Dismiss text",
			widget.NewLabel("Content"), func(b bool) {
				log.Printf("Confirm dismissed with value %v\n", b)
			}, myWindow)
	})
	myWindow.SetContent(container.NewCenter(button))
	myWindow.Resize(fyne.NewSize(400, 300))
	myWindow.ShowAndRun()
}
```

### Fyne version

2.4.3

### Go compiler version

go1.21.1

macOS Ventura M2

### Additional Information

_No response_

andydotxyz:
Fixes #4574

### Checklist:

- [x] Tests included. <- existing graphical tests verify the look is unchanged
- [x] Tests all pass.

### DIFF
--- a/dialog/base.go
+++ b/dialog/base.go
@@ -108,8 +108,13 @@ func (d *dialog) hideWithResponse(resp bool) {
 func (d *dialog) create(buttons fyne.CanvasObject) {
 	label := widget.NewLabelWithStyle(d.title, fyne.TextAlignLeading, fyne.TextStyle{Bold: true})
 
+	image := &canvas.Image{Resource: d.icon}
+	if d.icon == nil {
+		image.Hide()
+	}
+
 	content := container.New(&dialogLayout{d: d},
-		&canvas.Image{Resource: d.icon},
+		image,
 		newThemedBackground(),
 		d.content,
 		buttons,
--- a/dialog/base.go
+++ b/dialog/base.go
@@ -8,6 +8,7 @@ import (
 	"fyne.io/fyne/v2/canvas"
 	"fyne.io/fyne/v2/container"
 	col "fyne.io/fyne/v2/internal/color"
+	"fyne.io/fyne/v2/layout"
 	"fyne.io/fyne/v2/theme"
 	"fyne.io/fyne/v2/widget"
 )
@@ -108,9 +109,11 @@ func (d *dialog) hideWithResponse(resp bool) {
 func (d *dialog) create(buttons fyne.CanvasObject) {
 	label := widget.NewLabelWithStyle(d.title, fyne.TextAlignLeading, fyne.TextStyle{Bold: true})
 
-	image := &canvas.Image{Resource: d.icon}
-	if d.icon == nil {
-		image.Hide()
+	var image fyne.CanvasObject
+	if d.icon != nil {
+		image = &canvas.Image{Resource: d.icon}
+	} else {
+		image = &layout.Spacer{}
 	}
 
 	content := container.New(&dialogLayout{d: d},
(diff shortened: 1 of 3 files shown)


dialog/base.go
```diff
@@ -108,8 +108,13 @@ func (d *dialog) hideWithResponse(resp bool) {
 func (d *dialog) create(buttons fyne.CanvasObject) {
 	label := widget.NewLabelWithStyle(d.title, fyne.TextAlignLeading, fyne.TextStyle{Bold: true})
 
+	image := &canvas.Image{Resource: d.icon}
```
dweymouth, line 111: If the dialog is immutable once created (ie if it doesn't have an image it will never have one), why not avoid adding the image entirely, or if it's easier code-wise, place a `layout.NewSpacer()` in place of the image in the container, as it's a less expensive object. That's what I often use as an "empty" canvas object.

andydotxyz: You're right it was a code complexity issue that led me to leave it there.
My concern is that by placing a spacer we have possibly different objects in that index - is a spacer much less expensive than a hidden image with nil content?

dweymouth: I checked `canvas.Image` implementation and the only reason I even bothered with the suggestion is that Resize on an image results in a Refresh: https://github.com/fyne-io/fyne/blob/master/canvas/image.go#L174

dweymouth: I am fine with merging this as-is though (but I didn't actually test it myself) and we could file an optimization ticket to the backlog if we want.

dweymouth (approved):

dialog/base.go
andydotxyz: > I checked `canvas.Image` implementation and the only reason I even bothered with the suggestion is that Resize on an image results in a Refresh

You talked me into it thank :) spacer is actually OK - all tests and usages passing

dweymouth (approved):
nice one

Jacalz:
Should we not make sure that the renderer doesn't render nil resources and get stuck in a rendering loop?
I agree that this is a good change but is it also not leaving the underlying issue unfixed?

dweymouth:
> Should we not make sure that the renderer doesn't render nil resources and get stuck in a rendering loop? I agree that this is a good change but is it also not leaving the underlying issue unfixed?

There's a separate ticket tracking the underlying issue (which may be a harder fix), but I think this PR is still desirable even if the underlying issue were to be fixed, and this also allows us to fix it fast, which is a high-impact issue, whereas the underlying issue only impacts user code that shows an empty image.

andydotxyz:
Yeah the change I proposed on the linked issue didn't really solve the underlying issue it just reduced it - as the not-drawing-nil-image didn't remove the loop completely. Something about image handling needs more time spent on it in the nil case as it didn't seem to "stick" that rendering an empty image was done when nothing was drawn.

PR #4620, opened 2024-02-07
Fix two possible panics in closing

andydotxyz:
Fix issues recently introduced into closing windows.
One was seen running hello app, the other using fyne_demo

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/app/lifecycle.go
+++ b/internal/app/lifecycle.go
@@ -82,7 +82,7 @@ func (l *Lifecycle) OnStarted() func() {
 func (l *Lifecycle) OnStopped() func() {
 	stopped := l.onStopped.Load()
 	stopHook := l.onStoppedHookExecuted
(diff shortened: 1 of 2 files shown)


internal/app/lifecycle.go
```diff
@@ -90,7 +90,7 @@ func (l *Lifecycle) OnStopped() func() {
 		return *stopped
 	}
 
-	if *stopped == nil {
```
dweymouth, line 93: Is is possible for `stopped` to be non-nil but `*stopped` to be still nil? Should the check be `if stopped == nil || *stopped == nil`?

andydotxyz: Yes, it is possible someone stored nil good catch. Fixed

dweymouth (approved):

PR #4622, opened 2024-02-07
Convert more widgets

andydotxyz:
TODO:

- [x] Accordion
- [x] Activity
- [x] Card
- [x] Menus 
- [x] Toolbar

This just leaves the collections and containers for another PR

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/accordion.go
+++ b/widget/accordion.go
@@ -147,8 +147,10 @@ type accordionRenderer struct {
 }
 
 func (r *accordionRenderer) Layout(size fyne.Size) {
-	pad := theme.Padding()
-	dividerOff := (pad + theme.SeparatorThicknessSize()) / 2
+	th := r.container.Theme()
+	pad := th.Size(theme.SizeNamePadding)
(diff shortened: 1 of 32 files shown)


dweymouth (commented):
Is this work going to be obsolete when we move to the approach of theming the icons in the renderer so that custom widgets can have theme overrides "just work"?

andydotxyz:
> Is this work going to be obsolete when we move to the approach of theming the icons in the renderer so that custom widgets can have theme overrides "just work"?

No. Colouring icons will just work, but widgets must all look up the current theme for their widget - the globalm APIs we used before have no way of knowing what context the incoming request for icon/colour etc is referring to.

andydotxyz:
This qualifies for auto-acceptance as its been a week with no change requests, but it seems a bit large to do so.
Any thoughts @Jacalz @dweymouth ?

dweymouth (approved):
Seems fine to me

PR #4630, opened 2024-02-11
Dialog wrapping

dweymouth:
### Description:

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


andydotxyz:
This is fantastic work thanks so much.

One question as I was testing - if the window is large and is made smaller it will resize to 100% of the window width instead of the 90% (which looks good btw).
Could it be fixed so they are consistent?

dweymouth:
> One question as I was testing - if the window is large and is made smaller it will resize to 100% of the window width instead of the 90% (which looks good btw).
> Could it be fixed so they are consistent?

Hmm, interesting question and I agree that would be nice to have that capability - but I think it would need to be hooked in at a higher level (window canvas)? as it's up to the window to control the dialog's size after it's shown. Maybe you could advise on where that hook-in might go? It would probably need type assertion to check if an overlay is a `*dialog` to handle it specially

andydotxyz:
Sorry I forgot to reply. The suggestion would need updates at the (canvas/window) level - we already help Menus (and popups) to be within bounds, in this place I think the same could be done for dialogs to be within a 90% inset.

Having said that I realise that some apps don't want this (for example I have seen some folk open a window only to show a (full size) dialog. So maybe it was a bad suggestion?

dweymouth:
> Sorry I forgot to reply. The suggestion would need updates at the (canvas/window) level - we already help Menus (and popups) to be within bounds, in this place I think the same could be done for dialogs to be within a 90% inset.
> 
> Having said that I realise that some apps don't want this (for example I have seen some folk open a window only to show a (full size) dialog. So maybe it was a bad suggestion?

If not all apps want it, I wonder if maybe we shouldn't add it. I imagine users resizing a window while a modal dialog is shown isn't super common anyway. Or the logic could only apply the 90% width to a dialog that was previously shown at <= 90% window width (ie was not initially shown at the full parent window size), though that could get complicated

dialog/text.go
```diff
@@ -0,0 +1,47 @@
...
+
+	return d
+}
+
+// returns a beforeShowHook that sets the desired width of the dialog to the min of:
+// - width needed to show message without wrapping
+// - maxTextDialogAbsoluteWidth
+// - current window width * maxTextDialogWinPcntWidth
+func createBeforeShowHook(d *dialog, message string) func() {
+	noWrapWidth := widget.NewLabel(message).MinSize().Width + padWidth + theme.Padding()*2
```
Jacalz, line 35: It is better to use `fyne.MeasureText` rather than allocating a temp label.

dialog/information.go
```diff
@@ -7,22 +7,20 @@ import (
 	"fyne.io/fyne/v2/widget"
 )
 
-func createTextDialog(title, message string, icon fyne.Resource, parent fyne.Window) Dialog {
-	d := newDialog(title, message, icon, nil, parent)
-
+func createInformationDialog(title, message string, icon fyne.Resource, parent fyne.Window) Dialog {
+	d := newTextDialog(title, message, theme.ErrorIcon(), parent)
```
Jacalz, line 11: It seems like this shouldn't ignore `icon` and pass `theme.ErrorIcon()` for everything?

Jacalz (changes requested):
Just a few notes inline on top of the discussion with Andrew. Nice work :)

dialog/text.go
dweymouth: It resulted in a different (slightly) size, and failing tests. But if we want to update all the rendering tests (I don't know how to regenerate the XMLs), I suppose we could do that

dialog/information.go
dweymouth: Good catch!

dialog/text.go
dweymouth: Created issue #4648 and added an explanatory comment to the code, referencing it

dweymouth:
Any further thoughts on whether we want to maintain dialogs to be sized to 90% of the window width as the window is resized down?

andydotxyz:
> Any further thoughts on whether we want to maintain dialogs to be sized to 90% of the window width as the window is resized down?

I'm happy to leave that out for now I guess - could come back to it later if folk do want that behaviour too.

andydotxyz (approved):
Nice one thanks so much for solving this!

PR #4632, opened 2024-02-11
Add Desktop directory to file dialogs

codesoap:
I believe that the "Desktop" directory is heavily used by many users. Thus I think it would be convenient, if it appeared in the favorites of file dialogs.

The changes are basically pure Cargo cult, but seem sensible to me and seem to work. Please correct me, if I've forgotten something.

I have not assigned an icon to this entry in the file dialogs, since the most common symbol seems to be a monitor and I couldn't find such a symbol in the `theme` package.

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -860,6 +860,7 @@ func getFavoriteIcons() map[string]fyne.Resource {
 	if runtime.GOOS == "darwin" {
 		return map[string]fyne.Resource{
 			"Documents": theme.DocumentIcon(),
+			"Desktop":   theme.DesktopIcon(),
 			"Downloads": theme.DownloadIcon(),
 			"Music":     theme.MediaMusicIcon(),
 			"Pictures":  theme.MediaPhotoIcon(),
@@ -869,6 +870,7 @@ func getFavoriteIcons() map[string]fyne.Resource {
 
 	return map[string]fyne.Resource{
 		"Documents": theme.DocumentIcon(),
+		"Desktop":   theme.DesktopIcon(),
 		"Downloads": theme.DownloadIcon(),
 		"Music":     theme.MediaMusicIcon(),
 		"Pictures":  theme.MediaPhotoIcon(),
@@ -878,6 +880,7 @@ func getFavoriteIcons() map[string]fyne.Resource {
 
 func getFavoriteOrder() []string {
 	order := []string{
+		"Desktop",
 		"Documents",
 		"Downloads",
 		"Music",
--- a/dialog/file_xdg.go
+++ b/dialog/file_xdg.go
@@ -45,6 +45,7 @@ func getFavoriteLocations() (map[string]fyne.ListableURI, error) {
 
 	favoriteNames := getFavoriteOrder()
 	arguments := map[string]string{
+		"Desktop":   "DESKTOP",
 		"Documents": "DOCUMENTS",
 		"Downloads": "DOWNLOAD",
 		"Music":     "MUSIC",
--- a/test/markup_renderer.go
+++ b/test/markup_renderer.go
@@ -437,6 +437,7 @@ func knownResource(rsc fyne.Resource) string {
 		theme.ContentRemoveIcon():      "contentRemoveIcon",
 		theme.ContentUndoIcon():        "contentUndoIcon",
 		theme.DeleteIcon():             "deleteIcon",
+		theme.DesktopIcon():            "desktopIcon",
 		theme.DocumentCreateIcon():     "documentCreateIcon",
 		theme.DocumentIcon():           "documentIcon",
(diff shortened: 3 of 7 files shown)


andydotxyz:
Thanks for this, but can you add it for Windows and macOS as well? We keep feature parity between all the systems (and file dialog is only shown on desktop)

codesoap:
I'm afraid I don't see what is missing. I had grepped for the term "Videos" as a guidance for my change and found no other relevant places like this.

The `getFavoriteLocations()` functions in `dialog/file_darwin.go` and `dialog/file_windows.go` both seem to use the favorites from the `getFavoriteOrder()` function, which I adapted.

I had tried my change on Linux and on Windows and the "Desktop" favorite was available on both.

andydotxyz:
I see that the locations are looked up, good point.
But I think the `getFavouriteIcons` is missing an understanding of Desktop - so this new item does not match the others.

codesoap:
You're right, I didn't assign an icon. The most common symbol for the desktop seems to be a monitor and I couldn't find that in the `theme` package. What is the preferred way; should I try to find a different icon, that already exists in the `theme` package, or should a new icon be added?

andydotxyz:
> should I try to find a different icon, that already exists in the `theme` package, or should a new icon be added?

If you find one that works use it, if not you can add a new one :).
There is a good repository of material icons that will fit right in (see how we embed the current ones to add a new).

codesoap:
I couldn't find a suitable icon, so I added one (well, actually two). After some research it seemed to me, that (at least some of) the existing symbols where taken from Polymer ([source I found during research](https://github.com/googlearchive/core-icons/blob/337080bebd90c2f0706b6cc25b60ceac10a8722d/hardware-icons.html)). Since Polymer allows use under BSD style licenses and fyne uses a BSD style license, I saw no problem in copying the icons.

Polymer had actually defined two separate icons for the desktop: desktop-mac and desktop-windows. I kept this decision and added the two icons to fyne.

`go run ./theme/gen.go` did generate a change for `fynelogo`, but I didn't commit it here, since it has nothing to do with this issue.

**I have speculated, that this change will be released with fyne 2.5 and added according "Since" comments to `IconNameDesktopMac` and `IconNameDesktopWindows`.**

andydotxyz:
Nice improvement thanks. But is having two different icons for desktops needed? I'm in 2 minds and curious about @Jacalz, @dweymouth or @Bluebugs thoughts

Jacalz:
I think the can just have the Windows one and it just "Desktop"

andydotxyz:
Please don't force push a PR that is going through reviews - in general we prefer to keep history.

Jacalz (changes requested):
All of the tests are failing. Would you mind having a look at that?
I also think you need to update the `test` package lookup functions for the renderer markup output.

codesoap:
I took a closer look and found that 3 fails are due to timeouts and one is unhappy about the use of `theme.DarkTheme`. I feel like these are not problems I created with this PR.

Am I missing something? I'm afraid I'm not very experienced with interpreting the errors...

Jacalz:
The test failures may not be your fault but I haven't seen them all fail like this. It is strange. Will try to look at it when I can.

andydotxyz:
It looks like a deadlock in theme lookup code - my guess would be that the test themes are not bringing along the required features that have been added.
This should happen locally if you just run "go test ./...".

codesoap:
I'm afraid I cannot reproduce the problem locally. I only get a test failure in the `TestWindow_CaptureTypedShortcut` Test case, but that one also fails on `master` for me. The tests in `fyne.io/fyne/v2/dialog` and `fyne.io/fyne/v2/theme` do not fail on my system.

andydotxyz:
I cannot replicate but also cannot seem to kick the CI to try again (not sure why but these workflows seem frozen?).
Is it possible to try at your end, or push an empty commit that will trigger it @codesoap ?

Jacalz:
Rebase on latest develop might fix things maybe?

codesoap:
I have performed the suggested rebase and the tests don't fail anymore. Thanks!

Jacalz (changes requested):
Thanks. Looks better now. You just need to teach the test package about this new icon. See https://github.com/fyne-io/fyne/blob/d8e9fe02c9bb9e4b4c42778f5cae8ffcc43f36e2/test/markup_renderer.go#L413.

codesoap:
I have added the requested entry in the `test` package, although I must admit that I don't understand what for.

andydotxyz:
> although I must admit that I don't understand what for.

The test package runs the apps when in unit test mode. Without it code that used that icon would not function correctly.

Jacalz (approved):
Sorry for the late review. Thanks for the contribution.

codesoap:
Thanks for your thorough reviews!

PR #4633, opened 2024-02-12
Implement persistent storage for preferences on web

### LINKED ISSUE #2734: Add persistent settings and preferences capability to Web driver
The web driver in work does not currently have the ability to remember settings between run. It would be nice to have this feature working also in the web.

This will enable the settings API to be supported on the web, no new API should be necessary.

I think that the web local storage API should work for this use:

https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API

dweymouth:
Adds a persistent backend for the Preferences API for web, using browser local storage. The entire preferences map is saved as a single localStorage object, serialized in JSON under the key "preferences.json" (open to discussion if this key name should be different - user code can inadvertently access this using the `syscall/js` APIs themselves if they use the same key name).

This was done for maximum code sharing between platforms (only the "file-like storage object" is abstracted, the serialization/deserialization code is the same for all platforms), and also so that we don't need to worry about cleaning out individual user key-value pairs from local storage if we later add a Delete API to preferences - the whole preferences blob will just be updated.

Fixes #2734 

### Checklist:

- [ ] Tests included - test manually with wasm fyne_demo - note that it saves the active tutorial page across browser reloads
- [x] Tests all pass.

### DIFF
--- a/app/preferences_wasm.go
+++ b/app/preferences_wasm.go
@@ -1,0 +1,16 @@
+//go:build wasm
+
+package app
+
+import (
+	"errors"
+	"io"
+)
+
+func (p *preferences) storageReader() (io.ReadCloser, error) {
+	return nil, errors.New("not implemented")
+}
+
+func (p *preferences) storageWriter() (writeSyncCloser, error) {
+	return nil, errors.New("not implemented")
+}
--- a/app/preferences_wasm.go
+++ b/app/preferences_wasm.go
@@ -7,10 +7,18 @@ import (
 	"io"
 )
 
+func (a *fyneApp) storageRoot() string {
+	return "" // no storage root for web driver yet
+}
+
 func (p *preferences) storageReader() (io.ReadCloser, error) {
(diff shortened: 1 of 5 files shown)


app/preferences_wasm.go
```diff
@@ -0,0 +1,62 @@
...
+package app
+
+import (
+	"bytes"
+	"io"
+	"strings"
+	"syscall/js"
+)
+
+const preferencesLocalStorageKey = "preferences.json"
```
andydotxyz, line 12: possibly "fyne/preferences.json" or "Fyne-preferences.json" - as you say to reduce the possibility of overlap.

andydotxyz (commented):
This is fantastic, thanks so much.
I just tested with my todo list app in a browser and it works perfectly :).

Just a note about the key name inline

app/preferences_wasm.go
dweymouth: Done

PR #4637, opened 2024-02-13
xml permisions

MatejMagat305:
### Description:
this is sugestion how add permision without putting own xml where can be many obstacle .....

#### Where applicable:
android future - camera, bluetooth, wokeup ...


MatejMagat305:
It does not seems my error...

andydotxyz (changes requested):
Thanks for working on this - but adding platform specific imports to get permissions fixed is not a way forward that makes sense.
The general concept of "enable bluetooth" etc does sound useful, but there must be a cleaner way to enable it.
In the future where we have APIs for bluetooth (for example) that will be platform agnostic and it will be used to turn on the right permissions for each platform.

In the meantime turning it on using build parameters that are interpreted for the current platform seems more extensible and will not require code changes at the Fyne level.

Remember that an public API you add will have to be supported for many years.

MatejMagat305:
> Thanks for working on this - but adding platform specific imports to get permissions fixed is not a way forward that makes sense. The general concept of "enable bluetooth" etc does sound useful, but there must be a cleaner way to enable it. In the future where we have APIs for bluetooth (for example) that will be platform agnostic and it will be used to turn on the right permissions for each platform.
> 
> In the meantime turning it on using build parameters that are interpreted for the current platform seems more extensible and will not require code changes at the Fyne level.
> 
> Remember that an public API you add will have to be supported for many years.

Well, good if this is not clear, could I close and sugest command "fyne package ...[flags] -print xml" ?

andydotxyz:
I don't think printing XML is the solution either - fyne package creates the manifest as you have clearly worked on ion this PR.
It could be configured based on flags (or metadata in FyneApp.toml) to output the appropriate manifest.
The generic concept of functionality or permissions would feed into requesting the required configuration for each platform.

Jacalz:
FYI: The `cmd/fyne` package is deprecated on develop. Any changes there should land in https://github.com/fyne-io/tools/tree/main/cmd/fyne.

MatejMagat305:
So thank for advise and atantion

PR #4647, opened 2024-02-18
Add ScrollToOffset/GetScrollOffset APIs for List

### LINKED ISSUE #3544: List scroll API enhancements
### Checklist

I would like to be able to get the current scroll position of a list, and programmatically scroll a list to a specified position (not list item ID). In my case this would be useful because I would like to be able to save just the data about a list (the backing model plus the current scroll position) so I can rebuild the list widget later exactly as it appeared before saving it.

Not really, `ScrollTo` exists to scroll to a specific item, but there is no API to get the current scroll position. `ScrollTo` is also limited since it can't scroll to an arbitrary position (i.e. in between two items)

Related to #2105, I'd like to see further extensions to the list scrolling API to enable getting the current scroll position, and scrolling to a specified position. Something like

```
func (l *List) GetScrollPosition() fyne.Position
func (l *List) ScrollToPosition(pos fyne.Position)
```

And maybe also something like:

```
func (l *List) GetFirstVisibleItemID() list.ListItemID
```

dweymouth:
### Description:
These APIs already exist on GridWrap and are useful for two things: 
- saving the state of a list view and re-creating it with a new widget that is scrolled to the same position (eg in apps that have "pages" and can navigate back/forward
- implementing keyboard scrolling when nothing is focused (ie key events hit the window canvas) for apps that have just one list view as their main UI element

Fixes #3544 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -261,10 +261,17 @@ func (l *List) ScrollToTop() {
 //
 // Since: 2.5
 func (l *List) ScrollToOffset(offset float32) {
-	if l.scroller != nil {
-		l.scroller.Offset.Y = offset
-		l.offsetUpdated(l.scroller.Offset)
+	if l.scroller == nil {
+		return
 	}
+	if offset < 0 {
+		offset = 0
+	} else if h := l.contentMinSize().Height; offset > h {
+		offset = h
+	}
+	l.scroller.Offset.Y = offset
+	l.offsetUpdated(l.scroller.Offset)
+	l.Refresh()
 }
 
 // GetScrollOffset returns the current scroll offset position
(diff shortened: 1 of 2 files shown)


Jacalz:
Nice. Can you add some tests please? :)

dweymouth:
need to address #4655 for list as well in this PR before merging

widget/list.go
```diff
@@ -257,6 +257,30 @@ func (l *List) ScrollToTop() {
...
+// ScrollToOffset scrolls the list to the given offset position.
+//
+// Since: 2.5
+func (l *List) ScrollToOffset(offset float32) {
+	if l.scroller == nil {
+		return
+	}
+	if offset < 0 {
+		offset = 0
+	} else if h := l.contentMinSize().Height; offset > h {
```
dweymouth, line 269: contentMinSize() is O(n) in the number of items with set heights. If we want to assume that using setItemHeight to make a height shorter than itemMin.Height is not allowed, we can have an O(1) check to skip calling contentMinSize if the offset is within the lower bound on content size of `numItems*itemMin.Height + separatorHeight*numSeparators`

widget/list.go
```diff
@@ -321,6 +345,36 @@ func (l *List) UnselectAll() {
...
 
+func (l *List) contentMinSize() fyne.Size {
+	l.propertyLock.Lock()
+	defer l.propertyLock.Unlock()
+	items := 0
+	if f := l.Length; f == nil {
+		return fyne.NewSize(0, 0)
+	} else {
+		items = f()
+	}
```
Jacalz, line 356:
```suggestion
	if l.Length == nil  {
		return fyne.NewSize(0, 0)
	}
	items := l.Length()
```

Jacalz (changes requested):
Nice. I just left a minor code simplification comment inline

andydotxyz:
Don't we normally omit the `Get` prefix?

andydotxyz:
> > Don't we normally omit the `Get` prefix?
> 
> Usually, but GridWrap already had a `GetSrollOffset` API so I thought in this case, consistency was better.

A bummer, yes that did slip into 2.4.
Quite right consistency trumps this.

Jacalz:
Would it not be better to do it correctly here and then deprecate the one with incorrect naming for List?

dweymouth:
> Would it not be better to do it correctly here and then deprecate the one with incorrect naming for List (and add one with the fright naming of course)?

If we can find a good name, maybe. I do think that having both `ScrollOffset() float64` and `ScrollToOffset(float64)` is confusing naming, even though it matches the "no Get prefix" convention

Jacalz (approved):
Good point. I'm happy in that case :)

PR #4650, opened 2024-02-18
Fix recognition of missing XDG user directories

codesoap:
### Description:
In `dialog/file_xdg.go` there was some code to check, if a user directory does not exist. This code did not work for me, because `locURI.String()` was, for example, `file:///home/richard/` when the "Videos" directory wasn't found. `homeUri.Strin()` was `file:///home/richard`.

I am not sure if `xdg-user-dir` always outputs the trailing slash, so I took a conservative approach and used `strings.TrimRight`.

I've added a second commit, which prevents the errors from `getFavoriteLocation` to be overwritten by errors from `storage.ListerForURI`.

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/file_xdg.go
+++ b/dialog/file_xdg.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/storage"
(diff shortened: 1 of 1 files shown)


dialog/file_xdg.go
```diff
@@ -27,7 +28,7 @@ func getFavoriteLocation(homeURI fyne.URI, name, fallbackName string) (fyne.URI,
 	loc = loc[:len(loc)-1]
 	locURI := storage.NewFileURI(string(loc))
 
-	if locURI.String() == homeURI.String() {
+	if strings.TrimRight(locURI.String(), "/") == homeURI.String() {
```
andydotxyz, line 31: Can you be certain the home directory does not include a slash?
It may be better to ensure both are trimmed to cover all cases

codesoap: I have looked into it a bit, and think it could be possible. `homeURI` is constructed from `os.UserHomeDir()`, which will return the value of the environment variable `$HOME` (if set). Usually this shouldn't contain a trailing slash, but if a user manually set their `$HOME` variable and accidentally included the trailing slash, it would be problematic.

Thus I have added another `strings.TrimRight`. It can't hurt.

andydotxyz (approved):
Thanks

PR #4652, opened 2024-02-18
Fix SetOnClosed for FileDialog

### LINKED ISSUE #4651: FileDialog.SetOnClosed not always working
### Checklist

### Describe the bug

If `SetOnClosed` is called on a `FileDialog` before `Show` is called, the callback will not be called when the dialog is closed.

### How to reproduce

Run the example code, close the file dialog and see that `foo` is not printed to standard output.

### Screenshots

_No response_

### Example code

```go
package main

import "fyne.io/fyne/v2"
import "fyne.io/fyne/v2/app"
import "fyne.io/fyne/v2/dialog"

func main() {
	a := app.New()
	w := a.NewWindow("test")

	fd := dialog.NewFileOpen(func(rc fyne.URIReadCloser, err error) {}, w)
	fd.SetOnClosed(func() { println("foo") })
	fd.Show()

	w.ShowAndRun()
}
```

### Fyne version

2.4.4

### Go compiler version

1.22.0

MX Linux, I believe version 21; it's basically Debian

### Additional Information

It works, if `SetOnClosed` is called after `Show`:

```go
package main

import "fyne.io/fyne/v2"
import "fyne.io/fyne/v2/app"
import "fyne.io/fyne/v2/dialog"

func main() {
	a := app.New()
	w := a.NewWindow("test")

	fd := dialog.NewFileOpen(func(rc fyne.URIReadCloser, err error) {}, w)
	fd.Show()
	fd.SetOnClosed(func() { println("foo") })

	w.ShowAndRun()
}
```

The problem seems to be that `f.dialog == nil` [is checked in `SetOnClosed`](https://github.com/fyne-io/fyne/blob/e332a5e47813bff1af9631cbbfe84654c0143189/dialog/file.go#L706), but `f.dialog` [is only set once `Show` is called](https://github.com/fyne-io/fyne/blob/e332a5e47813bff1af9631cbbfe84654c0143189/dialog/file.go#L634).

My initial impulse was to just remove the early return in `SetOnClosed`, but I suppose it's there for a reason (which I don't see right now), so I didn''t open a PR.

codesoap:
### Description:
`SetOnClose` did not work if called before `Show` for file dialogs.

Fixes #4651 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -736,13 +736,13 @@ func (f *FileDialog) SetLocation(u fyne.ListableURI) {
 // SetOnClosed sets a callback function that is called when
 // the dialog is closed.
 func (f *FileDialog) SetOnClosed(closed func()) {
-	if f.dialog == nil {
-		return
-	}
 	// If there is already a callback set, remember it and call both.
 	originalCallback := f.onClosedCallback
 
 	f.onClosedCallback = func(response bool) {
+		if f.dialog == nil {
+			return
+		}
 		closed()
 		if originalCallback != nil {
 			originalCallback(response)
--- a/dialog/file_test.go
+++ b/dialog/file_test.go
@@ -671,3 +671,13 @@ func TestCreateNewFolderInDir(t *testing.T) {
 	folderNameInputCreate := folderNameInputUI.Objects[3].(*fyne.Container).Objects[1].(*widget.Button)
 	assert.Equal(t, theme.ConfirmIcon(), folderNameInputCreate.Icon)
 }
+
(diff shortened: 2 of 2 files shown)


andydotxyz (changes requested):
This is a good catch, thanks.
Would it be possible to add a test so we don't regress?

codesoap (commented):
Sure thing, I've added a test.

andydotxyz (approved):
Excellent, thanks so much

Issue #4654, opened 2024-02-18
New threading model


andydotxyz:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [X] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

To work on race conditions and many locking challenges we agreed to revise the threading model (the "App thread").
This issue tracks the work required.

### Describe the solution you'd like to see.

First:
- [x] Back out the BaseWidget.SetFieldsAndRefresh API - it will be obsoleted by "fyne.Go"
- [x] Re-merge draw and event into one thread (#4656)
- [x] Move animation ticks to the main thread (#4656)
- [x] Check that all widget callbacks happen on this same thread (#5384)
- [x] Combine data binding into same thread (and remove test waits) (#5361)
- [x] Add ability to schedule code on the App thread (new "fyne.Go" API #6599)
- [x] Enable race on all our CI test runs (#5384)
- [ ] Add new tests that show that things work as expected from goroutines
- [x] Explicitly document the specific APIs that are still allowed to be called from goroutines without using the above new API. (Possibly only data binding, and various "pure" funcs that don't mutate state?)
- [x] Test on all devices. Now validated on macOS, Linux, Windows, iOS and Android

Then:
- [x] Clean up now-unneeded locks/synchronization in widgets, etc (#5388)
- [x] Revise unit testing to do this all sync and remove waits etc no longer needed (#5384)
- [x] Make the transition less painful for users upgrading to v2.6.0 (#5411)

Also to consider:
~- [ ] Fix issues with re-draw on window resize that led to draw/event separation (moved to next release for new event loop)~
- [ ] Consider adding a `noCopy` struct that is a dummy implementation of `sync.Locker` to BaseWidget so `go vet` can still complain about passing widgets by value (#5520)

andydotxyz:
> > Re-merge draw and event into one thread
> 
> This might be the easiest path, but it isn't necessarily the only one. We can keep them on separate threads as long as we have them lock when reading or writing state.

Yes, there could be locking based solutions - but part of this is trying to find the simplest solution, so reverting the thread split seems like a good start.

andydotxyz:
Work is begun on this - I updated the TODO above to highlight what is needed in the order.
I have a proof of concept for most of the top list working locally and it is indeed race-free but with a significant lag on heavy event usage... more work to do before I open more PRs

[...]

PR #4656, opened 2024-02-18
Revert using a separate draw thread

andydotxyz:
Progresses #4654 

### Checklist:

- [ ] Tests included. <- currently manual testing to check all platforms
  - [x] macOS arm
  - [x] macOS Intel
  - [x] Windows
  - [x] Linux X11
  - [x] Linux Wayland
- [x] Tests all pass.

Notice that on macOS the UI will stop drawing if we resize. This is the reason draw thread was added and we will need to find another solution...

### DIFF
--- a/internal/animation/runner.go
+++ b/internal/animation/runner.go
@@ -43,7 +43,6 @@ func (r *Runner) Start(a *fyne.Animation) {
 			r.animations = make([]*anim, 0, 16)
 		}
 		r.animations = append(r.animations, newAnim(a))
-		r.runAnimations()
 	} else {
 		if r.pendingAnimations == nil {
 			// initialize with excess capacity to avoid re-allocations
@@ -85,18 +84,18 @@ func (r *Runner) Stop(a *fyne.Animation) {
 	r.pendingAnimations = newList
 }
 
-func (r *Runner) runAnimations() {
-	draw := time.NewTicker(time.Second / 60)
-	go func() {
-		for done := false; !done; {
-			<-draw.C
-			done = r.runOneFrame()
-		}
+func (r *Runner) TickAnimations() {
+	if !r.runnerStarted {
+		return
+	}
+
+	done := r.runOneFrame()
+
+	if done {
 		r.animationMutex.Lock()
 		r.runnerStarted = false
 		r.animationMutex.Unlock()
-		draw.Stop()
-	}()
+	}
 }
 
 func (r *Runner) runOneFrame() (done bool) {
--- a/internal/animation/runner.go
+++ b/internal/animation/runner.go
@@ -84,6 +84,8 @@ func (r *Runner) Stop(a *fyne.Animation) {
 	r.pendingAnimations = newList
 }
 
+// TickAnimations progresses all running animations by one tick.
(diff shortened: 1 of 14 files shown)


internal/animation/runner.go
```diff
@@ -61,32 +60,34 @@ func (r *Runner) Stop(a *fyne.Animation) {
...
+	oldList := r.animations
+	r.animationMutex.Unlock()
+	newList := make([]*anim, 0, len(oldList))
+	for _, a := range oldList {
+		if !a.isStopped() && r.tickAnimation(a) {
+			newList = append(newList, a)
 		}
+	}
+	r.animationMutex.Lock()
+	r.animations = append(newList, r.pendingAnimations...)
```
dweymouth, line 81: The runner can get a whole lot cleaner once we assume the new threading model. While Start and Stop can still be called when invoking user code within a Tick, we will know they won't be called elsewhere, so we can do some simplification and cleanup around the synchronization between Start, Stop and Tick. That can be a follow-up PR though.

andydotxyz:
Putting this on hold until we begin work on 2.6

andydotxyz:
Reworked on top of `develop`.

This seems to work smoothly in limited testing but we need to get it checked on all OS.
Lots more work is required to simplify but I didn't want to go too deep yet.

andydotxyz:
P.S. in a follow up PR where I merge the event thread in as well I have **hello** and **fyne_demo** completely without race errors!

Keeping them separate as this PR works smoothly but the event management change has significant performance impact!

andydotxyz:
All Mac tests passing. There is a performance impact but it is actually clear where some latency is coming from - hovering events are triggering multiple event updates etc. This change may help with hunting down other performance bugs!

Jacalz:
Nice. That should help us decrease power consumption and unnecessary cpu usage :)

internal/driver/glfw/loop.go
```diff
@@ -55,16 +48,8 @@ func runOnMain(f func()) {
...
-		runOnMain(func() { w.RunWithContext(f) })
-		return
-	}
-	done := common.DonePool.Get()
-	defer common.DonePool.Put(done)
-
-	drawFuncQueue <- drawData{f: f, win: w, done: done}
-	<-done
+func runOnMainWithContext(w *window, f func()) {
+	runOnMain(func() { w.RunWithContext(f) }) // TODO remove this completely
```
Jacalz, line 52: Is this a TODO for an upcoming PR or should it be removed?

Jacalz (approved):
Well done. Performance seems okay and everything appears to be working here on Linux. Just left one minor comment line but I'll approve before hand :)

internal/driver/glfw/loop.go
andydotxyz: Yes it is part of the future refactoring. Lots more to be done as mentioned above and in the other tickets.

There are 3 different ways to queue onto what is essentially the same thread but I didn't want to rip out too much code until the direction is established :)

Jacalz: Cool. Just wanted to check. Still approved from my end :)

PR #4663, opened 2024-02-22
Properly validate offset in GridWrap.ScrollToOffset

### LINKED ISSUE #4655: GridWrap ScrollToOffset API doesn't properly validate offset when viewport is larger than the full content
### Checklist

### Describe the bug

If a gridwrap is showing items in a viewport larger than needed to show everything in one screen, ScrollToOffset doesn't validate the offset correctly and you can scroll "down", causing items to disappear from the viewport, and "up" in a way that can result in a crash

### How to reproduce

It shows up in Supersonic on main (development) branch now that I've added keyboard scrolling. May add example code later...

### Screenshots

_No response_

### Example code

might add later

### Fyne version

2.4.4

### Go compiler version

go1.21.1

macOS Ventura M2

### Additional Information

_No response_

dweymouth:
### Description:
Ensures that ScrollToOffset cannot be used to scroll beyond the content size, even if the viewport (ie size of the widget itself) is larger than the content size. Also fixes a missed Refresh call in ScrollToOffset.

Fixes #4655 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/gridwrap.go
+++ b/widget/gridwrap.go
@@ -328,13 +328,13 @@ func (l *GridWrap) UnselectAll() {
 
 func (l *GridWrap) contentMinSize() fyne.Size {
 	padding := theme.Padding()
-	if lenF := l.Length; lenF != nil {
-		cols := l.ColumnCount()
-		rows := float32(math.Ceil(float64(lenF()) / float64(cols)))
-		return fyne.NewSize(l.itemMin.Width,
-			(l.itemMin.Height+padding)*rows-padding)
+	if l.Length == nil {
+		return fyne.NewSize(0, 0)
 	}
-	return fyne.NewSize(0, 0)
+	
+	cols := l.ColumnCount()
+	rows := float32(math.Ceil(float64(l.Length()) / float64(cols)))
+	return fyne.NewSize(l.itemMin.Width, (l.itemMin.Height+padding)*rows-padding)
 }
 
 // Declare conformity with WidgetRenderer interface.
--- a/widget/gridwrap.go
+++ b/widget/gridwrap.go
@@ -215,8 +215,17 @@ func (l *GridWrap) ScrollToTop() {
 
 // ScrollToOffset scrolls the list to the given offset position
 func (l *GridWrap) ScrollToOffset(offset float32) {
+	if l.scroller == nil {
+		return
+	}
+	if offset < 0 {
+		offset = 0
+	} else if contentH := l.contentMinSize().Height; offset > contentH {
+		offset = contentH
+	}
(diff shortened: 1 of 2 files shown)


widget/gridwrap.go
```diff
@@ -317,6 +326,17 @@ func (l *GridWrap) UnselectAll() {
...
 
+func (l *GridWrap) contentMinSize() fyne.Size {
+	padding := theme.Padding()
+	if lenF := l.Length; lenF != nil {
+		cols := l.ColumnCount()
+		rows := float32(math.Ceil(float64(lenF()) / float64(cols)))
+		return fyne.NewSize(l.itemMin.Width,
+			(l.itemMin.Height+padding)*rows-padding)
+	}
+	return fyne.NewSize(0, 0)
```
Jacalz, line 337: I realise that this just was copied from the other method more or less but lets get the logic ordered up to our usual standard with edge-cases indented and main path inline:
```suggestion
	if l.Length == nil {
		return fyne.NewSize(0, 0)
	}
	
	cols := l.ColumnCount()
	rows := float32(math.Ceil(float64(l.Length()) / float64(cols)))
	return fyne.NewSize(l.itemMin.Width, (l.itemMin.Height+padding)*rows-padding)
```

Jacalz (changes requested):
Sorry. I realised that the code wasn't conforming to our style guides. I know it is more or less fixing up old code but lets get it solved now that you are touching it :)

widget/gridwrap.go
dweymouth: Interestingly, the old way was more thread-safe than the new way (user could unset the Length callback in between checking it for nil), but given that we are moving to the single-threaded model it's not an issue.

Jacalz: Without actually using any locks they are both half broken but yeah the old one was better in that regard. You can always save the callback to a variable if you like.

dweymouth: Nah, it won't be an issue at all with the new threading model :) (b/c you'll be disallowed from changing it while the logic here is running)

Jacalz: But you have labelled this as going into v2.4.5?

dweymouth: As you mentioned it was half-broken anyway, and unsetting the Length callback after having set it before for List is basically a developer error anyway (I can't think of a legitimate use case). I wouldn't worry about it, but if you want I can make the change

Jacalz (approved):

dweymouth:
@Jacalz needs re-approval since there was a trailing whitespace picked up that failed static analysis :)

Jacalz (approved):

PR #4677, opened 2024-02-26
Rework markdown parser to fix #4613

### LINKED ISSUE #4613: RichText swallowing whitespace after Markdown links
### Checklist

### Describe the bug

Spaces and linebreaks following a markdown link in a RichText widget get swallowed.

### How to reproduce

1. Run example code
2. See missing whitespace after link

### Screenshots

Missing space after link:
fyne-richtext-url-squeeze

### Example code

```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

var md = `# headline for a wide window

[https://docs.fyne.io/started/hello](https://docs.fyne.io/started/hello)

foo

bar
`

func main() {
	a := app.New()

	rt := widget.NewRichTextFromMarkdown(md)

	w.SetContent(rt)
	w.ShowAndRun()
}
```

### Fyne version

2.4.3

### Go compiler version

1.21.1

OpenBSD 7.4

### Additional Information

_No response_

codesoap:
### Description:
First of all: Sorry for the huge diff, without any previous conversation. I had an idea for a rework, but before suggesting it, I wanted to see if the idea would even work. Before I knew it, I fell into a rabbit hole and now here I am with this PR.

Inspired by #4613, I read through `markdown.go` and investigated goldmark a little bit (here's a tool to visualize goldmark's ASTs: [goldmarkvis](https://github.com/codesoap/goldmarkvis)). I got an understanding of where spaces should be added, but couldn't find an easy way to incorporate my newfound knowledge into the existing design. I felt like it would be easier, if the AST were rendered with recursive code.

I went ahead and tried it out. With this approach, I was able to fix all the problems mentioned in #4613 (including the "split link"). As a side effect, the code also got a little more compact and slightly more readable (but that's highly subjective, of course).

I have also added a benchmark, which shows a slight decline of performance with the new implementation, but it's not too bad and I would assume that performance would have decreased anyway, no matter how #4613 would have been fixed:
- Before: `BenchmarkMarkdownParsing-4  8527  135023 ns/op  34864 B/op  306 allocs/op`
- After: `BenchmarkMarkdownParsing-4  7798  149466 ns/op  35920 B/op  328 allocs/op`

Since the new implementation adds leading spaces instead of trailing ones, and also adds new segments for some spaces and the end of paragraphs, I had to adapt the existing tests slightly.

Fixes #4613 

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/markdown_test.go
+++ b/widget/markdown_test.go
@@ -225,3 +225,22 @@ func TestRichTextMarkdown_Separator(t *testing.T) {
 		t.Error("Segment should be a separator")
 	}
 }
+
+func BenchmarkMarkdownParsing(b *testing.B) {
+	md := `# Test heading
+This is some test markdown. It contains some different markdown
+elements in an effort to put a realistic load onto the parser.
+
+> Here is a quote.
+> It streches over two lines and contains some ` + "`inline code`" + `.
+
+Moreover, we've got a list:
+- foo
+- bar
+- baz
+`
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		NewRichTextFromMarkdown(md)
+	}
+}
--- a/widget/markdown_test.go
+++ b/widget/markdown_test.go
@@ -165,13 +165,13 @@ func TestRichTextMarkdown_Lines(t *testing.T) {
 
 	assert.Equal(t, 3, len(r.Segments))
 	if text, ok := r.Segments[0].(*TextSegment); ok {
-		assert.Equal(t, "line1", text.Text)
+		assert.Equal(t, "line1 ", text.Text)
 		assert.True(t, text.Inline())
 	} else {
 		t.Error("Segment should be Text")
 	}
 	if text, ok := r.Segments[1].(*TextSegment); ok {
-		assert.Equal(t, " line2", text.Text)
+		assert.Equal(t, "line2", text.Text)
 	} else {
 		t.Error("Segment should be Text")
 	}
--- a/widget/markdown_test.go
+++ b/widget/markdown_test.go
@@ -11,8 +11,8 @@ import (
 func TestRichTextMarkdown_Blockquote(t *testing.T) {
 	r := NewRichTextFromMarkdown("p1\n\n> quote\n\np2")
 
-	assert.Equal(t, 3, len(r.Segments))
-	if text, ok := r.Segments[1].(*TextSegment); ok {
+	assert.Equal(t, 5, len(r.Segments))
+	if text, ok := r.Segments[2].(*TextSegment); ok {
 		assert.Equal(t, "quote", text.Text)
 		assert.Equal(t, RichTextStyleBlockquote, text.Style)
 	} else {
@@ -23,7 +23,7 @@ func TestRichTextMarkdown_Blockquote(t *testing.T) {
 func TestRichTextMarkdown_Code(t *testing.T) {
 	r := NewRichTextFromMarkdown("a `code` inline")
 
-	assert.Equal(t, 3, len(r.Segments))
+	assert.Equal(t, 4, len(r.Segments))
 	if text, ok := r.Segments[1].(*TextSegment); ok {
 		assert.Equal(t, "code", text.Text)
 		assert.Equal(t, RichTextStyleCodeInline, text.Style)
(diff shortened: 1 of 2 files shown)


andydotxyz:
> Since the new implementation adds leading spaces instead of trailing ones, and also adds new segments for some spaces and the end of paragraphs, I had to adapt the existing tests slightly.

Moving where spaces are inserted seems OK, so the test changes where it is moved from one segment to another seems OK.
However adding trailing spaces to handle the newline doesn't seem right - because these spaces will be inserted into the content and the `String()` method no longer correct.

Unless I misunderstood what you meant?

codesoap:
Sorry, I didn't explain it enough. `String()` should not be affected. No superfluous spaces are inserted. Besides the swap from trailing to leading spaces, there are these two new special segments:
1. goldmark seems to signal a single newline after emph/strong/link with an `ast.Text` node, which has no content (`""`). This is translated to a `TextSegment` with the `Text: " "`. We need to play along with this quirk of goldmark, because (as far as I can see) this is the only way we can distinguish `*foo*\nbar` from `*foo*bar`.
2. Previously the last segment of a paragraph got modified with `text.Style.Inline = false`. As far as I can see, this won't work if the last element of a paragraph is, for example, a `HyperlinkSegment`, because this doesn't have a `Style`. Thus I now always add a `TextSegment` with `Style: RichTextStyleParagraph` and no `Text` at the end of paragraphs.

An example will probably explain it better:
```markdown
*foo*
bar
```

Will now be parsed to the following 4 `TextSegment`s:
1. A segment with `RichTextStyleEmphasis` and `Text: "foo"`.
2. A segment with `RichTextStyleInline` and `Text: " "` (result of 1. from above).
3. A segment with `RichTextStyleInline` and `Text: "bar"`.
4. A segment with `RichTextStyleParagraph` and `Text: ""` (result of 2. from above).

andydotxyz:
I think I follow.
To confirm this could you check in the tests that the newly inserted text segments for line break are indeed `Text` with "" and not " " just to be sure? Or at least some of them, just so we can't regress this special marker to something else in the future.

codesoap:
I have added an explicit test for the last segment of a paragraph. Is this OK? I can also adapt the existing test functions, if you'd prefer that.

montovaneli:
Sorry to jump into the discussion, but I'm really interested in this pull request.

What would be the outcome on the screen when we pass a string with escape characters? For example, using https://markdownlivepreview.com/, we get this result:

image


Now, using this pull request and a string `s:= "FIRST LINE\n\nTHIRD LINE"` we have:

image

I also tested with "FIRST LINE  \n\rTHIRD LINE" and "FIRST LINE  \n\nTHIRD LINE".

Sorry if I misunderstood something, but it seems that a new paragraph isn't being considered.

andydotxyz:
> Sorry if I misunderstood something, but it seems that the "New paragraph" is not a new paragraph at all.

To know this you should look at the object tree of `Segments`. I suspect it is an unrelated bug where the gap between a block element and a paragraph are not spaced enough. Try testing before and after this PR patch applied to see if it changes things or not. I think they are unrelated.

widget/markdown_test.go
```diff
@@ -163,29 +163,29 @@ func TestRichTextMarkdown_Image(t *testing.T) {
...
 	if text, ok := r.Segments[0].(*TextSegment); ok {
-		assert.Equal(t, "line1 ", text.Text)
+		assert.Equal(t, "line1", text.Text)
 		assert.True(t, text.Inline())
 	} else {
 		t.Error("Segment should be Text")
 	}
 	if text, ok := r.Segments[1].(*TextSegment); ok {
-		assert.Equal(t, "line2", text.Text)
+		assert.Equal(t, " line2", text.Text)
```
andydotxyz, line 174: Just coming back to this apologies for the delay. Is this a desirable change? Putting a space at the leading of a string seems like it could cause a problem.

codesoap: Previously trailing spaces were inserted, now leading spaces are inserted instead. I cannot remember why I changed it, but I guess it made the implementation a little easier. I don't see how either choice would be better than the other. What kind of problem do you foresee?

andydotxyz: The problem I foresee is that strings are indexed from the beginning, so if anyone was actually working with the content of a TextSegment then the characters inside them are now off-by-1.

codesoap: Good point. I'll try to revert back to trailing spaces when I find some time.

codesoap: I have now revisited the code and was able to return to trailing spaces instead of leading ones. In the process, the code even became a little more compact and the `renderNode` function requires one less argument.

I hope I didn't forget anything that I had considered when originally creating this pull request, but the tests cannot find any issues.

andydotxyz (approved):
Thanks so much for working this out

codesoap:
Thanks for accepting this PR! I kinda feared it would be rejected because it changes too much and really appreciate you taking the time to consider it.

andydotxyz:
The benefit of a robust unit test suite is that the size of the change isn't usually a problem. It just takes longer.

I always go for the smallest possible fix, but in some cases refactoring is the way forward :).

PR #4680, opened 2024-02-27
Fix double refresh in entry and unnecessary string buffer allocs

### LINKED ISSUE #4397: Entry refreshes too often

Jacalz:
### Description:

This is a reimplementation of the excellent work done by #4465 but without the strange test failures. I have not managed to figure out why they were failing but I managed to get these ones to not do so. I also fixed up a few cases where erasing selection was allocating the string buffer twice and a few extra refreshes being removed compared to the other PR.

We need to take care in the future to not refresh twice and not to run `provider.String()` as often given that it does allocate a string buffer every time. I believe that this should remove all cases of duplicated calls to these APIs. It would be nice to get this into v2.4.5 as it does speed things up nicely.

Replaces #4465
Fixes #4397 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -901,16 +901,13 @@ func (e *Entry) TypedRune(r rune) {
 	// if we've typed a character and we're selecting then replace the selection with the character
 	cb := e.OnChanged
 	if e.selecting {
-		e.OnChanged = nil // don't propagate this change to binding etc
 		e.eraseSelection()
-		e.OnChanged = cb // the change later will then trigger callback
 	}
-
-	provider := e.textProvider()
-	e.selecting = false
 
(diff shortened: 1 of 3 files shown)


dweymouth (approved):
Nice work :)

PR #4681, opened 2024-02-27
RFC: Add caches to avoid recomputing MinSize multiple times

Jacalz:
### Description:

This is a request for comments about adding an automatic and built-in cache for MinSize values in BaseWidget and containers. This avoids having to recalculate the size by walking the entire tree of objects each time. It does incur a slight change of behaviours for widgets that overwrite the `MinSize()` method and use the size given from the BaseWidget to add or subtract to. I have updated all widgets to make sure that the renderer defines the needed MinSize and as such all widgets are updated to use the cache. 

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/container.go
+++ b/container.go
@@ -10,8 +10,7 @@ var _ CanvasObject = (*Container)(nil)
 type Container struct {
 	size     Size     // The current size of the Container
 	position Position // The current position of the Container
-	minCache Size
-	Hidden   bool // Is this Container hidden
+	Hidden   bool     // Is this Container hidden
 
 	Layout  Layout // The Layout algorithm for arranging child CanvasObjects
 	lock    sync.Mutex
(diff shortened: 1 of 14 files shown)


andydotxyz:
Hmm, it might be a bit out there - but perhaps this does not have to be on the widget at all.
If it is to be reset when Refresh is called then we could probably cache it in a similar way to the texture is cache?
That way it would not need any widgets to change at all...

Looking at the bigger picture the tree walk leans heavily on MinSize but does not know if it is changed - perhaps we could even save the tree walk if we are able to determine that no minimum has changed between walks?

Jacalz:
Interesting thoughts. I will try to get it to work and get all the tests to pass first using this solution and then I can looking into how the other cache works.

FYI: I think using the cache for container might be entirely flawed? We don't want to have to refresh the whole container when a widget changes, right?

Jacalz:
Continuing this in https://github.com/fyne-io/fyne/pull/4827 instead

PR #4682, opened 2024-02-27
widget: Move some minsize calculations to the renderer

Jacalz:
### Description:

Part of https://github.com/fyne-io/fyne/pull/4681 ripped out into a separate PR.
Moving calculations of the MinSize into the renderer allows us to clean up some of the code
because `BaseWidget.MinSize()` will create the renderer if it does not exists and as such do any setup necessary.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/activity.go
+++ b/widget/activity.go
@@ -34,8 +34,7 @@ func NewActivity() *Activity {
 
 func (a *Activity) MinSize() fyne.Size {
 	a.ExtendBaseWidget(a)
-
-	return fyne.NewSquareSize(a.Theme().Size(theme.SizeNameInlineIcon))
+	return a.BaseWidget.MinSize()
 }
 
 // Start the activity indicator animation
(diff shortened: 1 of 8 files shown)


andydotxyz (commented):
For Separator and Activity we don't need a renderer to know our minsize - those shortcuts were added as optimisations... so should they really be removed?

Jacalz:
Hmm. Good point. The idea was that if we move all our MinSize logic to the renderer, we can add caching to the base widget more easily and we have better separation of the code paths by keeping size logic in one place. What do you think?

Jacalz:
There is also less code duplication. Updating the sizing only has to happen in one place and there is no risk of them getting out of sync. If getting the MinSize form the renderer is slow then maybe we ought to optimise that code?

andydotxyz:
> If getting the MinSize form the renderer is slow then maybe we ought to optimise that code?

It's not slow per-se - but it requires creating a renderer, and when we are calculating how to display something that is not visible its potential overhead. Maybe not an issue in the bigger picture though?

Jacalz:
Hmm. I personally think that the less code duplication and separation of concerns is a bigger win. If we can get the minsize cache working it should be even faster going forward so hard coding renderer details into the widget itself should not be necessary.

PR #4689, opened 2024-03-01
Fix an issue where canvas.Text in form was not aligned

### LINKED ISSUE #4665: layout.formLayout do not handle canvas.Text well in second column
### Checklist

### Describe the bug

The form layout do not handle canvas.Text in the second column very well.
The texts at each line in first column do not align vertically with matching texts in second column.

### How to reproduce

Create a formlayout with canvas.Text's in both columns.

### Screenshots

image

### Example code

```go
	form := container.New(layout.NewFormLayout(),
		canvas.NewText("label1", color.White), canvas.NewText("value1", color.White),
		canvas.NewText("label2", color.White), canvas.NewText("value2", color.White),
		canvas.NewText("label3", color.White), canvas.NewText("value3", color.White),
	)
	window.SetContent(form)
```

### Fyne version

2.4.4

### Go compiler version

1.21.3

macOS Sonoma 14.2.1

### Additional Information

A hint to a solution and where the problem might lie.
This is a fix I've used in my own project.
The solution for rendering second column is merely trying to mirror the special case handling of canvas.Text in first column. 

**DO NOTE: The solution is not verified or tested but within my own code with acceptable visual result.**

```go
File: formlayout.go

func (f *formLayout) Layout(objects []fyne.CanvasObject, size fyne.Size) {
  // [...]

		tableRow := table[row]
		if _, ok := objects[i].(*canvas.Text); ok {
			objects[i].Move(fyne.NewPos(innerPadding, y+innerPadding))
			objects[i].Resize(fyne.NewSize(tableRow[0].Width-innerPadding*2, objects[i].MinSize().Height))
		} else {
			objects[i].Move(fyne.NewPos(0, y))
			objects[i].Resize(fyne.NewSize(tableRow[0].Width, tableRow[0].Height))
		}

		if i+1 < len(objects) {
                         // Here is the fix with special case for canvas.Text in second column as well
			if _, ok := objects[i+1].(*canvas.Text); ok { 
				objects[i+1].Move(fyne.NewPos(padding+tableRow[0].Width+innerPadding, y+innerPadding))
				objects[i+1].Resize(fyne.NewSize(tableRow[1].Width-innerPadding*2, objects[i+1].MinSize().Height))
			} else {
				objects[i+1].Move(fyne.NewPos(padding+tableRow[0].Width, y))
				objects[i+1].Resize(fyne.NewSize(tableRow[1].Width, tableRow[0].Height))
			}
		}

  // [...]
}
```

andydotxyz:
As suggested in the issue the only sensible course is to add special case to content just like label.
It feels dirty, but we cannot remove the label special case, so this is all that makes sense.

Fixes #4665

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/layout/formlayout.go
+++ b/layout/formlayout.go
@@ -110,8 +110,13 @@ func (f *formLayout) Layout(objects []fyne.CanvasObject, size fyne.Size) {
 		}
 
 		if i+1 < len(objects) {
-			objects[i+1].Move(fyne.NewPos(padding+tableRow[0].Width, y))
-			objects[i+1].Resize(fyne.NewSize(tableRow[1].Width, tableRow[0].Height))
+			if _, ok := objects[i+1].(*canvas.Text); ok {
+				objects[i+1].Move(fyne.NewPos(padding+tableRow[0].Width+innerPadding, y+innerPadding))
+				objects[i+1].Resize(fyne.NewSize(tableRow[1].Width-innerPadding*2, objects[i+1].MinSize().Height))
+			} else {
+				objects[i+1].Move(fyne.NewPos(padding+tableRow[0].Width, y))
+				objects[i+1].Resize(fyne.NewSize(tableRow[1].Width, tableRow[0].Height))
+			}
 		}
 		row++
 	}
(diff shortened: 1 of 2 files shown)


dweymouth (approved):
Makes sense, even though it can feel a bit like a hack. As you said, I don't think there's another way around it.

PR #4693, opened 2024-03-03
fix: Make gradient resizing refresh and not use cached textures

### LINKED ISSUE #4608: Gradient doesn't work!
### Checklist

### Describe the bug

I write simple code to try how gradient work, but when I run it, there in the window isn't a gradient from white to black, but only grey color.

### How to reproduce

See the example code.

### Screenshots

Снимок экрана от 2024-02-04 13-10-29

### Example code

```go
package main

import (
	"image/color"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/canvas"
)

func main() {
	myApp := app.New()
	w := myApp.NewWindow("Gradient")

	gradient := canvas.NewHorizontalGradient(color.White, color.Black)
	w.SetContent(gradient)

	w.Resize(fyne.NewSize(100, 100))
	w.ShowAndRun()
}
```

### Fyne version

v2.4.2

### Go compiler version

1.21.5

Linux Mint 21.3 Virginia

### Additional Information

_No response_

adamantike:
### Description:

Avoid using cached textures for gradients, by refreshing the object on resize.

Fixes #4608.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/canvas/gradient.go
+++ b/canvas/gradient.go
@@ -73,6 +73,16 @@ func (g *LinearGradient) Move(pos fyne.Position) {
 	repaint(g)
 }
 
+// Resize the gradient to a new size. Causes a refresh, as cached textures become invalid on resize.
+func (g *LinearGradient) Resize(size fyne.Size) {
+	if size == g.Size() {
+		return
+	}
+	g.baseObject.Resize(size)
+
+	g.Refresh()
+}
+
 // Refresh causes this gradient to be redrawn with its configured state.
 func (g *LinearGradient) Refresh() {
 	Refresh(g)
@@ -134,6 +144,16 @@ func (g *RadialGradient) Move(pos fyne.Position) {
 	g.baseObject.Move(pos)
 
 	repaint(g)
+}
+
+// Resize the gradient to a new size. Causes a refresh, as cached textures become invalid on resize.
+func (g *RadialGradient) Resize(size fyne.Size) {
+	if size == g.Size() {
+		return
+	}
+	g.baseObject.Resize(size)
+
+	g.Refresh()
 }
 
 // Refresh causes this gradient to be redrawn with its configured state.
--- a/canvas/gradient.go
+++ b/canvas/gradient.go
@@ -147,13 +147,14 @@ func (g *RadialGradient) Move(pos fyne.Position) {
 	repaint(g)
 }
 
-// Resize the gradient to a new size. Causes a refresh, as cached textures become invalid on resize.
+// Resize resizes the gradient to a new size.
 func (g *RadialGradient) Resize(size fyne.Size) {
 	if size == g.Size() {
 		return
 	}
 	g.baseObject.Resize(size)
 
+	// refresh needed to invalidate cached textures
 	g.Refresh()
 }
 
--- a/canvas/gradient.go
+++ b/canvas/gradient.go
@@ -73,13 +73,14 @@ func (g *LinearGradient) Move(pos fyne.Position) {
 	repaint(g)
 }
 
-// Resize the gradient to a new size. Causes a refresh, as cached textures become invalid on resize.
+// Resize resizes the gradient to a new size.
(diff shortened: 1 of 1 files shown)


canvasobject.go
```diff
@@ -105,3 +105,9 @@ type Tabbable interface {
 type Tappable interface {
 	Tapped(*PointEvent)
 }
+
+// CacheExpirable describes any CanvasObject that can expire its cache.
+// Its methods could gain more arguments, as objects require other information to determine if the cache is expired.
+type CacheExpirable interface {
+	IsCacheExpired(Size) bool
+}
```
Jacalz, line 113: Do we really need this interface? We have access to `.Size()` and the size we pass in so I think it seems unnecessary to have the interface as well.

Jacalz (changes requested):
Thanks for looking into this. I just have a comment on the interface perhaps being unnecessary.

canvasobject.go
adamantike: My understanding is that the cache texture is used by many other Canvas objects besides the `LinearGradient`/`RadialGradient`, so this interface avoids invalidating the cache for those other objects.

I haven't tested the rest of the objects that use the cache, but can go through that to understand if all of them should invalidate it on size change, or if we should maintain this as an opt-in for each Canvas object.

andydotxyz: It seems like there may be a simpler solution - when `Resize` of a gradient is called we can invalidate the texture with `Refresh`...

That is how image and rectangle does it

andydotxyz:
I would check out how Rectangle (when stroke is on) or the Image canvas types, they already reset the cache when resized.
I agree with @Jacalz this should not require any new API.

canvasobject.go
adamantike: Great feedback. I found that calling `Resize` completely fixes the issue. My previous approach was to trigger `repaint` instead, but that sometimes worked and sometimes didn't, apparently because of a race condition between marking cache as dirty and a different process actually cleaning it up?

canvas/gradient.go
```diff
@@ -73,6 +73,15 @@ func (g *LinearGradient) Move(pos fyne.Position) {
 	repaint(g)
 }
 
+func (g *LinearGradient) Resize(size fyne.Size) {
+	if size == g.Size() {
+		return
+	}
+	g.baseObject.Resize(size)
+
+	g.Refresh()
```
dweymouth, line 82: I would just add a quick comment here (and in RadialGradient) saying that the reason a Refresh is needed is that the texture will need to be regenerated. It's not a common pattern to Refresh in a resize (but it makes total sense here), so let's make sure someone wouldn't remove it later thinking they're making an optimization

adamantike: Added a small comment to both functions, to have more context. I expect people trying to delete existing code to be good friends with git history, but that's not always the case :)

dweymouth (approved):

canvas/gradient.go
```diff
@@ -73,6 +73,16 @@ func (g *LinearGradient) Move(pos fyne.Position) {
 	repaint(g)
 }
 
+// Resize the gradient to a new size. Causes a refresh, as cached textures become invalid on resize.
```
dweymouth, line 76: Ah, sorry I have one more suggestion - the second sentence about refreshing should just be moved to an inline code comment just above the g.Refresh() call, as it's more of an internal implementation detail than an API documentation. (And in RadialGradient too)

dweymouth (changes requested):

dweymouth (approved):

dweymouth:
> I forgot to review this one. Sorry about that. Plan to do so tomorrow evening

Oh whoops, I just merged before I saw this

Jacalz:
No problem. I thought you couldn't review before I had approved so I'm kind of just glad that you relieved me of my duties :)

PR #4694, opened 2024-03-03
Mimic desktop behaviour for FileSave/FileOpen dialogs: focus and submitting by keyboard

ivoras:
### Description:
Two commits, each introducing one change:

1. Submitting (clicking ENTER) while in the fileName Entry the SaveDialog submits the form (clicks the Save button)
2. When the Save / Open file dialog is shown, the fileName entry is focused immediately

This mimics the native behaviour on desktop environments (at least Windows & Linux, can't test on OSX).

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -8,7 +8,6 @@ import (
 	"runtime"
 	"strings"
 	"sync"
-	"time"
 
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/container"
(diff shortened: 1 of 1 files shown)


dialog/file.go
```diff
@@ -644,6 +648,13 @@ func showFile(file *FileDialog) *fileDialog {
 
 	d.setLocation(file.effectiveStartingDir())
 	d.win.Show()
+	if file.save {
+		// Focus on the filename Entry as soon as the dialog is shown
+		go func() {
+			time.Sleep(25 * time.Millisecond)
+			file.dialog.win.Canvas.Focus(file.dialog.fileName.(*widget.Entry))
+		}()
```
Jacalz, line 656: You can focus things before they are shown as long as the items are part of the window contents. Spinning up a goroutine and sleeping for a random amount of time seems like a hacky workaround.

Jacalz (changes requested):
Thanks for these qol improvements. I left a comment inline about cleaning up the code a bit. Also, would you mind giving the PR a descriptive title?

dialog/file.go
ivoras: Done. I see why.

Jacalz: Remove the comment as well. It no longer valid (technically wasn't 100% valid before either).

ivoras: Done

Jacalz (approved):
Thanks for the nice QOL improvements :+1:

PR #4701, opened 2024-03-05
Widget scrolling plus

### LINKED ISSUE #4695: home/end & page up/down keys dont work
### Checklist

### Describe the bug

Cant scroll with the keys specified in the title inside the list collection within the fyne demo!

### How to reproduce

Open fyne demo, go to Collections -> List and try to scroll using one of the 4 keys.

### Screenshots

_No response_

### Example code

There is no code.

### Fyne version

2.4.0

### Go compiler version

1.21.7

Fedora 39

### Additional Information

Should work in all containers.

m-rei:
### Description:

Fixes #4695

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


widget/table.go
```diff
@@ -344,6 +344,17 @@ func (t *Table) TouchCancel(*mobile.TouchEvent) {
 // Implements: fyne.Focusable
 func (t *Table) TypedKey(event *fyne.KeyEvent) {
 	switch event.Name {
+	case fyne.KeyHome:
+		t.ScrollToTop()
+		t.Refresh()
```
dweymouth, line 349: The Refresh() calls here should not be necessary since the Scroll APIs already refresh

widget/table.go
```diff
@@ -541,6 +552,33 @@ func (t *Table) ScrollToTop() {
 	t.finishScroll()
 }
 
+// scrollByOnePage scrolls down or up by table height
+func (t *Table) scrollByOnePage(down bool) {
+	if t.Length == nil || t.content == nil {
```
dweymouth, line 557: You may need to add some offset validation here - please test what happens if you try to keep scrolling when you're already at the top or bottom of the list

widget/table.go
m-rei: You were partially right.

The default ScrollToTop/ScrollToBottom did not work correctly, a t.content.Refresh() was missing before t.finishScroll(), so I added it. Without it, it does scroll to top or bottom, but the table becomes blank.

widget/table.go
m-rei: There is one bug in the tree widget, but its not introduced with this PR, it just became obvious after implementing it.

When you call tree.ScrollToBottom() twice, the last row's top border vanishes.

Other than that, there are no problems. Not sure if I should investigate it, since its not part of this PR.

I can share some example code (120 lines).

widget/table.go
andydotxyz: Sorry this has sat for so long. Something doesn't seem right though. Are you saying that the current "ScrollToTop" or "ScrollToBottom" don't work unless there is a Refresh call in the developers code?

widget/list.go
```diff
@@ -274,6 +290,16 @@ func (l *List) ScrollToOffset(offset float32) {
 	l.Refresh()
 }
 
+// ScrollUpOnePage scrolls up one page (table height)
+func (l *List) ScrollUpOnePage() {
```
dweymouth, line 294: I'm not sure these need to be public APIs (you can do the equivalent by `ScrollToOffset(l.GetScrollOffset() + l.Size().Height)`) but if they are going to be new public APIs they need the `// Since: 2.5` designation

widget/table.go
andydotxyz: If its not related then open an issue - the example code will be useful there :)

PR #4713, opened 2024-03-11
Move from go get to git clone for fyne get

### LINKED ISSUE #4684: cmd/fyne: command ”get” is broken with Go 1.22
### Checklist

### Describe the bug

Generally my Golang installation is working OK. I have previously used an old version of Fyne (1.x) and I don't recall any issues. I'm now trying to install fyne_settings with the command `fyne get github.com/fyne-io/fyne/cmd/fyne_settings` and I can't get past the dreaded error "go: modules disabled by GO111MODULE=off" when in fact, that is not the case.

Here is the output:

```
fyne get github.com/fyne-io/fyne/cmd/fyne_settings
go: modules disabled by GO111MODULE=off; see 'go help modules'
exit status 1

go env | grep GO111MODULE
GO111MODULE='auto'

echo $GO111MODULE

 
```

### How to reproduce

1. From command prompt enter `fyne get github.com/fyne-io/fyne/cmd/fyne_settings`
2. Observe that it probably worsks for you.
3. It doesn't for me!

### Screenshots

_No response_

### Example code

Code is not relevant to this issue

### Fyne version

v2.4.4

### Go compiler version

1.22.0

macOS Monterey

### Additional Information

Clearly, given nobody else seems to be reporting this problem, this is highly likely a 'me' problem but I'm not sure where to start.

andydotxyz:
This clearly belongs on the new repo - but first I wanted to land it here to pick onto the release branch for quicker fix :)

Fixes #4684

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


cmd/fyne/internal/commands/get.go
```diff
@@ -10,6 +10,9 @@ import (
 
 	"github.com/urfave/cli/v2"
 	"golang.org/x/sys/execabs"
+
+	//lint:ignore SA1019 The recommended replacement does not solve the use-case
+	"golang.org/x/tools/go/vcs"
```
Jacalz, line 15: I understand what the ignore comment is saying but I'm a bit worried about starting to import a module that never will get updated any more. Is there not a better package we can use? Or some other way to do it?

Jacalz (changes requested):
See note inline about importing a deprecated package. Maybe there is something better we can use?

cmd/fyne/internal/commands/get.go
andydotxyz: The only other way to do it would be to copy a load of code out of the go/cmd/go internal module resolver code.
See the ongoing feature request at https://github.com/golang/go/issues/18387

Jacalz: Alright. That makes a lot of sense. I was just wondering if there was some other third-party package that did the same thing. If that's not the case then I think your approach makes complete sense.

Jacalz: Alright. I see that now. Will test the code later today and approve if everything looks good :)

cmd/fyne/internal/commands/get.go
```diff
@@ -57,20 +60,36 @@ func NewGetter() *Getter {
 
 // Get automates the download and install of a named GUI app package.
 func (g *Getter) Get(pkg string) error {
-	cmd := execabs.Command("go", "get", "-u", "-d", pkg)
-	cmd.Env = append(os.Environ(), "GO111MODULE=off") // cache the downloaded code
+	name := filepath.Base(pkg)
+	path := filepath.Join(os.TempDir(), name)
+	if util.Exists(path) {
+		_ = os.Remove(path)
```
Jacalz, line 66: See the generic review comment for background. Is it not better to do `os.CreateTemp()` to get a fresh random temp directory and then delete it afterwards instead?

Jacalz (changes requested):
This does not seem to clean up the temporary directory after installation. Trying to install twice (because I got an error the first time) results in this:
```
fatal: destination path '/tmp/v3' already exists and is not an empty directory.
exit status 128
```
I think you are missing a `defer os.Remove()` or similar.

I am not sure if this is due to this PR or not (my Go install did apparently not work at all with `fyne get` before this change) but I'm getting this error when trying to install Rymdport (fyne_demo works fine):
```
$ fyne get github.com/Jacalz/rymdport/v3

Cloning into '/tmp/v3'...
remote: Enumerating objects: 78, done.
remote: Counting objects: 100% (78/78), done.
remote: Compressing objects: 100% (75/75), done.
remote: Total 78 (delta 1), reused 39 (delta 1), pack-reused 0
Receiving objects: 100% (78/78), 170.37 KiB | 2.43 MiB/s, done.
Resolving deltas: 100% (1/1), done.

failed to set up installer: Missing application icon at "/tmp/v3/v3/Icon.png"
```

andydotxyz:
> The error seems to be that the installer expects an Icon.png file and doesn't read
> FyneApp.toml? I'm quite sure that Rymdport used to work when installing through fyne get 
> but the last time I tested was likely over a year ago (or more likely even longer)

This is not a regression, the icon handling is somewhat special for get but it should be resolved separately. You can pass "-icon" command to work around it for now.

cmd/fyne/internal/commands/get.go
andydotxyz: CreateTemp creates files not directories

andydotxyz: I think that works nicely, thanks.

Jacalz (commented):
Hmm. This just seems to get stuck after cloning the repository from what I can tell. It resolves the deltas and then just seems to do absolutely nothing when trying to download fyne_demo. I'm not sure what is going on.

Jacalz:
Actually: It might just be my Go install being stupidly slow at compiling for some reason. Will test again tomorrow.

Jacalz (approved):

PR #4715, opened 2024-03-11
Read from FyneApp.toml when developing with go tools

### LINKED ISSUE #4688: Please add a way to embed the `FyneApp.toml` metadata into the `go build`.
### Checklist

Simply put, I want to disable `-H=windowsgui` and embed the `FyneApp.toml` metadata.
See #3781 and #3604.

The option to disable `-H=windowsgui` in `fyne build` was not accepted.
However, if there is a way to add metadata when `go build`, the problem is solved.

_No response_

Add a new subcommand to the fyne command to generate metadata source code.
The source code to set the metadata is generated at `fyne build` time.
The source code for the following section will be helpful.

https://github.com/fyne-io/fyne/blob/e332a5e47813bff1af9631cbbfe84654c0143189/cmd/fyne/internal/commands/build.go#L377C6-L377C28

Or maybe there is a better way. Is there anything?

andydotxyz:
Fixes the go run and go build finding metadata during development.

Fixes #4688 and https://github.com/fyne-io/tools/issues/12

### Checklist:

- [ ] Tests included. <- I didn't want to add Go compile (times) to the test process...
- [x] Tests all pass.


Jacalz:
I’m more entirely sure about this. Now every app will try to load FyneApp.toml on startup from the local directory. That doesn’t go well with the fact that this is a very specific developer problem mostly on Windows.

I think we either need a better solution or make sure that this functionality is turned off when building a release binary. I prefer the former though as I can’t build Flatpak apps with -tags release.

Jacalz:
Maybe the better option just simply is to allow the developer to turn off the windowsgui linker flag?

andydotxyz:
> Maybe the better option just simply is to allow the developer to turn off the windowsgui linker flag?

I don't understand this comment? The change enables usage of "go run" lifecycle and using the "run" button inside IDEs.
It isn't only about fixing the issue where the windows console is not connected.

andydotxyz:
> Now every app will try to load FyneApp.toml on startup from the local directory.

Unless I coded it wrong I don't think this is the case. It will only look for local metadata if none was packaged. Which means for any packaged app, or released bundle, this code will never execute.

> That doesn’t go well with the fact that this is a very specific developer problem mostly on Windows. 99% of all non development installs will not use this.

Erm... this is something that is useful to anybody wanting to use standard Go tools for their development cycle instead of packaging then running if their app uses metadata...

Jacalz:
I’m sorry if my comments didn’t summarise clearly what I meant. The use case in the issue describes Windows and not wanting `-H windowsgui` as a linker flag and hence wanting to use the regular `go` commands. We added “fyne build” to have a way for embedding metadata for builds in the first place. Why work around that and add runtime lookup when not using it? I’d be more in favour of adding a “fyne run” command and potentially an option to turn off `-H windowsgui` for Windows developers that care about it.

If a user wants to have metadata during development, then switching to “fyne build” from “go build” is not such a bug change. We did design the build command to basically a drop in replacement. If that’s not the case for some use cases we might want to look into that.

Jacalz:
> Unless I coded it wrong I don't think this is the case. It will only look for local metadata if none was packaged. Which means for any packaged app, or released bundle, this code will never execute.

Yes, I expressed myself incorrectly there. What I mean is that there are cases where we have production builds that just can't use the `fyne` command for building. This bacillus applies to Flatpak and many Linux package build systems where there is sandboxing and no network access to install the command. I.e. all of those apps will now try to load a FyneApp.toml from the local directory which I'm not a huge fan of.

andydotxyz:
> This bacillus applies to Flatpak and many Linux package build systems where there is sandboxing and no network access to install the command. I.e. all of those apps will now try to load a FyneApp.toml from the local directory which I'm not a huge fan of.

I think this is a different problem and not one I had fully comprehended - they will not support metadata at all in the current state, or after the PR lands (as the file will be missing at runtime)

Jacalz:
> The Go tools create simple binaries and the fyne tool is for bundling graphical apps. This seems like a strong divide that we should stick to.

I don't quite agree with that given that we added the "fyne build" command for the very reason of building regular binaries in the same sense as "go build" without any bundling but with customisations to make the binaries more adapted to guis.

andydotxyz:
> > The Go tools create simple binaries and the fyne tool is for bundling graphical apps. This seems like a strong divide that we should stick to.
> 
> I don't quite agree with that given that we added the "fyne build" command for the very reason of building regular binaries in the same sense as "go build" without any bundling but with customisations to make the binaries more adapted to guis.

Maybe I used the wrong words, but the sentiments are basically the same - go tools for CLI and fyne tools for GUI apps. If you add an option to create a GUI app without it being a GUI app then we will have to explain complex technical issues specific to OS that (if we have done our job right) people should never have to worry about.

app/meta_development.go
```diff
@@ -0,0 +1,40 @@
...
+	"os"
+	"path/filepath"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/internal/metadata"
+)
+
+func checkLocalMetadata() {
+	file, _ := filepath.Abs("FyneApp.toml")
+	ref, err := os.Open(file)
```
dweymouth, line 15: missing defer Close?

app/app_openurl_web.go
```diff
@@ -7,6 +7,6 @@ import (
 	"net/url"
 )
 
-func (app *fyneApp) OpenURL(url *url.URL) error {
+func (a *fyneApp) OpenURL(url *url.URL) error {
```
Jacalz, line 10: Wonderful. That has been bothering me a while :)

app/meta_development.go
```diff
@@ -0,0 +1,40 @@
+//go:build !no_metadata
```
Jacalz, line 1: Instead of more complexity (more files) in the `app/` folder, can you not use the new `internal/build` package and add a `build.NoMetadata` constant?

app/meta_development.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"os"
+	"path/filepath"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/internal/metadata"
+)
+
+func checkLocalMetadata() {
+	file, _ := filepath.Abs("FyneApp.toml")
```
Jacalz, line 14: You might want to check the error here, I think

Jacalz (changes requested):
I added some comments inline. Also, should we try to not compile in the runtime metadata lookup if we know that the metadata was added? I.e. should we avoid compiling it in if we have `release` as build mode?

One thought (not necessarily something you need to fix): This can theoretically lead to strange behaviour if a binary was built without metadata in the past and runs in a directory that has a `FyneApp.toml` file. Might not be a meaningful problem but I feel it is worth noting.

Jacalz:
Sorry, I realised something. It is worth noting that this does bring in the toml dependency into every binary. You ought to check binary sizes and add that checkbox back to the PR template.

Jacalz:
Thinking a bit more about this, I'm afraid that think this kind of brings more downsides for applications than upsides in its current implementation. Bundling the toml parsing into every app, apps without metadata will try to load metadata from PWD and an app without metadata potentially loading the wrong metadata if launched from another directory seems bad in my opinion.

I'd instead propose adding a "runtime_metadata" tag to explicitly enable this for those that want it. If we introduce a "fyne run" command, it can use that tag automatically.

andydotxyz:
I take your point - but adding a flag to do this essentially removes the point of it.
The idea was that go tools would just work. Requiring an additional flag is a different, but still additional, requirement for metadata to function.

If you do add that flag then the same problem you report with taking the wrong data still occurs for any apps built with that turned on.

Jacalz:
> I take your point - but adding a flag to do this essentially removes the point of it.
The idea was that go tools would just work. Requiring an additional flag is a different, but still additional, requirement for metadata to function.

We introduced the "fyne build" command basically to support metadata because "go build" couldn't. I agree that we for the most part shouldn't force people to use our cmd commands but I do see a tag for runtime metadata lookup as the best option given the tradeoffs mentioned above.

> If you do add that flag then the same problem you report with taking the wrong data still occurs for any apps built with that turned on.

Yes but there is a very important distinction there. The potentially problematic behaviour is no longer by default and won't he open without you knowing.

andydotxyz:
> Bundling the toml parsing into every app, apps without metadata will try to load metadata from PWD (unless it uses that tag to avoid it) and an app without metadata potentially loading the wrong metadata if launched from another directory seems bad in my opinion.

This has been addressed with the latest commit - loading only from the source directory (or PWD if it's a temporary go build i.e. "go run").

andydotxyz:
As a general comment I don't like the idea that "just working" is something we should /not/ strive for...
There has been so much work put into making the idiosyncrasies of platforms and tools go away. For example:

* macOS apps can only display notifications if they have been installed as a full app bundle - so we add a fallback for when running as go run or other pre-release package format.
* building for web browser required specific versions of Go runtime to be installed so we stripped out GopherJS so that it would work with whatever versions the developer is running.

I would hope that everything we do is to make either the user or the developers life easier, and to take the path of least surprise.

Jacalz:
I kind of think it is easier to put the "you need to look up this compile tag to improve things" on the developer of the Fyne app rather than the package maintainer.

andydotxyz:
I'm sorry but I don't agree with this. The go tools and source build are used a lot more by developers and consumers of GitHub code than packagers.
Developer tools are primarily for development and not packaging.

With the current exception of flatpak app packaging is done with "fyne" tool.

Jacalz:
> With the current exception of flatpak app packaging is done with "fyne" tool.

Not Flatpak only. I'd say that it is more like all of the Linux packages in repositories. I.e. Rymdport I'm AUR, Solus, Nix and so on. That's the majority of Linux packages there. Might even apply to BSD* if their build systems behave in a similar way.

Jacalz:
Let me present my problem in Rymdport. I should have done so earlier, sorry. This might make things more sensible to reason about. So, I have some code to make sure that I'm not bundling the icon twice. If the application is built with `-tags release`, I do nothing. If it isn't, I use `go:embed` to bundle the icon in automatically. This works great for development builds and for Flatpak, Solus and the other Linux packages.

Given what I wrote below, I can't pass `-tags release` to the Linux packages because they wouldn't get the icon to work. With some extra thought, I realise now that the `no_metadata` flag really would solve my issues as long as we make sure to have `-tags release` turn it off as well (so release binaries don't compile in the whole TOML parser).

Jacalz:
Just a side thought: What if we approached this in a different way? My writing about `go:embed` made me realise something. Like Drew commented on above, we are essentially side-stepping the official Go build tooling in a sense. For "fyne build" this makes a lot of sense to make the metadata compile time but what if we could get development builds to work without doing so at runtime?

The idea (pseudo-code ish):
We have the `SetMetadata(fyne.Metadata)` method already. If we expose the metadata parser as a `metadata.Parse([]byte) fyne.Metadata` function, it would allow the user to do this if they want to:

```go
//go:embed FyneApp.toml
var fyneAppToml []byte

func main() {
    a := app.New()

    a.SetMetadata(metadata.Parse(fyneAppToml))
```

The metadata parser wouldn't have to be imported anywhere else in the codebase. Go automatically embeds FyneApp.toml into the binary and all is well. Having it not embed metadata twice for release builds requires a bit more code if the user cares about it but I kind of like the idea.

andydotxyz:
> Not Flatpak only. I'd say that it is more like all of the Linux packages in repositories. I.e. Rymdport I'm AUR, Solus, Nix and so on. That's the majority of Linux packages there. Might even apply to BSD* if their build systems behave in a similar way.

I think we are losing sight of the purpose of the tools we have created. The `fyne` command is to help with packaging and other aspects of GUI app creation that are required along the path to getting your app out there. It is not a development tool. When I am preparing linux builds I use the `fyne` commands and it works great. The exception we are dealing with is that the flatpak build system apparently won't allow the use of this 3rd-party tool so we have to work around it. It feels like we are focusing on the exceptions rather than the design principles in this discussion. During development this should all just work.

andydotxyz:
>  I realise now that the no_metadata flag really would solve my issues as long as we make sure to have -tags release turn it off as well (so release binaries don't compile in the whole TOML parser).

Good idea - done.

> This might make things more sensible to reason about. So, I have some code to make sure that I'm not bundling the icon twice.

There should be no code required in applications to make sure that metadata is all bundled correctly. With these changes it will work in all cases except when you are manually turning metadata off. If that is a desire of your app then the same build flags can be used to insert your own version of metadata instead. No overlap and no ambiguity.

andydotxyz:
>  If we expose the metadata parser as a metadata.Parse([]byte) fyne.Metadata function, it would allow the user to do this if they want to

This is returning to the old problem of "why is my icon not working" bugs where the code was missing, so we added automatic metadata handling so that it just worked. Reverting to the old method is going to introduce new unexpected corner cases and require that people a) read the documentation and b) insert more code into their app to get it working for the default tooling. Both are things I would like to avoid, especially as we saw this exact thing play out with app icons before it.

Jacalz:
Would you mind addressing the review comments that I have inline from the last review? I think they are still valid.

Jacalz:
> This is returning to the old problem of "why is my icon not working" bugs where the code was missing, so we added automatic metadata handling so that it just worked.

It was just a thought experiment around using more `go:embed` to spark some ideas :)

Jacalz:
> The exception we are dealing with is that the flatpak build system apparently won't allow the use of this 3rd-party tool so we have to work around it.

All I was trying to say is that it isn't an exception given that it basically applies to all Linux package systems because we can't rely on the `cmd/fyne` version being the same as what we are statically linking in out apps (if it even exists as a package). Anyway, I think this the issues that I brought up have been addressed with the latest changes.

andydotxyz:
> All I was trying to say is that it isn't an exception given that it basically applies to all Linux package systems because we can't rely on the cmd/fyne version being the same as what we are statically linking

I think that "packager does not have the latest packaging tool installed" can be considered a corner case. It's surely in the same camp as "packager's Go version is not new enough for this app"?

app/meta_development.go
andydotxyz: This went away

andydotxyz:
> Would you mind addressing the review comments that I have inline from the last review? I think they are still valid.

Done. As far as I can see it was only a missed defer that still applied.

Jacalz:
> Done. As far as I can see it was only a missed defer that still applied.

I think this comment still is valid:
https://github.com/fyne-io/fyne/pull/4715#discussion_r1521673410

Jacalz:
> I think that "packager does not have the latest packaging tool installed" can be considered a corner case. It's surely in the same camp as "packager's Go version is not new enough for this app"?

I wouldn't directly put it in the same camp. One makes it impossible to compile and the other would just not support metadata or some other missing functionality.

It is also worth noting that on an LTS version of Linux, the fyne command will likely never be as new as the Fyne version of the app (if we assume they stick to a minor release for the whole life cycle). A rolling release would potentially be less of an issue but there it might instead be too new if you haven't updated the app in a while.

andydotxyz:
> > Done. As far as I can see it was only a missed defer that still applied.
> 
> I think this comment still is valid: [#4715 (comment)](https://github.com/fyne-io/fyne/pull/4715#discussion_r1521673410)

Apologies, for some reason GitHub had completely hidden that.

app/meta_development.go
Jacalz: You are right that it won't result in less files but it would avoid adding more files to the already crowded app/ folder. The files in the other package are easier to grasp what they do. It is also consistent with how we check other build tags using that internal/build package.

andydotxyz: OK done.

Just to be clear this change means:
* 1 extra file
* A package public constant checked in 1 place
* the toml package cannot be ignored during compile

Jacalz: The first two are correct but will have less meaning if we use them in more places in the future.

The third one, about toml not being ignored should not be correct. As the NoMetadata values is constant, the compiler can dead-code eliminate the branches where it is used.

andydotxyz: > The third one, about toml not being ignored should not be correct. As the NoMetadata and Type values are constant, the compiler can dead-code eliminate the branches where they are used.

I guess this is getting a little into semantics but I think we are both right. The package cannot be ignored during compile - however it may be possible that the compiler later optimises it out as you say.

Jacalz: Yup. It will likely be downloaded but it should not be part of the binary :)

andydotxyz:
OK, I think this is now a solution that 

a) "just works" for most go based development flow
b) will only use the FyneApp.toml next to the executable, unless invoked during "go run" in which case it uses PWD (as the bin is actually in /tmp)
c) can be switched off manually and will be disabled for release builds

dweymouth:
I would just check for binary size increases with a release build, and if it doesn't go up, or goes up minimally (a few %?), I'm fine with this now

Jacalz:
Sounds very sensible. Bigger difference for a regular development build?

dweymouth (approved):

andydotxyz:
> Sounds very sensible. Bigger difference for a regular development build?

yeah it was around 2/3 of 1%

andydotxyz:
I think the changes requested are all sorted @Jacalz

Jacalz:
I think so too after a quick glance. Will review once I'm back at my computer on Monday

app/meta_development.go
```diff
@@ -0,0 +1,64 @@
...
+		fyne.LogError("failed to lookup build executable", err)
+		return work
+	}
+
+	temp := os.TempDir()
+	if strings.Contains(exe, temp) { // this happens with "go run"
+		return work
+	}
+
+	// we were called with after "go build" completed
```
Jacalz, line 62: I think there is a mistake in this comment. The wording does not quite make sense?

Jacalz (changes requested):
Just one minor comment that has some strange wording. Approved in all other ways :)

Issue #2637, opened 2021-11-11
Add support for building without bundled fonts


Jacalz:
### Is your feature request related to a problem? Please describe:

Similar to https://github.com/fyne-io/fyne/issues/2636 but for systems like Linux or BSD when you don't wan to bundle fonts because you wan to specify different fonts at runtime. This could be useful for package managers and Linux/BSD distros where they want to globaly define one font instead of bundling it into every binary.

As an example, let's say that we are building FyshOS and each system applications is a Fyne application. We thus have FyneDesk, a clock app, a calculator, a calendar,  a contacts app, a weather app and a software center, to name a few. We might want to then set the `FYNE_FONT=` environment variables globally and avoid many unnecessary megabytes of duplicated fonts.

It could also be useful for people using other languages which simply use different fonts in the first place and thus don't need or care about the bundled fonts. 

### Is it possible to construct a solution with the existing API?

Might be possible to add build tags.

### Describe the solution you'd like to see:

A build tag to build without bundled fonts.

andydotxyz:
In theory yes, but I don't think we should support this with only a FYNE_FONT environment fallback. Then applications launched without the env will be unusuable.
So I think we may need to solve the issue of looking up fonts before this makes sense

PR #4721, opened 2024-03-14
Some text improvements including falling back to system fonts

andydotxyz:
Fixes are in separate commits, the fallback code is basically first-pass.
We include lookup for OS font for the current language and whatever font the OS has for japanese, which typically includes most script languages.

This provides good (tested) coverage of rendering out of the box for macOS, iOS, Linux and Android (windows still to be tested).
Memory increase is currently 5% - 20% depending on how many fonts the app uses.
There are more optimisations to come that will counteract the increase.
The code is now faster than before so that's the current tradeoff

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
Still some tests to revise, but this is pretty much code complete I think.

image

Jacalz:
How much work is left for this to be ready?

dweymouth:
Same question as Jacalz, also wondering if this will cover Chinese and Korean as well even if the locale is EN. (Those along with Japanese are the most common non-Latin scripts in peoples' music libraries ;)

andydotxyz:
Yes. It falls back on a rune basis so unidentified characters should all be looked up.
Functionally the only thing missing is when you need locale info to shape the glyphs in a run - but that is a step later than looking up the font.

I will try and address the rest issues this weekend to get the PR landed.

andydotxyz:
On Windows and macOS (not Linux, iOS or Android) sometimes the symbol fonts align slightly differently on the sub-pixel.
I have disabled the test for now, it makes no difference to the functionality and I did not want to hold up people checking this out...

Jacalz (commented):
I wonder if it perhaps would be easier to test this if we could easily disable the built in fonts similar to https://github.com/fyne-io/fyne/issues/2637?

andydotxyz:
``

> I wonder if it perhaps would be easier to test this if we could easily disable the built in fonts similar to #2637?

I don't see why that makes it easier - if we disable the builtin fonts then it will be down to the OS/setup as to what fonts are found - basically guaranteed to be different unexpectedly on different computers - and may even break on CI.

dweymouth:
I think @Jacalz means that if you can build with no bundled fonts, at least for testing, you don't need an app that has CJK or other less common characters to test the lookup

dweymouth:
Plus now that lookup is in place we probably want to give apps the option to opt out of having any bundled fonts in the binary

Jacalz:
> I think @Jacalz means that if you can build with no bundled fonts, at least for testing, you don't need an app that has CJK or other less common characters to test the lookup

Indeed. I figured it would make things not only easier to test but also allow us to test more parts of the font lookup.

> Plus now that lookup is in place we probably want to give apps the option to opt out of having any bundled fonts in the binary

Yup. That would be nice as well down the road as more of a feature than a testing helper :)

andydotxyz:
These are good points and were agreed on Slack too as follow up work. This PR still needs a 👍 or 👎 to progress.

dweymouth (approved):
This seems good to me, at least as a first pass. More improvements can be done in follow-ups if needed. Also will be easier to continue testing with it integrated into develop.

PR #4722, opened 2024-03-14
Wire up remaining XDG Desktop Portal usage into upstream package 

Jacalz:
### Description:

This is a refcatoring of  the XDG app code to use the corresponding functionality of the rymdport/portal package. With this landed, most of the portal changes are ready. There is more work to be done but we do at least have a shared portal package now that includes all of the work we had in Fyne initially plus the file choosers and such. 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
neat :)

Jacalz:
Thanks. I'll continue working on the package by adding both more supported portals and improvements to the API. It will be nice :)

Issue #1121, opened 2020-06-21
Add option for tabs to fill the tab bar


Jacalz:
### Is your feature request related to a problem? Please describe:
I think that it might be a good idea to have a setting for the tab bar to let the tabs fill the space instead of being small. I am talking about something in line of the tabs in for example Nautilus (Gnome file manager) or Finder (on macOS) where tabs generally fill the space that they have to work with:

![image](https://user-images.githubusercontent.com/25466657/85222723-dc092880-b3bd-11ea-8c49-f25c4c6825e4.png)

![image](https://user-images.githubusercontent.com/25466657/85222743-fc38e780-b3bd-11ea-8fb8-68161058feec.png)

### Is it possible to construct a solution with the existing API?
In general yes it probably should with some changes to the layout code. It might however be necessary to have #371 to make it better with more tabs.

### Describe the solution you'd like to see:

Possibly having a field `tab.FillSpace = true` to enable it or something along those lines.

andydotxyz:
This seems like a style choice rather than functionality or a use-case?
Maybe the current setup is not ideal and can be improved, but not sure if it needs to be an API for developer to choose?

Jacalz:
Yeah I think you are right about that. I will go ahead and close this.

stuartmscott:
Reopening as I find this "stretch to fill" style desirable, tho I agree it should not be an API for devs.

Probably depends on #1310

itsmewes:
I was looking through the GitHub issues for just this. Is there any update on this by any chance? Sorry not trying to push or anything, this is the only place I've really found exactly what I'm looking for and not sure if this has been implemented yet or if there is a way to currently do this in v2.

Really enjoying getting into Fyne at the moment.

andydotxyz:
Now that `Tabs` are split to `AppTabs` and `DocTabs` I think we could clarify that this (or does it?) applies to `DocTabs`.
The reason I say that is that `AppTabs` are typically adjusted per-platform but doctabs are a more abstract concept.

atljoseph:
Stretch to fill should be available parameter.

atljoseph:
How about just add func SetStretchToFit(val bool) to both AppTabs and DocTabs structs? It would just use the mobile layout for the AppTabs/DocTabs widgets. 

But then there is a question of theme usage. Seems that the mobile layout forces a thicker tab bar and bigger icons.

andydotxyz:
[...]

PR #4756, opened 2024-03-30
Add option UseMobileLayout to force AppTab TabItem buttons to be stretched.

atljoseph:
### Description:
Add option UseMobileLayout to force AppTab TabItem buttons to be spaced evenly along the tab bar's axis instead of being aligned to the top and left. This option will give app developers the ability to have app tabs with a more refined look, if they choose, or if their needs are not met by the layout inferred for the given device. 

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/cmd/fyne_demo/tutorials/container.go
+++ b/cmd/fyne_demo/tutorials/container.go
@@ -32,6 +32,7 @@ func makeAppTabsTab(_ fyne.Window) fyne.CanvasObject {
 	for i := 4; i <= 12; i++ {
 		tabs.Append(container.NewTabItem(fmt.Sprintf("Tab %d", i), widget.NewLabel(fmt.Sprintf("Content of tab %d", i))))
 	}
+	tabs.UseMobileLayout()
 	locations := makeTabLocationSelect(tabs.SetTabLocation)
 	return container.NewBorder(locations, nil, nil, nil, tabs)
 }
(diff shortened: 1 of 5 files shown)


atljoseph:
image
image

Jacalz:
I'm not entirely sure that this is the right solution. See https://github.com/fyne-io/fyne/issues/1121 for more information.

andydotxyz:
I can't shake that the window looks like a mobile app - I wonder if that relates to this request?
What AppTabs is trying to do is adjust to the different usages that platforms default to, I guess we didn't think of what to do when effectively it's a mobile UI on a desktop computer.

andydotxyz:
If this is to be accepted then the naming would have to be more along the lines of the desired outcome instead of the internal implementation. That this is how it is done on mobile is not a factor that people coding with the API will know (or care about?) so it's more about fill instead.

Maybe it would be inspired by the fill mode used on other objects / widgets.
Also please don't add new commands to the repository just to show an API call addition.

atljoseph:
Yeah just decline it. There won’t be any more contributions from me. I’m gonna either use something else or live with it looking like butt or let someone else handle all that back and forth. But def gonna live the rest of my life without stressing over somebody else’s baby. There are more important things in life. Peace.

PR #4764, opened 2024-04-03
Add support for the default icon in Windows notifications

### LINKED ISSUE #2592: Windows: App Icon in Notification

montovaneli:
### Description:

Fixes #2592 

Adds support for the default icon in Windows notifications by changing the ToastTemplateType from ToastText02 to ToastImageAndText02 and adding the app's default icon. Also, since Windows ToastNotificationManager requests a path for the icon, we write the icon file in the temporary folder.

Example:

examplenotification

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.


Jacalz (changes requested):
Thanks for the contribution. This is a very nice improvement. I just have a suggestion inline for improving the code a bit.

app/app_windows.go
```diff
@@ -73,11 +73,21 @@ func (a *fyneApp) SendNotification(n *fyne.Notification) {
 	title := escapeNotificationString(n.Title)
 	content := escapeNotificationString(n.Content)
 	appID := a.UniqueID()
+	iconFilePath := filepath.Join(os.TempDir(), fmt.Sprintf("fyne-%s.ico", appID))
 	if appID == "" || strings.Index(appID, "missing-id") == 0 {
 		appID = a.Metadata().Name
 	}
 
-	script := fmt.Sprintf(notificationTemplate, title, content, appID)
+	if a.Icon() != nil {
```
Jacalz, line 81: I think it would be a good idea to also not save the icon each time a notification is sent. You might want to cache the icon insular to Joe we do in https://github.com/fyne-io/fyne/blob/6b2c99e1a3a79e7463a3083350070fc0c583f412/app/app_xdg.go#L122.

Jacalz (changes requested):

app/app_windows.go
```diff
@@ -73,11 +73,21 @@ func (a *fyneApp) SendNotification(n *fyne.Notification) {
...
+		if _, err := os.Stat(iconFilePath); os.IsNotExist(err) {
+			err := os.WriteFile(iconFilePath, a.Icon().Content(), 0600)
```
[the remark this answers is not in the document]
montovaneli: yes, it was for that reason that I used os.Stat to check if the icon already existed, because saving the icon every time seemed like bad design/performance. But the idea behind the cachedIconPath is superb. I will implement it that way.

app/app_windows.go
montovaneli: what do you think about using the rootConfigDir() instead of os.TempDir() to save the icon?

Jacalz: Good catch. Maybe name it "icon.png" to be consistent with what the linked code does on Linux?

montovaneli: I'm getting ahead of myself and thinking about a feature where the notification icon can be different from the app's default icon. Should we stick with 'notification-icon.png' or is it better to consider this in another PR? It's up to you :)

Jacalz: I'd say that we can be consistent with the old code and then implement it everywhere when we add support for using other icons for notifications :)

andydotxyz:
This looks excellent thanks so much. Can I just ask however what happens if there is no icon set? Is passing "" as the icon path OK?

montovaneli:
> This looks excellent thanks so much. Can I just ask however what happens if there is no icon set? Is passing "" as the icon path OK?


This is not a problem. The XML Handler simply ignores the image when it is not found or when it is an invalid path.
Testing, for example, by forcing iconFilePath = "":

```
func (a *fyneApp) SendNotification(n *fyne.Notification) {
	title := escapeNotificationString(n.Title)
	content := escapeNotificationString(n.Content)
	appID := a.UniqueID()
	iconFilePath := a.cachedNotificationIconPath()
	if appID == "" || strings.Index(appID, "missing-id") == 0 {
		appID = a.Metadata().Name
	}

	iconFilePath = "" // Empty path

	script := fmt.Sprintf(notificationTemplate, title, content, iconFilePath, appID)
	go runScript("notify", script)
}
```
image


And with iconFilePath = "myWrongPath/icon.png" we have the same result:

```
func (a *fyneApp) SendNotification(n *fyne.Notification) {
	title := escapeNotificationString(n.Title)
	content := escapeNotificationString(n.Content)
	appID := a.UniqueID()
	iconFilePath := a.cachedNotificationIconPath()
	if appID == "" || strings.Index(appID, "missing-id") == 0 {
		appID = a.Metadata().Name
	}

	iconFilePath = "myWrongPath/icon.png" // Wrong path

	script := fmt.Sprintf(notificationTemplate, title, content, iconFilePath, appID)
	go runScript("notify", script)
}
```

image

andydotxyz:
Oops I missed this before, please base your change on the `develop` branch not `master`.

montovaneli:
> Oops I missed this before, please base your change on the `develop` branch not `master`.

Oops, sorry. Is it ok now?

montovaneli:
> > Oops I missed this before, please base your change on the `develop` branch not `master`.
> 
> Oops, sorry. Is it ok now?

It seems not, sorry, I didn't realize I wasn't on develop. It might be easier to close this PR and I'll open another one pointing to the correct base.

Jacalz (changes requested):
Thanks for implementing the changes from https://github.com/fyne-io/fyne/blob/6b2c99e1a3a79e7463a3083350070fc0c583f412/app/app_xdg.go#L122. However, would you mind reusing the code to avoid having duplicate functions that do the same thing? You should  be able to move out the linked function (and any functions it calls) to another file that isn't inside a platform specific Go-file. That seems better than implementing a second function that does the same thing.

montovaneli:
> Thanks for implementing the changes from
> 
> https://github.com/fyne-io/fyne/blob/6b2c99e1a3a79e7463a3083350070fc0c583f412/app/app_xdg.go#L122
> 
> . However, would you mind reusing the code to avoid having duplicate functions that do the same thing? You should be able to move out the linked function (and any functions it calls) to another file that isn't inside a platform specific Go-file. That seems better than implementing a second function that does the same thing.

No problem, I was being cautious in modifying the least possible number of files. I'll close this PR and create a new one based on the develop branch.

PR #4765, opened 2024-04-04
Add support for the default icon in Windows notifications

### LINKED ISSUE #2592: Windows: App Icon in Notification

montovaneli:
### Description:

Fixes #2592
(now based on develop branch 🙂)

Adds support for the default icon in Windows notifications by changing the ToastTemplateType from ToastText02 to ToastImageAndText02 and adding the app's default icon. Also, since Windows ToastNotificationManager requests a path for the icon, we write the icon file in the temporary folder.

Example:

319294268-43faee05-ea36-43cd-a94b-2087d158b7ad

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/icon_cache_file.go
+++ b/app/icon_cache_file.go
@@ -1,0 +1,59 @@
+package app
+
+import (
+	"os"
+	"path/filepath"
+	"sync"
+
+	"fyne.io/fyne/v2"
+)
+
+var once sync.Once
+
+func (a *fyneApp) CachedIconPath() string {
+	if a.Icon() == nil {
+		return ""
+	}
+
+	dirPath := filepath.Join(rootCacheDir(), a.UniqueID())
+	filePath := filepath.Join(dirPath, "icon.png")
+	once.Do(func() {
+		err := a.saveIconToCache(dirPath, filePath)
+		if err != nil {
+			filePath = ""
+		}
+	})
+
+	return filePath
+}
+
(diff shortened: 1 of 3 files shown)


app/icon_cache_file.go
```diff
@@ -0,0 +1,59 @@
...
+	"os"
+	"path/filepath"
+	"sync"
+
+	"fyne.io/fyne/v2"
+)
+
+var once sync.Once
+
+func (a *fyneApp) CachedIconPath() string {
```
andydotxyz, line 13: I don't think this needs to be changed to a public function...

andydotxyz (changes requested):
Nice change thanks - but let's keep it private.

app/icon_cache_file.go
montovaneli: Done. Even I can't say why I changed it (?)

andydotxyz (approved):
Fantastic thanks

andydotxyz:
Congratulations and many thanks for your first Fyne contribution.

montovaneli:
First of many! It's been awesome using Fyne to build my apps. Thanks!

PR #4768, opened 2024-04-05
Fix memory leak in RichText when replacing segments

### LINKED ISSUE #4723: RichText leaks memory when replacing segments
### Checklist

### Describe the bug

Changing `RichText` content by invoking `RichText.ParseMarkdown` leads to linear growth of memory consumption. In the practical case I faced it took gigabytes of RAM in a couple of hours.

### How to reproduce

To reproduce — call `RichText.ParseMarkdown` many times with different content. Or just use the snippet I attached below.

### Screenshots

Memory growth (one line per second):
image

### Example code

```go
package main

import (
	"fmt"
	"runtime"
	"time"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
	"github.com/google/uuid"
)

func main() {
	a := app.New()
	w := a.NewWindow("Fyne's RichText Memory Leak")
	w.Resize(fyne.NewSize(128, 128/1.618))
	rt := widget.NewRichText()
	go func() {
		var m runtime.MemStats
		for {
			time.Sleep(time.Second)
			// Force GC just to make measurements a bit more precise.
			runtime.GC()
			// Report memory consumption.
			runtime.ReadMemStats(&m)
			fmt.Printf(
				"Sys: %v MiB\tHeapAlloc: %v MiB\tTotalAlloc: %v MiB\tNumGC: %v\n",
				m.Sys/1024/1024, m.HeapAlloc/1024/1024, m.TotalAlloc/1024/1024, m.NumGC,
			)
		}
	}()
	go func() {
		for {
			rt.ParseMarkdown(uuid.NewString())
		}
	}()
	w.SetContent(rt)
	w.ShowAndRun()
}
```

### Fyne version

2.4.4

### Go compiler version

1.22.1

macOS Sonoma 14.4

### Additional Information

_No response_

dweymouth:
### Description:
Clean out visualCache entries that are not currently needed on Refresh

Fixes #4723 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/richtext.go
+++ b/widget/richtext.go
@@ -226,6 +226,9 @@ func (t *RichText) cachedSegmentVisual(seg RichTextSegment, offset int) fyne.Can
 }
 
 func (t *RichText) cleanVisualCache() {
+	if len(t.visualCache) <= len(t.Segments) {
+		return
+	}
 	var deletingSegs []RichTextSegment
 	for seg1 := range t.visualCache {
 		found := false
--- a/widget/richtext.go
+++ b/widget/richtext.go
@@ -225,6 +225,26 @@ func (t *RichText) cachedSegmentVisual(seg RichTextSegment, offset int) fyne.Can
 	return vis
 }
 
+func (t *RichText) cleanVisualCache() {
+	var deletingSegs []RichTextSegment
+	for seg1 := range t.visualCache {
+		found := false
+		for _, seg2 := range t.Segments {
+			if seg1 == seg2 {
+				found = true
+				break
+			}
+		}
+		if !found {
+			// cached segment is not currently in t.Segments, clear it
+			deletingSegs = append(deletingSegs, seg1)
+		}
+	}
+	for _, seg := range deletingSegs {
+		delete(t.visualCache, seg)
+	}
+}
+
 // insertAt inserts the text at the specified position
 func (t *RichText) insertAt(pos int, runes string) {
 	index := 0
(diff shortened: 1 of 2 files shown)


widget/richtext.go
```diff
@@ -225,6 +225,26 @@ func (t *RichText) cachedSegmentVisual(seg RichTextSegment, offset int) fyne.Can
...
+		found := false
+		for _, seg2 := range t.Segments {
+			if seg1 == seg2 {
+				found = true
+				break
+			}
+		}
+		if !found {
+			// cached segment is not currently in t.Segments, clear it
+			deletingSegs = append(deletingSegs, seg1)
```
Jacalz, line 240: Does the map change order when we delete from ut? I wonder if we can just delete it directly instead of heap-allocating a slice on each refresh?

dweymouth: It does look like it's *probably safe* but the actual spec isn't super explicit about it: https://stackoverflow.com/questions/56048259/is-it-safe-to-remove-dictionary-keys-while-iterating-over-it

dweymouth: I added an early return condition so we will skip the allocation and iteration if we don't need to clean anything

widget/richtext.go
```diff
@@ -225,6 +225,29 @@ func (t *RichText) cachedSegmentVisual(seg RichTextSegment, offset int) fyne.Can
 	return vis
 }
 
+func (t *RichText) cleanVisualCache() {
+	if len(t.visualCache) <= len(t.Segments) {
```
andydotxyz, line 229: With this things may still leak once... is that OK?
(i.e. if you replace 1 or 2 or more with exactly the same number then it will expand to fit the new items, which will get removed later.
Probably a fair trade-off, just wanted to check

dweymouth: I think so, what's important is that it is cleared eventually and there's an upper bound on the cache size

andydotxyz (approved):
Thanks. I think the clarification on the early escape marks this good.

PR #4789, opened 2024-04-18
fix: close deskFile

testwill:
### Description:

close deskFile
Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


andydotxyz:
Thanks for the fix - but I wonder if perhaps the close should happen before the tar command runs?

Jacalz:
We really should be checking the error when creating that file as well.

Side-note: Should this not be in the fyne-io/tools repo now that we have moved the command line tool?

Jacalz:
If you could close this PR and open it against https://github.com/fyne-io/tools instead, that would be grand. Sorry for the inconvenience and thanks for your contribution :)

PR #4794, opened 2024-04-19
Double-click to select a file in FileOpen / FileSave dialogs

### LINKED ISSUE #4676: Double-click to select a file in FileOpen / FileSave dialogs
### Checklist

On Windows and Linux at the least, the common way to select a file from a system "select file" dialog is by double-clicking on the file name or icon. Fyne doesn't do this, which makes it counter-intuitive for people used to the native behaviour (a "why isn't the app working?" moment). Mobile environments have their own rules and the current behaviour is fine for them.

-

Enable confirming file selection in FileOpen / FileSave etc. dialogs by double-clicking on the file name or icon in the list, for desktop environments.

Dorbmon:
### Description:
Support Double-click to select a file in FileOpen / FileSave dialogs

Fixes #4676

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -555,7 +555,8 @@ func (f *fileDialog) setView(view ViewLayout) {
 			parent := id == 0 && len(dir.Path()) < len(f.dir.Path())
 			_, isDir := dir.(fyne.ListableURI)
 			o.(*fileDialogItem).setLocation(dir, isDir || parent, parent, id)
-			o.(*fileDialogItem).setChooseAndOpenCallBack(choose, f.open.OnTapped)
+			o.(*fileDialogItem).choose = choose
+			o.(*fileDialogItem).open = f.open.OnTapped
 		}
 	}
 	// Acutally, during the real interaction, the choose won't be called.
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -554,12 +554,13 @@ func (f *fileDialog) setView(view ViewLayout) {
 		if dir, ok := f.getDataItem(id); ok {
 			parent := id == 0 && len(dir.Path()) < len(f.dir.Path())
 			_, isDir := dir.(fyne.ListableURI)
-			o.(*fileDialogItem).setLocation(dir, isDir || parent, parent, id)
+			o.(*fileDialogItem).setLocation(dir, isDir || parent, parent)
 			o.(*fileDialogItem).choose = choose
+			o.(*fileDialogItem).id = id
 			o.(*fileDialogItem).open = f.open.OnTapped
 		}
 	}
-	// Acutally, during the real interaction, the choose won't be called.
+	// Actually, during the real interaction, the OnSelected won't be called.
 	// It will be called only when we directly calls container.select(i)
 	if f.view == GridView {
 		grid := widget.NewGridWrap(count, template, update)
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -558,14 +558,17 @@ func (f *fileDialog) setView(view ViewLayout) {
 			o.(*fileDialogItem).setChooseAndOpenCallBack(choose, f.open.OnTapped)
 		}
 	}
-
+	// Acutally, during the real interaction, the choose won't be called.
+	// It will be called only when we directly calls container.select(i)
 	if f.view == GridView {
 		grid := widget.NewGridWrap(count, template, update)
+		grid.OnSelected = choose
 		f.files = grid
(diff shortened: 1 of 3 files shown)


andydotxyz:
This is a lot more functionality into the fileItem to handle one new callback - can you talk through why this was necessary?

I would have thought that either you could add DoubleTapped to the item (it should not delay as it does not actually have a single tapped handler - that is in the parent).

Thanks for working on this, I just want to check this is the simplest approach.

Dorbmon:
> This is a lot more functionality into the fileItem to handle one new callback - can you talk through why this was necessary?
> 
> I would have thought that either you could add DoubleTapped to the item (it should not delay as it does not actually have a single tapped handler - that is in the parent).
> 
> Thanks for working on this, I just want to check this is the simplest approach.

Since fileDialogItem also needs to handle click events (which were previously handled by grid or list), the click event of the list seems to override the event of the item. Therefore, I moved the click event handling into fileDialogItem as well.

dialog/fileitem.go
```diff
@@ -55,6 +64,31 @@ func (i *fileDialogItem) setLocation(l fyne.URI, dir, up bool) {
...
+	}
+	now := time.Now()
+	if !i.dir && now.Sub(i.lastClick) < fyne.CurrentApp().Driver().DoubleTapDelay() && i.open != nil {
+		// It is a double click, so we ask the dialog to open
+		i.open()
+	}
+	i.lastClick = now
+}
+
+func (i *fileDialogItem) TouchDown(*mobile.TouchEvent) {}
```
dweymouth, line 79: I think you can still use Tappable instead of Mouseable/Touchable. If your object is Tappable and not DoubleTappable then there will be no delay and you can just do the same logic you already have here to treat it as a single tap to select or a double tap to open.

dweymouth (changes requested):

dialog/fileitem.go
Dorbmon: Yes. It is a more elegant way. I have switched to that.

dialog/fileitem.go
```diff
@@ -77,6 +96,11 @@ func (f *fileDialog) newFileItem(location fyne.URI, dir, up bool) *fileDialogIte
 	return item
 }
 
+func (i *fileDialogItem) setChooseAndOpenCallBack(choose func(id int), open func()) {
```
dweymouth, line 99: A minor style thing - I'm not sure it's worth creating a function for just this. Instead what you could do is where you invoke this function now, do a cast to (*fileDialogItem) and save it to a local variable, then you can simply access .choose and .open (and .setLocation).

dialog/file.go
```diff
@@ -547,26 +554,25 @@ func (f *fileDialog) setView(view ViewLayout) {
...
 		}
 	}
-	choose := func(id int) {
-		if file, ok := f.getDataItem(id); ok {
-			f.selectedID = id
-			f.setSelected(file, id)
-		}
-	}
-
+	// Acutally, during the real interaction, the choose won't be called.
```
dweymouth, line 562: I'm not sure what this comment is really communicating? Also there is a typo misspelling in "Actually"

dialog/fileitem.go
```diff
@@ -20,8 +21,13 @@ type fileDialogItem struct {
 	picker *fileDialog
 
 	name     string
+	id       int // id in the father container
```
dweymouth, line 24:
```suggestion
	id       int // id in the parent container
```

dialog/fileitem.go
```diff
@@ -39,7 +45,8 @@ func (i *fileDialogItem) CreateRenderer() fyne.WidgetRenderer {
 	}
 }
 
-func (i *fileDialogItem) setLocation(l fyne.URI, dir, up bool) {
+func (i *fileDialogItem) setLocation(l fyne.URI, dir, up bool, id int) {
```
dweymouth, line 48: Should this function be renamed since it's doing more than setting location? Like maybe just `update` or `updateForURI`?

dweymouth (changes requested):

dialog/file.go
Dorbmon: Yes, a typo. I mean The choose function won't be called through select because we override the tapped function.

dialog/fileitem.go
Dorbmon: Yes. I can use the same way as choose function.

dweymouth (approved):
LGTM now, but I think we should give @andydotxyz a chance to take another look

PR #4800, opened 2024-04-21
Fix timeout if the portal DBus is not present

andydotxyz:
Just update a cache when the value is found and refresh app

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -23,7 +23,12 @@ import (
 var currentVariant atomic.Value
 
 func defaultVariant() fyne.ThemeVariant {
-	return currentVariant.Load().(fyne.ThemeVariant)
+	val := currentVariant.Load()
+	if val == nil {
+		return nil
+	}
+
+	return val.(fyne.ThemeVariant)
 }
 
 func (a *fyneApp) OpenURL(url *url.URL) error {
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -25,7 +25,7 @@ var currentVariant atomic.Value
 func defaultVariant() fyne.ThemeVariant {
 	val := currentVariant.Load()
 	if val == nil {
-		return nil
+		return theme.VariantDark
 	}
 
 	return val.(fyne.ThemeVariant)
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -20,15 +20,10 @@ import (
 	"fyne.io/fyne/v2/theme"
 )
 
-var currentVariant atomic.Value
+var currentVariant atomic.Uint64
 
 func defaultVariant() fyne.ThemeVariant {
-	val := currentVariant.Load()
-	if val == nil {
-		return theme.VariantDark
-	}
-
-	return val.(fyne.ThemeVariant)
+	return fyne.ThemeVariant(currentVariant.Load())
 }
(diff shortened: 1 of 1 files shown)


app/app_xdg.go
```diff
@@ -20,8 +20,10 @@ import (
 	"fyne.io/fyne/v2/theme"
 )
 
+var currentVariant = theme.VariantDark
+
 func defaultVariant() fyne.ThemeVariant {
-	return findFreedestktopColorScheme()
+	return currentVariant
```
Jacalz, line 26: Hmm. I think this might be introducing a race condition given that it is modified in a different goroutine? Might be worth using something like an `atomic.Pointer` or a lock?

andydotxyz:
Looks like some Microsoft dependency has taken down the build CI setup :(

andydotxyz:
Whatever the problem was it is fixed upstream.
Just added a nil check before casting out the atomic.Value and it should be ready now.

Jacalz (approved):

app/app_xdg.go
```diff
@@ -20,8 +20,15 @@ import (
 	"fyne.io/fyne/v2/theme"
 )
 
+var currentVariant atomic.Value
```
Jacalz, line 23: Hmm. I actually have a better idea. Given that `fyne.ThemeVariant` is a uint, you can get cleaner and faster code by just using an `atomic.Uint64` and cast the value to a `fyne.ThemeVariant` where loading and storing.

app/app_xdg.go
```diff
@@ -20,8 +20,15 @@ import (
...
+
 func defaultVariant() fyne.ThemeVariant {
-	return findFreedestktopColorScheme()
+	val := currentVariant.Load()
+	if val == nil {
+		return theme.VariantDark
+	}
+
+	return val.(fyne.ThemeVariant)
```
Jacalz, line 31: With the suggestion above, this would just be something like this (default/zero value is zero, aka `theme.VariantDark`):
```suggestion
	return fyne.ThemeVariant(currentVariant.Load())
```

app/app_xdg.go
```diff
@@ -119,9 +126,15 @@ func rootConfigDir() string {
...
 
 func watchTheme() {
-	go portalSettings.OnSignalSettingChanged(func(changed portalSettings.Changed) {
-		if changed.Namespace == "org.freedesktop.appearance" && changed.Key == "color-scheme" {
-			fyne.CurrentApp().Settings().(*settings).setupTheme()
-		}
-	})
+	go func() {
+		// with portal this may not be immediate, so we update a cache instead
+		currentVariant.Store(findFreedesktopColorScheme())
```
Jacalz, line 131: Similar suggestion here:
```suggestion
		currentVariant.Store(fyne.ThemeVariant(findFreedesktopColorScheme()))
```

Jacalz (changes requested):
Sorry, I realised that there was a cleaner and faster way to do this :sweat_smile:

app/app_xdg.go
andydotxyz: Thanks, done

Jacalz (approved):
Perfect. Thanks

PR #4808, opened 2024-04-25
feat: support set nil for boundUntyped.Set()

### LINKED ISSUE #4807: binding.Untyped crashes when set to nil
### Checklist

### Describe the bug

When setting a `binding.Untyped` to nil, it crashes.  I expect this case to work - since the type I'm using is a pointer and could be nil.  The code seems to handle nil properly in several code paths, but not for `.Set()`

### How to reproduce

Set binding.Untyped to something other than nil, then set it to nil.

### Screenshots

_No response_

### Example code

```go
 a := binding.NewUntyped()
 a.Set(nil)  // this is fine since it's already set to nil
 b := 1
 a.Set(&b)
 a.Set(nil) // crashes
```

### Fyne version

2.4.4

### Go compiler version

1.19.3

Windows 10

### Additional Information

_No response_

Dorbmon:
### Description:
We should support setting nil for Untyped Binding. For the original code, when we pass nil for the Set function, it will call `b.val.Set(reflect.ValueOf(nil))` which will panic due to the lack of type information of nil. 

Fixes #4807

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

### DIFF
--- a/data/binding/binding.go
+++ b/data/binding/binding.go
@@ -117,8 +117,12 @@ func (b *boundUntyped) Set(val any) error {
 	if b.val.Interface() == val {
 		return nil
 	}
-
-	b.val.Set(reflect.ValueOf(val))
+	if val == nil {
+		zeroValue := reflect.Zero(b.val.Type())
+		b.val.Set(zeroValue)
+	} else {
+		b.val.Set(reflect.ValueOf(val))
+	}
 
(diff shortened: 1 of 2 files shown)


andydotxyz (approved):
Excellent thanks for providing this improvement

PR #4811, opened 2024-04-27
Software Driver POC

chabad360:
### Description:
This POC adds a new Driver that allows one to include their own event source and framebuffer (through an `image.Image`). It's based almost entirely on a combination of the [`testDriver`](https://github.com/fyne-io/fyne/blob/master/test/testdriver.go) and some choice bits from the [`mobileDriver`](https://github.com/fyne-io/fyne/blob/master/internal/driver/mobile/driver.go) (mainly the basic event loop and touch handling).

This is an incredibly rough draft, I'm putting it up so that it can be discussed. Moreover, the actual API surface and basic design hasn't really been discussed yet, so this couldn't be merged either way. But most importantly, it does work, which is a good start.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


app/app_software.go
```diff
@@ -1,9 +1,12 @@
-//go:build ci
+//go:build ci || software
 
 package app
 
 import (
+	"image"
+
 	"fyne.io/fyne/v2"
+	software2 "fyne.io/fyne/v2/driver/software"
```
andydotxyz, line 9: I'd probably switch the naming here as the first of the two is the one called "software" in the public API

app/app_software.go
```diff
@@ -13,3 +16,9 @@ import (
 func NewWithID(id string) fyne.App {
 	return newAppWithDriver(test.NewDriverWithPainter(software.NewPainter()), id)
 }
+
+// NewWithSoftwareDriver returns a new app instance using the Software (custom) driver.
+// The ID string should be globally unique to this app.
+func NewWithSoftwareDriver(id string, painter func(image.Image), events chan any) fyne.App {
```
andydotxyz, line 22: Won't the painter always be the one we have in the software package?

chabad360: TBH I didn't really look at the logic for how that worked, but I needed a way to add some extra params and wasn't sure a good way to do that with the existing interfaces.

chabad360:
> but I would hesitate from pulling so much internal API code into a public driver package.

Yea, you're right about that. It can pretty easily move into an internal package, I just didn't put it there. I'll move things around once the code is in a more complete state.

internal/painter/software/draw.go
```diff
@@ -19,16 +21,22 @@ type gradient interface {
...
+	t := time.Now()
 	pad := painter.VectorPad(circle)
 	scaledWidth := scale.ToScreenCoordinate(c, circle.Size().Width+pad*2)
 	scaledHeight := scale.ToScreenCoordinate(c, circle.Size().Height+pad*2)
 	scaledX, scaledY := scale.ToScreenCoordinate(c, pos.X-pad), scale.ToScreenCoordinate(c, pos.Y-pad)
 	bounds := clip.Intersect(image.Rect(scaledX, scaledY, scaledX+scaledWidth, scaledY+scaledHeight))
 
-	raw := painter.DrawCircle(circle, pad, func(in float32) float32 {
-		return float32(math.Round(float64(in) * float64(c.Scale())))
+	raw, err := p.getTexture(circle, func(object fyne.CanvasObject) Texture {
```
chabad360, line 32: This is a problem for tests (wrong type for `TextureType`), and I'm not quite sure what's a good way around it...

Maybe we could make the painter functions internal and add a wrapper file (i.e. `draw_extern_software.go`, `draw_extern.go`) that handles storing in the cache (or does nothing)? being that it's only a few lines, it should be eligible for inlining, so the wrapper should be functionally free.

dialog/form_test.go
```diff
@@ -90,7 +91,7 @@ func TestFormDialog_Hints(t *testing.T) {
 	test.NewApp()
 	defer test.NewApp()
 	test.ApplyTheme(t, internalTest.LightTheme(theme.DefaultTheme()))
-	w := test.NewWindow(nil)
+	w := software.NewWindow(nil)
 	w.SetFullScreen(true)
```
chabad360, line 96: this test is broken and I have no clue how or why. pretty much every other test is fine, and this one the failure is visually identical. So I've no clue what's going wrong here...

lostdusty:
Cool and good work there, friend. Keep it up :)

andydotxyz:
Sorry I did not get back to this sooner. 
I see you moved lots of the test code to use the software package - but really it should have stayed with the `test` usage.

Also there have been lots of test code changes to use new API as we headed to 2.5, so some of the conflicts are extensive.

I would be tempted to check out a new branch and cherry pick the work onto that - broadly skipping the commits that changed all the test code to use your new package.

I hope that helps.

chabad360:
> Sorry I did not get back to this sooner.
 
Np.

> I see you moved lots of the test code to use the software package - but really it should have stayed with the test usage.

Yea, fair, but I'm curious if you can give me any particular reasons why. I wonder if maybe there's a happy middle?

> Also there have been lots of test code changes to use new API as we headed to 2.5, so some of the conflicts are extensive.

I see that.

> I would be tempted to check out a new branch and cherry pick the work onto that

 - broadly skipping the commits that changed all the test code to use your new package.

I hope that helps.

redawl:
I think this is unneeded, now that https://github.com/fyne-io/fyne/pull/5821 has landed. Any objections to closing?

andydotxyz:
That's a good point, thanks so much

andydotxyz:
I appreciate your hard work on this @chabad360 - the implementation that delivers in v2.7.0 should support what you need to do. With the model we chose external packages can provide a variety of driver implementations, and you can see some in the fyne-x package.

chabad360:
If I get around to touching this project again, I will have a look. Thanks for remembering this tho!

andydotxyz:
OK cool, thanks for understanding.

You can see a demo project using it at https://github.com/ApptrixAI/fyne-noos-demo

Issue #1031, opened 2020-05-23
Layout with custom padding


ababo:
Currently I implement a canvas object for rich text which will use a very limited subset of HTML for styling. And I struggle with the horizontal box layout, since it uses a theme padding, but I don't need any padding. Is there a way to reuse existing layout with no padding without changing the theme globally? Thanks in advance.

andydotxyz:
The default layouts use theme padding, just like the standard widgets use theme colours.
You can easily pass a custom layout into NewContainerWithLayout, https://fyne.io/develop/tutorials/custom-layout .

ababo:
So I had to replicate the box layout code with a couple changed lines to support custom (zero) padding. It would be nice to support NewVBoxWithPadding() and NewHBoxWithPadding() functions for the existing box layout having really minimal changes. If I'll create a PR would you accept that?

andydotxyz:
The box widgets are designed for laying out other widgets in a consistent manner.
Your work sounds closer to a custom widget - that would normally provide it's own layout anyhow.

ababo:
Oh, I meant layouts, not the widgets, i.e. `NewVBoxLayoutWithPadding()` and `NewHBoxLayoutWithPadding()`.

andydotxyz:
I think that it's probably still the same case - you are working on what should probably be a widget so adding customisation to standard components to support that doesn't make sense to the core API. At some point we should complete the more advanced layout options (like #4) that make "anything possible" but the simple layout API should probably always use the theme like widgets

dweymouth:
Added on develop for Fyne 2.5

PR #4813, opened 2024-04-27
feat: Allow setting custom Layout paddings

adamantike:
### Description:

Introduce a `LayoutOption` concept, and allow setting custom paddings for a Layout. The `BaseLayout` struct implements the default paddings retrieval, to reduce the amount of changes required by project maintainers to comply with the new interface methods.

This change uses the Option pattern in case other customizations can be added to Layouts in the future, without changing the public API.

Main motivation for this change has been Supersonic needing to copy the entire Vbox/Hbox implementation, just to add custom paddings [1]. Also related to feature request #1031.

[1] https://github.com/dweymouth/supersonic/blob/02d4082f3dae4d1d54a4384489785a5984a7a7de/ui/layouts/hboxcustompadding.go

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/layout.go
+++ b/layout.go
@@ -8,4 +8,10 @@ type Layout interface {
 	// MinSize calculates the smallest size that will fit the listed
 	// CanvasObjects using this Layout algorithm.
 	MinSize(objects []CanvasObject) Size
+	// GetPaddings returns the top, bottom, left and right paddings used
+	// by this layout.
+	GetPaddings() (float32, float32, float32, float32)
+	// SetPaddingFn sets the function that will be called to get the
+	// padding values for this layout.
+	SetPaddingFn(fn func() (float32, float32, float32, float32))
 }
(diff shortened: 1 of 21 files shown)


layout.go
```diff
@@ -8,4 +8,10 @@ type Layout interface {
 	// MinSize calculates the smallest size that will fit the listed
 	// CanvasObjects using this Layout algorithm.
 	MinSize(objects []CanvasObject) Size
+	// GetPaddings returns the top, bottom, left and right paddings used
+	// by this layout.
+	GetPaddings() (float32, float32, float32, float32)
```
andydotxyz, line 13: Just FYI such an addition to a published interface is a breaking change.

Anyhow this topic has been discussed a lot and you can see replacement PRs coming in that take a less-sweeping change that can be adopted more easily.
Hopefully this resolves the use-case we missed?

andydotxyz (changes requested):
I think this may want to close in preference to #4823 and #4825 along with others I expect.

adamantike:
This is indeed solved by the linked PRs, thank you!

PR #4820, opened 2024-04-30
RFC: Introduce a new Edge container

Jacalz:
### Description:

This new container is a replacement for container.NewBorder but without the footguns. There is a centre item instead of a slice of objects that act like a Stack container. The centre item can optionally be nil also.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/container/layouts.go
+++ b/container/layouts.go
@@ -49,6 +49,34 @@ func NewCenter(objects ...fyne.CanvasObject) *fyne.Container {
 	return New(layout.NewCenterLayout(), objects...)
 }
 
+// NewEdge creates a new container with the specified objects and using the edge layout.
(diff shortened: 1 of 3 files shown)


Jacalz:
This uses the `layout.NewEdge` name which stutters less than `layout.NewEdgeLayout` but is less similar to the other layouts. Should we rename (I.e. create new functions and deprecate the old ones) with better naming?

andydotxyz:
Hmm, if we think that having no 5th parameter for border we could potentially move it from 4 to 5 parameters and the optional becomes 6 onwards in case people did use the ability to have more than a single item in the centre.

I'm not totally sure it's worth a new layout (and new container?) to fix this one issue.

Jacalz:
yeah, probably not

Issue #4821, opened 2024-04-30
Form layout allocates a new 2d slice on each MinSize and Layout call


Jacalz:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

I noticed that the form layout allocates a two-dimensional slice on each call to Layout and MinSize:
fyne-io/fyne/blob/d8e9fe02c9bb9e4b4c42778f5cae8ffcc43f36e2/layout/formlayout.go#L38

This was likely a way to write simpler code some time but it does mean that (especially noticeable for large forms) the layout can allocate a lot of garbage each second when layout and minsize has to be computed often.

### How to reproduce

Look at memory profiles or use `b.ReportAllocs()` in benchmarks.

### Fyne version

v2.4.5

### Go compiler version

1.22.1

### Operating system and version

Fedora Silverblue 40

### Additional Information

I think the best option is to do a large refactor of the layout and remove the allocations from the fast path.
The code is also, in my opinion at least, quite convoluted and could do with a cleanup to make it more readable.

dweymouth:
Hmm, I'm actually not sure there's a way to avoid allocation without the possibly worse tradeoff of invoking MinSize() twice per cell instead of once per cell. (Remembering that MinSize can be user code for a custom widget that may not be efficient). It could possibly be turned into a 1D slice of row heights that is allocated. And might be a candidate for using a widget.Pool since there are instances where MinSize and Layout will be called very frequently (eg user resizing a window or split container)

Jacalz:
Using a pool would likely be a big improvement  given how often it can be called. I'd need to look at the code more but I feel like there should be a better way to refactor the algorithm to not need that table. Maybe use some modulo operation into the object slice?

Jacalz:
It seems to me (and initial prints seem to show the same thing) that the table is just a slice of the two exact same widths with only the height potentially differing.

Jacalz:
@dweymouth Did you intend to land the stack-allocated change code that you talked about?

dweymouth:
Not for 2.5.0 at least, maybe I'll get around to it at some point but I am busy enough trying to push Supersonic toward a new release and closing out my draft/open PRs for Fyne 2.5

Jacalz:
I understand. Would you mind opening a ticket to track that so we don't forget?

PR #4822, opened 2024-05-01
Try to decrease allocations in form layout

Jacalz:
### Description:

This PR does some refactoring to remove unnecessary code and allocate only as much as we need.
The code is cleaner, tidier and does allocate 75% less memory but we are still allocating in the fast path so this is not a final solution but a step in the right direction. Performance wise, it is only ever so slightly faster. 

```
BenchmarkFormMinSizeBefore-8   	 6294230	       188.3 ns/op	      32 B/op	       1 allocs/op
BenchmarkFormMinSizeAfter-8   	 6604521	       175.1 ns/op	       8 B/op	       1 allocs/op

BenchmarkFormLayoutBefore-8   	 3236059	       363.9 ns/op	      32 B/op	       1 allocs/op
BenchmarkFormLayoutAfter-8   	 3091474	       350.5 ns/op	       8 B/op	       1 allocs/op
```

For #4821 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/layout/formlayout.go
+++ b/layout/formlayout.go
@@ -28,22 +28,17 @@ func (f *formLayout) countRows(objects []fyne.CanvasObject) int {
 	return count
 }
 
-type cellHeights struct {
-	labelHeight   float32
-	contentHeight float32
-}
-
 // tableCellsSize defines the size for all the cells of the form table.
 // The height of each row will be set as the max value between the label and content cell heights.
 // The width of the label column will be set as the max width value between all the label cells.
 // The width of the content column will be set as the max width value between all the content cells
 // or the remaining space of the bounding containerWidth, if it is larger.
-func (f *formLayout) tableCellsSize(objects []fyne.CanvasObject, containerWidth float32) (float32, float32, []cellHeights) {
+func (f *formLayout) tableCellsSize(objects []fyne.CanvasObject, containerWidth float32) (float32, float32, []float32) {
 	rows := f.countRows(objects)
-	table := make([]cellHeights, rows)
+	heights := make([]float32, rows)
 
 	if (len(objects))%formLayoutCols != 0 {
-		return 0, 0, table
+		return 0, 0, heights
 	}
 
 	padding := theme.Padding()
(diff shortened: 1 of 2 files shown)


layout/formlayout.go
```diff
@@ -33,15 +33,14 @@ func (f *formLayout) countRows(objects []fyne.CanvasObject) int {
 // The width of the label column will be set as the max width value between all the label cells.
```
dweymouth, line 33: I think the comment for this function should be updated to make it clear that it now returns, in order, the label col width, the content col width, and the row heights. Maybe it should be renamed too like `calculateTableSizes`?

layout/formlayout.go
```diff
@@ -126,23 +120,19 @@ func (f *formLayout) Layout(objects []fyne.CanvasObject, size fyne.Size) {
...
 
-	added := false
-	minSize.Width = table[0][0].Width + table[0][1].Width + padding
-	for row := 0; row < len(table); row++ {
-		minSize.Height += table[row][0].Height
-		if added {
-			minSize.Height += padding
-		}
-		added = true
+	for row := 0; row < len(heights); row++ {
```
dweymouth, line 134: Maybe a `for _, height := range heights` would be cleaner?

layout/formlayout.go
```diff
@@ -126,23 +120,19 @@ func (f *formLayout) Layout(objects []fyne.CanvasObject, size fyne.Size) {
...
+	}
 
 	padding := theme.Padding()
-	minSize := fyne.NewSize(0, 0)
-
-	if len(table) == 0 {
-		return minSize
+	minSize := fyne.Size{
+		Width:  labelWidth + contentWidth + padding,
+		Height: padding * float32(len(heights)-1),
```
dweymouth, line 131: nit: could just inline `theme.Padding()` call here since it's only used once

layout/formlayout.go
```diff
@@ -126,23 +120,19 @@ func (f *formLayout) Layout(objects []fyne.CanvasObject, size fyne.Size) {
 // For a FormLayout this is the width of the widest label and content items and the height is
 // the sum of all column children combined with padding between each.
 func (f *formLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
-	table := f.tableCellsSize(objects, 0)
+	labelWidth, contentWidth, heights := f.tableCellsSize(objects, 0)
+	if len(heights) == 0 {
```
dweymouth, line 124: Nit: `len(heights)` is invoked 3 times in this func. Compiler is probably smart enough to inline it but it wouldn't hurt to cache it in a variable

dweymouth (changes requested):
Just some style comments but looks good!

layout/formlayout.go
Jacalz: I tried that as well but look on the line above and you will see that that isn't used once.

layout/formlayout.go
Jacalz: It is just a field on the slice struct. It basically has zero cost and the compiler should be able to optimise it. Coaching it might honestly be slower and prevent the compiler from optimising it.

dweymouth (approved):

PR #4823, opened 2024-05-01
Add a custom padded layout

dweymouth:
### Description:
This PR seeks to add a CustomPaddedLayout that behaves like PaddedLayout except allowing for a custom amount of padding applied to each side rather than the theme padding value.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/layout/custompaddedlayout.go
+++ b/layout/custompaddedlayout.go
@@ -9,6 +9,8 @@ var _ fyne.Layout = (*CustomPaddedLayout)(nil)
 
 // CustomPaddedLayout is a layout similar to PaddedLayout, but uses
 // custom values for padding on each side, rather than the theme padding value.
+//
+// Since: 2.5
 type CustomPaddedLayout struct {
 	TopPadding    float32
 	BottomPadding float32
--- a/layout/custompaddedlayout.go
+++ b/layout/custompaddedlayout.go
@@ -1,0 +1,72 @@
+package layout
+
+import (
+	"fyne.io/fyne/v2"
+)
+
+// Declare conformity with Layout interface
+var _ fyne.Layout = (*CustomPaddedLayout)(nil)
+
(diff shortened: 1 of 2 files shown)


Jacalz (commented):
I have nothing against this idea. Sounds like a good solution to the custom padding problem you have discussed earlier.

dweymouth:
> I have nothing against this idea. Sounds like a good solution to the custom padding problem you have discussed earlier.

This is half the solution. The other would be to add CustomPaddedHBox and VBox layouts which space the items using a custom padding value instead of theme.Padding

Jacalz (approved):
Looks good to me but I'd appreciate if Andy have go ahead before you merge (given that it is a slight stray from the other layouts) 👍

layout/custompaddedlayout.go
```diff
@@ -0,0 +1,74 @@
...
+		LeftPadding:   padLeft,
+		RightPadding:  padRight,
+	}
+}
+
+// NewSquareCustomPaddedLayout creates a new CustomPaddedLayout instance
+// with the same custom padding applied to all four sides.
+//
+// Since: 2.5
+func NewSquareCustomPaddedLayout(pad float32) fyne.Layout {
```
andydotxyz, line 38: Hmm, not sure about the naming here as we have used "SquareSize" to indicate equal width and height.
Clearly this is not creating square content. Is there a better name?

andydotxyz (changes requested):
Not sure about the shorthand naming.
Could land without it and think more - or come up with a clearer name in this PR...

layout/custompaddedlayout.go
dweymouth: Nothing great is coming to mind (NewEqualCustomPaddedLayout maybe, but it's a bit awkward). So I'll remove this shorthand for now so we can land it

andydotxyz (approved):
Thanks

PR #4825, opened 2024-05-02
Add custom padded box layouts

### LINKED ISSUE #1031: Layout with custom padding

dweymouth:
### Description:
Add custom padded HBox and VBox layouts that use a specified amount of padding between objects instead of theme.Padding

Fixes #1031 (orig. closed as won't fix)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/layout/boxlayout.go
+++ b/layout/boxlayout.go
@@ -10,7 +10,9 @@ import (
 // at their vertical MinSize. Use a different layout if the objects are intended
 // to be larger then their vertical MinSize.
 func NewVBoxLayout() fyne.Layout {
-	return vBoxLayout{}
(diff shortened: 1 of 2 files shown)


andydotxyz (approved):
Thanks. I had thought the `CustomPadded` was too long and `Custom` worked better, but looking at it seems to be more clear, thanks.

PR #4827, opened 2024-05-02
Add a cache for MinSize in BaseWidget

Jacalz:
### Description:

This attempts to cache MinSize values of widgets to provide for faster lookups by avoiding recalculations when nothing has changed. I think it feels quite a bit faster when just using fyne_demo. Placebo?  

Supersedes #4681

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/richtext.go
+++ b/widget/richtext.go
@@ -45,7 +45,6 @@ type RichText struct {
 
 	visualCache map[RichTextSegment][]fyne.CanvasObject
 	cacheLock   sync.Mutex
-	minCache    fyne.Size
 }
 
 // NewRichText returns a new RichText widget that renders the given text and segments.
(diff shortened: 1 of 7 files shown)


widget/richtext.go
```diff
@@ -89,19 +88,18 @@ func (t *RichText) CreateRenderer() fyne.WidgetRenderer {
 // MinSize calculates the minimum size of a rich text widget.
 // This is based on the contained text with a standard amount of padding added.
 func (t *RichText) MinSize() fyne.Size {
-	// we don't return the minCache here, as any internal segments could have caused it to change...
 	t.ExtendBaseWidget(t)
 
-	min := t.BaseWidget.MinSize()
-	t.minCache = min
-	return min
+	// We don't return the cached value as any internal segments could have caused it to change...
```
Jacalz, line 93: We should likely fix this as it kind of goes against the idea of refreshing a widget when it changes state. SOunds like a bug but it isn't new behaviour to this PR so I'll let it slide for now.

toaster:
I thought, that I had added some sort of minSize cache back in the days when I worked on the walkers, didn’t I?

Jacalz:
There appears to be a cache in the renderer cache but that is only used to tell us when the entire min size of the window needs to be recalculated as far as I can tell. The idea of this is to speed up lookups by not needing to look up the renderer and ask the widget to recompute it.

andydotxyz:
> There appears to be a cache in the renderer cache but that is only used to tell us when the entire min size of the window needs to be recalculated as far as I can tell.

There is a tree of nodes mapping the minsize for each element isn't there?

PR #4834, opened 2024-05-09
Update golang.org/x packages to fix x/net issue

Jacalz:
### Description:

Not something that we directly are vulnerable to in Fyne but GitHub keeps warning downstream projects so let's be on the safe side.

Fixes https://github.com/fyne-io/fyne/security/dependabot/8

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
Oh. This has already been fixed but the security advisory is still open.

PR #4835, opened 2024-05-09
Improve performance of content size calculation in List

### LINKED ISSUE #3427: List/Table size iteration could be faster
### Checklist

### Describe the bug

As noted during development there is a combination of iteration and map indexing that could be streamlined.
This applies to old table code and new list code.

@Bluebugs suggestion was:
```
height := 0
total := 0
for idx, itemHeight := range l.list.itemHeights {
   if idx < items {
       total++
       height += itemHeight
   }
}
height += (items - total) * templateHeight
```

Using some benchmark code on #3408 we should check this is better for majority of usage and update code accordingly.

### How to reproduce

See PR #3408

### Screenshots

_No response_

### Example code

as above

### Fyne version

v2.3.0

### Go compiler version

1.18.4

### Operating system

Other (specify below)

### Operating system version

all

### Additional Information

_No response_

dweymouth:
### Description:
Makes calculating the content size in List O(n) in the number of custom-height items rather than the number of total items. Unsurprisingly, a huge performance improvement for long lists with only a few custom item heights (as is a common use case).

Fixes #3427 (for list)

### Checklist:

- [ ] Tests included. <- already covered by existing tests
- [x] Tests all pass.

Benchmark (before):

`goos: darwin
goarch: arm64
pkg: fyne.io/fyne/v2/widget`

`BenchmarkContentMinSize-8   	     198	   6114649 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	fyne.io/fyne/v2/widget	2.062s`

Benchmark (after):

`goos: darwin
goarch: arm64
pkg: fyne.io/fyne/v2/widget`

`BenchmarkContentMinSize-8   	19891040	        51.04 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	fyne.io/fyne/v2/widget	1.565s`

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -365,15 +365,15 @@ func (l *List) contentMinSize() fyne.Size {
 	}
 
 	height := float32(0)
+	totalCustom := 0
 	templateHeight := l.itemMin.Height
-	for item := 0; item < items; item++ {
-		itemHeight, ok := l.itemHeights[item]
-		if ok {
+	for id, itemHeight := range l.itemHeights {
+		if id < items {
+			totalCustom++
 			height += itemHeight
-		} else {
-			height += templateHeight
 		}
 	}
+	height += float32(items-totalCustom) * templateHeight
 
 	return fyne.NewSize(l.itemMin.Width, height+separatorThickness*float32(items-1))
 }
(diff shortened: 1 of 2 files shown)


Jacalz (changes requested):
Similar to https://github.com/fyne-io/fyne/pull/4841, I think we should add more benchmarks to List so these optimizations are easier to reason about. Apart from that, the change seems like a nice one :)

Jacalz (approved):
Thanks. That makes it easier to reason about

PR #4841, opened 2024-05-11
List optimization - handle SetItemHeight with default height as a special case

dweymouth:
### Description:
Handles the case of SetItemHeight(h) where h is equal to the default item height specially. Instead of storing it in the itemHeights map, delete the entry in the map if it exists for the given item, since list operations are more efficient with fewer entries in the itemHeights map.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -139,14 +139,22 @@ func (l *List) RefreshItem(id ListItemID) {
 func (l *List) SetItemHeight(id ListItemID, height float32) {
 	l.propertyLock.Lock()
 
-	if l.itemHeights == nil {
-		l.itemHeights = make(map[ListItemID]float32)
+	var refresh bool
+	if height == l.itemMin.Height {
+		// special case - unset custom item height for this item
+		if l.itemHeights != nil {
+			_, refresh = l.itemHeights[id]
+			delete(l.itemHeights, id)
+		}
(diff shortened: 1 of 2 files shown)


Jacalz (changes requested):
Thanks. This seems very sensible but I think we might want a benchmark to back this up?

dweymouth:
What would we want to benchmark in this case? Setup by setting a custom height for every row and then unsetting it, then benchmarking how long contentMinSize takes?

Jacalz:
I don't really know but I just find it hard to grasp or understand why this is an obvious performance improvement and without any benchmark to back it up, I have a hard time motivating an approval.

dweymouth:
It's an improvement since list calculations are more expensive with a lot of custom item heights (especially now that https://github.com/fyne-io/fyne/pull/4835 is in making size O(n) in custom heights rather than O(n) in total items). So I guess the appropriate benchmark will be to set and unset a bunch of item heights in setup, and then benchmark that contentMinSize is faster with these changes. I'll look into adding one

Jacalz:
Just a friendly reminder to add a benchmark. Would be nice to have this in 2.5 :)

Jacalz:
It would be good to get this landed :)

PR #4846, opened 2024-05-18
Various improvements to the progress bar renderer

Jacalz:
### Description:

This rewrites part of the renderer for the progress bar to be more efficient and smarter.
The following changes have been implemented:
- Refactor MinSize to use less duplicated code.
- Keep ratio calculations in float64 to avoid potential rounding errors.
- Don't allocate and set text when doing a layout (i.e. perform part of a refresh). Keep that path fast by using stored ratio.
- Don't render the background under the progress indicator when at 100% progress.
- Keep the progress indicator hidden until progress goes above 0%.

Other related changes:
- The constructor now extends the base widget correctly instead of going through the cached renderer and then forcing a layout step before being rendered.

### Benchmarks:

This change makes the layout call roughly 3.8x faster. If a complex text formatter is set, the gain might be bigger.
```
goos: linux
goarch: amd64
pkg: fyne.io/fyne/v2/widget
BenchmarkProgressBarLayoutBefore-8   	29877590	        38.35 ns/op	       2 B/op	       1 allocs/op
BenchmarkProgressBarLayoutAfter-8   	100000000	        10.18 ns/op	       0 B/op	       0 allocs/op
```

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/progressbar.go
+++ b/widget/progressbar.go
@@ -17,6 +17,7 @@ type progressRenderer struct {
 	widget.BaseRenderer
 	background, bar canvas.Rectangle
 	label           canvas.Text
+	ratio           float32
 	progress        *ProgressBar
 }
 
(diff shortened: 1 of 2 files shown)


dweymouth:
Should this have a benchmark?

Jacalz:
I think it falls on the edge but it is probably wise to add one for the layout call to make sure we don't regress in allocating text in that path

dweymouth (approved):

PR #4847, opened 2024-05-18
Don't create the renderer in the constructor for infinite progressbar

Jacalz:
### Description:

This fixes the infinite progress bar code to not create the renderer inside the constructor. The old behaviour was not only a potential performance and memory leak and just seems generally incorrect from a best practices standpoint.

The tests had to be modified to actually render the object so this is a change in behaviour but I would not call it breaking given that I consider the old behaviour a incorrect and a bug.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


dweymouth (approved):

Jacalz:
Sorry, I realised that I needed to close the test windows to avoid a memory leak.

dweymouth (approved):

PR #4848, opened 2024-05-18
Avoid internal packages when starting and stopping infinite progress bar

Jacalz:
### Description:

This refactors the code to use a better mechanism for triggering start and stop in the renderer. The code now used Refresh() to indicate a state change to the renderer; this is a lot more consistent with other widgets.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/progressbarinfinite.go
+++ b/widget/progressbarinfinite.go
@@ -146,16 +146,30 @@ func (p *ProgressBarInfinite) Hide() {
 
 // Start the infinite progress bar animation
 func (p *ProgressBarInfinite) Start() {
-	p.SetFieldsAndRefresh(func() {
-		p.running = true
-	})
+	p.propertyLock.Lock()
+	if p.running {
+		p.propertyLock.Unlock()
+		return
+	}
+
+	p.running = true
+	p.propertyLock.Unlock()
+
+	p.BaseWidget.Refresh()
 }
 
 // Stop the infinite progress bar animation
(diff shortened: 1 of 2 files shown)


widget/progressbarinfinite.go
```diff
@@ -111,49 +107,62 @@ func (p *infProgressRenderer) stop() {
...
+	p.running = false
+	p.propertyLock.Unlock()
+
 	p.BaseWidget.Hide()
 }
 
 // Start the infinite progress bar animation
 func (p *ProgressBarInfinite) Start() {
-	cache.Renderer(p).(*infProgressRenderer).start()
+	p.SetFieldsAndRefresh(func() {
```
dweymouth, line 149: as discussed in Slack, let's just use the basewidget internal lock directly, instead of SetFieldsAndRefresh, in case we decide to remove or deprecate it

Jacalz: Yup, good point. That did also made make me realise that we don't need to refresh if we are already running or already stopped :)

Jacalz: That was easier to implement without using SetFieldsAndRefresh.

Jacalz:
Requesting a review from you Andrew, because this change can also be applied to the Activity widget later on if this change looks like a good one.

andydotxyz (approved):
Nice, this is how it should have been all along :)

PR #4849, opened 2024-05-18
Introduce test.NewTempWindow() to avoid memory leak in test windows

Jacalz:
### Description:

I noticed that we had a quite substantial memory leak in our test suite due to various tests forgetting to call .Close() on the test window that was created. This led to larger memory usage than what would be ideal. I fixed all the issues I could find and introduced a `test.NewTempWindow()` function to create a window that automatically closes at the end of the test.

Any places where we forgot to call Close() now uses the new function. Old uses that close the window have been left as is. Should we deprecate the old function to point people towards this safer new function?

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
I don't think we should deprecate test.NewWindow as it is a legitimate thing to do.

test/testwindow.go
```diff
@@ -18,6 +20,17 @@ type testWindow struct {
 	menu      *fyne.MainMenu
 }
 
+// NewTempWindow creates and registers a new window for test purposes.
+// This window will get removed automatically once the running test ends.
+//
+// Since: 2.5
+func NewTempWindow(t testing.TB, content fyne.CanvasObject) fyne.Window {
+	window := fyne.CurrentApp().NewWindow("")
```
andydotxyz, line 28: Shouldn't this be test.NewWindow?

widget/entry_test.go
```diff
@@ -2142,11 +2112,6 @@ func setupSelection(t *testing.T, reverse bool) (*widget.Entry, fyne.Window) {
 	return e, window
 }
 
-func teardownImageTest(w fyne.Window) {
-	w.Close()
-	test.NewApp()
```
andydotxyz, line 2147: This line is no longer called from many tests. Is that ok?

Jacalz: Yes, that is expected. I wired up `setupImageTest` and `setupPasswordTest` to both use `NewTempWindow` and create a new application when cleaning up after the test.

andydotxyz (approved):
Nice improvement thanks

Jacalz:
Thanks. Had to fix a merge conflict. Needs a re-approval, I'm afraid

andydotxyz (dismissed):

Jacalz:
Of course there were merce conflicts again. Sigh 😮‍💨

andydotxyz (approved):

PR #4850, opened 2024-05-18
Add test.NewTempApp() to tear down app at the end of the test

Jacalz:
### Description:

Similar to https://github.com/fyne-io/fyne/pull/4849 but more about convenience than memory leaks.

There are also a few cases where we did not create a new app but only created a new one afterwards. 
This PR should get us one step close towards being able to run tests in random order and hopefully get rid of strange order constraints when running the tests.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz (dismissed):
Thanks, this does read better than the defer and app creation pair

Jacalz:
Thanks. Had to fix a merge conflict. Needs a re-review, I'm afraid

andydotxyz (approved):

PR #4851, opened 2024-05-18
Create tempdir in test using t.TempDir()

Jacalz:
### Description:

Go 1.18 introduced the .TempDir() method for automatically creating temporary directories in tests that then are removed automatically at the end of the tests. use that instead of the functions in the os package.

This change initially introduced errors that then turned out to be cases of actual bugs where we did not close file handles.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/preferences_test.go
+++ b/app/preferences_test.go
@@ -72,8 +72,7 @@ func TestPreferences_Save_OverwriteFast(t *testing.T) {
 		val["key"] = "value"
 	})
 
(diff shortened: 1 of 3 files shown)


Jacalz:
Sigh. Windows… It might just be a case of some readers not being closed but of course you’ll have to dial the whole test for that. I’ll have a look tomorrow

andydotxyz (commented):
Oh that is so cool - nice to see testing has this feature!

andydotxyz (approved):

PR #4857, opened 2024-05-21
fix mobile mobile driver absolute position computation for padded child windows

toaster:
### Description:

This PR fixes a bug where the absolute position of an element is computed wrong by the mobile driver for child windows. The chrome box is not taken into account when the windows is padded and therefore the position is not adjusted accordingly.
This affects the position of drop-downs.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Great thanks for this fix

Issue #4483, opened 2023-12-24
Add function to obtain Window handle.


MiyamuraMiyako:
### Checklist

- [x] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [X] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

I need use handle pointer to set Window's other attribute.

### Describe the solution you'd like to see.

```golang
w := d.CreateSplashWindow()
w.Show()
var hwnd := w.Handle()
winapi.SetWindowPos(hwnd,..............)
```

andydotxyz:
How do you imagine this working universally for all platforms?
The "Other attributes" may not work either, can you be more specific?

For the most part API that are not exposed are done so for good reason.

MatejMagat305:
maybe with funkcion RunOnNative, it could be posible ...

Jacalz:
Partially related to https://github.com/fyne-io/fyne/issues/1155

dweymouth:
I can see why this is desired and sometimes necessary for certain things outside of Fyne's scope, like there are certain WinAPIs that must be called with an HWND (like integrating with System Media Transport Controls for media player apps [(API)](https://learn.microsoft.com/en-us/windows/win32/api/systemmediatransportcontrolsinterop/nf-systemmediatransportcontrolsinterop-isystemmediatransportcontrolsinterop-getforwindow). I wonder if there could be a case for a Fyne `unsafe` package or something where we expose APIs that most apps shouldn't need, and don't have guaranteed similar behavior on all platforms?

aynakeya:
For those need a temporary way to obtain a window handle in desktop environment (linux,windows.macos).

Here is a working solution.

```golang
var w fyne.Window

func main() {
	a := app.NewWithID("io.fyne.demo")
	w = a.NewWindow("Fyne Demo")
	go func() {
                time.Sleep(2 * time.Second) 
		println("Window handle:", xfyne.GetWindowHandle(w))
	}()
	w.Resize(fyne.NewSize(1080, 720))
	w.ShowAndRun()
}
```

also noted that a window handle is assigned only after the window has been displayed.

`xfyne` utils is defined by following file

`window.go`

```golang
package xfyne

import (
	"fyne.io/fyne/v2"
	"github.com/go-gl/glfw/v3.3/glfw"
	"reflect"
	"unsafe"
)

[...]

Issue #1155, opened 2020-07-02
Reposition Window programmatically


mortzdk:
### Is your feature request related to a problem? Please describe:
I have a window, which needs repositioning according to some other application. I now the screen position, width and height of the other application and can hence place my new window besides it.

### Is it possible to construct a solution with the existing API?
Looking through the code, the only positioning of the window is done with CenterOnScreen. Hence one is not able to reposition the window programmatically.

### Describe the solution you'd like to see:
I guess this feature could be handled by adding another API method "Move" that uses almost the same logic as CenterOnScreen, but repositions the window according to input parameter x and y.

andydotxyz:
Thanks for this, I guess I can understand what you are trying to do.
I am a little nervous about how this could be mis-used as internally the coordinates are scalable and so don't correspond to pixel positions. 
I'm mostly just frustrated because some of my apps continue to position themselves on a screen that does not exist becuase last time I had an external monitor plugged in.

How can we ensure that such an API does exaclt what users expect as well as what developers want?

andydotxyz:
There is a concern amongst the team that this function enables potentially user-unhelpful behaviours and encourages multi-window (I.e. palette) window apps which are discouraged in favour of laying out items in a larger window.

mortzdk:
While I do understand you concern, I still believe this is a very useful and valid feature.

In the end, how the application is layed out is up to the application developer, not the framework. That being said, one could easily write comments that encourages specific use of the method.

andydotxyz:
That's a tough point actually - many systems would say it's up to the OS/Window Manager. In macOS an the main APIs moved to require specific pemissions to move or resize windows, so it's not quite so trivial.

Potentially it we could be limited it to only being called before Show(), also make sure that the requests are within visible bounds, and ignore any requests after that to avoid usability issues.

One other thing to note - what coordinate system would you use? When calling Resize() it is not using pixels - so a Window that has Resize() with 200, 200 that is placed at 0, 0 will not have bottom right 200, 200. There will be the windowmanager specific borders added and then also the monitor scale is a factor. Mostly likely on a modern monitor it is more likely bottom right 432, 408 - how do you factor that into a Move() API?

itschrono:
> That's a tough point actually - many systems would say it's up to the OS/Window Manager. In macOS an the main APIs moved to require specific pemissions to move or resize windows, so it's not quite so trivial.
> 
[...]

Issue #1129, opened 2020-06-25
Ability to create alway-on-top window


mortzdk:
Should be something as simple as setting window hint: GLFW_FLOATING.

Generally it would be nice if the user could set the window hint settings themselves.

andydotxyz:
There is a discussion related to this on #476 - it's a complex topic as that is not something that will work across all platforms and also is discouraged by some platforms.

Can I ask what the user-facing requirement for this is, are you trying to do dialogs, kiosks - or just forcing user to interact with your app at some point? :)

hick:
I want to make an alarm or some task notification application,  just like an old desktop application named "alarm++" ,   the alert window should be always on top in case other operation & windows covered the alert. just like the following screenshot. 

the tip on systray can only stay for a few seconds, and can not display many alerts at the same time.   and can not operate like Snooze and Off   like the following screenshot

It's ok if only some system like MS Windows & MacOS are supported.   the QT and goqt can do like this but it's so heavy a project, I do not want to use goqt

![image](https://user-images.githubusercontent.com/38886/86318857-aa1d7f00-bc65-11ea-9e67-00d73a95b727.png)

mortzdk:
I'm trying to build a "Deck Tracker" for the Hearthstone game. Most of the work is done reading log files but the actual presentation of the information I would like to use "fyne" for. Currently I have the following window:

![Screenshot from 2020-07-02 08-08-49](https://user-images.githubusercontent.com/2938444/86322473-5a29c280-bc3b-11ea-9d4a-52d6ee65d177.png)

The always-on-top part should be used when the game runs in fullscreen mode or when the game is using the whole screen. In these cases the user will not be able to use the tracker unless it is always on top.

mortzdk:
On the glfw website: https://www.glfw.org/docs/3.1/news.html they say:

> GLFW now supports floating windows, also called topmost or always on top, for easier debugging with the GLFW_FLOATING window hint.

Hence this could also be useful in general, for easier debugging of certain windows, while also supporting cases where it is absolute necessary for the application to either gain the users attention or provide extra services.

andydotxyz:
@mortzdk Thanks for the input and great screenshot :)
Unfortunately I don't know if we could support always-on-top over a fullscreen window. The way that fullscreen is handled on most OS is to not allow any other apps to be visible, and so even if on top you would not be able to see it.
[...]

PR #4858, opened 2024-05-21
Add new driver.NativeWindow interface with RunNative function to access window handle or other context

### LINKED ISSUE #2363: How to get "fyne" to get hwnd under windows
I want to get hwnd without changing the fyne code

dweymouth:
### Description:

Adds a new extension interface `driver.NativeWindow` with a `RunNative` function similar to the existing `driver.RunNative`, for executing native code that requires the window handle or other contexts. 

Fixes #2363, #4483 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


dweymouth:
hmm not sure why the checks aren't starting

andydotxyz:
I'm not sure about top level for interfaces that are not great abstractions of behaviour based APIs. How about the driver package alongside the definition of `RunNative`?

andydotxyz (commented):
General direction looks good - but I would probably not put such a complex interface at the top level, maybe keep it within the driver package along with the platform types.

dweymouth:
Really just that I don't know Android development at all and aren't set up with a mobile simulator - so I don't really know what to do for Android if we wanted to add the View(?) handle to the window context

andydotxyz:
OK, here is my diff for getting Android fixed up and a little test code in the main package to show it working...

[android.diff.txt](https://github.com/user-attachments/files/15747582/android.diff.txt)

driver/native.go
```diff
@@ -1,14 +1,74 @@
...
+// when it is executed on a device with the Wayland windowing system.
+//
+// Since: 2.5
+type WaylandWindowContext struct {
+	// WaylandSurface is the handle to the native Wayland surface.
+	WaylandSurface uintptr
+
+	// EGLSurface is the handle to the native EGL surface.
+	EGLSurface uintptr
+}
```
Jacalz, line 74: Would it be or be more helpful to give access to the underlying `*glfw.Window` type with the desktop driver? It makes the interfaces much more useful but also a bit harder to use (which I think is fine if someone wants to use this in the first place).

Jacalz (changes requested):
I think this likely will have to wait until 2.6 given that we are in the feature freeze?

dweymouth:
> I think this likely will have to wait until 2.6 given that we are in the feature freeze?

I believe Andy said that open PRs before feature freeze are fair game for 2.5 - also see his above comments specifically mentioning landing this for 2.5 :)

driver/native.go
dweymouth: I wanted to keep GLFW as an implementation detail - since theoretically we could move to a different windowing manager (or do it all native), even though we currently have no plans to. Thoughts?

dweymouth: Plus we can always add it to the window context structs later if needed, since it would be a purely additive change

andydotxyz: > I wanted to keep GLFW as an implementation detail and not expose it in the public API 

Agreed - there is no guarantee that GLFW remains the native implementation. Pointers to the native system are at least specific to the platform the developer is looking to get access to.

Jacalz: Hmm, yeah. I do agree with you there to be honest. My idea was to allow access to the glfw window so we could have a nice way of closing out issues like https://github.com/fyne-io/fyne/issues/1129 and https://github.com/fyne-io/fyne/issues/1155 by saying that it is now doable but not something that we officially support through our own APIs.

internal/driver/glfw/window_x11.go
```diff
@@ -2,10 +2,30 @@
...
+
+// assert we are implementing driver.NativeWindow
+var _ driver.NativeWindow = (*window)(nil)
+
+func (w *window) RunNative(f func(any) error) error {
+	var err error
+	done := make(chan struct{})
+	runOnMain(func() {
+		err = f(driver.X11WindowContext{
+			WindowHandle: w.GetWindowHandle(),
```
Jacalz, line 25: I don't think this should be using the method above. I realise now that the "x11:" prefix is likely just how XDG Desktop Protocol wants the string formatted. See comment above about using the value of `GetX11Window()` as the type instead.

If you want to, you can remove `GetWindowhandle()` entirely and making the usage of it use this new functionality instead (by adding "x11:" as a prefix and formatting a base 16 string locally instead). I can also do that in a follow-up PR if you prefer.

internal/driver/glfw/window_darwin.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"fyne.io/fyne/v2/driver"
+)
+
+// assert we are implementing driver.NativeWindow
+var _ driver.NativeWindow = (*window)(nil)
+
+func (w *window) RunNative(f func(any) error) error {
+	var err error
+	done := make(chan struct{})
```
Jacalz, line 14: Do we need the done-channel here? It seems like `runOnMain` already waits as long as the driver already is started. We are waiting twice in most cases. Will this functionality make any sense in the case when the driver hasn't started (we won't have any window)?

internal/driver/glfw/window_darwin.go
```diff
@@ -0,0 +1,23 @@
...
+	runOnMain(func() {
+		err = f(driver.MacWindowContext{
+			NSWindow: uintptr(w.view().GetCocoaWindow()),
```
Jacalz, line 17: The call to `w.view()` can return nil. This might crash if called at the wrong time? Applies to most of the implementations.

internal/driver/glfw/window_wayland.go
```diff
@@ -2,7 +2,30 @@
...
+// assert we are implementing driver.NativeWindow
+var _ driver.NativeWindow = (*window)(nil)
+
+func (w *window) RunNative(f func(any) error) error {
+	var err error
+	done := make(chan struct{})
+	runOnMain(func() {
+		err = f(driver.WaylandWindowContext{
+			WaylandSurface: uintptr(unsafe.Pointer(w.view().GetWaylandWindow())),
+			EGLSurface:     uintptr(unsafe.Pointer(w.view().GetEGLSurface())),
```
Jacalz, line 25: Do we need convert to `unsafe.Pointer()` here? I see you are not doing it for the `darwin` target.

driver/native.go
```diff
@@ -1,14 +1,74 @@
...
-// AndroidContext is passed to the `RunNative` callback when it is executed on an Android device.
-// The VM, Env and Ctx pointers are reqiured to make various calls into JVM methods.
+// NativeWindow is an extension interface for `fyne.Window` that gives access
+// to platform-native features of application windows.
+//
+// Since: 2.5
+type NativeWindow interface {
+	// RunNative  provides a way to execute code within the platform-specific runtime context for a window.
+	// The context types are defined in the `driver` package and the specific context passed will differ by platform.
+	RunNative(func(context any) error) error
```
Jacalz, line 10: All of the implementations for `RunNative()` just return the error from the inner function. Should the function really return an error?

driver/native.go
```diff
@@ -1,14 +1,74 @@
...
+	NSWindow uintptr
+}
+
+// X11WindowContext is passed to the NativeWindow.RunNative callback
+// when it is executed on a device with the X11 windowing system.
+//
+// Since: 2.5
+type X11WindowContext struct {
+	// WindowHandle is the window handle for the native X11 window.
+	WindowHandle string
```
Jacalz, line 61: See comment further down about thing probably not making sense as a string. My string conversion for `.Windowhandle()` is likely a wrong assumption based on how XDG Desktop Portal wants it formatted.

internal/driver/glfw/window_wayland.go
Jacalz, line 25: We can avoid grabbing the lock twice here by storing `w.view()` to a variable.

Jacalz (changes requested):
I have a few more comments inline on the implementation.

Jacalz:
Trying this out on Wayland I get the following panic when doing a `.RunNative()`:
```go
panic: NoWindowContext: The specified window has no context

github.com/go-gl/glfw/v3.3/glfw.acceptError({0x0, 0x0, 0xc0002f6c50?})
        /var/home/jacob/go/pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20240506104042-037f3cc74f2a/error.go:177 +0x170
github.com/go-gl/glfw/v3.3/glfw.panicError(...)
        /var/home/jacob/go/pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20240506104042-037f3cc74f2a/error.go:188
github.com/go-gl/glfw/v3.3/glfw.(*Window).GetEGLSurface(0xc0004114a0?)
        /var/home/jacob/go/pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20240506104042-037f3cc74f2a/native_linbsd_wayland.go:46 +0x25
fyne.io/fyne/v2/internal/driver/glfw.(*window).RunNative.func1()
        /var/home/jacob/git/fyne/internal/driver/glfw/window_wayland.go:20 +0x59
fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).runGL(0xc0002f6c40?)
        /var/home/jacob/git/fyne/internal/driver/glfw/loop.go:132 +0x199
fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).Run(0xc0002f2000)
        /var/home/jacob/git/fyne/internal/driver/glfw/driver.go:168 +0x6f
fyne.io/fyne/v2/internal/driver/glfw.(*window).ShowAndRun(0xc0002f6c40)
        /var/home/jacob/git/fyne/internal/driver/glfw/window.go:230 +0x26
main.main()
        /var/home/jacob/git/fyne/cmd/fyne_demo/main.go:76 +0x46c
```

Related to https://github.com/go-gl/glfw/issues/389?

driver/native.go
```diff
@@ -1,14 +1,74 @@
...
+}
+
+// WaylandWindowContext is passed to the NativeWindow.RunNative callback
+// when it is executed on a device with the Wayland windowing system.
+//
+// Since: 2.5
+type WaylandWindowContext struct {
+	// WaylandSurface is the handle to the native Wayland surface.
+	WaylandSurface uintptr
```
Jacalz, line 70: I also wonder to which degree this uintptr to a `wl_surface` struct is useful. I don't even understand how to use that type in the first place. Can this be converted back into the C-type?

driver/native.go
dweymouth: Those should still be do-able by going through the platform-native C APIs I'd guess.

internal/driver/glfw/window_wayland.go
dweymouth: Yes, as the type is not directly convertible to uintptr

internal/driver/glfw/window_darwin.go
dweymouth: Removed it

internal/driver/glfw/window_darwin.go
dweymouth: Fixed - they will now pass a null uintptr to the callback if executed before the window is created

driver/native.go
dweymouth: Got rid of the errors in the signature

internal/driver/glfw/window_wayland.go
dweymouth: Fixed

driver/native.go
dweymouth: I know nothing about Wayland - this was based on a suggestion from @Geo52rey on Discord

dweymouth:
> Trying this out on Wayland I get the following panic when doing a `.RunNative()`:
> 
> ```go
> panic: NoWindowContext: The specified window has no context
> 
> goroutine 1 [running, locked to thread]:
> github.com/go-gl/glfw/v3.3/glfw.acceptError({0x0, 0x0, 0xc0002f6c50?})
>         /var/home/jacob/go/pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20240506104042-037f3cc74f2a/error.go:177 +0x170
> github.com/go-gl/glfw/v3.3/glfw.panicError(...)
>         /var/home/jacob/go/pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20240506104042-037f3cc74f2a/error.go:188
> github.com/go-gl/glfw/v3.3/glfw.(*Window).GetEGLSurface(0xc0004114a0?)
>         /var/home/jacob/go/pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20240506104042-037f3cc74f2a/native_linbsd_wayland.go:46 +0x25
> fyne.io/fyne/v2/internal/driver/glfw.(*window).RunNative.func1()
>         /var/home/jacob/git/fyne/internal/driver/glfw/window_wayland.go:20 +0x59
> fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).runGL(0xc0002f6c40?)
>         /var/home/jacob/git/fyne/internal/driver/glfw/loop.go:132 +0x199
> fyne.io/fyne/v2/internal/driver/glfw.(*gLDriver).Run(0xc0002f2000)
>         /var/home/jacob/git/fyne/internal/driver/glfw/driver.go:168 +0x6f
> fyne.io/fyne/v2/internal/driver/glfw.(*window).ShowAndRun(0xc0002f6c40)
>         /var/home/jacob/git/fyne/internal/driver/glfw/window.go:230 +0x26
> main.main()
>         /var/home/jacob/git/fyne/cmd/fyne_demo/main.go:76 +0x46c
> ```
> 
> Related to [go-gl/glfw#389](https://github.com/go-gl/glfw/issues/389)? Seems to be happening when getting the EGL surface.

Can you try it again after these changes and if it's still crashing this way, we may have to remove the EGLSurface from the WaylandWindowContext struct and not try to retrieve it at all - at least for now, it can always be added back in 2.6 if the GLFW bug is fixed

internal/driver/glfw/window_x11.go
dweymouth: Changed it to just pass the uintptr from view.GetX11Window()

driver/native.go
dweymouth: Updated it to the raw uintptr

internal/driver/glfw/window_x11.go
```diff
@@ -2,10 +2,29 @@
...
+
+func (w *window) RunNative(f func(any)) {
+	var handle uintptr
+	if v := w.view(); v != nil {
+		handle = uintptr(v.GetX11Window())
+	}
+	runOnMain(func() {
+		f(driver.X11WindowContext{
+			WindowHandle: handle,
+		})
```
Jacalz, line 28: Should we handle the case of someone passing a `nil` callback `f` for these different implementations?

Jacalz (commented):
This looks great. I just have a question inline. Otherwise approved basically :)

internal/driver/glfw/window_x11.go
dweymouth: Eh, I think a panic/crash is fine in this case - the API is being used entirely improperly - and it's the same behavior as driver.RunNative

Jacalz: Sure. I figured as much. That's why I added it as a comment and not request for changes :)

Jacalz (approved):

CarlosHleb:
How to actually use this feature, i tried with main and develop branch(i am trying to set position where window should appear on screen):

```
    a := app.New()
    w := a.NewWindow("Top Right Popup")

    w.SetContent(widget.NewLabel("Hello from top right corner!"))
    w.Resize(fyne.NewSize(300, 150))
	w.RunNative()


w.RunNative undefined (type fyne.Window has no field or method RunNative)compiler[MissingFieldOrMethod](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#MissingFieldOrMethod)
```

```
when i try to type assert:
	w.(driver.X11WindowContext)

impossible type assertion: w.(driver.X11WindowContext)
	driver.X11WindowContext does not implement fyne.Window (missing method Canvas)compiler[ImpossibleAssert](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#ImpossibleAssert)
```

Jacalz:
This PR was merged over a year ago. Please use one of our support channels like GitHub Discussions, Slack or Discord (see https://fyne.io/support/) for questions. The short answer to your question however can be found here: https://github.com/fyne-io/fyne/blob/c4b5c694ac7ebc943ab5ee939ca907520f3ae5af/dialog/file_xdg_flatpak.go#L127

PR #4869, opened 2024-05-23
Add AppendMarkdown function to RichText

### LINKED ISSUE #4866: Add new RichText API for appending markdown to an existing widget
### Checklist

title

Yes, somewhat clunkily
```go
rt2 := widget.NewRichTextFromMarkdown(markdownText)
rt1.Segments = append(rt1.Segments, rt2.Segments...)
rt1.Refresh()
```

RichText.AppendMarkdown(string)

dweymouth:
### Description:
Allows appending additional markdown text to an existing RichText

Fixes #4866 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/markdown.go
+++ b/widget/markdown.go
@@ -29,6 +29,11 @@ func (t *RichText) ParseMarkdown(content string) {
 
 // AppendMarkdown parses the given markdown string and appends the
 // content to the widget, with the appropriate formatting.
+// This API is intended for appending complete markdown documents or
+// standalone fragments, and should not be used to parse a single
+// markdown document piecewise. Use ParseMarkdown for this case.
+//
+// Since: 2.5
 func (t *RichText) AppendMarkdown(content string) {
 	t.Segments = append(t.Segments, parseMarkdown(content)...)
 	t.Refresh()
--- a/widget/markdown.go
+++ b/widget/markdown.go
@@ -31,7 +31,7 @@ func (t *RichText) ParseMarkdown(content string) {
 // content to the widget, with the appropriate formatting.
 // This API is intended for appending complete markdown documents or
 // standalone fragments, and should not be used to parse a single
-// markdown document piecewise. Use ParseMarkdown for this case.
+// markdown document piecewise.
 //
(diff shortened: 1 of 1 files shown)


andydotxyz:
This carries potential risks which we might need to think about or document.
For example a markdown document should only have one title (h1) - but with this addition you could have any number.

Also the interpretation of markdown may depend on what came before it.
I.e. this function cannot be used for streaming multiple chunks of a markdown file - unless it is carefully parsed first. Possibly it's only useful for concatenation of documents?

dweymouth:
Concatenation of documents - or rather fragments of text that might have markdown elements like lists but aren't full documents *per se* - was the use case from Discord that made me think of this feature addition. Is there really anything preventing a markdown text from having multiple titles anyway?

```
# Title 1

### subsubheading
* some stuff

# Title 2
```

andydotxyz:
Yes you can technically have multiple headings, it's the spec that disagrees.

Not sure if it's an API naming issue or just documentation. "ParseMarkdown" also requires that it be a full document so we shouldn't need to introduce new naming to reflect that.

I agree that fragments are ok, but they do need to be complete in their own right. Maybe that's what the doc needs?

dweymouth:
Updated API doc to include the caveat against processing a single markdown document piecewise.

widget/markdown.go
```diff
@@ -29,6 +29,11 @@ func (t *RichText) ParseMarkdown(content string) {
 
 // AppendMarkdown parses the given markdown string and appends the
 // content to the widget, with the appropriate formatting.
+// This API is intended for appending complete markdown documents or
+// standalone fragments, and should not be used to parse a single
+// markdown document piecewise. Use ParseMarkdown for this case.
```
andydotxyz, line 34: The indication that "ParseMarkdown" can work piecewise doesn't seem correct - that code is being used within this helper so I don't know how it could be true.

andydotxyz (approved):
Thanks, I'm sure this will be helpful

PR #4871, opened 2024-05-23
Various cleanups to the test package

Jacalz:
### Description:

A bit less code, faster in some situations but mostly just better readability.

PS: Tests on develop failed even before these changes so the updated images are included.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Nice set of tidies thanks

PR #4875, opened 2024-05-24
Make animation runner and Start zero-alloc

### LINKED ISSUE #4445: Animation runner allocates new slices every frame

dweymouth:
### Description:

Fixes #4445

This is like the third try of fixing this but this way is the smallest diff and most logically similar to the preexisting code. The main change is to keep a new slice reference that can re-use the memory allocated for the animations slice that was previously discarded for garbage collection each tick.

Stopping an animation early is still not zero-alloc but it happens rare enough (relative to start and tick) that it shouldn't be a big issue, at least for now.

### Allocs benchmark:

== old code ==
goos: darwin
goarch: arm64
pkg: fyne.io/fyne/v2/internal/animation
BenchmarkRunnerAllocs-8      1843419           650.5 ns/op        80 B/op          1 allocs/op

we allocate once per frame (at 60fps in real-world) (new []*anim slice)

== new code ==

goos: darwin
goarch: arm64
pkg: fyne.io/fyne/v2/internal/animation
BenchmarkRunnerAllocs-8   	 1871389	       635.2 ns/op	       0 B/op	       0 allocs/op

We do not allocate at all on an individual frame of animations. We are also a tiny bit faster per frame, which is fairly negligible in the benchmark (allocating a slice *is* fast after all), but in real-world should mean less garbage and less GC triggers when it runs longer

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/animation/runner.go
+++ b/internal/animation/runner.go
@@ -20,12 +20,12 @@ type Runner struct {
 	// by the runner goroutine to be ticked each frame
 	pendingAnimations []*anim
 
-	// nextAnimations is the list of animations that will be ticked in the next frame.
+	// nextFrameAnimations is the list of animations that will be ticked in the next frame.
 	// It is accessed only by the runner goroutine and accumulates the continuing animations
 	// during a tick that are not completed, plus the pendingAnimations picked up at the end of the frame.
 	// At the end of a full frame of animations, the nextAnimations slice is swapped with
 	// the current `animations` slice which is then cleared out, while holding the mutex.
-	nextAnimations []*anim
+	nextFrameAnimations []*anim
 
 	runnerStarted bool
 }
@@ -105,18 +105,18 @@ func (r *Runner) runOneFrame() (done bool) {
 	r.animationMutex.Unlock()
 	for _, a := range oldList {
 		if !a.isStopped() && r.tickAnimation(a) {
-			r.nextAnimations = append(r.nextAnimations, a)
+			r.nextFrameAnimations = append(r.nextFrameAnimations, a)
 		}
 	}
 
 	r.animationMutex.Lock()
-	// nil out old r.animations for re-use as next r.nextAnimations
+	// nil out old r.animations for re-use as next r.nextFrameAnimations
 	tmp := r.animations
 	for i := range tmp {
 		tmp[i] = nil
 	}
-	r.animations = append(r.nextAnimations, r.pendingAnimations...)
-	r.nextAnimations = tmp[:0]
+	r.animations = append(r.nextFrameAnimations, r.pendingAnimations...)
+	r.nextFrameAnimations = tmp[:0]
 	// nil out r.pendingAnimations
 	for i := range r.pendingAnimations {
 		r.pendingAnimations[i] = nil
--- a/internal/animation/runner.go
+++ b/internal/animation/runner.go
@@ -74,40 +74,44 @@ func (r *Runner) Stop(a *fyne.Animation) {
 
 func (r *Runner) runAnimations() {
 	draw := time.NewTicker(time.Second / 60)
-
 	go func() {
 		for done := false; !done; {
 			<-draw.C
-			r.animationMutex.Lock()
-			oldList := r.animations
-			r.animationMutex.Unlock()
-			for _, a := range oldList {
-				if !a.isStopped() && r.tickAnimation(a) {
-					r.nextAnimations = append(r.nextAnimations, a)
-				}
-			}
-
-			r.animationMutex.Lock()
-			// nil out old r.animations for re-use as next r.nextAnimations
-			tmp := r.animations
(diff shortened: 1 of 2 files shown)


internal/animation/runner.go
```diff
@@ -23,9 +24,19 @@ func (r *Runner) Start(a *fyne.Animation) {
 
 	if !r.runnerStarted {
 		r.runnerStarted = true
+		if r.animations == nil {
```
andydotxyz, line 27: I think it could avoid duplication if moved up 4 lines? (Before the outer if)

andydotxyz (commented):
A clever approach thanks, just a comment on the duplication inline.

internal/animation/runner.go
dweymouth: One is for initializing `r.animations` and the other for `r.pendingAnimations` though. I could create a helper function if you think it's cleaner, invoked like `r.animations = initIfNil(r.animations)`?

ccoVeille: I think both will be fine. So please choose the one you prefer.

I just raised a point it could have been moved out

Another solution is a function where you pass a 

```
func initL(l *[]anim) {
if l == nil {
l= make....
}
}
```

Then you call initL(r.animations) then same for r.pendingAnimations

But maybe I missed something

Jacalz: I think it is fine as it is. It is not much in the way of duplication

dweymouth:
Any further thoughs?

Jacalz:
Sorry, I've been away over the weekend. I can have a look tomorrow

internal/animation/runner.go
```diff
@@ -12,6 +12,7 @@ type Runner struct {
 	animationMutex    sync.RWMutex
 	animations        []*anim
 	pendingAnimations []*anim
+	nextAnimations    []*anim // used by runner goroutine only
```
Jacalz, line 15: This seems a bit strange naming wise. Pending and next seem like two synonyms for the same thing?

Jacalz (changes requested):
Thanks. I have a little problem with the clash between next and pending not really making sense and making the code harder to follow. Also, I think this really needs a benchmark to show that it makes a difference.

PS: This is not zero-alloc in a strict sense. I'd rather that it uses a buffer to hopefully avoid allocating as often; if possible.

internal/animation/runner.go
andydotxyz: Sorry I missed that it was not identical

internal/animation/runner.go
dweymouth: `pendingAnimations` is the animations that have been `Start`ed during a single tick and haven't been picked up by the runner goroutine yet.

`nextAnimations` is basically "the pre-allocated slice that will be used by the runner goroutine on the next tick to become the new `animations` slice. Yeah the naming is a bit awkward, if you have a better suggestions I'm open to any!

dweymouth:
> Thanks. I have a little problem with the clash between next and pending not really making sense and making the code harder to follow. Also, I think this really needs a benchmark to show that it makes a difference.
> 
> PS: This is not zero-alloc in a strict sense. I'd rather that it uses a buffer to hopefully avoid allocating as often; if possible.

Added a benchmark to show that we are now allocating nothing per frame of animations. You are correct that it is not *technically* zero-alloc but it is amortized zero-alloc since we only need to allocate if we've accumulated more running animations than could fit in the existing allocated slice memory (and then we won't allocate anymore unless the number of concurrent animations exceeds *that* slice's capacity)

internal/animation/runner.go
dweymouth: It can also be thought of as "the slice for accumulating the animations that will run in the next frame"

Jacalz: Hmm. Yeah, I don't quite know how to name it but I think we need something better than what is there currently. Anyone reading the code for the first time will likely be confused.

dweymouth: How about a descriptive comment?

dweymouth: @Jacalz I added descriptive comments to all the slice declarations - hopefully that should make it understandable :)

Jacalz (approved):
Thanks. The extra comments and minor naming change helped a lot :)

PR #4881, opened 2024-05-27
Add HideSeparators option to collection widgets

### LINKED ISSUE #3756: Add option in List to hide separators
### Checklist

The List widget has a lot of nice builtins but if the developer wants a non-interactive list (eg to display a log file as recently discussed in the discord), or if the row widgets themselves handle interactivity, then it is not really suitable to use the existing List without forking it to comment out builtin behavior. 

_No response_

The `List` widget should have new publicly exported flags:

```go
type List struct {
    // ...
    DisableSelection bool
    DisableRowFocus bool
    DisableHover bool
    NoSeparator bool
    // ...
}
```

Or perhaps DisableSelection should disable rows being focusable and hoverable as well, and in that case be named DisableInteractivity?

Also, maybe instead of DisableSelection, have a new enum SelectionMode once multi-select is actually implemented:

```go
type SelectionMode int

const (
    SelectionModeSingle SelectionMode = iota // default, since existing behavior is single-select
    SelectionModeMulti
    SelectionModeNone
)
```

dweymouth:
### Description:

Adds a `HideSeparators bool` property to List, Tree, and Table (GridWrap does not have separators) to toggle whether the separator lines between elements are shown.

Fixes #3756 

### Checklist:

- [ ] Tests included. <- tested by building a toggle checkbox into the List tab of fyne_demo locally
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -36,6 +36,11 @@ type List struct {
 	OnSelected   func(id ListItemID)                         `json:"-"`
 	OnUnselected func(id ListItemID)                         `json:"-"`
 
+	// HideSeparators hides the separators between list rows
+	//
+	// Since: 2.5
+	HideSeparators bool
+
 	currentFocus  ListItemID
 	focused       bool
 	scroller      *widget.Scroll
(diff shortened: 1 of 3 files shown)


andydotxyz:
If it's not too much trouble would it be possible to make this available across all the collection widgets for parity?

andydotxyz (approved):
This will make many people happy, thanks :)

PR #4884, opened 2024-05-28
Fix #4068: take window head into account for interactive area and use latter more consistently

### LINKED ISSUE #4068: widget.Select doesn't display all items on mobile
### Checklist

### Describe the bug

When I have enought options in widget.Select to have to scroll down to choose one, I noticed one or more elements in the end doesn't show up on Android 13.

Works fine on Windows.
Can't select the last option on android emulator and real android phone.

In my real app, where there are more options, this is worse. I can't select the last 3 options.

### How to reproduce

1/ Compile the sample code attached with fyne package -os android -appID select.size
2/ Run on emulator or real android device
3/ Try to select the option 24

### Screenshots

https://github.com/fyne-io/fyne/assets/1087443/2d6fea97-0879-41cb-925f-b7287acdaf62

### Example code

```
package main

import (
	"fmt"

	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.NewWithID("select.size")
	w := a.NewWindow("select.size")

	lstr := make([]string, 25)

	for i := 0; i < 25; i++ {
		lstr[i] = fmt.Sprintf("%d", i)
	}

	sel := widget.NewSelect(lstr, func(s string) {})

	w.SetContent(sel)

	w.ShowAndRun()

}

```

### Fyne version

2.3.5

### Go compiler version

1.20.5

Android 13

### Additional Information

_No response_

toaster:
### Description:

This PR changes the mobile interactive area to not include a displacing window head. Thus, all logic respecting the interactive area automagically handles the window head.

It also changes the menu to only use the interactive area for layout.

In combination, this fixes #4068.

Piggy backed: a lot of code style fixes.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


internal/driver/mobile/canvas.go
```diff
@@ -10,124 +10,132 @@ import (
...
+	onTypedRune     func(rune)
+	padded          bool
+	scale           float32
+	size            fyne.Size
 	touchCancelFunc context.CancelFunc
 	touchLastTapped fyne.CanvasObject
+	touchTapCount   int
+	touched         map[int]mobile.Touchable
+	windowHead      fyne.CanvasObject
 }
```
Jacalz, line 42: The code was a lot easier to read when the fields were spaced out. Would you mind reverting that?

internal/driver/mobile/canvas.go
```diff
@@ -10,124 +10,132 @@ import (
...
-	ret.lastTapDownPos = make(map[int]fyne.Position)
-	ret.lastTapDown = make(map[int]time.Time)
+func newCanvas(dev fyne.Device) fyne.Canvas {
+	d, _ := dev.(*device)
+	ret := &canvas{
+		Canvas: common.Canvas{
+			OnFocus:   handleKeyboard,
+			OnUnfocus: hideVirtualKeyboard,
+		},
+		device:         d,
```
Jacalz, line 51: No need to do the interface cast outside and ignore the `ok` parameter. You can just do an explicit cast inline:
```suggestion
	ret := &canvas{
		Canvas: common.Canvas{
			OnFocus:   handleKeyboard,
			OnUnfocus: hideVirtualKeyboard,
		},
		device:         dev.(*device),
```

Jacalz (changes requested):
Thanks. Nice fix. Just left two small suggestions otherwise thinks look ok. I would kind of have preferred to have the code style changes in a different PR but this is fine.

internal/driver/mobile/canvas.go
toaster: Not if you give me a clear, unambiguous and exhaustive rule for the grouping.
And if you find such rule, give me a hint on how to do so. I tried for myself many times in my career and never was satisfied.
The best things I came up with were functionality groups covering a subset of all properties. In those cases, it turned out to be better to use a describing prefix for the props. This not only automates the grouping by sorting but also improves the readability in the rest of the code where a property is easier to be recognized as being part of a specific functionality.

internal/driver/mobile/canvas.go
toaster: I’m quite sure that your proposal will crash if `dev` would be a test device.

toaster: `x, _ := a.(*typ)` and `x := a.(*typ)` are semantically different.

Jacalz: Surely your code will crash as well given that you don’t check the “ok” value (second parameter)?

Jacalz: Never mind. Of course you get a nil in that case and any other codes does a nil check before checking. Right...

internal/driver/mobile/canvas.go
Jacalz: What do you mean? That doesn't change the fact that I think it would be a lot clearer code to have everything related to dragging in one group, touch related stuff in one; kind of like it was before.

andydotxyz:
Thanks for this. I see that all the desktop test files show the menus moving down to the right. Is that intended/desired?

internal/driver/mobile/canvas.go
andydotxyz: Maybe we can justify or keep this "unordered" groupings which I too find helpful by putting a comment at the top of each to communicate why it is a block - and keep each of them individually in order?

toaster: If you think that the previous grouping had logic in it, you probably can name the groups, don’t you?
And of course the code is clearer if the properties that belong to the same functionality are grouped together. But this should better happen automatically when applying the code style rules, i.e., sorting.
The three dragging props are already grouped by their prefix.

toaster: I could move these out of the “everything else” group like so:
```go
type canvas struct {
	common.Canvas
	content         fyne.CanvasObject
	device          *device
	initialized     bool
	lastTapDown     map[int]time.Time
	lastTapDownPos  map[int]fyne.Position
	menu            fyne.CanvasObject
	onTypedKey      func(event *fyne.KeyEvent)
	onTypedRune     func(rune)
	padded          bool
	scale           float32
	size            fyne.Size
	touched         map[int]mobile.Touchable
	windowHead      fyne.CanvasObject

	dragOffset      fyne.Position
	dragStart       fyne.Position
	dragging        fyne.Draggable

	touchCancelFunc context.CancelFunc
	touchLastTapped fyne.CanvasObject
	touchTapCount   int
}
```

toaster:
> Thanks for this. I see that all the desktop test files show the menus moving down to the right. Is that intended/desired?

Yes, because the test canvas now has a safety margin, i.e., it has an interactive area smaller than the canvas.

internal/driver/mobile/canvas.go
toaster: > Maybe we can justify or keep this "unordered" groupings which I too find helpful by putting a comment at the top of each to communicate why it is a block - and keep each of them individually in order?

Don’t get me wrong, I’m pro-grouping. But the former grouping was a mess. The only reasonable group was the `touch` group. And there were no comments.
IMO, comments are superfluous most of the time. Take `touchTapCount`: Its prefix clearly says where it belongs to. If you see this in the code, you immediately know, that it is about touch stuff. And if sorted alphabetically, it automagically groups with the other `touch*` properties. If you separate them from the others with blank lines, you don’t need a comment because the common prefix already says everything. If it would just be named `tapCount` you would loose these benefits.

andydotxyz: Sounds like we are on the same page then - grouping without need for comments.
Add the event handler fields in another group and what you have must be fairly close to what was there already?

toaster: Not really. I the old grouping there was
```go
	touched map[int]mobile.Touchable
	padded  bool
```
which is totally unclear to me and
```go
	inited                bool
	lastTapDown           map[int]time.Time
	lastTapDownPos        map[int]fyne.Position
	dragging              fyne.Draggable
	dragStart, dragOffset fyne.Position
```
where I can only recognize the drag stuff.

andydotxyz: Maybe "fairly close" to me means different to you. I am happy with a logical tidy up, and some areas probably needed it - I do think we're on the same page.

Jacalz (approved):

PR #4906, opened 2024-06-03
Add support for setting a custom resource as the font source for text

andydotxyz:
### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/canvas/text.go
+++ b/canvas/text.go
@@ -21,6 +21,11 @@ type Text struct {
 	Text      string         // The string content of this Text
 	TextSize  float32        // Size of the text - if the Canvas scale is 1.0 this will be equivalent to point size
 	TextStyle fyne.TextStyle // The style of the text content
+
+	// FontSource defines a resource that can be used instead of the theme for looking up the font.
+	//
+	// Since: 2.5
+	FontSource fyne.Resource
 }
 
 // Hide will set this text to not be visible
@@ -33,7 +38,8 @@ func (t *Text) Hide() {
 // MinSize returns the minimum size of this text object based on its font size and content.
 // This is normally determined by the render implementation.
 func (t *Text) MinSize() fyne.Size {
-	return fyne.MeasureText(t.Text, t.TextSize, t.TextStyle)
+	s, _ := fyne.CurrentApp().Driver().RenderedTextSize(t.Text, t.TextSize, t.TextStyle, t.FontSource)
+	return s
 }
 
 // Move the text to a new position, relative to its parent / canvas
--- a/canvas/text.go
+++ b/canvas/text.go
@@ -23,6 +23,8 @@ type Text struct {
 	TextStyle fyne.TextStyle // The style of the text content
 
 	// FontSource defines a resource that can be used instead of the theme for looking up the font.
+	// When a font source is set the `TextStyle` may not be effective, as it will be limited to the styles
+	// present in the data provided.
 	//
 	// Since: 2.5
 	FontSource fyne.Resource
--- a/internal/painter/font.go
+++ b/internal/painter/font.go
@@ -95,10 +95,23 @@ func lookupFaces(theme, fallback fyne.Resource, family string, style fyne.TextSt
 }
 
 // CachedFontFace returns a Font face held in memory. These are loaded from the current theme.
-func CachedFontFace(style fyne.TextStyle, fontDP float32, texScale float32) *FontCacheItem {
+func CachedFontFace(style fyne.TextStyle, source fyne.Resource, fontDP float32, texScale float32) *FontCacheItem {
+	if source != nil {
+		val, ok := fontCustomCache.Load(source)
+		if !ok {
+			face := loadMeasureFont(source)
+			faces := &dynamicFontMap{family: source.Name(), faces: []font.Face{face}}
+
+			val = &FontCacheItem{Fonts: faces}
+			fontCustomCache.Store(source, val)
+		}
+		return val.(*FontCacheItem)
+	}
+
 	val, ok := fontCache.Load(style)
 	if !ok {
 		var faces *dynamicFontMap
+
 		switch {
 		case style.Monospace:
 			faces = lookupFaces(theme.TextMonospaceFont(), theme.DefaultTextMonospaceFont(), fontscan.Monospace, style)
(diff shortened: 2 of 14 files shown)


internal/painter/font.go
```diff
@@ -313,7 +326,8 @@ type FontCacheItem struct {
 	Fonts shaping.Fontmap
 }
 
-var fontCache = &sync.Map{} // map[fyne.TextStyle]*FontCacheItem
+var fontCache = &sync.Map{}       // map[fyne.TextStyle]*FontCacheItem
+var fontCustomCache = &sync.Map{} // map[string]*FontCacheItem for custom resources
```
dweymouth, line 330: I think we are missing a clean task for fontCustomCache

canvas/text.go
```diff
@@ -21,6 +21,11 @@ type Text struct {
 	Text      string         // The string content of this Text
 	TextSize  float32        // Size of the text - if the Canvas scale is 1.0 this will be equivalent to point size
 	TextStyle fyne.TextStyle // The style of the text content
+
+	// FontSource defines a resource that can be used instead of the theme for looking up the font.
+	//
+	// Since: 2.5
+	FontSource fyne.Resource
```
dweymouth, line 28: What is the behavior if the FontSource doesn't actually match the TextStyle? eg FontSource is a bold font but TextStyle is default? I assume we will fall back to the theme but do we want to document this?

canvas/text.go
```diff
@@ -21,6 +21,11 @@ type Text struct {
 	Text      string         // The string content of this Text
 	TextSize  float32        // Size of the text - if the Canvas scale is 1.0 this will be equivalent to point size
 	TextStyle fyne.TextStyle // The style of the text content
+
+	// FontSource defines a resource that can be used instead of the theme for looking up the font.
```
dweymouth, line 25: We may want to think about compatability with the future where we may expose a public API to retrieve system fonts. This could probably work with that if that API were a way to obtain `fyne.Resource`s by looking up on the OS by font family name. Just something to think about

canvas/text.go
dweymouth, line 28: Is fyne.Resource becoming too overloaded? (it can be a font, an SVG, anything really) - Do we need a new type fyne.Font and an API to attempt to parse a fyne.Resource into a fyne.Font and return an error if it cannot?

canvas/text.go
dweymouth: Actually a FontFamilyName might be a better API for that - so the developer doesn't have to think about using different FontSource resources for different text styles?

canvas/text.go
andydotxyz: Yes it can literally be anything. It was never image specific but was mostly used there.
Perhaps we should look into the idea of specific types of resources like image/font etc... but there is no precedent at this time.

canvas/text.go
andydotxyz: As discussed on Slack the lookups etc could be added but family etc is just metadata that operates at a higher level than this simple configuration of a canvas item metadata.

canvas/text.go
andydotxyz: Updated thanks

internal/painter/font.go
andydotxyz: It should be there - line 154

dweymouth: OK, yeah I see it, thanks. The possible issue is- that particular clean task only runs on settings or theme change. Which was fine when we could only load a limited number of fonts because it was only what the theme supports. And even with OS font lookup we technically have a cap on the number of fonts that will be loaded because it's at most one per Unicode script.

But with this API the app can load an unlimited number of fonts so it is more of a true memory leak now since it's unbounded. Not sure if it would be blocking but at least we will need to file a follow-up ticket to address this sooner than later probably.

canvas/text.go
dweymouth: I want to test this myself before I resolve this conversation, but looks good assuming that is the behavior in-practice

internal/painter/font.go
andydotxyz: I don't know exactly how or when we should clean out fonts that have been used in the app but not used for a while. A follow up sounds reasonable.

andydotxyz: #4908

dweymouth: thanks

dweymouth:
I think we need better error handling before we can merge this. Setting a non-supported font resource (ie a jpg for example) causes crashes in the ResolveFace call. Even a .ttc font (which we do not support, but it's pretty likely users will try to use this API with) crashes.

canvas/text.go
dweymouth: Confirmed that the documented behavior (FontSource takes precedent over TextStyle if the source does not have glyphs for the style) matches the runtime behavior.

dweymouth (changes requested):
We need to make sure this doesn't crash with arbitrary resources set as the FontSource

dweymouth:
I guess the behavior for a not-supported FontResource should be to log a Fyne error and then fall back to the default font. Though it would be harder for developers to spot the error than if we introduce a FontResource type that is only creatable through a constructor that can return an error. Though as you said, there isn't a precedent for this right now.

andydotxyz:
> Even a .ttc font (which we do not support, but it's pretty likely users will try to use this API with) crashes.

That is the same as if they specify something incompatible on the FYNE_FONT environment.

dweymouth:
> > Even a .ttc font (which we do not support, but it's pretty likely users will try to use this API with) crashes.
> 
> That is the same as if they specify something incompatible on the FYNE_FONT environment.

Didn't know that - if this is true I suppose it's not a blocker for this feature but should be a high-priority separate ticket to look at and fix. In a way, FYNE_FONT being able to crash the app is even worse since if it is set improperly OS-wide, no Fyne apps will be able to run.

andydotxyz:
I will have a look to see if there is an obvious fix that could resolve both as it is pretty unfortunate.

andydotxyz:
Thanks for the suggestion @dweymouth - I fixed it for this PR and the existing FYNE_FONT issue. And added a test to show it's working correctly :)

dweymouth (approved):

PR #4910, opened 2024-06-04
Fix last row occasionally missing from list

### LINKED ISSUE #4909: Sometimes the last list row that should be visible doesn't show
### Checklist

### Describe the bug

Sometimes the last should-be-visible row in a list doesn't show until you scroll a bit past where it should start being visible, and if it's the last row, depending on the list widget's size (ie viewport size) it may not show at all

### How to reproduce

it's been a known issue for awhile but just never had a ticket tracking it

### Screenshots

_No response_

### Example code

any app with a list

### Fyne version

develop

### Go compiler version

all

all

### Additional Information

_No response_

dweymouth:
### Description:
The calculation calculates maxRow as the max row that could be visible, but then the rest of the code treated it as an exclusive bound (ie first row that *won't* be visible no matter what) instead of inclusive

Fixes #4909 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -411,11 +411,11 @@ func (l *listLayout) calculateVisibleRowHeights(itemHeight float32, length int) 
 			offY = 0
 		}
 
-		if maxRow > length {
-			maxRow = length
+		if maxRow > length-1 {
+			maxRow = length - 1
 		}
 
-		for i := 0; i < maxRow-minRow; i++ {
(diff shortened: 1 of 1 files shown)


dweymouth:
OK tests are broken. Will close in favor of new PR from my actual fork instead of just using Github editor

PR #4911, opened 2024-06-04
Fix last row sometimes being missing in List

### LINKED ISSUE #4909: Sometimes the last list row that should be visible doesn't show

dweymouth:
### Description:

Fixes #4909 

The `calculateVisibleRowHeights` calculation derived `maxRow` as the max row that could possibly be visible, but the rest of the function treated it as the first row that couldn't be visible (i.e. inclusive vs exclusive bound issue).

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/cmd/fyne_demo/main.go
+++ b/cmd/fyne_demo/main.go
@@ -253,10 +253,10 @@ func makeNav(setTutorial func(tutorial tutorials.Tutorial), loadPrevious bool) f
 
 	themes := container.NewGridWithColumns(2,
 		widget.NewButton("Dark", func() {
-			a.Settings().SetTheme(&forcedVariant{Theme: theme.DefaultTheme(), variant: theme.VariantDark})
+			//a.Settings().SetTheme(&forcedVariant{Theme: theme.DefaultTheme(), variant: theme.VariantDark})
 		}),
 		widget.NewButton("Light", func() {
-			a.Settings().SetTheme(&forcedVariant{Theme: theme.DefaultTheme(), variant: theme.VariantLight})
+			//a.Settings().SetTheme(&forcedVariant{Theme: theme.DefaultTheme(), variant: theme.VariantLight})
 		}),
 	)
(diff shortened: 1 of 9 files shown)


dweymouth:
A bunch of list tests are failing as they seem to depend on the old incorrect behavior. I'd appreciate if someone can check my assumptions before I update them all though.

dweymouth:
> It does seem as though you've found an off-by-one. Just make sure it does not relate to a header row/column counting...

List doesn't have header rows builtin like table

andydotxyz:
Haha you got me :) I should sleep more!

dweymouth:
> Haha you got me :) I should sleep more!

Also just that I'm quite familiar with the List API as something like 40% of my app UI is different kind of lists (with another 40% being gridwraps maybe) 😄

Jacalz (commented):
Hmm. Is it possible to get some kind of code to reproduce this or a test that clearly shows it? I notice no difference in lists on my computer.

dweymouth:
It was a rounding error so it only reproduced at certain list widget sizes relative to the row size. Larger rows (ie taller than just a label) seemed to reproduce the problem more often too. Drag the window size around while being scrolled right near the bottom of the list and eventually you should get the original problem to repro

andydotxyz:
I was able to replicate the issue with one of my apps and this does indeed fix it.

andydotxyz (approved):
Thanks so much for solving this

PR #4913, opened 2024-06-04
Ios local notification

kaicherry:
### Description:

Fixed Local Notifications on iOS
    Added conditional to isDark to return false iOS < 12.0
    Added app entry point to handle being opened via URL Scheme
    Added preliminary support for query options received if app is launched via URL Scheme
Fixes #(???)
I don't know if there were issues on these, didn't look. I simply realized I needed this functionality working on iOS too, discovered the issue. Maybe it will help someone else as well.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.


andydotxyz:
I'm not sure why, but your PR seems to include a lot of other commits, please check it is based correctly on develop at your side.

internal/driver/mobile/app/darwin_ios.m
```diff
@@ -26,11 +27,39 @@ @interface GoInputView : UITextField<UITextFieldDelegate>
 @interface GoAppAppDelegate : UIResponder<UIApplicationDelegate>
 @property (strong, nonatomic) UIWindow *window;
 @property (strong, nonatomic) GoAppAppController *controller;
+@property (strong, nonatomic) NSString *parameters;
 @end
 
 @implementation GoAppAppDelegate
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
-    int scale = 1;
+    [self setUpAppDisplay];
```
sdassow, line 35: Looks like a typo. Further down it's called  `setUpAppDisplay`.

andydotxyz:
In addition to the comments above this PR does not compile (for iOS, which I was testing).
Note also that osascript is not available on iOS.

I would recommend a smaller PR where each fixes a known issue (we are not aware of notifications not working...)

andydotxyz:
Thanks to your tips @kaicherry the local notification could be fixed in d6261cf3b18b7, will be in v2.5.0

PR #4914, opened 2024-06-04
Remove theme.OnPrimaryColor()

toaster:
### Description:

When adding this method, I was unaware of the fact that these color accessors are deprecated in favor of `theme.Color(…)`.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
As discussed on Slack I think this makes sense as the direction of things. Would be good to get another voice to check it though. Context is that more "on..." colours are coming and the theme static API is large and also not useful for most widgets going forward @Jacalz @dweymouth

dweymouth (approved):

PR #4918, opened 2024-06-07
Refactor test theme

toaster:
### Description:

This PR refactors the test theme to:

- enforce unique color values
- warn about undefined color, font or size values

The unique color values are important for testing to be able to distinguish between the used colors in the outcome.
The change already found a lot of wrong color names in markup masters because the markup renderer uses the first matching color value for the name lookup.

The missing symbol fonts already logged an error which was ignored.
It might be reasonable to add an output scanner for Fyne errors to CI and make tests fail if there are errors logged.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


test/theme.go
```diff
@@ -118,24 +130,50 @@ func Theme() fyne.Theme {
...
 
+func newConfigurableTheme(name string, colors map[fyne.ThemeColorName]color.Color, fonts map[fyne.TextStyle]fyne.Resource, sizes map[fyne.ThemeSizeName]float32) fyne.Theme {
+	seen := map[color.Color]fyne.ThemeColorName{}
+	for cn, c := range colors {
+		if seen[c] != "" {
+			// This panic will never happen on released code because it will happen on CI runs prior release.
+			panic(fmt.Sprintf("duplicate color value %#v for color %s already used for color %s in theme %s", c, cn, seen[c], name))
+		}
+		seen[c] = cn
+	}
```
Jacalz, line 146: Wouldn't it be wiser to run this in a test for our theme instead of doing it for each new theme created? We could even have some sort of helper to check that there are no duplicate colours.

Jacalz (changes requested):
Thanks. Nice improvement, seems like sensible changes. I just have a question inline.

test/theme.go
toaster: I can see your point. With a test one have to be aware that a third theme method (like `UglyAndPrettyTheme()`) would need to be checked but I think this is okay. With the current approach, a third method also might bypass the constructor, so there is no real advantage in my approach.
Will implement the tests.

andydotxyz (approved):
Great job catching incorrect test assertions

Jacalz (approved):

PR #4926, opened 2024-06-11
Remove the need to hook resources in for overriding in theme override

### LINKED ISSUE #4596: ThemedResource Icons in custom widgets don't follow the overridden theme when inside a themed container
### Checklist

### Describe the bug

The new theme override container handles resource colouring internally which is not available to other developers to adapt their widget resources.

### How to reproduce

1. Create a custom widget with a standard widget like label
2. Add a resource and render it using an image object
3. Put this into a theme override container
4. Make sure the foreground colour in the theme is different
5. Watch that icon and other content are not the same colour

### Screenshots

_No response_

### Example code

tbc

### Fyne version

pre-2.5

### Go compiler version

1.20

all

### Additional Information

_No response_

andydotxyz:
### Description:

Fixes #4596 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


canvas/image.go
```diff
@@ -270,7 +270,16 @@ func (i *Image) updateReader() (io.ReadCloser, error) {
 	i.isSVG = false
 	if i.Resource != nil {
 		i.isSVG = svg.IsResourceSVG(i.Resource)
-		return io.NopCloser(bytes.NewReader(i.Resource.Content())), nil
+		content := i.Resource.Content()
+		if res, ok := i.Resource.(fyne.ThemedResource); i.isSVG && ok {
+
```
toaster, line 275: I think, this empty line can be deleted.

internal/cache/svg.go
```diff
@@ -57,3 +59,13 @@ func destroyExpiredSvgs(now time.Time) {
 		svgs.Delete(exp)
 	}
 }
+
+func overriddenName(name string, o fyne.CanvasObject) string {
+	if o != nil { // for overridden themes get the cache key right
```
toaster, line 64: When will `o` ever be `nil`?

container/theme_test.go
```diff
@@ -0,0 +1,37 @@
...
+	b := widget.NewButtonWithIcon("", theme.HomeIcon(), func() {})
+	o := NewThemeOverride(b, test.Theme())
+	w := test.NewWindow(o)
+	plain := w.Canvas().Capture().(*image.NRGBA)
+
+	o.Theme = test.NewTheme()
+	o.Refresh()
+	changed := w.Canvas().Capture().(*image.NRGBA)
+
+	assert.NotEqual(t, plain.Pix, changed.Pix)
```
toaster, line 23: Tests only containing `NotEqual` are always a little strange. I think an image comparison or at least a markup comparison would have been better here (and in the text test below).

widget/testdata/accordion/layout_expanded_multiple_open_multiple_items.xml
```diff
@@ -8,15 +8,15 @@
 					<widget pos="32,8" size="9x20" type="*widget.RichText">
 						<text alignment="center" bold size="9x19">A</text>
 					</widget>
-					<image fillMode="contain" pos="8,8" rsc="menuDropDownIcon" size="iconInlineSize" themed="foreground"/>
```
toaster, line 11: AFAICS, this (and other) previously were explicitly themed with the (default) foreground color. Now, it says that it is no longer themed anymore and thus will use the image’s colors, right? Is this really wanted?

toaster (changes requested):
- [ ] verify that no longer theme some icons is really intended
- [ ] remove stray empty line

In general: The PR was quite hard to grasp despite it isn’t that big. It would have helped to have commits organized like so:
1. Add the new param to `GetSvg` and `SetSvg` in a first commit without using them yet but including all the necessary test adjustments.
2. Move the logic and include the necessary tests and test adjustments.

In short, a commit should be focused (only address a single aspect) and complete (in the sense that it does not fail nor miss tests).

dweymouth (approved):
LGTM with addressing Tilo's comments

container/theme_test.go
andydotxyz: > at least a markup comparison would have been better here (and in the text test below).

Unfortunately a markup comparison won't show the differences because the widget state is identical, it is the theme context applied that changes. So no markup difference - only render pixel colour.

internal/cache/svg.go
andydotxyz: Yes, plenty of times. If the context object is nil. Like how the tests call GetSvg with a nil context. I suppose it is possible to step through every call to that and ensure that GetSvg and SetSvg are always providing an object for context - but this is a handy shortcut and a safety that doesn't seem superfluous.

container/theme_test.go
toaster: Then, we should add this.

widget/testdata/accordion/layout_expanded_multiple_open_multiple_items.xml
andydotxyz: There was inconsistency before - named resources did not seem to have theme information (at least not some of them?) for example this happened on one test:
```
								<image pos="66,7" rsc="menuExpandIcon" size="iconInlineSize"/>
								<image pos="8,7" rsc="account.svg" size="iconInlineSize" themed="default"/>
```

I have updated the markup test renderer to be more consistent which has added changes. A smaller change set than before which seems promising - but overall it is now explicit about what theme is applied on all resources I think...

andydotxyz:
> In short, a commit should be focused (only address a single aspect) and complete (in the sense that it does not fail nor miss tests).

I agree, and sorry for the lack of cleanliness - I have been travelling and desperately trying to get the required work done between flights, presentations and other commitments during this release process.

container/theme_test.go
andydotxyz: Added image tests as requested

toaster (approved):

toaster (approved):

toaster (approved):

PR #4929, opened 2024-06-12
more “on” colors

toaster:
### Description:

This PR adds additional “on” colors for the “error”, “warning” and “success” colors.

Later, these will be used to improve the contrast on buttons or other widgets using these colors.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


theme/color.go
```diff
@@ -106,11 +106,26 @@ const (
...
+
 	// ColorNameOnPrimary is the name of theme lookup for a contrast color to the primary color.
 	//
 	// Since: 2.5
 	ColorNameOnPrimary fyne.ThemeColorName = "onPrimary"
 
+	// ColorNameOnSuccess is the name of theme lookup for a contrast color to the success color.
+	//
+	// Since: 2.5
+	ColorNameOnSuccess fyne.ThemeColorName = "onSuccess"
```
Jacalz, line 122: I'm a bit sceptical to the "On" naming (applies to the colour we already have?). It clashes a bit too much with the namning we have for callbacks in my opinion. Can we not call it something like `ColorNameContrastSuccess` or similar?

toaster: I like the naming approach (the color to be used _on_ some other color) and don’t think that it clashes with names in a completely different field (event handling).

toaster: I like the naming approach (the color to be used _on_ some other color) and don’t think that it clashes with names in a completely different field (event handling).

Jacalz: I really do think it sounds as if "ColorNameOnSuccess" is a colour that in some way "happens" when something succeeds. If it is a contrasting colour to it, I'd prefer to mention contrast in the name. Do you have any thoughts on it @dweymouth?

dweymouth: I agree with Jacob - I find the "on" names a little confusing. But maybe prefixing the color names with "Foreground" would work? eg:
`ColorNameForegroundOnSuccess fyne.ThemeColorName = "foregroundOnSuccess"`

That would:
* Make it clear that these "on" colors are variants of the foreground color, not the success color (in this example)
* Make the naming clash with events less confusing as those names all *begin* with `On`

I also think though with the feature freeze this should wait until 2.6 to go in

Jacalz: We already have one "On" colour in and we need to decide on what to do here (but yeah, I do agree on this waiting for 2.6). ColorNameForegroundOnSuccess is a big improvement. There is still room for confusion though. Maybe we can use "Over" or "Above" instead of "On" in that case? Or tie in contrast in some way?

dweymouth: We can still fix the existing "On" color naming before 2.5 release if we want to go with a different scheme. Or back it out as well for 2.6

andydotxyz: instead of `ForegroundNoSuccess` maybe just `SucccessForeground`? I agree on reflection the `On` can be misread.
Another alternative is `Over`, though it is close to overlay.

Could also consider `Contrast`?

toaster: “Over” and “Above” are wrong IMO.
I’m not sure about “foregroundOnSuccess”. This reads as if there might be a “backgroundOnSuccess”.

Also, I think that my future commits for the on colors might need a different approach. Currently, I tweaking the colors until the contrast is good but I lost track of the fact that the existing colors are also used as foreground color on the default background. My adjustments would make the contrast there worse.

What if we introduce not one but two new colors per old color:
- `error`/`success`/`warning`/`primary` stay as they are (as color with good contrast to the default background, maybe adjusted to increase this contrast) and are used where they are currently used as foreground color on default (or similar) background
- for each of these there is a new `*Foreground` and a new `*Background` color which will be used by buttons, progress bar and whatever where the old color was used as background

WDYT?

dweymouth: Maybe we should deprecate the original colors and move to a *Foreground and *Background for each of them? And are we in agreement that until we figure out the naming for these all, maybe the first "on" color that was merged should be backed out of develop and made a part of this PR instead?

toaster: If this step is too radical, here is a smaller one which goes into the same direction:

Rename all the “on” colors from “onColor” to “colorForeground”. This makes the contrast improvements harder and probably pushes them back to 2.6. Maybe I can find a fairly good “intermediate” solution for 2.5.

andydotxyz: We should get agreement on this before the new "OnPrimary" is landed in the coming release!

toaster: I just implemented my suggestion.

toaster: @dweymouth @Jacalz would you go with this recent commit?

toaster: Added some minor fixes, so the commit I was referring to is 9de5027.

Jacalz: Seems good to me. I'm happy

Jacalz (approved):
Big contrast improvements. Nice!

Jacalz (approved):

Jacalz (approved):

PR #4931, opened 2024-06-13
Fix settings app system theme preview

toaster:
### Description:

The settings app always shows a dark preview for “system default” no matter what the system actually said.

This PR moves the `app.defaultVariant()` to `internal/app.DefaultVariant()` and thus gives the settings app access to it.

Note that someone need to check that this still compiles for all platforms.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz (approved):
Nice :)

andydotxyz (commented):
On this version the system default preview always seems to be dark?

andydotxyz:
Compiles on Darwin, iOS, Android, WASM - but does not seem to get the preview right on my macOS machine

toaster:
Found and fixed.

cmd/fyne_settings/settings/appearance.go
```diff
@@ -11,14 +11,18 @@ import (
...
 	"fyne.io/fyne/v2/theme"
 	"fyne.io/fyne/v2/widget"
 )
 
 const (
-	systemThemeName = "system default"
+	themeNameDark           = "dark"
+	themeNameLight          = "light"
+	themeNameSystemLabel    = "system default"
+	themeNameSystemSettings = ""
```
andydotxyz, line 25: I was a little confused by the naming - would `themeNameSystem` be OK? It feels like that matches the other keys, leaving the `Label` variant just used for label purposes?

andydotxyz (changes requested):
Looks good thanks, I just got stuck on naming.

andydotxyz (approved):

Jacalz (approved):

PR #4934, opened 2024-06-13
Allow an About menu item in File to be used as special

andydotxyz:
to replace the macOS about in the app menu

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/menu_darwin.go
+++ b/internal/driver/glfw/menu_darwin.go
@@ -139,11 +139,12 @@ func exceptionCallback(e *C.char) {
 
 func handleSpecialItems(w *window, menu *fyne.Menu, nextItemID int, addSeparator bool) (*fyne.Menu, int) {
 	for i, item := range menu.Items {
-		if item.Label == "Settings" || item.Label == "Settings…" || item.Label == "Preferences" || item.Label == "Preferences…" {
+		switch item.Label {
+		case "Settings", "Settings…", "Preferences", "Preferences…":
 			items := make([]*fyne.MenuItem, 0, len(menu.Items)-1)
 			items = append(items, menu.Items[:i]...)
 			items = append(items, menu.Items[i+1:]...)
-			menu, nextItemID = handleSpecialItems(w, fyne.NewMenu(menu.Label, items...), nextItemID, false)
+			menu.Items = items
 
 			insertNativeMenuItem(C.darwinAppMenu(), item, nextItemID, 1)
 			if addSeparator {
@@ -160,7 +161,29 @@ func handleSpecialItems(w *window, menu *fyne.Menu, nextItemID int, addSeparator
 				)
 			}
 			nextItemID = registerCallback(w, item, nextItemID)
-			break
+		case "About":
+			items := make([]*fyne.MenuItem, 0, len(menu.Items)-1)
+			items = append(items, menu.Items[:i]...)
+			if i < len(menu.Items)-1 {
+				items = append(items, menu.Items[i+1:]...)
+			}
+			menu.Items = items
+
+			insertNativeMenuItem(C.darwinAppMenu(), item, nextItemID, 1)
+			if addSeparator {
+				C.insertDarwinMenuItem(
+					C.darwinAppMenu(),
+					C.CString(""),
+					C.CString(""),
+					C.uint(0),
+					C.int(nextItemID),
+					C.int(1),
+					C.bool(true),
+					unsafe.Pointer(nil),
+					C.uint(0),
+				)
+			}
+			nextItemID = registerCallback(w, item, nextItemID)
 		}
 	}
 	return menu, nextItemID
--- a/internal/driver/glfw/menu_darwin.go
+++ b/internal/driver/glfw/menu_darwin.go
@@ -138,13 +138,18 @@ func exceptionCallback(e *C.char) {
 }
 
 func handleSpecialItems(w *window, menu *fyne.Menu, nextItemID int, addSeparator bool) (*fyne.Menu, int) {
-	for i, item := range menu.Items {
+	menu = fyne.NewMenu(menu.Label, menu.Items...) // copy so we can manipulate
+	for i := 0; i < len(menu.Items); i++ {
+		item := menu.Items[i]
 		switch item.Label {
 		case "Settings", "Settings…", "Preferences", "Preferences…":
 			items := make([]*fyne.MenuItem, 0, len(menu.Items)-1)
 			items = append(items, menu.Items[:i]...)
-			items = append(items, menu.Items[i+1:]...)
+			if i < len(menu.Items)-1 {
+				items = append(items, menu.Items[i+1:]...)
+			}
 			menu.Items = items
+			i--
 
 			insertNativeMenuItem(C.darwinAppMenu(), item, nextItemID, 1)
 			if addSeparator {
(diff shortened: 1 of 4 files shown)


Jacalz:
I suppose this is targeted towards 2.6 given that we are in a feature freeze right?

andydotxyz:
Hmm, in a sense yes. But this is a small enabler that should have landed months ago as part of the new About window in fyne-x. Given that I'm tempted to try and convince you there is no new feature here (and certainly no new api :) )

andydotxyz:
Sending my apologies as this was discovered in git stashes very late in the release process and I should have processed them all ages ago.

dweymouth:
Tested it and it works, I can't say I'm a fan of the "magic strings" approach but it's what was already in place for settings/preferences. We should document this better because I didn't even know this existed. Also we'll need to think about the magic string approach now that we have localization/translation, but that doesn't block this PR.

internal/driver/glfw/menu_darwin.go
```diff
@@ -139,11 +139,35 @@ func exceptionCallback(e *C.char) {
...
 func handleSpecialItems(w *window, menu *fyne.Menu, nextItemID int, addSeparator bool) (*fyne.Menu, int) {
 	for i, item := range menu.Items {
-		if item.Label == "Settings" || item.Label == "Settings…" || item.Label == "Preferences" || item.Label == "Preferences…" {
+		switch item.Label {
+		case "Settings", "Settings…", "Preferences", "Preferences…":
 			items := make([]*fyne.MenuItem, 0, len(menu.Items)-1)
 			items = append(items, menu.Items[:i]...)
 			items = append(items, menu.Items[i+1:]...)
-			menu, nextItemID = handleSpecialItems(w, fyne.NewMenu(menu.Label, items...), nextItemID, false)
+			menu.Items = items
```
toaster, line 147: I can see that removing the recursion is a win but now you are modifying the input.
In the end you are modifying the window’s `mainMenu` which is provided by the app developer and I think we should not do that.
So, either we have three options:
- create a new menu for each special item (like the recursion did)
- create a menu copy up front and then modify this (useless copy when there are no special items but this is a setup task, only)
- create a menu copy on demand (probably leads to more complicated code)

internal/driver/glfw/menu_darwin.go
```diff
@@ -139,11 +139,35 @@ func exceptionCallback(e *C.char) {
...
+					C.uint(0),
+				)
+			}
+			nextItemID = registerCallback(w, item, nextItemID)
+		case "About":
+			items := make([]*fyne.MenuItem, 0, len(menu.Items)-1)
+			items = append(items, menu.Items[:i]...)
+			if i < len(menu.Items)-1 {
+				items = append(items, menu.Items[i+1:]...)
+			}
```
toaster, line 169: I guess, this is a countermeasure against `i > len(menu.Items)` because the latter shrinks with the new implementation.
This reveals several flaws of the new implementation:
1. You did not apply the same protection to the old branch but only to the copied new branch.
2. You did not apply the protection to `menu.Items[i:]` which might also crash now.
3. Your `i` and `item` are no longer in sync with `menu.Items` after a change to `menu.Items` leading to unwanted effects.

Instead we should use
```go
for i := 0; i < len(menu.Items); i++ {
  item := menu.Items[i]
  …
    i-- // if special item
  …
}
```

internal/driver/glfw/menu_darwin.m
```diff
@@ -109,10 +109,18 @@ void handleException(const char* m, id e) {
         }
     }
 
-    if (index > -1) {
-        [menu insertItem:item atIndex:index];
+    if (strcmp(label, "About") == 0) {
+        item = [menu itemArray][0];
+        [item setAction:@selector(tapped:)];
+        [item setTarget:[FyneMenuHandler class]];
+        [item setTag:id+menuTagMin];
```
toaster, line 116: `id` is a keyword. What does it do here?

internal/driver/glfw/menu_darwin.m
```diff
@@ -109,10 +109,18 @@ void handleException(const char* m, id e) {
         }
     }
 
-    if (index > -1) {
-        [menu insertItem:item atIndex:index];
+    if (strcmp(label, "About") == 0) {
```
toaster, line 112: This branch looks like it should be done _before_ initializing a then unused item (which is also not released here).

toaster (changes requested):
- [ ] fix `handleSpecialItems` (add test cases revealing the flaws if they don’t already exist)
- [ ] fix memory leak in (and probably optimize) `insertDarwinMenuItem`

internal/driver/glfw/menu_darwin.go
andydotxyz: Thanks for catching, this should all be sorted now.

Jacalz:
I'm sorry; I like the idea behind the change but I'm not that keen on letting things through, it defeats the purpose of having a feature freeze. Given that idea with it is to let things stabilise and allow us to focus more on bug fixing, I'd be tempted to say that merging PR:s that already are open is a step too far already. What's the point in having a process for the feature freeze if we don't follow it?

andydotxyz:
> What's the point in having a process for the feature freeze if we don't follow it?

Essentially the purpose of a release schedule on a large open source project is about corralling energy towards the release. If we cut off at a specific time all work because a date has passed then the enthusiasm that was going into those items wanes and we have to work again to resuscitate the item 3-4 months later after we have been working on bug fixing. I don't wish to break process but I do want to see the best release we can make in the approximate timescales.

andydotxyz:
In this specific instance it is absolutely my fault - the about screen landed in fyne-x about 3 months ago and I forgot about this tweak that makes it more useful. The result will be macOS apps with multiple About menu entries for the next 4-6 months...

internal/driver/glfw/menu_darwin.go
```diff
@@ -138,13 +138,18 @@ func exceptionCallback(e *C.char) {
...
+	for i := 0; i < len(menu.Items); i++ {
+		item := menu.Items[i]
 		switch item.Label {
 		case "Settings", "Settings…", "Preferences", "Preferences…":
 			items := make([]*fyne.MenuItem, 0, len(menu.Items)-1)
 			items = append(items, menu.Items[:i]...)
-			items = append(items, menu.Items[i+1:]...)
+			if i < len(menu.Items)-1 {
+				items = append(items, menu.Items[i+1:]...)
+			}
```
toaster, line 150: This condition is not necessary. I vote to remove it again.

toaster: Also, the two cases of the switch contain the same code, AFAICS. That should be consolidated.

toaster: And: Where are the tests?

Please add an “About” case to the existing test in `menu_darwin_test.go`.
Also, I’d really hoped for a test checking the flaws that were introduced without being detected.

toaster (changes requested):
- [ ] remove unnecessary condition
- [ ] deduplicate code
- [ ] add “About” menu test case

Nice to have:
- [ ] add test which would have detected the misbehaviour of the first implementation

internal/driver/glfw/menu_darwin.go
andydotxyz: I reverted to the recursion version after all

internal/driver/glfw/menu_darwin.go
andydotxyz: I reverted the code to the older recursive version - some of the test cases seemed to only work that way. About is now added to tests

toaster (approved):

PR #4941, opened 2024-06-14
Make sure initial rendering is correct

### LINKED ISSUE #4937: Wrong font rendering when overriding theme
### Checklist

### Describe the bug

When one uses a theme for overriding which differs with some font metrics (yet unknown to me) from the main theme, the fonts are rendered wrong.

I discovered this when trying to change the hyperlink theme override test to use `test.Theme()` (default test theme) and `test.newTheme()` (ugly test theme) to remove its dependency from the production theme.

The test fails when doing so and further research revealed that the text was rendered differently.

### How to reproduce

Change the hyperlink theme override test in `widget/hyperlink_test.go` like so:
```diff
func TestHyperlink_ThemeOverride(t *testing.T) {
	_ = test.NewApp()
	defer test.NewApp()
-	test.ApplyTheme(t, internalTest.LightTheme(theme.DefaultTheme()))
+	test.ApplyTheme(t, test.Theme())

	hyperlink := &Hyperlink{Text: "Test"}
	bg := canvas.NewRectangle(color.Gray{Y: 0xc0})
	w.Resize(hyperlink.MinSize())

	light := w.Canvas().Capture()
-	test.ApplyTheme(t, test.Theme())
+	test.ApplyTheme(t, test.NewTheme())
	hyperlink.Refresh()
	ugly := w.Canvas().Capture()
	assertPixelsMatch(t, false, ugly, light)

-	cache.OverrideTheme(hyperlink, internalTest.LightTheme(theme.DefaultTheme()))
+	cache.OverrideTheme(hyperlink, test.Theme())
	hyperlink.Refresh()
	override := w.Canvas().Capture()
	assertPixelsMatch(t, true, override, light)
}
```

The test will then fail. One could use `AssertImageMatches` to get image files of the captured images or write a small test app to have it on screen.

### Screenshots

_No response_

### Example code

-

### Fyne version

develop

### Go compiler version

1.22.1

macOS 14.5

### Additional Information

_No response_

andydotxyz:
Relates to #4937

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container/theme.go
+++ b/container/theme.go
@@ -30,6 +30,7 @@ func NewThemeOverride(obj fyne.CanvasObject, th fyne.Theme) *ThemeOverride {
 	t.ExtendBaseWidget(t)
 
 	cache.OverrideTheme(obj, th)
+	obj.Refresh() // required as the widgets passed in could have been initially rendered with default theme
 	return t
 }
 


dweymouth (approved):
Seems fine, but we may have to have @toaster verify the issue as he reported it is indeed fixed before closing it out

andydotxyz:
Ah, no - there is a separate issue where different font styles from themes which are represented by the same font style in memory does not update as you could think.

I'm not sure what to do about that - it will need further study but this fix should land in the meantime (marking it as not a fix)

andydotxyz:
So I'll land this but not close the issue as fixed.

PR #4944, opened 2024-06-15
Use the right cache scope for overridden font resources

### LINKED ISSUE #4937: Wrong font rendering when overriding theme

andydotxyz:
Fixes #4937

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


internal/painter/font.go
```diff
@@ -111,21 +111,29 @@ func CachedFontFace(style fyne.TextStyle, source fyne.Resource, fontDP float32,
 		return val.(*FontCacheItem)
 	}
 
-	val, ok := fontCache.Load(style)
+	scope := ""
+	if o != nil { // for overridden themes get the cache key right
+		scope = cache.WidgetScopeID(o)
+	}
```
toaster, line 117: FYI: Moving the `nil` check into `WidgetScopeID` would simplify the callers.

PR #4946, opened 2024-06-17
Add desktop support for screen blanking

andydotxyz:
Fixes #4538 

### Checklist:

- [ ] Tests included. <- pretty much impossible and/or very slow!
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/driver_xdg.go
+++ b/internal/driver/glfw/driver_xdg.go
@@ -1,0 +1,38 @@
+//go:build linux || freebsd || openbsd || netbsd
+
+package glfw
+
+import "C"
+import (
+	"fyne.io/fyne/v2"
+	"github.com/godbus/dbus/v5"
+)
+
+var inhibitCookie uint32
+
+func setDisableScreenBlank(disable bool) {
+	conn, err := dbus.SessionBus() // shared connection, don't close
+	if err != nil {
+		fyne.LogError("Unable to connect to session D-Bus", err)
+		return
+	}
+
+	if !disable {
+		if inhibitCookie != 0 {
+			obj := conn.Object("org.freedesktop.ScreenSaver", "/org/freedesktop/ScreenSaver")
+			call := obj.Call("org.freedesktop.ScreenSaver.Uninhibit", 0, inhibitCookie)
+			if call.Err != nil {
+				fyne.LogError("Failed to send message to bus", call.Err)
+			}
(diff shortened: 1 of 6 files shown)


dweymouth:
Just to confirm, this will disable sleep AND display dimming for desktop? (As opposed to just disabling sleep but allowing the display to dim/shut off which is a separate thing)

internal/driver/glfw/driver_xdg.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+		return
+	}
+
+	if inhibitCookie != 0 {
+		return
+	}
+	obj := conn.Object("org.freedesktop.ScreenSaver", "/org/freedesktop/ScreenSaver")
+	call := obj.Call("org.freedesktop.ScreenSaver.Inhibit", 0, fyne.CurrentApp().Metadata().Name,
+		"App disabled screensaver")
```
Jacalz, line 38: What kind of standard do we have in regards to strings being localized now that we support that?

Jacalz (changes requested):
My computer didn't seem to want to dim the screen on Linux even without these changes and one minute timer. Strange behaviour but not related to this PR. Left two comments inline :)

internal/driver/glfw/driver_xdg.go
```diff
@@ -0,0 +1,44 @@
...
+			inhibitCookie = 0
+		}
+		return
+	}
+
+	if inhibitCookie != 0 {
+		return
+	}
+	obj := conn.Object("org.freedesktop.ScreenSaver", "/org/freedesktop/ScreenSaver")
+	call := obj.Call("org.freedesktop.ScreenSaver.Inhibit", 0, fyne.CurrentApp().Metadata().Name,
```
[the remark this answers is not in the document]
andydotxyz: Thanks, I missed that. The examples I found didn't follow that spec ;)

internal/driver/glfw/driver_xdg.go
andydotxyz: Good point. It could be localised, but this felt like it was developer info not user info - am I wrong about that too?

Jacalz: Yeah, I didn't seem to get a notification from it when enabling inhibit so maybe it isn't even always presented to the user? I think we can avoid localising it for now :)

Jacalz (approved):

PR #4960, opened 2024-06-22
Last of the theme override code

andydotxyz:
Just the remaining support for theme override:

- [x] Containers
- [x] Collections
- [x] Tests
- [x] Internal widgets

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


container/tabs.go
```diff
@@ -79,6 +79,7 @@ type baseTabs interface {
 
 	items() []*TabItem
 	setItems([]*TabItem)
+	scope() fyne.Widget
```
dweymouth, line 82: sorry, I don't really know why this is needed. If it really is needed, can we add a comment explaining why (and what)?

container/tabs.go
```diff
@@ -795,17 +815,20 @@ func (r *tabCloseButtonRenderer) Layout(size fyne.Size) {
...
-	return fyne.NewSquareSize(theme.IconInlineSize())
+	return fyne.NewSquareSize(r.button.Theme().Size(theme.SizeNameInlineIcon))
 }
 
 func (r *tabCloseButtonRenderer) Objects() []fyne.CanvasObject {
 	return r.objects
 }
 
 func (r *tabCloseButtonRenderer) Refresh() {
+	th := r.button.Theme()
```
dweymouth, line 826: Having to do this everywhere seems cumbersome since only the theme, not the variant, can be overridden by ThemeOverrideContainer. So unless we plan on introducing variant overrides as well later, maybe it's worth thinking about adding a new theme helper that looks up the color given the theme and using the current app variant? Something to think about but not blocking this PR

andydotxyz: There is already `theme.ColorForWidget` which does just that, but it does not make use of the baseWidget cache we have for faster lookups internally.

container/tabs.go
andydotxyz: It is because of the baseTabs is not a widget type, it represents the commonality of two types of widget but is itself not a widget.

andydotxyz: Added a note

dweymouth: If both the baseTabs implementations (AppTabs and DocTabs) are widgets, why not add `fyne.Widget` to the baseTabs interface and then the scope() is not needed?

andydotxyz: I didn't think of that as I was using *BaseWidget before where that would not work. I'll try with the interface now it's a possibility thanks.

andydotxyz: Fixed, nice suggestion thanks.

andydotxyz:
This work has found some issues with the theme application of existing containers (tabs mostly, but also tree, table) which I suppose is a good thing.
I will continue to work on them, it would be nice if it could be a separate PR, as this improvement is demonstrated and tested to be working.

There is one TODO that the scroll container on its own has a 1bit RGB flip that i cannot track down which is not found in any scroll usage, so I'm not too concerned but will continue to explore.

widget/gridwrap.go
```diff
@@ -82,7 +83,12 @@ func (l *GridWrap) CreateRenderer() fyne.WidgetRenderer {
 	l.ExtendBaseWidget(l)
 
 	if f := l.CreateItem; f != nil && l.itemMin.IsZero() {
-		l.itemMin = f().MinSize()
+		item := f()
```
dweymouth, line 86: Could this logic (create new item, then maybe refresh depending on theme) be extracted into a helper function? it appears 3 times. Also I'm a little concerned about calling new internal functions in widgets because does that suggest this would not be possible if writing something like the GridWrap as a custom widget?

andydotxyz: This is not the first internal call within GridWrap, though I appreciate the point.
I think that this new API will go away when we fix the issue that containers do not currently track when the gain or lose child items.
There are two possible approaches that would mean this becomes unneeded but both will take a lot of internal refactoring and this seemed like a good approach for now.

andydotxyz:
The updates uncovered an issue in tree that explained a flakey test, which is now working well - thanks @dweymouth

widget/gridwrap.go
dweymouth: OK so this inspired me to write up a short test case and it seems that yes, containers do not properly apply the theme override to their contents that are added after the fact. I think this is another blocker on theme override being complete, do you not see it that way?

andydotxyz: It works well enough in the Fysion use-case which is the only metric I have at this time.
When you override a container and later add items to it they would be overridden as well. It doesn't seem like a big deal to me but others may disagree.

dweymouth (approved):
Rubber stamp approval - also filed https://github.com/fyne-io/fyne/issues/4982 to track the follow-up work

PR #4967, opened 2024-06-24
[For 2.6] Add drag and drop API to list

dweymouth:
### Description:

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


dweymouth:
This should probably wait for the drag-and-drop design so it can fit with that, including possibly accepting drops from other widgets or outside the window

PR #4968, opened 2024-06-24
Update portal dialogs to use native window API

Jacalz:
### Description:

Look up the window handle using the new native window API instead of using custom methods on the window.
These nice new APIs really do help make the code cleaner. Now I only have to get the xdg_foreign protocol hooked up so I can get the same support on the Wayland side.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Cleaning is good :)

PR #4969, opened 2024-06-24
Improve logic for deciding if scrollbar should be large

Jacalz:
### Description:

This fixes some weird behaviour where the scrollbar on desktop previously would go back to not being expanded if the mouse left the area of the bar while dragging it. It also fixes an issue where the scrollbar on mobile would stay small even when it is used to initiate a scroll.

### Checklist:

- [x] Tests included.
- [x] Lint and formatter run with no errors.
- [x] Tests all pass.

PR #4973, opened 2024-06-25
Ensure scroll bar shrinks if drag ends with mouse outside scroll bar area

dweymouth:
### Description:
Follow-up to https://github.com/fyne-io/fyne/pull/4969

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/widget/scroller.go
+++ b/internal/widget/scroller.go
@@ -82,7 +82,7 @@ func (b *scrollBar) Cursor() desktop.Cursor {
 }
 
 func (b *scrollBar) DragEnd() {
-	b.area.isDragged = false
+	b.area.isDragging = false
 
 	if fyne.CurrentDevice().IsMobile() {
 		b.area.MouseOut()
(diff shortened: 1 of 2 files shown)


internal/widget/scroller.go
```diff
@@ -82,16 +82,17 @@ func (b *scrollBar) Cursor() desktop.Cursor {
 }
 
 func (b *scrollBar) DragEnd() {
-	b.area.isDragged = false
+	b.area.isDragging = false
 
 	if fyne.CurrentDevice().IsMobile() {
 		b.area.MouseOut()
 	}
+	b.area.Refresh()
```
Jacalz, line 90: I think this is causing a double refresh on mobile. You might want to add a return inside the if-statement.

Jacalz (changes requested):
Nice catch. Feels perfect. Just one note about double refresh on mobile :)

internal/widget/scroller.go
dweymouth: Done

Jacalz (approved):

PR #4983, opened 2024-06-29
Document the requirement to refresh if adding items

andydotxyz:
### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container/theme.go
+++ b/container/theme.go
@@ -24,6 +24,9 @@ type ThemeOverride struct {
 // Containers will be traversed and all child widgets will reflect the theme in this container.
 // This should be used sparingly to avoid a jarring user experience.
 //
+// If the content `obj` of this theme override is a container and items are later added to the container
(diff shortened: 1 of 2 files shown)


container/theme.go
```diff
@@ -24,6 +24,9 @@ type ThemeOverride struct {
 // Containers will be traversed and all child widgets will reflect the theme in this container.
 // This should be used sparingly to avoid a jarring user experience.
 //
+// If the content `obj` of this theme override is a container and items are later added to the container
```
dweymouth, line 27: maybe rephrase - "are later added to the container **or any sub-containers**"?

dweymouth (approved):
looks good with one optional suggestion

container/theme.go
andydotxyz: Thanks, I pushed that in as well

PR #4986, opened 2024-07-03
Add support for multiple windows in web driver

### LINKED ISSUE #2740: Add multi window support to the web driver
The web driver currently only support one window. It is an unexpected requirement on Fyne application that should be lifted to get the web driver out of alpha.

The goal of this issue is to make sure that existing API work properly on the web driver, so no new API should be added.

The idea would be to reproduce the same logic the mobile driver use and instead of having multiple real window, we reuse the only window we can create and stack all requested window on top. It might be possible to refactor the mobile code somehow and reuse it for this feature.

andydotxyz:
Adding basic multiple window support to web. The idea is this:

* First window fills the space, as it did before
* The canvas has a (hidden) MultipleWindow container overlay
* Additional windows are created as InnerWindows
* These are added to the root canvas overlay which is then shown

This behaves pretty well and delivers the necessary functionality.
Open now for feedback, though I intend to do more tidying of the underlying code.

Fixes #2740

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
Cool. Nice work. I can't seem to reach the underlying canvas when there is a window open. That seems unintuitive?

Jacalz:
The reason that I mentioned it was because the button in fyne_demo for showing and hiding a window wasn't usable with this approach.

andydotxyz:
Funny, I always just closed the window in that test - it then gets re-used with the show when you tap the button.

Is this real-world problem or not? I’m not sure what other approach works beyond putting the first window into a frame and losing the “full sized primary” approach that most apps will benefit from.

Thanks,
Andrew
—
https://andy.xyz
***@***.***

Jacalz:
I'm not saying that it is a problem, just that the button in the demo makes no sense on the web platform in this case (and the same goes for mobile come to think of it) because the parent window can't be clicked when the child is opened :)

andydotxyz:
I can see your point, but at the same time it never occurred to me that was a way of interacting. It is just a button to test a function, and the calling of `Hide()` is completely functional which is the important thing.

> I'm not saying that it is a problem

Cool, if you have better ideas that's fine but if this is not problematic then I'm still happy to go ahead.

internal/driver/glfw/window.go
```diff
@@ -944,6 +949,34 @@ func (w *window) runOnMainWhenCreated(fn func()) {
...
 
 func (d *gLDriver) CreateWindow(title string) fyne.Window {
+	var root fyne.Window
+	d.windowLock.RLock()
+	count := 0
+	for _, w := range d.windows {
+		if w.(*window).visible {
+			count++
+			if count == 1 {
+				root = w
```
Jacalz, line 959: I think you might get away with a `break` here and a boolean for if there are any visible windows?

internal/driver/glfw/canvas.go
```diff
@@ -34,7 +35,8 @@ type glCanvas struct {
 
 	scale, detectedScale, texScale float32
 
-	context driver.WithContext
+	context         driver.WithContext
+	webExtraWindows *container.MultipleWindows
```
Jacalz, line 39: Do we need this field? Is it not possible to always have the web driver have the canvas content be this, even for cases where this is one window?

Jacalz (changes requested):
Nice work. Just a few notes from a quick review on my phone. Will be away from my laptop for a few days so might have to dismiss my review once changes have been implemented (just so you know) :)

internal/driver/glfw/canvas.go
andydotxyz: I could not find a way to do that - as the multiple window overlay will take events as per previous discussion.
Bear in mind also that because all the canvas code is common if we want to inject a layer below we would have to override a lot of functionality in that common space to handle the visibility and event handling (like earlier commit on this PR).

internal/driver/glfw/window.go
```diff
@@ -944,6 +949,34 @@ func (w *window) runOnMainWhenCreated(fn func()) {
...
+		multi := c.webExtraWindows
+		if multi == nil {
+			multi = container.NewMultipleWindows()
+			multi.Resize(c.Size())
+			c.webExtraWindows = multi
+		}
+		inner := container.NewInnerWindow(title, canvas.NewRectangle(color.Transparent))
+		multi.Add(inner)
+
+		wrap := wrapInnerWindow(inner, root, d)
```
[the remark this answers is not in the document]
andydotxyz: Hmm, this is a tricky one. It was messy so I have tidied it up.
However all of the exit points here are a window creation, so I cannot think of a smart way to extract it into other methods...

internal/driver/glfw/window.go
andydotxyz: Good hint thanks

dweymouth (approved):

PR #4995, opened 2024-07-09
Working around missing screen dimensions for Steam Deck

### LINKED ISSUE #4896: Incorrect scaling on Steam Deck screen
### Checklist

### Describe the bug

Copying over from https://github.com/Jacalz/rymdport/issues/159, Fyne applications are entirely unusable on Steam Deck (maybe even other small screens?). This is really problematic because anyone using a Steam Deck and then installing something like Rymdport or Supersonic on Flathub will be very annoyed because the apps have stopped working.

### How to reproduce

1. Start any Fyne application on Steam Deck.
2. Notice that scaling makes it impossible to use the app.

### Screenshots

image
image

### Example code

Use Fyne demo or install Rymdport or Supersonic from Flathub.

### Fyne version

v2.4.5

### Go compiler version

1.22.3

SteamOS

### Additional Information

See https://github.com/Jacalz/rymdport/issues/159 for more information.

andydotxyz:
And maybe more?

Fixes #4896

Results in this (not the right DPI, but not unexpected given how other software looks)

Screenshot_20240709_095130

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window_desktop.go
+++ b/internal/driver/glfw/window_desktop.go
@@ -315,7 +315,7 @@ func (w *window) detectScale() float32 {
 	}
 
 	widthMm, heightMm := monitor.GetPhysicalSize()
-	if widthMm == 60 && heightMm == 60 { // some sort of failure - mostly on Steam Deck
+	if runtime.GOOS == "linux" && widthMm == 60 && heightMm == 60 { // some sort of failure - mostly on Steam Deck
 		return 1
 	}
 	widthPx := monitor.GetVideoMode().Width
--- a/internal/driver/glfw/window_desktop.go
+++ b/internal/driver/glfw/window_desktop.go
@@ -307,15 +307,15 @@ func (w *window) getMonitorForWindow() *glfw.Monitor {
 
 func (w *window) detectScale() float32 {
 	if build.IsWayland { // Wayland controls scale through content scaling
-		return 1.0
+		return 1
 	}
 	monitor := w.getMonitorForWindow()
 	if monitor == nil {
-		return 1.0
+		return 1
 	}
 
 	widthMm, heightMm := monitor.GetPhysicalSize()
-	if runtime.GOOS == "linux" && widthMm == 60 && heightMm == 60 { // some sort of failure - mostly on Steam Deck
+	if runtime.GOOS == "linux" && widthMm == 60 && heightMm == 60 { // Steam Deck incorrectly reports 6cm square!
 		return 1
 	}
 	widthPx := monitor.GetVideoMode().Width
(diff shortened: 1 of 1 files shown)


internal/driver/glfw/window_desktop.go
```diff
@@ -314,7 +314,10 @@ func (w *window) detectScale() float32 {
 		return 1.0
 	}
 
-	widthMm, _ := monitor.GetPhysicalSize()
+	widthMm, heightMm := monitor.GetPhysicalSize()
+	if widthMm == 60 && heightMm == 60 { // some sort of failure - mostly on Steam Deck
+		return 1
```
toaster, line 319: Nit: I think the `return 1` statements should look all the same. Either with or without `.0`. Personally, I would leave it out.

internal/driver/glfw/window_desktop.go
```diff
@@ -314,7 +314,10 @@ func (w *window) detectScale() float32 {
 		return 1.0
 	}
 
-	widthMm, _ := monitor.GetPhysicalSize()
+	widthMm, heightMm := monitor.GetPhysicalSize()
+	if widthMm == 60 && heightMm == 60 { // some sort of failure - mostly on Steam Deck
```
toaster, line 318: Where does these `60` come from? AFAIU the GLFW source, it would return `0` if the monitor did not provide a number.
If these values are based on experience, the comment should explain this.

toaster: Also, this might result in a wrong scale for real-life 60mmx60mm displays with a huge DPI 😁.

toaster (changes requested):
Just some small things:
- [ ] consolidate scale fallback values
- [ ] explain “failure screen dimensions”

internal/driver/glfw/window_desktop.go
Jacalz, line 318: Should this be not be protected by a `runtime.GOOS == "linux"` check? Maybe even `!build.IsWayland` from what I understand of the issue?

andydotxyz: The Wayland check is above and it will exit early.
Linux specific... I guess so, updated

andydotxyz: That is true. If there was ever a 6cm square screen that we had to support. If that ever happens I will update this for sure!

toaster (approved):

internal/driver/glfw/window_desktop.go
Jacalz: Good point :)

PR #5002, opened 2024-07-15
Fix leak in theme overrides map when widgets expire

### LINKED ISSUE #5000: Using container.NewThemeOverride leaks memory until window closing
### Checklist

### Describe the bug

When adding a widget to a container.NewThemeOverride, the widget and its override theme are stored in the `overrides` map. This is never cleaned out except when a window is closing. This is because the deletion of the widget's entry in the `overrides` map happens in `cache.DestroyRenderer`, but that function is only called when a window is closing. Elsewhere (eg in the regular clean task), renderers are directly destroyed instead of by calling this function, which misses the deletion of the entry from the overrides map, if that widget had a theme override.

### How to reproduce

just check the code as described above. Or create an app that cycles through new widgets, applying an override to them, and then not using them. It will leak memory

### Screenshots

_No response_

### Example code

n/a

### Fyne version

2.5.0

### Go compiler version

n/a

n/a

### Additional Information

_No response_

dweymouth:
### Description:
Ensures that wherever we destroy a widget renderer in the case that it is expired, we also ensure that widget's entry from the theme override map is removed, if any.

Fixes #5000 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/cache/base.go
+++ b/internal/cache/base.go
@@ -84,11 +84,12 @@ func CleanCanvas(canvas fyne.Canvas) {
 		if !ok {
 			continue
 		}
-		winfo, ok := renderers[wid]
+		rinfo, ok := renderers[wid]
 		if !ok {
 			continue
 		}
-		winfo.renderer.Destroy()
+		rinfo.renderer.Destroy()
+		overrides.Delete(wid)
 		delete(renderers, wid)
 	}
 	renderersLock.Unlock()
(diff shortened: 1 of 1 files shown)


Jacalz (commented):
Is it possible to add a test for this?

dweymouth:
> Is it possible to add a test for this?

I don't believe we really have tests for any of the cache cleaning stuff.. If you have any ideas feel free to suggest something but it could be a big refactor...

andydotxyz (approved):
Thanks so much for getting to this, clean fix.

PR #5012, opened 2024-07-19
Don't refresh if no sizes changed

### LINKED ISSUE #5007: stack overflow when calling SetRowHeight in table UpdateCell callback
### Checklist

### Describe the bug

I create a table widget . But the row height is uncertain when it created, only until use other variable return fields.

runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0x1407a780420 stack=[0x1407a780000, 0x1409a780000]
fatal error: stack overflow

runtime stack:
runtime.throw({0x104b739ce?, 0x16ee4adb8?})
        /usr/local/go/src/runtime/panic.go:1023 +0x40 fp=0x16ee4ad50 sp=0x16ee4ad20 pc=0x103016420
runtime.newstack()
        /usr/local/go/src/runtime/stack.go:1103 +0x478 fp=0x16ee4aef0 sp=0x16ee4ad50 pc=0x103030fa8
runtime.morestack()
        /usr/local/go/src/runtime/asm_arm64.s:341 +0x70 fp=0x16ee4aef0 sp=0x16ee4aef0 pc=0x103049010

goroutine 115 gp=0x1400f26ce00 m=5 mp=0x14000100008 [running]:
fyne.io/fyne/v2/widget.splitLines(0x140002491f0)
        /Users/xjy/go/pkg/mod/fyne.io/fyne/v2@v2.5.0/widget/richtext.go:1097 +0x3e8 fp=0x1407a780420 sp=0x1407a780420 pc=0x104999418
fyne.io/fyne/v2/widget.lineBounds(0x140002491f0, 0x0, 0x0, 0x436a0000, {0x436a0000, 0x4193e000}, 0x1407a780be8)
        /Users/xjy/go/pkg/mod/fyne.io/fyne/v2@v2.5.0/widget/richtext.go:943 +0x54 fp=0x1407a780a20 sp=0x1407a780420 pc=0x104997714
fyne.io/fyne/v2/widget.(*RichText).updateRowBounds.func1({0x140104a03f0, 0x1, 0x1})

when i remove this code : 

table.SetRowHeight(id.Row, 50).  

Everything has become normal

### How to reproduce

just use i provide code

### Screenshots

Longshot-20240718234817

### Example code

when i use this code:

``` go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

var data = [][]string{
	{"this is a long string", "Row 1, Col 2"},
	{"Row 2, Col 1", "Row 2, Col 2"},
	{"Row 3, Col 1", "Row 3, Col 2"},
	{"Row 4, Col 1", "Row 4, Col 2"},
	{"Row 5, Col 1", "Row 5, Col 2"},
	{"Row 6, Col 1", "Row 6, Col 2"},
}

func main() {
	mainApp := app.NewWithID("a.b.c.d")
	mainWin := mainApp.NewWindow("test table")
	table := getTable()
	mainWin.SetContent(container.NewStack(table))
	mainWin.CenterOnScreen()
	mainWin.Resize(fyne.NewSize(600, 400))
	mainWin.ShowAndRun()
}

func getTable() *widget.Table {
	var table *widget.Table
	table = widget.NewTableWithHeaders(func() (int, int) { return len(data), 2 },
		func() fyne.CanvasObject {
			label := widget.NewLabel("unload")
			return label
		},
		func(id widget.TableCellID, cell fyne.CanvasObject) {
			c := cell.(*widget.Label)
			if len(data[id.Row][id.Col]) > 10 {
				table.SetRowHeight(id.Row, 50)
			}
			c.SetText(data[id.Row][id.Col])
		},
	)
	table.SetColumnWidth(0, 100)
	table.SetColumnWidth(1, 100)
	return table
}

```
### Fyne version

v2.5.0

### Go compiler version

1.21

macOS Sonoma 14.5 Apple M2

### Additional Information

_No response_

andydotxyz:
Fixes #5007

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


widget/table.go
```diff
@@ -297,13 +297,21 @@ func (t *Table) Select(id TableCellID) {
...
+			return
+		}
 		t.headerSize.Width = width
 	}
 
 	if t.columnWidths == nil {
 		t.columnWidths = make(map[int]float32)
 	}
 
+	if t.columnWidths[id] == width {
```
dweymouth, line 311: Should we add an `else if width == templateCell.Width` here? ie if setting column width to the same thing as the default width it is a no-op (same for row height)

andydotxyz: I don't think that would work, because the `columnWidths[id]` being set would mean it does not use the template wouldn't it?
I guess we could add an optimisation to remove custom widths when they are the same as template, but I'd say that is separate to this fix...

dweymouth: I meant the else case where `id` was not an entry in t.columnWidths. Now that I look again, I see that you are blindly accessing t.columnWidths[id] when it may be unset - that may be the cause of the unit test failures.

andydotxyz: Ah I see, thanks for the catch. I'll get that addressed.

andydotxyz: > that may be the cause of the unit test failures

The test failures seem to relate to something strange in the release branch merge. I cannot replicate it locally either so I am struggling to find the cause. It's on the main develop branch as well as this PR.

andydotxyz:
Rebasing to try and get on top of earlier test failures.

andydotxyz:
Not sure why the actions are not running...

dweymouth:
If you just push a rebasing commit to bring this up-to date with the base branch, it might trigger the checks to run again. Also, shouldn't this target release/v2.5.x instead of develop?

andydotxyz:
> shouldn't this target release/v2.5.x instead of develop?

Really all development is based off `develop` - the release process is to cherry pick fixes from that branch. Often after a release there is tidyup on the release branch which confuses this.
We could be more strict and have all bug fixing on release branches but getting external contributors to do this is pretty difficult, and it's easier to default everyone to the same develop branch. It's not a big deal and we can ensure nothing is missed using milestones instead.

dweymouth (approved):

Issue #4998, opened 2024-07-11
Japanese text doesn't truncate properly using fyne.TextTruncateEllipsis


dweymouth:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

When truncating text in a label, richtext, etc. Japanese text (and possibly other scripts too) don't truncate at the right position and instead overstep the boundary of the widget.

### How to reproduce

Run the example app

### Screenshots

<img width="412" alt="image" src="https://github.com/fyne-io/fyne/assets/5303666/967e9aad-83cb-41c1-8c26-a96b575b6591">

### Example code

```go
package main

import (
	"image/color"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/canvas"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/layout"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Test truncate")
	l1 := widget.NewLabel("hereisakindoflonglabeltextwithnospaces")
	l1.Truncation = fyne.TextTruncateEllipsis
	l2 := widget.NewLabel("これはテキストの切り捨てがどのように機能するかをテストするための日本語の文です")
	l2.Truncation = fyne.TextTruncateEllipsis
	rect := &canvas.Rectangle{
		StrokeColor: &color.NRGBA{R: 255, A: 255},
		StrokeWidth: 2,
	}
	w.SetContent(container.New(layout.NewGridLayoutWithColumns(2),
		container.NewVBox(l1, l2),
		rect,
	))
	w.Resize(fyne.NewSize(300, 150))
	w.ShowAndRun()
}
```

### Fyne version

develop

### Go compiler version

n/a

### Operating system and version

mac

andydotxyz:
I have a local fix that should resolve the issue and will open a PR soon.
However it does not completely fix the test case provided as that is a mix of different types of text which triggers a bug in go-text as far as I can see. Reported upstream and will see what the team say (https://github.com/go-text/typesetting/issues/166)

andydotxyz:
Fixed on develop, will merge to release before .2

PR #5014, opened 2024-07-19
Correctly shape input before truncating

andydotxyz:
Partial fix for #4998

Need to resolve an underlying issue before this can be correctly tested - current state breaks existing tests too

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


dweymouth:
What is the status of this PR - what's the underlying issue that needs to be resolved? I tried this out and it seems to be truncating the Japanese script way too early now:

image

andydotxyz:
> it seems to be truncating the Japanese script way too early now:

Yup, that's the issue, Before it was treating it as a full block of unshaped characters and now it splits it into runs correctly but it will wrap early if they are not all part of exactly the same character / font / face type - the related issue I have opened is https://github.com/go-text/typesetting/issues/166

andydotxyz:
That should cover it now @dweymouth  :)

dweymouth (approved):
Seems to work fine now, thanks for getting this sorted!

PR #5023, opened 2024-07-24
[TASK] rename setting localizer function

Sharrnah:
Just a small followup merge, because the function is not a setter, so renamed it `updateLocalizer()` as suggested.


andydotxyz (approved):
Thanks for this, I think the original got merged a little prematurely I'm glad it's tidied now.

PR #5030, opened 2024-07-26
Fix OS theme variant detection for Linux XDG

### LINKED ISSUE #5029: Light/Dark mode detection no longer works on Ubuntu with Fyne 2.5
### Checklist

### Describe the bug

Fyne apps normally start in light or dark mode, depending on the current preferences of the OS.

Since Fyne 2.5 this does no longer seam to work on Ubuntu. I tested this with the demo app and it consistently starts in dark mode, even though the OS is set to light mode (see screenshot).

However, if I compile an app with Fyne 2.4.5 it correctly runs in light mode, so this looks to me like a new bug that was introduced with 2.5.

I also tested this on Windows 10, and there the theme detection still works with Fyne 2.5 as expected.

### How to reproduce

1. Chose "Light" style under Settings/Appearance
2. Start the Fyne demo app
3. The app runs in dark mode

### Screenshots

Screenshot from 2024-07-25 14-35-04

### Example code

N/A

### Fyne version

2.5.0

### Go compiler version

1.22.4

Ubuntu 22.04.4 LTS

### Additional Information

_No response_

dweymouth:
### Description:

The theme variant detection from rymdport/portal appears to be broken - until the issue can be resolved there we can switch back to the working XDG variant detection code from 2.4.x. Also fix a missed setupTheme() call to get the app to theme correctly to the detected variant at startup.

Fixes #5029 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -13,7 +13,6 @@ import (
 	"github.com/rymdport/portal/notification"
 	"github.com/rymdport/portal/openuri"
 	portalSettings "github.com/rymdport/portal/settings"
-	"github.com/rymdport/portal/settings/appearance"
 
 	"fyne.io/fyne/v2"
 	internalapp "fyne.io/fyne/v2/internal/app"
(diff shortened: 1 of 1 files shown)


Jacalz:
I'm not a fan of this. Let's try to figure out with this happens and fix this in that package instead of just reverting back to the previous code.

dweymouth:
I did try a bit to investigate the portal package but didn't get anywhere - I'm not familiar with the DBus APIs or the portal codebase so I'll let you dig into this when you have some time.

andydotxyz:
I'm not a fan of backing things out, but if we can't help @Jacalz get a fix into the portal lib and we get to the 2.5.1 release window then let's land this fix.

Jacalz:
I haven't been home but I'll try to have a look into this later this week. During the rc-testing, I did test out the theme switching and it seemed to work correctly so I'm a bit confused by this but I'll dig in to it and see what I can find.

Jacalz:
What is there to fix? Everything is working fine on Gnome 46 for me on Fedora 40.
[Screencast from 2024-08-07 14-14-53.webm](https://github.com/user-attachments/assets/282e9f46-9ff1-4f63-8c00-29fd67815be5)

dweymouth:
Closing in favor of #5061

PR #5036, opened 2024-07-27
Read double click delay time from OS

### LINKED ISSUE #4448: Get DoubleTapDelay from the OS if an API is available
### Checklist

Currently the double click delay is hard-coded to 300 ms on desktop and 500 ms on mobile. This is a good default, but for OSes that provide an API to read the system double click delay (eg [Windows](https://learn.microsoft.com/sv-se/windows/win32/api/winuser/nf-winuser-getdoubleclicktime)), we should use that instead.

no

reading the system double click delay

dweymouth:
### Description:
On Mac and Windows, use OS APIs to read the system double click interval

Fixes #4448 

### Checklist:

- [ ] Tests included. <- tested manually on Mac and Windows
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/driver_windows.go
+++ b/internal/driver/glfw/driver_windows.go
@@ -71,6 +71,7 @@ func setDisableScreenBlank(disable bool) {
 const defaultDoubleTapDelay = 300 * time.Millisecond
 
 func (g *gLDriver) DoubleTapDelay() time.Duration {
+	// https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getdoubleclicktime
 	user32 := syscall.NewLazyDLL("user32.dll")
 	if user32 == nil {
 		return defaultDoubleTapDelay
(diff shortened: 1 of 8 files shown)


internal/driver/glfw/driver_windows.go
```diff
@@ -66,3 +67,19 @@ func setDisableScreenBlank(disable bool) {
 
 	syscall.Syscall(executionState.Addr(), 1, uintptr(uType), 0, 0)
 }
+
+const defaultDoubleTapDelay = 300 * time.Millisecond
```
andydotxyz, line 71: Could this const be left as applying to all drivers rather than adding it in multiple places - I think we should all fall back to the same value.

andydotxyz (commented):
Nice one thanks for this.

Just a note inline that the new methods returning a hard-coded number should all use the same const which was moved just to the windows driver. (note: not tested yet, just reviewing the code)

internal/driver/glfw/driver_windows.go
dweymouth: Updated. Mobile is at 500 milliseconds by design I think since it's harder to tap quickly there, so I left it and defined a constant for each platform (desktop, web, mobile -_pre existing_) instead of a single Fyne-wide value

andydotxyz (approved):
Nice on the DLL cleanup too

PR #5037, opened 2024-07-27
Auto-capitalize error messages in dialog.NewError

### LINKED ISSUE #4876: Error dialogs should be auto capitalized
### Checklist

According to Go's wiki:
> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context.

However, our context (error dialogs) are usually showed to the end/final user, and I believe texts in error dialogs should start capitalized, and end with a period.

[See discussion on Discord.](https://discord.com/channels/953020122690359327/953031065331695666/1243630292372295681)

Yes

By default, every error dialog should start capitalized and end with a period. I believe most languages follows this "convention" (?).
To disable, maybe you could use a function like:
```go
a := dialog.NewError(fmt.Errorf("something went very wrong in our side, error code: %v", errInt))
a.DisableAutoCaps()
a.Show()
```

dweymouth:
### Description:

Fixes #4876 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/information.go
+++ b/dialog/information.go
@@ -1,6 +1,9 @@
 package dialog
 
 import (
+	"unicode"
+	"unicode/utf8"
+
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/container"
 	"fyne.io/fyne/v2/lang"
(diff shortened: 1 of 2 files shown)


andydotxyz:
Elsewhere I think we have used helpers from the x/text package which have a much simpler API

dweymouth:
> Elsewhere I think we have used helpers from the x/text package which have a much simpler API

Do you know what the APIs were? I can't find it with quick codebase search and googling

andydotxyz:
This is what I was thinking of: https://pkg.go.dev/golang.org/x/text/cases#Upper

PR #5043, opened 2024-07-30
Fall back to English if there is no sensible match

### LINKED ISSUE #5040: If there is not a good translation match for an app with the user's OS language settings, the first translation alphabetically is used
### Checklist

### Describe the bug

For example, if the app supplies a German (de) translation, and a user's OS language is Chinese, and the app does not have a Chinese translation available, it will load German instead of English. English (if available) being hard-coded as the fallback in the case that there is no translation matching seems like a good default, other possibilities include providing an API to set the app's translation fallback language, or preferred language per-app (which could be exposed in the app's settings UI)

### How to reproduce

load a Fyne app that has translations alphabetically before en, with the OS language set to a language for which the app has no translations.

### Screenshots

_No response_

### Example code

n/a

### Fyne version

develop

### Go compiler version

n/a

n/a

### Additional Information

_No response_

andydotxyz:
Fixes #5040

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/lang/lang.go
+++ b/lang/lang.go
@@ -169,6 +169,7 @@ func init() {
 	bundle = i18n.NewBundle(language.English)
 	bundle.RegisterUnmarshalFunc("json", json.Unmarshal)
 
+	translated = []language.Tag{language.Make("en")} // the first item in this list will be the fallback if none match
 	err := AddTranslationsFS(translations, "translations")
 	if err != nil {
 		fyne.LogError("Error occurred loading built-in translations", err)
--- a/lang/lang_test.go
+++ b/lang/lang_test.go
@@ -4,7 +4,6 @@ import (
 	"testing"
 
 	"fyne.io/fyne/v2"
-	"github.com/nicksnyder/go-i18n/v2/i18n"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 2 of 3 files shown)


lang/lang.go
```diff
@@ -169,6 +169,7 @@ func init() {
 	bundle = i18n.NewBundle(language.English)
 	bundle.RegisterUnmarshalFunc("json", json.Unmarshal)
 
+	translated = []language.Tag{language.Make("en")} // the first item in this list will be the fallback if none match
```
dweymouth, line 172: This is kind-of assuming the app has supplied its own English translations, which to be fair is a pretty reasonable bet, but if not, it will lead to lots of "translation failure" logs since Fyne does supply an English translation and it will fail on all the strings unique to the app itself.

Do we want to track if the app developer has supplied their own english translation dictionary, and only if that's true, put english at the front?

andydotxyz:
Hmm, I dunno.
It feels like defaulting to a language that the toolkit does not include is worse, because then the app has certain unhandled elements that the developer can’t fix?

I’m not sure that we should re-introduce the old behaviour for apps when English is not included, but if there is a better system that can be introduced I’d be open to that.

Mostly I’d like to resolve the immediate issue so that we can accept more translation packs.

Sharrnah:
maybe https://github.com/fyne-io/fyne/pull/5026 could work? I added English as a default but it can be changed by the app developer in case the app itself has no english translation?

I am just not 100% sure about the function names in my PR and if how i do it is even considered okay.

lang/lang_test.go
```diff
@@ -23,6 +24,29 @@ func TestAddTranslations(t *testing.T) {
 	assert.Equal(t, "Match2", L("Test2"))
 }
 
+func TestLocalize_Default(t *testing.T) {
```
dweymouth, line 27: Strangely, this test passes even with the code change commented out, and I played around with various permutations of the test code here and I couldn't get it to fail for the old code and pass for the new code, so I'm not sure what it's actually testing

dweymouth:
> maybe #5026 could work? I added English as a default but it can be changed by the app developer in case the app itself has no english translation?
> 
> I am just not 100% sure about the function names in my PR and if how i do it is even considered okay.
> 
> I am even not sure if we really want to add fyne included translations at all. Since i think we can be quite sure that an App will have its own strings anyway, so relying on fyne included translations will fail with missing translations.
> 
> Question then is, how to handle apps that do have no translation pack at all. As then we definetly need to add the fyne core translation packs. 🤔

#5026 has to wait for Fyne 2.6, being a change that introduces new public APIs. So we do need this or a similar no-api fix in the meantime

lang/lang_test.go
andydotxyz: I was attempting to test the fallback logic separately from the init code - instead I have just pulled in a translation from Weblate and checked the fallback based on internal translations.

dweymouth (approved):

PR #5048, opened 2024-07-31
Update the size and position of Entry accessories to be more consistent with our tap sizes

### LINKED ISSUE #4430: Odd looking SelectEntry with long PlaceHolder - overlap by dropdown ActionItem
### Checklist

### Describe the bug

Using SelectEntry with a long PlaceHolder looks odd because the ActionItem to open the dropdown is rendered partially over the PlaceHolder text.

### How to reproduce

1. Run example code
2. Hover with mouse over ActionItem to open dropdown
3. See overlap as shown in screenshot

### Screenshots

selectentry-placeholder-overlap

### Example code

```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("PlaceHolder SelectEntry Overlap Test")

	e := widget.NewSelectEntry([]string{})
	e.PlaceHolder = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua."

	w.SetContent(container.NewCenter(e))
	w.ShowAndRun()
}
```

### Fyne version

2.4.2

### Go compiler version

different versions

various deskop oses (not tested on mobile)

### Additional Information

_No response_

andydotxyz:
Improved layout and fixing a bug and making way for a new widget I have in mind too :)

Fixes #4430

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

320883980-932de6d6-f7e4-4318-9741-0c1bed3836e9


andydotxyz:
Here is the password accessory. 

Screenshot 2024-07-31 at 22 18 23

widget/entry_password.go
```diff
@@ -67,7 +67,7 @@ func (r *passwordRevealerRenderer) Layout(size fyne.Size) {
 
 func (r *passwordRevealerRenderer) MinSize() fyne.Size {
 	iconSize := r.entry.Theme().Size(theme.SizeNameInlineIcon)
-	return fyne.NewSquareSize(iconSize)
+	return fyne.NewSquareSize(iconSize + r.entry.Theme().Size(theme.SizeNameInnerPadding)*2)
```
dweymouth, line 70: It looks like you're reserving the inner padding here in the actionItem. Should this be applied in the Entry renderer code instead, so that using a standard widget.Icon as an entry actionItem would result in the same layout and spacing as the password entry's revealer?

andydotxyz: Good point, thanks

andydotxyz: Actually after working on this it doesn't seem like a good plan - that would mean that either Entry adds big padding around all action items (including buttons etc) or it has a special case for the revealer or peeks into the renderer of an action item to see if it's a button.

Really it belongs here because the tappable area should not have been just the size of an icon. This is a change and so if you're uneasy about this in a point release we could push back to v2.6.0 with a changelog item mentioning that action items match the expected tappable area by adding inner padding around them if just icons?

dweymouth: It does seem like something developers should know about, in order to add the padding around their ActionItem icons to match the rendering of the builtin ones (password entry, etc). So maybe it is better in 2.6

andydotxyz:
Thanks @dweymouth - I have moved this to `develop` and added a `CHANGELOG` note.

andydotxyz:
@dweymouth happy with this now? I think the points were addressed.

dweymouth (approved):

PR #5055, opened 2024-08-06
Update so we call the systray functions on main thread

### LINKED ISSUE #5039: `SetSystemTrayMenu` doesn't work when run in goroutine
### Checklist

### Describe the bug

I changed my code so it would set the System Tray in a goroutine after another condition was met.  The app still showed the system tray icon but right-clicking it did nothing.  Moving it back so `SetSystemTrayMenu` is always run before `App.Run()` appears to fix the problem.  It may be fine that this is a requirement - but its not documented anywhere and it would be good to at least return or log an error if SetSystemTrayMenu doesn't actually work in this case.

### How to reproduce

See example code.  Run `SetSystemTrayMenu` in a goroutine while `App.Run()` is executing.

### Screenshots

_No response_

### Example code

```go
fyneApp := app.NewWithID("TestApp")
win := fyneApp.NewWindow("Master")
win.SetMaster()
go func () {
  time.Sleep(1 * time.Second)
  trayMenu := fyne.NewMenu("TestMenu",
			fyne.NewMenuItem("Test", func() {
				win.Show()
				fmt.Println("test"
			})}
  if desk, ok := fyneApp.(desktop.App); ok {
    desk.SetSystemTrayMenu(trayMenu)
  }
}
fyneApp.Run()
```

### Fyne version

2.4.5

### Go compiler version

1.22

Windows 10

### Additional Information

_No response_

andydotxyz:
It's a bit of a stab in the dark as I cannot replicate but maybe this fixes the issue?

Fixes #5039

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


internal/driver/glfw/driver_desktop.go
```diff
@@ -136,10 +136,12 @@ func (d *gLDriver) refreshSystray(m *fyne.Menu) {
 	defer d.systrayLock.Unlock()
 
 	d.systrayMenu = m
-	systray.ResetMenu()
-	d.refreshSystrayMenu(m, nil)
+	runOnMain(func() {
+		systray.ResetMenu()
+		d.refreshSystrayMenu(m, nil)
```
dweymouth, line 141: I think we have a double `runOnMain` here since `refreshSystrayMenu` also does a `runOnMain`? Also I believe the systray unlock has to be moved inside the runOnMain block here to prevent another call to refreshSystray from starting until the update is fully complete?

dweymouth: Alternatively, get rid of the lock entirely, since we now have the runOnMain queue to enforce atomicity, but move d.systrayMenu = m inside as well?

andydotxyz: Nice suggestion, thanks

dweymouth (approved):

PR #5056, opened 2024-08-06
If canvas cannot be located assume HiDPI

andydotxyz:
This is a better default for most modern systems.
Could be useful pre-attach to context or if using a software canvas mixed with OpenGL

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/canvas/image.go
+++ b/canvas/image.go
@@ -353,6 +353,9 @@ func (i *Image) renderSVG(width, height float32) (image.Image, error) {
 	if c != nil {
 		// We want real output pixel count not just the screen coordinate space (i.e. macOS Retina)
 		screenWidth, screenHeight = c.PixelCoordinateForPosition(fyne.Position{X: width, Y: height})
+	} else { // no canvas info, assume HiDPI
+		screenWidth *= 2
+		screenHeight *= 2
 	}
 
 	tex := cache.GetSvg(i.name(), i, screenWidth, screenHeight)


dweymouth:
This one kind-of seems like a 2.6 change to me too, since it's more of a change in assumptions about the system than a straightforward bug fix that has no risk of introducing any downgrade in behavior (in this case, for users of non-hidpi systems where we can't detect the scale)

andydotxyz:
What sort of behaviour downgrade are you imagining?

dweymouth:
I was imagining non-HiDPI displays where we can't detect the scaling - apps would currently draw at a reasonable size but with this change may draw way too big? I agree this is a good change, but feels more like a change for a feature release than a bugfix release since it is changing an assumption that may be better for most systems but worse for some

andydotxyz:
> apps would currently draw at a reasonable size but with this change may draw way too big?

No, the draw size is the same. We will draw too many pixels and this will be scaled back down on the GPU.
I don't think there is any behaviour change or broken assumptions in this approach.

andydotxyz:
@dweymouth can you comment on the response? The change in behaviour you were worried about is not present.

dweymouth:
It seems like this was based against the wrong branch, but otherwise seems fine

andydotxyz:
It is targetting the release, can we not land it on the release branch?

dweymouth:
> It is targetting the release, can we not land it on the release branch?

I mean the diff picked up a lot of unrelated changes, that if we want to land those on the release branch as well it's probably best in a separate PR/cherry-pick?

andydotxyz:
My apologies somehow I missed what you meant. I will look at that in the morning.

andydotxyz:
No idea what happened with that history - cherry-picked to a fresh branch and force-pushed

dweymouth (approved):

PR #5057, opened 2024-08-07
Fall back to old Gnome interface when FreeDesktop Appearance fails

### LINKED ISSUE #5029: Light/Dark mode detection no longer works on Ubuntu with Fyne 2.5

Jacalz:
### Description:

Should hopefully fix theme lookup on Ubuntu 22.04 and other Linux distros using too old Gnome versions.

Fixes #5029
Replaces #5030

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -121,7 +121,7 @@ func watchTheme() {
 		internalapp.CurrentVariant.Store(uint64(findFreedesktopColorScheme()))
 
 		portalSettings.OnSignalSettingChanged(func(changed portalSettings.Changed) {
-			if changed.Namespace == "org.freedesktop.appearance" && changed.Key == "color-scheme" {
+			if (changed.Namespace == "org.gnome.desktop.interface" || changed.Namespace == "org.freedesktop.appearance") && changed.Key == "color-scheme" {
 				internalapp.CurrentVariant.Store(uint64(findFreedesktopColorScheme()))
 				fyne.CurrentApp().Settings().(*settings).setupTheme()
 			}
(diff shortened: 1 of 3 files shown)


Jacalz:
I'd appreciate some help testing this given that everything works fine on my end already.

ErikKalkoken:
I tested this fix on Ubuntu 22.04 and it does not appear to work:
- Light mode still not detected on startup
- Live switching the theme in Gnome has no effect on the app's theme

Jacalz:
Hmm. Strange. Thanks for testing. I'll have a look at it again and see if I can figure out why it suddenly stopped working for older Linux systems.

dweymouth:
The code from #5029 (ie the old implementation from 2.4.x) works on Ubuntu 22.04 at least. Maybe figuring out the differences between that and what DBus calls are executed in `portal` would be one angle of investigation?

dweymouth:
In case it's helpful, Fedora 40 is another system where the current code wasn't working but the older code was, at least according to a bug report against Supersonic

Jacalz:
> Maybe figuring out the differences between that and what DBus calls are executed in portal would be one angle of investigation?

Yup. That's my plan as well.

> In case it's helpful, Fedora 40 is another system where the current code wasn't working but the older code was

Given that it works perfect on my Fedora 40 installation, it just confuses things even more...

PR #5061, opened 2024-08-09
Fix color lookup not working as expected on Linux

Jacalz:
### Description:

The theme value was looked up before being set for the first time. I also updated to a new version of the portal package that lets us not only use the value when the theme changes (instead of manually loading it again) but also falls back to use the old Read function to look up settings if ReadOne fails.

Should hopefully fix https://github.com/fyne-io/fyne/issues/5029

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/app_xdg.go
+++ b/app/app_xdg.go
@@ -21,10 +21,6 @@ import (
 	"fyne.io/fyne/v2/theme"
 )
 
-func init() {
-	internalapp.CurrentVariant.Store(uint64(findFreedesktopColorScheme()))
-}
-
 func (a *fyneApp) OpenURL(url *url.URL) error {
 	if build.IsFlatpak {
 		err := openuri.OpenURI("", url.String(), nil)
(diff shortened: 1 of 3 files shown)


app/app_xdg.go
```diff
@@ -21,6 +21,10 @@ import (
 	"fyne.io/fyne/v2/theme"
 )
 
+func init() {
+	internalapp.CurrentVariant.Store(uint64(findFreedesktopColorScheme()))
```
andydotxyz, line 25: This reverts the ability to load the app whilst DBus/Gnome calls are hanging.
Having pulled this out from the goroutine will return to non-gnome desktops having apps hang for 60 sec until the call returns.

No idea why, this just seems to be a thing that many Gnome apps do and it's not good that Fyne apps would do the same.

app/app_xdg.go
```diff
@@ -116,15 +124,11 @@ func rootConfigDir() string {
 }
 
 func watchTheme() {
-	go func() {
-		// with portal this may not be immediate, so we update a cache instead
```
andydotxyz, line 120: This is the comment that explained why it was in a goroutine before the move to `init`

andydotxyz (changes requested):
Apps in fyneDesk block for 60 seconds before showing with this change applied.

app/app_xdg.go
Jacalz: Well, that change was also the one that broke theme lookup at startup but sure. I can do an ugly fix to repaint twice on startup instead.

andydotxyz: Of course ugly workarounds are not desirable.
If you have another way of avoiding the portal library calls hanging that would be better :)

Jacalz: Ugly workaround might be a bit harsh. I have a fix ready now. Will look into the Dbus code to see if there is some flag to pass for it not to hang. Where is the hang happening? All non-Gnome, just FyneDesk or somethign else? If it only happens on desktops that do not support the appearance settings, I would guess it is a bug in my library :)

andydotxyz: > Ugly workaround might be a bit harsh.

Sorry, I thought I was re-using your words - I may have accidentally exaggerated them.

> All non-Gnome, just FyneDesk

I don't know about all, but XFCE is the same.

Jacalz: > Sorry, I thought I was re-using your words - I may have accidentally exaggerated them.

I meant that my wordings for it was a bit too harsh. It is not as ugly of a fix as I thought :)

> I don't know about all, but XFCE is the same.

That sounds like it might be related to desktops not supporting the spec then. I have an idea incoming (no idea if it works or not) but lets merge this fix for now :)

Jacalz: I would appreciate if you could try https://github.com/rymdport/portal/pull/12 together with a version of Fyne that doesn't have the cache. Might make life easier for us all :)

andydotxyz: > That sounds like it might be related to desktops not supporting the spec then. I have an idea incoming (no idea if it works or not) but lets merge this fix for now if it looks good :)

Possibly, but why would the method call hang instead of returning a message that it's not supported?

andydotxyz (approved):
If this fixes things that is great - it is still working for me on the test machine.

dweymouth (approved):

dweymouth:
Works for me on Ubuntu 22.04

app/app_xdg.go
Jacalz: Because I think it is trying to auto start the service but it fails to start (because it doesn't exist) so it has some kind of retry loop. That was my idea.

PR #5063, opened 2024-08-09
Simplify sync.Pool code using generics

Jacalz:
### Description:

Inspired by the syncs package from tailscale, let's use generics to clean up and improve the code for managing pools within Fyne.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


internal/async/pool.go
```diff
@@ -0,0 +1,30 @@
+package async
+
+import "sync"
+
+// Implementation inspired by https://github.com/tailscale/tailscale/blob/main/syncs/pool.go.
+
+// Pool is the generic version of sync.Pool.
+type Pool[T any] struct {
```
dweymouth, line 8: I wonder if we want to add a constructor that takes the New func as a parameter, to make initialization a bit cleaner since we always use a New func? (the constructor can still return the Pool[T] struct by value I think)

Jacalz: I don't think we always use the `New` function. My initial implementation of this was without one.

dweymouth (approved):

PR #5069, opened 2024-08-12
Fix double-tap in file dialog

### LINKED ISSUE #5053: Single click in file dialog enters two directories
### Checklist

### Describe the bug

When clicking a directory in a file dialog, the dialog will go two directories deeper instead of one, whenever possible.

This has been reported to me by a @vaibhav-kaushal: https://github.com/codesoap/spind/issues/1

### How to reproduce

1. Run `go run fyne.io/fyne/v2/cmd/fyne_demo@v2.5.0`.
2. Go to the "Dialogs" tab.
3. Click "Folder Open".
4. Now click a directory which also contains a directory "at the same position"; this means if the first directory is, e.g., the third entry, this directory should also contain another directory as its third entry.

In case I didn't explain 4. well enough, here is an example: Create `mkdir -p /tmp/foo/bar/baz`, then go to `/tmp/foo` with the file dialog and click "bar". Instead of just changing directory to "bar", fyne will directly go to "bar/baz".

### Screenshots

_No response_

### Example code

`go run fyne.io/fyne/v2/cmd/fyne_demo@v2.5.0`

### Fyne version

54b25af3ce9ed44005867999a7fc8ee69da42cc1

### Go compiler version

go version go1.22.1 openbsd/amd64

OpenBSD 7.5

### Additional Information

`git bisect` shows me that 54b25af3ce9ed44005867999a7fc8ee69da42cc1 introduced the bug.

andydotxyz:
Looks like a bad merge in 2.5.0 prep
Fixes #5053

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -535,7 +535,6 @@ func (f *fileDialog) setView(view ViewLayout) {
 	fyne.CurrentApp().Preferences().SetInt(viewLayoutKey, int(view))
 	var selectF func(id int)
 	choose := func(id int) {
-		selectF(id)
 		if file, ok := f.getDataItem(id); ok {
 			f.selectedID = id
 			f.setSelected(file, id)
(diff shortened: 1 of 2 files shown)


dweymouth (dismissed):

codesoap:
The fix works in my test. Thanks!

However, the new test still succeeds if I revert the fix in `dialog/file.go`, so will not prevent a future regression. Is this intended?

dweymouth:
> The fix works in my test. Thanks!
> 
> However, the new test still succeeds if I revert the fix in `dialog/file.go`, so it will not prevent a future regression. Is this intended?

Good catch, the test is intended to catch regressions so should be fixed before merging

andydotxyz:
Thanks for noticing, updated

dweymouth (approved):

PR #5075, opened 2024-08-15
Update serve mode to accept --tags and --metadata flags

### LINKED ISSUE #5074: Make `fyne serve` accept `--metadata` flags
### Checklist

`fyne package` exposes some options that are not exposed to `fyne serve`, but could still be relevant:

- `--metadata`
- `--tags`

And a couple others that may be technically relevant, but of less obvious value, as they seem purely aesthetic for local development:

- `--name`
- `--appVersion`

_No response_

I'd like to see at least `--metadata` passed through when provided to the `fyne serve` command.

flimzy:
### Description:
Pass `--tags` and `--metadata` flag values from `fyne serve` to package phase.

Fixes #5074

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/serve.go
+++ b/cmd/fyne/internal/commands/serve.go
@@ -12,8 +12,10 @@ import (
 // Server serve fyne wasm application over http
 type Server struct {
 	*appData
-	port        int
-	srcDir, dir string
+	port           int
+	srcDir, dir    string
+	tags           string
+	customMetadata keyValueFlag
 }
 
 // Serve return the cli command for serving fyne wasm application over http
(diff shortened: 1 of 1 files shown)


andydotxyz (changes requested):
Hmm, I think some of those `p`s should be `s`.
Double check that you tested the right thing, not sure how something like that slipped in.

flimzy:
:facepalm: 

Muscle memory had me type `git push origin -f serveMetadata` when I fixed that, which of course failed, because I don't have write permission to this repo. And I didn't notice the error message. I've now updated the _correct_ remote.

andydotxyz (approved):
Great thanks

PR #5087, opened 2024-08-23
Use float comparison within delta for gradient angle

dweymouth:
### Description:
Ensures that float rounding errors will not prevent a linear gradient at 0, 90, 180, 270 degrees from using the optimized code

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Thanks

Issue #5093, opened 2024-08-26
Tree widget sometimes doesn't render elements


williambrode:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

When adding elements to a data-bound tree widget and then selecting an element, it can make parts of the tree disappear.

For Developers:
I discussed this issue in discord here: https://discord.com/channels/953020122690359327/953030515152281650/1276684982202663015

TLDR;
There is a race when creating the Tree renderer which creates two TreeContent widgets.  This is somehow causing the problem.  I have a fix PR incoming that moves the initialization logic of the tree content to avoid the issue.  It may be worth reviewing all renderers to ensure they don't create widgets?

### How to reproduce

1. Clone the test case from my repo: https://github.com/williambrode/fyne-test-case/tree/bug/TreeDisappearing
2. Run the app
3. Click "Repo" to add a few elements to the tree in a hierarchy
4. Click around.

I believe the issue reproduces most of the time but not always.  So you may have to try a few times.

### Screenshots

![8-26-2024 (15-50-28)](https://github.com/user-attachments/assets/751b77b7-5e34-46ca-bd60-6d05f3a67563)

### Example code

https://github.com/williambrode/fyne-test-case/tree/bug/TreeDisappearing

### Fyne version

2.5.0

### Go compiler version

1.22.3

### Operating system and version

Windows 10

williambrode:
This was fixed by https://github.com/fyne-io/fyne/pull/5096

PR #5094, opened 2024-08-26
Fix an issue where the Tree scroll and content are created twice beca…

williambrode:
…use the renderer is created twice. Fixes https://github.com/fyne-io/fyne/issues/5093.

 

### Description:

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/tree.go
+++ b/widget/tree.go
@@ -52,6 +52,7 @@ type Tree struct {
 	offset        fyne.Position
 	open          map[TreeNodeID]bool
 	scroller      *widget.Scroll
+	content       *treeContent
 	selected      []TreeNodeID
 }
 
@@ -65,6 +66,7 @@ type Tree struct {
 func NewTree(childUIDs func(TreeNodeID) []TreeNodeID, isBranch func(TreeNodeID) bool, create func(bool) fyne.CanvasObject, update func(TreeNodeID, bool, fyne.CanvasObject)) *Tree {
 	t := &Tree{ChildUIDs: childUIDs, IsBranch: isBranch, CreateNode: create, UpdateNode: update}
 	t.ExtendBaseWidget(t)
+	t.initContentAndScroll()
 	return t
 }
 
@@ -87,7 +89,6 @@ func NewTreeWithData(data binding.DataTree, createItem func(bool) fyne.CanvasObj
 			}
 			updateItem(item, branch, o)
 		})
-
 	data.AddListener(binding.NewDataListener(t.Refresh))
 	return t
 }
@@ -114,6 +115,7 @@ func NewTreeWithStrings(data map[string][]string) (t *Tree) {
 		},
 	}
 	t.ExtendBaseWidget(t)
+	t.initContentAndScroll()
 	return
 }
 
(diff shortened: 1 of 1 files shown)


widget/tree.go
```diff
@@ -52,6 +52,7 @@ type Tree struct {
 	offset        fyne.Position
 	open          map[TreeNodeID]bool
 	scroller      *widget.Scroll
+	content       *treeContent
```
dweymouth, line 55: This does kind-of feel like a band-aid fix since conceptually, treeRenderer should own the treeContent widget. I'd rather get to the bottom of how CreateRenderer is actually called twice in a race condition. Maybe @andydotxyz has ideas?

williambrode: I'm certainly in favor of a more generic fix if there is one.  Here are the two backtraces that both called CreateRenderer:

```go
fyne.io/fyne/v2/widget.newTreeContent(tree.go:580)
fyne.io/fyne/v2/widget.(*Tree).CreateRenderer(tree.go:142)
fyne.io/fyne/v2/internal/cache.Renderer(widget.go:33)
fyne.io/fyne/v2/widget.(*BaseWidget).MinSize(widget.go:74)
fyne.io/fyne/v2/widget.(*Tree).MinSize(tree.go:196)
fyne.io/fyne/v2/layout.(*borderLayout).MinSize(borderlayout.go:79)
fyne.io/fyne/v2.(*Container).MinSize(container.go:90)
fyne.io/fyne/v2/internal/driver/glfw.(*glCanvas).SetContent(canvas.go:158)
fyne.io/fyne/v2/internal/driver/glfw.(*window).SetContent(window.go:256)
main.main(main.go:83)
runtime.main(proc.go:271)
runtime.goexit(asm_amd64.s:1695)
runtime.newproc(:0)


fyne.io/fyne/v2/widget.newTreeContent(tree.go:580)
fyne.io/fyne/v2/widget.(*Tree).CreateRenderer(tree.go:142)
fyne.io/fyne/v2/internal/cache.Renderer(widget.go:33)
fyne.io/fyne/v2/widget.(*BaseWidget).Refresh(widget.go:136)
fyne.io/fyne/v2/widget.(*BaseWidget).Refresh-fm(:0)
fyne.io/fyne/v2/data/binding.(*listener).DataChanged(binding.go:56)
fyne.io/fyne/v2/data/binding.DataListener.DataChanged-fm(:0)
fyne.io/fyne/v2/data/binding.queueItem.func1.1(queue.go:19)
runtime.goexit(asm_amd64.s:1695)
fyne.io/fyne/v2/data/binding.queueItem.func1(queue.go:17)
```

The code for getting the renderer clearly doesn't hold a lock when doing `CreateRenderer` which feels like the underlying issue here:

```go
        renderersLock.RLock()
	rinfo, ok := renderers[wid]
	renderersLock.RUnlock()
	if !ok {
		rinfo = &rendererInfo{renderer: wid.CreateRenderer()}
		renderersLock.Lock()
		renderers[wid] = rinfo
		renderersLock.Unlock()
	}
```

But I experimented a bit with locking there and found that Renderers can create other renderers which would create deadlock in some places.  I believe we either need to find a way to lock when calling CreateRenderer (at least for a single widget instance), or ensure all CreateRenderer methods have no side affects.  It seems more difficult to enforce that CreateRenderer doesn't have side affects BUT users are supposed to expect them to be destroyed and recreated - so I question if side affects should be avoided anyways.

dweymouth: Oh, I think the problem may have to do with bindings as they run on a separate thread IIRC. Yeah we might need @andydotxyz to suggest the best plan for a fix

dweymouth:
There are some tree-related test failures

williambrode:
I just tested this again on the latest `develop` code.  It looks like my fix for the other tree rendering issue also fixes this one: https://github.com/fyne-io/fyne/pull/5096

I still think there are likely other issues that arise from the dueling renderers (if not just resource usage), but the priority is likely greatly reduced now that the user-facing issue is fixed.

widget/tree.go
andydotxyz: I see this is closed but I wanted to comment as pinged incase it helps.
Locking access to renderers whilst one is created wouldn't be the right way - but locking on a specific key would be OK. However as this seems to step from a binding issue then it will likely be resolved by our threading changes for 2.6.

The overall question of when is a render created still applies, however we have no guarantee that `CreateRenderer` is called only once (in fact we say it may be called many times). As long as the new renderer replaces the old which will be garbage collected then I think everything should continue to work as planned?

PR #5098, opened 2024-08-27
Modify fyne package -os web to generate a PWA with empty service worker, update package unit tests

### LINKED ISSUE #5097: Add manifest.json and service-worker to make web app an installable PWA
### Checklist

Using `fyne package -os web` creates an index.html that cannot be installed as a PWA.

It is possible to hack the index.html and add files to the web application generated by `fyne package -os web`, but this is undesirable. A change to the packager is required.

Modify the packager to include a manifest.json, service-worker.js and update index.html to run the service-worker.js. This allows installing the PWA using the browser. The resulting PWA is a binary for the target platform that is generated by the browser itself. Of course, it still runs a captive browser in the application.

The proposed fix has an empty service-worker.js, which allows installing the PWA but the functions normally performed by the service worker, such as caching data and support for updates is not included in this enhancement. Adding a functional service worker requires consideration for what makes sense to support when running a WASM payload in a browser. This enhancement makes it possible for someone in the future to craft functional service worker based on features it should support. It will likely need someone expert in the construction of service workers for browser to design and implement.

joebologna:
### Description:
Using fyne package -os web creates an index.html that cannot be installed as a PWA.

Modified the packager to include a manifest.json, service-worker.js and update index.html to run the service-worker.js. This allows installing the PWA using the browser. The resulting PWA is a binary for the target platform that is generated by the browser itself. Of course, it still runs a captive browser in the application.

The proposed fix has an empty service-worker.js, which allows installing the PWA but the functions normally performed by the service worker, such as caching data and support for updates is not included in this enhancement. Adding a functional service worker requires consideration for what makes sense to support when running a WASM payload in a browser. This enhancement makes it possible for someone in the future to craft functional service worker based on features it should support. It will likely need someone expert in the construction of service workers for browser to design and implement.

Fixes fyne-io/tools#67 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


joebologna:
You are correct, a service worker is not required. The quality of a PWA is measured by its Lighthouse score. The score is generated using the Developer Tools in Chrome.

Not having a service worker will reduce the score by not providing optimization. Since the webpage is just running a WASM payload there isn't a need for a service worker to cache other data. A functional service worker and/or more JavaScript may be required to indicate an update is available and allow it to be refreshed in the PWA. Without the refresh capability may be difficult to get the new WASM payload into the app.

In summary, without doing some research on the refresh capability, the value installing the app as a PWA is dubious.

In my opinion, these are the options:

1. Abandon the enhancement until more research can be done
2. Determine what PWA features are desired/required, refreshing the PWA might be the only one to consider
3. Leave the code to generate service-worker.js and run it but include comments in the .js file to allow modifying itto add the functionality after the web app is built - this kludgy. It would probably be best to allow customization akin to the method used with  FyneApp.toml or AndroidManifest.xml

I'm partial to #2.

Jacalz:
Closing per the comment above. Sorry that this stalled before the move to the new repository. This is something that I really would want to see added so I hope you can continue the work there. Thanks

Issue #3886, opened 2023-05-12
Deadlock acquiring read lock twice with rich text


williambrode:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

Here is the code path showing how we take the same read lock twice in the same thread.  When another thread is waiting on the Write lock between the first read lock and the second, it causes deadlock (which just happened to me while running the debugger).

```golang
func (r *entryContentRenderer) moveCursor() {
	// build r.selection[] if the user has made a selection
	r.buildSelection()
	r.content.entry.propertyLock.RLock()
	provider := r.content.entry.textProvider()
-->	provider.propertyLock.RLock()
	size := provider.lineSizeToColumn(r.content.entry.CursorColumn, r.content.entry.CursorRow)

func (t *RichText) lineSizeToColumn(col, row int) fyne.Size {
	bound := t.rowBoundary(row)
	
func (t *RichText) rowBoundary(row int) *rowBoundary {
-->	t.propertyLock.RLock()
```

### How to reproduce

Unsure of repro steps, but code is clearly incorrect as the description shows.

### Example code

See description for code that causes the issue.

### Fyne version

2.2.3

### Go compiler version

1.19.1

### Operating system and version

Windows 10

williambrode:
Just hit this twice in a row.

Should `func (r *entryContentRenderer) moveCursor() {` even be taking the provider.propertyLock?  Seems like the provider itself should manage the lock on its properties?

@andydotxyz any thoughts on that?  If you agree then I can make a PR for removing the the lock taking there.

andydotxyz:
It shouldn't have to - but it will have been added for a reason so maybe a public call that used it was not locking properly.
As in the other discussion:

> In general private methods assume a lock, but public API must attain the lock if required...

P.S. you don't need my opinion to open a PR - you can leave questions open for any of our contributors to answer.

andydotxyz:
@dweymouth can you provide some context for the re-opening? did we back out the fix of was the exact same problem encountered again somehow?

dweymouth:
[...]

PR #5115, opened 2024-09-05
Avoid deadlock in button renderer by aquiring read lock twice.

### LINKED ISSUE #5114: Deadlock in Button `CreateRenderer()`
### Checklist

### Describe the bug

UI can deadlock due to acquiring the same read lock twice in a single goroutine.  If any other goroutine acquires the same lock for write between the two reads - the second read will hang and never release the first, meaning the entire UI deadlocks.

Here is the sequence that causes the double read lock:
```
#	0xe91ce4	sync.runtime_SemacquireRWMutexR+0x24							runtime/sema.go:82
#	0x14de187	sync.(*RWMutex).RLock+0x47								sync/rwmutex.go:70
#	0x14de163	fyne.io/fyne/v2/widget.(*BaseWidget).Visible+0x23					fyne.io/fyne/v2@v2.5.0/widget/widget.go:85
#	0x14a2a6d	fyne.io/fyne/v2/widget.(*buttonRenderer).updateIconAndText+0x2d				fyne.io/fyne/v2@v2.5.0/widget/button.go:396
#	0x14a14af	fyne.io/fyne/v2/widget.(*Button).CreateRenderer+0x4cf					fyne.io/fyne/v2@v2.5.0/widget/button.go:123
#	0x11892d7	fyne.io/fyne/v2/internal/cache.Renderer+0x117						fyne.io/fyne/v2@v2.5.0/internal/cache/widget.go:33
#	0x14de0f0	fyne.io/fyne/v2/widget.(*BaseWidget).MinSize+0x30					fyne.io/fyne/v2@v2.5.0/widget/widget.go:74
#	0x14a166e	fyne.io/fyne/v2/widget.(*Button).MinSize+0x2e						fyne.io/fyne/v2@v2.5.0/widget/button.go:148
#	0x1243181	fyne.io/fyne/v2/layout.hBoxLayout.MinSize+0x101						fyne.io/fyne/v2@v2.5.0/layout/boxlayout.go:203
#	0x11776b0	fyne.io/fyne/v2.(*Container).MinSize+0x50						fyne.io/fyne/v2@v2.5.0/container.go:90
#	0x124254a	fyne.io/fyne/v2/layout.(*borderLayout).MinSize+0x2aa					fyne.io/fyne/v2@v2.5.0/layout/borderlayout.go:91
#	0x11776b0	fyne.io/fyne/v2.(*Container).MinSize+0x50						fyne.io/fyne/v2@v2.5.0/container.go:90
#	0x124489a	fyne.io/fyne/v2/layout.stackLayout.MinSize+0x9a						fyne.io/fyne/v2@v2.5.0/layout/stacklayout.go:47
#	0x11776b0	fyne.io/fyne/v2.(*Container).MinSize+0x50						fyne.io/fyne/v2@v2.5.0/container.go:90
#	0x1242ba1	fyne.io/fyne/v2/layout.vBoxLayout.MinSize+0x101						fyne.io/fyne/v2@v2.5.0/layout/boxlayout.go:120
#	0x11776b0	fyne.io/fyne/v2.(*Container).MinSize+0x50						fyne.io/fyne/v2@v2.5.0/container.go:90
#	0x123a364	fyne.io/fyne/v2/internal/widget.(*SimpleRenderer).MinSize+0x24				fyne.io/fyne/v2@v2.5.0/internal/widget/simple_renderer.go:46
```

`(*Button).CreateRenderer` takes the `propertyLock.RLock()`, and so does `(*BaseWidget).Visible` which is further down the stacktrace.

### How to reproduce

It is a race condition so it will be hard to reproduce without modifying the code to enforce the ordering that causes the deadlock.  It is easy to see from the code that the possibility of deadlock exists.  In my case - the thread waiting on the write lock was trying to call `.Hide()` on the button.

### Screenshots

_No response_

### Example code

No code needed.

### Fyne version

2.5.0

### Go compiler version

1.19.3

Windows 10

### Additional Information

_No response_

williambrode:
### Description:

Fixes #5114

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/button.go
+++ b/widget/button.go
@@ -393,7 +393,7 @@ func (r *buttonRenderer) padding(th fyne.Theme) fyne.Size {
 
 // must be called with r.button.propertyLock RLocked
 func (r *buttonRenderer) updateIconAndText() {
-	if r.button.Icon != nil && r.button.Visible() {
+	if r.button.Icon != nil && !r.button.Hidden {
 		icon := r.button.Icon
 		if r.icon == nil {
 			r.icon = canvas.NewImageFromResource(icon)


dweymouth:
Acquiring a read lock twice should not cause a deadlock - RWMutex allows for multiple read locks at the same time. Are you sure this PR is actually solving a deadlock or have you just not seen it because of timing luck when you try to reproduce it again?

williambrode:
Yes, it will cause deadlock - it's a bit hard for me to explain but this answer seems to do a better job: https://stackoverflow.com/a/30549188/1139197

The TLDR is you can never acquire the same read lock twice in the same goroutine.

dweymouth (approved):

beeblebrox:
@dweymouth these can be tricky to spot. Do you think there might be any other double read locks hidden in the engine? Coincidently I think I hit this deadlock using the demo app yesterday and was coming to report it and why I felt obliged to comment :) However I lost the stack trace and not 100% sure it was the same deadlock.

But here is some clarity if anyone else stumbles upon this and asks why you can't take RLock() twice in the same goroutine without a potential deadlock situation. 

Taking the read lock twice (RLock) in the same routine won't cause issues unless there is another go routine that calls Lock on the same mutex between the multiple RLock's.  Go uses a writer prefer RW lock strategy and in addition will cause Lock() to force all other future RLocks to block until the Lock() obtains the lock and releases it.  

The problem here is with this timeline:
```
go1 RLock()'A - takes the lock
go2 Lock() - blocks waiting for go1 RLock()'A release.
go1 RLock()'B - blocks until go2 Lock() is released.
```
Deadlock because:
1. go1 RLock()'A will never release waiting for go1 RLock()'B to obtain the lock and finish.
2.  go1 RLock()'B always blocked waiting for go2 Lock()/
3. go2 Lock() will always be blocked waiting for RLock()'A

When go2 Lock occurs anywhere else go won't deadlock so this kind of issue can hide for quite awhile.  I suspect that this is only seen in Fyne when someone has implemented a GUI doing some dynamic updates to the button icon's asynchronously with some other mechanisms force a re-render.

Most UI systems get around this issue (and locks in general) requiring all UI property state changes to happen on a single thread, or go routine in the case of go.  The databinding is a nice new feature which I think helps prevents these bugs due to the channels and maybe that can be extended to more properties?

One last thing, I thought maybe `go run -race` might catch this RLock() called twice without the need of a Lock() running at the same time but I just checked and it doesn't seem to. Funny enough though there is another race condition I detected though just when updating a button with an icon, but I'll save that for a different issue and I am unsure if it has any real impact in that case.

williambrode:
@beeblebrox I opened this one last year: https://github.com/fyne-io/fyne/issues/3886 and doing a quick review of the code I see its still an issue.

beeblebrox:
I see.

 It looks like one option as implied in those comments might be to make
each public API simply a grab/defer of the lock the operation required and
call a non public version with a strict rule that it may never call an
exported method and must always call a private one to prevent this error.
The private ones probably should also be named stating if they expect the
write lock instead of the read lock to also prevent unintentional write
races in the stack. I have seen other issues like this already.


I'm guessing without a significant design shift/discipline like above it is
really easy to
Introduce more potential instances of this double+ lock scenario that is
easily missed because you'd have to understand all possible call
hierarchies in the system to know it occurred and -race won't detect this
without actually hitting the scenario with the Lock too.

For now the easier thing to do is to keep reporting the issues we find and
submitting PRs to fix them to help out the team. You're doing well in that
regard. I think what you did here was a good way for this specific case.

What do you think if in future PRs if we can fix the  private functions
that call the exported ones by breaking out the exported function into one
that only calls the lock and defers it and call a new private function that
does the implementation?  We should probably only do this only if it runs
in the double Lock scenario because of a call further up the stack and
incrementally as we see the issue. That way maybe we can slowly improve the
state and hopefully make future changes less fragile if there is no need to
call the exported versions from the private ones any longer. We'd need to
be careful though because this would naturally have the tradeoff of having
less fine grain read locks which could affect performance. I know the team
already had a plan to fix slot of this late 2024, maybe they already have a
best practice or redesign in the locking pattern already? Thoughts?

Fyne seems to be one of the better gui frameworks for go and I hope we, the
community, can help make it better. I am also using Windows 10 and like you
I seem to be getting unlucky and hitting the locks more frequently than
perhaps the team is used to seeing on other platforms.

williambrode:
@beeblebrox as I understand it the fyne team is working on a single-threaded approach which likely will resolve most race conditions.  I wasn't able to wait on this one because it was actively affecting my customers.

andydotxyz:
That's right. More info to come, will be discussed in Fyne Conf 2024 at the latest :)

PR #5118, opened 2024-09-07
widget.SimpleRenderer: use an array instead of a slice

pierrec:
### Description:
The widget.SimpleRenderer only renders one object and was using a slice to hold it (24 bytes).
By using a single item array instead, its size comes down to 16 bytes.

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

### DIFF
--- a/internal/widget/simple_renderer.go
+++ b/internal/widget/simple_renderer.go
@@ -9,7 +9,7 @@ var _ fyne.WidgetRenderer = (*SimpleRenderer)(nil)
 //
 // Since: 2.1
 type SimpleRenderer struct {
-	objects []fyne.CanvasObject
+	objects [1]fyne.CanvasObject
 }
 
 // NewSimpleRenderer creates a new SimpleRenderer to render a widget using a
@@ -17,7 +17,7 @@ type SimpleRenderer struct {
 //
 // Since: 2.1
 func NewSimpleRenderer(object fyne.CanvasObject) *SimpleRenderer {
-	return &SimpleRenderer{[]fyne.CanvasObject{object}}
+	return &SimpleRenderer{[1]fyne.CanvasObject{object}}
 }
 
 // Destroy does nothing in this implementation.
@@ -52,7 +52,7 @@ func (r *SimpleRenderer) MinSize() fyne.Size {
 //
 // Since: 2.1
 func (r *SimpleRenderer) Objects() []fyne.CanvasObject {
-	return r.objects
+	return r.objects[:]
 }
 
 // Refresh requests the underlying object to redraw.


andydotxyz:
Thanks for this suggestion. However the call to `Objects()` happens on a very frequent basis, won't the CPU required constantly converting [1] to [] negate the benefits of saved ram?

pierrec:
I dont htink so but there is only one way to find out... Will add a benchmark.

pierrec:
Here are the benchmark results:

```
func BenchmarkSimpleRenderObjects(b *testing.B) {
	var obj fyne.CanvasObject
	r := widget.NewSimpleRenderer(obj)
	for i := 0; i < b.N; i++ {
		r.Objects()
	}
}
```

```
go test -bench Simple -count 10 > before.txt
goos: darwin
goarch: amd64
pkg: fyne.io/fyne/v2/internal/widget
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3875 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3872 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3871 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3875 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3874 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3876 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3876 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3877 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3872 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3873 ns/op
PASS
ok  	fyne.io/fyne/v2/internal/widget	6.401s
```

```
go test -bench Simple -count 10 > after.txt
goos: darwin
goarch: amd64
pkg: fyne.io/fyne/v2/internal/widget
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3884 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3874 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3872 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3872 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3871 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3879 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3885 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3891 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3892 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         0.3883 ns/op
PASS
ok  	fyne.io/fyne/v2/internal/widget	6.419s
```

```
~/go/bin/benchstat before.txt after.txt
goos: darwin
goarch: amd64
pkg: fyne.io/fyne/v2/internal/widget
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
                       │  before.txt  │           after.txt            │
                       │    sec/op    │    sec/op     vs base          │
SimpleRenderObjects-12   0.3875n ± 0%   0.3881n ± 0%  ~ (p=0.157 n=10)
```

Jacalz:
Those benchmarks are not helping you anything. Because you are not saving the result from the call to Objects(), it is optimising away the entire call. That is why you are seeing abnormal low numbers that are almost as fast as the CPU can execute an instruction.

pierrec:
> Those benchmarks are not helping you prove anything. Because you are not saving the result from the call to Objects(), it is optimising away the entire call. That is why you are seeing abnormal low numbers that are the same before and after as well as being as fast as the CPU can execute a single instruction.
> 
> You need to save the objects slice to a local variable on each iteration and save that variable to a global variable at the end of the benchmark if you want correct results.

Ah yes. No impact still after updating the benchmark.

```
var benchmarkObjects []fyne.CanvasObject

func BenchmarkSimpleRenderObjects(b *testing.B) {
	var obj fyne.CanvasObject
	r := widget.NewSimpleRenderer(obj)
	for i := 0; i < b.N; i++ {
		benchmarkObjects = r.Objects()
	}
}
```

```
cat before.txt
goos: darwin
goarch: amd64
pkg: fyne.io/fyne/v2/internal/widget
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkSimpleRenderObjects-12    	1000000000	         1.163 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.164 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.163 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.164 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.162 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.162 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.163 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.162 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.163 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.162 ns/op
PASS
ok  	fyne.io/fyne/v2/internal/widget	14.959s
```

```
cat after.txt
goos: darwin
goarch: amd64
pkg: fyne.io/fyne/v2/internal/widget
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkSimpleRenderObjects-12    	954153519	         1.162 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.162 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.162 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.163 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.162 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.163 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.163 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.162 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.170 ns/op
BenchmarkSimpleRenderObjects-12    	1000000000	         1.162 ns/op
PASS
ok  	fyne.io/fyne/v2/internal/widget	14.898s
```

```
~/go/bin/benchstat before.txt after.txt
goos: darwin
goarch: amd64
pkg: fyne.io/fyne/v2/internal/widget
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
                       │ before.txt  │           after.txt           │
                       │   sec/op    │   sec/op     vs base          │
SimpleRenderObjects-12   1.163n ± 0%   1.162n ± 0%  ~ (p=0.496 n=10)
```

andydotxyz:
Is this good now @Jacalz? I am no good with benchmarks :)

Jacalz:
I mean, the benchmarks mostly look good but would have been more accurate if they used a local variable in the loop and then assigned that local variable to the global one at the end of the benchmark (right now it is including the overhead of global assignment in each benchmark) but it probably won't make much difference to the results. However, that is likely of no importance.

I'm a bit on the fence about this though. Feels a bit like code churn and extra complexity for the sake of very little to gain for no extra performance? You need an insane amount of renderers for these extra bytes to matter in any memory usage situation.

Jacalz:
For the record: I have a vague recollection of trying to do this optimization sometime before and avoiding it due to being afraid that the conversion to a slice would allocate. It might be possible to gain more insight by looking at the generated assembly in something like https://go.godbolt.org/ or use the linker flag to have Go compiler tell us when allocations are happening (the latter can also be done interactively in VSCode quite nicely). That would likely be a better source of information than a small micro benchmark but I don't know.

pierrec:
I am new to Fyne, so not sure how often the SimpleRenderer is actually used. :)
However this seems like it is not worth it (the benchmark does not show any allocation either (go1.22) but it might on other versions). Going to close this so as not make you waste more time. Thanks for the feedback though.

Jacalz:
Thanks for looking at it and trying to optimize Fyne. Even if this PR wasn't merged, it is still greatly appreciated work. Feel free to look into more areas and ask any questions when the arise. Welcome to Fyne :)

PR #5119, opened 2024-09-07
Proposing a DateEntry widget to input dates as a time.Time

andydotxyz:
Looking for feedback and thoughts on this, more to come but is this a great start?

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

Screenshot 2024-09-07 at 18 30 15


widget/date_entry.go
```diff
@@ -0,0 +1,158 @@
...
+	e.Entry.Resize(size)
+	if e.popUp != nil {
+		e.popUp.Resize(fyne.NewSize(size.Width, e.popUp.Size().Height))
+	}
+}
+
+func (e *DateEntry) SetDate(d *time.Time) {
+	if d == nil {
+		e.Date = nil
+		e.Entry.SetText("")
```
dweymouth, line 124: Should there be a return after this line? (thanks staticcheck for detecting null dereference below)

widget/calendar.go
```diff
@@ -0,0 +1,213 @@
...
+			col++
+		}
+		i++
+	}
+}
+
+// MinSize sets the minimum size for the calendar
+func (g *calendarLayout) MinSize(_ []fyne.CanvasObject) fyne.Size {
+	pad := theme.Padding()
+	largestMin := NewLabel("22").MinSize()
```
dweymouth, line 195: I think this template label can be declared once as a private global, to avoid allocating new widgets every MinSize?

widget/calendar.go
```diff
@@ -0,0 +1,213 @@
...
+		dayIndex += daysPerWeek
+	}
+
+	//add spacers if week doesn't start on Monday
+	for i := 0; i < dayIndex-1; i++ {
+		buttons = append(buttons, layout.NewSpacer())
+	}
+
+	for d := start; d.Month() == start.Month(); d = d.AddDate(0, 0, 1) {
+
```
dweymouth, line 123: blank lines here seem kind of random and not following an existing grouping style

widget/date_entry.go
```diff
@@ -0,0 +1,158 @@
...
+
+	return e.Entry.CreateRenderer()
+}
+
+// Enable this widget, updating any style or features appropriately.
+//
+// Implements: fyne.DisableableWidget
+func (e *DateEntry) Enable() {
+	if e.ActionItem != nil {
+		e.ActionItem.(fyne.Disableable).Enable()
```
dweymouth, line 78: This is a public field, I'm not sure it's safe to always assume it is a fyne.Disableable if it is non-nil?

widget/calendar.go
```diff
@@ -0,0 +1,213 @@
...
+func (c *Calendar) dateForButton(dayNum int) time.Time {
+	oldName, off := c.currentTime.Zone()
+	return time.Date(c.currentTime.Year(), c.currentTime.Month(), dayNum, c.currentTime.Hour(), c.currentTime.Minute(), 0, 0, time.FixedZone(oldName, off)).In(c.currentTime.Location())
+}
+
+func (c *Calendar) daysOfMonth() []fyne.CanvasObject {
+	start := time.Date(c.currentTime.Year(), c.currentTime.Month(), 1, 0, 0, 0, 0, c.currentTime.Location())
+	var buttons []fyne.CanvasObject
+
+	//account for Go time pkg starting on sunday at index 0
```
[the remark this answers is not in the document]
andydotxyz: Correct, the localisation items are not included - so the first day of week and format aren't updated yet

widget/calendar.go
andydotxyz: It was ported in from fyne-x. Should we add a new commit just to remove strange empty lines?

widget/calendar.go
```diff
@@ -0,0 +1,223 @@
...
+
+func (c *Calendar) calendarObjects() []fyne.CanvasObject {
+	var columnHeadings []fyne.CanvasObject
+	for i := 0; i < daysPerWeek; i++ {
+		j := i + 1
+		if j == daysPerWeek {
+			j = 0
+		}
+
+		t := NewLabel(shortDayName(time.Weekday(j).String()))
```
dweymouth, line 96: Does time.Weekday(j).String() always return the English names regardless of locale? Since this is just now being used as a localization key anyway, it may make sense to switch on the value of time.Weekday(j) itself, which may be more robust and easier to reason about

andydotxyz: Yes I looked at the source and that string doesn't seem to be localised.
If we use it directly then the key would be something like "day.1" is that clear enough do you think?

dweymouth: I guess I was confused why we were still indexing the first 3 bytes of a string in shortDayName and thought this might help with that.

andydotxyz: I don't think it is indexing the first 3 - that is the fallback.
So if a language is missing the translation it uses the start of the word...

widget/calendar.go
dweymouth: I thought they can just be cleaned up alongside any commit for any other reason

widget/calendar.go
dweymouth: Sorry, didn't realize lang.X was a 'with fallback' API - I'm used to using lang.L

widget/calendar.go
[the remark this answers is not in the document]
dweymouth: I think we should update that before releasing this in 2.6 - could be a follow-up PR but if we do that let's file a blocker ticket to make sure it gets addressed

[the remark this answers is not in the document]
andydotxyz: Yea, a new PR sounds good as it will require extensions to the localisation API and/or capabilities. #5141

andydotxyz:
I think that is all (except the extra localisation work, linked) @dweymouth

dweymouth:
And filed https://github.com/fyne-io/fyne/issues/5142 to make sure we don't release without addressing the follow-up locale work

dweymouth:
Whoops, closed as duplicate of the one you already filed

dweymouth (approved):

PR #5121, opened 2024-09-08
Decouple the Clipboard API from fyne.Window

### LINKED ISSUE #4418: Decouple clipboard from fyne.Window
### Checklist

The clipboard in GLFW was previously tied to the window. This is no longer the case so I think it seems strange to have a `window.Clipboard()` method when clipboards generally are per app or per system and not per window.

See https://github.com/fyne-io/fyne/pull/4417 for step one.

Yeah. We can move the clipboard struct anywhere we'd like.

Add a `app.Clipboard()` method and deprecate the one on the window.

pierrec:
### Description:

This PR implements the changes for #4418 by:
-  updating the fyne.App, fyne.Driver and fyne.Window interfaces
- migrating tests and cmd programs using the clipboard to the new interface

Fixes #4418

### Checklist:

- [ ] Tests included.
  I dont see any automated tests for this but maybe I missed them. I did test manually and ctrl-c, ctrl-v and ctrl-x work as expected on the desktop.
- [ ] Tests all pass.
I get quite a few errors on my system, some related to the locale (labels are emitted in French while tests expect English), others to something else which I dont know about.
They all seem unrelated to this change though.

#### Where applicable:

  (no new module imported)


cmd/fyne_demo/main.go
```diff
@@ -251,14 +251,14 @@ func makeNav(setTutorial func(tutorial tutorials.Tutorial), loadPrevious bool) f
 	return container.NewBorder(nil, themes, nil, nil, tree)
 }
 
-func shortcutFocused(s fyne.Shortcut, w fyne.Window) {
+func shortcutFocused(s fyne.Shortcut, a fyne.App, w fyne.Window) {
```
Jacalz, line 254: I wonder if it perhaps makes more sense to pass in the focused object now that we don't need the window for other purposes?

Jacalz (commented):
Nice work. I changed the base branch to develop and that made the changes look a lot more sensible. Just added a small comment about that we perhaps should consider changing a function signature (if you agree with me there) :)

cmd/fyne_demo/main.go
pierrec: Sure, just did a mechanical fix, patch sent.

driver.go
```diff
@@ -43,4 +43,9 @@ type Driver interface {
 	//
 	// Since: 2.5
 	SetDisableScreenBlanking(bool)
+
+	// Clipboard returns the system clipboard.
+	//
+	// Since: 2.6
+	Clipboard() Clipboard
```
dweymouth, line 50: This PR exposes the clipboard as a new public API on both the App and the Driver. I think we should just choose one location for the new API - I think the app makes more sense but open to suggestions

dweymouth (changes requested):

driver.go
Jacalz: Good point. I agree that the app makes more sense.

internal/driver/glfw/window.go
```diff
@@ -237,7 +237,7 @@ func (w *window) ShowAndRun() {
 
 // Clipboard returns the system clipboard
 func (w *window) Clipboard() fyne.Clipboard {
-	return &clipboard{}
```
dweymouth, line 240: Rather than allocating a new clipboard, should we just return the singleton instance from the app here?

Jacalz:
@pierrec Please don't force push your changes very time. Descriptive commits with are lot better as we more easily can see what has changed and GitHub also has a tendency to sometimes lose track of old review comments. We'll squash on merge if needed.

pierrec:
> @pierrec Please don't force push your changes very time. Descriptive commits with are lot better as we more easily can see what has changed and GitHub also has a tendency to sometimes lose track of old review comments. We'll squash on merge if needed.

OK, so sorry, will just push next time.

pierrec:
Question while working through the CI failures (a lot of them due to my lack of knowledge of the Fyne codebase and all the build flags!):
what to do with the `TestWindow_Clipboard` and `TestWindow_ClipboardCopy_DisabledEntry` tests in `internal/driver/glfw/window_test.go`? Should they be moved to `app/app_test.go`?

internal/driver/mobile/driver.go
```diff
@@ -100,6 +100,10 @@ func (d *driver) currentWindow() *window {
 	return last
 }
 
+func (d *driver) Clipboard() fyne.Clipboard {
+	return &mobileClipboard{}
```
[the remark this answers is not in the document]
Jacalz: I would rather want to not use singletons honestly. It is a poor coding standard and I don't think the optimisation actually helping anything. We now have a global variable that is always on the heap while newly allocated object might be stack allocated by the compiler as it sees that it doesn't escape.

app/app_test.go
```diff
@@ -48,3 +48,26 @@ func TestFyneApp_SetIcon(t *testing.T) {
 
 	assert.Equal(t, setIcon, app.Icon())
 }
+
+func TestFynaApp_Clipboard(t *testing.T) {
+	app := NewWithID("io.fyne.test")
+	app.NewWindow("test")
```
Jacalz, line 54: I haven't tried but I suspect you might want to use `test.NewTempApp` and `test.NewTempWindow`?

pierrec: Both actually work, but I have updated the test with your suggestion.

andydotxyz: Both will work, but the ones Jacob suggested clean up at the end of the test for you :)

andydotxyz:
I think there may have been a confusion about word usage here - so I want to ensure we are on the same page.

> Rather than allocating a new clipboard, should we just return the singleton instance from the app here?

I think that @dweymouth was suggesting that the clipboard exist inside the app instance instead of inside the window where it was before. That is neither a true singleton, but exists only once per application. It will be much more efficient than the current implementation which seems to allocate a new clipboard for every situation where it may be used.

Jacalz:
> It will be much more efficient than the current implementation which seems to allocate a new clipboard for every situation where it may be used.

I'm really not sure that it is a correct assumption. All of our clipboard structs are zero size as I can remember and don't actually contain any information so allocating extra size for the interface is unnecessary? We can probably just avoid returning a pointer to the struct and help it not allocate that way.

I don't think it is worth making the app struct larger just for an arbitrary optimisation that we don't if it is worth it or not. Getting the clipboard is likely not a hot path in any application and if anyone cares, they can just save it to a variable themselves and be happy there.

Jacalz:
Sorry for the long turnaround time for me reviewing this. I've had quite a busy first weeks with the first weeks of my masters program. I'll try to have a look one of the coming days.

pierrec:
> Sorry for the long turnaround time for me reviewing this. I've had quite a busy first weeks with the first weeks of my masters program. I'll try to have a look one of the coming days.

Thank you for the update, there is no rush :).

Jacalz (changes requested):
Sorry for the very long wait. Just one minor change needed and we're golden :)

internal/driver/mobile/driver.go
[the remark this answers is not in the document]
pierrec: This one slipped through. Will do.

Jacalz:
The static analysis test failure seemed a bit strange. Kicked it to see if it is temporary.

pierrec:
> The static analysis test failure seemed a bit strange. Kicked it to see if it is temporary.

I had a look but I dont even have that statement on line 406 of the file; grepping doesnt return any instance of it either.

Jacalz:
Hmm, interesting. I've never really seen any flukes before there; Staticcheck is about a solid as it gets in static analysis tools for Go. I'll have a look locally when I get back after a short weekend-holiday here and see what the problem can be :)

pierrec:
> Hmm, interesting. I've never really seen any flukes before there; Staticcheck is about a solid as it gets in static analysis tools for Go. I'll have a look locally when I get back after a short weekend-holiday here and see what the problem can be :)

Enjoy your break!

andydotxyz:
Perhaps there was a merge and an out-of-order run. kicked it to try again.

Jacalz:
Same error strangely enough. I've opened https://github.com/fyne-io/fyne/pull/5175 for now to hopefully get the ball rolling again :)

Jacalz:
Try rebasing on develop now :)

pierrec:
Still erroring out. At this point I might close this PR and make another one.

Jacalz:
Let me just check out your branch locally to see if there is anything strange going on there before you do anything :)

Jacalz:
It does indeed still error out but the line it points to happens to be a perfect valid error that easily can be fixed.

Jacalz (approved):
Thanks for working on this. I'm really sorry for the long turnaround time and all the problems that popped up. Hopefully it will be smooth sailing for any future contributions :)

pierrec:
Thanks for the review!

andydotxyz (changes requested):
Questioning some removed code.

The test failure looks like a reduction in code coverage - It isn't an obvious lack in this PR but I wonder if somehow it has redirected code to be less well tested. The Coveralls report should lend a hint.

cmd/fyne_demo/main.go
```diff
@@ -251,16 +251,8 @@ func makeNav(setTutorial func(tutorial tutorials.Tutorial), loadPrevious bool) f
 	return container.NewBorder(nil, themes, nil, nil, tree)
 }
 
-func shortcutFocused(s fyne.Shortcut, w fyne.Window) {
-	switch sh := s.(type) {
-	case *fyne.ShortcutCopy:
-		sh.Clipboard = w.Clipboard()
```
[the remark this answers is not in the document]
Jacalz: From what I can tell from the code, it was always setting the clipboard twice. First when creating the struct and then this function made a type switch and set it again?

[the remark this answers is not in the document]
pierrec: @andydotxyz the shortcut is already setup with the clipboard at instanciation. No need to do it a second time right?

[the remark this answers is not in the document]
andydotxyz: > the shortcut is already setup with the clipboard at instanciation.

In the instantiation you make, yes - but check other usages too - like in Entry...

`e.shortcut.AddShortcut(&fyne.ShortcutCopy{}, ...)`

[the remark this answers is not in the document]
pierrec: I see. It may be better to get ShortcutCopy/Cut/Paste from the App though instead of relying on the user to properly instantiate it. In the meantime, I will restore the check in the fyne_demo app.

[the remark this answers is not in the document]
andydotxyz: > It may be better to get ShortcutCopy/Cut/Paste from the App though instead of relying on the user to properly instantiate it

I agree with this statement, however having the fields be `nil` would be a breaking change, so thanks for re-instating it.

andydotxyz (approved):

andydotxyz:
Excellent thanks so much, I think we just need the approval of changes from @dweymouth now too. The coverage issue persists but without it going to coveralls I cannot see what the issue is. I'm happy that you have added tests so I can merge this after changes are approved and we can continue to investigate the 63% coverage failure.

dweymouth (approved):

PR #5122, opened 2024-09-08
widget.Table: show first row/column when navigating with keyboard

### LINKED ISSUE #5088: Table Keyboard Navigation with Header Row Hides First Data Row
### Checklist

### Describe the bug

I created a Table widget and included a header row.  I filled the row with enough data to allow the table to have a scroll bar.

When I use the keyboard to navigate up and down, before scrolling, the "current focus" cell moves correctly up to row 0.  However, if you scroll with the keyboard such that the table scrolls down, then scroll back to the top, the "current focus" row will move what seems like under the header and will no longer be visible.  It means the top row (row 0) will never be viewable in this table view without using the mouse.

### How to reproduce

1. Create a table with a header, and fill the cells with simple data.
2. Click on a cell to allow keyboard navigation.
3. Press the "up" key.  The active cell correctly remains visible even when the up key is pressed with row zero active.
4. Scroll down with the "down" key such that the displayed cells scroll.
5. Press the "up" key to return to the top.  The displayed active cell will disappear, and the top row does not show up, as though it's hidden under the table header.

### Screenshots

TableKeyboardScroll

### Example code

```
package main

import (
	"strconv"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Table Scroll")

	t := widget.NewTable(
		func() (rows int, cols int) { return 1000, 1 },
		func() fyne.CanvasObject { return widget.NewLabel("000") },
		func(tci widget.TableCellID, co fyne.CanvasObject) {
			co.(*widget.Label).SetText(strconv.FormatInt(int64(tci.Row), 10))
		},
	)
	t.ShowHeaderRow = true
	t.CreateHeader = func() fyne.CanvasObject { return widget.NewLabel("Row") }
	t.UpdateHeader = func(id widget.TableCellID, template fyne.CanvasObject) {}
	w.SetContent(container.NewStack(t))
	w.Resize(fyne.Size{Width: 100, Height: 250})

	w.ShowAndRun()
}
```

### Fyne version

2.5.0

### Go compiler version

1.22.6

Arch Linux

### Additional Information

_No response_

pierrec:
Fixes #5088.

### Description:

When a Table has its header(s) shown, keyboard navigation fails to show the first row or column
if the list was scrolled. This is because the header size is always added to the offset even when 
the first row/column is reached. Disabling this for the first row or column fixes the issue.

Fixes #5088

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

### DIFF
--- a/widget/table.go
+++ b/widget/table.go
@@ -474,7 +474,6 @@ func (t *Table) ScrollTo(id TableCellID) {
 		cellX -= t.stuckXOff + t.stuckWidth
 	}
 	if t.ShowHeaderColumn {
-		cellX += t.headerSize.Width
 		stickCols--
 	}
 	if stickCols == 0 || id.Col > stickCols {
@@ -491,7 +490,6 @@ func (t *Table) ScrollTo(id TableCellID) {
 		cellY -= t.stuckYOff + t.stuckHeight
 	}
 	if t.ShowHeaderRow {
-		cellY += t.headerSize.Height
 		stickRows--
 	}
 	if stickRows == 0 || id.Row >= stickRows {
--- a/widget/table.go
+++ b/widget/table.go
@@ -370,39 +370,31 @@ func (t *Table) TypedKey(event *fyne.KeyEvent) {
 	case fyne.KeyDown:
 		if f := t.Length; f != nil {
 			rows, _ := f()
-			if t.currentFocus.Row >= rows-1 {
-				return
+			if t.currentFocus.Row+1 < rows {
+				t.currentFocus.Row++
 			}
 		}
-		t.RefreshItem(t.currentFocus)
-		t.currentFocus.Row++
 		t.ScrollTo(t.currentFocus)
 		t.RefreshItem(t.currentFocus)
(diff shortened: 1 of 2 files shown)


andydotxyz (changes requested):
Thanks so much for fixing this and adding tests.

Unfortunately it only works for when the ID for row or col is 0. 
But also we should ensure the focused cell highlight is visible when the cell index is 1 or above, otherwise the indicator hides behind the header until you reach the first row/col

pierrec:
I tested by moving too fast, I see the issue now. Will get back to it.

andydotxyz:
@pierrec can you look at these failures? Because of the force-push it has lost the build history so it is not possible to see where it broke.
Please do avoid force-push after reviews have started so we get more info on what is changing.

pierrec:
@andydotxyz closing this PR and resubmitting on lastest develop branch.

PR #5123, opened 2024-09-08
Add string binding to the Selected field of the Select widget

jimorc:
### Description:
Adds string binding to the Selected field of the Select widget.

Partial implementation for issue 2836. Binding of the Options field is outside the current scope of the binding implementation. A couple of suggestions were made:
1. Implement functionality in fyne-x.
2. Discuss BindXxx, e.g. BindSelected, BindOptions, but this is a discussion I would much prefer to leave to the main fyne developers as this may result in breaking changes.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.
- [x] Formatter runs with no errors. Formatting actually done by my IDE (VSCode).
- [x] golang-ci run with no additional errors over those already produced in develop branch

#### Where applicable:


andydotxyz (changes requested):
This looks good thanks so much, but the version numbers are rather old

jimorc:
> This looks good thanks so much, but the version numbers are rather old

I assume you mean the "Since 2.0" lines. If yes, what should it be: 2.5? 2.6? IF not, then I don't know what version numbers you are referring to.

andydotxyz:
> > This looks good thanks so much, but the version numbers are rather old
> 
> I assume you mean the "Since 2.0" lines. If yes, what should it be: 2.5? 2.6? IF not, then I don't know what version numbers you are referring to.

That's correct, it should be 2.6 - there should be a comment inline in the PR review.

andydotxyz (approved):
Great, thanks very much!

PR #5138, opened 2024-09-12
Remember the folder that file dialogs use

andydotxyz:
### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -895,7 +895,7 @@ func getFavoriteOrder() []string {
 	}
 
 	if runtime.GOOS == "darwin" {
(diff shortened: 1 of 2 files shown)


dialog/file.go
```diff
@@ -619,6 +623,18 @@ func (f *FileDialog) effectiveStartingDir() fyne.ListableURI {
 
 	}
 
+	// last used
```
dweymouth, line 626: Let's add 'Last used' to the order or preference doc comment for this function

dweymouth (approved):

PR #5144, opened 2024-09-13
added a search bar to the file dialog

billyb2:
the search bar takes advantage of the existing filtering system in order to only show files or filders that match a specific prompt. for now, there is no way to recursively search through files, or use something like regex, though that could be added as options later.

the main internal change needed is, in order to be backwards compatible, having a list of filters instead of just a single filter at a time. that way, library users can still have many filters while including the search bar.

 

### Description:

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


billyb2:
Still have yet to add tests to this, so I haven't put it up for proper review yet. I don't expect to change much about the implementation though, unless the fyne team want changes :)

andydotxyz:
Yeah apologies I slipped and tapped the wrong button when trying to kick off tests.
I have a backlog of reviews so won't get to this immediately.

billyb2:
> Yeah apologies I slipped and tapped the wrong button when trying to kick off tests. I have a backlog of reviews so won't get to this immediately.

no worries :). any time reviewing this is great, i appreciate the amazing work yall are doing.

andydotxyz:
> Still have yet to add tests to this, so I haven't put it up for proper review yet.

Before adding more tests better fix the ones that broke ;)

PR #5146, opened 2024-09-14
 cache: use one lock/unclock cycle to remove expired items (Issue5129)

### LINKED ISSUE #5129: Race condition in renderer cache can lead to widget being rendered with destroyed renderer
### Checklist

### Describe the bug

Thanks to @knusbaum for discovering this bug

The [implementation](https://github.com/fyne-io/fyne/blob/master/internal/cache/base.go#L184) of the widget renderer cache cleaning task acquires a lock on the renderer cache map, identifies the expired renderers and calls Destroy() on each of them, **and then releases the lock** before it grabs it again to actually remove the expired renderers from the map. This leads to a race condition where the renderer for a widget could be looked up after Destroy has been called but before it was removed from the map.

A possible fix would be to make the clean task atomic, so it holds a lock for the entirety of the clean task, but we need to make sure this can't introduce deadlocks.

### How to reproduce

it's a race condition, hard to reproduce, but it is clear from looking at the code

### Screenshots

_No response_

### Example code

see implementation linked in the bug description

### Fyne version

develop

### Go compiler version

n/a

n/a

### Additional Information

_No response_

pierrec:
### Description:

Previous implementation collected expired items while under an Rlock then proceeded to
remove them from the map they sat in while under a Lock. A race can occur while the map
is changed or accessed between the RUnlock and the Lock.
Deleting a map item while iterating it is actually fine, the only issue might be if you delete 
an item not yet seen by the loop. This is never the case here as we delete the item being
iterated. Therefore, we can simply iterate the map and remove expired items right away while
being protected by a Lock.

Fixes #5129.

### DIFF
--- a/internal/cache/base.go
+++ b/internal/cache/base.go
@@ -12,7 +12,6 @@ var (
 	cacheDuration     = 1 * time.Minute
 	cleanTaskInterval = cacheDuration / 2
 
-	expiredObjects                = make([]fyne.CanvasObject, 0, 50)
 	lastClean                     time.Time
 	skippedCleanWithCanvasRefresh = false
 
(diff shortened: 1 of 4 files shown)


andydotxyz:
Please use more descriptive PR titles so they can be understood more quickly.
Also make sure the base in GitHub is the same as the base you branch from (develop).

andydotxyz (approved):
Looks good from a code review point. I'll need to test it (not on a computer right now) but if anyone else gets there first feel free to merge.

PR #5149, opened 2024-09-17
added SetItemHeight to the Tree widget, copying from the List widget

kothawoc:
### Description:

This adds `SetItemHeight` to the Tree widget. It is copied from the List widget to keep it similar.

I require dynamic heights for items in the tree, which wasn't supported, this implements them using the same public API as the List widget.

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/tree.go
+++ b/widget/tree.go
@@ -48,6 +48,7 @@ type Tree struct {
 	branchMinSize fyne.Size
 	currentFocus  TreeNodeID
 	focused       bool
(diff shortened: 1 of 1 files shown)


andydotxyz:
@kothawoc are you still looking to take this forward? it would be a great addition for v2.6.0 but it's currently missing tests.

PR #5151, opened 2024-09-17
internal/driver/mobile: return storage.ErrNotExists instead of custom error

### LINKED ISSUE #3353: app.Storage.Open() to return a known error when file not exist

pierrec:
### Description:

Use a standard error instead of a custom one on Android.

Fixes #3353.

### Checklist:
Did not add any test as the change is really trivial.

### DIFF
--- a/internal/driver/mobile/file_android.go
+++ b/internal/driver/mobile/file_android.go
@@ -25,6 +25,7 @@ import (
 
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/internal/driver/mobile/app"
+	"fyne.io/fyne/v2/storage"
 	"fyne.io/fyne/v2/storage/repository"
 )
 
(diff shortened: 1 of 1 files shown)


pierrec:
Hmm not sure why github insists of putting it into master...

Jacalz (approved):
Thanks. Nice catch :)

PR #5152, opened 2024-09-18
widget: make Button trigger unfocus when tapped.

### LINKED ISSUE #5107: Entry don't get focus on tap after entry.FocusLost()
### Checklist

### Describe the bug

I have button and entry.
When i tap any button, entry do not loose focus.
When i tap any label, entry loose focus.
If i use entry.FocusLost() in button handler or entry.onSubmitted, entry loose focus, but if i tap back into entry, it can't get focus.
If i tap on some label, and then into entry, it get focus. 

### How to reproduce

You can clone and run:
https://github.com/antongumenov/tong_simulator.git

Or look code here:
https://github.com/antongumenov/tong_simulator/blob/main/internal/view/main_screen_components/dump_level_widget.go

### Screenshots

_No response_

### Example code

	entry.OnSubmitted = func(s string) {
		if s != "" {
			val, _ := strconv.ParseInt(s, 10, 64)
			e.ctrl.SetDumpLevel(val)
		}
		entry.FocusLost()
	}
	btn := widget.NewButton("SET MAX TORQUE", func() {
		if entry.Text != "" {
			val, _ := strconv.ParseInt(entry.Text, 10, 64)
			e.ctrl.SetDumpLevel(val)
		}
		entry.FocusLost()
	})

### Fyne version

v2.5.1

### Go compiler version

go1.23.0 linux/amd64

Arch Linux

### Additional Information

_No response_

pierrec:
Fixes #5107.

### Description:
Previously, widget.Button did not grab the focus when tapped, resulting in any focused widget on the canvas remaining focused. Now it does gab the focus and then releases it as this should be transient for a button.

Fixes #5107.

### Checklist:

- [X] Tests included.
- [X] Tests all pass. 
  (besides language related ones)

### DIFF
--- a/widget/button.go
+++ b/widget/button.go
@@ -191,6 +191,12 @@ func (b *Button) Tapped(*fyne.PointEvent) {
 	b.tapAnimation()
 	b.Refresh()
 
+	if !b.focused {
+		// Grab the focus to unfocus any previous widget, then release it.
+		focusIfNotMobile(b.super())
+		b.FocusLost()
+	}
+
 	if onTapped := b.OnTapped; onTapped != nil {
 		onTapped()
 	}
--- a/widget/button_internal_test.go
+++ b/widget/button_internal_test.go
@@ -235,3 +235,47 @@ func TestButtonRenderer_TapAnimation(t *testing.T) {
 	button.tapAnim.Tick(0.5)
 	test.AssertImageMatches(t, "button/tap_animation.png", w.Canvas().Capture())
 }
+
+func TestButton_TappedFocus(t *testing.T) {
+	test.NewApp()
+	entry := NewEntry()
+	button := NewButton("ok", nil)
+	w := test.NewTempWindow(t, newCont(entry, button))
+
+	c := w.Canvas()
+	c.Focus(entry)
+	assert.True(t, entry.focused, "entry is not focused")
+
+	// NB. we cannot use test.Tap(button) as it handles focus,
+	// and we want to specifically test Button's handling of focus.
+	button.Tapped(&fyne.PointEvent{})
+	assert.False(t, entry.focused, "entry still has focus")
+	assert.False(t, button.focused, "button still has focus")
+}
+
+type cont struct {
+	BaseWidget
+	objects []fyne.CanvasObject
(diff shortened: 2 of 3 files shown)


andydotxyz:
That's not how other widgets behave. If someone wants to switch to keyboard handling then it should start from the last item they interacted with.

This may relate to a request for "remember potential focus in case keyboard controls are used" which would be a behaviour change and would have to apply to most widgets (not Entry ;) ) in a major release like 2.6.

pierrec:
> That's not how other widgets behave. If someone wants to switch to keyboard handling then it should start from the last item they interacted with.
> 
> This may relate to a request for "remember potential focus in case keyboard controls are used" which would be a behaviour change and would have to apply to most widgets (not Entry ;) ) in a major release like 2.6.

So you are saying that clicking the button should not unfocus anything?

andydotxyz:
That's not what I am saying.
To be consistent with other widgets when you interact with a widget it would take focus, but it would not immediately lose focus.

pierrec:
I see. The reason I removed focus was that otherwise the button appears in focus mode (blue background with the default theme) once clicked. Is this expected?

andydotxyz:
That is how other widgets behave isn't it?
I agree it doesn't look quite right - which is why I mentioned a discussion that had happened regarding "potentially focused" or "invisible focus" in case the keyboard handing is invoked.

Jacalz:
I think this might be a sensible change. It seems like many applications on my computer at least do unfocus when you press a button.

andydotxyz:
Rather than considering just the visual state we need to consider the context. When another toolkit "unfocuses" the button after tapping with a mouse what happens if you hit space or enter? What happens if you hit tab?

In my experience the item still has focus it just doesn't indicate that for this situation.

pierrec:
So would only removing the visuals of the focus for a clicked button but keepig its focus be fine?

andydotxyz:
There was a simpler solution, now landed.
There is still an interesting discussion around the "next to focus" that should be explored for the desktop where mouse should not lead to indicators but that keyboard focus should pick up from where mouse left off.

PR #5165, opened 2024-09-27
Check with display PoC

flimzy:
### Description:

Proof-of-concept for adding optional display text to Check widgets, following Slack discussion.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
It's not clear here what `Display` adds here over `Text`. Both are just putting a string representation for describing the item being turned on/off

PR #5168, opened 2024-09-27
Add command to scan for new translation

sdassow:
### Description:

Adds a translate command that scans for new translation and updates the translations file.

No new external modules are imported. 
The resulting size increase for the fyne command binary is 830kb on this version of macOS:
```
$ ls -al fyne*
-rwxr-xr-x  1 simon  staff  17764240 Nov  2 13:48 fyne
-rwxr-xr-x  1 simon  staff  16913632 Nov  2 13:47 fyne.develop
```

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

- [X] No breaking changes (new command).


andydotxyz (changes requested):
Can you please factor code out of the 'Translate' method? 120 lines seems very long for something that could be broken down. That would also facilitate adding a test :).

Lastly do note to check the boxes on your PR template so we can see where it's at

cmd/fyne/internal/commands/translate.go
```diff
@@ -0,0 +1,374 @@
...
+			// without any argument find all .go files in the source directory
+			if len(files) == 0 {
+				sources, err := findFilesExt(sourceDir, ".go")
+				if err != nil {
+					return err
+				}
+				files = sources
+			}
+
+			// update the translation file by scanning the given source files
```
andydotxyz, line 72: We prefer to omit inline documentation for the most part - here it is clear from the factored our function name what is happening

cmd/fyne/internal/commands/translate.go
```diff
@@ -0,0 +1,374 @@
...
+
+	// only try to parse translations that exist
+	if f != nil {
+		dec := json.NewDecoder(f)
+		if err := dec.Decode(&translations); err != nil {
+			return err
+		}
+	}
+
+	//if ctx.Bool("verbose") {
```
andydotxyz, line 122: This looks like commented out code. Either way it's not adding anything to the clear line below

cmd/fyne/internal/commands/translate.go
```diff
@@ -0,0 +1,374 @@
...
+// Create or add to translations file by scanning the given files for translation calls
+func updateTranslationsFile(opts *translateOpts, file string, files []string) error {
+	translations := make(map[string]interface{})
+
+	// try to get current translations first
+	f, err := os.Open(file)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
+	defer f.Close()
```
andydotxyz, line 112: This line will be reached when the file does not exist - seems like it should not be running.

cmd/fyne/internal/commands/translate.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+
+	// serialize in readable format for humas to change
+	b, err := json.MarshalIndent(translations, "", "\t")
+	if err != nil {
+		return err
+	}
+
+	// avoid writing an empty file
+	if len(translations) == 0 {
```
andydotxyz, line 139: Should this happen before json encoding?

cmd/fyne/internal/commands/translate.go
```diff
@@ -0,0 +1,374 @@
...
+		fmt.Fprintln(os.Stderr, "No translations found")
+		return nil
+	}
+
+	// stop without making any changes
+	if opts.DryRun {
```
andydotxyz, line 145: We are not doing anything with the json - is this check too late?

cmd/fyne/internal/commands/translate.go
```diff
@@ -0,0 +1,374 @@
...
+	// support writing to stdout
+	if file == "-" {
+		fmt.Printf("%s\n", string(b))
+		return nil
+	}
+
+	return writeTranslationsFile(b, file, f)
+}
+
+// Write data to given file, optionally using same permissions as the original file
```
andydotxyz, line 158: Optionally doesn't quite describe what this function is doing with permissions.

andydotxyz (changes requested):
Found a few logic glitches which I commented inline.

However the biggest comment would be that we avoid code comments unless really required. More often than not an inline comment shows a lack of clarity in the code.

cmd/fyne/internal/commands/translate.go
sdassow: It is meant to keep going even if the file doesn't exist. The idea is to make it work without having to create a translation file first. Need to make sure this is mentioned in the documentation, thanks.

cmd/fyne/internal/commands/translate.go
sdassow: Yes, old code I forgot to remove.

cmd/fyne/internal/commands/translate.go
sdassow: Another way to look at it: we want to reach this line when there is an error other than the file not existing (something fatal).

cmd/fyne/internal/commands/translate.go
sdassow: Yes, good catch.

cmd/fyne/internal/commands/translate.go
sdassow: I'm actually going to move it down a bit more so it just prevents file modifications and otherwise still can output to Stdout.

cmd/fyne/internal/commands/translate.go
andydotxyz: Sorry I missed this comment. As far as I can tell this code will execute for one of two reasons:
a) The file opened OK (no error)
b) the file could not be opened (does not exist).

For b) we should not be closing the file! I think all that needs to happen is that the defer moves into the `if` below

andydotxyz (changes requested):
This looks pretty epic. Just continuing a comment from before.

cmd/fyne/internal/commands/translate.go
sdassow: Oh, I think I might have misunderstood. Yes, moving the defer is cleaner..

andydotxyz:
Those Windows errors seem to indicate you haven't closed a file before trying to remove it.

sdassow:
> Those Windows errors seem to indicate you haven't closed a file before trying to remove it.

Sort of like that. The problem was doing chdir into the temp directory in the test as it prevents removal.

sdassow:
> This looks pretty epic. Just continuing a comment from before.

Did I miss the comment or was it addressed? I'm not sure at the moment.

andydotxyz:
Thanks so much, I tried this out but I think it doesn't create the folder if it doesn't exist.

```
[sdassow-add-translate-command] ><> go run ../fyne translate *.go   
open translations/en.json-182321854: no such file or directory
exit status 1
```

My first attempt was to pass a directory but it didn't work - should it? :)

Last comment. Our translations were 2 spaces, weblate default is 4 spaces and this is tabs.
It would be good to get this and our internal files consistent at least - so we should agree on the approach.

sdassow:
> Thanks so much, I tried this out but I think it doesn't create the folder if it doesn't exist.
> 
> ```
> [sdassow-add-translate-command] ><> go run ../fyne translate *.go   
> open translations/en.json-182321854: no such file or directory
> exit status 1
> ```

That's not a nice experience. Changed it.
The first argument is now mandatory instead and specifies the file to use to avoid any confusion.
When the first argument is missing it shows this:
```
Missing argument: translationsFile
usage: fyne translate [command options] translationsFile [source ...]
```


> My first attempt was to pass a directory but it didn't work - should it? :)
> 

Absolutely, and it does now :)


> Last comment. Our translations were 2 spaces, weblate default is 4 spaces and this is tabs. It would be good to get this and our internal files consistent at least - so we should agree on the approach.

Since we use weblate I've changed it to 4 spaces.

sdassow:
Additionally it's now checking the first argument:
```
$ ./fyne translate
Missing argument: translationsFile
usage: fyne translate [command options] translationsFile [source ...]
$ ./fyne translate *.go
Need .json extension for translationsFile
usage: fyne translate [command options] translationsFile [source ...]
$
```

sdassow:
> Additionally it's now checking the first argument:
> 
> ```
> $ ./fyne translate
> Missing argument: translationsFile
> usage: fyne translate [command options] translationsFile [source ...]
> $ ./fyne translate *.go
> Need .json extension for translationsFile
> usage: fyne translate [command options] translationsFile [source ...]
> $
> ```

Why make the translations file a mandatory argument you ask?
 - works regardless of the current directory or sub-directories
 - makes no assumptions about the default language, leaving the decision to the developer

andydotxyz:
Apologies for the delay. I have tried this latest version but can't seem to figure it out.
In the Fyne repo I removed one of the English translation lines and ran the command:

```
go run ./cmd/fyne translate --dry-run lang/translations/base.en.json .
```

But I see no output at all - any idea what I am doing wrong?

When I ran it without --dry-run it re-added the translation as expected.

sdassow:
> ```
> go run ./cmd/fyne translate --dry-run lang/translations/base.en.json .
> ```
> 
> But I see no output at all - any idea what I am doing wrong?

You're doing it right. I think `--dry-run` is useless now, and definitely pointless without `--dry-run`.
When the translation file name was implicit it seemed useful, but with the latest changes it doesn't.
I'm going to remove it - if someone wants to see changes to existing files they can use a different filename and check the delta. Or use version control.

andydotxyz (approved):
Nice one thanks.

One last question... should it be in the new tools repo?
If we are going to push to move there this should at least be copied over - but is maintaining it here as well a little unneeded effort?

Feel free to land then pick over if you think that's the right approach. Or add this code to tools and we can get a quick review on that PR instead.

PR #5169, opened 2024-09-28
Update to the latest breaking API changes in go-text

andydotxyz:
As happens sometimes the repo had an API break and so people upgrading all their deps get compile errors.
This resolves this and will be picked to release branch as well.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (approved):
Gosh how I wish people would stop doing bad stuff :/

Anyway, nice change and hopefully no problems going forward then :)

andydotxyz:
> Gosh how I wish people would stop doing bad stuff :/
> 
> Anyway, nice change and hopefully no problems going forward then :)

Yeah. Though isn't that what we said last time?

I wonder if we should go ahead and fork so we can manage in the unfortunate situation there are ever future breaks?

PR #5171, opened 2024-09-29
Improve documentation for file dialogs

ErikKalkoken:
### Description:

When using file dialogs it is currently unclear from the documentation whether the reader/writer passed in the callbacks need to be closed. This may lead to obscure bugs, since the file is often still read/created. I therefore like to propose extending the comments to clarify that those in fact must be closed by the callback.

I further like to propose adding variable names in the callbacks to make it clearer to the user what they are (e.g. those will be added to the users code though auto-complete). 

See also related conversation with Andy on [Discord](https://discord.com/channels/953020122690359327/953030515152281650/1289557458804805694).

### Checklist:

- [ ] Tests included. (N/A)
- [x] Tests all pass.

#### Where applicable:

N/A

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -831,33 +831,40 @@ func (f *FileDialog) SetView(v ViewLayout) {
 }
 
 // NewFileOpen creates a file dialog allowing the user to choose a file to open.
-// The callback function will run when the dialog closes. The URI will be nil
-// when the user cancels or when nothing is selected.
+//
+// The callback function will run when the dialog closes and provide a reader for the chosen file.
+// The reader must be closed by the callback.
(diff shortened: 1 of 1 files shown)


andydotxyz (changes requested):
Thanks for this. Probably important to note that the reader should only be closed when it is not nil - perhaps changing the order of the last two lines in each block?

Also please fix the grammar of the last one - the comma is not wanted in that position.

ErikKalkoken:
> Thanks for this. Probably important to note that the reader should only be closed when it is not nil - perhaps changing the order of the last two lines in each block?
> 
> Also please fix the grammar of the last one - the comma is not wanted in that position.

Thanks for your feedback. I incorporated both of your points.

PR #5176, opened 2024-09-30
Add storage API to open a file for appending, without truncating it's content first

mbaklor:
Implemented in both memory and desktop, need to implement for mobile still, setting this as a draft for now, I don't know who can do mobile and how but for now this is what I could pull off. 

 

### Description:

Needed an option to add to a file in the storage without truncating it first, specifically to write long lasting log files.

Fixes #(issue)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/internal/driver/mobile/repository.go
+++ b/internal/driver/mobile/repository.go
@@ -67,11 +67,11 @@ func (m *mobileFileRepo) Reader(u fyne.URI) (fyne.URIReadCloser, error) {
 }
 
 func (m *mobileFileRepo) Writer(u fyne.URI) (fyne.URIWriteCloser, error) {
-	return fileWriterForURI(u)
+	return fileWriterForURI(u, true)
 }
 
 // TODO: need someone who understands native code to write this, I'm not sure
 // how it works
 func (m *mobileFileRepo) Appender(u fyne.URI) (fyne.URIWriteCloser, error) {
-	return fileWriterForURI(u)
+	return fileWriterForURI(u, false)
 }
--- a/internal/driver/mobile/repository.go
+++ b/internal/driver/mobile/repository.go
@@ -11,6 +11,7 @@ var _ repository.Repository = (*mobileFileRepo)(nil)
 var _ repository.HierarchicalRepository = (*mobileFileRepo)(nil)
 var _ repository.ListableRepository = (*mobileFileRepo)(nil)
 var _ repository.WritableRepository = (*mobileFileRepo)(nil)
+var _ repository.AppendableRepository = (*mobileFileRepo)(nil)
 
 type mobileFileRepo struct {
 }
@@ -70,8 +71,6 @@ func (m *mobileFileRepo) Writer(u fyne.URI) (fyne.URIWriteCloser, error) {
 	return fileWriterForURI(u, true)
 }
 
-// TODO: need someone who understands native code to write this, I'm not sure
-// how it works
 func (m *mobileFileRepo) Appender(u fyne.URI) (fyne.URIWriteCloser, error) {
 	return fileWriterForURI(u, false)
 }
--- a/internal/driver/mobile/repository.go
+++ b/internal/driver/mobile/repository.go
@@ -69,3 +69,9 @@ func (m *mobileFileRepo) Reader(u fyne.URI) (fyne.URIReadCloser, error) {
 func (m *mobileFileRepo) Writer(u fyne.URI) (fyne.URIWriteCloser, error) {
 	return fileWriterForURI(u)
 }
+
+// TODO: need someone who understands native code to write this, I'm not sure
+// how it works
+func (m *mobileFileRepo) Appender(u fyne.URI) (fyne.URIWriteCloser, error) {
+	return fileWriterForURI(u)
+}
--- a/internal/docs.go
+++ b/internal/docs.go
@@ -115,6 +115,30 @@ func (d *Docs) Save(name string) (fyne.URIWriteCloser, error) {
 	return storage.Writer(u)
 }
 
+// Append will open a document ready for writing without truncating it's content,
+// you close the returned writer for the save to complete.
+// If the document for this app with that name does not exist a storage.ErrNotExists error will be returned.
+func (d *Docs) Append(name string) (fyne.URIWriteCloser, error) {
+	if d.RootDocURI == nil {
+		return nil, errNoAppID
+	}
+
+	u, err := storage.Child(d.RootDocURI, name)
+	if err != nil {
+		return nil, err
+	}
+
+	exists, err := storage.Exists(u)
+	if err != nil {
+		return nil, err
+	}
+	if !exists {
+		return nil, storage.ErrNotExists
+	}
+
+	return storage.Appender(u)
+}
+
 func (d *Docs) ensureRootExists() error {
 	exists, err := storage.Exists(d.RootDocURI)
 	if err != nil {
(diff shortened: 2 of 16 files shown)


andydotxyz (commented):
Clear and clean, thanks for looking at this - a good addition.

I think it will need the mobile code added to land though.
Would tips on how to progress be sufficient, or are you handing this on to someone else?

mbaklor:
Thank you that was super encouraging and got me the drive to try the mobile code myself! 

Let me know if this mobile code makes sense to you, and if someone with an android or ios setup - well - set up could check that this code works I'd super appreciate it as well!

andydotxyz:
Looking great. The code compiles and runs on my iOS test devices.
But there is no test for the file repository on mobile.
Is there a snippet of code that could be run, or an example we could drop into fyne_demo that would verify?

Thanks!

mbaklor:
I have discovered in writing this test snippet that I in fact forgot to implement the last part of the puzzle, the storage interface itself. Fixed and pushed, and here's the snippet I used to check it on my PC

```go
func main() {
	app := app.NewWithID("io.fyne.appendtest")
	win := app.NewWindow("Append Test")
	win.Resize(fyne.NewSize(400, 300))

	app.Storage().Create("test.txt")

	btnAppend := widget.NewButton("append to file", func() {
		doc, err := app.Storage().Append("test.txt")
		if err != nil {
			dialog.ShowError(err, win)
			return
		}
		_, err = doc.Write([]byte("test line\n"))
		if err != nil {
			dialog.ShowError(err, win)
			return
		}
		err = doc.Close()
		if err != nil {
			dialog.ShowError(err, win)
			return
		}
		dialog.ShowInformation("Success", "Wrote to file", win)
	})
	btnRead := widget.NewButton("read file", func() {
		doc, err := app.Storage().Open("test.txt")
		if err != nil {
			dialog.ShowError(err, win)
			return
		}
		b := make([]byte, 64)
		_, err = doc.Read(b)
		if err != nil {
			dialog.ShowError(err, win)
			return
		}
		dialog.ShowInformation("Success", "Successfully read from file\n\ncontents:\n"+string(b), win)
		println(string(b))
	})

	content := container.NewGridWithRows(2, btnAppend, btnRead)
	win.SetContent(container.NewCenter(content))
	win.ShowAndRun()
}
```

storage.go
```diff
@@ -8,6 +8,7 @@ type Storage interface {
 	Create(name string) (URIWriteCloser, error)
 	Open(name string) (URIReadCloser, error)
 	Save(name string) (URIWriteCloser, error)
+	Append(name string) (URIWriteCloser, error)
```
andydotxyz, line 11: Making this change into an existing interface is a breaking change and we cannot do it (unless a v3 is made).
Additional features added on top of the initial storage API will have to be optional.

andydotxyz (changes requested):

storage.go
mbaklor: Okay, fair, didn't think about how people who implemented the interface themselves will lose it like this. In that case though, how do we make this addition optional while allowing for `app.Storage().Append` or other convenient usage of the new API?

dweymouth: You can create a new interface like Appender or StorageAppender that defines the single function. And then the builtin one implements both Storage and Appender

mbaklor: for now I deleted this commit, as there's no real reason to change the storage interface at all, I'll think about how to deal with the `internal.Docs` questions that came up for me here but for now I can resolve this at least

mbaklor:
the new "it works on my system" with the storage interface reverted to how it was is:

```go
func main() {
	app := app.NewWithID("io.fyne.appendtest")
	win := app.NewWindow("Append Test")
	win.Resize(fyne.NewSize(400, 300))
	root := app.Storage().RootURI()
	uri, err := storage.Child(root, "test.txt")
	if err != nil {
		dialog.ShowError(err, win)
		return
	}

	btnAppend := widget.NewButton("append to file", func() {
		doc, err := storage.Appender(uri)
		defer doc.Close()
		if err != nil {
			dialog.ShowError(err, win)
			return
		}
		_, err = doc.Write([]byte("test line\n"))
		if err != nil {
			dialog.ShowError(err, win)
			return
		}
		dialog.ShowInformation("Success", "Wrote to file", win)
	})
	btnRead := widget.NewButton("read file", func() {
		doc, err := storage.Reader(uri)
		defer doc.Close()
		if err != nil {
			dialog.ShowError(err, win)
			return
		}
		b := make([]byte, 64)
		_, err = doc.Read(b)
		if err != nil {
			dialog.ShowError(err, win)
			return
		}
		dialog.ShowInformation("Success", "Successfully read from file\n\ncontents:\n"+string(b), win)
		println(string(b))
	})

	content := container.NewGridWithRows(2, btnAppend, btnRead)
	win.SetContent(container.NewCenter(content))
	win.ShowAndRun()
}
```

internal/driver/mobile/repository.go
```diff
@@ -67,5 +67,11 @@ func (m *mobileFileRepo) Reader(u fyne.URI) (fyne.URIReadCloser, error) {
 }
 
 func (m *mobileFileRepo) Writer(u fyne.URI) (fyne.URIWriteCloser, error) {
-	return fileWriterForURI(u)
+	return fileWriterForURI(u, true)
+}
+
+// TODO: need someone who understands native code to write this, I'm not sure
```
andydotxyz, line 73: Seems like this TODO could go?

mbaklor: OOPS! deleted

andydotxyz:
This doesn't work work on Android - I finally tracked it down to the `modes` always being `NULL` which caused all write operations to fail.

This patch fixes Android!

```diff
diff --git a/internal/driver/mobile/android.c b/internal/driver/mobile/android.c
index 14e72b66b..db4f55c06 100644
--- a/internal/driver/mobile/android.c
+++ b/internal/driver/mobile/android.c
@@ -154,9 +154,9 @@ void* saveStream(uintptr_t jni_env, uintptr_t ctx, char* uriCstr, bool truncate)
        jobject uri = parseURI(jni_env, ctx, uriCstr);
        jstring modes = NULL;
        if (truncate) {
-               jstring modes = (*env)->NewStringUTF(env, "wt"); // truncate before write
+               modes = (*env)->NewStringUTF(env, "wt"); // truncate before write
        } else {
-               jstring modes = (*env)->NewStringUTF(env, "wa");
+               modes = (*env)->NewStringUTF(env, "wa");
        }
        jobject stream = (jobject)(*env)->CallObjectMethod(env, resolver, saveOutputStream, uri, modes);
        jthrowable loadErr = (*env)->ExceptionOccurred(env);
```

andydotxyz:
Unfortunately on iOS it reports writing successfully but the read back just gets EOF not the file content

andydotxyz:
The following change made it work on iOS:

```diff
diff --git a/internal/driver/mobile/file_ios.m b/internal/driver/mobile/file_ios.m
index 18b08de24..9ff83770d 100644
--- a/internal/driver/mobile/file_ios.m
+++ b/internal/driver/mobile/file_ios.m
@@ -39,7 +39,7 @@ bool iosExistsPath(const char* path) {
     NSFileHandle* handle = [NSFileHandle fileHandleForWritingToURL:url error:&err];
 
     if (!truncate) {
-        [handle truncateFileAtOffset:[handle seekToEndOfFile]];
+        [handle seekToEndOfFile];
     }
 
     return handle;
```

but in part it was because the read operation can return EOF error which we were treating as a problem, so try this:

```go
              _, err = doc.Read(b)
              if err != nil && err != io.EOF {
                      dialog.ShowError(err, win)
                      return
              }
```

andydotxyz (approved):
Thanks so much for this

PR #5191, opened 2024-10-12
Decay mobile scroll as users expect on a touch display

andydotxyz:
### Checklist:

- [ ] Tests included. <- I can't think of how, open to suggestions
- [x] Tests all pass.


andydotxyz:
No objections or comments in 7 day so defaulting accept.

PR #5197, opened 2024-10-15
Add optional Source metadata to the FyneApp.toml

andydotxyz:
Use this to write an extension of the .desktop file for Linux. I am not sure if we can extend other package formats with this information also?

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


sdassow (approved):
Looks good.

Issue #3576, opened 2023-01-16
Add a third state ("indeterminate" or "partial") to the check widget


dweymouth:
### Checklist

- [x] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [X] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

A common UI pattern for selection of multiple related items is to have a dual-state checkbox next to each item and a 3-state checkbox in a header, where the 3-state checkbox will show a "partial" or "indeterminate" state when some, but not all, of the selectable items are checked. It would be nice to implement this in Fyne with the built-in widgets.

### Is it possible to construct a solution with the existing API?

Not without creating a custom widget or custom override of `widget.Check`

### Describe the solution you'd like to see.

Add a third state to the check widget. The third (indeterminate, or partially selected) state should only be able to be set programmatically, not directly via user taps (`check.SetPartial()` ?), and if a partial check is clicked, it should go to the unchecked state (and of course fire `OnChanged(false)`, so the application code can then unselect all the selected items).

dweymouth:
This overview of HTML's 3-state checkboxes might be useful https://css-tricks.com/indeterminate-checkboxes/

andydotxyz:
Just to be clear, as stated in the document linked, HTML does not have 3-state check boxes.
Thankfully Material does have it in the design https://m2.material.io/components/checkboxes.
Getting a simple API for this is probably the complex part, and making it backward compatible too

dweymouth:
I wonder if keeping it similar to HTML would help make it backward-compatible, as it would still be two-state, but with Indeterminate being an additional new property alongside the checked state. (Or you could enforce that `SetIndeterminate` always additionally sets the checked state to checked). The only challenging part for backward compatibility would be if it's desired to communicate additional info about the indeterminate state to listeners of the OnChanged callback. (One could also argue that since setting the widget to indeterminate state is only possible programmatically, you don't need to communicate any additional info in OnChanged, since the application code could keep track of when the widget was set to indeterminate, and then know that OnChanged always clears the indeterminate state.)

andydotxyz:
It's possible that a separate widget may be cleaner - which uses an enum instead of a bool for state.

andydotxyz:
[...]

PR #5201, opened 2024-10-16
Add `Partial` mode to `Check` so we can indicate the indeterminate state

andydotxyz:
This is API controlled only - this is not a three-state check as it cannot be selected by the user.

Screenshot 2024-10-16 at 12 33 29

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


sdassow (approved):
Good stuff!

dweymouth:
I think this can close #3576 actually, as the user could now override Check with their own Tapped and TypedKey handlers to allow it to cycle through the three states with user interaction

dweymouth (approved):

Issue #3799, opened 2023-04-08
When will ttc fonts be supported?


sh1luo:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

I noticed the previous issues, someone mentioned this problem, and some developers said that they will fix it in the next minor version, but I found that it will still panic directly in version v233

### How to reproduce

```golang
var (
	HmTTf []byte
	//go:embed msyh.ttc
	backup []byte
)

func init() {
	fontPaths, err := findfont.Find("msyh.ttc")
	if err != nil {
		HmTTf = backup
		return
	}

	HmTTf, err = os.ReadFile(fontPaths)
	if err != nil {
		log.Logger.Println(err)
	}
}

```

```golang
func (*CustomerTheme) Font(s fyne.TextStyle) fyne.Resource {
	return fyne.NewStaticResource("font", fonts.HmTTf)
}
```

### Screenshots

2023/04/09 08:17:27 Fyne error: font load error
2023/04/09 08:17:27 Cause: unsupported font format
2023/04/09 08:17:27 At: C:/Users/zhonger/go/pkg/mod/fyne.io/fyne/v2@v2.3.3/internal/painter/font.go:150
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x4ec pc=0x7ff7fd7549e0]

goroutine 1 [running, locked to thread]:
github.com/benoitkugler/textlayout/fonts/truetype.(*Font).Upem(0x1?)
C:/Users/zhonger/go/pkg/mod/github.com/benoitkugler/textlayout@v0.3.0/fonts/truetype/metrics.go:34
github.com/benoitkugler/textlayout/harfbuzz.NewFont({0x7ff7fdf2b078?, 0x0})

### Example code

```golang
var (
	HmTTf []byte
	//go:embed msyh.ttc
	backup []byte
)

func init() {
	fontPaths, err := findfont.Find("msyh.ttc")
	if err != nil {
		HmTTf = backup
		return
	}

	HmTTf, err = os.ReadFile(fontPaths)
	if err != nil {
		log.Logger.Println(err)
	}
}

```

```golang
func (*CustomerTheme) Font(s fyne.TextStyle) fyne.Resource {
[...]

Issue #2761, opened 2022-02-03
Failed to load TTC font: freetype: invalid TrueType format: bad TTF version


ipfans:
### Describe the bug:

When I trying to load font with `FYNE_FONT` env but it failed with error.

### To Reproduce:

`go run main.go`. Code as Example Code.

### Screenshots:

```
2022/02/03 10:28:18 Fyne error:  font load error
2022/02/03 10:28:18   Cause: freetype: invalid TrueType format: bad TTF version
2022/02/03 10:28:18   At: /Users/User/go/pkg/mod/fyne.io/fyne/v2@v2.1.2/internal/painter/font.go:27
```

### Example code:

```go
package main

import (
	"os"
	"runtime"
	"strings"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
	"github.com/flopp/go-findfont"
)

func main() {
	paths := findfont.List()
	for _, path := range paths {
		if runtime.GOOS == "darwin" {
			if strings.Contains(path, "PingFang.ttc") {
				os.Setenv("FYNE_FONT", path)
				break
			}
		}
	}
	a := app.New()
	w := a.NewWindow("Hello World")
	w.SetFixedSize(true)
	w.Resize(fyne.NewSize(300, 300))
	w.SetContent(widget.NewLabel("你好，世界!"))
	w.ShowAndRun()
}
```

### Device (please complete the following information):
 - **OS:** macOS
 - **Version:** 12.1
 - **Go version:** Go 1.17.6
 - **Fyne version:** 2.1.2

Bluebugs:
Does the license of the font allow you to share it with us? Or can you point us to a website where we can get it?

andydotxyz:
There is no need to see the file - you can tell from the extension that it is not a TTF file. That is the only type of font file we support (as the error shows).

ZHOUYue67:
> There is no need to see the file - you can tell from the extension that it is not a TTF file. That is the only type of font file we support (as the error shows).

There's a simple implementation: now fyne use "font.ParseTTF" to load a single font, you can simply change it to "font.ParseTTC" and choose the first one. "ParseTTC" works for both ttf and ttc. There's no way to choose which font in ttc using this method though.

PR #5210, opened 2024-10-18
ttc fonts supported

### LINKED ISSUE #4053: Support for font collections in theme / user font
### Checklist

I want to use .ttc to load my fonts

I would have to load, parse and extract a font from the collection then pass it into fyne as a re-packaged font file (not sure if that is even possible).

The ability to load .ttc and .otc files.
Ideally it would not have to parse and load every font in the collection - but somehow make them available for selecting when needed.

bon-ami:
theme.CheckFontParsable added
font.ParseTTC replaces font.ParseTTF. The first font of the collection is taken.

 

### Description:

Fixes #4053,#3799,#2761

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

- [ ] Public APIs match existing style and have Since: .


andydotxyz:
You do realise that this will parse the entire font collection and then return only the first font? Could be quite a bad performance drop.

Also, why all the nil checks added? Having font rendering need to worry about nil seems bad - because we cannot possible render a string if the font is nil. Unless I misunderstood the change this seems like a bad thing.

dweymouth:
Yes, this certainly isn't really solving TTC support. TTC support means loading all the fonts in the collection when needed to display specific glyphs. The point of TTC is usually achieving larger glyph coverage than any single TTF can support on its own

bon-ami:
@andydotxyz , @dweymouth , 
1. this is not a final solution. I cannot use current API to fully support ttc. Since I remember seeing the font mechanism/lib is to be switched, I only want to make ttc partially supported.
2. I want to protect the whole app being crashed by the rendering issue, by not showing the texts. Because I am using a customized theme with fonts, if the font is invalid, I have no API to check that and it will crash my whole app.
3. So I also added an API to check the validity of a font resource.

as I commented in https://github.com/fyne-io/fyne/issues/4053#issuecomment-2418643857,
I think we should provide a temporary change according to https://github.com/fyne-io/fyne/issues/2761#issuecomment-2041367042
as a quick, though not full, support to https://github.com/fyne-io/fyne/issues/3799

andydotxyz:
Thanks for this, some thoughts:

> 1. this is not a final solution. I cannot use current API to fully support ttc. Since I remember seeing the font mechanism/lib is to be switched, I only want to make ttc partially supported.

The font lib was switched out a couple of releases ago, there is no future plans to change font loading.
If it is a temporary solution it should certainly not add new public APIs - we have to support them forever.

> 2. I want to protect the whole app being crashed by the rendering issue, by not showing the texts. Because I am using a customized theme with fonts, if the font is invalid, I have no API to check that and it will crash my whole app.

If there is a crash we should fix that directly - no Fyne API call or render should ever panic.
For the case where you are worried a font is not supported it is possible to query the go-text/typesetting loader directly in your app.

> 3. So I also added an API to check the validity of a font resource.

As noted above APIs should not be part of a temporary fix but have clear utility and purpose. I get the feeling that this has been added to avoid a crash - which is not how we do things as the panic itself should be resolved.

andydotxyz:
I see that updates were merged but not addressing the comments left, is this PR still progressing towards something for v2.6.0?

bon-ami:
> I see that updates were merged but not addressing the comments left, is this PR still progressing towards something for v2.6.0?

I merged on my branch, not here. It is a incomplete solution just for myself.

andydotxyz:
> > I see that updates were merged but not addressing the comments left, is this PR still progressing towards something for v2.6.0?
> 
> I merged on my branch, not here. It is a incomplete solution just for myself.

OK thanks for the clarity. Please do open a PR on this project if you have a solution that would be something to be shared with all users.

Issue #5195, opened 2024-10-14
Package command fails on OpenBSD


sdassow:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

OpenBSD [tar](https://man.openbsd.org/tar) does not support the -J flag and fails during `fyne package`.

### How to reproduce

Run `fyne package --os openbsd --release` on OpenBSD.

### Example code

Any code.

### Fyne version

any

### Go compiler version

any

### Operating system and version

OpenBSD any version

andydotxyz:
Ah shoot. When we moved to .xz checks were made to ensure it was supported everywhere. we missed this.

Can it be enabled through an external plugin or will we need to revert to .tar.gz for BSD usages?

sdassow:
We could use `gtar` (GNU tar) instead, which can be installed from packages/ports.

andydotxyz:
> Other than that, using gtar wouldn't be the best user experience since it means installing something else first.

Unusual for someone to reply to their own suggestion going against it ;).

It sounds like we have no choice - should we just create .gz when run on OpenBSD? I'm not sure it matters too much.
It may also mean we have to use that when packaging /for/ openBSD as well, or you won't be able to open the archives.

sdassow:
> > Other than that, using gtar wouldn't be the best user experience since it means installing something else first.
> 
> Unusual for someone to reply to their own suggestion going against it ;).

Just thinking out loud ;).

> It sounds like we have no choice - should we just create .gz when run on OpenBSD? I'm not sure it matters too much.

I'll have a look and brew something up. When I found this I just created an issue for later.

> It may also mean we have to use that when packaging /for/ openBSD as well, or you won't be able to open the archives.

For distribution it doesn't matter as much.
For source distribution not at all since that's usually handled in the ports tree for proper OpenBSD packaging as a build dependency.
For binary distribution it means a small inconvenience having to manually install GNU tar and/or xz first.

andydotxyz:
> For binary distribution it means a small inconvenience having to manually install GNU tar and/or xz first.

Sorry I missed this before commenting on the PR.
Typically we want to put the package install / additional dependency on the developer not the end user. Is it a bad thing to adapt the fix so it avoids that compression for both developer and user?

sdassow:
[...]

PR #5212, opened 2024-10-18
Fix tar flags (and filename) on OpenBSD

sdassow:
This PR addresses #5195 and makes sure the compression is supported by `tar` in the base system, as well as using the corresponding file name.

Without this patch:
```
$ fyne package --os openbsd --release
failed to create archive with tar: tar: unknown option -- J
usage: tar {crtux}[014578befHhjLmNOoPpqsvwXZz]
           [blocking-factor | archive | replstr] [-C directory] [-I file]
           [file ...]
       tar {-crtux} [-014578eHhjLmNOoPpqvwXZz] [-b blocking-factor]
           [-C directory] [-f archive] [-I file] [-s replstr] [file ...]
```

### DIFF
--- a/cmd/fyne/internal/commands/package-unix.go
+++ b/cmd/fyne/internal/commands/package-unix.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"fyne.io/fyne/v2/cmd/fyne/internal/templates"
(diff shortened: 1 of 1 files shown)


cmd/fyne/internal/commands/package-unix.go
```diff
@@ -96,8 +97,14 @@ func (p *Packager) packageUNIX() error {
 			return fmt.Errorf("failed to write Makefile string: %w", err)
 		}
 
+		tarCmdArgs := []string{"-Jcf", filepath.Join(p.dir, p.Name+".tar.xz")}
+		if runtime.GOOS == "openbsd" {
```
andydotxyz, line 101: if OpenBSD cannot handle writing tar files with .xz compression should we not also ensure that we only produce .gz files when it is the /target/ system as well? Otherwise won't the user be unable to open the package if it was produced on, say, a Linux host?

sdassow: Good point with regards to logic and consistency, even though practically I haven't seen anyone targeting OpenBSD in a cross build from another system. It seems impractical to achieve and maintain as the ABI changes pretty much every six months with a new release.
Still, it should depend on the target instead of the runtime OS, and I'll adjust it.

sdassow: Better now, thanks.

Jacalz (approved):

PR #5214, opened 2024-10-18
Recover panic "send on closed channel"

xaionaro:
### Description:
In [my application](https://github.com/xaionaro-go/streamctl) I have a lot of concurrent processes, and I get panics in fyne.
My understanding is that if I close a window in one goroutine, but refresh something else in another, it sometimes may cause a panic.

To mitigate the issue adding a "recover" to the places where it happens.

Feel free to suggest a better solution (I'll try to implement it) :)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/async/chan_go1.21.go
+++ b/internal/async/chan_go1.21.go
@@ -2,7 +2,11 @@
 
 package async
 
-import "fyne.io/fyne/v2"
+import (
+	"sync"
+
+	"fyne.io/fyne/v2"
+)
 
 // UnboundedFuncChan is a channel with an unbounded buffer for caching
 // Func objects. A channel must be closed via Close method
@@ -34,9 +38,10 @@ func NewUnboundedCanvasObjectChan() *UnboundedChan[fyne.CanvasObject] {
 // UnboundedChan is a channel with an unbounded buffer for caching
 // Func objects. A channel must be closed via Close method.
 type UnboundedChan[T any] struct {
-	in, out chan T
-	close   chan struct{}
-	q       []T
+	in, out   chan T
+	close     chan struct{}
+	closeOnce sync.Once
+	q         []T
 }
 
 // NewUnboundedChan returns a unbounded channel with unlimited capacity.
@@ -61,7 +66,11 @@ func (ch *UnboundedChan[T]) In() chan<- T { return ch.in }
 func (ch *UnboundedChan[T]) Out() <-chan T { return ch.out }
 
 // Close closes the channel.
-func (ch *UnboundedChan[T]) Close() { ch.close <- struct{}{} }
+func (ch *UnboundedChan[T]) Close() {
+	ch.closeOnce.Do(func() {
+		close(ch.close)
+	})
+}
 
 func (ch *UnboundedChan[T]) processing() {
 	// This is a preallocation of the internal unbounded buffer.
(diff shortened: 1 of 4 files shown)


andydotxyz:
> My understanding is that if I close a window in one goroutine, but refresh something else in another, it sometimes may cause a panic.

Not that I am aware. Anyhow panics should be resolved not worked around. Sorry you have found this bug, but what we need to do is fix the panic instead of hiding it. Hopefully you have a stack trace that we can use to help track it down?

andydotxyz (changes requested):
As noted this isn't the right way to go - we should not hide panics, we should fix their cause.

xaionaro:
> [...] but what we need to do is fix the panic instead of hiding it. 

I totally agree.

> Hopefully you have a stack trace that we can use to help track it down?

It happens rarely and very difficult to reproduce, and unfortunately I do not have the stack traces anymore.

The rationale behind this PR was:
* Bug#1: The fact that `Close` can panic at all (for example on a double call of `UnboundedChan[T].Close`) seems like a bug by itself, so why not fix it (especially if it mitigates the problem I have as a side effect)? It felt right to log this, but I decided let's publish a PR and see what would be the feedback overall.
* Bug#2: If the `eventQueue` is already closed then any event after that does not make any sense. I believe it is generally erroneous by design if a channel is used for writing by one entity (e.g. `window.Close`), but closed by another entity (`gLDriver.runGL`). So for example if I run `runGL` in one goroutine, but invoke `window.Close` from another, I can easily get a panic. But since this is my first commit to the package, I should not do any design changes, so I resorted to just discarding the events.

But I agree that one should address root causes and not hide bugs, so I was hesitant to publish this PR and mostly expected it to be closed without merging :)

andydotxyz:
Thanks for understanding and discussing.

> Bug#1: The fact that Close can panic at all (for example on a double call of UnboundedChan[T].Close) seems like a bug by itself

But isn't that what Go does for regular `chan`? 

> Bug#2: If the eventQueue is already closed then any event after that does not make any sense. I believe it is generally erroneous by design if a channel is used for writing by one entity (e.g. window.Close), but closed by another entity (gLDriver.runGL). So for example if I run runGL in one goroutine, but invoke window.Close from another, I can easily get a panic.

I think this is the bug you are fixing, yes. We should understand how it happens and resolve that. It may help to know that we are moving the threading model to a simpler setup in 2.6 which moves the event and render to the same goroutine - does that impact and/or resolve the root cause?

xaionaro:
> But isn't that what Go does for regular `chan`?

It does that for regular channels, but AFAIR it does not do that in Close functions, e.g. you can easily double-close a file:
https://play.golang.com/p/ydQoTGo_IlK

IMO, it is still a valid argument that double-close is generally a bug by itself and should be surfaced (not hidden). However personally my opinionated approach is: surface as a log entry, but do not fail if unnecessary.

> [...] moves the event and render to the same goroutine - does that impact and/or resolve the root cause?

Yeah, I think it should address the root cause and there will be no more panics (and no need for the fix I proposed in the PR).

andydotxyz:
> It does that for regular channels, but AFAIR it does not do that in Close functions, e.g. you can easily double-Close a file:
https://play.golang.com/p/ydQoTGo_IlK

Fair point, but I think the semantic of replacing a channel overrides the look of a file because of the presence of `Close`. We would have used `close()` if we could, but of course we cannot override these primitive functions.

Issue #5207, opened 2024-10-18
func (f *fileDialog) setLocation(dir fyne.URI) error { does not work with non "file" repository


brucealthompson:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

There is "file" respository specific code in func (f *fileDialog) setLocation(dir fyne.URI) error {}

Here is the code where the issue is:
newDir := storage.NewFileURI(buildDir)
		isDir, err := storage.CanList(newDir)
		if err != nil {
			return err
		}

Notice that setLocation takes a fyne URL as input (dir fyne.URI). If the fyne URL has a scheme of "file" then the call to  storage.NewFileURI(buildDir) above will work. If the fyne URL in the parameter (dir fyne.URI) has a scheme different than "file". Then the call to storage.NewFileURI(buildDir) will return a URL with the "file" scheme instead of the scheme passed in dir fyne.URI.

I found this bug when trying to use the respository "httpfile" that I created.

### How to reproduce

Create a non "file" repository. 
Create a new file dialog pointing to the non "file" repository
Try traversing directories in the non "file" repository. The UI will stop once storage.NewFileURI(buildDir) resturns an invalid "file" URI.

### Example code

Create a new file dialog using the following code:
```
locationURI, err := storage.ParseURI("httpfile:///")
	if err != nil {
		return err
	}
	listableURI, err := storage.ListerForURI(locationURI)
	if err != nil {
		return err
	}
	folderdialog := dialog.NewFolderOpen(callback, parent)
	folderdialog.SetLocation(listableURI)
	folderdialog.Show()
```

### Fyne version

2.5.2

### Go compiler version

1.23.1

### Operating system and version

Windows 11

brucealthompson:
Here's the fix:
               ```
 **newURL := dir.Scheme() + "://" + filepath.ToSlash(buildDir)
		newDir, err := storage.ParseURI(newURL)
		if err != nil {
			return err
		}**
		isDir, err := storage.CanList(newDir)
		if err != nil {
			return err
		}
```

Issue #5204, opened 2024-10-16
dialog.New(File/Folder/...)Open() only works with "file" respository


brucealthompson:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

I have written a new repository interface which is intended to allow dialog.New(File/Folder/...)Open() to work with remote filesystems in a similar manner as WebDav. I have found that the current implementation of dialog.New(File/Folder/...)Open() has
a number of ties to the operating system / filesystem that fyne is running on. This works fine when it is displaying content from the "file" repository. However, it pretty much breaks down when you try to use it with a different repository. Here is an example of code that looks like it is currently broken for anything other than the "file" repository:
[fyne.io](http://fyne.io/)\fyne\v2@v2.5.2\dialog\file.go

```
func (f *fileDialog) loadFavorites() {
  favoriteLocations, err := getFavoriteLocations()
  if err != nil {
    fyne.LogError("Getting favorite locations", err)
  }
  favoriteIcons := getFavoriteIcons()
  favoriteOrder := getFavoriteOrder()
```

getFavoriteLocations() is platform (not repository) specific code.

I suspect that the file/folder dialog objects are going to need some pretty major work to get them to work with repositories other than "file". 

### How to reproduce

This is bug is not possible to reproduce with the current codebase as a new non "file" repository needs to be written to exercise the code. I have written a new repository called "httpfile" which supports a subset of webdav functionality.

### Example code

I would provide the package "httpfile" I have written but it looks like attaching go files is not supported.

### Fyne version

2.5.2

### Go compiler version

1.23.1

### Operating system and version

Windows 11

brucealthompson:
I have produced a fix for this issue on https://github.com/brucealthompson/fyne/tree/release/v2.5.x.
The fix essentially removes the elements in the fyne file dialog if the scheme of the initial location is not "file". It works for me and I believe in general.
I am not going to push this fix as I it is not a long term solution to the issue.

andydotxyz:
Hmm, is it right that setting the starting location to non-file should remove the favourites? That is not how things like Windows file explorer or a macOS finder window work.

If we have to hide them because they are not supported on wasm then let's make that change instead.

brucealthompson:
[...]

PR #5215, opened 2024-10-19
Fix for https://github.com/fyne-io/fyne/issues/5207

brucealthompson:
### Description:

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -474,8 +474,11 @@ func (f *fileDialog) setLocation(dir fyne.URI) error {
 			d = buildDir
 			buildDir = d + string(os.PathSeparator)
 		}
-
-		newDir := storage.NewFileURI(buildDir)
+		newURL := dir.Scheme() + "://" + filepath.ToSlash(buildDir)
+		newDir, err := storage.ParseURI(newURL)
+		if err != nil {
+			return err
+		}
 		isDir, err := storage.CanList(newDir)
 		if err != nil {
 			return err


andydotxyz (changes requested):
This doesn't seem right - the new URI is constructed essentially from a file path, you can see it starts with:

```
buildDir := filepath.VolumeName(localdir)
```

but then is created as:

```
newURL := dir.Scheme() + "://" + filepath.ToSlash(buildDir)
```

This does not seem right. I cannot tell exactly what is happening here, but simply swapping the correct scheme where "file://" is used will not be the right fix as file uris/paths are being assumed in this whole code chunk.

I think the right fix will be to traverse this properly with the storage API to iterate on the `storage.Parent` result to walk back up the "tree".

brucealthompson:
I was trying to make as minimal a change as possible. The fix certainly works for the "file" repository and my "httpfile" repository that I created.

> This does not seem right. I cannot tell exactly what is happening here, but simply swapping the correct scheme where "file://" is used will not be the right fix as file uris/paths are being assumed in this whole code chunk.
> 
> I think the right fix will be to traverse this properly with the storage API to iterate on the storage. Parent result to walk back up the "tree".

I did not touch the code that manipulated the path. I believe the data structure represented by a fyne URI is the same as an RFC 3986 URI and the same as is represented by the golang URI https://pkg.go.dev/go.lsp.dev/uri#URI. I believe the path manipulation code you are referring to is correct for manipulating an RFC 3986 compliant path.

The only potential issue with the code you are referring to is that the "file" repository as well as the the "httpfile" repository can include DOS drive letters in the path. IE, a valid "file" repository URL is "file://c:/foo". This is a valid URI according to RFC 8089 https://datatracker.ietf.org/doc/html/rfc8089#page-13. The fyne storage.ParseURI() call does not parse this type of URI correctly and ends up putting the drive letter in the Authority field of the URI. I had to work around this issue in my "httpfile" repository code. So there is a bug in the fyne fyne storage.ParseURI() code. I think the storage.ParseURI() implementation should call the golang uri.ParseURI function instead of having the fyne repository interface provide its own implementation for this.

The only issue I could see with the fix I made is this:
newURL := dir.Scheme() + "://" + filepath.ToSlash(buildDir)

The filepath.ToSlash() call should not be necessary since buildDir should be an RFC 3986 compliant path, I do not remember why I did this. I will try removing it and test to make sure everything works.

brucealthompson:
OK. I fixed the path manipulation code. Here are the changes
fyne fix

Notice newURL := dir.Scheme() + "://" + filepath.ToSlash(buildDir) >> newURL := dir.Scheme() + "://" + buildDir

I also found the issue in the fyne storage.ParseURI() code. Here it is:
	if strings.EqualFold(scheme, "file") {
		// Does this really deserve to be special? In principle, the
		// purpose of this check is to pass it to NewFileURI, which
		// allows platform path seps in the URI (against the RFC, but
		// easier for people building URIs naively on Windows). Maybe
		// we should punt this to whoever generated the URI in the
		// first place?

This code should be independent of scheme. This is a separate issue.

brucealthompson:
> I also found the issue in the fyne storage.ParseURI() code. Here it is:
> if strings.EqualFold(scheme, "file") {
> // Does this really deserve to be special? In principle, the
> // purpose of this check is to pass it to NewFileURI, which
> // allows platform path seps in the URI (against the RFC, but
> // easier for people building URIs naively on Windows). Maybe
> // we should punt this to whoever generated the URI in the
> // first place?
> 
> This code should be independent of scheme. This is a separate issue.

I have fixed this issue for my httpfile repository. The fix is to add a ParseURI method to the repository. Here is what I imples code should be independent of scheme. This is a separate issue.

I have fixed this issue for my httpfile repository. The fix is to add a ParseURI method to the repository. Here is what I implemented for the httpfile repository:

```
func (r *HttpFileRepository) ParseURI(uristring string) (fyne.URI, error) {
	localurl, err := url.Parse(uristring)
	if err != nil {
		return nil, err
	}
	uri := &uri{
		scheme: localurl.Scheme,
		// url.Parse interprets windows drive letters as a Host. Include this in the path.
		path:      localurl.Host + localurl.Path,
		authority: "",
		fragment:  "",
		query:     "",
	}
	return uri, nil
}
```

I believe the same fix cam be applied to the "file" repository. Then the repository specific code in fyne storage.ParseURI() that I highlighted above can be removed.

andydotxyz:
> I believe the path manipulation code you are referring to is correct for manipulating an RFC 3986 compliant path

The issue is that the code assumes a URI is just scheme://path - this Is not true when there is a host component, or port (or query, but that may be out of scope).

> The fix is to add a ParseURI method to the repository.

I don't understand why this would be the case - as you say a URI parse is pretty standard (if there is a fix in our parser we should fix it). We can iterate through the file tree of any URI using `storage.Parent` iteratively as I suggested before. I don't think new parsing or file path manipulation is needed to do this in a way that works for all repository schemes.

andydotxyz:
I think you should check out an earlier piece of work for handling windows volume names and remote servers for windows format at https://github.com/fyne-io/fyne/pull/4606

brucealthompson:
The 2nd topic is about an issue I founding during code inspection. Here is the issue I found during code inspection:

> I also found the issue in the fyne storage.ParseURI() code. Here it is:
> if strings.EqualFold(scheme, "file") {
> // Does this really deserve to be special? In principle, the
> // purpose of this check is to pass it to NewFileURI, which
> // allows platform path seps in the URI (against the RFC, but
> // easier for people building URIs naively on Windows). Maybe
> // we should punt this to whoever generated the URI in the
> // first place?
>
> This code should be independent of scheme. This is a separate issue.

This issue definitely impacted my httpfile repository. If you remove this code from storage.ParseURI() then the file repository interface will not work with dialog.NewFileOpen()

> I don't understand why this would be the case - as you say a URI parse is pretty standard (if there is a fix in our parser we should fix it). We can iterate through the file tree of any URI using storage.Parent iteratively as I suggested before. I don't think new parsing or file path manipulation is needed to do this in a way that works for all repository schemes.

I believe my testing from above shows that there is an issue with the fyne parser. It does not work with "file"//" URIs that include DOS paths. The code above must be a workaround for this issue. You can either:
1. Leave things alone.
2.  fix the fyne parser to make it completely general
3. create a  ParseURI method for the "file" repository

dialog/file.go
```diff
@@ -469,13 +481,16 @@ func (f *fileDialog) setLocation(dir fyne.URI) error {
...
 			d = "/"
 		} else if i > 0 {
-			buildDir = filepath.Join(buildDir, d)
+			buildDir = buildDir + "/" + d
 		} else {
 			d = buildDir
-			buildDir = d + string(os.PathSeparator)
+			buildDir = d + "/"
+		}
+		newURL := dir.Scheme() + "://" + buildDir
```
andydotxyz, line 489: This assumes that a URI  (not URL!) is a scheme followed by a path. This is only true for file URIs. Most contain a host and/or port portion as well.

andydotxyz (changes requested):
Is hiding favourites necessary because you have specified a non-file starting location? The two feel unconnected.

Anyhow, the reason for asking for changes is the false URI assumption mentioned inline and in the issue this fixes.

brucealthompson:
> Is hiding favourites necessary because you have specified a non-file starting location? The two feel unconnected.
> 
> Anyhow, the reason for asking for changes is the false URI assumption mentioned inline and in the issue this fixes.

Ugh...

This is github issues. I did a pull request for the file revision before this one. The revision you are looking at includes a local I fix I made for https://github.com/fyne-io/fyne/issues/5204. This fix works but I did not intend to push this back to fyne.

andydotxyz:
> Ugh...
> 
> This is github issues. I did a pull request for the file revision before this one. The revision you are looking at includes a local I fix I made for #5204. This fix works but I did not intend to push this back to fyne.

Ah. If you want to have two Pull Requests where one does not depend on another you'll need a different branch.

dialog/file.go
brucealthompson: > This assumes that a URI (not URL!) is a scheme followed by a path. This is only true for file URIs. Most contain a host and/or port portion as well.

This variable is actually for a URI, not a URL. I will rename it and check to see if there are other components of the fyne URI with variable name "dir" that need to be included in this constructed URI.

I will create a separate branch for this fix based on your dev branch and resubmit. Give me a little time as I am working on something else right now.

brucealthompson: I did not write the code you are referring to. I was trying to do the simplest thing possible with rewriting a bunch of existing code. Not sure why the original author did not choose to use storage.Child(). Was it not evailable when this was originally written?

In any case, I can check into that solution as well.

andydotxyz: > I did not write the code you are referring to. I was trying to do the simplest thing possible without rewriting a bunch of existing code.

I appreciate that, but sometimes when the existing code contains the bug we need to fix that directly instead of following the old pattern and working around it to minimise the change.

> Not sure why the original author did not choose to use storage.Child(). Was it not evailable when this was originally written?

Well, when this only worked with file:// the assumptions held - because it does not (or did not) have any other URI elements like host or port. Those assumptions no longer hold.

Thanks for looking into it.

andydotxyz:
Can you please update the title and description so it is clear what should (vs should not!) be in this PR/review?

PR #5225, opened 2024-10-26
Allow a theme to override the form factor of the tabs to be mobile

andydotxyz:
This avoids being forced to respect the singleton device reported from hardware

Opening for discussion about this use for the `Size` callback avoiding a breaking change or new API requirement for a "theme feature".

Also test to follow.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container/apptabs.go
+++ b/container/apptabs.go
@@ -215,7 +215,7 @@ func (t *AppTabs) SetItems(items []*TabItem) {
 
 // SetTabLocation sets the location of the tab bar
 func (t *AppTabs) SetTabLocation(l TabLocation) {
-	t.location = tabsAdjustedLocation(l)
+	t.location = tabsAdjustedLocation(l, t)
 	t.Refresh()
 }
 
@@ -357,7 +357,7 @@ func (r *appTabsRenderer) buildTabButtons(count int) *fyne.Container {
 	buttons := &fyne.Container{}
 
 	var iconPos buttonIconPosition
-	if fyne.CurrentDevice().IsMobile() {
+	if isMobile(r.tabs) {
 		cells := count
 		if cells == 0 {
 			cells = 1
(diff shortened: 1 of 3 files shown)


dweymouth:
I think a new API would be *much* cleaner than overriding the meaning of Size in themes like this

andydotxyz:
> I think a new API would be _much_ cleaner than overriding the meaning of Size in themes like this

What would such an API be like? Would it be an optional extension? The part that I'm not sure about is that some sort of "config" section could be all sorts of types int/bool/float/string...

dweymouth:
> > I think a new API would be _much_ cleaner than overriding the meaning of Size in themes like this
> 
> What would such an API be like? Would it be an optional extension? The part that I'm not sure about is that some sort of "config" section could be all sorts of types int/bool/float/string...

I guess I need to know more about the planned use cases. If this is just about the tabs widget, I would just add a new property API to AppTabs and DocTabs to control this, similar to the Orientation property of RadioGroup and CheckGroup. If it's about overriding device settings in general, maybe a new set of APIs related to device and not the theme? If it's about general, open-ended settings that are tied to the theme, then some more API design may be needed (and more use case discovery to guide the API design)

andydotxyz:
It's not the same thing as orientation, as the tabs respond differently on different platforms rather than just switch configuration. 
If the options are exposed separately then something overriding the context becomes tied to the render implementation for AppTabs.

I wonder if it could instead be an api on ThemeOverride if being on theme isn't looking like a good approach?
It would probably still need to be passed internally through theme or a whole new cache that indexes on theme. I was trying to keep it simple by passing it in Size but I'll keep thinking.

andydotxyz:
This was merged by mistake somehow. The commit was reverted on `develop` and re-applied on this branch so the conversation can continue where we left off.

PR #5226, opened 2024-10-26
Add Option to CheckGroup to make Rows/Columns

BieHDC:
### Description:
This adds an option to CheckGroup to make Rows/Columns instead of one long list.
I made this for a program i am currently writing and it makes sense to upstream.
It is fully backwards compatible.

Video: 
https://github.com/user-attachments/assets/79bf6be3-0d93-4e51-8f94-f94c1244da75

Example Code for testing:
```go
package main

import (
	"fmt"
	"strconv"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/dialog"
	"fyne.io/fyne/v2/widget"
)

type gui struct {
	a fyne.App
	w fyne.Window
}

func main() {
	g := gui{}

	g.a = app.NewWithID("biehdc.fynedev.checkgroupcolumns")
	g.w = g.a.NewWindow("CheckGroup with Rows/Columns")
	g.w.CenterOnScreen()
	g.w.Resize(fyne.NewSize(800, 600))

	g.w.SetContent(g.content())
	g.w.ShowAndRun()
}

func (g *gui) content() fyne.CanvasObject {
	contenttainer := container.NewStack()
	setcontentainer := func(co fyne.CanvasObject) {
		contenttainer.Objects = []fyne.CanvasObject{co}
		contenttainer.Refresh()
	}

	numitems := widget.NewEntry()
	numitems.PlaceHolder = "Amount of items to generate"
	numcols := widget.NewEntry()
	numcols.PlaceHolder = "Amount of Rows/Columns"
	togglehorizonal := widget.NewCheck("Horizontal", nil)
	makeui := widget.NewButton("Make Example", func() {
		ni, err := strconv.Atoi(numitems.Text)
		if err != nil {
			dialog.ShowError(err, g.w)
			return
		}

		nr, err := strconv.Atoi(numcols.Text)
		if err != nil {
			dialog.ShowError(err, g.w)
			return
		}

		var items []string
		for i := range ni {
			items = append(items, fmt.Sprintf("Check Item %d", i))
		}

		checkgroup := widget.NewCheckGroup(items, nil)
		checkgroup.SetColumns(nr)
		checkgroup.Horizontal = togglehorizonal.Checked

		setcontentainer(checkgroup)
	})
	numitems.OnSubmitted = func(_ string) { makeui.OnTapped() }
	numcols.OnSubmitted = func(_ string) { makeui.OnTapped() }

	return container.NewBorder(
		container.NewHBox(
			NewMinSizer(numitems, fyne.NewSize(300, 0)),
			NewMinSizer(numcols, fyne.NewSize(300, 0)),
			togglehorizonal, makeui,
		), nil, nil, nil,
		container.NewScroll(contenttainer),
	)
}

type minSizer struct {
	widget.BaseWidget
	child   fyne.CanvasObject
	minsize fyne.Size
}

func NewMinSizer(child fyne.CanvasObject, minsize fyne.Size) *minSizer {
	ms := &minSizer{child: child, minsize: minsize}
	ms.ExtendBaseWidget(ms)
	return ms
}

func (ms *minSizer) MinSize() fyne.Size {
	return ms.child.MinSize().Max(ms.minsize)
}

func (ms *minSizer) CreateRenderer() fyne.WidgetRenderer {
	return widget.NewSimpleRenderer(ms.child)
}
```

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/widget/check_group.go
+++ b/widget/check_group.go
@@ -1,6 +1,7 @@
 package widget
 
 import (
+	"math"
 	"strings"
 
 	"fyne.io/fyne/v2"
@@ -15,6 +16,7 @@ import (
 type CheckGroup struct {
 	DisableableWidget
 	Horizontal bool
+	numColumns int
 	Required   bool
 	OnChanged  func([]string) `json:"-"`
 	Options    []string
@@ -30,12 +32,20 @@ var _ fyne.Widget = (*CheckGroup)(nil)
 // Since: 2.1
 func NewCheckGroup(options []string, changed func([]string)) *CheckGroup {
 	r := &CheckGroup{
-		Options:   options,
-		OnChanged: changed,
+		Options:    options,
+		OnChanged:  changed,
+		numColumns: 1,
 	}
 	r.ExtendBaseWidget(r)
 	r.update()
 	return r
+}
+
+func (r *CheckGroup) SetColumns(columns int) {
+	if columns < 1 {
+		columns = 1
+	}
+	r.numColumns = columns
 }
 
 // Append adds a new option to the end of a CheckGroup widget.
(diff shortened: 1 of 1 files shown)


Jacalz:
> It is fully backwards compatible.

Sorry but it actually isn’t. Your variable is indexed from one meaning that the default value of zero when creating a struct without using the constructor results in incorrect behaviour.

BieHDC:
> > It is fully backwards compatible.
> 
> Sorry but it actually isn’t. Your variable is indexed from one meaning that the default value of zero when creating a struct without using the constructor results in incorrect behaviour.

True, but i can fix that. Right now i would more like to know if that feature was even wished in the main repo. Otherwise i would demote it to a new widget for fyne-x.

Jacalz:
Absolutely, I meant it as a review comment. Sorry if that wasn't clear. I personally think it is a useful addition :)

BieHDC:
I reworked the code on the basis of gridLayout which also fixed a rendering bug and the list now also has the natural flow. And of course fixed the backwards compatibility issue.

widget/check_group.go
```diff
@@ -179,54 +185,87 @@ type checkGroupRenderer struct {
...
+		primaryObjects, secondaryObjects = secondaryObjects, primaryObjects
 	}
 
-	itemSize := fyne.NewSize(itemWidth, itemHeight)
-	x, y := float32(0), float32(0)
-	for _, item := range r.items {
-		item.Resize(itemSize)
-		item.Move(fyne.NewPos(x, y))
+	size := r.checks.MinSize()
+	cellWidth := float64(size.Width) / float64(primaryObjects)
```
andydotxyz, line 218: This doesn't seem quite right - normally a grid would spread out over the avilable space, but here it's using the minimum size instead. Is that intentional?

Initially it didn't matter because it was just 1 column but with horizontal and this grid layout I wonder if we should be respecting size...

widget/check_group.go
```diff
@@ -179,54 +185,87 @@ type checkGroupRenderer struct {
...
+		r.numColumns = 1
+	}
+	return r.numColumns
+}
+
+func (r *CheckGroup) countRows() int {
+	return int(math.Ceil(float64(len(r.items)) / float64(r.countColumns())))
+}
+
+func getLeading(size float64, offset int) float32 {
```
andydotxyz, line 199: functions not attached to types should be at the end of the file as utilities

andydotxyz (commented):
Opened a question about the layout. Perhaps folk have thoughts?

Thanks for adding this though1

widget/check_group.go
andydotxyz: Lets leave it as the PR and if we want to change then review it separately - and probably match radio and anything else needing to also update.

BieHDC: Fixed.

andydotxyz:
From the test failures it seems like perhaps the algorithm is now laying out items from a single row/column differently to before.
If that is part of what was agreed then the test golden files will need to be updated.

There should also be a test added for this new feature.

andydotxyz (changes requested):
From the test failures it seems like perhaps the algorithm is now laying out items from a single row/column differently to before.
If that is part of what was agreed then the test golden files will need to be updated.

There should also be a test added for this new feature.

andydotxyz:
Sorry this got delayed, I am struggling with the substantial impact to all of the test changes.

With this change I think that RadioGroup and CheckGroup have now diverged to incompatible layouts and I think really they should remain consistent. It might be easier to review if it was addressed in two separate stages so we can verify the new layout, applied to both, and then discuss the approach/API to how this wrap/column application is set (and also maybe add it to checks as well?)

PR #5231, opened 2024-10-28
Defer to system theme when a custom theme delegates to DefaultTheme

andydotxyz:
I discovered that an app with a partial custom theme will delegate to the builtin theme instead of the rendered one. These can be different if you have a system theme installed.

- [x] I still have to refactor the base config dir code

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -172,10 +172,6 @@ func newAppWithDriver(d fyne.Driver, clipboard fyne.Clipboard, id string) fyne.A
 	return newApp
 }
 
-func rootConfigDir() string {
-	return app.RootConfigDir()
-}
-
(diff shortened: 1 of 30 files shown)


app/app.go
```diff
@@ -172,6 +172,10 @@ func newAppWithDriver(d fyne.Driver, clipboard fyne.Clipboard, id string) fyne.A
 	return newApp
 }
 
+func rootConfigDir() string {
+	return app.RootConfigDir()
```
Jacalz, line 176: Should we not just call that function directly instead of indirectly through this function?

Jacalz (approved):

PR #5239, opened 2024-11-05
Allow a theme to override the form factor of the tabs to be mobile

andydotxyz:
Continuing from the unfortunately-merged-unmerged https://github.com/fyne-io/fyne/pull/5225

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/theme/feature.go
+++ b/internal/theme/feature.go
@@ -11,8 +11,6 @@ const FeatureNameDeviceIsMobile = FeatureName("deviceIsMobile")
 
 // FeatureTheme defines the method to look up features that we use internally to apply functional
 // differences through a theme override.
-//
-// Since: 2.6
 type FeatureTheme interface {
 	Feature(FeatureName) any
 }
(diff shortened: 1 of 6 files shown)


internal/theme/feature.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+type FeatureName string
+
+const FeatureNameDeviceIsMobile = FeatureName("deviceIsMobile")
+
+// FeatureTheme defines the method to look up features that we use internally to apply functional
+// differences through a theme override.
+//
+// Since: 2.6
```
dweymouth, line 15: Is this supposed to be a public API (since it has the Since: line)? If so, this should be in the public theme package rather than internal?

andydotxyz: No it wasn't - that doc was just over-enthusiastic

andydotxyz: Fixed

andydotxyz:
Requests were fixed 2 weeks ago, merging

PR #5241, opened 2024-11-05
Fix color picker dismissed callback

### LINKED ISSUE #5237: color picker callback isn't called in case of dismissal
### Checklist

### Describe the bug

The callback provides when creating a new color picker is called when a color is chosen - great.
For the file dialogs the same callback is called with a `nil` value for dismissal - great.
This does **not** happen in the case of the color picker.

### How to reproduce

1. Create a color picker with a callback function.
2. Show the color picker.
3. Hit the cancel button.
4. The callback function isn't called.

### Screenshots

_No response_

### Example code

```go
package main

import (
	"fmt"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/dialog"
	"image/color"
)

func main() {
	fapp := app.NewWithID("app-with-unique-id")
	win := fapp.NewWindow("pick color")

	cp := dialog.NewColorPicker("", "", func(c color.Color) {
		fmt.Println("Color callback called with:", c)
		win.Close()
		fapp.Quit()
	}, win)

	cp.Show()
	win.Show()
	fapp.Run()
}
```

### Fyne version

2.5.2

### Go compiler version

1.23.2

Linux 6.1.0-26-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.112-1 (2024-09-30) x86_64 GNU/Linux

### Additional Information

I will be happy to fix this bug.
I have to find out how this is done at Fyne first.

ole108:
### Description:
Call the callback provided to the color picker dialog with a `nil` color if the dialog is dismissed.
So users can easily distinguish between confirmation and dismissal and react to it.
This is inline with how all the file related dialogs work (fileSave, fileOpen and folderOpen).

Fixes #5237

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/color.go
+++ b/dialog/color.go
@@ -141,6 +141,9 @@ func (p *ColorPickerDialog) updateUI() {
 		p.dialog.create(container.NewGridWithColumns(2, p.dialog.dismiss, confirm))
 	} else {
 		p.dialog.content = container.NewVBox(p.createSimplePickers()...)
+		p.dialog.dismiss.OnTapped = func() {
+			p.callback(nil)
+		}
 		p.dialog.create(container.NewGridWithColumns(1, p.dialog.dismiss))
 	}
 }
--- a/dialog/color.go
+++ b/dialog/color.go
@@ -135,6 +135,9 @@ func (p *ColorPickerDialog) updateUI() {
 				p.selectColor(p.color)
 			},
 		}
+		p.dialog.dismiss.OnTapped = func() {
+			p.callback(nil)
+		}
 		p.dialog.create(container.NewGridWithColumns(2, p.dialog.dismiss, confirm))
 	} else {
 		p.dialog.content = container.NewVBox(p.createSimplePickers()...)
--- a/dialog/color_test.go
+++ b/dialog/color_test.go
@@ -95,26 +95,50 @@ func TestColorDialog_SetColor(t *testing.T) {
 
 func TestColorDialog_Buttons(t *testing.T) {
 	for name, tt := range map[string]struct {
-		action     func(d *ColorPickerDialog, w fyne.Window)
+		action          func(d *ColorPickerDialog, w fyne.Window)
+		advanced        bool
 		expectedCalled  bool
 		expectedColored bool
 	}{
-		"confirm": {
+		"confirmAdvanced": {
 			action: func(d *ColorPickerDialog, w fyne.Window) {
 				test.FocusNext(w.Canvas()) // advanced accordion
 				test.FocusNext(w.Canvas()) // dismiss button
 				test.FocusNext(w.Canvas()) // confirm button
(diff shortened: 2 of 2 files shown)


andydotxyz:
I would prefer that we match the data input dialogs not the file dialogs.
I thought the bug you reported was that the close and callback were not called in the right order?

ole108:
I didn't look at the data input dialogs at all yet.
That's next.
Other order of close and callback would fix it, too.
I can dig into that. I suggest that both would be best.

ole108:
OK. I will look into the order of callbacks tomorrow.
That would work for me, too.
And consistency would be great.
I might have started my Fyne journey at the wrong end. :shrug:

dialog/color.go
```diff
@@ -107,7 +110,10 @@ func (p *ColorPickerDialog) updateUI() {
 		w.Hide()
 	}
 	p.dialog.dismiss = &widget.Button{Text: lang.L("Cancel"), Icon: theme.CancelIcon(),
-		OnTapped: p.dialog.Hide,
+		OnTapped: func() {
+			p.callback(nil)
```
andydotxyz, line 114: I don’t think this part of the change is ok.
It will **panic** any apps which assume a colour chosen cannot be nil. Which is probably a lot of apps.

andydotxyz (changes requested):
The change of order seems reasonable, but the adding of possible panics worries me.

dialog/color.go
ole108: OK. I will change. 
But this is all a bit inconsistent. The file dialogs explicitly work with such a nil argument.
And the confirmation dialog tests that the order of callbacks is the other way around.
For Fyne3 it would be great to get this straight.

I will go out for the rest of the day in a few minutes. So I will do the change tomorrow.

ole108: Change is done.
All the best with your workshop at GoLab!

andydotxyz:
I have looked into this to approve and only just realised (sorry if there was a discussion before) that this makes the close order different for the colour picker than for all others.
I agree that the change makes sense, but probably it needs to apply to `dialog/base.go:101` as well so that dialogs that use the builtin callback get data before dismissal.

It could be checked by using the other data input dialogs (Form, Entry) in the context of your dialog library to see if you get data from either of those.

ole108:
I am happy to make this change.
Funny thing is that I had made that change already before.
But I realized that one test (for the confirmation dialog) would fail because of it.
It tests specifically the order to be the other way around.
So I reverted that part of my change. :disappointed: 

I am on my way to a conference in Berlin (DevFest, where I will give a talk about the `fdialog` project).
So I am slow to react this time.

andydotxyz:
We have started the release process for v2.6.0, so it would be a good time to get this landed :).

andydotxyz:
There's just a couple of days left - what do you think @ole108 ?

PR #5243, opened 2024-11-06
Update NewFileURI to support relative paths

### LINKED ISSUE #5234: FileDialog does not handle relative file URIs well.
### Checklist

### Describe the bug

When using a relative directory to file dialog's SetLocation method and it is a relative directory it will not produce any contents.  If you select the listing icon then files will show, but not the breadcrumb buttons. You can then click grid again to see the files, but again no breadcrumb buttons.

### How to reproduce

Simply create  a file dialog and set the storage uri to a relative directory that exists to the running working directory of the program.  Then click the list/grid buttons to see behaviors as described.

### Screenshots
This is an example of what the bug looks like:
___1go_build_ttstick_runtimes_traceReader_x1Wtl3O6WC

Gif showing that the file dialog still sorta works when you hit the list/grid buttons after but due to the nature of this issue the breadcrumb buttons will not be available:
___go_build_ttstick_runtimes_traceReader_S6exROEjpN

The workaround is simple of course, first make sure the directory is an absolute directory before setting the location:
___go_build_ttstick_runtimes_traceReader_t4d10mk6Lc

### Example code

Of course make sure ./logs exist with some files from the runtime directory when the program starts for this example to work.
```go
package main

import (
	"context"
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/dialog"
	"fyne.io/fyne/v2/storage"
)

func main() {
	_, cancel := context.WithCancel(context.Background())
	a := app.New()
	w := a.NewWindow("Trace Reader")
	w.Resize(fyne.NewSize(800, 600))
	w.SetOnClosed(cancel)

	var fileOpener *dialog.FileDialog
	fileOpener = dialog.NewFileOpen(func(reader fyne.URIReadCloser, err error) {
		if err != nil {
			dialog.ShowError(err, w)
		}
	}, w)

	p := "./logs"
	// Workaround is as follows
	// p, err := filepath.Abs(p)
	// if err != nil {
	// 	panic(err)
	// }
	u := storage.NewFileURI(p)
	d, err := storage.ListerForURI(u)
	if err != nil {
		panic(err)
	}
	fileOpener.SetLocation(d)
	fileOpener.Show()
	w.SetMaster()
	w.ShowAndRun()
}
```

### Fyne version

v2.5.2

### Go compiler version

1.23.1

Windows 10

### Additional Information

I tracked down this issue to the code that creates the breadcrumbs loop
`/fyne/dialog/file.go:468` 
image
This will take the location "./logs" and end up parsing it to "/" then "/logs" which does not exist. so the isDir checks fail.

Although the workaround is simple, the fact fyne doesn't return an error wit SetLocation there is no way from a client code perspective to know there was a problem at runtime (setLocation does return an error, but SetLocation does not).  

I think if an error was returned from SetLocation stating it does not support relative file uri locations that would of been fine.

I suspect this should work though instead of throwing an error, but maybe file dialog needs a way to recognize if the storage provider url support an absolute function of some sort and use that before creating the bread crumbs?

Also, given the way the bread crumbs are made, I suspect this bug hits in other scenarios like when the user has permissions to list in the directory they opened and navigated in but in the parent paths they do not have listing permissions which is a possible edge case in some OS's.  That kind of issue would manifest in the same way.  It might make more sense if the breadcrumbs were made from the current directory up to parents instead of starting with the root parent; in this way a partial bread crumb could be listed at all times, even if it is only for the current directory, up until the path breaks.

andydotxyz:
Fix up docs to clarify as well.

Fixes #5234 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/repository/file.go
+++ b/internal/repository/file.go
@@ -176,9 +176,8 @@ func (r *FileRepository) Parent(u fyne.URI) (fyne.URI, error) {
 		return nil, repository.ErrURIRoot
 	}
 
-	parent := ""
-	// use the system native path resolution
-	parent = filepath.Dir(s)
+	child := filepath.Base(s)
+	parent := s[:len(s)-len(child)]
 	if parent[len(parent)-1] != filepath.Separator {
 		parent += "/"
 	}
--- a/internal/repository/file.go
+++ b/internal/repository/file.go
@@ -177,8 +177,8 @@ func (r *FileRepository) Parent(u fyne.URI) (fyne.URI, error) {
 	}
 
 	child := filepath.Base(s)
-	parent := s[:len(s)-len(child)]
-	if parent[len(parent)-1] != '/' {
+	parent := s[:len(s)-len(child)] // avoid filepath.Dir as it follows platform rules
+	if parent == "" || parent[len(parent)-1] != '/' {
 		parent += "/"
 	}
 
--- a/internal/repository/file.go
+++ b/internal/repository/file.go
@@ -171,14 +171,14 @@ func (r *FileRepository) Parent(u fyne.URI) (fyne.URI, error) {
 	// trim the scheme
 	s = strings.TrimPrefix(s, fileSchemePrefix)
 
-	// Completely empty URI with just a scheme
-	if s == "" {
+	// Completely empty URI or only root component
+	if s == "" || s == "/" || (len(s) == 2 && s[1] == ':') {
 		return nil, repository.ErrURIRoot
 	}
 
 	child := filepath.Base(s)
 	parent := s[:len(s)-len(child)]
-	if parent[len(parent)-1] != filepath.Separator {
+	if parent[len(parent)-1] != '/' {
 		parent += "/"
 	}
 
--- a/storage/uri.go
+++ b/storage/uri.go
@@ -11,8 +11,12 @@ import (
 // Relative paths will be converted to absolute using filepath.Abs if required.
 func NewFileURI(path string) fyne.URI {
 	assumeAbs := false
-	if len(path) >= 2 {
-		assumeAbs = path[1] == ':' || path[0] == '/'
+	if len(path) >= 1 {
+		if path[0] == '/' {
+			assumeAbs = true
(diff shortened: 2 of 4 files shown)


internal/repository/file.go
```diff
@@ -171,15 +171,14 @@ func (r *FileRepository) Parent(u fyne.URI) (fyne.URI, error) {
...
+	if s == "" || s == "/" || (len(s) == 2 && s[1] == ':') {
 		return nil, repository.ErrURIRoot
 	}
 
-	parent := ""
-	// use the system native path resolution
-	parent = filepath.Dir(s)
-	if parent[len(parent)-1] != filepath.Separator {
+	child := filepath.Base(s)
+	parent := s[:len(s)-len(child)]
```
Jacalz, line 180: Was there something wrong with `filepath.Dir`? If so, this might warrant a comment as to why we are hard coding the logic.

storage/uri.go
```diff
@@ -1,12 +1,31 @@
...
+	"path/filepath"
+
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/storage/repository"
 )
 
 // NewFileURI creates a new URI from the given file path.
+// Relative paths will be converted to absolute using filepath.Abs if required.
 func NewFileURI(path string) fyne.URI {
+	assumeAbs := false
```
Jacalz, line 13: How about using `filepath.IsAbs()`?

Jacalz (changes requested):
Thanks. Just a few comments inline :)

internal/repository/file.go
andydotxyz: It's because the `filepath` utilities are platform specific and the `Dir()` command won't work as expected on a windows path when you are running on Unix.

storage/uri.go
andydotxyz: Documented as per the comment above

Jacalz (approved):

PR #5248, opened 2024-11-09
Add menu shortcut texts for non-darwin OS

### LINKED ISSUE #5108: Improve how shortcut special keys for menu items are rendered on Windows and Linux
### Checklist

With Fyne it is possible to define shortcuts and assign them to menu items. Those shortcuts are then rendered alongside the menu item to inform the user about them.

However, special keys (e.g. ALT, SHIFT, ..) are always rendered as symbols  - and apparently in macOS style -, which will most likely be confusing to many users on Windows and Linux.

Example from the Demo App:
Screenshot from 2024-09-03 19-44-23

e.g. the 2nd symbol from the left is actually the ALT key

No

I would suggest to render all special keys as plain text on Windows and Linux. E.g. "Alt" and use the "+" key as separator, e.g. "Alt+G" or "Ctrl+C". This appears to be how most apps on Linux and Windows are doing it.

The following example is VS Code on Ubuntu/Gnome:

Screenshot from 2024-09-03 19-58-09

And here LibreOffice Word on Ubuntu/Gnome:

image

ErikKalkoken:
### Description:

The current implementation renders menu shortcuts exclusively in macOS style, which might be confusing for users on Windows and Linux (see also #5108).

This PR adds a new generic style for rending menu shortcut texts on non-Darwin OS, while keeping the existing macOS style for Darwin OS. The generic style is mostly spelling out the names of keys instead of using symbols and should be much easier to understand for Windows and Linux users. The style is based on the VS Code shortcut documentation for [Windows](https://code.visualstudio.com/shortcuts/keyboard-shortcuts-windows.pdf) and [Linux](https://code.visualstudio.com/shortcuts/keyboard-shortcuts-linux.pdf).

#### Additional changes

- Refactored existing code
- Added unit tests for function that generates shortcut texts

#### Example screenshot

Screenshot from 2024-11-09 14-23-03

Fixes #5108

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

N/A


widget/menu_item.go
```diff
@@ -37,6 +41,25 @@ var keySymbols = map[fyne.KeyName]rune{
 	fyne.KeyUp:        '↑',
 }
 
+// key default text
```
Jacalz, line 44: This comment does not add any value over the variable name. Either remove it or write something more helpful.

widget/menu_item.go
```diff
@@ -13,13 +14,16 @@ import (
...
-	runeModifierControl = '⌃'
-	runeModifierShift   = '⇧'
+	runeModifierAltDarwin     = '⌥'
+	runeModifierControlDarwin = '⌃'
+	runeModifierShiftDarwin   = '⇧'
+	runeModifierSuperDarwin   = '⌘'
 )
 
-var keySymbols = map[fyne.KeyName]rune{
+// key symbols for darwin OS
```
Jacalz, line 25: Ditto.

widget/menu_item.go
```diff
@@ -13,13 +14,16 @@ import (
 
 // TODO
 
+// modifier symbols for darwin OS
```
Jacalz, line 17: Ditto.

widget/menu_item.go
```diff
@@ -13,13 +14,16 @@ import (
 
 // TODO
```
Jacalz, line 15: Would you mind removing this while at it? :)

widget/menu_item.go
```diff
@@ -392,33 +415,63 @@ func shortcutColor(th fyne.Theme) color.Color {
...
-func textsForShortcut(s fyne.KeyboardShortcut, th fyne.Theme) (texts []*canvas.Text) {
-	b := strings.Builder{}
+// textsForShortcut return texts for rendering a shortcut.
+func textsForShortcut(s fyne.KeyboardShortcut, th fyne.Theme) []*canvas.Text {
+	return textsForShortcutAndOS(s, th, runtime.GOOS)
+}
+
+// textsForShortcut return texts for rendering a shortcut for a specific OS.
+func textsForShortcutAndOS(s fyne.KeyboardShortcut, th fyne.Theme, os string) []*canvas.Text {
+	var b strings.Builder
```
Jacalz, line 425: I would rather keep the previous `b := strings.Builder{}` syntax if you don't mind. That's more consistent with other code in the repo as far as I know.

widget/menu_item.go
```diff
@@ -392,33 +415,63 @@ func shortcutColor(th fyne.Theme) color.Color {
...
+		}
+		if mods&fyne.KeyModifierAlt != 0 {
+			b.WriteString("Alt+")
+		}
+		if mods&fyne.KeyModifierShift != 0 {
+			b.WriteString("Shift+")
+		}
+		if mods&fyne.KeyModifierSuper != 0 {
+			b.WriteString("Super+")
+		}
```
Jacalz, line 467: This seems like quite a bit of code duplication. Can you somehow move the darwin rune constants to a separate file and define the values specifically on darwin and have the other values in another file? This way we could have `runeModifierControlDarwin` return the darwin value on darwin and "Ctrl+" on other operating systems.

Jacalz (changes requested):
Nice idea. Thanks for this. I've left some of my own ideas on the code but I think it would be wise to have @toaster have a look as well.

widget/menu_item.go
ErikKalkoken: Thank you for taking the time to review my PR. 

I will rework the comments.

widget/menu_item.go
ErikKalkoken: I chose this form, because it is unnecessary to initialize a string builder and I think this form is more Go idiomatic (e.g. this is what is used in the [official go documentation](https://pkg.go.dev/strings#example-Builder). And you actually find both variants in the repo (e.g. [here](https://github.com/fyne-io/fyne/blob/81b3cd1496f0a41dc8dee09f8f60722575274824/widget/markdown.go#L176).
So I would prefer to keep this change, would that be ok with you?

widget/menu_item.go
ErikKalkoken: If you look more closely you will see that the code for each case is actually quite different. The former adds rune, the later strings. The former can add two texts with different formatting, the later only one. 

My intention here was to minimize changes to the existing code. I would advise against mapping the shortcut text via platform dependent constants though. That would make this function harder to test and understand. But I am happy to look into combining the two key maps into one, which should allow me to reduce the lines of code quite a bit.

widget/menu_item.go
Jacalz: There is no code difference between the two as all Go variables, unlike in C, are initialized to matter what so there is no performance win here. I much prefer using `:=` where possible as there is less words to parse until I get to the type.

ErikKalkoken: I am aware that both variants are equivalent and I understand that you personally prefer the other form. My point is that using var form is more idiomatic Go code. e.g. the Go standard library uses the var form [exclusively](https://github.com/search?q=repo%3Agolang%2Fgo+%2Fvar+%5Cw%2B+strings.Builder%2F&type=code).

Jacalz: I see your point and I appreciate that the compiler does aeon thing else but let's keep consistent with our own code base please. https://github.com/search?q=repo%3Afyne-io%2Ffyne%20strings.Builder&type=code

widget/menu_item.go
ErikKalkoken: There appear to be some flaky tests and I do not have the rights to re-run tests in this repo, but they are now passing in my branch.

Jacalz: I'm sorry but there seems like there have been some communication mishaps as some information got lost in translation.

You didn't have to rework the code that much, I find it a lot harder to read now, sorry. What I was trying to say was that the four if statements were duplicated. What I meant was this:
- Move darwin rune constants to a `_darwin.go` file so they are used only when building for that architecture. Use single length strings and remove darwin specific naming.
- The other ctrl+ string etc. for other architectures can be put in another file that only is compiled in on non darwin desktop os. Constants have the same names as corresponding darwin ones.
- Move the four if statements out of the darwin/other-os checks so they happen before that and use the containers that switch depending on the os in use. The previous code you had in this function for checking operating system can be kept.
- One test runs on darwin the other for other architectures. There should already be existing code/files for this (as parts of the menus are darwin specific already), I think. 

Again, I'm sorry if this was unclear but it was what I was trying to hunt at with my previous messages.

ErikKalkoken: There are 19 platform dependent strings, not 4. Are you suggesting to map all of those via target specific code or just the modifiers?

As I tried to explain before, it is not only the strings that are different, but there also is different code per platform. Do you want me to put that platform specific logic in target specific files too or keep the current approach via runtime constant?

Would it not make much more sense to use one approach to handle all platform specific variations, not mix approaches?

Again, would it not be better to avoid the additional complexity that target specific code adds and keep everything together in one file (i.e. use the runtime constant approach for everything)?

With all due respect, but it appears to me that you are asking changes without having a complete understanding of the  code. So in order to avoid any more miscommunication, I would very much appreciate it if you could take the time to read through the complete code (and my above questions) and then revise your comments accordingly.

andydotxyz: > With all due respect, but it appears to me that you are asking changes without having a complete understanding of the code. 

In general this seems like an unhelpful thing to say - maintainers of this project generally have a good handle of the code. If in doubt try to assume that.

I do not understand where the miscommunication exists but this diff seems to be getting longer somehow.
Previously there were 3 consts that defined the runes used. And I understand that there are now 19 strings. But even so the code complexity should not have needed to increase.

We already had the OS specific (for darwin) code through build flags which I am sure @Jacalz was indicating could be kept.

Personally I would have split this into two commits - the first moving from the old rune system to new strings (where darwin are 1-length and others vary) which should have been an almost direct replacement. And then a second commit introducing the idea that the key symbol/name lookup could be set per-system as well.

It is possible that I have over simplified here, and if so please do share why that wouldn't be desirable - but it seems like it would be the minimal approach.

ErikKalkoken: > In general this seems like an unhelpful thing to say - maintainers of this project generally have a good handle of the code. If in doubt try to assume that.

I did not say that. I was referring to the code of my PR, not the existing code of the Fyne project.

I am sorry, but it is still not clear to me what approach you want me to implement. 
Please allow me summarize the issue as I see it:

The baseline is my initial commit, not the latest commit (That is what I understand @Jacalz  is referring to).

There are 4 kinds of things that have to be defined for each platform:
- 4 modifiers (string mapping)
- 15 keys (string mapping)
- text format for modifiers (symbol for mac, default for others)
- text format for keys (symbol or monospace for mac, default for others)

There are 2 approaches to define something platform dependent:

- A: target specific code / build tags
- B: runtime OS condition

There are 2 solutions being discussed:

1. I would suggest to use approach B for everything, because it is simpler and allows you to unit test everything. While with approach A you can only test the whole thing in platform tests (e.g. on Github).

1. @Jacalz latest request would use approach A for the 4 modifiers and approach B for the rest. I find it pretty ugly to mix both approaches in the same solution.

I would appreciate if you could confirm that 2. is really the solution you want me to implement.

andydotxyz:
Yes a runtime check might seem simpler (not sure how using runtime checks is easier to unit test TBH).
But for efficiency the build flags are preferable because there is less code to compile and the compiler can eliminate or inline the static const/var declarations etc.
I don't think that Jacob was asking for the "2) hybrid" I would have thought that using A) for all of them matches the current code style most effectively.

P.S. do check that the font must be different before coding that in - I think that if you choose symbol it would fall back to the normal font if the item is not in fact a symbol.

Jacalz:
I entirely agree that A matches what we already have the best. It is good to be consistent with that and also not compiling in unnecessary code can be a win (sometimes also for readability).

To be fair, I did actually suggest B initially because that better described what I meant with "moving out the four if-statements" and because it was a smaller change from the existing code in the PR. However, I do agree that A is more consistent with what we already have and likely is the better path forward.

ErikKalkoken:
tyvm both for the clarification and your patience. I will rework the solution to exclusively use build flags.

@andydotxyz 

> Yes a runtime check might seem simpler (not sure how using runtime checks is easier to unit test TBH).

With runtime checks you can make the current OS a parameter of the `textsForShortcuts()` function and then test all OS cases with unit tests (see also my initial commit for an example). This works here, because the "platform specific" code is just some different texts and formatting and runs on all platforms. But I fully understand your point about consistency; this was just to answer your question.

andydotxyz:
Thanks, A helpful answer. It is indeed true that with the build flags the unit tests cannot cover all systems on one run, but CI provides the checks for those systems - so overall we get the optimised path but get the test runs too.
You'll find that on this path some tests /have/ to run on the target system (such as visual confirmation that symbols render in menus) - so the infrastructure is already set up that way.

ErikKalkoken:
Here is my new attempt ready for review.

Jacalz:
Thanks. I had a very quick look through the code on my phone and it looked great. Will follow up with a proper review later in the week when I have more time :)

widget/testdata/menu/desktop/layout_shortcuts_other_theme_changed.png
Jacalz, line 1: This test image looks wrong position wise now. Would you mind updating the size so it displays correctly?

Jacalz (approved):
Perfect. Thanks for working on this 👍

ErikKalkoken:
> Perfect. Thanks for working on this 👍

Thank you again for your review and patience with my PR. This is my first feature submission for Fyne and I am still learning how everything works.

The tests are now passing on my branch. Seams to be those flaky tests again acting up.

Jacalz:
> Thank you again for your review and patience with my PR. This is my first feature submission for Fyne and I am still learning how everything works.

No worries. Glad we all got to the same page in the end. Congratulations on your first PR then :)

> The tests are now passing on my branch. Seams to be those flaky tests again acting up.

Yeah, sorry about that. Seems like the coverage upload is a bit flakey. Strange. Either way, kicking the tests made it all pass now so I'll merge :+1:

PR #5250, opened 2024-11-11
Don't reload whole theme when variant changes on XDG

Jacalz:
### Description:

This is a small improvement in the XDG theme watching code to make sure that we only change the variant and not do a full theme reload each time given that we know that just the variant changed. The delay when switching between dark and light mode is decreased.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


dweymouth (approved):
Code looks good, I didn't test it myself but I trust that you verified it :)

PR #5251, opened 2024-11-12
Minor cleanup for handling of disabled animations

Jacalz:
### Description:

Just a very simple cleanup PR to move the `no_animations` tag handling to use existing infrastructure in the `internal/build` package. The biggest win here is reducing the number of files in the `app` package but it might theoretically (due to now using a constant for the value) also help the compiler deadcode eliminate more branches when building without animations (although the latter is unlikely to matter in most cases). 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app/settings.go
+++ b/app/settings.go
@@ -12,8 +12,6 @@ import (
 	"fyne.io/fyne/v2/theme"
 )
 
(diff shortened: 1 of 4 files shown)


sdassow (approved):
Nice. And always glad to avoid `init()`.

PR #5278, opened 2024-11-20
Add Known issues running on Windows on README.md

xalves:
### Description:
This change includes a new section on the README. It adds instructions regarding the Windows steps to make Fyne work(MSYS2).
The motivation to add this was that this information is on the Fyne website but I didn't find it easily, so the motivation for this is to move it to a place where usually we read first.


Jacalz:
Thanks but I don't think we can merge this as it is now. The wording makes it seem like running applications on Windows is not working at all. I'm also not sure what this fixes? As far as I know, `CGO_ENABLED` is already set to 1 for everything that isn't cross compilations. If something is missing at https://docs.fyne.io/started/, I think we should update that instead?

andydotxyz:
Agreed. The "problem" was that the tools were not installed. The "fix" is following getting started.

If we need to make that more obvious then we can, but we don't need to list known issues or ay ENV setup to make things work.

PR #5280, opened 2024-11-21
Resolve some issues with reopening full screen windows

xfaris:
### Description:
Resolve the issue of incorrect position when opening a full screen window after it is hidden. 
Resolve the issue of flickering when the mouse is placed on the taskbar when the full screen window is hidden and opened again.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -1023,6 +1023,13 @@ func (w *window) doShowAgain() {
 		w.viewLock.Lock()
 		w.visible = true
 		w.viewLock.Unlock()
+
+		if w.fullScreen { // this does not work if called before viewport.Show()
+			go func() {
+				time.Sleep(time.Millisecond * 100)
+				w.SetFullScreen(true)
+			}()
+		}
 	})
 }
 


Jacalz:
If the issue is a result of a reach condition then we should solve the race condition instead of sleeping and re-running a function. If we are certain that it is not, we should document clearly why we do it and which issue it relates to.

Do you have an issue number for this problem?

Jacalz:
> I think they just copied the code from the initial show function.
Perhaps factoring that out into a function called from both places would be clearer?

I'm just generally sceptical about creating a new goroutine and then sleeping. How do we know that the arbitrary sleep time is the right one? Why do we even have to spawn a goroutine and then do the operation at a later time? Seems to me like we are either working around a bug or some strange race condition.

Excuse my ramblings though. If this is the only way to solve the problem then sure, a common function with an explanatory comment would be much better.

andydotxyz:
> If this is the only way to solve the problem then sure, a common function with an explanatory comment would be much better.

This sounds like a good conclusion.

In 2.6.0 there may be a better solution as the window display sequence may be affected by other threading changes.

Jacalz (changes requested):
Per discussion above, lets move this out to a separate function

Jacalz:
It looks like this timeout isn't needed with the changes in https://github.com/fyne-io/fyne/pull/5328?

andydotxyz:
Yes, with that merged @xfaris should rebase on latest `develop` and the problem goes away.

andydotxyz:
@xfaris are you able to rebase this and complete the PR? it should be cleaner now that window handling on develop has been tidied up.

Jacalz:
I took the liberty of rebasing this onto latest develop. It would be good with a reproduction example for the issue so we can verify that this fixes the issue. As I have added a commit, I am no longer able to approve and merge this myself.

andydotxyz (approved):

Jacalz:
This is now merged. Sorry for the long wait.

Issue #4997, opened 2024-07-11
Scrolling is laggy with high speed gaming mice


dweymouth:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

When users use Fyne apps with high-speed gaming mice, e.g. 1000 Hz, the scrolling is very laggy and continues to process a backlog of scroll events after the user stops dragging the scroll bar.

A possible solution is to merge scroll events in the driver if they arrive too fast so they are added together and processed only once per frame with the combined delta of all the scroll events that were merged.

### How to reproduce

Run any Fyne app with a scroll container using a mouse that reports very fast, say 1000 Hz

### Example code

fyne_demo

### Fyne version

develop

### Go compiler version

n/a

### Operating system and version

all

dweymouth:
See [githu](https://github.com/dweymouth/supersonic/issues/418) for more details and a user-submitted screen capture

andydotxyz:
I don't believe this can apply to all operating systems - the way they handle stepping is very different.
I'm guessing this is a Linux report - and may relate specifically to the X / Wayland / WM configuration.

Please include at least the minimal info for it to be reproducible.

In addition this seems to confuse two items - the refresh rate and the scroll speed. Scrolling is normally communicated through button presses, but the refresh rate of the mouse refers to the precision of the location on screen and speed of the move...

dweymouth:
From the user reporting the bug to Supersonic:

OS: Arch Linux
DE: GNOME 46.3.1
WM: Mutter (Wayland)

They are also using xwayland as far as I can tell since I don't think a Wayland build is packaged in Arch

Bush-cat:
> I don't believe this can apply to all operating systems - the way they handle stepping and scrolling is very different. I'm guessing this is a Linux report - and may relate specifically to the X / Wayland / WM configuration.

I'm having the same issue on windows, scrolling takes ages.

It's not as laggy as in the video below but laggy in a sense that it feels like it only runs at 30fps on my 144hz display, scrolling in any other app is much more responsive.

https://github.com/dweymouth/supersonic/assets/126769537/0e4ad93e-bfbc-4760-8715-35883f958412

_Originally posted by @kimjongbing in https://github.com/dweymouth/supersonic/issues/418#issuecomment-2222591874_

[...]

Issue #4058, opened 2023-07-13
widget.List is very slow on android


blaize:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

I need to use a List to display my data. It's work fine on windows but on android, scrolling the list is very slow, almost unusable.

### How to reproduce

Compile the code attached with
`fyne package -os android -appID scroll.test`
Drag & drop the apk on android studio emulator or your android device
Run the app and scroll the list

### Screenshots

This with real data. But It's the same behaviour with the sample code attached.

https://github.com/fyne-io/fyne/assets/1087443/f26f6f93-6ef1-438b-a0af-b460c16b5033

### Example code

```
package main

import (
	"fmt"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/data/binding"
	"fyne.io/fyne/v2/theme"
	"fyne.io/fyne/v2/widget"
)

func main() {

	a := app.NewWithID("scroll.test")
	w := a.NewWindow("Scroll Test")

	items := []string{}
	for i := 0; i < 100; i++ {
		items = append(items, fmt.Sprintf("%d", i))
	}

	l := widget.NewListWithData(binding.BindStringList(&items),
		func() fyne.CanvasObject {
			return container.NewHBox(
				widget.NewCheck("", func(b bool) {}),
				widget.NewLabel("01/01/23"),
				widget.NewLabel("F"),
				widget.NewLabel("client"),
				widget.NewIcon(theme.DocumentIcon()),
				widget.NewLabel("money"),
			)
		}, func(di binding.DataItem, co fyne.CanvasObject) {
			cont := co.(*fyne.Container)
			lbl := cont.Objects[3].(*widget.Label)
			txt, _ := di.(binding.String).Get()
			lbl.SetText(txt)
		})

	w.SetContent(l)

	w.ShowAndRun()

}

```

### Fyne version

2.3.5

### Go compiler version

1.20.5

### Operating system and version

Android 13

andydotxyz:
This may be a duplicate of #1062

blaize:
I have looked the code about GL rendering in fyne and it's way out of my competences to find the bug.
I hope a talented coder will find it soon.
[...]

Issue #3374, opened 2022-11-03
Fyne splitters & scrollers lag behind commands when there is too much content on screen


fuskiid:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

Fyne is unable to keep up with task when there is too much content on the screen, while this may be an expected behavior for extremely large amounts of content, it is coming into affect (in to what I believe to be) way less content than should be expected before this occurs; and on the hardware I am running I would expect to have to deliberately make my system lag before this behavior starts.

I have found this to be most easily noticeable in defyne as a verticle splitter is in between a (relatively) long list of files and a terminal.

Video across all 3 OS included below.

Please let me know if I can assist by providing any further information.

Thanks.

### How to reproduce

1. Reproducible using most recent defyne commit [7405238](https://github.com/fyne-io/defyne/commit/74052386bbf7530bc2bfd6b8ba6818b4ed6122f1)
2. (Possibly) reproducible using fyne_demo (v2.2.3) on the Theme Icons tab, as that seems to have enought items between the splitter to make it lag. I have noticed a lag on my linux machine but none on my mac with fyne_demo (mac does lag with defyne)

### Screenshots

Vertical and Horizontal splitter lag on defyne (linux) https://www.youtube.com/watch?v=tGm3gQgKfo8

Vertical and Horizontal splitter lag on defyne (windows)https://youtu.be/gTMTVzXCwXU

Vertical splitter lag on fyne_demo (linux) https://youtu.be/3wl88qoNCOI - you'll notice the "Animations" section works as expected, in contrast you can see the Theme Icon start to show signs of the lag - although much less than defyne (probably due to less content?)

Vertical splitter lag on defyne mac https://youtu.be/rSRIRwEtOZQ (ignore broken mouse, due to remote desktop to take video)

For all of these I deliberately stop moving the mouse so you know the application is just lagging behind my previous commands

### Example code

See videos under "screenshots" for examples.

### Fyne version

v2.2.3 & commit 5248b46

### Go compiler version

1.18.7/1.19.3 - i've tried with both

### Operating system

Other (specify below)

### Operating system version

Windows 11, Ventura+Monterey, I use Arch btw

### Additional Information

Linux 
AMD Ryzen 9 5950X (32) @ 3.400GHz
[...]

Issue #1062, opened 2020-06-01
mobile apps are lagging when enter text and scrolling


andydotxyz:
Since recent changes that enforce layout the general responsiveness, particularly around text and scroll, has gone down.

Some optimisations that exist in glfw may not have been ported to mobile.

andydotxyz:
A couple of minor fixes are in. We need to do more.
It does not seem to be an issue with the GlScissor which we considered earlier.

andydotxyz:
Thanks to the very generous sponsorship of @hkparker the bounty on this ticket has increased to $1000!

andydotxyz:
In discussion with sponsors of this work we have agreed that substantial improvements have been made - in fact performance in almost all circumstances is now meeting expectations. There is still work to be done so this ticket will remain open to track the remaining optimisations expected around text handling (moving to glyph based render cache instead of per-text-object caches).

50% of the bounty attached here is being awarded to @redawl for the excellent optimisations in the render pipeline and to @Jacalz for all the optimisations made across the library.

The bounty for the remaining text handling optimisations now sits at $500.

repairman29:
Hey there! This looks like something I could help with. I'd be happy to take a look and see what we can do.

*(Found via [Echeo](https://echeo.io))*

andydotxyz:
It will take a lot of careful design of how to sub-divide a single visible object into multiple catchable elements.

Are you sure it's an issue you want to tackle first? If you're coming from a vibe code bounty tracker it may not be a good fit.

PR #5283, opened 2024-11-22
WIP: Improve scroll speed

sdassow:
### Description:

First attempt at improving scroll speed by reducing the amount of calls to Draggable.Dragged.

Big diff because it introduces an interface type for queued events to allow differentiation for optimization purposes.

References:
 - #4997 
 - #4058
 - #1062 
 - #3374

### Checklist:

- [ ] Tests included.
- [X] Tests all pass.

#### Where applicable:


event.go
```diff
@@ -35,3 +35,26 @@ type DragEvent struct {
...
+	ev  *DragEvent
+}
+
+func NewDragEventFunc(wid Draggable, ev *DragEvent) *DragEventFunc {
+	return &DragEventFunc{wid, ev}
+}
+
+func (drag *DragEventFunc) Execute() {
+	drag.wid.Dragged(drag.ev)
+}
```
Jacalz, line 60: I know this is a draft PR but just FYI: this adds a public API for something that I suppose only was intended to be used internally? You might want to use one of the packages in the `internal/` folder.

Issue #4345, opened 2023-10-26
Empty but visible images consume high CPU on 2.4.x


dweymouth:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

A canvas.Image that is empty (no image.Image, fyne.Resource or file set) but not hidden will cause high CPU use as it causes constant texture cache misses and a new GL texture to be created for the "image" on every redraw. This is a regression from 2.3.x.

### How to reproduce

Run the example app

### Example code

```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/canvas"
)

func main() {

	myApp := app.New()
	myWindow := myApp.NewWindow("Empty image perf regression")

	myWindow.SetContent(canvas.NewImageFromImage(nil))
	myWindow.Resize(fyne.NewSize(400, 400))
	myWindow.ShowAndRun()
}
```

### Fyne version

2.4.1

### Go compiler version

go1.21.1

### Operating system and version

macOS Ventura M2

dweymouth:
OK I've managed to work out that it's texture cache misses. newGlImageTexture which is in the pprof trace is only referenced [here](https://github.com/fyne-io/fyne/blob/ccc6a1c19a6d16988da3e468f08f25d8100cd4b4/internal/painter/gl/draw.go#L38) in drawImage, and print statements in [getTexture](https://github.com/fyne-io/fyne/blob/master/internal/painter/gl/texture.go#L32) (which is called by drawTextureWithDetails) reveal that on 2.3.x there are a handful of cache misses on startup (as expected), but then none until mousing over an image. On 2.4.x however it continues to have dozens of cache misses a second even when not interacting with the app

andydotxyz:
Should we mark this as blocker for v2.4.3 (i.e. the point release after current work)?

dweymouth:
> Should we mark this as blocker for v2.4.3 (i.e. the point release after current work)?

I imagine this should be a quick fix to someone familiar with the image painter but I don't know what the fix is. On the other hand it has a very easy workaround (hide empty images) so personally I don't feel it needs to be high priority, unless others complain about it.

andydotxyz:
[...]

PR #5290, opened 2024-11-30
Fixing CPU tight loop if we have a visible image with nil content

andydotxyz:
### Checklist:

- [ ] Tests included. <- runtime CPU is high if you include an image object with no file or resource content
- [x] Tests all pass.

### DIFF
--- a/canvas/image.go
+++ b/canvas/image.go
@@ -100,7 +100,9 @@ func (i *Image) Hide() {
 // MinSize returns the specified minimum size, if set, or {1, 1} otherwise.
 func (i *Image) MinSize() fyne.Size {
 	if i.Image == nil || i.aspect == 0 {
-		i.Refresh()
+		if i.File != "" || i.Resource != nil {
+			i.Refresh()
+		}
 	}
 	return i.baseObject.MinSize()
 }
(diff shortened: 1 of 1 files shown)


canvas/image.go
```diff
@@ -100,7 +100,9 @@ func (i *Image) Hide() {
 // MinSize returns the specified minimum size, if set, or {1, 1} otherwise.
 func (i *Image) MinSize() fyne.Size {
 	if i.Image == nil || i.aspect == 0 {
-		i.Refresh()
+		if i.File != "" || i.Resource != nil {
```
Jacalz, line 103: You can move this into the upper if-statement to remove some unnecessary nesting.

Jacalz (changes requested):
Nice catch. Minor suggestion inline :)

canvas/image.go
andydotxyz: I agree that it could - but it didn't feel right because they are checking different things. Outer is "do we have valid state" and inner is "can we do anything about it".

Jacalz: Good point

Jacalz (approved):

dweymouth:
This probably fixes #4345

andydotxyz:
Thanks so much - I thought there was an issue but could not find it

PR #5292, opened 2024-12-01
Fix `lang` tests

toaster:
### Description:
Fixes a bug in the `lang` tests for `AddTranslations` on systems without English system locale.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/lang/lang.go
+++ b/lang/lang.go
@@ -205,5 +205,4 @@ func updateLocalizer() {
 	}
 	str := closestSupportedLocale(all).LanguageString()
 	setupLang(str)
-	localizer = i18n.NewLocalizer(bundle, str)
 }
--- a/lang/lang.go
+++ b/lang/lang.go
@@ -203,6 +203,5 @@ func updateLocalizer() {
 		fyne.LogError("Failed to load user locales", err)
 		all = []string{"en"}
 	}
-	str := closestSupportedLocale(all).LanguageString()
-	setupLang(str)
+	setupLang(closestSupportedLocale(all).LanguageString())
 }
--- a/lang/lang_test.go
+++ b/lang/lang_test.go
@@ -10,19 +10,19 @@ import (
 )
 
 func TestAddTranslations(t *testing.T) {
-	setupLang("en")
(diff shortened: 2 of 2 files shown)


lang/lang.go
```diff
@@ -203,7 +203,5 @@ func updateLocalizer() {
 		fyne.LogError("Failed to load user locales", err)
 		all = []string{"en"}
 	}
-	str := closestSupportedLocale(all).LanguageString()
-	setupLang(str)
-	localizer = i18n.NewLocalizer(bundle, str)
```
andydotxyz, line 208: Are you sure this was superfluous?
The intention was that it would replace whatever previous configuration was set up...

lang/lang_test.go
```diff
@@ -11,14 +11,14 @@ import (
 func TestAddTranslations(t *testing.T) {
 	setupLang("en")
 	err := AddTranslations(fyne.NewStaticResource("en.json", []byte(`{
-  "Test": "Match"
-}`)))
+		"Test": "Match"
```
andydotxyz, line 14: This doesn't seem right - why would a single indentation of a JSON string be 2 tabs?

andydotxyz (commented):
Thanks for this. I have two questions inline because I'm not sure of the change

lang/lang.go
toaster: `setupLang` does the exact same thing, see https://github.com/fyne-io/fyne/blob/develop/lang/lang.go#L196

lang/lang_test.go
toaster: JSON does not bother about indentation at all. This is not something written into a text file and formatted to be readable there. It is something that should be readable right here in the test code.

andydotxyz: Ah, thanks - I thought it was a comparison sorry

lang/lang.go
andydotxyz: Excellent, sorry for missing that - I was on my mobile

andydotxyz (approved):
👍

PR #5293, opened 2024-12-02
allow to set a custom title for the FileDialog

ftl:
### Description:

Allow to set a custom title text for the FileDialog.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/dialog/file.go
+++ b/dialog/file.go
@@ -55,6 +55,7 @@ type fileDialogPanel interface {
 type fileDialog struct {
 	file             *FileDialog
 	fileName         textWidget
+	title            *widget.Label
 	dismiss          *widget.Button
 	open             *widget.Button
 	breadcrumb       *fyne.Container
@@ -87,6 +88,7 @@ type FileDialog struct {
 	parent           fyne.Window
 	dialog           *fileDialog
 
+	titleText                string
 	confirmText, dismissText string
 	desiredSize              fyne.Size
 	filter                   storage.FileFilter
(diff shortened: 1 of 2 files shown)


dialog/file.go
```diff
@@ -745,6 +751,18 @@ func (f *FileDialog) Hide() {
...
+// SetTitleText allows custom text to be set in the dialog title
+//
+// Since: 2.6
+func (f *FileDialog) SetTitleText(label string) {
+	f.titleText = label
+	if f.dialog == nil {
+		return
+	}
+	f.dialog.title.SetText(label)
+	f.dialog.win.Refresh()
```
andydotxyz, line 763: This line should not be needed - setting text will update visuals required.

andydotxyz: p.s. please try and keep methods in alphabetical order per-type (i.e. SetConfirmText would go before SetTitleText)

andydotxyz (changes requested):
Thanks so much for this.
Just a note inline here - Refresh should not be used unless required especially at the top level as it will cause a lot of redrawing...

dialog/file.go
andydotxyz: I guess you're right consistent with what was there before works for this PR.

andydotxyz (approved):

andydotxyz:
Many thanks for this @ftl and congratulations on your first Fyne feature addition!

PR #5296, opened 2024-12-04
Fix race in window fullscreen test

andydotxyz:
### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/canvas_other_test.go
+++ b/internal/driver/glfw/canvas_other_test.go
@@ -12,7 +12,7 @@ import (
 )
 
 func TestGlCanvas_FocusHandlingWhenActivatingOrDeactivatingTheMenu(t *testing.T) {
-	w := createWindow("Test", false)
+	w := createWindow("Test")
 	w.SetPadded(false)
 	w.SetMainMenu(
 		fyne.NewMainMenu(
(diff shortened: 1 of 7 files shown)


dweymouth (approved):

Jacalz (changes requested):
It seems like you are passing in a "false" parameter into almost all uses of the function. Would another solution not be better?

andydotxyz:
I'm not sure what other solution you have in mind.
This helper function was doing too much in one block so either we pass in this extra state or it has to split into separate functions and all the tests have more code than before?

Jacalz:
If I counted correctly, you are only passing in true in one of the tests. That just makes me think that the parameter shouldn't be there and that that one test needs to be modified, not all the other.

andydotxyz:
> If I counted correctly, you are only passing in true in one of the tests. That just makes me think that the parameter shouldn't be there and that that one test needs to be modified, not all the other.

I understand. Updated.

andydotxyz:
I guess the tests don't agree - for some reason that kicked me back to square one.
Off to think more.

andydotxyz:
Merging in with the requested change to get this release together

andydotxyz:
Not a problem. I think the items on GitHub appeared out of order for some reason. The commit above was the result of the thinking.

Issue #3900, opened 2023-05-20
Some SVG resources don't update appearance correctly with the theme


dweymouth:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

I have some SVG resources that aren't displaying properly with the NewThemedResource wrapper. Rather than updating the fill color appropriately, it seems that the fill is becoming transparent and the SVG is not drawing at all. They draw normally without the themed resource wrapper but using the original fill color.

### How to reproduce

run the example app and switch the OS theme from light to dark. observe that the Fyne app updates its theming but one button themes correctly and the other doesn't show its icon at all

### Screenshots

<img width="377" alt="Screen Shot 2023-05-20 at 10 23 37 AM" src="https://github.com/fyne-io/fyne/assets/5303666/67bc1c1a-2727-45aa-b6d7-46df9a9d3189">

<img width="377" alt="Screen Shot 2023-05-20 at 10 23 54 AM" src="https://github.com/fyne-io/fyne/assets/5303666/367dd55f-bc4c-462c-88bb-96d2b65151fd">

### Example code

```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/layout"
	"fyne.io/fyne/v2/theme"
	"fyne.io/fyne/v2/widget"
)

var ResDiscSvg = &fyne.StaticResource{
	StaticName: "disc.svg",
	StaticContent: []byte(
[...]

PR #5301, opened 2024-12-07
Add fix and test for serializing SVG without width or height attributes

sdassow:
### Description:

TL;DR: serializing SVG without width and height leads to empty but existing attributes which causes issues.

People worked around this issue by adding width and height, and even went as far as building tools to work around this.
And it came up before: https://github.com/fyne-io/fyne/issues/3900#issuecomment-2052726026

As this affects the developer experience considerably enough in my opinion, I'm marking this as a blocker to get it into the upcoming release.

### Checklist:

- [X] Tests included.
- [X] Tests all pass.


andydotxyz (approved):
Great catch thanks

PR #5302, opened 2024-12-08
Switch back to previously working image for now

sdassow:
See reference: https://github.com/actions/runner-images/issues/10636

 

### Description:

Fixing the platform tests.

### Checklist:

- [X] Tests all pass.


Jacalz:
What is the problem with 24.04? Is there anything that we can fix there to get it working?

andydotxyz:
I think this was a mistaken fix for an intermittent issue that was resolved on a different PR.
I landed it to get the next rc together so I think this change is not needed.

PR #5305, opened 2024-12-09
enhance functionality of InnerWindow

roffe:
### Description:

2 changes:

**First:**
allow customization of button location on InnerWindow via InnerWindow.ButtonAlignment

The widget.ButtonAlign constant is used to configure

    ButtonAlignment specifies where the window buttons (close, minimize, maximize) should be placed.
    The default is widget.ButtonAlignCenter which will auto select based on the OS.
    - On Darwin this will be `widget.ButtonAlignLeading`
    - On all other OS this will be `widget.ButtonAlignTrailing`

**Second**
Added `Title() string` function to retrieve the title of the current inner window

### DIFF
--- a/container/innerwindow.go
+++ b/container/innerwindow.go
@@ -73,7 +73,6 @@ func (w *InnerWindow) CreateRenderer() fyne.WidgetRenderer {
 			w.Close()
 		}
 	}}
-	close.Resize(fyne.NewSize(15, 15))
 
 	var icon fyne.CanvasObject
 	if w.Icon != nil {
--- a/container/innerwindow.go
+++ b/container/innerwindow.go
@@ -11,6 +11,14 @@ import (
 
 var _ fyne.Widget = (*InnerWindow)(nil)
 
+// InnerWindowButtonAlignment defines the alignment of the buttons in the window title.
+type InnerWindowButtonAlignment int
+
+const (
+	InnerWindowButtonAlignLeft  InnerWindowButtonAlignment = iota // OSX style (default)
+	InnerWindowButtonAlignRight                                   // Linux, Windows style
+)
+
 // InnerWindow defines a container that wraps content in a window border - that can then be placed inside
 // a regular container/canvas.
 //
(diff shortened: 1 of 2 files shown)


roffe:
that should do it

andydotxyz:
Would it be OK to re-use the existing button align consts?

andydotxyz (changes requested):
A nice addition, thanks

All new APIs need the:
```go
//
// Since: 2.6
```
lines at the end of their documentation

Jacalz:
For the record: I realised that my comment about Fedora defaulting to buttons on the right side is technically incorrect, at least on Gnome. Anyhow, Ubuntu is the largest and they default to the right like most other :)

Jacalz (changes requested):
Thanks. Some minor comments inline :)

container/innerwindow.go
```diff
@@ -53,15 +65,13 @@ func (w *InnerWindow) CreateRenderer() fyne.WidgetRenderer {
...
-			}
-		}},
-		min, max)
+	close := &widget.Button{Icon: theme.WindowCloseIcon(), Importance: widget.DangerImportance, OnTapped: func() {
+		if f := w.CloseIntercept; f != nil {
+			f()
+		} else {
+			w.Close()
+		}
+	}}
```
Jacalz, line 74: I think it would be better to move this down to where the variable is used.

container/innerwindow.go
```diff
@@ -104,6 +130,13 @@ func (w *InnerWindow) SetPadded(pad bool) {
 	w.content.Refresh()
 }
 
+// Title returns the current title of the window
+//
+// Since: 2.6
+func (w *InnerWindow) Title() string {
+	return w.title
+}
```
Jacalz, line 138: Adding this method in a PR that says it allows changing the button location might become a bit confusing if we ever have to look back on this in the future but I'll let it slide this time. Lets just add a dot at the end of the first comment though.

Jacalz (changes requested):
Sorry, I realised that I missed some things in my last review

container/innerwindow.go
roffe: done

andydotxyz:
Hi @roffe, it's not clear why this was deleted, was it a mistake?

Issue #1967, opened 2021-02-17
Form not rendering correct Items


stuartmscott:
### Describe the bug:
Changes to `widget.Form.Items` are not reflected once the Form is refreshed

### Example code:
```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
	"log"
	"time"
)

func main() {
	a := app.New()
	w := a.NewWindow("Form")

	form := widget.NewForm(
		widget.NewFormItem("Foo", widget.NewLabel("Template Object")),
	)

	go func() {
		time.Sleep(5 * time.Second)
		button := widget.NewButton("Click Me!", func() {
			log.Println("Clicked!")
		})
		form.Items = form.Items[:0] // Cut slice to remove all items
		log.Println(len(form.Items))
		form.Append("Bar", button) // Append a new item
		/* This doesn't work either
		form.Items = []*widget.FormItem{
			widget.NewFormItem("Bar", button),
		}
		*/
		form.Refresh()
		log.Println("Form Updated")
	}()

	w.SetContent(form)
	w.Resize(fyne.NewSize(200, 200))
	w.ShowAndRun()
}
```

### Screenshot
I expect only one Item in the form, but there are two
<img width="312" alt="Screen Shot 2021-02-16 at 18 52 03" src="https://user-images.githubusercontent.com/1117080/108149600-124f9280-7088-11eb-8609-e89ee3b6ab11.png">

### Device (please complete the following information):
 - **OS:** MacOS
 - **Version:** Big Sur
 - **Go version:** go1.15 darwin/amd64
 - **Fyne version:** 1ac1d1d

andydotxyz:
Does calling Refresh after you edit the fields help?

```go
form.Items = form.Items[:0] // Cut slice to remove all items
form.Refresh()
```

stuartmscott:
No change with the code below;
```go
form.Items = form.Items[:0] // Cut slice to remove all items
log.Println(len(form.Items))
form.Refresh()
form.Append("Bar", button) // Append a new item
form.Refresh()
log.Println("Form Updated")
```

PR #5308, opened 2024-12-11
Fix changes to widget.Form.Items not being reflected on refresh

### LINKED ISSUE #1966: Cannot replace FormItem Widget
### Describe the bug:
Changes to FormItem Widgets are not visible when Form is refreshed

### Example code:
```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
	"log"
	"time"
)

func main() {
	a := app.New()
	w := a.NewWindow("Form")

	form := widget.NewForm(
		widget.NewFormItem("Foo", widget.NewLabel("Template Object")),
	)

	go func() {
		time.Sleep(5 * time.Second)
		button := widget.NewButton("Click Me!", func() {
			log.Println("Clicked!")
		})
		form.Items[0].Text = "Bar"		// This change works :)
		form.Items[0].Widget = button		// This change does not :(
		form.Refresh()
		log.Println("Form Updated")
	}()

	w.SetContent(form)
	w.Resize(fyne.NewSize(200, 200))
	w.ShowAndRun()
}
```

### Screenshot
Before
Screen Shot 2021-02-16 at 18 55 05

After
Screen Shot 2021-02-16 at 18 55 26

 - **OS:** MacOS
 - **Version:** Big Sur
 - **Go version:** go1.15 darwin/amd64
 - **Fyne version:** 1ac1d1d

coder-in-go:
### Description:
This pull request fixes the FormItem replacement issue in the Form Widget

Fixes  #1966, #1967

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/form.go
+++ b/widget/form.go
@@ -256,31 +256,35 @@ func (f *Form) checkValidation(err error) {
 }
 
 func (f *Form) ensureRenderItems() {
-	existingItemCount := len(f.itemGrid.Objects) / 2
-	if existingItemCount > len(f.Items) {
-		f.itemGrid.Objects = f.itemGrid.Objects[:len(f.Items)*2]
-		return
+	// Calculate the required capacity based on the number of items in the form
+	requiredCapacity := len(f.Items) * 2
+
+	// Pre-allocate capacity if necessary
+	if cap(f.itemGrid.Objects) < requiredCapacity {
+		newObjects := make([]fyne.CanvasObject, len(f.itemGrid.Objects), requiredCapacity)
+		copy(newObjects, f.itemGrid.Objects)
+		f.itemGrid.Objects = newObjects
 	}
+
+	// Adjust the length to match the number of items (each with label and widget)
+	f.itemGrid.Objects = f.itemGrid.Objects[:requiredCapacity]
 
 	for i, item := range f.Items {
 		labelIndex := i * 2
 		widgetIndex := labelIndex + 1
 
-		if labelIndex < len(f.itemGrid.Objects) {
+		// Update or create label for the item
+		if labelIndex < len(f.itemGrid.Objects) && f.itemGrid.Objects[labelIndex] != nil {
 			f.itemGrid.Objects[labelIndex] = f.createLabel(item.Text)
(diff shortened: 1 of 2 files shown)


widget/form.go
```diff
@@ -256,27 +256,32 @@ func (f *Form) checkValidation(err error) {
...
+			f.itemGrid.Objects = append(f.itemGrid.Objects, f.createLabel(item.Text))
 		}
 
-		objects[off] = f.createLabel(item.Text)
-		off++
-		f.setUpValidation(item.Widget, i)
-		objects[off] = f.createInput(item)
-		off++
+		if widgetIndex < len(f.itemGrid.Objects) {
+			f.setUpValidation(item.Widget, i)
```
Jacalz, line 276: This code is duplicated in both branches. Let's just move it out to before the branch.

Jacalz (changes requested):
Thanks for the PR. I left two quick comments inline. However, I do also see that this change makes the form tests fail so I think there must be something wrong with what the algorithm is trying to do. Also, it would be nice to add a test for this to verify that it does fix the issues you mention it fixing.

widget/form.go
```diff
@@ -256,27 +256,36 @@ func (f *Form) checkValidation(err error) {
...
+		} else {
+			f.itemGrid.Objects[labelIndex] = f.createLabel(item.Text)
 		}
 
-		objects[off] = f.createLabel(item.Text)
-		off++
 		f.setUpValidation(item.Widget, i)
-		objects[off] = f.createInput(item)
-		off++
+		f.itemGrid.Objects[widgetIndex] = item.Widget
```
geodimm, line 284:
```suggestion
		f.itemGrid.Objects[widgetIndex] = f.createInput(item)
```

I believe this is what breaks the tests

andydotxyz (commented):
Please only use comments in code when 
a) the code is not clear without it
b) there is no way to improve the code so that it would be clear.

We expect code to be self-documenting.

PR #5314, opened 2024-12-13
Fix clicking on empty part of widget.List not unfocusing

### LINKED ISSUE #4770: Clicking on empty part of widget.List does not unfocus other widgets
### Checklist

### Describe the bug

I noticed in Rymdport (see corresponding issue at https://github.com/Jacalz/rymdport/issues/142) that it wasn't possible to unfocus an Entry. I thought that it related to my extended entry (with accepts tabs instead of moving focus along) breaking something but I narrowed it down to `widget.List` not unfocusing other widgets in the case of it receiving a click that doesn't select any items.

### How to reproduce

1. Start the example code below.
2. Focus the entry.
3. Try to click in between the list item and notice that the entry stays focused.
4. Now click the text at the bottom of the window and notice that it does unfocus correctly.

### Screenshots

[list-unfocus-other-widgets.webm](https://github.com/fyne-io/fyne/assets/25466657/0e497d64-e331-4209-9fc4-9c9e85c4f5a5)

### Example code

```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Hello")

	unfocus := widget.NewLabel("Try to click in an empty part of the list or on this text to unfocus!")
	focus := &widget.Entry{PlaceHolder: "Focus me first!"}
	list := widget.NewList(func() int { return 1 }, func() fyne.CanvasObject { return widget.NewLabel("Item") }, func(lii widget.ListItemID, co fyne.CanvasObject) {})
	w.SetContent(container.NewBorder(focus, unfocus, nil, nil, list))
	w.Resize(w.Canvas().Size().AddWidthHeight(0, 100))

	w.ShowAndRun()
}
```

### Fyne version

v2.4.4 and develop as of dc6141425e20c0aea2877936f979070e60e0801e

### Go compiler version

1.21.8

Fedora Silverblue 39

### Additional Information

_No response_

coder-in-go:
### Description:
This PR fixes the issue where clicking on an empty part of the widget list does not unfocus other widgets.

Fixes #4770 

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -130,6 +130,7 @@ func (l *List) RefreshItem(id ListItemID) {
 		return
 	}
 	l.BaseWidget.Refresh()
+	l.Unselect(id)
 	lo := l.scroller.Content.(*fyne.Container).Layout.(*listLayout)
 	lo.renderLock.RLock() // ensures we are not changing visible info in render code during the search
 	item, ok := lo.searchVisible(lo.visible, id)
@@ -384,6 +385,17 @@ func (l *List) contentMinSize() fyne.Size {
 	height += float32(items-totalCustom) * templateHeight
 
 	return fyne.NewSize(l.itemMin.Width, height+separatorThickness*float32(items-1))
+}
+
+func (l *List) Tapped(event *fyne.PointEvent) {
+	canvas := fyne.CurrentApp().Driver().CanvasForObject(l)
+	if canvas != nil {
+		// First, unfocus the currently focused widget if any
+		if l.focused {
+			l.FocusLost()
+		}
+		canvas.Focus(l)
+	}
 }
 
 // fills l.visibleRowHeights and also returns offY and minRow


Jacalz (changes requested):
I don't understand the driver internals well enough to comment on the approach in question but this does seem to fix the issue and the code looks good to me. Would you mind adding a test case for this?

widget/list.go
```diff
@@ -130,6 +130,8 @@ func (l *List) RefreshItem(id ListItemID) {
 		return
 	}
 	l.BaseWidget.Refresh()
+
+	l.Unselect(id)
```
dweymouth, line 134: This seems like an unrelated change? Why are the semantics of RefreshItem changing to that you can't use it to refresh a selected item?

internal/driver/glfw/window.go
```diff
@@ -511,7 +511,7 @@ func (w *window) processMouseClicked(button desktop.MouseButton, action action,
 
 	co, pos, _ := w.findObjectAtPositionMatching(w.canvas, mousePos, func(object fyne.CanvasObject) bool {
 		switch object.(type) {
-		case fyne.Tappable, fyne.SecondaryTappable, fyne.DoubleTappable, fyne.Focusable, desktop.Mouseable:
+		case fyne.Tappable, fyne.SecondaryTappable, fyne.DoubleTappable, fyne.Focusable, desktop.Mouseable, desktop.Hoverable:
```
andydotxyz, line 514: I don't agree with this change.

If you put a hoverable item over a tappable item then the event will no longer get sent to the tappable item because It will be swallowed by this check for coverable.

internal/driver/glfw/window.go
```diff
@@ -540,9 +540,11 @@ func (w *window) processMouseClicked(button desktop.MouseButton, action action,
 	if wid, ok := co.(fyne.Focusable); !ok || wid != w.canvas.Focused() {
 		ignore := false
 		_, _, _ = w.findObjectAtPositionMatching(w.canvas, mousePos, func(object fyne.CanvasObject) bool {
-			switch object.(type) {
-			case fyne.Focusable:
+			if focusable, isFocusable := object.(fyne.Focusable); isFocusable {
 				ignore = true
+				if focusable != w.canvas.Focused() {
+					w.canvas.Focus(focusable)
```
andydotxyz, line 546: No, we have specifically set it so that a focusable item does not automatically get focus when tapped - the objects would have to request it.

andydotxyz (changes requested):
Sorry but this breaks driver event propagation as described in my feedback

andydotxyz:
I don't think a driver change is the right way to solve this.
The list itself gets tap events and it will know if the tap is over an event or over space...

Jacalz:
A little note on PR naming: Please avoid `Fixes issue XYZ:` in the beginning. As very few of us actually know which problem a given number corresponds to, it is better to just keep that information in the `Fixes #XYZ` part of the description. In this case I would have labelled the PR something along the lines: "Fix clicking on empty part of widget.List not unfocusing".

widget/list.go
```diff
@@ -386,6 +387,17 @@ func (l *List) contentMinSize() fyne.Size {
...
 }
 
+func (l *List) Tapped(event *fyne.PointEvent) {
+	canvas := fyne.CurrentApp().Driver().CanvasForObject(l)
+	if canvas != nil {
+		// First, unfocus the currently focused widget if any
+		if l.focused {
+			l.FocusLost()
+		}
+		canvas.Focus(l)
```
andydotxyz, line 397: the PR says it is unfocusing but this code shows focussing - is this correct?

widget/list.go
andydotxyz: But this code is not allowing an item to remain selected when it is refreshed - so scrolling will lose the selected status

andydotxyz (changes requested):
The PR looks to be doing the opposite of what the code specifies, is this a mistake?

With a unit test we would be able to verify this corrected behaviour.

dweymouth:
PR seems to have been abandoned and as mentioned in review is not really fixing the root cause. Closing

PR #5322, opened 2024-12-16
Move events to main thread and remove unneeded locks

andydotxyz:
This means for race-free apps except for data binding and goroutines, which are covered later, separately.

Progresses #4654

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/widget.go
+++ b/widget/widget.go
@@ -2,7 +2,6 @@
 package widget // import "fyne.io/fyne/v2/widget"
 
 import (
-	"sync"
 	"sync/atomic"
 
 	"fyne.io/fyne/v2"
@@ -19,9 +18,8 @@ type BaseWidget struct {
 	position async.Position
 	Hidden   bool
 
-	impl         atomic.Pointer[fyne.Widget]
-	propertyLock sync.RWMutex
-	themeCache   fyne.Theme
+	impl       atomic.Pointer[fyne.Widget]
+	themeCache fyne.Theme
 }
 
 // ExtendBaseWidget is used by an extending widget to make use of BaseWidget functionality.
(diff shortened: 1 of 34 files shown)


widget/widget.go
```diff
@@ -19,9 +18,8 @@ type BaseWidget struct {
 	position async.Position
 	Hidden   bool
 
-	impl         atomic.Pointer[fyne.Widget]
-	propertyLock sync.RWMutex
-	themeCache   fyne.Theme
+	impl       atomic.Pointer[fyne.Widget]
```
Jacalz, line 21: Should atomic types and things from the `async` package also be removed or is that perhaps better to have as a separate PR?

Jacalz (commented):
Nice work. Just one minor comment (not necessarily request for changes) inline. On more question: are there any parts where we should still support calling methods on different threads? Just asking so we don't get carried away and rip out too many locks at once :)

Jacalz:
FYI: There seems to be compilation issues?

andydotxyz:
> FYI: There seems to be compilation issues?

Yeah I just noteiced I didn't compile the tests for last refactor. Updated

widget/widget.go
andydotxyz: I was wondering the same. I think later PR as it could impact semantics too

andydotxyz:
> On more question: are there any parts where we should still support calling methods on different threads?

I think it's just data binding which requires more work for various reasons.

andydotxyz:
Looks like there may be a deadlock on Linux only - I'll check it out tomorrow

Jacalz (approved):
Seems to work fine on Linux for me. Maybe deadlock is for tests only?

andydotxyz:
> Seems to work fine on Linux for me. Maybe deadlock is for tests only?

Yes it was, or a codepath that was not executed on the testing. Fixed.

andydotxyz:
One thing led to another and I just refactored the event queue out in this PR too.

Fixes a few flakey tests :)

PR #5327, opened 2024-12-17
driver/glfw: Merge waitForStart and done channels

Jacalz:
### Description:

With the threading changes, we no longer need to have separate channels for done and waitForStart given that showing a window no longer is expected to happen on the wrong thread. This allows allows us to first wait for starting and then reopen the channel to wait for shutting down. I.e. we no longer carry around a useless startup channel once we are running.

Follow up to https://github.com/fyne-io/fyne/pull/5322

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/driver.go
+++ b/internal/driver/glfw/driver.go
@@ -33,10 +33,9 @@ var curWindow *window
 var _ fyne.Driver = (*gLDriver)(nil)
 
 type gLDriver struct {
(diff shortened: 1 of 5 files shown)


Jacalz:
Better solution was implemented in https://github.com/fyne-io/fyne/pull/5328#pullrequestreview-2519582934.

PR #5330, opened 2024-12-17
Update go-locale version for full language integration on macos

sdassow:
### Description:

With this the preferred language from the languages list is used as it should on macOS.

### Checklist:

- [ ] Tests included.
- [X] Tests all pass.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/go-text/typesetting v0.2.0
 	github.com/godbus/dbus/v5 v5.1.0
 	github.com/jackmordaunt/icns/v2 v2.2.6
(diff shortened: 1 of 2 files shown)


Jacalz:
Please don't remove the PR checklist. It is meant to be filled out :)

sdassow:
Ok, added them back. Only removed them because they don't mean much for this PR I think.

Jacalz:
Sorry to be picky but they are meant to stay. However, you should not fill in the one about adding tests if you haven't added any.

sdassow:
Right, changed. There are no new tests, but existing tests cover it, which I've used as qualifier ;-).

dweymouth (approved):

PR #5334, opened 2024-12-20
Add entry.Insert API to insert text at current cursor position

### LINKED ISSUE #3445: widget.Entry.InsertText
### Checklist

A method that will insert text at current caret position to an Entry widget. If some text is selected, it will replace it.

Yes, but it's complicated: you must get the current text, get the caret position (col,row) and convert this to a position in the text stream (no simple method to do this, especialy in multiline Entry), insert the text, then set the text.

`widget.Entry.InsertText(text string)`

dweymouth:
### Description:

Fixes #3445 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -908,6 +908,53 @@ func (e *Entry) Unbind() {
 	e.binder.Unbind()
 }
 
+// Insert inserts the given text into the Entry at the current cursor position.
+// If text is currently selected, the selected text will be replaced
+// with the inserted content.
+//
+// Since: 2.6
+func (e *Entry) Insert(text string) {
+	if text == "" {
+		changed := e.selecting && e.eraseSelection()
+
+		if changed {
+			e.Refresh()
+		}
+
+		return // Nothing to paste into the text content.
+	}
+
+	if !e.MultiLine {
+		// format clipboard content to be compatible with single line entry
+		text = strings.Replace(text, "\n", " ", -1)
+	}
+
+	if e.selecting {
+		e.eraseSelection()
+	}
+
+	runes := []rune(text)
+	pos := e.cursorTextPos()
+	provider := e.textProvider()
+	provider.insertAt(pos, runes)
+
+	e.undoStack.Add(&entryModifyAction{
+		Position: pos,
+		Text:     runes,
+	})
+	content := provider.String()
+	e.updateText(content, false)
+	e.CursorRow, e.CursorColumn = e.rowColFromTextPos(pos + len(runes))
+	cb := e.OnChanged
+
+	e.validate()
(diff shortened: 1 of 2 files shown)


andydotxyz:
I think the review comments were removed by the force push.

dweymouth:
> I think the review comments were removed by the force push.

I actually think conversation had only been on the linked issue so far and not the PR. But noted on the force push, will use merge commit next time :)

andydotxyz:
> > I think the review comments were removed by the force push.
> 
> I actually think conversation had only been on the linked issue so far and not the PR. But noted on the force push, will use merge commit next time :)

Ah, that makes sense, sorry for the confusion.

widget/entry.go
```diff
@@ -908,6 +908,53 @@ func (e *Entry) Unbind() {
...
+		changed := e.selecting && e.eraseSelection()
+
+		if changed {
+			e.Refresh()
+		}
+
+		return // Nothing to paste into the text content.
+	}
+
+	if !e.MultiLine {
```
andydotxyz, line 927: I think this code is specific to clipboard so maybe it should be outside of the new insert API?

If you think it does belong here please document it in the API doc.

andydotxyz (changes requested):
Thanks for this - just one note about potentially unexpected behaviour...

widget/entry.go
dweymouth: @matwachich Do you have an opinion on whether the InsertAtCursor API should also reformat multiline text when inserting into a single-line entry (like paste does)? I feel like it maybe shouldn't since it'd be the developer who controls what is being inserted, vs a user action like paste.

matwachich:
Default behaviour could be to convert line breaks to spaces

Le ven. 3 janv. 2025 à 20:01, Drew Weymouth ***@***.***> a
écrit :

dweymouth:
> Default behaviour could be to convert line breaks to spaces Le ven. 3 janv. 2025 à 20:01, Drew Weymouth ***@***.***> a 

There won't be a *default* behavior but a single, unchangable behavior since the API doesn't have any other arguments for options. For that reason I'm leaning toward inserting the text exactly as-is, and documenting it as such. It's more flexible since the programmer is in control of the text passed to the API, so if they want to remove newlines, or process the string in any way, they can do so. But if we process the string, there would be no way to disable it.

matwachich:
Makes sense...
Agree

Le ven. 3 janv. 2025 à 21:29, Drew Weymouth ***@***.***> a
écrit :

dweymouth:
Merge conflicts... would be easier to start fresh

PR #5335, opened 2024-12-21
Ensure text and SVG textures re-rasterize when window is dragged to new monitor

### LINKED ISSUE #5323: Text strings become blurry when dragging window from low DPI to high DPI monitor
### Checklist

### Describe the bug

New in 2.5.3: Text that was first rendered on a low DPI monitor doesn't re-render when the window moves to a high DPI monitor, instead the GPU just rescales the pre-existing texture(?), resulting in blurriness.

SVG rasterizations also seem to be affected, though the root cause may be different

### How to reproduce

launch a Fyne app on a low DPI monitor, then drag it to a high DPI monitor

### Screenshots

_No response_

### Example code

fyne_demo

### Fyne version

2.5.3

### Go compiler version

n/a

ARM Mac, though probably all affected

### Additional Information

_No response_

dweymouth:
### Description:
A regression in 2.5.3 resulted in text and SVG rasterizations not re-rasterizing when the window was moved between monitors, resulting in blurriness if the window went from low DPI to high DPI.

Fixes #5323 

### Checklist:

- [ ] Tests included. -- Tested manually
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window_wasm.go
+++ b/internal/driver/glfw/window_wasm.go
@@ -11,6 +11,7 @@ import (
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/container"
 	"fyne.io/fyne/v2/driver/desktop"
+	"fyne.io/fyne/v2/internal/cache"
 	"fyne.io/fyne/v2/internal/painter/gl"
 	"fyne.io/fyne/v2/internal/scale"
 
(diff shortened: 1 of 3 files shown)


internal/driver/glfw/window_wasm.go
```diff
@@ -474,7 +475,11 @@ func (w *window) rescaleOnMain() {
 		scale.ToFyneCoordinate(w.canvas, w.width),
 		scale.ToFyneCoordinate(w.canvas, w.height))
 	w.canvas.Resize(scaledFull)
-	return
+
+	// Ensure textures re-rasterize at the new scale
+	cache.DeleteTextTexturesFor(w.canvas)
+	w.canvas.content.Refresh()
+	//return
```
Jacalz, line 482: I think committing commented out code goes against our contribution guide. I guess that applies to the code below as well?

Jacalz (commented):
I can't quite comment on the approach but I left a minor comment inline.

internal/driver/glfw/window_wasm.go
dweymouth: Done. I suppose the git history will always have the commented-out lines if we ever need to switch back (I have no idea why it was commented-out in the first place)

andydotxyz (approved):
Thanks for fixing this oversight

PR #5339, opened 2024-12-24
Fix issue where linux app may crash with a big icon

### LINKED ISSUE #5272: FyneApp.toml causes BadLength error
### Checklist

### Describe the bug

Current go app runs without a FyneApp.toml; upon inserting one (even copying the example in the Docs) causes the following error at runtime: 
```
X Error of failed request:  BadLength (poly request too large or internal Xlib length error)
  Major opcode of failed request:  18 (X_ChangeProperty)
  Serial number of failed request:  246
  Current serial number in output stream:  246
```
Removing the toml results in a successful build and run. 

### How to reproduce

1. Add FyneApp.toml
2. Try `go build`, `fyne build`, or `go run main.go` and run executable
3. Error occurs

### Screenshots

_No response_

### Example code

```go
func main() {
	prog := app.New()
	mainWindow := prog.NewWindow("Expense Report Builder")
	mainWindow.Resize(fyne.Size{Height: 412, Width: 720})

	templateRow, templatePathBox := inputRow(&mainWindow, "Template: ", FILE)
	rosterRow, rosterPathBox := inputRow(&mainWindow, "Roster: ", FILE)
	outputRow, outputPathBox := inputRow(&mainWindow, "Output: ", FOLDER)
	runButton := widget.NewButton("Create Expense Reports", func() {
		err := business(rosterPathBox.Text, templatePathBox.Text, outputPathBox.Text)
		if err != nil {
			dialog.ShowError(err, mainWindow)
		}
	})

	mainWindow.SetContent(container.NewStack(
		container.NewVBox(
			templateRow, rosterRow, outputRow, runButton,
		),
	))

	mainWindow.ShowAndRun()
	// TODO Build ui
	// TODO Verify additions from Github
}
```

### Fyne version

2.5.2

### Go compiler version

1.23.2

Arch Linux

### Additional Information

This only appears to be an issue with the example code provided; simple "Hello, world" programs work (as well as the fyne demo app).

andydotxyz:
Fixes #5272

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


internal/metadata/icon.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func ScaleIcon(data fyne.Resource, size int) fyne.Resource {
+	img, err := png.Decode(bytes.NewReader(data.Content()))
+	if err != nil {
+		fyne.LogError("Failed to decode app icon", err)
+		return data
+	}
+
+	if img.Bounds().Dx() <= size {
```
dweymouth, line 20: How do we handle non-square icons? We are only checking the X dimension here, it's possible the Y dimension could be larger, unless we explicitly don't support non-square?

internal/metadata/icon.go
```diff
@@ -0,0 +1,41 @@
...
+		return data
+	}
+
+	sized := resize.Resize(uint(size), uint(size), img, resize.Lanczos3)
```
dweymouth, line 24: Same concern here, it appears we're resizing to square always rather than maintaining the aspect ratio?

internal/metadata/icon.go
andydotxyz: Only square icons are supported

internal/metadata/icon.go
andydotxyz: Yup, square is the only format we support. Most OS expect square so resizing here may even be an improvement if people accidentally pass off-square...

dweymouth (approved):

PR #5344, opened 2024-12-28
Add a new InfiniteAnimation

### LINKED ISSUE #4735: Add a way to create an infinite animation
### Checklist

The `fyne.NewAnimation` API takes a duration argument and ends the animation after the given duration. I had thought that a duration of 0 might be used to create an infinite animation, but it does not, nor is it documented as such.

No. Closest thing is to use math.MaxInt64 as the duration which is about 290 years (which is good enough for all practical purposes but not clean code).

An API `fyne.NewIndefiniteAnimation(tick func(f float64))` or a clarification of the existing `fyne.NewAnimation` that 0 can be used for infinite duration (since an animation of duration 0 is not useful anyway.

dweymouth:
### Description:
Adds a new type of animation that has no curve or duration, and simply ticks a function every frame until stopped.
This will be useful in 2.6 to schedule an ongoing "update loop" to run on the main app thread, for example in the `fyne-io/life` repo to replace the goroutine that currently drives the updates.

**Note:** Since this is implemented entirely using the existing animation type, it's debatable whether it's even needed as it can be fully implemented in user code with the current set of public APIs. However, it's more clean to have a public API specifically for this purpose and a Tick function that takes no arguments, instead of user code having to supply one that simply ignores its argument.

Also, it's possible other names could be more descriptive of its more general use case, like `fyne.FrameTicker`?

Fixes #4735 

### Checklist:

- [ ] Tests included. // we don't currently have animation tests
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/animation.go
+++ b/animation.go
@@ -44,6 +44,17 @@ type Animation struct {
 	Tick        func(float32)
 }
 
+// IndefiniteAnimation represents an animation that continues indefinitely. It has no duration
+// or curve, and when started, the Tick function will be called every frame until Stop is invoked.
+//
+// Since: 2.6
+type IndefiniteAnimation struct {
+	Tick func()
+
+	isSetup   bool
+	animation Animation
+}
+
 // NewAnimation creates a very basic animation where the callback function will be called for every
 // rendered frame between [time.Now] and the specified duration. The callback values start at 0.0 and
 // will be 1.0 when the animation completes.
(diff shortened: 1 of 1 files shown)


animation.go
```diff
@@ -53,16 +64,50 @@ func NewAnimation(d time.Duration, fn func(float32)) *Animation {
...
+		i.animation = Animation{
+			Tick: func(_ float32) {
+				i.Tick()
+			},
+			Curve:       AnimationLinear, // any curve will work
+			Duration:    1 * time.Second, // anything positive here will work
+			RepeatCount: AnimationRepeatForever,
+		}
+		i.isSetup = true
+	}
```
Jacalz, line 108: It would match our style to indent less by instead inverting your if-statement and retuning if it is set up.

animation.go
```diff
@@ -53,16 +64,50 @@ func NewAnimation(d time.Duration, fn func(float32)) *Animation {
...
+// Start registers the animation with the application run-loop and starts its execution.
+func (i *IndefiniteAnimation) Start() {
+	i.setupAnimation()
+	i.animation.Start()
+}
+
 // Stop will end this animation and remove it from the run-loop.
 func (a *Animation) Stop() {
 	CurrentApp().Driver().StopAnimation(a)
 }
```
Jacalz, line 89: I'm not a fan of how the methods of different types are interleaved. I'd much rather have all methods for one typ after each other in alphabetical order before the next type.

Jacalz (changes requested):

Jacalz:
Whoops. Forgot to add the message again. I like the idea of this but I'm not entirely sure if this is the best API for it (not necessarily against but an animation that just passes random curve and duration seems a bit strange). :)

dweymouth:
> Whoops. Forgot to add the message again. I like the idea of this but I'm not entirely sure if this is the best API for it (not necessarily against but an animation that just passes random curve and duration seems a bit strange). :)

That part is just an implementation detail hidden away from the public API. I did it that way to avoid having to do a much larger code change making the animation scheduler aware of a new type. The implementation details could always be changed later, though if you have any other suggestions for how to implement this, I'm all ears.

animation.go
Jacalz: To clarify: I'd rather have struct, methods, struct, methods and so on. I think that's how Go files generally are structured. Without your latest change it is struct, struct, methods, methods?

andydotxyz (changes requested):
I think it makes sense to have a clean way to have such an infinite animation, rather than exposing the details in docs or example.

However the word "indefinite" sounds somewhat confusing - would infinite not be more intuitive (and match the progress widget wording).

andydotxyz (approved):
Great thanks

Jacalz (approved):

PR #5345, opened 2024-12-28
Add new canvas.RecolorSVG API

### LINKED ISSUE #4815: Internal support for custom colored fyne.Resource theming.
### Checklist

It is possible to change the color of fyne.Resource objects?

It is possible to do so, like this:
`theme.NewErrorThemedResource(theme.AccountIcon())`

Which makes the icon red, but I would like to able to define any kind of custom color for the resource.

It might be fairly simple to add this to the theme package.

For the code, I used the exact same approach as it is already present in the theme/icons.go file from the fyne internal library, but appended it with an additional `color` property

```go
type CustomColoredResource struct {
	source fyne.Resource
	color  color.Color
}

func NewCustomColoredResource(orig fyne.Resource, color color.Color) *CustomColoredResource {
	res := &CustomColoredResource{source: orig, color: color}
	return res
}

// Name returns the underlying resource name (used for caching).
func (res *CustomColoredResource) Name() string {
	return "custom_" + res.source.Name() // I do not know what is the relevance of caching here, might need to change the 'custom_' prefix with a unique value based on the color + the resource Name? Not sure.
}

func (res *CustomColoredResource) Content() []byte {
	return Colorize(res.source.Content(), res.color)
}

func (res *CustomColoredResource) Original() fyne.Resource {
	return res.source
}
```

I would like to see something like this:
```go
myCustomColor := color.NRGBA{R: 0x99, G: 0xff, B: 0x99, A: 0xff}
theme.NewCustomColoredResource(theme.AccountIcon(), myCustomColor)
```

Do you think this would be a nice idea?

dweymouth:
### Description:

Fixes #4815

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/canvas/canvas.go
+++ b/canvas/canvas.go
@@ -20,7 +20,7 @@ func Refresh(obj fyne.CanvasObject) {
 	}
 }
 
-// ColorizeSVG takes a fyne.Resource containing SVG content, and returns
+// RecolorSVG takes a fyne.Resource containing SVG content, and returns
 // the raw SVG content, re-colorized to be monochrome with the given color.
 // The content can be assigned to a new fyne.StaticResource with an appropriate name
 // to be used in a widget.Button, canvas.Image, etc.
@@ -29,7 +29,7 @@ func Refresh(obj fyne.CanvasObject) {
 // and a non-nil error is returned.
 //
 // Since: 2.6
-func ColorizeSVG(resource fyne.Resource, color color.Color) ([]byte, error) {
+func RecolorSVG(resource fyne.Resource, color color.Color) ([]byte, error) {
 	return svg.ColorizeError(resource.Content(), color)
 }
 
--- a/canvas/canvas.go
+++ b/canvas/canvas.go
@@ -1,6 +1,11 @@
 package canvas
 
-import "fyne.io/fyne/v2"
+import (
+	"image/color"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/internal/svg"
+)
 
 // Refresh instructs the containing canvas to refresh the specified obj.
 func Refresh(obj fyne.CanvasObject) {
@@ -13,6 +18,19 @@ func Refresh(obj fyne.CanvasObject) {
 	if c != nil {
 		c.Refresh(obj)
 	}
+}
+
+// ColorizeSVG takes a fyne.Resource containing SVG content, and returns
+// the raw SVG content, re-colorized to be monochrome with the given color.
+// The content can be assigned to a new fyne.StaticResource with an appropriate name
+// to be used in a widget.Button, canvas.Image, etc.
+//
+// If an error occurs, the returned content will be the content of the input Resource,
+// and a non-nil error is returned.
+//
+// Since: 2.6
+func ColorizeSVG(resource fyne.Resource, color color.Color) ([]byte, error) {
+	return svg.ColorizeError(resource.Content(), color)
 }
 
 // repaint instructs the containing canvas to redraw, even if nothing changed.
--- a/canvas/canvas.go
+++ b/canvas/canvas.go
@@ -20,17 +20,17 @@ func Refresh(obj fyne.CanvasObject) {
 	}
 }
 
-// RecolorSVG takes a fyne.Resource containing SVG content, and returns
-// the raw SVG content, re-colorized to be monochrome with the given color.
+// RecolorSVG takes a []byte containing SVG content, and returns
+// new SVG content, re-colorized to be monochrome with the given color.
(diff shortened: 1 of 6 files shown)


internal/svg/svg.go
```diff
@@ -22,19 +22,32 @@ import (
...
+// ColorizeError is the same as Colorize, except returning instead of logging any error.
+func ColorizeError(src []byte, clr color.Color) ([]byte, error) {
+	var err error
+	content := colorizeImpl(src, clr, func(s string, e error) {
+		err = fmt.Errorf("%s %s", s, e.Error())
+	})
+	return content, err
+}
+
+func colorizeImpl(src []byte, clr color.Color, errFn func(string, error)) []byte {
```
Jacalz, line 37: This error function seems a bit strange in my opinion. Why not just use the return value of the function directly and call fyne.LogError or similar on that?

Jacalz (changes requested):

internal/svg/svg.go
dweymouth: Yeah now that I look at it again, it is a bit strange. It was to minimize both code changes and duplicated code. Fyne.LogError takes 2 arguments, so I guess the alternative is for the signature to be `func colorizeImpl(src []byte, clr color.Color) ([]byte, string, error)` so that the actual error and the context we add to it can still be passed separately to fyne.LogError. What do you think?

Jacalz: Do you even need the extra string information? It just seems unintuitive from how Go error handling usually is made? We can lose a bit in the FyneLog error and add any necessary additional information in the error text?

dweymouth: Updated to just returning a regular error

dweymouth:
CI Failures appear to be unrelated flakiness

canvas/canvas.go
```diff
@@ -15,6 +20,19 @@ func Refresh(obj fyne.CanvasObject) {
...
+// RecolorSVG takes a fyne.Resource containing SVG content, and returns
+// the raw SVG content, re-colorized to be monochrome with the given color.
+// The content can be assigned to a new fyne.StaticResource with an appropriate name
+// to be used in a widget.Button, canvas.Image, etc.
+//
+// If an error occurs, the returned content will be the content of the input Resource,
+// and a non-nil error is returned.
+//
+// Since: 2.6
+func RecolorSVG(resource fyne.Resource, color color.Color) ([]byte, error) {
```
andydotxyz, line 32: It feels like an impedance mismatch to take in a Resource but return a []byte...

If we are expecting people to create an appropriate name for a resource we probably need to be more specific that it should contain the colour name to avoid cache collisions - can we solve this problem for them?

I think it may need to be []byte in and []byte out because this is used internally to ThemedResource handling, so being in canvas package probably implies the lower level API.

andydotxyz (changes requested):

canvas/canvas.go
dweymouth: I am in favor of []byte in and []byte out vs trying to automatically derive a resource name. I'll switch it to that

internal/svg/svg.go
```diff
@@ -22,22 +22,28 @@ import (
...
 func Colorize(src []byte, clr color.Color) []byte {
+	content, err := ColorizeError(src, clr)
+	if err != nil {
+		fyne.LogError("", err)
+	}
+	return content
+}
+
+// ColorizeError is the same as Colorize, except returning instead of logging any error.
+func ColorizeError(src []byte, clr color.Color) ([]byte, error) {
```
Jacalz, line 33: Sorry for the pickiness but I'm not sure about the function name here. ColorizeError sounds as if it is changing the colour of an error and not an svg?

Jacalz (changes requested):

internal/svg/svg.go
dweymouth: Do you have a better name in mind? `ColorizeReturnError`?

Jacalz: Not really, no. I think having in the name that something returns an error will be strange. Also the idea of having a version that returns errors and one that doesn't is strange in most cases. I think there are two options:

1. Depending on how many call sites there are, it might make sense to only use the version that returns errors and either look at it or ignore it completely.
2. The standard library seems to follow the convention that a function always returns an error and in the case where there is a version of it that doesn't, it is called MustXYZ (although the latter usually panics on errors).

It is a bit of a strange pattern in general come to think of it. Any ideas?

andydotxyz: I would agree with 1 - there is little point in exposing the non-error as a public helper API if errors can be generated.

andydotxyz (approved):
Thanks, this looks like a great new helper

PR #5349, opened 2025-01-01
Add tap-to-scroll to the scroll bars

### LINKED ISSUE #4922: When scroll bar is expanded, clicking above or below the bar should scroll up or down
### Checklist

Fyne's scroll bars do not support the standard behavior of clicking above or below the scroll bar itself, when it is "activated" (ie in its expanded state), to scroll up or down.

_No response_

Clicking above or below the scroll bar scrolls up or down.

dweymouth:
### Description:
Tapping above, below, left, or right of a scroll bar will scroll the content up, down, left, or right by a (nearly) full page, respectively. When the scroll bar is large, a new background rectangle is drawn behind it to indicate the taps will go to the scroller and not the underlying content.

Fixes #4922 

https://github.com/user-attachments/assets/978a52a8-fac4-45e3-a098-15637ca1fa9a

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (changes requested):
Thanks. I just tested this some more and the behaviour is not quite what I would expect. Should the scroller not go to where I ave clicked? Look at this page in the browser and you should see that clicking on the scroller moves the middle of the scroller to where it was clicked. This does not happen if you for example go to fyne_demo and the Collections -> List page.

dweymouth:
I based it on what Firefox does. I checked with this page https://worlds-highest-website.com/ - when you tap below the scroll bar, no matter how far down, it barely moves the bar because the page is so long. I just tested with Chromium too and it does the same thing

Jacalz:
This is how Firefox behaves on my end. It seems much more intuitive in my opinion to move to where I clicked rather than just some set of pages at a time:
[Screencast From 2025-01-07 21-09-51.webm](https://github.com/user-attachments/assets/4428d5b0-af34-4c53-b457-db96204c98c6)

dweymouth:
Oh no.. I guess it may be platform-specific. In which case we probably want to do the same, since we try to fit in with platform conventions. Will test browser behavior on my Linux system too, and Windows VM, but I feel like as a MacOS user, I'm already subconsciously expecting the scroll behavior I implemented here.

dweymouth:
Oh, it's a MacOS system setting. Yeah, we should probably try to respect that and read it from the ObjC API, and implement your suggested behavior for all other platforms:

https://www.reddit.com/r/Windows10/comments/kjxz1v/scrollbar_jump_to_position_setting/

andydotxyz:
I agree we should try to read the system setting, on macOS and if the same is set on other systems.

However I think the default may need to be "tap to location" because even on macOS apps like VS Code and GoLand do that irrespective of the OS setting...

sdassow:
Really nice feature!

> I agree we should try to read the system setting, on macOS and if the same is set on other systems.
> 
> However I think the default may need to be "tap to location" because even on macOS apps like VS Code and GoLand do that irrespective of the OS setting...

That sounds almost like it should be configurable, even though I wouldn't know on what level. Maybe as a theme settings so it doesn't affect existing APIs?

andydotxyz:
> That sounds almost like it should be configurable, even though I wouldn't know on what level. Maybe as a theme setting so it doesn't affect existing APIs?

I think that would be worse than just having our preference because apps that look the same will behave differently based on the developers preference.
We could make it a fyne setting like primary colour or the animations disabling?

dweymouth:
I think we could just start with the scroll-to-tapped behavior (I have to update the PR), and later add in the scroll-one-page based on OS settings. I couldn't even find out how to read the MacOS setting programmatically though.

Jacalz (approved):
Looking good on my end now. Thanks :)

dweymouth:
> Looking good on my end now. Thanks :)

Whoops, just fixed another bug - needs re-review :)

Jacalz (approved):

PR #5359, opened 2025-01-04
Add an environment variable to disable scale adjustment based on detected DPI

### LINKED ISSUE #5164: Ability to turn off re-scaling when a window moves between monitors
### Checklist

See the details here - https://github.com/dweymouth/supersonic/issues/489 some users of Fyne apps find the auto-scaling when windows move between monitors jarring and would like to be able to disable it.

_No response_

An API or environment variable setting for disabling re-scaling of windows when moved between monitors. They should keep the same pixel size.

dweymouth:
### Description:

Adds a new environment variable `FYNE_DISABLE_DPI_DETECTION` which can be set to `true`, `True`, `TRUE` or `1` to disable detecting the screen DPI and adjusting scaling accordingly.

I do not have a dual-monitor Linux setup so I can only confirm that it results in a new Fyne app being drawn at 1:1 pixel scale on a Hi-DPI monitor.

Fixes #5164 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window_desktop.go
+++ b/internal/driver/glfw/window_desktop.go
@@ -9,6 +9,7 @@ import (
 	_ "image/png" // for the icon
 	"os"
 	"runtime"
+	"strings"
 	"sync"
 
 	"fyne.io/fyne/v2"
@@ -301,16 +302,16 @@ func (w *window) getMonitorForWindow() *glfw.Monitor {
 }
 
 func (w *window) detectScale() float32 {
-	// check if DPI detection is disabled
-	env := os.Getenv(disableDPIDetectionEnvKey)
-	switch env {
-	case "True", "TRUE", "true", "T", "t", "1":
-		return 1
-	}
-
 	if build.IsWayland { // Wayland controls scale through content scaling
 		return 1
 	}
+
+	// check if DPI detection is disabled
+	env := os.Getenv(disableDPIDetectionEnvKey)
+	if strings.EqualFold(env, "true") || strings.EqualFold(env, "t") || env == "1" {
+		return 1
+	}
+
 	monitor := w.getMonitorForWindow()
 	if monitor == nil {
 		return 1
--- a/internal/driver/glfw/window_desktop.go
+++ b/internal/driver/glfw/window_desktop.go
@@ -303,7 +303,8 @@ func (w *window) getMonitorForWindow() *glfw.Monitor {
 func (w *window) detectScale() float32 {
 	// check if DPI detection is disabled
 	env := os.Getenv(disableDPIDetectionEnvKey)
-	if len(env) > 0 && (env[0] == 'T' || env[0] == 't' || env[0] == '1') {
+	switch env {
+	case "True", "TRUE", "true", "T", "t", "1":
 		return 1
 	}
 
--- a/internal/driver/glfw/window_desktop.go
+++ b/internal/driver/glfw/window_desktop.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"image"
 	_ "image/png" // for the icon
+	"os"
 	"runtime"
 	"sync"
 
@@ -24,7 +25,10 @@ import (
 	"github.com/go-gl/glfw/v3.3/glfw"
 )
 
-const defaultTitle = "Fyne Application"
+const (
+	defaultTitle              = "Fyne Application"
+	disableDPIDetectionEnvKey = "FYNE_DISABLE_DPI_DETECTION"
+)
 
 // Input modes.
 const (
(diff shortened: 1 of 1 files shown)


Jacalz:
I doubt many Linux users will set this environment variable. Is it not better to fix the issue and/or rework how we handle window scaling? Something is obviously wrong there from what I can tell from the ticket.

dweymouth:
There's two separate things being discussed in the ticket. The original conversation was about something that's not a bug, but a user preference to disable a window re-adjusting its scale when moving between monitors

Jacalz:
Alright then but maybe it is better to add it into fyne_settings instead in that case? We could add a page for advanced options.

My gripe with this is relay that a user needs to read the documentation to know that the environment variable exists (that needs to be maintained and not rot away as well) and then define it either for each app individually or globally. I doubt many users will do so.

dweymouth:
> Alright then but maybe it is better to add it into fyne_settings instead in that case? We could add a page for advanced options.

We could do that as well, similar to how FYNE_SCALE is done (I think). I'd like to avoid a Fyne app user having to download and run a different app (fyne settings) to adjust this.

andydotxyz:
> I'd like to avoid a Fyne app user having to download and run a different app (fyne settings) to adjust this.

You can embed the Fyne settings panel in your app if you like :)

dweymouth:
> > I'd like to avoid a Fyne app user having to download and run a different app (fyne settings) to adjust this.
> 
> You can embed the Fyne settings panel in your app if you like :)

Won't that adjust it for all apps though instead of just mine? Which is not what the user would expect. And also I already have my own settings UI with other settings specific to my app.

I can add this to the settings app as well, keeping in mind it would need a new public getter API like `fyne.CurrentApp().Settings().DPIDetectionDisabled()`. But I strongly would urge keeping the env variable too, like we have for `FYNE_SCALE`

What do you think?

internal/driver/glfw/window_desktop.go
```diff
@@ -297,6 +301,12 @@ func (w *window) getMonitorForWindow() *glfw.Monitor {
 }
 
 func (w *window) detectScale() float32 {
+	// check if DPI detection is disabled
+	env := os.Getenv(disableDPIDetectionEnvKey)
+	if len(env) > 0 && (env[0] == 'T' || env[0] == 't' || env[0] == '1') {
```
andydotxyz, line 306: So "Talse" or "tomatoes" is to be treated as true?

andydotxyz:
> Won't that adjust it for all apps though instead of just mine?

Yes

internal/driver/glfw/window_desktop.go
dweymouth: I can switch to a full proper list of supported terms (1, true, TRUE, True, t, T)

Jacalz:
> Won't that adjust it for all apps though instead of just mine? Which is not what the user would expect.

If the app is sandboxed, like Flatpak for example, then it is only for your application.

internal/driver/glfw/window_desktop.go
```diff
@@ -297,6 +301,13 @@ func (w *window) getMonitorForWindow() *glfw.Monitor {
 }
 
 func (w *window) detectScale() float32 {
+	// check if DPI detection is disabled
+	env := os.Getenv(disableDPIDetectionEnvKey)
+	switch env {
+	case "True", "TRUE", "true", "T", "t", "1":
```
Jacalz, line 307: You might want to use `strings.EqualFold` instead of hardcoding values. That should allow the word "true" no matter how it is capitalised.

dweymouth:
> If the app is sandboxed, like Flatpak for example, then it is only for your application.

Yes, though not all installations will be Flatpak. It sounds like you're in favor of adding a settings app control for this as well as the env var?

internal/driver/glfw/window_desktop.go
```diff
@@ -297,6 +301,13 @@ func (w *window) getMonitorForWindow() *glfw.Monitor {
...
+		return 1
+	}
+
 	if build.IsWayland { // Wayland controls scale through content scaling
```
Jacalz, line 311: I think it would be wise to move the env check to after this check. In the case of Wayland, there is no need to check the environment variable at all and the compiler can just deadcode eliminate most of the function.

Jacalz (changes requested):

Jacalz:
> Yes, though not all installations will be Flatpak.

Indeed. I just wanted to add that information to the discussion :)

> It sounds like you're in favor of adding a settings app control for this as well as the env var?

Was this meant as a setting mostly to help out on Linux or is it useful on all platforms? I was initially in favour of the settings page but in my latest review I looked at the code and saw that Wayland handles scaling themselves. If it is Linux specific, it seems strange to add a settings toggle for it when Wayland will take over eventually (not talking about the ENV var).

dweymouth:
I do think it is mainly a Linux thing, since at least MacOS handles it at the OS level I think (not really sure about Windows). So yeah maybe just env var makes sense.

andydotxyz:
> I do think it is mainly a Linux X11 thing, since at least MacOS handles it at the OS level I think (not really sure about Windows). So yeah maybe just env var makes sense.

macOS and Windows provide a system check first - so global user settings are respected.

dweymouth:
Are there any other code suggestions for this?

andydotxyz (approved):
Cool, thanks for this, I'm sure a few folk will like to control scaling at a lower level.

Jacalz (approved):

PR #5360, opened 2025-01-04
Prevent tapping within a popup from dismissing it, even if non-modal

### LINKED ISSUE #5358: Basic PopUp should not close when tapped inside
### Discussed in https://github.com/fyne-io/fyne/discussions/5340

<div type='discussions-op-text'>

<sup>Originally posted by **typenoob** December 27, 2024</sup>
Hi, I want to make a popup which would be closed when tap on the backdrop but remain on the screen if tap inside of it. The `NewPopUp` or `NewModalPopUp` method does not satisfy my requirement perfectly. Is there any solution for it conveniently? Appreciate for any help!  </div>

dweymouth:
### Description:

Fixes #5358 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/popup.go
+++ b/widget/popup.go
@@ -78,16 +78,16 @@ func (p *PopUp) ShowAtRelativePosition(rel fyne.Position, to fyne.CanvasObject) 
 
 // Tapped is called when the user taps the popUp outside the content area
 // if not modal then dismiss this widget
-func (p *PopUp) Tapped(_ *fyne.PointEvent) {
-	if !p.modal {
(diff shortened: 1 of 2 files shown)


dweymouth:
Hmm, somehow this affected image tests and I'm not sure why...

andydotxyz:
This could have been a simpler change by leaving the object hierarchy as-is and just doing a bounds check on Tapped to see if the tap is outside of the content area...

That might avoid whatever problem the tests have kicked up?

andydotxyz (approved):
Nice, thanks.

PR #5361, opened 2025-01-05
Make data binding thread safe

andydotxyz:
Get/Set can be called from anywhere, Reload too.
Remove all the tedious waiting in data binding tests.
This is not the final public API for goroutine work, but as drivers needed to add this support first I added it to that interface.

Relates to #4654 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/binding_test.go
+++ b/data/binding/binding_test.go
@@ -29,8 +29,6 @@ func TestBase_AddListener(t *testing.T) {
 	})
 	data.AddListener(fn)
 	assert.Equal(t, 1, syncMapLen(&data.listeners))
-
-	waitForItems()
 	assert.True(t, called)
 }
 
@@ -46,7 +44,6 @@ func TestBase_RemoveListener(t *testing.T) {
 	data.RemoveListener(fn)
 	assert.Equal(t, 0, syncMapLen(&data.listeners))
 
-	waitForItems()
 	data.trigger()
 	assert.False(t, called)
 }
@@ -57,7 +54,6 @@ func TestNewDataItemListener(t *testing.T) {
 		called = true
 	})
 
-	waitForItems()
 	fn.DataChanged()
 	assert.True(t, called)
 }
--- a/data/binding/queue.go
+++ b/data/binding/queue.go
@@ -1,30 +1,12 @@
 package binding
 
 import (
-	"runtime"
-
 	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/internal/async"
 )
 
-var mainGoroutineID uint64
-
-func init() {
-	runtime.LockOSThread()
-	mainGoroutineID = goroutineID()
-}
-
-func goroutineID() (id uint64) {
-	var buf [30]byte
-	runtime.Stack(buf[:], false)
-	for i := 10; buf[i] != ' '; i++ {
-		id = id*10 + uint64(buf[i]&15)
-	}
-
-	return id
-}
-
 func queueItem(f func()) {
-	if goroutineID() == mainGoroutineID {
+	if async.IsMainGoroutine() {
 		f()
 		return
 	}
--- a/data/binding/queue.go
+++ b/data/binding/queue.go
@@ -1,30 +1,33 @@
 package binding
 
 import (
-	"sync"
+	"runtime"
 
-	"fyne.io/fyne/v2/internal/async"
+	"fyne.io/fyne/v2"
 )
 
-var (
-	once  sync.Once
-	queue *async.UnboundedFuncChan
-)
+var mainGoroutineID uint64
 
-func queueItem(f func()) {
-	once.Do(func() {
-		queue = async.NewUnboundedFuncChan()
-		go func() {
-			for f := range queue.Out() {
-				f()
-			}
-		}()
-	})
-	queue.In() <- f
+func init() {
+	runtime.LockOSThread()
+	mainGoroutineID = goroutineID()
 }
 
-func waitForItems() {
-	done := make(chan struct{})
-	queue.In() <- func() { close(done) }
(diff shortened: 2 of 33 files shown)


data/binding/queue.go
```diff
@@ -1,30 +1,33 @@
...
-		queue = async.NewUnboundedFuncChan()
-		go func() {
-			for f := range queue.Out() {
-				f()
-			}
-		}()
-	})
-	queue.In() <- f
+func init() {
+	runtime.LockOSThread()
```
dweymouth, line 12: With other runtime.LockOSThreads in the driver, can we be sure that this one will lock the same main thread? (ie is is a guarantee of Go that all init funcs call on the OS main thread, and redundant LockOSThread calls elsewhere will just be no-ops instead of getting a different Goroutine?)

data/binding/queue.go
```diff
@@ -1,30 +1,33 @@
...
 }
 
-func waitForItems() {
-	done := make(chan struct{})
-	queue.In() <- func() { close(done) }
-	<-done
+func goroutineID() (id uint64) {
+	var buf [30]byte
+	runtime.Stack(buf[:], false)
+	for i := 10; buf[i] != ' '; i++ {
```
dweymouth, line 19: Is this basically a hash code? This makes me *slightly* nervous about hash collisions. If the [30]byte we get out of runtime.Stack is already unique maybe we just make `type goroutineID [30]byte` (or [20]byte since it seems we only care about buf[10:]) and use `slices.Equal`?

driver.go
```diff
@@ -43,4 +43,9 @@ type Driver interface {
 	//
 	// Since: 2.5
 	SetDisableScreenBlanking(bool)
+
+	// CallFromGoroutine provides a way to queue a function that is running on a goroutine back to the main thread.
+	// This is required when background tasks want to execute code safely in the graphical context.
+	// Since: 2.6
+	CallFromGoroutine(func())
```
dweymouth, line 50: Slightly nitpicky, but I wonder if the name `RunFromGoroutine` that was discussed is better. To me, `CallFromGoroutine` could be interpreted as "it will call the supplied func from a new goroutine" ie a synonym for `go f()`. I'm sure it's fine either way though, just something to think about.

data/binding/binding_test.go
```diff
@@ -29,8 +29,6 @@ func TestBase_AddListener(t *testing.T) {
 	})
 	data.AddListener(fn)
 	assert.Equal(t, 1, syncMapLen(&data.listeners))
-
```
dweymouth, line 32: I think a new unit test where binding APIs are invoked in a goroutine could be worthwhile. The test when run normally just confirms the expected behavior, but when run with `go test -race` we can also prove that those APIs don't have any races.

dweymouth (commented):
Really nice to see the threading work nearing the finish line! Just a few comments so far

driver.go
dweymouth: Hmm, now that I think about it, the `RunFromGoroutine` name could have the same misinterpretation perhaps. Not sure it matters

data/binding/binding_test.go
Jacalz: Good point. This is likely true for more parts of the API. We should run more tests with race instrumentation enabled (once we get there).

data/binding/queue.go
Jacalz: All init functions always run before the main function so there really are no other threads to run on. However, I have no idea how multiple calls to runtime.LockOSThread() behave.

data/binding/queue.go
Jacalz: It would be good to reuse the other function instead of copying it here.

driver.go
Jacalz: Yeah, I think both have that misinterpretation. I'm not a huge fan of it for that exact reason. I think we need something different.

Jacalz: I think RunOnMain is a better name.

andydotxyz: We cannot use the `Run` prefix, particularly not in a situation that implies an app can be run from a goroutine.

I could use `CallOnMain` - but then there is the slight complication that "main thread" is not part of our lexicon in the public API nor the Go familiar terms either.

data/binding/binding_test.go
andydotxyz: I agree. I would prefer to do this as a follow-up PR. The intent of this work was to show that things do work exactly as before

andydotxyz: I have added an item to the TODO list on the parent ticket to come back to this.

andydotxyz:
Thanks for the comments, I think that's all good now - with an ongoing discussion for public API for enqueuing functions

driver.go
dweymouth: This is a new API on the public driver interface though, even if we introduce another one later, likely at the root package. I suppose we can amend the name of this one at that point, since we don't promise no breaking changes on develop

dweymouth (approved):
Approved, so long as we don't care about potential breaking changes on `develop` in case we rename or move the new public "CallFromGoroutine" function introduced here

Jacalz:
> Approved, so long as we don't care about potential breaking changes on `develop` in case we rename or move the new public "CallFromGoroutine" function introduced here

We never care about that. It is one of the reasons to have such a branch. Anyone pulling latest master or release/v2.x.y will never get breaking changes :)

But yeah, I agree we need a better name in the future

Jacalz (approved):
Nice step in the right direction :)

driver.go
andydotxyz: Agreed, that is what I meant by "with a change to driver if required."

PR #5364, opened 2025-01-05
Test suite: code linting

deining:
This PR brings improvements to several tests of the test suite. The changes are based on the results of some code linting done by [testifylint](https://github.com/Antonboom/testifylint).

- [x] Tests all pass.


Jacalz:
The changes seem entirely sensible but I kind of also think that the assertion methods used already are used because of their consistency. It is easier to stick to only using Equal(), True() etc and use them consistently. I'm not saying that this is a bad change but I'm worried that we're just going to continue with new code just like the code looks now (not that there necessarily is anything wrong with that).

deining:
> The changes seem entirely sensible but I kind of also think that the assertion methods used already are used because of their consistency. It is easier to stick to only using Equal(), True() etc and use them consistently. I'm not saying that this is a bad change but I'm worried that we're just going to continue with new code just like the code looks now (not that there necessarily is anything wrong with that).

I see.

I'm citing the [opinion](https://github.com/Antonboom/testifylint?tab=readme-ov-file#problem-statement) of the author of testifylint checker:

> Most checkers are stylistic, but checkers like [error-is-as](https://github.com/Antonboom/testifylint?tab=readme-ov-file#error-is-as), [require-error](https://github.com/Antonboom/testifylint?tab=readme-ov-file#require-error), [expected-actual](https://github.com/Antonboom/testifylint?tab=readme-ov-file#expected-actual), [float-compare](https://github.com/Antonboom/testifylint?tab=readme-ov-file#float-compare) are really helpful.

I did a quick check for the code base:

```
$ testifylint --disable-all --enable=expected-actual ./... 2>&1 | wc -l
32
```

Presumably there are 32 cases where actual and expected value are in wrong order. These assertions should be fixed IMHO.

Concerning the more stylistic changes: For me, this PR results in a easier readable code which is a good thing IMHO. But I understand your arguments, too.

dweymouth:
I am inclined to agree with @deining here. I'll approve this PR but leave it unmerged for further discussion. I don't think absolute consistency is important either if future tests still end up using assert.True unnecessarily, but I do think this is a (small, but every bit counts) readability improvement.

dweymouth (approved):

deining:
@dweymouth: I'll appreciate your quick feedback, thanks a lot!

> I do think this is a (small, but every bit counts) readability improvement.

Great to here that, that's exactly my opinion, too.

> I'll approve this PR

Thanks for your approval, I see this as a signal that we can go further down the road here 😄.

> but leave it unmerged for further discussion.

So let's continue discussion here: This PR is a mix of everyting that doesn't cover all test cases. It's certainly a valid PR, but was also meant to ignite discussion on this topic (which obviously worked out 😄).

What I would propose now:

- We close this PR unmerged and take a more stuctured approach by splitting up the whole linting task into several smaller chunks/PRs that are easier to consume/digest.
- Each of these PRs is focused on **one single linting rule only** and addresses violations of this rule in **all test cases**.
- I propose to start with linting rules that are more relevant, like [error-is-as](https://github.com/Antonboom/testifylint?tab=readme-ov-file#error-is-as), [require-error](https://github.com/Antonboom/testifylint?tab=readme-ov-file#require-error) and [expected-actual](https://github.com/Antonboom/testifylint?tab=readme-ov-file#expected-actual), eventually followed by more stylistic changes.

Is this something we all can agree upon?

Jacalz (approved):
I agree that it is a good step. The wrong actual and expected are the most important in my opinion 👍

Jacalz:
I took a look at that linter and it seems to be very useful. I've sometimes toyed with the idea of switching to a different testing framework (mainly https://github.com/alecthomas/assert) as testify is huge and has its share of problems but that would be a big change. As long as we are udon testify, then that linter seems like a great addition. I'd be in favour of adding it to our already existing workflow infrastructure 👍

PR #5365, opened 2025-01-06
Minor dependency upgrades for v2.6.0

Jacalz:
### Description:

Some minor fixes and improvements in fsnotify plus updated golang.org/x/ packages. The latter updates so we don't use a version which had security issues, in turn stopping GitHub's security scanning from complaining. The security issues did not affect us but were reported nonetheless as they don't check if the code is used at all but merely if the package is imported.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (dismissed):
👍

Jacalz:
Windows CI did not like the fsnotify upgrade. Reverted that.

Jacalz:
There seems to be a lot of strange errors with badvalue when trying to set up some display parameter.

andydotxyz:
> There seems to be a lot of strange errors with badvalue when trying to set up some display parameter.

An intermittent error on the Ubuntu tests seems to always hit on the first run which is weird indeed

dweymouth (approved):

PR #5367, opened 2025-01-06
Test suite. code linting

deining:
This PR brings the work that started with #5364 to an end.

### Checklist:
- [x] Tests all pass.


andydotxyz:
Sorry there has been a lot of change to the testing infrastructure with the thread changes landing this week.
Could you bring in latest develop and make sure you're happy with the results? :) Thanks

deining:
> Sorry there has been a lot of change to the testing infrastructure with the thread changes landing this week. Could you bring in latest develop and make sure you're happy with the results? :) Thanks

I rebased my PR and made the necessary adjustments. All checkers are green now, except for `static-analysis`, which is addressed in PR #5379.

andydotxyz:
Hi @deining the develop branch is getting more stable now and we are tagging an alpha build so we are aiming for no more API changes (only additions and fixes). This would be a good time to rebase and finalise this PR.

internal/repository/memory_test.go
```diff
@@ -353,12 +354,12 @@ func TestInMemoryRepositoryListing(t *testing.T) {
 	assert.True(t, canList)
 
 	foo, _ := storage.ParseURI("mem:///foo")
-	canList, err = storage.CanList(foo)
```
andydotxyz, line 356: foo -> empty seems wrong

andydotxyz (changes requested):
Thanks for all this, looks like a possible typo though?

deining:
> Hi @deining the develop branch is getting more stable now and we are tagging an alpha build so we are aiming for no more API changes (only additions and fixes). This would be a good time to rebase and finalise this PR.

I rebased my PR as requested, all checkers are green now! :smile:

internal/repository/memory_test.go
deining: Can you please take that over and change if needed? Thanks.

deining:
> Thanks for all this

You are welcome.

Now it's time for sleep. :smile:

andydotxyz:
Please don't force-push after a review. I cannot tell which lines changed and now I have to go through every line again!

deining:
> Please don't force-push after a review. I cannot tell which lines changed and now I have to go through every line again!

I will follow your advice in the future, sorry for the inconvenince caused!

internal/repository/memory_test.go
andydotxyz: I don't understand this question - do you mean you do not want to make the required change?

andydotxyz:
After the force push I don't think the output changed - am I missing something?

On another note @Jacalz and @toaster are more familiar with unit testing tools, are you happy with the changes? (Other than the mistake I noted).

Jacalz:
I like the changes to correct usages of the assert functions but I'm not sure about the switch to use the `require` package. We are used to using the former package and it seems to me like we will continue writing code that way.

internal/repository/memory_test.go
deining: > I don't understand this question

And I don't really understand what you would consider a typo and would you would like to see changed.

My impression is that you would like to change `foo` to `empty` in line 356. Is that correct? 

> - do you mean you do not want to make the required change?

No not at all. This was meant as a friendly offer to you, kindly asking you for your help here. I'm not sure what's correct here, so can you please take over (or alternatively, tell me what to do here).

Hope this helps for clarify the issue.

andydotxyz: OK, in more detail:

You have taken the line:
`canList, err = storage.CanList(foo)`
and replaced it with:
`canList, err = storage.CanList(empty)`.

So now the test is not actually validating that `foo` can be listed.

andydotxyz:
> Looks like it's more a convention you are used to and that you would like to keep!?

This is probably more than just style as require vs assert actually changes the flow of the error.
If a test has 20 assertions then changing to requires could mean a longer process of fixing because it would show 1 error at a time... (I realise, having read up on the subject since yesterday).

Jacalz:
> Looks like it's more a convention you are used to and that you would like to keep!?

Indeed. I would prefer to stick to the convention that we are used to and we know works well.

> My proposal is to add a new GitHub workflow action that runs testifylint after each pull request/commit.

I'm not entirely against that but I am a bit sceptical about the need for a linter just to make sure that the assert functions are used correctly. It is the assert package there that is the problem and we've not had all too many problems in the past. I'm leaning a bit against https://github.com/fyne-io/fyne/issues/5447 instead honestly.

internal/repository/memory_test.go
deining: > OK, in more detail:
> 
> You have taken the line: `canList, err = storage.CanList(foo)` and replaced it with: `canList, err = storage.CanList(empty)`.
> 
> So now the test is not actually validating that `foo` can be listed.

Mea culpa, I was on the wrong branch! After switching to the correct branch, things are evident to me now 😄.
I fixed the typo by adding an additional commit. This should resolve the issue.

PR #5372, opened 2025-01-07
Improvements in test runner

andydotxyz:
Relates to #4654 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window_test.go
+++ b/internal/driver/glfw/window_test.go
@@ -40,7 +40,7 @@ func TestMain(m *testing.M) {
 
 	waitForStart := make(chan struct{})
 	go func() {
-		// Wait for GLFW loop to be running (plus a moment in case od scheduling switches).
+		// Wait for GLFW loop to be running (plus a moment in case of scheduling switches).
 		// If we try to create windows before the context is created, this will fail with an exception.
 		<-waitForStart
 		time.Sleep(time.Millisecond * 100)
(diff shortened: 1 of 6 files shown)


andydotxyz:
This is 1 of 2 PRs for turning on race. This is just CI fixes that benefit everyone :)

The race detection has highlighted an issue with a race in the test runner / driver itself which will take some more time to resolve.

internal/driver/glfw/window_test.go
```diff
@@ -1942,3 +1949,27 @@ func (t *tabbable) AcceptsTab() bool {
 	t.acceptTabCallCount++
 	return true
 }
+
+func safeResize(w fyne.Window, size fyne.Size) {
+	runOnMain(func() {
+		w.Resize(size)
+	})
+}
```
Jacalz, line 1957: I wonder if it perhaps would be easier, and less chance of new tests here not using these, if we had a wrapper around the window? We could extend it like we usually do with widgets basically and then have all of the glfw tests use that specific glfwTestWindow wrapper? Just an idea :)

Jacalz (commented):
Looks great. Just one possible improvement idea inline :)

internal/driver/glfw/window_test.go
Jacalz: Absolutely. I'll open an issue for it

Jacalz (approved):
Thanks. Nice fixes :)

internal/driver/glfw/window_test.go
Jacalz: https://github.com/fyne-io/fyne/issues/5375

PR #5374, opened 2025-01-07
Remove rendererCacheNode painterData field

### LINKED ISSUE #4523: painterData in RendererCacheNode is only used in tests
### Checklist

### Describe the bug

I noticed that the field `painterData` is used extensively for the tests related to the renderer cache. We should try to remove the field while keeping the tests useful, i.e. find a way to test without the painter data. Removing this field will have a positive effect on memory usage as each cached object will be smaller. My testing shows that the struct becomes 25% smaller (64 bytes to 48 bytes) when removing the field. We are basically wasting 16 bytes for each cached renderer.

https://github.com/fyne-io/fyne/blob/a09bbd790b98120bc406a064f80aeafb4ca97656/internal/driver/common/canvas.go#L536-L540

### How to reproduce

Look at struct size for `RenderCacheNode` and references of the `painterData` field.

### Screenshots

_No response_

### Example code

Not a regular bug so to say:
https://github.com/fyne-io/fyne/blob/a09bbd790b98120bc406a064f80aeafb4ca97656/internal/driver/common/canvas.go#L536-L540

### Fyne version

v2.4.3

### Go compiler version

1.21.5

Fedora Linux 39 

### Additional Information

_No response_

Jacalz:
### Description:

This was only used in tests to verify that nodes were not reused. Update tests to use a map for that; this makes the size of each renderer cache 25% smaller with now using 48 bytes instead of 64 bytes.

Fixes #4523

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/common/canvas.go
+++ b/internal/driver/common/canvas.go
@@ -494,11 +494,6 @@ type RenderCacheNode struct {
 	parent      *RenderCacheNode
 	// cache data
 	minSize fyne.Size
-	// painterData is some data from the painter associated with the drawn node
-	// it may for instance point to a GL texture
-	// it should free all associated resources when released
-	// i.e. it should not simply be a texture reference integer
-	painterData any
 }
 
 // Obj returns the node object.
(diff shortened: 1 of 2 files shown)


andydotxyz (approved):
Nice catch thanks

PR #5376, opened 2025-01-08
Add a generic sync.Map wrapper/helper

Jacalz:
### Description:

Adding useful methods (like `Clear()` and `Len()`) plus adding some type safety on top of `sync.Map`. The added type safety will be useful to reduce chances of crashes and other incorrect usages. I also made sure to include code so that compiling with Go 1.23 or newer uses the faster builtin `Clear()` instead of O(n) range and delete.

I also updated the texture and renderer caches to use this map. It should decrease lock contention when deleting objects. Fyne Demo still runs without races.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Great cleanup thanks!

PR #5383, opened 2025-01-09
Minor refactor of refresh code in base widgets

Jacalz:
### Description:

Removing some code duplication for how methods were refreshing the widget. I also reordered unexported methods to be in the correct order. This does move `.Hide()` and `.Show()` away from doing a `canvas.Refresh()` which I think is both correct and a performance win, right?

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/widget.go
+++ b/widget/widget.go
@@ -5,7 +5,6 @@ import (
 	"sync/atomic"
 
 	"fyne.io/fyne/v2"
-	"fyne.io/fyne/v2/canvas"
 	"fyne.io/fyne/v2/internal/async"
 	"fyne.io/fyne/v2/internal/cache"
 	internalWidget "fyne.io/fyne/v2/internal/widget"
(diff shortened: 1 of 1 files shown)


andydotxyz (changes requested):
This confuses `Widget.Refresh()` and `canvas.Refresh(Widget)`.
This means that when hiding a `BaseWidget` it will now refresh the tree of widgets which it did not do before, as it just told the driver that those items should be removed from cache.

Jacalz:
So it should only refresh on show if I understand you correctly?

Jacalz:
I'm really confused about the difference between canvas.Refresh() for a widget and refresh on that widget directly.

andydotxyz:
Calling `Widget.Refresh` cascades into the widget renderer and probably all of its children to redraw each canvas object.
`canvas.Refresh(obj)` just tells the driver that the item should be purged from the cache - so if it is then hidden no further code will be executed until it is shown again.

Jacalz:
Ah, alright. I'll close this and do the refactoring in some future work after https://github.com/fyne-io/fyne/pull/5388 has been merged

PR #5385, opened 2025-01-09
Proposal: CustomMessage dialogs

lostdusty:
### Description:
I've wanted to include for some time a new API that allows the developer to set whatever resource for the text-based dialog(s) icon. This allows for an example, a success dialog, confirming the user changes.

### Checklist:

- [ ] Tests included. (didn't make tests, but it uses the same base as NewInformation)
- [X] Tests all pass.

#### Where applicable:

- [X] Public APIs match existing style and have Since: line. (will be added)


lostdusty:
Moved to #5386

PR #5386, opened 2025-01-09
Add .SetIcon() to custom dialogs

lostdusty:
### Description:
I've wanted to include for some time a new API that allows the developer to set whatever resource for the text-based dialog(s) icon. This allows for an example, a success dialog, confirming the user changes.

### Checklist:

- [x] Tests included. 
- [X] Tests all pass.

#### Where applicable:


Jacalz (commented):
We discussed new dialogs a long time ago with some v2 dialog stuff that I intended to work on but never got around to. The problem now is that a lot of details are internal (like your new dialogs just being exported functions that call a single internal one) and the dialog API is very old and fragile without much happening. It just doesn't scale well and feels unintuitive compared to more modern and developer friendly APIs that we have in other places.

I think we really should be very careful about adding more dialogs until we have a better API and can expose settings in a tidier way. Adding a new exported constructor API just to change a single internal parameter does not scale well.

Jacalz:
A better API in the meantime would perhaps be to just allow setting the icon on some of the other dialogs? You can look at the button row changes that I did a while ago.

PS: It should be `Since: 2.6.0` for the new APIs.

andydotxyz (commented):
I don't think this belongs in the info files - we have other source for custom dialogs don't we?

Jacalz (changes requested):
Yeah, I would personally just add a method similar to https://github.com/fyne-io/fyne/blob/4a875d9d5677ce8a98bd15787d6f74d4263cac16/dialog/custom.go#L53 on the custom dialog struct.

lostdusty:
I have tried adding a `SetIcon` on custom but it did not work, at least not with `NewCustomDialog`

Jacalz:
It definitely be possible to do, or might require tweaking some minor stuff, but you might have been bitten by the strange semantics of how dialogs work. You basically have to recreate the dialog with a call to `.create()` to populate the internal layout and objects. I worked around this, or optimized it, with setButtons to avoid recreating the whole dialog.

lostdusty:
I think I managed to do it
image

dialog/base.go
```diff
@@ -140,6 +140,11 @@ func (d *dialog) setButtons(buttons fyne.CanvasObject) {
 	d.win.Refresh()
 }
 
+func (d *dialog) setIcon(icon fyne.Resource) {
+	d.win.Content.(*fyne.Container).Objects[0] = &canvas.Image{Resource: icon}
```
Jacalz, line 144: Please see the discussion in https://github.com/fyne-io/fyne/pull/4615. Would you mind making sure that the object is set to a spacer in case of no icon (icon == nil). This reduces CPU usage of trying to draw an empty image.

dialog/custom.go
```diff
@@ -55,6 +55,14 @@ func (d *CustomDialog) SetButtons(buttons []fyne.CanvasObject) {
 	d.setButtons(container.NewGridWithRows(1, buttons...))
 }
 
+// SetIcon sets an icon to be shown on the right of the dialog, like the NewInformation dialog.
```
Jacalz, line 58: I think we should mention top right and ignore the example.
```suggestion
// SetIcon sets an icon to be shown in the top right of the dialog.
```

Jacalz (changes requested):
Thanks. Looking good. Just a few suggestions inline

dialog/base.go
```diff
@@ -140,6 +140,15 @@ func (d *dialog) setButtons(buttons fyne.CanvasObject) {
 	d.win.Refresh()
 }
 
+func (d *dialog) setIcon(icon fyne.Resource) {
+	if icon == nil {
+		d.win.Content.(*fyne.Container).Objects[0] = &layout.Spacer{}
+		return
```
Jacalz, line 146: You are not refreshing in this case.

Jacalz (changes requested):
Thanks. I noticed a small bug when setting the icon to nil, left a note inline.

I forgot to write this earlier but would you please add some tests? You can look at how the tests around the setButton functionality was done.

FYI: We usually work on the principle that the one adding the review comment is the one that decide when it is resolved (and marks it as such).

lostdusty:
How do I make the test image? Should I use the window.Canvas().Capture() function?

lostdusty:
Used it, all tests passed for me :)
image

dialog/base.go
lostdusty: Added a refresh

Jacalz: Thanks :)

dialog/custom_test.go
```diff
@@ -142,3 +142,33 @@ func TestCustomConfirm_Importance(t *testing.T) {
...
+	size := fyne.NewSize(200, 300)
+	w.Resize(size)
+
+	test.ApplyTheme(t, test.Theme())
+	label := widget.NewLabel("Test was successful.")
+	d := NewCustom("Fyne test", "Dimiss", label, w)
+	d.SetIcon(nil)
+	d.Show()
+
+	test.AssertRendersToImage(t, "dialog-custom-seticon-nil.png", w.Canvas())
```
Jacalz, line 173: You don't need two different test functions just for setting different values. There is just unnecessary code duplication between them. I'd rather call the test `TestCustom_SetIcon` to be consistent with the other tests (that is the standard for naming) and then just at the end of the first one you have, set the icon to nil and do a new assert to image.

Jacalz (changes requested):
Thanks. Well done with the tests. Really close now. Let's just combine the tests into one and use correct naming :)

andydotxyz:
I am pretty sure the failures were resolved in more recent commits on develop branch, if you rebase or merge it in things should be looking pretty good.

lostdusty:
Does it looks better now? I'll try rebasing now...

lostdusty:
Hope its good now :)

Jacalz:
Oh no. You are running into the horrible test failure that I encountered

Jacalz:
It is a test that hard-codes the lookup of the amount of files in the folder where you put your images. Now there are 11 instead of 9 so it is failing everywhere

Jacalz:
See https://github.com/fyne-io/fyne/blob/4a875d9d5677ce8a98bd15787d6f74d4263cac16/dialog/file_test.go#L434-L435 and https://github.com/fyne-io/fyne/blob/4a875d9d5677ce8a98bd15787d6f74d4263cac16/dialog/file_test.go#L449-L450. That test is really smelly. We need to do something about that

lostdusty:
Mind if I ask why its hard-coded?

Jacalz:
I don't know. I suppose the one who wrote the test didn't think of the consequences of having it fail any time someone adds a new file to the folder but it doesn't matter much. We all make mistakes. I added those warning comments last time I ran into it. The fact that it is problematic remains :)

andydotxyz:
At the same time it's a pretty easy fix. Files added, test fails, update number, tests pass.

lostdusty:
I mean I thought it had a particular reason, I'll update the file :)

lostdusty:
Just to make sure, I just need to change from 9 to 11, other numbers should be fine, right?

Jacalz:
You can run the test locally to see what it says. It should print an expected and an actual number. Just edit that and you should see when it passes

lostdusty:
Should be working now, had to update two numbers.

andydotxyz (approved):
Clean addition thanks.

Jacalz (approved):
Great addition indeed. Well done!

PR #5387, opened 2025-01-09
document the way to use gendex and dex.go for new GoNativeActivity.java behaviour

dolanor:


Jacalz (commented):
FYI: We are planning on deprecating this tool here in favour of moving it into https://github.com/fyne-io/tools. I think I changed how the generation is done in that project.

internal/driver/mobile/app/GoNativeActivity.java
```diff
@@ -27,6 +27,11 @@
 import android.widget.TextView;
 import android.widget.TextView.OnEditorActionListener;
 
+// GoNativeActivity is the java implementation that helps Go map to android via the NDK.
```
andydotxyz, line 30: I'm not sure this first sentence is right - the activity is the entry point for our Android boot loader.
The NDK connects to this via C - so it's not really providing any connections itself I don't think?...

andydotxyz (changes requested):
Great additions thanks, but I think one of the docs is a bit misleading.

dolanor:
Ok.

What's the new procedure to generate a new dex, then?
I kind of need it

internal/driver/mobile/app/GoNativeActivity.java
dolanor: What I meant (from my VERY limited understanding) is that it was the java SDK object that was a utility for Go to access via the C and the NDK.
I guess I don't know how to express it correctly :)

andydotxyz:
> Ok.
> 
> What's the new procedure to generate a new dex, then? I kind of need it

We are in the process of moving `cmd/fyne` into the `tools` repo, but nothing else changes. So this could/should land as I believe the tools repo is pending a merge as things have got behind again. Is that about right @sdassow ?

internal/driver/mobile/app/GoNativeActivity.java
andydotxyz: Go accesses the Java object via NDK. The Java provides us certain functionality which is accessed through JNI and C functionality.
Is that clearer?

sdassow:
Yes, please. I've just done a merge, so we should be able to move ahead without too much delay: https://github.com/fyne-io/tools/pull/24

sdassow:
Can you please file PR against the Fyne tools repo at https://github.com/fyne-io/tools? That way you can also just file against the `main` branch there.

Jacalz:
That reminds me, please note that PRs should go to the develop branch in this repository and that the PR template should be filled out accordingly, not removed.

andydotxyz:
Thanks @Jacalz I missed that!

dolanor:
So the `fyne` CLI will be build from [fyne-io/tools](/fyne-io/tools).

So it's gonna base on the GoNativeActivity that would be referenced from the `fyne.io/fyne/v2` dependency in the go.mod, right?
So if we want to integrate updates from a modified `GoNativeActivity`, we would need to add some `replace` in the `tools` repo, right?

Jacalz:
I already implementing the support for updating the Dex stuff in that repository. It relies on the cloned `fyne` and `tools` repositories being in the same parent folder as it just does a relative lookup. No need to replace anything in the module.

andydotxyz:
Yes it should work the same way, though perhaps we need to document the additional assumption that "tools" and "fyne" repos are checked out next to each other.

andydotxyz:
Just a prod that it would be good to land an updated version of this that references the tools API now that it is fully up to speed so we can get this in the release candidate?

PR #5390, opened 2025-01-10
Feature/selective lang loading

aksdb:
### Description:
At the moment, fyne loads all base translations, sets the default language to English and merges in translations provided by the user.

This approach has a few downsides, though:

1. If the user registers only a subset (or a different set) of languages, than the base translations provided, the `closestSupportedLocale` function is not (and cannot be) aware of this. It might pick a language that has only partial (our base) translations and misses user translations. Example: the user provides custom translations for English but the app runs on a German system. Some buttons and menu items are now translated into German, but all others fall back to the provided English translation.

2. The user may not develop their app with English in mind, so falling back to English might not be the best choice for them; especially when paired with the issue above.

This PR proposes a solution to both issues that is (mostly) backwards compatible:

- We still load all base translations and consider them "available" for matching in `closestSupportedLocale`. So for apps without any custom translations, everything behaves as before.
- Once the user adds their first own translation, we reset the known languages and re-initialize the bundle.
  - The first language becomes the new fallback language. This gives the user (implicit) control, by allowing them to register the languages in their preferred order.
  - The list of known languages is reduced to the ones registered _explicitly_. So if, as in the example above, the user registers a custom translation for English, only English will be considered until they also register further translations for other languages.

The implementation also has two downsides, but IMO smaller ones:
- That the first added translation becomes the default language is more or less implicit. When using `AddTranslationsFS` as the only (or at least first) call, one would have to make sure, that the order makes sense. It can easily be circumvented by either using file name prefixes in that case or by calling one of the other `AddTranslations*` functions first (with an empty JSON).

  > **Note**: Potential improvement: offer an API to explicitly set the order of languages; overwriting the `translations` slice.

- If someone relied on base translations being in place, while they register their own, different, translations, this PR would change behavior; I think it changes for the better, though and IMO it prevents (likely) unexpected side-effects.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/lang/lang.go
+++ b/lang/lang.go
@@ -7,6 +7,7 @@ import (
 	"embed"
 	"encoding/json"
 	"log"
+	"path/filepath"
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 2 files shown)


Jacalz:
I'm sorry that this got stuck without a review for so long. Let's make sure that someone with more knowledge of the translations backend gets to it so we hopefully can land it sometime soon :)

PR #5392, opened 2025-01-11
Add a fyne.Do method for developers

andydotxyz:
and update our naming/docs

Testing the threading model is a separate PR as it refers to the driver work not this top level package change.

Nearing the end of #4654 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -5,7 +5,8 @@ import "time"
 // Driver defines an abstract concept of a Fyne render driver.
 // Any implementation must provide at least these methods.
 type Driver interface {
-	// CreateWindow creates a new UI Window.
+	// CreateWindow creates a new UI Window for a certain implementation.
+	// Developers should use [App.NewWindow].
 	CreateWindow(string) Window
 	// AllWindows returns a slice containing all app windows.
 	AllWindows() []Window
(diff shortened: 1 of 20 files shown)


Jacalz (approved):
I'm happy with the new API but let's hear what the other think first :)

driver.go
```diff
@@ -44,8 +44,10 @@ type Driver interface {
...
 	SetDisableScreenBlanking(bool)
 
-	// CallFromGoroutine provides a way to queue a function that is running on a goroutine back to the main thread.
+	// DoFromGoroutine provides a way to queue a function that is running on a goroutine back to
+	// the central thread for Fyne updates.
 	// This is required when background tasks want to execute code safely in the graphical context.
+	//
 	// Since: 2.6
-	CallFromGoroutine(func())
+	DoFromGoroutine(func())
```
dweymouth, line 52: Do we want to un-export this one for public use to avoid confusion, so we just have `fyne.Do`? Or at least have the documentation here make clear that it is entirely equivalent to calling `fyne.Do` and `fyne.Do` should generally be preferred (for brevity and for standardization)?

dweymouth (approved):

Jacalz (approved):

PR #5394, opened 2025-01-11
Fix races within flatpak portal dialogs

Jacalz:
### Description:

This cleans up and refactors the dialog code for desktop portals on flatpak and it does so while fixing some races in it. I also fixed RunNative not behaving correctly (or as expected) with the new threading model.

For #4654 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Thanks for that, neater indeed :)

PR #5403, opened 2025-01-12
add StartSelect Collumn and Row in entry

mykytaserdiuk:
### Description:

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -398,6 +398,14 @@ func (e *Entry) KeyUp(key *fyne.KeyEvent) {
 	}
 }
 
+func (e *Entry) StartSelectRow() int {
+	return e.selectRow
+}
+
+func (e *Entry) StartSelectColumn() int {
+	return e.selectColumn
+}
+
 // MinSize returns the size that this widget should not shrink below.
 //
 // Implements: fyne.Widget


mykytaserdiuk:
@andydotxyz People who want to get the starting point of the text selection will not be able to get it. I think it would be better if they had a method that would give them this opportunity.
Maybe "Start" is really not a good naming. I make a new name

andydotxyz:
Is it possible to expand on the use-case? "Needing to access the variable" doesn't tell us much about why - and because Fyne naming tries to be focused on outcomes then it may be important.

Also there may even be an easier way to do the thing you're trying to achieve!

mykytaserdiuk:
@andydotxyz okey. in one of the project I added and refactored the ability in the "text editor" to change the "parameters" of the text by pressing a button. And I found the fact that I can't "out of the box" take the "beginning of the selection", and I had to implement it through available paths. But it was very difficult and unpleasant, after which I decided to make the selectRow and selectColumn fields private

Do I understand correctly that the problem is not only in the implementation, but also in the lack of understanding of the need?

mykytaserdiuk:
“start of selection” is what I need to insert in front of the text a text tag specific for display (local text engine for the project, like Markdown)

andydotxyz:
Yea trying to understand the need is key ("use-case").
If your need was to get selected text then Entry.SelectedText() would have helped no? Or is there a display element that I'm not understanding as well?

mykytaserdiuk:
image

here is a regular entry window. And in it I selected text. There is a point where I started to select it and finished. The position where I finished selecting can be found out using

"entry.SelectRow" "entry.SelectColumn"
But where I started to see - no

In my case, it will be "row:0 column:1", "row:3 column:3" respectively
But we can’t find out the second one’s data without machinations, mathematics or getting inside pkg.

andydotxyz:
Yes thanks I understand exactly what you are describing as start of selection - but I'm still it grasping what you're looking to do - text manipulations and selection handling are complex for sure.

In 2.6 we have "InsertAtCursor" and if you send it "textToInsert" + Entry.SelectedText() then it will inset the string at the beginning of the selection.

Is that what you're trying to do?

mykytaserdiuk:
@andydotxyz no, using the method you indicated, we can't insert text BEFORE starting to select text

Or is there something I don't understand anymore?

mykytaserdiuk:
Maybe have sense create just method for it (insert text before "start pos")

mykytaserdiuk:
@andydotxyz yes, the logic is very close, but in this case, we inserted the text at the cursor location, although we started selecting in a different place (we started selecting from the end, but inserted it almost at the beginning of the text)

andydotxyz:
> @andydotxyz yes, the logic is very close, but in this case, we inserted the text at the cursor location, although we started selecting in a different place (we started selecting from the end, but inserted it almost at the beginning of the text)
> 
> +it's a replace of the text, if I'm right understand

Sorry I don't understand this message. The unit test I pasted replaces the selected text with alternative text which is the selection plus some other text - so it could be added at the beginning or the end.

Does that help?

mykytaserdiuk:
There is no way, so to speak, to insert text at the same time and the end, and in the beginning of the text (without the help of external means). When we have seen the text, we can find the point where we stopped selecting, but not where we started

CursorRow - the place where we stopped highlighting, but where we started selected, we have no way of knowing this place
Sorry if i write something strange

andydotxyz:
> There is no way, so to speak, to insert text at the same time and the end, and in the beginning of the text (without the help of external means). When we have seen the text, we can find the point where we stopped selecting, but not where we started
> 
> CursorRow - the place where we stopped highlighting, but where we started selected, we have no way of knowing this place Sorry if i write something strange

Forgive me, but doesn't this code very that scenario?

```go
func Test_InsertBefore(t *testing.T) {
	test.NewTempApp(t)

	initial := widget.NewEntry()
	initial.SetText("123456")

	right := &fyne.KeyEvent{
		Name: fyne.KeyRight,
	}
	initial.TypedKey(right)
	initial.TypedKey(right)

	initial.KeyDown(&fyne.KeyEvent{
		Name: desktop.KeyShiftLeft,
	})
	initial.TypedKey(right)
	initial.TypedKey(right)
	initial.KeyUp(&fyne.KeyEvent{
		Name: desktop.KeyShiftLeft,
	})

	if initial.SelectedText() != "34" {
		t.Fail()
	}

	initial.InsertAtCursor("start34stop")
	if initial.Text != "12start34stop56" {
		t.Fail()
	}
}
```

PR #5407, opened 2025-01-14
List and GridWrap threading simplifications

dweymouth:
### Description:
Previously, complicated logic had been added in List and GridWrap to ensure that the renderer update function was concurrency-safe if the contents of the collection widget were asynchronously updated during a render. Now that is no longer a concern; the only concern would be if the collection widget contents were updated from within an UpdateItem callback, which isn't good practice, and this PR adds documentation to these callbacks that they should only update the item's rendering, not modify properties of the collection as a whole.

Some variable names were updated in GridWrap -- the logic changes are the same between List and GridWrap and easier to see in List since no other changes were made there.

### Checklist:

- [ ] Tests included. // Not adding new behavior, current tests sufficient to catch a regression
- [x] Tests all pass.

### DIFF
--- a/widget/gridwrap.go
+++ b/widget/gridwrap.go
@@ -31,11 +31,26 @@ type GridWrapItemID = int
 type GridWrap struct {
 	BaseWidget
 
-	Length       func() int                                      `json:"-"`
-	CreateItem   func() fyne.CanvasObject                        `json:"-"`
-	UpdateItem   func(id GridWrapItemID, item fyne.CanvasObject) `json:"-"`
-	OnSelected   func(id GridWrapItemID)                         `json:"-"`
-	OnUnselected func(id GridWrapItemID)                         `json:"-"`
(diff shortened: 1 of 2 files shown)


Jacalz (approved):
Nice one. Thanks. Is this the case with the other collection widgets as well? I think those need to fixed then also (especially the doc comments on the fields would be great as well).

andydotxyz (approved):
Thanks, good to simplify as a result of these changes. Speedups everywhere :)

dweymouth:
> Is this the case with the other collection widgets as well? I think those need to fixed then also (especially the doc comments on the fields would be great as well).

The doc comments would be good, but can be done in a follow-up. The logic that this PR is undoing was added by me just to List and GridWrap. (I was motivated to optimize List and GridWrap because they're the core of my app :) ) The other collections widgets don't have the zero-alloc optimizations.

Issue #5378, opened 2025-01-08
data/binding: Use generics instead of generator script


Jacalz:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [X] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

With https://github.com/fyne-io/fyne/pull/5377 landed, there is only the code generation for data binding left. As there is quite a lot of code duplication between the binding types, this would be a wise place to apply generics and decrease our reliance on code generation. This should improve developer ergonomics and make the code easier to understand.

### Is it possible to construct a solution with the existing API?

I think so. It might be possible to rewrite some internal types to use generics without impacting the public API but I think a nice change for v3.0.0 further down the line would be to have a public generic API for bindings (far into the future but anyway). 

### Describe the solution you'd like to see.

Clean up data binding using  generics to try and remove code generation as much as possible.

andydotxyz:
We need to pay attention to the possible need for comparators as well - the custom types in tree/list need to be able to do change detection which we are falling short on at the moment.

Jacalz:
Indeed. The `cmp` package may be useful for that: https://pkg.go.dev/cmp

Jacalz:
We might wish to add a generic alternative to `Untyped` instead and deprecate that one.

Jacalz:
> Add yes - but I don't think we need to deprecate just because there is a generic alternative. Unless there is a reason to remove the non-generic version?

If the generic does basically the same thing and we avoid having to rely on slow (and in my opinion being a bad practice) reflect code was my idea. Type safety is always nicer in my opinion :)

Jacalz:
Done :)

PR #5408, opened 2025-01-14
Initial binding cleanup using generics

Jacalz:
### Description:

I honestly should be studying but I couldn't hold my self back with this one. The cleanup gains are too substantial.

For #5378 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


data/binding/preference_test.go
```diff
@@ -13,12 +13,13 @@ func TestBindPreference_DataRace(t *testing.T) {
 	a := test.NewApp()
 	p := a.Preferences()
 	key := "test-key"
-	const n = 100
 
+	const n = 100
 	var wg sync.WaitGroup
+	wg.Add(n)
```
andydotxyz, line 19: Isn't this considered bad practice? I thought I read you should only increment as each item is added

andydotxyz: Anyhow I'd be tempted to say that in a refactor like this I'd prefer to see no test changes ;)

andydotxyz (commented):
Thanks for this, it looks good indeed - cool stuff.

I've made a comment inline as the test changes seem unnecessary.

On a side note (unimportant) the gen script could have been updated to generate all the API in its original place leading to a clearer diff. With this approach we have to rely on the tests to verify that the function signatures are the same as they were before.

data/binding/preference_test.go
Jacalz: Yeah, this seems like some sort of thing that I accidentally committed after some race fix testing. I have no idea what is the best practice. The official example adds one at a time but I also can't find anything to say that adding them all at once would be bad. It is certainly faster 😅

Jacalz:
Thanks. Yeah, I noticed that too late and the test change was not meant to be there. Will fix :)

andydotxyz (approved):
Sweet thanks so much!

PR #5414, opened 2025-01-15
Fix a delay in double tap that meant it would not fire until timeout …

andydotxyz:
…instead of on second tap

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window.go
+++ b/internal/driver/glfw/window.go
@@ -578,9 +578,7 @@ func (w *window) mouseClickedHandleTapDoubleTap(co fyne.CanvasObject, ev *fyne.P
 
 func (w *window) waitForDoubleTap(co fyne.CanvasObject, ev *fyne.PointEvent) {
 	ctx, mouseCancelFunc := context.WithDeadline(context.TODO(), time.Now().Add(w.driver.DoubleTapDelay()))
-	defer func() {
-		runOnMain(mouseCancelFunc)
-	}()
+	defer runOnMain(mouseCancelFunc)
 	runOnMain(func() {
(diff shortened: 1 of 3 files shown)


internal/driver/glfw/window.go
```diff
@@ -572,11 +578,12 @@ func (w *window) mouseClickedHandleTapDoubleTap(co fyne.CanvasObject, ev *fyne.P
...
 func (w *window) waitForDoubleTap(co fyne.CanvasObject, ev *fyne.PointEvent) {
 	ctx, mouseCancelFunc := context.WithDeadline(context.TODO(), time.Now().Add(w.driver.DoubleTapDelay()))
-	if mouseCancelFunc != nil {
-		defer func() {
-			runOnMain(mouseCancelFunc)
-		}()
-	}
+	defer func() {
+		runOnMain(mouseCancelFunc)
+	}()
```
Jacalz, line 583: Just a small suggestion while I ask another question elsewhere. Would not have bothered with this otherwise:
```suggestion
	defer runOnMain(mouseCancelFunc)
```

Jacalz (changes requested):
Would it be possible to write a test to make sure that it fires after the correct time?

Jacalz:
Indeed. Maybe it is possible to capture the time in the first click and the one for the double click and just subtract them to make sure that the time is less than the delay?

andydotxyz:
> Indeed. Maybe it is possible to capture the time in the first click and the one for the double click and just subtract them to make sure that the time is less than the delay?

Hopefully adding timings to what was there covers it. Seems to be relatively stable.

Jacalz (approved):

Issue #5411, opened 2025-01-14
Migrate to the new threading model more gently


andydotxyz:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

As pointed out by some of our testers, an app that has any amount of background processing will probably crash on basic usage when upgrading to v2.6.0.

### How to reproduce

Run on `develop` and include in your app this minimal code:

```go
go func() {
    w.setTitle("Crash on some systems")
}
```

### Example code

```go
go func() {
    w.setTitle("Crash on some systems")
}
```

### Fyne version

v2.6.0-pre

### Go compiler version

all

### Operating system and version

macOS particularly, but many variants

### Additional Information

Plan:

- [x] Where previously we had `runOnMain` for window handling / OS impacting code we should have a thread check that will be removed later (#5415)
- [x] Add this same safety to common entry-points (`Widget.Refresh()` for one)
- [x] Print clear warnings / errors that the thread management was not correct (#5415)
- [x] Add a way for apps to opt out of this (potentially slow) thread checking
- [x] Make a statement about when this migration path will be disabled (2.8?) - https://github.com/fyne-io/fyne/wiki/Threading-migration

andydotxyz:
Linking this to the parent ticket as well #4654

andydotxyz:
So far I have a utility function that we can use appropriately:

```go
func EnsureMain(fn func()) {
	if IsMainGoroutine() {
		fn()
		return
	}

	fyne.Do(fn)
}
```

And I propose log output along the lines of:

```
2025/01/14 22:18:59 *** Error in Fyne call thread, this should have been called in fyne.Do ***
2025/01/14 22:18:59   At: /Users/andy/Code/Fyne/fyne/cmd/hello/main.go:21
```

If this looks like a good start I can open a PR for that which will avoid a lot of crashes, then we can look at how it would be applied wider and also how to disable it for the project once people have ported over

Jacalz:
As noted in https://github.com/fyne-io/fyne/pull/5420, we need to wait with moving internal maps over because it will cause crashes for anyone with old code still updating elements in goroutines.

andydotxyz:
The main points are all delivered now :)

PR #5415, opened 2025-01-15
Moving window handling functions into a safer transition mode

andydotxyz:
Avoid the most likely crashes as people try out 2.6

 

### Description:

Progressing #5411

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

I tested using the following code:

```go
package main

import (
	"time"

	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Hello")

	w.SetContent(widget.NewButton("Hi!", func() {}))

	go func() {
		time.Sleep(time.Second * 2)
		w2 := a.NewWindow("Two")
		w2.SetContent(widget.NewLabel("Small"))
		w2.Show()

		time.Sleep(time.Second)
		w2.SetContent(widget.NewLabel("Updated"))

		time.Sleep(time.Second)
		w2.Hide()
	}()

	w.ShowAndRun()
}
```

Which worked, but logged this:
```
2025/01/15 14:08:39 *** Error in Fyne call thread, this should have been called in fyne.Do ***
2025/01/15 14:08:39   From: /Users/andy/Code/Fyne/fyne/cmd/hello/main.go:27
2025/01/15 14:08:39 *** Error in Fyne call thread, this should have been called in fyne.Do ***
2025/01/15 14:08:39   From: /Users/andy/Code/Fyne/fyne/cmd/hello/main.go:28
2025/01/15 14:08:40 *** Error in Fyne call thread, this should have been called in fyne.Do ***
2025/01/15 14:08:40   From: /Users/andy/Code/Fyne/fyne/cmd/hello/main.go:31
2025/01/15 14:08:41 *** Error in Fyne call thread, this should have been called in fyne.Do ***
2025/01/15 14:08:41   From: /Users/andy/Code/Fyne/fyne/cmd/hello/main.go:34
```

### DIFF
--- a/internal/async/goroutine.go
+++ b/internal/async/goroutine.go
@@ -1,5 +1,14 @@
 package async
 
+import (
+	"log"
+	"runtime"
+	"strings"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/internal/build"
+)
+
 // mainGoroutineID stores the main goroutine ID.
 // This ID must be initialized in main.init because
 // a main goroutine may not equal to 1 due to the
@@ -12,4 +21,27 @@ func init() {
 
 func IsMainGoroutine() bool {
 	return goroutineID() == mainGoroutineID
+}
+
+func EnsureMain(fn func()) {
+	if build.DisableThreadChecks || IsMainGoroutine() {
+		fn()
+		return
+	}
+
+	log.Println("*** Error in Fyne call thread, this should have been called in fyne.Do ***")
+
+	pc := make([]uintptr, 2)
+	count := runtime.Callers(3, pc)
+	frames := runtime.CallersFrames(pc)
+	frame, more := frames.Next()
+	if more && count > 1 {
+		nextFrame, _ := frames.Next()                     // skip an occasional driver call to itself
+		if !strings.Contains(nextFrame.File, "runtime") { // don't descend into Go
+			frame = nextFrame
+		}
+	}
+	log.Printf("  From: %s:%d", frame.File, frame.Line)
+
+	fyne.Do(fn)
 }
--- a/internal/async/goroutine.go
+++ b/internal/async/goroutine.go
@@ -23,6 +23,18 @@ func IsMainGoroutine() bool {
 	return goroutineID() == mainGoroutineID
 }
 
+func EnsureNotMain(fn func()) {
+	if build.DisableThreadChecks || !IsMainGoroutine() {
+		fn()
+		return
+	}
+
+	log.Println("*** Error in Fyne call thread, fyne.Do called from main goroutine ***")
+
+	logStackTop(2)
+	go fn()
+}
+
 func EnsureMain(fn func()) {
 	if build.DisableThreadChecks || IsMainGoroutine() {
 		fn()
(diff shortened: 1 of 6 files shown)


internal/build/build.go
```diff
@@ -1,2 +1,4 @@
 // Package build contains information about they type of build currently running.
 package build
+
+const DisableThreadChecks = false
```
Jacalz, line 4: This is always false. There is no build tag?

internal/async/goroutine.go
```diff
@@ -13,3 +22,41 @@ func init() {
...
+		return
+	}
+
+	log.Println("*** Error in Fyne call thread, fyne.Do called from main goroutine ***")
+
+	logStackTop(2)
+	go fn()
+}
+
+func EnsureMain(fn func()) {
```
Jacalz, line 38: I kind of feel like these should have comments as they are exported (although in an internal package but anyway)? Applies to the ones above as well.

Jacalz (changes requested):
Thanks. Will test this in Rymdport but this was just a quick drive-by review on my phone in the meantime :)

internal/build/build.go
Jacalz: As for a build tag, I'd rather have something like `new-threading-model` or maybe `using-correct-threads` (or similar) over `disable-thread-checks`. The latter sounds a bit negative like you are now using unsafe and might shoot yourself in the foot 😅

dweymouth: Hmm, not sure I agree with that, to me the current name is more clear what it's doing, where the `new-threading-model` name suggests it would be enabling something bigger (like if we supported both the old and new threading models together and the app could choose which). And the naming of `using-correct-threads` doesn't really suggest what it does at all IMO.

andydotxyz:
@dweymouth @Jacalz thanks for the comments.
How and when to disable this functionality is a follow up PR as outlined in the parent ticket. I don't think that a build flag is the way because it is important that the project is marked as transitioned at a source level IMHO.

My aim is to land this code ASAP so developers on this branch get the immediate benefit and we get more testing.

internal/build/build.go
andydotxyz: Sorry for the confusion, there is a local TODO in my code maybe i should have left it here ;).
Disabling is a follow up as that is where the naming discussions come in!

Jacalz:
> I don't think that a build flag is the way because it is important that the project is marked as transitioned at a source level IMHO.

How would you otherwise do it? I need another way of turning it off for Flatpak compared to FyneApp.toml. What I would propose, like on Slack, is having a build tag that gets set by cmd/fyne when a value in FyneApp.toml is set.

Jacalz:
Thanks for the doc comments. I don't think we need "Since:" for internal packages but you don't have to remove it :)

andydotxyz:
> How would you otherwise do it? I need another way of turning it off for Flatpak compared to FyneApp.toml. What I would propose, like on Slack, is having a build tag that gets set by cmd/fyne when a value in FyneApp.toml is set.

The point of following up with another PR is so we can have that discussion - I don't know the answer for sure and want to discuss it. It's pretty clear from people testing develop that we need this - so can we get it added before deciding how and when it can be disabled?

Jacalz (approved):
I'm still in favour of a build tag to disabled this but approved for now as we obviously need that transition for older codebases.

andydotxyz:
> I'm still in favour of a build tag to disabled this but approved for now as we obviously need that transition for older codebases.

Thanks. I will open the follow-on PR shortly and we can discuss further.

PR #5420, opened 2025-01-16
One big last lock cleanup

Jacalz:
### Description:

This removes most of the unnecessary atomic types everything except for app, bindings and so on. It also fixes various small races and cleans up some code leftovers. Maps are still concurrency safe as anyone relying on goroutines might see crashes otherwise.

For #4654

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/color_channel.go
+++ b/dialog/color_channel.go
@@ -2,7 +2,6 @@ package dialog
 
 import (
 	"strconv"
-	"sync/atomic"
 
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/canvas"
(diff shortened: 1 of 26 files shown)


Jacalz:
I don't quite understand why those races are happening. fyne_demo is fine and I can't quite figure out what is triggering in the race in those tests. Do you have any idea? @andydotxyz

Jacalz:
I backed out all changes to maps. It causes crashes for anyone updating widgets outside of the correct threads.

andydotxyz (approved):
Watch out that too many unrelated things don't get mixed in.
But thanks for all the tidyup.

PR #5421, opened 2025-01-16
Clear out pending window operations after start

Jacalz:
### Description:

Found a small memory leak when looking through the code. Let's make sure to clear out the pending operations once when have executed them. This is hard to test with how glfw tests works so did not figure a way to add a test.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Thanks :)

PR #5422, opened 2025-01-16
Implement more efficient run loop

### LINKED ISSUE #2506: An idle application with no visible windows still uses noticeable CPU time

Jacalz:
### Description:

* The main loop has a new "idle" mode, where it is blocked on a call to `glfw.WaitEvents()` waiting for input events from the OS/GLFW. Starting an animation, calling fyne.Do[AndWait], and canvas.SetDirty (without threading migration) post empty events to wake up the main loop.

* Most GLFW event callbacks now do not process the event immediately, but queue it to be processed in the main loop. The exception is the Resize callback, since glfw.WaitEvents does not return until a window resize is fully completed.

* When `glfw.WaitEvents` returns, the main loop enters "active" mode, where a 60 Hz timer is started, and each tick, it runs the draw single frame actions, which now include processing the queued GLFW events. It returns to "idle" mode only when no events or animations were available to run within a tick.

Fixes #2506

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/loop.go
+++ b/internal/driver/glfw/loop.go
@@ -50,11 +50,11 @@ func runOnMainWithWait(f func(), wait bool) {
 		defer common.DonePool.Put(done)
 
 		funcQueue.In() <- funcData{f: f, done: done}
-		postEmptyEvent()
+		wakeUpDriver()
 		<-done
 	} else {
 		funcQueue.In() <- funcData{f: f}
-		postEmptyEvent()
+		wakeUpDriver()
 	}
 }
 
@@ -141,6 +141,18 @@ func (d *gLDriver) runGL() {
 		f()
 	}
 
+<<<<<<< HEAD
+=======
+	if stg, ok := fyne.CurrentApp().Settings().(interface{ AddChangeListenerFunc(func(fyne.Settings)) }); ok {
+		stg.AddChangeListenerFunc(func(s fyne.Settings) {
+			settingsMutex.Lock()
+			settingsToApply = s
+			settingsMutex.Unlock()
+			wakeUpDriver()
+		})
+	}
+
+>>>>>>> 3174f9b66 (Nicer naming for waking up driver)
 	for {
 		d.waitEvents()
 
--- a/internal/driver/glfw/loop.go
+++ b/internal/driver/glfw/loop.go
@@ -220,7 +220,9 @@ func (d *gLDriver) runSingleFrame() (exit, animationsDone bool) {
 		select {
 		case f := <-funcQueue:
 			f.f()
-			f.done <- struct{}{}
+			if f.done != nil {
+				f.done <- struct{}{}
+			}
 		default:
 			funcsDone = true
 		}
--- a/internal/driver/glfw/loop.go
+++ b/internal/driver/glfw/loop.go
@@ -136,84 +136,19 @@ func (d *gLDriver) runGL() {
 		f()
 	}
 
-<<<<<<< HEAD
-	eventTick := time.NewTicker(time.Second / 60)
-	for {
-		select {
-		case <-d.done:
-			eventTick.Stop()
-			d.Terminate()
-			l := fyne.CurrentApp().Lifecycle().(*app.Lifecycle)
-			if f := l.OnStopped(); f != nil {
-				l.QueueEvent(f)
-			}
-			return
-		case f := <-funcQueue.Out():
-			f.f()
-			if f.done != nil {
-				f.done <- struct{}{}
-			}
-		case <-eventTick.C:
-			d.pollEvents()
-			for i := 0; i < len(d.windows); i++ {
-				w := d.windows[i].(*window)
-				if w.viewport == nil {
-					continue
-				}
-
-				if w.viewport.ShouldClose() {
-					d.destroyWindow(w, i)
-					i-- // Trailing windows are moved forward one step.
-					continue
-				}
-
-				expand := w.shouldExpand
-				fullScreen := w.fullScreen
-
-				if expand && !fullScreen {
-					w.fitContent()
-					shouldExpand := w.shouldExpand
-					w.shouldExpand = false
-					view := w.viewport
-
-					if shouldExpand && runtime.GOOS != "js" {
-						view.SetSize(w.shouldWidth, w.shouldHeight)
-					}
-				}
-			}
-
-			d.animation.TickAnimations()
-			d.drawSingleFrame()
-=======
-	settingsChange := make(chan fyne.Settings)
-	fyne.CurrentApp().Settings().AddChangeListener(settingsChange)
-
 	for {
 		d.waitEvents()
 
 		if d.animation.HasAnimations() {
-			// Switch to running at 60 fps with d.pollEvents
-			// until we have no more animations to tick
-			t := time.NewTicker(time.Second / 60)
-			for range t.C {
-				d.pollEvents()
-				exit, animationsDone := d.runSingleFrame(settingsChange)
-				if exit {
-					t.Stop()
-					return
-				}
-				if animationsDone {
-					t.Stop()
-					break
-				}
+			// run animations while available
+			if exit := d.runAnimationLoop(settingsChange); exit {
+				return
 			}
 		} else {
 			// idle mode: run single frame and sleep on d.waitEvents() above
-			exit, _ := d.runSingleFrame(settingsChange)
-			if exit {
+			if exit, _ := d.runSingleFrame(settingsChange); exit {
 				return
 			}
->>>>>>> 58a8be1bf (Update loop.go - switch between waitEvents and pollEvents at 60fps with animations)
 		}
 	}
 }
@@ -235,8 +170,23 @@ func (d *gLDriver) destroyWindow(w *window, index int) {
 	}
 }
 
-func (d *gLDriver) repaintWindow(w *window) bool {
-=======
+// runs a loop that invokes runSingleFrame at 60 fps until there are no more animations
+// uses pollEvents to check for events every frame
+func (d *gLDriver) runAnimationLoop(settingsChange <-chan fyne.Settings) bool {
+	t := time.NewTicker(time.Second / 60)
+	defer t.Stop()
+	for range t.C {
+		d.pollEvents()
+		exit, animationsDone := d.runSingleFrame(settingsChange)
+		if exit {
+			return exit
+		} else if animationsDone {
+			break
+		}
+	}
+	return false
+}
+
 func (d *gLDriver) runSingleFrame(settingsChange <-chan fyne.Settings) (exit, animationsDone bool) {
 	// check if we're shutting down
 	if !running.Load() {
@@ -344,7 +294,6 @@ func (d *gLDriver) removeWindows(count int) {
 }
 
 func (d *gLDriver) repaintWindow(w *window) {
->>>>>>> 58a8be1bf (Update loop.go - switch between waitEvents and pollEvents at 60fps with animations)
 	canvas := w.canvas
 	freed := false
 	w.RunWithContext(func() {
--- a/internal/driver/glfw/loop.go
+++ b/internal/driver/glfw/loop.go
@@ -136,6 +136,7 @@ func (d *gLDriver) runGL() {
 		f()
 	}
 
+<<<<<<< HEAD
 	eventTick := time.NewTicker(time.Second / 60)
 	for {
 		select {
@@ -183,10 +184,40 @@ func (d *gLDriver) runGL() {
 
 			d.animation.TickAnimations()
 			d.drawSingleFrame()
+=======
+	settingsChange := make(chan fyne.Settings)
+	fyne.CurrentApp().Settings().AddChangeListener(settingsChange)
+
+	for {
+		d.waitEvents()
+
+		if d.animation.HasAnimations() {
+			// Switch to running at 60 fps with d.pollEvents
+			// until we have no more animations to tick
+			t := time.NewTicker(time.Second / 60)
+			for range t.C {
+				d.pollEvents()
+				exit, animationsDone := d.runSingleFrame(settingsChange)
+				if exit {
+					return
+				}
+				if animationsDone {
+					t.Stop()
+					break
+				}
+			}
+		} else {
+			// idle mode: run single frame and sleep on d.waitEvents() above
+			exit, _ := d.runSingleFrame(settingsChange)
+			if exit {
+				return
+			}
+>>>>>>> 58a8be1bf (Update loop.go - switch between waitEvents and pollEvents at 60fps with animations)
 		}
 	}
 }
 
+<<<<<<< HEAD
 func (d *gLDriver) destroyWindow(w *window, index int) {
 	w.visible = false
 	w.viewport.Destroy()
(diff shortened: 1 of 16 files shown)


Jacalz:
@dweymouth Feel free to test this out and help out getting it to the finish line if you'd like. There are some problems remaining but it is working out quite well. With no animations on screen, we will wait for events to occur.

Jacalz:
This might not even be relevant any more with threading changes? It seems to me like CPU usage is quite unchanged between the two, even for completely idle windows with no animations to run.

dweymouth:
I quickly tried out something that was closer to the idea I was thinking of - removing the time.Ticker entirely that currently drives the event loop (the `<-t.C` case in the select became a default). With that, CPU does truly go down to 0 on idle, though it would still need some more work to handle the animation case then. My idea would be to switch to a loop being driven by the Ticker and pollEvents while there exist animations to run, and switching to waitEvents with no Ticker running for idle. Also, fyne.Do will need to post an empty event as well to wake up the main thread. (And during the transitional period, maybe canvas.Refresh as well in case people call it directly or indirectly (via Refresh on a widget) without using fyne.Do)

dweymouth:
also, looking at this code again, I think I discovered an existing bug. Here: https://github.com/fyne-io/fyne/pull/5422/files#diff-c4f15824f3446f8e7e44764227aa45dd5f43c4981c462270d40dfaccef9b2198L142

The TickAnimations and drawSingleFrame calls should be outside the per-window loop, not within it, as far as I can tell.

Jacalz:
Thanks for the notes and ideas. To be fair, fyne.Do already does post an empty event in this version (through runOnMain) but it did not seem to have much effect.

I might have a look at this in the coming days but I just want to say that you are absolutely welcome to commit to this branch of you want to collaborate :)

Jacalz:
> The TickAnimations and drawSingleFrame calls should be outside the per-window loop, not within it, as far as I can tell.

Cool. Feel free to open a PR if you'd like :)

dweymouth:
Pushed updates to switch between sleeping (waitEvents) and polling at 60fps depending on whether there are animations to run. This brings CPU down to 0 and no idle wakeups showing, according to Mac activity monitor, when we have no animations to tick. I'm seeing the window fail to exit when closed for some reason in fyne_demo but I was also seeing that before my changes, so something that still needs to be addressed.

The logic of the loop.go updates was to move the select body out of the ticker and into its own "runSingleFrame" func. Also we are checking all the channels (done, funcs, settings) per frame now instead of just selecting one of them.

Jacalz:
Well done. I will look through the code more but I can verify that CPU usage is at 0% with no animations here as well.

andydotxyz:
Should we use this PR to resolve the "freezes on resize" issue as well - the solution may be the same code change.
It may already be fixed - I have not run this code yet.

dweymouth:
> Should we use this PR to resolve the "freezes on resize" issue as well - the solution may be the same code change.
It may already be fixed - I have not run this code yet.

This doesn't fix the "freezes on resize" issue - that seems to be caused by GLFW as both `waitEvents` and `pollEvents` block while the window is being resized. I'm actually not sure we can fix it in pure Go code (or if it's even fixable?).

dweymouth:
I ran this again on M1 Mac and it is not freezing on window exiting. I think I had something wrong in my local state last time since I applied the patch manually to my branch, vs now I'm running it straight from cloning jacalz/fyne and it seems to work fine, no bugs that I can tell clicking around in fyne_demo. (Except it doesn't fix the aforementioned freeze-on-resize)

andydotxyz:
> > Should we use this PR to resolve the "freezes on resize" issue as well - the solution may be the same code change.
> > It may already be fixed - I have not run this code yet.
> 
> This doesn't fix the "freezes on resize" issue - that seems to be caused by GLFW as both `waitEvents` and `pollEvents` block while the window is being resized. I'm actually not sure we can fix it in pure Go code (or if it's even fixable?).

There must be something we can do, as it was working before falling back to one thread ;)

dweymouth:
This may be relevant (for the freeze-on-resize, which is orthogonal to this PR):

https://stackoverflow.com/questions/45880238/how-to-draw-while-resizing-glfw-window

Jacalz:
Can we not post an empty event when doing a resize?

dweymouth:
From the StackOverflow, it sounds like we may need to hook into the GLFW window_refresh_callback (and/or the framebuffer_size_callback) and maybe just post an empty event from within those callbacks. Anyhow, I think that should be a follow-up, not done in this PR, since this PR achieves the CPU improvement without changing the behavior at all of the draw-on-resize issue

Jacalz:
Yeah, I agree

Jacalz:
This PR is almost working perfectly on my end. Changing the theme from fyne_setting or the dark/light theme control on Gnome does not update until I move my mouse to the window when there are no animations running.

Also: Do note that I think animations for objects that are not visible are still running? I notice that if I go to the progressbar tab and then move away from there, there are still animations ticking in the backround making the window use the less efficient path.

internal/driver/glfw/loop.go
```diff
@@ -97,94 +98,134 @@ func (d *gLDriver) runGL() {
...
-			for _, win := range d.windowList() {
-				w := win.(*window)
-				if w.viewport == nil {
-					continue
+		d.waitEvents()
+
+		if d.animation.HasAnimations() {
+			// Switch to running at 60 fps with d.pollEvents
+			// until we have no more animations to tick
+			t := time.NewTicker(time.Second / 60)
```
Jacalz, line 107: FYI @dweymouth: This is a memory leak for anyone using Go 1.22 or earlier, see https://pkg.go.dev/time#NewTicker.

internal/driver/glfw/loop.go
```diff
@@ -97,93 +98,137 @@ func (d *gLDriver) runGL() {
...
-					}
-
-					newWindows = append(newWindows, win)
-				}
+			c.applyThemeOutOfTreeObjects()
+			c.reloadScale()
+		})
+	default:
+		break
+	}
```
Jacalz, line 167: I'd need to find a source for it but I am quite sure that having a single select, if possible, would be a lot more efficient with how Go deals with non-blocking channels.

Jacalz: Yeah, I thought about that as well. Do we even need channels with the new threading model?

internal/driver/glfw/loop.go
dweymouth: Updated

internal/driver/glfw/loop.go
dweymouth: I kept the channels for the func queue and settings listeners for now, to avoid a bigger change. (Implementing our own non-channeled queue would take a bit of work). But could be nice to move to eventually. Also I changed it so it will run all available queued funcs in one frame, vs only picking at most one per frame.

Jacalz: Nice. Sounds good

internal/driver/glfw/loop.go
```diff
@@ -97,93 +98,138 @@ func (d *gLDriver) runGL() {
...
+				return
 			}
-			return
+		}
+	}
+}
+
+func (d *gLDriver) runSingleFrame(settingsChange <-chan fyne.Settings) (exit, animationsDone bool) {
+	// check if we're shutting down
+	if d.done.Load() {
```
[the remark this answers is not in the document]
dweymouth: Good point

internal/driver/glfw/loop.go
Jacalz: Ah, never mind. You call stop on it everywhere. Is that fine?

Jacalz: I suppose it probably is but it seems bound to fail us in the future when someone updates the code and misses a stop. Can defer work?

dweymouth: Only if we create an anonymous func to encompass this little code block I guess. I suppose it's up to you, I prefer it as it is but I don't think it's important either way

Jacalz: Someone not running Go 1.23 might get huge memory buildup if the app is running a long time. I think it is worth it but we can add a comment to remove the workaround code once 1.23 is our base.

Jacalz:
Do you think it would be worth it to add some check if we are running in `glfw.WaitEvents()` mode and not post an empty event in that case? I assume we want to do as few CGo calls as possible but I'm not sure it is worth it?

internal/driver/glfw/loop.go
dweymouth: Updated

dweymouth:
> Do you think it would be worth it to add some check if we are running in `glfw.WaitEvents()` mode and not post an empty event in that case? I assume we want to do as few CGo calls as possible but I'm not sure it is worth it?

Probably not, or at the least it would be premature optimization. It's the waitEvents case where we *do* have to post an event to wake up main, and that's the default mode. Only if there are actively running animations do we use pollEvents where we could theoretically avoid posting an empty event. (I suppose also we could avoid posting an empty event from fyne.Do if we know there are already queued funcs that haven't run yet, now that we run them all in one go, but that would require complex signaling and locking I think)

dweymouth:
> Also: Do note that I think animations for objects that are not visible are still running? I notice that if I go to the progressbar tab and then move away from there, there are still animations ticking in the background making the window use the less efficient path.

This has always been the case and I'm not sure what we can do about it -- we don't currently have a system for tracking when a widget first becomes invisible. Fyne_demo could use the OnChanged hook from the tabs to call .Stop() on the infinite progress bar maybe? Or we just don't care since it's just fyne_demo :)

Jacalz:
Good point. I think we likely want to create a good example and hide it when changing tabs :)

dweymouth:
Settings change working now for me. Had to add a new API to the settings (keeping it internal for now with anonymous interface casting) to invoke listeners via a func instead of a channel, since a func allows us to post empty event to wake up main. Attempting to post empty event directly from the settings didn't work because of timing with the postEmptyEvent

dweymouth:
Oh I noticed tables don't scroll smoothly with the Mac trackpad. I think it sends more than 60 events per second perhaps and table is doing too much per scroll... may need to look into combining events now and rate limiting drawing somehow

dweymouth:
Alternatively, we could switch back to 60fps polling when we receive an event, and go back to idle mode only when we receive no events within 1/60 sec. That way we should be invoking widget code similarly to before with max 60fps events (except maybe for that first one that wakes the loop out of idle mode)

dweymouth:
> Do you think it would be worth it to add some check if we are running in `glfw.WaitEvents()` mode and not post an empty event in that case? I assume we want to do as few CGo calls as possible but I'm not sure it is worth it?

Returning to this, I think it may be worth it if we move to a custom slice+lock-based func queue, to avoid posting the empty event if there are multiple funcs in the queue (since we know the driver will be awake processing them then). We cannot do it with a simple `asleep atomic.Bool`, since if the main thread were to set `asleep.Store(true)` and then call `glfw.WaitEvents()` there is still the unavoidable race that: 1) goroutine checks asleep and sees it's false, 2) main thread sets asleep = true, then calls glfw.WaitEvents to go to sleep`, 3) goroutine adds func to the queue and doesn't post empty event since it thinks main is awake.

But I believe it could be done with the custom func queue, because pulling the head from the queue, as well as adding a func to the queue and posting an event if needed, can each be made atomic operations

dweymouth:
@Jacalz if you have some time could you rebase this to the latest develop? I want to see how much Andy's scroller PR may have smoothed things up with resizing the Table/Tree in fyne_demo

andydotxyz (changes requested):
I haven't fully reviewed this code but we cannot merge at an uncapped frame rate.

We should not be painting to the screen any faster than it's FPS (or 60 fallback) irrespective of how many background processes tell us there are changes to paint.

nagisa:
What sort of work remains here? Is there anything people not super familiar with fyne can do to help? The persistently running tick loop is currently one of the larger power consumers on my PC due to an application built upon fyne producing a significant number of processor events/wakeups in the background.

andydotxyz:
> What sort of work remains here? Is there anything people not super familiar with fyne can do to help? The persistently running tick loop is currently one of the larger power consumers on my PC due to an application built upon fyne producing a significant number of processor events/wakeups in the background.

Testing is what it needs. This was put on hold until the previous release was completed but it should probably be revived now as 2.8.0 is the intended target.

nagisa:
Alright, I crudely patched supersonic to use this branch and been running it throughout the day. It launches and appears to run.

The only new issue (comparing to a build against fyne @ af19611ffc5087fcc47a2c8a6448256458b3c52e, which is the parent commit to this PR) I could find is that application exits when I click around tabs of supersonic's about or preferences window. The GPL v3 tab takes a little while to load up, and switching to another tab after that will have the application exit. No error I could copy though.

Outside of that I don't see any other new problems!

cubancorona:
Quick note from some outside testing on this branch — we hit the same idle-loop issue on mobile, which led us here.

Good news first: we checked whether the loop reliably wakes for everything that can change the screen, and it looks fully covered. We couldn't find a case where the UI changes but silently fails to redraw until you nudge the mouse.

One thing worth a look: a rare, pre-existing animation glitch this change could expose. Now that the loop sleeps when idle, an animation started from a background task at the wrong instant could occasionally get stranded and never play — something the old always-on loop quietly papered over. Narrow and probably rare, but easy to miss before a release.

How we checked: we walked through every category that updates the UI (refreshes, data/background updates, animations, theme changes, menus, scrolling…), confirmed each wakes the loop, and re-verified the suspicious ones to rule out false alarms.

Happy to share specifics and a small proposed fix, or help test toward 2.8.0. (The tab-switch crash @nagisa hit looks separate — an actual crash, not a wake issue — glad to help repro that too.)

andydotxyz:
@dweymouth should we get this fixed to pull in at the last minute, or is that too risky now for 2.8?

dweymouth:
> @dweymouth should we get this fixed to pull in at the last minute, or is that too risky now for 2.8?

I know of an issue on Mac where an interactive window resize sometimes gets "stuck" as in the window retains the resize cursor after releasing the drag, no hover events are sent to the window, and then the first click inside the window gets interpreted as the drag being let go at that window size. It is rare and hard to trigger but I don't think it happens without this change and I don't know the root cause either

dweymouth:
Though I guess it's worth considering if introducing this rare bug is worth the trade off to reduce the wasted cpu and power consumption of the always-on loop

andydotxyz:
I think it would be classified as a regression which isn't something I'd support adding?

cubancorona:
Heads up — on macOS I checked out this branch fresh and `go test ./internal/driver/glfw/` hangs: `TestWindow_NoDragEndWithoutDraggedEvent` times out blocked in `funcQueue.Push` (funcqueue.go:51), while that same test passes on `develop`.

PR #5424, opened 2025-01-17
Use async.Map for canvases cache

Jacalz:
### Description:

This fixes a crash for unmigrated code bases where a concurrent map write happened. This was due to a lock I removed internally. The lock protected other things than just the map so we should still be faster but now with the lock removed, it made it possible to see a small simplification/optimization that could be done so I cleaned up that code as well. 

For #5411

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/cache/canvases.go
+++ b/internal/cache/canvases.go
@@ -2,13 +2,14 @@ package cache
 
 import (
 	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/internal/async"
 )
 
-var canvases = make(map[fyne.CanvasObject]*canvasInfo, 1024)
+var canvases async.Map[fyne.CanvasObject, *canvasInfo]
 
 // GetCanvasForObject returns the canvas for the specified object.
 func GetCanvasForObject(obj fyne.CanvasObject) fyne.Canvas {
-	cinfo, ok := canvases[obj]
+	cinfo, ok := canvases.Load(obj)
 	if cinfo == nil || !ok {
 		return nil
 	}
@@ -21,9 +22,8 @@ func GetCanvasForObject(obj fyne.CanvasObject) fyne.Canvas {
 func SetCanvasForObject(obj fyne.CanvasObject, c fyne.Canvas, setup func()) {
 	cinfo := &canvasInfo{canvas: c}
 	cinfo.setAlive()
-	old, found := canvases[obj]
-	canvases[obj] = cinfo
 
+	old, found := canvases.LoadOrStore(obj, cinfo)
 	if (!found || old.canvas != c) && setup != nil {
 		setup()
 	}
--- a/internal/driver/glfw/canvas_test.go
+++ b/internal/driver/glfw/canvas_test.go
@@ -365,9 +365,10 @@ func TestGlCanvas_MinSizeShrinkTriggersLayout(t *testing.T) {
 	})
 	repaintWindow(w)
 
+	assert.Equal(t, oldCanvasSize, c.Size())
+	expectedRightColSize := oldRightColSize.Subtract(fyne.NewSize(20, 0))
+
 	runOnMain(func() {
-		assert.Equal(t, oldCanvasSize, c.Size())
-		expectedRightColSize := oldRightColSize.Subtract(fyne.NewSize(20, 0))
 		assert.Equal(t, expectedRightColSize, rightCol.Size())
 		assert.Equal(t, fyne.NewSize(100, 40), leftObj1.Size())
 		assert.Equal(t, fyne.NewSize(80, 30), rightObj1.Size())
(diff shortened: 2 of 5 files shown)


internal/cache/canvases.go
```diff
@@ -2,13 +2,14 @@ package cache
 
 import (
 	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/internal/async"
 )
 
-var canvases = make(map[fyne.CanvasObject]*canvasInfo, 1024)
```
dweymouth, line 7: Can we put a comment here that this is an async.Map just for compatibility reasons and should be switched back to regular map at 3.0 or whenever we decide to fully run down the threading migration plan?

Jacalz: I don't think that is necessary. I reverted all my other async.Map removals so there is basically a lot more to move over and it is tracked in the migration plan.

andydotxyz (dismissed):
Thanks

andydotxyz:
Looks good but I think there seems to be a race associated...

Jacalz:
Indeed. I think I managed to fix the test now

internal/driver/glfw/canvas_test.go
```diff
@@ -348,24 +348,31 @@ func TestGlCanvas_MinSizeShrinkTriggersLayout(t *testing.T) {
 	w.SetContent(content)
 
 	oldCanvasSize := fyne.NewSize(200+3*theme.Padding(), 100+3*theme.Padding())
-	assert.Equal(t, oldCanvasSize, c.Size())
+	runOnMain(func() {
+		assert.Equal(t, oldCanvasSize, c.Size())
```
andydotxyz, line 352: This doesn't make sense as `c.Size()` is on safeCanvas and already on main

andydotxyz (commented):
Thanks makes sense - bit is it possible that some of the test fix is not needed (in fact possibly wrong as it would be a runOnMain in another?)

internal/driver/glfw/canvas_test.go
Jacalz: Ah, it did not deadlock because you have a check for gorutine ID in there. Will fix!

Jacalz: Fixed it up just now

andydotxyz:
I think there is still an issue:

```
  fyne.io/fyne/v2/internal/cache.Clean()
      /home/runner/work/fyne/fyne/internal/cache/base.go:54 +0x204
  fyne.io/fyne/v2/test.(*canvas).Capture()
      /home/runner/work/fyne/fyne/test/canvas.go:91 +0x38
  fyne.io/fyne/v2/internal/driver/glfw.TestMenuBar.func10.3.2()
      /home/runner/work/fyne/fyne/internal/driver/glfw/menu_bar_test.go:318 +0x1c1
```

Jacalz:
Gash. It is running free from races locally for me :/

Jacalz:
I put in some safety around that canvas capture

Jacalz:
I think the problem is that the tests are all using the same canvas cache in the bottom and the driver that is running continuously for the driver tests (from `TestMain`) is also accessing it as it is spinning in the runloop. There is basically artificial races even in test code that should not race.

Jacalz:
Hmm. I actually managed to get a race in fyne_demo just now. Might be something strange going on

Jacalz:
Hmm. Strange race. Not sure if it is related to this. Will investigate

Jacalz:
Hmm, nope. The advanced screen is giving me races now on latest develop as well.

Jacalz:
It looks like the fyne.Do call in the advanced screen actually doesn't run on the right thread for some reason. Anyway, not related to this.

andydotxyz:
> Gash. It is running free from races locally for me :/

Yeah I grapple with that a lot myself too - I guess we should be glad that the CI finds a lot more!

internal/driver/glfw/menu_bar_test.go
```diff
@@ -428,18 +428,20 @@ func TestMenuBar(t *testing.T) {
...
-						lastAction = ""
-						for _, key := range tt.keys {
-							c.Focused().TypedKey(&fyne.KeyEvent{
-								Name: key,
-							})
-						}
-						test.AssertRendersToMarkup(t, "menu_bar_kbdctrl_"+name+".xml", c)
-						assert.Equal(t, tt.wantAction, lastAction, "last action should match expected")
-					}
+					captured = c.Capture()
```
andydotxyz, line 433: Capture on safeCanvas is already safe isn't it?

andydotxyz: Oh, I see this is a test window. Maybe that is where some races are coming from? Mixing the test and GLFW windows?

Jacalz: Yeah, exactly. I got strange errors when I changed to the safe window. Mixing it in the same package does make the app run in the background and race on canvas internals

Jacalz:
Indeed. It could be the other way around

Jacalz:
All green now :)

andydotxyz (approved):

PR #5425, opened 2025-01-17
Allow disabling of the thread checks as a "migration" - a pattern we can re-use

andydotxyz:
Reasoning:
* Build tags are great to run a specific build with a feature on or off but can be forgotten
* App metadata is great for marking a project as migrated but are not part of the binary
* So we have both. Use the tag if you want and add the migration to file when you're looking to make it permanent.

With this pattern we could (like the Go module tooling approach) change how apps work based on metadata, but build it in to the app through compiler flags.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/hello/FyneApp.toml
+++ b/cmd/hello/FyneApp.toml
@@ -4,3 +4,6 @@
   ID = "io.fyne.hello"
   Version = "2.5.3"
   Build = 2
+
+[Migrations]
+  fyneDo = true
--- a/internal/async/goroutine.go
+++ b/internal/async/goroutine.go
@@ -30,7 +30,7 @@ func IsMainGoroutine() bool {
 //
 // This will be removed later and should never be public
 func EnsureNotMain(fn func()) {
-	if (build.DisableThreadChecks || !build.HasHints) || !IsMainGoroutine() {
+	if build.DisableThreadChecks || !build.HasHints || !IsMainGoroutine() {
 		fn()
 		return
 	}
@@ -47,7 +47,7 @@ func EnsureNotMain(fn func()) {
 //
 // This will be removed later and should never be public
 func EnsureMain(fn func()) {
-	if (build.DisableThreadChecks || !build.HasHints) || IsMainGoroutine() {
+	if build.DisableThreadChecks || !build.HasHints || IsMainGoroutine() {
 		fn()
 		return
 	}
--- a/internal/async/goroutine.go
+++ b/internal/async/goroutine.go
@@ -30,7 +30,7 @@ func IsMainGoroutine() bool {
 //
 // This will be removed later and should never be public
 func EnsureNotMain(fn func()) {
-	if build.DisableThreadChecks || !build.HasHints || !IsMainGoroutine() {
+	if build.MigratedToFyneDo() || !build.HasHints || !IsMainGoroutine() {
 		fn()
 		return
 	}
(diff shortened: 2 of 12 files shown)


internal/async/goroutine.go
```diff
@@ -30,7 +30,7 @@ func IsMainGoroutine() bool {
 //
 // This will be removed later and should never be public
 func EnsureNotMain(fn func()) {
-	if build.DisableThreadChecks || !IsMainGoroutine() {
+	if (build.MigratedToFyneDo() || !build.HasHints) || !IsMainGoroutine() {
```
Jacalz, line 33: You can remove the unnecessary parenthesis, assuming you did not mean to have an `&&` in there?

cmd/hello/FyneApp.toml
```diff
@@ -4,3 +4,6 @@
   ID = "io.fyne.hello"
   Version = "2.5.3"
   Build = 2
+
+[Migrations]
+  fyneDo = true
```
Jacalz, line 9: Same comment as further down. Maybe calling this `threading = true` would make more sense?

Jacalz (changes requested):
Thanks. I really like this approach. Added some comments inline.

Are we still adding `cmd/fyne` changes here or should we focus on the tools repo instead?

andydotxyz:
> Are we still adding `cmd/fyne` changes here or should we focus on the tools repo instead?

Good question. I wanted to do it here as one PR for discussion, will migrate to tools once agreed.

internal/build/build.go
```diff
@@ -1,4 +1,16 @@
 // Package build contains information about they type of build currently running.
 package build
 
-const DisableThreadChecks = false
+import "fyne.io/fyne/v2"
+
+func MigratedToFyneDo() bool {
```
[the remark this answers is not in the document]
andydotxyz: I removed it for thread safety reasons

internal/build/migrated_notfynedo.go
```diff
@@ -0,0 +1,5 @@
+//go:build !migrated_fynedo
+
+package build
+
+const DisableThreadChecks = false
```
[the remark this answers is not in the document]
andydotxyz: Sorry I missed this as it was just a renamed file

Jacalz:
FYI: With the last commit, the thread check lookup being variables means that stuff aren't optimized away any more. You could just revert that commit and use a `sync.One` to store that metadata value globally once instead.

internal/build/migrated_fynedo.go
```diff
@@ -0,0 +1,5 @@
+//go:build migrated_fynedo
```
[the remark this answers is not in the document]
andydotxyz: I'd thing that "threading" is generic which is potentially problematic too.
The reasoning here is that the name matches an API that developers will have used to prepare this migration.

cmd/hello/FyneApp.toml
Jacalz: Hmm, true but I'm not all too enthusiastic about the previous name either. I would like to hear what @dweymouth thinks.

andydotxyz:
@Jacalz i think it's good now?

Jacalz:
Three is the naming discussion that I pinged Drew about still.

cmd/hello/FyneApp.toml
Jacalz: If we are to do a v3.0.0, then we can remove the tag and use the same tag again in the future in case we want to do one more thread migration (which I kind of think is unlikely before v3.0.0). I am still in favour of migrating threads as a better name (`fynedo` sounds a bit strange) but I approved this for now so it can land and be useful for more people.

PR #5427, opened 2025-01-17
Remove closed windows in place without allocating a new slice

Jacalz:
### Description:

Just a small improvement that I noticed could be done, thanks to everything happening on one thread, while working on #5422. Will give a bit of a merge conflict but this seems to make more sense separately (especially if that for some reason wouldn't be ready in time for 2.6).

This also includes a fix from Drew from the same PR to address window repaints and animation ticking happening once for every window on each frame tick. We were basically doing a frame render `O(n^2)` times, where n is amount of visible windows.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/canvas_test.go
+++ b/internal/driver/glfw/canvas_test.go
@@ -427,7 +427,10 @@ func TestGlCanvas_ResizeWithOtherOverlay(t *testing.T) {
 	content := widget.NewLabel("Content")
 	over := widget.NewLabel("Over")
 	w.SetContent(content)
-	w.Canvas().Overlays().Add(over)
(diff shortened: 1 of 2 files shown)


andydotxyz:
Looks like a race or two opened up in the tests

Jacalz:
The race seems like the same one I fixed in the other PR, I think

andydotxyz:
> The race seems like the same one I fixed in the other PR, I think

Ah, perhaps - let's get that in and rebase to make sure.

Jacalz:
The race does indeed look to be the same :)

andydotxyz (approved):

PR #5428, opened 2025-01-17
Loading is now synchronous so we don't have to track it

### LINKED ISSUE #5393: Investigate suspicious preference write delay
### Checklist

Noted in one of the race fixes PRs, this sleep looks suspicious. Sleeping to avoid race conditions is almost always a hacky solution around what really is lack of synchronization.  We should look at if this sleep is correct.

https://github.com/fyne-io/fyne/blob/4a875d9d5677ce8a98bd15787d6f74d4263cac16/app/preferences.go#L50-L63

Yes

We need to find out how these writes are happening and in what way operations need to be synchronized. Perhaps running them on the same thread does not cause any problems? Maybe we just need some kind of locking inbetween?

andydotxyz:
Fixes #5393

### Checklist:

- [x] Tests included. <- existing tests are sufficient
- [x] Tests all pass.

### DIFF
--- a/app/preferences.go
+++ b/app/preferences.go
@@ -15,7 +15,6 @@ type preferences struct {
 	*internal.InMemoryPreferences
 
 	prefLock            sync.RWMutex
-	loadingInProgress   bool
 	savedRecently       bool
 	changedDuringSaving bool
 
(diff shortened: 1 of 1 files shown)


Jacalz (approved):
I must admit that I love when a diff looks this red :)

PR #5431, opened 2025-01-19
Add new callback-based settings listener API

dweymouth:
### Description:

The existing Settings.AddChangeListener needs to be deprecated, since it is channel-based, and user code cannot block on channel receives from the main thread, and communicating to user code via a goroutine is against the new threading model.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


internal/driver/glfw/loop.go
```diff
@@ -81,6 +81,10 @@ func (d *gLDriver) drawSingleFrame() {
 	refreshingCanvases = refreshingCanvases[:0]
 }
 
+var (
+	pendingSettings fyne.Settings
```
andydotxyz, line 85: I think this can stay in the scope of the `runGL` method?

andydotxyz:
Not to sure how to resolve the conflict through web UI sorry @dweymouth

app/settings.go
```diff
@@ -170,7 +168,13 @@ func (s *settings) setupTheme() {
 		variant = theme.VariantDark
 	}
 
-	s.applyTheme(effectiveTheme, variant)
+	if async.IsMainGoroutine() {
```
andydotxyz, line 171: Is it OK that setupTheme is sometimes called from non-main?

dweymouth: That was the reasoning behind adding the IsMainGoroutine check. setupTheme is called from platform-specific code and I wasn't sure which of those would be on main, and which may be from a goroutine

Jacalz: I don't think we should rely too much on `async.IsMainGoroutine()`. There is a reason that the Go developers don't expose the goroutine ID in a simple way as it allows people to more easily write bad code. Can we not use `async.EnsureMain()` for now until the migration is complete?

internal/driver/glfw/loop.go
```diff
@@ -149,23 +151,24 @@ func (d *gLDriver) runGL() {
...
 			cache.ResetThemeCaches()
-			app.ApplySettingsWithCallback(set, fyne.CurrentApp(), func(w fyne.Window) {
+			app.ApplySettingsWithCallback(pendingSettings, fyne.CurrentApp(), func(w fyne.Window) {
 				c, ok := w.Canvas().(*glCanvas)
 				if !ok {
 					return
 				}
 				c.applyThemeOutOfTreeObjects()
 				c.reloadScale()
 			})
```
Jacalz, line 170: Is there any specific reason as to why we can't do this in the change listener directly if it already runs on the main thread anyhow?

internal/driver/mobile/driver.go
```diff
@@ -258,6 +250,19 @@ func (d *driver) Run() {
...
+					cache.ResetThemeCaches()
+					intapp.ApplySettingsWithCallback(pendingSettings, fyne.CurrentApp(), func(w fyne.Window) {
+						c, ok := w.Canvas().(*canvas)
+						if !ok {
+							return
+						}
+						c.applyThemeOutOfTreeObjects()
+					})
+					pendingSettings = nil
+				}
```
Jacalz, line 265: Same as above.

Jacalz (changes requested):
Well done! I really like this new API and you are entirely correct that it is well worth it to replace the old one :)

app/settings.go
andydotxyz: All OS specific code should come in on main. If it doesn't then other things will break.

Jacalz: Then this should not do the ID check at all, right?

Jacalz: Cool. Agreed :)

test/app.go
```diff
@@ -253,6 +261,12 @@ func (s *testSettings) apply() {
 		cache.ResetThemeCaches()
 		intapp.ApplySettings(s, s.app)
 		s.app.propertyLock.Unlock()
+
+		s.listenersLock.Lock()
+		defer s.listenersLock.Unlock()
+		for _, l := range s.listeners {
+			l(s)
+		}
```
Jacalz, line 269: Hmm. This read is using listenersLock and the append is using propertyLock. That does not sound ideal. I suppose both ought to be the latter and then we don't need a new lock?

app/settings.go
```diff
@@ -116,6 +103,11 @@ func (s *settings) AddChangeListener(listener chan fyne.Settings) {
 	s.changeListeners.Store(listener, true) // the boolean is just a dummy value here.
 }
 
+func (s *settings) AddListener(listener func(fyne.Settings)) {
+	s.listeners = append(s.listeners, listener)
+}
+
+// must be called from main goroutine
```
Jacalz, line 110: I think we can remove this comment.

Jacalz (changes requested):
Thanks. Looking really good. Just a few small notes inline :)

Jacalz (approved):
Perfect. Thanks

PR #5432, opened 2025-01-19
Fix progressbar not showing label until first refresh

Jacalz:
### Description:

This fixes a whole bunch of issues that I noticed with the progressbar. It had a lot of incorrect assumptions about the renderer being set up in a specific way and relied on a manual refresh happening after the object had been created. With these changes, an initial progress bar from struct looks like it should and methods like `.Refresh()` and `.Layout()` does what is expected from them (no more and no less).

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Nice catch thanks

PR #5434, opened 2025-01-19
Adding scrolling to TextGrid and a new Append method too

andydotxyz:
Super fast monospace text is now possible.

Congratulations to anyone using Fyne for their first app by logging text ;)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

```go
package main

import (
	"strconv"
	"time"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Hello")

	tg := widget.NewTextGrid()
	tg.ShowLineNumbers = true
	tg.Scroll = container.ScrollBoth
	w.SetContent(tg)
	w.Resize(fyne.NewSize(120, 100))

	go func() {
		i := 0
		for {
			time.Sleep(time.Millisecond * 25)
			fyne.Do(func() {
				tg.Append("Another line added, this time it's..." + strconv.Itoa(i))
				i++
			})
		}
	}()
	w.ShowAndRun()
}
```

### DIFF
--- a/widget/textgrid.go
+++ b/widget/textgrid.go
@@ -475,7 +475,6 @@ type textGridContentRenderer struct {
 
 	visible  []fyne.CanvasObject
 	itemPool async.Pool[*textGridRow]
-	current  fyne.Canvas
 }
 
 func (t *textGridContentRenderer) Destroy() {
@@ -611,7 +610,6 @@ type textGridRowRenderer struct {
 	cols int
 
 	objects []fyne.CanvasObject
-	current fyne.Canvas
 }
 
 func (t *textGridRowRenderer) appendTextCell(str rune) {
@@ -678,12 +676,12 @@ func (t *textGridRowRenderer) setCellRune(str rune, pos int, style, rowStyle Tex
 		text.Text = newStr
 		text.Color = fg
 		text.TextStyle = textStyle
-		t.refresh(text)
+		text.Refresh()
 	}
 
 	if underlineStrokeWidth != underline.StrokeWidth || underlineStrokeColor != underline.StrokeColor {
 		underline.StrokeWidth, underline.StrokeColor = underlineStrokeWidth, underlineStrokeColor
-		t.refresh(underline)
+		underline.Refresh()
 	}
 
 	bg := color.Color(color.Transparent)
@@ -694,7 +692,7 @@ func (t *textGridRowRenderer) setCellRune(str rune, pos int, style, rowStyle Tex
 	}
 	if rect.FillColor != bg {
 		rect.FillColor = bg
-		t.refresh(rect)
+		rect.Refresh()
 	}
 }
 
@@ -833,11 +831,6 @@ func (t *textGridRowRenderer) MinSize() fyne.Size {
 }
 
 func (t *textGridRowRenderer) Refresh() {
-	// we may be on a new canvas, so just update it to be sure
-	if fyne.CurrentApp() != nil && fyne.CurrentApp().Driver() != nil {
-		t.current = fyne.CurrentApp().Driver().CanvasForObject(t.obj.text.text)
-	}
-
 	th := t.obj.text.text.Theme()
 	v := fyne.CurrentApp().Settings().ThemeVariant()
 	TextGridStyleWhitespace = &CustomTextGridStyle{FGColor: th.Color(theme.ColorNameDisabled, v)}
@@ -853,19 +846,4 @@ func (t *textGridRowRenderer) Objects() []fyne.CanvasObject {
 }
 
 func (t *textGridRowRenderer) Destroy() {
-}
-
-func (t *textGridRowRenderer) refresh(obj fyne.CanvasObject) {
-	if t.current == nil {
-		if fyne.CurrentApp() != nil && fyne.CurrentApp().Driver() != nil {
-			// cache canvas for this widget, so we don't look it up many times for every cell/row refresh!
-			t.current = fyne.CurrentApp().Driver().CanvasForObject(t.obj.text.text)
-		}
-
-		if t.current == nil {
-			return // not yet set up perhaps?
-		}
-	}
-
-	t.current.Refresh(obj)
 }
--- a/widget/textgrid.go
+++ b/widget/textgrid.go
@@ -385,10 +385,7 @@ type textGridRenderer struct {
 }
 
 func (t *textGridRenderer) Layout(s fyne.Size) {
-	if t.scroll.Content != nil {
-		t.scroll.Resize(s)
-	}
-	t.text.Resize(s)
+	t.Objects()[0].Resize(s)
 }
 
 func (t *textGridRenderer) MinSize() fyne.Size {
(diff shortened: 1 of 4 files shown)


widget/textgrid.go
```diff
@@ -341,17 +378,239 @@ func nextTab(column int, tabWidth int) int {
...
+func (t *textGridRenderer) Refresh() {
+	content := t.text
+	if t.text.text.Scroll != widget.ScrollNone {
+		t.scroll.Direction = t.text.text.Scroll
+	}
+	if t.text.text.Scroll == widget.ScrollNone && t.scroll.Content != nil {
+		t.scroll.Hide()
+		t.scroll.Content = nil
+		content.Resize(t.text.Size())
+		t.SetObjects([]fyne.CanvasObject{t.text})
```
Jacalz, line 411: I think it would be faster and potentially even cleaner, to do just not use a base renderer (or expose its internal object slice) to just set `t.objects[0] = t.text`.

widget/textgrid.go
```diff
@@ -341,17 +378,239 @@ func nextTab(column int, tabWidth int) int {
...
+	} else if (t.text.text.Scroll != widget.ScrollNone) && t.scroll.Content == nil {
+		t.scroll.Content = content
+		t.scroll.Show()
+
+		t.scroll.Resize(t.text.Size())
+		content.Resize(content.MinSize())
+		t.SetObjects([]fyne.CanvasObject{t.scroll})
```
Jacalz, line 418: See comment above.

widget/textgrid.go
```diff
@@ -341,17 +378,239 @@ func nextTab(column int, tabWidth int) int {
...
+
+		off += t.text.text.Size().Height
+		end = int(math.Ceil(float64(off / t.text.cellSize.Height)))
+	}
+
+	var toRemove []*textGridRow
+	for _, row := range t.visible {
+		if row.(*textGridRow).row < start || row.(*textGridRow).row > end {
+			toRemove = append(toRemove, row.(*textGridRow))
+			break
```
Jacalz, line 531: It looks like `toRemove` is only ever going to be of length one? Seems strange.

widget/textgrid.go
Jacalz: I've honestly thought about removing that helper sometimes. It rarely adds much in my opinion and many renderers don't even use it.

widget/textgrid.go
```diff
@@ -444,65 +703,65 @@ func (t *textGridRenderer) addCellsIfRequired() {
 	}
 }
 
-func (t *textGridRenderer) refreshGrid() {
-	line := 1
+func (t *textGridRowRenderer) refreshCells() {
 	x := 0
+	if t.obj.row >= len(t.obj.text.text.Rows) {
+		return // empty?
```
Jacalz, line 709: The question mark in the comment makes me worried. Is there something strange going on here? Are unsure about that the effect of returning here does?

widget/textgrid.go
```diff
@@ -444,65 +703,65 @@ func (t *textGridRenderer) addCellsIfRequired() {
...
+		x++
+	}
+	if t.obj.text.text.ShowWhitespace && i < t.cols && t.obj.row < len(t.obj.text.text.Rows)-1 {
+		t.setCellRune(textAreaNewLineSymbol, x, TextGridStyleWhitespace, rowStyle) // newline
+		i++
+		x++
+	}
+	for ; i < t.cols; i++ {
+		t.setCellRune(' ', x, TextGridStyleDefault, rowStyle) // blanks
+		x++
```
Jacalz, line 762: I'm not sure if there is much to do about this but I find that the `x++`, `i++` and so on everywhere makes the code rather hard to read.

widget/textgrid.go
```diff
@@ -341,17 +378,239 @@ func nextTab(column int, tabWidth int) int {
...
+	text   *textGridContent
+	scroll *widget.Scroll
+}
 
-	cols, rows int
+func (t *textGridRenderer) Layout(s fyne.Size) {
+	if t.scroll.Content != nil {
+		t.scroll.Resize(s)
+	}
+	t.text.Resize(s)
```
Jacalz, line 391: With the suggestion about accessing the objects slice below, this could be:
```suggestion
	t.objects[0].Resize(s)
```

widget/textgrid.go
```diff
@@ -292,15 +295,27 @@ func (t *TextGrid) SetStyleRange(startRow, startCol, endRow, endCol int, style T
...
+		scroll = widget.NewScroll(nil)
+		objs[0] = content
+	} else {
+		scroll = widget.NewScroll(content)
+		scroll.Direction = t.Scroll
+		objs[0] = scroll
+	}
+	t.scroll = scroll
+	r := &textGridRenderer{text: content, scroll: scroll}
+	r.SetObjects(objs)
```
Jacalz, line 317: This can be cleaned up a bit:
```suggestion
	var inScroller fyne.CanvasObject
	content := newTextGridContent(t)
	objs := []fyne.CanvasObject{content}
	if t.Scroll != widget.ScrollNone {
		inScroller = content
		scroll.Direction = t.Scroll
		objs[0] = scroll
	}
	t.scroll = widget.NewScroll(inScroller)
	r := &textGridRenderer{text: content, scroll: t.scroll}
	r.SetObjects(objs)
```

Jacalz (changes requested):
Thanks. This is a really nice change. I left a few suggestions and a comment about a  potential bug I found. Would it not be nice to also update fyne_demo so this can be showcased in some way?

andydotxyz:
> Would it not be nice to also update fyne_demo so this can be showcased in some way?

I didn't bother with the fyne_demo because, as you saw in another PR, I'm hoping to replace the tutorials with our new code style assisting demo.

widget/textgrid.go
andydotxyz: All I did was un-indent the existing code here by working on a single row instead of them all

widget/textgrid.go
andydotxyz: This feels subjective - can it be definitive that creating "inScroller" to avoid an else is better?

widget/textgrid.go
andydotxyz: Good catch

widget/textgrid.go
Jacalz: Ah, sorry. That was not apparent from the diff but I switched to side-by-side and that's more clear

andydotxyz:
I've addressed what I think are the objective fixes. The other items are commenting on pre-existing algorithms or are suggestions about how the code could be - but I think we should stick to what needs to be better for a clear reason TBH.

widget/textgrid.go
Jacalz: Hmm, yeah. Might be subjective. I just try to avoid branches as much as possible because it is less code indented to the right. Do as you wish :)

andydotxyz: I see that but it is not removing an indent (as that level deep was already there).
I was a little confused because the nil content of Scroller becomes implicit which could look like a bug.

Jacalz:
Sorry if it wasn't clear but I basically tried to give suggestions and so on as well. I think your points are fair :)

Jacalz (approved):

PR #5442, opened 2025-01-20
Implement dead window cleanup without niling out

Jacalz:
### Description:

Reduces the chance of a potential crash for someone not expecting a nil window and is honestly a bit cleaner code wise.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Thanks. Those nils were cropping up all over the place.

Jacalz:
Yeah, I noticed that as well which I do find a bit strange actually. The idea was to clean up directly after the loop and since this is single threaded now there are no races there but I suppose I missed some functions being called in between 😅

PR #5443, opened 2025-01-20
Embed theme contents instead of bundle

Jacalz:
### Description:

This has code modernisation has various benefits. A shallow git clone of the project no longer includes all icons twice (in source and .go file), opening the huge files with bundled stuff no longer cause editors to grind to a halt but most importantly, just editing the file makes the data in the binary change without needing to do anything at all. It is all automatic and cached by the compiler. There is also less text for the compiler to parse so there might be faster compile times (assuming it can just copy the file directly into the binary) but that's just speculation.

Building `fyne_demo` using `go build -ldflags="-s -w"` yields:
- Before: 24224176
- After: 24255312

Binaries are ever so slightly larger but we are just talking ~30kb so nothing much to write home about.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
I don't know if it's just me but the workflows did seem to be done quite quickly. I suppose the double file download for that shallow git clone already is paying off? If we can do the same for the font, then I'm quite enthusiastic.

andydotxyz (approved):

andydotxyz:
Merging to avoid a re-base due to the branch settings change

PR #5445, opened 2025-01-21
Better ci infrastructure

Jacalz:
### Description:

Various changes to make it easier to maintain (or rather have it maintain itself) and also some cleanups plus simplifications.

- Moved the licence check workflow run into the static analysis test.
- Stop running benchmarks on CI, no one looks at this output and it doesn't fail or notify us if performance goes bad so there is really very little gain from the extra time it takes. Let's add some specific benchmark workflow in the future instead if necessary.
- Use latest Go version automatically to avoid having to change required versions each time.
- Switch to `actions/setup-go` which is better maintained and has built in dependency caching infrastructure (a fresh build after `go.mod` changes is likely slower but we avoid downloading dependencies for all else).
- Load version from `go.mod` automatically using workaround until https://github.com/actions/setup-go/issues/450 is implemented.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Cool

Jacalz:
Thanks. Feel free to update required workflows and merge when you are ready :)

andydotxyz:
Perfect thanks, I think that is the settings updated

Jacalz:
Great. Just an approval needed before it can be merged :)

PR #5448, opened 2025-01-21
Creating a version of fyne.Do that does not wait

andydotxyz:
Naming and implementation open for discussion

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/mobile/canvas.go
+++ b/internal/driver/mobile/canvas.go
@@ -407,7 +407,7 @@ func (c *canvas) waitForDoubleTap(co fyne.CanvasObject, ev *fyne.PointEvent, tap
 		c.touchCancelFunc = nil
 		c.touchLastTapped = nil
 		c.touchCancelLock.Unlock()
-	})
+	}, true)
 }
 
 func (c *canvas) windowHeadIsDisplacing() bool {
--- a/driver.go
+++ b/driver.go
@@ -47,11 +47,11 @@ type Driver interface {
 	// Since: 2.5
 	SetDisableScreenBlanking(bool)
 
-	// DoFromGoroutine provides a way to queue a function that is running on a goroutine back to
-	// the central thread for Fyne updates.
+	// DoFromGoroutine provides a way to queue a function `fn` that is running on a goroutine back to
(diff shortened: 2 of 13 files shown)


Jacalz:
I will review this further tomorrow but this does look great. The name is want I was going to suggest actually :)

internal/driver/mobile/canvas.go
```diff
@@ -407,7 +407,7 @@ func (c *canvas) waitForDoubleTap(co fyne.CanvasObject, ev *fyne.PointEvent, tap
 		c.touchCancelFunc = nil
 		c.touchLastTapped = nil
 		c.touchCancelLock.Unlock()
-	})
+	}, true)
```
dweymouth, line 410: should this function call just be converted to `fyne.DoAsync`?

andydotxyz: I'm not sure - that is going out of the driver to a utility function which does thread checks then back into the driver code - feels inefficient. Either way wouldn't it be the Sync/wait/fyne.Do version?

andydotxyz:
I am wondering about naming - and defaults..

Should we change so fyne.Do, fyne.DoAsync = fyne.Wait, fyne.Do?

Although our driver code will more often than not want to wait (though even the async version maintains order of operations queued...) but for app developers won't the async version be more common? i.e. "I updated a value, now go refresh"...

Jacalz:
Hmm. I think you are right. That does seem like a good change, though I hink I'd prefer `fyne.WaitFor()` over `fyne.Wait()` as it sounds more like you are waiting on what you passed in to be finished.

dweymouth:
I don't like fyne.Wait on its own, but fyne.DoWait could be OK. I think having the "Do" verb in common for both names is important for clarity.

andydotxyz:
What about "DoAndWait" or "DoBlock"?

andydotxyz:
It seems like DoAndWait may satisfy all the comments above but I will sleep on it :).

Jacalz:
I'm not 100% sold on `DoAndWait` but it isn't terrible so I don't wish to hold this up :)

PR #5450, opened 2025-01-23
Clean up background processes when changing tutorial in fyne_demo

Jacalz:
### Description:

All animations and tickers were recreated on loading a new page. Animations resulted in just background CPU usage but the tickers can not (until go.mod defaults to Go 1.23) be GC'd before stop so they would tick in the background forever while basically leaking CPU and memory.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
I should note that this makes life a lot easier when testing https://github.com/fyne-io/fyne/pull/5422. That's the motivation behind this so there is 0% cpu more often :)

andydotxyz (approved):
It's not amazing having those string IDs re-declared but it seems ok for this fix. I'm still opting to get the revised demo system in for 2.6 which will be able to refactor this further.

PR #5459, opened 2025-01-24
Clean up cleanup function queue in fyne_demo

Jacalz:
### Description:

Right after having merged and read https://github.com/fyne-io/fyne/pull/5450#pullrequestreview-2573571888, I realised that the checks and page input string just was a leftover from before I cleared the queue of functions after having executed them. This means that I can just remove that part and use `func()` directly.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
cleaner :)

PR #5465, opened 2025-01-25
Improve fyne.Do error reporting

andydotxyz:
### Checklist:

- [ ] Tests included. <- should stop seeing the reports in tests
- [x] Tests all pass.

### DIFF
--- a/internal/driver/common/canvas.go
+++ b/internal/driver/common/canvas.go
@@ -314,10 +314,8 @@ func (c *Canvas) Painter() gl.Painter {
 
 // Refresh refreshes a canvas object.
 func (c *Canvas) Refresh(obj fyne.CanvasObject) {
-	async.EnsureMain(func() {
-		c.refreshQueue.In(obj)
-		c.SetDirty()
-	})
+        c.refreshQueue.In(obj)
+	async.EnsureMain(c.SetDirty)
 }
 
 // RemoveShortcut removes a shortcut from the canvas.
--- a/internal/driver/common/canvas.go
+++ b/internal/driver/common/canvas.go
@@ -314,8 +314,10 @@ func (c *Canvas) Painter() gl.Painter {
 
 // Refresh refreshes a canvas object.
 func (c *Canvas) Refresh(obj fyne.CanvasObject) {
-	c.refreshQueue.In(obj)
-	c.SetDirty()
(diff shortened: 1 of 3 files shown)


internal/driver/common/canvas.go
```diff
@@ -314,8 +314,10 @@ func (c *Canvas) Painter() gl.Painter {
 
 // Refresh refreshes a canvas object.
 func (c *Canvas) Refresh(obj fyne.CanvasObject) {
-	c.refreshQueue.In(obj)
-	c.SetDirty()
+	async.EnsureMain(func() {
+		c.refreshQueue.In(obj)
+		c.SetDirty()
+	})
```
Jacalz, line 320: Suggestion but not request: It seems unnecessary to do move the refreshQueue.In() call to another thread as it is an async channel anyway. Maybe this is cleaner?
```suggestion
        c.refreshQueue.In(obj)
	async.EnsureMain(c.SetDirty)
```

Jacalz (changes requested):
Thanks for this but it still doesn't seem to get the right line. Adding a `go` statement before the first `w.Resize()` call in fyne_demo gives me this:
```
2025/01/26 20:49:04 *** Error in Fyne call thread, this should have been called in fyne.Do[AndWait] ***
2025/01/26 20:49:04   From: /var/home/jacob/git/fyne/internal/driver/glfw/window.go:242
```

internal/driver/common/canvas.go
Jacalz: Sorry that the indenting was off. Apparently GitHub doesn't handle that very well. Looked good on my computer

andydotxyz:
> Thanks for this but it still doesn't seem to get the right line. Adding a `go` statement before the first `w.Resize()` call in fyne_demo gives me this:
> 
> ```
> 2025/01/26 20:49:04 *** Error in Fyne call thread, this should have been called in fyne.Do[AndWait] ***
> 2025/01/26 20:49:04   From: /var/home/jacob/git/fyne/internal/driver/glfw/window.go:242
> ```

When testing inside the fyne codebase it is not possible to deduce there the line between library and app are.
Try with one of your apps or anything in a different repo...

Jacalz (approved):
Needs a merge for test fixes

PR #5468, opened 2025-01-26
Improving inner window theme

andydotxyz:
Also added some dummy handlers to the demo to show difference of disabled / default.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

https://github.com/user-attachments/assets/2151d672-d1a1-4d01-bf89-8938178b7881

### DIFF
--- a/cmd/fyne_demo/tutorials/container.go
+++ b/cmd/fyne_demo/tutorials/container.go
@@ -3,6 +3,7 @@ package tutorials
 import (
 	"fmt"
 	"image/color"
+	"log"
 	"strconv"
 
 	"fyne.io/fyne/v2"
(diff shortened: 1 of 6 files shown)


Jacalz (changes requested):
Thanks. This looks so much better. Can you please make the contrast for hovering the buttons higher because I find it on the edge of being hard to make out, especially if you move the mouse quickly.

The close button is also "very read" in an almost distracting way and pops out a bunch when hovered. I wonder if we should just make it match hover for the other buttons?

andydotxyz:
> Thanks. This looks so much better. Can you please make the contrast for hovering the buttons higher because I find it on the edge of being hard to make out, especially if you move the mouse quickly.
> 
> The close button is also "very read" in an almost distracting way and pops out a bunch when hovered. I wonder if we should just make it match hover for the other buttons?

What you're asking for really requires a whole new set of colours for hover and therefore likely others like focus etc. 
I think that is way beyond what we need so instead I have removed the custom colours completely.

Revised:


https://github.com/user-attachments/assets/803c9657-600a-4b84-9d83-76cb03805385

Jacalz (approved):
Cool. That looks better contrast wise :+1:

andydotxyz:
Failures in pref test only - that's on another PR for discussion

PR #5470, opened 2025-01-26
Proposed pref test fix

andydotxyz:
The new async way of working with prefs is not OK with the test system not having complete safety of thread handling.
So take a round trip to ensure it.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (changes requested):
If you remove the wait group and use `fyne.DoAndWait()` you should get the same result. However, in this case I think we should rename the test because with this change i no longer actually tests any races but just that we get the right object (which also is a bit strange given that we add no other objects).

andydotxyz:
> If you remove the wait group and use fyne.DoAndWait() you should get the same result. However, in this case I think we should rename the test because with this change it no longer actually tests any races

Which is why I didn't do that approach - instead spawning lots of "async" fyne.Do calls as a rough approximation of what it was doing before...

Jacalz (approved):

PR #5472, opened 2025-01-27
Only refresh scroll chrome when scroller is resized

### LINKED ISSUE #5456: Tree and Table widgets refresh full content on any scroll or resize
### Checklist

### Describe the bug

The Tree and Table widgets currently refresh the entire contents whenever they are scrolled or resized, regardless of whether new elements appear in the viewport or not. This is inefficient, though with the 2.5.x driver implementation it was not severe enough to be a bottleneck at least on modern systems. But with #5422, where we are currently processing scroll / mouse drag events from the OS as fast as they arrive, it causes severe lagging when scrolling or resizing a Tree or Table (visible in fyne demo). This does not happen with List, which has been optimized to only refresh the necessary rows.

### How to reproduce

Run fyne_demo on the #5422 branch. Scroll the tree or table, on a Macbook with the trackpad (which seems to deliver scroll events much faster than 60 fps), or resize the split container containing the Tree, also with the Macbook trackpad.

### Screenshots

_No response_

### Example code

fyne_demo

### Fyne version

develop

### Go compiler version

n/a

MacOS

### Additional Information

_No response_

andydotxyz:
Progresses #5456

### Checklist:

- [ ] Tests included.<- fewer refreshes, same test results :)
- [x] Tests all pass.

### DIFF
--- a/internal/widget/scroller.go
+++ b/internal/widget/scroller.go
@@ -222,7 +222,7 @@ func (a *scrollBarArea) CreateRenderer() fyne.WidgetRenderer {
 
 func (a *scrollBarArea) MouseIn(*desktop.MouseEvent) {
 	a.isMouseIn = true
-	a.scroll.Refresh()
+	a.scroll.refreshBars()
 }
 
 func (a *scrollBarArea) MouseMoved(*desktop.MouseEvent) {
@@ -234,7 +234,7 @@ func (a *scrollBarArea) MouseOut() {
 		return
 	}
 
-	a.scroll.Refresh()
+	a.scroll.refreshBars()
 }
 
 func (a *scrollBarArea) moveBar(offset float32, barSize fyne.Size) {
@@ -432,12 +432,13 @@ func (s *Scroll) CreateRenderer() fyne.WidgetRenderer {
 // ScrollToBottom will scroll content to container bottom - to show latest info which end user just added
 func (s *Scroll) ScrollToBottom() {
 	s.scrollBy(0, -1*(s.Content.MinSize().Height-s.Size().Height-s.Offset.Y))
-	s.Refresh()
+	s.refreshBars()
 }
 
 // ScrollToTop will scroll content to container top
 func (s *Scroll) ScrollToTop() {
 	s.scrollBy(0, -s.Offset.Y)
+	s.refreshBars()
 }
 
 // DragEnd will stop scrolling on mobile has stopped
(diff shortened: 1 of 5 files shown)


dweymouth:
This is a good change (though haven't tested it yet) but just noting that it doesn't fix #5456 - the root cause there is the widgets themselves. e.g. here in Tree, it refreshes the entire content in the OffsetUpdated callback from the scroller: https://github.com/fyne-io/fyne/blob/develop/widget/tree.go#L491 (and similarly in table)

widget/widget.go
```diff
@@ -112,16 +111,14 @@ func (w *BaseWidget) Hide() {
...
-		impl := w.super()
-		if impl == nil {
-			return
-		}
+	impl := w.super()
+	if impl == nil {
+		return
+	}
 
-		w.themeCache = nil
```
Jacalz, line 121: This change does not look like it belongs here?

Jacalz (changes requested):
Thanks. Looking really good. Just a small change I think you accidentally left in there (or due to being based on the wrong commit compared to develop).

widget/widget.go
andydotxyz: The commit comment "Widget.Refresh running thread handling code was not great for potential loop scenarios" explains it I think...

andydotxyz: The added bonus being that you the long log of thread warnings from the test driver that were being complained about is gone.

Jacalz: Ah, sorry

Jacalz (approved):

internal/widget/scroller.go
```diff
@@ -489,7 +493,19 @@ func (s *Scroll) Resize(sz fyne.Size) {
...
 
 	s.Base.Resize(sz)
-	s.Refresh()
+	s.refreshBars()
+}
+
+// SetOffset will update the location of the content of this scroll container.
+//
+// Since: 2.6
+func (s *Scroll) SetOffset(p fyne.Position) {
```
dweymouth, line 502: Should this perhaps be named `ScrollToOffset` to match the naming of `ScrollToBottom` etc, and also the `ScrollToOffset` that exists in List and GridWrap?

andydotxyz: Thanks, good eye.

Jacalz (approved):

dweymouth (approved):

PR #5477, opened 2025-01-29
Fixing Scroller.ScrollToTop, alias to a new fyne.ScrollDirection type and align all collections on ScrollToOffset

andydotxyz:
### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/scroll.go
+++ b/scroll.go
@@ -1,6 +1,8 @@
 package fyne
 
 // ScrollDirection represents the directions in which a scrollable container or widget can scroll its child content.
+//
+// Since: 2.6
 type ScrollDirection int
 
 // Constants for valid values of ScrollDirection used in containers and widgets.
@@ -12,7 +14,5 @@ const (
 	// ScrollVerticalOnly specifies the scrolling should only happen top to bottom.
 	ScrollVerticalOnly
 	// ScrollNone turns off scrolling for this container.
-	//
-	// Since: 2.6
 	ScrollNone
 )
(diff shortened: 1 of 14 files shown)


scroll.go
```diff
@@ -0,0 +1,18 @@
...
+const (
+	// ScrollBoth supports horizontal and vertical scrolling.
+	ScrollBoth ScrollDirection = iota
+	// ScrollHorizontalOnly specifies the scrolling should only happen left to right.
+	ScrollHorizontalOnly
+	// ScrollVerticalOnly specifies the scrolling should only happen top to bottom.
+	ScrollVerticalOnly
+	// ScrollNone turns off scrolling for this container.
+	//
+	// Since: 2.6
```
Jacalz, line 16: All of these above are new within the `fyne` package scope. I understand why this was added here but it is slightly confusing 😅

Jacalz (commented):
Perfect. Well done. I just added a question inline :)

scroll.go
andydotxyz: I get your point but what would be more correct? I can switch to whatever you suggest, but I'm not sure what is right.

Jacalz: I'd suggest adding the line to the type only as it is new in this release but I don't know really.

andydotxyz: That works, done thanks :)

Jacalz (approved):
Splendid 👍

PR #5483, opened 2025-01-31
Rework bindings to avoid generator script entirely

### LINKED ISSUE #5378: data/binding: Use generics instead of generator script

Jacalz:
### Description:

Working on removing the last bits of code duplication and generator scripts within data bindings. With this landing, the code cleanup is complete, I think. Conversions are now handled by generic implementations and the other generic implementations in `generic.go` have been moved into their own files again (I also removed a generic interface that accidentally existed twice using two different names).

Fixes #5378 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
FYI: The `convert.go` related files are the ones with the only new code. All other binding files are basically just code moved out from the `generic.go` file except for one duplicated interface that was removed.

andydotxyz (approved):
Thanks so much for this cleaner code and for getting our coverage back up :)

PR #5484, opened 2025-01-31
Data binding file structure rework

Jacalz:
### Description:

Just a small cleanup renaming files so they don't start with "bind" as that's given from the package name, moves some stuff into files that no longer are generated and just generally cleans up the amount of files in there.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Nice thanks for this

PR #5485, opened 2025-01-31
Hide empty spaces where validation is enabled but nothing to show

andydotxyz:
This includes a slight change in behaviour that errors would be shown after initial input is completed.
It is helpful to avoid forms jumping around when focus changes.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


dweymouth:
Does this cause "jumps" in the layout positioning when a validation error is shown then? If so, I wonder if it should be somehow themeable because some may prefer the compact layout while others prefer no layout jumps?

andydotxyz:
Yes if and only if an error appears with no hint. If a developer wants to prop it open they could use " " as a hint.

dweymouth:
> Yes if and only if an error appears with no hint. If a developer wants to prop it open they could use " " as a hint.

Good point that " " can prop it open if desired

andydotxyz:
Does that mean we're good @dweymouth ?

dweymouth (approved):

PR #5487, opened 2025-01-31
Fix deadlock in untyped binding

andydotxyz:
Unlock before triggering callbacks for untyped bind

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


dweymouth (approved):

andydotxyz:
Sorry @dweymouth I missed the early return and missed that tests had not run locally...

PR #5490, opened 2025-02-02
Avoid allocating for each binding trigger

Jacalz:
### Description:

This reworks data listeners for bindings to be a bit smarter. Instead of using a map (with unnecessary bool parameter) to store the listener, we just use a regular slice. We iterate over this slice in the queued operation without needing to append to a new slice first. Delete is a bit more expensive (and more code because with don't use Go 1.21 yet and can't use the `slices` package) but add is the one that is more common and trigger is the most common so it makes sense to optimise for the common path. A second lock is needed to not deadlock tests for some reason and it also avoids contention between value updates.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/binding.go
+++ b/data/binding/binding.go
@@ -9,7 +9,6 @@ import (
 	"sync"
 
 	"fyne.io/fyne/v2"
-	"fyne.io/fyne/v2/internal/async"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


data/binding/binding.go
```diff
@@ -58,31 +57,43 @@ func (l *listener) DataChanged() {
...
-	b.listeners.Range(func(listener DataListener, _ bool) bool {
-		listeners = append(listeners, listener)
-		return true
-	})
-
 	queueItem(func() {
-		for _, listen := range listeners {
+		b.listenerLock.Lock()
+		defer b.listenerLock.Unlock()
+		for _, listen := range b.listeners {
```
dweymouth, line 96: Worth noting that registering a new listener (or unsubscribing) on the same binding in the DataChanged callback will now deadlock. The two slice copy-and-swap-under-lock technique could be used here if we want to work around this limitation.

Jacalz: Are you sure it will deadlock? The queued operations run after each other so I can't quite see why it would.

Jacalz: I do however have another potential idea for avoiding the lock entirely. How about queuing all operations on the slice onto the main channel? Triggers are the ones that are the most common of course with a fee adds and very few removes.

dweymouth: `listenerLock` is held during the invocation of the listeners, and it's held in Add/Remove listener. So it would deadlock if one of the listeners were to try to unregister itself in its callback, for example. I think you're right that queueing all slice operations to main (effectively making them async/promise-like) would avoid locking entirely. I also think duplicating the listeners slice under lock before invoking the callbacks without the lock would have a negligible perf impact, as long as we don't allocate a new slice for it but re-use one.

Jacalz: I kind of feel like having two slices wouldn't be worth it for each and every bind able item someone might want to use. That doesn't scale well. Either allocating a slice copy on each trigger (which creates lots of unnecessary garbage if items update often) or queuing to main when also adding or removing listeners are the best options that I find. The latter might have scalability issues but seems better from a performance standpoint given that all triggered listeners already need to be on main (lower frequency of add/remove of listeners than triggers)?

Jacalz: I tried the latter. Seems to work fine and be quite a clean implementation

dweymouth (approved):

PR #5491, opened 2025-02-02
Add a storage helper to compare two URIs

Jacalz:
### Description:

This introduces a `storage.EqualURI` (matching `bytes.Equal` in naming kind of) helper that allows to easily compare two URI types. Having this in the repository directly also makes it possible to compare struct fields directly instead of always allocating a new string for each comparison (we still have to as a fallback though).

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


storage/repository/uri.go
```diff
@@ -10,6 +10,23 @@ import (
 	"fyne.io/fyne/v2"
 )
 
+// EqualURI returns true if the two URIs are equal.
+//
+// Since: 2.6
+func EqualURI(t1, t2 fyne.URI) bool {
```
dweymouth, line 16: Do we need to expose the same public function in two packages? Could we just move the implementation to the root `storage` package and just have that one?

storage/repository/uri.go
```diff
@@ -10,6 +10,23 @@ import (
...
+	if t1 == nil || t2 == nil {
+		return t1 == t2
+	}
+
+	u1, ok1 := t1.(*uri)
```
andydotxyz, line 21: URI is an interface, we cannot assert that it is always backed by our internal type.

andydotxyz (dismissed):
It's an interface and we can't assume the only implementations are our storage one.

storage/repository/uri.go
Jacalz: Indeed. I realised that I forgot to add tests for it but the reason for checking it he underlying type is that we can do a fast comparison directly without allocating the whole string. If you look at the comparison at the bottom of the function, you'll see that it doesn't compare the type asserted types but the interfaces.

storage/repository/uri.go
Jacalz: We don't have access to the internal uri type in the storage package to do a fast path for comparison.

storage/repository/uri.go
andydotxyz: Apologies you're quite right

storage/repository/uri.go
dweymouth: Could this move to `internal/storage` maybe without import loops (not sure)? But I just feel like having two public APIs that are identical isn't ideal. It's one thing in the case of e.g. fyne.Do, where one is a static helper for accessing the implementation from the loaded driver, but these are both just static functions.

Jacalz: It's the same with almost all other APIs in the repository package

dweymouth: What if we make it a method on the URI interface, e.g. `uri1.Equal(uri2)`? Then we can have the implementation on the implementing type here in the repository package (which maybe just calls this function, but we make it private), and we can avoid having repeat public API definitions

Jacalz: Changing the interface would be a breaking change for anyone implementing their own URI

Jacalz: We could theoretically add the method and look for that in an interface assertion but using the method also wouldn't work in the case for bindings where we pass it in as a function parameter (general form `(T, T) bool` where T is any type).

Jacalz: I don't see any solution around this, sorry. The repository package likely should have been internal from the start. I think people generally treat it as such.

dweymouth: Unfortunate, but OK. I guess we can consider moving the repository package to internal in Fyne 3.0

Jacalz: Yup. We might consider not adding a since: line to the repository but I don't know.

storage/repository/uri.go
Jacalz: No worries. I switched around the logic and added an extra comment. I hope it is clearer now :)

storage/repository/uri.go
dweymouth: nah, I think every public API should probably have them for consistency, and it is technically a public API (as is any exported symbol within a non-internal package)

Jacalz: Good point

dweymouth (approved):

PR #5495, opened 2025-02-03
Make GridWrap more efficient by avoiding unnecessary content refreshes when resizing, scrolling

dweymouth:
### Description:
Avoids refreshing content when not necessary during GridWrap.Resize and scrolling. Applies the same optimization from List to only refresh newly visible items unless doing the full Refresh call.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Nice optimisation for the horizontal resize :)

PR #5496, opened 2025-02-03
Make table scrolling and resizing more efficient

dweymouth:
### Description:

Only refresh newly visible cells when scrolling and resizing the table

Progresses #5456 (for Table)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/table.go
+++ b/widget/table.go
@@ -9,15 +9,19 @@ import (
 	"fyne.io/fyne/v2/driver/desktop"
 	"fyne.io/fyne/v2/driver/mobile"
 	"fyne.io/fyne/v2/internal/async"
-	"fyne.io/fyne/v2/internal/cache"
 	"fyne.io/fyne/v2/internal/widget"
 	"fyne.io/fyne/v2/theme"
 )
 
 const noCellMatch = math.MaxInt
 
-// allTableCellsID represents all table cells when refreshing requested cells
-var allTableCellsID = TableCellID{-1, -1}
+var (
+	// allTableCellsID represents all table cells when refreshing requested cells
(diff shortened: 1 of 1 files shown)


widget/table.go
```diff
@@ -1168,7 +1169,17 @@ func (c *tableCells) CreateRenderer() fyne.WidgetRenderer {
 
 func (c *tableCells) Resize(s fyne.Size) {
 	c.BaseWidget.Resize(s)
-	c.Refresh() // trigger a redraw
+	c.RefreshForID(onlyNewTableCellsID) // trigger a redraw
+}
+
+func (c *tableCells) RefreshForID(id TableCellID) {
```
andydotxyz, line 1175: This should probably be left unexported

andydotxyz (changes requested):
Thanks for doing this, nice solution. 
Just a drive-by comment as I've not managed to test the code yet - but let's unexport methods by default

widget/table.go
```diff
@@ -1388,6 +1384,31 @@ func (r *tableCellsRenderer) refreshForID(toDraw TableCellID) {
...
+		fyne.LogError("Missing UpdateCell callback required for Table", nil)
+		return
+	}
+
+	if toDraw == onlyNewTableCellsID {
+		for id, cell := range visible {
+			if _, ok := wasVisible[id]; !ok {
+				updateCell(id, cell)
+			}
+		}
```
Jacalz, line 1399: I think it would be cleaner with an early return here to avoid two indented code blocks.

Jacalz (changes requested):
Thanks. +1 on Andy's review and just a tip about early return

PR #5498, opened 2025-02-04
Support canvas.Image to parse gif images.

HuXin0817:
### Description:

Support canvas.Image to parse gif images. Like `canvas.NewImageFromFile("xx.gif")`

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/canvas/image.go
+++ b/canvas/image.go
@@ -4,15 +4,18 @@ import (
 	"bytes"
 	"errors"
 	"image"
+	gif2 "image/gif"
 	_ "image/jpeg" // avoid users having to import when using image widget
 	_ "image/png"  // avoid the same for PNG images
 	"io"
 	"os"
 	"path/filepath"
 	"sync"
+	"time"
 
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/internal/cache"
+	"fyne.io/fyne/v2/internal/gif"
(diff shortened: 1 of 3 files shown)


andydotxyz:
I think I agree. If this was just decoding the first frame it would fit in fine (that would just be an "import _ image/gif" I think.

But automatically starting any gif loaded and running it until that specific asset is hidden, I'm not sure about - apps typically want more control and that is what the gif widget in Fyne-x provides.

dweymouth:
I agree as well, though I wouldn't be opposed to cleaning up and moving the gif widget from fyne-x into fyne core (widget.AnimatedGIF)?

Jacalz:
I kind of feel like GIF as a format doesn't quite jive well with scaling such? We are focused on vector graphics and getting HIDPI working well but GIF seems to be a legacy format that doesn't fit in there. Do we really want it in the standard set of widgets?

andydotxyz:
> I kind of feel like GIF as a format doesn't quite jive well with scaling such? We are focused on vector graphics and getting HIDPI working well but GIF seems to be a legacy format that doesn't fit in there. Do we really want it in the standard set of widgets?

A fair question.

I am going to close this as the functionality exists in fyne-x with a control API for developers.

Thanks so much for your work on this one @HuXin0817 but as discussed above the gif format doesn't map ideally to a single image in the traditional sense.

Issue #1166, opened 2020-07-04
NavigationContainer Widget


lherman-cs:
### Is your feature request related to a problem? Please describe:
I want to organize my application to "pages" similar to how web pages usually work so that my application can be scalable.

### Is it possible to construct a solution with the existing API?
No, the closest solution I could get was by using some flags to mark some widgets to be hidden/shown. This approach is hard to scale, it becomes more annoying when I want to navigate from one page to another.

### Describe the solution you'd like to see:
I've created my own router which you can see the implementation here, https://github.com/pion/obs-wormhole/blob/master/router.go and an example how to use the API, https://github.com/pion/obs-wormhole/blob/master/main.go. The concept is very similar to react-router.

I'm more than happy to create my implementation as a library or as a part of fyne!! But, before I do that, I'd like to know if there's already an existing router widget (I don't want to reinvent the wheel) or an alternative approach that basically can do what I want, separate pages to make applications more scalable.

By the way, this project is pretty awesome! Great job authors and maintainers!

andydotxyz:
Sounds really cool. Could such a widget be called “NavigationContainer’? If so I’d imagined something that provides user navigation as well.
I think your API is overall quite “React” instead of “Go/Fyne” but the concept is something I could see is including

lherman-cs:
`NavigationContainer` sounds good to me. Yes, it'll provide navigation as well, you can see an example here, https://github.com/pion/obs-wormhole/blob/master/home.go#L37-L39.

Yup, I was inspired how they declare routing declaratively. I'm down with changing the API to be more "Go/Fyne", but I'm not sure exactly what you mean or how to make it more "Go".

andydotxyz:
I can't relly see the navigation in that - do you have a screenshot so we can more easily tell how it would work?
If you'd like to consider a core widget using this perhaps propose here, or in a wiki page how the API might be adapted to feel more fitting with our other widgets?

lherman-cs:
# Design
This proposal is about adding a new widget called `NavigationContainer`. The widget will be used as a page view and navigation. The design is inspired by declarative routing design from react-router and has been modified to be suitable for "Go/Fyne".

## New interfaces

These interfaces should go under https://pkg.go.dev/fyne.io/fyne?tab=doc

### Navigator
This interface will be used to hide implementation details and meant to be passed around between pages so that they can use the interface to navigate.

```go
type Navigator interface {
	Push(path string, ctx interface{}) error
	Pop() error
	Reset()
[...]

PR #5500, opened 2025-02-04
Initial implementation of navigation container

sdassow:
### Description:

This add a basic navigation container type with back button and title. Sponsored by Fyne Labs.

How it looks like:

https://github.com/user-attachments/assets/e8e70946-1dae-4d9d-824e-6905717ef5d0

Example code:
```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.NewWithID("fyne-navigation")

	var nav *container.Navigation
	var view1, view2, view3 *fyne.Container

	view1 = container.NewVBox(
		widget.NewLabel("Foo"),
		widget.NewButton("Push!", func() {
			nav.PushWithTitle(view2, "Bar Title")
		}),
	)

	view2 = container.NewVBox(
		widget.NewLabel("Bar"),
		widget.NewButton("More!", func() {
			nav.PushWithTitle(view3, "Baz Title")
		}),
	)

	view3 = container.NewVBox(
		widget.NewLabel("Baz"),
	)

	nav = container.NewNavigation("NavNavNav", view1)

	w.SetContent(nav)
	w.Resize(fyne.NewSize(400, 400))
	w.ShowAndRun()
}
```

### Checklist:

- [X] Tests included.
- [X] Tests all pass.

#### Where applicable:

### DIFF
--- a/container/navigation.go
+++ b/container/navigation.go
@@ -1,0 +1,118 @@
+package container
+
+import (
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/theme"
+	"fyne.io/fyne/v2/widget"
+)
+
+// Navigation container is used to provide your application with a control bar and an area for content objects.
+// Objects can be any CanvasObject, and only the most recent one will be visible.
+//
+// Since: 2.6
+type Navigation struct {
+	widget.BaseWidget
+	control *fyne.Container
+	stack   *fyne.Container
+	button  *widget.Button
+	label   *widget.Label
+	title   string
+	titles  []string
+}
+
+// NewNavigation creates a new navigation container allowing to pass in one or more objects.
+//
+// Since: 2.6
+func NewNavigation(s string, objs ...fyne.CanvasObject) *Navigation {
+	button := widget.NewButtonWithIcon("", theme.NavigateBackIcon(), nil)
+	label := widget.NewLabelWithStyle(s, fyne.TextAlignCenter, fyne.TextStyle{Bold: true})
+	titles := make([]string, len(objs))
+	for n := 0; n < len(objs); n++ {
+		titles[n] = s
+	}
+	nav := &Navigation{
+		button:  button,
+		control: NewStack(NewHBox(button), label),
+		stack:   NewStack(objs...),
+		label:   label,
+		title:   s,
+		titles:  titles,
+	}
+	nav.ExtendBaseWidget(nav)
+	nav.button.OnTapped = func() {
+		nav.Pop()
+	}
+	if len(objs) <= 1 {
+		nav.button.Disable()
+	}
+
+	return nav
+}
+
+// Push puts the given object on top of the navigation stack and hides the object below.
+//
+// Since: 2.6
+func (nav *Navigation) Push(obj fyne.CanvasObject) {
+	nav.PushWithTitle(obj, nav.titles[len(nav.titles)-1])
+}
+
+// PushWithTitle puts the given CanvasObject on top, hides the object below, and uses the given title as label text.
+//
+// Since: 2.6
+func (nav *Navigation) PushWithTitle(obj fyne.CanvasObject, s string) {
+	objs := nav.stack.Objects
+	if len(objs) > 0 {
+		objs[len(objs)-1].Hide()
+	}
+	nav.stack.Objects = append(objs, obj)
+	if len(nav.stack.Objects) > 1 {
+		nav.button.Enable()
+	}
+
+	nav.titles = append(nav.titles, s)
+	nav.label.SetText(s)
+}
+
+// Pop return the top level CanvasObject, adjusts the title accordingly, and disabled the back button
+// when no more objects are left to go back to.
+//
+// Since: 2.6
+func (nav *Navigation) Pop() fyne.CanvasObject {
+	objs := nav.stack.Objects
+	if len(objs) == 0 {
+		return nil
+	}
+	obj := objs[len(objs)-1]
+	objs = objs[:len(objs)-1]
+	if len(objs) > 0 {
+		objs[len(objs)-1].Show()
+	}
+	nav.stack.Objects = objs
+	if len(nav.stack.Objects) <= 1 {
+		nav.button.Disable()
+	}
+
+	if len(nav.titles) > 0 {
+		nav.titles = nav.titles[:len(nav.titles)-1]
+	}
+	title := nav.title
+	if len(nav.titles) > 0 {
+		title = nav.titles[len(nav.titles)-1]
+	}
+	nav.label.SetText(title)
+
+	return obj
+}
+
+// SetTitle changes the navigation title and the title for the current object.
+//
+// Since: 2.6
+func (nav *Navigation) SetTitle(s string) {
+	nav.title = s
+	nav.titles[len(nav.titles)-1] = s
+	nav.label.SetText(s)
+}
+
+func (nav *Navigation) CreateRenderer() fyne.WidgetRenderer {
+	return widget.NewSimpleRenderer(NewBorder(nav.control, nil, nil, nil, nav.stack))
+}
--- a/container/navigation_test.go
+++ b/container/navigation_test.go
@@ -1,0 +1,104 @@
+package container
+
+import (
+	"testing"
+
+	"fyne.io/fyne/v2/widget"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNavigation_Basics(t *testing.T) {
+	l := widget.NewLabel("something")
+	nav := NewNavigation("Title", l)
+
+	assert.Equal(t, true, nav.button.Disabled())
+	assert.Equal(t, 1, len(nav.stack.Objects))
+	assert.Equal(t, 1, len(nav.titles))
+
+	assert.Equal(t, l, nav.Pop())
+
+	assert.Equal(t, "Title", nav.label.Text)
+}
+
+func TestNavigation_NewWithMultiple(t *testing.T) {
+	a := widget.NewLabel("a")
+	b := widget.NewLabel("b")
+	c := widget.NewLabel("c")
+	nav := NewNavigation("Title", a, b, c)
+
+	assert.Equal(t, false, nav.button.Disabled())
+	assert.Equal(t, 3, len(nav.stack.Objects))
+	assert.Equal(t, 3, len(nav.titles))
+
+	assert.Equal(t, c, nav.Pop())
+	assert.Equal(t, "Title", nav.label.Text)
+	assert.Equal(t, 2, len(nav.stack.Objects))
+	assert.Equal(t, 2, len(nav.titles))
+	assert.Equal(t, false, nav.button.Disabled())
+
+	assert.Equal(t, b, nav.Pop())
+	assert.Equal(t, "Title", nav.label.Text)
+	assert.Equal(t, 1, len(nav.stack.Objects))
+	assert.Equal(t, 1, len(nav.titles))
+	assert.Equal(t, true, nav.button.Disabled())
+}
+
+func TestNavigation_PushWithTitle(t *testing.T) {
+	a := widget.NewLabel("a")
+	b := widget.NewLabel("b")
+	c := widget.NewLabel("c")
+	nav := NewNavigation("Title")
+
+	assert.Equal(t, true, nav.button.Disabled())
+	assert.Equal(t, 0, len(nav.stack.Objects))
+	assert.Equal(t, 0, len(nav.titles))
+
+	nav.PushWithTitle(a, "A")
+	assert.Equal(t, "A", nav.label.Text)
+	assert.Equal(t, true, nav.button.Disabled())
+
+	nav.PushWithTitle(b, "B")
+	assert.Equal(t, "B", nav.label.Text)
+	assert.Equal(t, false, nav.button.Disabled())
+
+	nav.PushWithTitle(c, "C")
+	assert.Equal(t, "C", nav.label.Text)
+	assert.Equal(t, false, nav.button.Disabled())
+
+	assert.Equal(t, c, nav.Pop())
+	assert.Equal(t, "B", nav.label.Text)
+	assert.Equal(t, 2, len(nav.stack.Objects))
+	assert.Equal(t, 2, len(nav.titles))
+	assert.Equal(t, false, nav.button.Disabled())
+
+	assert.Equal(t, b, nav.Pop())
+	assert.Equal(t, "A", nav.label.Text)
+	assert.Equal(t, 1, len(nav.stack.Objects))
+	assert.Equal(t, 1, len(nav.titles))
+	assert.Equal(t, true, nav.button.Disabled())
+
+	assert.Equal(t, a, nav.Pop())
+	assert.Equal(t, "Title", nav.label.Text)
+	assert.Equal(t, 0, len(nav.stack.Objects))
+	assert.Equal(t, 0, len(nav.titles))
+	assert.Equal(t, true, nav.button.Disabled())
+
+	assert.Nil(t, nav.Pop())
+	assert.Equal(t, "Title", nav.label.Text)
+	assert.Equal(t, 0, len(nav.stack.Objects))
+	assert.Equal(t, 0, len(nav.titles))
+	assert.Equal(t, true, nav.button.Disabled())
+}
+
+func TestNavigation_Empty(t *testing.T) {
+	nav := NewNavigation("Title")
+
+	assert.Equal(t, true, nav.button.Disabled())
+	assert.Equal(t, 0, len(nav.stack.Objects))
+	assert.Equal(t, 0, len(nav.titles))
+
+	assert.Nil(t, nav.Pop())
+
+	assert.Equal(t, "Title", nav.label.Text)
+}


container/navigation.go
```diff
@@ -0,0 +1,118 @@
...
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/theme"
+	"fyne.io/fyne/v2/widget"
+)
+
+// Navigation container is used to provide your application with a control bar and an area for content objects.
+// Objects can be any CanvasObject, and only the most recent one will be visible.
+//
+// Since: 2.6
+type Navigation struct {
```
dweymouth, line 13: The way this is written it can only be constructed via `NewNavigation` and not easily extended, or constructed by a struct literal instead, as has currently been a requirement for all builtin Fyne widgets. Most of this struct's contents may be able to move to the renderer instead.

container/navigation.go
```diff
@@ -0,0 +1,118 @@
+package container
+
+import (
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/theme"
+	"fyne.io/fyne/v2/widget"
+)
+
+// Navigation container is used to provide your application with a control bar and an area for content objects.
```
dweymouth, line 9: If the idea is for this to be the final API, it should probably be called StackNavigation, to distinguish it from other navigation schemes, such as a browser-like queue of past pages that can be navigated back and forward. Unless the idea is to make this single component support multiple navigation schemes in the future

container/navigation.go
```diff
@@ -0,0 +1,118 @@
...
+		control: NewStack(NewHBox(button), label),
+		stack:   NewStack(objs...),
+		label:   label,
+		title:   s,
+		titles:  titles,
+	}
+	nav.ExtendBaseWidget(nav)
+	nav.button.OnTapped = func() {
+		nav.Pop()
+	}
```
Jacalz, line 44:
```suggestion
	nav.button.OnTapped = nav.Pop
```

container/navigation.go
```diff
@@ -0,0 +1,118 @@
...
+
+	return nav
+}
+
+// Push puts the given object on top of the navigation stack and hides the object below.
+//
+// Since: 2.6
+func (nav *Navigation) Push(obj fyne.CanvasObject) {
+	nav.PushWithTitle(obj, nav.titles[len(nav.titles)-1])
```
Jacalz, line 56: I think this will crash if there are 0 titles? Or can that not happen?

Jacalz (changes requested):
A very useful addition. I agree with the comments from Drew but I also added some tips/suggestions inline.

container/navigation.go
andydotxyz: But couldn't you build a browser back/forward with this too? It could be extended to have forward navigation without needing Stack vs not-stack navigation.
Conceptually it feels like the browser is a stack as well?

Jacalz: Is that really the best option? Do we know that all the other navigators will have the same API?

With more and more types in the container package that actually aren't simple containers but technically widgets, I wonder if a "navigation" package would make sense? Tabs would fit in quite well there as well, I'd think. Just an idea to bring up before we commit to anything :)

andydotxyz: I'm not sure that tabs fit a mental model of navigation - you're not "going" anywhere...
Container has diversified but the general concept of "you put other widgets in there" mostly holds.

dweymouth: I don't really think you could build a browser-like navigation with this component without just forking it and adding a different API - the browser navigation model is only a bit "stack-like", but it's really more of a list with a pointer that can advance back and forward, and when you push a new page, it deletes the whole "tail" of the history after the current pointer position. (ie back, back, back, new page would remove the last 3 items in the old history)

container/navigation.go
sdassow: Can't happen, but it actually points at something related I was wondering: can the stack be empty or is there always at least one element? Or should it be configurable?

container/navigation.go
sdassow: Different function signature, forgot to explicitly ignore the return value. Still raising the API question, should Pop() return the object?

container/navigation.go
sdassow: Good point. Have to look from that angle again and compare to other widgets.

container/navigation.go
sdassow: I'm now considering to change the logic to support an optional "Next" button.

Jacalz:
FYI: This likely closes https://github.com/fyne-io/fyne/issues/1166.

container/navigation.go
andydotxyz: Does that mean the naming works now? If it needs a more explicit name then OK but I'm not convinced that "Stack" or "Simple" add anything meaningful to a developer's understanding of what it is.

dweymouth: I think the naming works if we plan to add (at least eventually) the forward navigation as well. I wonder if users may also confuse this with a "routing" system though, which would be more complex still.

container/navigation.go
sdassow: This needs more thought than I expected to make sure everything is initialized correctly, but I started making some parts public.

container/navigation.go
sdassow: The forward navigation is now in the PR.

andydotxyz: I'm happy with this now. If a routing system is to be added that would not be presented as a navigation container in this sense - more likely it would have to be a parser of path through a UI which then populates the various containers / tabs along the way to that content. 

Maybe that's overly complex but I'm happy that routing is not a widget/container concern so there is no name conflict.

andydotxyz:
What is the status of this? We have just over 2 days to get any new API in for 2.6.

I'm happy with this change but given the history of the feature I don't feel it would be appropriate for me to accept it.

Jacalz:
The comment by Drew about it only being possible to use this when created using the constructor is still a valid one in my opinion and it hasn't been resolved.

dweymouth:
> The comment by Drew about it only being possible to use this when created using the constructor is still a valid one in my opinion and it hasn't been resolved.

Agree - though I think Navigation.stack may be the last one left that needs work for its zero value to be valid/not crash

andydotxyz:
Thanks, sounds like a good summary - @sdassow is this something you can resolve before Friday?

Jacalz:
If it helps, my implementation at https://github.com/Jacalz/rymdport/blob/rewrite/internal/ui/components/stack.go does not have that issue. Feel free to use it as inspiration if it helps.

container/navigation.go
sdassow: Got it working by adding a setup method and a bunch of checks.

container/navigation.go
```diff
@@ -0,0 +1,196 @@
...
+// NavigationObject allows using any object implementing the `fyne.Disableable`,
+// `fyne.CanvasObject`, and `fyne.Tappable` interfaces to be used as Back/Next
+// navigation control.
+//
+// Since: 2.6
+type NavigationObject interface {
+	fyne.CanvasObject
+	fyne.Disableable
+	fyne.Tappable
+}
```
Jacalz, line 37: Like indicated above, I'm sceptical about this interface and the custom solution for changing the backwards and next buttons. I'd prefer to just have an `OnBack` and an `OnNext` function to keep things simple. I'd rather begin without this and we can add this later if there is enough demand.

container/navigation.go
```diff
@@ -0,0 +1,196 @@
...
+	nav := &Navigation{
+		Root:  root,
+		Title: s,
+		Label: widget.NewLabelWithStyle(s, fyne.TextAlignCenter, fyne.TextStyle{Bold: true}),
+	}
+	nav.setup()
+	nav.Back = widget.NewButtonWithIcon("", theme.NavigateBackIcon(), func() { _ = nav.Pop() })
+	nav.Back.Disable()
+	nav.Next = widget.NewButtonWithIcon("", theme.NavigateNextIcon(), func() { _ = nav.Forward() })
+	nav.Next.Disable()
```
Jacalz, line 59: This doesn't quite solve the review comment about making sure that the type should be usable from the struct. All fields or exported and part of the struct so sure, you can use the struct but there is a lot of setup required. Have a look at how https://github.com/Jacalz/rymdport/blob/rewrite/internal/ui/components/stack.go has all of the widgets internal in the renderer and handles the setup of them in the `CreateRenderer` call. That is more consistent with out other APIs and avoid the developer having to do a lot of setup to get the API to work.

Basically, let the renderer contain the internal widgets and do most of the setup in the `CreateRenderer` call. If the data inside the Navigation struct changes after a refresh, the methods on our custom renderer change the contents of our internal widgets to reflect the change.

container/navigation.go
```diff
@@ -0,0 +1,196 @@
...
+		box = append(box, nav.Back)
+	}
+	if nav.Next != nil {
+		box = append(box, layout.NewSpacer(), nav.Next)
+	}
+	control := NewStack(NewHBox(box...))
+	if nav.Label != nil {
+		control.Objects = append(control.Objects, nav.Label)
+	}
+	return widget.NewSimpleRenderer(NewBorder(control, nil, nil, nil, &nav.stack))
```
Jacalz, line 195: With the comments above, I think we basically can't use the simple renderer. There needs to be an actual custom renderer with the internal widgets and properties reachable so we can have the methods on that renderer change the internal widgets when values on our widget is changed. You can use https://github.com/Jacalz/rymdport/blob/4707bc88f31670e2de0293f914f0894e30b78755/internal/ui/components/stack.go#L79 for inspiration.

container/navigation.go
```diff
@@ -0,0 +1,196 @@
...
+// Objects can be any CanvasObject, and only the most recent one will be visible.
+//
+// Since: 2.6
+type Navigation struct {
+	widget.BaseWidget
+
+	Root  fyne.CanvasObject
+	Title string
+	Back  NavigationObject
+	Next  NavigationObject
```
Jacalz, line 20: I think this likely adds more complexity over having the CreateRenderer just create buttons internally and let the developer wire up functions for what should happen on next and back. Just doing `OnBack func()` and `OnNext func()` will likely suffice. There are two options for that approach:

1. Always show backwards and forwards buttons and use a sensible approach for selecting what they do when no functions are set (just moving backwards and forwards).
2. Don't show buttons at all when functions are not set.

container/navigation.go
```diff
@@ -0,0 +1,196 @@
...
+	Label *widget.Label
+
+	level  int
+	stack  fyne.Container
```
Jacalz, line 24: While I do like this approach, it might make more sense (and be more consistent) with just a `[]fyne.CanvasObject` slice and have the container handled in the renderer. See comment further down.

container/navigation.go
```diff
@@ -0,0 +1,196 @@
...
+
+	if nav.Next != nil {
+		nav.Next.Enable()
+	}
+	nav.level--
+
+	return objs[nav.level]
+}
+
+// Forward shows the next object in the again.
```
Jacalz, line 132: There is a word missing here at the end.

Jacalz (changes requested):
I'm so sorry for the late review. I really like the approach of the global title and the root object being handled like they are now; it is a big improvement over my navigation container. The support for moving forward to an object that already is added is nice. I think we also want a method to go back without popping?

However, I am a bit sceptical to there being a lot of complexity and manual setup required for a developer to use this (especially when using the struct). It does not quite work like how many of our other APIs work in this regard; it feels to me like internal details are exposed as exported fields in the main struct. I left a few comments inline to try and explain more thoroughly what I mean. I propose handling our internal widgets inside the renderer (by not using a simple renderer), slimming down the custom navigation objects into creating our own buttons internally and letting the developer set `OnNext` and `OnBack` functions if they want a different behaviour than default.

container/navigation.go
andydotxyz: This is a good point - but is slightly ironic because the "forward nagivation" was added in response to it not being sufficient to be called `Navigation` without it - so we might need a little discussion to agree the scope.

container/navigation.go
```diff
@@ -0,0 +1,196 @@
...
+// Navigation container is used to provide your application with a control bar and an area for content objects.
+// Objects can be any CanvasObject, and only the most recent one will be visible.
+//
+// Since: 2.6
+type Navigation struct {
+	widget.BaseWidget
+
+	Root  fyne.CanvasObject
+	Title string
+	Back  NavigationObject
```
andydotxyz, line 19: I'm not a fan of exporting these objects either - a standard back and forward buttons should be sufficient

container/navigation.go
```diff
@@ -0,0 +1,196 @@
...
+	Next  NavigationObject
+	Label *widget.Label
```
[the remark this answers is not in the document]
andydotxyz: Yeah, `string` is what I would expect here too

container/navigation.go
andydotxyz: Actually maybe I misunderstood - with or without forward navigation this interface should not be needed.

sdassow: I didn't really like it either - now gone.

container/navigation.go
sdassow: Adopted this approach, thanks!

dweymouth (commented):
To be compatible with a wider variety of use cases, I wonder if this should be configurable to not show the title bar - ie act only as a stack of "pages". E.g. for apps that would show the back/forward buttons elsewhere in the UI (e.g. Supersonic)

andydotxyz:
> To be compatible with a wider variety of use cases, I wonder if this should be configurable to not show the title bar - ie act only as a stack of "pages". E.g. for apps that would show the back/forward buttons elsewhere in the UI (e.g. Supersonic)

At that point it's basically just a Stack container where you can hide/show items...

dweymouth:
> > To be compatible with a wider variety of use cases, I wonder if this should be configurable to not show the title bar - ie act only as a stack of "pages". E.g. for apps that would show the back/forward buttons elsewhere in the UI (e.g. Supersonic)
> 
> At that point it's basically just a Stack container where you can hide/show items...

Well it would still handle the history management, e.g push 5 items followed by going Back 3, forward 1, then pushing a new item would drop the last 2 from the history. But that logic is simple enough that it's easy for such apps to just write their own custom component based on a slice of CanvasObject

container/navigation.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	nav.level--
+
+	return objs[nav.level]
+}
+
+// Forward shows the next object in the again.
+//
+// Since: 2.6
+func (nav *Navigation) Forward() fyne.CanvasObject {
```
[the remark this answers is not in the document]
dweymouth: Maybe just renaming Pop to Back, so the API is Push, Back, Forward? I'm not really sure what the difference between Pop and Back would be. Maybe Pop removing from history (so you can't go forward anymore) and Back leaving it?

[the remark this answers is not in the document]
andydotxyz: well, with "back" the "forward" is still possible. but with "pop" it is not...

Is there some ambiguity here? Or does the Forward not belong any more?

(also isn't there a length check missing here?)

container/navigation.go
```diff
@@ -0,0 +1,207 @@
...
+type Navigation struct {
+	widget.BaseWidget
+	level  int
+	stack  fyne.Container
+	titles []string
+
+	Root   fyne.CanvasObject
+	Title  string
+	OnBack func()
+	OnNext func()
```
andydotxyz: I think this should be `OnForward` no?

container/navigation.go
```diff
@@ -0,0 +1,207 @@
...
+	"fyne.io/fyne/v2/widget"
+)
+
+// Navigation container is used to provide your application with a control bar and an area for content objects.
+// Objects can be any CanvasObject, and only the most recent one will be visible.
+//
+// Since: 2.7
+type Navigation struct {
+	widget.BaseWidget
+	level  int
```
andydotxyz, line 22: Can the private fields be after the public ones? at least for initial commit of a widget that fits our ordering better

container/navigation.go
```diff
@@ -0,0 +1,207 @@
...
+	nav.stack.Objects[nav.level].Show()
+	nav.level++
+
+	return nav.stack.Objects[nav.level-1]
+}
+
+// SetTitle changes the navigation title and the title for the current object.
+//
+// Since: 2.7
+func (nav *Navigation) SetTitle(s string) {
```
andydotxyz, line 109: This function makes me a little nervous - what it does varies with what the current level is but the field `Title` does not vary... 

I think for consistency this should be setting the root title. Perhaps another method that more clearly indicates it is setting the title of a level.

andydotxyz (changes requested):
This looks excellent thanks so much, just a few things inline

container/navigation.go
sdassow: Indeed, and fixed it in the renderer as well.

container/navigation.go
sdassow: Added SetCurrentTitle, which seems a lot more precise, and SetTitle only does the root now.

andydotxyz (approved):
Thanks so much for getting this over the line.

Jacalz (approved):

Jacalz:
Thanks. I took the liberty of merging this given that it includes a fix to make the static analysis pass on develop. Makes it easier than juggling around merge conflicts.

Issue #5395, opened 2025-01-11
Aggressive refresh on resize of Tree widget


Jacalz:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

I noticed in some code refactoring that the Tree widget overrides the `.Resize()` methods using some internal parts of the base widget and manually does a refresh instead of repaint. This is technically wrong both from how widgets should be have and from a performance stand point.

https://github.com/fyne-io/fyne/blob/4a875d9d5677ce8a98bd15787d6f74d4263cac16/widget/tree.go#L239-L248

https://github.com/fyne-io/fyne/blob/4a875d9d5677ce8a98bd15787d6f74d4263cac16/widget/tree.go#L599-L607

### How to reproduce

1. Remove extended resize methods
2. Notice that the test fail

### Example code

See reproduction steps above. Easier to see in tests and than demo.

### Fyne version

v2.5.3

### Go compiler version

1.23.4

### Operating system and version

Fedora Silverblue 41

PR #5504, opened 2025-02-05
Make Tree scrolling and resizing more efficient by only refreshing newly visible nodes

### LINKED ISSUE #5456: Tree and Table widgets refresh full content on any scroll or resize

dweymouth:
### Description:

Fixes #5456, #5395

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/tree.go
+++ b/widget/tree.go
@@ -2,7 +2,6 @@ package widget
 
 import (
 	"fmt"
-	"slices"
 
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/canvas"
@@ -794,7 +793,7 @@ func (r *treeContentRenderer) refreshForID(toDraw TreeNodeID) {
 
 	if toDraw == onlyNewTreeNodesID {
 		for id, b := range r.branches {
-			if slices.Contains(r.visible, id) && !slices.Contains(r.wasVisible, id) {
+			if contains(r.visible, id) && !contains(r.wasVisible, id) {
 				b.Refresh()
 			}
 		}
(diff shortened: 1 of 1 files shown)


dweymouth:
> If this fixes #5395, then I think the overrides for the Resize methods need to be removed. With those there still, it likely is refreshing on resize still.

All the collection widgets override Resize because they do need to (partially) refresh on resize, to update and draw any new items that become visible when resizing larger. The fix here is that we used to do a full refresh, but now it is only refreshing any new nodes which may appear during a resize, and just move/resizing everything else

Jacalz (approved):
Sounds good. Thanks for the explanation :)

Issue #5502, opened 2025-02-05
Replace untyped bindings with generic implementations


Jacalz:
### Checklist

- [x] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [x] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

We do currently have a couple untyped bindings that rely on runtime reflection to look up types and set their values appropriately. This reflection is slow, does not give the developer any type safety (which may lead to crashes or potential security issues popping up) and it might lead to larger binaries as the Go compiler needs to embed more type information that can be used at runtime. We should avoid runtime reflection to as large extent as we can.

It has been proven with #5378 that generics are a great fit for reducing code duplication within binding types. A generic implementation for bindings should be possible to use for this purpose.

### Is it possible to construct a solution with the existing API?

Yes

### Describe the solution you'd like to see.

I propose that we expose a generic binding that lets the compiler generate a type safe binding implementation for the specific type that the developer wants to use instead.

PR #5505, opened 2025-02-05
Add a generic bindable item

Jacalz:
### Description:

Introduce a generic bindable item so we can deprecate the untyped bindings. Should be both faster and type safe.

For #5502 

### Checklist:

- [ ] Tests included. <- Actually covered already thanks to generic implementation.
- [x] Tests all pass.

#### Where applicable:


Jacalz:
I accidentally pushed two commits onto this branch. Removed them.

andydotxyz (changes requested):
This is a great addition thanks - but I don't think we should deprecate the `Untyped` at this time.

By making that change we are telling people to start using generics and that's not a stance I think we have to take.

Feel free to call out the alternative - but marking deprecated seems a little aggressive for a style question.

Jacalz:
> By making that change we are telling people to start using generics and that's not a stance I think we have to take.

Why? I see no point in being against generics when they have an obvious benefit. They have existed since Go 1.18 and work well. Several types in the standard library are already generic, like `atomic.Pointer[T]`. The untyped binding is not type safe in any way and there were even recently on Discord a case where someone was confused about the code crashing because two different types could not be compared. Using the generic item with `any` achieves basically the same thing as untyped but without runtime reflection. However, it is of course better to specify the custom type to use instead of `any`.

> Feel free to call out the alternative - but marking deprecated seems a little aggressive for a style question.

See notes about runtime reflection and type safety above. This is not a style question. Instead of using `Untyped` for your custom type, it is better to use the generic version with you specific type to both get better performance and type safety.

Jacalz:
Perhaps the deprecation notices should be clarified?

Jacalz:
I don't understand why this was closed. Strange.

Jacalz:
I added a clarification about type safety to the deprecation notes.

andydotxyz:
> Why? I see no point in being against generics when they have an obvious benefit

I am not being "against generics" however the code works perfectly well without them. So why not support both?

My understanding was that generics would be an enabler for our data binding but that we would not require it in the public API.

Jacalz:
> I am not being "against generics" however the code works perfectly well without them. So why not support both?

The generic API gives us type safety and improved performance. I see literally no use for an old API that does the same thing without type safety. Runtime reflection is slow and increases binary size because the compiler has to embed more type data into the binary so it can do stuff at runtime.

> My understanding was that generics would be an enabler for our data binding but that we would not require it in the public API.

It certainly is an enabler but why have two things that do the same hing when one is obviously better?

Jacalz:
The requirements around the item having to be a pointer into the Untyped binding is also strange. The generic API works much more like the other bindings we already have (obviously because they are implemented with it).

zSnails:
Would've been nice if you also added a generic List

andydotxyz:
> Would've been nice if you also added a generic List

Please don't post comments on closed items. If you want to request that then open a new feature request.

PR #5506, opened 2025-02-05
Fix moving of folder/listable content

### LINKED ISSUE #5493: storage.Move() fails to move directories
### Checklist

### Describe the bug

$path is just the path of the folder/directory

On Linux:
write $path: copy_file_range: is a directory

On Windows:
read $path: Incorrect function.

### How to reproduce

When trying to use the move function on a directory/folder, I get the following error listed above

### Screenshots

The code below results with RootURI looking like this on Linux
Image

### Example code

```go
package main

import (
	"fmt"

	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/storage"
)

func main() {
	fyneApp := app.NewWithID("MoveFailTestCode")

	path, err := storage.Child(fyneApp.Storage().RootURI(), "Folder")
	fmt.Println(err)

	exists, err := storage.Exists(path)
	fmt.Println(err)

	if !exists {
		err := storage.CreateListable(path)

		if err != nil {
			fmt.Println(err)
		}
	}

	newPath, err := storage.Child(fyneApp.Storage().RootURI(), "RenamedFolder")
	fmt.Println(err)

	err = storage.Move(path, newPath)
	fmt.Println(err)
}
```

### Fyne version

2.5.4

### Go compiler version

1.23.4

Linux OpenSuse Tumbleweed && Windows 10

### Additional Information

_No response_

andydotxyz:
Fixes #5493

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/repository/file.go
+++ b/internal/repository/file.go
@@ -301,6 +301,15 @@ func (r *FileRepository) Copy(source, destination fyne.URI) error {
 //
 // Since: 2.0
 func (r *FileRepository) Move(source, destination fyne.URI) error {
+	listSrc, _ := r.CanList(source)
+	if listSrc {
+		err := os.Rename(source.Path(), destination.Path())
+		if err == nil {
+			return nil
+		}
+		// fallthrough to slow move
+	}
+
 	// NOTE: as far as I can tell, golang does not have an optimized Move
 	// function - everything I can find on the 'net suggests doing more
 	// or less the equivalent of GenericMove(), hence why that is used.
--- a/internal/repository/file_test.go
+++ b/internal/repository/file_test.go
@@ -410,6 +410,39 @@ func TestFileRepositoryMove(t *testing.T) {
 	assert.False(t, ex)
 }
 
+func TestFileRepositoryMoveDirectory(t *testing.T) {
+	dir := t.TempDir()
+
+	// Create a file in a dir to test with
+	parentPath := path.Join(dir, "parent")
+	fooPath := path.Join(parentPath, "foo")
+	newParentPath := path.Join(dir, "newParent")
+	newFooPath := path.Join(newParentPath, "foo")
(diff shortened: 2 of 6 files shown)


internal/repository/file.go
```diff
@@ -301,6 +301,15 @@ func (r *FileRepository) Copy(source, destination fyne.URI) error {
...
+		err := os.Rename(source.Path(), destination.Path())
+		if err == nil {
+			return nil
+		}
+		// fallthrough to slow move
+	}
+
 	// NOTE: as far as I can tell, golang does not have an optimized Move
 	// function - everything I can find on the 'net suggests doing more
 	// or less the equivalent of GenericMove(), hence why that is used.
```
Jacalz, line 315: Does the above change not invalidate this comment? I think it can be removed. I also kind of don't like that it says "golang" lol so that motivates me more to remove it :)

Jacalz (changes requested):
Thanks. Looks great. I think there just is a comment that no longer applies with these changes.

internal/repository/file.go
Jacalz: Is it not possible to use the same `os.Rename` function for files?

andydotxyz: It might be - I was only looking at folders.

Whatever we have to leave the slow path in as well because any Move (/Rename) can fail due to filesystem traversal or cross-volume issues.

Jacalz: Alright. Lets add a fast path for regular files as a followup

Jacalz (approved):

internal/repository/file.go
Jacalz: Opened https://github.com/fyne-io/fyne/issues/5517

PR #5508, opened 2025-02-06
Fix the thread we are firing on from binding callbacks

andydotxyz:
### Checklist:

- [ ] Tests included. <- macOS apps would crash when data binding updates window / menu
- [x] Tests all pass.

### DIFF
--- a/data/binding/binding.go
+++ b/data/binding/binding.go
@@ -139,7 +139,7 @@ func (b *boundUntyped) Set(val any) error {
 	}
 	b.lock.Unlock()
 
-	b.trigger()
+	fyne.Do(b.trigger)
 	return nil
 }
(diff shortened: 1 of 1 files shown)


dweymouth (approved):

dweymouth:
Worth noting if those *are* called from main, they will call fyne.Do from main as well. (Which is not a problem for fyne.Do, ony for fyne.DoAndWait). We should probably update our threading checks to exempt fyne.Do from the "called from main" warning

andydotxyz:
> We should probably update our threading checks to exempt fyne.Do from the "called from main" warning

A good point. Probably with an issue or follow up PR. now we are iterating the stack it could be done.

PR #5509, opened 2025-02-06
We can ignore warnings if the request for background execution is not waiting

andydotxyz:
as discussed on #5508 , main (not waiting) from main is OK now. 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/mobile/driver.go
+++ b/internal/driver/mobile/driver.go
@@ -97,9 +97,7 @@ func (d *driver) DoFromGoroutine(fn func(), wait bool) {
 	}
 
 	if wait {
-		async.EnsureNotMain(func() {
-			caller()
-		})
+		async.EnsureNotMain(caller)
 	} else {
 		caller()
 	}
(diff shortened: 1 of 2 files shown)


internal/driver/mobile/driver.go
```diff
@@ -94,7 +94,16 @@ func (d *driver) DoFromGoroutine(fn func(), wait bool) {
 		if wait {
 			<-done
 		}
-	})
+	}
+
+	if wait {
+		async.EnsureNotMain(func() {
+			caller()
+		})
```
Jacalz, line 102:
```suggestion
		async.EnsureNotMain(caller)
```

Jacalz (changes requested):
This is a great idea. With `fyne.Do` and the unbounded channel this makes a lot of sense. Just one minor nitpick inline that makes the stack output a bit cleaner.

Jacalz (approved):

PR #5510, opened 2025-02-06
Queue binding operations without checking thread

Jacalz:
### Description:

Once #5509 is merged, running binding callbacks from main no longer print a warning so we cna avoid checking the thread. The result is much faster performance.

I also fixed a bug where we were scheduling s.DataChanged() onto main, which in itself then runs a trigger() which schedules the work onto main. That was unnecessary overhead.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


data/binding/convert.go
```diff
@@ -258,7 +258,7 @@ func (s *toStringFrom[T]) Set(str string) error {
 		return err
 	}
 
-	queueItem(s.DataChanged)
+	s.DataChanged()
```
dweymouth, line 261: Does this need to be in fyne.Do?

Jacalz: Like I wrote in the description, `s.DataChanged()` just runs `.trigger()` which queues the operation into main. Having a second fyne.Do here is not needed.

dweymouth (approved):

data/binding/convert.go
andydotxyz: I'm not totally sure I agree here - the `DataChanged` is part of the public API and must also be called on the correct thread.

andydotxyz (changes requested):
I /think/ that `DataChanged` should be called on main as it is part of the public API and could be called back into user facing code.

> I also fixed a bug where we were scheduling s.DataChanged() onto main, which in itself then runs a trigger() which schedules the work onto main. That was unnecessary overhead.

There may be a bug in here, but I think the right solution is slightly different.

Jacalz:
> There may be a bug in here, but I think the right solution is slightly different.

I'm not sure I agree, åt least not entirely. We said that data binding are thread safe, did we not? Also, the old code was inherently wrong as it scheduled a call onto main which then just scheduled a call onto main. I don't see this being any worse than it already was.

Jacalz:
DataChanged() for the conversion types literally only calls trigger() which itself schedules the data listener updates onto main. We don't need to call DataChanged() on ourselves from main.

Jacalz:
Ah, I kind of see what you mean. We would match other binding types better by just calling .trigger() in those instances.

andydotxyz:
> DataChanged() for the conversion types literally only calls trigger() which itself schedules the data listener updates onto main. We don't need to call DataChanged() on ourselves from main.

The core of this is that `DataChanged` is in the listener API - anyone that called `AddListener` would expect the Fyne callback to be on the correct thread.

I think this is now correct.

andydotxyz (approved):
Thanks, I think that's the threading sorted now.

Jacalz:
> I think this is now correct.

Thanks. Yeah, I agree and your explanation matches my understanding as well now. Anyone actually calling DataChanged on those types would get a double bounce on main though. Might be a problem for a future PR though

PR #5511, opened 2025-02-06
Add a OpenURI helper

andydotxyz:
### Checklist:

- [ ] Tests included. <- any tips? :)
- [x] Tests all pass.

### DIFF
--- a/uri.go
+++ b/uri.go
@@ -111,7 +111,7 @@ type URIWithIcon interface {
 func OpenURI(u URI) error {
 	a := CurrentApp()
 	if a == nil {
-		return errors.New("oepnURI requires a running Fyne application")
+		return errors.New("calling openURI requires a running Fyne application")
 	}
 
 	parsed, err := url.Parse(u.String())
--- a/uri.go
+++ b/uri.go
@@ -1,8 +1,10 @@
 package fyne
 
 import (
+	"errors"
 	"fmt"
 	"io"
+	"net/url"
 )
 
 // URIReadCloser represents a cross platform data stream from a file or provider of data.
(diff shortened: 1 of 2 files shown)


uri.go
```diff
@@ -101,3 +103,21 @@ type URIWithIcon interface {
...
 }
+
+// OpenURI asks the operating system to open the specified resource.
+// This is commonly used for a URL or a file URI to open the appropriate system application.
+//
+// Since: 2.6
+func OpenURI(u URI) error {
+	a := CurrentApp()
+	if a == nil {
+		return errors.New("oepnURI requires a running Fyne application")
```
Jacalz, line 114: Fixing spelling mistake and prefix with a word so the name can be spelled with a capital case in the beginning:
```suggestion
		return errors.New("running OpenURI requires a running Fyne application")
```

Jacalz (changes requested):
I wonder if it perhaps would make more sense to the app interface instead (and possibly deprecate the old method or not)? I’m not sure Inlike the idea of having it in the global namespace. Feels a bit strange, especially if it is using parts of the app to open the uri.

andydotxyz:
> I wonder if it perhaps would make more sense to the app interface instead (and possibly deprecate the old method or not)? I’m not sure I like the idea of having it in the global namespace.

I agree it could indeed go there - but it just felt like a more serious change to alter the definition of `App` for trivial change.

uri.go
andydotxyz: Good catch. I used a different word to avoid the duplication.

Jacalz: Thanks. I'd still recommend going with the same function name though. It seems like you forgot to use a capital letter there.

Jacalz:
> I agree it could indeed go there - but it just felt like a more serious change to alter the definition of App for trivial change.

Hmm, maybe. Feels a bit strange though to have it globally. We have altered the driver and window interfaces before so I don't see altering app interface being too bad :)

andydotxyz:
I decided it was not really important enough to pull in at all. Nothing wrong with the URL implementation for communicating with the OS.

PR #5512, opened 2025-02-07
Support getting the cursor / relative positions from each in a text grid

andydotxyz:
Useful when popups or other items wish to interact with the text which is laid out completely internally.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (approved):

dweymouth:
This would be useful to have on Entry as well (though probably harder to calculate)

andydotxyz:
> This would be useful to have on Entry as well (though probably harder to calculate)

Yes it would, on both counts. Entry hasn't got the love it needs this release cycle, maybe next.

PR #5514, opened 2025-02-07
Provide ability to choose side of window a border button set will be

### LINKED ISSUE #5469: Change position of window controls for InnerWindow
### Checklist

The InnerWindow currently is very macOS-ish with buttons on the left side and maximize and minimize are switched compared to the positions where they would be on Windows or most Linux distributions (except for Gnome which by default only has a close button, sigh).

Image
Image

Yes

For this to match user expectations, I think we need to both set a default depending on the operating system (macOS has left with buttons in current order and all other desktop systems have it on the right with max/min swapped).

andydotxyz:
Opening for feedback just now

TODO:
- [ ] Tests
- [ ] OS level defaults (for mac and Windows at least)

Fixes #5469

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/container/innerwindow.go
+++ b/container/innerwindow.go
@@ -33,6 +33,7 @@ type InnerWindow struct {
 	OnDragged, OnResized                                func(*fyne.DragEvent) `json:"-"`
 	OnMinimized, OnMaximized, OnTappedBar, OnTappedIcon func()                `json:"-"`
 	Icon                                                fyne.Resource
+	Alignment                                           widget.ButtonAlign
 
 	title      string
 	borderIcon *borderButton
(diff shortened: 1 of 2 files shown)


Jacalz (commented):
Thanks for this. Trailing buttons looks like I would expect. I kind of feel like the default would have to be set depending on which platform we are on though. Windows and Linux likely want to default to trailing positions (while Gnome technically defaults to the left, Ubuntu has trailing by default and it is the most popular distribution plus KDE Plasma has them on the right as default also so there seems to be a majority on the right according to my research).

andydotxyz:
> I kind of feel like the default would have to be set depending on which platform we are on though.

This was checkbox 2 above :).

Jacalz:
Hehe yeah. Although I would consider Linux to also be on the list :)

andydotxyz:
> Hehe yeah. Although I would consider Linux to also be on the list :)

OK, not sure I am 100% in agreement but it is just a default, so I added it as requested.

Jacalz (approved):

PR #5519, opened 2025-02-08
Clean up WASM platform lookups

Jacalz:
### Description:

Avoids keeping global references to some WASM state that was unnecessary and also moves the macOS check to only be looked up once.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Thanks

PR #5520, opened 2025-02-08
Add a noCopy struct so `go vet` complains about copying widgets

dweymouth:
### Description:
Pre-2.6 we had an actual lock in BaseWidget so we got this warning "for free". Adding back a dummy `sync.Locker` implementation so we can still warn if widget structs are copied or passed by value.

image

For #4654 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/widget.go
+++ b/widget/widget.go
@@ -9,12 +9,6 @@ import (
 	"fyne.io/fyne/v2/theme"
 )
 
-type noCopy struct{}
-
-func (*noCopy) Lock() {}
-
-func (*noCopy) Unlock() {}
-
 // BaseWidget provides a helper that handles basic widget behaviours.
 type BaseWidget struct {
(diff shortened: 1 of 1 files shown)


andydotxyz (changes requested):
Thanks for this, but the utility type should be at the end of the file (or at least after the public types)

Jacalz:
I think it is common practice to have it at the top actually. It also avoids adding extra padding to have it at the top if I remember correctly.

Jacalz (dismissed):

Jacalz:
Oh, sorry. I read that comment incorrectly. I thought you said bottom of the structure. My bad :/

Jacalz (approved):

andydotxyz (approved):

PR #5533, opened 2025-02-14
Add datepicker widget

hazemhashem:
### Description:
This PR introduces a new `DatePicker` widget to allow users to select dates using dropdowns for year, month, and day. The widget is fully customizable and integrates seamlessly with the existing Fyne API.
Fixes #(issue)

### Checklist:
- [x] Tests included.
  - I have added unit tests for the `DatePicker` widget in `datepicker_test.go`.
  - I ran `go fmt` and `golint` with no errors.
- [x] Tests all pass.
  - All tests pass locally. Here's the output of  `go test -v -run TestDatePicker`:
    ```
=== RUN   TestDatePicker
--- PASS: TestDatePicker (0.00s)
PASS
ok      fyne.io/fyne/v2/widget  0.047s

    ```
### Output :
Screenshot from 2025-02-14 18-34-29

### DIFF
--- a/widget/datepicker.go
+++ b/widget/datepicker.go
@@ -1,0 +1,212 @@
+// DatePicker is a widget for selecting a date. It consists of three dropdowns
+// for year, month, and day. The days dropdown is dynamically updated based
+// on the selected month and year.
+//
+// Usage:
+//
+//	dp := widgets.NewDatePicker()
+//	date, err := dp.GetDate()
+//	if err != nil {
+//	    log.Println("No date selected")
+//	} else {
+//	    log.Println("Selected date:", date)
+//	}
+
+package widget
+
+import (
+	"fmt"
+	"strconv"
+	"time"
+
+	"fyne.io/fyne/v2"
+)
+
+// DatePicker is a custom widget for selecting a date.
+type DatePicker struct {
+	BaseWidget
+	YearSelect  *Select
+	MonthSelect *Select
+	DaySelect   *Select
+}
+
+// NewDatePicker creates a new DatePicker widget.
+func NewDatePicker() *DatePicker {
+	dp := &DatePicker{}
+	dp.ExtendBaseWidget(dp)
+
+	// Generate lists of years and months
+	years := generateYears()
+	months := generateMonths()
+
+	// Get Current Time to use as placeholders
+	currentTime := time.Now()
+
+	// Extract the year, month, and day and convert them to strings
+	year := currentTime.Year()
+	month := currentTime.Month()
+	day := currentTime.Day()
+
+	// Convert year and day to strings
+	yearStr := strconv.Itoa(year)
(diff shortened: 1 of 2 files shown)


andydotxyz:
How does this fit alongside the `DateEntry`?

I'm not sure about the 3 separate pickers - on mobile aren't these presented in a single keyboard/picker instead of directly in the user interface?

https://assets.website-files.com/6231812313e8661d7da3a4e7/62820ad35553f03e5fb5e1a0_13-Good-user-experience-datepicker-examples.png

hazemhashem:
You're absolutely right about the mobile usability concern. It slipped my mind because I don’t currently have an Android/iOS development environment to test mobile interfaces. My initial implementation is focused on desktop use cases, such as sign-up forms or other scenarios where a structured date selection might be more intuitive.

For desktop applications, having 3 separate pickers (year, month, day) can be beneficial in scenarios like:
- **Sign-up forms**: Users can quickly select their date of birth without needing to open a popup.
- **Filters or dashboards**: Users can easily adjust date ranges without navigating through a calendar.

I’d love to hear your thoughts on the following:
- Should the 3 separate pickers remain as a single widget for desktop use?
- Would it be better to move them into a custom dialog for better organization?
- Should we consider adapting this for mobile use in a future update?

Your guidance would be greatly appreciated as I want to ensure this widget aligns with Fyne’s design principles and usability standards.

andydotxyz:
> I’d love to hear your thoughts on the following:
> 
> * Should the 3 separate pickers remain as a single widget for desktop use?
> * Would it be better to move them into a custom dialog for better organization?
> * Should we consider adapting this for mobile use in a future update?

I'm not sure about all of these - why not use the Calendar popped up from the DateEntry widget?
I don't think we have mobile optimised that yet either - but it is much more touch-screen friendly.

sdassow:
Maybe moving the month and year Select fields into DateEntry popup header would be a good compromise?

hazemhashem:
> > I’d love to hear your thoughts on the following:
> > 
> > * Should the 3 separate pickers remain as a single widget for desktop use?
> > * Would it be better to move them into a custom dialog for better organization?
> > * Should we consider adapting this for mobile use in a future update?
> 
> I'm not sure about all of these - why not use the Calendar popped up from the DateEntry widget? I don't think we have mobile optimised that yet either - but it is much more touch-screen friendly.

Thanks for the insight! I actually wasn’t aware that the Calendar could pop up from the DateEntry widget.

Jacalz:
Is this still relevant with the DateEntry widget?

andydotxyz:
> Is this still relevant with the DateEntry widget?

That's an interesting question - the PR was opened after DateEntry existed.
I guess you could argue that it's a non-entry version of that same thing, but if that was the case it should probably be a button that opens the calendar for consistency

PR #5536, opened 2025-02-16
Adding selection option to Label

andydotxyz:
Selectable label incoming.

Got some tests to write/fix and a little refactoring to do - but I am pretty sure it's working well and has the right public API.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -255,7 +255,7 @@ func (e *Entry) DragEnd() {
 // Implements: fyne.Draggable
 func (e *Entry) Dragged(d *fyne.DragEvent) {
 	d.Position = d.Position.Add(fyne.NewPos(0, e.Theme().Size(theme.SizeNameInputBorder)))
-	e.sel.Dragged(d)
+	e.sel.dragged(d, false)
 	e.updateMousePointer(d.Position, false)
 }
 
@@ -373,7 +373,7 @@ func (e *Entry) MouseDown(m *desktop.MouseEvent) {
 	e.syncSelectable()
 
 	if isTripleTap(e.sel.doubleTappedAtUnixMillis, time.Now().UnixMilli()) {
-		e.sel.selectCurrentRow()
+		e.sel.selectCurrentRow(false)
 		e.CursorColumn = e.sel.cursorColumn
 		e.Refresh()
 		return
@@ -576,7 +576,7 @@ func (e *Entry) TouchDown(ev *mobile.TouchEvent) {
 		e.requestFocus()
 	}
 	if isTripleTap(e.sel.doubleTappedAtUnixMillis, now) {
-		e.sel.selectCurrentRow()
+		e.sel.selectCurrentRow(false)
 		e.CursorColumn = e.sel.cursorColumn
 		e.Refresh()
 		return
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -2,6 +2,7 @@ package widget
 
 import (
 	"image/color"
+	"log"
 	"runtime"
 	"strings"
 	"time"
@@ -272,6 +273,8 @@ func (e *Entry) FocusGained() {
 	e.setFieldsAndRefresh(func() {
 		e.dirty = true
 		e.focused = true
+
+		log.Println("FOCUS")
 	})
 	if e.onFocusChanged != nil {
 		e.onFocusChanged(true)
@@ -430,6 +433,7 @@ func (e *Entry) Refresh() {
 	if e.sel != nil {
 		e.sel.style = e.TextStyle
 		e.sel.theme = e.Theme()
+		e.sel.focussed = e.focused
 		e.sel.Refresh()
 	}
 	e.BaseWidget.Refresh()
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -273,8 +273,6 @@ func (e *Entry) FocusGained() {
 	e.setFieldsAndRefresh(func() {
 		e.dirty = true
 		e.focused = true
-
-		log.Println("FOCUS")
 	})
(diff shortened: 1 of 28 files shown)


widget/selectable.go
```diff
@@ -0,0 +1,376 @@
...
+	cursorRow, cursorCol := r.sel.cursorRow, r.sel.cursorColumn
+	selectRow, selectCol := -1, -1
+	if r.sel.selecting {
+		selectRow = r.sel.selectRow
+		selectCol = r.sel.selectColumn
+	}
+
+	if selectRow == -1 || (cursorRow == selectRow && cursorCol == selectCol) {
+		r.sel.selections = r.sel.selections[:0]
+
```
Jacalz, line 309: Same as above.

widget/selectable.go
```diff
@@ -0,0 +1,376 @@
...
+		if c == nil {
+			return
+		}
+
+		m := fyne.NewMenu("",
+			fyne.NewMenuItem(lang.L("Copy"), func() {
+				fyne.CurrentApp().Clipboard().SetContent(s.SelectedText())
+			}))
+		ShowPopUpMenuAtPosition(m, c, ev.AbsolutePosition)
+
```
Jacalz, line 122: This newline seems like it is a bit strange to have here in my opinion.

widget/entry.go
```diff
@@ -102,16 +95,14 @@ type Entry struct {
...
-	// used for deciding whether the next MouseDown/TouchDown is a triple-tap or not
-	doubleTappedAtUnixMillis int64
 }
 
 // NewEntry creates a new single line entry widget.
 func NewEntry() *Entry {
 	e := &Entry{Wrapping: fyne.TextWrap(fyne.TextTruncateClip)}
 	e.ExtendBaseWidget(e)
+
+	e.syncSelectable()
```
Jacalz, line 105: I don't think this should be here. If it is needed here then we can't guarantee that someone using the widget directly from the struct get consistent result. Better to just let the CreateRenderer call sync stuff like we do with the test of the providers.

Jacalz (changes requested):
Thanks. Just some minor comments inline.

Jacalz:
Hmm. I was testing this some more now before intending to approve and it seems like clicking outside of the label does not remove the selection?

Jacalz (commented):
Hmm. I was testing this some more now before intending to approve and it seems like clicking outside of the label does not remove the selection?

widget/label.go
```diff
@@ -92,15 +105,17 @@ func (l *Label) Refresh() {
...
-	l.BaseWidget.Resize(s)
-	if l.provider != nil {
-		l.provider.Resize(s)
+// Since: 2.6
+func (l *Label) SelectedText() string {
+	if !l.Selectable {
+		return ""
 	}
+
+	return l.selection.SelectedText()
```
dweymouth, line 118: I think we need a nil check around l.selection in case SelectedText is called before the renderer is created

widget/label.go
```diff
@@ -162,3 +177,36 @@ func (l *Label) updateFromData(data binding.DataItem) {
...
+	r.l.selection.Resize(s)
+	r.l.provider.Resize(s)
+}
+
+func (r *labelRenderer) MinSize() fyne.Size {
+	return r.l.provider.MinSize()
+}
+
+func (r *labelRenderer) Objects() []fyne.CanvasObject {
+	if !r.l.Selectable {
```
dweymouth, line 198: nice optimization for the non-selectable case :)

widget/label.go
```diff
@@ -162,3 +177,36 @@ func (l *Label) updateFromData(data binding.DataItem) {
...
+		return []fyne.CanvasObject{r.l.provider}
+	}
+
+	return []fyne.CanvasObject{r.l.selection, r.l.provider}
+}
+
+func (r *labelRenderer) Refresh() {
+	r.l.provider.Refresh()
+
+	sel := r.l.selection
```
dweymouth, line 208: I think we could do the same optimization here to avoid refreshing sel if the label is not selectable?

dweymouth (changes requested):

widget/label.go
andydotxyz: This allows us to keep the same XML for old usages of label too!

andydotxyz:
> Hmm. I was testing this some more now before intending to approve and it seems like clicking outside of the label does not remove the selection?

Thanks, all fixed plus I found some issues with mobile details as well, working well now I think.

andydotxyz:
Gah, missed the change in one file. Could have force pushed but forgot. Ah well, should be green now.

Jacalz (approved):
Tests seem to be failing but otherwise approved :+1:

widget/entry.go
```diff
@@ -288,6 +273,8 @@ func (e *Entry) FocusGained() {
 	e.setFieldsAndRefresh(func() {
 		e.dirty = true
 		e.focused = true
+
+		log.Println("FOCUS")
```
sdassow, line 277: Found a left-over.

andydotxyz: How embarrassing! Fixed

Jacalz: The import was accidentally left in leading to a compile error unfortunately

andydotxyz: Apologies, I should stop doing code commits on the bus!

Jacalz: Relatable. No problem

Jacalz (approved):

widget/entry.go
sdassow: Different kind of bus factor, and we're all in for the ride 😉

dweymouth (approved):

PR #5466, opened 2025-01-26
Fix widget renderer leak from cache never being cleaned

dweymouth:
### Description:

Calls the cache clean task in the main loop, which was previously not called due to a bug. Also periodically walks inactive windows to mark the canvasForObject cache entries as alive, to prevent freezes from cleaning the canvas cache in inactive windows.

Fixes #4903 

### Checklist:

- [ ] Tests included.
- [x] Lint and formatter run with no errors.
- [x] Tests all pass.

PR #2543, opened 2021-10-11
Fix freezes and crashes related to canvas/renderer cache cleanup

smcio:
### Description:

Fixes freeze and crashes related to canvas/renderer cache cleanup.

Fixes #2536

### Checklist:

- [ ] Tests included.
- [ ] Lint and formatter run with no errors.
- [ ] Tests all pass.

PR #5541, opened 2025-02-21
Update cache clean to never delete renderers of widgets still in the object tree

dweymouth:
### Description:

Follow-up for #5466 

Digging into this more I discovered why there are `Clean` and `CleanCanvases` tasks in the first place. This PR #2543 introduced CleanCanvases to address the issue of window freezing when deleting the cache entries for widgets still in the window. But it has a possible leak (fixed in #5466) where widgets that had a renderer created but were not placed in an object tree within a canvas would leak that renderer into the cache forever. #5466 fixed this but also ended up deleting renderers (but not CanvasForObject entries) for visible widgets. This fix avoids deleting renderers for widgets still in an active object tree (even if the window is hidden/ not redrawn), and also consolidates the cache clean task back to just the single `Clean` (since we're walking the object trees to mark the cache entries alive, we are avoiding the original issue)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/internal/cache/base.go
+++ b/internal/cache/base.go
@@ -107,19 +107,6 @@ func destroyExpiredRenderers(now time.Time) {
 	})
 }
 
-// matchesACanvas returns true if the canvas represented by the canvasInfo object matches one of
-// the canvases passed in 'canvases', otherwise false is returned.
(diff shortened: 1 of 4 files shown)


Jacalz (approved):
Nice. I'm glad to see that slice removed as well. Good find :+1:

PR #5545, opened 2025-02-22
Add safety around the fontMap

andydotxyz:
Fixing a single instance where concurrency in migration could cause crashes.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
With the lock around the code, I don’t think we actually need the `sync.Once` value. You should get away with an if-condition, I think.

Jacalz:
You don't set it to true so it likely is doing the setup each time now. Would it not be possible to look at the value of the fm to see if it is set up or not?

Jacalz (approved):

PR #5549, opened 2025-02-24
Go124 lib wasm path

adamdecaf:
### Description:
Go 1.24 [changed the location of `wasm_exec.js`](https://go.dev/doc/go1.24#wasm) under GOROOT. 

> The support files for WebAssembly have been moved to lib/wasm from misc/wasm.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/package-web.go
+++ b/cmd/fyne/internal/commands/package-web.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"bytes"
+	"os"
 	"path/filepath"
 	"runtime"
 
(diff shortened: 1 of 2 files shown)


Jacalz:
I don't think there were any plans for a v2.5.5 release and I didn't think about backporting the fix into the old cmd/fyne tool in time for v2.5.4 unfortunately. However, I do admit that having it not working on the latest Go version is far from ideal :/

adamdecaf:
I didn't realize cmd/fyne was being replaced. I have a locally compiled binary that works for me if cmd/fyne is being replaced.

Jacalz:
I think we are going to have to do a v2.5.5 release just to patch this up. Do note that your test changes makes tests only work on Go 1.24 which isn't ideal (and it is better to be consistent with the util.Exists function). I'd suggest moving the changes from my PR above and putting this onto the releases/v2.5.x branch.

adamdecaf:
Yea, your change makes sense. I figured pre-1.24 support would be needed in tests. 

I can bring the fix to that release branch if you'd like.

Jacalz:
Thanks. That sounds like a good plan. You can probably ignore the GOROOT changes from that PR. They are not necessary for the simple fix :+1:

PR #5552, opened 2025-02-25
implement async.Map with plain map for much improved performance with migrated_fynedo

dweymouth:
### Description:
The `sync.Map` type in Go is very slow for our use pattern. With migrated_fynedo we can use a plain map and assume single-threaded access, which gives a very significant performance improvement.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


internal/async/map_migratedfynedo.go
```diff
@@ -0,0 +1,71 @@
+//go:build migrated_fynedo
+
+package async
+
+// Map is a generic wrapper around [sync.Map].
+type Map[K any, V any] struct {
+	// once go1.20 is base, can use map[K]V
+	// with K being Comparable instead of any
+	// (CanvasObject, etc aren't Comparable for go1.19)
+	m map[any]V
```
Jacalz, line 10: This should be [K]V, right?

Jacalz (changes requested):
Great plan. I think the map type is wrong though.

internal/async/map_migratedfynedo.go
Jacalz: Ah, I missed the comment. Sorry. Might it make sense to instead just remove the methods of the map that require them to be comparable? I don’t think they are used in too many places for it to matter. Should be a good performance boost.

Jacalz: Or is it the amp itself that complains because it can't hash the values?

Jacalz:
Like I wrote on Slack, it is likely worth it to remove the `internal/cache/texture_*_test.go` files entirely as they assume the cache is thread safe. Doing so should remove a lot of strange race conditions from your tests.

andydotxyz (approved):
Nice optimisation!

Jacalz (approved):
This is a great change. Thanks. It is worth noting though that we need to be aware of anywhere we might want to use the `async` map in a place that always needs synchronization might result in crashes.

I updated https://github.com/fyne-io/fyne/issues/3242 to add the `map[any]V` problem there.

PR #5554, opened 2025-02-25
Fixing semantics of test canvas resize

### LINKED ISSUE #5548: `SoftwareCanvas` resize only works properly if it's the last step
### Checklist

### Describe the bug

Following up on the suggestion from Discord to report this here... If i take the following code:

```go
width := 1024
height := 758

myCanvas := playground.NewSoftwareCanvas()

clockText := canvas.NewText(time.Now().Format("15:04:05"), color.White)
clockText.TextSize = 150
clockText.Alignment = fyne.TextAlignCenter

clockContainer := container.NewCenter(clockText)

myCanvas.SetContent(clockContainer)
myCanvas.Resize(fyne.Size{Width: width, Height: height})

img := myCanvas.Capture()
```

And save the result as an image, I get the expected result:

Image

But if I move `myCanvas.Resize(fyne.Size{Width: width, Height: height})` directly after `myCanvas := playground.NewSoftwareCanvas()`, the result will be wrong:

Image

### How to reproduce

1. Copy my provided example code
2. `go run .`
3. See that the resulting `clock.png` has the wrong resolution, it should be 1024x758.
4. Put `myCanvas.Resize(fyne.Size{Width: width, Height: height})` under `myCanvas.SetContent(clockContainer)`.
5. Notice that the image is rendered properly.

### Screenshots

_No response_

### Example code

```go
package main

import (
	"fmt"
	"image"
	"image/color"
	"image/png"
	"os"
	"time"

	"fyne.io/fyne"
	"fyne.io/fyne/canvas"
	"fyne.io/fyne/container"
	"fyne.io/fyne/tools/playground"
)

func main() {
	width := 1024
	height := 758

	myCanvas := playground.NewSoftwareCanvas()
	myCanvas.Resize(fyne.Size{Width: width, Height: height})

	clockText := canvas.NewText(time.Now().Format("15:04:05"), color.White)
	clockText.TextSize = 150
	clockText.Alignment = fyne.TextAlignCenter

	clockContainer := container.NewCenter(clockText)

	myCanvas.SetContent(clockContainer)

	img := myCanvas.Capture()

	saveImage(img, "clock.png")
}

func saveImage(img image.Image, filename string) {
	f, err := os.Create(filename)
	if err != nil {
		fmt.Println(err)
		return
	}
	defer f.Close()

	err = png.Encode(f, img)
	if err != nil {
		fmt.Println(err)
		return
	}
}
```

### Fyne version

v2.5.4

### Go compiler version

1.24.0

Arch Linux amd64

### Additional Information

_No response_

andydotxyz:
Fixes #5548

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/list.go
+++ b/widget/list.go
@@ -762,7 +762,7 @@ func (l *listLayout) updateList(newOnly bool) {
 			l.setupListItem(vis.item, vis.id, l.list.focused && l.list.currentFocus == vis.id)
 		}
 
-		// a full refresh means drain the pool of excess old items as well
+		// a full refresh may change theme, we should drain the pool of unused items instead of refreshing them.
 		for l.itemPool.Get() != nil {
 		}
 	}
--- a/widget/list.go
+++ b/widget/list.go
@@ -761,6 +761,10 @@ func (l *listLayout) updateList(newOnly bool) {
 		for _, vis := range l.visible {
 			l.setupListItem(vis.item, vis.id, l.list.focused && l.list.currentFocus == vis.id)
 		}
+
+		// a full refresh means drain the pool of excess old items as well
+		for l.itemPool.Get() != nil {
+		}
 	}
 
 	// we don't need wasVisible now until next call to update
--- a/widget/tree.go
+++ b/widget/tree.go
@@ -766,6 +766,9 @@ func (r *treeContentRenderer) Objects() []fyne.CanvasObject {
 
 func (r *treeContentRenderer) Refresh() {
 	r.refreshForID(r.treeContent.nextRefreshID)
+	for _, s := range r.separators {
+		s.Refresh()
+	}
 }
 
 func (r *treeContentRenderer) refreshForID(toDraw TreeNodeID) {
--- a/test/canvas.go
+++ b/test/canvas.go
@@ -61,7 +61,6 @@ func NewCanvas() WindowlessCanvas {
 		focusMgr: intapp.NewFocusManager(nil),
 		padded:   true,
 		scale:    1.0,
-		size:     fyne.NewSize(100, 100),
 	}
 	c.overlays = &internal.OverlayStack{Canvas: c}
 	return c
@@ -223,7 +222,7 @@ func (c *canvas) SetContent(content fyne.CanvasObject) {
 	if c.padded {
 		minSize = minSize.Add(fyne.NewSquareSize(theme.Padding() * 2))
 	}
-	c.Resize(minSize)
+	c.Resize(c.Size().Max(minSize))
 }
 
 func (c *canvas) SetOnTypedKey(handler func(*fyne.KeyEvent)) {
--- a/test/canvas.go
+++ b/test/canvas.go
@@ -27,8 +27,9 @@ type WindowlessCanvas interface {
 }
 
 type canvas struct {
-	size  fyne.Size
-	scale float32
+	size    fyne.Size
+	resized bool
+	scale   float32
 
 	content     fyne.CanvasObject
 	overlays    *internal.OverlayStack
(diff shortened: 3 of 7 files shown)


widget/tree.go
```diff
@@ -766,6 +766,9 @@ func (r *treeContentRenderer) Objects() []fyne.CanvasObject {
 
 func (r *treeContentRenderer) Refresh() {
 	r.refreshForID(r.treeContent.nextRefreshID)
+	for _, s := range r.separators {
```
dweymouth, line 769: Why was this needed? I think this will lead to RefreshItem, and maybe scrolling, refreshing all separators always, which shouldn't be needed.

andydotxyz: Without it the resize after changing theme was picking up bad separators.

It seemed like a logical change because the `Refresh` of the content was being called and the items were being refreshed - so shouldn't the separators?

andydotxyz: (You can see that List has essentially the same - when the content `Refresh()` is called we loop through the separators and refresh them)

andydotxyz:
This change uncovered a bug in both tree and list refreshing (theme change then resize) which I think I have fixed appropriately...

widget/tree.go
dweymouth: OK - may be something that can be optimized later since it should only be needed for a full Refresh, and not RefreshItem, etc which will still go through this path. But as you point out it's the same as List for now

widget/list.go
```diff
@@ -761,6 +761,10 @@ func (l *listLayout) updateList(newOnly bool) {
 		for _, vis := range l.visible {
 			l.setupListItem(vis.item, vis.id, l.list.focused && l.list.currentFocus == vis.id)
 		}
+
+		// a full refresh means drain the pool of excess old items as well
```
dweymouth, line 765: Can we elaborate on the why here - that the old items may have an old theme applied? I can imagine this piece of code, and the current comment, becoming cryptic with time since it's not obvious why we'd be draining the pool. Also would be useful if we revisit this some time later with smarter refreshing to avoid having to do this.

andydotxyz: When we call Refresh it is iterating visible items to refresh them.
But the items in the cache are being left in old theme.
Is the sentence there just needing more "why" in it?

widget/tree.go
andydotxyz: Good point. I'm not too worried as a separator refresh is so minimal :)

widget/list.go
dweymouth: yeah I think the code comment needs the info about the old theme - so someone coming to this code later on after this PR is long forgotten knows why it was added

andydotxyz: OK thanks, I hope the update helps

widget/tree.go
Jacalz: I think we should open an issue to make sure we don't forget to do that optimization in the future

widget/list.go
```diff
@@ -761,6 +761,10 @@ func (l *listLayout) updateList(newOnly bool) {
 		for _, vis := range l.visible {
 			l.setupListItem(vis.item, vis.id, l.list.focused && l.list.currentFocus == vis.id)
 		}
+
+		// a full refresh may change theme, we should drain the pool of unused items instead of refreshing them.
+		for l.itemPool.Get() != nil {
```
Jacalz, line 766: I don't understand why this is the case. Is it not defeating the purpose of the pool if we do this for each refresh?

andydotxyz: The pool does not exist to make refresh fast it is to make scroll fast...
When `Refresh()` is called we have to re-draw all of the list items whether they are in the pool or not. So either a) drain the pool and don't refresh them or b) Redraw all the items not currently on screen.

I went with a - but b would be OK too if you have a strong feeling.

Jacalz: Makes sense. Let's go with this approach for now

Jacalz (approved):

PR #5556, opened 2025-02-26
Implement missing functionality in portal dialogs

Jacalz:
### Description:

Not many relevant changes in the dependency upgrade apart from less code duplication and allowing us to use a common formatter for the X11 window handle. Other than that, this PR also completes the functionality of portal dialogs to match the built in ones (supporting file filters and l18n).

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


dialog/file_xdg_flatpak.go
```diff
@@ -139,7 +139,7 @@ func convertFilterForPortal(fyneFilter storage.FileFilter) (list []*filechooser.
 		return nil, nil
 	}
 
-	out := &filechooser.Filter{Name: lang.L("Filter")}
+	out := &filechooser.Filter{Name: "Filter"}
```
andydotxyz, line 142: From the API docs I see filter Name is "User-visible name for the filter." - so I'm not sure about this - where will it be shown and/or who will translate it instead?

Jacalz: I removed the translation as I was unsure if I added it the right way and I didn't know if it was worth it with a new translation string just for the portal dialogs but I can definitely add it back if you prefer. I also tried to not set the field but that results in an error and no dialog being shown. The idea is to let the user have a set of filters that they can select like one for PNG, one for JPG and so on. There is a select widget that lets the user choose from multiple filters.

Jacalz: image

Jacalz: I added back the translations for the filter. Added the Swedish translation for it as well given that it is the same

andydotxyz:
Ah I understand. 

I think adding the generic "Filter" text (translated or otherwise) isn't great - user has no info at all.

Instead it would be cool if you could type-check the filters to see if it is file extension or mime and provide more helpful texts - perhaps with an "Unknown filter" fallback?

andydotxyz (changes requested):
More notes in the comments but I think that "Filter" shouldn't be used plain or translated, as it doesn't provide meaningful info to the user.

Jacalz:
Those are some good points. I have implemented more descriptive names and made them translatable. I can't quite create a filter with "Unknown filter" as name because we have no way to convert a `Match(string)` function into a list of globs unless we have access to the underlying slice of globs.

andydotxyz:
I like this implementation - but for the names won't user expect something more like "File extensions: .png, .jpeg" in the name?

Honestly I don't know because I haven't seen the dialogs in use, but I thought I would ask

Jacalz:
I agree that it would be perhaps more appropriate but can we sensibly do so? The slice can be of any length really and encoding any arbitrary amount of matches might be weird. Should we just show the first two or all? I think there are a lot of complications given that the developer doesn't give us a name of the file filter in the way our API is written at the moment 😅

andydotxyz:
I guess I don't know the answer to those questions - what are filter names like in other usages of this API and/or is there a specification?

Jacalz:
The problem is that I don't know either. That was why I ignored translating it as there are just to many variables to get right. The specification is at https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.FileChooser.html#org-freedesktop-portal-filechooser-openfile but it doesn't say much more.

Jacalz:
This is an example of how it looks in other apps: 
image

Jacalz:
Perhaps we can add this as is for now and then expland the filters on our end to have an API to give them names and have more than one filter for the user to choose from? That would achieve the same outcome as the picture above and I think there is an issue open for the "multiple filters" idea already.

andydotxyz:
I agree that there is space for an enhancement here that we can look at later. However I think that adding these strings isn't really helpful to our users - as is it does not tell me anything about what the filter is.

I'd rather not add translations if we don't intend to use them going forward. A smaller impact temporary solution would probably be to join the strings in the filter, perhaps limiting at a number (and appending an ellipsis?).

Jacalz:
I have updated the code now to concat the first three patterns together and add an ellipsis if there are more patterns in the filters than those three. See images below for the extension filter in fyne_demo set to show two and three items:

Screenshot From 2025-03-04 21-21-31

image

dialog/file_xdg_flatpak.go
```diff
@@ -107,33 +106,84 @@ func fileSaveOSOverride(d *FileDialog) bool {
...
+
+		name := formatFilterName(filter.MimeTypes, 3)
+		converted := &filechooser.Filter{Name: name, Rules: rules}
+		return []*filechooser.Filter{converted}, converted
+	}
+
+	return nil, nil
+}
+
+func formatFilterName(patterns []string, count int) string {
```
andydotxyz, line 178: I think this would benefit from a unit test.

I'm pretty sure that if you pass ["1", "2", "3"] and count=4 the output won't be correct.

andydotxyz (changes requested):
A quick test for formatting for the 3 main cases of count less than, equal to, and greater than filter length would be good. I think the last one isn't correct at the moment.

dialog/file_xdg_flatpak.go
Jacalz: Yup. Good catch. Fixed now

andydotxyz (approved):
Sweet thanks

PR #5559, opened 2025-02-26
Remove old windows mobile workaround fixed in Go 1.8

Jacalz:
### Description:

While we don't support the mobile simulator on Windows yet, I found some assembly code lying around fixing an issue that was fixed in Go 1.8  by https://github.com/golang/go/commit/9c02c75639b893cea6dbce1092d07e886ec5f44e. Let's avoid carrying unmaintained assembly code.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Good catch!

PR #5565, opened 2025-02-27
Feature/bindpreferencelists

andydotxyz:
This has annoyed me for a while that it was missing!

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/data/binding/preference.go
+++ b/data/binding/preference.go
@@ -22,6 +22,21 @@ func BindPreferenceBool(key string, p fyne.Preferences) Bool {
 		})
 }
 
+// BindPreferenceBoolList returns a bound list of bool values that is managed by the application preferences.
+// Changes to this value will be saved to application storage and when the app starts the previous values will be read.
+//
+// Since: 2.6
+func BindPreferenceBoolList(key string, p fyne.Preferences) BoolList {
+	return bindPreferenceListComparable[bool](key, p,
+		func(p fyne.Preferences) (func(string) []bool, func(string, []bool)) {
+			return p.BoolList, p.SetBoolList
+		},
+		func(a, b bool) bool {
+			return a == b
+		},
+	)
+}
+
 // BindPreferenceFloat returns a bindable float64 value that is managed by the application preferences.
 // Changes to this value will be saved to application storage and when the app starts the previous values will be read.
 //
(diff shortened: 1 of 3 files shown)


data/binding/preference.go
```diff
@@ -119,3 +195,79 @@ func (b *prefBoundBase[T]) checkForChange() {
...
+	}
+}
+
+func (b *prefBoundList[T]) replaceProvider(p fyne.Preferences) {
+	b.get, b.set = b.setLookup(p)
+}
+
+type internalPrefs = interface{ WriteValues(func(map[string]any)) }
+
+func bindPreferenceListComparable[T bool | float64 | int | string](key string, p fyne.Preferences,
```
Jacalz, line 237: If we are to be consistent with the other functions here, the one called "comparable" should just pass a hard-coded `func(T, T) bool` into the function without the "comparable" suffix. Doing the same here would avoid the code duplication of all the comparison functions you have above.

data/binding/preference.go
```diff
@@ -22,6 +22,21 @@ func BindPreferenceBool(key string, p fyne.Preferences) Bool {
 		})
 }
 
+// BindPreferenceBoolList returns a bound list of bool values that is managed by the application preferences.
+// Changes to this value will be saved to application storage and when the app starts the previous values will be read.
+//
+// Since: 2.6
+func BindPreferenceBoolList(key string, p fyne.Preferences) BoolList {
```
Jacalz, line 29: Should we really fill the API with new functions for each type? We could do a single `BindPreferenceList[T bool | float64 | int | string]()` and potentially expand that typeset in the future.

Jacalz (changes requested):
Thanks. Nice one. Just a few questions/suggestions inline.

data/binding/preference.go
Jacalz: See note in slack, we might even be able to just do `[T comparable]` and match all comparable types.

andydotxyz: I would rather complete the API as it exists already (as it was basically missing). We could have a generic version but as I doubt other items will be added in the future I don't think it would be pertinent to add a new pattern for this PR.

data/binding/preference.go
andydotxyz: Ah good catch thanks.

data/binding/preference.go
Jacalz: Yeah, I kind of agree. In general, I think we should focus on using generics now in these cases where it makes a lot of sense over flooding the API surface with a bunch of new functions but like you say, it should have been there from the start. Will approve 👍

Jacalz (approved):

PR #5570, opened 2025-02-28
Add SizeName property to label to control text size

### LINKED ISSUE #5561: Add ability to choose a text size for label widgets
### Checklist

The Label widget currently supports one text size only (size 14 in the default Fyne theme).

However, users may want to render text in different sizes in their apps. For example I made a mail client with Fyne and wanted to render the subject text with in a larger size then the body text. I ended up re-creating the label widget, so I can use the same API but have different sizes. (See example screenshot below). 

I am aware that a similar feature has been requested before and that this may be a bit of a controversial topic. But I think I can propose a solution that satisfies both the apparent wish for such a feature from the user community and and still ensure it would not break the theme.

## Example

A mail client with different text sizes:

Image

Yes, there are three alternatives, but they each have significant disadvantages.

## canvas.Text

First, one can create a text of any size with `canvas.Text`:

```go
x := canvas.NewText("Heading", theme.Color(theme.ColorNameForeground))
x.TextSize = 20
```

However, these objects require the user to manually integrate them into the current theme (padding, color, size, ...). It also lacks all the advanced API features of a label like for example data binding, which again the user has to add manually. Other important feature like truncation are also missing too.

In my opinion `canvas.Text` is a low-level primitive, that should not be used directly in an app, but instead for building custom widgets. I also think that new users looking for a solution might often end up using this primitive and then get poor results.

## widget.RichText with markdown

Second, one can use a `widget.RichText` with Markdown to create larger text sizes:

```go
widget.NewRichTextFromMarkdown("# Heading")
```

This is much better then using canvas.Text, since it has theming, truncation and wrapping. But it still lacks some important features of the label API, e.g. data binding.  The approach also feels a bit like a hacky.

### widget.RichText with segments

Third, one can construct a text with a custom size through `widget.RichText` and segments:

```go
widget.NewRichText(&widget.TextSegment{
	Style: widget.RichTextStyle{
		SizeName: theme.SizeNameHeadingText,
	},
	Text: "Heading",
})
```

This also is much better then using canvas.Text, since it also has theming, truncation and wrapping. This approach also does not feel like a hack. Seasoned Fyne developed will recognize this the same approach used by `widget.Label` under the hood. However, this approach is very verbose and probably not very accessible to Fyne beginners. The RichText API itself is also not that easy to work with for a single piece of text, since it is designed to work with larger texts consisting of multiple segments. e.g. changing the text required something like this:

```go
x.Segments[0].(*widget.TextSegment).Text = "Other text"
```
And the API also lacks comfortable features like data binding.

I would suggest adding the following function to the label API:

```
func NewLabelWithSize(text string, sizeName fyne.ThemeSizeName) *Label{
    ...
}
```

This function would create a new `widget.Label`, but only accept text related sizes,. Those currently are:

Name | Size
-- | --
SizeNameCaptionText|11
SizeNameHeadingText|24
SizeNameSubHeadingText|18
SizeNameText|14

This will ensure that any label created with this new function would still fit nicely into the current theme, since it would use the same text sizes, which are already used by other Fyne widgets, e.g. `widget.Card`.

In addition I would propose to add `SizeName` as a public field. This would enable users to use the new size feature together with data binding.

In case this proposal is met with positive feedback, I would also offer to implement it.

dweymouth:
### Description:
Allows for controlling the text size of a Label widget. Especially useful with selection support landing in 2.6 as the only other way to customize the size of selectable text would be a ThemeOverride container, which is heavyweight and verbose.

This mirrors the API that already exists for Hyperlink sizing.

Fixes #5561 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz (changes requested):
I think this really needs some tests, make sure the label is scaling appropriately, updates on change and that the selection follows it...

andydotxyz (changes requested):
Thanks for adding a test.

I'm surprised that the images are the same size - does this indicate that the MinSize function is not updating, or that the test isn't respecting it?

I assume that a label with larger or smaller text would have a MinSize adjustment accordingly?

ErikKalkoken:
tyvm for implementing this feature!

sorry for my late comment (literally just saw this). But wouldn't it be better to limit the user to text-related sizes? e.g. by defaulting to normal size when a non-text related size is specified?

dweymouth:
> tyvm for implementing this feature!
> 
> sorry for my late comment (literally just saw this). But wouldn't it be better to limit the user to text-related sizes? e.g. by defaulting to normal size when a non-text related size is specified?

you mean preventing something like `theme.SizeNameInputRadius` from being used? That's not possible, at least in a clean way, since both "text-related" and "non-text-related" size names have the same type, and are the same concept at the theme level. This also mirrors the API already on `widget.Hyperlink` for controlling size as well as what TextSegments in `widget.RichText` support

ErikKalkoken:
> > tyvm for implementing this feature!
> > sorry for my late comment (literally just saw this). But wouldn't it be better to limit the user to text-related sizes? e.g. by defaulting to normal size when a non-text related size is specified?
> 
> you mean preventing something like `theme.SizeNameInputRadius` from being used? That's not possible, at least in a clean way, since both "text-related" and "non-text-related" size names have the same type, and are the same concept at the theme level. This also mirrors the API already on `widget.Hyperlink` for controlling size as well as what TextSegments in `widget.RichText` support

ty for the clarification. did not realize that the RichText and Hyperlink API also allows it and in that regards it makes indeed more sense to keep it the same.

PR #5571, opened 2025-03-02
Finish localising the DateEntry and Calendar

### LINKED ISSUE #5141: Calendar and DateEntry are not fully localised
### Checklist

### Describe the bug

The first day of the week and the date format are static instead of matching the user's locale.

### How to reproduce

Use Calendar widget and switch to a locale that has Sunday first instead of Monday.

Use DateEntry and notice that the date format is the same as you switch from English to US English.

### Screenshots

_No response_

### Example code

n/a

### Fyne version

future pre-2.6 (not a landed feature yet)

### Go compiler version

all

all

### Additional Information

_No response_

andydotxyz:
Fixes #5141

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/lang/translations/base.en_US.json
+++ b/lang/translations/base.en_US.json
@@ -1,0 +1,4 @@
+{
+  "settings.date.format": "01/02/2006",
+  "settings.date.startweek": "Sunday"
+}
--- a/lang/locale.go
+++ b/lang/locale.go
@@ -41,9 +41,16 @@ func closestSupportedLocale(locs []string) fyne.Locale {
 }
 
 func localeFromTag(in language.Tag) fyne.Locale {
-	b, _ := in.Base()
-	r, _ := in.Region()
-	s, _ := in.Script()
+	b, s, r := in.Raw()
+	ret := b.String()
 
-	return fyne.Locale(b.String() + "-" + r.String() + "-" + s.String())
+	if r.String() != "ZZ" {
+		ret += "-" + r.String()
+
+		if s.String() != "Zzzz" {
+			ret += "-" + s.String()
+		}
+	}
+
+	return fyne.Locale(ret)
 }
--- a/lang/translations/base.en.json
+++ b/lang/translations/base.en.json
@@ -41,5 +41,8 @@
     "tuesday": "Tuesday",
     "tuesday.short": "Tue",
     "wednesday": "Wednesday",
-    "wednesday.short": "Wed"
+    "wednesday.short": "Wed",
+
+    "settings.date.format": "02/01/2006",
+    "settings.date.startweek": "Monday"
 }
--- a/widget/calendar.go
+++ b/widget/calendar.go
@@ -12,6 +12,8 @@ import (
 	"fyne.io/fyne/v2/theme"
 )
 
+const weekStartKey = "settings.date.startweek"
+
 // Declare conformity with Layout interface
 var _ fyne.Layout = (*calendarLayout)(nil)
 
@@ -85,20 +87,22 @@ func (c *Calendar) CreateRenderer() fyne.WidgetRenderer {
 }
 
 func (c *Calendar) calendarObjects() []fyne.CanvasObject {
+	offset := 0
+	switch lang.X(weekStartKey, "Monday") {
+	case "Saturday":
+		offset = 6
+	case "Sunday":
+	default:
+		offset = 1
+	}
+
 	var columnHeadings []fyne.CanvasObject
 	for i := 0; i < daysPerWeek; i++ {
-		j := i + 1
-		if j == daysPerWeek {
-			j = 0
-		}
-
-		t := NewLabel(shortDayName(time.Weekday(j).String()))
+		t := NewLabel(shortDayName(time.Weekday((i + offset) % daysPerWeek).String()))
 		t.Alignment = fyne.TextAlignCenter
 		columnHeadings = append(columnHeadings, t)
 	}
-	columnHeadings = append(columnHeadings, c.daysOfMonth()...)
-
-	return columnHeadings
+	return append(columnHeadings, c.daysOfMonth()...)
 }
 
 func (c *Calendar) dateForButton(dayNum int) time.Time {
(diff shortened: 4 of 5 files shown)


lang/translations/base.en_US.json
```diff
@@ -0,0 +1,4 @@
+{
+  "settings.date.format": "01/02/2006",
```
dweymouth, line 2: should these settings be loaded from a different file that is just associated with region, to avoid the combinatorial explosion of language * region? (e.g. es_US, fr_CA, etc)

andydotxyz: That is an interesting thought - do other localisation systems work that way? I thought the lookup was strictly hierarchical, i.e. en_GB_Latn falls back to en_GB falls back to en?

dweymouth: Not sure, but when testing it with my system set to Spanish (es_US), Monday was the first day of the calendar

dweymouth: I would suggest they could even start as hard-coded before we think of a better place to configure them (and if they even need to be configured in a text-based metadata file)

andydotxyz:
Thanks for adding the Sunday starts - but I don't think they all use US date format ...

dweymouth:
> Thanks for adding the Sunday starts - but I don't think they all use US date format ...

I have them using the day/month/year format?

andydotxyz:
So you do - I can't read sorry.

widget/locale.go
```diff
@@ -0,0 +1,104 @@
...
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/lang"
+)
+
+const (
+	localeDateFormatKey = "locale.date.format"
+	localeWeekStartKey  = "locale.date.startweek"
+)
+
+var localeSettings = map[string]map[string]string{
```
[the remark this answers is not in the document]
andydotxyz: This was included to be a replacement for the lan file lookups until a better solution was decided upon - you can see the suggestion from @dweymouth.

Yes there are many more efficient ways to pack this data in, but they would not work if we ever make this more configurable again through something like weblate. Optimising for storage in code seems premature with that in mind. Similarly using "0" to represent Sunday (or Monday) makes it hard to read if we move to such a file later.

andydotxyz:
> If you wish to add Swedish to the list then it is Monday and `time.DateOnly` that is correct.

Thanks for this - but given that Swedish is the language do you mean Sweden? (i.e. (*_SE_*)?

> This seems like a very memory inefficient data structure with quite a lot of duplicated data.

I have significantly reduced the memory requirement, hopefully this makes it good for now?

Jacalz:
> Thanks for this - but given that Swedish is the language do you mean Sweden? (i.e. (SE)?

Oh, yes of course. That's what I meant

widget/locale.go
```diff
@@ -0,0 +1,90 @@
...
+func getLocaleWeekStart() string {
+	return lookupLocaleSetting(localeWeekStartKey, lang.SystemLocale())
+}
+
+func lookupLocaleSetting(key string, l fyne.Locale) string {
+	region := ""
+	lang := l.LanguageString()
+	if pos := strings.Index(lang, "-"); pos != -1 {
+		region = strings.Split(lang, "-")[1]
+	}
```
Jacalz, line 81: I think this might be cleaner to do using `strings.Cut`.

widget/locale.go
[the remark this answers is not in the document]
Jacalz: Yeah, those are good points. I wonder if just using a struct instead of the inner map would be a good balance between the two?

Jacalz (changes requested):

widget/locale_test.go
```diff
@@ -7,17 +7,15 @@ import (
...
+	assert.Equal(t, "Monday", lookupLocaleSetting("").weekStartDay.String())
+	assert.Equal(t, "Monday", lookupLocaleSetting("en").weekStartDay.String())
+	assert.Equal(t, "Monday", lookupLocaleSetting("en-GB").weekStartDay.String())
+	assert.Equal(t, "Sunday", lookupLocaleSetting("en-US").weekStartDay.String())
+	assert.Equal(t, "Sunday", lookupLocaleSetting("es-US").weekStartDay.String())
 }
 
 func TestGetLocaleDateFormat(t *testing.T) {
-	assert.Equal(t, "02/01/2006", lookupLocaleSetting(localeDateFormatKey, ""))
-	assert.Equal(t, "02/01/2006", lookupLocaleSetting(localeDateFormatKey, "en"))
```
andydotxyz: Any reason these two were removed?

widget/locale.go
andydotxyz: Doesn't that require the second instance of the separator? It is optional in this instance.

widget/locale_test.go
dweymouth: I'd have to change the expected value to empty string, or else test via the `getLocaleDateFormat` function, which uses the actual system locale. That function now handles the "" -> defaultDateFormat substitution for table entries that don't have a date format set.

andydotxyz: I don't understand - pasting this line in passes as expected, but is not included in date format tests whereas it still is tested in the week start test:
```go
	assert.Equal(t, "02/01/2006", lookupLocaleSetting("en").dateFormat)
```

andydotxyz:
Tests finally passing again

widget/locale_test.go
dweymouth: You are right - I added them back. Those 2 are both hitting the default entry in the map which does explicitly specify the date format.

widget/locale.go
Jacalz: I figured it would be possible to do something like this to just use one strings API instead of two to achieve the same result. However, I do see now that the following code assumes that the region is the end of the file so it might be moot.

```
_, trailing, found := strings.Cut(lang, "-")
if found {
    region = trailing
}
```

Jacalz (approved):
Looking good 👍

PR #5579, opened 2025-03-08
load markdown images from origin when not a URL

hkparker:
### Description:

Markdown supports loading local images off disk.  Currently, in fyne, when in the web, this doesn't work, because the browser can't access the local file system like that.  However, what I've seen from other markdown implementations is that in the web, a "local" file means from the same origin as the markdown file.  For example `` in a markdown loaded from `fyne.io` should be interpreted as ``.

This change adds that behavior to Fyne's markdown parser.  This has been tested locally for a website I'm working on with Fyne.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/widget/markdown.go
+++ b/widget/markdown.go
@@ -3,16 +3,13 @@ package widget
 import (
 	"io"
 	"net/url"
-	"runtime"
 	"strings"
-	"syscall/js"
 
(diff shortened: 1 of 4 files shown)


andydotxyz (approved):
Excellent thanks for this

PR #5583, opened 2025-03-08
Add test to check built-in translations

sdassow:
### Description:

Fixes #(issue)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz (approved):
Nice one, thanks for this - somehow I missed that we could fail to load the defaults!

PR #5594, opened 2025-03-12
Allow validation errors on entries without interaction

### LINKED ISSUE #5590: Allow validation errors on entries without interaction
### Checklist

Sometimes validation is more then just checking the string of an input. In my case, I have a lot of validation done after the user has already entered some data which can be from just entering data, loading the application or manual / periodic checks. 

These validations are too heavy to be put in the `Validator` of an entry and so they are done in a different place. What I would like to achieve is to still make the entry look like it has a validation error. 

Currently it is not possible to achieve this. Entry does have `SetValidationError()`, but that doesn't show the error if there is not a validator set **and** if the entry is not interacted with (aka dirty). 

I have disussed the problem and possible solutions in the Discord, which resulted in the following:

An entry should have an optional flag (boolean) that can be set, something along the lines of `ShowValidationErrorImmediately` (I had `AlwaysShowValidationError`  in mind because we always want to show validation errors if there are any)

Setting this flag will allow validation errors to be triggered without interaction. This comes down to two situations:

1. The entry has a validator: It will run the validator instantly without user interaction and reflect if it has any validation errors.  If the entry does not have a validation error, it should show a checkmark because it passed the validator given.
2. The entry does not have a validator: It will reflect if a validation error has been set with `SetValidationError()`. If the entry does not have a validation error, it should not show a checkmark because it has no validator and thus behave like nothing happend. 

I think these changes will benefit people who use more complex validators to still let the user know which inputs to change, while also providing benefits to people who add validators to their entries.

kruitbosdotdev:
### Description:

Implements a way to show errors without user interaction as described in the issue mentioned below. The main goal was to implement this change without modifying existing tests, ensuring that current behavior remains intact. The default functionality continues to work as expected without setting `AlwaysShowValidationError`.

All the tests create their testbed from the ground up instead of expanding a single test and doing `SetText()` for example. This way the tests test the behaviour when the entry is rendered instead of relying on user interaction to tigger a `Refresh()`.

Fixes #5590 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -63,6 +63,10 @@ type Entry struct {
 	onValidationChanged func(error)
 	validationError     error
 
+	// When true, the entry will always display an error (if there is any) without user interaction.
+	// Validator is called when rendered
+	AlwaysShowValidationError bool
+
 	CursorRow, CursorColumn int
 	OnCursorChanged         func() `json:"-"`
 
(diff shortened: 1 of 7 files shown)


andydotxyz:
Thanks for this, unfortunately it came in after the feature freeze and we are now in full release mode.
This should be good to go after that, but will need a "Since: 2.7" line instead.

kruitbosdotdev:
Will do. I'll also take a look at the failing tests

andydotxyz (approved):
Thanks for getting this worked out - let's land it and get 2.7 work started :)

PR #5596, opened 2025-03-13
Focus the extended widget not embedded in a collection tap

andydotxyz:
Fixes an issue from discussion https://github.com/fyne-io/fyne/discussions/5591

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


dweymouth (approved):
LGTM, though I wonder if this issue exists for any other builtin widgets, and do we need to think of a way (later) to expose this functionality for 3rd party custom widgets as well? They can't access super or impl.

andydotxyz:
Good questions. I have at the back of my mind to do away with this, but it will take some experimentation.

PR #5599, opened 2025-03-13
change dialog callback order

hkparker:
### Description:

This changes the order of callback execution in dialogs so that the original callback is executed before any of the `SetOnClosed` functions.

Marking as a draft until I go through and look for any other dialog inconsistencies.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/dialog/base.go
+++ b/dialog/base.go
@@ -111,10 +111,10 @@ func (d *dialog) SetOnClosed(closed func()) {
 	originalCallback := d.callback
 
 	d.callback = func(response bool) {
-		closed()
 		if originalCallback != nil {
 			originalCallback(response)
 		}
+		closed()
 	}
 }
 
--- a/dialog/color.go
+++ b/dialog/color.go
@@ -91,7 +91,6 @@ func (p *ColorPickerDialog) createSimplePickers() (contents []fyne.CanvasObject)
 }
 
 func (p *ColorPickerDialog) selectColor(c color.Color) {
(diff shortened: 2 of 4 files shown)


dweymouth:
Were there bugs that are fixed by this PR? If so, we usually reference them in the description. (E.g. "Fixes #<issueNum>"), but this seems like a good change to me

hkparker:
Not that I'm aware of, this allows a new behavior that I was prevented from doing without extending dialog, but I didn't open a feature request for it first

dweymouth (approved):

andydotxyz:
Just to clarify (because the definition of "original callback" is referring to a variable name more than it is a feature).
Does this mean that the "action" callback should always be called before `OnClosed` for all dialogs now?

If so, I think that the colour dialog is now an odd one out.

andydotxyz (changes requested):
Marking as requiring changes referring to the clarification posted rather than any specific code change.

hkparker:
Yes, the intention is that the "action" happens before any `SetOnClosed` callbacks.

I took another look though color, took a second to wrap my head around what was going on there but I think I just pushed something up that will bring it in line with this new expectation.

hkparker:
Test failures appear to be flake

andydotxyz (approved):
Thanks for this. We probably need to document it better and I will add a note to our change notes for the upgrade too.

Issue #4634, opened 2024-02-12
Implement storage APIs for web driver


dweymouth:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [X] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

The storage APIs are currently unimplemented for web.

Related: #2734 

### Describe the solution you'd like to see.

Storage APIs implemented for web. Local storage will probably not be usable here due to size limitations.

dweymouth:
This could probably be done with the [IndexedDB API](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) which is suitable for storing larger amounts of data than the 5MB-10MB limit of localStorage

dweymouth:
This [comment](https://github.com/fyne-io/fyne/issues/2734#issuecomment-1939732591) from @kkoreilly may be relevant here

dweymouth:
It's probably more heavyweight (eg we don't need unix permissions, etc) and pulls in more dependencies than we want, but could be a good guide for how to build our own filesystem-like abstraction on indexeddb

kkoreilly:
Yep, if you're not trying to implement the stdlib os API, you should need a lot less code. Most of the logic in jsfs is implementing different os functions that aren't needed for simple file storage.

bilalshaikh292:
Hey, I am new to this Org . I just wanted to know that is this issue open and if anyone is already working on this.

andydotxyz:
Nobody was working on this - however it is part of the upcoming 2.5 release which is due imminently. So we are looking for a quick turnaround (i.e. days remaining) - is that possible?

bilalshaikh292:
@andydotxyz just got a need to work on a university Project. Sorry won't be able to do this. Thanks for support.

andydotxyz:
Let's not push it that far. No new APIs required so we should be able to pull it into 2.5.x (hopefully the .1 at the latest, but who knows...)

sdassow:
I'm working on this now.

andydotxyz:
Thanks for this @sdassow - we will get the bounty paid out after release.

PR #5603, opened 2025-03-15
Initial code for web driver storage API using IndexDB

sdassow:
### Description:

Fix for https://github.com/fyne-io/fyne/issues/4634

Still need to figure out how to do proper WASM testing.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:

No change in binary output. The WASM output is 514631 bytes bigger for the example that @andydotxyz shared, if that is of interest.

### DIFF
--- a/internal/repository/indexdb_file_wasm.go
+++ b/internal/repository/indexdb_file_wasm.go
@@ -1,0 +1,178 @@
+//go:build wasm
+
+package repository
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"syscall/js"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/storage"
+
+	"github.com/hack-pad/go-indexeddb/idb"
+)
+
+var (
+	blob       js.Value
+	uint8Array js.Value
+)
+
+func init() {
+	blob = js.Global().Get("Blob")
+	uint8Array = js.Global().Get("Uint8Array")
+}
+
+var _ fyne.URIReadCloser = (*file)(nil)
+var _ fyne.URIWriteCloser = (*file)(nil)
+
+type file struct {
+	db          *idb.Database
+	path        string
+	parent      string
+	isDir       bool
+	truncate    bool
+	isTruncated bool
+	parts       []interface{}
+	add         bool
+	isAdding    bool
+}
+
+func (f *file) Close() error {
+	return nil
+}
+
+func (f *file) URI() fyne.URI {
+	u, _ := storage.ParseURI(fileSchemePrefix + f.path)
+	return u
+}
+
+func (f *file) rwstore(name string) (*idb.ObjectStore, error) {
+	txn, err := f.db.Transaction(idb.TransactionReadWrite, name)
+	if err != nil {
+		return nil, err
+	}
+	store, err := txn.ObjectStore(name)
+	if err != nil {
+		return nil, err
+	}
+	return store, nil
+}
+
+func (f *file) Write(data []byte) (int, error) {
+	p := f.path
+	ctx := context.Background()
+
+	if f.truncate && !f.isTruncated || f.add && !f.isAdding {
+		store, err := f.rwstore("meta")
+		if err != nil {
+			return 0, err
+		}
+
+		f := map[string]interface{}{
+			"parent": f.parent,
+		}
+		req, err := store.PutKey(js.ValueOf(p), js.ValueOf(f))
+		if err != nil {
+			return 0, err
+		}
+
+		if _, err := req.Await(ctx); err != nil {
+			return 0, err
+		}
+	}
+
+	if f.truncate && !f.isTruncated {
+		store, err := f.rwstore("data")
+		if err != nil {
+			return 0, err
+		}
+		delreq, err := store.Delete(js.ValueOf(p))
+		if err != nil {
+			return 0, err
+		}
+		if err := delreq.Await(ctx); err != nil {
+			return 0, err
+		}
+		f.isTruncated = true
+	}
+
+	if f.add && !f.isAdding {
+		store, err := f.rwstore("data")
+		if err != nil {
+			return 0, err
+		}
+		req, err := store.Get(js.ValueOf(f.path))
+		if err != nil {
+			return 0, err
+		}
+		b, err := req.Await(ctx)
+		if err != nil {
+			return 0, err
+		}
+
+		bytes := b.Call("bytes")
+		outch := make(chan js.Value)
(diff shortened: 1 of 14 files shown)


dweymouth:
Did you consider using the `hackpadfs` implementation of a filesystem on top of IndexedDB? When I briefly investigated this I came across hackpadfs but wasn't sure we wanted to bring another large dependency into Fyne. (And if its Apache license was compatible). But if we're bringing in hackpad dependencies, why not go all the way and use hackpadfs?

sdassow:
> Did you consider using the `hackpadfs` implementation of a filesystem on top of IndexedDB? When I briefly investigated this I came across hackpadfs but wasn't sure we wanted to bring another large dependency into Fyne. (And if its Apache license was compatible). But if we're bringing in hackpad dependencies, why not go all the way and use hackpadfs?

Yes, I had a look and saw more layers of abstraction then I am comfortable with, and the complexity is beyond what is needed for the storage APIs as well. Just the go-indexdb driver and safejs module are reasonably small I think.

Apache 2.0 license seems compatible, as we have other (Fyne) dependencies with it already (checked with go-licenses):
```
fyne.io/systray,https://github.com/fyne-io/systray/blob/v1.11.0/LICENSE,Apache-2.0
github.com/fyne-io/image/ico,https://github.com/fyne-io/image/blob/v0.1.0/ico/LICENSE-reader,Apache-2.0
```

internal/repository/indexdb_file_wasm.go
```diff
@@ -0,0 +1,178 @@
...
+	b, err := req.Await(ctx)
+	if err != nil {
+		return 0, err
+	}
+
+	if b.IsUndefined() {
+		return 0, fmt.Errorf("idbfile undefined")
+	}
+
+	bytes := b.Call("bytes")
```
andydotxyz, line 169: When I run the code I get an error from JS that this is undefined:

`panic: syscall/js: Value.Call: property bytes is not a function, got undefined`

andydotxyz:
Sorry for the log of merge commits, that was not my intent. However I have added a commit (the last one there) which hooks this into the app documents store provider correctly, so things are actually working. The test app I have used is:

```go
package main

import (
	"io"
	"log"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/dialog"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.NewWithID("wasmstoretest")
	w := a.NewWindow("Hello")

	files := a.Storage().List()

	list := widget.NewList(func() int {
		return len(files)
	}, func() fyne.CanvasObject {
		return widget.NewLabel("File name")
	}, func(id widget.ListItemID, o fyne.CanvasObject) {
		o.(*widget.Label).SetText(files[id])
	})

	list.OnSelected = func(id widget.ListItemID) {
		r, err := a.Storage().Open(files[id])
		log.Println("E", err)

		data, _ := io.ReadAll(r)
		r.Close()

		dialog.ShowInformation("File content", string(data), w)
	}

	input := widget.NewEntry()
	w.SetContent(container.NewBorder(nil,
		container.NewBorder(nil, nil, nil,
			widget.NewButton("Add", func() {
				name := input.Text
				input.SetText("")

				w, err := a.Storage().Create(name)
				log.Println("ERR", err)
				io.WriteString(w, "Hello world, "+name)
				w.Close()

				files = a.Storage().List()
				list.Refresh()
			}), input), nil, nil, list),
	)

	w.ShowAndRun()
}
```

As noted in the comments it works in all except getting the file content.

If you can resolve that in the next day or two @sdassow we might even get this in to 2.6.0!

andydotxyz (changes requested):
As noted inline the file read does not seem to work. Otherwise this is now very close!

internal/repository/indexdb_file_wasm.go
sdassow: Oh, fun, browser compatibility issues. Works on Gecko, broken on Blink - fixed now.

dweymouth:
Also please make sure to add the results of 

> Check for binary size increases when importing new modules.

to this PR descrption. At least for a desktop build (and maybe mobile as well?) Ideally this should only affect the binary size of wasm apps I think

sdassow:
> I didn't give a super close review since I thought this was going for 2.7 given 2.6 is in feature freeze. But are you storing any stats about the files (length, created/modified time, etc)? For future compatability with new Fyne APIs it may be a good idea to store the same kind of metadata filesystems store so that if/when we add metadata capabilities to the storage APIs the historical storage files "out in the wild" wouldn't have to have empty/defaulted metadata

Added code to store file size and times now.

dweymouth:
Are all the testdata fixture updates a carry-over from rebasing? Or else how did this change the rendering of everything slightly?

sdassow:
> Are all the testdata fixture updates a carry-over from rebasing? Or else how did this change the rendering of everything slightly?

Yes - the storage API changes are only in `app/` and `internal/repository/`.

Jacalz:
I'm not saying that 500kB is too much of an increase but WASM is one place where size matters the most. Larger binaries make loading times a lot worse. Just a theoretical question: does a storage API backend for web really have to be this complicated?

andydotxyz:
> > Are all the testdata fixture updates a carry-over from rebasing? Or else how did this change the rendering of everything slightly?
> 
> Yes - the storage API changes are only in `app/` and `internal/repository/`.

You should be able to rebase the branch, sorry for somehow messing up the history.

andydotxyz:
Just had a quick look and this seems excellent now thanks.

Unfortunately somewhere an encoding was missed - possibly in the handling of the storage files, because they assumed file path before but we have to have URI here, so if you try to store a file with a space you get an error that URI path contains an invalid character.

Am running for a flight just now so can't look into it further. If anyone gets a chance great, otherwise I might have time later today.

andydotxyz:
I had a moment and a thought so I pushed it in - I hope that is suitable?
As far as I can see this is working well now.

Jacob's question is an interesting one though - is there a way we can (now or in the future) have a lighter implementation to avoid the extra download?

sdassow:
> I had a moment and a thought so I pushed it in - I hope that is suitable? As far as I can see this is working well now.
> 

Good catch, thanks!

> Jacob's question is an interesting one though - is there a way we can (now or in the future) have a lighter implementation to avoid the extra download?

Future, yes. Still have to iron out a bug or two and clean up, and not sure how long that will take.

andydotxyz (approved):
Excellent thanks so much for contributing this!

PR #5608, opened 2025-03-16
Move the around drop-shadow to light from above as per Material2

andydotxyz:
This was requested by a user who pointed out our shadows looked too much like a blur.
We should do more than just gradients but in the meantime we can make it look like the shadow "falls" down like in the M2 spec https://m2.material.io/design/environment/light-shadows.html#light.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


Jacalz (approved):
This looks great. Thanks :+1:

PR #5612, opened 2025-03-17
Add RowWrap layout

ErikKalkoken:
### Description:

Adds a new layout for dynamically arranging objects with same height in rows with wrapping. 

In my opinion this layout solves a common layout problem and would be a good addition the Fyne toolkit. For example it would allow to dynamically arranging a set of filter chips in multiple rows:

image

I went with "RowWrap", because it arranges objects in rows and wraps them. This layout is similar to the existing GridWrap layout, so a similar name makes sense (i.e. it arranges in rows instead of a grid). But I am of course open to other suggestions.

### Demo

Here is an example render:

[Screencast from 17.03.2025 15:39:08.webm](https://github.com/user-attachments/assets/3f16b2dc-af1a-47d7-b996-737fdd3889ed)

Code to reproduce the example:

```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/layout"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("RowWrap layout")
	c := container.New(
		layout.NewRowWrapLayout(),
		widget.NewButton("Alpha", nil),
		widget.NewButton("Bravo Charlie", nil),
		widget.NewButton("Delta", nil),
		widget.NewButton("Echo Foxtrott Golf Hotel", nil),
		widget.NewButton("India Juliet", nil),
		widget.NewButton("Kilo Limo Mike November", nil),
	)
	w.SetContent(c)
	w.Resize(fyne.NewSize(50, 50))
	w.ShowAndRun()
}

```

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### Open

- [x] Add tests for custom padding variant

### DIFF
--- a/layout/rowwrap.go
+++ b/layout/rowwrap.go
@@ -13,7 +13,7 @@ type rowWrapLayout struct {
 
 // NewRowWrapLayout returns a layout that dynamically arranges objects of similar height
 // in rows and wraps them dynamically.
-// Objects are separated with horizontal and vertical inner padding.
+// Objects are separated with horizontal and vertical padding.
 //
 // Since: 2.7
 func NewRowWrapLayout() fyne.Layout {
@@ -25,7 +25,7 @@ func NewRowWrapLayout() fyne.Layout {
 }
 
 // NewRowWrapLayoutWithCustomPadding returns a new RowWrapLayout instance
-// with custom horizontal and vertical inner padding.
+// with custom horizontal and inner padding.
 //
 // Since: 2.7
 func NewRowWrapLayoutWithCustomPadding(horizontal, vertical float32) fyne.Layout {
--- a/layout/rowwrap.go
+++ b/layout/rowwrap.go
@@ -6,7 +6,9 @@ import (
 )
 
 type rowWrapLayout struct {
-	rowCount int
+	rowCount          int
+	horizontalPadding float32
+	verticalPadding   float32
 }
 
 // NewRowWrapLayout returns a layout that dynamically arranges objects
@@ -16,11 +18,29 @@ type rowWrapLayout struct {
 //
 // Since: 2.7
 func NewRowWrapLayout() fyne.Layout {
-	return &rowWrapLayout{}
+	return &rowWrapLayout{
+		horizontalPadding: theme.Padding(),
+		verticalPadding:   theme.Padding(),
+	}
+}
+
+// NewRowWrapLayoutWithCustomPadding creates a new RowWrapLayout instance
+// with the specified paddings.
+//
+// Since: 2.7
+func NewRowWrapLayoutWithCustomPadding(horizontal, vertical float32) fyne.Layout {
+	return &rowWrapLayout{
+		horizontalPadding: horizontal,
+		verticalPadding:   vertical,
+	}
 }
 
 var _ fyne.Layout = (*rowWrapLayout)(nil)
 
+// MinSize finds the smallest size that satisfies all the child objects.
+// For a RowWrapLayout this is the width of the widest child
+// and the height of the first child multiplied by the number of children,
+// with appropriate padding between them.
 func (l *rowWrapLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
 	if len(objects) == 0 {
 		return fyne.NewSize(0, 0)
@@ -30,22 +50,23 @@ func (l *rowWrapLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
 		rows = 1
 	}
 	rowHeight := objects[0].MinSize().Height
-	var w float32
+	var width float32
 	for _, o := range objects {
 		size := o.MinSize()
-		if size.Width > w {
-			w = size.Width
+		if size.Width > width {
+			width = size.Width
 		}
 	}
-	s := fyne.NewSize(w, rowHeight*float32(rows)+theme.Padding()*float32(rows-1))
+	s := fyne.NewSize(width, rowHeight*float32(rows)+l.verticalPadding*float32(rows-1))
 	return s
 }
 
+// Layout is called to pack all child objects into a specified size.
+// For RowWrapLayout this will arrange all objects into rows of equal size.
 func (l *rowWrapLayout) Layout(objects []fyne.CanvasObject, containerSize fyne.Size) {
 	if len(objects) == 0 {
 		return
 	}
-	padding := theme.Padding()
 	rowHeight := objects[0].MinSize().Height
 	pos := fyne.NewPos(0, 0)
 	rows := 1
@@ -55,9 +76,9 @@ func (l *rowWrapLayout) Layout(objects []fyne.CanvasObject, containerSize fyne.S
 		}
 		size := o.MinSize()
 		o.Resize(size)
-		w := size.Width + padding
+		w := size.Width + l.horizontalPadding
 		if pos.X+w > containerSize.Width {
-			pos = fyne.NewPos(0, float32(rows)*(rowHeight+padding))
+			pos = fyne.NewPos(0, float32(rows)*(rowHeight+l.verticalPadding))
 			rows++
 		}
 		o.Move(pos)
--- a/layout/rowwrap.go
+++ b/layout/rowwrap.go
@@ -15,7 +15,7 @@ type rowWrapLayout struct {
 // Object visibility is supported.
 //
 // Since: 2.7
-func NewRowWrapLayout() *rowWrapLayout {
+func NewRowWrapLayout() fyne.Layout {
 	return &rowWrapLayout{}
 }
 
--- a/layout/rowwrap.go
+++ b/layout/rowwrap.go
@@ -11,16 +11,16 @@ type rowWrapLayout struct {
 	verticalPadding   float32
 }
 
-// NewRowWrapLayout returns a layout that dynamically arranges objects
-// with the same height in rows and wraps them as necessary.
-//
-// Object visibility is supported.
+// NewRowWrapLayout returns a layout that dynamically arranges objects of similar height
+// in rows and wraps them as necessary.
+// The height of the rows is determined by the tallest object and the same for all rows.
 //
 // Since: 2.7
 func NewRowWrapLayout() fyne.Layout {
+	p := theme.Padding()
 	return &rowWrapLayout{
-		horizontalPadding: theme.Padding(),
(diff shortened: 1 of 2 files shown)


layout/rowwrap.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+}
+
+var _ fyne.Layout = (*rowWrapLayout)(nil)
+
+// MinSize finds the smallest size that satisfies all the child objects.
+// For a RowWrapLayout this is the width of the widest child
+// and the height of the first child multiplied by the number of children,
+// with appropriate padding between them.
+func (l *rowWrapLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
```
dweymouth, line 44: For something with wrapping like this, (and Label/RichText etc with wrapping), the behavior needed IIRC is for the MinSize returned to depend on the current size of the object. If the object hasn't been Resized (IE its size is 0, 0), then it returns the MinSize as you currently calculate it. But if it has been resized to a width >= the min width, the MinSize height is based on how many rows it needs to lay things out *at the current width*. Otherwise, placing it as the child of another layout will always make it take up more height than needed.

Not sure how to do this with a layout though, given that they are stateless by design. I suppose in Layout you could remeber the size in the `rowWrapLayout` struct, and then reference it in MinSize

layout/rowwrap.go
```diff
@@ -0,0 +1,88 @@
...
+type rowWrapLayout struct {
+	rowCount          int
+	horizontalPadding float32
+	verticalPadding   float32
+}
+
+// NewRowWrapLayout returns a layout that dynamically arranges objects
+// with the same height in rows and wraps them as necessary.
+//
+// Object visibility is supported.
```
dweymouth, line 17: "Object visibility is supported" probably isn't needed, since all first-party layouts support object visibility, and as such it can just be assumed.

dweymouth (changes requested):

layout/rowwrap.go
```diff
@@ -0,0 +1,88 @@
...
+	rowHeight := objects[0].MinSize().Height
+	var width float32
+	for _, o := range objects {
+		size := o.MinSize()
+		if size.Width > width {
+			width = size.Width
+		}
+	}
+	s := fyne.NewSize(width, rowHeight*float32(rows)+l.verticalPadding*float32(rows-1))
+	return s
```
Jacalz, line 61: Cleaner to just return directly:
```suggestion
	return fyne.NewSize(width, rowHeight*float32(rows)+l.verticalPadding*float32(rows-1))
```

layout/rowwrap.go
```diff
@@ -0,0 +1,88 @@
...
+	if rows == 0 {
+		rows = 1
+	}
+	rowHeight := objects[0].MinSize().Height
+	var width float32
+	for _, o := range objects {
+		size := o.MinSize()
+		if size.Width > width {
+			width = size.Width
+		}
```
Jacalz, line 58: I think there is a `fyne.Max()` function you can use here instead.

layout/rowwrap.go
```diff
@@ -0,0 +1,88 @@
...
+	if len(objects) == 0 {
+		return fyne.NewSize(0, 0)
+	}
+	rows := l.rowCount
+	if rows == 0 {
+		rows = 1
+	}
+	rowHeight := objects[0].MinSize().Height
+	var width float32
+	for _, o := range objects {
```
Jacalz, line 54: This will add sizes of invisible objects as well. Is that expected?

Jacalz (changes requested):
I really like the idea of this. Just left a few minor suggestions inline before I noticed it was a draft PR but I'm attaching them anyway :)

layout/rowwrap.go
andydotxyz: Layouts don't have to be stateless (see Border), the thing they lack is any connection to the container they are laying out.

layout/rowwrap.go
ErikKalkoken: No. good point. I will correct that.

layout/rowwrap.go
ErikKalkoken: I spoke too soon as my attempt did not fully work. I have reverted back to the previous solution for now and will make another attempt.

andydotxyz:
Do try not to force-push after reviews, it means that every review has to be from scratch not incremental.

layout/rowwrap.go
andydotxyz: Can you comment on the status of this comment? I'm not entirely sure by looking at the force-pushed code sorry.

ErikKalkoken: @andydotxyz I tried to implement this comment as suggested by dwey. But I could not get it to work and have rolled back those changes completely with the force-push (sorry). All other comments have been implemented though, so this is currently the last open point.

Jacalz:
Generally yes, though I'd say that it makes sense to force push for changes in a PR that is still in draft. I only submitted my review because I didn't realise it being drafted at the time

layout/rowwrap.go
ErikKalkoken: @dweymouth I must confess I was a bit confused about your comment at first, because MinSize() was already returning the actual height based on the number of rows as calculated in Layout(). Nevertheless I have now extended it as per your suggestion to return both the actual width and height once it is known (i.e. after Layout() has run).

ErikKalkoken:
This PR is now finally ready for review. 

All review comments so far have been addressed. I am expecting the reviewers to resolve their issues if they agree.

layout/rowwrap.go
```diff
@@ -0,0 +1,107 @@
...
+
+type rowWrapLayout struct {
+	horizontalPadding float32
+	minSize           fyne.Size
+	verticalPadding   float32
+}
+
+// NewRowWrapLayout returns a layout that dynamically arranges objects of similar height
+// in rows and wraps them dynamically.
+// Objects are separated with horizontal and vertical inner padding.
```
Jacalz: This comment does not seem correct with regular padding being passed in?

layout/rowwrap.go
```diff
@@ -0,0 +1,107 @@
...
+func NewRowWrapLayout() fyne.Layout {
+	p := theme.Padding()
+	return &rowWrapLayout{
+		horizontalPadding: p,
+		verticalPadding:   p,
+	}
+}
+
+// NewRowWrapLayoutWithCustomPadding returns a new RowWrapLayout instance
+// with custom horizontal and vertical inner padding.
```
Jacalz: Same here but you are passing in a custom amount.

layout/rowwrap.go
```diff
@@ -0,0 +1,107 @@
...
+		s := o.MinSize()
+		maxW = fyne.Max(maxW, s.Width)
+		maxH = fyne.Max(maxH, s.Height)
+	}
+	return fyne.NewSize(maxW, l.minHeight(maxH, objCount))
+}
+
+func (l *rowWrapLayout) minHeight(rowHeight float32, rowCount int) float32 {
+	height := rowHeight*float32(rowCount) + l.verticalPadding*float32(rowCount-1)
+	return height
```
Jacalz:
```suggestion
	return rowHeight*float32(rowCount) + l.verticalPadding*float32(rowCount-1)
```

layout/rowwrap.go
```diff
@@ -0,0 +1,107 @@
...
+	for _, o := range objects {
+		if !o.Visible() {
+			continue
+		}
+		maxH = fyne.Max(maxH, o.MinSize().Height)
+	}
+	var minSize fyne.Size
+	pos := fyne.NewPos(0, 0)
+	rows := 1
+	isFirst := true
```
Jacalz, line 87: I think you can just remove this and do `i == 0` in the if-statement instead.

Jacalz (changes requested):
Thanks. Just a few minor suggestions inline.

layout/rowwrap.go
ErikKalkoken: Are you sure? If the first object is invisible, wouldn't we then not need to handle i==1 as first (or i=x+1 if the first x are invisible).

Jacalz: Good point. You're right. Sorry for the fuss

Jacalz (approved):
Nice one. Thanks 👍

layout/rowwrap.go
```diff
@@ -0,0 +1,106 @@
...
+	for _, o := range objects {
+		if !o.Visible() {
+			continue
+		}
+		objCount++
+		s := o.MinSize()
+		maxW = fyne.Max(maxW, s.Width)
+		maxH = fyne.Max(maxH, s.Height)
+	}
+	return fyne.NewSize(maxW, l.minHeight(maxH, objCount))
```
dweymouth, line 63: It looks to me like you're passing objCount instead of row count?

ErikKalkoken: I think those are the same in this context. This is the part where minSize is estimated. And the code counts the number of visible objects = number of rows. I could rename the variable to rowCount to make it more clear perhaps?

dweymouth: They are different. The current implementation will always return a MinSize whose height is big enough for each item to be on its own row - even if the current size of the container is large enough that multiple objects fit within a given row. This means it will always be given more vertical space than it needs when nested into another container.

For things with wrapping, like this, and Label / RichText with wrapping, the MinSize has to be based on the *current* width.

In other words, the layout needs to remember the last width from the last call to Layout, and then use that in the MinSize calculation to figure out how many rows it actually needs with the current width allocated to it.

dweymouth: Wait, OK, I may be confused, since it looks like you actually do this in the Layout implementation. And the MinSize code here will short circuit at line 50 most of the time. I think this is good.

layout/rowwrap.go
ErikKalkoken: @dweymouth Github is still showing a request for change from you, but I can not find it. Might have to do with my force push earlier. Are you maybe able to close this or is there maybe another open issue that I missed?

ErikKalkoken:
@dweymouth This PR is still technically blocked by your request for change, but I can not find that comment anymore and I think it should be already resolved. Could you please check and either point me to what is missing or remove the blocker?

dweymouth (approved):

PR #5618, opened 2025-03-22
Remove atomic.Pointer from fyne.CurrentApp()

Jacalz:
### Description:

The main benefit here is that someone using the race detector will know that they are using fyne.CurrentApp() from the wrong goroutine in a migrated app and it should be safe for non migrations given the fact that the app is set once only. Another benefit is of course that frequest app lookups should be much faster (not only are we avoiding the atomic lookup but fyne.CurrentApp() is now also inlineable).

I would argue that this would be good to add for v2.6.0 mainly due to the fact that it will make the race detector trigger on using the API from the wrong place.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/app.go
+++ b/app.go
@@ -1,9 +1,6 @@
 package fyne
 
-import (
-	"net/url"
-	"sync/atomic"
-)
+import "net/url"
 
 // An App is the definition of a graphical application.
 // Apps can have multiple windows, by default they will exit when all windows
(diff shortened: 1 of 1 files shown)


Jacalz:
Hmm. Test's with race detector seem to not like this but it is fine in the app :/

andydotxyz:
> Hmm. Test's with race detector seem to not like this but it is fine in the app. I don't know how to go about fixing this in a good way. The performance improvement of this would be so nice thanks to inlining :/

Perhaps on `ci` tag there could be safety around it (such as now?) That is not present when we do local runs?

andydotxyz:
I think that you'll need to update the windows tests to turn on the ci tag now that it's needed to pass race checks on actions?

Just out of interest - is this not an issue for running tests locally?

Jacalz:
Yes, that is a problem. This is fundamentally broken for all kind of tests unfortunately. I will have to think more about this and see if there is a possible solution. I don't like the idea of having to modify code just to make tests work correct when the code behaves correctly for apps.

PR #5622, opened 2025-03-24
Better solution for faster internal maps when migrated

Jacalz:
### Description:

This uses a lock that turns off when migrated instead of migrating the whole async map. Firstly, this avoids potentialy confusion with async.Map potentially not being safe when it should be. Secondly, it avoids the issue with generics not supporting comparable map keys on Go 1.19 so should result in a performance improvement for all maps.

TODO: Fix deadlock on startup for non-migrated app.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
Faster but uglier code. Let's hope that we get Go 1.20 as base soon at least so we don't have to use 'any' as the map key...

dweymouth:
> Faster but uglier code. Let's hope that we get Go 1.20 as base soon at least so we don't have to use 'any' as the map key...

Once 2.7 is released and we can scrap the legacy thread-safe code entirely and just use map without any abstraction layer.

Issue #5597, opened 2025-03-13
`Hide()` doesn't work at startup for widgets/containers


GuillaumeLescure:
### Checklist

- [x] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [x] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

Container `DocTabs` and its children should be hidden at startup if the method `Hide()` is called ... but it's not.

### How to reproduce

See the example code below.

### Example code

```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
	"fyne.io/fyne/v2/container"
)

func main() {
	a := app.New()
	w := a.NewWindow("Test")

	t := container.NewDocTabs()
	b := widget.NewButton("now it works", func() {t.Hide()})
	t.Append(container.NewTabItem("tab1", b))

	t.Hide() // <=== should be hidden at startup

	w.SetContent(t)
	w.ShowAndRun()
}
```

Like the comment says, it should be hidden at startup ... but it's only hidden when le button is clicked.

### Fyne version

v2.5.4

### Go compiler version

go1.22.12

### Operating system and version

linux/amd64 (Fedora 40)

### Additional Information

Maybe it's related to [#4453](https://github.com/fyne-io/fyne/issues/4453) and/or [#3434](https://github.com/fyne-io/fyne/issues/3434).

It seems to work on `widget.Button`.

andydotxyz:
This seems to be true of many widgets, just updating title.

andydotxyz:
Hold on, I have run some tests and can't see the sense in this.
The issue is "widgets at the top level will be shown when the window appears" - that does not sound like a bug on the surface.

What is the use-case for having a window where there is no content visible at all?

frenkoes:
> How can I start my application with hidden content?

This seems to work for me:

`t.Hide()` to `a.Lifecycle().SetOnEnteredForeground(t.Hide)`

andydotxyz:
So the desire is to open a blank window? In general hiding top level content makes no sense to me - but put your hidden content into a `container.NewStack()` which remains visible will solve your problem.

Seems like you're probably working on something bigger we don't understand though.

dweymouth:
While I agree I don't see the use case for opening a blank window with all content hidden, on the other hand I agree with this statement from @GuillaumeLescure 

[...]

PR #5624, opened 2025-03-25
Don't manage content visibility at window level

andydotxyz:
Fixes requested feature to have window visible with hidden top level content.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
See #5597

sdassow (approved):
Less side-effects are good in my experience, and widgets aren't hidden by default anyways, so this makes sense to me.

PR #5626, opened 2025-03-25
Don't scroll on FocusGained For Collections

### LINKED ISSUE #5605: After scrolling, first selection in a list jumps that item to the bottom of the container
### Checklist

### Describe the bug

After scrolling in a list and then selecting an item for the first time, that item scrolls to the bottom of the visual frame.  All subsequent selections, even after a scroll, behave as expected.

### How to reproduce

1. Go to Fyne Demo
2. Go to Collections
3. Go to List
4. Scroll down a good ways
5. Select an Item

### Screenshots

_No response_

### Example code

Fyne Demo

### Fyne version

Develop Branch

### Go compiler version

1.24.3

Arch Linux

### Additional Information

_No response_

okratitan:
Fixes #5605

### Description:
Collections widgets already scroll to items when they are selected via clicking, keys, or by calling the API.  They also scroll via normal scroll means such as clicking scrollbars and scrolling.  There is no need to scroll on FocusGained as it is duplicative of other scrolls that are happening/will happen.

Fixes #5605 

### Checklist:

- [ ] Tests included.
- [X] Tests all pass.

### DIFF
--- a/widget/gridwrap.go
+++ b/widget/gridwrap.go
@@ -114,7 +114,6 @@ func (l *GridWrap) CreateRenderer() fyne.WidgetRenderer {
 // Implements: fyne.Focusable
 func (l *GridWrap) FocusGained() {
 	l.focused = true
-	l.scrollTo(l.currentFocus)
 	l.RefreshItem(l.currentFocus)
 }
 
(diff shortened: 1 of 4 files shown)


andydotxyz (approved):
Given the zero item ID for focus is valid I guess this is the best we can do.
Though maybe using pointer for optional focus tracking could be useful.
Happy to land this to resolve the issue for pending release.

PR #5659, opened 2025-04-09
Add tests for svg render issue

sdassow:
### Description:

This adds a test to reproduce a problem with a circular mask image.

It comes with two images made with Inkscape by just cutting out a circle and exporting it,
only difference is that the "simplified" version is the after using the "simplify" option in Inkscape, which results in a different SVG path.

The error output of "circle mask normal" is what's interesting, as only a fraction of the circle is rendered.

### Checklist:

- [X] Tests included.
- [ ] Tests all pass.


Jacalz:
I wonder if this (partly) belong in our oksvg fork?

PR #5669, opened 2025-04-14
Correctly reset an image when resource goes to nil

andydotxyz:
An image optimisation triggered a bug where it could not be reset to nil after a resource was used

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/canvas/image.go
+++ b/canvas/image.go
@@ -133,8 +133,9 @@ func (i *Image) Refresh() {
 			return
 		}
 		rc = io.NopCloser(r)
-	} else if !i.previousRender {
+	} else if i.previousRender {
 		i.previousRender = false
+	} else {
 		return
 	}
 
--- a/canvas/image.go
+++ b/canvas/image.go
@@ -135,6 +135,9 @@ func (i *Image) Refresh() {
 		rc = io.NopCloser(r)
 	} else if i.previousRender {
 		i.previousRender = false
+
+		Refresh(i)
+		return
 	} else {
 		return
 	}
--- a/canvas/image.go
+++ b/canvas/image.go
@@ -70,6 +70,8 @@ type Image struct {
 	Translucency float64    // Set a translucency value > 0.0 to fade the image
 	FillMode     ImageFill  // Specify how the image should expand to fill or fit the available space
 	ScaleMode    ImageScale // Specify the type of scaling interpolation applied to the image
+
+	previousRender bool // did we successfully draw before? if so a nil content will need a reset
 }
 
 // Alpha is a convenience function that returns the alpha value for an image
@@ -131,7 +133,8 @@ func (i *Image) Refresh() {
 			return
 		}
 		rc = io.NopCloser(r)
-	} else {
+	} else if !i.previousRender {
+		i.previousRender = false
 		return
 	}
 
@@ -165,6 +168,7 @@ func (i *Image) Refresh() {
 		}
 	}
 
+	i.previousRender = true
 	Refresh(i)
 }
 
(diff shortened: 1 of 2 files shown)


canvas/image.go
```diff
@@ -131,7 +133,8 @@ func (i *Image) Refresh() {
 			return
 		}
 		rc = io.NopCloser(r)
-	} else {
+	} else if !i.previousRender {
+		i.previousRender = false
```
hkparker, line 137: What is this line doing?  Since we only enter this block if `i.previousRender` is already false, as far as I can tell

andydotxyz: Good catch, that was an optimisation I added which wasn't executed. Re-worked the else so it catches as expected.

hkparker: Cool, this bit makes sense to me now.  If the previous render was a success, but now there is nothing defined that we could render, we mark this current render as unsuccessful.

I don't understand how this field is being used for flow control though.  If there is something for us to render (if line 129 is true), or if there's nothing for us to render but the previous render was a success, then we arrive at line 142.  But there, if there is nothing we can render we skip that block and set this field to true again on line 172.  I'm still a bit confused as to how this field/block is causing the image to be set to nil.

It seems like what we want is to set `i.Image = nil` here then early return, but according to the passing test somehow this is already happening.

hkparker (commented):
waiting for clarification

canvas/image.go
andydotxyz: > It seems like what we want is to set i.Image = nil here then early return, but according to the passing test somehow this is already happening.

Thanks for this. Yes it won't fail the test because it was just an optimisation that as you say was not actually optimising much at all.
I have returned early with the key refresh which was why it was continuing before.

hkparker: Looking at it now, this early return makes it clear what's going on, but doesn't seem to optimize too much, since while we avoid running the check on line 145 which we already know will be false, but we're still calling `Refresh` either way.

If it were me I'd block it out a bit different, moving lines 145-175 to just above 136, so something like:

```go
func (i *Image) Refresh() {
    // 119-127...
    
    if i.File != "" || i.Resource != nil || i.Image != nil {
        // 130-135 ...
        if i.File != "" || i.Resource != nil {
            // 146-172 ...
        }
        i.previousRender = true
    } else if i.previousRender {
        i.previousRender = false
    } else {
        return
    }

    Refresh(i)
}
```

As far as I can tell, the above refactor is functionally identical to the current state of the PR, but looking at it this way it doesn't appear that there's a reason to keep track of `previousRender`.  So I might be missing something about why my snippet there isn't identical, or maybe this check isn't really saving us anything

andydotxyz: If you don't keep track of previousRender then as soon as all items are nil it will exit early and never call refresh (the original bug). 

We have to call refresh again after the items are set to nil, but I want to do it only one time to avoid unnecessary refreshes. As far as I can see, even using your refactored code, removing the previousRender check will mean a nilled image won't refresh.

hkparker: Ah I see that now!  Thanks.  Ok now that I see what's going on there this looks good to me.

dweymouth (approved):
Approve for the bug fix, but looking at the test I notice another thing I never noticed - that if you set a Resource on an image and then render it, it will set the public field Image to non-nil? That seems like a potentially confusing behavior, and we may want to address it later

sdassow (approved):

andydotxyz:
> Approve for the bug fix, but looking at the test I notice another thing I never noticed - that if you set a Resource on an image and then render it, it will set the public field Image to non-nil? That seems like a potentially confusing behavior, and we may want to address it later

It was added some time ago as a way to render images on the correct thread. I'm considering reverting it now that threading has been sorted in another way.

This bug fix stands either way, I can look at the other after.

PR #5692, opened 2025-04-21
Add storage function to delete recursively 

### LINKED ISSUE #5691: storage.Delete() fails to delete non-empty folders
### Checklist

### Describe the bug

Delete function will error out when trying to delete a folder that contains files/folders.

### How to reproduce

Create a folder, within that folder create a file.
Then try to delete it with storage.Delete

### Screenshots

_No response_

### Example code

package main

import (
	"fmt"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"

	"fyne.io/fyne/v2/storage"
)

func main() {
	a := app.NewWithID("delete-recursive-test")

	createTestData(a)

	deleteTestData(a)
}

func createTestData(a fyne.App) {
	// create empty folder
	path, _ := storage.Child(a.Storage().RootURI(), "EmptyFolder")
	_ = storage.CreateListable(path)

	// create folder with data
	path, _ = storage.Child(a.Storage().RootURI(), "DataFolder")
	_ = storage.CreateListable(path)

	path, _ = storage.Child(path, "data.txt")
	file, _ := storage.Writer(path)
	file.Write([]byte{})

}

func deleteTestData(a fyne.App) {
	path, _ := storage.Child(a.Storage().RootURI(), "EmptyFolder")
	err := storage.Delete(path)

	fmt.Println(err)

	path, _ = storage.Child(a.Storage().RootURI(), "DataFolder")
	err = storage.Delete(path)

	fmt.Println(err)
}

### Fyne version

2.6.0

### Go compiler version

1.24.1

Linux

### Additional Information

_No response_

rad756:
### Description:

This tries to fix the below error, instead of modifying the delete method itself, this adds DeleteRecursive method to fyne storage to keep backwards compatibility (to possibly not break prior programs).

Fixes #5691 storage.Delete() fails to delete non-empty folders by adding a function to delete all children of a URI.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
> I'd personally say that it is a bug and better to fix than adding a specific recursive function.

I'm not sure about that - the docs refer to removing a resource, and if you compare it to the Go `os.Remove` it is consistent.

If we are looking at them for inspiration then maybe `RemoveAll` would be more familiar than `RemoveRecursive`?

rad756:
Could there be an agreed path forward and I will try to implement a fix that way? Just do not want to put in time in something that will be rejected or asked to do another way.

rad756:
So just rename the method to DeleteAll?

charlesdaniels:
The code does appear correct, although I think with some care you could write it to use iteration instead of recursion which might be better from a performance perspective with very deep file trees. I think Go does support tail call recursion, so if you prefer recursion stylistically, you should probably make sure your code is being correctly picked up by the compiler for that optimization.

From a design perspective, I do think that separating "remove a scalar item" from "remove an entire tree of items" makes sense, and many existing APIs make this distinction.

* Go: distinguishes between [`os.Remove()`](https://www.man7.org/linux/man-pages/man1/rm.1.html) and [`os.RemoveAll()`](https://www.man7.org/linux/man-pages/man1/rm.1.html)
* POSIX shell: [`rm(1)`](https://www.man7.org/linux/man-pages/man1/rm.1.html) separates single vs recursive deletion with the `-r` flag.
* POSIX C: does not support recursive deletion directly; [`unlink(2)`](https://man7.org/linux/man-pages/man2/unlink.2.html) "deletes" a "file" and [`rmdir(2)`](https://man7.org/linux/man-pages/man2/rmdir.2.html) deletes an empty directory.
* Python 3: [`os.remove()`](https://docs.python.org/3/library/os.html#os.remove) is similar to POSIX `unlink(2)`, [`os.rmdir`](https://docs.python.org/3/library/os.html#os.rmdir) behaves similarly to POSIX `rmdir(2)`. [`shutil.rmtree()`](https://docs.python.org/3/library/shutil.html#shutil.rmtree) provides a helper that acts similarly to Go's `os.RemoveAll()`.

Those are the languages and tools I'm personally most familiar with, but all of them (except POSIX C) support a distinction between "delete a thing", and "recursively delete many things". Given that Fyne is a Go library, I think that mirroring the Go `os` package and calling it either `RemoveAll()` or `DeleteAll()`. In retrospect, perhaps it would have been better to call [`Delete()`](https://github.com/fyne-io/fyne/blob/49eb265b399d2b28e546d2a26d6c67fdf946d239/storage/uri.go#L199) `Remove()`, but it's already part of the public API.

In the interest of sticking with our existing nomenclature (`Delete()`) while also staying in keeping with the Go standard library conventions, I think I would advocate for `DeleteAll()`, although I will defer the (currently active) members of the core team for a final decision.

storage/uri.go
```diff
@@ -211,6 +211,42 @@ func Delete(u fyne.URI) error {
 
 }
 
+func DeleteRecursive(u fyne.URI) error {
```
charlesdaniels, line 214: P.S. you should write a documentation comment for this 😉

Jacalz:
Good point about using iteration instead of recursion. It is likely easier than trying to make sure tail call recursion applies

Jacalz:
Good point about using iteration instead of recursion. It is likely easier than trying to make sure tail call recursion applies

rad756:
Besides renaming it to DeleteAll() and adding a comment (will try to follow fyne style). Do I need to change it to iterative approach or will recursive be fine?

rad756:
Seems to me that the recursive approach is  simpler and it already seems to be working correctly, at least in my limited testing

Jacalz:
> Besides renaming it to DeleteAll() and adding a comment (will try to follow fyne style)


As mentioned above, please also make sure to target the develop branch and try to tick the (by performing the corresponding tasks if necessary) checkboxes in the PR template. Adding tests for this would also be good.

storage/uri.go
```diff
@@ -211,6 +211,42 @@ func Delete(u fyne.URI) error {
...
+	list, err := List(u)
+
```
Jacalz, line 216: Please remove lines like this between the function call and checking errors. This isn't idiomatic Go code. Same applies in more places.

storage/uri.go
```diff
@@ -211,6 +211,42 @@ func Delete(u fyne.URI) error {
...
+	if err != nil {
+		return err
+	}
+
+	for _, v := range list {
+		listable, err := CanList(v)
+		if err != nil {
+			return err
+		}
+		if !listable {
```
Jacalz, line 226: I would prefer a new line between these.

storage/uri.go
```diff
@@ -211,6 +211,42 @@ func Delete(u fyne.URI) error {
...
+				return err
+			}
+			continue
+		} else {
+			err = DeleteRecursive(v)
+
+			if err != nil {
+				return err
+			}
+		}
```
Jacalz, line 238: You can remove the else statement due to the continue call.
```suggestion
			continue
		}

		err = DeleteRecursive(v)
		if err != nil {
			return err
		}
```

Jacalz (changes requested):
I have no strong opinion on iterative vs recursive but doing an iterative solution certainly wouldn't hurt. You avoid potentially crashing due to running out of stack space for long folders (as each recursive call increases the stack) and performance should likely be a lot better.

I left a few stylistic comments on the code inline.

storage/uri.go
```diff
@@ -211,6 +211,42 @@ func Delete(u fyne.URI) error {
...
+	}
+
+	err = Delete(u)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
```
Jacalz, line 247: No need to check the value and return the same value.
```suggestion
	return Delete(u)
```

Jacalz (changes requested):

rad756:
Could you please tell me if the following psuedo code will be sufficient or I am missing something
Check if the delete target is file or folder
If file delete as normal
If folder && empty > delete as normal
If folder and its not empty:
find all paths of files within all subdirectories
delete all files
delete the top folder **(OR would I have to delete the folders from the bottom of the folder structure, going up?)**

andydotxyz:
I would think a little recursion could clear this up. instead of:

```
If folder && empty > delete as normal
If folder and its not empty:
find all paths of files within all subdirectories
delete all files
```

Just this should work:

```
If folder 
delete all child items <- this item would descend depth first through the hierarchy I would think...
delete the folder
```

rad756:
Just to clarify, would I have to only delete the files in the subfolder and then be able to delete the target folder
OR
would I have to delete files of the subdirectory first, then delete subdirectory itself, then the target folder?

andydotxyz:
> Just to clarify, would I have to only delete the files in the subfolder and then be able to delete the target folder OR would I have to delete files of the subdirectory first, then delete subdirectory itself, then the target folder?

I would think the latter almost certainly. When I said "all child items" I meant files and folders. Then the containing folder can delete.

rad756:
EDIT: Corrected code block
Am I on the right path here and is this on track to be in acceptable state if I properly implement the below
```
func Delete(u fyne.URI) error {
	listable, err := CanList(u)
	if err != nil {
		return err
	}

	// will delete if target is not a folder OR will delele if it is an empty folder
	if !listable {
		repo, err := repository.ForURI(u)
		if err != nil {
			return err
		}

		wrepo, ok := repo.(repository.WritableRepository)
		if !ok {
			return repository.ErrOperationNotSupported
		}

		return wrepo.Delete(u)
	} 

	list, err := List(u)
	if err != nil {
		return err
	}

	// will delele if it is an empty folder
	if list == nil {
		repo, err := repository.ForURI(u)
		if err != nil {
			return err
		}

		wrepo, ok := repo.(repository.WritableRepository)
		if !ok {
			return repository.ErrOperationNotSupported
		}

		return wrepo.Delete(u)
	}

	// Will delete all contents of target folder then delete the folder itself
}
```
I will then iteritively get paths of all files in all subdirectories and paths of all subdirectories themselves. Will delete the files in reverse order (staring from the deepest paths going up) and then do the same for the subdiectories and then delete the target folder itself.

rad756:
I re-wrote (massively expanded) the Delete method to be able to handle cases were its asked to delete a folder containing subdirectories and/or files.

I ensured that it works with a small test script. 

Since this is my first pull not only to Fyne but overall in git, I assume there will be things that need to be improved. This is more meant as a proof of concept that this works and with feedback would be glad to refactor (hopefully using the right term) this when time permits.

rad756:
Also, in case of error happening in the middle of deleting a non empty folder, it might be a good idea to include a specific error about this, that the contents of the folder could have been partially deleted.

rad756:
Refactored commit to remove repeated code and added comments

storage/uri.go
```diff
@@ -197,6 +197,78 @@ func Exists(u fyne.URI) (bool, error) {
...
+		return err
+	}
+
+	// Will delete if target is not a folder
+	if !listable {
+		return delete(u)
+	}
+
+	// Makes list of all contents of the folder
+	list, err := List(u)
```
Jacalz: If you name `list` as something more deskriptive, the rest of the code will be easier to reason about. How about `folderContents`?

storage/uri.go
```diff
@@ -197,6 +197,78 @@ func Exists(u fyne.URI) (bool, error) {
...
+	if err != nil {
+		return err
+	}
+
+	// Will delele if it is an empty folder
+	if list == nil {
```
Jacalz: Per above comment, doing `if len(folderContents) == 0` is easier to follow than whatever a `nil` list might mean (if you don't know what the function that returns it does).

storage/uri.go
```diff
@@ -197,6 +197,78 @@ func Exists(u fyne.URI) (bool, error) {
...
+		return delete(u)
+	}
+
+	// Will delete all contents of target folder then delete the folder itself
+	queue := list
```
Jacalz: This likely won't be necessary with the better name and a suggestion below.

storage/uri.go
```diff
@@ -197,6 +197,78 @@ func Exists(u fyne.URI) (bool, error) {
...
+	var folders []fyne.URI
+	var files []fyne.URI
+
+	// Lists paths of all sub-folders and files
+	for len(queue) > 0 {
```
Jacalz: I think you can just iterate forward and check for length of the contents without having to slice off the queue.

storage/uri.go
```diff
@@ -197,6 +197,78 @@ func Exists(u fyne.URI) (bool, error) {
...
+			queue = append(queue, list...)
+		} else {
+			files = append(files, currentPath)
+		}
+	}
+
+	// Deletes all the files inside the folder
+	for len(files) > 0 {
+		fileToDelete := files[len(files)-1]
+		files = files[:len(files)-1]
```
Jacalz: This would be easier to read and more compact if you just used a range loop.

storage/uri.go
```diff
@@ -197,6 +197,78 @@ func Exists(u fyne.URI) (bool, error) {
...
+		err = delete(fileToDelete)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Deletes all the sub-folders
+	for len(folders) > 0 {
+		folderToDelete := folders[len(folders)-1]
+		folders = folders[:len(folders)-1]
```
Jacalz: Same as above.

Jacalz:
I also think the current approach doesn't add `RemoveAll` as the function name that we agreed upon above?

storage/uri.go
rad756: I made it this way since GO does not have a while keyword. This is basically while the queue is not empty do this.

I did it this way since I need to check each subdirectory depth separately and add its contents to the queue. This is the only way I figured how to perform this, _at least so far_

storage/uri.go
rad756: Would range still work correctly if I edit the range while still being in the for loop?

rad756:
@Jacalz I can refactor it back to being a singular Delete() and having a DeleteAll() in addition to this but it would be a little later in the week, I will also make the renamings of queue and list. If you could reply back to the above conversations I would be grateful.

storage/uri.go
Jacalz: I don't know but there is no point in editing the slice in this nor the loop below. Just iterate through and delete for each element

storage/uri.go
Jacalz: Yeah, what you have written is basically a while loop but I think my suggestion above should work.

Surely `for i := 0; i < len(folderContents); i++` should work even if you are appending to it while iterating?

storage/uri.go
rad756: Now I understand your comment. I have to delete the items in reverse order, meaning deepest first otherwise it will error out that the folder is not empty

Jacalz: Just iterate backwards then using a regular for-loop and indexing. That you avoid the need of subslicing

Jacalz: I can see that folders would need to be iterated over backwards but why files?

storage/uri.go
andydotxyz: `children` or `childItems` might be more indicative of a recursive descent.

andydotxyz (commented):
The complexity of ordering and modifying slices inline seems to be unnecessary.
If the `delete` (or `deleteAll`) had the logic of handling listable items then this would descend recursively and you would not have to assemble the list before deleting it.

rad756:
Reverted the Delete method back to original, created DeleteAll method along with all the changes made that were requested

storage/uri.go
rad756: Files don't need to be iterated backwards, I think I just made both the same since it was easier and make it look more uniform

andydotxyz:
It took me a long time to understand this code, but I now see how it works.

However won't it be wasteful to track multiple copies of the data and to constantly grow the children, files and folders lists in preparation for a single delete loop.
I'm not sure why this is preferable to the more efficient, and simpler, recursive solution:

```go
func DeleteAll(u fyne.URI) error {
	listable, err := CanList(u)
	if err != nil {
		return err
	}

	if !listable {
		return Delete(u)
	}

	children, err := List(u)
	if err != nil {
		return err
	}

	for _, child := range children {
		err = DeleteAll(child)
		if err != nil {
			return err
		}
	}

	return Delete(u)
}
```

p.s. Remember you'll need to add documentation to the method and a "Since: 2.7" comment as well

p.p.s. I think this is still against Fyne `master` branch and not `develop`.

andydotxyz:
Also, don't forget a unit test. I knocked one up myself that might work OK for you...

```go
func TestDeleteAll(t *testing.T) {
	m := intRepo.NewInMemoryRepository("uritest")
	repository.Register("uritest", m)
	m.Data["/foo/"] = []byte{}
	m.Data["/foo/bar/"] = []byte{}
	m.Data["/foo/bar/baz/"] = []byte{}
	m.Data["/foo/bar/baz/more/"] = []byte{}
	m.Data["/foo/bar/baz/more/test.txt"] = []byte("test.txt content")

	foo, _ := storage.ParseURI("uritest:///foo/")
	data, _ := storage.ParseURI("uritest:///foo/bar/baz/more/test.txt")

	fooExists, err := storage.Exists(foo)
	assert.True(t, fooExists)
	assert.NoError(t, err)
	bazExists, err := storage.Exists(data)
	assert.True(t, bazExists)
	assert.NoError(t, err)

	err = storage.DeleteAll(foo)
	assert.NoError(t, err)

	fooExists, err = storage.Exists(foo)
	assert.False(t, fooExists)
	assert.NoError(t, err)
	bazExists, err = storage.Exists(data)
	assert.False(t, bazExists)
	assert.NoError(t, err)
}
```

rad756:
@andydotxyz I reimplemented it in iterative approach as Charles and Jackalz suggested it would be better.

Also you need to delete the files first and then delete the folders starting from deepest first to ensure you are deleting empty folders, as otherwise it will throw an error that it cannot delete because the folder has contents.

So can we agree in which style this needs to be implemented in as I do not want to keep changing it between recursive and iterative methods every few weeks.

andydotxyz:
> @andydotxyz I reimplemented it in iterative approach as Charles and Jackalz suggested it would be better.

Very well, I missed that sorry. In which case I'll leave it to them to review because I don't understand why the more complex code is more desirable.

> Also you need to delete the files first and then delete the folders starting from deepest first to ensure you are deleting empty folders.

Yes, the simple algorithm I posted above is a complete depth-first deletion - no need to differentiate folders vs files or maintaining any order.

> So can we agree in which style this needs to be implemented in as I do not want to keep changing it between recursive and iterative methods every few weeks. 

Of course, as above I'll leave it to those that requested this approach to review it.

andydotxyz:
I have rebased this existing PR to develop to assist. It's not perfect but probably good enough.
Losing all the context above with a new PR seems like it would be unfortunate.

rad756:
Hi, what still needs to be done to merge this into fyne?

Jacalz:
Sorry. I've been away and rather busy most of the summer. This is ended up far down the inbox and I forgot about it. I'll try to have a look when I'm back home next week.

storage/uri.go
```diff
@@ -208,7 +208,59 @@ func Delete(u fyne.URI) error {
 	}
 
 	return wrepo.Delete(u)
+}
+
+func DeleteAll(u fyne.URI) error {
```
Jacalz, line 212: All functions should have a function comment and a `Since:` line.

storage/uri.go
```diff
@@ -208,7 +208,59 @@ func Delete(u fyne.URI) error {
...
+	}
+
+	if len(children) == 0 {
+		return Delete(u)
+	}
+
+	var folders []fyne.URI
+	var files []fyne.URI
+	for i := 0; i < len(children); i++ {
+		listable, _ := CanList(children[i])
```
Jacalz: Should this error not be checked?

.github/FUNDING.yml
```diff
@@ -1 +1 @@
-github: [fyne-io, andydotxyz, toaster, Jacalz, changkun, dweymouth, lucor]
+github: [fyne-io, andydotxyz, Jacalz, dweymouth, toaster, lucor]
```
Jacalz: You might have to rebase your branch on the latest develop. These changes seem unrelated.

storage/uri.go
rad756: I added a small comment and since 2.7

storage/uri.go
rad756: Returned error.

rad756:
Made a small commit to hopefully resolve the issues listed above. The added comment might be too short, some guidance on proper commenting would be appreciated.

storage/uri.go
Jacalz: Thanks. It might make sense to mention that it removed any child resources if it is listable. I don't know how to formulate that in a good way.

storage/uri.go
```diff
@@ -208,7 +208,67 @@ func Delete(u fyne.URI) error {
...
+	if len(children) == 0 {
+		return Delete(u)
+	}
+
+	var folders []fyne.URI
+	var files []fyne.URI
+	for i := 0; i < len(children); i++ {
+		listable, err = CanList(children[i])
+
+		if err != nil {
```
Jacalz:
```suggestion
		listable, err = CanList(children[i])
		if err != nil {
```

rad756: I added a comment above DeleteAll func, is it not visible?

// DeleteAll destroys, deletes, or otherwise removes the resource referenced
// by the URI even if the resource is a non-empty directory.

I thought that the even if part would make it clear that it also deletes folders and all their contents

rad756:
I would appreciate if you Jacalz fixed it as a follow up. That might be a bit out of my depth to tackle as this is my first merge request not only to Fyne but any repo.

Is this in a state that it could make its way into next develop release?

storage/uri.go
Jacalz: I think you added the comment in the wrong thread. All I'm saying here is that you shouldn't add a new line between the error check.

storage/uri.go
Jacalz: > I added a comment above DeleteAll func, is it not visible?

// DeleteAll destroys, deletes, or otherwise removes the resource referenced
// by the URI even if the resource is a non-empty directory.

I thought that the even if part would make it clear that it also deletes folders and all their contents

It is visible. What I'm saying is that I don't think "the resource" captures the intent of what the function does as it removes more than one resource. To instead say that it deletes all of the children of a resource would be clearer.

Jacalz:
> I would appreciate if you Jacalz fixed it as a follow up. That might be a bit out of my depth to tackle as this is my first merge request not only to Fyne but any repo.

Sure thing. I'll tidy that up given that it is your first PR. Sorry if you think this has been a long road to get through.

> Is this in a state that it could make its way into next develop release?

The changes requested are far from big, it's just some minor stuff. With that fixed and the conflict resolved, I can merge it.

storage/uri.go
rad756: It actually won't let me comment on the above thread for some reason that's why I stated it here, but if having no line between making err and checking it is the correct way or more go like, I will do it this way going forward

rad756:
I wasn't expecting it to be a short easy road, it wasn't that bad.
Thanks for your reviews and suggestions Jacalz.

storage/uri.go
Jacalz: It is idiomatic Go, basically how all existing lines in this project are set up and how most other are as well

Jacalz:
I tidied this up to make sure it conforms to existing repository functionality where repositories can provide their own implementations of specific functions. The generic repository delete all function that @rad756 implemented is now in the `storage/repository` package. Only tests remain now.

Jacalz (approved):
I'm personally happy with this now but I need someone else to review as well given my commits to the branch.

rad756:
Is there anything else needed from me on here or we just need other maintainers to review this and then it could be added to next release?

Also massive thanks for these commits Jacalz!

Jacalz:
> Is there anything else needed from me on here or we just need other maintainers to review this and then it could be added to next release?

Nothing is needed on your end. I can fix any future review comments. Thanks for working on this :)

> Also massive thanks for these commits Jacalz!

No worries. Sorry for this taking so long. This is a great addition.

rad756:
Glad that I could have helped the project.

internal/repository/file.go
```diff
@@ -171,6 +171,13 @@ func (r *FileRepository) Delete(u fyne.URI) error {
 	return os.Remove(u.Path())
 }
 
+// DeleteAll implements repository.DeleteAllRepository.DeleteAll
```
andydotxyz, line 146: Referencing the discussion about removing "Implements: " comments I feel that this documentation could be more helpful.
Is this type missing the interface assertion in the code to assert the implementation?

dweymouth (approved):

internal/repository/file.go
Jacalz: Ah, yes. I added that to be consistent given that all other methods use that. I agree that it should be removed but I'd rather do so in a follow-up PR.

PR #5697, opened 2025-04-25
bugfix: fix keyboard not working on mobile emulator mode

LeHoangLong:
### Description:

Fixes #(issue) 
Keyboard not working on mobile emulator mode (Linux).

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/internal/driver/mobile/app/x11.c
+++ b/internal/driver/mobile/app/x11.c
@@ -57,7 +57,7 @@ new_window(Display *x_dpy, EGLDisplay e_dpy, int w, int h, EGLContext *ctx, EGLS
 	}
 
 	attr.event_mask = StructureNotifyMask | ExposureMask |
-		ButtonPressMask | ButtonReleaseMask | ButtonMotionMask;
+		ButtonPressMask | ButtonReleaseMask | ButtonMotionMask | KeyPressMask | KeyReleaseMask;
 	Window win = XCreateWindow(
 		x_dpy, root, 0, 0, w, h, 0, visInfo->depth, InputOutput,
 		visInfo->visual, CWColormap | CWEventMask, &attr);
@@ -141,6 +141,7 @@ void
 processEvents(void) {
 	while (XPending(x_dpy)) {
 		XEvent ev;
+		KeySym keysym;
 		XNextEvent(x_dpy, &ev);
 		switch (ev.type) {
 		case ButtonPress:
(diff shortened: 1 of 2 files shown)


LeHoangLong:
image

LeHoangLong:
image

Please help to advise if this fix is reasonable before I add tests. 

Also, some test in glfw is failing sporadically, not sure if it's because of this change?

andydotxyz:
Yes it seems like a good approach thanks. Though I think the X methods should not be exported.

Jacalz (changes requested):
Thanks. This looks good and works well. However; do get warnings about `XKeycodeToKeysym` being deprecated. Can anything else be used instead?

```
# fyne.io/fyne/v2/internal/driver/mobile/app
x11.c:154:13: warning: 'XKeycodeToKeysym' is deprecated [-Wdeprecated-declarations]
/usr/include/X11/Xlib.h:1682:1: note: 'XKeycodeToKeysym' has been explicitly marked deprecated here
/usr/include/X11/Xfuncproto.h:136:40: note: expanded from macro '_X_DEPRECATED'
x11.c:158:13: warning: 'XKeycodeToKeysym' is deprecated [-Wdeprecated-declarations]
/usr/include/X11/Xlib.h:1682:1: note: 'XKeycodeToKeysym' has been explicitly marked deprecated here
/usr/include/X11/Xfuncproto.h:136:40: note: expanded from macro '_X_DEPRECATED'
```

andydotxyz:
> Though I think the X methods should not be exported.

I think my comment still stands though

Jacalz:
> I think my comment still stands though

Why not? They are being called from the C-code as far as I can see

andydotxyz:
> > I think my comment still stands though
> 
> Why not? They are being called from the C-code as far as I can see

Just for clarity, and for example, the function `X11KeySymToFyneKeyCode` is only called from within the same Go file as far as I can tell, it has no need to be exposed.

Jacalz:
> Just for clarity, and for example, the function X11KeySymToFyneKeyCode is only called from within the same Go file as far as I can tell, it has no need to be exposed.

Ah, I see. I think we were talking about different kinds of exported. I thought you meant the two functions which were exported to the C code.,

andydotxyz:
> > Just for clarity, and for example, the function X11KeySymToFyneKeyCode is only called from within the same Go file as far as I can tell, it has no need to be exposed.
> 
> Ah, I see. I think we were talking about different kinds of exported. I thought you meant the two functions which were exported to the C code.,

Ah yes I see the ambiguity. However I don't think the X... method(s) are exposed to C - only publicly in Go. Which it shouldn't be.

Issue #5639, opened 2025-04-01
Unsafe use of map in RichText on 2.6.0


hkparker:
### Checklist

- [x] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [x] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

Rarely, using RichText I see this crash at startup while using develop.  Some locks were removed for 2.6, but this seems to have resulted in a race condition on this map.

```
fatal error: concurrent map writes

goroutine 1 [running, locked to thread]:
internal/runtime/maps.fatal({0x143c971?, 0x16121a0?})
        /usr/lib/go/src/runtime/panic.go:1058 +0x18
fyne.io/fyne/v2/widget.(*RichText).cachedSegmentVisual(0xc0047ab080, {0x1610980, 0xc0002f3c00}, 0x0)
        /home/hayden/Development/Go/pkg/mod/fyne.io/fyne/v2@v2.6.0-beta1.0.20250330210720-ecaa745b72dc/widget/richtext.go:218 +0x197
fyne.io/fyne/v2/widget.(*textRenderer).Refresh(0xc004689720)
        /home/hayden/Development/Go/pkg/mod/fyne.io/fyne/v2@v2.6.0-beta1.0.20250330210720-ecaa745b72dc/widget/richtext.go:693 +0x711
fyne.io/fyne/v2/widget.(*RichText).CreateRenderer(0xc0047ab080)
        /home/hayden/Development/Go/pkg/mod/fyne.io/fyne/v2@v2.6.0-beta1.0.20250330210720-ecaa745b72dc/widget/richtext.go:83 +0x29d
fyne.io/fyne/v2/internal/cache.Renderer({0x1614a30, 0xc0047ab080})
        /home/hayden/Development/Go/pkg/mod/fyne.io/fyne/v2@v2.6.0-beta1.0.20250330210720-ecaa745b72dc/internal/cache/widget.go:20 +0x46
fyne.io/fyne/v2/widget.(*BaseWidget).Refresh(0xc0047ab080?)
        /home/hayden/Development/Go/pkg/mod/fyne.io/fyne/v2@v2.6.0-beta1.0.20250330210720-ecaa745b72dc/widget/widget.go:123 +0x49
fyne.io/fyne/v2/widget.(*RichText).Refresh(0xc0047ab080)
[...]

PR #5699, opened 2025-04-27
Clean up how app code starts and ends to fix crashes and false warnings

andydotxyz:
This fixes a few crashes such as #5639 plus seeing warnings incorrectly during start or end.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/loop.go
+++ b/internal/driver/glfw/loop.go
@@ -21,7 +21,7 @@ type funcData struct {
 
 // channel for queuing functions on the main thread
 var funcQueue = async.NewUnboundedChan[funcData]()
-var running atomic.Bool
+var running, drained atomic.Bool
 
 // Arrange that main.main runs on main thread.
 func init() {
@@ -38,7 +38,7 @@ func runOnMain(f func()) {
 func runOnMainWithWait(f func(), wait bool) {
 	// If we are on main just execute - otherwise add it to the main queue and wait.
 	// The "running" variable is normally false when we are on the main thread.
-	if !running.Load() {
+	if (!running.Load() && async.IsMainGoroutine()) || drained.Load() {
 		f()
 		return
 	}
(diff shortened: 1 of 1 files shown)


internal/driver/glfw/loop.go
```diff
@@ -38,7 +38,7 @@ func runOnMain(f func()) {
 func runOnMainWithWait(f func(), wait bool) {
 	// If we are on main just execute - otherwise add it to the main queue and wait.
```
dweymouth, line 39: I think this comment should be updated, something like

// If the app main loop is not yet running, run f directly only if we are on the main goroutine.

internal/driver/glfw/loop.go
```diff
@@ -130,6 +130,15 @@ func (d *gLDriver) runGL() {
 			if f := l.OnStopped(); f != nil {
 				l.QueueEvent(f)
 			}
+
+			for len(funcQueue.Out()) > 0 {
```
dweymouth, line 134: A comment here could be nice too to explain the goal of this code block (I assume it's so that goroutines blocked on a `fyne.DoAndWait` will not be blocked forever)

internal/driver/glfw/loop.go
```diff
@@ -130,6 +130,15 @@ func (d *gLDriver) runGL() {
 			if f := l.OnStopped(); f != nil {
 				l.QueueEvent(f)
```
dweymouth, line 131: Does QueueEvent add f to the funcQueue? If so, will the added code below inadvertently make it so that the OnStopped callback is never invoked now? We should just be safe to call OnStopped directly here, since we're on main?

andydotxyz: I'm not sure I follow, but with my added comment it should illustrate that things are still getting called because we drain that function queue.
It may be safe to call directly, but I think that is a different tidy where lifecycle is updated for the new threading.

internal/driver/glfw/loop.go
```diff
@@ -36,9 +36,9 @@ func runOnMain(f func()) {
 
 // force a function f to run on the main thread and specify if we should wait for it to return
 func runOnMainWithWait(f func(), wait bool) {
-	// If we are on main just execute - otherwise add it to the main queue and wait.
-	// The "running" variable is normally false when we are on the main thread.
-	if !running.Load() {
+	// If we are on main during app run just execute - otherwise add it to the main queue and wait.
```
dweymouth, line 39: I think this should be "If we are on main _before_ app run..."

andydotxyz: Thanks @dweymouth , fixed

dweymouth (approved):

PR #5712, opened 2025-05-01
fix entry undo and redo don't trigger OnChanged

### LINKED ISSUE #5710: Entry OnChanged does not get triggered on undo and redo
### Checklist

### Describe the bug

When using a entry input, the OnChanged does not gets triggered when Undo and redo is done.

### How to reproduce

In the Fyne App for Entry widget, write some content
used CTRL+Z for undo, and see the OnChanged not getting triggered.
Similarly for CTRL+Y for redo the OnChanged for Entry does not gets triggered.

### Screenshots

_No response_

### Example code

```go
package main

import (
	"fmt"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Text Grid")

	entry := widget.NewEntry()
	entry.OnChanged = func(v string) {
		fmt.Println(v)
	}

	w.SetContent(entry)

	w.Resize(fyne.NewSize(400, 400))

	w.ShowAndRun()
}

```

### Fyne version

2.6.0

### Go compiler version

1.24

Windows 11

### Additional Information

_No response_

mbaklor:
### Description:
using undo and redo on an entry widget didn't trigger OnChanged, so I added a call to the callback in both methods.

Fixes #5710 

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -426,6 +426,10 @@ func (e *Entry) Redo() {
 	e.updateText(newText, false)
 	e.CursorRow, e.CursorColumn = e.rowColFromTextPos(pos)
 	e.syncSelectable()
+	cb := e.OnChanged
+	if cb != nil {
+		cb(newText) // We know that the text has changed.
+	}
 	e.Refresh()
 }
 
@@ -720,6 +724,10 @@ func (e *Entry) Undo() {
 	e.updateText(newText, false)
 	e.CursorRow, e.CursorColumn = e.rowColFromTextPos(pos)
 	e.syncSelectable()
+	cb := e.OnChanged
+	if cb != nil {
+		cb(newText) // We know that the text has changed.
+	}
 	e.Refresh()
 }
 
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -426,9 +426,8 @@ func (e *Entry) Redo() {
 	e.updateText(newText, false)
 	e.CursorRow, e.CursorColumn = e.rowColFromTextPos(pos)
 	e.syncSelectable()
-	cb := e.OnChanged
-	if cb != nil {
-		cb(newText) // We know that the text has changed.
+	if e.OnChanged != nil {
+		e.OnChanged(newText)
 	}
 	e.Refresh()
 }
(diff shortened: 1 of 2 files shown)


widget/entry.go
```diff
@@ -426,6 +426,10 @@ func (e *Entry) Redo() {
 	e.updateText(newText, false)
 	e.CursorRow, e.CursorColumn = e.rowColFromTextPos(pos)
 	e.syncSelectable()
+	cb := e.OnChanged
```
andydotxyz, line 429: This could be simplified:

```go
	if cb := e.OnChanged; cb != nil {
```

widget/entry.go
```diff
@@ -426,6 +426,10 @@ func (e *Entry) Redo() {
...
+	if cb != nil {
+		cb(newText) // We know that the text has changed.
```
andydotxyz, line 431: Avoid comments unless it is required and cannot be communicated through cleaner code - this seems self-evident so no need

widget/entry.go
mbaklor: this is fair but I wanted to keep it in line with the other places we call the callback, I'll fix when I get home, we can remove the whole `cb := ` at this point since we no longer have to mutex lock

widget/entry.go
mbaklor: oops I was sure I removed them that's my bad, I'll fix when I get home 😅

widget/entry.go
andydotxyz: I did do that check to make sure it wasn't suitably matching - but in the other cases the callback was taken a long time before the callback. Perhaps they do indeed all need fixed with the mutex changes, not sure.

mbaklor:
Fixed the stylistic things, no need to copy the callback now that there isn't a mutex lock, and removed the comment that I accidentally yanked from elsewhere

andydotxyz (approved):
Thanks!

PR #5739, opened 2025-05-12
Generalise Form Validation

jimorc:
### Description:
The Form widget only handles validation of Entry widget. These changes generalise validation handling by:
1. Creating a new FormValidatable interface that encapsulates Entry functionality that the Form widget was referencing directly. For example: instead of referencing Entry.dirty directly, the interface defines a new IsDirty() method.
2. Modifying the Entry widget to implement the FormValidatable interface.
3. Modifying the Form widget to interact with the Entry widget only using the methods in the FormValidatable interface.
4. Remove all Entry specific references from Form.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/cmd/fyne_demo/tutorials/dialog.go
+++ b/cmd/fyne_demo/tutorials/dialog.go
@@ -120,9 +120,9 @@ func dialogScreen(win fyne.Window) fyne.CanvasObject {
 		}),
 		widget.NewButton("Form Dialog (Login Form)", func() {
 			username := widget.NewEntry()
-			username.Validator = validation.NewRegexp(`^[A-Za-z0-9_-]+$`, "username can only contain letters, numbers, '_', and '-'")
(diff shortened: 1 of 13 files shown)


andydotxyz:
I think the rebase did not work properly - this is including lots of changes that are unrelated.

Jacalz:
See discussion at https://github.com/fyne-io/fyne/pull/4250.

andydotxyz:
@jimorc I'm not sure why you closed this, I don't think @Jacalz was saying this is duplicative just that feedback has been given in this area before.
Sometimes there are multiple people solving a problem and it will be the most elegant, idiomatic solution that will land.

andydotxyz:
Thanks for the clarification. Smaller PRs in general are easier to review. If the issue was already existing there is no harm in opening the first and working on the issue separately. Though if it was introduced or somehow interlinked with this work maybe post them together.

Issue #5391, opened 2025-01-10
Change widget.Entry.onFocusChanged to widget.Entry.OnFocusChanged


jimorc:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [X] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

The `Entry.onFocusChanged` field is provided so that validation can be performed externally to the Entry widget whenever focus on the widget is lost. However, there is no way to set the field outside the widget package. Code in the Entry widget checks if the field is set and calls the registered callback, but the field is never set anywhere in the widget package. It is therefore not a usable field.

### Is it possible to construct a solution with the existing API?

The following is possible and is what I am currently doing:
```go
type Entry2 struct {
    widget.Entry
}
func (e Entry2) FocusLost() {
// override the Entry.FocusLost code to do what you would have put in OnFocusChanged
}
```

### Describe the solution you'd like to see.

Change `onFocusChanged` to `OnFocusChanged` in the Entry struct.

jimorc:
go version 1.23.4 although the fyne go.mod sets it to version 1.21
fyne version 1.5.
I have also checked the fyne develop branch.

Jacalz:
Please see the discussion in https://github.com/fyne-io/fyne/pull/4250. I need to get that sorted and completed. There are way more issues than just `onFocusChanged` being specific to Entry. Not sure exporting it is the right solution but there needs to be more discussions about it.

andydotxyz:
I don't think that this is the right solution. Any public API that is exposed so that it can be used in another widget (like Form) is sure to cause problems because if the developer wants to use it then either their callback will override form or it won't be called because the form overrides developer.
If one widget depends on another it needs to be done without the `OnSomething` public APIs.

Jacalz:
Yup. I'll try to get my PR back up and running sometime going forward. There is bound to be a better solution

PR #5741, opened 2025-05-15
Support form validation of widgets other than Entry

jimorc:
Form and FormDialog currently support validation of Entry widgets. With this PR, validation can be extended to other widgets.
This PR is an alternative to any changes submitted with issue #4250. Issue #4147 is also fixed.

### Description
FormValidatable interface has been added. Entry has been modified to implement this interface. Form has been modified to validate widgets that implement FormValidatable interface. This removes Entry specific code from Form, and also any need for reflection.

### Checklist:

- [X] Tests included.
- (No additional tests were required)
- [x] Tests all pass.

### DIFF
--- a/dialog/form_test.go
+++ b/dialog/form_test.go
@@ -119,12 +119,12 @@ func TestFormDialog_Hints(t *testing.T) {
 
 func TestFormDialog_Submit(t *testing.T) {
 	validatingEntry := widget.NewEntry()
-	validatingEntry.Validator = func(input string) error {
+	validatingEntry.SetValidator(func(input string) error {
 		if input != "abc" {
 			return errors.New("only accepts 'abc'")
 		}
 		return nil
-	}
+	})
 	validatingItem := &widget.FormItem{Widget: validatingEntry}
 
 	confirmed := false
@@ -150,12 +150,12 @@ func TestFormDialog_Submit(t *testing.T) {
 
 func validatingFormDialog(result *formDialogResult, parent fyne.Window) *FormDialog {
 	validatingEntry := widget.NewEntry()
-	validatingEntry.Validator = func(input string) error {
+	validatingEntry.SetValidator(func(input string) error {
 		if input != "abc" {
 			return errors.New("only accepts 'abc'")
 		}
 		return nil
-	}
+	})
 	validatingItem := &widget.FormItem{
 		Text:   "Only accepts 'abc'",
 		Widget: validatingEntry,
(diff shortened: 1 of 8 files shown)


jimorc:
There are two Validatable interfaces: Validatable and FormValidatable which is defined in this PR. FormValidatable is meant to be used by any widget that needs to be validated within a form. Validatable was retained because the Form iteself uses this interface and currently does not need to be FormValidatable.

A few days ago, Bob Hyman discussed forms within another form on Slack and provided a use case. To implement forms within a form, then the form would need to implement the FormValidatable interface, so there is no need for both Validatable and FormValidatable. I could fold the methods specified in FormValidatable into Validatable so that there is only one interface. However, I am not sure if anyone is using Validatable outside of the fyne widget package.

Jacalz:
Like I commented on in the previous, please have a look at the discussion at  https://github.com/fyne-io/fyne/pull/4250#pullrequestreview-1870719464. I believe that many of the review comments in that PR applies to this one as the approach mostly is similar.

On another note, I would suggest not locking our self into the interface have “form” in the name in case validation gets to be useful in some other component.

Jacalz:
> Validatable was retained because the Form iteself uses this interface and currently does not need to be FormValidatable.

Removing or changing an existing interface would not be possible either way given that stability guarantees.

jimorc:
@Jacalz wrote:
> Like I commented on in the previous, please have a look at the discussion at [#4250 (review)](https://github.com/fyne-io/fyne/pull/4250#pullrequestreview-1870719464). I believe that many of the review comments in that PR applies to this one as the approach mostly is similar.

Yes, I did just provide new methods to access the Entry fields that are referenced directly in Form. I will go back and see if I can clean this up somewhat.

@jacalz also wrote:
> On another note, I would suggest not locking our self into the interface having “form” in the name in case validation gets to be useful in some other component in the future.

I originally called it ValidatableWidget rather than FormValidatable. Would that be a better choice? If not, do you have a recommendation? I am not always the best at naming things.

Jacalz:
> I originally called it ValidatableWidget rather than FormValidatable. Would that be a better choice? If not, do you have a recommendation? I am not always the best at naming things.

Adding "Widget" at the end doesn't quite match the existing naming and is confusing compared to the existing interface. I went for "ExtendedValidatable" to indicate that it was an improvement over the existing interface but I'm sure there is something better.

jimorc:
Some time ago I submitted a change request (#5391) that is related to this. In my suggested changes above, I have SetOnFocusChanged as an alternative to the OnFocusChanged in 5391.

Perhaps an alternative to a single onFocusChanged field is to allow multiple functions to be called on focus change. That would require adding another RemoveOnFocusChanged method to the interface.

PR #5754, opened 2025-05-26
Multiple performance improvements for TextGrid

redawl:
### Description:

While profiling an application I am writing that uses TextGrid, I noticed that the largest amount of time was being spent inside 
`updateGridSize`, not any render specific code. This led me down the path of trying to optimize it.

Optimizations:
1. Remove dead code from `textGridContentRenderer.updateGridSize`
    I noticed that bufCols is computed, but then never used, so I removed it.
2. Use direct comparisons and integer division instead of `math.Max` and `math.Floor` respectively.
    Max and Floor are general purpose utilities, and as such have to do a lot of extra work that we don't need. For example, they must properly handle `Infinity`, as well as negative numbers. This introduces a lot of overhead. Since we are checking lengths of rows and columns, we can guarantee that the numbers involved are positive, we can beat the performance of Max and Floor by doing the computation ourselves directly.

With these two performance improvements, I see a noticeable difference in performance. 

Take a look at this recording, with optimizations on the left, without optimizations on the right:
https://github.com/user-attachments/assets/e3f8ab9f-cd59-4b72-81c7-59413a9decf8

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/textgrid.go
+++ b/widget/textgrid.go
@@ -496,13 +496,13 @@ func (t *textGridContent) refreshCell(row, col int) {
 
 func (t *textGridContentRenderer) updateGridSize(size fyne.Size) {
 	bufRows := len(t.text.text.Rows)
-	bufCols := 0
-	for _, row := range t.text.text.Rows {
-		bufCols = int(math.Max(float64(bufCols), float64(len(row.Cells))))
-	}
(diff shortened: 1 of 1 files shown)


andydotxyz (approved):
Thanks, some clever optimisations there :)

PR #5769, opened 2025-06-06
all: Run gofumpt

Jacalz:
### Description:

This runs [gofumpt](https://github.com/mvdan/gofumpt) on the whole project. It is a lot stricter than gofmt but in a way such that running gofmt on it again does not complain. I would propose that we stop using the deprecated goimports and switch to this instead as a future step. We would unfortunately lose automatic reordering of import statement but I think it might be worth it given that we get a much better formatter.

I have used this tool in my own projects for years now and it really is great!

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Jacalz:
Yeah, good point on using both. I'll fix up that formatting when I update the workflows 👍

PR #5772, opened 2025-06-06
Fix non-atomic access to expiringCache internals

cognusion:
### Description:
expiringCache is used in a number of places internally. When a function, e.g. GetSVG() is called when destroyExpiredSvgs() is running, there is a race to read and write the internal time.Time.

This changes the time.Time to an atomic.Value, and updates the accessor functions accordingly. This could also be handled with an RWMutex. I am uncertain whether the time is read-most or write-most, so went with the least intrusive overall solution. If that or another pattern is preferred, I would gladly implement it accordingly.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/cache/base.go
+++ b/internal/cache/base.go
@@ -2,6 +2,7 @@ package cache
 
 import (
 	"os"
+	"sync/atomic"
 	"time"
 
 	"fyne.io/fyne/v2"
@@ -108,15 +109,19 @@ func destroyExpiredRenderers(now time.Time) {
 }
 
 type expiringCache struct {
-	expires time.Time
+	expires atomic.Value
 }
 
 // isExpired check if the cache data is expired.
 func (c *expiringCache) isExpired(now time.Time) bool {
-	return c.expires.Before(now)
+	exp := c.expires.Load()
+	if exp == nil {
+		return time.Time{}.Before(now)
+	}
+	return exp.(time.Time).Before(now)
 }
 
 // setAlive updates expiration time.
 func (c *expiringCache) setAlive() {
-	c.expires = timeNow().Add(ValidDuration)
+	c.expires.Store(timeNow().Add(ValidDuration))
 }


Jacalz (commented):
Are you sure there really is a race here? All our tests are running with race detector on in the CI. Are all your widgets being updated from the `fyne.Do[AndWait]` functions as expected? Race conditions in this part of the cache will be reported for anyone that hasn’t been migrated to the new threading model yet.

cognusion:
I can find races using Fyne tests. Some are harder than others to trigger. Here's a reliable one: Close to 100% of the time vs develop.
`
fyne/internal/driver/glfw$ go test -v -race -run TestWindow_Shortcut
...
    testing.go:1490: race detected during execution of test
--- FAIL: TestWindow_Shortcut (0.40s)
FAIL
`

If the objects that implement expiringCache (canvas, svg, text, widget, etc.) are never supposed to be accessed outside of a single goro, then this can be closed. I was having an intermittent problem with a fairly trivial app I wrote, when the system was busy doing other things, and started looking for places that were time sensitive or brittle. I found quite a few, but fixing expiringCache fixed my SVG rendering problem. 

Attached is a -race test run using Fyne's showing a bunch of races, including the expiringCache race that this PR fixes. Again, if I'm the problem, and none of that code should ever be accessed outside a single goro, then please accept my apologies for wasting your time. I've certainly spent a lot chasing these ghosts.

[fyne.test.out.dump.txt](https://github.com/user-attachments/files/20635679/fyne.test.out.dump.txt)

Jacalz:
> Again, if I'm the problem, and none of that code should ever be accessed outside a single goro, then please accept my apologies for wasting your time. I've certainly spent a lot chasing these ghosts.

The new threading model assumes that you are updating all graphical applications from the main goroutine. It was not long ago that I removed this lock actually. Looking at https://docs.fyne.io/api/v2.6/upgrading.html is perhaps helpful. The idea is to use fyne.Do[AndWait] for scheduling updates and turn on the migration in FyneApp.toml when you are ready.

Don't worry about wasting anyone's time. Race conditions are important and I appreciate you taking the time to try and fix this regardless of confusion. Thanks for the contribution :)

Jacalz:
Closing per comment above

PR #5774, opened 2025-06-07
Fix bug in refreshCell causing multiple renderers to be created

### LINKED ISSUE #5767: Calling SetStyleRange on a scrolled TextGrid causes the internal scroll to always have offset 0, 0
### Checklist

### Describe the bug

When calling SetStyleRange on a TextGrid that is scrolled away from the origin, the internal scroll always has an offset of 0, 0. This causes noticeable visual glitches, as well as causing functions like `CursorLocationForPosition` to report the wrong data.

Video reproduction:

https://github.com/user-attachments/assets/08b3fee9-a748-429a-981b-501759c64fdb

### How to reproduce

1. Run example code
2. Scroll down a bit
3. Click the button. This calls `SetStyleRange`, and causes the bug.
4. Notice that the textgrid only displays the first part of the text, up to the size of the scroll container. This is because the scroll is always 0,0. 

### Screenshots

_No response_

### Example code

```go
package main                                                                                                                                                20:40:34 [22/1947]

import (
        "fmt"
        "strconv"
        "strings"

        "fyne.io/fyne/v2"
        "fyne.io/fyne/v2/app"
        "fyne.io/fyne/v2/container"
        "fyne.io/fyne/v2/widget"
)

var _ desktop.Hoverable = (*TextGridWrapper)(nil)

type TextGridWrapper struct {
        widget.TextGrid
}

func (t *TextGridWrapper) MouseIn(evt *desktop.MouseEvent) {}
func (t *TextGridWrapper) MouseMoved(evt *desktop.MouseEvent) {
        row, col := t.CursorLocationForPosition(evt.Position)
        fmt.Printf("Reported location: row = %d, col = %d\n", row, col)
}
func (t *TextGridWrapper) MouseOut() {}
func multiplyString(str string, i int) string {
        builder := strings.Builder{}
        inc := i
        for inc > 0 {
                builder.WriteString(strconv.Itoa(i - inc))
                builder.WriteString(": ")
                builder.WriteString(str)
                builder.WriteByte('\n')
                inc--
        }

        return builder.String()
}

func main() {
        a := app.New()
        w := a.NewWindow("Test")

        textGridWrapper := &TextGridWrapper{}
        textGridWrapper.Scroll = fyne.ScrollBoth
        textGridWrapper.SetText(multiplyString("Hello, world", 10000))

        textGridWrapper.ExtendBaseWidget(textGridWrapper)

        button := widget.NewButton("Click me", func() {
                textGridWrapper.SetStyleRange(0, 0, 100, 100, widget.TextGridStyleWhitespace)
        })

        layout := container.NewBorder(button, nil, nil, nil, textGridWrapper)

        w.SetContent(layout)
        w.ShowAndRun()
}
```

### Fyne version

develop

### Go compiler version

1.24.3

Gentoo

### Additional Information

_No response_

redawl:
### Description:

Fixes #5767

The bug was caused by `cache.Renderer` creating a renderer for TextGrid, even when TextGrid was extended with a custom widget. Instead, we can retrieve textContent directly from TextGrid, without needing to retrieve the renderer at all.

After that was fixed, it also uncovered an out-of-bounds error in `textGridRowRenderer`, which also needed to be fixed. 

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/textgrid.go
+++ b/widget/textgrid.go
@@ -663,9 +663,11 @@ func (t *textGridRowRenderer) refreshCell(col int) {
 		return
 	}
 
-	if len(t.obj.text.text.Rows[t.obj.row].Cells) > col {
-		cell := t.obj.text.text.Rows[t.obj.row].Cells[col]
-		t.setCellRune(cell.Rune, pos, cell.Style, t.obj.text.text.Rows[t.obj.row].Style)
+	row := t.obj.text.text.Rows[t.obj.row]
+
+	if len(row.Cells) > col {
+		cell := row.Cells[col]
+		t.setCellRune(cell.Rune, pos, cell.Style, row.Style)
 	}
 }
 


widget/textgrid.go
```diff
@@ -663,8 +663,10 @@ func (t *textGridRowRenderer) refreshCell(col int) {
 		return
 	}
 
-	cell := t.obj.text.text.Rows[t.obj.row].Cells[col]
-	t.setCellRune(cell.Rune, pos, cell.Style, t.obj.text.text.Rows[t.obj.row].Style)
+	if len(t.obj.text.text.Rows[t.obj.row].Cells) > col {
```
dweymouth: Minor style suggestion - I think readability can benefit from assigning `row := t.obj.text.text.Rows[t.obj.row]` to avoid the repeats of this slightly cumbersome access chain

redawl: Good idea, done in https://github.com/fyne-io/fyne/pull/5774/commits/8b49aa368e5d73ca48e39f192cb6ae4663cd9b6b :)

redawl:
> LGTM - If you can think of a not-too-convoluted way to unit test this, could be nice to get a test, but it's also such a simple fix I'm fine with it being merged as-is too. (Side note - I love any change that can remove `cache.Renderer` calls within widget code, since I think it's kind-of a broken pattern and seems to be bug-prone)

I did attempt to write a unit test, but I was unable to reproduce the bug in the test environment. Since it is a simple fix (no real logic changes besides the bounds check), I gave up. I did test it thoroughly with a small example app, as well as my larger production app, and it fixed the issue with no side effects.

Good to know about the cache.Renderer calls being problematic in widget code; I did also test changing to `cache.Renderer(t.super())`, which also fixed the bug, but then I realized there was no need to invoke the renderer cache at all.

PR #5775, opened 2025-06-09
Try to use CC=clang for faster CI build on Linux

Jacalz:
### Description:

Windows is the slowest building system in CI but that doesn't stop us from trying to make Linux compiles faster. I have been seeing clang get much faster compiles locally so let's see if it can work its magic in CI as well. I also removed the apt package installation from the web tests given that the WASM port doesn't use CGO so it runs blazingly fast now.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Nice thanks

PR #5776, opened 2025-06-11
cmd/fyne: Remove everything but the public APIs

Jacalz:
### Description:

Phasing out the old cmd/fyne tool in favour of the tools repo was done as part of 2.6.0. We have discussed potentially removing part of command to clean up the go.mod here and avoid carrying around the bulk of the old command. This PR removes the cli stuff but leaves the APIs used by the `cmd/fyne/commands` package. I don't know to which degree those still work or are usable but they are there.

The added benefit of this change is that it makes it more obvious to users that they should migrate to the new tool.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

#### Where applicable:


andydotxyz:
I like the initiative, but isn't it a bit fast? The CLI only added the "deprecated" notice in the last major release.
This essentially makes it non-functional in the next release.
I would prefer that have at least the same notice as Go for breaking tools changes (2 major releases - approx 1 year).

Jacalz:
Yeah, good point. Forgot about that

Jacalz:
I think it may be time to re-evaluate this again for v2.8. It would have been nice to remove the APIs but I still think the fyne installer uses the old API?

andydotxyz:
> I think it may be time to re-evaluate this again for v2.8. It would have been nice to remove the APIs but I still think the fyne installer uses the old API?

Yup. Let's get some work into the installer and setup tools to get them updated to all the latest features and APIs then we can come back and reconsider.

Jacalz:
There are no new APIs available yet and this PR does not remove any of the old ones. This can probably land before that.

andydotxyz:
I misunderstood sorry - you're removing all of the functionality but not the public APIs.

Wasn't the first release of the tools CLI in August this year? That means that when 2.8.0 releases we would be pulling out features where the replacement is barely half a year old. Thats twice as aggressive as Google and I don't see what we gain from forcing that on people.

Jacalz:
Ah, I thought it was longer ago than that. These last couple months have all drifted together; it's been stressful. Two release cycles does match what they are doing although it is over a shorter time frame.

As this leaves everything that was reachable through the public APIs in place but only removes the cli part, I'd personally say that it might be time for this. If people want to continue using the old frozen code, they can just not update to this release. I've seen various reports of people being confused still using this old version so I genuinely think that removing the cli functionality (and leaving the warning in when running it) for the tool may have a positive impact in moving people over to the new tool. I do understand if you think it still is too early for that though. I genuinely believed, perhaps incorrectly, when I commented yesterday that it would be about a year between once this release is out.

sdassow:
I think this is a good step as well.

Since we have a deprecation warning and know the version numbers now, should we include those to reduce confusion even further (people were asking about it in the chat)?

andydotxyz:
We have discussed it before, but to be clear for anyone new - I don't want to be more aggressive with breaking changes than the Go team are. 1 year is an absolute minimum to be deprecated and ideally it would be more (see related discussion about the thread handling changes).

With all that it makes no sense at all that we would remove these features before our tools have stopped using them - and fyne-cross has not yet had a release that supports the new tool...

Jacalz:
Alright let's go for a year. Sorry, my mind really tricked me into thinking the tool had been out longer

PR #5779, opened 2025-06-11
TextGrid.SetText update scroll if new text is shorter

### LINKED ISSUE #5762: TextGrid is glitchy when calling SetText on a scrolled container.
### Checklist

### Describe the bug

When calling SetText on a scrolled TextGrid, if the new next doesn't reach down as far as the previous text, there are visual glitches, like the text not appearing, or appearing only partially.

### How to reproduce

1. Run example code
2. Click button twice
3. Scroll down a ways
4. Click button again
5. Notice that new text doesn't show until the TextGrid is touched.

### Screenshots

_No response_

### Example code

```go
package main

import (
        "strconv"
        "strings"

        "fyne.io/fyne/v2"
        "fyne.io/fyne/v2/app"
        "fyne.io/fyne/v2/container"
        "fyne.io/fyne/v2/widget"
)

func multiplyString(str string, i int) string {
        builder := strings.Builder{}
        inc := i
        for inc > 0 {
                builder.WriteString(strconv.Itoa(i - inc))
                builder.WriteString(": ")
                builder.WriteString(str)
                builder.WriteByte('\n')
                inc--
        }

        return builder.String()
}

func main() {
        a := app.New()
        w := a.NewWindow("Test")
        text := "Example text"
        toggle := false

        textGridWrapper := &widget.TextGrid{
                Scroll: fyne.ScrollBoth,
        }

        button := widget.NewButton("Click me!", func() {
                if toggle {
                        textGridWrapper.SetText(multiplyString(text, 10000))
                } else {
                        textGridWrapper.SetText(multiplyString(text, 50))
                }
                toggle = !toggle
        })

        textGridWrapper.SetText(text)
        layout := container.NewBorder(nil, nil, nil, button, textGridWrapper)

        w.SetContent(layout)
        w.ShowAndRun()
}
```

### Fyne version

develop

### Go compiler version

1.24.3

Gentoo

### Additional Information

_No response_

redawl:
### Description:

Fixes #5762

_Note: I had to add an explicit refresh in one of the unit tests, otherwise I was getting a crash related to 
`t.scroll.Content` being nil in `Scroll.updateOffset`._

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/textgrid.go
+++ b/widget/textgrid.go
@@ -175,8 +175,9 @@ func (t *TextGrid) SetText(text string) {
 	oldRowsLen := len(t.Rows)
 	t.Rows = rows
 
-	// Update scroll position if new text is shorter than previous text
-	if t.Scroll != fyne.ScrollNone && len(rows) < oldRowsLen {
+	// If we don't update the scroll offset when the text is shorter,
+	// we may end up with no text displayed or text appearing partially cut off
+	if t.Scroll != fyne.ScrollNone && len(rows) < oldRowsLen && t.scroll.Content != nil {
 		offset := t.PositionForCursorLocation(len(rows), 0)
 		t.scroll.ScrollToOffset(fyne.NewPos(offset.X, t.scroll.Offset.Y))
 		t.scroll.Refresh()
--- a/widget/textgrid.go
+++ b/widget/textgrid.go
@@ -177,7 +177,7 @@ func (t *TextGrid) SetText(text string) {
 
 	// If we don't update the scroll offset when the text is shorter,
 	// we may end up with no text displayed or text appearing partially cut off
-	if t.Scroll != fyne.ScrollNone && len(rows) < oldRowsLen && t.scroll.Content != nil {
+	if t.scroll != nil && t.Scroll != fyne.ScrollNone && len(rows) < oldRowsLen && t.scroll.Content != nil {
 		offset := t.PositionForCursorLocation(len(rows), 0)
 		t.scroll.ScrollToOffset(fyne.NewPos(offset.X, t.scroll.Offset.Y))
 		t.scroll.Refresh()
--- a/widget/textgrid.go
+++ b/widget/textgrid.go
@@ -172,7 +172,16 @@ func (t *TextGrid) Resize(size fyne.Size) {
 func (t *TextGrid) SetText(text string) {
 	rows := t.parseRows(text)
 
+	oldRowsLen := len(t.Rows)
 	t.Rows = rows
+
+	// Update scroll position if new text is shorter than previous text
+	if t.Scroll != fyne.ScrollNone && len(rows) < oldRowsLen {
+		offset := t.PositionForCursorLocation(len(rows), 0)
+		t.scroll.ScrollToOffset(fyne.NewPos(offset.X, t.scroll.Offset.Y))
+		t.scroll.Refresh()
+	}
+
 	t.Refresh()
 }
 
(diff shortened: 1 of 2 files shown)


widget/textgrid.go
```diff
@@ -172,7 +172,16 @@ func (t *TextGrid) Resize(size fyne.Size) {
 func (t *TextGrid) SetText(text string) {
 	rows := t.parseRows(text)
 
+	oldRowsLen := len(t.Rows)
 	t.Rows = rows
+
+	// Update scroll position if new text is shorter than previous text
```
Jacalz: This comment just says the same as reading the code on the line below. If it has to be commented to be clearer, I’d prefer to explain why we have to update the scroll position for shorter text (maybe I’m just not understanding the code well enough to know)

redawl: Good point. I am really just saying what the condition already says by itself.

I'm thinking a comment could still be useful, with more of a why than a how. Something like:
`If we don't update the scroll offset when the text is shorter, we may end up with no text displayed or text appearing partially cut off`.

I'll make the update if you think that is useful, otherwise I'll remove the comment altogether.

widget/textgrid.go
```diff
@@ -172,7 +172,16 @@ func (t *TextGrid) Resize(size fyne.Size) {
...
+	if t.Scroll != fyne.ScrollNone && len(rows) < oldRowsLen {
+		offset := t.PositionForCursorLocation(len(rows), 0)
+		t.scroll.ScrollToOffset(fyne.NewPos(offset.X, t.scroll.Offset.Y))
```
dweymouth, line 182: I think the fact that you had to add a Refresh to the test is because of here, where `t.scroll` will be nil if the widget has not yet been rendered. I think adding a `&& t.scroll != nil` to the if conditions here should avoid this issue.

widget/textgrid.go
redawl: Updated the comment in https://github.com/fyne-io/fyne/pull/5779/commits/11809011e036fd85b96e28e7606956be4564b66b.

widget/textgrid.go
redawl: `t.scroll != nil` didn't work, but `t.scroll.Content != nil` did, thanks!
 https://github.com/fyne-io/fyne/pull/5779/commits/11809011e036fd85b96e28e7606956be4564b66b

dweymouth: I think you still need a `t.scroll != nil` condition before `t.scroll.Content != nil` because `t.scroll` is not initialized at construction. I haven't tested but I think a NewTextGrid with Scroll set to something other than ScrollNone immediately followed by a SetText call will crash

redawl: `TextGrid.CreateRenderer()` initializes `t.scroll`, so it is always non-nil.
There are tests that cover the scenario you describe, and they do pass.

dweymouth: There's no guarantee CreateRenderer is called before SetText though. It does seem like in practice, `len(rows) < oldRowsLen` is false before it tries to access `t.scroll.Content`, preventing a nil crash. But if I remove that condition I can get it to crash on accessing `t.scroll.Content`

redawl: Ah, good point. Would you prefer an explicit check that t.scroll != nil, or is the `len(rows) < oldRowsLen` enough?

dweymouth: Adding the check for  t.scroll != nil to the whole existing list of checks is best IMO

redawl: Done! https://github.com/fyne-io/fyne/pull/5779/commits/fecf5f8f32e5603e3aafd432fc4eb344bccca277

dweymouth (approved):

PR #5787, opened 2025-06-16
Run gofumpt in CI workflow

Jacalz:
### Description:

Much stricter checking of formatting. Followup for #5769. Keeping gofumpt for import statement reordering until https://github.com/mvdan/gofumpt/issues/38 is fixed.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -8,56 +8,57 @@ env:
 jobs:
   static_analysis:
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
 
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/static_analysis.yml
```diff
@@ -8,56 +8,57 @@ env:
 jobs:
   static_analysis:
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
```
Jacalz, line 12: This was never actually used in the static analysis workflow. The GitHub workflow linter in Zed complained about it. The formatting changes to the file are also from Zed FYI.

PR #5790, opened 2025-06-16
Minor cleanups to wasm driver

Jacalz:
### Description:

This removes some GOOS checks (likely leftovers from desktop driver) which were never going to be true, makes sure we clean up the pending function to run and also includes some minor cleanups.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


internal/driver/glfw/window_wasm.go
```diff
@@ -552,6 +531,7 @@ func (w *window) create() {
 	for _, fn := range w.pending {
 		fn()
 	}
+	w.pending = nil
```
andydotxyz: Can you comment on this line?

Jacalz: I just added in the same as what the general desktop driver does on the corresponding line https://github.com/fyne-io/fyne/blob/66c8de3d389aa7d33c72ce371e93c09c397ad8ba/internal/driver/glfw/window_desktop.go#L787. I can refactor it out to a shared function between the two implementations if you'd like?

Jacalz: I extracted into a separate function and also cleaned up a few other minor things I saw

Jacalz: I have fixed the conflict. Would you kindly re-review? @andydotxyz

sdassow (dismissed):
Diff looks good to me.

internal/driver/glfw/window_wasm.go
```diff
@@ -227,26 +220,12 @@ func (w *window) mouseScrolled(viewport *glfw.Window, xoff, yoff float64) {
 func convertMouseButton(btn glfw.MouseButton, mods glfw.ModifierKey) (desktop.MouseButton, fyne.KeyModifier) {
 	modifier := desktopModifier(mods)
 	var button desktop.MouseButton
-	rightClick := false
-	if runtime.GOOS == "darwin" {
```
andydotxyz, line 254: I'm not sure about removing this code - although it was not executing it has purpose.
What we really need to do is move it to a runtime check of the client not the system.

internal/driver/glfw/window_wasm.go
```diff
@@ -227,26 +220,12 @@ func (w *window) mouseScrolled(viewport *glfw.Window, xoff, yoff float64) {
...
-		}
-		if modifier&fyne.KeyModifierSuper != 0 {
-			modifier |= fyne.KeyModifierControl
-			modifier &^= fyne.KeyModifierSuper
-		}
-	}
+
 	switch btn {
 	case glfw.MouseButton1:
-		if rightClick {
```
andydotxyz, line 266: As above, ctrl+left on macOS should be right

internal/driver/glfw/window_wasm.go
andydotxyz: Thanks for this, sorry for the delay

internal/driver/glfw/window_wasm.go
Jacalz: Ah, sorry. I thought it was just working around inconsistencies in GLFW. Will fix.

Jacalz: I fixed it up and also noticed that support for mouse button key 3 was missing compared to desktop. Added that in as well.

andydotxyz (approved):
Nice tidies thanks

PR #5791, opened 2025-06-16
Add support for changing cursor in WASM

Jacalz:
### Description:

Update web driver to support changing cursors. Previously it was always using the pointer cursor.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


internal/driver/glfw/device_wasm.go
```diff
@@ -26,32 +28,18 @@ func (*glDevice) SystemScaleForWindow(w fyne.Window) float32 {
 }
 
 func (*glDevice) hideVirtualKeyboard() {
-	d := dummyEntry()
-	if d.IsNull() {
```
andydotxyz, line 30: This clearly caught when "dummyEntry" was not found... this is safety because it wasn't always there.
Now the same case will panic

andydotxyz (changes requested):
Noted inline this could now panic where before it wouldn't

internal/driver/glfw/device_wasm.go
Jacalz: This will only panic if you are using a very old fyne tool to build the WASM bundle. The dummyEntry already exists in the new cmd/fyne plus if anyone installs cmd/fyne from v2.6.x in the old location. Would it not be safe safe to assume that it exists by 2.7? If not, I think we need a policy for how long to support old tools.

Jacalz: I can add the check back if you like but I'm just saying that we shouldn't need to have it forever.

andydotxyz: It shouldn't be there forever, no - but let's not pull it out until there is a reason to :)

> Would it not be safe safe to assume that it exists by 2.7

As in other related discussions I think it's bad practice to have a break in the tools upgrade cycle in under a year of support time. 2 would be better.

Jacalz: Fair point. Fixed now

Issue #5789, opened 2025-06-16
Leading icon in Entry widget


Vinci10:
This is my first issue to the Fyne project. I would like to thank you for this awesome framework — I really enjoy using it!

### Checklist

- [x] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [x] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

Material design specs has a leading icon (an icon at the start/left of the input field) in text entry widgets. This is useful for indicating the purpose of the field (e.g., a search icon for an search input etc.). Currently, Fyne's Entry widget only supports an action item (trailing icon) but does not provide a built-in way to add a leading icon.

### Describe the solution you'd like to see.

I would like the Entry widget to support a leading icon property (e.g., Icon fyne.Resource), similar to the existing ActionItem for trailing icons. This icon should be rendered inside the Entry, aligned to the left, and should respect the Entry's padding, background, and focus/disabled/validation states. The API could look like:
```go
entry := widget.NewEntry()
entry.SetIcon(theme.MailIcon())
// or
entry.Icon = theme.MailIcon()
```

I have a working solution for this feature. If this request is accepted, I can create a pull request.

![Image](https://github.com/user-attachments/assets/4cac488b-0200-4004-bfbd-513f52543a1e)

dweymouth:
I think this is a reasonable feature request - especially since it's specified in material design. Feel free to open a PR for discussion :)

andydotxyz:
Agreed, though something more descriptive than `Icon` may be good if possible - it feels potentially confused with the action item just now.

But maybe it's OK, if that is what people would expect.

Vinci10:
I have opened the PR. Thanks for the suggestions. I agree that the name `Icon` may not be descriptive enough.

PR #5793, opened 2025-06-20
Add leading icon to the entry widget

Vinci10:
### Description:

Feature Request [#5789](https://github.com/fyne-io/fyne/issues/5789)

Add `Icon fyne.Resource` to the Entry widget. Icon is displayed at the outer left of the entry.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -69,13 +69,13 @@ type Entry struct {
 	// Since: 2.7
 	AlwaysShowValidationError bool
 
+	CursorRow, CursorColumn int
+	OnCursorChanged         func() `json:"-"`
+
 	// Icon is displayed at the outer left of the entry.
 	// It is not clickable, but can be used to indicate the purpose of the entry.
 	// Since: 2.7
 	Icon fyne.Resource `json:"-"`
-
-	CursorRow, CursorColumn int
-	OnCursorChanged         func() `json:"-"`
 
 	cursorAnim *entryCursorAnimation
 
@@ -1539,8 +1539,7 @@ func (r *entryRenderer) leadingInset() float32 {
 	xInset := float32(0)
 
 	if r.entry.Icon != nil {
-		iconSpace := th.Size(theme.SizeNameInlineIcon) + th.Size(theme.SizeNameLineSpacing) + th.Size(theme.SizeNameInnerPadding)
-		xInset = iconSpace
+		xInset = th.Size(theme.SizeNameInlineIcon) + th.Size(theme.SizeNameLineSpacing)
 	}
 
 	return xInset
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -69,6 +69,11 @@ type Entry struct {
 	// Since: 2.7
 	AlwaysShowValidationError bool
 
+	// Icon is displayed at the outer left of the entry.
+	// It is not clickable, but can be used to indicate the purpose of the entry.
+	// Since: 2.7
+	Icon fyne.Resource `json:"-"`
+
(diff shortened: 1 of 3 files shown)


dweymouth (dismissed):
This looks good to me. API seems sensible and I can't think of a better name than `.Icon` for the property. The only thing is that to my eye there seems to be a bit more horizontal space between the icon and the text than I'd expect, but only when the text isn't scrolled. A minor thing that could be fixed up later or we could just leave it as-is.

Also I'm not sure how to get the flaky Ubuntu tests to retry for some reason

widget/entry_internal_test.go
```diff
@@ -295,6 +295,33 @@ func TestEntry_CallbackLocking(t *testing.T) {
...
+	entry.Wrapping = fyne.TextWrapOff
+	entry.Scroll = fyne.ScrollNone
+	icon := theme.MailComposeIcon()
+	entry.SetIcon(icon)
+	entry.SetText("SomeText")
+	r := test.TempWidgetRenderer(t, entry)
+	r.Layout(entry.MinSize())
+
+	var iconObj *canvas.Image
+	for _, obj := range r.Objects() {
```
dweymouth: You can cast r to a `*entryRenderer` and then access `.icon` directly instead of looping like this to find it.

(I'm mostly just trying to find something that can be changed for the PR to get the checks to run again :D but this (casting to the known concrete render type) is generally the pattern used in tests elsewhere)

Vinci10: I updated tests. Please review.

widget/entry.go
```diff
@@ -69,6 +69,11 @@ type Entry struct {
 	// Since: 2.7
 	AlwaysShowValidationError bool
 
+	// Icon is displayed at the outer left of the entry.
```
andydotxyz, line 75: I think this should go after the other exposed fields - in case anyone is constructing this widget using positional arguments (`&widget.Entry{"Hello", TextStyle{}, "Placeholder", ...}`.

widget/entry.go
```diff
@@ -1511,6 +1534,18 @@ func (r *entryRenderer) trailingInset() float32 {
 	return xInset
 }
 
+func (r *entryRenderer) leadingInset() float32 {
+	th := r.entry.Theme()
+	xInset := float32(0)
+
+	if r.entry.Icon != nil {
+		iconSpace := th.Size(theme.SizeNameInlineIcon) + th.Size(theme.SizeNameLineSpacing) + th.Size(theme.SizeNameInnerPadding)
```
andydotxyz: Looking at how `trialingInset` is calculated I don't think the `Padding` item is needed here.
Plus probably no point in the new `iconSpace` variable.

widget/entry.go
Vinci10: thanks, fixed

widget/entry.go
Vinci10: thanks, done

Vinci10:
> This looks good to me. API seems sensible and I can't think of a better name than `.Icon` for the property. The only thing is that to my eye there seems to be a bit more horizontal space between the icon and the text than I'd expect, but only when the text isn't scrolled. A minor thing that could be fixed up later or we could just leave it as-is.
> 
> Also I'm not sure how to get the flaky Ubuntu tests to retry for some reason

After removing `Padding` as suggested by @andydotxyz, the space between the icon and the text has decreased. I hope it looks better now.

dweymouth (approved):

PR #5794, opened 2025-06-20
Minor refactor/cleanup of desktop cursor and key handling

Jacalz:
### Description:

Use an array for looking up the cursor, we don't need a map. Refactor some code and use a fast way to do lowercase to uppercase.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/driver/glfw/window_desktop.go
+++ b/internal/driver/glfw/window_desktop.go
@@ -555,7 +555,7 @@ func keyCodeToKeyName(code string) fyne.KeyName {
 
 	char := code[0]
 	if char >= 'a' && char <= 'z' {
-		return fyne.KeyName(char ^ 32) // Simple conversion to uppercase by adding 32.
+		return fyne.KeyName(char ^ ('a' - 'A')) // Corresponding KeyName is uppercase. Convert with simple bit flip.
 	}
 
 	switch char {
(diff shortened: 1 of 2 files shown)


internal/driver/glfw/window_desktop.go
```diff
@@ -555,8 +555,7 @@ func keyCodeToKeyName(code string) fyne.KeyName {
 
 	char := code[0]
 	if char >= 'a' && char <= 'z' {
-		// Our alphabetical keys are all upper case characters.
-		return fyne.KeyName('A' + char - 'a')
+		return fyne.KeyName(char ^ 32) // Simple conversion to uppercase by adding 32.
```
andydotxyz: This feels like a case where the comment was added because the code became unclear. Can you find a happy medium perhaps? Was the previous code too slow?

Jacalz: What do you think would be a happy medium? The old code was likely more than fast enough but this should be faster as it is just a bit flip. Flipping the fifth bit to convert to uppercase is a common way to do it in ASCII (https://stackoverflow.com/questions/25821178/convert-lowercase-to-uppercase-using-ascii).

Jacalz: I must admit that I kind of find the new code to be easier to read 😅

Jacalz: I think I've made it a bit clearer now :)

andydotxyz (approved):

Issue #3660, opened 2023-02-13
When trying to resize the window in Mac (arm & amd), The content inside the window is shaking / glitching / resizing. There seems to be a lag in rendering the content.


S-Varunn:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

When i try to resize the window in Mac (M1 and Intel chip) the contents inside the window is shaking uncontrollably. The window resize is also affected by this. 

In arm the glitch is minimal but still exists. In amd it can be observed very clearly. 
I was able to reduce the glitch a bit by modifying the 

```
func (d *gLDriver) startDrawThread() {
	settingsChange := make(chan fyne.Settings)
	fyne.CurrentApp().Settings().AddChangeListener(settingsChange)
	var drawCh <-chan time.Time
	if d.drawOnMainThread {
		drawCh = make(chan time.Time) // don't tick when on M1
	} else {
		drawCh = time.NewTicker(time.Second / 60).C
	} 
```
 To 
```
func (d *gLDriver) startDrawThread() {
	settingsChange := make(chan fyne.Settings)
	fyne.CurrentApp().Settings().AddChangeListener(settingsChange)
	var drawCh <-chan time.Time
	if d.drawOnMainThread {
		drawCh = make(chan time.Time) // don't tick when on M1
	} else {
		drawCh = time.NewTicker(time.Second / 15).C    // Change here!
	}
```
But this resulted in the internal logic to slow down.

### How to reproduce

1. Run the fyne demo
2. Resize the window using the mouse
3. Observer shaking of contents in Mac m1 and Intel chips. 
4. The Irregular renders can be seen in intervals where you stop dragging or drag continuously and fast.

### Screenshots

Mac M1 :  

https://user-images.githubusercontent.com/85671418/218434296-7e243597-775e-4cf8-a67a-c5492e7880e0.mov

Mac Intel Chip (arm64) :

https://user-images.githubusercontent.com/85671418/218435771-c617138d-1c3d-4106-bd2b-34b4d9d0ff2a.mov

### Example code

Run the fyne_demo

### Fyne version

v2.3.0

### Go compiler version

1.19.5

### Operating system

macOS

### Operating system version

Ventura 13.2 

### Additional Information

Mac arm:
- OS: Ventura 13.2
- Version: 13.2
[...]

PR #5797, opened 2025-06-22
Scaling and Window Size Confusion Fixes

cwarden:
_[see https://github.com/fyne-io/examples/issues/42 for context]_

### Fix Window Unresponsiveness When Canvas Size Doesn't Match Window Size

When a Fyne application window is opened in a tiled window manager or
constrained by other window management, the canvas size could become out
of sync with the actual window dimensions, causing the window to be
unresponsive to input. This occurred because the canvas coordinate system
didn't match the window's actual size, resulting in input events being
misaligned with visual elements.

The issue was in RescaleContext(), which assumed it could resize the window
to match the canvas size. However, when the window size is constrained
(by a tiling window manager, during scale changes, or other scenarios),
this approach fails.  Use its actual dimensions to recalculate the canvas size.

This fix resolves the coordinate system mismatch that prevented proper
input handling in tiled windows and also fixes issues when moving between
monitors with different DPI scales.

### Fix Premature Scale Changes When Dragging Windows Near Monitor Boundaries

When dragging a window toward a monitor with a different DPI scale, the
window would prematurely switch to the new monitor's scale while still
on the original monitor.

Change the monitor detection algorithm from center-point detection to
intersection area calculation. The window is now associated with the
monitor that has the largest overlapping area with the window bounds.
This provides more natural behavior when dragging windows between
monitors with different scales.

### Fix Widget Layout Issues When Window Size Is Constrained By Tiling Window Manager

When running a Fyne application in a tiling window manager, widgets
could be positioned incorrectly or move out of view after content
changes.  This occurred because the application assumed its resize
requests would succeed, but tiling window managers often constrain or
ignore these requests, creating a mismatch between the expected and
actual window dimensions.

Add window size verification to the canvas Resize() function. When the
canvas is resized, the code now checks if the actual window size differs
from what would be expected for the requested canvas size. If a mismatch
is detected (indicating the window manager has constrained the size),
the canvas size is recalculated based on the actual window dimensions.


internal/driver/glfw/window_desktop.go
```diff
@@ -261,26 +261,89 @@ func (w *window) fitContent() {
 	}
 }
 
+// monitorInfo contains the information needed for monitor detection
```
andydotxyz: This is a lot of new code where simpler solutions should exist.
Following discussion on the source issue it seems that scaling the xOff and yOff then the original code should work leading to a much smaller diff.

Perhaps this belongs in a different PR though - although it's all related to windows this is about position and the other two fixes are inter-related resizing issues.

internal/driver/glfw/window_desktop.go
```diff
@@ -705,9 +768,31 @@ func (w *window) RescaleContext() {
 		return
 	}
 
-	size := w.canvas.size.Max(w.canvas.MinSize())
-	newWidth, newHeight := w.screenSize(size)
-	w.viewport.SetSize(newWidth, newHeight)
+	// When scale changes (e.g., moving between monitors), we need to recalculate
+	// the canvas size based on the current window size, not the other way around
+	if w.viewport != nil && w.visible {
```
andydotxyz: Has something changed whereby the `w.viewport` can be nil where it couldn't before? There was not a nil check before.

internal/driver/glfw/window_desktop.go
```diff
@@ -705,9 +768,31 @@ func (w *window) RescaleContext() {
...
+
+			// Update our internal size tracking
+			w.width = currentWidth
+			w.height = currentHeight
+		}
+	} else {
+		// Window not visible yet - use original behavior
+		size := w.canvas.size.Max(w.canvas.MinSize())
+		newWidth, newHeight := w.screenSize(size)
+		if w.viewport != nil {
```
andydotxyz: As above, is this nil viewport a new consideration?

internal/driver/glfw/window_desktop_monitor_test.go
```diff
@@ -0,0 +1,308 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestFindMonitorWithLargestIntersection tests the actual monitor detection algorithm
+func TestFindMonitorWithLargestIntersection(t *testing.T) {
+	// Two monitors side by side, like the original issue:
```
andydotxyz: I don't think these comments are necessary, the setup code reads well

internal/driver/glfw/window_desktop_monitor_test.go
```diff
@@ -0,0 +1,308 @@
...
+package glfw
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestFindMonitorWithLargestIntersection tests the actual monitor detection algorithm
+func TestFindMonitorWithLargestIntersection(t *testing.T) {
```
andydotxyz: It may just be a naming thing - but do be sure to test for the desired behaviour rather than the specific implementation.

internal/driver/glfw/window_desktop_monitor_test.go
```diff
@@ -0,0 +1,308 @@
...
+			result := findMonitorWithLargestIntersection(
+				tt.windowX, tt.windowY, tt.windowWidth, tt.windowHeight, tt.monitors)
+			assert.Equal(t, tt.expectedMonitor, result, tt.description)
+		})
+	}
+}
+
+// TestFindMonitorWithLargestIntersection_ScaleScenario tests the specific scenario from the bug report
+func TestFindMonitorWithLargestIntersection_ScaleScenario(t *testing.T) {
+	// This test verifies that the monitor detection works correctly when
```
andydotxyz: Also, feels duplicative of the code below - the inline comments inside the code are helpful though

internal/driver/glfw/window_test.go
```diff
@@ -2215,3 +2215,51 @@ func (s *safeWindow) keyPressed(w *glfw.Window, key glfw.Key, scancode int, acti
 		s.window.keyPressed(w, key, scancode, action, mods)
 	})
 }
+
```
andydotxyz: These two new tests don't appear to actually test anything.
The comments even talk about how the desired changes to test cannot be tested without hardware.
If that is true then I think there is no benefit to adding this

andydotxyz (changes requested):
Thanks so much for working on these fixes - great to have the underlying issues figured out.
As discussed on the issue it looks like it might be good to split between the resize and the re-scale/move issues to two PRs.

One overall PR comment is that comments should only be used where the intent of the code cannot be communicated directly. i.e. don't duplicate what is written in code.

internal/driver/glfw/window_desktop.go
cwarden: I removed this in 7f112e390cdaa728c9bc3134e7b7187ed4a67261.  Replaced by the simpler fix in #5802.  (This original fix only worked with the changes in 3a9707c36b26b8cbdbd7c79f5f0477022a194c59).

internal/driver/glfw/window_test.go
cwarden: I'll get rid of them.  It would probably be a pretty intrusive change to introduce the abstraction over glfw needed for testing, and not worth it.

internal/driver/glfw/window_desktop.go
cwarden: I thought I was getting failures in existing tests when not checking for a non-nil viewport, but I'm not seeing that now so I'll remove.

internal/driver/glfw/window_rescale_test.go
cwarden, line 1: I'm not sure how to thoroughly test these changes in a race-safe way.  Replaced these broken ones with simpler ones that use `runOnMain` in 141c41d10764ceedd7e81d9bb077d15569a8c235.

internal/driver/glfw/window_desktop.go
```diff
@@ -261,26 +261,89 @@ func (w *window) fitContent() {
 	}
 }
 
+// monitorInfo contains the information needed for monitor detection
+type monitorInfo struct {
+	X, Y, Width, Height int
```
andydotxyz: Best not to make these public fields, even though it's a private structure we go for private by default as a safety / style feature.

internal/driver/glfw/window_desktop.go
```diff
@@ -261,26 +261,89 @@ func (w *window) fitContent() {
...
+	windowLeft := windowX
+	windowRight := windowX + windowWidth
+	windowTop := windowY
+	windowBottom := windowY + windowHeight
+
+	var bestMonitorIndex int = -1
+	var bestArea int
+
+	for i, mon := range monitors {
+		// Calculate monitor bounds
```
andydotxyz: The comments inside this method are not needed - the code is clean and clear.

internal/driver/glfw/window_desktop.go
```diff
@@ -261,26 +261,89 @@ func (w *window) fitContent() {
...
-		for _, monitor := range glfw.GetMonitors() {
-			x, y := monitor.GetPos()
 
-			if x > xOff || y > yOff {
-				continue
-			}
-			if videoMode := monitor.GetVideoMode(); x+videoMode.Width <= xOff || y+videoMode.Height <= yOff {
-				continue
-			}
+		// Get all monitors and convert to monitorInfo
```
andydotxyz: Good naming = no need for these comments

internal/driver/glfw/window_desktop.go
cwarden: The struct was removed in 7f112e390cdaa728c9bc3134e7b7187ed4a67261.

internal/driver/glfw/window_desktop.go
cwarden: Code removed in 7f112e390cdaa728c9bc3134e7b7187ed4a67261.

internal/driver/glfw/window_desktop.go
cwarden: Removed in 7f112e390cdaa728c9bc3134e7b7187ed4a67261.

internal/driver/glfw/window_desktop.go
cwarden: Removed viewport check

internal/driver/glfw/window_desktop_monitor_test.go
cwarden: Removed in 7f112e390cdaa728c9bc3134e7b7187ed4a67261.

internal/driver/glfw/window_desktop_monitor_test.go
cwarden: Removed in 7f112e390cdaa728c9bc3134e7b7187ed4a67261.

internal/driver/glfw/window_desktop_monitor_test.go
cwarden: Removed in 7f112e390cdaa728c9bc3134e7b7187ed4a67261.

andydotxyz:
Please try to avoid force-pushing after reviews have start. It means we have to review from scratch on every change.

andydotxyz:
The code looks good, but I am away from my multi-monitor setup for another week so cannot really test this.

If anyone else is able to verify it that would be great.

cwarden:
> Please try to avoid force-pushing after reviews have start. It means we have to review from scratch on every change.

Sorry, that's actually one of my own guidelines.  I thought it was worth breaking based on the history of this PR, but I should have checked with you first.

andydotxyz:
> > Please try to avoid force-pushing after reviews have start. It means we have to review from scratch on every change.
> 
> Sorry, that's actually one of my own guidelines. I thought it was worth breaking based on the history of this PR, but I should have checked with you first.

cool cool. When it feels like a squash or a re-base is better you can note it on the PR and the person landing it can do that action after the reviews etc all complete :).

Jacalz (approved):
Thanks. This seems to be working great on my end using Gnome. It also seems like it is removing almost all of the shaking content problems that I have seen previously. It might resolve https://github.com/fyne-io/fyne/issues/3660 potentially.

andydotxyz:
It does make things a lot smoother - but it doesn't remove flicker all together so let's not close out the other issue.

This change seems to essentially stop the window resizing when moving from one display to another (previously it would adjust so when the scale changes the window size adapts to make the same amount of content visible).

This is a complex area and I think perhaps if the trade-off is required then it's a reasonable one - let the window be whatever size the desktop thinks it should be and just adjust the content so scale works as expected. Thoughts?

internal/driver/glfw/window_desktop.go
```diff
@@ -718,9 +718,24 @@ func (w *window) RescaleContext() {
 		return
 	}
 
-	size := w.canvas.size.Max(w.canvas.MinSize())
-	newWidth, newHeight := w.screenSize(size)
-	w.viewport.SetSize(newWidth, newHeight)
+	// When scale changes (e.g., moving between monitors), we need to recalculate
+	// the canvas size based on the current window size, not the other way around
```
andydotxyz, line 722: This seems like the source of the change in behavior - with previous setup it was correct to adjust the window size based on the content and scale so that when moving when scale changes the window changed too (worked on all except gnome moving from a low scale to high scale monitor - desktop would not allow the window to shrink).

I wonder if there is a way to fix tiling whilst maintaining the old behavior?

I don't really know why tiling breaks in this code so I'm not sure if there is a happy medium.

cwarden: Isn't the issue that fyne can't control the window size in every environment?  But it can control the canvas size so when it doesn't get the window size it wants, it needs to adjust the canvas size.

andydotxyz: That would make sense.

So instead of trying to control it with this PR it doesn't try any more? Can we somehow attempt to resize and fallback if it doesn't work?

cwarden: I don't regularly work with non-tiled windows, but the automatic resizing feels awkward to me.

as-of-19dcdbdb5

The behavior with the changes in this branch feel more intuitive to me.

with-scaling-fix-branch

Jacalz: I would want to test this PR again but I do agree that using this PR felt more natural when I moved the window between screens.

andydotxyz: > I don't regularly work with non-tiled windows, but the automatic resizing feels awkward to me

I appreciate the comparison but I would recommend doing the same with more app content :).

andydotxyz:
I think the race conditions in the CI are actually caused by this change btw - I can't see at a glance why

internal/driver/glfw/window_desktop.go
Jacalz: I think we need to compare how other apps behave when moving them between windows. This is a GTK4 app:

[Screencast From 2025-09-06 08-20-01.webm](https://github.com/user-attachments/assets/626eddf2-d66f-4f4d-a432-e954c51c5873)

This is Fyne before this change (jumps a bit in size also):

[Screencast From 2025-09-06 08-20-38.webm](https://github.com/user-attachments/assets/91c6dc9d-7995-4f82-bb80-62eeb71fdecf)

This is with this PR:

[Screencast From 2025-09-06 08-23-05.webm](https://github.com/user-attachments/assets/1011f5f6-dbe3-4e1b-8f22-821641bcf8bd)

With this, it actually behaves more like the other apps on my system and like what I would expect. The scaling and/or window size seems to potentially be a bit smaller than before though, I think?

andydotxyz: I'm not sure that comparing to GTK+ is a good idea - they have declared that fractional scaling on X11 is not possible and rely on gnome desktop hinting instead of DPI detection.
Not apples and Oranges.

If you compare with how apps work on Windows or macOS you will see the scale changes *and* the size changes so that the UI is a consistent size unless user specifies other default for the monitors.

andydotxyz: If you want the GTK version then turn off DPI detection completely rather than removing it from the codebase.

Jacalz: > I'm not sure that comparing to GTK+ is a good idea - they have declared that fractional scaling on X11 is not possible and rely on gnome desktop hinting instead of DPI detection.
Not apples and Oranges.

What does their opinion on X11 and fractional scaling have to do with it? I am using Wayland and have 125% fractional scaling enabled on the right monitor. It works perfectly. The behaviour of this app seems more sane after this change than before.

andydotxyz: I'm sorry - are you doing all of this testing with pure-wayland compiled Fyne? I did not realise that was the case. If so it is a different code path and I will have to re-analyse as the wayland scale handling is not the same as X11.

PR #5802, opened 2025-06-23
Use Scaled Monitor Size For Monitor Detection

cwarden:
Fix monitor detection by using the candidate monitor's scaled
dimensions.

Fixes bug in which scale changes prematurely as window gets near monitor with a different scale.

### DIFF
--- a/internal/driver/glfw/window_desktop.go
+++ b/internal/driver/glfw/window_desktop.go
@@ -272,13 +272,13 @@ func getMonitorScale(monitor *glfw.Monitor) float32 {
 }
 
 // getScaledMonitorSize returns the monitor dimensions adjusted for scaling
-func getScaledMonitorSize(monitor *glfw.Monitor) (int, int) {
+func getScaledMonitorSize(monitor *glfw.Monitor) fyne.Size {
 	videoMode := monitor.GetVideoMode()
 	scale := getMonitorScale(monitor)
 
-	scaledWidth := int(float32(videoMode.Width) / scale)
-	scaledHeight := int(float32(videoMode.Height) / scale)
-	return scaledWidth, scaledHeight
+	scaledWidth := float32(videoMode.Width) / scale
+	scaledHeight := float32(videoMode.Height) / scale
+	return fyne.NewSize(scaledWidth, scaledHeight)
 }
 
 func (w *window) getMonitorForWindow() *glfw.Monitor {
@@ -297,8 +297,8 @@ func (w *window) getMonitorForWindow() *glfw.Monitor {
 				continue
 			}
 
-			scaledWidth, scaledHeight := getScaledMonitorSize(monitor)
-			if x+scaledWidth <= xOff || y+scaledHeight <= yOff {
+			scaledSize := getScaledMonitorSize(monitor)
+			if x+int(scaledSize.Width) <= xOff || y+int(scaledSize.Height) <= yOff {
 				continue
 			}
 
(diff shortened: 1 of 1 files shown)


internal/driver/glfw/window_desktop.go
```diff
@@ -261,6 +261,21 @@ func (w *window) fitContent() {
 	}
 }
 
+// getScaledMonitorSize returns the monitor dimensions adjusted for scaling
+func getScaledMonitorSize(monitor *glfw.Monitor) (int, int) {
+	videoMode := monitor.GetVideoMode()
+	widthMm, heightMm := monitor.GetPhysicalSize()
+
+	var scale float32 = 1.0
+	if runtime.GOOS != "linux" || widthMm != 60 || heightMm != 60 { // Steam Deck incorrectly reports 6cm square!
```
andydotxyz: Let's avoid duplicating this workaround - I know the code is not identical is but the intent is the same but inverted, can it be factored out in some way so they use the same check for scale being 1 on the steam deck?

cwarden:
> Thanks for this. A comment inline about a potential tidy.
> 
> However a larger question... what are the return units? We usually use int for pixels only but this is no longer a pixel size I don't think. If it is back to regular Fyne coordinates then float32 is what we use.

It is pixels.  Here's layout I'm using where I ran into the issue.

2025-06-24-075739_985x263_scrot

The monitor on the left has a native resolution of 2560x1600, which is scaled to 1920x1200, the ~1.6 scaling.  So the function should return (1920, 1200) for the first monitor.  The other two are not scaled so it returns their native resolutions of (2560, 1440).

cwarden:
> > Thanks for this. A comment inline about a potential tidy.
> > However a larger question... what are the return units? We usually use int for pixels only but this is no longer a pixel size I don't think. If it is back to regular Fyne coordinates then float32 is what we use.
> 
> It is pixels. Here's layout I'm using where I ran into the issue.
> 
> 2025-06-24-075739_985x263_scrot
> 
> The monitor on the left has a native resolution of 2560x1600, which is scaled to 1920x1200, the ~1.6 scaling. So the function should return (1920, 1200) for the first monitor. The other two are not scaled so it returns their native resolutions of (2560, 1440).

Hmm, it's actually not returning (1920, 1200), but (1625, 1015).  So I think you're right that these are Fyne coordinates.  (My mental model of Fyne's scaling is clearly still incomplete.)

internal/driver/glfw/window_desktop.go
cwarden: Added a `getMonitorScale` function in 7dd0e39fb7e890e8f7f8dca746c2a7820d1bfa62.

internal/driver/glfw/window_desktop.go
```diff
@@ -261,6 +261,26 @@ func (w *window) fitContent() {
...
+	widthMm, heightMm := monitor.GetPhysicalSize()
+	if runtime.GOOS == "linux" && widthMm == 60 && heightMm == 60 { // Steam Deck incorrectly reports 6cm square!
+		return 1.0
+	}
+	widthPx := monitor.GetVideoMode().Width
+	return calculateDetectedScale(widthMm, widthPx)
+}
+
+// getScaledMonitorSize returns the monitor dimensions adjusted for scaling
+func getScaledMonitorSize(monitor *glfw.Monitor) (int, int) {
```
andydotxyz: Thanks this is cleaner now, but I do think that the return from this should be `fyne.Size` because (if I understand it correctly) that is the coordinate system being used for the return.

cwarden: Updated in fc67f022064bbb8cff6ec7cf1edfc1f0d3097058.

andydotxyz:
Thanks so much @cwarden and congratulations on landing your first Fyne code contribution :)

PR #5805, opened 2025-06-24
Adjust backspace handling so we ignore hardware key events during soft keyboard editing

### LINKED ISSUE #2774: Text entry widget backspace deletes two characters on Android
Possibly related issues: #1941, #165

### Describe the bug:

On Android, pressing backspace in an editable text entry widget deletes two characters.

### To Reproduce:

1. Clone [this github repository](https://github.com/xxxserxxx/fyneissue2)
2. Package for Android (`fyne package -os android`)
3. Push to an android device, install, and run
4. Select the text entry widget at the top of the screen
5. Enter some text (e.g., "abcdefg")
6. Press the keyboard's backspace (⌫) key

Two characters will be deleted. This happens from wherever in the string the cursor is, e.g. the end or the middle.

### Example code:

Sample code is in a [github repository](https://github.com/xxxserxxx/fyneissue). It is, essentially, a text entry widget with a list widget below. The text entry widget has an `OnSubmitted` lambda that prepends the text to the list.  The behavior happens without the `OnSubmitted` closure as well.

 - **OS:** Android
 - **Version:** Samsung One UI; Android 12. Built-in Samsung soft keyboard.
 - **Go version:** 1.17.7
 - **Fyne version:** v2.1.2

andydotxyz:
(More of a mystery is why those duplicate keys happen backwards!?)

Fixes #2774

### Checklist:

- [ ] Tests included. <- I cannot replicate on emulator or code
- [x] Tests all pass.

### DIFF
--- a/internal/driver/mobile/device.go
+++ b/internal/driver/mobile/device.go
@@ -9,6 +9,8 @@ import (
 
 type device struct {
 	safeTop, safeLeft, safeBottom, safeRight int
+
+	keyboardShown bool
 }
 
 //lint:file-ignore U1000 Var currentDPI is used in other files, but not here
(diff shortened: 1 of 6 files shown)


internal/driver/mobile/device.go
```diff
@@ -45,14 +47,14 @@ func (*device) HasKeyboard() bool {
 	return false
 }
 
-func (*device) ShowVirtualKeyboard() {
-	showVirtualKeyboard(mobile.DefaultKeyboard)
+func (dev *device) ShowVirtualKeyboard() {
```
Jacalz, line 50: Isn't single letter names for polymer receivers the standard/ideomatic way in Go?

Jacalz (commented):
Looking good. Just a quick question inline.

internal/driver/mobile/device.go
andydotxyz: I was just matching what was there. I can tidy that up on another PR - there is a conflict internally to do that refactor which I will get to the bottom of.

Issue #5799, opened 2025-06-23
Ability to set password keyboard


rad756:
### Checklist

- [x] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [x] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

On android when apps usually ask me to input password, I get the option on my keyboard to open password manager which will paste the selected password into the entry, would be a good addition to support this in Fyne. 

### Describe the solution you'd like to see.

Add ability to set keyboard/entry to include password manager prompt on mobile.

andydotxyz:
Hmm, password should be working with `widget.NewPasswordEntry()` - are you not seeing the right keyboard type when you use that?
Or does it need some sort of button added to open password manager if it is present?

Can you be more specific about the interaction expected, maybe with a screenshot?

rad756:
First screenshot will show how the keyboard should appear and the second how it appears on the fyne demo app

Working correctly 
![Image](https://github.com/user-attachments/assets/13d507ee-04ed-4905-8e4f-7fb599da72c4)

Not working
![Image](https://github.com/user-attachments/assets/c38b9be6-341c-4d91-8553-e9657ecbc2b9)

If you can notice in fyne demo the keyboard is missing open proton pass button, which is a password manager.

andydotxyz:
OK, thanks! 

Can you confirm which this keyboard is, perhaps not all will support the integration (for our testing notes)

rad756:
I tried it with Samsung and FUTO keyboard (this one is pictured above) and both functioned the same way.

andydotxyz:
See if that PR improves things. It may have not been the correct password hint that we set.

rad756:
So I would need to pull the commit and build fyne demo against that, install the new APK to my phone and check if it pulls up the password manager on the password field?

andydotxyz:
Yup - but also build with the `fyne` command from tools with the updated dex applied too - the change is across two repos unfortunately.

andydotxyz:
Hmm, in fact for the password fix you might only need the PR from the tools repo as the .java is not actually loaded by the library.

rad756:
Built fyne with https://github.com/andydotxyz/fyne-tools/tree/fix/updatedex-5666
Packaged fyne demo for android, password manager still does not appear in the form widget's password field

rad756:
I had a change in the samsung keyboard!
[...]

PR #5806, opened 2025-06-24
Fix up. some keyboard issues on Android

### LINKED ISSUE #5666: On mobile, first character typed into Entry is prefixed with "0" after focus
### Checklist

### Describe the bug

On mobile, when a widget.Entry is focused and the user types something for the first time, the input always starts with a "0" followed by the typed character (e.g. typing S results in 0S). This only happens once per focus. If the user changes focus and types again, it repeats. It also affects fields like passwords where the first character might be "0" intentionally.

### How to reproduce

Create a basic Entry widget (e.g. a username or password input).

Run the app on a mobile device.

Tap into the Entry field and type a letter (e.g. "S").

The result will be "0S" instead of just "S".

Delete the input and type again — subsequent input works as expected.

Change focus to another Entry and return — issue occurs again on first input.

### Screenshots

Image

### Example code

```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/dialog"
	"fyne.io/fyne/v2/widget"
)

func main() {
	myApp := app.New()
	myWindow := myApp.NewWindow("Entry Bug Example")

	usernameEntry := widget.NewEntry()
	usernameEntry.PlaceHolder = "Username..."

	passwordEntry := widget.NewEntry()
	passwordEntry.PlaceHolder = "Password..."
	passwordEntry.Password = true

	dialog.NewCustomConfirm("Login", "Login", "Cancel",
		container.NewVBox(usernameEntry, passwordEntry),
		func(confirmed bool) {},
		myWindow).Show()

	myWindow.ShowAndRun()
}

```

### Fyne version

fyne cli version: v1.0.0-alpha1

### Go compiler version

go version go1.24.0 windows/amd64

windows 11 home

### Additional Information

_No response_

andydotxyz:
Fixes #5666

Attempting fix on #5799 in passing.

Important: to test this the java needs to be applied into the `fyne` cli from tools repo, then the android app re-packaged.

### Checklist:

- [ ] Tests included. <- cannot replicate either, so no idea how to add test
- [x] Tests all pass.

### DIFF
--- a/internal/driver/mobile/app/GoNativeActivity.java
+++ b/internal/driver/mobile/app/GoNativeActivity.java
@@ -121,7 +121,7 @@ public class GoNativeActivity extends NativeActivity {
 
                 // always place one character so all keyboards can send backspace
                 ignoreKey = true;
-                mTextEdit.setText("0");
+                mTextEdit.setText(" ");
                 mTextEdit.setSelection(mTextEdit.getText().length());
                 ignoreKey = false;
 
@@ -257,7 +257,7 @@ public class GoNativeActivity extends NativeActivity {
                 addContentView(mTextEdit, mEditTextLayoutParams);
 
                 // always place one character so all keyboards can send backspace
-                mTextEdit.setText("0");
+                mTextEdit.setText(" ");
                 mTextEdit.setSelection(mTextEdit.getText().length());
 
                 mTextEdit.addTextChangedListener(new TextWatcher() {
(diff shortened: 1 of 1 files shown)


internal/driver/mobile/app/GoNativeActivity.java
```diff
@@ -121,7 +121,7 @@ public boolean onEditorAction(TextView v, int actionId, KeyEvent event) {
 
                 // always place one character so all keyboards can send backspace
                 ignoreKey = true;
-                mTextEdit.setText("0");
+                mTextEdit.setText(" ");
```
Jacalz, line 124: I don't know how this works but given that the issue notes a 0 being prefixed, would a space not show up as well? I'm thinking that setting empty string might make more sense?

andydotxyz: No, it must be 1 rune wide at least or the "backspace" will not be possible. This is required because what is in the Fyne Entry can be much more than the user has just typed into the current Android context.

My thinking is that suggestions on "0ar" doesn't offer sensible autosuggestion but " ar" will suggest "are" etc, based on even this peculiar keyboard using space characters as word separators.

Issue #5398, opened 2025-01-11
Submenus not working on mobile


pneumaticdeath:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

When I have menuItmes with child menus they can't be accessed when in a mobile layout

### How to reproduce

1. Run code below with `--tags miobile`
2. Select menu Main
3. Select menu Foo
4. Unable to select Bar or Baz menu

### Screenshots

<img width="712" alt="Screenshot 2025-01-11 at 13 54 50" src="https://github.com/user-attachments/assets/7ed85a9a-8037-4b72-8510-fa6b754e4ec3" />

### Example code

```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Multi layered menus")

	entry := widget.NewEntry()

	barMI := fyne.NewMenuItem("Bar", func() { entry.Text = "Bar"; entry.Refresh() })
	bazMI := fyne.NewMenuItem("Baz", func() { entry.Text = "Baz"; entry.Refresh() })
	barbazMenu := fyne.NewMenu("Foo", barMI, bazMI)
	fooMI := fyne.NewMenuItem("Foo", func() {})
	fooMI.ChildMenu = barbazMenu
	fooMenu := fyne.NewMenu("Main", fooMI)
	mainMenu := fyne.NewMainMenu(fooMenu)

	w.SetMainMenu(mainMenu)
	w.SetContent(entry)

	w.ShowAndRun()
}
```

### Fyne version

v2.5.3

### Go compiler version

1.23.4

### Operating system and version

macOS Sonoma 14.6.1

### Additional Information

works fine in desktop mode

pneumaticdeath:
Then I'd classify this as a documentation bug, since the behavior is inconsistent with desktop.

andydotxyz:
Consistency is more important than better docs in this situation I think, opened a PR to resolve that - sorry for the confusion...

andydotxyz:
Fixed on develop. Beware that we are going through some threading changes there right now so I wouldn't switch immediately but in a couple of days it should be safer - we are making a smoother migration path so that health warnings are not required ;)

PR #5807, opened 2025-06-24
menu item with SubMenu was no longer tappable on desktop (Action neve…

matwachich:
### Description:
MenuItems with SubMenus where no longer responding to user tap on desktop (Action never called)

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/widget/menu_item.go
+++ b/widget/menu_item.go
@@ -122,21 +122,21 @@ func (i *menuItem) Tapped(*fyne.PointEvent) {
 	if i.Item.Disabled {
 		return
 	}
-	if i.Item.Action == nil {
-		if fyne.CurrentDevice().IsMobile() {
+	if fyne.CurrentDevice().IsMobile() {
+		// idea to discuss: if the item is already activated, then call trigger
+		if i.Item.Action == nil || i.Item.ChildMenu != nil {
 			i.activate()
+			return
 		}
 
-		return
-	} else if i.Item.ChildMenu != nil {
-		if fyne.CurrentDevice().IsMobile() {
-			i.activate()
+		i.trigger()
+	} else {
+		if i.Item.Action == nil {
+			return
 		}
 
-		return
+		i.trigger()
 	}
-
-	i.trigger()
 }
 
 func (i *menuItem) activate() {


matwachich:
I made a clear separation between mobile path and desktop path.
In mobile, I keep the same behaviour: item with submenu is not tappable : when tapped, it call .activate and not .trigger
In desktop, no need to call .activate, .trigger is called whenever the is submenu or not

andydotxyz:
I'm not sure this is a great approach - it means that some items will be tappable on desktop but not on mobile. Unless there is a clear reason to such inconsistencies should be avoided.

See original cause for this change https://github.com/fyne-io/fyne/issues/5398

matwachich:
totally agree.
Another idea: in mobile, secondary (long press) tapping an item will trigger Action ?

andydotxyz:
I guess to be able to comment on that we need to explore the use-case rather than just technically what is possible.

I tapped through some apps and cannot find one where tapping a parent menu item does something - so what's the journey here that we are solving for?

matwachich:
We could be the first toolkit to propose this behaviour hhhh

IMHO, we are trying to solve the fact that before, tapping an item with submenu would trigger Action, and now, it does not.
So I thought of this solution, that will not change the current behaviour.

matwachich:
Another idea: tapping an already selected/activated item will trigger action?
(but here, I don't know how to get activation status of an item... will have to investigate)

andydotxyz:
I guess tapping an active item leading to calling the `Action` would be almost consistent and allow access to all features on all platforms.
Though I'm still not sure what is the expected use-case - as a user I had always assumed the action of a menu item was either to tap or to expand...

matwachich:
I don't know other usecases, but here is mine :
I made a dynamic menu to select predefined time periodes: years, months, weeks, and days
Tapping a menu item with sub menu is absolutely needed here.

Yes, I could do it another way, but this is the most clear way I found.

image

andydotxyz:
As we have to make sure functionality works everywhere the best suggestion so far is "tapping an active item with action will call it".
Personally I would add an item at the top of each something like "All Days" or "Complete Year" which is 100% discoverable.
I'm just not a fan of user interactions that are hidden - and in the case of menus this feels unexpected.

matwachich:
Agreed, I made it in my app, it seems more idiomatic.
Should we close the PR?

andydotxyz:
Thanks for understanding, I think that sounds good.

PR #5811, opened 2025-06-25
Calendar_additions

matwachich:
### Description:

Some enhancements to Calendar widget and DateEntry

- added localized month name
- added a function to programmatically change the currently displayed date
- added the ability to select a date, and a function to programmatically change the selected date
- fixed crash when OnChanged callback in not set
- calendar of DateEntry will now open with currently entered date selected
- corrected Calendar popup positioning (do not resize the popup, keep it always at its minSize)

What to discuss next:

- adding a localized button under the Calendar in DateEntry to select today's date
- adding a config variable WeekStart to Calendar to override configured first Week day
- adding something similar for date format to DateEntry (did not do it, will have to study it)
- discussing is it judicious and idiomatic that tapping the selected date in the calendar will unselect it (setting SelectedDate = time.Time{})

Another one I just want to understand:
- Why using in DateEntry a *time.Time and not a time.Time ??? We can just set Date = time.Time{} (Zero-date) instead on nil

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/widget/calendar.go
+++ b/widget/calendar.go
@@ -1,10 +1,12 @@
 package widget
 
 import (
+	"fmt"
 	"math"
 	"strconv"
 	"strings"
 	"time"
+	"unicode"
 
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/lang"
@@ -27,13 +29,17 @@ var minCellContent = NewLabel("22")
 // Since: 2.6
 type Calendar struct {
 	BaseWidget
-	currentTime time.Time
+	displayedMonth time.Time
 
 	monthPrevious *Button
 	monthNext     *Button
 	monthLabel    *Label
 
 	dates *fyne.Container
+
+	SelectedDate time.Time // currently selected date
+
+	// TODO discuss a config variable WeekStart time.Weekday to override week's first day
 
 	OnChanged func(time.Time) `json:"-"`
 }
@@ -43,30 +49,49 @@ type Calendar struct {
 // Since: 2.6
 func NewCalendar(cT time.Time, changed func(time.Time)) *Calendar {
 	c := &Calendar{
-		currentTime: cT,
-		OnChanged:   changed,
+		displayedMonth: cT,
+		OnChanged:      changed,
 	}
 
 	c.ExtendBaseWidget(c)
 	return c
 }
 
+// SetDisplayedDate sets the currently displayed year and month
+func (c *Calendar) SetDisplayedDate(date time.Time) {
+	if date.IsZero() {
+		date = time.Now()
+	}
+
+	// Dates are 'normalised', forcing date to start from the start of the month ensures move from March to February
+	c.displayedMonth = time.Date(date.Year(), date.Month(), 1, 0, 0, 0, 0, time.Local)
+
+	// check if label has been created to avoid nil pointer access panic if function is called before
+	// widget instanciation
+	if c.monthLabel != nil {
+		c.monthLabel.SetText(c.monthYear())
+		c.dates.Objects = c.calendarObjects()
+	}
+}
+
+// SetSelectedDate sets the currently selected date
+//
+// Date selection works only if .Selectable = true
+func (c *Calendar) SetSelectedDate(date time.Time) {
+	c.SelectedDate = date
+	c.updateSelection()
+}
+
 // CreateRenderer returns a new WidgetRenderer for this widget.
 // This should not be called by regular code, it is used internally to render a widget.
 func (c *Calendar) CreateRenderer() fyne.WidgetRenderer {
 	c.monthPrevious = NewButtonWithIcon("", theme.NavigateBackIcon(), func() {
-		c.currentTime = c.currentTime.AddDate(0, -1, 0)
-		// Dates are 'normalised', forcing date to start from the start of the month ensures move from March to February
-		c.currentTime = time.Date(c.currentTime.Year(), c.currentTime.Month(), 1, 0, 0, 0, 0, c.currentTime.Location())
-		c.monthLabel.SetText(c.monthYear())
-		c.dates.Objects = c.calendarObjects()
+		c.SetDisplayedDate(c.displayedMonth.AddDate(0, -1, 0))
 	})
 	c.monthPrevious.Importance = LowImportance
 
 	c.monthNext = NewButtonWithIcon("", theme.NavigateNextIcon(), func() {
-		c.currentTime = c.currentTime.AddDate(0, 1, 0)
-		c.monthLabel.SetText(c.monthYear())
-		c.dates.Objects = c.calendarObjects()
+		c.SetDisplayedDate(c.displayedMonth.AddDate(0, 1, 0))
 	})
 	c.monthNext.Importance = LowImportance
 
@@ -110,12 +135,12 @@ func (c *Calendar) calendarObjects() []fyne.CanvasObject {
 }
 
 func (c *Calendar) dateForButton(dayNum int) time.Time {
-	oldName, off := c.currentTime.Zone()
-	return time.Date(c.currentTime.Year(), c.currentTime.Month(), dayNum, c.currentTime.Hour(), c.currentTime.Minute(), 0, 0, time.FixedZone(oldName, off)).In(c.currentTime.Location())
+	oldName, off := c.displayedMonth.Zone()
+	return time.Date(c.displayedMonth.Year(), c.displayedMonth.Month(), dayNum, c.displayedMonth.Hour(), c.displayedMonth.Minute(), 0, 0, time.FixedZone(oldName, off)).In(c.displayedMonth.Location())
 }
 
 func (c *Calendar) daysOfMonth() []fyne.CanvasObject {
-	start := time.Date(c.currentTime.Year(), c.currentTime.Month(), 1, 0, 0, 0, 0, c.currentTime.Location())
+	start := time.Date(c.displayedMonth.Year(), c.displayedMonth.Month(), 1, 0, 0, 0, 0, c.displayedMonth.Location())
 	var buttons []fyne.CanvasObject
 
 	dayIndex := int(start.Weekday())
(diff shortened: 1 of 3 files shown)


widget/calendar.go
```diff
@@ -27,14 +29,18 @@ var minCellContent = NewLabel("22")
...
-	currentTime time.Time
+	displayedMonth time.Time
 
 	monthPrevious *Button
 	monthNext     *Button
 	monthLabel    *Label
 
 	dates *fyne.Container
 
+	SelectedDate time.Time // currently selected date
```
dweymouth: We probably should name this something that is future-compatible with supporting a selected date range. Maybe `SelectedDateStart`? But it is a bit confusing before we add `SelectedDateEnd`, so not sure.

Also, this and all new public APIs will need a `Since: 2.7` line.

widget/date_entry.go
```diff
@@ -12,7 +14,7 @@ import (
 // Since: 2.6
 type DateEntry struct {
 	Entry
-	Date      *time.Time
+	Date      *time.Time       // TODO discuss: why not just use time.Time{} (Zero-time) ?
```
dweymouth: This can't be changed to `time.Time` since it's a public API. Maybe a non-pointer field may have been better but it's too late to change it (except for maybe Fyne 3.0)

widget/calendar.go
```diff
@@ -43,30 +49,49 @@ type Calendar struct {
...
+	// widget instanciation
+	if c.monthLabel != nil {
+		c.monthLabel.SetText(c.monthYear())
+		c.dates.Objects = c.calendarObjects()
+	}
+}
+
+// SetSelectedDate sets the currently selected date
+//
+// Date selection works only if .Selectable = true
```
dweymouth: Did you forget to add the `.Selectable` property that was mentioned here?

widget/date_entry.go
```diff
@@ -117,9 +119,9 @@ func (e *DateEntry) Move(pos fyne.Position) {
 // Implements: fyne.Widget
 func (e *DateEntry) Resize(size fyne.Size) {
 	e.Entry.Resize(size)
-	if e.popUp != nil {
+	/*if e.popUp != nil {
```
dweymouth: why is this code now commented out?

widget/calendar.go
matwachich: Agree for this.
Why not 2 variables in the struct: `SelectedDateStart` and `SelectedDateEnd` ; and another boolean `RangeSelection`
if this boolean is false (default behaviour), then `SelectedDateEnd` is always ignored

Then we add 2 setter functions:
`SetSelectedDate` that will set only `SelectedDateStart`, and `SetSelectedDateRange` that wil set the two variables

Edit: ah yes, also forgot the OnChanged callback should be adapted `OnChanged(start, end time.Time)`

widget/date_entry.go
matwachich: Too late yes...

widget/date_entry.go
matwachich: IMO it should be removed, but I kept it to discuss it.
Calendar popup should never be resized I think.

andydotxyz: Fair. I don't have a strong feeling either way. 
Except commented out code is bad ;)

widget/date_entry.go
andydotxyz: My understanding is that a zero time is a valid time but a null time represents the field missing. 
Apologies if I got that wrong.

widget/calendar.go
andydotxyz: This ties into the multi/range selection discussions of table/list etc and needs to match the pattern discussed there.

We should not implement a range system quickly here where it is not figured out in general.

As this selector does not currently support ranges `SelectDate` works in a future-proof way

andydotxyz: I imagine a date range selector would need at least a different constructor for the different OnChanged.

widget/calendar.go
```diff
@@ -43,30 +49,49 @@ type Calendar struct {
...
+		displayedMonth: cT,
+		OnChanged:      changed,
 	}
 
 	c.ExtendBaseWidget(c)
 	return c
 }
 
+// SetDisplayedDate sets the currently displayed year and month
+func (c *Calendar) SetDisplayedDate(date time.Time) {
```
andydotxyz: Isn't the current /display/ at month granularity not date granularity?

widget/calendar.go
andydotxyz: Looks like that was removed. I don't think it is needed

widget/calendar.go
```diff
@@ -27,14 +29,18 @@ var minCellContent = NewLabel("22")
...
 
 	monthPrevious *Button
 	monthNext     *Button
 	monthLabel    *Label
 
 	dates *fyne.Container
 
+	SelectedDate time.Time // currently selected date
+
+	// TODO discuss a config variable WeekStart time.Weekday to override week's first day
```
andydotxyz: See widget/locale.go

widget/date_entry.go
```diff
@@ -154,11 +160,27 @@ func (e *DateEntry) setupDropDown() *Button {
...
+			e.dropDown.SetDisplayedDate(*e.Date)
+		} else {
+			e.dropDown.SetSelectedDate(time.Time{})
+			e.dropDown.SetDisplayedDate(time.Time{})
+		}
+
 		c := fyne.CurrentApp().Driver().CanvasForObject(e.super())
 
-		e.popUp = NewPopUp(e.dropDown, c)
+		e.popUp = NewPopUp(&fyne.Container{
```
andydotxyz, line 163: container.NewVBox is a lot cleaner.

widget/date_entry.go
matwachich: From golang documentation:

Programs using times should typically store and pass them as values, not pointers. That is, time variables and struct fields should be of type [time.Time](https://pkg.go.dev/time#Time), not *time.Time.

...

The zero value of type Time is January 1, year 1, 00:00:00.000000000 UTC. As this time is unlikely to come up in practice, the [Time.IsZero](https://pkg.go.dev/time#Time.IsZero) method gives a simple way of detecting a time that has not been initialized explicitly.

---

Is it worth introducing a small breaking change?

widget/calendar.go
matwachich: Agree.
There should be 2 "different" widgets
`NewCalendar` ; `OnChanged(time.Time)` ; `SetSelectedDate(time.Time)` ; `GetSelectedDate() time.Time`
`NewCalendarRange` ; `OnChanged(time.Time, time.Time)` ; `SetSelectedRange(time.Time, time.Time)` ; `GetSelectedRange() (time.Time, time.Time)`

What do you think of making SelectedDate a private field?

widget/date_entry.go
matwachich: Impossible due to import cycle... (container imports widget and widgets imports container...)

widget/calendar.go
matwachich: Yes, I saw the mecanism used to selected first week day.
But I thought it would good for people like me, in obscure non-typical countries, to be able to modify first week day

widget/calendar.go
matwachich: Yes, `SetDisplayedMonth` seems more idiomatic

widget/date_entry.go
dweymouth: > Is it worth introducing a small breaking change?

No. Fyne commits fully to the rule of no breaking API changes in minor (2.x.y) releases. We could consider adding it to the list of potential changes in 3.0 if it is ever needed.

Jacalz:
Thanks for this. Though, I would like to point out that PRs that introduce new bigger features are usually easier to review and land if they focus on one at a time. Fixes and smaller changes can often be grouped together though. Not something you need to change now but it is worth keeping in mind m

widget/date_entry.go
```diff
@@ -12,7 +14,7 @@ import (
 // Since: 2.6
 type DateEntry struct {
 	Entry
-	Date      *time.Time
+	Date      *time.Time       // TODO discuss: using time.Time
```
Jacalz: I'd say yes, this is probably the first time I've seen a pointer to it.

widget/calendar.go
```diff
@@ -238,3 +299,9 @@ func shortDayName(in string) string {
 	long := lang.X(lower, in)
 	return strings.ToUpper(lang.X(key, long[:3]))
 }
+
+func monthName(in string) string {
+	r := []rune(lang.X(strings.ToLower(in), in))
+	r[0] = unicode.ToUpper(r[0])
```
Jacalz: Should the months always be displayed in upper case? In Swedish for example, we don't write months (or week days for that matter) with capital letters unless required for some other reason (like start of sentence etc.).

widget/calendar.go
```diff
@@ -145,20 +168,58 @@ func (c *Calendar) daysOfMonth() []fyne.CanvasObject {
...
+			if b, ok := c.dates.Objects[i].(*Button); ok {
+				b.Importance = LowImportance
+				b.Refresh()
+			}
+		}
+		return
+	}
+
+	for i := 0; i < len(c.dates.Objects); i++ {
+		if b, ok := c.dates.Objects[i].(*Button); ok {
```
Jacalz, line 215: Not a requested change but just a note: I wonder if there is some more clever way this data can be stored internally so we don't need to type assert this? Might make the code a bit cleaner

widget/date_entry.go
dweymouth: It's a breaking change though, it's a pre-existing field. Could be added to the 3.0 list but can't be done now

widget/calendar.go
matwachich: Mmm yes
So translation files should reflect how is the names written in each language, and then no transformation should be done in the code

widget/date_entry.go
Jacalz: Good point. We should update the 3.0 issue about that

widget/calendar.go
andydotxyz: That would match how we are handling it in collections at this time

widget/calendar.go
andydotxyz: It could be an additional Fyne setting I guess, if you want to override what we get from the system..

widget/date_entry.go
Jacalz: There is no such thing as "a small braking change" IMHO. We can't know how people are using the API.

widget/calendar.go
matwachich: OK, so I'll make a feature request to track this

widget/date_entry.go
matwachich: Yep understood.
This wont resolve until 3.0

widget/calendar.go
matwachich: the simplest way that comes to my mind is to track the first and last index of actual buttons.
but is it worth the complexification ? I don't think so.
traversing at most 40'ish interfaces should not be a big performance bottleneck

widget/date_entry.go
andydotxyz: Thanks everyone :) glad to defend the no-breakages rule!

widget/calendar.go
```diff
@@ -43,30 +46,52 @@ type Calendar struct {
...
-		OnChanged:   changed,
+		displayedMonth: cT,
+		OnChanged:      changed,
 	}
 
 	c.ExtendBaseWidget(c)
 	return c
 }
 
+// SetDisplayedMonth sets the currently displayed year and month
```
Jacalz: Let's add a terminating dot at the end of the sentence and a `Since: 2.7` line two rows below.

widget/calendar.go
```diff
@@ -43,30 +46,52 @@ type Calendar struct {
...
+	}
+}
+
+// SetSelectedDate sets the currently selected date. Passing ZeroDate (time.Time{}) will cause no selection.
+func (c *Calendar) SetSelectedDate(date time.Time) {
+	c.selectedDate = date
+	c.updateSelection()
+}
+
+// GetSelectedDate returns the currently selected date, or ZeroTime (time.Time{}) if no date is selected.
```
Jacalz, line 82: We need `Since:` lines here as well.

Jacalz (changes requested):
Sorry for the late review. I added some minor comments inline. It would also be good to add some tests for this. Also, please do tick the boxes in the PR template as appropriately.

matwachich:
Ok done for the inline comments.
For the tests, I have to admit that... I never wrote tests for fyne, nor for golang...
I vaguely know the principle, but have no clue how to really write usefull tests...
If someone could help about this :)

Issue #5815, opened 2025-06-28
Two textgrids inside NewHSplit are considerably laggy


redawl:
### Checklist

- [x] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [x] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

When I have two textgrids inside a HSplit container, dragging the splitter from left to right causes noticeable lag.

### How to reproduce

1. Run example code
2. Drag split back and forth.
3. Observe considerable lagginess when dragging.

### Example code

```go
package main

import (
        "fyne.io/fyne/v2"
        "fyne.io/fyne/v2/app"
        "fyne.io/fyne/v2/container"
        "fyne.io/fyne/v2/widget"
)

func main() {
        a := app.New()
        w := a.NewWindow("Test")
        w.Resize(fyne.NewSize(800, 800))

        grid1 := widget.NewTextGrid()
        grid2 := widget.NewTextGrid()

        w.SetContent(container.NewHSplit(grid1, grid2))
        w.ShowAndRun()
}
```

### Fyne version

develop

### Go compiler version

1.24.4

### Operating system and version

Gentoo

redawl:
It seems like part of the problem here is that walking the object tree takes a significant amount of time the larger the `TextGrid` is, given there are three canvas objects per cell. This is a problem in this case because moving the split from left to right causes both grids to increase in size permanently, since there is currently no mechanism in `TextGrid` to shrink row object slices when their text content shrinks.

redawl:
Fixed on develop

PR #5816, opened 2025-06-29
canvas: only move if pos is different than previous

redawl:
### Description:
Only `Move` the canvas primitive if the passed in pos is different from the previous position.

Since these canvas primitives call `repaint` inside `Move`, there can be a significant performance penalty in some cases without this optimization.

For example, #5815 is helped significantly by this change.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/canvas/circle.go
+++ b/canvas/circle.go
@@ -41,6 +41,9 @@ func (c *Circle) MinSize() fyne.Size {
 
 // Move the circle object to a new position, relative to its parent / canvas
 func (c *Circle) Move(pos fyne.Position) {
+	if c.Position1.X == pos.X && c.Position1.Y == pos.Y {
+		return
+	}
 	size := c.Size()
 	c.Position1 = pos
 	c.Position2 = c.Position1.Add(size)
--- a/canvas/gradient.go
+++ b/canvas/gradient.go
@@ -68,6 +68,9 @@ func (g *LinearGradient) Hide() {
 
 // Move the gradient to a new position, relative to its parent / canvas
 func (g *LinearGradient) Move(pos fyne.Position) {
+	if g.Position().X == pos.X && g.Position().Y == pos.Y {
+		return
+	}
 	g.baseObject.Move(pos)
 
 	repaint(g)
--- a/canvas/line.go
+++ b/canvas/line.go
@@ -57,6 +57,11 @@ func (l *Line) Position() fyne.Position {
 // Move the line object to a new position, relative to its parent / canvas
 func (l *Line) Move(pos fyne.Position) {
 	oldPos := l.Position()
+
+	if oldPos.X == pos.X && oldPos.Y == pos.Y {
+		return
+	}
+
 	deltaX := pos.X - oldPos.X
 	deltaY := pos.Y - oldPos.Y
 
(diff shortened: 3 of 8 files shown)


canvas/circle.go
```diff
@@ -41,6 +41,9 @@ func (c *Circle) MinSize() fyne.Size {
 
 // Move the circle object to a new position, relative to its parent / canvas
 func (c *Circle) Move(pos fyne.Position) {
+	if c.Position1.X == pos.X && c.Position1.Y == pos.Y {
```
Jacalz: I think you can do `c.Position == pos` instead. The same applies to all places below.

Jacalz (changes requested):
Nice catch. Thanks. One note on making the if-statements easier to read.

Jacalz:
If you want to, you can likely also apply the same optimization to the base widgets (`internal/widget/base.go` and `widget/widget.go`).

canvas/circle.go
redawl: Done, thanks!

internal/widget/base.go
```diff
@@ -53,6 +53,9 @@ func (w *Base) Position() fyne.Position {
 // Move the widget to a new position, relative to its parent.
 // Note this should not be used if the widget is being managed by a Layout within a Container.
 func (w *Base) Move(pos fyne.Position) {
+	if w.Position() == pos {
+		return
+	}
 	w.position = pos
 
 	Repaint(w.super())
```
Jacalz: This is a bit nitpicky but would you mind updating the two base widgets to have the lines match in terms of empty lines? I.e. move the line setting the position down one line and keep the empty line between the if statement instead.

canvas/gradient.go
```diff
@@ -68,6 +68,9 @@ func (g *LinearGradient) Hide() {
 
 // Move the gradient to a new position, relative to its parent / canvas
 func (g *LinearGradient) Move(pos fyne.Position) {
+	if g.Position() == pos {
+		return
+	}
 	g.baseObject.Move(pos)
 
 	repaint(g)
```
Jacalz: See the comment below, I think it would be nice to be consistent and apply the same approach here and everywhere else.

canvas/circle.go
```diff
@@ -41,6 +41,9 @@ func (c *Circle) MinSize() fyne.Size {
 
 // Move the circle object to a new position, relative to its parent / canvas
 func (c *Circle) Move(pos fyne.Position) {
+	if c.Position1 == pos {
+		return
+	}
```
Jacalz, line 46: And an empty line after this line.

Jacalz (changes requested):
Thanks. So close. I just noticed that the line structure of the two base widgets no longer match so I added some suggestions to improve consistency when we're at it :)

canvas/line.go
```diff
@@ -57,6 +57,11 @@ func (l *Line) Position() fyne.Position {
 // Move the line object to a new position, relative to its parent / canvas
 func (l *Line) Move(pos fyne.Position) {
 	oldPos := l.Position()
+
```
Jacalz: And remove this line.

PR #5822, opened 2025-06-30
create wrapper in entry renderer for action item

### LINKED ISSUE #5823: ActionItem cannot be added to bound entry
### Checklist

### Describe the bug

Setting the ActionItem on a bound entry does not cause it to display.  Working on #5822 to fix it.

### How to reproduce

Image

### Screenshots

_No response_

### Example code

```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/theme"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("entries")

	empty := binding.NewString()
	boundNoAction := widget.NewEntryWithData(empty)

	bound := widget.NewEntryWithData(empty)
	bound.ActionItem = widget.NewButtonWithIcon("", theme.ContentCopyIcon(), func() {})

	b := binding.NewString()
	b.Set("string")
	boundNoActionWithString := widget.NewEntryWithData(b)

	boundWithString := widget.NewEntryWithData(b)
	boundWithString.ActionItem = widget.NewButtonWithIcon("", theme.ContentCopyIcon(), func() {})

	unboundNoAction := widget.NewEntry()

	unbound := widget.NewEntry()
	unbound.ActionItem = widget.NewButtonWithIcon("", theme.ContentCopyIcon(), func() {})

		widget.NewLabel("Bound entry without an action item, string set:"),
		boundNoActionWithString,
		widget.NewLabel("Bound entry with an action item, string set:"),
		boundWithString,
		widget.NewLabel("Bound entry without an action item, empty:"),
		boundNoAction,
		widget.NewLabel("Bound entry with an action item, empty:"),
		bound,
		widget.NewLabel("Normal entry without an action item:"),
		unboundNoAction,
		widget.NewLabel("Normal entry with an action item:"),
		unbound,
	))

	w.ShowAndRun()
}
```

### Fyne version

develop

### Go compiler version

go1.24.2 linux/amd64

Arch Linux

### Additional Information

Fixing this might solve #4993

hkparker:
### Description:

When creating a renderer, the entry includes its action item in the list of objects.  However, if the action item is added after the renderer if created, it won't be included in the list of objects (until the next renderer is created).  It seems like something about the bound entry is causing the renderer to be created right away, and this makes it impossible to add an action item.

This PR creates a stack container that is always part of the renderer's object list.  When an action item is present during Refresh, it is placed in this stack container.

Fixes #5823

<details>
  <summary>Demo code</summary>

```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/theme"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("entries")

	empty := binding.NewString()
	boundNoAction := widget.NewEntryWithData(empty)

	bound := widget.NewEntryWithData(empty)
	bound.ActionItem = widget.NewButtonWithIcon("", theme.ContentCopyIcon(), func() {})
	bound.Refresh()

	b := binding.NewString()
	b.Set("string")
	boundNoActionWithString := widget.NewEntryWithData(b)

	boundWithString := widget.NewEntryWithData(b)
	boundWithString.ActionItem = widget.NewButtonWithIcon("", theme.ContentCopyIcon(), func() {})
	boundWithString.Refresh()

	unboundNoAction := widget.NewEntry()

	unbound := widget.NewEntry()
	unbound.ActionItem = widget.NewButtonWithIcon("", theme.ContentCopyIcon(), func() {})
	unbound.Refresh()

		container.NewHBox(
			widget.NewLabel("Bound entry without an action item, string set:"),
			widget.NewButton("toggle", func() {
				if boundNoActionWithString.ActionItem == nil {
					boundNoActionWithString.ActionItem = widget.NewButtonWithIcon("", theme.ContentCopyIcon(), func() {})
					boundNoActionWithString.Refresh()
				} else {
					boundNoActionWithString.ActionItem = nil
					boundNoActionWithString.Refresh()
				}
			}),
		),
		boundNoActionWithString,
		container.NewHBox(
			widget.NewLabel("Bound entry with an action item, string set:"),
			widget.NewButton("toggle", func() {
				if boundWithString.ActionItem == nil {
					boundWithString.ActionItem = widget.NewButtonWithIcon("", theme.ContentCopyIcon(), func() {})
					boundWithString.Refresh()
				} else {
					boundWithString.ActionItem = nil
					boundWithString.Refresh()
				}
			}),
		),
		boundWithString,
		container.NewHBox(
			widget.NewLabel("Bound entry without an action item, empty:"),
			widget.NewButton("toggle", func() {
				if boundNoAction.ActionItem == nil {
					boundNoAction.ActionItem = widget.NewButtonWithIcon("", theme.ContentCopyIcon(), func() {})
					boundNoAction.Refresh()
				} else {
					boundNoAction.ActionItem = nil
					boundNoAction.Refresh()
				}
			}),
		),
		boundNoAction,
		container.NewHBox(
			widget.NewLabel("Bound entry with an action item, empty:"),
			widget.NewButton("toggle", func() {
				if bound.ActionItem == nil {
					bound.ActionItem = widget.NewButtonWithIcon("", theme.ContentCopyIcon(), func() {})
					bound.Refresh()
				} else {
					bound.ActionItem = nil
					bound.Refresh()
				}
			}),
		),
		bound,
		container.NewHBox(
			widget.NewLabel("Normal entry without an action item:"),
			widget.NewButton("toggle", func() {
				if unboundNoAction.ActionItem == nil {
					unboundNoAction.ActionItem = widget.NewButtonWithIcon("", theme.ContentCopyIcon(), func() {})
					unboundNoAction.Refresh()
				} else {
					unboundNoAction.ActionItem = nil
					unboundNoAction.Refresh()
				}
			}),
		),
		unboundNoAction,
		container.NewHBox(
			widget.NewLabel("Normal entry with an action item:"),
			widget.NewButton("toggle", func() {
				if unbound.ActionItem == nil {
					unbound.ActionItem = widget.NewButtonWithIcon("", theme.ContentCopyIcon(), func() {})
					unbound.Refresh()
				} else {
					unbound.ActionItem = nil
					unbound.Refresh()
				}
			}),
		),
		unbound,
	))

	w.ShowAndRun()
}

```
action_items

</details>

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -209,13 +209,6 @@ func (e *Entry) CreateRenderer() fyne.WidgetRenderer {
 
 	e.actionWrapper = &fyne.Container{Layout: layout.NewStackLayout()}
 	objects = append(objects, e.actionWrapper)
-	//if e.ActionItem != nil {
-	//	e.actionWrapper.(*fyne.Container).Objects = []fyne.CanvasObject{e.ActionItem}
-	//	e.actionWrapper.Show()
-	//} else {
-	//	e.actionWrapper.(*fyne.Container).Objects = []fyne.CanvasObject{}
-	//	e.actionWrapper.Hide()
-	//}
 
 	icon := canvas.NewImageFromResource(e.Icon)
 	icon.FillMode = canvas.ImageFillContain
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -104,7 +104,6 @@ type Entry struct {
 
 	// ActionItem is a small item which is displayed at the outer right of the entry (like a password revealer)
 	ActionItem      fyne.CanvasObject `json:"-"`
-	actionWrapper   fyne.CanvasObject
 	binder          basicBinder
 	conversionError error
 	minCache        fyne.Size
@@ -207,9 +206,6 @@ func (e *Entry) CreateRenderer() fyne.WidgetRenderer {
 		e.ActionItem = newPasswordRevealer(e)
 	}
 
-	e.actionWrapper = &fyne.Container{Layout: layout.NewStackLayout()}
-	objects = append(objects, e.actionWrapper)
-
 	icon := canvas.NewImageFromResource(e.Icon)
 	icon.FillMode = canvas.ImageFillContain
 	objects = append(objects, icon)
@@ -218,7 +214,7 @@ func (e *Entry) CreateRenderer() fyne.WidgetRenderer {
 	}
 
 	e.syncSegments()
-	return &entryRenderer{box, border, e.scroll, icon, objects, e}
+	return newEntryRenderer(box, border, e.scroll, icon, objects, e)
 }
 
 // CursorPosition returns the relative position of this Entry widget's cursor.
@@ -1481,12 +1477,28 @@ func (e *Entry) setFieldsAndRefresh(f func()) {
 var _ fyne.WidgetRenderer = (*entryRenderer)(nil)
 
 type entryRenderer struct {
-	box, border *canvas.Rectangle
-	scroll      *widget.Scroll
-	icon        *canvas.Image
+	box, border   *canvas.Rectangle
+	scroll        *widget.Scroll
+	icon          *canvas.Image
+	actionWrapper *fyne.Container
 
 	objects []fyne.CanvasObject
 	entry   *Entry
+}
+
+func newEntryRenderer(box, border *canvas.Rectangle, scroll *widget.Scroll, icon *canvas.Image, objects []fyne.CanvasObject, e *Entry) *entryRenderer {
+	actionWrapper := &fyne.Container{Layout: layout.NewStackLayout()}
+	objects = append(objects, actionWrapper)
+
+	return &entryRenderer{
+		box:           box,
+		border:        border,
+		scroll:        scroll,
+		icon:          icon,
+		actionWrapper: actionWrapper,
+		objects:       objects,
+		entry:         e,
+	}
 }
 
 func (r *entryRenderer) Destroy() {
@@ -1525,11 +1537,11 @@ func (r *entryRenderer) leadingInset() float32 {
 
 func (r *entryRenderer) Layout(size fyne.Size) {
 	if r.entry.ActionItem != nil {
-		r.entry.actionWrapper.(*fyne.Container).Objects = []fyne.CanvasObject{r.entry.ActionItem}
-		r.entry.actionWrapper.Show()
+		r.actionWrapper.Objects = []fyne.CanvasObject{r.entry.ActionItem}
+		r.actionWrapper.Show()
 	} else {
-		r.entry.actionWrapper.(*fyne.Container).Objects = []fyne.CanvasObject{}
-		r.entry.actionWrapper.Hide()
+		r.actionWrapper.Objects = []fyne.CanvasObject{}
+		r.actionWrapper.Hide()
 	}
 
 	th := r.entry.Theme()
@@ -1547,9 +1559,11 @@ func (r *entryRenderer) Layout(size fyne.Size) {
 
 	pad := theme.InputBorderSize()
 	actionIconSize := fyne.NewSize(0, size.Height-pad*2)
-	actionIconSize.Width = r.entry.actionWrapper.MinSize().Width
-	r.entry.actionWrapper.Resize(actionIconSize)
-	r.entry.actionWrapper.Move(fyne.NewPos(size.Width-actionIconSize.Width-pad, pad))
+	if r.entry.ActionItem != nil {
+		actionIconSize.Width = r.actionWrapper.MinSize().Width
+		r.actionWrapper.Resize(actionIconSize)
+		r.actionWrapper.Move(fyne.NewPos(size.Width-actionIconSize.Width-pad, pad))
+	}
 
 	validatorIconSize := fyne.NewSize(0, 0)
 	if r.entry.Validator != nil || r.entry.AlwaysShowValidationError {
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -89,6 +89,9 @@ type Entry struct {
 	placeholder RichText
 	content     *entryContent
 	scroll      *widget.Scroll
+	box         *canvas.Rectangle
+	border      *canvas.Rectangle
+	objects     []fyne.CanvasObject
 
 	// useful for Form validation (as the error text should only be shown when
 	// the entry is unfocused)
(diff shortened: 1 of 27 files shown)


Jacalz:
This solution means that getting the objects will allocate each time if the entry has an action item. I also think it may be added multiple times to the objects slice? I think it would be better to make sure that the action item is only added once but I suppose we may run into issues with old action items in the objects slice if the developer changes to another action item?

hkparker:
Yeah exactly.  If it should be possible for the user to change these around just by setting the field, then it needs to be added to the objects like this every time, as far as I can tell.

An alternative that might work best is to not export `ActionItem` as a field, but instead export a `SetActionItem(func())` method on the entry, that way we have a place we can adjust the object list and we don't need to do it every time `Objects` is called, but that would be an API change.

Jacalz:
I think you might be able to create some sort of wrapper object which shows the action item as part of its own contents if it exists but stays hidden otherwise. This way we avoid allocating a bunch of times. Not sure if that approach is feasible though, just an idea.

hkparker:
Oh yeah, that's a good idea.  I guess it depends on how important it is to avoid allocating here, I don't have a good sense of the performance impact from this.

Jacalz:
I think the .Objects() call may technically be called on every frame, I.e. 60 times per second, (put in a counter and check) so it is definitely in the hot path. As noted above, I think the current code also appends the action item once more each time so memory growth is unbounded.

hkparker:
wow ok yeah, definitely need an alternative approach then!  What about adding it to the object list (when needed) during calls to Refresh?

Jacalz:
That might work. I suppose it also might be possible to look at the last element or length of the objects slice and be a bit smarter about adding it twice.

hkparker:
I don't see how it's being added twice yet in this code, since each time Objects is called I just add the ActionItem to the existing objects if needed, but never change the existing objects.  But either way I'll see if I can update it with a technique that requires Refresh to update the objects and avoids the allocation here.

Jacalz:
Ah, sorry. You're right about that. I read the code wrong previously

hkparker:
Ok I have something up now that 1) doesn't require an API change, 2) doesn't allocate on the calls to `Objects`, and 3) allows the user to set and change the `ActionItem` as they see fit and apply the change with `Refresh`.  But it's weird and I'm guessing there's a reason we don't do things this way.

I moved the objects over to a field on the entry itself, not the renderer (since widgets can't reference their renderers, but renderers can reference their widgets).  Then I made a function to set all the objects on that field (starting from scratch each time so that adding a different action item after one has already been set doesn't result in both ending up in the object list).  This new function then gets called when creating the entry or refreshing.

Seems to work.  If this is a viable technique I can do some further testing and experimentation as requested.  I'm not sure how else to accomplish the three constraints I approached this with, but also happy to try alternative suggestions as well.

dweymouth:
> Ok I have something up now that 1) doesn't require an API change, 2) doesn't allocate on the calls to `Objects`, and 3) allows the user to set and change the `ActionItem` as they see fit and apply the change with `Refresh`. But it's weird and I'm guessing there's a reason we don't do things this way.
> 
> I moved the objects over to a field on the entry itself, not the renderer (since widgets can't reference their renderers, but renderers can reference their widgets). Then I made a function to set all the objects on that field (starting from scratch each time so that adding a different action item after one has already been set doesn't result in both ending up in the object list). This new function then gets called when creating the entry or refreshing.
> 
> Seems to work. If this is a viable technique I can do some further testing and experimentation as requested. I'm not sure how else to accomplish the three constraints I approached this with, but also happy to try alternative suggestions as well.

I think @Jacalz's suggestion of using a wrapper object for the action item that can always be in the Objects slice on the renderer may be best. The separation between the widget and renderer was designed for the widget struct to just be the "state", and the renderer struct everything related to drawing it. This allows the renderers to be disposed of by Fyne if the widget hasn't been rendered in awhile, and to be re-created as needed.

hkparker:
I'll give this approach a try soon and share what I come up with

hkparker:
I've been thinking about this again more recently and it's not obvious to me what this wrapper object would look like.  The idea as I understand it is you have some wrapper widget that is never nil and always in the entry's object list, and when this entry has an action item set, it's done by setting it in the wrapper widget, which becomes visible when it has an action item.  But isn't that pushing the same problem just down into this wrapper widget?  Because the wrapper widget is going to have it's own renderer and object list, and it's going to need to handle the case where the action item it's wrapping is nil and dynamically added to it's object list, etc

Jacalz:
I don't think that's a problem. The wrapper can cheaply add or remove the containing widget from its object slice depending on if the object is nil or not. Allocating a slice of length one is a lot cheaper than allocating the whole entry renderer slice on each call to .Objects().

Jacalz:
But yeah, it does add a lot of complexity for something that out to be quite simple. However, I don't think this is easily solved with how the Fyne API works unfortunately.

andydotxyz:
I would have thought the Stack container would fit the bill?

hkparker:
Ok yeah maybe, if whenever the action item is updated, the stack's Objects list is updated, and the stack is hidden / shown as needed.  Let me play around with this today and see if I can get it working.

hkparker:
I'm getting an import cycle when I try to import container from within widget, is that expected?

```
$ go run main.go
package command-line-arguments
	imports fyne.io/fyne/v2/app from main.go
	imports fyne.io/fyne/v2/internal/driver/glfw from app_gl.go
	imports fyne.io/fyne/v2/container from canvas.go
	imports fyne.io/fyne/v2/widget from apptabs.go
	imports fyne.io/fyne/v2/container from entry.go: import cycle not allowed
```

Jacalz:
Yeah, you can't use that package in widgets as some containers are widgets. That would be nice to fix but I don't think we can do so in any easy way. You can work around it by creating a fyne.Container struct directly and setting the layout parameter instead.

widget/entry.go
```diff
@@ -99,6 +100,7 @@ type Entry struct {
 
 	// ActionItem is a small item which is displayed at the outer right of the entry (like a password revealer)
 	ActionItem      fyne.CanvasObject `json:"-"`
+	actionWrapper   fyne.CanvasObject
```
dweymouth: This is probably better suited for the renderer struct

hkparker: Moved it over to the renderer

hkparker:
@Jacalz @dweymouth let me know if this approach seems correct whenever you've got the time please, if it looks good I'll get the tests corrected for this and we can get it in

widget/entry.go
```diff
@@ -1523,14 +1520,30 @@ func (e *Entry) setFieldsAndRefresh(f func()) {
...
+	box, border   *canvas.Rectangle
+	scroll        *widget.Scroll
+	icon          *canvas.Image
+	actionWrapper *fyne.Container
 
 	objects []fyne.CanvasObject
 	entry   *Entry
 }
 
+func newEntryRenderer(box, border *canvas.Rectangle, scroll *widget.Scroll, icon *canvas.Image, objects []fyne.CanvasObject, e *Entry) *entryRenderer {
```
Jacalz: This looks a bit strange that almost all of the setup code for the entry renderer is outside of this function except for one part of it. I would recommend bringing down the complexity of the function in some other way.

Jacalz (commented):
It looks very reasonable to me. Nice work :)

widget/entry.go
hkparker: huh I'm actually not sure why I did it that way originally, yeah, let me fix that

andydotxyz:
Just to ping on this there are 2 days left on the bug fix window - do you need a hand on getting this wrapped up @hkparker ?

hkparker:
Hey @andydotxyz sorry I missed the window.  I could use some help with the tests I think, yeah

andydotxyz:
The test failures are because the layout is not being run during the current refresh loop when the content is inserted. This diff (and then the associated test file fixes) worked for me:

```diff
diff --git a/widget/entry.go b/widget/entry.go
index e1c984eb5..b7e6b9980 100644
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -1655,6 +1655,7 @@ func (r *entryRenderer) Refresh() {
        if r.entry.ActionItem != nil {
                r.actionWrapper.Objects = []fyne.CanvasObject{r.entry.ActionItem}
                r.actionWrapper.Show()
+               r.Layout(r.entry.Size())
        } else {
                r.actionWrapper.Objects = []fyne.CanvasObject{}
                r.actionWrapper.Hide()
```

hkparker:
@andydotxyz I tried that patch (as well as calling Layout after that check, so we layout in all cases), and rebased on develop for good measure, but I'm still getting a bunch of unexpected failures in the pipeline.  Were you able to get tests to pass locally with that change?

andydotxyz:
I needed to also fix some of the golden files, but yes otherwise it worked as far as I can remember.

hkparker:
@andydotxyz I figured out that when I fix some of the golden files for one test, often another test uses that same golden file as well and is now broken.  So, I can make different copies of the files and have the other tests use the versions that are modified differently, but I have to admit I'm not sure why this is going on and don't think I'll be able to figure out each case.

andydotxyz:
Perhaps make a note of the files coped so whoever reviews can look into it. I've had the issue before as well.

hkparker:
It also looks like I'm seeing a few different failures between the platforms, which is surprising to me considering the nature of the change

andydotxyz:
The select_entry_internal_test golden file change seems to be fixing a corrupt golden file (removing a container output that has the same indentation of another)...

Perhaps rebase on develop and run tests locally, see where you get to?

PR #5824, opened 2025-06-30
wasm: Fix memory leaks in uses of js.FuncOf

Jacalz:
### Description:

The documentation for `js.FuncOf` notes that you need to call `.Release()` on the function to release the value from the internal map. I noticed that there was two places in our code where this memory leak may happen. I removed the two leaks and refactored the notification code to be more readable (plus removed the fyne.LogError on success). The only remaining `js.FuncOf` is the theme watcher but that lives the entirety of the program so it should need to be cleaned up.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


sdassow (approved):
Good catch, thanks!

PR #5825, opened 2025-06-30
Improve performance of wasm syscalls in driver

Jacalz:
### Description:

Use a common device setup init to get values and functions we invoke a lot. Use these stored functions when we can. Hiding/showing virtual keyboard and changing cursors should be faster. Startup may be a bit faster as well.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


sdassow (approved):
Looks good.

PR #5826, opened 2025-07-01
added german date format

generikvault:
### Description:
Added date format for german.

### Checklist:

- [X] Tests included.
- [ X] Tests all pass.


generikvault:
Thank you for the approval @Jacalz 
How can I deal with the failed driver/glfw test? That should be totally independet from my change.

Jacalz:
You're welcome. That flaky test is known but we haven't found the root cause yet unfortunately. I'll rerun and merge when it passes :)

PR #5833, opened 2025-07-02
Add functionality to detect existing syso files during packaging

alechamaker:
### Description:

Adds the ability to detect existing syso files during packaging.

I ran into this issue as I was trying to cross-compile with fyne-cross linux -> windows. I found that the `.syso` file I had specified in the build directory, as I would with a normal `go build`, was conflicting with the `fyne.syso` created during the packaging stage of the build. I have tested this fix, and it worked for my use case.

(still working out the kinks of the unit test)

Fixes #(issue)
conflicts with user-specified `.syso` files.

### Checklist:

- [x] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/cmd/fyne/internal/commands/package-windows.go
+++ b/cmd/fyne/internal/commands/package-windows.go
@@ -51,57 +51,67 @@ func (p *Packager) packageWindows(tags []string) error {
 		return fmt.Errorf("failed to close image file: %w", err)
 	}
 
-	// write manifest
-	manifest := p.exe + ".manifest"
-	manifestGenerated := false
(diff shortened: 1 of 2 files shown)


Jacalz:
The cmd/fyne tool here is deprecated and has moved into https://github.com/fyne-io/tools instead. Please open a PR there instead.

On another note, I think it might make more sense to try add support for more metadata fields in FyneApp.toml instead of using existing .sysp files. That approach is what I added for Linux metadata.

Jacalz:
Closing per the above comment. Feel free to continue working on this over there. Thanks

Issue #5827, opened 2025-07-02
Widgets should not use cache.Renderer


Jacalz:
### Checklist

- [x] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [x] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

Per discussion with @dweymouth on Discord, we should make sure to update all of the widgets to not rely on `cache.Renderer()` from the internal packages. This breaks the model we have where a widget should not reference it's renderer and is bad practice as it makes it hard to replicate the same function outside of this repository (given that it depends on internal details from the renderer cache). The renderer lookup is also quite expensive (at least a map lookup and some other logic) so in the cases where we are doing a refresh anyway, it is much cheaper to just set a field and refresh.

### Is it possible to construct a solution with the existing API?

Yes, I rewrote Activity (see https://github.com/fyne-io/fyne/pull/4861) and InfiniteProgressbar (see https://github.com/fyne-io/fyne/pull/4848/files) a while back to avoid this. The general approach is to set fields inside the widget instead and let the renderer look at them when it is being refreshed.

### Describe the solution you'd like to see.

Widgets (or containers that are widgets) which do use `cache.Renderer` should be reworked not to. A quick grep on `develop` gives:
- [ ] internal/widget.Scroller
- [ ] widget.Entry
- [ ] widget.Form
- [ ] widget.RichText
- [x] widget.TextGrid
- [ ] widget.Tree

andydotxyz:
As noted on the issue let's not forget that some containers can have this problem as well - see #5780

PR #5834, opened 2025-07-02
Remove uses of renderer cache in textgrid

redawl:
### Description:
Remove usage of renderer cache in TextGrid, per https://github.com/fyne-io/fyne/issues/5827

I also fixed a bug where `textGridRow` was only calling `ExtendBaseWidget` once its renderer was created. This was causing a couple of tests to fail, because with the `cache.Renderer` calls removed, we were no longer creating the renderer during unit test setup.

Because we are now extending BaseWidget on creation of rows, a lot of snapshots needed updating. I took a look at the changes, and I think they were actually wrong before. LMK if I'm wrong about that :)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/textgrid.go
+++ b/widget/textgrid.go
@@ -9,7 +9,6 @@ import (
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/canvas"
 	"fyne.io/fyne/v2/internal/async"
-	"fyne.io/fyne/v2/internal/cache"
 	"fyne.io/fyne/v2/internal/painter"
 	"fyne.io/fyne/v2/internal/widget"
 	"fyne.io/fyne/v2/theme"
@@ -490,6 +489,8 @@ type textGridContent struct {
 
 	rows     int
 	cellSize fyne.Size
+
+	visible []fyne.CanvasObject
 }
 
 func newTextGridContent(t *TextGrid) *textGridContent {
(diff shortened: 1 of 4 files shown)


widget/textgrid.go
```diff
@@ -663,14 +664,10 @@ func (t *textGridRow) setRow(row int) {
...
 	obj *textGridRow
-
-	cols int
-
-	objects []fyne.CanvasObject
 }
 
-func (t *textGridRowRenderer) appendTextCell(str rune) {
-	th := t.obj.text.text.Theme()
+func (t *textGridRow) appendTextCell(str rune) {
```
Jacalz, line 665: The method ordering here is a little strange now that the widgets methods are below the renderer struct.

Jacalz (commented):
Thanks for this. This is great stuff and an important change. I'm not entirely sure how to solve this some other way but I think this still might be breaking the promise of widgets not referencing their renders because some of the widgets that make up the rows etc. are now referencing the objects that the renderer is rendering. See discussion on https://github.com/fyne-io/fyne/pull/5822.

redawl:
> Thanks for this. This is great stuff and an important change. I'm not entirely sure how to solve this some other way but I think this still might be breaking the promise of widgets not referencing their renders because some of the widgets that make up the rows etc. are now referencing the objects that the renderer is rendering. See discussion on #5822.

I could be wrong here, but I think this is a different case. `textGridRow` and `textGridContent` are wrapper components that are only known about by the renderers. `TextGrid`, `TextGridRow` and `TextGridCell` are the structs that contain the "state" that is being referred to here.

widget/textgrid.go
redawl: Moved `textGridContentRenderer` and `textGridRowRenderer` typedefs so that they are above their methods, thanks!
https://github.com/fyne-io/fyne/pull/5834/commits/d27888292afe629c143d98afa92b0852bdd37dba

Jacalz:
Hmm. Good point. If you think about those sub widgets being opened by the main renderer it kind of makes sense. However, they are still widgets so I don't know how to reason about it. I think it is best to check with Andy.

andydotxyz:
I think this change makes sense. Many widgets will be children of the parent widgets, the pattern here doesn't seem to go against the pattern of referencing renderers.

Jacalz:
Great :)

PR #5830, opened 2025-07-02
Rectangle shadow and partially rounded corners, edge smoothness improvements

Vinci10:
### Description:

Fixes:
- [#5352](https://github.com/fyne-io/fyne/issues/5352)
- [#4129](https://github.com/fyne-io/fyne/issues/4129) (shadows only)
- [#4411](https://github.com/fyne-io/fyne/issues/4411) (shadows on the right and bottom can be added using proper offset)

This pull request introduces enhancements to the canvas objects, focusing on improved visual flexibility and shadow rendering:

- **Per-Corner Radius Support**
  - Rectangle and Square now support specifying a different corner radius for each corner (`TopRightCornerRadius`, `TopLeftCornerRadius`, `BottomRightCornerRadius`, `BottomLeftCornerRadius`).
  - The `CornerRadius` field remains as the default; per-corner fields override it if set with higher value.
  - Added a new constant `FullyRoundedCornerRadius` for pill/circular shapes.

- **Shadow Support for Rectangle, Square, and Circle**
  - Introduced a `baseShadow` struct and `Shadowable` interface to provide reusable shadow logic.
  - Added fields for shadow color, softness, offset, and type (drop/box shadow).
  - The `ExpandForShadow` field allows the object to keep its requested/original size, expanding the total bounds to include the shadow.
  - Updated shader and painter logic to render shadows with configurable softness and offset, and to blend shadows correctly.
  - If a canvas object has a shadow added, the canvas size is reduced by the shadow paddings on each side. This ensures that the entire object—including its shadow—fits within the originally requested frame size. The content area (without shadow) will be smaller to accommodate the shadow within the allocated space.

- **Shader and Painter Updates**
  - Rectangle and round-rectangle shaders updated to support shadow rendering and per-corner radii.
  - Painter logic updated to pass new parameters and handle shadow bounds.

- **Rounded Edge Softness Fix**
  - Fixed an issue where rounded edge was not rendered correctly.

- **API Additions**
  - New methods and fields are documented in code comments.
  - Backwards compatibility is maintained for existing code.

## Status

- **This is a draft PR.**
- I am looking for feedback and suggestions from the Fyne team regarding:
  - API design and naming
  - Performance considerations
- **Tests are not yet included.**
- **Performance has not been fully evaluated.**

#### Additional Notes

[...]

PR #5838, opened 2025-07-04
rectangle edge smoothness fix

Vinci10:
### Description:

(Based on #5830)
Fixed an issue where rounded edge was not rendered correctly.

##### Before
edge_smoothness_before
##### After
edge_smoothness_after

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.


Vinci10:
Good point. Fortunately, if there is added border then sharp edge is not visible, what about changing `edge_softness` to 0 when border is added ? `edge_softness` has already been added as a shader parameter

IMO significantly wider outline looks better but I understand that someone may want to draw thin border

andydotxyz:
> Good point. Fortunately, if there is added border then sharp edge is not visible, what about changing `edge_softness` to 0 when border is added ? `edge_softness` has already been added as a shader parameter

I'm not certain of the details but if the edge is only used when it is fill to background then that sounds like a fine approach.

Vinci10:
Maybe we should add edge_softness as a new rectangle parameter to allow user to change it? High `edge_softness` value may create something looking like drop shadow

andydotxyz:
> Maybe we should add edge_softness as a new rectangle parameter to allow user to change it? High `edge_softness` value may create something looking like drop shadow

No. 

The Fyne API is a semantic API - the parameters do what they mean. Something called "EdgeSoftness" that can create a shadow-like effect makes no sense IMHO.

andydotxyz:
This is looking excellent now thanks.
Unfortunately a tidy in how shaders are loaded ("embed" package) is blocking this now.
Could you re-base on latest develop and we can get this merged?

Thanks!

Vinci10:
I updated the PR with a different implementation of the stroke and fill colors via `smoothstep` to make the thin border look smooth for rounded corners .  Please check if it works as expected. Thanks.

andydotxyz:
> I updated the PR with a different implementation of the stroke and fill colors via `smoothstep` to make the thin border look smooth for rounded corners . Please check if it works as expected. Thanks.

Thanks so much for going back and looking at it again :)

internal/painter/gl/shaders/round_rectangle.frag
```diff
@@ -22,26 +22,24 @@ void main() {
...
-    if (distance < 0.0)
+    else
     {
-        to_color = fill_color;
-    } 
-
-    distance = abs(distance) - stroke_width_half;
-
-    float blend_amount = smoothstep(-1.0, 1.0, distance);
+        // --- Fill-Only Rectangle (No Stroke) ---
```
andydotxyz: Is the comment needed? if so then it doesn't need all the decoration

andydotxyz:
Thanks - this version looks cool.
The comment style is pretty weird though - if it's needed try to limit it to the content.

Vinci10:
I removed that comment. I think it is not needed. I tested the code only on `WSL Ubuntu 20.04`. I am not able to perform tests on other platforms and I am not sure about performance impact. Could you please verify correctness of this new code ? Thanks.

internal/painter/gl/shaders/round_rectangle.frag
Vinci10: comment is not needed. I removed it.

Vinci10:
I had to update PR because I found out that edges were cropped so I expanded frame without changing rectangle size

andydotxyz:
Congratulations on landing your first Fyne PR and thanks so much!

PR #5839, opened 2025-07-04
support fully rounded corner radius

Vinci10:
### Description:

(Based on #5830)
Added a new constant `FullyRoundedCornerRadius` for pill/circular shapes as per Material Design.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/canvas.go
+++ b/canvas.go
@@ -1,6 +1,16 @@
 package fyne
 
-import "image"
+import (
+	"image"
+	"math"
+)
+
+const (
+	// FullyRoundedCornerRadius can be applied to a canvas corner radius to achieve fully rounded corners.
+	// This constant represents the maximum possible corner radius, resulting in a pill or circular appearance.
+	// Since: 2.7
+	FullyRoundedCornerRadius float32 = math.MaxFloat32
+)
 
 // Canvas defines a graphical canvas to which a [CanvasObject] or Container can be added.
 // Each canvas has a scale which is automatically applied during the render process.
--- a/internal/painter/gl/draw.go
+++ b/internal/painter/gl/draw.go
@@ -27,11 +27,6 @@ func (p *painter) defineVertexArray(prog Program, name string, size, stride, off
 }
 
 func (p *painter) drawCircle(circle *canvas.Circle, pos fyne.Position, frame fyne.Size) {
-	size := circle.Size()
-	radius := size.Width / 2
-	if size.Height < size.Width {
-		radius = size.Height / 2
-	}
 	program := p.roundRectangleProgram
 
 	// Vertex: BEG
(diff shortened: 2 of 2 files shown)


andydotxyz:
I don't understand this API sorry. I think it's the same as setting the radius on a rectangle to be half the width or height.
Is that right or is it something different?

Vinci10:
Right, the API is essentially the same as setting the radius on a rectangle to be half the width or height. Originally, this approach was reserved for circles, but I wanted to extend similar functionality to rectangles as well.
Developers can simply implement this logic on their side, but if the size of the shape changes, the radius would also need to be adjusted accordingly. By handling this directly within the painter, where the size is known before drawing, it simplify the development process and reduce the need for repetitive calculations.

andydotxyz:
> but if the size of the shape changes, the radius would also need to be adjusted accordingly

Thanks, this makes sense as a use-case.

Essentially this could be called `RadiusMaximum` to avoid having to learn what "fully rounded" means then?

p.s. it would need to be added into the software painter as well - in this case it should be straight forward.

Vinci10:
> > but if the size of the shape changes, the radius would also need to be adjusted accordingly
> 
> Thanks, this makes sense as a use-case.
> 
> Essentially this could be called `RadiusMaximum` to avoid having to learn what "fully rounded" means then?
> 
> p.s. it would need to be added into the software painter as well - in this case it should be straight forward.

Great! I will update the PR, should I also add some unit tests ?

andydotxyz:
> should I also add some unit tests ?

Yes please - always (unless not possible)

Vinci10:
I added `RadiusMaximum` to software painter and included tests.

canvas.go
```diff
@@ -1,6 +1,16 @@
...
+import (
+	"image"
+	"math"
+)
+
+const (
+	// RadiusMaximum can be applied to a canvas corner radius to achieve fully rounded corners.
+	// This constant represents the maximum possible corner radius, resulting in a pill or circular appearance.
+	// Since: 2.7
+	RadiusMaximum float32 = math.MaxFloat32
```
andydotxyz: On reading more I think this belongs in the `canvas` package as it relates to there in terms of usage.

I can't tell if `RadiusMaximum` is the best name, it is unambiguous but is there an alternative that reads better - any thoughts @Jacalz @sdassow @renlite ?

dweymouth: My only other suggestion is `RadiusFullyRounded` but I honestly think RadiusMaximum is fine

andydotxyz: Thanks, do you also think the `canvas` package is a better fit than `fyne` toplevel?
(like the animation constants)

dweymouth: Yeah I think so, good call

andydotxyz: @Vinci10 if you can make that change then I think this could be landed for 2.7l :)

Vinci10: thanks for suggestion, I have moved new constant to `canvas` package

andydotxyz (approved):
Nice one, thanks for this!

Issue #5352, opened 2025-01-02
Support for partially rounded rectangles


dweymouth:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [X] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

Currently, rounding of rectangle corners is always applied to all 4 corners and there is no way to round only some corners.

The ability to selectively round corners would enable drawing things like progress bars or seek bars styled like this:
![image](https://github.com/user-attachments/assets/24a3d2e1-f77f-4d48-b3b2-d926a58e0096)

Or segmented controls with Material styling:
![image](https://github.com/user-attachments/assets/875ae097-9b0a-4c7e-a7c1-05897811fb3b)

### Is it possible to construct a solution with the existing API?

Not with the canvas API. I suppose you could rasterize an image with a 3rd party drawing library although this would be inefficient and lacking GPU acceleration.

### Describe the solution you'd like to see.

A way to apply the corner radius to only some corners. A possibly backward-compatible way of doing this could be

```go
package canvas

type RectangleCorner int
const (
    RectangleCornerTopLeft RectangleCorner = 0x01
    RectangleCornerTopRight RectangleCorner = 0x02
    RectangleCornerBottomLeft RectangleCorner = 0x04
    RectangleCornerBottomRight RectangleCorner = 0x08
)

// bitwise and of the above constants
type RectangleCorners int

type Rectangle struct {
    // existing fields

    // RoundedCorners specifies which corners the rounding will be applied to.
    // The default is to round all corners.
    RoundedCorners RectangleCorners
}
```
    
One downside of this backward-compatible approach is the inability to round corners with a different radius per corner but this would be quite uncommon in UI elements.

andydotxyz:
I challenge that it cannot be done with current API.
For the first image I would use two rounded rectangles and two regular ones the same colour.

For the second picture the rounded rectangle makes a border and then lines in the separator spots.

dweymouth:
Actually you are right :) I didn't think about overlapping two rectangles. It still may be a worthwhile future feature though (since it's not necessarily intuitive to solve it this way) but this request can certainly be treated as just a "nice to have" rather than something to prioritize.

andydotxyz:
Thanks so much for all the work on this @Vinci10, this is now landed!

PR #5840, opened 2025-07-04
feat: rectangle per-corner radius

Vinci10:
### Description:

(Based on #5830)
Fixes: [#5352](https://github.com/fyne-io/fyne/issues/5352)

- **Per-Corner Radius Support**
  - Rectangle and Square now support specifying a different corner radius for each corner (`TopRightCornerRadius`, `TopLeftCornerRadius`, `BottomRightCornerRadius`, `BottomLeftCornerRadius`).
  - The `CornerRadius` field remains as the default; per-corner fields override it if set with higher value.

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,1 +1,1 @@
-github: [fyne-io, andydotxyz, toaster, Jacalz, changkun, dweymouth, lucor]
+github: [fyne-io, andydotxyz, Jacalz, dweymouth, toaster, lucor]
--- a/canvas/text_test.go
+++ b/canvas/text_test.go
@@ -3,6 +3,9 @@ package canvas_test
 import (
 	"image"
 	"image/color"
+	"io"
+	"log"
+	"os"
 	"testing"
 
 	"fyne.io/fyne/v2"
@@ -14,6 +17,11 @@ import (
 )
 
 func TestText_FontSource(t *testing.T) {
+	// Discarding log output for tests
+	// The following method logs an error:
+	// test.NewWindow(text).Canvas().Capture()
+	log.SetOutput(io.Discard)
+	t.Cleanup(func() { log.SetOutput(os.Stderr) })
 	text := canvas.NewText("Test", color.NRGBA{0, 0, 0, 0xff})
 	c := test.NewWindow(text).Canvas()
 
--- a/cmd/fyne/internal/commands/build_test.go
+++ b/cmd/fyne/internal/commands/build_test.go
@@ -1,6 +1,8 @@
 package commands
 
 import (
+	"io"
+	"log"
 	"os"
 	"path/filepath"
 	"testing"
@@ -9,6 +11,11 @@ import (
 )
 
 func Test_BuildWasmVersion(t *testing.T) {
+	// Discarding log output for tests
+	// The following method logs an error:
+	// err := b.build()
+	log.SetOutput(io.Discard)
+	t.Cleanup(func() { log.SetOutput(os.Stderr) })
 	expected := []mockRunner{
 		{
 			expectedValue: expectedValue{args: []string{"mod", "edit", "-json"}},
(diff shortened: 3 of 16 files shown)


canvas/rectangle.go
```diff
@@ -26,6 +26,17 @@ type Rectangle struct {
 	//
 	// Since: 2.7
 	Aspect float32
+
+	// The `CornerRadius` field sets the default radius for all corners. If a specific corner
+	// radius (such as `TopRightCornerRadius`, `TopLeftCornerRadius`, `BottomRightCornerRadius`,
+	// or `BottomLeftCornerRadius`) is set to a value greater than `CornerRadius`, that value
+	// will override `CornerRadius` for the respective corner. Otherwise, `CornerRadius` is used.
+	//
+	// Since: 2.7
```
andydotxyz, line 42: Unfortunately (I think...) This document only applies to the first item in the list of fields.
The note about version needs to be on both and the documentation re-jigged to describe the field rather than the concept.

Wider documentation could be added to the rectangle page on docs.fyne.io.

Vinci10:
> This is a neat addition - however I question 1 item of logic:
> 
> > is set to a value greater than `CornerRadius`
> 
> What if 1 corner should be smaller than the `CornerRadius`? I think it would be better to always override a corner if that radius is set to non-zero.
> 
> Please note this change needs to be applied to the software renderer as well. And unit tests added.

Great idea! So if the user wants to set the radius to zero for just one corner, they shouldn't use the `CornerRadius` value and instead set the radius individually for the other corners. Sounds good.

Sure, I will add unit tests and also try to update software renderer.

andydotxyz:
> Great idea! So if the user wants to set the radius to zero for just one corner, they shouldn't use the CornerRadius value and instead set the radius individually for the other corners. Sounds good.

Yes exactly. Once this goes out in a release we can make some examples for the documentation site to clear this up if required.

canvas/rectangle.go
Vinci10: Ok, so I need to add `Since: 2.7` to every new field along with some short one-line comment ?

Jacalz:
Just a bold idea that might not make sense: we are adding a lot of parameters to rectangles, many of which might not be used in the majority of cases. Should we add a more specialised rectangle object that allows changing corner radius of all four corners and also the fully rounded corners form the other PR?

I'm not saying that I necessarily think it is the best option. Just throwing the idea out there.

andydotxyz:
> Just a bold idea that might not make sense: we are adding a lot of parameters to rectangles, many of which might not be used in the majority of cases. Should we add a more specialised rectangle object that allows changing corner radius of all four corners and also the fully rounded corners form the other PR?
> 
> I'm not saying that I necessarily think it is the best option. Just throwing the idea out there.

It's a good question to ask, but personally I don't see the need for a second type of rectangle that is more configurable.
All these fields are optional and default perfectly well to zero values. Seems good API to me.

canvas/rectangle.go
andydotxyz: Yes, apologies for not confirming sooner

Vinci10: I updated struct documentation and changed logic to always override a corner if the individual radius is set to non-zero

Vinci10:
I applied changes to the software renderer and added unit tests.

andydotxyz (approved):
Amazing thanks so much!

andydotxyz:
This looks excellent, unfortunately I cannot resolve the conflicts on GitHub - can you merge in develop and see what needs fixed @Vinci10 ?

Vinci10:
> This looks excellent, unfortunately I cannot resolve the conflicts on GitHub - can you merge in develop and see what needs fixed @Vinci10 ?

Done, please review, thanks.

andydotxyz (approved):
Fantastic thanks so much!

Issue #4411, opened 2023-11-22
Remove shadow above expanded Menu Items


Rldeckard:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [X] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

There is a shadow that should not exist above/on top of expanded Menus. This makes the Menu Items look disconnected from the dropdown like a pop up window. Removing the top shadow will make the menu appear attached to the calling button as is expected functionality.

### How to reproduce

1. Create Fyne Menu.
2. Add NewMenuItem.
3. Run code and click on Menu.
4. Observe Shadow above submenu / list.
5. Open notepad and click file to observe shadows on the right and bottom, but not above the list.

### Screenshots

![image](https://github.com/fyne-io/fyne/assets/30917551/6f14e69c-9fd5-4c15-8f93-97f9ad8543e8)

![image](https://github.com/fyne-io/fyne/assets/30917551/229e3837-b050-43e8-9f5d-14258a55867c)

### Example code

```go
menu := fyne.NewMenu("Help", fyne.NewMenuItem("About", func() {}))
```

### Fyne version

2.3.5

### Go compiler version

1.20.2

### Operating system and version

Windows 10

andydotxyz:
This should now be fixed on develop

Issue #4129, opened 2023-08-03
Support rounded gradients and shadows


Jacalz:
### Checklist

- [X] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same feature, before opening a new one.
- [X] This issue only relates to a single feature. I will open new issues for any other features.

### Is your feature request related to a problem?

A big part of the remaining work for the transition to a theme with rounded corners rely on implementing rounded corners on shadows and gradients so we can get menus and dialogs to use rounded corners. Until this is implemented, we will have some parts of the theme that must remain square.

### Is it possible to construct a solution with the existing API?

No. We need to implement the support for it.

### Describe the solution you'd like to see.

Support rounded corners for gradients and shadows so we can complete the transition.

andydotxyz:
For the purposes listed above I don't think we need round gradients *beyond the radial we already have) so calling this complete.

PR #5841, opened 2025-07-04
rectangle shadow

Vinci10:
### Description:

(Based on #5830)
Fixes:
- [#4129](https://github.com/fyne-io/fyne/issues/4129) (shadows only)
- [#4411](https://github.com/fyne-io/fyne/issues/4411) (shadows on the right and bottom can be added using proper offset)

**Shadow Support for Rectangle, Square, and Circle**
  - Introduced a `Shadow` struct ~and `Shadowable` interface~ to provide reusable shadow logic.
  - Added fields for shadow color, blur radius, offset, spread and type (drop/box shadow).
  - ~~The `ExpandForShadow` field allows the object to keep its requested/original size, expanding the total bounds to include the shadow.~~
  - Updated shader and painter logic to render shadows with configurable blur radius, spread and offset, and to blend shadows correctly.
  - ~~If a canvas object has a shadow added, the canvas size is reduced by the shadow paddings on each side. This ensures that the entire object—including its shadow—fits within the originally requested frame size. The content area (without shadow) will be smaller to accommodate the shadow within the allocated space.~~ If a canvas object has a shadow added, the canvas coords are expanded to accommodate the shadow.

- **Shader and Painter Updates**
  - Rectangle and round-rectangle shaders updated to support shadow rendering.
  - Painter logic updated to pass new parameters and handle shadow bounds.

- **API Additions**
  - New methods and fields are documented in code comments.
  - Backwards compatibility is maintained for existing code.

## Status

- I am looking for feedback and suggestions from the Fyne team regarding:
  - API design and naming
  - Performance considerations
- **Performance has not been fully evaluated.**

#### Additional Notes

This work is a prerequisite for adding shadows and rounded corners to popups, dialogs, and menus in Fyne. I have already applied these changes to those components and have working code locally. Rectangle shadow provides the necessary foundation for theme-level enhancements.

### Examples
shadow_examples
### Dialog, menu shadow (preview)
dialog_shadow
menu_shadow

---

**Please let me know if you have any suggestions, concerns, or requests for changes.**

### Checklist:

- [x] Tests included.
- [ ] Tests all pass.

#### Where applicable:


Vinci10:
By default, adding a shadow currently impacts the geometry: the canvas size is reduced by the shadow paddings so that both the object and its shadow fit inside the requested frame. This means the content (e.g., a rectangle) is shrunk to ensure the shadow does not overflow the frame.

To address this, I introduced the `ExpandForShadow` option. When set to `true`, the object keeps its requested/original size, and the overall canvas size is increased to fit the shadow outside the content.

### Example Calculations

#### Example 1

Suppose you have a rectangle with a requested frame size of **150 x 250**.

- **Shadow parameters:**
  - Offset: **(-10, 15)**
  - Softness: **8**

**Shadow paddings calculation:**
- Left:   `offsetX + softness`   = -10 + 8 = **-2** (clamped to 0)
- Top:    `-offsetY + softness` = -15 + 8 = **-7** (clamped to 0)
- Right:  `-offsetX + softness` = 10 + 8  = **18**
- Bottom: `offsetY + softness`  = 15 + 8  = **23**

After clamping negatives to zero:
- Left: **0**
- Top: **0**
- Right: **18**
- Bottom: **23**

| Mode                      | Rectangle Size | Canvas Size | Rectangle Position | Canvas Position |
|---------------------------|---------------|-------------|-------------------|----------------|
| No shadow                 | 150 x 250     | 150 x 250   | (0, 0)            | (0, 0)         |
| Shadow, ExpandForShadow=false | 132 x 227     | 150 x 250   | (0, 0)            | (0, 0)         |
| Shadow, ExpandForShadow=true  | 150 x 250     | 168 x 273   | (0, 0)            | (0, 0)         |

---

#### Example 2

Suppose you have a rectangle with a requested frame size of **100 x 100**.

- **Shadow parameters:**
  - Offset: **(12, -6)**
  - Softness: **5**

**Shadow paddings calculation:**
- Left:   12 + 5  = **17**
- Top:    6 + 5   = **11**
- Right:  -12 + 5 = **-7** (clamped to 0)
- Bottom: -6 + 5  = **-1** (clamped to 0)

After clamping negatives to zero:
- Left: **17**
- Top: **11**
- Right: **0**
- Bottom: **0**

| Mode                      | Rectangle Size | Canvas Size | Rectangle Position | Canvas Position |
|---------------------------|---------------|-------------|-------------------|----------------|
| No shadow                 | 100 x 100     | 100 x 100   | (0, 0)            | (0, 0)         |
| Shadow, ExpandForShadow=false | 83 x 89       | 100 x 100   | (17, 11)          | (0, 0)         |
| Shadow, ExpandForShadow=true  | 100 x 100     | 117 x 111   | (0, 0)            | (-17, -11)       |

---

**Notes:**

- When `ExpandForShadow` is **false**, the rectangle is shrunk and shifted inside the canvas to fit the shadow, so the rectangle's position changes (by the left/top padding), but the canvas position remains the same.
- When `ExpandForShadow` is **true**, the rectangle keeps its original position and size, but the canvas itself is shifted (by the left/top padding) to ensure the rectangle appears at the requested position, and the overall canvas size is increased to fit

This approach allows developers to choose whether the shadow should be included inside the original frame (shrinking the content) or expand the overall bounds to preserve the content size.

Please let me know if you’d like further clarification or adjustments to this behavior!

andydotxyz:
Copying from the wider discussion as it pertains to the API design here.

> You realise that an object is not restricted to drawing in its "frame"?

> I don't think we should add more APIs to control how the geometry will be interpreted, it will just lead to confusion.
Simple and standard is the way - provide the "path of least surprise".

I don't think the presence of a shadow should change how the geometry of a widget or shape is interpreted at all. With the maths illustrations above complexity goes up substantially and widgets with shadows will no longer line up with others which breaks the simplicity of our layout model.

andydotxyz:
> https://github.com/fyne-io/fyne/issues/4411 (shadows on the right and bottom can be added using proper offset)

Just FYI on that one, the bug is not about the drop of the shadow or its depth - the report is that a menu popping up from a menu bar should be at the same level so there should be shadow on 3 sides of the menu and the underside of the bar, but not between them (top of the menu)

canvas/base_shadow.go
```diff
@@ -0,0 +1,101 @@
...
+// Content refers to the area without shadow, while the shadow is rendered outside this area.
+// The canvas object size is reduced to include the shadow within the allocated frame.
+//
+// Since: 2.7
+type Shadowable interface {
+	// SizeAndPositionWithShadow returns the total size and position needed to render the shadow around the content
+	// with the input size. Total size is larger than the input size.
+	SizeAndPositionWithShadow(size fyne.Size) (fyne.Size, fyne.Position)
+	// ShadowPaddings returns the paddings (left, top, right, bottom) for the shadow.
+	ShadowPaddings() [4]float32
```
andydotxyz: This seems like the only one of the interface methods that adds (rather than works around geometry complications).

However I wonder if ShadowOffset and ShadowDepth really are all we need to draw shadows around objects - all these calculations could be removed and made a rendering only complexity.

canvas/base_shadow.go
```diff
@@ -0,0 +1,101 @@
...
+
+// Ensure baseShadow implements Shadowable.
+var _ Shadowable = (*baseShadow)(nil)
+
+// baseShadow provides base functionality for objects that can have a shadow.
+// Intended to be embedded in other structs to add shadow support.
+type baseShadow struct {
+	baseObject
+
+	ShadowColor    color.Color   // Color of the shadow.
```
andydotxyz: One API consideration for this PR should be:

* Is the shadow a themed colour or a canvas feature

Which impacts how it should be implemented. If widgets are to have shadows it must be in the theme - but for canvas objects you get lower level control. Worth considering in the bigger picture how this replaces the existing (gradient) shadow implementation. 

We should replace old with new in the scope of this PR rather than having two different shadow implementations - plus it will inform the correct API.

andydotxyz:
Very cool indeed thanks. I have made a few API /design comments inline.

Another thing to note would be that we have also the software renderer which would need to understand this new feature and draw something similar for compatibility.
This would also open up easier unit testing.

Vinci10:
> > #4411 (shadows on the right and bottom can be added using proper offset)
> 
> Just FYI on that one, the bug is not about the drop of the shadow or its depth - the report is that a menu popping up from a menu bar should be at the same level so there should be shadow on 3 sides of the menu and the underside of the bar, but not between them (top of the menu)

You're right, the top shadow is the problem there. I just wanted to point out that a rectangular shadow may help with that.

> Copying from the wider discussion as it pertains to the API design here.
> 
> > You realise that an object is not restricted to drawing in its "frame"?
> 
> > I don't think we should add more APIs to control how the geometry will be interpreted, it will just lead to confusion.
> > Simple and standard is the way - provide the "path of least surprise".
> 
> I don't think the presence of a shadow should change how the geometry of a widget or shape is interpreted at all. With the maths illustrations above complexity goes up substantially and widgets with shadows will no longer line up with others which breaks the simplicity of our layout model.

I also prefer simple solutions :) To be honest, I am not an expert in such graphical implementations. The biggest challenge was to draw the shadow correctly and that's what I focused all my attention on. However, all changes in geometry were made only to "see" the desired effect, because without any change the "frame" clipped the shadow and I was not able to assess whether it was drawing correctly. That's why I am open to any suggestions regarding the API and changes on the library side.

andydotxyz:
>  because without any change the "frame" clipped the shadow

This will be a limitation of the rendering and not part of the Fyne geometry design. If you look at rectangle, line and others you will see a `vectorPad` value which is how we expand the frame to draw outside of the original rectangle if required. Expanding that to accommodate your new shadow should fix the issue and all the geometry changes can be reverted.

Vinci10:
> > because without any change the "frame" clipped the shadow
> 
> This will be a limitation of the rendering and not part of the Fyne geometry design. If you look at rectangle, line and others you will see a `vectorPad` value which is how we expand the frame to draw outside of the original rectangle if required. Expanding that to accommodate your new shadow should fix the issue and all the geometry changes can be reverted.

That would be great, so I can just extend `vectorPad` by shadow paddings and it would work? I will test that approach

I don't fully understand how Fyne's design geometry works, and that's why I added so much unnecessary code.

canvas/base_shadow.go
Vinci10: What about gradient shadow implementation if only one-side shadow is needed like for scroll ? I think it still may be useful, but I could be wrong

canvas/base_shadow.go
Vinci10: So all calculations should be moved to painter `draw.go` file ?

andydotxyz:
> That would be great, so I can just extend `vectorPad` by shadow paddings and it would work? I will test that approach

Yes

> I don't fully understand how Fyne's design geometry works, and that's why I added so much unnecessary code.

In essence it is really simple - a device agnostic coordinate system so it's consistent across all devices. Every item is positioned relative to the parent. More info at https://docs.fyne.io/architecture/geometry

Vinci10:
I reverted geometry changes. As of now shadow paddings expand `coords` in `vecRectCoordsWithPad` method. Please let me know if that is correct.

Result:
shadow

canvas/base_shadow.go
Vinci10: The interface has been updated. Now only the `ShadowPaddings` method is available

Vinci10:
I added shadow to software painter along with test cases

andydotxyz:
> I reverted geometry changes. As of now shadow paddings expand `coords` in `vecRectCoordsWithPad` method. Please let me know if that is correct.

Thanks, I appreciate that when packed together it looks peculiar but I think this is better for the geometry.
It does raise the question of whether shadows should be drawn below all other items, but I think that is maybe not needed and maybe for another time.

internal/painter/software/draw.go
```diff
@@ -246,13 +256,82 @@ func drawOblong(c fyne.Canvas, obj fyne.CanvasObject, fill, stroke color.Color,
...
 	scaledWidth := scale.ToScreenCoordinate(c, width)
 	scaledHeight := scale.ToScreenCoordinate(c, height)
 	scaledX, scaledY := scale.ToScreenCoordinate(c, pos.X), scale.ToScreenCoordinate(c, pos.Y)
 	bounds := clip.Intersect(image.Rect(scaledX, scaledY, scaledX+scaledWidth, scaledY+scaledHeight))
-	draw.Draw(base, bounds, image.NewUniform(fill), image.Point{}, draw.Over)
+
+	if shadowColor != color.Transparent && shadowColor != nil && (!shadowOffset.IsZero() || shadowSoftness > 0.0) {
+		bounds = drawShadow(c, obj, fyne.NewSize(width, height), shadowSoftness, shadowOffset, shadowColor, 0, bounds, base, clip)
+		// due to shadow draw rectangle with a certain width and height
+		raw := painter.DrawRectangle(canvas.NewRectangle(fill), width, height, 0, func(in float32) float32 {
```
andydotxyz: I don't understand this - is it the cause of test failures?
If the geometry is not changed then isn't it the same rectangle but drawn at a different offset?

canvas/shadow.go
```diff
@@ -0,0 +1,66 @@
...
+	BoxShadow
+)
+
+// Shadowable defines an interface for objects that can render a shadow.
+// It provides methods to retrieve the shadow paddings.
+//
+// Since: 2.7
+type Shadowable interface {
+	// ShadowPaddings returns the paddings (left, top, right, bottom) of the shadow.
+	ShadowPaddings() [4]float32
```
andydotxyz: This seems to be derived from the public shadow struct data - what is the purpose of having a new interface with this method when it could be internal to the rendering?

canvas/rectangle.go
```diff
@@ -12,6 +12,10 @@ var _ fyne.CanvasObject = (*Rectangle)(nil)
 // Rectangle describes a colored rectangle primitive in a Fyne canvas
 type Rectangle struct {
 	baseObject
+	// Support shadow configuration
```
andydotxyz: Embedding at the beginning of the struct like this will break any code that has done `&Rectangle{color.Transparent}` - though that is not best practice we try to avoid breaking if we can.

andydotxyz (changes requested):
Thanks for working more on this, it is coming together.

I have a few questions at this time:
1) Isn't the geometry in the software painter wrong? The objects are moving out of their position to accommodate the shadow
2) The Shadowable interface isn't clear - what is its purpose? It doesn't seem to be needed 
3) The public `Shadow` type can only be used as embedded in other types - what are your thoughts on this balance?

canvas/base_shadow.go
andydotxyz: Yes a fair point - but for the major shadow usages (where it is a complete shadow) we should replace and test

canvas/base_shadow.go
andydotxyz: Thanks. As noted on the recent review now that it is reduced to padding calculations it seems not needed at all...

internal/painter/software/draw.go
Vinci10: to accomodate shadow the frame must be extended, without this change rectangle always fill in the entire space and the shadow is not visible

canvas/rectangle.go
Vinci10: thanks, I will move the embedding at the end of the struct

canvas/shadow.go
Vinci10: The interface is useful when I have `obj fyne.CanvasObject` as an argument and want to quickly parse it to the new interface `s, ok := obj.(canvas.Shadowable)` to get paddings `ShadowPaddings()` but maybe there is a better approach without using new interface.

Vinci10:
> Thanks for working more on this, it is coming together.
> 
> I have a few questions at this time:
> 
> 1. Isn't the geometry in the software painter wrong? The objects are moving out of their position to accommodate the shadow
> 2. The Shadowable interface isn't clear - what is its purpose? It doesn't seem to be needed
> 3. The public `Shadow` type can only be used as embedded in other types - what are your thoughts on this balance?

Thanks for review.
1. Unfortunately, without changing object position when I expand the frame (to accommodate the shadow) the object is always located on the upper-left corner, in such case I cannot render and test shadow with upper-left offset properly. I am testing changes on `WSL2 Ubuntu` and the new tests succeed. Not sure what could be wrong :(
2. With interface it is easy to parse `fyne.CanvasObject` object to the new `Shadowable` interface and call `ShadowPaddings()` method. But probably it is not a good argument. I also thought the new interface would help with adding new shadow types in the future, but I'm not sure if it makes sense.
3. Sure, sounds good. So should it be a public struct and how should I call `ShadowPaddings()` method for `fyne.CanvasObject` objects ? Thanks in advance.

Which implementation is expected ?
a (embedded). 
```
type Rectangle struct {
	baseObject

        // ....

	// Support shadow configuration
	//
	// Since: 2.7
	Shadow
}
``` 

b (public field).
```
type Rectangle struct {
	baseObject

        // ....

	// Support shadow configuration
	//
	// Since: 2.7
	Shadow shadow
}
```

internal/painter/software/draw.go
Vinci10: I see there is issue on macos platform, could you please show me the rendered images on macos so I can try to investigate the problem because on my local windows platform all tests succeed.

Vinci10:
I have removed `ShadowingRenderer` and applied rectangle shadow to the following objects:
## Containers:
- InnerWindow
## Widgets:
- Card
- Menu
- Popup
## Internal GLFW:
- MenuBar

I would like to highlight that I set some initial values for the shadow softness and offset to show what it might look like and how the test might change. I was able to update some platform tests to be successful.
Could you please help me with finding proper values for shadow parameters for the objects so I can start fixing tests for mobile and other platforms.

canvas/base_shadow.go
Vinci10: I replaced ShadowingRenderer and updated tests. I started with some initial values for shadow parameters which should be adjusted to make shadow looking good. I still need to update macos and mobile tests. I can start fixing tests once final shadow parameters per object (dialogs, widgets, etc.) are determined.

Vinci10:
If you don't mind, I'd like to continue working on this feature for version 2.8.
I removed Shadowable interface, but the tests still need updating.

andydotxyz:
How is this looking for next release? I am trying to plan for 2.8 and not sure if this will make it.

There are a few questions left I think (other than the tests)

- Can we keep geometry consistent as shadows are turned on/off
- On naming would we be better to find a way to "apply shadow to" an object rather than "object contains a shadow" (just from an API point of view)

I'm struggling on the api/naming side of things - maybe what is proposed is the best balance, I would be keen to know what others think...

Vinci10:
I can continue working on this feature, but I need help with the shadow configuration values ​​for specific widgets like dialogs, card, etc.

> * Can we keep geometry consistent as shadows are turned on/off

Do you mean software or gl painter ? I am not sure what the behavior is expected, for the gl object position is not changing when the shadow is enabled or disabled. Software painter geometry is changing, from the tests perspective it looks fine not sure about the impact on "live" apps because I ran only apps with gl painter.

> * On naming would we be better to find a way to "apply shadow to" an object rather than "object contains a shadow" (just from an API point of view)

Do you want a generic API that accepts any canvas object (rectangle, image, polygon) and creates a shadow for it? I'm fine with that, but I don't see a better solution than using separate shadow logic in each shader, because shadows require a distance from the object's edge, which is calculated in the shader.

andydotxyz:
> Do you mean software or gl painter ? I am not sure what the behavior is expected, for the gl object position is not changing when the shadow is enabled or disabled. Software painter geometry is changing

I mean that the geometry should not change when a shadow is turned on. And the painters should absolutely be consistent. It sounds like the GL shaders are doing what I would have expected but that the software painter is causing change in layout that is not desirable.

> Do you want a generic API that accepts any canvas object (rectangle, image, polygon) and creates a shadow for it? I'm fine with that, but I don't see a better solution than using separate shadow logic in each shader,

It's likely that the public API and the internal implementation are not exactly the same. I am not certain what the right API should be, but coding the effect into each shader is likely required as you say. But that does not mean we have to expose the configuration alongside the shape specification publicly. Keen to get other peoples thoughts. @sdassow @redawl @dweymouth

Vinci10:
I understand that geometry cannot change after adding a shadow. I will try to fix this in the software painter. Regarding API, I will adjust the code to your suggestions.

Vinci10:
To start fixing the tests, it would be good to know the default theme shadow values ​​(offset, radius, etc.) for each widget, because any shadow change results in different test data. I don't think I'm the right person to decide how the shadow should look like for individual widgets.

Vinci10:
@andydotxyz Geometry for the software painter should be consistent with gl now, it means object position is not changed when the shadow is turned on.

Tests for canvas with shadow fail on macOS, I am not sure why, on Ubuntu and Windows tests pass. I remember that for other objects like polygon I had to increase object size to pass tests on macos but for shadow even higher size not help.

andydotxyz:
> Tests for canvas with shadow fail on macOS, I am not sure why, on Ubuntu and Windows tests pass.

Just FYI that's not what CI is showing - all OSes are failing with mismatched images or xml. Are you sure you copied in all the new golden masters?

Vinci10:
For dialogs and widgets, the tests fail because I'm not sure what the correct shadow configuration is, so I haven't updated the test data (image, xml) yet.

But `canvas` and `internal/painter/software` tests should pass for all OSes however in CI logs I see mismatch errors only for macos.

Below errors occur only on macos
```
--- FAIL: TestPainter_paintCircle_shadow (0.21s)
    util_helper.go:61: 
        	Error Trace:	/Users/runner/work/fyne/fyne/internal/test/util_helper.go:61
        	            				/Users/runner/work/fyne/fyne/test/test_helper.go:78
        	            				/Users/runner/work/fyne/fyne/internal/painter/software/painter_test.go:144
        	Error:      	Images not equal.
        	Test:       	TestPainter_paintCircle_shadow
        	Messages:   	Image did not match master. Actual image written to file:///Users/runner/work/fyne/fyne/internal/painter/software/testdata/failed/draw_circle_box_shadow.png.
```

```
--- FAIL: TestRectangle_shadow (0.31s)
    util_helper.go:61: 
        	Error Trace:	/Users/runner/work/fyne/fyne/internal/test/util_helper.go:61
        	            				/Users/runner/work/fyne/fyne/test/test_helper.go:78
        	            				/Users/runner/work/fyne/fyne/test/test_helper.go:89
        	            				/Users/runner/work/fyne/fyne/canvas/rectangle_test.go:138
        	Error:      	Images not equal.
        	Test:       	TestRectangle_shadow
        	Messages:   	Image did not match master. Actual image written to file:///Users/runner/work/fyne/fyne/canvas/testdata/failed/rounded_rect_stroke_shadow.png.
```

After applying this patch https://github.com/fyne-io/fyne/pull/5889#issuecomment-3178515117 and increase size some `canvas` test cases pass on macos but most of the files still don't match

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.19
 require (
 	fyne.io/systray v1.12.0
 	github.com/BurntSushi/toml v1.5.0
+	github.com/anthonynsimon/bild v0.13.0
```
sdassow, line 8: Looks like a duplicate.

widget/testdata/select/desktop/theme_changed.png
sdassow, line 1: This one might be an example why some shadow visibility on the top and left could improve the visual outline, but then again it's an extreme case. But at least in the context of a shadow configuration this might be relevant.

internal/driver/glfw/testdata/menu_bar_hovered_file_recent_older.png
sdassow, line 1: If anyone would ask me why I prefer the new shadow style to the previous one, this is it.
The visual effect before made it seem as if UI elements extruded towards the users face, while the new drop shadow style appears more observational and as part of one surface, but also makes more sense in hierarchical terms as can be seen with the cascading menu.

test/markup_renderer.go
```diff
@@ -278,6 +278,7 @@ func (r *markupRenderer) writeCircle(c *fynecanvas.Circle, attrs map[string]*str
 	r.setColorAttr(attrs, "fillColor", c.FillColor)
 	r.setColorAttr(attrs, "strokeColor", c.StrokeColor)
 	r.setFloatAttr(attrs, "strokeWidth", float64(c.StrokeWidth))
+	r.setShadowAttrs(c.Shadow, attrs)
```
sdassow: Compared to the other setters, the argument order for `setShadowAttrs` seems to be reversed (the calls above all have `attrs` first).

sdassow (changes requested):
Really cool!

sdassow:
> For dialogs and widgets, the tests fail because I'm not sure what the correct shadow configuration is, so I haven't updated the test data (image, xml) yet.
> 

To me it sounds like two related questions:
- How close to the previous shadow configuration should it be?
- Should we use the opportunity to change it?

And if the answer to the second question is "yes", the first question doesn't matter much, which I interpret as our current situation.

I for one like the new suggested shadow configuration so far.

Vinci10:
Thanks for the review, as changing the shadow causes a lot of golden files to need updating and since the biggest advantage of the new shadow implementation is the ability to round corners, do you mind if I add new theme variables to round corners for dialog, menu and card in this PR similarly to #6268 ?

sdassow:
> Thanks for the review, as changing the shadow causes a lot of golden files to need updating and since the biggest advantage of the new shadow implementation is the ability to round corners, do you mind if I add new theme variables to round corners for dialog, menu and card in this PR similarly to #6268 ?

Making the shadow configuration themeable sounds like a good idea. Not sure to what extent tho, since I can imagine ending up with a lot of new theme variables, but maybe that's okay?

go.mod
Vinci10: thanks, fixed

test/markup_renderer.go
Vinci10: fixed

Vinci10:
I was thinking about corner radii, but if you want shadow configuration can also themeable.

I implemented a shadow configuration for the widgets trying to imitate Material Design along with the new theme variables for the corner radii:
- SizeNameCardRadius 
- SizeNameDialogRadius 
- SizeNamePopupRadius 
- SizeNameMenuRadius

Tests for Ubuntu and Windows are running successfully, although there is a problem with tests on macOS (unfortunately I am unable to verify that)

widget/testdata/select/desktop/theme_changed.png
Vinci10: I updated the test case, please check if it's better now

internal/driver/glfw/testdata/menu_bar_hovered_file_recent_older.png
Vinci10: I updated the test case, I hope it still meets your expectations

andydotxyz (changes requested):
I like some of this API - however the `ShadowConfig` seems like it could be internal only (or not needed at all?) if the `ApplyShadowConfig` was actually taking the level directly (i.e. ApplyShadowForLevel).

Vinci10:
Good catch. `ShadowConfig` was redundant. I've removed it and replaced `ApplyShadowConfig` with `ApplyShadowForLevel`.

andydotxyz:
The macOS seems to have problems - any thoughts on that? Do you need someone on macOS to show the exact image failures?

Vinci10:
I have no idea why the tests don't work on macOS :( I am not able to generate and validate images for macOS

Vinci10:
I realized that sometimes rectangles are smaller than expected. I guess it was casued by rounding. 
image
Hence, I have modified `scale` function in software painter https://github.com/fyne-io/fyne/pull/5841/changes/29eb39ed02379a23b0df10be58dbe4f43b86ff32
Now the menu background looks fine and also it has positive impact on other widgets 
image

andydotxyz:
This is a tricky subject and I would recommend using what is in other areas of the same file - `ToScreenCoordinate` for consistency.

Rounding will sometimes be smaller than you might expect - but Ceil will sometimes be bigger.

If you use Ceil in this place but elsewhere we have other resolution then the rectangles will not match position of other items exactly.

Vinci10:
I'm not sure I fully understand. I haven't modified or replaced `ToScreenCoordinate`. Basically, all objects in `internal/painter/draw.go` have the same `scale` function passed to them. It could be generalized to single shared function for consistency.

andydotxyz:
Sorry, I was looking at the software renderer - but my point is that rounding should not make things too small - the left and right edge should both be rounded. If there is an unexpected gap it's probably due to an inconsistency.

If you look at the history of the files we have tweaked this a lot over the years to remove such gap/overlap issues.

Vinci10:
I dont' want to complicate this PR. Let me revert that commit. I see some inconsistency occurring in the test cases but it can be handled in separate PR.

Vinci10:
I modified github actions to upload failed testdata. For macOS failed images look the same so I changed `pixCloseEnough`.

andydotxyz:
The conflicts were severe so I took care of it myself to land for 2.8.

Also I:
* Removed the constructor (misleading as it did not create a CanvasObject)
* Renamed FillColor as Color
* Extended it to add the same shadow to the Ellipse type.

PR #5842, opened 2025-07-04
Add cursor functions to Entry

### LINKED ISSUE #5753: Entry widget - find position of cursor in the source text
### Checklist

There is a need for the API to expose the rune length of each visible line of a multi-line entry (especially with TextWrapWord selected).

Let us say that there is an entry widget spanning several lines whereby there are no line breaks in the text. Suppose there are 127 runes in the text. There is no way to currently know where my cursor is with respect to the entire text in the widget. The visible row and column of the cursor is exposed, but it is necessary to know the total rune length of each row - in order to calculate where the cursor is with respect to the entire text.

Were the API to expose the rune length of each visible line, this could be calculated.

Thank you in advance,
Isaac

I have tried to use Monospace font and a fixed entry width, and with this one can solve the problem. However, this is too limiting.

The API should either expose:

1) The rune length of each line, or
2) The absolute position of the cursor with respect to the entire text

andydotxyz:
Fixes #5753

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/widget/entry.go
+++ b/widget/entry.go
@@ -208,6 +208,30 @@ func (e *Entry) CreateRenderer() fyne.WidgetRenderer {
 	return &entryRenderer{box, border, e.scroll, objects, e}
 }
 
+// CursorPosition returns the relative position of this Entry widget's cursor.
+//
+// Since: 2.7
+func (e *Entry) CursorPosition() fyne.Position {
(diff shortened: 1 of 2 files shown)


Jacalz (approved):
Nice one :)

PR #5846, opened 2025-07-07
Make sure tab container's don't refresh content when creating renderer

Jacalz:
### Description:

I noticed that there were quite a few places where the tab code was refreshing content when creating the renderer even though that is handled automatically. There is likely more we can clean up and improve with tabs in this regard (I am not a big fan of layouts handling content refresh) but this is a good first step.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


redawl (commented):
Would it be better to just explicitly call `Refresh` after `updateTabs` and `applyTheme` in the places where it is needed, so we don't need the `refresh` argument and condition?

Jacalz:
That's a nice idea. I'll check it out tomorrow. I might also see if it makes sense to tackle layouts doing refresh as well. That one is bugging me :)

Jacalz:
That's a nice idea. I'll check it out tomorrow. I might also see if it makes sense to tackle layouts doing refresh as well. That one is bugging me :)

Jacalz:
Gah. I don't understand why the tests are failing. Giving up for now...

Jacalz:
Alright. Tests are constantly failing on mobile locally even without this change. That explains why I was going mad

Jacalz:
I can't get the tests to pass with the more advanced changes and I am running out of time before I'm going on a holiday tomorrow. Marking this as ready for review with the trivial change I initially started work on.

redawl (approved):
Looks good. Enjoy your holiday :)

Jacalz:
Thanks :)

Issue #5486, opened 2025-01-31
Unbounded internal cache sizes can lead to unbounded RAM use


dweymouth:
### Checklist

- [x] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [x] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

Internal caches are currently unlimited in size, and are cleaned out on a fixed 1-minute schedule. This means that within a minute, apps can grow in RAM use unboundedly, possibly leading to out of memory errors, if they create a huge number of cached objects (especially text textures) within that time. For example, scrolling quickly through 10s of thousands of text-heavy list rows (which could contain on the order of 10 canvas.Texts each).

Possible ideas are to take inspiration from the Go GC itself - run our clean tasks more frequently based on use patterns, and have a "soft limit" on cache size which is based on the past patterns of use.

### How to reproduce

scroll through many many rows of a text-heavy collection widget (eg extend the list or table in fyne_demo to be a million rows long)

### Example code

fyne_demo

### Fyne version

develop

### Go compiler version

n/a

### Operating system and version

n/a

PR #5849, opened 2025-07-08
Adding a blur graphical primitive and use it under modal popups & dialogs

andydotxyz:
Tested on:

- [x] macOS
- [x] Linux
- [x] Windows
- [x] Android
- [x] iOS
- [x] WASM <- depends on fix fyne-io/gl-js#17

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/fogleman/gg v1.3.0
 	github.com/fredbi/uri v1.1.1
 	github.com/fsnotify/fsnotify v1.9.0
-	github.com/fyne-io/gl-js v0.2.0
+	github.com/fyne-io/gl-js v0.2.1-0.20260315212741-029c47fd27e8
 	github.com/fyne-io/glfw-js v0.3.0
 	github.com/fyne-io/image v0.1.1
 	github.com/fyne-io/oksvg v0.2.0
--- a/go.sum
+++ b/go.sum
@@ -27,8 +27,8 @@ github.com/fredbi/uri v1.1.1/go.mod h1:4+DZQ5zBjEwQCDmXW5JdIjz0PUA+yJbvtBv+u+adr
 github.com/fsnotify/fsnotify v1.4.7/go.mod h1:jwhsz4b93w/PPRr/qN1Yymfu8t87LnFCMoQvtojpjFo=
 github.com/fsnotify/fsnotify v1.9.0 h1:2Ml+OJNzbYCTzsxtv8vKSFD9PbJjmhYF14k/jKC7S9k=
 github.com/fsnotify/fsnotify v1.9.0/go.mod h1:8jBTzvmWwFyi3Pb8djgCCO5IBqzKJ/Jwo8TRcHyHii0=
-github.com/fyne-io/gl-js v0.2.0 h1:+EXMLVEa18EfkXBVKhifYB6OGs3HwKO3lUElA0LlAjs=
-github.com/fyne-io/gl-js v0.2.0/go.mod h1:ZcepK8vmOYLu96JoxbCKJy2ybr+g1pTnaBDdl7c3ajI=
+github.com/fyne-io/gl-js v0.2.1-0.20260315212741-029c47fd27e8 h1:0kdPD/GEntpWmZEK5Zu/xE6Tr37jYCVDf9QP8lA/QK8=
+github.com/fyne-io/gl-js v0.2.1-0.20260315212741-029c47fd27e8/go.mod h1:ZcepK8vmOYLu96JoxbCKJy2ybr+g1pTnaBDdl7c3ajI=
 github.com/fyne-io/glfw-js v0.3.0 h1:d8k2+Y7l+zy2pc7wlGRyPfTgZoqDf3AI4G+2zOWhWUk=
 github.com/fyne-io/glfw-js v0.3.0/go.mod h1:Ri6te7rdZtBgBpxLW19uBpp3Dl6K9K/bRaYdJ22G8Jk=
 github.com/fyne-io/image v0.1.1 h1:WH0z4H7qfvNUw5l4p3bC1q70sa5+YWVt6HCj7y4VNyA=


internal/painter/gl/draw.go
```diff
@@ -26,6 +27,79 @@ func (p *painter) defineVertexArray(prog Program, name string, size, stride, off
...
 
+func (p *painter) drawBlur(b *canvas.Blur, pos fyne.Position, frame fyne.Size) {
+	if b.Radius == 0 {
+		return
+	}
+	radius := b.Radius * p.pixScale
+
+	w := roundToPixel(frame.Width*p.pixScale, 1.0)
+	h := roundToPixel(frame.Height*p.pixScale, 1.0)
+	pixels := make([]uint8, int(w*h)*4)
```
dweymouth: This function is called every screen repaint, not just every refresh of the blur object itself, so this can be a potentially large allocation that ends up running 60 times a second (e.g. a modal dialog with an Activity widget animating). We should reuse the allocated pixels slice, either caching it on the *canvas.Blur itself, or maybe using a pool in the painter of pixbufs.

internal/painter/gl/draw.go
```diff
@@ -26,6 +27,79 @@ func (p *painter) defineVertexArray(prog Program, name string, size, stride, off
...
+
+	vertAttrib := p.ctx.GetUniformLocation(p.blurProgram, "radius")
+	p.ctx.Uniform1f(vertAttrib, radius)
+
+	vertAttrib = p.ctx.GetUniformLocation(p.blurProgram, "size")
+	p.ctx.Uniform2f(vertAttrib, w, h)
+
+	sum := float32(0.0)
+	length := int(radius)*2 + 1
+	values := make([]float32, length)
```
dweymouth: Same here re. allocation at 60 fps. But at least this one is small, compared to the pixbuf. But we could (and maybe should) still take an approach like storing this slice on the canvas.Blur itself so the allocated space can be reused next time.

internal/painter/gl/draw.go
andydotxyz: Nice catch thanks.

This will be a tricky object to optimise for because if anything in the window refreshes we might have to update.
But yes we should avoid the allocations, thanks. I know there are other potential optimisations but others will be less obvious!

internal/painter/gl/painter.go
```diff
@@ -36,6 +36,8 @@ type Painter interface {
 // If it is a master painter it will also initialise OpenGL
 func NewPainter(c fyne.Canvas, ctx driver.WithContext) Painter {
 	p := &painter{canvas: c, contextProvider: ctx}
+	p.blurKernels = make(map[float32][]float32)
```
dweymouth: Is there a reason not to cache the kernel on the `canvas.Blur` itself? With a map like this we need a clean task / expiry too, and it seems that caching it on the blur object would avoid the overhead of a map lookup (which, granted, is small). You can have
```go
package "canvas"
type Blur struct {
    // ... public fields
    
    cachedKernelRadius float32
    cachedKernel []float32
}
```

Then you can know whether the kernel needs to rebuilt for a new radius (if the .Radius property on the blur has been updated).

andydotxyz: Well, it is a detail of the renderer not the blur. and a 5pt kernel is the same for all blurs - seemed like the cache would be smaller.

dweymouth: For all the ones the theme applies to dialog backgrounds, yes. But user code could do something like animate a blur over a range of radii, causing a potentially unbounded number of entries into the blurKernels cache. We can keep it a map, but we'll have to make sure we have a system to delete from it, like all caches.

dweymouth: Another option to cache on the canvas.Blur but to more closely follow this
> Well, it is a detail of the renderer not the blur. 

would be to create a field on the Blur struct: `painterData any`, that whichever loaded painter is allowed to use, so the logic to generate the kernel would be kept in the gl painter, and it would type-assert that field to its own `blurKernel` struct or whatever; it just is allowed to tack it onto the Blur object itself instead of needing a map.

canvas/blur.go
```diff
@@ -0,0 +1,53 @@
...
+
+// Hide will set this blur to not be visible
+func (b *Blur) Hide() {
+	b.baseObject.Hide()
+
+	repaint(b)
+}
+
+// Move the blur to a new position, relative to its parent / canvas
+func (b *Blur) Move(pos fyne.Position) {
```
Jacalz, line 28: We might want to apply the same approach as in https://github.com/fyne-io/fyne/pull/5816 here.

internal/driver/mobile/gl/work.h
```diff
@@ -12,7 +12,7 @@
 #include <GLES3/gl3.h> // install on Ubuntu with: sudo apt-get install libegl1-mesa-dev libgles2-mesa-dev libx11-dev
 #elif os_openbsd
 #include <GLES3/gl3.h>
-#elif os_freebsd
+#elif os_freebsdUniform4
```
Jacalz: This looks like a copy-paste error?

internal/painter/gl/painter.go
Jacalz: I think the caching here makes more sense like it is now. Breaking the barrier between objects and internal details seems undesirable. However, I do think that the potentially unbounded map growth is a problem that needs to be fixed.

internal/painter/gl/draw.go
Jacalz: We probably don't even need a pool given that it is single threaded code. Just having a reusable buffer available here in the painter should be good enough. Caching it on the object will just lead to more memory usage and be unnecessarily complicated.

Jacalz:
Nitpick but the checklist item for binary size increase when importing module should probably be checked :)

internal/painter/gl/painter.go
dweymouth: Agreed on the unbounded map growth being a must-solve before we can merge this. I'm still not convinced that slightly breaking the barrier between object and painter details isn't worth it for easier to understand or maintain code. But I'm fine with the map approach, as long as we add the `expiringCache` metadata to the entries and add it to the cache clean task along with the others (`fontMetrics`, etc).

dweymouth: worth noting that [this issue](https://github.com/fyne-io/fyne/issues/5486) would apply here as well - but I had a draft PR to address this (which I really should pick up again sometime) and can add the `blurKernel` cache to this new system as well.

andydotxyz: Bear in mind the kernel cache is a OpenGL requirement not all renderers. 

Can you expand on why caching a rendering lookup table on the object is clearer?

Also I'm unsure how a private field on a struct in a different package is even something the renderer code can reference... I feel I'm missing something.

dweymouth: > Also I'm unsure how a private field on a struct in a different package is even something the renderer code can reference...

Oh, of course you're right about that 🤦 

My other original thought was that computing the kernel itself doesn't seem expensive compared to the allocation (just a pair of non-nested loops over what usually will be like 8 iterations only). Since we're single threaded now and only one blur will ever be painted at a time, even a static []float32 slice declared in the gl package that the painter can reset the length to zero (NOT setting it to nil) and re-fill the same memory buffer with a freshly-computed kernel for each paint would probably be viable.

Same idea applies for keeping a static []uint8 slice that can be zeroed out and reused to capture the pixel data under the blur on each paint, for the other allocation

andydotxyz: I'm not getting why we would choose to re-compute every kernel for every frame over using some memory to cache them...

dweymouth: We can use memory to cache them but we **must** add the cache clean task then, so that we don't retain that memory indefinitely for every blur radius ever asked for by the app. Right now we are only ever putting things into the `blurKernels` map and never deleting.

andydotxyz:
iOS is good but performance is not good enough IMHO.
Will work on an optimisation that should also fix WASM support too.

andydotxyz:
WASM is working AND performance is significantly better than before, getting close to ready I think...

andydotxyz:
dialog-blur
dialog-blur2

andydotxyz:
Caching is also in now, clearing on the same loop as textures etc

canvas/blur.go
```diff
@@ -0,0 +1,56 @@
...
+var _ fyne.CanvasObject = (*Blur)(nil)
+
+// Blur creates a rectangular blur region on the output.
+// All objects drawn under this will be blurred, any above will not be affected.
+//
+// Since: 2.7
+type Blur struct {
+	baseObject
+
+	// Radius refers to how far from a pixel should be used to calculate the blur.
```
dweymouth: Is this really based on pixels (ie not scaled dp) or is the comment just a typo?

andydotxyz: Apologies, quite right

PR #5851, opened 2025-07-09
Don't override user choice if a) xdg lookup fails or b) system updates

andydotxyz:
### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz:
I found that FYNE_THEME was being ignored if the xdg/portal lookup failed.
But also if it succeeds or any system change happens then the fact it was set is overwritten.

Jacalz (approved):
Nice :)

PR #5853, opened 2025-07-09
TextGrid: Only render cells inside the viewport

### LINKED ISSUE #5815: Two textgrids inside NewHSplit are considerably laggy

redawl:
### Description:
The idea here is to reduce the amount of canvas objects fyne has to walk in order to draw very large textgrids that extend out in the horizontal direction.

This is a work in progress. Any ideas are appreciated :)

Fixes #5815

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.
- [x] Works with all types of scrolls
- [ ] Works without ScrollNone
- [ ] Attempt to remove Refresh call, hopefully replace with Layout instead.

### DIFF
--- a/widget/textgrid.go
+++ b/widget/textgrid.go
@@ -505,8 +505,8 @@ func (t *textGridContent) CreateRenderer() fyne.WidgetRenderer {
 
 	r.updateCellSize()
 	t.text.scroll.OnScrolled = func(_ fyne.Position) {
-		r.addRowsIfRequired()
 		r.Layout(t.Size())
+		r.Refresh()
 	}
 	return r
 }
@@ -542,9 +542,12 @@ func (t *textGridContentRenderer) Destroy() {
 func (t *textGridContentRenderer) Layout(s fyne.Size) {
 	size := fyne.NewSize(s.Width, t.text.cellSize.Height)
 	t.updateGridSize(s)
-
+	xOffset := float32(0)
+	if t.text.text.Scroll != fyne.ScrollNone {
+		xOffset = float32(math.Floor(float64(t.text.text.scroll.Offset.X)))
+	}
 	for _, o := range t.text.visible {
-		o.Move(fyne.NewPos(0, float32(o.(*textGridRow).row)*t.text.cellSize.Height))
+		o.Move(fyne.NewPos(xOffset, float32(o.(*textGridRow).row)*t.text.cellSize.Height))
 		o.Resize(size)
 	}
 }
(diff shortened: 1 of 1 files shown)


widget/textgrid.go
```diff
@@ -755,15 +760,19 @@ func (t *textGridRow) addCellsIfRequired() {
 	if len(t.objects) == cellCount*3 {
 		return
 	}
+	if len(t.objects) > cellCount*3 {
+		newObjects := make([]fyne.CanvasObject, cellCount*3)
+		copy(newObjects, t.objects)
+		t.objects = newObjects
```
Jacalz, line 766: Am I reading this right that the code is trying to empty out the objects slice and making it shorter? You can likely avoid the allocation by looping over and clearing the fields (a call to `clear()` when we have a newer Go as base) and then slicing off the remainder.

redawl: Yeah, basically we don't want to keep the objects around that are no longer inside the viewport if the textgrid is resized.

Hmm last I read this:
https://100go.co/#slices-and-memory-leaks-26

I thought it said the recommended way to reduce the size of a backing array for a slice was to do a copy, but I guess we could also set the trailing objects to `nil` before slicing, which would be better. Thanks for the suggestion :)

Jacalz: No problem. I guess it depends on if we deem that we might want to reuse the remaining space again in the future or not (as long as we nil out all the objects). Your call :)

andydotxyz:
Just scanning an eye over this and the test failures seem to be valid - it's reporting that the rectangles of the cell are no longer the right size / position I think?

The files will of course change as un-needed items won't be rendered any more, but I wonder if it's seeing that a layout is not right?

redawl:
> Just scanning an eye over this and the test failures seem to be valid - it's reporting that the rectangles of the cell are no longer the right size / position I think?
> 
> The files will of course change as un-needed items won't be rendered any more, but I wonder if it's seeing that a layout is not right?

Yeah, this isn't ready to be merged, plenty of work still to do.

I think the test failures are related to the changes in when refreshes occur, so size calculations aren't happening before the assertions run anymore. 

I need to take another look at this, it's still blocking my work on one of my projects, thanks for the reminder :)

PR #5854, opened 2025-07-10
Take into account starting number when rendering ordered list

### LINKED ISSUE #5803: RichTextFromMarkDown not correctly setting first numbered list index
### Checklist

### Describe the bug

Numbered lists in markdown should increase in numerical order, even if there is content in between. widget.NewRichTextFromMarkdown does not properly handle this case, starting numbering over if there is content that breaks up the list.

### How to reproduce

1. Run the example code
2. Observe that numbering starts over at 1. for the second list item.

### Screenshots

Fyne `NewRichTextFromMarkdown`:
Image

Stackedit.io with correct implementation (Editor on the left, display on the right):

Image

### Example code

```go
package main

import (
        "fyne.io/fyne/v2/app"
        "fyne.io/fyne/v2/widget"
)

func main() {
        a := app.New()

        w := a.NewWindow("Test Window")

        richText := widget.NewRichTextFromMarkdown("# Title\n1. Item 1\n\nContent under item 1\n\n2. Item 2\n")

        w.SetContent(richText)

        w.ShowAndRun()
}
```

### Fyne version

2.6.1

### Go compiler version

1.24.4

Gentoo

### Additional Information

_No response_

redawl:
### Description:

Fixes #5803

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:


widget/richtext_objects.go
```diff
@@ -232,10 +237,13 @@ func (l *ListSegment) Inline() bool {
 // Segments returns the segments required to draw bullets before each item
 func (l *ListSegment) Segments() []RichTextSegment {
 	out := make([]RichTextSegment, len(l.Items))
+	if l.Start == 0 {
+		l.Start = 1
+	}
```
redawl: TBH I'm not sure about this approach. Maybe its better to set a local variable here instead of mutating the struct state, i.e:
```suggestion
    start := l.Start
	if start == 0 {
		start = 1
	}
```
LMK if that's better

andydotxyz: Yes I think local state would be better - This is (sort of) rendering the state and it feels wrong to mutate at that point.

Thanks for fixing this though :)

redawl: Done, thanks https://github.com/fyne-io/fyne/pull/5854/commits/d4b62d58c7ae56dfb7ece013b712e08c2c80106a

I also figured we should also round all `Start`s that are less than 1 up to 1 while we're at it.

andydotxyz:
I just realised that this solution may not be sufficient... the user could have specified 0 as their initial index.

Therefore I wonder, do we have to check if it was specified and/or make it an optional field somehow?

We need to ensure it fulfils the markdown spec but also that people using the Segments API directly don't get any surprises.

I think the index tests need to include raw api usage but also the 0/1/2 start cases.

redawl:
Ok, so I took a slightly different tactic to allow the zero value of the struct to have the correct meaning of an ordered list starting at 1., while still allowing the user to manually set the start to be 0.

The API is less clean now, but I couldn't think of any other way.

widget/richtext_objects.go
```diff
@@ -222,6 +222,23 @@ func (i *ImageSegment) Unselect() {
 type ListSegment struct {
 	Items   []RichTextSegment
 	Ordered bool
+
+	startIndex int
```
andydotxyz, line 232: This might be a misleading name now.
I wonder if `previousIndex` or something like that commenting on its "-1" nature could be a good idea?

redawl: I'm not sure if `previousIndex` conveys the correct meaning either. What about adding a doc comment above `startIndex` with an explanation why it is set to Start - 1?

andydotxyz: Sounds good. As it's unexported we should not spend too much time worrying ;)

widget/richtext_objects.go
```diff
@@ -222,6 +222,29 @@ func (i *ImageSegment) Unselect() {
...
+//
+// Since: 2.7
+func (l *ListSegment) SetStart(s int) {
+	l.startIndex = s - 1
+}
+
+// Start return the starting number for an ordered list.
+//
+// Since: 2.7
+func (l *ListSegment) Start() int {
```
andydotxyz: Sorry I didn't think before - but this reads like a verb not a noun.

Is it needed?
If so maybe Start -> StartIndex? (Possibly the setter too?) or some other naming that is just noun and not a verb?

redawl: What about "StartNumber", since thats what the markdown spec refers to it as:
https://spec.commonmark.org/0.31.2/#start-number

PR #5857, opened 2025-07-11
Single threaded pool buffer

Jacalz:
### Description:

The pool was used in places where concurrency actually was important. I solved this by introducing a new `internal/async/migrated` package which has all of the async data structures but with build tags to become single threaded when migrated. This has the added benefit of avoid issues where we accidentally uses the single threaded version although we need safety.

Benchmark results are promising:

Before:
```
BenchmarkPool/GetOnly-22      	11466109	        90.45 ns/op	       0 B/op	       0 allocs/op
BenchmarkPool/PutOnly-22      	28852821	        38.67 ns/op	      35 B/op	       1 allocs/op
```

After:
```
BenchmarkPool/GetOnly-22      	370900233	         3.386 ns/op	       0 B/op	       0 allocs/op
BenchmarkPool/PutOnly-22      	59720872	        16.87 ns/op	       8 B/op	       1 allocs/op
```

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/async/migrated/pool_migratedfynedo.go
+++ b/internal/async/migrated/pool_migratedfynedo.go
@@ -1,6 +1,6 @@
 //go:build migrated_fynedo
 
-package async
+package migrated
 
 // Pool (when migrated) is just a simple single threaded data buffer.
 type Pool[T any] struct {
--- a/app/settings.go
+++ b/app/settings.go
@@ -7,7 +7,7 @@ import (
 
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/internal/app"
-	"fyne.io/fyne/v2/internal/async"
+	"fyne.io/fyne/v2/internal/async/migrated"
 	"fyne.io/fyne/v2/internal/build"
 	"fyne.io/fyne/v2/theme"
 )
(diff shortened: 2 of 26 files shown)


internal/async/migrated/pool_migratedfynedo.go
```diff
@@ -0,0 +1,32 @@
+//go:build migrated_fynedo
+
+package migrated
+
+// Pool (when migrated) is just a simple single threaded data buffer.
+type Pool[T any] struct {
+	buffer *T
```
dweymouth, line 7: This looks like it can only hold one item at a time? Replicating Go's `sync.Pool` even without the thread safety is a more complicated task than this. Basically we'd want the pool to be able to hold multiple items, and to hook into the Go GC invocation (or Fyne cache clean invocations) and drop references for items that haven't been accessed in awhile when that happens.

Jacalz: Ah, yes. I though we were pooling slices everywhere. However, I think hooking into the GC or setting up a clean interval likely is overkill. We can probably just do something similar to the ring buffer in my queue implementation. Also, with the weak pointers that were added recently, we can likely just use a map of that which cleans itself (when we have bumped to a new enough Go version).

PR #5858, opened 2025-07-11
Use single threaded canvas object queue when migrated

Jacalz:
### Description:

This feels, and is, significantly faster. It is only used for the refresh queue
so when we no longer have migrations, we can likely remove the async
version entirely.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/async/queue_migratedfynedo.go
+++ b/internal/async/queue_migratedfynedo.go
@@ -4,35 +4,49 @@ package async
 
 import "fyne.io/fyne/v2"
 
-// CanvasObjectQueue represents a simple single threaded queue for managing canvas objects.
+// CanvasObjectQueue represents a single-threaded queue for managing canvas objects using a ring buffer.
 type CanvasObjectQueue struct {
-	items []fyne.CanvasObject
+	buffer []fyne.CanvasObject
+	head   int
+	size   int
 }
 
-// NewCanvasObjectQueue returns a queue for caching values.
+// NewCanvasObjectQueue returns a queue for caching values with an initial capacity.
 func NewCanvasObjectQueue() *CanvasObjectQueue {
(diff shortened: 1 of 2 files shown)


internal/async/queue_migratedfynedo.go
```diff
@@ -0,0 +1,38 @@
...
+// Out removes and returns the value at the head of the queue.
+// It returns nil if the queue is empty.
+func (q *CanvasObjectQueue) Out() fyne.CanvasObject {
+	if len(q.items) == 0 {
+		return nil
+	}
+
+	head := q.items[0]
+	q.items[0] = nil
+	q.items = q.items[1:]
```
dweymouth: Definitely a nice idea for an improvement! I think we want an expandable ring buffer implementation, so that we can re-use the array space for objects that have been dequeued. As is, I believe this will re-allocate a new backing slice at a minimum every 32 items enqueued, even if only a few are actually in the queue at a time before being dequeued.

Jacalz: That's a great suggestion. I fixed that up just now :)

internal/async/queue_migratedfynedo.go
```diff
@@ -0,0 +1,59 @@
...
+		return nil
+	}
+
+	first := q.buffer[q.head]
+	q.buffer[q.head] = nil
+	q.head = (q.head + 1) % len(q.buffer)
+	q.size--
+
+	if q.size == 0 && len(q.buffer) > 4*defaultQueueCapacity {
+		q.buffer = make([]fyne.CanvasObject, defaultQueueCapacity)
```
dweymouth, line 50: I don't think it really makes a difference but resetting `q.head = 0` here may be slightly cleaner if you feel like changing it

PR #5860, opened 2025-07-13
WIP: rounded image corners

### LINKED ISSUE #5679: Support rounded corners for images
### Checklist

In "modern" UIs, it is often desirable to round image corners slightly.

Yes, by manipulating `image.Image` in Go, but it would not be GPU accelerated

`canvas.Image` should have a `CornerRadius` property similar to rectangles that allows the corners of an image to be rounded slightly. It should be applied in the GL painter to be GPU accelerated.

Another related concept is rounding images completely to form a circle or oval (depending on the aspect ratio of the image). It would be nice to have an API for this as well, or at least design the corner radius API in a forward-compatible way. Note that these are different concepts since rounded corners typically would have the same radius in pixels or dp regardless of image size, whereas the full rounding to a circle/oval is obviously relative to the size of the image.

dweymouth:
### Description:
Adds CornerRadius to canvas.Image

Fixes #5679 

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


andydotxyz:
Is it possible to make this less jagged? Our other corners use fading/anti-aliasing particularly with the smoother edges PR landing this could stand out in comparison.

dweymouth:
Yeah, I tried, but hadn't gotten anything completely working with my lack of full understanding of how the shaders work and reaching the limits of ChatGPT vibe coding ;)

dweymouth:
I imagine the rounded rectangle shader could be adapted for images, but I didn't figure it out

Vinci10:
if it makes sense, you can apply a similar anti-aliasing function as used for rectangles
Exemplary code:
```
#version 110

uniform sampler2D tex;
uniform float radius;  // in pixels
uniform vec2 size;     // in pixels: size of the rendered image quad

varying vec2 fragTexCoord;
varying float fragAlpha;

void main() {
    float alpha = 1.0;
    if (radius > 0.5) {
        vec2 pos = fragTexCoord * size;
        vec2 halfSize = size * 0.5;
        float dist = length(max(abs(pos - halfSize) - halfSize + radius, 0.0)) - radius;
        alpha = 1.0 - smoothstep(-1.0, 1.0, dist);
    }

    vec4 texColor = texture2D(tex, fragTexCoord);
    texColor.a *= fragAlpha * alpha;
    texColor.rgb *= fragAlpha * alpha;

    if(texColor.a < 0.01)
        discard;
    gl_FragColor = texColor;
}
```

Result:
rounded_image

dweymouth:
@Vinci10 This shader works well but seems to stretch the corner radius as well when showing an image with a non-square aspect ratio (using FillMode = canvas.ImageFillContain on the canvas.Image). I don't think this is what we want; it makes sense for the corner radius to always be circular regardless of the image aspect ratio (IMO).

Vinci10:
@dweymouth I think it depends on the radius value passed to the shader. You can get a circular corner by using a proper radius value, I guess.

andydotxyz:
I'm pretty sure all the Go is good - the issue looks to be on the shader, you are operating on the texture coordinates rather than the output coordinates. Normally they have the same aspect ratio but on Stretch fill that is not true.

Two options would likely be to move the shader calculation to use output not texture coordinates - or the radius could be stretched in Go to counteract the squashing.

andydotxyz:
The last two commits here resolve conflicts and then update to smoother corners that don't squash.

Apologies if I accidentally broke the history - I'm not sure what happened.

If you're happy with the changes can you pick them to a corrected branch @dweymouth, or do I need to do something to tidy this up?
Apologies for the trouble! I think the develop merge didn't get history right for some reason.

PR #5862, opened 2025-07-14
Fix the theme override for padding (and other static values)

### LINKED ISSUE #5019: Theme Override container cannot be used to override layout paddings
### Checklist

### Describe the bug

The standard layouts all use the deprecated `theme.Padding()` to look up the padding value which does not pick up the overrides if the layout is used in a container or internally within a widget in an override container.

### How to reproduce

create a theme override container with a theme that overrides padding, and put a container with a layout inside that theme override container

### Screenshots

_No response_

### Example code

it's trivial from a code audit of the layout package to see where the problem is

### Fyne version

2.5.0

### Go compiler version

n/a

all

### Additional Information

_No response_

andydotxyz:
In theory some areas of code could be simplified - but that relies on there being no miss-use of the thread/fyne.Do.
So I have left all the widget/theme handling as-is for now and we can simplify it down the road once multiple threads are completely not supported any more.

Fixes #5019

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/theme/render.go
+++ b/internal/theme/render.go
@@ -1,0 +1,29 @@
+package theme
+
+import (
+	"fyne.io/fyne/v2"
+)
+
+var themeStack []fyne.Theme
+
+// CurrentlyRenderingWithFallback returns the theme that is currently being used during rendering or layout
+// calculations. If there is no override in effect then the fallback is returned.
+func CurrentlyRenderingWithFallback(f fyne.Theme) fyne.Theme {
+	if len(themeStack) == 0 {
+		return f
+	}
+
+	return themeStack[len(themeStack)-1]
+}
+
+// PushRenderingTheme is used by the ThemeOverride container to stack the current theme during rendering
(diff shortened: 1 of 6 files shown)


internal/theme/render.go
```diff
@@ -0,0 +1,29 @@
...
+// PushRenderingTheme is used by the ThemeOverride container to stack the current theme during rendering
+// and calculations.
+func PushRenderingTheme(th fyne.Theme) {
+	themeStack = append(themeStack, th)
+}
+
+// PopRenderingTheme is used by the ThemeOverride container to remove an overridden theme during rendering
+// and calculations.
+func PopRenderingTheme() {
+	themeStack = themeStack[:len(themeStack)-1]
```
dweymouth, line 29:
```suggestion
	themeStack[len(themeStack)-1] = nil
	themeStack = themeStack[:len(themeStack)-1]
```

